PR #1209, opened 2023-06-27
dialects: (affine) use maps in affine.for bounds, and add some helpers

superlopuh:
@Groverkss

### DIFF
--- a/xdsl/ir/affine/affine_map.py
+++ b/xdsl/ir/affine/affine_map.py
@@ -16,7 +16,7 @@ class AffineMap:
     results: list[AffineExpr]
 
     @staticmethod
-    def from_constant(value: int) -> AffineMap:
+    def constant_map(value: int) -> AffineMap:
         return AffineMap(0, 0, [AffineExpr.constant(value)])
 
     @staticmethod
--- a/xdsl/ir/affine/affine_map.py
+++ b/xdsl/ir/affine/affine_map.py
@@ -15,6 +15,18 @@ class AffineMap:
     num_symbols: int
     results: list[AffineExpr]
 
+    @staticmethod
+    def from_constant(value: int) -> AffineMap:
+        return AffineMap(0, 0, [AffineExpr.constant(value)])
+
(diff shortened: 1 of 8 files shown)


adutilleul (approved):

compor (approved):
LGTM

xdsl/ir/affine/affine_map.py
```diff
@@ -15,6 +15,18 @@ class AffineMap:
     num_symbols: int
     results: list[AffineExpr]
 
+    @staticmethod
+    def from_constant(value: int) -> AffineMap:
+        return AffineMap(0, 0, [AffineExpr.constant(value)])
```
Groverkss, line 20: Can we call this constant_map?

xdsl/ir/affine/affine_map.py
```diff
@@ -15,6 +15,18 @@ class AffineMap:
...
+
+    @staticmethod
+    def identity(rank: int) -> AffineMap:
+        return AffineMap(rank, 0, [AffineExpr.dimension(dim) for dim in range(rank)])
+
+    @staticmethod
+    def empty() -> AffineMap:
+        return AffineMap(0, 0, [])
+
```
Groverkss, line 29: Can you make the empty method with dims = 0, syms = 0?

xdsl/ir/affine/affine_map.py
superlopuh: How about just `constant`?

Groverkss: I prefer `constant_map`. It's also what mlir uses.

xdsl/ir/affine/affine_map.py
superlopuh: you mean explicitly specifying the arguments?

Groverkss: No, I meant the api. But it's fine, I can extend this later.

Groverkss (approved):

PR #1210, opened 2023-06-27
[core] Introduce LocationAttr (unknown only)

Moxinilian:
This commit introduces basic support for location attributes. This only
supports unknown locations for now.

Printers and parsers now support locations on arbitrary operations and
block arguments, along with locations as attributes.

### DIFF
--- a/tests/filecheck/dialects/builtin/parse_with_location.mlir
+++ b/tests/filecheck/dialects/builtin/parse_with_location.mlir
@@ -1,12 +1,10 @@
 // RUN: xdsl-opt %s --print-op-generic | filecheck %s
 // CHECK: module
 "builtin.module"() ({
-  "func.func"() ({
-    // CHECK: ^{{.*}}(%{{.*}}: i32):
-    ^bb0(%arg: i32 loc(unknown)):
-      // CHECK: "test.op"
-      %0 = "test.op"() : () -> !test.type<"int"> loc(unknown)
-      "test.op"() : () -> () loc(unknown)
-      "func.return"() : () -> () loc(unknown)
-  }) {"function_type" = (i32) -> (), "sym_name" = "builtin"} : () -> ()
+  // CHECK: ^{{.*}}(%{{.*}}: i32):
+  ^bb0(%arg: i32 loc(unknown)):
+    // CHECK: "test.op"() : () -> !test.type<"int">
+    %0 = "test.op"() : () -> !test.type<"int"> loc(unknown)
+    // CHECK: "test.op"() : () -> ()
+    "test.op"() : () -> () loc(unknown)
 }) : () -> ()
--- a/tests/filecheck/dialects/builtin/parse_with_location.mlir
+++ b/tests/filecheck/dialects/builtin/parse_with_location.mlir
@@ -1,0 +1,11 @@
+// RUN: xdsl-opt %s --print-op-generic | filecheck %s
+// CHECK: module
+"builtin.module"() ({
+  "func.func"() ({
+    // CHECK: ^{{.*}}(%{{.*}}: i32):
+    ^bb0(%arg: i32 loc(unknown)):
+      // CHECK: "arith.constant"
+      %x1 = "arith.constant"() {"value" = 0 : i64} : () -> i64 loc(unknown)
+      "func.return"() : () -> () loc(unknown)
+  }) {"function_type" = (i32) -> (), "sym_name" = "builtin"} : () -> ()
+}) : () -> ()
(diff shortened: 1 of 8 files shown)


Moxinilian:
Codecov this was a draft PR...

xdsl/dialects/builtin.py
```diff
@@ -292,6 +292,16 @@ class UnitAttr(ParametrizedAttribute):
     name = "unit"
 
 
+@irdl_attr_definition
+class LocationAttr(ParametrizedAttribute):
```
math-fehr, line 296: Can you add it to the `Builtin` dialect at the bottom of the file?

xdsl/xdsl_opt_main.py
```diff
@@ -329,7 +329,11 @@ def register_all_targets(self):
 
         def _output_mlir(prog: ModuleOp, output: IO[str]):
             printer = Printer(
-                stream=output, print_generic_format=self.args.print_op_generic
+                stream=output,
+                print_generic_format=self.args.print_op_generic,
+                print_debuginfo=self.args.print_debuginfo
+                if "print_debuginfo" in self.args
+                else False,
```
math-fehr, line 336: Can you add an option for `print_debuginfo` instead? So it appears in `--help` for instance?

math-fehr (approved):
Nice, that looks good!

math-fehr:
Could you also duplicate your test to pass through MLIR? That way we make sure it works well with it!

tests/filecheck/dialects/builtin/parse_with_location.mlir
webmiche, line 1: Could you use the testing dialect here instead? That way the builtin tests won't depend on the func and arith dialect.

tests/test_printer.py
webmiche, line 1: Same here.

tests/filecheck/dialects/builtin/parse_with_location.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: xdsl-opt %s --print-op-generic | filecheck %s
+// CHECK: module
+"builtin.module"() ({
+  "func.func"() ({
+    // CHECK: ^{{.*}}(%{{.*}}: i32):
```
superlopuh, line 5: Am I understanding correctly that this checks only for parsing and not printing? Do we want a test for printing also?

superlopuh (changes requested):
Would love full location support, thank you for starting this

xdsl/xdsl_opt_main.py
Moxinilian: I looked for this but have no idea where this is happening. Any pointer?

tests/filecheck/dialects/builtin/parse_with_location.mlir
Moxinilian: Tests for printing happen in `test_printer.py`. The reason they are done separately is that right now this printing and parsing is only here for compatibility purposes. It does not add any information to the IR.

tests/test_printer.py
Moxinilian: I'm not sure I understand. I copied the test above to make this test.

tests/filecheck/dialects/builtin/parse_with_location.mlir
Moxinilian: What is the testing dialect?

superlopuh: It's in the dialects folder in xdsl

webmiche: The testing dialect is our way of making tests independent. Check for example [the affine tests](https://github.com/xdslproject/xdsl/blob/main/tests/filecheck/dialects/affine/affine_ops.mlir). Here we use the test dialect to create the SSAValues that are the operands of other ops and to get the types. Overall, the dialect is quite small ([here](https://github.com/xdslproject/xdsl/blob/main/xdsl/dialects/test.py)).

tests/test_printer.py
webmiche: This comment was about using the testing dialect. I know, this is not (yet) done in all the tests, but I think it makes sense to do it for new tests.

math-fehr (approved):
:+1:

math-fehr:
@superlopuh, if that's good for you, we can merge it

tests/filecheck/dialects/builtin/parse_with_location.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: xdsl-opt %s --print-op-generic | filecheck %s
+// CHECK: module
+"builtin.module"() ({
+  "func.func"() ({
```
webmiche, line 4: Why not remove the functions as well?

AntonLydike (approved):

superlopuh (approved):
my b

webmiche (approved):
LGTM now 👍

PR #1211, opened 2023-06-27
core: Extract AttrParser out of Parser

math-fehr:
This PR moves the attribute parsing functions out of Parser, and moves them to a new `AttrParser`.
`AttrParser` is now given to attributes and types, so all the unrelated methods are not shown anymore in the IDE.
Also, this cleans up a bit the file as we now have two smaller files instead of a huge one.

There are no changes in the methods moved to `AttrParser`, so no need to review the content of the methods moved.


webmiche (approved):
I'm surprised. Don't we have parser specific tests? Shouldn't those be split as well? o/w LGTM.

superlopuh (approved):

math-fehr:
> I'm surprised. Don't we have parser specific tests? Shouldn't those be split as well? o/w LGTM.

Right now, `Parser` inherits from `AttrParser`, so that's why nothing had to be changed!
We might want to rename `Parser` into `OpParser` and then make both `AttrParser` and `OpParser` inherit from the `BaseParser`. However, that should be done in multiple steps here I believe!

math-fehr:
(I'm waiting for #1210 to be merged to merge this, as these are conflicting)

AntonLydike (approved):
:+1:

PR #1212, opened 2023-06-28
dialects: (riscv) initial arith lowering

adutilleul:
- [X] Lowering scalar types (i.e: not Vector)
- [X] Lowering for ALU-like ops
- [X] Lowering for FPU-like ops
- [X] Tests


superlopuh (approved):
I think this is ready to merge modulo greedy pattern applier

tests/filecheck/backend/riscv/riscv_arith_lowering_rv32.mlir
```diff
@@ -0,0 +1,145 @@
...
+    %lhsi32 = "arith.constant"() {value = 1 : i32} : () -> i32
+    // CHECK: %{{.*}} = "riscv.li"() {"immediate" = 1 : si32} : () -> !riscv.reg<>
+    %rhsi32 = "arith.constant"() {value = 2 : i32} : () -> i32
+    // CHECK-NEXT: %{{.*}} = "riscv.li"() {"immediate" = 2 : si32} : () -> !riscv.reg<>
+    %lhsindex = "arith.constant"() {value = 1 : index} : () -> index
+    // CHECK-NEXT: %{{.*}} = "riscv.li"() {"immediate" = 1 : si32} : () -> !riscv.reg<>
+    %rhsindex = "arith.constant"() {value = 2 : index} : () -> index
+    // CHECK-NEXT: %{{.*}} = "riscv.li"() {"immediate" = 2 : si32} : () -> !riscv.reg<>
+    %lhsf32 = "arith.constant"() {value = 1.000000e+00 : f32} : () -> f32
+    // CHECK-NEXT: %{{.*}} = "riscv.li"() {"immediate" = 1065353216 : si32} : () -> !riscv.reg<>
```
JosseVanDelm, line 12: Sorry if I missed this before, but why are we using signed types here?
Seems like a no-no since LLVM does not use it?

xdsl/backend/riscv/lowering/rv32_arith_lowering.py
```diff
@@ -0,0 +1,436 @@
...
+                xori = riscv.XoriOp(sltu, 1)
+                rewriter.replace_matched_op([sltu, xori])
+            case _:
+                raise NotImplementedError("Cmpi predicate not supported")
+
+
+class LowerArithSelect(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: arith.Select, rewriter: PatternRewriter) -> None:
+        raise NotImplementedError("Select is not supported")
```
JosseVanDelm, line 211: Feels like ideally we should have a python test that tests for this exception and all of the others

xdsl/xdsl_opt_main.py
```diff
@@ -112,6 +113,7 @@ def get_all_passes() -> list[type[ModulePass]]:
         LowerSnitchRuntimePass,
         PrintToPrintf,
         RISCVRegisterAllocation,
+        RISCVLowerArithRV32,
```
JosseVanDelm, line 116: Wondering whether we at some point should integrate the ABI spec thing from @AntonLydike here.
Not in this PR?

xdsl/backend/riscv/lowering/rv32_arith_lowering.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+class LowerArithTruncFOp(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: arith.TruncFOp, rewriter: PatternRewriter) -> None:
+        raise NotImplementedError("TruncF is not supported")
+
+
+class RISCVLowerArithRV32(ModulePass):
+    name = "riscv-lower-arith-rv32"
```
JosseVanDelm, line 388: Is there a canonical way to name conversion passes? I would have named this `lower-arith-to-rv32` or `convert-arith-to-rv32`. Not sure if this is right, but this order in naming looks a bit redundant and backwards.

JosseVanDelm (approved):
Hi @adutilleul,

Thanks for your contribution! This is an important step forward, but I do have some questions

I missed the part where we introduced signed integers to the Riscv dialect (or I forgot that I did that? Not sure anymore).
After learning about signless types it seems a bit weird to me that we would be introducing them after arith, since riscv assembly is pretty much signless all the way?

I also don't like the naming of the current pass. 

Also, this PR introduces a lot of exceptions that don't seem to be tested.
Can you add some negative tests to increase coverage please?

Thanks!

JosseVanDelm (changes requested):
Forgot to press "request changes" on my last review!

tests/filecheck/backend/riscv/riscv_arith_lowering_rv32.mlir
adutilleul: That's because the `LiOp` constructor does the following : 
```python 
immediate = IntegerAttr(immediate, IntegerType(32, Signedness.SIGNED))
```

xdsl/backend/riscv/lowering/rv32_arith_lowering.py
adutilleul: I'm not sure how useful it is right now since the coverage of these ops might be improved very soon.

xdsl/xdsl_opt_main.py
adutilleul: Some part of the lowering is really specific and assume a 32 bits target. I just wanted to make it clear.

xdsl/backend/riscv/lowering/rv32_arith_lowering.py
superlopuh: That's a good point. I think `arith-to-riscv` or `lower-arith-to-riscv` are good. not sure we want to specify `rv32`, we haven't so far, might want to stay consistent and rename things later

xdsl/backend/riscv/lowering/rv32_arith_lowering.py
JosseVanDelm: Good point!

tests/filecheck/backend/riscv/riscv_arith_lowering_rv32.mlir
JosseVanDelm: I will make an issue for this!

JosseVanDelm: #1216 was made!

xdsl/xdsl_opt_main.py
JosseVanDelm: @superlopuh does your comment about the pass name also apply here?

superlopuh:
I don't think we need the tests for coverage, seems like an inversion of the purpose of coverage. It seems that the implemented lowerings are tested in the filecheck, so I think we're good. About the name, I don't think we're that careful about non platform-specific lowerings elsewhere, I don't think it's worth worrying about that too much for the time being. At some point we'll come up with actual infrastructure for targets and stick to it, but I'd rather just assume that riscemu is the only riscv that exists until that becomes a problem.

compor (approved):

adutilleul:
I've changed the naming @JosseVanDelm, so I think we're good for merging ?

JosseVanDelm (approved):

JosseVanDelm:
@adutilleul sure! go ahead! sorry for the slow response!

PR #1213, opened 2023-06-29
Resture RISCV Dialect (WIP)

kingiler:
This PR splits the RISCV dialect into multiple files by the extension.

The current approach tries to minimise the impact on other codes depending on the dialect.

TODO:
- Make a generic Register type, IntegerRegister and FloatRegister will inherite from it.
- Make generic Instruction types, such as R-Type and I-Type. Extensions may narrow it down to specific register type.
- Make the verification work. Currently there is a problem where AddOp will take in FloatRegister and does not raise verfication error.


compor:
In general, I like this, but this is a massive PR and it also seems to do more things than it "advertises" like the inheritance in the RegisterType.

I'd suggest moving this to a draft PR and once things in the TODO are closer to ready you can break off self-contained smaller PRs.

I'm also not 100% certain we need separate dialects for each extension ATM, but that can be further discussed.

AntonLydike:
Can we close this PR?

PR #1217, opened 2023-06-30
interpreter: (cf) add cf interpreter functions and extend arith

superlopuh:


tests/interpreters/test_cf_interpreter.py
```diff
@@ -0,0 +1,69 @@
...
+from xdsl.dialects import arith, cf, func
+from xdsl.dialects.builtin import ModuleOp, i32
+from xdsl.interpreter import Interpreter
+from xdsl.interpreters.arith import ArithFunctions
+from xdsl.interpreters.cf import CfFunctions
+from xdsl.interpreters.func import FuncFunctions
+from xdsl.ir.core import Block, Region
+
+
+def triangle_fn(n: int) -> int:
```
webmiche, line 11: Where does this name vome from? Isn't this just the sum from 1 to n? 😅

tests/interpreters/test_cf_interpreter.py
```diff
@@ -0,0 +1,69 @@
...
+    (result,) = interpreter.call_op("triangle", (n,))
+    return result
+
+
+def test_triangle():
+    assert triangle_fn(0) == triangle_interp(0)
+    assert triangle_fn(1) == 1
+    assert triangle_fn(2) == 3
+    assert triangle_fn(3) == 6
+    assert triangle_fn(4) == 10
```
webmiche, line 69: Why are you mainly testing the python function here? Why not make this a parametric test taking in a range of values and comparing the two results?

xdsl/interpreter.py
```diff
@@ -362,25 +362,6 @@ def run_block(self, block: Block, args: PythonValues) -> PythonValues | None:
         The terminator of this block is expected either to call its successor or return
         the results for the region directly.
         """
```
webmiche, line 364: Is this git diffs messing with me or is this function now just empty? 🤔

xdsl/interpreters/arith.py
webmiche, line 1: I guess we see here one of the limitations of the Pyhton-based interpreter: Modellling fixed-size integers is nasty in Python. That's also why signed an unsigned are not distinguished in Python...

But I guess as long as we stay within reasonable boundaries, this should not matter.

xdsl/interpreters/cf.py
```diff
@@ -0,0 +1,29 @@
...
+class CfFunctions(InterpreterFunctions):
+    @impl_terminator(cf.Branch)
+    def run_br(self, interpreter: Interpreter, op: cf.Branch, args: tuple[Any, ...]):
+        return Successor(op.successor, args), ()
+
+    @impl_terminator(cf.ConditionalBranch)
+    def run_cond_br(
+        self, interpreter: Interpreter, op: cf.ConditionalBranch, args: tuple[Any, ...]
+    ):
+        cond: int = args[0]
```
webmiche, line 23:
```suggestion
        cond: bool = args[0]
```

Why not like this?

compor: condition is `i1` in cf, but I guess in Python-land we can cast it?

compor (approved):

adutilleul (approved):

tests/interpreters/test_cf_interpreter.py
superlopuh: In my head it made sense :) it's the number of elements in a lower triangular matrix of size n. I'll rename to `sum_to`

xdsl/interpreters/cf.py
superlopuh: fun fact :)
```
>>> issubclass(bool, int)
True
```

tests/interpreters/test_cf_interpreter.py
superlopuh: I just forgot to finish writing this bit of code, will do the parametric thing

xdsl/interpreters/cf.py
superlopuh: So it really depends on whether you philosophically think of i1 as bool or int, at runtime it's still going to be an instance of bool there's no actual casting going on

xdsl/interpreter.py
superlopuh: Hmm I'll check the whole function should be deleted

tests/interpreters/test_cf_interpreter.py
webmiche: I see, that's true :sweat_smile:

xdsl/interpreters/cf.py
webmiche: Ah okay, that's fine then.

xdsl/interpreters/arith.py
math-fehr: Yeah, I guess the problem here is that our interpreter should also take into account bitwidth, which it doesn't now!

math-fehr (approved):
Nice!

tests/interpreters/test_cf_interpreter.py
```diff
@@ -0,0 +1,69 @@
...
+from xdsl.interpreters.arith import ArithFunctions
+from xdsl.interpreters.cf import CfFunctions
+from xdsl.interpreters.func import FuncFunctions
+from xdsl.ir.core import Block, Region
+
+
+def triangle_fn(n: int) -> int:
+    result = 0
+    i = 0
+    while i <= n:
```
georgebisbas, line 14: This can be done better?

superlopuh: Is that a question or an assertion? the whole function is `n * (n-1) // 2` so can be done better indeed :)

webmiche (approved):

PR #1219, opened 2023-06-30
misc: Add coverage generation rules to Makefile

compor:
Adds coverage generation and report (html and markdown) rules to Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ VENV_DIR ?= venv
 
 # these targets don't produce files:
 .PHONY: clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
+.PHONY: coverage coverage-report-html coverage-report-md
 
 # remove all caches and the venv
 clean:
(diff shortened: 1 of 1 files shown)


webmiche (approved):

superlopuh (changes requested):
Could you please replicate the behaviour of `ci-mlir.yml` and check that the coverage report is identical to the CI?

compor:
I replicated the coverage behaviour of the CI and also allow optional execution of coverage in filechecks if `mlir-opt` is not in the `PATH` env variable.

math-fehr:
Just something I don't understand, is there a reason why we don't want to try the coverage when `mlir-opt` is not in the PATH? AFAIK, it should still work, right? It will just not output the `mlir-opt` tests, but that should be fine?

compor:
> Just something I don't understand, is there a reason why we don't want to try the coverage when `mlir-opt` is not in the PATH? AFAIK, it should still work, right? It will just not output the `mlir-opt` tests, but that should be fine?

It does work indeed.
I sort of assumed from the prior comments that we don't want coverage there if `mlir-opt` is not available.
I'll amend.

math-fehr (approved):
Looks good to me!

superlopuh (approved):
thank you!

PR #1220, opened 2023-06-30
dialects/riscv: Add liveness block naive register allocation strategy

compor:
This PR adds a liveness use-def register allocation strategy that works on blocks.
It traverses the use-def SSA chain backwards (i.e., from uses to defs) and 
1. allocates registers for operands
2. frees registers for results (since that will be the last time they appear when going backwards)

```
for op in block.walk_reverse():
  for operand in op.operands:
     if operand is not allocated:
          allocate(operand)
  
for result in op.results:
  if result is not allocated:
      allocate(result)
      free_before_next_instruction.append(result)
  else:
      free(result)
```


xdsl/backend/riscv/register_allocation.py
```diff
@@ -24,6 +25,95 @@ def allocate_registers(self, module: ModuleOp) -> None:
...
+        )
+
+    def _allocate(self, reg: SSAValue) -> bool:
+        if isinstance(reg.typ, RegisterType | FloatRegisterType) and (
+            available_regs := self.register_sets.get(type(reg.typ))
+        ):
+            if reg.typ.data.name is None:
+                # if we run out of real registers, allocate a j register
+                reg_type = type(reg.typ)
+                if not available_regs:
```
superlopuh, line 63: Isn't this unreachable? the truthiness check is on line 58

xdsl/backend/riscv/register_allocation.py
```diff
@@ -24,6 +25,95 @@ def allocate_registers(self, module: ModuleOp) -> None:
         raise NotImplementedError()
 
 
+class RegisterAllocatorLivenessBlockNaive(RegisterAllocator):
```
superlopuh, line 28: Can you please document this class with a detailed but high-level description of what its doing? The code is a bit hard to follow, would be nice to have a strategy outline

compor: I added a docstring, but this is a textbook algorithm and rather simplified.

xdsl/backend/riscv/register_allocation.py
compor: Good spot, I simplified this. It needs a test where it runs out of registers

superlopuh (approved):

adutilleul (approved):
LGTM, I feel that more code could be shared amongst all strategies, like for instance having a proper abstraction for a `RegisterSet`;  an helper like `walk_blocks` would also make the code easier to read but this should be done in other PRs.

PR #1223, opened 2023-07-03
documentation: (toy) toy cli interprets by default

superlopuh:


docs/Toy/toy/compiler.py
```diff
@@ -20,3 +24,25 @@ def parse_toy(program: str, ctx: MLContext | None = None) -> ModuleOp:
...
+
+    if target == "toy-inline":
+        return
+
+    ShapeInferencePass().apply(ctx, module_op)
+
+    if target == "toy-infer-shapes":
+        return
+
+    raise ValueError(f"Unknown target option {target}")
```
webmiche, line 48: Why are you doing all the transforms just to fail later on? Wouldn't it make more sense to just fail early?

superlopuh: Mostly because modifying the strings in three places is tedious. `argparse` will already have checked for the options before calling this function, so it's only if for some reason someone calls it from python code somehow, and to make Pyright happy since it doesn't know that there is a closed set of options possible.

PR #1224, opened 2023-07-03
core: Add composition to AffineMap and AffineExpr

Groverkss:


tests/test_affine_builtins.py
```diff
@@ -57,3 +57,39 @@ def test_quasiaffine_map():
...
+
+
+def test_composition_simple():
+    # map1 = (x, y) -> (x - y)
+    map1 = AffineMap(2, 0, [AffineExpr.dimension(0) - AffineExpr.dimension(1)])
+    # map2 = (x, y) -> (y, x)
+    map2 = AffineMap(2, 0, [AffineExpr.dimension(1), AffineExpr.dimension(0)])
+    # Compose
+    # map3 = (x, y) -> (y - x)
+    map3 = map1.compose(map2)
```
superlopuh, line 69: Do maps have equality? Feels like we should be checking that the expression is what we expect. Maybe we'd want to write "integration" tests with some values in addition to the unit test that checks for structural equality.

superlopuh (changes requested):
can you take a look at the coverage for the code and make sure that we cover all the interesting branches? Not sure why coverage is broken online but it's fairly easy to see in vscode also

tests/test_affine_builtins.py
Groverkss: That is not easy. It's like you can have a single arithmetic expression represented as multiple ASTs. Getting a canonical form for them is non-trivial and is not possible to check equality without a solver. Checking using evaluation is the best way.

Groverkss:
> can you take a look at the coverage for the code and make sure that we cover all the interesting branches? Not sure why coverage is broken online but it's fairly easy to see in vscode also

I don't know how to do this, but the tests of this patch mostly cover all of the code in this patch. I'm not sure what you are referring to. Could you be more specific?

webmiche (approved):
LGTM. I must admit though I am a bit out of the loop on the polyhedral stuff. It looks like good, reasonable Python code, but you are the expert on what it actually does/should do :wink:

tests/test_affine_builtins.py
superlopuh: Yes but we know what the AST should be and it would be nice to catch when it changes. I find it easier to reason about than evaluation testing, because you don't have to figure out what the representative set of inputs and outputs is, it's just a simple ADT comparison.

superlopuh:
Alright, disregard my codecov comment :) I'm mostly just surprised that the action didn't run on this PR

superlopuh:
Ah I see why the codecov didn't run, it's because this is a PR from a fork

superlopuh (approved):

math-fehr:
Once this is updated we can merge it!

PR #1229, opened 2023-07-04
interpreter: (ShapedArray) add some methods to ShapedArray

superlopuh:


xdsl/interpreters/shaped_array.py
```diff
@@ -17,34 +21,64 @@ class ShapedArray(Generic[_T]):
...
         """
-        self.data[self.offset(indices)] = value
+        self.data[self.offset(index)] = value
+
+    def indices(self) -> Iterable[tuple[int, ...]]:
+        """
+        Iterates over the indices of this shaped array.
+        """
+        for indices in product(*(range(dim) for dim in self.shape)):
+            yield tuple(indices)
```
AntonLydike, line 65: `for/yield` can be folded to a `yield from`:

```suggestion
        yield from product(*(range(dim) for dim in self.shape))
```

AntonLydike (approved):

xdsl/interpreters/shaped_array.py
superlopuh: Good spot I thouhgt it wasn't a tuple for some reason

PR #1236, opened 2023-07-05
dialects: (llvm) add constant, return, call intrinsic

superlopuh:


AntonLydike (approved):
Thanks for the renaming of the LLVM operations! That was needed!

xdsl/dialects/llvm.py
```diff
@@ -673,31 +677,74 @@ def get(ptr_type: LLVMPointerType | None = None):
...
+class ExtractValueOp(IRDLOperation):
+    """
+    https://mlir.llvm.org/docs/Dialects/LLVM/#llvmextractvalue-mlirllvmextractvalueop
+    """
+
     name = "llvm.extractvalue"
 
     position: DenseArrayBase = attr_def(DenseArrayBase)
-    container: Operand = operand_def(AnyAttr())
+    container: Operand = operand_def(Attribute)
```
webmiche, line 688:
```suggestion
    container: Operand = operand_def()
```

I am getting confused again with our syntax, shouldn't this work as well?

webmiche (approved):

xdsl/dialects/llvm.py
superlopuh: All three are equivalent, you're right, as well as `container = operand_def()`

superlopuh: I don't really have a strong preference for either quite yet, maybe one will form over time

PR #1238, opened 2023-07-05
core: allow insertion anywhere in the block containing matched operation

superlopuh:
This sort of thing is done in the Toy dialect, and I think @PapyChacal has had a similar need for allocation/deallocation with the tensor work. @math-fehr mentioned that the walker might need to be changed somehow, but I'm not sure how exactly, the changes in this PR seem to work


adutilleul:
One of the issue I encountered applying a similar fix is that the walker will continue matching deleted operations of the same kind.
I don't know how much an issue this is, because my use-case was basically rewriting something similar to the pass `reconcile-unrealized-casts`.

Overall, aside from this specific case I believe this temporary fix looks good to me, but I'm interested in hearing others' opinions.

webmiche (commented):
Personally, I like to think of these things in a very conceptual manner, trying to understand advantages/disadvantages.

AFAIK, the old system tried to restrict the scope as much as possible to enforce that people write peephole rewrites. This has several advantages, in particular being able to reason very locally and being able to executed rewrites in parallel (though that has not been implemented (yet)). If someone wanted to change the block, they had to match the parent op (though that is not quite true as we already have [inline_block_{before,after}](https://github.com/xdslproject/xdsl/blob/main/xdsl/rewriter.py#L106).). But this was a very local and specific kind of rewrite that would not make us lose all advantages of peephole rewrites and we felt was still relatively easy to reason about.

This new thing allows to add/change ops much more freely and in many more ways, which might make it hard to reason about the possibilities. 

So, overall, I think we should think about what we want to do and where we want to get. To me, the pattern rewriter is supposed to be a peephole rewriter and your suggestion changes that.

Still, I understand that this certainly is a kind of rewrite we do want to support in our infrastructure in a better way than "just match the parent, bro". So, I feel we might want to separate things a bit in the rewriter, specifying a part that really is a peephole rewriter with all its advantages and another part that is used for this "the peephole actually is the entire block"-rewrite style, maybe even another one where "the peephole is the entire region".

That being said, I am also fine with upstreaming this patch for the time being and opening an issue/discussion on this topic. In particular, considering that we really aren't leveraging any advantages of peephole rewrites currently and a lot of what this PR adds as functionality is already possible, but with "worse syntax" (say you define a new block in a region of your op and inline that before it or whatever).

AntonLydike (approved):

compor:
I'm OK with this, although not so familiar with the rewriter.
Opening an issue might be a good idea.
I'm not sure if there is any other workaround for this.

That's where I stopped trying to add labels from cf to rsicv.

math-fehr:
While having peephole rewrites is nice, sometimes it makes things too annoying, so that's why I think this is the right direction.

Note that currently, you will have weird issues when doing a rewrite using this feature. If you add a new operation at the start of the block for instance, it will not be matched again by the rewriter, as the rewriter is walking in order.
We should change this at some point to have a work-list in our rewriter, so that we can also recurse on operations that were just created.

math-fehr (commented):
That looks good to me!
But I have a question. Is there a reason why you still restrict the rewrite to happen on the block? We should probably remove that restriction as well, right?

superlopuh:
I was under the impression that this was getting the rewriter closer in behaviour to the MLIR one. The MLIR rewriter definitely allows for rewrites within the same block, but I don't know if it's more permissive. IMO one good way to get the best of both worlds would be with subclassing, i.e. make PatternRewriter fully permissive, add a subclass called PeepholePatternRewriter with the old checks, and somehow make that explicit on the RewritePattern level also. These sorts of changes start making much more sense once we actually have ways to reason about rewrites, which I'd like to make much easier over the summer, so maybe it would be ok to be a little more permissive in the meantime?

adutilleul:
I've tested it and now I'm pretty sure that the MLIR allows rewriting across blocks, so I would also agree to allow it.

PR #1239, opened 2023-07-06
dialects: (affine) freeze AffineMap

superlopuh:
This is kind of an RFC. It seems that most of the time we use affine maps as an immutable data structure, thought I'd leverage some python tools to enforce it. @Groverkss WDYT?

### DIFF
--- a/tests/dialects/test_affine.py
+++ b/tests/dialects/test_affine.py
@@ -8,8 +8,8 @@ from xdsl.ir.affine.affine_expr import AffineExpr
 
 def test_simple_for():
     f = For.from_region([], [], 0, 5, Region())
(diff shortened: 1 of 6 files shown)


math-fehr (approved):
:+1:, making attributes pure is what we should aim for.
But I'll wait for @Groverkss as he may have needs to keep this mutable?

compor (approved):

Groverkss (changes requested):
Blocking until we have a mature implementation of AffineMap.

webmiche:
I guess we can close this PR for the time being, then. What do you think @superlopuh?

superlopuh:
Chatted with Kunwar on Zulip, looks like he has upcoming changes, I'll rebase on those when they're in

Issue #1174, opened 2023-06-21
API: rename `typ` properties to `type`


superlopuh:

webmiche:
Why do we want to do this? `type` is a keyword right? Feels better to stay with `typ`, IMO :thinking:

superlopuh:
I started on this and came to a similar conclusion... It's not a keyword though, it's a type/function, and it's perfectly fine to shadow it, we do so in a few places. I'll close this issue though

PR #1240, opened 2023-07-06
core: Rename ```typ``` properties to ```type```

kingiler:
Address #1174


docs/Toy/toy/dialects/toy.py
```diff
@@ -3,58 +3,47 @@
...
-    Block,
-    Region,
-    OpResult,
-)
-
 from xdsl.dialects.builtin import (
+    DenseIntOrFPElementsAttr,
     Float64Type,
     FunctionType,
+    StringAttr,
```
compor, line 14: Can you please adjust your IDE/linters/etc to avoid reordering imports?
It adds a lot of noise that is unrelated to the changes in this PR

AntonLydike:
I'm curious why this PR also re-orders all imports, are you running a custom formatting script?

It would help immensely with reviewing if we only had relevant changes in the diff!

math-fehr:
I'll look at it once the imports changes are removed, but otherwise it looks good to me!

kingiler:
The import changes issue is addressed in #1243. Now the only changes should the the apis.

docs/Toy/toy/dialects/toy.py
kingiler: Addressed in #1243. Rebased to main.

tests/dialects/test_mpi_lowering.py
```diff
@@ -32,8 +32,8 @@ def check_emitted_function_signature(
         # check that the argument type is correct (if constraint present)
         if typ is not None:
```
adutilleul, line 33: This should probably be renamed `type` also.

xdsl/dialects/arith.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, value: AnyIntegerAttr | FloatAttr[AnyFloat]):
         if isinstance(value, FloatAttr):
             typ = value.type
         else:
-            typ = value.typ
+            typ = value.type
```
adutilleul, line 65: The variable should also be named `type` here.

xdsl/dialects/func.py
```diff
@@ -378,7 +380,7 @@ def parse(cls: type[Return], parser: Parser) -> Return:
                 parser.raise_error("Expected the same number of types and arguments!")
             for arg, typ in zip(args, types):
```
adutilleul, line 381: Also `type` here.

xdsl/dialects/memref.py
```diff
@@ -177,21 +177,21 @@ class Load(IRDLOperation):
...
 
-        if self.memref.typ.get_num_dims() != len(self.indices):
+        if self.memref.type.get_num_dims() != len(self.indices):
             raise Exception("expected an index for each dimension")
 
     @staticmethod
     def get(ref: SSAValue | Operation, indices: Sequence[SSAValue | Operation]) -> Load:
         ssa_value = SSAValue.get(ref)
-        typ = ssa_value.typ
+        typ = ssa_value.type
```
adutilleul, line 194: Another `type` issue here.

xdsl/dialects/printf.py
```diff
@@ -68,7 +68,7 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
             args.append(arg := parser.parse_operand())
             parser.parse_characters(":", " - all arguments must have a type")
             typ = parser.parse_type()
```
adutilleul, line 70: Yet another `type` issue here.

xdsl/dialects/scf.py
```diff
@@ -233,7 +233,7 @@ def verify_(self) -> None:
         # operand type)
         for reduction in range(num_reductions):
             typ = self.get_arg_type_of_nth_reduction_op(reduction)
```
adutilleul, line 235: Yet another `type` issue here. (x2)

xdsl/dialects/scf.py
```diff
@@ -198,7 +198,7 @@ def verify_(self) -> None:
         num_reductions = self.count_number_reduction_ops()
         for reduction in range(num_reductions):
             typ = self.get_arg_type_of_nth_reduction_op(reduction)
```
adutilleul, line 200: Yet another `type` issue here. (x1)

xdsl/printer.py
```diff
@@ -675,20 +675,22 @@ def print_op_with_default_format(self, op: Operation) -> None:
...
         # Print the operation type
         self.print(" : (")
-        self.print_list(op.operands, lambda operand: self.print_attribute(operand.typ))
+        self.print_list(op.operands, lambda operand: self.print_attribute(operand.type))
         self.print(") -> ")
         if len(op.results) == 0:
             self.print("()")
         elif len(op.results) == 1:
-            typ = op.results[0].typ
+            typ = op.results[0].type
```
adutilleul, line 683: Yet again another `type` issue.

tests/dialects/test_llvm.py
```diff
@@ -86,7 +86,7 @@ def test_llvm_getelementptr_op_invalid_construction():
 def test_llvm_getelementptr_op():
     size = arith.Constant.from_int_and_width(1, 32)
     ptr = llvm.AllocaOp.get(size, builtin.i32)
-    ptr_typ = llvm.LLVMPointerType.typed(ptr.res.typ)
+    ptr_typ = llvm.LLVMPointerType.typed(ptr.res.type)
```
adutilleul, line 89: should be named `ptr_type` here

xdsl/dialects/gpu.py
```diff
@@ -150,8 +150,8 @@ class AllocOp(IRDLOperation):
 
     def verify_(self) -> None:
         ndyn = len(self.dynamicSizes)
-        assert isinstance(self.result.typ, memref.MemRefType)
-        typ: memref.MemRefType[Attribute] = self.result.typ
+        assert isinstance(self.result.type, memref.MemRefType)
+        typ: memref.MemRefType[Attribute] = self.result.type
         ndyn_typ = len([i for i in typ.shape.data if i.value.data == -1])
```
adutilleul, line 155: should be named `ndyn_type` here

docs/Toy/toy/dialects/toy.py
```diff
@@ -230,7 +230,7 @@ def verify_(self):
             raise VerifyException("Expected last op of FuncOp to be a ReturnOp")
 
         operand = last_op.input
-        operand_typ = None if operand is None else operand.typ
+        operand_typ = None if operand is None else operand.type
 
         return_typs = self.function_type.outputs.data
```
adutilleul, line 235: should be named `return_types` here

webmiche:
Let me repeat my recent comment on the issue here: 
> Why do we want to do this? `type` is a keyword right? Feels better to stay with `typ`, IMO :thinking:

adutilleul (changes requested):
If we want to be consistent, you should also rename all the stray variable names containing `typ` or `typs`, not sure I catched all of them in my review but you should take a look at this.

adutilleul:
> Let me repeat my recent comment on the issue here:
> 
> > Why do we want to do this? `type` is a keyword right? Feels better to stay with `typ`, IMO thinking

`type` is a built-in function, so you could prefix `__builtin__`  (__builtin__.type) to still use it but that might be confusing indeed

math-fehr:
> Let me repeat my recent comment on the issue here:
> 
> > Why do we want to do this? `type` is a keyword right? Feels better to stay with `typ`, IMO 🤔

`type` is not a keyword, right? It's a builtin function though. I think it should be safe to have `type` as a field, as it anyway needs `self.type` to access it.

I personally think that we should use `type` rather than `typ` for fields, which is the case here. What do you think?

webmiche:
> > Let me repeat my recent comment on the issue here:
> > > Why do we want to do this? `type` is a keyword right? Feels better to stay with `typ`, IMO thinking
> 
> `type` is not a keyword, right? It's a builtin function though. I think it should be safe to have `type` as a field, as it anyway needs `self.type` to access it.
> 
> I personally think that we should use `type` rather than `typ` for fields, which is the case here. What do you think?

Okay, sure. I think my point still stands though. I feel that we will run into cases where this will get annoying, so I would try to circumvent it and do `typ` from the get go. But I don't have a strong opinion on it. If you feel `type` is the way, I am happy to take `type`.

kingiler:
I am thinking about the scope of this PR. I think having a class field named ```type``` is fine. But for the others variables named ```typ```, should they put into next PR?

math-fehr:
> I am thinking about the scope of this PR. I think having a class field named `type` is fine. But for the others variables named `typ`, should they put into next PR?

I think that's exacly @michel-steuwer point here. I believe class fields should be named `type`, as they don't shadow anything. However, variables named `typ` might be a bit dangerous, and that's where I believe Michel wants to avoid them. I'm fine with both for the variables, as anyway we have pyright telling us that we are doing something wrong.

superlopuh (approved):
I closed the issue before seeing this PR, and I actually think that the diff is a strict improvement. I don't feel very strongly about this, but I think we should merge this PR.

adutilleul (approved):

Issue #1128, opened 2023-06-13
Safe erase is not honoured by rewiter.erase_op


1rre:
See here - `safe_erase` is not passed to `self.erase_matched_op()` on line 157

https://github.com/xdslproject/xdsl/blob/4a62f6cb1a3ea953c32484615bfba175019bd52a/xdsl/pattern_rewriter.py#L149-L163

This can lead to issues eg. by making the following impossible to erase:
`%4 = "flow2.binop"(%3, %4) {"uid" = "sum", "op" = "+"} : (!flow2.node, !flow2.node) -> !flow2.node`

kingiler:
@1rre would you mind providing a minimal working example showing the issue. It is quite hard to see how the issue is triggered.

PR #1241, opened 2023-07-06
core: Honour safe erase in rewriter.earse_op

kingiler:
Address #1128


math-fehr (approved):
Yes perfect, can you write a test that was failing before but working now?

kingiler:
I am asking the one who create the issue for a MWE. Or anyone can come up with a MWE as I personally do not come into this problem before.

math-fehr:
In a program like:
```
%0 = "test.op"() : () -> i32
%1 = "test.op"(%0) : (i32) -> ()
```
Removing the first operation with "safe_erase=False" should fail on the previous commit, but succeed after this one.

kingiler:
@math-fehr  @AntonLydike , test case added, should be okay?

tests/test_rewriter.py
```diff
@@ -441,3 +442,34 @@ def transformation(module: ModuleOp, rewriter: Rewriter) -> None:
...
+
+    def transformation_unsafe(module: ModuleOp, rewriter: Rewriter) -> None:
+        constant_op = module.ops.first
+        assert constant_op is not None
+        rewriter.erase_op(constant_op, safe_erase=False)
+
+    with pytest.raises(Exception):
+        rewrite_and_compare(prog, expected, transformation_safe)
+
+    rewrite_and_compare(prog, expected, transformation_unsafe)
```
superlopuh, line 475: This feels slightly off to me, what are the guarantees about the state of things after raising an error? I don't think we make a guarantee that the module will be in a usable state. It might be worth swapping this line with the two above

superlopuh (approved):

tests/test_rewriter.py
kingiler: Reading the source code of ```rewrite_and_compare```, when it is called, the string will be parsed into a newly created module.

kingiler: Anyway, I can swap those lines if you think it is more appropriate.

superlopuh: Ah, I missed that they are strings. In that case it's fine, but maybe worth swapping since it's low overhead and might avoid similar confusion in the future. But I don't feel very strongly about this.

PR #1243, opened 2023-07-06
dev: Add isort to sort import, compatible with black

kingiler:
```isort``` is added to ```.pre-commit-config.yaml```


kingiler:
I'm curious why unable to upload coverage report?

AntonLydike (approved):

AntonLydike:
> I'm curious why unable to upload coverage report for Python 3.10 but successful for Python 3.11?

Sometimes coverage upload fails, but that's "normal". I restarted the workflow, should be green soon!

webmiche:
> I'm curious why unable to upload coverage report for Python 3.10 but successful for Python 3.11?

The 3.11 run does not upload coverage, we only need it for one of the two.

kingiler:
Should ```isort``` add to ```requirements-optional.txt```?

webmiche:
> This adds unnecessary noise IMHO. There is no need to sort imports considering the capabilities of modern IDEs.

I second this opinion. The other thing to mention here is that we used to have an approach that needed a certain import order for some modules to interact with the MLIR-Python bindings (I know this is bad style in Python, but we did not find a better way at the time). If we ever want to revive the Python bindings, isort might turn out to be a footgun.

Also, I do not consider merge conflicts in imports too much of an issue, at least it has never been for me. IMO, there is more pressing, unnecessary merge conflict potential to be addressed before thinking about import order.

compor:
> I think I like it, it makes merge conflicts in imports even less likely than before.

My troubles with merge conflicts were never at the imports, if that were the case alone I'd be very happy ;)

kingiler:
If some order has to be maintained, for monkey patching etc, there are methods telling ```isort``` do not sort particular imports, see [here](https://pycqa.github.io/isort/docs/configuration/action_comments.html). Just like ```# type: ignore``` stuff.

kingiler:
Another common case is when there are duplicated imports.

adutilleul (commented):
I think the trade-off for this dependency is fine, looking at the documentation of `isort`  (https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) however they also suggest adding to `pyproject.toml` some configuration to make it independent (pre-commit, CLI, or editor integration) from where it is called, so probably a good thing to do ?

webmiche:
I guess the obvious con is that this PR itself will probably have more merge conflicts in the imports than what we have had until now. But that is a one-time cost compared to a running one, so I guess we can argue that this is still beneficial.

The other thing is that this enforces people to install stuff and setup their environment in a specific way, that might not be compatible with other options. This does not feel like a worth trade-off for the very few merge conflicts this circumvents.

Overall, I don't have a strong opinion. I don't think the trade off is worth, but I can live with it.

Before we can merge this though, you will need to make it compliant with our [PR format](https://github.com/xdslproject/xdsl/wiki/Tags-for-commit-messages-and-PR-titles).

Issue #1244, opened 2023-07-06
Generation of variadic functions with no arguments broken


gabrielrodcanal:
Hi,
I've noticed that whenever we generate a pure variadic function (no arguments) we get ```"llvm.func"() ({
}) {"sym_name" = "test_function", "function_type" = !llvm.func<void (, ...)>, "CConv" = #llvm.cconv<ccc>, "linkage" = #llvm.linkage<"external">, "visibility_" = 0 : i64} : () -> ()```. Notice how there is a ```,``` before the ellipsis that shouldn't be there. This breaks the translation to LLVM IR in mlir-translate. Variadic functions with no arguments used to be legal in LLVM 7 and I don't know if that has changed, but otherwise this should be supported.

Code for reproduction:
```
test_function = llvm.FuncOp(
    "test_function",
    llvm.LLVMFunctionType(
        [], is_variadic=True
    ),
    linkage=llvm.LinkageAttr("external"),
)
print(test_function)
```

PR #1245, opened 2023-07-06
dialects: Fix LLVM variadic functions

kingiler:
Address #1244

### DIFF
--- a/tests/dialects/test_llvm.py
+++ b/tests/dialects/test_llvm.py
@@ -1,7 +1,9 @@
+from io import StringIO
+
 import pytest
-from conftest import assert_print_op
 
 from xdsl.dialects import arith, builtin, llvm
+from xdsl.printer import Printer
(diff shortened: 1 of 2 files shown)


webmiche:
Does this solve the issue @gabrielrodcanal?

I think we should add an MLIR interoperation test as well. One that uses our API, generates a function and then feeds this into mlir-opt.  I think there is a test that does something similar, but I don't remember where. The frontend certainly runs something with python and then filechecks it. One could just add a roundtrip through mlir-opt there...

gabrielrodcanal:
@webmiche Yes, this solves the issue!

AntonLydike (approved):

math-fehr (commented):
👍

math-fehr (approved):

webmiche (approved):
Yes

> I think we should add an MLIR interoperation test as well

But if everyone else agrees, we can forget about that.

PR #1251, opened 2023-07-07
python: Use f-string

kingiler:
- python: Use f-string instead of .format or %

### DIFF
--- a/.github/parse_pyright_version.py
+++ b/.github/parse_pyright_version.py
@@ -1,6 +1,4 @@
-import os
-
-f = open("requirements-optional.txt", "r")
+f = open("requirements-optional.txt")
 
 line = f.readlines()[-1]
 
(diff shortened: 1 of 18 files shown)


math-fehr (approved):
:+1:

superlopuh:
This broke pyright in main... Can you please avoid merging without the CI running successfully? The CI won't run by default with PRs from forks, so probably easier to use the xDSL repo to fix

compor:
@kingiler can you please refrain from merging with just a single approval?
there is no rush, especially for these kind of changes.

kingiler:
I can't see pyright is broken? Merging won't be successful without tasks successfully checked.

------- Original Message -------

superlopuh:
Yeah I don't know why my VSCode was complaining. Maybe I was on a different version of pyright or pylance. It no longer complains

PR #1252, opened 2023-07-07
python: Use generators or direct syntax

kingiler:


math-fehr (approved):
:+1:

compor (approved):

superlopuh:
How did you find these? Do you have a linter that flags this sort of thing?

kingiler:
@superlopuh I am using ```ruff```, configure some rules that flags this things which may impact Python's performance.

superlopuh:
Adding ruff would be a worthwhile PR on its own IMO, I gave myself a few minutes to experiment with it and didn't get to a stage where I was happy before giving up. If you have the time to open a PR with it I think that would be pretty great.

compor:
AFAIU, `ruff` is still in beta? In addition to the "being fast" selling point, I think it might also help to understand the current level of support and maturity of that linter.
I've added it in my editor's setup, but as @superlopuh hinted, I'm not sure where it surpasses other linters.

superlopuh:
I use many non-beta products that would do well to reach the quality of ruff's beta :)

superlopuh:
I'd still recommend opening a PR with your local ruff config so we can discuss adding it as a dev dependency.

PR #1256, opened 2023-07-09
dialects: (riscv) add custom format for lw and sw instructions

superlopuh:
riscemu currently has a bug in sw, mixing up the rs1 and rs2 operations, but only for `sw rs1 rs2 imm`, not the `sw rs2 imm(rs1)` format. This PR introduces the custom format printing for those two ops, as well as a bit of refactoring in the instruction printing code.


compor (approved):

adutilleul (approved):

superlopuh:
@AntonLydike would appreciate your review!

PR #1257, opened 2023-07-09
core: add MLIROptPass, calling mlir-opt

superlopuh:
This adds a Python class to pipe code through `mlir-opt`. This has been quite useful to me while writing the pass pipeline in Python, to leverage things like cse and lowering through builtin dialects. I'm planning to take mlir-opt out of the Toy tutorial for the sake of having an end-to-end compiler with no dependencies, but I think it will be useful to future projects using xDSL.

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
@@ -1,0 +1,15 @@
+// RUN: xdsl-opt %s -p mlir-opt{arguments='--cse','--mlir-print-op-generic'} | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() ({
+    %0 = "arith.constant"() {"value" = 1 : i32} : () -> i32
+    "func.return"() : () -> ()
+  }) {function_type = () -> (), sym_name = "do_nothing"} : () -> ()
+}) : () -> ()
+
+
+// CHECK:         "builtin.module"() ({
+// CHECK-NEXT:      "func.func"() ({
+// CHECK-NEXT:        "func.return"() : () -> ()
+// CHECK-NEXT:      }) {function_type = () -> (), sym_name = "do_nothing"} : () -> ()
+// CHECK-NEXT:    }) : (index) -> ()
--- a/tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt %s -p mlir-opt{arguments='--cse','--mlir-print-op-generic'} | filecheck %s
+// RUN: xdsl-opt %s -p mlir-opt{arguments='--cse','--mlir-print-op-generic'} --print-op-generic | filecheck %s
 
 "builtin.module"() ({
   "func.func"() ({
@@ -11,5 +11,5 @@
 // CHECK:         "builtin.module"() ({
 // CHECK-NEXT:      "func.func"() ({
 // CHECK-NEXT:        "func.return"() : () -> ()
-// CHECK-NEXT:      }) {function_type = () -> (), sym_name = "do_nothing"} : () -> ()
-// CHECK-NEXT:    }) : (index) -> ()
+// CHECK-NEXT:      }) {"function_type" = () -> (), "sym_name" = "do_nothing"} : () -> ()
+// CHECK-NEXT:    }) : () -> ()
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -92,11 +92,7 @@ class ModulePass(ABC):
 
         # if not all args were removed we raise an error
         if len(spec.args) != 0:
-            arguments_str = ", ".join(f'"{arg}"' for arg in spec.args)
-            fields_str = ", ".join(f'"{field.name}"' for field in fields)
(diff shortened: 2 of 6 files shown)


PapyChacal (approved):
Lovely :heart_eyes:

xdsl/transforms/mlir_opt.py
```diff
@@ -0,0 +1,50 @@
...
+    A pass for calling the `mlir-opt` tool with specified parameters.
+    """
+
+    name = "mlir-opt"
+
+    arguments: list[str] = field(default_factory=list)
+
+    def apply(self, ctx: MLContext, op: ModuleOp) -> None:
+        if not shutil.which("mlir-opt"):
+            raise ValueError("mlir-opt is not available")
```
webmiche, line 26: Does this work platform-independently?

tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: xdsl-opt %s -p mlir-opt{arguments='--cse','--mlir-print-op-generic'} --print-op-generic | filecheck %s
```
webmiche, line 1: Not sure how mlir structures it's optimizations, but this feels more like dce than cse. Maybe change the example (or ignore the comment 😉 )

tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
```diff
@@ -0,0 +1,15 @@
...
+    "func.return"() : () -> ()
+  }) {function_type = () -> (), sym_name = "do_nothing"} : () -> ()
+}) : () -> ()
+
+
+// CHECK:         "builtin.module"() ({
+// CHECK-NEXT:      "func.func"() ({
+// CHECK-NEXT:        "func.return"() : () -> ()
+// CHECK-NEXT:      }) {"function_type" = () -> (), "sym_name" = "do_nothing"} : () -> ()
+// CHECK-NEXT:    }) : () -> ()
```
webmiche, line 15: Why not use `CHECK-NOT`? Feels like it's intended for exactly this. Plus. this will be more stable to printing changes for all of these ops.

xdsl/passes.py
```diff
@@ -92,7 +92,11 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
 
         # if not all args were removed we raise an error
         if len(spec.args) != 0:
-            raise ValueError(f'Unrecognised pass arguments "{", ".join(spec.args)}"')
+            arguments_str = ", ".join(f'"{arg}"' for arg in spec.args)
+            fields_str = ", ".join(f'"{field.name}"' for field in fields)
+            raise ValueError(
+                f"Pass arguments [{arguments_str}] not found in [{fields_str}]"
+            )
```
webmiche, line 99: This feels a bit snuck in here. Can you open a separate PR? Or atleast test it?

webmiche (commented):
I like the general idea of this. However, I wonder about how platform-independent this is. And I feel you should link to the commit hash in the README (or wherever that is nowadays).

xdsl/passes.py
superlopuh: Ah sorry this was a part of my debugging while working on this, I'll split it out.

- [ ] split out

xdsl/transforms/mlir_opt.py
superlopuh: I have no idea :) what platforms should I test this on?

webmiche: Not sure, not even sure we need somethgin else than linux, we aren't weirdos after all.

Maybe documentation is enough? Also, from looking at the docu, shutil should at least work relatively well with several platforms, so....

tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
superlopuh: I like check-next for the whole thing because then if it prints an error or something unexpected then at least we catch it. I'm not sure how robust check-not would be if mlir-opt prints "error"

tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
superlopuh:
```
mlir-opt: Unknown command line argument '--dce'.  Try: 'mlir-opt --help'
mlir-opt: Did you mean '--cse'?
```

xdsl/transforms/mlir_opt.py
```diff
@@ -0,0 +1,51 @@
...
+
+        completed_process = subprocess.run(
+            ["mlir-opt", *self.arguments],
+            input=my_string,
+            stdout=subprocess.PIPE,
+            text=True,
+        )
+
+        # Get the stdout output
+        stdout_output = completed_process.stdout
```
math-fehr, line 43: Can you look at `completed_process.returncode`, and return an exception if it is not 0?
That way, if mlir-opt fails, we will see it.

math-fehr (approved):
Nice! Could you just add a test when `mlir-opt` fails?

xdsl/transforms/mlir_opt.py
superlopuh: - [ ] return code

superlopuh:
- [ ] add test case for fail

PR #1259, opened 2023-07-10
transformations: Add `reconcile-unrealized-casts` pass to remove no-op casts

adutilleul:
During multiple partial conversions when lowering to a target dialect, casts operations (`builtin.unrealized_conversion_cast`) are added to a module to ensure it is still type-checkable at any stage (by MLIR invariant).

Thus, cast operations that became no-op can be removed at the end of a given lowering.
This PR adds such a pass that might be useful in the general case and should soon be used for the `riscv` infrastructure.

### DIFF
--- a/tests/filecheck/transforms/reconcile_unrealized_casts.mlir
+++ b/tests/filecheck/transforms/reconcile_unrealized_casts.mlir
@@ -1,0 +1,93 @@
+// RUN: xdsl-opt %s -p reconcile-unrealized-casts | filecheck %s
+
+builtin.module {
+    // CHECK:  builtin.module {
+
+    func.func @unused_cast(%arg0: i64) -> i64 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i16
+        "func.return"(%arg0) : (i64) -> ()
+    }
+
+    // CHECK-NEXT:  func.func @unused_cast(%arg0 : i64) -> i64 {
+    // CHECK-NEXT:    func.return %arg0 : i64
+    // CHECK-NEXT:  }
+
+    // -----
+
+    func.func @simple_cycle(%arg0: i64) -> i64 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i32
+        %1 = "builtin.unrealized_conversion_cast"(%0) : (i32) -> i64
+        "func.return"(%1) : (i64) -> ()
+    }
+
+    // CHECK-NEXT:  func.func @simple_cycle(%arg0_1 : i64) -> i64 {
+    // CHECK-NEXT:    func.return %arg0_1 : i64
+    // CHECK-NEXT:  }
+
+    // -----
+
+    func.func @cycle_singleblock(%arg0: i64) -> i64 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i16
+        %1 = "builtin.unrealized_conversion_cast"(%0) : (i16) -> i1
+        %2 = "builtin.unrealized_conversion_cast"(%1) : (i1) -> i64
+        %3 = "builtin.unrealized_conversion_cast"(%1) : (i1) -> i16
+        %4 = "builtin.unrealized_conversion_cast"(%3) : (i16) -> i64
+        %5 = "test.op"(%2, %4) : (i64, i64) -> i64
+        func.return %5 : i64
+    }
+
+    // CHECK-NEXT:    func.func @cycle_singleblock(%arg0_2 : i64) -> i64 {
+    // CHECK-NEXT:      %0 = "test.op"(%arg0_2, %arg0_2) : (i64, i64) -> i64
+    // CHECK-NEXT:      func.return %0 : i64
+    // CHECK-NEXT:    }
+
+    // -----
+
+    func.func @cycle_multiblock(%arg0: i64) -> i64 {
+        %c0 = "test.op"() : () -> i32
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i16
+        %1 = "builtin.unrealized_conversion_cast"(%0) : (i16) -> i1
+        %2 = "builtin.unrealized_conversion_cast"(%1) : (i1) -> i64
+        %3 = "builtin.unrealized_conversion_cast"(%1) : (i1) -> i16
+        "cf.br"(%c0)[^bb1] : (i32) -> ()
+        ^bb1(%33: i32):  // pred: ^bb0
+        %4 = "builtin.unrealized_conversion_cast"(%3) : (i16) -> i64
+        %5 = "test.op"(%2, %4) : (i64, i64) -> i64
+        func.return %5 : i64
+    }
+
+    // CHECK-NEXT:    func.func @cycle_multiblock(%arg0_3 : i64) -> i64 {
+    // CHECK-NEXT:      %c0 = "test.op"() : () -> i32
+    // CHECK-NEXT:      "cf.br"(%c0) [^0] : (i32) -> ()
+    // CHECK-NEXT:    ^0(%1 : i32):
+    // CHECK-NEXT:      %2 = "test.op"(%arg0_3, %arg0_3) : (i64, i64) -> i64
+    // CHECK-NEXT:      func.return %2 : i64
+    // CHECK-NEXT:    }
+
+    // -----
+
+    func.func @failure_simple_cast(%arg0: i64) -> i32 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i32
+        func.return %0 : i32
+    }
+
+    // CHECK-NEXT:    func.func @failure_simple_cast(%arg0_4 : i64) -> i32 {
+    // CHECK-NEXT:      %3 = "builtin.unrealized_conversion_cast"(%arg0_4) : (i64) -> i32
+    // CHECK-NEXT:      func.return %3 : i32
+    // CHECK-NEXT:    }
+
+    // -----
+
+    func.func @failure_cycle(%arg0: i64) -> i32 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i1
+        %1 = "builtin.unrealized_conversion_cast"(%0) : (i1) -> i32
+        func.return %1 : i32
+    }
+
+    // CHECK-NEXT:    func.func @failure_cycle(%arg0_5 : i64) -> i32 {
+    // CHECK-NEXT:      %4 = "builtin.unrealized_conversion_cast"(%arg0_5) : (i64) -> i1
+    // CHECK-NEXT:      %5 = "builtin.unrealized_conversion_cast"(%4) : (i1) -> i32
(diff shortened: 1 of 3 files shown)


superlopuh (changes requested):

tests/filecheck/transforms/reconcile_unrealized_casts.mlir
```diff
@@ -0,0 +1,93 @@
...
+
+    // -----
+
+    func.func @failure_simple_cast(%arg0: i64) -> i32 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i32
+        func.return %0 : i32
+    }
+
+    // CHECK-NEXT:    func.func @failure_simple_cast(%arg0_4 : i64) -> i32 {
+    // CHECK-NEXT:      %3 = "builtin.unrealized_conversion_cast"(%arg0_4) : (i64) -> i32
```
superlopuh, line 75: Can you replace the exact names for regexes? Makes it easier to move test code around
```suggestion
    // CHECK-NEXT:      %{{.*}} = "builtin.unrealized_conversion_cast"(%{{.*}}) : (i64) -> i32
```

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+    that is used to enable the inter-mixing of different type systems.
+
+    This pattern rewriter removes `unrealized_conversion_cast` operations that are not needed.
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: builtin.UnrealizedConversionCastOp, rewriter: PatternRewriter
+    ) -> None:
+        # We remove severals cast in a single pass, which should not happen
```
compor, line 31: typo
```suggestion
        # We remove several casts in a single pass, which should not happen
```

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class RemoveUselessConversionCast(RewritePattern):
```
compor, line 19: why not `ReconcileUnrealizedCasts` like in MLIR? Isn't it pretty much the same functionality?
moreover, I'm not really fond of the "useless" in the name; there was utility once but now it's a noop.

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if op.parent is not None:
+            # casts that either have no uses or have at least
+            # one user that isn't an unrealized cast.
+            finalized_casts = list[builtin.UnrealizedConversionCastOp]()
+
+            # casts with only other unrealized casts as users.
+            pending_casts = deque[builtin.UnrealizedConversionCastOp]()
+
+            # casts that neeed to be visited in the DFS traversal of the use-def chain.
```
compor, line 44: great comments.

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+            # DFS traversal of the use-def chain.
+            while casts_to_visit:
+                cast = casts_to_visit.pop()
+
+                is_live = False
+                has_any_uses = False
+
+                for use in gen_all_uses_cast(cast):
+                    if isinstance(use.operation, builtin.UnrealizedConversionCastOp):
+                        # early check, for sure we are not able to remove such a cast
```
compor, line 62:
```suggestion
                        # early check, for making sure that we are not able to remove such a cast
```

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+
+def reconcile_unrealized_casts(op: ModuleOp):
+    """
+    Removes all `builtin.unrealized_conversion_cast` operations that are not needed in a module.
+    """
+    walker = PatternRewriteWalker(RemoveUselessConversionCast())
+    walker.rewrite_module(op)
+
+
+class ReconcileUnrealizedCasts(ModulePass):
```
compor, line 113: maybe suffix this with `Pass` so you can rename the rewrite pattern?

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+                for use in gen_all_uses_cast(cast):
+                    if isinstance(use.operation, builtin.UnrealizedConversionCastOp):
+                        # early check, for sure we are not able to remove such a cast
+                        if any(
+                            r != i
+                            for r, i in itertools.zip_longest(
+                                use.operation.inputs, cast.results
+                            )
+                        ):
+                            warn(
```
compor, line 69: is there a test for this condition?

tests/filecheck/transforms/reconcile_unrealized_casts.mlir
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    // CHECK-NEXT:    func.func @failure_chain(%{{.*}} : i64) -> i32 {
+    // CHECK-NEXT:      %4 = "builtin.unrealized_conversion_cast"(%{{.*}}) : (i64) -> i1
+    // CHECK-NEXT:      %5 = "builtin.unrealized_conversion_cast"(%4) : (i1) -> i32
+    // CHECK-NEXT:      func.return %5 : i32
+    // CHECK-NEXT:    }
+
+
+    func.func @cycle_singleblock_var_ops(%arg0: i64, %arg1: i64) -> i64 {
```
compor, line 94: good stuff!

compor (approved):

xdsl/transforms/reconcile_unrealized_casts.py
adutilleul: I have added one test, but the latter check should fail in that case anyway.

tests/filecheck/transforms/reconcile_unrealized_casts.mlir
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // CHECK-NEXT:  func.func @simple_cycle(%arg0_1 : i64) -> i64 {
+    // CHECK-NEXT:    func.return %arg0_1 : i64
+    // CHECK-NEXT:  }
+
+    // -----
+
+    func.func @cycle_singleblock(%arg0: i64) -> i64 {
+        %0 = "builtin.unrealized_conversion_cast"(%arg0) : (i64) -> i16
```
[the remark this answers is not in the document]
superlopuh: Thank you, can you please add something like this:
```
def test(a: i32, b: i64) -> i16:
  c: i16, d: i16 = a, b
  return c
```
To test that a variadic cast reduces the number of operands

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+    that is used to enable the inter-mixing of different type systems.
+
+    This pattern rewriter removes `unrealized_conversion_cast` operations that are not needed.
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: builtin.UnrealizedConversionCastOp, rewriter: PatternRewriter
+    ) -> None:
+        # We remove several casts in a single pass, which should not happen
```
superlopuh, line 31: Are we sure we want this behaviour? The similar pass in the workshop-all branch only modifies one cast at a time, and the only difference seems to be that it changes cast chains like `abcd` to `ad`, unlike MLIR. I would argue that if we want to go with the MLIR approach then we should write this not as a rewrite pattern but a straight-up function that iterates over the IR itself.

superlopuh (changes requested):

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,117 @@
...
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class ReconcileUnrealizedCasts(RewritePattern):
+    """
+    An `unrealized_conversion_cast` operation represents an unrealized conversion from one set of types to another,
```
math-fehr, line 21: Can you wrap this line to 88 characters? (And the other lines in this comment)

math-fehr (approved):
Nice!
I'm wondering though how much we should in the long term implement these things, and how much we should use MLIR for that instead!
For now, let's use this though.

xdsl/transforms/reconcile_unrealized_casts.py
adutilleul: I've just done it, and if that's alright with you I think we can merge @superlopuh ?

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,107 @@
...
+            # remove other casts in the chain in the right order
+            PatternRewriter(cast).erase_matched_op()
+
+
+def reconcile_unrealized_casts(module: ModuleOp):
+    """
+    Removes all `builtin.unrealized_conversion_cast` operations
+    that are not needed anymore in a module.
+    """
+    for op in filter(
```
superlopuh, line 96:
```suggestion
    for op in module.walk():
        if not isinstance(op, builtin.UnrealizedConversionCastOp):
            continue
```

superlopuh (changes requested):
thank you, one last small change

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -0,0 +1,107 @@
...
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import PatternRewriter
+
+
+def _try_remove_cast_chain(op: builtin.UnrealizedConversionCastOp):
+    # We remove several casts in a single pass, so we need to
+    # ignore casts that have already been removed earlier in the pass
+    # (i.e: no parent anymore).
+
+    if op.parent is not None:
```
superlopuh, line 19: Will this ever fail? I don't think it actually can

xdsl/transforms/reconcile_unrealized_casts.py
compor: is this a runtime overhead concern?

xdsl/transforms/reconcile_unrealized_casts.py
superlopuh: Thinking about it some more, I think modifying the IR structure while walking it is not allowed in python, by the contract of iterators. We should probably at least cache the operations we want to work on before calling this function, so we iterate over a list that is not mutated, even if the operations inside it are

xdsl/transforms/reconcile_unrealized_casts.py
superlopuh: it's a check for class name twice concern really, this seems neater just in terms of not duplicating logic and adding a cast. I think in general filter and map is recommended against because they're both slower and less well supported by pyright

superlopuh (approved):
Thank you!

PR #1260, opened 2023-07-10
misc: Rename Typ to Type

kingiler:
This is a follow up to a previous PR #1240 .


docs/Toy/Toy_Ch2.ipynb
```diff
@@ -392,8 +392,14 @@
   }
  ],
  "metadata": {
+  "kernelspec": {
```
compor, line 395: Can you please clear this metadata? The `Makefile` has a rule that shows how this is done.

docs/Toy/Toy_Ch2.ipynb
```diff
@@ -392,8 +392,14 @@
...
+   "display_name": ".venv",
+   "language": "python",
+   "name": "python3"
+  },
   "language_info": {
-   "name": "python"
+   "name": "python",
```
compor, line 401: Same here regarding metadata.

compor (changes requested):
While in general, I like this kind of change, I find the minor irk that seeing `typ` instead of `type` personally causes me to be vastly inferior to the effect of such a repo-wide change on the codebase.

I'm OK with this going either way.

superlopuh (approved):
I like this change a lot. If anyone asked me whether it would be worth spending the time to do this kind of refactoring I'd say no, but if the change is made and is an improvement to readability then the effort of clicking "merge" is more than worth it. It's not even like it's difficult to review, there are no functional changes.

webmiche:
I just wonder about downstream changes for currently open PRs. To me, if anyone merge conflicts because of this PR, we have already lost everything that we gain from it.

math-fehr (approved):
I think I'm a bit conflicted on this PR.
On one hand, I'm in favor of changing the *public* `Typ` to `Type`, as we should have coherent names in our public API.
However, on the other hand, this contains a lot of changes of variable names, that are local to functions, and not at all public. I don't know if this is worth changing these names, as they do not really harm anything, and we can always rename them when we are changing the code. Now, there is going to be a lot of conflicts on these private parts because of this PR.

I think that since this is already implemented, we could merge it, but I don't think it's worth making these changes anymore, unless we clearly gain from it. I think we gain from improving public API (to certain extent, since this introduce a change to downstream users), but I don't think we should pass much time making small grammar changes in the internal code.

superlopuh:
I think we all agree that the time making this change could have been spent more productively but I still think we're better off merging it now it's done.

kingiler:
So should this PR be closed, or merged, or still need extra time for discussion? I am happy to listen to public opinions.

PR #1261, opened 2023-07-11
tests: (misc) Replace scf with test dialect dependency in pattern rewriter tests

compor:
This PR replaces the dependency on the `scf` dialect to `test`.

The nesting and region/block structure is preserved, but instead of using `scf.if` it is replaced with `test.testop`.
This also piggybacks a few Pyright suggestions:
- unused method variables renamed (e.g., from `op` to `__op__`
- inconsistent return values (e.g., `return` and `return None`)

Also, squeezed in the removal of a tiny `func` dependency.

This came about while I was looking into traits and I thought it suits better a separate PR.


tests/test_pattern_rewriter.py
```diff
@@ -1,10 +1,9 @@
 from conftest import assert_print_op
 
-from xdsl.dialects import test
 from xdsl.dialects.arith import Addi, Arith, Constant, Muli
 from xdsl.dialects.builtin import Builtin, IntegerAttr, ModuleOp, StringAttr, i32, i64
-from xdsl.dialects.scf import If, Scf
-from xdsl.ir import Block, MLContext, Operation, Region
+from xdsl.dialects.test import Test, TestOp, TestType
```
webmiche, line 5: Once upon a time, we decided to try to make these look like this:
```
from xdsl.dialects import test
```

In this way, we can then write `test.testop` which looks a bit more like the SSA-form.
Not saying you should do this, just saying that you could, if you wanted to :wink:

tests/test_pattern_rewriter.py
```diff
@@ -540,27 +538,31 @@ def match_and_rewrite(self, op: ModuleOp, rewriter: PatternRewriter):
...
 
     class Rewrite(RewritePattern):
         @op_type_rewrite_pattern
-        def match_and_rewrite(self, op: If, rewriter: PatternRewriter):
-            rewriter.modify_block_argument_type(op.true_region.blocks[0].args[0], i64)
+        def match_and_rewrite(self, matched_op: TestOp, rewriter: PatternRewriter):
+            if matched_op.regs and matched_op.regs[0].blocks:
+                rewriter.modify_block_argument_type(
+                    matched_op.regs[0].blocks[0].args[0], i64
+                )
```
webmiche, line 565: Not sure how others feel, but would it make sense to had a RegOp in the testing dialect, that has arbitrarily many regions? This way, you could get rid of this if, which feels not quite as nice to me.

webmiche (approved):
Thanks a lot, as a wise man once said, you are doing the lord's work!

tests/test_pattern_rewriter.py
compor: Nice idea.

tests/test_pattern_rewriter.py
compor: Hmm, I have no problem with this, it does look nicer to me as well.
Maybe that's something for documentation on code conventions? This is the first time I "hear" this.

tests/test_pattern_rewriter.py
webmiche: Personally, I think for rewriting with the testing dialect, it might make sense to add these structures. I would make sure to restrict ourselves to very few, but I don't think a couple more operations (with clear purpose) makes the testing dialect too big. What do you think, @math-fehr?

webmiche: And @superlopuh I guess

tests/test_pattern_rewriter.py
webmiche: Yes, we should put this on our todo list. And opening a todo list should be on there too :thinking:

webmiche:
There it is. It was probably just hungover again, my guess is that codecov is a heavy drinker that sometimes just talks bs or turns up late for work :thinking:

compor:
> There it is. It was probably just hungover again, my guess is that codecov is a heavy drinker that sometimes just talks bs or turns up late for work thinking

It needs alcohol to forget what it sees daily.

tests/test_pattern_rewriter.py
superlopuh: Seems consistent, would probably make life a little easier. It's the sort of thing I'd do as a subtask of a task, not as a separate goal TBH.

compor: @superlopuh I know what you mean, but I can add it separately in another PR and apply it here, since there is no immediate need for it.
I'd rather not stall this PR for that, but I don't really mind, so up to you guys.

webmiche: Sure, we can merge this and open an issue or something.

webmiche:
I think we can merge this. The test dialect discussion is orthogonal, IMO.

compor:
> I think we can merge this. The test dialect discussion is orthogonal, IMO.

I'll open an issue to examine your suggestion @webmiche

PR #1263, opened 2023-07-12
reqs: add typing-extension dependency

superlopuh:
Self is a very useful type to have available, introduced in 3.11. In Python 3.10 a third-party module is required for the same functionality. I reckon the extra dependency is worth it for the kind of thing we'd like to express.

It seems to me like it's necessary to make it a required dependency as core code should be able to use Self, so users of pip install should inherit it as well. It's a bit annoying to have a dependency that's only required for python 3.10 but it seems worth it to me.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 pip<24.0
 immutabledict<2.2.6
-typing-extensions<4.8
+typing-extensions>=4.7,<5
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 pip<24.0
 immutabledict<2.2.6
+typing-extensions<4.8
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -11,7 +11,6 @@ pre-commit==3.3.3
 riscemu==2.1.0
 asv<0.6
 isort==5.12.0
-typing-extensions<4.8
 # pyright has to be the last line and fixed with `==`. The CI parses this file
 # in `.github/parse_pyright_version.py` and installs the according version for
 # typechecking.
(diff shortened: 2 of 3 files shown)


webmiche (commented):
What do you think when will python 3.11 be shipped with ubuntu? I guess at that point we could remove 3.10 in general, IMO.

Until then, I am fine with having this additional dependency.

superlopuh:
Yeah ideally we'd drop it as soon as we drop 3.10

compor (approved):

adutilleul (approved):

kingiler:
@webmiche Since Ubuntu Lunar (23.04), `python3` is `3.11` as in [here](https://packages.ubuntu.com/lunar/python3).

requirements.txt
```diff
@@ -1,2 +1,3 @@
 pip<24.0
 immutabledict<2.2.6
+typing-extensions<4.8
```
kingiler, line 3:
```suggestion
typing-extensions>=4.8,<5
```
as suggested on the homepage.

kingiler (changes requested):
LGTM

PapyChacal (approved):

math-fehr (approved):
:+1: 
I don't think we want to remove Python 3.10 support any time soon though (if we can easily keep compatibility, we should). But I think that depending on typing-extensions is quite safe!

requirements.txt
superlopuh: It worked! I was too pessimistic. Let's see how dependabot deals with the update

kingiler (approved):

PR #1264, opened 2023-07-12
interpreter: add cast mechanism

superlopuh:


compor (approved):

adutilleul (approved):

xdsl/interpreters/builtin.py
```diff
@@ -0,0 +1,19 @@
...
+    def run_cast(
+        self,
+        interpreter: Interpreter,
+        op: UnrealizedConversionCastOp,
+        args: tuple[Any, ...],
+    ):
+        return tuple(
+            interpreter.cast_value(o.type, r.type, arg)
+            for (o, r, arg) in zip(op.operands, op.results, args)
+        )
```
PapyChacal, line 19: Would be nice to have a test case for this one!

PapyChacal (approved):

xdsl/interpreters/builtin.py
superlopuh: Ah yes you're right

math-fehr (commented):
I have to say, I'm not sure that this is the right approach (but maybe I'm overlooking something).
Is there a reason why we have a special cast mechanism in place, when we could just have the `UnrealizedConversionCast` interpreter implementation? What do we gain from adding this complexity?

superlopuh:
> I have to say, I'm not sure that this is the right approach (but maybe I'm overlooking something).
> 
> Is there a reason why we have a special cast mechanism in place, when we could just have the `UnrealizedConversionCast` interpreter implementation? What do we gain from adding this complexity?

In the Toy pipeline, it's the fake accelerator for pointwise addition and multiplication that uses this mechanism, not the builtin dialects.

This is kind of a composition vs inheritance thing, if I have a toy vector extension with some casts, and maybe some IO thing with some casts, how do they both override the implementation of casting in the builtin interpreter functions that lets them both work without knowing about each other?

math-fehr:
Hmm, okay fair I think, works for me!

PR #1267, opened 2023-07-13
Change `parse_parameter` to `classmethod`

kingiler:


PapyChacal:
I have no objection, but could you give a reason for this change?

kingiler:
One would like to access to `self` or `cls` during parsing.

```
class Base(Data[str]):
    @classmethod
    def parse_parameter(cls, parser: AttrParser) -> Self:
        ops = parser.parse_operands()
        attrs = cls.parse_attributes(parser) # allow custom parsing in subclass
        return cls.create(ops, attrs)

    @classmethod
    @abstractmethod
    def parse_attributes(cls, parser: AttrParser) -> dict[str, Attribute]:
        raise NotImplementatedError()

class Derived(Base):
    @classmethod
    def parse_attributes(cls, parser: AttrParser) -> dict[str, Attribute]:
        ...
```

math-fehr (approved):
Yes, looks good to me!
The reason for this change is both to allow inheritance in ParametrizedAttributes, and also to make it consistent with how `Operation` works!

PapyChacal (approved):
Thanks :slightly_smiling_face:

PR #1268, opened 2023-07-13
dev: Experiment with ruff

kingiler:


tests/interpreters/test_riscv_emulator.py
```diff
@@ -10,7 +10,7 @@
 
 pytest.importorskip("riscemu", reason="riscemu is an optional dependency")
 
-from xdsl.interpreters.riscv_emulator import RV_Debug, run_riscv
+from xdsl.interpreters.riscv_emulator import RV_Debug, run_riscv  # noqa: E402
```
superlopuh, line 13: what rule is this?

superlopuh:
Thank you for opening this! I like all the changes. I think we should merge them, potentially in multiple PRs, one for each rule. So maybe start with a less nice ruff config for which the current code passes, or as close as possible, and then we can discuss the rules we've ignored one by one?

tests/interpreters/test_riscv_emulator.py
kingiler: [Import module](https://beta.ruff.rs/docs/rules/module-import-not-at-top-of-file/)

According to [PEP 8](https://peps.python.org/pep-0008/#imports), "imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants."

.pre-commit-config.yaml
```diff
@@ -2,15 +2,19 @@
...
+-   repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.278
     hooks:
-    -   id: isort
-        args: ["--profile", "black", "--filter-files"]
+    -   id: ruff
+        types_or: [
+            python, 
+            pyi, 
+            # jupyter,
```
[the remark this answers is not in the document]
kingiler: Yes, mainly with those import module issues.

kingiler:
There are some rules that I like but would cause a huge change in code base: [UP](https://beta.ruff.rs/docs/rules/#pyupgrade-up). These are mainly the Python syntax issues that I file PR before but there are still much more.

math-fehr (approved):
These changes seems okay to me, so I would be in favor of merging it!

tests/interpreters/test_riscv_emulator.py
superlopuh: Can you please move it to the top in a separate PE or ignore the rule in the project?

pyproject.toml
```diff
@@ -18,3 +18,42 @@ typeCheckingMode = "strict"
...
+[tool.ruff]
+select = ["E", "W", "I"]
+ignore = ["E741"]
+
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+fixable = ["A", "B", "C", "D", "E", "F", "G", "I", "N", "Q", "S", "T", "W", "ANN", "ARG", "BLE", "COM", "DJ", "DTZ", "EM", "ERA", "EXE", "FBT", "ICN", "INP", "ISC", "NPY", "PD", "PGH", "PIE", "PL", "PT", "PTH", "PYI", "RET", "RSE", "RUF", "SIM", "SLF", "TCH", "TID", "TRY", "UP", "YTT"]
+unfixable = []
+
+# Exclude a variety of commonly ignored directories.
+exclude = [
```
compor, line 31: can we skip this? it's part of the default config, right? I don't see any changes according to https://beta.ruff.rs/docs/configuration/#using-pyprojecttoml

pyproject.toml
```diff
@@ -18,3 +18,42 @@ typeCheckingMode = "strict"
...
+    ".venv",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "venv",
+]
+line-length = 300
```
compor, line 54: I think we use 120

pyproject.toml
```diff
@@ -18,3 +18,42 @@ typeCheckingMode = "strict"
 
 [tool.isort]
 profile = "black"
+
+[tool.ruff]
+select = ["E", "W", "I"]
+ignore = ["E741"]
+
+# Allow autofix for all enabled rules (when `--fix`) is provided.
```
compor, line 26: the comment says `ALL` but there's an explicit list? does this list correspond to the `ALL` option or is there a change? if so, why and which?

compor (changes requested):

pyproject.toml
kingiler: I don't why but there are lines over 200 characters in current repo.

math-fehr: We use 88 actually! (That's `black` config).
Lines over 200 characters should be reduced).

pyproject.toml
kingiler: Ok, let's remove default config.

pyproject.toml
```diff
@@ -18,3 +18,42 @@ typeCheckingMode = "strict"
...
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "venv",
+]
+line-length = 300
+dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
```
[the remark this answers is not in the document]
kingiler: Ok, let's remove default config.

pyproject.toml
kingiler: Ok, let's remove default config.

compor:
I've been using it for the last 2 weeks with a close-to default config and I'm quite happy with it.
Looking forward to this being added.

pyproject.toml
kingiler: Let do it in next PR to keep this PR small.

superlopuh: Ruff and black have different meanings for line lengths, for comments and maybe other things

tests/interpreters/test_riscv_emulator.py
superlopuh: Ah ok this makes sense. @kingiler can you please add a comment with this explanation? Then I think it's good to merge

.pre-commit-config.yaml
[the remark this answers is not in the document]
superlopuh: Ok, can you please file an issue to uncomment this line, and add a reference to it in this comment?
```suggestion
            # jupyter, # github.com/xdslproject/xdsl/123456
```

superlopuh (changes requested):

tests/interpreters/test_riscv_emulator.py
kingiler: I'm not sure what comment to add? There are quite some `# noqa:` in the code base already without comments.

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
         types_or: [
             python, 
             pyi, 
-            # jupyter,
+            # jupyter, # https://github.com/xdslproject/xdsl/issues/1273
```
superlopuh, line 15: 🙏

superlopuh (approved):

compor (approved):

PR #1205, opened 2023-06-27
dialect: RISCV custom syntax

kingiler:
A PoC addressing [issue](https://github.com/xdslproject/xdsl/issues/1067#issue-1740238906).

PR #1269, opened 2023-07-13
dialect: Refactor riscv register

kingiler:
Preparing for #1205

`IntegerRegister` and `FloatRegister` inherit from `RegisterType`
`IntegerRegisterType` and `FloatRegisterType` inherit from `RegisterType`


xdsl/dialects/riscv.py
```diff
@@ -166,9 +172,9 @@ def verify(self) -> None:
 
 
 @irdl_attr_definition
-class FloatRegisterType(Data[Register], TypeAttribute):
+class FloatRegisterType(RegisterType):
```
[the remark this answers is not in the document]
kingiler: Refactored in new commit.

xdsl/dialects/riscv.py
```diff
@@ -57,6 +59,11 @@ class Register:
     name: str | None = field(default=None)
     """The register name. Should be one of `ABI_INDEX_BY_NAME` or `None`"""
 
+    ABI_INDEX_BY_NAME: dict[str, int] = field(init=False)
```
superlopuh, line 62:
```suggestion
    abi_index_by_name: ClassVar[dict[str, int]] = field(init=False)
```

superlopuh: it might be better off as an abstract class function than a class variable TBH, I don't think it's useful to be able to specify this value at runtime.

xdsl/dialects/riscv.py
```diff
@@ -145,13 +159,21 @@ def register_name(self) -> str:
         return self.data.name
 
     @classmethod
-    def parse_parameter(cls, parser: AttrParser) -> Register:
+    def new_register(cls, name: str | None = None) -> RegisterT:
```
superlopuh, line 162: This is just the init, right?

xdsl/dialects/riscv.py
```diff
@@ -127,15 +138,18 @@ class Register:
...
     }
+    ABI_INDEX_BY_NAME = RV32F_INDEX_BY_NAME
 
 
-@irdl_attr_definition
-class RegisterType(Data[Register], TypeAttribute):
+RegisterT = TypeVar("RegisterT", bound="Register")
+
+
+class RegisterType(GenericData[RegisterT], TypeAttribute):
```
superlopuh, line 147: @math-fehr now that I think about it, maybe this should be `RegisterAttr` instead?

xdsl/dialects/riscv.py
```diff
@@ -57,6 +59,11 @@ class Register:
     name: str | None = field(default=None)
     """The register name. Should be one of `ABI_INDEX_BY_NAME` or `None`"""
 
+    ABI_INDEX_BY_NAME: dict[str, int] = field(init=False)
+
+
+@dataclass(frozen=True)
+class IntegerRegister(Register):
```
superlopuh, line 66: Do you think it might be better to call this `IntRegister` instead? I'm usually all for full length names of things but it's just used so often that I don't think it's that useful to have the extra `eger` everywhere. This is more of a nit.

xdsl/dialects/riscv.py
```diff
@@ -127,15 +138,18 @@ class Register:
...
+RegisterT = TypeVar("RegisterT", bound="Register")
+
+
+class RegisterType(GenericData[RegisterT], TypeAttribute):
     """
     A RISC-V register type.
     """
 
-    name = "riscv.reg"
+    register_class: type[RegisterT]
```
superlopuh, line 152: I don't think we need this tbh, let's start with a less flexible non-generic attribute with two subclasses, and use abstract methods rather than do full generic approach.

xdsl/dialects/riscv.py
```diff
@@ -160,116 +182,108 @@ def print_parameter(self, printer: Printer) -> None:
...
-    FS7 = FloatRegisterType(Register("fs7"))
-    FS8 = FloatRegisterType(Register("fs8"))
-    FS9 = FloatRegisterType(Register("fs9"))
-    FS10 = FloatRegisterType(Register("fs10"))
-    FS11 = FloatRegisterType(Register("fs11"))
-    FT8 = FloatRegisterType(Register("ft8"))
-    FT9 = FloatRegisterType(Register("ft9"))
-    FT10 = FloatRegisterType(Register("ft10"))
-    FT11 = FloatRegisterType(Register("ft11"))
+    ZERO = IntegerRegisterType(IntegerRegister("zero"))
```
superlopuh, line 221: Could you please add a helper to init register types with strings? I should probably have done this myself earlier but it seems more pressing with all the added letters in these lines :)

superlopuh (changes requested):
I like these changes a lot, just a few architecture changes would be nice before merging this

xdsl/dialects/riscv.py
kingiler: There is already a helper. Now just rewrite with the helper function.

xdsl/dialects/riscv.py
kingiler: Renamed.

xdsl/dialects/riscv.py
kingiler: Actually not. `new_register_type` is the init indeed.

superlopuh: Not sure why, but I think my comments got shifted in terms of line numbers.

superlopuh:
I played around with this branch a bit, and think there's an alternative direction worth considering, could you please take a look? #1281

kingiler:
Close in favour of #1281.

PR #1270, opened 2023-07-13
core: Restructure AffineExpr

Groverkss:
This patch allows users to walk on the AffineExpr AST by matching the corresponding AffineExpr class. Also significantly reduces the code size.


math-fehr (approved):
Really nice! Makes it much nicer!

webmiche (approved):
Nice, LGTM!

PR #1275, opened 2023-07-14
reqs: Improve pip command

kingiler:
It is recommanded to use `python -m pip` than plain `pip`. As in some cases `pip` may not come from the virtualenv while the former ensure the `pip` is bundled with the `python` used.


AntonLydike:
Why is is preferable to use the pip *not* from the virtualenv? Shouldn't you want to use that one?

kingiler:
What I mean is the `pip` may not come from the virtualenv, which I encounter a couple of times.

superlopuh:
I would recommend using `--require-venv` instead

AntonLydike:
Oooh, sorry, misunderstood that PR. Looks good, `--require-venv` may even be better!

kingiler:
Tested on my computer, `--require-virtualenv` seems not getting what is intended.

AntonLydike (approved):

kingiler:
`--require-virtualenv` is added for completeness.

georgebisbas (dismissed):
Why is this change happening?
Our ci setup looks fine with this.
Is this a problem you encountered locally?

georgebisbas:
Also, kindly check this: https://github.com/xdslproject/xdsl/wiki/Tags-for-commit-messages-and-PR-titles
to prepare your PR

math-fehr (approved):
:+1:

kingiler:
@georgebisbas, yes a little problem using the `make venv`.

georgebisbas:
> @georgebisbas, yes a little problem using the `make venv`.
What was the problem? Can you paste the log?

Makefile
```diff
@@ -88,5 +88,5 @@ coverage-report-md:
 venv: requirements-optional.txt requirements.txt
 	python3 -m venv ${VENV_DIR}
 	source ${VENV_DIR}/bin/activate
-	pip install -r requirements-optional.txt -r requirements.txt
-	pip install -e ".[extras]"
+	python3 -m pip --require-virtualenv install -r requirements-optional.txt -r requirements.txt
```
superlopuh, line 91: why the python3?

kingiler: Just be consistent with line 89 `python3 -m venv ${VENV_DIR}`

superlopuh: ah didn't spot that... @AntonLydike is that so people don't run it with Python 2 or Python 4? :)

AntonLydike: Just a weird habit I picked up a while back. Maybe related to some work on ancient servers with questionable defaults^^

superlopuh (approved):

georgebisbas (approved):

PR #1278, opened 2023-07-14
[WIP] interpreter: draft WGSL printer

cigarichard:
It is translation from MLIR to WebGpu Shading Language(WGSL), and planning to with tests.
It is planning to have jit compulation in WebGpu in the xDsl interpreter.


tobiasgrosser:
Hi @cigarichard, would it be possible to rebase this branch on main? Also, I feel some simple end-to-end filecheck test cases would be great. I wonder how we could achieve this best?

PapyChacal:
> Hi @cigarichard, would it be possible to rebase this branch on main? Also, I feel some simple end-to-end filecheck test cases would be great. I wonder how we could achieve this best?

So we worked with some examples, and a toy tool to parse IR and prints WGSL from it.
Thing is, I'm not sure this is a tool we want upstream; the translation is meant for individual gpu.;odules if not gpu.funcs, and I'm just not sure how to merge everything in one output making sense for example.

I added a more complete, but still pytest-based test, that @cigarichard used for first translations. Is it fitting what you expected?
Otherwise, I definitely see some more complete end-to-end tests once it's hooked in the interpreter! :slightly_smiling_face: 

Also, NB: The memref accesses are placeholders at this point. An interesting issue is that WGSL only has unidimensional arrays, and no arrays of arrays. So @cigarichard is working on delinearizing those for the known-size case we're interested in here, and generalizing to any memref can be a challenge for later.

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,249 @@
...
+    @print.register
+    def _(self, op: arith.Constant, out_stream: IO[str]):
+        value = int(str(op.attributes.get("value")).split()[0])
+        cons_type = op.result.type
+        if op.result.type.name == "index":
+            cons_type = "u32"
+        name_hint = self.wgsl_name(op.result)
+        if cons_type == "u32":
+            if value == -1:
+                value = 4294967295
```
PapyChacal, line 180: Could you rewrite this so it works for any negative value?

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,249 @@
...
+        value = self.wgsl_name(op.value)
+        store_ref = self.wgsl_name(op.memref)
+        indices = [self.wgsl_name(i) for i in op.indices]
+        index_values: list[str] = []
+        for i in range(memref_dimension):
+            product_of_dims = 1
+            for dim in memref_size[i + 1 :]:
+                product_of_dims *= dim
+            index_values.append(f"{product_of_dims} * {indices[i]}")
+        index_value = " + ".join(index_values)
```
PapyChacal, line 160: Could be factorized!

PapyChacal (approved):

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+from xdsl.ir import Operation, SSAValue
+from xdsl.ir.core import Attribute
+
+
+class WGSLPrinter:
+    name_dict: dict[SSAValue, str] = dict()
+    count = 0
+
+    def wgsl_name(self, v: SSAValue):
+        if v not in self.name_dict.keys():
```
AntonLydike, line 17: You can drop the `.keys()` here, `x in <some dict>` already checks if x is a key of the dict.

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+            if v.name_hint is not None:
```
AntonLydike, line 18: I think this won't work if you have two ssa values with the same name hint? You also need to keep track of the name hints you accepted (you can see the normal printer for that. Might be something worth factoring out later?)

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+            else:
+                self.name_dict[v] = f"v{self.count}"
+                self.count += 1
+        return self.name_dict[v]
+
+    @singledispatchmethod
+    def print(self, op: Operation, out_stream: IO[str]) -> None:
+        raise NotImplementedError(
+            f"Printing the {op.name} to WGSL is not implemented yet."
+        )
```
AntonLydike, line 29: :heart_eyes:

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @print.register
+    def _(self, op: gpu.ModuleOp, out_stream: IO[str]):
+        for o in op.body.ops:
+            if isinstance(o, gpu.FuncOp):
+                self.print(o, out_stream)
+
+    @print.register
+    def _(self, op: gpu.FuncOp, out_stream: IO[str]):
+        print(f"Thats a gpu func : {op}")
```
AntonLydike, line 39: remains of `printf` debugging?

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+        for arg in op.body.block.args:
+            auth = "read"
+            arg_type = ""
+            for use in arg.uses:
+                if isinstance(use.operation, memref.Store):
+                    auth = "read_write"
+            if isinstance(arg.type, builtin.Float32Type):
```
AntonLydike, line 46: `if arg.type == builtin.f32` would be easier I guess?

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+
+        out_stream.write(
+            """
+    @compute
+    @workgroup_size(1)
+    fn main(@builtin(global_invocation_id) global_invocation_id : vec3<u32>,
+    @builtin(workgroup_id) workgroup_id : vec3<u32>,
+    @builtin(local_invocation_id) local_invocation_id : vec3<u32>,
+    @builtin(num_workgroups) num_workgroups : vec3<u32>) {
+"""
```
AntonLydike, line 67: This looks wrong, hardcoded name `main`? Is all that really "fixed"?

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+        dim = str(op.dimension.value.param).replace('"', "")
+        name_hint = self.wgsl_name(op.result)
+        out_stream.write(
+            f"""
+        let {name_hint}: u32 = workgroup_id.{dim};"""
+        )
+
+    @print.register
+    def _(self, op: gpu.ThreadIdOp, out_stream: IO[str]):
+        dim = str(op.dimension.value.param).replace('"', "")
```
AntonLydike, line 92: you can use `.strip('"')` instead, it may be a bit more robust? idk. Feels more "proper".

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+        out_stream.write(
+            f"""
+        let {name_hint}: u32 = global_invocation_id.{dim};"""
+        )
+
+    @print.register
+    def _(self, op: memref.Load, out_stream: IO[str]):
+        memref_type = cast(MemRefType[Attribute], op.memref.type)
+        memref_dimension = memref_type.get_num_dims()
+        memref_size = memref_type.get_shape()
```
AntonLydike, line 130: You should assert that it has a fixed size, otherwise you may get some funny surprises where `memref<100x?>` has size -100 :sweat_smile:

xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -0,0 +1,248 @@
...
+    @print.register
+    def _(self, op: gpu.ModuleEndOp, out_stream: IO[str]):
+        # Nothing to print :)
+        pass
+
+    @print.register
+    def _(self, op: arith.Constant, out_stream: IO[str]):
+        value = int(str(op.attributes.get("value")).split()[0])
+        cons_type = op.result.type
+        if op.result.type.name == "index":
```
AntonLydike, line 174: I think we'd rather check the type than the name of the type (maybe using `isinstance`?)

AntonLydike (approved):
I got a million small little nitpicks, but nothing major!

xdsl/interpreters/experimental/wgsl_printer.py
PapyChacal: So this is all meant to be used as an intermediate representation to just throw at a WGSL compiler.
First goal is to just compile one function down. If we face issues with mutiple functions, we can deal with it later!
This will basically heavily depends on how it is all hooked in the rest of the API, and if the WGPU python package we use even allows to use those names as handles in the end.
So, short answer: we'll think about this later :slightly_smiling_face:

xdsl/interpreters/experimental/wgsl_printer.py
PapyChacal: It is a hazard at this point, but we'll improve that later with proper test cases! The first use plan is to just parse-and-translate, and the parser handles that case either way.
Just trying to get a first part in to improve the different aspects later !
NB: This is clearly one aspect that I would love to codeshare with the internal printer, which already handles this, right?

tests/interpreters/test_wgsl_printer.py
```diff
@@ -0,0 +1,364 @@
+from io import StringIO
+
+from xdsl.dialects import arith, builtin, gpu, memref, test
+from xdsl.dialects.builtin import IndexType, IntegerAttr, IntegerType, i32
+from xdsl.interpreters.experimental.wgsl_printer import WGSLPrinter
+from xdsl.ir.core import MLContext
+from xdsl.parser.core import Parser
+from xdsl.utils.test_value import TestSSAValue
+
+lhs_op = test.TestOp(operands=[[]], result_types=[IndexType()], regions=[[]])
```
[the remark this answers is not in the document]
PapyChacal: What do you mean? Tests pass :yum:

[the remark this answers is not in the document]
math-fehr: Yeah, I have to say I don't know how, but I'm fine with it ;)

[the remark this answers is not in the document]
PapyChacal: Yeah, honestly those tests are first draft placeholders more than anything else. I want to rush a bit toward integrating this, and think about the proper way to test the whole thing once the overall system is clearer!

PR #1279, opened 2023-07-14
dialects: (linalg) Add linalg methods to determine loop range

Groverkss:


xdsl/ir/affine/affine_map.py
```diff
@@ -47,6 +47,30 @@ def compose(self, map: AffineMap) -> AffineMap:
             results=results,
         )
 
+    def inverse_permutation(self) -> AffineMap | None:
```
math-fehr, line 50: Can you add a comment that this will not work when the AffineMap has symbols, and the cases for which this returns None?

xdsl/dialects/linalg.py
```diff
@@ -103,6 +105,45 @@ def __init__(
...
+        # shapes so this is not possible.
+        syms = 0
+
+        return AffineMap(dims, syms, result_exprs)
+
+    def get_shapes_to_loops_map(self) -> AffineMap:
+        loops_to_shapes = self.get_loops_to_shapes_map()
+        inverse = loops_to_shapes.inverse_permutation()
+        if not inverse:
+            raise NotImplementedError("Non-invertible map")
```
math-fehr, line 128: Is this actually not implemented, or is it just impossible to get it if the permutation is not invertible?

math-fehr (approved):
Just had two questions relative to documentation, but otherwise that looks good to me!

xdsl/dialects/linalg.py
Groverkss: If the map is not invertible, Linalg is supposed to use dynamic shapes for it. Since I do not have any support for dynamic shapes in Linalg, this is not implemented yet.

I don't think there will be support for this anytime soon, since I plan to focus on static shapes only.

xdsl/ir/affine/affine_map.py
Groverkss: Sure.

xdsl/dialects/linalg.py
webmiche: Maybe give a more descriptive error then?
```suggestion
            raise NotImplementedError("Non-invertible maps need dynamic shapes, whicha re not yet implemented.")
```

xdsl/dialects/linalg.py
```diff
@@ -103,6 +105,45 @@ def __init__(
...
+
+    def get_loops_to_shapes_map(self) -> AffineMap:
+        result_exprs: list[AffineExpr] = []
+        for map_attr in self.indexing_maps:
+            map = map_attr.data
+            result_exprs.extend(map.results)
+
+        dims = self.get_num_loops()
+        # FIXME: Check for symbols also. Currently, we do not allow dynamic
+        # shapes so this is not possible.
```
webmiche, line 119: Should this be an assertion?

webmiche (approved):
LGTM 👍

xdsl/dialects/linalg.py
```diff
@@ -103,6 +105,45 @@ def __init__(
             regions=[body],
         )
 
+    def get_num_loops(self) -> int:
+        return self.indexing_maps.data[0].data.num_dims
+
+    def get_loops_to_shapes_map(self) -> AffineMap:
+        result_exprs: list[AffineExpr] = []
```
superlopuh, line 112:
```suggestion
        result_exprs = [res for map in self.indexing_maps for res in map.data.results]
```

superlopuh (approved):
LGTM modulo some list comprehensions that IMO make code a little more readable, and that have better perf in Python than for loops and maps

xdsl/dialects/linalg.py
```diff
@@ -103,6 +105,45 @@ def __init__(
...
+        sizes: list[int] = []
+        for input in self.inputs:
+            if isinstance(input.type, ShapedType):
+                for shape in input.type.get_shape():
+                    sizes.append(shape)
+        for output in self.outputs:
+            if isinstance(output.type, ShapedType):
+                for shape in output.type.get_shape():
+                    sizes.append(shape)
+        return sizes
```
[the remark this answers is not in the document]
Groverkss: I personally find this less readable than what I have. I'll change it to a list comprehension for better performance, but simple clear loops are more readable to me than complicated list comprehension.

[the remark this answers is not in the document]
Groverkss: Also, this list comprehension does not work.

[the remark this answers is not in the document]
Groverkss: Sorry, I don't think I prefer the list comprehension. It's very unreadable for me.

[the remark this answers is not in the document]
superlopuh: I mixed up the order of the fors, but I still think it's better. I don't feel very strongly about it

xdsl/dialects/linalg.py
```diff
@@ -103,6 +105,52 @@ def __init__(
...
+            raise NotImplementedError(
+                "Non-invertible maps need dynamic shapes, which are not implemented."
+            )
+        return inverse
+
+    def get_static_shapes(self) -> list[int]:
+        sizes: list[int] = []
+        for input in self.inputs:
+            if isinstance(input.type, ShapedType):
+                for shape in input.type.get_shape():
```
superlopuh, line 142:
```suggestion
                for dim in input.type.get_shape():
```
Aren't elements of shapes dimensions?

PR #1281, opened 2023-07-15
dialects: (riscv) remove Register class

superlopuh:
When I initially wrote the register types, the reasoning in my mind was that I wanted an optional register name, and `Data` cannot accept an optional generic parameter. But actually an empty string seems more appropriate than None anyway. Now that we have floats in riscv, and some duplication, it feels like a good idea to simplify. This also includes a rename of RegisterType to RegisterAttr, since we don't exclusively use them as types.

After some discussions with @kingiler  on the changes in #1269, I thought I'd play around with the code to see how to make the generics work for both Register/FloatRegister and RegisterType/FloatRegisterType. There were always some issues with it locally, and quite a lot of complexity due to the extra step of indirection. This is what led me to the changes in this PR, which I propose as an alternative to #1269. It's probably a good idea for reviewers to take a look at both to have a better view of multiple potential ways forward.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -386,7 +386,7 @@ def _assembly_arg_str(arg: AssemblyInstructionArg) -> str:
         elif isinstance(arg.type, FloatRegisterAttr):
             reg = arg.type.register_name
             return reg
-    assert False
+    assert False, f"{arg}"
 
 
 def _assembly_line(
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -4,6 +4,8 @@ from abc import ABC, abstractmethod
 from io import StringIO
 from typing import IO, Sequence, TypeAlias
 
+from typing_extensions import Self
+
 from xdsl.dialects.builtin import (
     AnyIntegerAttr,
     IntegerAttr,
@@ -59,6 +61,11 @@ class RISCVRegisterAttr(Data[str], TypeAttribute, ABC):
         if not self.is_allocated:
             raise ValueError("Cannot get name for unallocated register")
         return self.data
+
+    @classmethod
+    @property
+    def unallocated(cls) -> Self:
+        return cls("")
 
     @property
     def is_allocated(self) -> bool:
@@ -386,6 +393,8 @@ def _assembly_arg_str(arg: AssemblyInstructionArg) -> str:
         elif isinstance(arg.type, FloatRegisterAttr):
             reg = arg.type.register_name
             return reg
+        else:
+            assert False, f"{arg.type}"
     assert False, f"{arg}"
 
 
@@ -443,7 +452,7 @@ class RdRsRsIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
         comment: str | StringAttr | None = None,
     ):
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
@@ -482,7 +491,7 @@ class RdImmIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
         elif isinstance(immediate, str):
             immediate = LabelAttr(immediate)
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
@@ -568,7 +577,7 @@ class RdRsImmIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
             immediate = LabelAttr(immediate)
 
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
@@ -684,7 +693,7 @@ class RdRsIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
         comment: str | StringAttr | None = None,
     ):
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
@@ -844,7 +853,7 @@ class CsrReadWriteOperation(IRDLOperation, RISCVInstruction, ABC):
         comment: str | StringAttr | None = None,
     ):
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
@@ -902,7 +911,7 @@ class CsrBitwiseOperation(IRDLOperation, RISCVInstruction, ABC):
         comment: str | StringAttr | None = None,
     ):
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
@@ -958,7 +967,7 @@ class CsrReadWriteImmOperation(IRDLOperation, RISCVInstruction, ABC):
         comment: str | StringAttr | None = None,
     ):
         if rd is None:
-            rd = RegisterAttr("")
+            rd = RegisterAttr.unallocated
         elif isinstance(rd, str):
             rd = RegisterAttr(rd)
         if isinstance(comment, str):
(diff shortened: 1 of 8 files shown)


xdsl/dialects/riscv.py
```diff
@@ -48,19 +49,66 @@
...
+    A RISC-V register type.
     """
 
-    name: str | None = field(default=None)
-    """The register name. Should be one of `ABI_INDEX_BY_NAME` or `None`"""
+    _unallocated: ClassVar[Self | None] = None
+
+    @classmethod
+    @property
+    def unallocated(cls) -> Self:
```
kingiler, line 61: `@classmethod` no longer wrap `@property` since 3.11.
https://docs.python.org/3/library/functions.html#classmethod

xdsl/dialects/riscv.py
```diff
@@ -498,24 +479,24 @@ class RdImmIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
...
-    rd: OpResult = result_def(RegisterType)
+    rd: OpResult = result_def(RegisterAttr)
     immediate: AnyIntegerAttr | LabelAttr = attr_def(AnyIntegerAttr | LabelAttr)
 
     def __init__(
         self,
         immediate: int | AnyIntegerAttr | str | LabelAttr,
         *,
-        rd: RegisterType | Register | None = None,
+        rd: RegisterAttr | str | None = None,
```
kingiler, line 489: Use of `str` may lost errors that can catch by type checkers in previous implementation. Can you revert it?

xdsl/dialects/riscv.py
```diff
@@ -550,15 +531,15 @@ def __init__(
...
```
kingiler, line 534: Same as here.

xdsl/dialects/riscv.py
```diff
@@ -636,7 +617,7 @@ def __init__(
         rs1: Operation | SSAValue,
         immediate: int | AnyIntegerAttr | str | LabelAttr,
         *,
-        rd: RegisterType | Register | None = None,
+        rd: RegisterAttr | str | None = None,
```
kingiler, line 620: See above

xdsl/dialects/riscv.py
superlopuh: I still don't see the benefit of the type checking here.

superlopuh: If a client wants to give a bogus string then so be it, I don't see the problem. Almost all the time they're expected to use one of the `RegisterAttr` instances in `Registers` or None. I'd rather remove `str` than add back `Register` TBH, but I also don't see it as a serious potential risk.

superlopuh: BTW this is pretty much equivalent to allowing int instead of an IntegerAttr, there can be type mismatches, but we allow it for convenience's sake

xdsl/dialects/riscv.py
```diff
@@ -48,19 +49,66 @@
...
+        assert self.data in type(self).abi_index_by_name()
+
+    @classmethod
+    @abstractmethod
+    def abi_index_by_name(cls) -> dict[str, int]:
+        raise NotImplemented
+
+
+@irdl_attr_definition
+class RegisterAttr(RISCVRegisterAttr):
```
compor, line 102: `InRegisterAttr`?

compor (approved):
I like this better:

- `unallocated` is exposed in the API and the internal representation can be anything
- it's sorter to instantiate
- and more intuitive, since the type *is* an attribute

xdsl/dialects/riscv.py
superlopuh: - [ ] IntRegisterAttr

xdsl/dialects/riscv.py
superlopuh: - [ ] remove property

adutilleul (approved):

xdsl/dialects/riscv.py
```diff
@@ -48,19 +49,66 @@
 from xdsl.utils.hints import isa
 
 
-@dataclass(frozen=True)
-class Register:
+class RISCVRegisterAttr(Data[str], TypeAttribute, ABC):
```
AntonLydike, line 52: I don't really see the point of removing the `Type` suffix, as it is still a type (it inherits from `TypeAttribute`). The argument that it is used as an attribute seems a bit moot to me, as types can be used as attributes anyway...

AntonLydike (approved):
LGTM overall, just not sure the name change from type to attr is necessary?

xdsl/dialects/riscv.py
superlopuh: It's a bit arbitrary but somehow feels right to me... I'll think about this some more to try and formulate a rationale.

superlopuh: Thought about this some more and it's not a good enough reason for a rename, I changed it back.

xdsl/dialects/riscv.py
```diff
@@ -48,19 +49,65 @@
...
+
+    def verify(self) -> None:
+        if not self.is_allocated or self.data.startswith("j"):
+            return
+        assert self.data in type(self).abi_index_by_name()
+
+    @classmethod
+    @abstractmethod
+    def abi_index_by_name(cls) -> dict[str, int]:
+        raise NotImplemented
```
kingiler, line 97: `NotImplementedError`

kingiler (changes requested):
LGTM

xdsl/dialects/riscv.py
kingiler: I don't think so. At least `IntegerAttr` has a verifier.

kingiler (changes requested):
At least `IntergerAttr` has a verifier.

xdsl/dialects/riscv.py
superlopuh: So does this class

xdsl/dialects/riscv.py
```diff
@@ -48,19 +49,67 @@
...
+        name = self.data
+        if not self.is_allocated or name.startswith("j"):
+            return
+        if name not in type(self).abi_index_by_name():
+            raise VerifyException(f"{name} not in RV32I")
+
+    @classmethod
+    @abstractmethod
+    def abi_index_by_name(cls) -> dict[str, int]:
+        raise NotImplementedError
```
kingiler, line 99: `NotImplementedError()`

xdsl/dialects/riscv.py
```diff
@@ -48,19 +49,67 @@
...
+    def unallocated(cls) -> Self:
+        if cls._unallocated is None:
+            cls._unallocated = cls("")
+        return cls._unallocated
+
+    @property
+    def register_name(self) -> str:
+        """Returns name if allocated, raises ValueError if not"""
+        if not self.is_allocated:
+            raise ValueError("Cannot get name for unallocated register")
```
kingiler, line 69: I'm not sure why we need the logic here. Can't we return empty string?

kingiler (changes requested):

xdsl/dialects/riscv.py
superlopuh: data is the empty string, this function guarantees that the register is allocated

tobiasgrosser: I feel that is a valid question. In theory, we could return an empty string for unallocated registers. The current code base does not call `register_name` for unallocated registers, so that codepath would be untested. I can see a valid argument, that we want for now to stay with an assertion as this is the only use case we consider. We can always change this if we feel other use-cases would benefit from register_name defaulting to `""` for unallocated registers. However, I can see this being confusing, so maybe lets stay with the current assertion?

xdsl/dialects/riscv.py
tobiasgrosser: I just went over the codebase and the one location where this matters is `xdsl/transforms/lower_riscv_func.py`:

```diff
            ops.append(
                riscv.MVOp(
                    arg,
-                    rd=riscv.Register(f"a{i}"),
+                    rd=f"a{i}",
                )
            )
```

as well as similar changes in this very same file. Looking into the implementation of Register, I see that Register also did not verify the name of the Register that was passed in its initializer. Hence, the current change seems to just implement what was already there. If there is a more typesafe way to write `lower_riscv_func.py`, we could explore that in the future, but I would suggest to remain permissive for now. In the spirit of reducing feature-creep in our PRs, maybe lets get this one in and then get the custom printing in next?

tobiasgrosser:
@superlopuh @kingiler and all, this looks very nice. I did a round of reviews and hope these help us to move forward with this patch (without sacrificing code quality).

xdsl/dialects/riscv.py
tobiasgrosser: This was addressed by @superlopuh in his latest commit. I mark this as resolved.

superlopuh:
@kingiler I've addressed all the comments, would be good to merge this soon

kingiler:
LGTM

tobiasgrosser:
Thank you all.

tobiasgrosser:
Thank you all.

PR #1172, opened 2023-06-21
dialects: (riscv) label region is multi block

superlopuh:
I put the return back in the middle because a lot more instructions are about to become terminators so it makes sense for them to be in order again

PR #1282, opened 2023-07-15
dialects: (riscv) make LabelOp region multi-block

superlopuh:
This PR adds the ability to have multiple blocks in a LabelOp, and amends the lowering from riscv_func.func to allow for multiple blocks. If we lower from func and cf, then we have no choice but to support multi-block.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -2000,14 +2000,14 @@ class LabelOp(IRDLOperation, RISCVOp):
     name = "riscv.label"
     label: LabelAttr = attr_def(LabelAttr)
     comment: StringAttr | None = opt_attr_def(StringAttr)
-    data: OptRegion = opt_region_def("single_block")
+    data: OptRegion = opt_region_def()
 
     traits = frozenset([NoTerminator()])
 
     def __init__(
         self,
         label: str | LabelAttr,
-        region: OptSingleBlockRegion = None,
+        region: OptRegion = None,
         *,
         comment: str | StringAttr | None = None,
     ):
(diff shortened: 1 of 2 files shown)


xdsl/transforms/lower_riscv_func.py
```diff
@@ -64,14 +64,14 @@ def match_and_rewrite(self, op: riscv_func.SyscallOp, rewriter: PatternRewriter)
 class LowerRISCVFuncOp(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: riscv_func.FuncOp, rewriter: PatternRewriter):
-        body = op.func_body.block
-        first_op = body.first_op
+        first_block = op.func_body.blocks[0]
+        first_op = first_block.first_op
         assert first_op is not None
```
compor, line 69: Can we handle this assert a bit more gracefully?

xdsl/dialects/riscv.py
```diff
@@ -2000,14 +2000,14 @@ def my_add():
     name = "riscv.label"
     label: LabelAttr = attr_def(LabelAttr)
     comment: StringAttr | None = opt_attr_def(StringAttr)
-    data: OptRegion = opt_region_def("single_block")
+    data: OptRegion = opt_region_def()
```
compor, line 2003: Augment the tests in `test_riscv.py`?

compor (changes requested):

xdsl/dialects/riscv.py
```diff
@@ -2000,14 +2000,14 @@ def my_add():
...
 
     traits = frozenset([NoTerminator()])
```
compor, line 2005: Does this trait work with multi-block regions? I don't think so, right?
Shouldn't this be removed? Which will then require terminators in all contained blocks?

superlopuh: Ah I didn't realise it wouldn't I removed it in the previous PR but for this one thought it might be ok without it. Maybe I should just reopen the old one, will take a look.

superlopuh:
Closing in favour of #1172

PR #1283, opened 2023-07-15
dialects: (riscv_cf) add riscv control flow instructions with successors

superlopuh:


compor (approved):

adutilleul (approved):

compor:
Although, I approved, I'm currently exploring alternatives for this, i.e., just using a pseudo-terminator instruction and going from cf to riscv directly, following it by the arith to riscv lowering.
So, I'd like to hold off on this for a bit.

superlopuh:
I would rather merge multiple versions that can work together TBH, just like this one can coexist with the branch instructions in the riscv dialect. The rationale for merging this is that it would unblock the rest of the Toy tutorial PRs that I'd love to get over with this week if possible. If there's an alternative that is functionally equivalent and superior in some ways then we can delete this later.

superlopuh:
Importantly, if we happen to have an input with cf, I don't see another way of dealing with an existing successor-based structure. Would we inline all the blocks? I'm not sure this can work in the general case.

compor:
> successor

what do you mean by "inline all the blocks"?

PR #1285, opened 2023-07-17
interpreter: (affine) add interpreter functions

superlopuh:


compor (approved):

xdsl/interpreters/affine.py
```diff
@@ -0,0 +1,74 @@
...
+        assert len(lower_bound) == 1
+        assert len(upper_bound) == 1
+
+        lower_bound = lower_bound[0]
+        upper_bound = upper_bound[0]
+        step = op.step.value.data
+
+        for i in range(lower_bound, upper_bound, step):
+            for_results = interpreter.run_ssacfg_region(op.body, (i,))
+            assert not for_results, "affine block results not supported yet"
```
PapyChacal, line 66: Why not a NotImplementedError? I feel we might want to have a structured way of displaying those errors at some point and it would ease this

tests/interpreters/test_affine_interpreter.py
```diff
@@ -0,0 +1,57 @@
...
+
+        @Builder.implicit_region((index,))
+        def rows_region(args: tuple[Any, ...]):
+            (row,) = args
+
+            @Builder.implicit_region((index,))
+            def cols_region(args: tuple[Any, ...]):
+                (col,) = args
+                sum_op = arith.Addi(row, col)
+                affine.Store(sum_op.result, alloc_op.memref, (row, col))
```
PapyChacal, line 35: Can you add a Load in the exemple?  NB I didn't foloow the memref implementation closely yet, not sure how hard this would be!

PapyChacal (approved):

xdsl/interpreters/affine.py
superlopuh: - [ ] NotImplementedError

tests/interpreters/test_affine_interpreter.py
superlopuh: Ah good point forgot about that.
- [ ] add load

PR #1288, opened 2023-07-17
CI: Add a retry on codecov upload.

PapyChacal:
Had an issue again with codecov, figured we might want to try something like this.
Just an auto-retry, 10 times spaced by 10 seconds here.

### DIFF
--- a/.github/workflows/ci-mlir.yml
+++ b/.github/workflows/ci-mlir.yml
@@ -113,10 +113,14 @@ jobs:
 
     - name: Upload coverage to Codecov
       if: matrix.python-version == '3.10'
-      uses: codecov/codecov-action@v3
+      uses: Wandalen/wretry.action@v1
       with:
-        fail_ci_if_error: true
-        verbose: true
-        directory: ${GITHUB_WORKSPACE}/../
-        files: coverage.xml
-        token: ${{ secrets.CODECOV_TOKEN }}
+        action: codecov/codecov-action@v3
+        attempt_delay: 10000
+        attempt_limit: 10
+        with: |
+          fail_ci_if_error: true
+          verbose: true
+          directory: ${GITHUB_WORKSPACE}/../
+          files: coverage.xml
+          token: ${{ secrets.CODECOV_TOKEN }}


PapyChacal:
Hard to figure out if this helps on a randomly occuring issue :thinking: But at least it seems to work on the happy case

superlopuh (approved):
I like this and I dislike this at the same time :)

webmiche:
Do you know if this actually sovles the issue? So does this even work? Do you have a case where a retry worked?

I am fine with merging this, but it might jsut be useless, as long as we don't understand the issue 😅

math-fehr:
Alternative idea.
Currently, we are using codecov to look online the coverage of our project.
Would there be a way to have this working offline? We compute the coverage using `coverage` anyway, maybe there is an offline tool that we could run in the CI and that would have a similar output?

It's just that we had so many problems with codecov, and we are rarely using it.

Otherwise, I'm fine merging it if you think it's worth it!

PapyChacal:
Michel: no, I'm not sure how to trigger an error, would you know ? We can randomly trigger it and wait to see a good example 😅 otherwise, this CI run shows that it seems to just work when codecov just works, so it doesn't seem to make the happy case worse

About understanding the issue, there's a lot of small different ones that can happen, since it's an online service. I can try and find the issue on the codecov action that discussed this. But ultimately, we can't do without the possibility of it failing. So sticking with an online service, I don't see a better option than wait&retry?

Mathieu: I don't know about other contributors, but I do use it and like it a lot! I would be fine with an alternative, I will see if I find something

PapyChacal:
Here the mentioned issue : https://github.com/codecov/codecov-action/issues/926

webmiche:
> Michel: no, I'm not sure how to trigger an error, would you know ? We can randomly trigger it and wait to see a good example 😅 otherwise, this CI run shows that it seems to just work when codecov just works, so it doesn't seem to make the happy case worse

No, I don't know when it happens. I think I once found an issue explaining that it was essentially them going over their bandwidth, but not sure where I found that.

I guess the line by line view is quite valuable and I do use it for that too sometimes. If we could have an offline version doing that, that would be fine for me as well. Though having a PR comment telling us how coverage changed can be quite useful at times, IMO.

PR #1292, opened 2023-07-18
tools: Split part of xDSLOpt in a CommandLineTool base class.

PapyChacal:
Just splitting what I think will be shared by all tools in a base class from xDSLOptMain.
That is: parsing input with any frontend available and the related arguments, along with the dialects registration and frontends registration.

### DIFF
--- a/xdsl/tools/xdsl_tool.py
+++ b/xdsl/tools/xdsl_tool.py
@@ -1,0 +1,203 @@
+import argparse
+from typing import IO, Callable
+
+from xdsl.backend.riscv.lowering.riscv_arith_lowering import RISCVLowerArith
+from xdsl.dialects.affine import Affine
+from xdsl.dialects.arith import Arith
+from xdsl.dialects.builtin import Builtin, ModuleOp
+from xdsl.dialects.cf import Cf
+from xdsl.dialects.cmath import CMath
+from xdsl.dialects.experimental.dmp import DMP
+from xdsl.dialects.experimental.fir import FIR
+from xdsl.dialects.experimental.math import Math
+from xdsl.dialects.func import Func
+from xdsl.dialects.gpu import GPU
+from xdsl.dialects.irdl.irdl import IRDL
+from xdsl.dialects.linalg import Linalg
+from xdsl.dialects.llvm import LLVM
+from xdsl.dialects.memref import MemRef
+from xdsl.dialects.mpi import MPI
+from xdsl.dialects.pdl import PDL
+from xdsl.dialects.printf import Printf
+from xdsl.dialects.riscv import RISCV
+from xdsl.dialects.riscv_func import RISCV_Func
+from xdsl.dialects.scf import Scf
+from xdsl.dialects.snitch import Snitch
+from xdsl.dialects.snitch_runtime import SnitchRuntime
+from xdsl.dialects.stencil import Stencil
+from xdsl.dialects.test import Test
+from xdsl.dialects.vector import Vector
+from xdsl.frontend.passes.desymref import DesymrefyPass
+from xdsl.frontend.symref import Symref
+from xdsl.ir import Dialect, MLContext
+from xdsl.parser import Parser
+from xdsl.passes import ModulePass
+from xdsl.transforms.dead_code_elimination import DeadCodeElimination
+from xdsl.transforms.experimental.convert_stencil_to_ll_mlir import (
+    ConvertStencilToLLMLIRPass,
+)
+from xdsl.transforms.experimental.dmp.stencil_global_to_local import (
+    GlobalStencilToLocalStencil2DHorizontal,
+    LowerHaloToMPI,
+)
+from xdsl.transforms.experimental.stencil_shape_inference import (
+    StencilShapeInferencePass,
+)
+from xdsl.transforms.experimental.stencil_storage_materialization import (
+    StencilStorageMaterializationPass,
+)
+from xdsl.transforms.lower_mpi import LowerMPIPass
+from xdsl.transforms.lower_riscv_func import LowerRISCVFunc
+from xdsl.transforms.lower_snitch import LowerSnitchPass
+from xdsl.transforms.lower_snitch_runtime import LowerSnitchRuntimePass
+from xdsl.transforms.mlir_opt import MLIROptPass
+from xdsl.transforms.printf_to_llvm import PrintfToLLVM
+from xdsl.transforms.reconcile_unrealized_casts import ReconcileUnrealizedCastsPass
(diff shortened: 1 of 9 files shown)


xdsl/tools/xdsl_tool.py
webmiche, line 1: Maybe `cli_tool` or `cmd_tool`?

I guess xDSL_tool is kind of not saying anything :sweat_smile:

xdsl/tools/xdsl_tool.py
```diff
@@ -0,0 +1,218 @@
...
+    StencilStorageMaterializationPass,
+)
+from xdsl.transforms.lower_mpi import LowerMPIPass
+from xdsl.transforms.lower_riscv_func import LowerRISCVFunc
+from xdsl.transforms.lower_snitch import LowerSnitchPass
+from xdsl.transforms.lower_snitch_runtime import LowerSnitchRuntimePass
+from xdsl.transforms.mlir_opt import MLIROptPass
+from xdsl.transforms.printf_to_llvm import PrintfToLLVM
+from xdsl.transforms.reconcile_unrealized_casts import ReconcileUnrealizedCastsPass
+from xdsl.transforms.riscv_register_allocation import RISCVRegisterAllocation
```
webmiche, line 58: Same here btw.

webmiche (changes requested):
AFAIU, this is essentially taking out all the things needed for any tool that will take textual input IR, parse it and then do things with it. Does this make sense? Are there other tools we could imagine?

IMO, one could imagine having tools that "do the opposite" and split everything out that is needed to properly print IR and then instantiate specific tools with different "frontends". Would there be value to a more general approach?

So fundamentally, I could see a class hierarchy with `cli_tool`(registration of dialects), `textual_parser_tool`, `textual_printer_tool` and `optimizer_tool` or something.

Maybe this is just overengineering at that point though, so I am fine with just merging the curent version (with a different name). This is just the spark some discussion and turn this into a case of PR creep :wink:

xdsl/tools/xdsl_tool.py
webmiche: it seems I have forgotten to add a comment :thinking: 

This is concerning the on lines 7-31 (which I will readd in a moment...)

xdsl/tools/xdsl_tool.py
```diff
@@ -0,0 +1,218 @@
...
+from xdsl.dialects.pdl import PDL
+from xdsl.dialects.printf import Printf
+from xdsl.dialects.riscv import RISCV
+from xdsl.dialects.riscv_func import RISCV_Func
+from xdsl.dialects.scf import Scf
+from xdsl.dialects.snitch import Snitch
+from xdsl.dialects.snitch_runtime import SnitchRuntime
+from xdsl.dialects.stencil import Stencil
+from xdsl.dialects.test import Test
+from xdsl.dialects.vector import Vector
```
webmiche, line 31: Could we import this like `from xdsl.dialects import affine, arith, ...` and then have `affine.Affine` below? That would be more consistent with the way we want dialect import to look like in the repo.

(I know this is irrelevant for this PR, but since we touch the code anyhow....)

math-fehr: How about we have something like `from xdsl.dialects import Arith, Builtin, ...` ? (In another PR though)
Otherwise, I think this current version makes more sense. `affine.Affine` seems like you have an `Affine` operation in the `affine` dialect.

webmiche: Yeah, that'd be even better :+1:

xdsl/tools/xdsl_tool.py
```diff
@@ -0,0 +1,218 @@
...
+        PrintfToLLVM,
+        RISCVRegisterAllocation,
+        RISCVLowerArith,
+        StencilShapeInferencePass,
+        StencilStorageMaterializationPass,
+        ReconcileUnrealizedCastsPass,
+    ]
+
+
+class xDSLTool:
```
math-fehr, line 116: Is there a reason we don't include `split-input-file` to the list of arguments here?
Do you think it should stay in xdsl-opt only?

xdsl/tools/xdsl_tool.py
superlopuh: yeah that's a good point I vote for CommandLineTool

math-fehr (approved):
Looks good to me!

superlopuh:
I like the suggestion of splitting the functionality of reading/printing into separate classes but it definitely feels like PR scope creep here. My understanding is that this is for the `xdsl-runner` tool which currently needs all the functionality split out in this PR. If there are other tools that don't need it in the future we could split them further. Or not :) But it can be a discussion for later, IMO.

math-fehr:
> AFAIU, this is essentially taking out all the things needed for any tool that will take textual input IR, parse it and then do things with it. Does this make sense? Are there other tools we could imagine?
> 
> IMO, one could imagine having tools that "do the opposite" and split everything out that is needed to properly print IR and then instantiate specific tools with different "frontends". Would there be value to a more general approach?

I had a quite long discussion with Emilien offline, because I had the same thought!
I believe for instance that long term, it would be better to have a single tool, where we can more easily parametrize what we want (i.e., you want a frontend but not a target, or you just want passes, etc...).
However, we came to the conclusion that this is probably overengineering things right now, and would just cost us a lot of time just to get things right for an unknown benefits. We think that it's probably better to keep this design right now, until we feel that this is not good anymore, and that we should move to a unified solution.
Also, that doesn't help that this is blocking the xdsl-run patch ;)

xdsl/tools/xdsl_tool.py
PapyChacal: Will rename those, I'm assuming you and @math-fehr came to an agreement to keep the dialect ones as is for now! Tell me if I misinterpreted :slightly_smiling_face:

xdsl/tools/xdsl_tool.py
PapyChacal: Will do this renaming, and I point to @math-fehr's comment below for the rest.
It's one of those PRs where I'm trying to do one step in the right direction, but I have other immediate objectives.
I find the proper splitting discussion interesting though! But I clearly don't want to solve it all before moving forwrd with the tool I have in mind now.

PapyChacal:
Opened https://xdsl.zulipchat.com/#narrow/stream/368604-Framework/topic/Splitting.20tools.20funtionnality for further discussion on the longer-term!

webmiche (approved):
Okay, letŝ ship it and address stuff once it comes up.

In terms of the name, I am fine with the new one, but I think I would prefer something along the lines of `ir_parsing_tool` or something, in particular with the modularity discussion in mind.

compor (approved):

PR #1293, opened 2023-07-18
dialect: (riscv) add support for trivial custom IR format

kingiler:
This change introduces trivial custom printing to the `riscv` dialect to bring in the core code changes that enable custom printing for RISCV. All code changes are introduced by expanding the `RVOp` operation, such that any subclass can in the future evolve its format by overloading the methods for custom parsing and printing.

The IR changes itself are trivial and only drop the `"` around the instruction names:

```diff
- %addi = riscv.addi(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+ %addi = riscv.addi(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
```

We also switch the format of the `riscv_ops.mlir` file to custom IR.

Future changes will bring the format closer to assembly code, e.g., by dropping the `()` around the operand list or by providing shorter syntax for immediates.

### DIFF
--- a/tests/filecheck/dialects/riscv/riscv_ops.mlir
+++ b/tests/filecheck/dialects/riscv/riscv_ops.mlir
@@ -1,269 +1,269 @@
-// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | filecheck %s
+// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | xdsl-opt | filecheck %s
 "builtin.module"() ({
-  "riscv.label"() ({
-    %0 = "riscv.get_register"() : () -> !riscv.reg<>
-    %1 = "riscv.get_register"() : () -> !riscv.reg<>
+  riscv.label() ({
+    %0 = riscv.get_register() : () -> !riscv.reg<>
+    %1 = riscv.get_register() : () -> !riscv.reg<>
     // RV32I/RV64I: 2.4 Integer Computational Instructions
 
     // Integer Register-Immediate Instructions
-    %addi = "riscv.addi"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK: %{{.*}} = "riscv.addi"(%{{.*}}) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %slti = "riscv.slti"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.slti"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %sltiu = "riscv.sltiu"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.sltiu"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %andi = "riscv.andi"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.andi"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %ori = "riscv.ori"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.ori"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %xori = "riscv.xori"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.xori"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %slli = "riscv.slli"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.slli"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %srli = "riscv.srli"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.srli"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %srai = "riscv.srai"(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.srai"(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %lui = "riscv.lui"() {"immediate" = 1 : i32}: () -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.lui"() {"immediate" = 1 : i32} : () -> !riscv.reg<>
-    %auipc = "riscv.auipc"() {"immediate" = 1 : i32}: () -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.auipc"() {"immediate" = 1 : i32} : () -> !riscv.reg<>
-    %mv = "riscv.mv"(%0) : (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK: %{{.*}} = "riscv.mv"(%{{.*}}) : (!riscv.reg<>) -> !riscv.reg<>
+    %addi = riscv.addi(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK: %{{.*}} = riscv.addi(%{{.*}}) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %slti = riscv.slti(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.slti(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %sltiu = riscv.sltiu(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.sltiu(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %andi = riscv.andi(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.andi(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %ori = riscv.ori(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.ori(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %xori = riscv.xori(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.xori(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %slli = riscv.slli(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.slli(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %srli = riscv.srli(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.srli(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %srai = riscv.srai(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.srai(%0) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
+    %lui = riscv.lui() {"immediate" = 1 : i32}: () -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.lui() {"immediate" = 1 : i32} : () -> !riscv.reg<>
+    %auipc = riscv.auipc() {"immediate" = 1 : i32}: () -> !riscv.reg<>
+    // CHECK-NEXT: %{{.*}} = riscv.auipc() {"immediate" = 1 : i32} : () -> !riscv.reg<>
+    %mv = riscv.mv(%0) : (!riscv.reg<>) -> !riscv.reg<>
+    // CHECK: %{{.*}} = riscv.mv(%{{.*}}) : (!riscv.reg<>) -> !riscv.reg<>
 
     // Integer Register-Register Operations
-    %add = "riscv.add"(%0, %1) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.add"(%{{.*}}, %{{.*}}) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    %slt = "riscv.slt"(%0, %1) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.slt"(%{{.*}}, %{{.*}}) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    %sltu = "riscv.sltu"(%0, %1) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.sltu"(%{{.*}}, %{{.*}}) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    %and = "riscv.and"(%0, %1) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    // CHECK-NEXT: %{{.*}} = "riscv.and"(%{{.*}}, %{{.*}}) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
-    %or = "riscv.or"(%0, %1) : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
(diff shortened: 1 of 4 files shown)


xdsl/dialects/riscv.py
```diff
@@ -331,6 +333,106 @@ class RISCVOp(Operation, ABC):
...
+    @classmethod
+    def parse_operands(cls, parser: Parser) -> Sequence[SSAValue]:
+        operands = parser.parse_comma_separated_list(
+            parser.Delimiter.PAREN, parser.parse_operand
+        )
+        return operands
+
+    @classmethod
+    def parse_successors(cls, parser: Parser) -> Sequence[Block]:
+        return parser.parse_optional_successors() or []
```
superlopuh, line 355: supernit:
```suggestion
        return parser.parse_optional_successors() or ()
```
tuples by default for Sequences seems like a good idea, but it's a minor code style thing

tests/filecheck/dialects/riscv/riscv_ops.mlir
```diff
@@ -1,267 +1,267 @@
-// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | filecheck %s
+// RUN: xdsl-opt --print-op-generic %s | xdsl-opt | filecheck %s
```
superlopuh, line 1: Why the print op generic? I think the previous version tested that the custom parser works. It feels like we might want to remove the generic printing entirely from this line.

superlopuh (approved):
Thank you, seems like a good first step.

tests/filecheck/dialects/riscv/riscv_ops.mlir
superlopuh: The code coverage should reveal this, or a `print("foo")` in the custom parsing, maybe I'm wrong.

AntonLydike:
I would prefer to add these as needed, so we can see how they are used. I feel that most of these are either already implemented in xDSL in the parser/printer classes, or should live there instead of on the RISCV operation class.

superlopuh:
Yeah that's a good point

AntonLydike:
It feels a bit like this is solving the issue from the wrong end? I'd rather add a custom parsing/printing to a single class, like `NullaryOperation`, instead of going the most generic way possible and making `RISCVOp` print custom format...

Then we can add one to `RdRsImmIntegerOperation`, and so on. Once we find some common operation, we can factor them out, etc.

tests/filecheck/dialects/riscv/riscv_ops.mlir
tobiasgrosser: The idea was that we test both generic and custom printing, generic printing via the printing and reparsing around the pipe and the custom parsing and printing by parsing and filechecking this file. What command line would you advice @superlopuh?

> The code coverage should reveal this, or a print("foo") in the custom parsing, maybe I'm wrong.

I do not understand this sentence. Could you potentially elaborate?

superlopuh: Ah you're right. I'm not sure we need to test that we print and parse custom syntax, though, since that's a well-tested part of the infrastructure already. What's the purpose of the second xdsl-opt? Maybe I'm misunderstanding.

superlopuh (approved):

tests/filecheck/dialects/riscv/riscv_ops.mlir
tobiasgrosser: We test four things:

- Parsing of custom syntax (via the first xdsl-opt command)
- Printing of generic syntax for this dialect (via the first xdsl-opt command)
- Parsing of generic syntax for this dialect (via the 2nd xdsl-opt command)
- Printing of custom syntax (via the 2nd xdsl-opt command) + filecheck

We can claim that generic parsing is already well tested, but given that we get it at the cost of just a 2nd xdsl-opt call, I feel having it tested per-dialect does not have a large cost.

superlopuh: Ok sounds good

xdsl/dialects/riscv.py
```diff
@@ -331,6 +333,106 @@ class RISCVOp(Operation, ABC):
...
+        if parser.parse_optional_punctuation("(") is None:
+            returns = [parser.parse_type()]
+        else:
+            returns = list[Attribute]()
+            if parser.parse_optional_punctuation(")") is None:
+                returns.append(parser.parse_type())
+                while parser.parse_optional_punctuation(",") is not None:
+                    returns.append(parser.parse_type())
+                parser.parse_punctuation(")")
+        return args, returns
```
tobiasgrosser, line 376: I am curious why this code is not available already. AFAIU types should already be parsed by the generic parser, no? @kingiler, can you explain why the generic code in the parser won't work?

kingiler: Because it is in a private method `Parser.parse_function_type` and `pyright` is not happy about accessing it so I just copy and paste for now. Making a PR to make it a public function.

kingiler: Use updated parser function in newest commit.

xdsl/dialects/riscv.py
AntonLydike: I think one rule we like to follow would be to use tuples unless mutability is required, instead of defaulting to lists. Tuples are hashable by default, and have some other nice benefits over lists.

superlopuh: In this case it's not all that important but they're also covariant due to the immutability

xdsl/dialects/riscv.py
```diff
@@ -331,6 +334,73 @@ class RISCVOp(Operation, ABC):
     def assembly_line(self) -> str | None:
         raise NotImplementedError()
 
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        operands = cls.parse_operands(parser)
+        successors = cls.parse_successors(parser)
+        regions = cls.parse_regions(parser)
```
AntonLydike, line 341: I would remove successors and regions until they are needed. Only a very small subset of RISC-V Operations actually require them.

xdsl/dialects/riscv.py
```diff
@@ -331,6 +334,73 @@ class RISCVOp(Operation, ABC):
...
+        return cls.create(
+            operands=operands,
+            result_types=result_types,
+            attributes=attributes,
+            successors=successors,
+            regions=regions,
+        )
+
+    @classmethod
+    def parse_operands(cls, parser: Parser) -> Sequence[SSAValue]:
```
AntonLydike, line 354: All these methods should either be private, or moved to a different place. I would like to be very careful with the public facing API of *all* the riscv operations...

xdsl/dialects/riscv.py
```diff
@@ -331,6 +334,73 @@ class RISCVOp(Operation, ABC):
...
+        printer.print_regions(self.regions)
+        self.print_attributes(printer)
+        printer.print(" : ")
+        self.print_types(printer)
+
+    def print_operands(self, printer: Printer) -> bool:
+        printer.print("(")
+        if self.operands:
+            printer.print_list(self.operands, printer.print_ssa_value)
+        printer.print(")")
```
AntonLydike, line 393: Let's just drop the brackets right away? No need to re-implement them in the custom syntax, no? And we are changing all the operations anyway.

xdsl/dialects/riscv.py
```diff
@@ -331,6 +334,73 @@ class RISCVOp(Operation, ABC):
...
+    def parse_successors(cls, parser: Parser) -> Sequence[Block]:
+        return parser.parse_optional_successors() or []
+
+    @classmethod
+    def parse_function_type(cls, parser: Parser) -> FunctionType:
+        return parser.parse_function_type()
+
+    @classmethod
+    def parse_attributes(cls, parser: Parser) -> Mapping[str, Attribute]:
+        return parser.parse_optional_attr_dict()
```
AntonLydike, line 369: I would remove these two and replace them with calls directly to the method they wrap. I don't see a point in having them otherwise.

xdsl/dialects/riscv.py
```diff
@@ -331,6 +334,73 @@ class RISCVOp(Operation, ABC):
...
+        printer.print(")")
+        return bool(self.operands)
+
+    def print_attributes(self, printer: Printer) -> Set[str]:
+        if self.attributes:
+            printer.print_op_attributes(self.attributes)
+        return self.attributes.keys()
+
+    def print_types(self, printer: Printer) -> None:
+        printer.print_operation_type(self)
```
AntonLydike, line 402: This one should also just be replaced with a call to the printer method as far as I can tell.

xdsl/dialects/riscv.py
```diff
@@ -331,6 +334,73 @@ class RISCVOp(Operation, ABC):
...
+
+    @classmethod
+    def parse_regions(cls, parser: Parser) -> Sequence[Region]:
+        if parser.parse_optional_punctuation("(") is not None:
+            regions = parser.parse_comma_separated_list(
+                parser.Delimiter.NONE, parser.parse_region
+            )
+            parser.parse_punctuation(")")
+            return regions
+        return []
```
AntonLydike, line 379: Similarly unused

AntonLydike (changes requested):
While I am still not sure this is the right direction to go in with custom syntax on RISC-V operations, I wanted to separate the concerns regarding the code from the more general ones.

If we want to go into this direction, I would like to see:

 - All methods be private that can be private
 - Good documentation, this means both on the PR and on the methods themselves. Docstrings to explain what you are doing and the design decisions you took. Also code comments on the more complex sections

I think we are doing some exciting work here, and we are starting to realize that our custom parsing/printing architecture is still lacking. I'd rather see the parser/printer extended to facilitate easier custom syntax than a whole new layer of abstraction on top of it that other dialects can't use themselves.

xdsl/dialects/riscv.py
tobiasgrosser: I agree.

xdsl/dialects/riscv.py
tobiasgrosser: I agree.

xdsl/dialects/riscv.py
tobiasgrosser: In this case, I see a merit to leave such changes for later. One can argue its easy, but there is a risk of getting into stylistic discussions which are typically faster if factored out. I feel exactly these "we can just sneak it in" habit leads to our PRs sometimes growing quite a bit.

tobiasgrosser: In particular, we then need to think about how instructions with immediates are handled, ... I feel keeping this discussion out of this PR will help us to move faster.

xdsl/dialects/riscv.py
tobiasgrosser: If that is possible that would be great. Did you test if this is indeed dead code?

xdsl/dialects/riscv.py
tobiasgrosser: Just to double check, did you test this?

AntonLydike: Ah, I think you are right, it's not unused! Good catch!

tobiasgrosser: This code should already exist in the parser, no?

xdsl/dialects/riscv.py
AntonLydike: Imo, removing the quotes from all operations, or removing quotes and brackets from all operations is roughly equivalent in terms of discussion required... But I can see the reason, let's not get hung up on this part.

tobiasgrosser: One is clearly minimal, the other is not. Also, I would actually not be too certain about this. Talking about operators, I see us already considering how to handle immediates. ... In any case, thank you for allowing us to focus first.

tobiasgrosser:
> I guess this discussion with "do we need to worry about regions and successors and such" is a good indicator that we are putting the cart before the horse so to speak. We are going up the steepest possible side of the hill that is custom syntax. I don't think we can do first-shot custom syntax for all possible (riscv) operations. That's just designing a new general-purpose custom format (like the xDSL syntax we had a while back).
> 
> If we find ourselves writing generic parser/printer code again, I think we are moving in the wrong direction.

I guess I disagree here. I feel this mostly point out that not enough of the public API is exposed yet. I feel this PR could actually be super productive. In an optimal world, it would be super tiny only using public APIs that allow us to reproduce the standard syntax, with a minimal change that drops the `""`. In fact, that's the entire point of this PR, helping us to expose the API that we need to reproduce the generic parsing as is.

xdsl/dialects/riscv.py
kingiler: I'm not sure if someone write a V-extension, not included in xDSL, they will need to use these APIs.

xdsl/dialects/riscv.py
kingiler: This is left for subsequent PR which will / may print custom function type: `(zero, t0) -> (t1)` instead of `(!riscv.reg<zero>, !riscv.reg<t0>) -> (!riscv.reg<t1>)`.

kingiler: Similarly for attributes. Current PR tries to minimize the diff in `tests/filecheck`.

xdsl/dialects/riscv.py
kingiler: Commented above.

xdsl/dialects/riscv.py
kingiler: There is currently no `parse_optional_regions` (there is `parse_region` though).

xdsl/dialects/riscv.py
tobiasgrosser: We may need these later, but for now I would just call `parser.parse_optional_attr_dict()` directly and do not yet introduce these `parse_attributes()` functions. They are not yet needed, and introducing abstractions without use cases is typically difficult to review, if the point of this PR is to not yet open too many discussions at once.

xdsl/dialects/riscv.py
tobiasgrosser: This one will be available promptly. We can then again drop the `parse_regions` function.

tobiasgrosser: There is no parse_region_list, which you can use here.

xdsl/dialects/riscv.py
tobiasgrosser: > I would remove these two and replace them with calls directly to the method they wrap. I don't see a point in having them otherwise.

Can you follow @AntonLydike's advice here? It is hard to establish abstractions without justifications, and the point of this PR is to limit the scope of discussions to a minimum. So we do not want to pull in other questions to provide evidence why these abstractions are needed. I feel we can introduce them iff they are needed (and I am not even sure we will need them).

xdsl/dialects/riscv.py
tobiasgrosser: I would not optimize for this case. If they need their own parsing/printing, they can just overwrite the parse/print functions as a whole. Supporting too generic use cases again unnecessarily broadens the scope of our discussion.

kingiler: Removed the function, will be exposed in next PR.

tobiasgrosser (approved):

xdsl/dialects/riscv.py
kingiler: Removed the function, will be exposed in next PR.

xdsl/dialects/riscv.py
kingiler: Now using the `parser.parse_region_list`. Mark as resolved.

xdsl/dialects/riscv.py
```diff
@@ -331,6 +331,32 @@ class RISCVOp(Operation, ABC):
...
+            result_types=result_types,
+            attributes=attributes,
+            regions=regions,
+        )
+
+    def print(self, printer: Printer) -> None:
+        printer.print("(")
+        printer.print_list(self.operands, printer.print_ssa_value)
+        printer.print(")")
+        printer.print_successors(self.successors)
```
tobiasgrosser, line 354: We can drop this call to print_successors, no?

tests/filecheck/dialects/riscv/riscv_ops.mlir
kingiler: Upon thinking, I'm not sure the first the `Parsing of custom syntax (via the first xdsl-opt command)` is actually run.

May suggest `xdsl-opt | xdsl-opt --print-op-generic %s | xdsl-opt | filecheck %s`

kingiler: Because the test file may in unknown format yet, (generic format in this case).

kingiler:
Hopefully now the PR is "minimal".

xdsl/dialects/riscv.py
```diff
@@ -331,6 +331,31 @@ class RISCVOp(Operation, ABC):
     def assembly_line(self) -> str | None:
         raise NotImplementedError()
 
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        operands = parser.parse_comma_separated_list(
+            Parser.Delimiter.PAREN, parser.parse_operand
+        )
```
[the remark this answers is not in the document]
tobiasgrosser: Maybe we could expose these in a separate PR, as we did with the other parse changes?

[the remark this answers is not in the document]
kingiler: Using `Parser.parse_op_args_list` and `Printer.print_operands` in newest commit. Mark as resolved.

tobiasgrosser:
We can probably use https://github.com/xdslproject/xdsl/pull/1302 to further simplify this code.

Also, given that the code is now small and easy to read, we can pick up @AntonLydike's suggestion to already drop the `()` around the parameters. AFAIU using the following instead of `parse_op_arg_list` should do the trick:
```python
self.parse_comma_separated_list(None, self.parse_unresolved_operand, " in operation argument list")
```

tests/filecheck/dialects/riscv/riscv_ops.mlir
tobiasgrosser: Indeed, we would need to change the input file to make sure that our custom parser is also tested. I think that would be a good idea, as this will also make the file more readable.

superlopuh: But we know that this file is input in generic format? If you wanted to make this purpose explicit you could rename it to `riscv_ops_generic_format.mlir` or something like that? I don't see the appeal of piping it through a bunch of times, what that extra time in testing for every commit for eternity gains in terms of expected bug minimisation.

kingiler:
Uses #1302 in newest commit. Will drop brackets around operands in next PR.

superlopuh:
I think this is good to merge. I'm still not sure what the benefit is of multiple xdsl-opts but don't feel very strongly about it.

superlopuh:
Thank you for making the suggested changes

superlopuh:
Yes I mean I still don't see the benefit of testing generic syntax in this unit test that's RISC-V dialect specific. This is not a blocker, I just thought I'd mention it.

kingiler:
If we only need to test custom syntax, we can use `xdsl-opt %s | xdsl | filecheck %s`. This will print and parse custom syntax.

kingiler:
I have no strong opinions on it.

kingiler:
Ready to get merged?

tobiasgrosser:
> Ready to get merged?

Let's wait until I and Anton had time to look into this. There is no need to rush this in.

tests/filecheck/dialects/riscv/riscv_ops.mlir
tobiasgrosser: I would rather suggest to turn the input into custom format.

AntonLydike (commented):
I still don't see the point in this, I guess I just give up at this point. :shrug:

tests/filecheck/dialects/riscv/riscv_ops.mlir
superlopuh: Sorry this was just a brain fart by me, I meant custom format...

kingiler: Input file is now in custom format.

tobiasgrosser:
I guess in that case, let's merge this in for now. We should probably meet sometime next week in person to talk about this. In that case, we can then always introduce a different piece of code if that is more sensible. Also, I personally feel these approaches will anyhow converge. Let's see.

PR #1296, opened 2023-07-18
dialects: (riscv) Add custom printing to NullaryOperation

AntonLydike:
A first step for custom printing, adding nullary op custom format.

Authored by @kingiler, @AntonLydike

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -42,7 +42,7 @@ from xdsl.irdl import (
     var_operand_def,
     var_result_def,
 )
-from xdsl.parser import AttrParser
+from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
 from xdsl.traits import IsTerminator, NoTerminator
 from xdsl.utils.exceptions import VerifyException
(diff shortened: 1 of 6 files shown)


superlopuh (approved):
perfect, thank you

xdsl/dialects/riscv.py
```diff
@@ -843,6 +843,16 @@ def __init__(
...
 
+    def print(self, printer: Printer):
+        if self.attributes:
+            printer.print(" ")
+            printer.print_op_attributes(self.attributes)
+
+    @classmethod
+    def parse(cls: type[Self], parser: Parser) -> Self:
+        attrs = parser.parse_optional_attr_dict()
+        return cls.create(attributes=attrs)
```
superlopuh, line 854: Can you please change this to init, and then set the attributes one by one?

compor (approved):

superlopuh:
Could you please update or close this PR? I still like the idea of using the typed init, but there's a bit more work to be done to allow for dynamic regions, if we want to support those in the same way that the `main` branch does today.

PR #1300, opened 2023-07-19
core: (parser) factor out parse_regions

tobiasgrosser:


georgebisbas (approved):

xdsl/parser/core.py
```diff
@@ -634,6 +634,14 @@ def parse_region(self, arguments: Iterable[Argument] | None = None) -> Region:
             self.raise_error("Expected region!")
         return region
 
+    def parse_regions(self) -> list[Region]:
```
math-fehr, line 637: Can you rename `parse_regions` to `parse_region_list`? To make it consistent with MLIR.
Also, can you add documentation like above on the syntax you are parsing? So it is clear for users that you are parsing `(` and `)` as well if present, and that you are returning `[]` if nothing was parsed.

math-fehr (approved):
Just added a comment on the name and the documentation, but otherwise it's good to me!

xdsl/parser/core.py
kingiler: I have updated the name and the doc. I'm not sure the parse format is correct or not?

xdsl/parser/core.py
```diff
@@ -634,6 +634,19 @@ def parse_region(self, arguments: Iterable[Argument] | None = None) -> Region:
             self.raise_error("Expected region!")
         return region
 
+    def parse_region_list(self) -> list[Region]:
+        """
+        Parse the list of operation regions.
+        If no regions are present, returns an empty list.
```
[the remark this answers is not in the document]
kingiler: Okay, I'm just blindly copying from `_parse_op_result_list` doc.

xdsl/parser/core.py
```diff
@@ -764,7 +777,7 @@ def _parse_op_result(self) -> tuple[Span, int]:
     def _parse_op_result_list(self) -> list[tuple[Span, int]]:
         """
         Parse the list of operation results.
-        If no results are present, returns an empty list.
+        If no results are present, return an empty list.
```
tobiasgrosser, line 780: Can you drop this change, as this is unrelated.

kingiler: Sorry for introducing unrelated changes.

PR #1302, opened 2023-07-19
core: (parser) factor out resolve_operands

tobiasgrosser:
Currently some hand-written parsers do not resolve operands correctly. This helper function will reduce the boilerplate needed to resolve operands.

### DIFF
--- a/xdsl/parser/core.py
+++ b/xdsl/parser/core.py
@@ -811,11 +811,12 @@ class Parser(AttrParser):
         and checks that the type is consistent.
 
         If the length of args and input_types does not match, an error is raised at
-        the the location error_pos.
+        the location error_pos.
         """
-        if len(args) != len(input_types):
+        length = len(list(input_types))
+        if len(args) != length:
             self.raise_error(
-                f"expected {len(input_types)} operand types but had {len(args)}",
+                f"expected {length} operand types but had {len(args)}",
                 error_pos,
             )
 
--- a/xdsl/parser/core.py
+++ b/xdsl/parser/core.py
@@ -807,17 +807,7 @@ class Parser(AttrParser):
         If the operand is not yet defined, it creates a forward reference.
         If the operand is already defined, it returns the corresponding SSA value,
         and checks that the type is consistent.
-
-        If the length of args and input_types does not match, an error is raised at
-        the current location.
         """
-        func_type_pos = self._current_token.span.start
-        if len(args) != len(input_types):
-            self.raise_error(
-                f"expected {len(input_types)} operand types but had {len(args)}",
-                func_type_pos,
-            )
-
         return [
             self.resolve_operand(operand, type)
             for operand, type in zip(args, input_types)
@@ -850,8 +840,16 @@ class Parser(AttrParser):
 
         self.parse_punctuation(":", "function type signature expected")
 
+        func_type_pos = self._current_token.span.start
+
         # Parse function type
         func_type = self.parse_function_type()
+
+        if len(args) != len(func_type.inputs):
+            self.raise_error(
+                f"expected {len(func_type.inputs)} operand types but had {len(args)}",
+                func_type_pos,
+            )
 
         self._parse_optional_location()
 
--- a/xdsl/parser/core.py
+++ b/xdsl/parser/core.py
@@ -23,7 +23,7 @@ from xdsl.ir import (
     SSAValue,
 )
 from xdsl.parser.attribute_parser import AttrParser
-from xdsl.parser.base_parser import ParserState
+from xdsl.parser.base_parser import ParserState, Position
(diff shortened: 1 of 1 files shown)


xdsl/parser/core.py
```diff
@@ -819,24 +850,13 @@ def _parse_generic_operation(self, op_type: type[Operation]) -> Operation:
...
-
         # Parse function type
         func_type = self.parse_function_type()
 
-        if len(args) != len(func_type.inputs):
-            self.raise_error(
-                f"expected {len(func_type.inputs)} operand types but had {len(args)}",
-                func_type_pos,
-            )
+        operands = self.resolve_operands(args, func_type.inputs)
```
kingiler, line 856: It feels better if `resolve_operands` is after all parsing.

tobiasgrosser: Done.

kingiler (approved):
It is a helpful function. Thanks!

xdsl/parser/core.py
```diff
@@ -792,6 +797,32 @@ def _parse_op_result_list(self) -> list[tuple[Span, int]]:
...
+        type the following happens:
+
+        If the operand is not yet defined, it creates a forward reference.
+        If the operand is already defined, it returns the corresponding SSA value,
+        and checks that the type is consistent.
+
+        If the length of args and input_types does not match, an error is raised at
+        the current location.
+        """
+        func_type_pos = self._current_token.span.start
```
superlopuh, line 814: This feels like a surprising coupling of the state of the parser with the behaviour of this function, which only declares inputs and outputs in the type signature. When called by clients outside of the parser, is it really expected that they call it when in a state where the current token is relevant to the error message? I have a feeling that clients might call this method once they've parsed the entire line, so the token wouldn't really make sense.

tobiasgrosser: Indeed. I got tempted but this is clearly not the right approach.

xdsl/parser/core.py
```diff
@@ -792,6 +797,32 @@ def _parse_op_result_list(self) -> list[tuple[Span, int]]:
     def parse_optional_attr_dict(self) -> dict[str, Attribute]:
         return self.parse_optional_dictionary_attr_dict()
 
+    def resolve_operands(
+        self, args: list[UnresolvedOperand], input_types: ArrayAttr[Attribute],
```
math-fehr, line 801:
```suggestion
        self, args: Sequence[UnresolvedOperand], input_types: Sequence[Attribute],
```
That way, this will also work with list, tuples, and ArrayAttr.

xdsl/parser/core.py
```diff
@@ -792,6 +797,32 @@ def _parse_op_result_list(self) -> list[tuple[Span, int]]:
...
+        error_pos: Span
```
math-fehr, line 802:
```suggestion
        error_pos: Position
```
Since we anyway don't use the `end` part.

math-fehr (approved):
I just added two comments, but otherwise that looks good to me!

xdsl/parser/core.py
tobiasgrosser: I tried to make input_types a Sequence, but I currently get the following pyright error:

```
Argument of type "ParameterDef[ArrayAttr[Attribute]]" cannot be assigned to parameter "input_types" of type "Sequence[Attribute]" in function "resolve_operands"
```

Are you sure this is a Sequence type? I guess we could make it one but maybe that would be a separate PR?

math-fehr: Oh, indeed I think at some point I might have seen a similar problem.
Can we switch it back to `Sequence` then, and instead pass `inputs.data` when calling it?
That way it should still work with everything else!

tobiasgrosser: I used `Iterable` instead.

math-fehr: `Iterable` cannot work because it does not have a `__len__` :/

xdsl/parser/core.py
```diff
@@ -811,11 +811,12 @@ def resolve_operands(
         and checks that the type is consistent.
 
         If the length of args and input_types does not match, an error is raised at
-        the the location error_pos.
+        the location error_pos.
         """
-        if len(args) != len(input_types):
+        length = len(list(input_types))
```
math-fehr, line 816: I have to say, that feels like a hack.
Here, the method expects a `Sequence` and an `Iterable`, but checks the length of both.
I feel that having `Sequence` for both would make more sense, and we would just need to add `.data` during the function call.

tobiasgrosser: That works for me as well.

superlopuh: It looks like the line stayed in. Could you please remove the temporary list allocation?

PR #1305, opened 2023-07-20
dialects: (arith) make parsing robust with respect to block ordering

tobiasgrosser:
The function Parse.parse_operand only works, if we are certain that all operands are defined before they are used. As arith can be used in arbitrary control flow graphs, we need to use the robust pattern of Parser.parse_unresolved_operand + Parser.resolve_operands.

Going forward we will remove remaining uses of parse_operand and eventually rename parse_unresolved_operand to parse_operand.


math-fehr (approved):
:+1:

tests/xdsl_opt/test_xdsl_opt.py
```diff
@@ -156,7 +157,7 @@ def test_diagnostic_exception():
 
     opt = xDSLOptMain(args=[filename_in])
 
-    with pytest.raises(DiagnosticException):
+    with pytest.raises(ParseError):
```
tobiasgrosser, line 160: This change is needed, as the Parser does now the type-checking and now throws an error: 
> operand is used with type f32, but has been previously used or defined with type i32

Is there a specific reason that we do type-checking in the parser. I feel this is somehow the wrong place to do type checking as it e.g. prevents further type errors at other locations to be emitted.

math-fehr: Let me check if we should instead change the program that is being parsed there.

The reason this is checked at parse time and not after, is because the program cannot even be constructed.
If you refer to a value that is expected to have type `i32`, and then you parse that value definition which has a different type, let's say `i64`, then you have a fundamental inconsistency in your program.
It's not even that it is ill-typed, it is that a value has two different types, and thus we should signal it directly.

math-fehr:
I quickly fixed the test so it still check the same thing as before, which is that a unverified program will return a diagnostic error.

tests/xdsl_opt/test_xdsl_opt.py
tobiasgrosser: I am confused. In which way would a type error be `fundamentally inconsistent`?

math-fehr: What I mean is that you cannot even create an IR that correspond to that program, because you have a value that has two different types.

Also, note that this is different than having an operation not verifying. This is handled later.

math-fehr: Also, this is exactly how it is done in MLIR, just in case.

tests/xdsl_opt/unverified_program.mlir
```diff
@@ -1,5 +1,5 @@
 builtin.module {
     %a = "arith.constant"() {value = 1 : i32} : () -> i32
     %b = "arith.constant"() {value = 1.0 : f32} : () -> f32
-    %c = arith.addf %a, %b : f32
+    %c = "arith.addf"(%a, %b) : (i32, f32) -> f32
```
superlopuh, line 4: why this change?

superlopuh (approved):
thank you

tests/xdsl_opt/unverified_program.mlir
superlopuh: Ah I missed this discussion in the main thread of this PR

tests/filecheck/dialects/arith/arith_cfg.mlir
```diff
@@ -0,0 +1,32 @@
+// RUN: xdsl-opt %s
```
webmiche, line 1: Could we use the test dialcet here?

Also is this really a property of arith or just a property of the ir-structure in general? If so, we could propably replace everything with the test dialect and move this to a new folder "ir-structure" or something.

webmiche (commented):
LGTM, modulo using the test dialect.

tests/filecheck/dialects/arith/arith_cfg.mlir
tobiasgrosser: No, we can unfortunately not. This test case tests the custom parsing of the `arith.add` operation, so we need at least arith to stick around. Or are you suggesting I replace the func.func and cf.br operation with the test dialect? I guess that could be possible.

webmiche: Yes, I think the func and cf dialects could certainly go.

Ah yes, I see the changes happen within arith, so that one can't go :+1:

tobiasgrosser: The test.termop does not allow `> 0` successors. Are you suggesting we expand termop to allow this?

webmiche: Yes, that would be needed too, I guess. IMO, you can do this in this PR, I don't mind. It feels like a rather small change...

tobiasgrosser: I tried this, but the build breaks as I cannot get var_successor_def to work.

AntonLydike (approved):
LGTM, modulo some pyright issues it seems?

tests/filecheck/dialects/arith/arith_cfg.mlir
tobiasgrosser: @webmiche, can you see if you can get the test dialect fixed here?

xdsl/dialects/test.py
```diff
@@ -34,7 +36,7 @@ class TestOp(IRDLOperation):
 
 
 @irdl_op_definition
-class TestTermOp(TestOp):
+class TestTermOp(IRDLOperation):
```
webmiche, line 39: @math-fehr it seems that this was broken before. This would only collect the constraints of the parent op for verification and not the ones specified within this class. Do we want it to be that way? Or did I just bandage over an IRDL bug?

xdsl/dialects/test.py
```diff
@@ -46,6 +48,10 @@ class TestTermOp(TestOp):
 
     name = "test.termop"
 
+    res: VarOpResult = var_result_def()
```
webmiche, line 51: Does it make sense for a terminator to have results?
This is something we used in some of the tests. That's why my earlier try failed ([CI run](https://github.com/xdslproject/xdsl/actions/runs/5611853160/jobs/10268770422))

tests/filecheck/dialects/arith/arith_cfg.mlir
webmiche: Fixed it. Let's see what Mathieu thinks about my fixes and then we can merge, I guess.

xdsl/dialects/test.py
webmiche: My one question was also about SSAValue scoping which I was not 100% sure about anymore. If SSAValues are block-scoped, this does not make sense as the terminator is the last operation in the block. If they are region-scoped (which from a quick google search, they seem to be in MLIR), I guess you could build IR that uses it.

And then the follow up question is: Will there ever be a use for it? Who would define a terminator that defines an SSAValue?

I guess we can leave it as-is. Better test more behavior than what people actually use. But, it was something that made me raise my eyebrow, so I decided to comment on it :sweat_smile:

compor: > I guess you could build IR that uses it.

Yeah, true, I missed that point.
But my doubts remain on the utility of this, and when in Rome, I'd rather do what MLIR does (or does not do in this case :-)

xdsl/dialects/test.py
math-fehr: Seems like a :bug: to me yes!

xdsl/dialects/test.py
math-fehr: Technically, a terminator can have regions, and even results.
Should they? I don't think so, but *maybe* someone one day will have a use case for it, so we can keep it as MLIR does.

math-fehr (approved):
:+1:

xdsl/dialects/test.py
webmiche: I added regions as well. Let's keep it like this...

xdsl/dialects/test.py
webmiche: opened #1312

webmiche:
I think we can merge after a rebase.

PR #1306, opened 2023-07-20
documentation: (toy) use MLIROptPass to lower affine

superlopuh:
This is the first non-xdsl pass in Toy, I'm not sure whether it would be worth reimplementing `--canonicalise`, `--cse`, and `--lower-affine` in the short term.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,13 @@ pytest-nb:
 	pytest -W error --nbval -vv docs --ignore=docs/mlir_interoperation.ipynb --nbval-current-env
 
 # run tests for Toy tutorial
-tests-toy:
+filecheck-toy:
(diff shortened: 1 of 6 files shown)


superlopuh:
Ah the tests don't install mlir-opt before running...

webmiche:
> Ah the tests don't install mlir-opt before running...

Yes, and that is on purpose as we want to test that things work without MLIR. I guess you will habe to disable these tests or something...

superlopuh:
Yeah that makes sense. I'll add a similar with-mlir mechanism to run this

JosseVanDelm (approved):
Cool!

compor (approved):

adutilleul (approved):

PR #1308, opened 2023-07-20
misc: make pyright only runs on .py files

superlopuh:
I frequently run `make tests` (thanks again) and this is minor noise when the diff contains files that aren't Python.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,6 @@ precommit-install:
 precommit:
 	pre-commit run --all
 
-
-
-
 # run pyright on all files in the current git commit
 pyright:
 	pyright $(shell git diff --staged --name-only  -- '*.py')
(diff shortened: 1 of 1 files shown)


adutilleul (approved):

Makefile
```diff
@@ -54,9 +54,12 @@ precommit-install:
 precommit:
 	pre-commit run --all
 
+
+
+
```
AntonLydike, line 59: Whitespace change? Or was formatting off before?

AntonLydike (approved):
LGTM

webmiche (approved):

kingiler:
Do we want `pyright` run on `.ipynb` files?

Makefile
superlopuh: ah my b this was the test :)

superlopuh:
That's a good question, I'll add it and see what happens

PR #1310, opened 2023-07-20
core: Rename _OpT to IRDLOperationInvT (NFC)

math-fehr:
This renames `_OpT` to IRDLOperationInvT, and also adds covariant and contravariant versions.

### DIFF
--- a/xdsl/irdl.py
+++ b/xdsl/irdl.py
@@ -434,7 +434,7 @@ def irdl_to_attr_constraint(
 #  \___/| .__/ \___|_|  \__,_|\__|_|\___/|_| |_|
 #       |_|
 
-_OpT = TypeVar("_OpT", bound="IRDLOperation")
+IRDLOperationInvT = TypeVar("IRDLOperationInvT", bound="IRDLOperation")
 
 
 class IRDLOperation(Operation):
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
This indeed seems more readable.

superlopuh (approved):

xdsl/irdl.py
```diff
@@ -434,7 +434,11 @@ def irdl_to_attr_constraint(
 #  \___/| .__/ \___|_|  \__,_|\__|_|\___/|_| |_|
 #       |_|
 
-_OpT = TypeVar("_OpT", bound="IRDLOperation")
+IRDLOperationInvT = TypeVar("IRDLOperationInvT", bound="IRDLOperation")
+IRDLOperationCovT = TypeVar("IRDLOperationCovT", bound="IRDLOperation", covariant=True)
+IRDLOperationContrT = TypeVar(
+    "IRDLOperationContrT", bound="IRDLOperation", contravariant=True
+)
```
webmiche, line 441: Why do you add all three and then only use one?

webmiche (approved):
LGTM :+1:

xdsl/irdl.py
math-fehr: It's for the public API, so we can have these once we need them, and won't need to change this again

PR #1311, opened 2023-07-20
[WIP] dialects: Add print_int, print_char ops.

JosseVanDelm:
Hi, this is an initial PR for adding more self-contained printing capabilities to the `printf` dialect.
I'm still unsure about a lot of stuff, and it is not finished yet, some initial help is appreciated!

## This PR adds:
Two new ops:
* A "printf.print_char" op. An operation that prints a single character, that is taken in as a 32-bit integer.
* A "printf.print_int" op. An operation that prints a 32 bit integer.
One pass `lower-printchar-to-putchar` (naming suggestions welcome!) that implements:
* A rewrite from `print_int` to a `func.func  inline_itoa` calls. The implementation of `inline_itoa` is inserted in the pass. `Inline_itoa` was written in MLIR assembly directly.
* A rewrite that lowers `print_char`s to C style `put_chars`. 

## Rationale
Some platforms don't ship with printf implementations or are very bad at printing. This implementation allows for printing integers:
* without relying much on external libraries (besides `put_char`)
* at (maybe?) increased performance, since now MLIR-passes can actually perform constant folding and the likes.
I think we can (if necessary) even write a pass that composes the current `printf.format()` into `print_char`s and `print_int`s if necessary.

Why a `print_char` op?
* Because I felt like using putchar directly was a bit ugly (it returns what it puts, which you typically don't really need, I believe, and because some other platforms might implement these calls in a different way than the standard C approach.

Why a `print_int` op?
* Some platforms implement this directly. It is for example available as a RISC-V syscall.

## Still missing/Questions/Concerns
* It feels like I'm not implementing signless semantics properly (should I make a signed print and unsigned print instead?). The implementation currently tests if the value is negative, and then prints a minus sign if necessary.
* I feel like the added itoa routine is not very easy to read and/or review. Any ideas on making this easier?
* Should I keep this limited to 32-bit integers?
* Some ops in the conversion seem to be still missing in xdsl (math.log10?), I'll work on these next.

### DIFF
--- a/tests/filecheck/dialects/print/print_basics.mlir
+++ b/tests/filecheck/dialects/print/print_basics.mlir
@@ -11,6 +11,7 @@ builtin.module {
     printf.print_format "{}", %144 : i32
     printf.print_format "{}", %144 : i32 {unit}
     "printf.print_char"(%12) : (i32) -> ()
+    "printf.print_int"(%12) : (i32) -> ()
 }
 
 // CHECK:       printf.print_format "Hello world!"
@@ -20,3 +21,4 @@ builtin.module {
 // CHECK-NEXT:  printf.print_format "{}", %0 : i32
 // CHECK-NEXT:  printf.print_format "{}", %0 : i32 {"unit"}
 // CHECK-NEXT:  "printf.print_char"(%{{.*}}) : (i32) -> ()
+// CHECK-NEXT:  "printf.print_int"(%{{.*}}) : (i32) -> ()
--- a/tests/filecheck/dialects/print/print_basics.mlir
+++ b/tests/filecheck/dialects/print/print_basics.mlir
@@ -10,6 +10,7 @@ builtin.module {
 
     printf.print_format "{}", %144 : i32
     printf.print_format "{}", %144 : i32 {unit}
+    "printf.print_char"(%12) : (i32) -> ()
 }
 
 // CHECK:       printf.print_format "Hello world!"
@@ -18,3 +19,4 @@ builtin.module {
 // CHECK-NEXT:  printf.print_format "Uses vals twice {} {} {} {}", %1 : i32, %0 : i32, %1 : i32, %0 : i32
 // CHECK-NEXT:  printf.print_format "{}", %0 : i32
 // CHECK-NEXT:  printf.print_format "{}", %0 : i32 {"unit"}
+// CHECK-NEXT:  "printf.print_char"(%{{.*}}) : (i32) -> ()
--- a/xdsl/dialects/printf.py
+++ b/xdsl/dialects/printf.py
@@ -99,7 +99,17 @@ class PrintCharOp(IRDLOperation):
     char: Operand = operand_def()
 
 
+@irdl_op_definition
+class PrintIntOp(IRDLOperation):
+    """
+    Print a single Integer
+    """
+
+    name = "printf.print_int"
+    int: Operand = operand_def(builtin.IntegerType)
+
+
 Printf = Dialect(
-    [PrintFormatOp, PrintCharOp],
+    [PrintFormatOp, PrintCharOp, PrintIntOp],
     [],
 )
--- a/xdsl/dialects/printf.py
+++ b/xdsl/dialects/printf.py
@@ -4,9 +4,11 @@ from xdsl.dialects import builtin
 from xdsl.ir import Dialect, Operation, SSAValue, VerifyException
 from xdsl.irdl import (
     IRDLOperation,
+    Operand,
     VarOperand,
     attr_def,
     irdl_op_definition,
+    operand_def,
     var_operand_def,
 )
 from xdsl.parser import Parser
@@ -85,9 +87,19 @@ class PrintFormatOp(IRDLOperation):
         return op
 
 
+@irdl_op_definition
+class PrintCharOp(IRDLOperation):
+    """
+    Print a single character
+
+    Equivalent to putchar in C.
+    """
+
+    name = "printf.print_char"
+    char: Operand = operand_def()
+
+
 Printf = Dialect(
-    [
-        PrintFormatOp,
-    ],
+    [PrintFormatOp, PrintCharOp],
     [],
 )
--- a/xdsl/transforms/printf_to_putchar.py
+++ b/xdsl/transforms/printf_to_putchar.py
@@ -1,0 +1,40 @@
+from xdsl.dialects import func
+from xdsl.dialects.builtin import ModuleOp, i32
+from xdsl.dialects.printf import PrintCharOp
+from xdsl.ir.core import MLContext
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    GreedyRewritePatternApplier,
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class LowerPrintCharToPutchar(RewritePattern):
+    """
+    Rewrite Pattern that rewrites printf.print_char to an
+    external function call
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: PrintCharOp, rewriter: PatternRewriter, /):
+        func_call = func.Call.get("putchar", [op.char], [i32])
+        # Add empty new_results, since a result is necessary for linking
+        # putchar, but the result does not exist anywhere.
+        rewriter.replace_matched_op(func_call, new_results=[])
+
+
+class LowerPrintCharToPutcharPass(ModulePass):
+    name = "lower-printchar-to-putchar"
+
+    # lower to func.call
+    def apply(self, ctx: MLContext, op: ModuleOp) -> None:
+        walker = PatternRewriteWalker(
+            GreedyRewritePatternApplier([LowerPrintCharToPutchar()]),
+            apply_recursively=True,
+        )
+        walker.rewrite_module(op)
+        # Add external putchar reference
+        op.body.block.add_ops([func.FuncOp.external("putchar", [i32], [i32])])
--- a/xdsl/transforms/printf_to_putchar.py
+++ b/xdsl/transforms/printf_to_putchar.py
@@ -1,7 +1,8 @@
-from xdsl.dialects import func
-from xdsl.dialects.builtin import ModuleOp, i32
-from xdsl.dialects.printf import PrintCharOp
-from xdsl.ir.core import MLContext
+from xdsl.builder import ImplicitBuilder
+from xdsl.dialects import arith, func, scf
+from xdsl.dialects.builtin import IndexType, ModuleOp, f32, i32
+from xdsl.dialects.printf import PrintCharOp, PrintIntOp
+from xdsl.ir.core import Block, MLContext
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
     GreedyRewritePatternApplier,
@@ -10,6 +11,7 @@ from xdsl.pattern_rewriter import (
     RewritePattern,
     op_type_rewrite_pattern,
 )
+from xdsl.printer import Printer
 
 
 class LowerPrintCharToPutchar(RewritePattern):
@@ -26,15 +28,125 @@ class LowerPrintCharToPutchar(RewritePattern):
         rewriter.replace_matched_op(func_call, new_results=[])
 
 
+class ConvertPrintIntToItoa(RewritePattern):
+    """
+    Rewrite Pattern that rewrites printf.print_int to an
+    inline_itoa function
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: PrintIntOp, rewriter: PatternRewriter, /):
+        func_call = func.Call.get("inline_itoa", [op.int], [])
+        rewriter.replace_matched_op(func_call)
+
+
+def get_inline_itoa():
+    inline_itoa_func = func.FuncOp("inline_itoa", ((i32,), ()))
+    with ImplicitBuilder(inline_itoa_func.body) as (integer,):
+        zero = arith.Constant.from_int_and_width(0, i32)
+        is_negative = arith.Cmpi.get(integer, zero, "slt")
+        # If the value is zero, just print one zero
+        is_zero = arith.Cmpi.get(integer, zero, "eq")
+        is_zero_block = Block()
+        # If is_zero is true:
+        with ImplicitBuilder(is_zero_block):
+            ascii_zero = arith.Constant.from_int_and_width(48, i32)
+            # The ascii value for zero is 48 in decimal
+            PrintCharOp((ascii_zero,))
+            scf.Yield.get()
+        # If is_zero is false:
+        is_not_zero_block = Block()
+        with ImplicitBuilder(is_not_zero_block):
+            # If the value is negative, print the minus sign, return abs value
+            is_negative_block = Block()
+            with ImplicitBuilder(is_negative_block):
+                ascii_minus = arith.Constant.from_int_and_width(52, i32)
+                PrintCharOp((ascii_minus,))
+                negative_input = arith.Subi(zero, integer)
+                scf.Yield.get(negative_input)
+            # If the value is positive, just continue
+            is_positive_block = Block()
+            with ImplicitBuilder(is_positive_block):
+                scf.Yield.get(integer)
+            absolute_value = scf.If.get(
+                is_negative, [i32], [is_negative_block], [is_positive_block]
+            )
+            one = arith.Constant.from_int_and_width(1, i32)
+            plus_one = arith.Addi(absolute_value, one)
+            # FIXME has to be unsigned
+            plus_one_fp = arith.SIToFPOp((plus_one,), (f32,))
+            # FIXME this has to be replaced with a log10 op!
+            # %size = math.log10 %plus_one_fp : f32
+            # %size_ceiled = math.ceil %size : f32
+            # FIXME has to be signed
+            size_int = arith.FPToSIOp((plus_one_fp,), (i32,))
+
+            # Using indexes here, since ints are not supported
+            # by scf.for (for now?)
+            zero_index = arith.Constant.from_int_and_width(0, IndexType())
+            one_index = arith.Constant.from_int_and_width(1, IndexType())
+            size_int_index = arith.IndexCastOp((size_int,), (IndexType(),))
+
+            loop_body = Block(arg_types=(IndexType(),))
+            with ImplicitBuilder(loop_body) as (index_var,):
+                size_minus_one = arith.Subi(size_int, one)
+                index_var_int = arith.IndexCastOp((index_var,), (i32,))
+                position = arith.Subi(size_minus_one, index_var_int)
+                # digit = (num // (10**pos)) % 10
+                ten = arith.Constant.from_int_and_width(10, i32)
+                # FIXME, line below actually has to be:
+                # %i_0 = "math.ipowi"(%ten, %position): (i32, i32)-> i32
+                i_0 = arith.Addi(ten, position)
+                i_1 = arith.DivUI(absolute_value, i_0)
+                digit = arith.RemUI(i_1, ten)
+                # ascii value for zero is 48 in decimal
+                ascii_offset = arith.Constant.from_int_and_width(48, i32)
+                char = arith.Addi(digit, ascii_offset)
+                PrintCharOp((char,))
+                scf.Yield.get()
+            scf.For.get(zero_index, size_int_index, one_index, (), loop_body)
+
+            scf.Yield.get()
+        scf.If.get(
+            is_zero,
+            [],
+            [is_zero_block],
+            [is_not_zero_block],
+        )
+        func.Return()
+
+    return inline_itoa_func
+
+
+if __name__ == "__main__":
+    printer = Printer()
+    printer.print(get_inline_itoa())
+
+
 class LowerPrintCharToPutcharPass(ModulePass):
     name = "lower-printchar-to-putchar"
 
     # lower to func.call
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
-        walker = PatternRewriteWalker(
+        print_int_walker = PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    ConvertPrintIntToItoa(),
+                ]
+            ),
+            apply_recursively=True,
+        )
+        print_int_walker.rewrite_module(op)
+        # This has to happen first, since inline itoa
+        # contains some references to printf.print_char
+        op.body.block.add_ops([get_inline_itoa()])
+        # Add function implementation for inline_itoa
+        # TODO Don't insert this if the operation is not there!
+        print_char_walker = PatternRewriteWalker(
             GreedyRewritePatternApplier([LowerPrintCharToPutchar()]),
             apply_recursively=True,
         )
-        walker.rewrite_module(op)
+        print_char_walker.rewrite_module(op)
         # Add external putchar reference
+        # TODO Don't insert this if the operation is not there!
         op.body.block.add_ops([func.FuncOp.external("putchar", [i32], [i32])])
--- a/tests/filecheck/dialects/print/printchar_to_putchar.mlir
+++ b/tests/filecheck/dialects/print/printchar_to_putchar.mlir
@@ -1,0 +1,14 @@
+// RUN: xdsl-opt -p lower-printchar-to-putchar %s | xdsl-opt | filecheck %s
+
+builtin.module {
+    // 4 in ascii
+    "func.func"() ({
+        %4 = arith.constant 52 : i32
+        "printf.print_char"(%4) : (i32) -> ()
+	"func.return"() : () -> ()
+    }) {"sym_name" = "main", "function_type" = () -> (), "sym_visibility" = "private"} : () -> ()
+}
+
+// CHECK: %{{.*}} = arith.constant 52 : i32
+// CHECK-NEXT: %{{.*}} = "func.call"(%{{.*}}) {"callee" = @putchar} : (i32) -> i32
+// CHECK: func.func private @putchar(i32) -> i32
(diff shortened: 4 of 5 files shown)


tests/filecheck/dialects/print/print_basics.mlir
```diff
@@ -10,6 +10,8 @@ builtin.module {
 
     printf.print_format "{}", %144 : i32
     printf.print_format "{}", %144 : i32 {unit}
+    "printf.print_char"(%12) : (i32) -> ()
```
superlopuh, line 13: hmm, what size should char be? feels like `u8` to me, but I'm really not sure

xdsl/dialects/printf.py
```diff
@@ -85,9 +87,29 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
...
+
+    name = "printf.print_int"
+    int: Operand = operand_def(builtin.IntegerType)
+
+
 Printf = Dialect(
-    [
-        PrintFormatOp,
-    ],
+    [PrintFormatOp, PrintCharOp, PrintIntOp],
```
superlopuh, line 113:
```suggestion
    [PrintFormatOp, PrintCharOp, PrintIntOp,],
```

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+    inline_itoa function
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: PrintIntOp, rewriter: PatternRewriter, /):
+        func_call = func.Call.get("inline_itoa", [op.int], [])
+        rewriter.replace_matched_op(func_call)
+
+
+def get_inline_itoa():
```
compor, line 44: I can't think of any other way to make this "prettier" other than trying to factor out parts of it to separate utility methods.
This might make testing easier as well, building incrementally.

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+        # If is_zero is true:
+        with ImplicitBuilder(is_zero_block):
+            ascii_zero = arith.Constant.from_int_and_width(48, i32)
+            # The ascii value for zero is 48 in decimal
+            PrintCharOp((ascii_zero,))
+            scf.Yield.get()
+        # If is_zero is false:
+        is_not_zero_block = Block()
+        with ImplicitBuilder(is_not_zero_block):
+            # If the value is negative, print the minus sign, return abs value
```
compor, line 61: The printing the minus sign and returning the absolute value could be one of those utility methods.

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+                negative_input = arith.Subi(zero, integer)
+                scf.Yield.get(negative_input)
+            # If the value is positive, just continue
+            is_positive_block = Block()
+            with ImplicitBuilder(is_positive_block):
+                scf.Yield.get(integer)
+            absolute_value = scf.If.get(
+                is_negative, [i32], [is_negative_block], [is_positive_block]
+            )
+            one = arith.Constant.from_int_and_width(1, i32)
```
compor, line 75: getting the number of digits could also be factored out.

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+            # FIXME has to be unsigned
+            plus_one_fp = arith.SIToFPOp((plus_one,), (f32,))
+            size = math.Log10Op((plus_one_fp,), (f32,))
+            size_ceiled = math.CeilOp((size,), (f32,))
+            # FIXME has to be unsigned
+            size_int = arith.FPToSIOp((size_ceiled,), (i32,))
+
+            zero_index = arith.Constant.from_int_and_width(0, IndexType())
+            one_index = arith.Constant.from_int_and_width(1, IndexType())
+            size_int_index = arith.IndexCastOp((size_int,), (IndexType(),))
```
compor, line 86: I found the name `size` confusing, maybe `digits` or `digits_num` is better?

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+
+            loop_body = Block(arg_types=(IndexType(),))
+            with ImplicitBuilder(loop_body) as (index_var,):
```
compor, line 89: I don't think a lot can be done about this section; it's rather straightforward.
Maybe clean up and provide a clear comment.

If it's worth factoring out (for readability, testing, etc), I think it will depend on what you decide to do in the previous parts, seeing how much they improve this.

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+        zero = arith.Constant.from_int_and_width(0, i32)
+        is_negative = arith.Cmpi.get(integer, zero, "slt")
+        # If the value is zero, just print one zero
+        is_zero = arith.Cmpi.get(integer, zero, "eq")
+        is_zero_block = Block()
+        # If is_zero is true:
+        with ImplicitBuilder(is_zero_block):
+            ascii_zero = arith.Constant.from_int_and_width(48, i32)
+            # The ascii value for zero is 48 in decimal
+            PrintCharOp((ascii_zero,))
```
compor, line 56: Seeing this bit here and the one for the minus sign below made me wondering if something like this is better:

```py
def _printchar(char: str, offset: int = 0) -> PrintCharOp:
    assert len(char) == 1
    c = arith.Constant.from_int_and_width(ord(char) + offset, i32)
    return PrintCharOp((c,))

_printchar("0")
```

But this does generalize well for the `0` + offset `for`-loop.
Maybe there is a way (i.e., accepting as offset the `arith.Constant` too)?

compor (commented):
Thanks for providing such detailed info for this PR, it made the review easier.

I mainly focused on `itoa`.

As for signless integers, my understanding is that treating them as signed is compatible with their semantics and it's up to the corresponding operation to interpret them as required.
Keeping this working just for 32-bit int sounds good to me for now.

xdsl/dialects/printf.py
```diff
@@ -85,9 +87,29 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
...
+@irdl_op_definition
+class PrintCharOp(IRDLOperation):
+    """
+    Print a single character
+
+    Equivalent to putchar in C.
+    """
+
+    name = "printf.print_char"
+    char: Operand = operand_def()
```
AntonLydike, line 99: I think you should restrict it to `i32` here?

xdsl/transforms/printf_to_putchar.py
AntonLydike, line 44: nit: this is no longer inline, I think I'm to blame for this, because I named the function this when I added the stub initially, and then we shifted the design later on without renaming... I would name it something like `generate_itoa_func_def`?

xdsl/dialects/printf.py
```diff
@@ -85,9 +87,29 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
...
+
+
+@irdl_op_definition
+class PrintIntOp(IRDLOperation):
+    """
+    Print a single Integer
+    """
+
+    name = "printf.print_int"
+    int: Operand = operand_def(builtin.IntegerType)
```
AntonLydike, line 109: You added the `PrintIntOp` and had a justification in your PR description for it, but then you don't use it in the lowering afaics?

I would just drop it, as `print_int` is just a subset of `print_char`, right?

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,148 @@
...
+    return inline_itoa_func
+
+
+if __name__ == "__main__":
+    printer = Printer()
+    printer.print(get_inline_itoa())
+
+
+class LowerPrintCharToPutcharPass(ModulePass):
+    name = "lower-printchar-to-putchar"
```
AntonLydike, line 123: `lower-printf-to-putchar`?

AntonLydike (approved):
LGTM with minor corrections!

tests/filecheck/dialects/print/print_basics.mlir
JosseVanDelm: I've used signless 8-bit types in the latest update

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: I've added a new constructor `PrintCharOp.from_constant_char()`
Woud be interesting to get your opinion on this

xdsl/dialects/printf.py
JosseVanDelm: I've currently refactored this to be constrained on i8's as suggested by @superlopuh

xdsl/dialects/printf.py
JosseVanDelm: It is used for sure, but i think i should write better filecheck tests for this

JosseVanDelm: Not sure why you consider `print_int` to be a subset of `print_char` I would actually consider it the other way around, `print_int` will actually generate multiple calls to `print_char`

AntonLydike: oooooh, print int prints more than just a single digit!

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,242 @@
...
+            position = arith.Subi(size_minus_one, index_var_int)
+            # digit = (num // (10**pos)) % 10
+            ten = arith.Constant.from_int_and_width(10, i32)
+            i_0 = math.IPowIOp((ten, position), (i32,))
+            i_1 = arith.DivUI(absolute_value, i_0)
+            digit = arith.RemUI(i_1, ten)
+            # ascii value for zero is 48 in decimal
+            ascii_offset = arith.Constant.from_int_and_width(48, i32)
+            char = arith.Addi(digit, ascii_offset)
+            char_i8 = arith.TruncIOp.get(char, IntegerType(8))
```
adutilleul, line 162: `nit`:  you could probably define again an alias `i8` at the top of the file, right ?

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,242 @@
...
+            size_minus_one = arith.Subi(digits, one)
+            index_var_int = arith.IndexCastOp((index_var,), (i32,))
+            position = arith.Subi(size_minus_one, index_var_int)
+            # digit = (num // (10**pos)) % 10
+            ten = arith.Constant.from_int_and_width(10, i32)
+            i_0 = math.IPowIOp((ten, position), (i32,))
+            i_1 = arith.DivUI(absolute_value, i_0)
+            digit = arith.RemUI(i_1, ten)
+            # ascii value for zero is 48 in decimal
+            ascii_offset = arith.Constant.from_int_and_width(48, i32)
```
adutilleul, line 160: `nit`: it would be less verbose to call `ord("0")`

adutilleul (approved):
Nice!

xdsl/dialects/printf.py
```diff
@@ -85,9 +89,41 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
...
+
+    @staticmethod
+    def from_constant_char(char: str) -> PrintCharOp:
+        """
+        This constructor returns a PrintCharOp that prints the value supplied
+        in "char" as a python char.
+        """
+        assert len(char) == 1, "Only single characters are supported"
+        ascii_value = ord(char)
+        assert ascii_value < 255, "Only ascii characters are supported"
```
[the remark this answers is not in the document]
JosseVanDelm: good catch!

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,243 @@
...
+            PrintCharOp.from_constant_char("-")
+            negative_input = arith.Subi(zero, integer)
+            scf.Yield.get(negative_input)
+        # If the value is positive, just return value itself as abs value
+        with ImplicitBuilder(is_positive_block):
+            scf.Yield.get(integer)
+
+        return absolute_value
+
+    def print_digits(digits: SSAValue, absolute_value: scf.If) -> scf.For:
```
kingiler, line 136: It looks weird that `absolute_value` has type `scf.If`.

kingiler:
I am happy that `printf` works and I'm thinking of a way that do not need to use `pow`.
The following python code looks good despite two while loops:
```python
def reverse(n: int) -> int: # n is absolute value
    m = 0
    while n:
        m = 10 * m + n % 10
        n //= 10
    return m


def print_integer(n: int) -> None:
    m = reverse(n)
    while m:
        print(m % 10, end="") # print single digit as char
        m //= 10


print_integer(123456)
```
Just some ideas.

JosseVanDelm:
Thanks for the idea @kingiler, i'm not entirely sure if what you are proposing is strictly necessary though, since the `mlir-opt` `--convert-math-to-funcs` seem to already implement this behaviour for us, as it inserts a function to perform powers on integers.

xdsl/dialects/printf.py
```diff
@@ -85,9 +89,41 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
...
+    name = "printf.print_char"
+    char: Operand = operand_def(i8)
+
+    @staticmethod
+    def from_constant_char(char: str) -> PrintCharOp:
+        """
+        This constructor returns a PrintCharOp that prints the value supplied
+        in "char" as a python char.
+        """
+        assert len(char) == 1, "Only single characters are supported"
```
superlopuh, line 109: Please raise ValueError instead

xdsl/dialects/printf.py
```diff
@@ -85,9 +89,41 @@ def parse(cls: type[PrintFormatOp], parser: Parser) -> PrintFormatOp:
...
+@irdl_op_definition
+class PrintCharOp(IRDLOperation):
+    """
+    Print a single character
+
+    Equivalent to putchar in C.
+    """
+
+    name = "printf.print_char"
+    char: Operand = operand_def(i8)
```
superlopuh, line 101: Please add an init

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,243 @@
...
+            print_digits(digits, absolute_value)
+            # Yield from is_not_zero_block
+            scf.Yield.get()
+        # Return from itoa function
+        func.Return()
+
+    return inline_itoa_func
+
+
+if __name__ == "__main__":
```
superlopuh, line 212:
```suggestion
```

xdsl/transforms/printf_to_putchar.py
superlopuh: Or just `printf-to-putchar`?

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,243 @@
...
+        op.body.block.add_ops([get_inline_itoa()])
+        # Add function implementation for inline_itoa
+        # TODO Don't insert this if the operation is not there!
+        print_char_walker = PatternRewriteWalker(
+            GreedyRewritePatternApplier([LowerPrintCharToPutchar()]),
+            apply_recursively=True,
+        )
+        print_char_walker.rewrite_module(op)
+        # Add external putchar reference
+        # TODO Don't insert this if the operation is not there!
```
superlopuh, line 242: Would be good to add this functionality as part of this PR. it should be pretty easy with the symbol table functionality, or just a manual scan of the module op body.

superlopuh (changes requested):

xdsl/dialects/printf.py
JosseVanDelm: Will do!

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: Could you give a reference/example to the symbol table functionality? I don't know it. Thanks!

superlopuh: It might not work here, but in theory can be something like in the `get_op_for_symbol` function in interpreter.py

``` python


    def get_op_for_symbol(self, symbol: str) -> Operation:
        if symbol in self.symbol_table:
            return self.symbol_table[symbol]
        else:
            raise InterpretationError(f'Could not find symbol "{symbol}"')
```

adutilleul (approved):

xdsl/transforms/printf_to_putchar.py
superlopuh: If it doesn't work I'd be interested :)

JosseVanDelm: Can you doublecheck now

JosseVanDelm:
As per @AntonLydike's suggestion I've removed the `inline_itoa` and refactored it to `mlir_itoa`, which seems to be what mlir-opt does when you lower a `math.ipowi` with `convert-math-to-funcs`

tests/dialects/test_printf.py
```diff
@@ -58,3 +58,13 @@ def test_global_symbol_name_generation():
     s3 = printf_to_llvm._key_from_str(")")  # pyright: ignore[reportPrivateUsage]
 
     assert s2 == s3
+
+
+def test_printchar_non_ascii():
+    with pytest.raises(ValueError):
+        print_dialect.PrintCharOp.from_constant_char("🔥")
```
superlopuh, line 65: ❤️‍🔥

tests/dialects/test_printf.py
```diff
@@ -58,3 +58,13 @@ def test_global_symbol_name_generation():
     s3 = printf_to_llvm._key_from_str(")")  # pyright: ignore[reportPrivateUsage]
 
     assert s2 == s3
+
+
+def test_printchar_non_ascii():
+    with pytest.raises(ValueError):
```
superlopuh, line 64: Can you please check that the text raised here is what we expect?

xdsl/dialects/printf.py
```diff
@@ -94,23 +94,31 @@ class PrintCharOp(IRDLOperation):
...
+
     @staticmethod
     def from_constant_char(char: str) -> PrintCharOp:
         """
         This constructor returns a PrintCharOp that prints the value supplied
         in "char" as a python char.
         """
-        assert len(char) == 1, "Only single characters are supported"
+        if len(char) != 1:
+            raise ValueError("Only single characters are supported")
```
superlopuh, line 116: Characters is ambiguous, maybe something like this?
```suggestion
            raise ValueError(f'Unexpected char value "{char}", input must be a single ascii character')
```

xdsl/dialects/printf.py
superlopuh: This would be a useful change to minimise spurious git diffs in the future

xdsl/transforms/printf_to_putchar.py
```diff
@@ -206,38 +206,49 @@ def print_digits(digits: SSAValue, absolute_value: scf.If) -> scf.For:
...
-        print_char_walker = PatternRewriteWalker(
-            GreedyRewritePatternApplier([LowerPrintCharToPutchar()]),
-            apply_recursively=True,
-        )
-        print_char_walker.rewrite_module(op)
-        # Add external putchar reference
-        # TODO Don't insert this if the operation is not there!
-        op.body.block.add_ops([func.FuncOp.external("putchar", [i32], [i32])])
+        # Check if there are any printints
+        contains_printint = False
```
superlopuh, line 218: nit:
```suggestion
        contains_printint = any(isinstance(op_in_module, PrintIntOp) for op_in_module in op.walk())
```

xdsl/transforms/printf_to_putchar.py
```diff
@@ -206,38 +206,49 @@ def print_digits(digits: SSAValue, absolute_value: scf.If) -> scf.For:
...
+        for op_in_module in op.walk():
+            if isinstance(op_in_module, PrintIntOp):
+                contains_printint = True
+                # One printint encountered is enough
+                break
+        if contains_printint:
+            print_int_walker = PatternRewriteWalker(
+                GreedyRewritePatternApplier(
```
superlopuh, line 226:
```suggestion
                ConvertPrintIntToItoa()
```
I don't thin ka greedy rewrite pattern applier with a single patern is any different to just that pattern. Also, I don't think we want to apply it recursively. Applying the pattern recursively would allow us to run the rewrite pattern on ops that it creates, but the `ConvertPrintIntToItoa` pattern does not create new `print_int` ops so it won't ever be applied in the recursive step.

superlopuh: Similarly below.

xdsl/transforms/printf_to_putchar.py
compor: I like this!

compor (approved):

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: :open_mouth:

tests/dialects/test_printf.py
```diff
@@ -61,10 +61,14 @@ def test_global_symbol_name_generation():
...
 
 
 def test_printchar_no_char():
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError) as e:
         print_dialect.PrintCharOp.from_constant_char("This should not work")
+    assert (
+        e.value.args[0]
+        == 'Unexpected char value "This should not work", input must be a single ascii character'
+    )
```
kingiler, line 74:
```suggestion
    with pytest.raises(
        ValueError,
        match='Unexpected char value "This should not work", input must be a single ascii character',
    ):
        print_dialect.PrintCharOp.from_constant_char("This should not work")

```

xdsl/transforms/printf_to_putchar.py
```diff
@@ -215,38 +214,28 @@ class PrintfToPutcharPass(ModulePass):
...
         # Check if there are any printchars
-        for op_in_module in op.walk():
-            if isinstance(op_in_module, PrintCharOp):
-                contains_printchar = True
-                # One printchar encountered is enough
-                break
+        contains_printchar = any(
+            isinstance(op_in_module, PrintCharOp) for op_in_module in op.walk()
+        )
         if contains_printchar:
```
kingiler, line 235:
```suggestion
        if any(
            isinstance(op_in_module, PrintCharOp)
            for op_in_module in op.walk()
        ):
```

kingiler: Can combine if only use once for if condition.

xdsl/transforms/printf_to_putchar.py
```diff
@@ -215,38 +214,28 @@ class PrintfToPutcharPass(ModulePass):
...
-        contains_printint = False
-        for op_in_module in op.walk():
-            if isinstance(op_in_module, PrintIntOp):
-                contains_printint = True
-                # One printint encountered is enough
-                break
+        contains_printint = any(
+            isinstance(op_in_module, PrintIntOp) for op_in_module in op.walk()
+        )
         if contains_printint:
```
kingiler, line 220:
```suggestion
        if any(
            isinstance(op_in_module, PrintIntOp)
            for op_in_module in op.walk()
        ):
```

kingiler: Can combine if only use once for if condition.

xdsl/dialects/printf.py
```diff
@@ -1,17 +1,21 @@
...
     VarOperand,
     attr_def,
     irdl_op_definition,
+    operand_def,
     var_operand_def,
 )
 from xdsl.parser import Parser
 from xdsl.printer import Printer
 
+i8 = builtin.IntegerType(8)
```
kingiler, line 17: I am not sure but `i8` seems useful to be put in `xdsl/dialect/builtin.py`?

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: IMO the explicit contains_printchar is more readable

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: Again, I think it is more readable if it is explicitly stated

xdsl/transforms/printf_to_putchar.py
superlopuh: Can you please pass an `SSAValue` here instead?

xdsl/transforms/printf_to_putchar.py
```diff
@@ -0,0 +1,243 @@
...
+        3. Get the number of digits for the absolute value
+        4. Go through the digits in reverse order and print them one by one
+
+    Several things to note about this specfic implementation:
+        * It does not require floating point arithmetic
+        * It does not require a printf implementation, only putchar.
+    """
+
+    # get_number_of_digits implementation
+    def get_number_of_digits(absolute_value: Operation) -> OpResult:
```
superlopuh, line 65:
```suggestion
    def get_number_of_digits(absolute_value: SSAValue) -> OpResult:
```
Can you please pass an SSAValue here instead? What is the advantage of Operation?

xdsl/dialects/printf.py
superlopuh: plz

superlopuh (changes requested):
Looking good, some minor changes would be nice to make or discuss

tests/dialects/test_printf.py
JosseVanDelm: agreed, that is clearer!

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: This was not a conscious decision :)

xdsl/transforms/printf_to_putchar.py
JosseVanDelm: Thanks for the remark @kingiler and the suggestion @superlopuh

xdsl/dialects/printf.py
JosseVanDelm: Sure, but what do you mean with 
> to minimise spurious git diffs in the future

?

superlopuh (approved):
Looking forward to using this :)

JosseVanDelm:
Thank you all for the reviews! I learned a lot while working on this!

PR #1313, opened 2023-07-20
core: Add declarative assembly format specification to IRDL

math-fehr:
This PR adds the first version of the [declarative assembly format](https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-format).
In its current form, it only supports the `attr-dict` directive, which specifies where the attribute dictionary should be parsed.

The structure of the PR is:
* `irdl.py` was modified (and moved to `irdl/irdl.py`) to allow specifying a declarative format with the `assembly_format` field.
* `declarative_assembly_format_parser.py` will parse the format and create the corresponding data structures, as well as checking that the format is correct. The parser is written by reusing the MLIR format lexer and parser. This is slightly hacky, but makes the change much smaller.
* `declarative_assembly_format.py` contains the data structure to represent the declarative format, and to parse/print a particular format.

Following PRs will incrementally add more features. I couldn't really remove more from this PR without making it impossible to understand.

### DIFF
--- a/tests/test_declarative_assembly_format.py
+++ b/tests/test_declarative_assembly_format.py
@@ -1,0 +1,205 @@
+from __future__ import annotations
+
+from io import StringIO
+
+import pytest
+
+from xdsl.dialects.builtin import ModuleOp
+from xdsl.ir.core import MLContext, Operation
+from xdsl.irdl import IRDLOperation, irdl_op_definition
+from xdsl.parser.core import Parser
+from xdsl.printer import Printer
+from xdsl.utils.exceptions import PyRDLOpDefinitionError
+
+################################################################################
+# Utils for this test file                                                     #
+################################################################################
+
+
+def check_roundtrip(program: str, ctx: MLContext):
+    """Check that the given program roundtrips exactly (including whitespaces)."""
+    parser = Parser(ctx, program)
+    ops: list[Operation] = []
+    while op := parser.parse_optional_operation():
+        ops.append(op)
+
+    res_io = StringIO()
+    printer = Printer(stream=res_io)
+    for op in ops[:-1]:
+        printer.print_op(op)
+        printer.print("\n")
+    printer.print_op(ops[-1])
+
+    assert program == res_io.getvalue()
+
+
+def check_equivalence(program1: str, program2: str, ctx: MLContext):
+    """Check that the given programs are structurally equivalent."""
+
+    parser = Parser(ctx, program1)
+    ops1: list[Operation] = []
+    while op := parser.parse_optional_operation():
+        ops1.append(op)
+
+    parser = Parser(ctx, program2)
+    ops2: list[Operation] = []
+    while op := parser.parse_optional_operation():
+        ops2.append(op)
+
+    assert ModuleOp(ops1).is_structurally_equivalent(ModuleOp(ops2))
+
+
+################################################################################
+# Tests that we cannot have both a declarative and Python-defined format       #
+################################################################################
+
+
+def test_format_and_print_op():
+    """
+    Check that an operation with an assembly format cannot redefine the print method.
+    """
+    with pytest.raises(
+        PyRDLOpDefinitionError, match="Cannot define both an assembly format"
+    ):
+
+        @irdl_op_definition
+        class FormatAndPrintOp(IRDLOperation):  # type: ignore[reportUnusedImport]
+            name = "test.format_and_print"
+
+            assembly_format = "attr-dict"
+
+            def print(self, printer: Printer) -> None:
+                pass
+
+
+def test_format_and_parse_op():
+    """
+    Check that an operation with an assembly format cannot redefine the parse method.
+    """
+    with pytest.raises(
+        PyRDLOpDefinitionError, match="Cannot define both an assembly format"
+    ):
+
+        @irdl_op_definition
+        class FormatAndParseOp(IRDLOperation):  # type: ignore[reportUnusedImport]
+            name = "test.format_and_parse"
+
+            assembly_format = "attr-dict"
+
+            @classmethod
+            def parse(cls, parser: Parser) -> FormatAndParseOp:
+                raise NotImplementedError()
+
+    with pytest.raises(
+        PyRDLOpDefinitionError, match="Cannot define both an assembly format"
+    ):
+
+        @irdl_op_definition
+        class FormatAndParseOp(IRDLOperation):  # type: ignore[reportUnusedImport]
+            name = "test.format_and_parse"
+
+            assembly_format = "attr-dict"
+
+            @classmethod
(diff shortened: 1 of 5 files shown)


tests/test_declarative_assembly_format.py
```diff
@@ -0,0 +1,205 @@
...
+        class FormatAndParseOp(IRDLOperation):  # type: ignore[reportUnusedImport]
+            name = "test.format_and_parse"
+
+            assembly_format = "attr-dict"
+
+            @classmethod
+            def parse(cls, parser: Parser) -> FormatAndParseOp:
+                raise NotImplementedError()
+
+    with pytest.raises(
```
kingiler, line 93: Is it a dup?

tests/test_declarative_assembly_format.py
```diff
@@ -0,0 +1,205 @@
...
+def test_format_and_print_op():
+    """
+    Check that an operation with an assembly format cannot redefine the print method.
+    """
+    with pytest.raises(
+        PyRDLOpDefinitionError, match="Cannot define both an assembly format"
+    ):
+
+        @irdl_op_definition
+        class FormatAndPrintOp(IRDLOperation):  # type: ignore[reportUnusedImport]
```
kingiler, line 66: nit: `# pyright: ignore[reportUnusedClass]`

tests/test_declarative_assembly_format.py
math-fehr: Oh yes good catch!

xdsl/irdl/irdl.py
```diff
@@ -1148,6 +1165,17 @@ def get_constraint(
 
                 raise wrong_field_exception(field_name)
 
+        op_def.assembly_format = pyrdl_def.assembly_format
+        if op_def.assembly_format is not None and (
+            pyrdl_def.print != Operation.print
+            or not hasattr(pyrdl_def.parse, "__func__")
+            or pyrdl_def.parse.__func__ != Operation.parse.__func__
```
kingiler, line 1172: I'm not sure why we are checking for `__func__`?

We can use `getattr(pyrdl_def.parse, "__func__", None) != getattr(Operation.parse, "__func__", None)`

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.parse_optional_keyword("attr-dict"):
+            return self.create_attr_dict_directive(False)
+        if self.parse_optional_keyword("attr-dict-with-keyword"):
+            return self.create_attr_dict_directive(True)
+        self.raise_error(f"unexpected token '{self._current_token.text}'")
+
+    def create_attr_dict_directive(self, with_keyword: bool) -> AttrDictDirective:
+        """Create an attribute dictionary directive, and update the parsing state."""
+        if self.has_attr_dict:
+            self.raise_error("'attr-dict' directive has already been seen")
```
webmiche, line 147:
```suggestion
            self.raise_error("'attr-dict' can only be defined once.")
```

webmiche (approved):
Cool idea :+1: 

LGTM in general, though I did not look very deeply.

xdsl/irdl/declarative_assembly_format_parser.py
webmiche: Or
```suggestion
            self.raise_error("'attr-dict' directive can only occur once in the assembly format description")
```

tests/test_declarative_assembly_format.py
```diff
@@ -0,0 +1,205 @@
...
+# Tests that we cannot have both a declarative and Python-defined format       #
+################################################################################
+
+
+def test_format_and_print_op():
+    """
+    Check that an operation with an assembly format cannot redefine the print method.
+    """
+    with pytest.raises(
+        PyRDLOpDefinitionError, match="Cannot define both an assembly format"
```
compor, line 62: 🙏🏽

compor:
I didn't understand why attributes first till I saw the [requirements][1]

Care to explain the bullet point in item 7. in that list?
I'm not sure if it is relevant with this patch, so if it's noise please ignore.

[1]: https://mlir.llvm.org/docs/DefiningDialects/Operations/#requirements

math-fehr:
> I didn't understand why attributes first till I saw the [requirements](https://mlir.llvm.org/docs/DefiningDialects/Operations/#requirements).
> 
> Care to explain the bullet point in item 7. in that list? I'm not sure if it is relevant to this patch, so if it's noise please ignore it.

So what I did is implemented item 6 with `attr-dict`, because it is mandatory.
The reason is that since you can attach any attribute to any operation, any operation might have an attribute that you need to print/parse.
So even if you don't have any attributes, you still need to print/parse the attribute dictionary, that's why I implemented this first, so it stays correct.

Point 7 is something for later. Essentially, `attr-dict` will show all attributes that are not yet printed/parsed by the other directives. For instance, if you have an attribute `value`, the format "$value attr-dict" will print an operation with the `{value = 3, foo = 5}` with the format `3 { foo = 5}`. Essentially, not printing `value` again in the attribute dictionary because it is already printed before.

xdsl/irdl/irdl.py
superlopuh: What would be the advantage of this change?

kingiler: `pyright` is unhappy about accessing `__func__`

compor: This is caught and reported by the CI, no?

kingiler: It is reported by CI so I'm suggesting ways to solve it.

kingiler: Because if one use `hasattr` and then access the attribute, mostly it can combined into `getattr`.

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -0,0 +1,152 @@
...
+    def verify_attr_dict(self):
+        """
+        Check that the attribute dictionary is present.
+        """
+        if not self.has_attr_dict:
+            self.raise_error("'attr-dict' directive not found")
+
+    def parse_directive(self) -> FormatDirective:
+        """
+        Parse a format directive, with the following forrmat:
```
tobiasgrosser, line 131: `forrmat -> format`

tobiasgrosser (approved):
This all looks great to me. Very nice work!

xdsl/irdl/irdl.py
superlopuh: That's a good point. I wonder if there's some pyright-blessed way to do this, otherwise might as well do `getattr`

superlopuh (approved):
v nice

PR #1314, opened 2023-07-20
core: Add punctuation directive to the assembly format

math-fehr:
This PR adds support for punctuation in the declarative assembly format.

For instance, "`(` `)` attr-dict" will parse correctly "test.op()".
Note that some punctuations will implicitly add spaces before/after them.

### DIFF
--- a/tests/test_declarative_assembly_format.py
+++ b/tests/test_declarative_assembly_format.py
@@ -189,3 +189,39 @@ def test_attr_dict(program: str, generic_program: str):
 
     check_roundtrip(program, ctx)
     check_equivalence(program, generic_program, ctx)
+
+
+################################################################################
+# Punctuations and keywords                                                    #
+################################################################################
+
+
+@pytest.mark.parametrize(
+    "format, program",
+    [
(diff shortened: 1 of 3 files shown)


tests/test_declarative_assembly_format.py
```diff
@@ -0,0 +1,241 @@
...
+from xdsl.parser.core import Parser
+from xdsl.printer import Printer
+from xdsl.utils.exceptions import PyRDLOpDefinitionError
+
+################################################################################
+# Utils for this test file                                                     #
+################################################################################
+
+
+def check_roundtrip(program: str, ctx: MLContext):
```
superlopuh, line 19: I feel like I've seen this function before. Maybe worth factoring out into a helper file?

math-fehr: Oh, that comment was from a change in the other patch (since Github is really annoying with stacked PRs).
I'll keep that in mind!

superlopuh (approved):

compor (approved):

AntonLydike (approved):
:heart:

PR #1316, opened 2023-07-20
core: Add operands in declarative assembly format

math-fehr:
This PR adds operands to the declarative assembly format.
An operand is parsed with the `$operand_name` syntax, and should also have its type parsed with the `type($operand_name)` syntax.
For instance, ``$lhs `,` $rhs `:` type($lhs) `,` type($rhs)`` will parse `test.op %0, %1 : i32, i64` correctly.

All operands should be parsed exactly once, and should have their type parsed exactly once as well.

Future PRs will add support for optional/variadic operands, as well as removing the need to parse their types if they have a unique type, or if their type can be deduced from the other parsed types.

### DIFF
--- a/tests/test_declarative_assembly_format.py
+++ b/tests/test_declarative_assembly_format.py
@@ -6,7 +6,7 @@ import pytest
 
 from xdsl.dialects.builtin import ModuleOp
 from xdsl.ir import MLContext, Operation
-from xdsl.irdl import IRDLOperation, irdl_op_definition
+from xdsl.irdl import IRDLOperation, irdl_op_definition, operand_def
 from xdsl.parser import Parser
 from xdsl.printer import Printer
 from xdsl.utils.exceptions import PyRDLOpDefinitionError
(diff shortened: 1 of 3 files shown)


tests/test_declarative_assembly_format.py
```diff
@@ -247,3 +248,140 @@ class PunctuationOp(IRDLOperation):
...
+
+def test_unknown_variable():
+    """Test that variables should refer to an element in the operation."""
+    with pytest.raises(
+        PyRDLOpDefinitionError,
+        match="expected variable to refer to an operand, attribute, region, result, or successor",
+    ):
+
+        @irdl_op_definition
+        class UnknownVarOp(IRDLOperation):  # type: ignore[reportUnusedImport]
```
superlopuh, line 266:
```suggestion
        class UnknownVarOp(IRDLOperation):  # pyright: ignore[reportUnusedImport]
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -24,20 +26,24 @@ class ParsingState:
     It contains the elements that have already been parsed.
     """
 
+    operands: list[None | UnresolvedOperand]
```
superlopuh, line 29: super merga ultra nit:
```suggestion
    operands: list[UnresolvedOperand | None]
```

This is so tiny I feel silly suggesting this change but None's always been the last type so far in unions

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -128,6 +130,58 @@ def verify_attr_dict(self):
         if not self.has_attr_dict:
             self.raise_error("'attr-dict' directive not found")
 
+    def parse_optional_variable(self) -> FormatDirective | None:
+        """
+        Parse a variable, if present, with the following format:
+          variable ::= `$` bare-ident
+        The variable should refer to an operand, attribute, region, result, or succesor.
```
superlopuh, line 137:
```suggestion
        The variable should refer to an operand, attribute, region, result, or successor.
```

superlopuh (approved):

xdsl/irdl/declarative_assembly_format.py
math-fehr: No no, you're right, my eyes cried looking at this again.

PR #1318, opened 2023-07-21
dialect: (riscv) initial custom syntax

kingiler:
This is an initial PR to custom RISCV syntax.

The major change is dropping bracket around operands

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -333,7 +333,13 @@ class RISCVOp(Operation, ABC):
 
     @classmethod
     def parse(cls, parser: Parser) -> Self:
-        args = parser.parse_op_args_list()
+        args = (
+            parser.parse_optional_undelimited_comma_separated_list(
+                parser.parse_optional_unresolved_operand,
+                parser.parse_unresolved_operand,
+            )
+            or ()
+        )
         regions = parser.parse_region_list()
         attributes = parser.parse_optional_attr_dict()
         parser.parse_punctuation(":")
@@ -347,7 +353,8 @@ class RISCVOp(Operation, ABC):
         )
 
     def print(self, printer: Printer) -> None:
-        printer.print_operands(self.operands)
+        printer.print(" ")
+        printer.print_list(self.operands, printer.print_operand)
         printer.print_regions(self.regions)
         printer.print_op_attributes(self.attributes)
         printer.print(" : ")
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -333,13 +333,12 @@ class RISCVOp(Operation, ABC):
 
     @classmethod
     def parse(cls, parser: Parser) -> Self:
-        args = (
-            parser.parse_optional_undelimited_comma_separated_list(
-                parser.parse_optional_unresolved_operand,
-                parser.parse_unresolved_operand,
-            )
-            or ()
+        args = parser.parse_optional_undelimited_comma_separated_list(
+            parser.parse_optional_unresolved_operand,
+            parser.parse_unresolved_operand,
         )
+        if args is None:
+            args = []
         regions = parser.parse_region_list()
         attributes = parser.parse_optional_attr_dict()
         parser.parse_punctuation(":")
--- a/tests/filecheck/dialects/riscv/riscv_ops.mlir
+++ b/tests/filecheck/dialects/riscv/riscv_ops.mlir
@@ -1,269 +1,269 @@
 // RUN: xdsl-opt %s | xdsl-opt --print-op-generic | xdsl-opt | filecheck %s
 "builtin.module"() ({
-  riscv.label() ({
-    %0 = riscv.get_register() : () -> !riscv.reg<>
-    %1 = riscv.get_register() : () -> !riscv.reg<>
+  riscv.label ({
+    %0 = riscv.get_register : () -> !riscv.reg<>
+    %1 = riscv.get_register : () -> !riscv.reg<>
     // RV32I/RV64I: 2.4 Integer Computational Instructions
 
     // Integer Register-Immediate Instructions
-    %addi = riscv.addi(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
-    // CHECK: %{{.*}} = riscv.addi(%{{.*}}) {"immediate" = 1 : i32} : (!riscv.reg<>) -> !riscv.reg<>
-    %slti = riscv.slti(%0) {"immediate" = 1 : i32}: (!riscv.reg<>) -> !riscv.reg<>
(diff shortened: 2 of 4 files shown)


xdsl/dialects/riscv.py
```diff
@@ -346,13 +351,55 @@ def parse(cls, parser: Parser) -> Self:
             regions=regions,
         )
 
+    @classmethod
+    def parse_args(cls, parser: Parser) -> Sequence[UnresolvedOperand]:
+        if arg := parser.parse_optional_unresolved_operand():
```
superlopuh, line 356: I think @math-fehr added an optional list without delimiter helper recently, could you please take a look?

xdsl/dialects/riscv.py
```diff
@@ -346,13 +351,55 @@ def parse(cls, parser: Parser) -> Self:
...
+        return []
+
+    @classmethod
+    def parse_attributes(cls, parser: Parser) -> Mapping[str, Attribute]:
+        return parser.parse_optional_attr_dict()
+
     def print(self, printer: Printer) -> None:
-        printer.print_operands(self.operands)
+        first = True
+        first &= not self.print_operands(printer, first)
```
superlopuh, line 371: what does first mean? it feels like this logic could be shifted out of the helper methods, and be here for clarity, even if that means checking for an empty sequence twice

xdsl/dialects/riscv.py
```diff
@@ -346,13 +351,55 @@ def parse(cls, parser: Parser) -> Self:
...
+            if not first:
+                printer.print(",")
+            printer.print(" ")
+            printer.print_list(self.operands, printer.print_operand)
+        return self.operands
+
+    def print_attributes(self, printer: Printer, first: bool) -> Set[str]:
+        """
+        Print list of attributes,
+        return the list of attributes being processed.
```
superlopuh, line 398: what does this mean? can you please elaborate in the comment? it feels like this is another case of adding helper methods before it is clear that they will be used later on. Adding a `self.print_attributes` method that just forwards to priner.print_op_attributes makes the code less clear in the short term. Am I missing something? Is it necessary?

superlopuh (changes requested):

xdsl/dialects/riscv.py
tobiasgrosser: I would suggest to first put up a pr that just drops the `()` around all operands. This should be a one-line change and does not yet need any of the helper functions. Let's avoid introducing all the helper functions, as these are controversial and I would like to discuss  them with Anton in person.

superlopuh: Ah no I see it is being overridden below, then please expand on the comment on how the logic works. To me, it seems like the list of attributes returned is not "processed" but rather "printed", maybe "printed in the custom format". I'd recommend changing the name of the method to something like "print_known_attributes" or, "print_attributes_custom_format"

xdsl/dialects/riscv.py
kingiler: It seems there are some issues with `parser.parse_optional_undelimited_comma_separated_list`.

math-fehr: This function should be put directly in the `parser` I would argue.
Also, what issue do you have with this function? `parser.parse_optional_undelimited_comma_separated_list(parser.parse_optional_unresolved_operand, parser.parse_unresolved_operand)` should work without problems.

xdsl/dialects/riscv.py
```diff
@@ -346,13 +351,55 @@ def parse(cls, parser: Parser) -> Self:
...
+            args = [arg]
+            while parser.parse_optional_punctuation(",") and (
+                arg := parser.parse_optional_unresolved_operand()
+            ):
+                args.append(arg)
+            return args
+        return []
+
+    @classmethod
+    def parse_attributes(cls, parser: Parser) -> Mapping[str, Attribute]:
```
math-fehr, line 366: Why do we redefine these things here? Can we just call `parser.parse_optional_attr_dict` directly?

xdsl/dialects/riscv.py
```diff
@@ -346,13 +351,55 @@ def parse(cls, parser: Parser) -> Self:
...
+            for name, attr in self.attributes.items()
+            if name not in processed_attrs
+        }
+        printer.print_op_attributes(remaining_attrs)
         printer.print_regions(self.regions)
-        printer.print_op_attributes(self.attributes)
         printer.print(" : ")
         printer.print_operation_type(self)
 
+    def print_operands(self, printer: Printer, first: bool) -> Sequence[SSAValue]:
```
math-fehr, line 383: Can we move that in the printer as well?

xdsl/dialects/riscv.py
kingiler: In the new custom syntax, both `args` and attributes are comma separated.

`riscv.op %0, %1, 123, "comment" (regions) ...`

xdsl/dialects/riscv.py
kingiler: `parse_attributes` will be override by classes which implement custom attribute syntax. By default it uses generic attribute dict. See [here](https://github.com/xdslproject/xdsl/pull/1318/files#diff-c2e583c286e17ace72eadfcf055304a684874cd8bfbb4c2b51ff7d9d21ff7be4R2141) for `riscv.DirectiveOp` example.

kingiler:
Now the PR is minimized to only drop brackets around operands.

tobiasgrosser (approved):
I guess this looks good to me so far.

tobiasgrosser:
@superlopuh, could you have a last look?

xdsl/dialects/riscv.py
```diff
@@ -333,7 +333,13 @@ def assembly_line(self) -> str | None:
...
     @classmethod
     def parse(cls, parser: Parser) -> Self:
-        args = parser.parse_op_args_list()
+        args = (
+            parser.parse_optional_undelimited_comma_separated_list(
+                parser.parse_optional_unresolved_operand,
+                parser.parse_unresolved_operand,
+            )
+            or ()
+        )
```
superlopuh, line 342:
```suggestion
        args = parser.parse_optional_undelimited_comma_separated_list(
            parser.parse_optional_unresolved_operand,
            parser.parse_unresolved_operand,
        )
        if args is None:
            args = ()
```

We try not to exploit the falseyness of `None` due to the ambiguity of double falseyness. Sometimes the result of some function is not `None` but still falsey, which can lead to hard-to-debug behaviour.

superlopuh (changes requested):
LGTM modulo small change

xdsl/dialects/riscv.py
kingiler: `parser.parse_optional_undelimited_comma_separated_list` returns either list or `None`. The return value is falsy only if it is empty list or `None` and either case returning an empty tuple is okay.

PR #1319, opened 2023-07-21
core: Fix bug in method, and improve documentation

math-fehr:
This fixes a bug in `parse_optional_undelimited_comma_separated_list`, and improve its documentation.

### DIFF
--- a/xdsl/parser/base_parser.py
+++ b/xdsl/parser/base_parser.py
@@ -252,14 +252,15 @@ class BaseParser:
         parse: Callable[[], _AnyInvT],
     ) -> list[_AnyInvT] | None:
         """
-        Parses greedily a list of elements separated by commas, and delimited
-        by the specified delimiter. Return None if no elements were parsed.
+        Parses greedily a list of elements separated by commas, if a first element is
+        present. The first element is parsed with `parse_optional`, and the remaining
+        are parsed with `parse`. The parsing stops either if the first element is not
+        present, or if no comma is present after parsing an element.
         """
         # Parse the first element, if it exist
         first_elem = parse_optional()
         if first_elem is None:
             return None
-        elems = parse_optional()
 
         # Parse the remaining elements
         elems = [first_elem]
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -475,6 +475,23 @@ def test_parse_comma_separated_list_none_delimiter_empty():
         )
 
 
+def test_parse_comma_separated_list_none_delimiter_two_no_comma():
+    """Test that a list without commas will only parse the first element."""
+    parser = Parser(MLContext(), "1 2")
+    res = parser.parse_comma_separated_list(
+        Parser.Delimiter.NONE, parser.parse_integer, " in test"
+    )
+    assert res == [1]
+    assert parser.parse_optional_integer() is not None
+
+    parser = Parser(MLContext(), "1 2")
+    parser.parse_optional_undelimited_comma_separated_list(
+        parser.parse_optional_integer, parser.parse_integer
+    )
+    assert res == [1]
+    assert parser.parse_optional_integer() is not None
+
+
 @pytest.mark.parametrize(
     "delimiter",
     [


tobiasgrosser (approved):

compor (approved):

xdsl/parser/base_parser.py
```diff
@@ -252,14 +252,15 @@ def parse_optional_undelimited_comma_separated_list(
         parse: Callable[[], _AnyInvT],
     ) -> list[_AnyInvT] | None:
         """
-        Parses greedily a list of elements separated by commas, and delimited
-        by the specified delimiter. Return None if no elements were parsed.
+        Parses greedily a list of elements separated by commas, if a first element is
+        present. The first element is parsed with `parse_optional`, and the remaining
+        are parsed with `parse`. The parsing stops either if the first element is not
+        present, or if no comma is present after parsing an element.
         """
```
kingiler, line 259:
```python
if (elem := parse_optional()) is not None:
    elems = [elem]
    while self._parse_optional_token(Token.Kind.COMMA) is not None:
        elems.append(parse())
    return elems
return None
```

kingiler (approved):
LGTM

tobiasgrosser:
Actually, I am wondering if we could improve the interface, by just taking a parse_optional method and returning "X, X, Y" just [X,X] for the previous example. Also, returning an empty list [] in case parse_optional does not match, could be nice.

tobiasgrosser:
This would turn:

```python
        args = (
            parser.parse_optional_undelimited_comma_separated_list(
                parser.parse_optional_unresolved_operand,
                parser.parse_unresolved_operand,
            )
            or ()
        )
```

into:

```

```python
        args = parser.parse_optional_undelimited_comma_separated_list(
                parser.parse_optional_unresolved_operand)
```

But maybe that somehow conflicts with our naming scheme.

kingiler:
@tobiasgrosser it would need the following
```python
if (elem := parse_optional()) is not None:
    elems = [elem]
    while (
        self._parse_optional_token(Token.Kind.COMMA) is not None
        and (elem := parse_optional()) is not None
    ):
        elems.append(elem)
    return elems
return []
```
but the pitfall is that it will parse the trailing comma. I'm not sure if this is good.

math-fehr:
> Actually, I am wondering if we could improve the interface, by just taking a parse_optional method and returning "X, X, Y" just [X,X] for the previous example. Also, returning an empty list [] in case parse_optional does not match, could be nice.

We could take only a `parse_optional`, the only reason I preferred the `parse` is because that way we can share the error reporting. But if people find it too annoying, I can move it to a `parse_optional`, and a string for the error.

I believe `x, x, y` should raise an error. This is both to stay consistent with MLIR (so people can expect the same API, instead of having these small differences), and because that doesn't involve backtracking, which MLIR (and xDSL) does not allow for users. Maybe I could add a `parse_trailing` version of this function, that would parse the trailing comma? But I'm not sure that you actually want that in your riscv case, since you know the number of operands.

For returning `[]` instead of `None`, I would personally prefer returning `None` to remain consistent throughout the API. Otherwise, we just get an inconsistent API, which is quite annoying. I feel the `or []` is a nice pattern that doesn't cost that much?

tobiasgrosser:
I see. In this case I guess at least this PR is fine. Sorry for the noise.

math-fehr:
> I see. In this case I guess at least this PR is fine. Sorry for the noise.

No worries, I'll try to take a look in more details in the MLIR repository to see if there is a function there that I'm missing.
If people feel that this should be changed though, we could remove the `parse` and replace it with an error message, that seems sensible to me.

kingiler:
Sadly, number of operands may be unknown due to `VarOperand`, for example `CustomAssemblyInstructionOp`. It would be another issue and not necessary become a blocker for this PR.

tobiasgrosser:
I feel we can merge this PR. We are going in small steps (which adds some noise), but allows us to make progress in areas that are easy.

xdsl/parser/base_parser.py
superlopuh: I prefer the version in the code, the successful path is down towards the end of the function, and it avoids nesting.

PR #1321, opened 2023-07-22
dialects: Fix wrong attribute type used for gpu.known_X_size.

PapyChacal:
Oops, seems I've been lazy with the filechecks.

### DIFF
--- a/tests/filecheck/dialects/gpu/ops.mlir
+++ b/tests/filecheck/dialects/gpu/ops.mlir
@@ -73,7 +73,7 @@
         "gpu.func"() ({
         ^bb0(%arg0: index):
             "gpu.return"() : () -> ()
-        }) {"sym_name" = "foo", "kernel", "function_type" = (index) -> ()} : () -> ()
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
This looks sensible to me. I am curious why we don't use the custom syntax here, but this is just an unrelated remark. I would also feel it would help the PR if you would actually state in the summary what was wrong and in which way the new attribute type is more correct.

superlopuh (approved):

PapyChacal:
Ah, sure!
The error: the `array<i32:...>` syntax, as used by MLIR for those attributes, are parsed as DenseArrayBase in xDSL rather than ArrayAttr. So it couldn't parse at all.
The custom syntax: I just tend to give low priority to it, focus on everything being functional before looking pretty, as long as it doesn't hurt other aspects of development.
As this is a very high-level and structural operation, you don't have to read dozens of it to debug examples or a flow, so I don't feel like it is that important here, as opposed to some ops of the stencil dialect for which I implemented syntax quicker indeed.

I'd welcome an implementation for sure, and might work on it once it becomes more important with our WGPU flow 🙂

PR #1322, opened 2023-07-22
Interpreter: translate gpu.known_size to wgsl 

cigarichard:
This pull request introduces an translation that allows the MLIR's gpu.known_block_size to be translated directly to WGSL's workgroup_size when the gpu.known_size is not None.


PapyChacal:
Cool, that should be performance-critical for this flow!
Looks like you should just update the pytest.

PapyChacal (approved):
Thanks!

tobiasgrosser:
Very cool!

superlopuh (approved):

PR #1323, opened 2023-07-22
API: Add reserved attribute names functionality to the printer.

PapyChacal:
The parser already have an optional parameter on `parse_optional_attr_dict_with_keyword` to handle attribute names that should not be in the generic attribute dictionnary of the custom syntax.
I'd like to do the same thing on the printing side, to simplify func here.

I'm preparing to merge an implementation of the SymbolTable trait, adding a *bunch* of `sym_name` attributes all over, I feel this might help. (e.g. next user is `builtin.module`)

### DIFF
--- a/xdsl/dialects/func.py
+++ b/xdsl/dialects/func.py
@@ -178,9 +178,7 @@ class FuncOp(IRDLOperation):
                 printer.print(" ")
         else:
             printer.print_attribute(self.function_type)
-        printer.print_op_attributes_with_keyword(
-            self.attributes, reserved, keyword=True
-        )
+        printer.print_op_attributes_with_keyword(self.attributes, reserved)
 
         if len(self.body.blocks) > 0:
             printer.print_region(self.body, False, False)
--- a/xdsl/dialects/func.py
+++ b/xdsl/dialects/func.py
@@ -158,6 +158,7 @@ class FuncOp(IRDLOperation):
         return func
 
     def print(self, printer: Printer):
+        reserved = {"sym_name", "function_type", "sym_visibility"}
         if self.sym_visibility:
             visibility = self.sym_visibility.data
             printer.print(f" {visibility}")
@@ -177,8 +178,6 @@ class FuncOp(IRDLOperation):
                 printer.print(" ")
         else:
             printer.print_attribute(self.function_type)
-
-        reserved = ("sym_name", "function_type", "sym_visibility")
         if any(k not in reserved for k in self.attributes.keys()):
             printer.print(" attributes")
             printer.print_op_attributes(self.attributes, reserved)
--- a/xdsl/dialects/func.py
+++ b/xdsl/dialects/func.py
@@ -178,9 +178,9 @@ class FuncOp(IRDLOperation):
                 printer.print(" ")
         else:
             printer.print_attribute(self.function_type)
-        if any(k not in reserved for k in self.attributes.keys()):
-            printer.print(" attributes")
-            printer.print_op_attributes(self.attributes, reserved)
+        printer.print_op_attributes_with_keyword(
+            self.attributes, reserved, keyword=True
+        )
 
         if len(self.body.blocks) > 0:
             printer.print_region(self.body, False, False)
--- a/xdsl/dialects/func.py
+++ b/xdsl/dialects/func.py
@@ -177,17 +177,11 @@ class FuncOp(IRDLOperation):
                 printer.print(" ")
         else:
             printer.print_attribute(self.function_type)
-        attr_dict = {
-            k: v
-            for k, v in self.attributes.items()
-            if k not in ("sym_name", "function_type", "sym_visibility")
-        }
-        if len(attr_dict) > 0:
-            printer.print(" attributes {")
-            printer.print_list(
-                attr_dict.items(), lambda i: printer.print(f'"{i[0]}" = {i[1]}')
-            )
-            printer.print("}")
+
+        reserved = ("sym_name", "function_type", "sym_visibility")
+        if any(k not in reserved for k in self.attributes.keys()):
+            printer.print(" attributes")
+            printer.print_op_attributes(self.attributes, reserved)
 
         if len(self.body.blocks) > 0:
             printer.print_region(self.body, False, False)
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -648,7 +648,7 @@ class Printer:
             i for i in attributes.items() if i[0] not in reserved_attr_names
         ]
 
-        if attribute_list:
+        if not attribute_list:
             return
         self.print(" {")
 
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -644,11 +644,11 @@ class Printer:
     def print_op_attributes(
         self, attributes: dict[str, Attribute], reserved_attr_names: Iterable[str] = []
     ) -> None:
-        attribute_list = list(
+        attribute_list = [
             i for i in attributes.items() if i[0] not in reserved_attr_names
-        )
+        ]
 
-        if len(attribute_list) == 0:
+        if attribute_list:
             return
         self.print(" {")
 
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -641,8 +641,22 @@ class Printer:
             self.print(f'"{attr_tuple[0]}" = ')
             self.print_attribute(attr_tuple[1])
 
-    def print_op_attributes(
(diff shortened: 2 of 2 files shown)


PapyChacal:
Now, here the custom printer of func still does some extra checks on the reserved keywords, to know if it should print the usual `attributes ` prefix before  the actual dict. Those checks are done by the dict too.
Do we want to add a `prefix` keyword argument to the dictionary printer, to simplify further? I guess that would cover most uses, but I,m concerned about API bloat :thinking:

xdsl/printer.py
```diff
@@ -641,13 +641,17 @@ def _print_attr_string(self, attr_tuple: tuple[str, Attribute]) -> None:
...
 
-    def print_op_attributes(self, attributes: dict[str, Attribute]) -> None:
-        if len(attributes) == 0:
-            return
+    def print_op_attributes(
+        self, attributes: dict[str, Attribute], reserved_attr_names: Iterable[str] = []
+    ) -> None:
+        attribute_list = list(
+            i for i in attributes.items() if i[0] not in reserved_attr_names
+        )
```
tobiasgrosser, line 649: Is this not just a [filter](https://www.programiz.com/python-programming/methods/built-in/filter).

tobiasgrosser (approved):
This is a great idea. I guess @math-fehr and maybe @AntonLydike or @superlopuh  might still have feedback.

xdsl/printer.py
math-fehr: That's the "pythonic" way of writing a filter

xdsl/dialects/func.py
```diff
@@ -177,17 +177,11 @@ def print(self, printer: Printer):
...
-        }
-        if len(attr_dict) > 0:
-            printer.print(" attributes {")
-            printer.print_list(
-                attr_dict.items(), lambda i: printer.print(f'"{i[0]}" = {i[1]}')
-            )
-            printer.print("}")
+
+        reserved = ("sym_name", "function_type", "sym_visibility")
+        if any(k not in reserved for k in self.attributes.keys()):
```
math-fehr, line 182: Actually, can we add that to `print_op_attributes`?
We probably can do something like adding a `with_keyword` option?
Or adding a new function `print_op_attributes_with_keyword`?
We can do it in another PR if you think this is feature creep ;)

math-fehr (approved):

xdsl/printer.py
tobiasgrosser: I was thinking of `list(filter(lambda i: i[0] not in reserved_attr_names, attributes))` or sth similar, but that's just a minor comment.

superlopuh: `filter` is significantly slower than list comprehensions, unfortunately

xdsl/dialects/func.py
```diff
@@ -177,17 +177,11 @@ def print(self, printer: Printer):
...
-            if k not in ("sym_name", "function_type", "sym_visibility")
-        }
-        if len(attr_dict) > 0:
-            printer.print(" attributes {")
-            printer.print_list(
-                attr_dict.items(), lambda i: printer.print(f'"{i[0]}" = {i[1]}')
-            )
-            printer.print("}")
+
+        reserved = ("sym_name", "function_type", "sym_visibility")
```
superlopuh, line 181: create this outside of function, and make a `set`? Seems like good practice even if it won't make a big difference here probably

xdsl/printer.py
```diff
@@ -641,13 +641,17 @@ def _print_attr_string(self, attr_tuple: tuple[str, Attribute]) -> None:
             self.print(f'"{attr_tuple[0]}" = ')
             self.print_attribute(attr_tuple[1])
 
-    def print_op_attributes(self, attributes: dict[str, Attribute]) -> None:
-        if len(attributes) == 0:
-            return
+    def print_op_attributes(
+        self, attributes: dict[str, Attribute], reserved_attr_names: Iterable[str] = []
+    ) -> None:
+        attribute_list = list(
```
superlopuh, line 647:
```suggestion
        attribute_list = [
```

xdsl/printer.py
```diff
@@ -641,13 +641,17 @@ def _print_attr_string(self, attr_tuple: tuple[str, Attribute]) -> None:
...
+            i for i in attributes.items() if i[0] not in reserved_attr_names
+        )
 
+        if len(attribute_list) == 0:
```
superlopuh, line 651:
```suggestion
        if not attribute_list:
```

xdsl/dialects/func.py
superlopuh: This is a nit

superlopuh (changes requested):
LGTM modulo minor changes

xdsl/dialects/func.py
PapyChacal: I like the idea, making it more declarative, but then it sounds to me like something that could be more integrated to PyRDL?
I'll do the small change, just giving this idea

superlopuh: Here I guess it's less about it being declarative and more about Python not being good at constant value optimisation, I think it'll create a tuple every time which seems like a waste if it's constant

xdsl/printer.py
math-fehr: I wouldn't argue that we should use `filter` because of the speed (unless in the hot section of the code), but rather because of style though.

xdsl/dialects/func.py
math-fehr: We could have it integrated with `PyRDL`, but we would need to slightly change the parser.
For instance, you could already do `reserved = self.irdl_definition.attributes.keys()`, but that would not be future-proof, as we don't actually check that all attributes are printed, so if you add a new attribute, we might miss it later on.
The declarative framework will ensure that though.

xdsl/printer.py
superlopuh: They're somewhat related but sure

xdsl/dialects/func.py
PapyChacal: Yup, I was wondering that! So how do you see this? I can add a `prefix` parameter defaulting to `attributes `?

xdsl/printer.py
PapyChacal: OOC; is it really different in that case? Aren't both of them calling the list constructor on the same Generator?
No feeling about it, applying your change, but wondering

superlopuh (approved):

xdsl/dialects/func.py
superlopuh: I think it makes sense to add this to the same PR tbh

superlopuh: Also seems reasonable to add it to the parser also, and add a roundtrip test that the funcitonality between the two functions is somewhat mirrored

AntonLydike:
On a more general level, I am not sure if we should put custom printing syntax into the generic printer. On the purely technical side this looks good!

PapyChacal:
Mh, I get this point.
On the other hand, a lot of custom syntaxes would use such a function, so I just think it makes sense to expose this somewhere; happy to hear ideas on where it would belong!

xdsl/dialects/func.py
math-fehr: The parser one is already there ^^'

PapyChacal: Yup, I wanted to add it as part of this PR and just waited for validation before wasting keystrokes 😋
Will do!

math-fehr: SO yeah, I would just put a boolean rather than a string. I don't think anyone would use a different keyword than `attributes `, and we probably don't want to give the option to people (otherwise our code will become less easy to read).

superlopuh (approved):

math-fehr:
Looks good to me! I would just remove the `with_keyword` in the name though, and that's good to merge for me!

xdsl/printer.py
```diff
@@ -652,6 +652,26 @@ def print_op_attributes(self, attributes: dict[str, Attribute]) -> None:
 
         self.print("}")
 
+    def print_op_attributes_with_keyword(
```
superlopuh, line 655: It feels like this could be fused with the previous method, with a pretty low overhead. I'd just check whether reserved attr names is not empty before filtering the attribute list

superlopuh: Oh and please remove the `if len(attributes) == 0` and replace with `if not attributes`

PapyChacal:
Discussed API naming issues in private, agreed on what I just pushed, there was just incompatible requests here.
NB: Agreed that it isn't the best naming, but I was just after factoring this first. Let's rework this API whenever useful!

PR #1325, opened 2023-07-22
Interpreter: index type arguments in mlir translate to wgsl

cigarichard:
This PR fix the issue when there is index type arguments in mlir translated into wgsl will not be converted to u32, and will keep index instead

### DIFF
--- a/xdsl/interpreters/experimental/wgsl_printer.py
+++ b/xdsl/interpreters/experimental/wgsl_printer.py
@@ -47,7 +47,7 @@ class WGSLPrinter:
                 arg_type = "f32"
             elif arg.type == builtin.IndexType():
                 arg_type = "u32"
-            elif isinstance(arg.type, MemrefType):
+            elif isa(arg.type, MemRefType[Attribute]):
                 if arg.type.element_type == builtin.IndexType():
                     arg_type = "u32"
                 else:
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/experimental/wgsl_printer.py
```diff
@@ -46,9 +47,12 @@ def _(self, op: gpu.FuncOp, out_stream: IO[str]):
                 arg_type = "f32"
             elif arg.type == builtin.IndexType():
                 arg_type = "u32"
-            elif isinstance(arg.type, MemRefType):
-                memref_typ = cast(MemRefType[Attribute], arg.type)
-                arg_type = f"array<{memref_typ.element_type}>"
+            elif isa(arg.type, MemRefType[Attribute]):
```
superlopuh, line 50:
```suggestion
            elif isinstance(arg.type, MemrefType):
```
It looks like the type of the generic parameter is not really tested against, so might as well use the simpler `isinstance`? This is a nit

superlopuh (approved):

PapyChacal:
Now that we have a tested base in, it would be nice to add tests alongside those changes!

PapyChacal (approved):
Thanks!

PR #1330, opened 2023-07-24
Implement is_optional_symbol for MLIR compliance.

PapyChacal:
That is, the `is_optional_symbol` method of `SymbolOpInterface`, matching MLIR's `isOptionalSymbol` on `Symbol`

Implement OptionalSymbolOpInterface for user-friendliness. Test it.

Our current implementation of SymbolOps is not compliant with MLIR's, which allows optionally symbol operations :upside_down_face: 
The most famous example being `builtin.module`, which can be anonymous as often for the top-level module, thus not containing any `sym_name` attribute.

I also take the opprtunity to make `builtin.module` an optional symbol operation. With this MLIR-compliant implementation, the only impact is to make the interpreter work with this compliance :slightly_smiling_face:

### DIFF
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -316,11 +316,11 @@ def test_optional_symbol_op_interface():
     assert interface is not None
     assert interface.is_optional_symbol(no_symbol)
     no_symbol.verify()
-    assert SymbolOpInterface.get_sym_attr_name(no_symbol) is None
+    assert interface.get_sym_attr_name(no_symbol) is None
 
     symbol = OptionalSymNameOp(attributes={"sym_name": StringAttr("main")})
     interface = symbol.get_trait(SymbolOpInterface)
     assert interface is not None
     assert interface.is_optional_symbol(symbol)
     symbol.verify()
-    assert SymbolOpInterface.get_sym_attr_name(symbol) == StringAttr("main")
+    assert interface.get_sym_attr_name(symbol) == StringAttr("main")
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -311,7 +311,7 @@ def test_optional_symbol_op_interface():
 
         traits = frozenset((OptionalSymbolOpInterface(),))
 
-    non_symbol = OptionalSymNameOp()
+    no_symbol = OptionalSymNameOp()
     interface = non_symbol.get_trait(SymbolOpInterface)
     assert interface is not None
     assert interface.is_optional_symbol(non_symbol)
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -312,11 +312,11 @@ def test_optional_symbol_op_interface():
         traits = frozenset((OptionalSymbolOpInterface(),))
 
     no_symbol = OptionalSymNameOp()
-    interface = non_symbol.get_trait(SymbolOpInterface)
+    interface = no_symbol.get_trait(SymbolOpInterface)
     assert interface is not None
-    assert interface.is_optional_symbol(non_symbol)
-    non_symbol.verify()
-    assert SymbolOpInterface.get_sym_attr_name(non_symbol) is None
+    assert interface.is_optional_symbol(no_symbol)
+    no_symbol.verify()
+    assert SymbolOpInterface.get_sym_attr_name(no_symbol) is None
 
     symbol = OptionalSymNameOp(attributes={"sym_name": StringAttr("main")})
     interface = symbol.get_trait(SymbolOpInterface)
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -25,9 +25,10 @@ from xdsl.irdl import (
     attr_def,
     irdl_op_definition,
     operand_def,
+    opt_attr_def,
     result_def,
 )
-from xdsl.traits import SymbolOpInterface
+from xdsl.traits import OptionalSymbolOpInterface, SymbolOpInterface
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.test_value import TestSSAValue
 
@@ -295,3 +296,31 @@ def test_symbol_op_interface():
 
     op2 = SymNameOp(attributes={"sym_name": StringAttr("symbol_name")})
     op2.verify()
+
+
+def test_optional_symbol_op_interface():
+    """
+    Test that operations that conform to OptionalSymbolOpInterface have necessary attributes.
+    """
+
+    @irdl_op_definition
+    class OptionalSymNameOp(IRDLOperation):
+        name = "no_sym_name"
+
+        sym_name = opt_attr_def(StringAttr)
+
+        traits = frozenset((OptionalSymbolOpInterface(),))
+
+    non_symbol = OptionalSymNameOp()
+    interface = non_symbol.get_trait(SymbolOpInterface)
+    assert interface is not None
+    assert interface.is_optional_symbol(non_symbol)
+    non_symbol.verify()
+    assert SymbolOpInterface.get_sym_attr_name(non_symbol) is None
+
+    symbol = OptionalSymNameOp(attributes={"sym_name": StringAttr("main")})
+    interface = symbol.get_trait(SymbolOpInterface)
+    assert interface is not None
+    assert interface.is_optional_symbol(symbol)
+    symbol.verify()
+    assert SymbolOpInterface.get_sym_attr_name(symbol) == StringAttr("main")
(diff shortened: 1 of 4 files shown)


xdsl/traits.py
```diff
@@ -236,6 +253,16 @@ def verify(self, op: Operation) -> None:
...
+class OptionalSymbolOpInterface(SymbolOpInterface):
+    """
+    Helper interface specialization for an optional Symbol.
+    """
+
+    @staticmethod
+    def is_optional_symbol(op: Operation) -> bool:
+        return True
+
+
```
PapyChacal, line 265: Digging in MLIR's sources, that's the only specialization I saw, and a few times, hence exposing it directly as is in the xDSL practical/firendly mindset!

xdsl/traits.py
```diff
@@ -212,21 +212,38 @@ class SymbolOpInterface(OpTrait):
     """
 
     @staticmethod
-    def get_sym_attr_name(op: Operation) -> StringAttr:
+    def get_sym_attr_name(op: Operation) -> StringAttr | None:
         """
         Returns the symbol of the operation
```
compor, line 217: maybe update the docstring wrt the updated return value (i.e., `None`)

tests/test_traits.py
```diff
@@ -295,3 +296,31 @@ class SymNameOp(IRDLOperation):
 
     op2 = SymNameOp(attributes={"sym_name": StringAttr("symbol_name")})
     op2.verify()
+
+
+def test_optional_symbol_op_interface():
+    """
+    Test that operations that conform to OptionalSymbolOpInterface have necessary attributes.
```
compor, line 303:
```suggestion
    Test that operations that conform to OptionalSymbolOpInterface have the necessary attributes.
```

tests/test_traits.py
```diff
@@ -295,3 +296,31 @@ class SymNameOp(IRDLOperation):
...
+
+    @irdl_op_definition
+    class OptionalSymNameOp(IRDLOperation):
+        name = "no_sym_name"
+
+        sym_name = opt_attr_def(StringAttr)
+
+        traits = frozenset((OptionalSymbolOpInterface(),))
+
+    non_symbol = OptionalSymNameOp()
```
compor, line 314:
```suggestion
    no_symbol = OptionalSymNameOp()
```

xdsl/traits.py
compor: Also, the class docstring?

compor (approved):

xdsl/traits.py
PapyChacal: Updated, WDYT?

xdsl/traits.py
```diff
@@ -203,30 +203,51 @@ class SymbolOpInterface(OpTrait):
     a `SymbolTable` (TODO). A Symbol operation should use the SymbolOpInterface interface to
     provide the necessary verification and accessors.
 
-    Currently the only requirement is a "sym_name" attribute of type StringAttr.
+    A Symbol operation may be optional or not. If - the dault - it is not optional,
```
math-fehr, line 206:
```suggestion
    A Symbol operation may be optional or not. If - the default - it is not optional,
```

xdsl/traits.py
```diff
@@ -203,30 +203,51 @@ class SymbolOpInterface(OpTrait):
...
+        Returns the symbol of the operation, if any
         """
         # import builtin here to avoid circular import
         from xdsl.dialects.builtin import StringAttr
 
+        concrete = op.get_trait(SymbolOpInterface)
+        assert concrete is not None
+        if concrete.is_optional_symbol(op) and "sym_name" not in op.attributes:
+            return None
         attr = op.attributes["sym_name"]
```
math-fehr, line 230: Here, I would write first a check that `sym_name` is in `op.attributes`.
If it is not, then I would raise an exception (probably a `VerifyException`)

xdsl/traits.py
```diff
@@ -203,30 +203,51 @@ class SymbolOpInterface(OpTrait):
...
+
     def verify(self, op: Operation) -> None:
         # import builtin here to avoid circular import
         from xdsl.dialects.builtin import StringAttr
 
+        # If this is an optional symbol, bail out early if possible.
+        concrete = op.get_trait(SymbolOpInterface)
+        assert concrete is not None
+        if concrete.is_optional_symbol(op) and "sym_name" not in op.attributes:
+            return
```
math-fehr, line 250:
```suggestion
        if self.is_optional_symbol(op) and "sym_name" not in op.attributes:
            return
```

xdsl/traits.py
```diff
@@ -203,30 +203,51 @@ class SymbolOpInterface(OpTrait):
...
+
+    More requirements are defined in MLIR; Please see MLIR documentation for Symbol and
+    SymbolTable for the requirements that are upcoming in xDSL.
 
     https://mlir.llvm.org/docs/SymbolsAndSymbolTables/#symbol
     """
 
     @staticmethod
-    def get_sym_attr_name(op: Operation) -> StringAttr:
+    def get_sym_attr_name(op: Operation) -> StringAttr | None:
```
math-fehr, line 219:
```suggestion
    def get_sym_attr_name(self, op: Operation) -> StringAttr | None:
```
I think that this design is better in general for traits.
This allows to not have to fetch the trait twice.
Most of the time, this will be used in a code like this:

xdsl/traits.py
```diff
@@ -203,30 +203,51 @@ class SymbolOpInterface(OpTrait):
...
+        concrete = op.get_trait(SymbolOpInterface)
+        assert concrete is not None
+        if concrete.is_optional_symbol(op) and "sym_name" not in op.attributes:
+            return None
         attr = op.attributes["sym_name"]
         assert isinstance(attr, StringAttr)
         return attr
 
+    @staticmethod
+    def is_optional_symbol(op: Operation) -> bool:
```
math-fehr, line 235:
```suggestion
    def is_optional_symbol(self, op: Operation) -> bool:
```
Somehow, I would feel that passing `self` would make more sense. I feel that this would make it less error-prone, by requiring a trait object.

PapyChacal: Oh, so we avoid people just calling SynbolOpInterface on the op rather than its concrete interface?

xdsl/traits.py
PapyChacal: This is the logic implemented in MLIR; what you say would raise a verifyexception on an optional symbol not being a symbol, or did I misunderstand ?

xdsl/traits.py
PapyChacal: Oh, that makes sense! I didn't know how to handle that case cleanly and I like this style 🙂

xdsl/traits.py
math-fehr: The reason I prefer calling the `get_trait` on the op directly is that this is less error-prone than just calling it on any operation, and then trigerring an assert/exception in the `is_optional_symbol` function.

xdsl/traits.py
PapyChacal: Okay I think I was confused, will push something and see what you think :)

PapyChacal: Pushed something, WDYT?
Mh, thinking of it, maybe I should just call the verifier and return the thing or None? :thinking:

xdsl/traits.py
```diff
@@ -203,30 +203,55 @@ class SymbolOpInterface(OpTrait):
...
+                    f'Operation {op.name} must have a "sym_name" attribute of type '
+                    f"`StringAttr` to conform to {SymbolOpInterface.__name__}"
+                )
         attr = op.attributes["sym_name"]
-        assert isinstance(attr, StringAttr)
+        if not isinstance(attr, StringAttr):
+            raise VerifyException(
+                f'Operation {op.name} must have a "sym_name" attribute of type '
+                f"`StringAttr` to conform to {SymbolOpInterface.__name__}"
+            )
```
math-fehr, line 238: Maybe you can merge this error with the previous one?
Just to not repeat the call.

math-fehr (approved):
Nice! Thanks!

xdsl/traits.py
PapyChacal: Okay so I'm rolling back on the condition :upside_down_face: Maybe My boolean-jitsu just got weak, not sure

PapyChacal: Okay, that was answer B :smiling_face_with_tear:

PR #1331, opened 2023-07-24
dialects: Add missing Symbol traits.

PapyChacal:
Add all missing direct SymbolOpInterface on implemented operations.
Our implemetations were alright, so no changes needed on anything else than declaring the trait :grin: Will allow more generic handlign of those though!

NB: some operations are "transitively" Symbols, because FunctionOpInterface => Symbol, and we don't have FunctionOpInterface yet. I decided to leave those alone for now!

@mesham I'll wait for your feedback, as this makes a change on the FIR dialect, and I'm not sure if you rely on this in your PsyClone flow? This didn't break any test, but I want to be sure I don't break anything untested before merging this!


compor (approved):
I'm fine with this, but I only checked `llvm`, `memref`, `pdl` and `gpu`.

xdsl/dialects/experimental/fir.py
```diff
@@ -1142,8 +1143,12 @@ class DispatchTable(IRDLOperation):
     """
 
     name = "fir.dispatch_table"
+
+    sym_name = attr_def(StringAttr)
```
math-fehr, line 1147: I guess that in the tests, `sym_name` was used but not required?

math-fehr (approved):
Nice! Thanks!

xdsl/dialects/experimental/fir.py
PapyChacal: I don't know, I'm not sure it's tested tbh! I chatted with @mesham and ensure this won't be an issue short term either way 🙂

PR #1332, opened 2023-07-24
backend: (riscv) Allow detection and exclusion of preallocated registers from register allocation

compor:
This PR adds:

- Detection of  RISCV registers that have been handed out
- Exclusion of the above from the selected allocation strategy
- Pass the commandline option to enable/disable the above (disabled by default)
- Filecheck tests of the above

This is in lieu of passing analysis results from one pass to another since there is no current support ATM.
Relevant Zulip discussion [here][1].

[1]: https://xdsl.zulipchat.com/#narrow/stream/364199-Beginners/topic/.E2.9C.94.20pass.20results


xdsl/backend/riscv/register_allocation.py
```diff
@@ -99,6 +126,14 @@ def _free(self, reg: SSAValue) -> None:
                 available_regs.append(reg_name)
 
     def allocate_registers(self, module: ModuleOp) -> None:
+        if self.exclude_preallocated:
```
superlopuh, line 129: so this will remove the preallocated registers from the set of all available registers, as opposed to just for the values that are live at the same time? Seems a bit coarse as a mechanism, or at least coarser than I expected. Maybe I'm missing the point of this functionality, but it feels like it would be good to add a big doc comment to `self.exclude_preallocated` to explain the objective.

adutilleul (approved):
LGTM.

superlopuh:
Thank you for updating this, it looks great! I'm afraid I caused some more merge conflicts in main, the main one is the addition of a helper for available register logic. In theory the only necessary change should be something like this:
``` python
used_registers = get_used_registers(my_func_op)
allocator.available_int_registers = [reg for reg in allocator.available_int_registers if reg not in used_registers]
allocator.available_float_registers = [reg for reg in allocator.available_float_registers if reg not in used_registers]
```

It would be nice to add some tests for the used register helpers, and maybe make sure it does what we expect for for loops and function arguments also, I don't think this patch handles block arguments, and we might want to do that to be safe.

compor:
> Thank you for updating this, it looks great! I'm afraid I caused some more merge conflicts in main, the main one is the addition of a helper for available register logic. In theory the only necessary change should be something like this:
> 
> ```python
> used_registers = get_used_registers(my_func_op)
> allocator.available_int_registers = [reg for reg in allocator.available_int_registers if reg not in used_registers]
> allocator.available_float_registers = [reg for reg in allocator.available_float_registers if reg not in used_registers]
> ```
> 
> It would be nice to add some tests for the used register helpers, and maybe make sure it does what we expect for for loops and function arguments also, I don't think this patch handles block arguments, and we might want to do that to be safe.

No worries, I'm just resurrecting this incrementally.

As for block args, loops, etc. I think that should be a separate PR, if it is to update this (`LivenessBlockNaive`).
What do you think?

superlopuh:
Your call, I think we'd benefit from having the infrastructure in that we can build on, so happy to merge this once the merge conflicts are resolved.

superlopuh:
I don't really know if all the refactoring is strictly speaking necessary, we could probably get away with having the logic duplicated for a while. I'll not merge my deduplication PRs that are up now and try to open one with the minimal changes required for the liveness allocator.

superlopuh (approved):

xdsl/backend/riscv/register_allocation.py
```diff
@@ -11,6 +12,23 @@
 from xdsl.ir import SSAValue
 
 
+def _gather_allocated(func: riscv_func.FuncOp) -> set[RISCVRegisterType]:
```
superlopuh, line 15: Can you add a pytest for this?

xdsl/backend/riscv/register_allocation.py
```diff
@@ -81,6 +100,15 @@ def _free(self, reg: SSAValue) -> None:
             self.available_registers.push(reg.type)
 
     def allocate_func(self, func: riscv_func.FuncOp) -> None:
+        if self.exclude_preallocated:
+            preallocated = _gather_allocated(func)
+
+            for pa_reg in preallocated:
```
superlopuh, line 106: We probably want to add preallocated registers to `available_registers.reserved_registers` so that we don't accidentally free them

superlopuh (changes requested):
Thinking about it a bit more, I think it would be good to add a couple more bits of functionality in before merging, and some unit tests. Sorry about the flip flop.

xdsl/backend/riscv/register_allocation.py
compor: That's a good suggestion, but I'd argue that we need a different set for tracking this.
I think reserved registers have had a different meaning so far.
What do you think?

superlopuh: maybe it's a docs/naming thing? In the RegisterQueue, functionally, it's the set of registers that cannot be pushed into the queue, which seems to fit the requirements here

superlopuh (approved):
Great, thank you!

Issue #1312, opened 2023-07-20
bug: unwanted behaviour when inheriting from class decorated with `irdl_op_definition`


webmiche:
As discussed in [this PR](https://github.com/xdslproject/xdsl/pull/1305/files/52a821026b808d7bb876f3aa095a5d555073c573#diff-b20593a7043e59a32670bf6c9dfb35daa9ed0719d379b3a53661dcf3b7607fe4) it seems that inheriting from an operation with the `irdl_op_definition`-decorator makes it impossible to add constraints to the child operation.

math-fehr:
Oh, now that you mention it, I'm not 100% sure we should allow inheriting from `irdl_op_definition`.
Inheriting an `IRDLOperation` subclass should not be a problem, it's just the decorator doing a lot of magic that we probably don't want to inherit

math-fehr:
`IRDLOperation` is not marknig operations as `final`, so this will not happen anymore.

PR #1334, opened 2023-07-24
core: Make inheritance work with irdl_op_definition

kingiler:
Address #1312


superlopuh (changes requested):
I'm not sure we want to support subclassing. It might be more useful to raise an error when subclassing from an operation marked with `irdl_op_definition`. This is mostly because there's more things that might subtlely go wrong with annotated subclasses than the verification, and the work to make sure that it goes well doesn't seem justified vs the small gains in API flexibility.

PR #1336, opened 2023-07-24
API: Merge exception cases in SymbolOpInterface.get_sym_attr_name.

PapyChacal:
Forgot to push this before merging #1330..

### DIFF
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -222,7 +222,9 @@ class SymbolOpInterface(OpTrait):
         # import builtin here to avoid circular import
         from xdsl.dialects.builtin import StringAttr
 
-        if "sym_name" not in op.attributes:
+        if "sym_name" not in op.attributes or not isinstance(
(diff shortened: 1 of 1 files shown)


xdsl/traits.py
```diff
@@ -222,7 +222,9 @@ def get_sym_attr_name(self, op: Operation) -> StringAttr | None:
         # import builtin here to avoid circular import
         from xdsl.dialects.builtin import StringAttr
 
-        if "sym_name" not in op.attributes:
+        if "sym_name" not in op.attributes or not isinstance(
+            op.attributes["sym_name"], StringAttr
+        ):
             if self.is_optional_symbol(op):
```
math-fehr, line 228: I think that if you have an optional symbol, and you have `sym_name` of the wrong type, then this will return `None` (and I think we should return an error there).

PapyChacal: Pushed something else, WDYT?

math-fehr (approved):
:+1:

PR #1337, opened 2023-07-24
dialect: Dropping `scf.While.get`

ShaolunWang:
- Fixing a bug where in `scf.While.get` we used
```Python
return While.build(
            operands=operands, result_types=result_types, regions=[before, after])
```
instead of 
```Python
return While.build(
            operands=[operands], result_types=[result_types], regions=[before, after])
```
- Marking `scf.While.get` as deprecated
- Adding respective tests for the bug and `scf.While.__init__()`

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -20,6 +20,8 @@ from xdsl.irdl import (
 from xdsl.traits import HasParent, IsTerminator, SingleBlockImplicitTerminator
 from xdsl.utils.exceptions import VerifyException
 
+from xdsl.utils.deprecation import deprecated
+
 
 @irdl_op_definition
 class While(IRDLOperation):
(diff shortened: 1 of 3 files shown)


xdsl/dialects/scf.py
```diff
@@ -46,6 +58,7 @@ def verify_(self):
                     f"got {self.after_region.block.args[idx].type}"
                 )
 
+    @deprecated("Use While() instead!!")
```
superlopuh, line 61:
```suggestion
    @deprecated("Use While() instead")
```

xdsl/dialects/scf.py
```diff
@@ -30,6 +31,17 @@ class While(IRDLOperation):
     before_region: Region = region_def()
     after_region: Region = region_def()
 
+    def __init__(
+        self,
+        operands: Sequence[SSAValue | Operation],
```
superlopuh, line 36:
```suggestion
        arguments: Sequence[SSAValue | Operation],
```

xdsl/dialects/scf.py
```diff
@@ -30,6 +31,17 @@ class While(IRDLOperation):
...
+        result_types: Sequence[Attribute],
+        before: Region | Sequence[Operation] | Sequence[Block],
```
superlopuh, line 38:
```suggestion
        before_region: Region | Sequence[Operation] | Sequence[Block],
```

xdsl/dialects/scf.py
```diff
@@ -30,6 +31,17 @@ class While(IRDLOperation):
...
+        after: Region | Sequence[Operation] | Sequence[Block],
```
superlopuh, line 39:
```suggestion
        after_region: Region | Sequence[Operation] | Sequence[Block],
```

superlopuh (changes requested):
LGTM modulo minor changes

superlopuh (approved):

math-fehr (approved):

PR #1338, opened 2023-07-24
API: Implement and use the SymbolTable Trait

PapyChacal:
The SymbolTable  trait offer Symbol lookup utilites and ensure Symbol uniqueness amongst its children Symbols.
This implements it, tests it, and use it on implemented operations, that is: `builtin.module`, `gpu.module`, and `irdl.dialect`.
For reference: https://mlir.llvm.org/docs/SymbolsAndSymbolTables/
NB: This documentation states "Described above are Symbols, which reside within a region of an operation defining a SymbolTable." but I could not find any enforcment of this anywhere in the codebase, so I didn't implement it. If anyone finds it, please tell :slightly_smiling_face:

### DIFF
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -9,6 +9,7 @@ from dataclasses import dataclass
 
 import pytest
 
+from xdsl.dialects import test
 from xdsl.dialects.builtin import (
     AnyIntegerAttr,
     IntegerAttr,
@@ -357,12 +358,6 @@ def test_symbol_table():
 
         traits = frozenset([SymbolOpInterface()])
 
-    @irdl_op_definition
-    class TerminatorOp(IRDLOperation):
-        name = "test.terminator"
-
-        traits = frozenset([IsTerminator()])
-
     # Check that having a single region is verified
     op = SymbolTableOp(regions=[Region(), Region()])
     with pytest.raises(
@@ -380,11 +375,13 @@ def test_symbol_table():
     ):
         op.verify()
 
+    terminator = test.TestTermOp([[]], [[]], regions=[[]], successors=[[]])
+
     # Check that symbol uniqueness is verified
     symbol = SymbolOp(attributes={"sym_name": StringAttr("name")})
     dup_symbol = SymbolOp(attributes={"sym_name": StringAttr("name")})
     op = SymbolTableOp(
-        regions=[Region(Block([symbol, dup_symbol, TerminatorOp()])), []]
+        regions=[Region(Block([symbol, dup_symbol, terminator.clone()])), []]
     )
     with pytest.raises(
         VerifyException,
@@ -395,7 +392,7 @@ def test_symbol_table():
     # Check a flat happy case, with symbol lookup
     symbol = SymbolOp(attributes={"sym_name": StringAttr("name")})
 
-    op = SymbolTableOp(regions=[Region(Block([symbol, TerminatorOp()])), []])
+    op = SymbolTableOp(regions=[Region(Block([symbol, terminator.clone()])), []])
     op.verify()
 
     assert SymbolTable.lookup_symbol(op, "name") is symbol
@@ -405,10 +402,10 @@ def test_symbol_table():
     symbol = SymbolOp(attributes={"sym_name": StringAttr("name")})
 
     nested = SymbolTableOp(
-        regions=[Region(Block([symbol, TerminatorOp()])), []],
+        regions=[Region(Block([symbol, terminator.clone()])), []],
         attributes={"sym_name": StringAttr("nested")},
     )
-    op = SymbolTableOp(regions=[Region(Block([nested, TerminatorOp()])), []])
+    op = SymbolTableOp(regions=[Region(Block([nested, terminator.clone()])), []])
     op.verify()
 
     assert SymbolTable.lookup_symbol(op, "name") is None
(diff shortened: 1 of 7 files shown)


tests/test_traits.py
```diff
@@ -324,3 +334,82 @@ class OptionalSymNameOp(IRDLOperation):
     assert interface.is_optional_symbol(symbol)
     symbol.verify()
     assert interface.get_sym_attr_name(symbol) == StringAttr("main")
+
+
+def test_symbol_table():
+    # Some helpers classes
```
superlopuh, line 340:
```suggestion
    # Some helper classes
```

xdsl/dialects/builtin.py
```diff
@@ -199,11 +205,11 @@ class SymbolRefAttr(ParametrizedAttribute):
     def __init__(
         self,
         root: str | StringAttr,
-        nested: list[str] | list[StringAttr] | ArrayAttr[StringAttr] = [],
+        nested: Sequence[str] | Sequence[StringAttr] | ArrayAttr[StringAttr] = [],
     ) -> None:
         if isinstance(root, str):
             root = StringAttr(root)
-        if isinstance(nested, list):
+        if isinstance(nested, Sequence):
```
superlopuh, line 212:
```suggestion
        if not isinstance(nested, ArrayAttr):
```
I'm not sure how exactly Python tests for interface conformance but it seems that testing for the target class directly is a bit simpler

xdsl/traits.py
```diff
@@ -197,6 +197,73 @@ def verify(self, op: Operation) -> None:
...
+
+        if len(op.regions) != 1:
+            raise VerifyException(
+                "Operations with a 'SymbolTable' must have exactly one region"
+            )
+        if len(op.regions[0].blocks) != 1:
+            raise VerifyException(
+                "Operations with a 'SymbolTable' must have exactly one block"
+            )
+        block = op.regions[0].block
```
superlopuh, line 220:
```suggestion
        block = op.regions[0].blocks[0]
```
the block helper does the checking for single block again, seems unnecessary

xdsl/traits.py
```diff
@@ -197,6 +197,73 @@ def verify(self, op: Operation) -> None:
...
+        SymbolTable parent.
+        """
+        # import builtin here to avoid circular import
+        from xdsl.dialects.builtin import StringAttr, SymbolRefAttr
+
+        anchor: Operation | None = op
+        while anchor is not None and not anchor.has_trait(SymbolTable):
+            anchor = anchor.parent_op()
+        if anchor is None:
+            raise ValueError(f"Operation {op} had no SymbolTable ancestor")
```
superlopuh, line 247:
```suggestion
            raise ValueError(f"Operation {op} has no SymbolTable ancestor")
```

xdsl/traits.py
```diff
@@ -197,6 +197,73 @@ def verify(self, op: Operation) -> None:
...
+        if isinstance(name, str | StringAttr):
+            name = SymbolRefAttr(name)
+        for o in anchor.regions[0].block.ops:
+            if (
+                sym_interface := o.get_trait(SymbolOpInterface)
+            ) is not None and sym_interface.get_sym_attr_name(o) == name.root_reference:
+                if not name.nested_references:
+                    return o
+                nested_root = name.nested_references.data[0]
```
superlopuh, line 256:
```suggestion
                nested_root, *nested_references = name.nested_references.data
```
Shouldn't this work? As long as there's at least one element it should be equivalent, and a little more direct

superlopuh (changes requested):
LGTM modulo minor changes, thank you for doing this

xdsl/traits.py
PapyChacal: I think so yes, thanks

superlopuh (approved):

tests/test_traits.py
```diff
@@ -324,3 +334,82 @@ class OptionalSymNameOp(IRDLOperation):
...
+
+        sym_name = attr_def(StringAttr)
+
+        traits = frozenset([SymbolOpInterface()])
+
+    @irdl_op_definition
+    class TerminatorOp(IRDLOperation):
+        name = "test.terminator"
+
+        traits = frozenset([IsTerminator()])
```
math-fehr, line 364: Can we reuse `test.termop` here?

xdsl/traits.py
```diff
@@ -197,6 +197,68 @@ def verify(self, op: Operation) -> None:
                         regions += child_op.regions
 
 
+class SymbolTable(OpTrait):
+    """
+    SymbolTable operations are containers for Symbol operations. They offer lookup
+    functionality for Symbols, and enforce unique symbols amongst its children.
+
+    A SymbolTable operation is constrained to have a sing-block single region.
```
math-fehr, line 205:
```suggestion
    A SymbolTable operation is constrained to have a single single-block region.
```

math-fehr (approved):
Very nice!

tests/test_traits.py
PapyChacal: Thought dump, we probably should have a test.symbol and test.symbol_table too?

math-fehr: We could, and for that there is something I would actually like to try.
So right now, MLIR puts all these methods in a `testd` dialect. But that means that the `testd` dialect is really big and it's really hard to know what's there.
On the other hand, what we could do is have a `test_symbol.py` file, that we could load dynamically (like having something like `xdsl-opt --load test_symbol.py my_file.mlir).
That could be useful to test these kind of things in filecheck.

But otherwise yes, we could put it in the `test` dialect in the meantime, if our goal is to put this in a filecheck test (otherwise, no one else will use it anyway so no need to).

PR #1339, opened 2023-07-25
dialects: (riscv) fix multi-block riscv func lowering

superlopuh:
While we're still not supporting multi-block riscv functions it's possible that the input functions have multiple blocks. This change makes this lowering compatible with that in the future, without functional changes to the code today.

### DIFF
--- a/xdsl/transforms/lower_riscv_func.py
+++ b/xdsl/transforms/lower_riscv_func.py
@@ -64,15 +64,15 @@ class LowerSyscallOp(RewritePattern):
 class LowerRISCVFuncOp(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: riscv_func.FuncOp, rewriter: PatternRewriter):
-        body = op.func_body.block
-        first_op = body.first_op
+        first_block = op.func_body.blocks[0]
+        first_op = first_block.first_op
         assert first_op is not None
-        while len(body.args):
+        while len(first_block.args):
             # arguments are passed to riscv functions via a0, a1, ...
             # replace arguments with `GetRegisterOp`s
-            index = len(body.args) - 1
-            last_arg = body.args[-1]
-            get_reg_op = riscv.GetRegisterOp(f"a{index}")
+            index = len(first_block.args) - 1
+            last_arg = first_block.args[-1]
+            get_reg_op = riscv.GetRegisterOp(riscv.IntRegisterType(f"a{index}"))
             last_arg.replace_by(get_reg_op.res)
             rewriter.insert_op_before(get_reg_op, first_op)
             first_op = get_reg_op
@@ -85,14 +85,16 @@ class LowerRISCVFuncOp(RewritePattern):
 
 class InsertExitSyscallOp(RewritePattern):
     @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: riscv_func.FuncOp, rewriter: PatternRewriter):
-        if op.sym_name.data != "main":
+    def match_and_rewrite(self, op: riscv_func.ReturnOp, rewriter: PatternRewriter):
+        parent_op = op.parent_op()
+        if (
+            not isinstance(parent_op, riscv_func.FuncOp)
+            or parent_op.sym_name.data != "main"
+        ):
             return
-        return_op = op.func_body.block.last_op
-        assert isinstance(return_op, riscv_func.ReturnOp)
 
         EXIT = 93
-        rewriter.insert_op_before(riscv_func.SyscallOp(EXIT), return_op)
+        rewriter.insert_op_before(riscv_func.SyscallOp(EXIT), op)
 
 
 class LowerRISCVFuncReturnOp(RewritePattern):


adutilleul (approved):

xdsl/transforms/lower_riscv_func.py
```diff
@@ -64,15 +64,15 @@ def match_and_rewrite(self, op: riscv_func.SyscallOp, rewriter: PatternRewriter)
...
-        while len(body.args):
+        while len(first_block.args):
             # arguments are passed to riscv functions via a0, a1, ...
             # replace arguments with `GetRegisterOp`s
-            index = len(body.args) - 1
-            last_arg = body.args[-1]
-            get_reg_op = riscv.GetRegisterOp(f"a{index}")
+            index = len(first_block.args) - 1
+            last_arg = first_block.args[-1]
+            get_reg_op = riscv.GetRegisterOp(riscv.IntRegisterType(f"a{index}"))
```
kingiler, line 75: It seems does not consider the case when there are more than 8 function arguments?

xdsl/transforms/lower_riscv_func.py
```diff
@@ -85,14 +85,16 @@ def match_and_rewrite(self, op: riscv_func.FuncOp, rewriter: PatternRewriter):
 
 class InsertExitSyscallOp(RewritePattern):
     @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: riscv_func.FuncOp, rewriter: PatternRewriter):
-        if op.sym_name.data != "main":
+    def match_and_rewrite(self, op: riscv_func.ReturnOp, rewriter: PatternRewriter):
+        parent_op = op.parent_op()
+        if (
+            not isinstance(parent_op, riscv_func.FuncOp)
+            or parent_op.sym_name.data != "main"
```
kingiler, line 92: I don't think checking for parent op is "main" is enough.

xdsl/transforms/lower_riscv_func.py
```diff
@@ -64,15 +64,15 @@ def match_and_rewrite(self, op: riscv_func.SyscallOp, rewriter: PatternRewriter)
 class LowerRISCVFuncOp(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: riscv_func.FuncOp, rewriter: PatternRewriter):
-        body = op.func_body.block
-        first_op = body.first_op
+        first_block = op.func_body.blocks[0]
+        first_op = first_block.first_op
         assert first_op is not None
```
kingiler, line 69: Why asserting `first_op` is not `None`? Could there be a possibility that the first block is empty but subsequent blocks are non-empty for some reason during rewrite?

xdsl/transforms/lower_riscv_func.py
```diff
@@ -64,15 +64,15 @@ def match_and_rewrite(self, op: riscv_func.SyscallOp, rewriter: PatternRewriter)
...
-        while len(body.args):
+        while len(first_block.args):
```
kingiler, line 70: I think it would be more readable to use for loop instead of `while` with extra variables.

Run and tested:
```python
        for index, arg in enumerate(first_block.args):
            get_reg_op = riscv.GetRegisterOp(riscv.IntRegisterType(f"a{index}"))
            rewriter.insert_op_before(get_reg_op, first_op)
            arg.replace_by(get_reg_op.res)
            rewriter.erase_block_argument(arg)
```

kingiler (changes requested):
Some comments.

xdsl/transforms/lower_riscv_func.py
superlopuh: Not allowed by the RISC-V function calling convention

xdsl/transforms/lower_riscv_func.py
superlopuh: it is not allowed to iterate over a collection that is being mutated in Python

xdsl/transforms/lower_riscv_func.py
AntonLydike: Care to elaborate? MLIR does not allow early returns, therefore this should be enough, right? Can you provide a counterexample?

xdsl/transforms/lower_riscv_func.py
AntonLydike: You could clone the list before iterating maybe? `enumerate(tuple(first_block.args))`?

xdsl/transforms/lower_riscv_func.py
superlopuh: I'm not sure what you mean @AntonLydike

xdsl/transforms/lower_riscv_func.py
superlopuh: Good question, I don't remember. I think it might be because we didn't have the necessary alternative API for the same logic, today I'd probably write it a little differently

xdsl/transforms/lower_riscv_func.py
AntonLydike: I was replying to @kingiler, your comment just ninja'd it way in-between, I'll edit it to be more clearer!

superlopuh: Ah yes I see, it's about return being possibly nested. Today we don't have the HasParent trait but we probably should eventually, guaranteeing that it's in the right place.

xdsl/transforms/lower_riscv_func.py
superlopuh: yeah that's a good point, although I'm not a big fan of the temporary allocation... A better way might be to do two loops

xdsl/transforms/lower_riscv_func.py
kingiler: My concern is whether the `riscv_func` allows more than 8 arguments and will rewrite auto spill the excess arguments on stacks.

kingiler (approved):

PR #1341, opened 2023-07-25
interpreter: add implementations for signed/unsigned comparisons

superlopuh:
Adds some helper functions for signed/unsigned comparisons for signless integers.


xdsl/interpreters/comparisons.py
```diff
@@ -0,0 +1,91 @@
...
+def unsigned_upper_bound(bitwidth: int) -> int:
+    return 1 << bitwidth
+
+
+def signed_lower_bound(bitwidth: int) -> int:
+    return -(1 << (bitwidth - 1))
+
+
+def signed_upper_bound(bitwidth: int) -> int:
+    return 1 << (bitwidth - 1)
```
math-fehr, line 45: Can we use `max` and `min` instead?
An upper bound is not really representative of what we want to convey here?
Also, for that reason, we should make the max inclusive (so adding a `-1` on the current upper bound functions).

math-fehr: Also, please document those as well ;)

xdsl/interpreters/comparisons.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def to_unsigned(signless: int, bitwidth: int) -> int:
+    """
+    Transforms values in range [MIN_SIGNED, MAX_UNSIGNED] to range [0, MAX_UNSIGNED].
+    """
+    # Normalise to unsigned range by adding the unsigned range and taking the remainder
+    m = unsigned_upper_bound(bitwidth)
```
math-fehr, line 53: This needs to be changed then, maybe adding a new function that conveys what this `m` should be? (renaming `_upper_bound` to something else)

xdsl/interpreters/comparisons.py
```diff
@@ -0,0 +1,91 @@
...
+    return (signless + m) % m
+
+
+def signless_equal(signless_lhs: int, signless_rhs: int, bitwidth: int) -> bool:
```
math-fehr, line 57: My understanding is that signless should have a single canonical representation in our interpreter. Either it is signed, or it is unsigned.
If we make that decision early on, this should help us a lot I think?

xdsl/interpreters/comparisons.py
```diff
@@ -0,0 +1,91 @@
...
+    negative number will become the larger positive number.
+    """
+    lhs_is_negative = signless_lhs < 0
+    rhs_is_negative = signless_rhs < 0
+    if lhs_is_negative == rhs_is_negative:
+        # Same signedness, normal comparison will work
+        return signless_lhs < signless_rhs
+    else:
+        # Negative number will be greater when bitcast to unsigned
+        return lhs_is_negative
```
math-fehr, line 77: Why do we need this?
Can we instead just use the `to_unsigned` and `to_signed` functions, and then work on that?
Otherwise, we are adding a lot of work to ourselves. This could just have been `to_unsigned(lhs) < to_unsigned(rhs)` ?

xdsl/interpreters/comparisons.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def signed_less_than(lhs: int, rhs: int, bitwidth: int) -> bool:
```
math-fehr, line 80: Same here?

math-fehr (commented):
I'm not completely sure that I agree with the design here.
I feel that if we just add `to_unsigned` and `to_signed`, this should make all code much simpler, instead of having to rewrite things like `unsigned_less_than`?

Also, I feel that if we decide on what a signless integer should look like, then our work could be made simpler (for instance, we could say that a signless integer should be represented with a signed one).

xdsl/interpreters/comparisons.py
superlopuh: Upper bound is what goes into `range`, meaning that it's `max + 1`, I'll document it

xdsl/interpreters/comparisons.py
superlopuh: That's not something possible I don't think. The point of signless integers is that they just have some bitwise representation, but not a value. In the interpreter we only have a value not a representation. That drove the design of this helper file.

xdsl/interpreters/comparisons.py
superlopuh: Yeah I guess we don't. I actually wrote these functions first, and then realised I was missing the conversion functions, so probably best to just use those directly for the time being. I'll swap them out

adutilleul (approved):

kingiler:
I'm not sure but I'm under the impression that for comparison we use semi-open convention?
i.e. `[MIN_SIGNED, MAX_SIGNED)` and `[0, MAX_UNSIGNED]`

tests/interpreters/test_comparisons.py
```diff
@@ -0,0 +1,44 @@
...
+    return u + SIGNED_LOWER_BOUND
+
+
+def test_bitwidth_2_values():
+    """
+    Above calculations are correct for bitwidth 2.
+    """
+    assert len(UNSIGNED_RANGE) == len(SIGNED_RANGE)
+    for i, u in zip(SIGNED_RANGE, UNSIGNED_RANGE):
+        assert i == unsigned_to_signed(u)
```
kingiler, line 28:
```suggestion
    assert list(SIGNED_RANGE) == list(map(unsigned_to_signed, UNSIGNED_RANGE))
```

kingiler (changes requested):

tests/interpreters/test_comparisons.py
superlopuh: this is less useful for unit tests, a list of mismatched elements is harder to compare than a specific failing element in a loop

AntonLydike (approved):

tests/interpreters/test_comparisons.py
superlopuh: Hmm I remember it being less helpful than this, this looks good, I'll change it, thank you

kingiler (approved):

PR #1342, opened 2023-07-25
documentation: (toy) add toy accelerator dialect

superlopuh:
The main idea is to show how to divert the compilation flow to a toy accelerator. This is much simpler than the plan with snitch etc, but still shows the vague idea of matching on affine loops and raising them to a different builtin operation.

As part of the Toy compilation, we know that the accelerator matches all the possible affine loops that we create, meaning that we can compile it end-to-end in python without relying on MLIR.

### DIFF
--- a/docs/Toy/toy/rewrites/lower_toy_affine.py
+++ b/docs/Toy/toy/rewrites/lower_toy_affine.py
@@ -441,17 +441,6 @@ class LowerToAffinePass(ModulePass):
     name = "toy-to-builtin"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
-        # We define the specific operations, or dialects, that are legal targets for this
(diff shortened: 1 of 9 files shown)


docs/Toy/toy/rewrites/lower_toy_affine.py
```diff
@@ -441,17 +441,6 @@ class LowerToAffinePass(ModulePass):
     name = "toy-to-builtin"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
-        # We define the specific operations, or dialects, that are legal targets for this
```
superlopuh, line 444: I initially copied this logic from the Toy tutorial, but it broke when I passed in an MLIR file directly, which ought to work. I don't think we really need to replicate this part of the tutorial for the time being

AntonLydike (approved):

math-fehr (approved):
👍

PR #1346, opened 2023-07-25
interpreter: Initial WGPU interpreter.

PapyChacal:
Here's a first draft/example for a WebGPU-based interpreter for the GPU dialect, including JIT-compilation through WGSL.
I only implement very specific cases of `gpu.func`, `gpu.launch_func`, `gpu.alloc` and `gpu.memcpy` here, to end up with a bearable PR!

NB: contains a minor fix on `gpu.memcpy`, this wasn't an issue anywhere else because the constructor swapped the argument again, but my first definition of this operation was wrong, as the destination is listed *before* the source for this op, as opposed to `memref.copy`.

This requires the addition of Vulkan drivers in the CI


PapyChacal:
Oops, forgot my denial of thys Pyright mess: the wgpu-py package really doesn't play nice with typing at time of writing. Soyunds like a nice PR to do there, but I'll try to fill the holes myself here first

PapyChacal:
Okay, so I'm blocked on this formatting issue, I think my local ruff disagree with the CI one? :upside_down_face: 

More importantly, the test fail because WGPU doesn't find a device to initialize on the CI runner. I'll see if I can fix this, meanwhile feel free to play around locally!

cigarichard (approved):

tests/interpreters/test_wgpu.py
```diff
@@ -0,0 +1,87 @@
...
+    parser = Parser(context, mlir_source)
+    module = parser.parse_module()
+
+    interpreter = Interpreter(module)
+    interpreter.register_implementations(ArithFunctions())
+    interpreter.register_implementations(MemrefFunctions())
+    interpreter.register_implementations(WGPUFunctions())
+    interpreter.register_implementations(PrintfFunctions())
+    f = StringIO("")
+    with redirect_stdout(f):
```
superlopuh, line 82: A) I didn't know this was possible, pretty cool
B) The interpreter has an optional StringIO that the interpreter functions should use for this purpose

xdsl/interpreters/experimental/wgpu.py
```diff
@@ -0,0 +1,214 @@
...
+
+    @impl(gpu.AllocOp)
+    def run_alloc(
+        self, interpreter: Interpreter, op: gpu.AllocOp, args: tuple[Any, ...]
+    ):
+        """
+        Allocate a GPUBuffer according to a gpu.alloc operation, return it as the memref
+        value.
+        """
+        if len(args) != 0 or op.asyncToken:
```
superlopuh, line 94:
```suggestion
        if args or op.asyncToken:
```

xdsl/interpreters/experimental/wgpu.py
```diff
@@ -0,0 +1,214 @@
...
+        return ()
+
+    @impl(gpu.LaunchFuncOp)
+    def run_launch_func(
+        self, interpreter: Interpreter, op: gpu.LaunchFuncOp, args: tuple[Any, ...]
+    ):
+        """
+        Launch a GPU kernel through the WebGPU API.
+        """
+        if op.asyncToken is not None or len(op.asyncDependencies) != 0:
```
superlopuh, line 155:
```suggestion
        if op.asyncToken is not None or op.asyncDependencies:
```

xdsl/interpreters/experimental/wgpu.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+        kernel_operands = op.kernelOperands
+
+        func = SymbolTable.lookup_symbol(op, op.kernel)
+        assert isinstance(func, gpu.FuncOp)
+        WGPUFunctions.compile_func(self, func)
+        shader_module = self.shader_modules[func]
+
+        # Compute the dispatch number
+        dispatch = list(gridSize)
```
superlopuh, line 172: Looks like we don't really need the list construction, does this work?
```suggestion
        dispatch = gridSize
```

superlopuh (changes requested):
Exciting, LGTM modulo small comments

tests/interpreters/test_wgpu.py
PapyChacal: This is copy pasta from other tests of ours 🙂
Will see if I can use this other parameter

xdsl/interpreters/experimental/wgpu.py
PapyChacal: Nope, tuple is immutable

superlopuh (approved):

.github/workflows/ci-mlir.yml
```diff
@@ -25,6 +25,12 @@ jobs:
     - uses: actions/checkout@v3
       with:
         path: xdsl
+        
+    - name: Install native dependencies
+      uses: awalsh128/cache-apt-pkgs-action@latest
+      with:
+        packages: mesa-vulkan-drivers
+        version: 1.0
```
webmiche, line 33: Personally, I would prefer to separate our testing environments a bit. Otherwise, we might run into versioning issues were certain packages are required, but not built with just the default config.

For that reason, I think I would have prefered this to be its own separate CI run.

What do others think? I guess our requirements-optional.txt and what is needed where is a big mess already, so this might not just be an issue with these tests, but something deeper. tagging @math-fehr @superlopuh and @PapyChacal here :)

PapyChacal: I would love to look into structuring those things for sure, but probably later I. August as we all have deadlines running currently 🙂

superlopuh: I would be interested in splitting the requirements-optional per subproject, with `extras` installing them all.

PR #1349, opened 2023-07-26
dialect: Deprecating `.get` in llvm dialect

ShaolunWang:
Making respective changes from .get to __init__ in mpi lowering and llvm/mpi tests

### DIFF
--- a/xdsl/dialects/llvm.py
+++ b/xdsl/dialects/llvm.py
@@ -52,6 +52,7 @@ from xdsl.irdl import (
 from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
 from xdsl.traits import IsTerminator, SymbolOpInterface
+from xdsl.utils.deprecation import deprecated
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import isa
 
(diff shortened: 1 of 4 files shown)


xdsl/dialects/llvm.py
```diff
@@ -455,6 +456,54 @@ class GEPOp(IRDLOperation):
...
+            attrs["inbounds"] = UnitAttr()
+
+        super().__init__(
+            operands=[ptr, ssa_indices], result_types=[result_type], attributes=attrs
+        )
+
+    @deprecated("Use GEPOp(...) instead")
     @staticmethod
     def get(
         ptr: SSAValue | Operation,
```
AntonLydike, line 509: Can you make this call `__init__` instead of duplicating code?

xdsl/dialects/llvm.py
```diff
@@ -548,6 +597,25 @@ class AllocaOp(IRDLOperation):
...
+            attrs["elem_type"] = elem_type
+        else:
+            ptr_type = LLVMPointerType.typed(elem_type)
+
+        super().__init__(operands=[size], attributes=attrs, result_types=[ptr_type])
+
+    @deprecated("Use Alloca(...) instead")
     @staticmethod
     def get(
         size: SSAValue | Operation,
```
AntonLydike, line 621: Same here, would be nice to make this just pass everything to `__init__`...

AntonLydike (changes requested):
I think this is much easier to review once you make `get` just call `__inti__`... Otherwise it looks good! Thanks!

kingiler (approved):

AntonLydike (approved):
Perfect! Thanks!

PR #1350, opened 2023-07-26
dialects: (riscv-scf) Add riscv_scf.for and yield ops

AntonLydike:
This adds the `rvscf.for` and `rvscf.yield` ops to the new `rvscf` dialect, which we want to use to do register allocation at higher levels

### DIFF
--- a/xdsl/dialects/rvscf.py
+++ b/xdsl/dialects/rvscf.py
@@ -10,7 +10,6 @@ from xdsl.irdl import (
     Operand,
     operand_def,
     var_operand_def,
-    AnyAttr,
     VarOperand,
     SSAValue,
     Operation,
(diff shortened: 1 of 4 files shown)


xdsl/dialects/rvscf.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self, *operands: SSAValue | Operation):
+        super().__init__(operands=[SSAValue.get(operand) for operand in operands])
+
+
+@irdl_op_definition
+class ForOp(IRDLOperation):
+    name = "rvscf.for"
+
+    lb: Operand = operand_def(RISCVRegisterType)
```
superlopuh, line 43:
```suggestion
    lb: Operand = operand_def(IntRegisterType)
```

adutilleul (approved):

PR #1351, opened 2023-07-27
dialects: (riscv) register allocation on scf level IR

AntonLydike:
Very naive and simple register allocation on an scf example. Very rough draft version for now, just to convey what we want to be doing.

Feel free to review as well @tobiasgrosser @adutilleul


xdsl/transforms/rvscf_regalloc.py
```diff
@@ -35,31 +36,48 @@ def loop_body(args: tuple[BlockArgument, ...]):
...
-            vals_to_regs=self.vals_to_regs.copy(),
-            used_regs=self.used_regs.copy(),
-        )
+        # TODO: we need to feed some updated stats back to the thing when we ascend out
+        #       of the region?
+        return RegAllocCtx(liveliness=self.liveliness.copy())
 
-    def free_reg(self):
+    def free_reg(self) -> riscv.IntRegisterType:
         for reg in [
```
superlopuh, line 61:
```suggestion
        for reg in chain(riscv.Registers.A, riscv.Registers.T, riscv.Registers.S):
```

nazavode (approved):

tests/filecheck/backend/rvscf_register_allocation.mlir
```diff
@@ -0,0 +1,115 @@
...
+  riscv_func.func @foo(%arg0 : !riscv.reg<a0>) {
+    %0 = riscv.li 0 : () -> !riscv.reg<>
+    %2 = riscv.li 0 : () -> !riscv.reg<>
+    %3 = riscv.li 1 : () -> !riscv.reg<>
+    %5 = "riscv_scf.for"(%2, %arg0, %3, %0) ({
+    ^1(%arg1 : !riscv.reg<>, %arg2 : !riscv.reg<>):
+      %6 = riscv.li 0 : () -> !riscv.reg<>
+      %7 = riscv.li 1 : () -> !riscv.reg<>
+      %9 = "riscv_scf.for"(%6, %arg0, %7, %arg2) ({
+      ^2(%arg3 : !riscv.reg<>, %arg4 : !riscv.reg<>):
```
JosseVanDelm, line 85: I find this `^1` and `^2` (xdsl) still a bit confusing, since technically they are both their regions `^bb0`(mlir), right?

xdsl/backend/riscv/register_rvscf_regalloc.py
```diff
@@ -0,0 +1,201 @@
...
+            *riscv.Registers.FS,
+        ):
+            if reg not in self.forbidden_vals and self.liveliness[reg] == 0:
+                return reg
+        raise NotImplementedError("Out of float registers, spilling not implemented")
+
+    def add_reg(self, val: SSAValue):
+        assert isinstance(val.type, riscv.RISCVRegisterType) and val.type.is_allocated
+        if self.liveliness[val.type] > 0:
+            # this may be interesting behaviour? idk
```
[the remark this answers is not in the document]
JosseVanDelm: Also don't know what this means

xdsl/backend/riscv/register_rvscf_regalloc.py
```diff
@@ -0,0 +1,201 @@
...
+        if self.liveliness[val.type] > 0:
+            self.liveliness[val.type] -= 1
+
+    def make_forbidden(self, reg: riscv.RISCVRegisterType):
+        self.forbidden_vals.add(reg)
+
+
+def register_allocate_function(func: riscv_func.FuncOp):
+    # create an empty context
+    ctx = RegAllocCtx()
```
JosseVanDelm, line 93: Maybe irrelevant right now, but will/should this still work for nested functions?

JosseVanDelm (approved):
Interesting to see where this is going!

tests/filecheck/backend/rvscf_register_allocation.mlir
superlopuh: these are basic blocks not regions, but yeah they could have the same name. I think we would do well to add custom syntax here, as then we wouldn't need the basic block argument printing.

xdsl/backend/riscv/register_rvscf_regalloc.py
```diff
@@ -0,0 +1,201 @@
...
+    it is "free" again.
+
+    It also carries a list of forbidden registers, as they are used by the "outside"
+    (i.e. parent region) and shall not be overwritten.
+
+    This is currently still a bit broken, as ssa values can have the same register
+    even before.
+    """
+
+    liveliness: dict[riscv.RISCVRegisterType, int] = field(
```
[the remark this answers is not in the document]
adutilleul: This mapping is only for allocated values (by assertion and construction), hence we don't run into the aforementioned issue. But you're right the naming and the documentation could me clearly improved to convey that.

tests/filecheck/backend/rvscf_register_allocation.mlir
JosseVanDelm: Yes I meant basic blocks, sorry.
But I agree this is not for this PR, either by fixing the way xdsl prints BB's or indeed by adding custom syntax :wink:

tobiasgrosser:
@superlopuh and @adutilleul, what is the plan for this PR?

xdsl/backend/riscv/register_rvscf_regalloc.py
```diff
@@ -0,0 +1,201 @@
...
+)
+
+
+@dataclass
+class RegAllocCtx:
+    """
+    This is the context object for allocating registers.
+
+    It keeps track of how many times each ssa value is expected to be seen before
+    it is "free" again.
```
tobiasgrosser, line 23: @AntonLydike and @adutilleul, do we indeed need a context object. AFAIU, if we allocate backwards we do not need to maintain use counts, no?

xdsl/backend/riscv/register_rvscf_regalloc.py
```diff
@@ -0,0 +1,201 @@
...
+        for op in block.ops:
+            if isinstance(
+                op,
+                riscv.RISCVOp,
+            ):
+                for result in op.results:
+                    assert isinstance(result.type, riscv.RISCVRegisterType)
+                    # keep track of the "alive" registers
+                    for operand in op.operands:
+                        ctx.register_use(operand)
```
tobiasgrosser, line 156: Is there a reason why this register allocator uses a different strategy for Blocks than `RegisterAllocatorLivenessBlockNaive`? I had hoped that our scf allocator would just be an extended version of `RegisterAllocatorLivenessBlockNaive`. Is there a reason we choose a different approach here?

superlopuh:
As per our discussions, closing this in favour of my stack of PRs

PR #1354, opened 2023-07-27
CI: Cache pip.

PapyChacal:
Try caching pip to speed up most workflows.


PapyChacal:
Mh, speeds up installation from 26s to 23s, definetely not worth the added complexity :sweat_smile:

PR #1355, opened 2023-07-27
dialects: (riscv) Add lowering from register-allocated rvscf to riscv labels

AntonLydike:
Whole pipeline runs: `python xdsl/transforms/rvscf_regalloc.py | xdsl-opt -p lower-riscv-scf-to-labels,lower-riscv-func -t riscv-asm`

This takes a small example, register allocates, lowers the scf to labels, and then prints assembly. Very rough around the edges.

When we elide unused moves and fold `li, add -> addi` we have one instruction less than gcc on `-O3`, whatever that's worth^^

Lots of improvement potential here as usual. 

@tobiasgrosser you may find this interesting.


xdsl/backend/riscv/riscv_scf_to_asm.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+        # Append the loop increment and the branch back to the loop header
+        # to the end of the body.
+
+        body = op.body.detach_block(0)
+        yield_op = body.last_op
+        assert isinstance(yield_op, riscv_scf.YieldOp)
+
+        body.insert_ops_after(
```
superlopuh, line 72: Is there a way to use the rewriter for this? Maybe before detaching the block?

superlopuh (changes requested):
Looks great overall, some small things to discuss. I would love @tobiasgrosser and @compor's reviews on this PR before merging

nazavode (approved):

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
...
+        %2 = riscv.mv %0 : (!riscv.reg<a0>) -> !riscv.reg<a0>
+        %3 = riscv.mv %1 : (!riscv.reg<a1>) -> !riscv.reg<a1>
+        %4 = riscv.li 1 : () -> !riscv.reg<a2>
+        %5 = riscv.li 0 : () -> !riscv.reg<a3>
+        %6 = riscv.fcvt.s.w %5 : (!riscv.reg<a3>) -> !riscv.freg<fa0>
+        %7 = "riscv_scf.for"(%2, %3, %4, %6) ({
+        ^1(%8 : !riscv.reg<a0>, %9 : !riscv.freg<fa0>):
+        %10 = riscv.fcvt.s.w %8 : (!riscv.reg<a0>) -> !riscv.freg<fa1>
+        %11 = riscv.fadd.s %9, %10 : (!riscv.freg<fa0>, !riscv.freg<fa1>) -> !riscv.freg<fa0>
+        "riscv_scf.yield"(%11) : (!riscv.freg<fa0>) -> ()
```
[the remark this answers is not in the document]
kingiler: Can use the output from `xdsl-opt tests/filecheck/backend/rvscf_lowering_labels.mlir --print-op-generic`

xdsl/backend/riscv/riscv_scf_to_asm.py
```diff
@@ -0,0 +1,106 @@
...
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+_REG_TYPES = riscv.IntRegisterType | riscv.FloatRegisterType
+
+
+def get_register_ops_from_values(
```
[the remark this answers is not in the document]
adutilleul: Inlining it requires a bit more code (duplication), I''m fine with either.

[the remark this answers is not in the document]
superlopuh: Let's leave it as-is

superlopuh (approved):

tests/filecheck/with-riscemu/rvscf_lowering_emu.mlir
```diff
@@ -0,0 +1,22 @@
...
+    %4 = "riscv_scf.for"(%0, %1, %2, %3) ({
+    ^0(%5 : !riscv.reg<a0>, %6 : !riscv.reg<a3>):
+      %7 = riscv.add %5, %6 : (!riscv.reg<a0>, !riscv.reg<a3>) -> !riscv.reg<a3>
+      "riscv_scf.yield"(%7) : (!riscv.reg<a3>) -> ()
+    }) : (!riscv.reg<a0>, !riscv.reg<a1>, !riscv.reg<a2>, !riscv.reg<a3>) -> !riscv.reg<a3>
+    %8 = riscv.mv %4 : (!riscv.reg<a3>) -> !riscv.reg<a0>
+    "riscv.custom_assembly_instruction"(%4) {"instruction_name" = "print"} : (!riscv.reg<a3>) -> ()
+    riscv.li 93 : () -> !riscv.reg<a7>
+    riscv.ecall : () -> ()
+    riscv.ret : () -> ()
```
kingiler, line 18: I'm not sure about the use of `custom_assembly_instruction` here?

adutilleul: As far as I know, this is the only way of getting the output from `riscemu` in the current filecheck infrastructure, see the other tests in the folder for reference.

kingiler: I see. 93 is syscall for exit instead of printing.

kingiler (approved):

compor:
> Looks great overall, some small things to discuss. I would love @tobiasgrosser and @compor's reviews on this PR before merging

will do tomorrow; no time today I'm afraid.

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
+// RUN: xdsl-opt -p lower-riscv-scf-to-labels --split-input-file %s | filecheck %s
+
+builtin.module {
+    "riscv_func.func"() ({
+    ^0(%0 : !riscv.reg<a0>, %1 : !riscv.reg<a1>):
+        %2 = riscv.mv %0 : (!riscv.reg<a0>) -> !riscv.reg<a0>
+        %3 = riscv.mv %1 : (!riscv.reg<a1>) -> !riscv.reg<a1>
```
tobiasgrosser, line 7: I am surprised that the input test case uses `riscv.mv` instructions? I feel these just are not needed and could be removed to reduce noise? Or do I miss sth here?

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
...
+        %4 = riscv.li 1 : () -> !riscv.reg<a2>
+        %5 = riscv.li 0 : () -> !riscv.reg<a3>
+        %6 = riscv.fcvt.s.w %5 : (!riscv.reg<a3>) -> !riscv.freg<fa0>
```
tobiasgrosser, line 10: Can we make this an integer loop? That would allow removing the fcvt instructions and would result in a more straightforward loop body. Also, given that floating-point is an extension over the base instruction set, maybe we should avoid using it in the base test cases for features that can work without floating point support

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
...
+        %7 = "riscv_scf.for"(%2, %3, %4, %6) ({
```
tobiasgrosser, line 11: I feel riscv_scf.for could really benefit from custom formatting, such that it is clear what the lb and upper bound are.

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
...
+// CHECK-NEXT:    ^0(%0 : !riscv.reg<a0>, %1 : !riscv.reg<a1>):
+// CHECK-NEXT:      %2 = riscv.mv %0 : (!riscv.reg<a0>) -> !riscv.reg<a0>
+// CHECK-NEXT:      %3 = riscv.mv %1 : (!riscv.reg<a1>) -> !riscv.reg<a1>
+// CHECK-NEXT:      %4 = riscv.li 1 : () -> !riscv.reg<a2>
+// CHECK-NEXT:      %5 = riscv.li 0 : () -> !riscv.reg<a3>
+// CHECK-NEXT:      %6 = riscv.fcvt.s.w %5 : (!riscv.reg<a3>) -> !riscv.freg<fa0>
+// CHECK-NEXT:      %7 = riscv.get_register : () -> !riscv.reg<a0>
+// CHECK-NEXT:      riscv.label "scf_cond_0_for" ({
+// CHECK-NEXT:      }) : () -> ()
+// CHECK-NEXT:      riscv.bge %7, %3, "scf_body_end_0_for" : (!riscv.reg<a0>, !riscv.reg<a1>) -> ()
```
tobiasgrosser, line 33: I guess the riscv.get_register() is how we move SSA-values around branches. That's an interesting approach that is different to what we did in ChocoPy. It mixes SSA-semantics within BBs and side-effectful semantics. I guess it also means that all instructions that write to register-allocated types are suddenly side-effectful as they will have write effects. Has this been documented somewhere? I feel getting this consistent might be difficult.

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
...
+            ^2(%arg3 : !riscv.reg<a4>, %arg4 : !riscv.reg<a1>):
+                %7 = riscv.add %arg1, %arg3 : (!riscv.reg<a2>, !riscv.reg<a4>) -> !riscv.reg<a0>
+                %8 = riscv.add %arg4, %7 : (!riscv.reg<a1>, !riscv.reg<a0>) -> !riscv.reg<a1>
+                "riscv_scf.yield"(%8) : (!riscv.reg<a1>) -> ()
+            }) : (!riscv.reg<a4>, !riscv.reg<a0>, !riscv.reg<a5>, !riscv.reg<a1>) -> !riscv.reg<a1>
+            "riscv_scf.yield"(%6) : (!riscv.reg<a1>) -> ()
+        }) : (!riscv.reg<a2>, !riscv.reg<a0>, !riscv.reg<a3>, !riscv.reg<a1>) -> !riscv.reg<a1>
+        "riscv_func.return"(%3) : (!riscv.reg<a1>) -> ()
+    }) {"sym_name" = "foo"} : () -> ()
+}
```
tobiasgrosser, line 72: Nice that there is an example with nesting. I like it. I would like to see an example with frep, but maybe in a separate PR?

tests/filecheck/backend/rvscf_lowering_labels.mlir
```diff
@@ -0,0 +1,112 @@
...
+// CHECK-NEXT:      riscv.blt %7, %3, "scf_body_0_for" : (!riscv.reg<a0>, !riscv.reg<a1>) -> ()
+// CHECK-NEXT:      riscv.label "scf_body_end_0_for" ({
+// CHECK-NEXT:      }) : () -> ()
+// CHECK-NEXT:      %13 = riscv.get_float_register : () -> !riscv.freg<fa0>
+// CHECK-NEXT:      %14 = riscv.fcvt.w.s %13 : (!riscv.freg<fa0>) -> !riscv.reg<a0>
+// CHECK-NEXT:      "riscv_func.return"(%14) : (!riscv.reg<a0>) -> ()
+// CHECK-NEXT:    }) {"sym_name" = "foo"} : () -> ()
+// CHECK-NEXT:  }
+
+// -----
```
tobiasgrosser, line 50: Is there a reason we are using --split-input-file? AFAIU this makes it harder to run individual test cases with `xdsl-opt`, no?

tobiasgrosser (commented):
Hi @superlopuh, thank you for this PR. I think this level of PR is great to get things moving. I added some review comments. Please don't feel blocked by them, as I am a little bit out of this codebase so just want to add some ideas. Some of them might be better considered in subsequent PRs. Having a baseline test case of what we do today may already be useful.

tests/filecheck/backend/rvscf_lowering_labels.mlir
superlopuh: This is an artefact of the register allocation approach. The idea is that the calling convention requires the funciton arguments to be in `a{k}` but we might want to reuse those registers inside the function. So the first thing we do is move them to a new register, which might end up being the same one, in which case the move will be eliminated during canonicalization.

tests/filecheck/backend/rvscf_lowering_labels.mlir
superlopuh: Good idea, I filed an issue #1454

tests/filecheck/backend/rvscf_lowering_labels.mlir
superlopuh: Yep, let's do this in another PR, we still don't have all the operations defined that we would need for that test, I don't think

tests/filecheck/backend/rvscf_lowering_labels.mlir
superlopuh: It's a tradeoff, using `--split-input-file` lets us reuse the loaded instance of the xdsl module, which takes hundreds of ms to import, which adds up, but more importantly it grous the tests by semantical meaning, so I find it a bit easier to navigate. You're right that it has its drawbacks also.

tests/filecheck/backend/rvscf_lowering_labels.mlir
superlopuh: I'll add another test with an integer loop in another PR, I think it would be good to keep both.

superlopuh:
Thank you everyone, merging

tests/filecheck/backend/rvscf_lowering_labels.mlir
tobiasgrosser: I feel it would be useful to drop this leftover artefacts from register allocation in this test case. This will make the test easier to understand.

superlopuh: https://github.com/xdslproject/xdsl/pull/1456

Issue #1218, opened 2023-06-30
Circular dependency between different classes through trait usage


compor:
Currently, a circular dependency between operations via the use of traits cannot be resolved.
Example using `HasParent` and `SingleBlockImplicitTerminator` traits:

```python
@irdl_op_definition
class FooTerminatorOp(IRDLOperation):
    name = "test.foo_terminator"

    # circular reference
    traits = frozenset([HasParent(FooOp), IsTerminator()])


@irdl_op_definition
class FooOp(IRDLOperation):
    name = "test.foo"
 
    # circular reference
    traits = frozenset([SingleBlockImplicitTerminator(FooOp)])
```

math-fehr:
Just as a note, this is very related to #12, which is about circular dependencies in operands/results.
I may try to look into it soon!

compor:
@math-fehr I didn't mean to create issue bloat. I can move this description to that issue and close this one if you want.

superlopuh:
I would vote for keeping both issues separate.

PR #1357, opened 2023-07-27
core: (trait) Implement LazyTrait to workaround circular dependency

kingiler:
Address #1218 

`LazyTrait` is implemented to postpone evaluation of provided traits.

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -78,10 +78,12 @@ class Yield(IRDLOperation):
     name = "scf.yield"
     arguments: VarOperand = var_operand_def(AnyAttr())
 
-    # TODO circular dependency disallows this set of traits
-    # tracked by gh issues https://github.com/xdslproject/xdsl/issues/1218
-    # traits = frozenset([HasParent((For, If, ParallelOp, While)), IsTerminator()])
-    traits = frozenset([IsTerminator()])
+    traits = frozenset(
+        [
+            HasParent((lambda: For, lambda: If, lambda: ParallelOp, lambda: While)),
+            IsTerminator(),
+        ]
+    )
 
     @staticmethod
     def get(*operands: SSAValue | Operation) -> Yield:
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -17,7 +17,12 @@ from xdsl.irdl import (
     var_operand_def,
     var_result_def,
 )
-from xdsl.traits import HasParent, IsTerminator, SingleBlockImplicitTerminator
+from xdsl.traits import (
+    HasParent,
+    IsTerminator,
+    LazyTrait,
+    SingleBlockImplicitTerminator,
+)
 from xdsl.utils.deprecation import deprecated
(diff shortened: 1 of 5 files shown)


PapyChacal (commented):
Oh wow, this still looks hacky, but is somehow minimally intrusive I guess?
I'm all for it personally, I'd rather have a functional hack in 🙂

That said, this makes me want to try wrapping it in a utility function though, which could show intent
Say `HasParent(forward(scf.For))`; do you think it would work? Also wondering What others think

Either way, thanks a lot!

PapyChacal (approved):

PapyChacal:
And I think it could work at the OpTrait level rather than a specific trait?

math-fehr (commented):
Nice!
I'm wondering if it would be better to do that directly in `traits`?
So that way, we get that for all traits for free, instead of having it here only for the `HasParent` trait?

xdsl/dialects/scf.py
```diff
@@ -78,10 +78,12 @@ class Yield(IRDLOperation):
     name = "scf.yield"
     arguments: VarOperand = var_operand_def(AnyAttr())
 
-    # TODO circular dependency disallows this set of traits
-    # tracked by gh issues https://github.com/xdslproject/xdsl/issues/1218
-    # traits = frozenset([HasParent((For, If, ParallelOp, While)), IsTerminator()])
-    traits = frozenset([IsTerminator()])
+    traits = frozenset(
+        [
+            HasParent((lambda: For, lambda: If, lambda: ParallelOp, lambda: While)),
```
superlopuh, line 83:
```suggestion
            HasParent((lambda: (For, If, ParallelOp, While))),
```

superlopuh: Would the following work? It feels more readable to me

superlopuh (commented):
This will work, but I wonder if we would benefit from a more general mechanism to separate the logic. I'm thinking something like a `LazyTrait(lambda: AnyTraitYouWant())`, which would let us test the logic separately, reuse it for other circular dependencies, and avoid the frozen workaround. Could you please give it a go and check whether it would actually simplify the code a bit?

superlopuh:
Ah I see others had the same idea :)

kingiler:
Do you want all traits to be lazy? We can have the implementation allows:
```python3
    traits = frozenset(
        LazyTrait(
            lambda: (
                IsolatedFromAbove(),
                SingleBlockImplicitTerminator(ModuleEndOp),
                SymbolOpInterface(),
                SymbolTable(),
            )
        )
    )
```

xdsl/traits.py
```diff
@@ -41,15 +43,31 @@ class HasParent(OpTrait):
...
+        if self._unresolved:
+            object.__setattr__(
+                self,
+                "parameters",
+                tuple(
+                    parameter if isclass(parameter) else parameter()
+                    for parameter in self.parameters
+                ),
+            )
+            object.__setattr__(self, "_unresolved", True)
```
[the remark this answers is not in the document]
kingiler: No longer have hacky ways in newest commit.

PR #1358, opened 2023-07-27
dialects: Adding integer width arith ops

ajrichins:
adding arith operations TruncIOp, ExtSIOp and ExtUIOp.

@Hatsunespica


tests/dialects/test_arith.py
```diff
@@ -245,3 +248,18 @@ def test_cmpi_index_type():
...
     Cmpi.get(a, b, "eq").verify()
+
+
+def test_extend_truncate_iops():
+    a = Constant.from_int_and_width(1, i32)
+    b = Constant.from_int_and_width(2, i64)
+    exts_op = ExtSIOp.get(a, i64)
+    extu_op = ExtUIOp.get(a, i64)
+    trunc_op = TruncIOp.get(b, i32)
+
```
math-fehr, line 259: Could you just add `exts_op.verify()`, and same for the two others?

math-fehr (approved):
Perfect, thanks a lot!
I just added a comment on the test, but otherwise that looks good to me!

xdsl/dialects/arith.py
```diff
@@ -619,6 +619,42 @@ def get(op: SSAValue | Operation, target_type: AnyFloat):
...
 
+@irdl_op_definition
+class TruncIOp(IRDLOperation):
+    name = "arith.trunci"
+
+    input: Operand = operand_def(IntegerType)
+    result: OpResult = result_def(IntegerType)
+
+    @staticmethod
+    def get(op: SSAValue | Operation, target_type: IntegerType):
```
superlopuh, line 630: Can you please use inits? We haven't migrated every file yet but I feel that new operations should use initialisers

math-fehr: Oh yes! The only change would be to have `def __init__(self, ...)` instead of `get` and the `staticmethod`!

math-fehr (approved):
Perfect, thanks a lot!

PR #1359, opened 2023-07-28
core: (trait) Implement Lazy Class to workaround circular dependency

kingiler:
Address #1218

~~`LazyTrait` is implemented to postpone evaluation of provided traits.~~
`Lazy` class is implemented to postpone evaluation of provided objects.


xdsl/traits.py
```diff
@@ -32,6 +33,23 @@ def verify(self, op: Operation) -> None:
...
+    """A trait that postpone evaluation of other traits. Can be used to workaround circular dependency."""
+
+    parameters: tuple[Callable[[], OpTrait | tuple[OpTrait, ...]], ...]
+
+    def __init__(self, *parameters: Callable[[], OpTrait | tuple[OpTrait, ...]]):
+        super().__init__(parameters)
+
+    def verify(self, op: Operation) -> None:
+        for func in self.parameters:
+            traits = func()
```
superlopuh, line 46: I think we might as well cache this to `self._traits`

superlopuh (approved):
Excellent, thank you

math-fehr (changes requested):
The current solution would not work with `has_traits`.
Can we instead make the entirety of `traits` lazy? I feel that this would reduce the complexity overall.
We would write `traits = lazy(lambda: frozenset([...]))`, and any time we would access `traits`, it would compute the traits and cache the results.
I think it might be possible to do that with a `lazy` function that cheat at pyright level, but we could also probably do that with descriptors https://docs.python.org/3/howto/descriptor.html (I'm just not sure if they are available for class fields)

xdsl/traits.py
kingiler: Cached in recent commit.

xdsl/ir/core.py
```diff
@@ -583,7 +583,7 @@ class Operation(IRNode):
     _prev_op: Operation | None = field(default=None, repr=False)
     """Previous operation in block containing this operation."""
 
-    traits: ClassVar[frozenset[OpTrait]]
+    traits: ClassVar[LazyTrait]
```
AntonLydike, line 586: Not sure I'm on board with this?

superlopuh:
Hmm, I kind of liked it more before. I'm not sure exactly what implementation @math-fehr had in mind, but it was nice that the thing stayed a frozenset, and that only lazy things had to be lazy

kingiler:
I'm not sure if you mean the implementation in #1375? It would be minimal and less intrusive, yet more generic with dirty hacks.

math-fehr:
@superlopuh What I have in mind is:
* We define traits with something like `traits = traits_def()`, or `traits = OpTraits(...)`
* We can either pass a `lambda` to `OpTraits`, or pass directly a `Sequence` of traits. In the end, it will always be a `frozenset` in the background, so that's fine
* I don't think a generic `Lazy` descriptor would be useful here, since we want to have a sequence of elements. So I would just have a custom class here for that.

math-fehr:
Closing this in favor of #1529

PR #1361, opened 2023-07-28
misc: ignore wgpu interpreter file

superlopuh:
The dependency has linting issues, I don't think we care about them

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,7 @@ typeCheckingMode = "strict"
     "tests/test_frontend_op_resolver.py",
     "tests/test_frontend_python_code_check.py",
     "tests/dialects/test_memref.py",
+    "xdsl/interpreters/experimental/wgpu.py",
 ]
 "ignore" = [
     "tests/filecheck/frontend/dialects/builtin.py",


pyproject.toml
```diff
@@ -9,6 +9,7 @@ typeCheckingMode = "strict"
     "tests/test_frontend_op_resolver.py",
     "tests/test_frontend_python_code_check.py",
     "tests/dialects/test_memref.py",
+    "xdsl/interpreters/experimental/wgpu.py",
```
superlopuh, line 12: eclude means the errors will still be there if you open the file in your IDE, but not otherwise

math-fehr (approved):
Nice, this would solve most of the remaining errors I have on the pyright bump.
Any reason why we didn't get errors before?

math-fehr:
Actually wait, I don't see them locally anymore, and actually I don't see it in the CI?

PapyChacal (approved):

superlopuh:
You probably accidentally updated your vscode and are now seeing the pylance errors from 318, I think that's what's happening to me locally and is why Emilien did not see them on his machine

PR #1362, opened 2023-07-28
interpreter: (riscv) add interpreter functions

superlopuh:
- [x] add tests for get_data

### DIFF
--- a/xdsl/interpreters/riscv.py
+++ b/xdsl/interpreters/riscv.py
@@ -1,0 +1,217 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable, Generic, Iterator, TypeAlias, TypeVar
+
+from xdsl.dialects import riscv
+from xdsl.dialects.builtin import AnyIntegerAttr, IntegerAttr, ModuleOp
+from xdsl.interpreter import (
+    Interpreter,
+    InterpreterFunctions,
+    PythonValues,
+    impl,
+    register_impls,
+)
+from xdsl.interpreters.comparisons import to_signed, to_unsigned
+from xdsl.ir.core import Operation
+from xdsl.utils.exceptions import InterpretationError
+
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/riscv.py
```diff
@@ -0,0 +1,217 @@
...
+        *,
+        bitwidth: int = 32,
+        data: dict[str, Any] | None = None,
+        custom_instructions: dict[str, CustomInstructionFn] | None = None,
+    ):
+        super().__init__()
+        self.module_op = module_op
+        self.bitwidth = bitwidth
+        if data is None:
+            data = RiscvFunctions.get_data(module_op)
```
PapyChacal, line 76: This is untested and the only used of `get_data`. Don't you want to test it if you use it?

PapyChacal: idem for `def pairs(...)` and the `__add__` helper, which AFAIU are meant to be used in `get_data`

PapyChacal (approved):
LGTM, just comments on what seems to be a significant part of the code untested? Not blocking for this though, mainly wondering!

xdsl/interpreters/riscv.py
superlopuh: Ah good point forgot to add tests for this

xdsl/interpreters/riscv.py
```diff
@@ -0,0 +1,217 @@
...
+
+    data: list[_T]
+    offset: int = 0
+
+    def __add__(self, offset: int) -> Buffer[_T]:
+        """
+        Aliases the data list, so storing into the offset stores for all other references
+        to the list.
+        """
+        return Buffer(self.data, self.offset + offset // 4)
```
adutilleul, line 48: I don't mind too much but feels strange to have a generic data structure and a hardcoded "size" constant here.

adutilleul (approved):
LGTM!

PR #1364, opened 2023-07-28
dialects: Updating arith to init

ajrichins:
swapping over arith dialect to using __init__ as opposed to static get methods. Update usages of arith in xdsl tests and experimental interpreters.  added a test for arith select op but unsure if it is actually helping
@superlopuh @math-fehr

### DIFF
--- a/docs/database_example.ipynb
+++ b/docs/database_example.ipynb
@@ -471,16 +471,8 @@
    "name": "python3"
   },
   "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
    "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.11.0"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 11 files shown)


tests/dialects/test_arith.py
```diff
@@ -116,9 +117,22 @@ def test_arith_ops(self, func: type[BinaryOperation[_BinOpArgT]]):
...
+    t = Constant.from_int_and_width(1, IntegerType(1))
+    f = Constant.from_int_and_width(0, IntegerType(1))
+    select_t_op = Select(t, t, f)
+    select_f_op = Select(f, t, f)
+    select_t_op.verify_()
+    select_f_op.verify_()
+
+    # wanting to verify it actually selected the correct operand, but not sure if in correct scope
+    assert select_t_op.result.type == t.result.type
+    assert select_f_op.result.type == f.result.type
```
ajrichins, line 130: Couldn't figure out how to connect operand to the result and I saw the self verify already does this assert

docs/database_example.ipynb
```diff
@@ -471,8 +471,16 @@
    "name": "python3"
   },
   "language_info": {
+   "codemirror_mode": {
```
superlopuh, line 474: Please remove this change

superlopuh (approved):
beautiful

math-fehr (commented):
Very nice!
Do you think you could add back the `.get` as well with a `@deprecated` decorator on it, so we don't break too much code for other people? (I'll take care of removing it after updating to the next xDSL version)

math-fehr (approved):
Nice! Thanks a lot!

PR #1365, opened 2023-07-29
documentation: (toy) add initial lowering to riscv

superlopuh:
Adds the structure of the riscv lowering and the first pass. I added some future changes as TODOs so that the remaining work is clear.

### DIFF
--- a/docs/Toy/toy/compiler.py
+++ b/docs/Toy/toy/compiler.py
@@ -1,8 +1,18 @@
 from pathlib import Path
 
-from xdsl.dialects import affine, arith, func, memref, printf, scf
+from xdsl.dialects import (
+    affine,
+    arith,
+    func,
+    memref,
+    printf,
+    riscv,
+    riscv_func,
+    scf,
+)
 from xdsl.dialects.builtin import Builtin, ModuleOp
 from xdsl.ir import MLContext
+from xdsl.transforms.dead_code_elimination import DeadCodeElimination
 from xdsl.transforms.mlir_opt import MLIROptPass
 
 from .dialects import toy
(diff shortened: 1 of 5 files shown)


docs/Toy/toy/rewrites/setup_riscv_pass.py
```diff
@@ -0,0 +1,41 @@
...
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class AddSections(RewritePattern):
```
superlopuh, line 13: With @AntonLydike's riscemu malloc and friends, the .bss section hack is not necessary, similarly we can probably get away with dropping the data section. I'd like to first merge the end-to-end flow as it's working today, and make it better later, if that's ok with everyone.

docs/Toy/toy/compiler.py
```diff
@@ -83,4 +96,22 @@ def transform(
...
+    SetupRiscvPass().apply(ctx, module_op)
+    # LowerFuncToRiscvFunc().apply(ctx, module_op)
+    # LowerToyAccelerator().apply(ctx, module_op)
+    # LowerMemrefToRiscv().apply(ctx, module_op)
+    # LowerPrintfRiscvPass().apply(ctx, module_op)
+    # LowerArithRiscvPass().apply(ctx, module_op)
+    DeadCodeElimination().apply(ctx, module_op)
+    # ReconcileUnrealizedCastsPass().apply(ctx, module_op)
+
+    DeadCodeElimination().apply(ctx, module_op)
```
AntonLydike, line 110: twice?

AntonLydike (approved):
LGTM!

docs/Toy/toy/compiler.py
superlopuh: Ah yes I guess not really necessary right now, I'll remove it in the next PR

PR #1366, opened 2023-07-29
transformations: Add TypeConversionPattern

PapyChacal:
So, I want to have something that looks like type conversion infrastructure in xDSL.

MLIR offers this as TypeConverters, with many nuances, because they have well-defined infrastructure for dialect conversion passes, including partial conversions. Their type conversion smartly builds on top. We don;t have any of this, and I don't have the bandwidth right now to start implementing it all.

So I'm offering something else, still helpful and nice to use but more brutal; a PatternRewrite subclass, on which we can override
```python
def convert_type(self, typ: Attribute, /) -> Attribute | None:
        ...
```
Rather than
```python
def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter):
    ...
```
 Right now, it has two knobs:
 
 - recursive : Set the conversion to be recursive or not (e.g. convert `memref<A>` to `memref<B>` when translating `A` to `B`)
 - ops: A whitelist of operation types to convert, defaulting to all operation types.
 
I might want to start generating `unrealized_cast`s in cases where a whitelist is used? Not sure, as this is not part of my own usecase. Happy to discuss how that should behave if somebody is interested in this feature!

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -420,7 +420,7 @@ class Data(Generic[DataElement], Attribute, ABC):
         attr = cls.__new__(cls, params)
 
         # Call the __init__ of Data, which will set the parameters field.
-        cls.__init__(attr, params)
+        Data.__init__(attr, params)
         return attr
 
     @classmethod
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -420,7 +420,7 @@ class Data(Generic[DataElement], Attribute, ABC):
         attr = cls.__new__(cls, params)
 
         # Call the __init__ of Data, which will set the parameters field.
-        Data[Any].__init__(attr, params)
+        cls.__init__(attr, params)
         return attr
 
     @classmethod
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -417,10 +417,10 @@ class Data(Generic[DataElement], Attribute, ABC):
         """
         # Create the new attribute object, without calling its __init__.
         # We do this to allow users to redefine their own __init__.
-        attr = cls.__new__(cls, params)
+        attr = cls.__new__(cls)
 
         # Call the __init__ of Data, which will set the parameters field.
-        Data.__init__(attr, params)
+        Data.__init__(attr, params)  # type: ignore
         return attr
 
     @classmethod
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -420,7 +420,7 @@ class Data(Generic[DataElement], Attribute, ABC):
         attr = cls.__new__(cls)
 
         # Call the __init__ of Data, which will set the parameters field.
-        Data.__init__(attr, params)  # type: ignore
+        Data.__init__(attr, params)  # pyright: ignore[reportUnknownMemberType]
         return attr
 
     @classmethod
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -457,19 +457,23 @@ def op_type_rewrite_pattern(
     return impl
 
 
+@dataclass
 class TypeConversionPattern(RewritePattern):
+    recursive: bool = True
+
     @abstractmethod
     def convert_type(self, typ: Attribute, /) -> Attribute | None:
         ...
 
     def convert_type_rec(self, typ: Attribute) -> Attribute | None:
         inp = typ
-        if isinstance(typ, ParametrizedAttribute):
-            parameters = list(self.convert_type_rec(p) or p for p in typ.parameters)
-            inp = type(typ).new(parameters)
-        if isa(typ, ArrayAttr[Attribute]):
-            parameters = tuple(self.convert_type_rec(p) or p for p in typ)
-            inp = type(typ).new(parameters)
+        if self.recursive:
+            if isinstance(typ, ParametrizedAttribute):
+                parameters = list(self.convert_type_rec(p) or p for p in typ.parameters)
+                inp = type(typ).new(parameters)
+            if isa(typ, ArrayAttr[Attribute]):
+                parameters = tuple(self.convert_type_rec(p) or p for p in typ)
+                inp = type(typ).new(parameters)
         return self.convert_type(inp) or inp
 
     @final
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -466,35 +466,44 @@ class TypeConversionPattern(RewritePattern):
         inp = typ
         if isinstance(typ, ParametrizedAttribute):
             parameters = list(self.convert_type_rec(p) or p for p in typ.parameters)
-            inp = ParametrizedAttribute.new(parameters)
+            inp = type(typ).new(parameters)
         if isa(typ, ArrayAttr[Attribute]):
             parameters = tuple(self.convert_type_rec(p) or p for p in typ)
-            inp = ArrayAttr.new(parameters)
+            inp = type(typ).new(parameters)
         return self.convert_type(inp) or inp
 
     @final
     def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter):
+        new_result_types: list[Attribute] = []
+        new_attributes: dict[str, Attribute] = {}
+        changed: bool = False
         for result in op.results:
-            # TODO: rewriter!
             converted = self.convert_type_rec(result.type)
-            if converted:
-                result.type = converted
-        for operand in op.operands:
-            # TODO: rewriter!
-            converted = self.convert_type_rec(operand.type)
-            if converted:
-                operand.type = converted
+            new_result_types.append(converted or result.type)
+            if converted and converted != result.type:
+                changed = True
         for name, attribute in op.attributes.items():
             # TODO: rewriter
             converted = self.convert_type_rec(attribute)
-            if converted:
-                op.attributes[name] = converted
+            new_attributes[name] = converted or attribute
+            if converted and converted != attribute:
+                changed = True
         for region in op.regions:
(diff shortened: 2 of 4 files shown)


PapyChacal:
Oh wait, thought dump, *maybe* it could be slightly flexible by greedily inserting unrealised casts every time 🤔 but more costly

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,97 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+            converted = self.convert_type_rec(attribute)
+            new_attributes[name] = converted or attribute
+            if converted and converted != attribute:
+                changed = True
+        for region in op.regions:
+            for block in region.blocks:
+                for arg in block.args:
+                    converted = self.convert_type(arg.type)
+                    if converted and converted != arg.type:
+                        rewriter.modify_block_argument_type(arg, converted)
```
kingiler, line 495: missing `changed=True`?

PapyChacal: No, because here I do the change in-place through the rewriter if needed.
This is al to go through the rewriter API, mostly for the statekeeping it does for recursive application
(This way, the rewrite infrastructure knows when applying this pattern didn't change anything and won't infinitely recurse!)

PapyChacal: So I record changed for things that would require to replace the op below

xdsl/ir/core.py
```diff
@@ -417,10 +417,10 @@ def new(cls: type[Self], params: Any) -> Self:
         """
         # Create the new attribute object, without calling its __init__.
         # We do this to allow users to redefine their own __init__.
-        attr = cls.__new__(cls, params)
+        attr = cls.__new__(cls)
 
         # Call the __init__ of Data, which will set the parameters field.
-        Data[Any].__init__(attr, params)
+        Data.__init__(attr, params)  # type: ignore
```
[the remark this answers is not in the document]
PapyChacal: No, this is a trick used in multiple places, precisely to allow `__init__` redefinition on child classes (who could be `cls` here).
See for example `Operation.create`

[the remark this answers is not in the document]
kingiler: I see. Why not?
```suggestion
        Data[Any].__init__(attr, params)
```

PapyChacal, line 423: Okay this is confusing as hell, but basically, the *object* `Data[Any]` *is not* `Data`, even if the two types are the same, because `Data[Any]` is a `GenericTypeAlias` :upside_down_face:

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+      parameters.
+      e.g. a recusrive `i32` to `index` conversion will convert `vector<i32>` to
+      `vector<index>`.
+
+    - `ops` (defaulting to any Operation) is a tuple of Operation types on which to apply
+      the defined attribute conversion.
+    """
+
+    recursive: bool = True
+    ops: tuple[type[Operation]] | None = None
```
AntonLydike: Should be `tuple[type[Operation], ...]`

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+        if self.recursive:
+            if isinstance(typ, ParametrizedAttribute):
+                parameters = list(
+                    self._convert_type_rec(p) or p for p in typ.parameters
+                )
+                inp = type(typ).new(parameters)
+            if isa(typ, ArrayAttr[Attribute]):
+                parameters = tuple(self._convert_type_rec(p) or p for p in typ)
+                inp = type(typ).new(parameters)
+        return self.convert_type(inp) or inp
```
AntonLydike, line 518: Should check explicitly for None, as you can return an instance of a type here that has a custom `__bool__` method returning false (e.g. when it's empty), which would trigger this thing. Instead you should just use `is None`.

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+    @abstractmethod
+    def convert_type(self, typ: Attribute, /) -> Attribute | None:
+        """
+        The method to implement to define a TypeConversionPattern
+
+        This defines how the input Attribute should be converted.
+        It allows returning None, meaning "this attribute should not
+        be converted".
+        """
+        ...
```
AntonLydike, line 501: please use `raise NotImplementedError()`

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+        """
+        if self.ops and not isinstance(op, self.ops):
+            return
+        new_result_types: list[Attribute] = []
+        new_attributes: dict[str, Attribute] = {}
+        changed: bool = False
+        for result in op.results:
+            converted = self._convert_type_rec(result.type)
+            new_result_types.append(converted or result.type)
+            if converted and converted != result.type:
```
AntonLydike: Here you should also explicitly check for `is None`

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+                changed = True
+        for name, attribute in op.attributes.items():
+            converted = self._convert_type_rec(attribute)
+            new_attributes[name] = converted or attribute
+            if converted and converted != attribute:
```
AntonLydike: same here

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+                changed = True
+        for region in op.regions:
+            for block in region.blocks:
+                for arg in block.args:
+                    converted = self.convert_type(arg.type)
+                    if converted and converted != arg.type:
```
AntonLydike: same here

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +457,140 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+        expected_types = get_args(expected_type)
+
+    if not all(issubclass(t, Attribute) for t in expected_types):
+        raise Exception(
+            "attr_type_rewrite_pattern expects the non-self argument "
+            "type hint to be an `Attribute` subclass or a union of `Attribute` "
+            "subclasses."
+        )
+
+    def impl(self: _TypeConversionPatternT, typ: Attribute) -> Attribute | None:
```
AntonLydike, line 593: nit: use `@wraps(func)`

AntonLydike (approved):
LGTM overall, just some minor nitpicks

AntonLydike:
For info on `functools.wraps` see [here](https://docs.python.org/3/library/functools.html#functools.wraps)

xdsl/pattern_rewriter.py
PapyChacal: Nice one :)

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +458,149 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+    method. It uses type hints to match on a specific operation type before
+    calling the decorated function.
+    """
+    # Get the operation argument and check that it is a subclass of Operation
+    params = [param for param in inspect.signature(func).parameters.values()]
+    if len(params) != 2:
+        raise Exception(
+            "op_type_rewrite_pattern expects the decorated function to "
+            "have one non-self argument."
+        )
```
math-fehr, line 581: We can remove this, as pyright will check this.

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +458,149 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+    expected_types = (expected_type,)
+    if get_origin(expected_type) in [Union, UnionType]:
+        expected_types = get_args(expected_type)
+
+    if not all(issubclass(t, Attribute) for t in expected_types):
+        raise Exception(
+            "attr_type_rewrite_pattern expects the non-self argument "
+            "type hint to be an `Attribute` subclass or a union of `Attribute` "
+            "subclasses."
+        )
```
math-fehr, line 593: This is also checked by pyright

math-fehr (approved):
Nice!

xdsl/ir/core.py
superlopuh, line 423: Can you please replace this with the exact pyright error?

xdsl/pattern_rewriter.py
```diff
@@ -439,6 +458,148 @@ def impl(self: _RewritePatternT, op: Operation, rewriter: PatternRewriter) -> No
...
+        """
+        ...
+
+    @final
+    def _convert_type_rec(self, typ: Attribute) -> Attribute | None:
+        """
+        Provided recursion over structed/parameterized Attributes.
+        """
+        inp = typ
+        if self.recursive:
```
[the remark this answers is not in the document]
PapyChacal: I'm not sure how you see it?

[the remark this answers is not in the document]
math-fehr: Wait nevermind actually, I missed something --'

PR #1367, opened 2023-07-29
core: Fix cornercase fail in `isa`

PapyChacal:
So I added a test, that fails because `isa` is failing  unexpectedly on that case, and I think it was because of the verifier of `ArrayAttr`. I solve it, then add a regression test for the related `DictionaryAttr`.

### DIFF
--- a/docs/irdl.ipynb
+++ b/docs/irdl.ipynb
@@ -566,7 +566,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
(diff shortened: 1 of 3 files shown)


PapyChacal:
@math-fehr Can you make sense of the Notebook error? I did not rephrase this error message myself. I guess it changed something somewhere in an exception cascade? :thinking:

superlopuh:
I think you just need to run the notebook?

superlopuh (approved):

PR #1370, opened 2023-07-31
dialects: (riscv) add `AssemblySectionOp` and split `DirectiveOp`

adutilleul:
**Rationale**: `DirectiveOp` has a broad scope handling both directives for defining sections and data like constants, so by splitting it in another operation `AssemblySectionOp` it makes things clearer and easier to write some lowering passes.

Taken from @superlopuh `workshop-all` branch and added a custom syntax over it.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -2055,7 +2055,7 @@ class LabelOp(IRDLOperation, RISCVOp):
 @irdl_op_definition
 class DirectiveOp(IRDLOperation, RISCVOp):
     """
-    The directive operation is used to emit assembler directives (e.g. .word; ..equ; etc.)
+    The directive operation is used to emit assembler directives (e.g. .word; .equ; etc.)
     without any associated region of assembly code.
     A more complete list of directives can be found here:
 
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -42,7 +42,7 @@ from xdsl.irdl import (
     var_operand_def,
     var_result_def,
 )
-from xdsl.irdl.irdl import region_def
+from xdsl.irdl.irdl import OptRegion, opt_region_def, region_def
 from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
 from xdsl.traits import IsTerminator, NoTerminator
@@ -1992,6 +1992,7 @@ class EcallOp(NullaryOperation):
     name = "riscv.ecall"
 
 
+
 @irdl_op_definition
 class LabelOp(IRDLOperation, RISCVOp):
     """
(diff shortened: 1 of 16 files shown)


xdsl/dialects/riscv.py
```diff
@@ -2055,7 +2055,8 @@ def assembly_line(self) -> str | None:
 @irdl_op_definition
 class DirectiveOp(IRDLOperation, RISCVOp):
     """
-    The directive operation is used to emit assembler directives (e.g. .word; .text; .data; etc.)
+    The directive operation is used to emit assembler directives (e.g. .word; ..equ; etc.)
```
superlopuh, line 2058: ..equ?

xdsl/dialects/riscv.py
```diff
@@ -2064,29 +2065,22 @@ class DirectiveOp(IRDLOperation, RISCVOp):
     name = "riscv.directive"
     directive: StringAttr = attr_def(StringAttr)
     value: StringAttr | None = opt_attr_def(StringAttr)
```
superlopuh, line 2067: It feels like maybe we could make this non-optional now? I'm not 100% sure but it feels like if a directive doesn't demark a section in the assembly code then it should probably have a value. If it doesn't and the empty string is inappropriate the user can always create an assembly section with an empty region. But maybe I'm missing something.

superlopuh (approved):

adutilleul:
> `AssemblySectionOp` seems will not print attributes that are not defined in the class. It is not able to round-trip.

As we discussed on Zulip, I'm not sure whether it's really relevant right now since I don't see any obvious use-cases for other attributes at that point on this operation.

kingiler (approved):

superlopuh:
With the recently added helpers by @math-fehr and @PapyChacal it should be pretty easy to add printing and parsing for remaining attributes, could you please take a look? Also it feels like potentially the whole thing could be implemented with the declarative syntax, could you please take a quick look with @math-fehr  if that's the case?

PapyChacal:
@superlopuh I don't think I have something to do with those helpers, or I'm just not sure which one you mean?

superlopuh:
I was referring to #1323

adutilleul:
> I was referring to #1323

I've done it but I'm curious do you have any actual use cases or is this just a general support you wanna have ? If the latter case, it might a good idea to open an issue for all the previous custom parsing in the codebase that doesn't handle this to be consistent.

math-fehr:
The declarative parsing/printing doesn't support attributes yet, so that wouldn't work for now! But it is planned

superlopuh:
Ah you've added it already. Fantastic, we're future-proof :)

JosseVanDelm:
@adutilleul what lowering passes are easier to write this way?
I think the rationale makes sense, just wondering why we need it

superlopuh:
I asked @adutilleul  to add this mostly so that the operations convey the semantics a bit more. Concretely, this helps us only walk the assembly sections to find the data section, as opposed to all directive ops, and might make it easier to find the text section in the future. I also find that it makes the filecheck file much easier to read.

JosseVanDelm:
@superlopuh that makes sense! Thanks!

PR #1373, opened 2023-07-31
transformations: Use TypeConversionPattern to simplify and improve the stencil lowering.

PapyChacal:
Using #1366, this trims a lot of undesired code from this lowering, while making it way more robust!
Adding a MFE for a case that would have required more random code in this lowering instead of having a more generic solution.


math-fehr (approved):
Very nice!

AntonLydike (approved):

superlopuh (approved):

PR #1374, opened 2023-07-31
transforms: Add canonicalization infrastructure with example

AntonLydike:
This adds canonicalization infrastructure.

 - A new subfolder `xdsl.transforms.canonicalization`
 - Contains the `CanonicalizationPass`
 - Contains an example riscv canonicalization pattern to fold immediates
 - Canonicalization patterns can be added using the `@is_canonicalization` decorator
 - Canonicalization patterns are only applied up to 5 times to prevent infinite loops

This PR also includes a RISC-V canonicalization pass I wrote to check out how this feels/works. Will remove once we are done with the general discussion and are ready to review PRs.

### DIFF
--- a/xdsl/transforms/canonicalize/__init__.py
+++ b/xdsl/transforms/canonicalize/__init__.py
@@ -1,4 +1,5 @@
 # ruff: noqa
+# type: ignore[reportUnusedImport]
 # import main module first
 from xdsl.transforms.canonicalize.canonicalize_pass import (
     CanonicalizationPass,
--- a/xdsl/transforms/canonicalize/__init__.py
+++ b/xdsl/transforms/canonicalize/__init__.py
@@ -1,0 +1,6 @@
+# list all canonicalization things here so they are included:
+import xdsl.transforms.canonicalize.canonicalize_riscv
+from xdsl.transforms.canonicalize.canonicalize_pass import (
+    CanonicalizationPass,
+    is_canonicalization,
+)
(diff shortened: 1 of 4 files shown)


xdsl/transforms/canonicalize/__init__.py
AntonLydike, line 1: Does anyone know how to tell pyright to calm down because this is an __init__ file that's supposed to expose the underlying module with a nicer interface?

superlopuh: No, I've hit this before, and it doesn't seem well supported. The other thing is that the language server recommends the original definition not the __init__ import when quick fixing missing imports, which makes the whole thing sort of tedious

xdsl/transforms/canonicalize/canonicalize_riscv.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def twos_complement_bitwidth(num: int):
+    """
+    Quick approximation of number of bits required to represent a number in two's
+    complement.
+    """
+    is_neg = 1 if num < 0 else 0
+    num = abs(num)
+    return ceil(log2(num)) + is_neg
```
superlopuh, line 35:
```suggestion
    return num.bit_length() + is_neg
```
;)

xdsl/transforms/canonicalize/__init__.py
kingiler: My recent PR has calm down `ruff` on `__init__.py` #1388.

xdsl/transforms/canonicalize/canonicalize_riscv.py
kingiler: Python doc [link](https://docs.python.org/3/library/stdtypes.html#int.bit_length), looks like `1 + int(log(abs(x), 2))` for non-zero `x`.

superlopuh:
It would be nice to get some of the mechanism from this PR merged, although I'm not sure where that fits with your plans. Maybe could close this PR in the meantime?

PR #1379, opened 2023-08-01
documentation: (toy) add toy accelerator interpreter functions

superlopuh:


docs/Toy/toy/rewrites/helpers.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def cast_value_to_register(operand: SSAValue, rewriter: PatternRewriter) -> OpResult:
+    types = (riscv.IntRegisterType.unallocated(),)
+    cast = builtin.UnrealizedConversionCastOp.get((operand,), types)
+    rewriter.insert_op_before_matched_op(cast)
+    return cast.results[0]
+
+
+def copy_registers(
```
adutilleul, line 15: `nit:` Since we went away from the `cf` lowering (were it was used previously iirc) and also that the helper is not used in this PR, do we really need it?

adutilleul (approved):

AntonLydike (approved):

docs/Toy/toy/rewrites/helpers.py
superlopuh: Ah it's not used in this PR but it's used in other ones, I just copied the whole helper file. We need it for riscv_func and riscv_scf lowerings, I think

PR #1381, opened 2023-08-01
interpreter: (riscv) add riscv_scf interpreter functions

superlopuh:


adutilleul (approved):

xdsl/interpreters/riscv_scf.py
```diff
@@ -0,0 +1,39 @@
...
+    ReturnedValues,
+    impl,
+    impl_terminator,
+    register_impls,
+)
+
+
+@register_impls
+class RiscvScfFunctions(InterpreterFunctions):
+    bitwidth = 32
```
kingiler, line 16: If this is a constant, it would be better to have capitalized variable name.

xdsl/interpreters/riscv_scf.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @impl(riscv_scf.ForOp)
+    def run_for(
+        self,
+        interpreter: Interpreter,
+        op: riscv_scf.ForOp,
+        args: tuple[Any, ...],
+    ):
+        lb, ub, step, *loop_args = args
+        loop_args = tuple(loop_args)
```
kingiler, line 26: I'm not sure this is redundant?

xdsl/dialects/riscv_scf.py
```diff
@@ -37,7 +37,7 @@ class YieldOp(IRDLOperation):
     traits = frozenset([IsTerminator()])
 
     def __init__(self, *operands: SSAValue | Operation):
-        super().__init__(operands=[SSAValue.get(operand) for operand in operands])
+        super().__init__(operands=[[SSAValue.get(operand) for operand in operands]])
```
kingiler, line 40: Why?

kingiler (approved):

xdsl/dialects/riscv_scf.py
superlopuh: each element in the operands array corresponds to an irdl operand definition. This class has one definition, so it should be a single-elemnt list. The one definition is variadic, hence the list.

xdsl/interpreters/riscv_scf.py
superlopuh: the lhs is a tuple, the rhs is a list, this was to make pyright happy, otherwise the variable has type list | tuple

xdsl/interpreters/riscv_scf.py
superlopuh: Good spot. I'm now thinking that maybe the `InterpreterFunctions` superclass should have the bitwidth set, a lot of interpreters subtlely depend on it.

xdsl/dialects/riscv_scf.py
kingiler: I see, thanks for explaining.

xdsl/interpreters/riscv_scf.py
kingiler: I see, forget that `*loop_args` returns a list instead of tuple.

xdsl/interpreters/riscv_scf.py
superlopuh: #1386

KGrykiel (approved):

PR #1383, opened 2023-08-01
transformations: Add tile-sizes flag to convert-stencil-to-ll-mlir.

PapyChacal:
NFC without the flag.
This will tile the `min(#tile-sizes, stencil-dimensionality)` outter dimensions of the `stencil.apply` in the CPU codegen case.


xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
```diff
@@ -526,7 +572,7 @@ def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         the_one_pass = PatternRewriteWalker(
             GreedyRewritePatternApplier(
                 [
-                    ApplyOpToParallel(return_targets, self.target),
+                    ApplyOpToParallel(return_targets, self.target, self.tile_sizes),
```
georgebisbas, line 575: Would it not be better to have a separate pass for tiling?
Like a pass to only tile and not parallelize?

PapyChacal: You mean one pattern for the GPU case and another one for the CPU/eventually tiled case? As a factorization thing?
I could do that, this function is terrifying and definitely needs cleaning.
Also just the tiling logic is terribly spaghetti. I was racing for the fastest correct thing and want to clean it if it works as intended for its goal. Again, the endgame would be to not tile at this level at all!

PR #1387, opened 2023-08-01
documentation: (toy) add lowering from `memref` to toy accelerator pass

adutilleul:
Memref lowering with tests as part of the toy flow upstreaming (https://github.com/xdslproject/xdsl/pull/1379).

### DIFF
--- a/docs/Toy/toy/rewrites/lower_memref_riscv.py
+++ b/docs/Toy/toy/rewrites/lower_memref_riscv.py
@@ -12,7 +12,7 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 
-from .lower_to_toy_accelerator import cast_value_to_register
+from .helpers import cast_value_to_register
 
 
 class LowerMemrefAllocOp(RewritePattern):
--- a/docs/Toy/toy/rewrites/lower_memref_riscv.py
+++ b/docs/Toy/toy/rewrites/lower_memref_riscv.py
@@ -45,19 +45,19 @@ def insert_shape_ops(
     """
     assert len(shape) == len(indices)
 
-    match len(shape):
-        case 1:
+    match indices:
+        case [idx1]:
             rewriter.insert_op_before_matched_op(
                 [
-                    ptr := riscv.AddOp(mem, indices[0]),
+                    ptr := riscv.AddOp(mem, idx1),
                 ]
             )
-        case 2:
+        case [idx1, idx2]:
             rewriter.insert_op_before_matched_op(
                 [
                     cols := riscv.LiOp(shape[1]),
-                    row_offset := riscv.MulOp(cols, indices[0]),
-                    offset := riscv.AddOp(row_offset, indices[1]),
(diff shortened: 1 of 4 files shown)


docs/Toy/toy/rewrites/lower_memref_riscv.py
```diff
@@ -0,0 +1,125 @@
...
+            [
+                cols := riscv.LiOp(shape[1]),
+                row_offset := riscv.MulOp(cols, indices[0]),
+                offset := riscv.AddOp(row_offset, indices[1]),
+                offset_bytes := riscv.SlliOp(offset, 2, comment="mutiply by elm size"),
+                ptr := riscv.AddOp(mem, offset_bytes),
+            ]
+        )
+    else:
+        raise NotImplementedError(f"Unsupported memref shape {shape}")
```
kingiler, line 65: Feel like it can be rewriting using pattern matching which will be more concise and readable.

kingiler: Something like this works, tested on my machine but you may rename the variable name for readability.
```suggestion
    match indices:
        case [el]:
            rewriter.insert_op_before_matched_op(
                [
                    ptr := riscv.AddOp(mem, el),
                ]
            )
        case [first, second]:
            rewriter.insert_op_before_matched_op(
                [
                    cols := riscv.LiOp(shape[1]),
                    row_offset := riscv.MulOp(cols, first),
                    offset := riscv.AddOp(row_offset, second),
                    offset_bytes := riscv.SlliOp(
                        offset, 2, comment="mutiply by elm size"
                    ),
                    ptr := riscv.AddOp(mem, offset_bytes),
                ]
            )
        case _:
            raise NotImplementedError(f"Unsupported memref shape {shape}")
```

adutilleul:
> It feels weird to check for length and then access the elements instead of pattern match directly on the sequence.



> It feels weird to check for length and then access the elements instead of pattern match directly on the sequence.

I forgot that Python had list deconstruction support when pattern matching, but yeah that's better.

docs/Toy/toy/tests/test_lower_memref_riscv.py
```diff
@@ -0,0 +1,190 @@
...
+from xdsl.pattern_rewriter import PatternRewriter
+from xdsl.utils.test_value import TestSSAValue
+
+from ..rewrites.lower_memref_riscv import LowerMemrefToRiscv, insert_shape_ops
+
+memref_type_2xf64 = memref.MemRefType.from_element_type_and_shape(f64, ([2]))
+memref_type_2x2xf64 = memref.MemRefType.from_element_type_and_shape(f64, ([2, 2]))
+memref_type_2x2x2xf64 = memref.MemRefType.from_element_type_and_shape(f64, ([2, 2, 2]))
+
+register_type = riscv.IntRegisterType.unallocated()
```
superlopuh, line 16:
```suggestion
REGISTER_TYPE = riscv.IntRegisterType.unallocated()
```

superlopuh (approved):
LGTM

kingiler (approved):

PR #1388, opened 2023-08-01
ci: (ruff) add F rule

kingiler:
[`F` rule](https://beta.ruff.rs/docs/rules/#pyflakes-f) enforce rules from pyflakes and attempt to auto fix.
Important rules such as
- removal of unused imports
- warn shadowing of same variables name
- remove redundant f-string usage


superlopuh (approved):
very nice, thank you

PR #1390, opened 2023-08-01
cli: Print parse error diagnostics instead of just the stacktrace

AntonLydike:
`xdsl-opt` will now print the parse error diagnostic message instead of the stacktrace.

Before:
```
> xdsl-opt ...
  File "/home/anton/unied/xdsl/venv/bin/xdsl-opt", line 8, in <module>
(many many lines omitted)
xdsl.utils.exceptions.ParseError: (Span[188:190](text='x2'), "'>' expected")
```

now:

```
> xdsl-opt ...
tests/filecheck/dialects/dmp/ops.mlir:7:26
        grid = #dmp.grid<2x2>,
                          ^^
                          '>' expected
```

Arguably, we may want to also print the stacktrace, but not 100% sure. Open to feedback!

### DIFF
--- a/xdsl/xdsl_opt_main.py
+++ b/xdsl/xdsl_opt_main.py
@@ -9,7 +9,7 @@ from xdsl.ir import MLContext
 from xdsl.passes import ModulePass
 from xdsl.printer import Printer
 from xdsl.tools.command_line_tool import CommandLineTool, get_all_passes
-from xdsl.utils.exceptions import DiagnosticException, ParseError
+from xdsl.utils.exceptions import DiagnosticException
 from xdsl.utils.parse_pipeline import parse_pipeline
 
 


superlopuh (approved):

superlopuh (approved):

xdsl/utils/exceptions.py
```diff
@@ -71,6 +71,9 @@ class ParseError(Exception):
     def __repr__(self) -> str:
         return self.span.print_with_context(self.msg)
 
+    def with_context(self) -> str:
+        return self.span.print_with_context(self.msg)
+
```
PapyChacal, line 76: Genuine question, it's the exact duplicate of `__repr__`, can't `__repr__` be used instead?
Otherwise make __repr__ use this one?

PapyChacal (approved):
Minor tweak, otherwise thanks, this looks headache-saving 🙏

superlopuh (approved):

xdsl/utils/exceptions.py
AntonLydike: Not sure why exactly I chose that way, I guess I didn't want to depend on the current behavior of `__repr__` (as it conflicts with the stated goal of `__repr__`)...

PR #1394, opened 2023-08-02
dialect: (riscv) Add custom syntax to RdRsImmIntegerOperation

kingiler:
Add custom syntax to RdRsImmIntegerOperation

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -600,11 +600,7 @@ class RdImmIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
 
     def custom_print_attributes(self, printer: Printer) -> Set[str]:
         printer.print(" ")
-        match self.immediate:
-            case IntegerAttr():
-                printer.print(self.immediate.value.data)
-            case LabelAttr():
-                printer.print_string_literal(self.immediate.data)
+        _print_immediate_value(printer, self.immediate)
         return {"immediate"}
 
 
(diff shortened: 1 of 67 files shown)


xdsl/dialects/riscv.py
```diff
@@ -715,6 +715,23 @@ def __init__(
...
+    def custom_parse_attributes(cls, parser: Parser) -> Mapping[str, Attribute]:
+        attributes = dict[str, Attribute]()
+        attributes["immediate"] = _parse_immediate_value(
+            parser, IntegerType(12, Signedness.SIGNED)
+        )
+        return attributes
+
+    def custom_print_attributes(self, printer: Printer) -> Set[str]:
+        printer.print(", ")
+        match self.immediate:
```
superlopuh, line 728: feels like this could be a helper next to _parse_immediate_value? We'll very likely want to reuse it

superlopuh (approved):
LGTM

xdsl/dialects/riscv.py
kingiler: Address in latest commit.

superlopuh (approved):

xdsl/dialects/riscv.py
```diff
@@ -707,6 +707,19 @@ def __init__(
     def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
         return self.rd, self.rs1, self.immediate
 
+    @classmethod
+    def custom_parse_attributes(cls, parser: Parser) -> Mapping[str, Attribute]:
+        attributes = dict[str, Attribute]()
+        attributes["immediate"] = _parse_immediate_value(
+            parser, IntegerType(12, Signedness.SIGNED)
```
adutilleul, line 714: what about the child class `RdRsImmShiftOperation` which uses 5 bits immediate ?

adutilleul (changes requested):

xdsl/dialects/riscv.py
kingiler: Address in #1416

PR #1399, opened 2023-08-03
core: print error on operation with mismatched types, not next one

superlopuh:
After parsing the types, `current_token` points to the next operation, so the diagnostic is misleading.


PapyChacal (approved):
Thanks 🙏 Confused me a few times indeed

math-fehr (approved):
Nice!

PR #1400, opened 2023-08-03
misc: disable type ignore

superlopuh:
Type ignore sometimes hides the wrong issue. We sometimes want to just get pyright to stop complaining about a private funciton call and then, when the function interface changes or is deleted, we miss the type error. When ignoring pyright, it's best to always write `pyright: ignore[specific, issues]` as opposed to a blanket `type: ignore`. I also found some `type: ignore[specificIssue]` but that's not actually a syntax that `type: ignore` supports, it just ignores the whole line.

### DIFF
--- a/xdsl/_version.py
+++ b/xdsl/_version.py
@@ -1,5 +1,3 @@
-# type: ignore
-
 # This file helps to compute a version number in source trees obtained from
 # git-archive tarball (such as those provided by githubs download-from-tag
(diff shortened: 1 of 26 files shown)


xdsl/_version.py
superlopuh, line 1: whole file is ignored in config

AntonLydike (approved):

PapyChacal (approved):

adutilleul (approved):

PR #1403, opened 2023-08-03
documentation: (toy) lower arith

superlopuh:
This is the last partial lowering of builtins, so I did a bit of cleanup also


docs/Toy/toy/rewrites/arith_float_to_int.py
```diff
@@ -0,0 +1,92 @@
...
+            return
+
+        assert isinstance(op.value, FloatAttr)
+
+        op_value = cast(FloatAttr[AnyFloat], op.value)
+
+        value = op.value.value.data
+        assert value == float(
+            int(value)
+        ), f"Only support integer values in arith.Constant, got {value}"
```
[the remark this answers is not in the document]
superlopuh: nice

adutilleul (approved):

docs/Toy/toy/rewrites/arith_float_to_int.py
```diff
@@ -0,0 +1,92 @@
...
+            value.is_integer()
+        ), f"Only support integer values in arith.Constant, got {value}"
+        value = int(value)
+
+        rewriter.replace_matched_op(
+            [
+                constant := arith.Constant.from_int_and_width(value, 32),
+                UnrealizedConversionCastOp.get(constant.results, (op_value.type,)),
+            ]
+        )
```
kingiler, line 51:
```suggestion
class CastConstantOp(RewritePattern):
    @op_type_rewrite_pattern
    def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter):
        match op.value:
            case IntegerAttr():
                return
            case FloatAttr():
                op_value = cast(FloatAttr[AnyFloat], op.value)
                value = op_value.value.data
                assert (
                    value.is_integer()
                ), f"Only support integer values in arith.Constant, got {value}"
                value = int(value)

                rewriter.replace_matched_op(
                    [
                        constant := arith.Constant.from_int_and_width(value, 32),
                        UnrealizedConversionCastOp.get(
                            constant.results, (op_value.type,)
                        ),
                    ]
                )
            case _:
                assert False


```

kingiler (approved):
LGTM except some nits

docs/Toy/toy/rewrites/arith_float_to_int.py
superlopuh: I don'f feel that thisd is an improvement.early returns are good, it communicates that there is a "happy path" down to the end of the function, whereas a match communicates that either case is equally interesting, I feel that the former applies better here.

PR #1404, opened 2023-08-03
dialect: (arith) Change from IntegerBinaryOp to SignlessIntegerBinaryOp for RemSI, ShLI and ShRSI

kingiler:
Change from IntegerBinaryOp to SignlessIntegerBinaryOp for RemSI, ShLI and ShRSI.
Preparing for #1403.

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -230,7 +230,7 @@ class RemUI(SignlessIntegerBinaryOp):
 
 
 @irdl_op_definition
-class RemSI(IntegerBinaryOp):
+class RemSI(SignlessIntegerBinaryOp):
     name = "arith.remsi"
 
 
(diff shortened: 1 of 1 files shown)


superlopuh (approved):
LGTM

adutilleul (approved):

xdsl/dialects/arith.py
```diff
@@ -230,7 +230,7 @@ class RemUI(SignlessIntegerBinaryOp):
 
 
 @irdl_op_definition
-class RemSI(IntegerBinaryOp):
+class RemSI(SignlessIntegerBinaryOp):
```
superlopuh, line 233: @PapyChacal Do you remember why this was previously `IntegerBinaryOp`? I just looked at the documentation, and while the operation treats the inputs as signed, lhs and rhs are signless-like. Was this intentional? Do you think this change is safe?

PR #1406, opened 2023-08-03
dialect: (builtin) Custom syntax for UnrealizedConversionCastOp

kingiler:
Custom syntax for UnrealizedConversionCastOp in par with MLIR.


AntonLydike:
Can you add proper interop tests as well? Just checking that mlir can parse our custom syntax and it doesn't fail

kingiler:
Tests are added for print and parsing custom syntax.

superlopuh:
I think @AntonLydike means a test like `tests/filecheck/mlir-conversion/with-mlir/dialects/builtin/unrealized_conversion_cast.mlir` but that would roundtrip the custom syntax as opposed to the custom syntax

tests/filecheck/dialects/builtin/unrealized_conv_cast.mlir
```diff
@@ -1,16 +1,22 @@
-// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | filecheck %s
+// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | xdsl-opt | filecheck %s
```
superlopuh, line 1: I'm not sure of the benefit of three `xdsl-opt`s

PapyChacal (commented):
This implementation doesn't handle nor test extra attributes in the syntax.
Just noticed the other day that we have this problem with quite a few custom syntaxes, we should be careful!

tests/filecheck/dialects/builtin/unrealized_conv_cast.mlir
```diff
@@ -1,16 +1,22 @@
...
-    // CHECK: %1 = "builtin.unrealized_conversion_cast"(%0) : (i64) -> f32
+    // CHECK: %1 = builtin.unrealized_conversion_cast %0 : i64 to f32
     %1 = "builtin.unrealized_conversion_cast"(%0) : (i64) -> f32
-    // CHECK: %2 = "builtin.unrealized_conversion_cast"(%0) : (i64) -> i32
+    // CHECK: %2 = builtin.unrealized_conversion_cast %0 : i64 to i32
     %2 = "builtin.unrealized_conversion_cast"(%0) : (i64) -> i32
-    // CHECK: %3 = "builtin.unrealized_conversion_cast"(%0) : (i64) -> i64
+    // CHECK: %3 = builtin.unrealized_conversion_cast %0 : i64 to i64
     %3 = "builtin.unrealized_conversion_cast"(%0) : (i64) -> i64
+    // CHECK: %4 = builtin.unrealized_conversion_cast to i64
```
PapyChacal, line 14: Oh, never seen that case, what is a cast from nothing to a type? 🤔
Looked for a case because I see it specially implemented in the parser.

kingiler: [UnrealizedConversionCastOp](https://mlir.llvm.org/docs/Dialects/Builtin/#builtinunrealized_conversion_cast-unrealizedconversioncastop) seems has 0-1 conversion.

tests/filecheck/dialects/builtin/unrealized_conv_cast.mlir
kingiler: For around-tripping printing and parsing of both generic and custom syntax.

kingiler:
Interop tests has been added.

kingiler:
Attribute dict has been handled.

superlopuh (approved):

PR #1356, opened 2023-07-27
add scf-to-riscv pass

AntonLydike:
Add a pass `scf-to-riscv` that converts scf for ops to `riscv.for`, for register allocation afterwards.

Missing tests currently.

PR #1407, opened 2023-08-03
misc: (toy) Rename cast_value_to_register to cast_value_to_int_register

kingiler:
Rename `cast_value_to_register` to `cast_value_to_int_register` in order to distinguish `cast_value_to_int_register` and `cast_value_to_float_register`.
Preparing for #1403.


superlopuh:
I think maybe we should just wait for @adutilleul's #1356 to be merged and use the helpers there. Might as well use this PR to do that once that PR is merged

PR #1409, opened 2023-08-03
dialects: (stencil) Add access pattern analysis

AntonLydike:
This adds `.get_accesses()` to the stencil apply operation, which returns an `AccessPattern` for each stencil field.

Each access pattern contains a list of accesses, and offers some helpers like `is_diagonal`, `get_diagonals`, etc.

### DIFF
--- a/tests/dialects/test_stencil.py
+++ b/tests/dialects/test_stencil.py
@@ -649,7 +649,7 @@ def test_access_patterns():
         AccessOp.get(t1, (1, 1), (1, 0))
         AccessOp.get(t1, (-1, -1), (1, 0))
 
-    apply = ApplyOp.get((temp, temp), apply_op_region.detach_block(0), [temp])
+    apply = ApplyOp.get((temp, temp), apply_op_region.detach_block(0), [typ])
 
     t0_acc, t1_acc = tuple(apply.get_accesses())
 
--- a/tests/dialects/test_stencil.py
+++ b/tests/dialects/test_stencil.py
@@ -1,5 +1,6 @@
 import pytest
 
+from xdsl.builder import Builder
 from xdsl.dialects.builtin import (
     AnyFloat,
     ArrayAttr,
@@ -35,7 +36,7 @@ from xdsl.dialects.stencil import (
     StoreResultOp,
     TempType,
 )
-from xdsl.ir import Attribute, Block
+from xdsl.ir import Attribute, Block, SSAValue
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import isa
 from xdsl.utils.test_value import TestSSAValue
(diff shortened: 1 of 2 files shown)


tests/dialects/test_access_pattern.py
kingiler, line 1: Empty file?

tests/dialects/test_stencil.py
```diff
@@ -631,3 +632,31 @@ def test_buffer():
     assert isinstance(buffer, BufferOp)
     assert buffer.temp == temp
     assert buffer.res.type == res_type
+
+
+def test_access_patterns():
```
georgebisbas, line 637: Nice

georgebisbas (approved):

xdsl/dialects/stencil.py
```diff
@@ -731,6 +755,66 @@ def verify_(self) -> None:
...
+            right = max(ax[axis], right)
+        return left, right
+
+    def halos(self) -> tuple[tuple[int, int], ...]:
+        n = self.dims
+        lefts, rights = [0] * n, [0] * n
+        for ax in self.accesses:
+            for axis in range(n):
+                lefts[axis] = min(ax[axis], lefts[axis])
+                rights[axis] = max(ax[axis], rights[axis])
```
PapyChacal, line 799: This could reuse `extent_in_axis`

xdsl/dialects/stencil.py
```diff
@@ -731,6 +755,66 @@ def verify_(self) -> None:
...
+        return tuple(zip(lefts, rights))
```
PapyChacal, line 800: I'd say this one could use a docstring

PapyChacal (approved):
Nice! Only minor nitpicks, nothing to block for if using it is urgent 🙂

xdsl/dialects/stencil.py
```diff
@@ -345,6 +346,29 @@ def get_rank(self) -> int:
...
         return res_type.get_num_dims()
 
+    def get_accesses(self) -> Iterable[AccessPattern]:
+        """
+        Return the access patterns of each input.
+
+        An access is a tuple if offsets: tuple[int, ...]
+        An access pattern is a sequence of accesses: Sequence[tuple[int, ...], ...]
+        This method returns an access pattern for each stencil field the apply operates
+        on.
```
PapyChacal, line 356: I would love to use "offset" rather than "access" to stick closer to the dialect's terminology, or do you find it confusing?

PapyChacal: As in the individual offsets; not the access pattern. I like that name 👍

xdsl/dialects/stencil.py
AntonLydike: Oh definitely^^

xdsl/dialects/stencil.py
PapyChacal: I don't see it?
If you swap the loops, the inner one is exactly extent_in_axis, and the total operations are the same, or am I missing something?

PapyChacal: Oh, because one of those is element-wise max?

I would prefer readability before this complexity becomes a problem, I'm not sure we'll ever have asymptotically big numbers of accesses.

But it's nitpicking 🙂

PR #1416, opened 2023-08-04
tests: (riscv) Check custom syntax gives plausible parsing result

kingiler:
Add a filecheck to ensure that parsing custom syntax gives desirable IR with correct attributes for all RISCV Op in xDSL.

A tiny bug was discovered when checking the output for the riscv shift operations, where it should parse back to 5 bit unsigned integer.


tests/filecheck/dialects/riscv/riscv_ops_parse.mlir
kingiler, line 1: This file is to check whether the parse result make sense. Avoid cases that a specific integer type is expected.

tests/filecheck/dialects/riscv/riscv_ops_parse.mlir
```diff
@@ -0,0 +1,233 @@
+builtin.module {
```
superlopuh, line 1: Looks like the RUN is missing

superlopuh: But also this feels like a pretty big test for this particular problem. In my understanding, we're trying to make sure that for each operation that can parse an immediate value, the Attribute has the correct bitwidth. Let's try to solve that separately? In the meantime, let's only test the RISCVOps that have interger immediate values?

superlopuh (changes requested):
Could you please reduce the test case to only the ones we care about? Also, could you please add a comment to the test file itself stating what its purpose is?

tobiasgrosser:
> Could you please reduce the test case to only the ones we care about? Also, could you please add a comment to the test file itself stating what its purpose is?

@kingiler, can you reduce this test case or is there a reason such a large test case is introduced with this small change?

kingiler:
The aim of the PR is to ensure custom parsing gives desirable IR with correct attributes.

tests/filecheck/dialects/riscv/riscv_ops_parse.mlir
kingiler: Merged into one file in newest commit.

superlopuh: Many of the operations added to the test do not have immediate values, so I do not see the benefit in testing them. Could you please reduce it to only include operations with immediate values? Also, could you please add a comment explaining the purpose of the addition to the test?

kingiler: The test is not only for immediate value, but to test both generic and custom syntax and printed and parsed appropriately. I don't understand your question?

superlopuh (approved):

tests/filecheck/dialects/riscv/riscv_ops_parse.mlir
superlopuh: Ok, let's merge as-is, there may be another benefit I'm not seeing right now.

adutilleul (approved):

PR #1417, opened 2023-08-04
dialects: (riscv) add custom assembly format for `flw` and `fsw` instructions

adutilleul:
`flw` and `fsw` were not updated in (https://github.com/xdslproject/xdsl/pull/1256), now they are also using the same custom assembly format.


kingiler (approved):
LGTM

superlopuh (approved):

PR #1418, opened 2023-08-04
backend: (riscv) Refactor lhs and rhs in LowerArithCmpf

kingiler:


superlopuh:
Let's close this, and rather split up the other PR

PR #1422, opened 2023-08-04
interpreter: (riscv) Add some libc runtime things to interpreter

AntonLydike:
Adds `malloc`, `free`, `calloc` and `putchar` to the riscv runtime. Still needs to be hooked up, should probably use a decorator and fancy class to handle everything.

### DIFF
--- a/xdsl/interpreters/memref.py
+++ b/xdsl/interpreters/memref.py
@@ -10,8 +10,8 @@ from xdsl.utils.exceptions import InterpretationError
 
 
 class MemrefValue(Enum):
-    Allocated = 1
-    Deallocated = 2
+    Initialized = 1
+    Freed = 2
 
 
 @register_impls
@@ -22,7 +22,7 @@ class MemrefFunctions(InterpreterFunctions):
 
         shape = memref_type.get_shape()
         size = prod(shape)
-        data = [MemrefValue.Allocated] * size
+        data = [MemrefValue.Initialized] * size
 
         shaped_array = ShapedArray(data, list(shape))
         return (shaped_array,)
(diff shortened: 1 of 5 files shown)


xdsl/interpreters/memref.py
```diff
@@ -10,8 +10,8 @@
 
 
 class MemrefValue(Enum):
-    Allocated = 1
-    Deallocated = 2
+    Initialized = 1
```
superlopuh, line 13: But it's not initialized? It's only allocated?

superlopuh (changes requested):
I'm not sure this is ready for review. I would recommend adding an isolated test that does not modify any existing interpreters. Those changes are useful but seem like working towards a future PR that actually leverages this functionality. I'd probably recommend adding a minimal module with a `main` function that calls an external `foo` function that is defined in Python with your proposed mechanism.

PapyChacal:
Interesting, I didn't think of such a solution!

I'm also thinking about using ctypes to actually link and call real compiled external functions at some point, instead of reimplementing a whole libc in Python or what not; but I do understand you at least want to get started before this happens; and maybe just want to have a way to call Python helpers from IR, even if linking capabilities would be there?

So I'm just stating that to see if we can find something that would fit both usecase.
I guess we could use this proposed approach, and fallback on linked libraries if no Python implementation was provided? Would that conflict with what you have in mind?

PapyChacal:
By the way, why implement such a generic hook for any external e.g. `malloc` symbol and present it as part of a RISC-V runtime?
AFAIU, this allows using it completely out of the RISC-V setting too, or am I missing something? (Which is nice!)

AntonLydike:
Closed in favor of #1536 and #1537

PR #1423, opened 2023-08-07
dialects: (riscv_scf) add while op

superlopuh:


xdsl/dialects/riscv_scf.py
```diff
@@ -74,10 +74,63 @@ def __init__(
...
+        before_region: Region | Sequence[Operation] | Sequence[Block],
+        after_region: Region | Sequence[Operation] | Sequence[Block],
+    ):
+        super().__init__(
+            operands=[arguments],
+            result_types=[result_types],
+            regions=[before_region, after_region],
+        )
+
+    # TODO verify dependencies between riscv_scf.condition, riscv_scf.yield and the regions
```
superlopuh, line 99: Comment copied over from scf.while

xdsl/dialects/riscv_scf.py
```diff
@@ -74,10 +74,63 @@ def __init__(
         )
 
 
+@irdl_op_definition
+class WhileOp(IRDLOperation):
+    name = "riscv_scf.while"
+    arguments: VarOperand = var_operand_def(IntRegisterType)
```
adutilleul, line 80: `ForOp` uses `RISCVRegisterType` for arguments (so either float or int), we should probably do the same here to be future proof

adutilleul (changes requested):

nazavode (approved):

xdsl/dialects/riscv_scf.py
kingiler: I'm not sure we allow float for `ForOp`? It's kind of weird.

xdsl/dialects/riscv_scf.py
```diff
@@ -74,10 +74,63 @@ def __init__(
...
+    arguments: VarOperand = var_operand_def(RISCVRegisterType)
+
+    res: VarOpResult = var_result_def(RISCVRegisterType)
+    before_region: Region = region_def()
+    after_region: Region = region_def()
+
+    def __init__(
+        self,
+        arguments: Sequence[SSAValue | Operation],
+        result_types: Sequence[IntRegisterType],
```
kingiler, line 89: `RISCVRegisterType`?

xdsl/dialects/riscv_scf.py
adutilleul: RISCVRegisterType is the parent class for both, so we do allow in fact.
The only limitation is the current lowering, but that'll be also changed in the nearby future.

xdsl/dialects/riscv_scf.py
```diff
@@ -74,10 +74,63 @@ def __init__(
...
+        super().__init__(
+            operands=[arguments],
+            result_types=[result_types],
+            regions=[before_region, after_region],
+        )
+
+    # TODO verify dependencies between riscv_scf.condition, riscv_scf.yield and the regions
+    def verify_(self):
+        for idx, arg in enumerate(self.arguments):
+            if self.before_region.block.args[idx].type != arg.type:
```
kingiler, line 102: Can we use `zip` instead of `enumerate` + `idx`?

kingiler (changes requested):

xdsl/dialects/riscv_scf.py
superlopuh: We could, but I prefer this version because when I break on a line like this with an exception breakpoint I get the index directly.

kingiler: Addressed in newest commit. Good to go.

kingiler (approved):
LGTM

PR #1424, opened 2023-08-07
 dialects: (riscv_scf) add verifier to `ForOp`

adutilleul:
The verifier from the `scf` dialect was not carried over to the `For` operation, hence some mistakes can be easily be made when using this operation manually or writing a lowering for it.

This PR adds such a verifier. Filechecks from `scf` have not been ported yet, but if people feel that it's necessary, will do it.


superlopuh (approved):

nazavode (approved):

kingiler (approved):
I'm not sure but can these be represented using traits?

xdsl/dialects/riscv_scf.py
```diff
@@ -73,6 +74,41 @@ def __init__(
...
+                f"variable and loop-carried variables as arguments."
+            )
+        if self.body.block.args and (iter_var := self.body.block.args[0]):
+            if not isinstance(iter_var.type, IntRegisterType):
+                raise VerifyException(
+                    f"The first block argument of the body is of type {iter_var.type}"
+                    " instead of riscv.IntRegisterType"
+                )
+        for idx, arg in enumerate(self.iter_args):
+            if self.body.block.args[idx + 1].type != arg.type:
```
kingiler, line 91: Can we use `zip` instead of `enumerate` + `idx`?

xdsl/dialects/riscv_scf.py
```diff
@@ -73,6 +74,41 @@ def __init__(
...
+                raise VerifyException(
+                    f"Block arguments with wrong type, expected {arg.type}, "
+                    f"got {self.body.block.args[idx].type}. Arguments after the "
+                    f"induction variable must match the carried variables."
+                )
+        if len(self.body.ops) > 0 and isinstance(self.body.block.last_op, YieldOp):
+            yieldop = self.body.block.last_op
```
kingiler, line 98: nit: 
```suggestion
        if len(self.body.ops) > 0 and isinstance(yieldop := self.body.block.last_op, YieldOp):
```

xdsl/dialects/riscv_scf.py
```diff
@@ -73,6 +74,41 @@ def __init__(
...
+            if len(yieldop.arguments) != len(self.iter_args):
+                raise VerifyException(
+                    f"Expected {len(self.iter_args)} args, got {len(yieldop.arguments)}. "
+                    f"The riscv_scf.for must yield its carried variables."
+                )
+            for idx, arg in enumerate(yieldop.arguments):
+                if self.iter_args[idx].type != arg.type:
```
kingiler, line 105: Can we use zip here as well?

kingiler (changes requested):
Some nits

xdsl/dialects/riscv_scf.py
superlopuh: I like the idx version more for exceptions, but it's a small difference anyway. I'm not sure it's worth the time changing it one way or another.

superlopuh:
In general, I would avoid blocking PRs on nit comments.

kingiler:
Okay, let me say they are not nit.

kingiler (approved):
LGTM

xdsl/dialects/riscv_scf.py
```diff
@@ -87,10 +87,16 @@ def verify_(self):
                     f"The first block argument of the body is of type {iter_var.type}"
                     " instead of riscv.IntRegisterType"
                 )
-        for arg, block_arg in zip(self.iter_args, self.body.block.args[1:]):
+        for idx, (arg, block_arg) in enumerate(
+            zip(
+                self.iter_args,
+                self.body.block.args[1:],
+                strict=True,
```
adutilleul, line 94: This `strict` parameter is not required since the verifier checks earlier that `(len(self.iter_args) + 1) == len(self.body.block.args)`.
Also, while I don't mind too much the change here, could you please use next time the comment feature like Sasha suggested ?

kingiler:
I do use GitHub comments. So you forbid code contribution?

adutilleul:
I prefer "suggesting" the change using the comment feature on it rather than committing on someone branch directly.
I mean the first change was a matter of opinion and I feel that's still the case for this commit and pushing it before discussing it just feels strange.

superlopuh:
Hey all, we seem to have some diverging development styles and it seems to harm the overall team productivity. In particular, the established development model in xDSL seems to not match the current workflow of @kingiler. I think that we should make sure that we’re on the same page on how best to collaborate going forward.

@kingiler, xDSL does not yet have a clear contribution guide, but some key guidelines that are well-established are: (a) we do not push to other peoples PRs unless requested to do so, (b) we block PRs with change-requests only once we are certain this blocking is necessary, especially if we are new in an area of xDSL, (c) we communicate in PRs to reach consensus on what changes to make to the project. We as the xDSL team really value your contributions over the last weeks and I would like for you to become an even more productive member of our team. The guidelines above have helped us work effectively so far. While it can be tempting to make the changes you think are best directly, it is important to ensure that the necessary consensus has been reached to keep us as a team productive. Could you please not push to other people’s branches, avoid using changes-requested for now, and also not merge pull requests while there is still a disagreement on the best way to move forward? There is no rush to get the changes in, and it is more important that we are on the same page as a team.

Don’t hesitate to ping me if you have any questions about other aspects of productive collaboration in xDSL. I value your feedback and would like to make the most out of our collaboration.

tobiasgrosser:
Hey @superlopuh , thank you for bringing this up. These are indeed well-established practices in xDSL and important to follow. I am sorry that I was unavailable when @kingiler  started on xDSL, so I missed sharing the community best practices. Thank you for doing this here for us. @kingiler, I am back in Edinburgh next Friday and am happy to chat if there are questions on how to be most productive in xDSL, but don’t hesitate to rely on @superlopuh to learn more about best practices in xDSL or to hit me up on zulip beforehand.

Thank you all for collaborating at this scale in xDSL. This is certainly not easy, also in a social sense, and seeing us all making it happen anyhow is actually quite impressive.

PR #1429, opened 2023-08-09
documentation: (toy) lower all the way to assembly and emulate in riscemu (no loops)

superlopuh:


nazavode (approved):

adutilleul (approved):

docs/Toy/toy/emulator/toy_accelerator_instructions.py
```diff
@@ -0,0 +1,203 @@
...
+        return str(data)
+    if len(shape):
+        size = reduce(lambda acc, el: acc * el, shape[1:], 1)
+        inner = (
+            tensor_description(shape[1:], data[start : start + size])
+            for start in range(0, size * shape[0], size)
+        )
+        return f'[{", ".join(inner)}]'
+    else:
+        return "[]"
```
kingiler, line 29: Any usage for this func?

superlopuh: You're right, it's no longer needed

docs/Toy/toy/emulator/toy_accelerator_instructions.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.set_reg(destination_ptr_reg, result_ptr)
+
+    def __eq__(self, __value: object) -> bool:
+        if not isinstance(__value, ToyAccelerator):
+            return False
+        return self.stream is __value.stream
+
+    def __hash__(self) -> int:
+        return hash(id(self.stream))
```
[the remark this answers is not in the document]
superlopuh: I don't think it is, strictly speaking, but my understanding of hashing is that some hash functions enable better performance than others. If the implementation of `id` returns the pointer to the python object, then it's likely to have some alignment, meaning that the lower bits are always 0, which may negatively impact some hash table implementations. It feels prudent to add a `hash` function call to make the probability of this being a problem lower, for a relatively low overhead. This is intuition over from other programming languages, maybe it doesn't hold for Python in general...

superlopuh:
#1427

PR #1438, opened 2023-08-09
core: (trait) Polish HasParent trait

kingiler:
Main change is the use of `*args` for `parameters`. Another change is using generators instead of list.

### DIFF
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -42,22 +42,20 @@ class HasParent(OpTrait):
 
     parameters: tuple[type[Operation], ...]
 
-    def __init__(self, parameters: type[Operation] | tuple[type[Operation], ...]):
-        if not isinstance(parameters, tuple):
-            parameters = (parameters,)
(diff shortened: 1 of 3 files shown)


xdsl/traits.py
```diff
@@ -42,22 +42,20 @@ class HasParent(OpTrait):
 
     parameters: tuple[type[Operation], ...]
 
-    def __init__(self, parameters: type[Operation] | tuple[type[Operation], ...]):
-        if not isinstance(parameters, tuple):
-            parameters = (parameters,)
+    def __init__(self, *parameters: type[Operation]):
         if len(parameters) == 0:
```
superlopuh, line 46: Nit:
```suggestion
        if parameters:
```

superlopuh (approved):
LGTM

xdsl/traits.py
kingiler: Done in latest commit.

PapyChacal (approved):
Nice 🙂

PR #1441, opened 2023-08-10
[WIP] dialect: (riscv) Custom function type

kingiler:
Implementing custom function type for riscv ops.

If there is variadic operands or result types, then generic function type is used for print and parsing.
We will drop the function type if both operands and results are empty.
If either operands or results are empty, we will only print and parse the other component.
Generic function type is used if both of operands and results are non-empty.


superlopuh (dismissed):
In my understanding, dropping `() -> ()` is worth doing when it's not informative, as in the operation can only have no operands and no results. Some operations have optional operands or results, and I think that we would still benefit from having the `() -> ()` there to signal that explicitly. Could you please move the functionality only to `NullaryOperation` for now, and leaving the operation type on the others as-is?

kingiler:
@superlopuh This is still a very early WIP draft PR. I don't understand why blocking it?

superlopuh:
Ah yes sorry I'll unblock it

PapyChacal:
> In my understanding, dropping `() -> ()` is worth doing when it's not informative, as in the operation can only have no operands and no results. Some operations have optional operands or results, and I think that we would still benefit from having the `() -> ()` there to signal that explicitly. Could you please move the functionality only to `NullaryOperation` for now, and leaving the operation type on the others as-is?

I don't agree in general, having no results is pretty clear (no defined values) and many operations use this for brevity with variadic operands. See `scf.yield` and `func.return`.
Then I don't know the RISC-V ones in depth, not sure for those

superlopuh:
Ah good point, I'm not as familiar with those, if that's what they do then maybe we should follow suit. I would still restricting the implementation to the specific operations we want to have this functionality, even if that means duplicating a bit of code.

kingiler:
@superlopuh I'm back. Looking for ideas on case where operands are empty or results are empty.

kingiler:
Currently is the prototype of custom function type:
```
%0 = riscv.result_only : !riscv.reg<>
riscv.operand_only %0 : !riscv.reg<>
riscv.none
%1 = riscv.both %0 :  (!riscv.reg<>) -> !riscv.reg<>
```
I am also open to alternative style
```
%0 = riscv.result_only -> !riscv.reg<>
riscv.operand_only %0 : !riscv.reg<>
riscv.none
%1 = riscv.both %0 :  (!riscv.reg<>) -> !riscv.reg<>
```

superlopuh:
I would recommend merging things one by one, because some decisions are much easier than others. For example, an operation with no results and no operands pretty clearly doesn't benefit from a type annotation:

```
riscv.ret
riscv.ecall
```

After that, I'd probably recommend doing the single return case, here I think it would make sense to mirror arith.constant et co
```
%c5 = riscv.li 5: !riscv.reg<>
```

After those are in, let's discuss the cases that are left?

superlopuh:
Hello, I've started working on this again, here's a first PR #1683, @kingiler do you feel like working on this still? There will be one more step after the NullaryOp, which is for ops that only benefit from the `rd` register type being printed, like `%rd = riscv.add %a, %b : !riscv.reg<a0>`.

tobiasgrosser:
This has meanwhile a lot of conflicts. I close this for now. We can add things incrementally through smaller PRs.

PR #1442, opened 2023-08-11
ci: (ruff) Add rule UP034

kingiler:
The rule is to avoid extraneous parentheses.

A bug was found when adding ruff rule UP034. The code intends to have a tuple of strings but a missing comma makes it becomes extraneous parentheses.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -2380,7 +2380,7 @@ class AssemblySectionOp(IRDLOperation, RISCVOp):
     @classmethod
     def parse(cls, parser: Parser) -> AssemblySectionOp:
         directive = parser.parse_str_literal()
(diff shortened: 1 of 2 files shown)


xdsl/dialects/riscv.py
```diff
@@ -2380,7 +2380,7 @@ def __init__(
     @classmethod
     def parse(cls, parser: Parser) -> AssemblySectionOp:
         directive = parser.parse_str_literal()
-        attr_dict = parser.parse_optional_attr_dict_with_keyword(("directive"))
+        attr_dict = parser.parse_optional_attr_dict_with_keyword(("directive",))
```
superlopuh, line 2383: Yay for python strings being iterables of string 😅

superlopuh (approved):
nice find

adutilleul (approved):

PR #1444, opened 2023-08-11
ci: (ruff) Add rule UP007

kingiler:
The rule is to use `X | Y` instead of `Union[X, Y]` for type annotations.


superlopuh (approved):
LGTM

math-fehr (approved):
:+1:

superlopuh:
Looks like this branch got a little messed up by the merges, the CI for the branch targeting this is almost green, so they should both be possible to merge pretty soon

PR #1445, opened 2023-08-11
ci: (ruff) Add rule UP038

kingiler:
The rule is to use `X | Y` instead of `(X, Y)` for `isinstance` and `issubclass` methods. See PEP604.


superlopuh (approved):
LGTM

adutilleul (approved):

math-fehr (approved):
:+1:

PR #1446, opened 2023-08-11
ci: (riscv) Add rule UP006

kingiler:
The rule is to replace the use of `typing.List` to `list` and some other standard libraries collections. See [PEP585](https://peps.python.org/pep-0585/).


superlopuh (approved):
LGTM

math-fehr (approved):
:+1:

PR #1447, opened 2023-08-11
ci: (ruff) Add rule UP032

kingiler:
This rule replaces `str#format` with `f-string` if possible.

### DIFF
--- a/xdsl/dialects/memref.py
+++ b/xdsl/dialects/memref.py
@@ -544,23 +544,17 @@ class DmaStartOp(IRDLOperation):
 
         if len(self.src.type.shape) != len(self.src_indices):
             raise VerifyException(
-                "Expected {} source indices (because of shape of src memref)".format(
-                    len(self.src.type.shape)
-                )
+                f"Expected {len(self.src.type.shape)} source indices (because of shape of src memref)"
             )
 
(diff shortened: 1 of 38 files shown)


xdsl/dialects/memref.py
```diff
@@ -544,23 +544,17 @@ def verify_(self) -> None:
...
+                f"Expected {len(self.dest.type.shape)} dest indices (because of shape of dest memref)"
             )
 
         if len(self.tag.type.shape) != len(self.tag_indices):
             raise VerifyException(
-                "Expected {} tag indices (because of shape of tag memref)".format(
-                    len(self.tag.type.shape)
-                )
+                f"Expected {len(self.tag.type.shape)} tag indices (because of shape of tag memref)"
             )
```
kingiler, line 558: It is unrelated to this PR but it seems the messages can be more helpful? In our code base, we have something like `expect ..., get ...`.

superlopuh: I 100% agree, it would be nice to work on this at some point soon. A more declarative verification API would allow us to have more unified error messages and some other goodies. I've been noting down use-cases for which a declarative API would be useful, and am just waiting for it to fit in with our other research plans :)

superlopuh (approved):

math-fehr (approved):

PR #1452, opened 2023-08-13
[WIP] little jit experiment

superlopuh:
What happens when we compile an MLIR module and execute it as a Python function? With a naive scf for loop implementation of a dot product, we're an order of magnitude slower than numpy, and an order of magnitude faster than Python.

Screenshot 2023-08-13 at 20 08 24


superlopuh:
I just wanted a record of this happening, I'll close this PR for now

PR #1457, opened 2023-08-15
backend: (riscv) fix riscv_scf lowering

superlopuh:
I added a test for the lowering in Toy, and found a bug in our for loop lowering code. Hurray for tests :)

### DIFF
--- a/xdsl/backend/riscv/riscv_scf_to_asm.py
+++ b/xdsl/backend/riscv/riscv_scf_to_asm.py
@@ -53,7 +53,7 @@ class LowerRiscvScfToLabels(RewritePattern):
         # upper bound and branching to the loop body if the condition is met.
         rewriter.insert_op_before_matched_op(
             [
-                get_loop_var := riscv.GetRegisterOp(loop_var_reg),
+                get_loop_var := riscv.MVOp(op.lb, rd=loop_var_reg),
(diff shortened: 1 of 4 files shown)


xdsl/backend/riscv/riscv_scf_to_asm.py
```diff
@@ -53,7 +53,7 @@ def match_and_rewrite(self, op: riscv_scf.ForOp, rewriter: PatternRewriter, /):
         # upper bound and branching to the loop body if the condition is met.
         rewriter.insert_op_before_matched_op(
             [
-                get_loop_var := riscv.GetRegisterOp(loop_var_reg),
+                get_loop_var := riscv.MVOp(op.lb, rd=loop_var_reg),
```
superlopuh, line 56: When in a nested loop, the iteration index variable was never reset to the beginning of the loop.

nazavode (approved):
asm ftw

xdsl/backend/riscv/riscv_scf_to_asm.py
tobiasgrosser: Should this not also update a filecheck test case?

superlopuh: yep, fixing

PR #1458, opened 2023-08-15
backend: (riscv) fix scf to riscv_scf lowering, add matmul test

superlopuh:
Similar to the index variable bug, we never reset the accumulator variables, so it just kept accumulating across all for loops.


xdsl/backend/riscv/register_allocation.py
```diff
@@ -194,12 +194,18 @@ def allocate_registers(self, module: ModuleOp) -> None:
         """
         for op in module.walk():
             if isinstance(op, riscv_scf.ForOp):
-                for arg in op.body.block.args:
+                for i, arg in enumerate(op.body.block.args):
```
superlopuh, line 197: I forgot this change was in this branch... I think I'll separate it out into another PR

nazavode (approved):

PR #1460, opened 2023-08-15
backend: (riscv) add register allocation for loop-carried scf_for vars

superlopuh:
This got a little trickier, since the operands, loop variables, yielded operands, and results all need to have the same registers.


nazavode (approved):

AntonLydike (approved):
The register allocator looks like it could use some factoring as it's getting quite larger, but otherwise this looks good (I think?)

adutilleul (approved):

PR #1463, opened 2023-08-16
backend: (riscv) move convert_float_to_int to own helpers file in utils

superlopuh:


tests/utils/test_bitwise_casts.py
```diff
@@ -0,0 +1,20 @@
...
+    "i, f",
+    [
+        (0b00000000000000000000000000000000, 0.0),
+        (0b10000000000000000000000000000000, -0.0),
+        (0b00111111100000000000000000000000, 1.0),
+        (0b01000000000000000000000000000000, 2.0),
+    ],
+)
+def test_float_bitwise_casts(i: int, f: float):
+    # assert i == 0
```
kingiler, line 17: What is this comment for?

tests/utils/test_bitwise_casts.py
```diff
@@ -0,0 +1,20 @@
...
+    assert convert_f32_to_u32(f) == i
+    # N.B. this is an IEEE float comparison, 0.0 == -0.0, but still a useful test
```
kingiler, line 19: Again what is this comment about?

superlopuh: Just a reminder to people reading the code that it's not a perfect comparison

tests/utils/test_bitwise_casts.py
superlopuh: debugging the test, will remove, ty

tests/utils/test_bitwise_casts.py
superlopuh: Clarified it a little

kingiler: So we want bit-wise float comparison? Or something else?

superlopuh: It would be nice but there is no such thing. It's explicitly not a supported operation, and converting both operands to int and comparing ints is the recommended way to do it, sort of what this test is testing in the first place.

tests/utils/test_bitwise_casts.py
```diff
@@ -0,0 +1,19 @@
...
+        (0b00000000000000000000000000000000, 0.0),
+        (0b10000000000000000000000000000000, -0.0),
+        (0b00111111100000000000000000000000, 1.0),
+        (0b01000000000000000000000000000000, 2.0),
+    ],
+)
+def test_float_bitwise_casts(i: int, f: float):
+    assert convert_f32_to_u32(f) == i
+    # N.B. this is an IEEE float comparison, 0.0 == -0.0, so there may be false positives
+    assert convert_u32_to_f32(i) == f
```
kingiler, line 19:
```suggestion
    assert struct.pack(">f", convert_u32_to_f32(i)) == struct.pack(">f", f)
```

kingiler (changes requested):
Bit-wise float compare.

tests/utils/test_bitwise_casts.py
kingiler: See requested changes for bit-wise float compare.

superlopuh: Good idea

adutilleul (approved):

kingiler:
After playing with python with bits and bytes. I come up with #1469 which have a consistent implementation without resorting to `ctypes` and pointers.

superlopuh:
@kingiler can you please update the review? This PR is now blocking some other work.

superlopuh:
I've addressed all the comments, will merge as-is, happy to address further comments in follow-up PRs.

kingiler:
@superlopuh I have been off from work recently. Sorry for the late reply and the blockage. I'm glad that the PR get merged.

superlopuh:
No worries, thank you for writing, please let me know if there are more changes that you think need to be made

PR #1464, opened 2023-08-16
backend: (riscv) unify GetRegister ops

superlopuh:


xdsl/dialects/riscv.py
```diff
@@ -211,6 +211,9 @@ def abi_index_by_name(cls) -> dict[str, int]:
     }
 
 
+RD = TypeVar("RD", bound=RISCVRegisterType)
```
kingiler, line 214: I'm not sure but it is confusing at first to see why it is `RD`. I would recommend `RdT` for naming conventions for type vars. Better with `RdInvT` as seen in code base.

xdsl/dialects/riscv.py
```diff
@@ -2506,8 +2509,7 @@ class WfiOp(NullaryOperation):
 # region RISC-V SSA Helpers
 
 
-@irdl_op_definition
-class GetRegisterOp(IRDLOperation, RISCVOp):
+class GetAnyRegisterOperation(Generic[RDInvT], IRDLOperation, RISCVOp):
     """
     This instruction allows us to create an SSAValue with for a given integer register name. This
```
nazavode, line 2514:
```suggestion
    This instruction allows us to create an SSAValue with for a given register name. This
```

nazavode (approved):

kingiler (approved):
Good and concise. Ty

adutilleul (approved):

PR #1466, opened 2023-08-16
ci: (ruff) Add rule PT for pytest

kingiler:
Since we are using `pytest`, it would be better if our tests are written in a consistent and concise way. Ruff's [PT](https://beta.ruff.rs/docs/rules/#flake8-pytest-style-pt) is based on flake8's pytest style.

Some rules are ignored in this PR to minimize the diff.
```diff
-select = ["E", "F", "W", "I", "UP"]
-ignore = ["E741"]
+select = ["E", "F", "W", "I", "UP", "PT"]
+ignore = ["E741", "PT006", "PT007", "PT011", "PT012", "PT015", "PT017", "PT018"]
```

This series of PR should not cause any API changes in the core xDSL code base but only for the tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,8 +25,8 @@ typeCheckingMode = "strict"
 profile = "black"
 
 [tool.ruff]
-select = ["E", "F", "W", "I", "UP"]
-ignore = ["E741"]
+select = ["E", "F", "W", "I", "UP", "PT"]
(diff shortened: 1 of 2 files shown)


adutilleul (approved):

AntonLydike (approved):

math-fehr (approved):
Nice! That makes sense!

PR #1467, opened 2023-08-16
core: unify some attribute printing/parsing methods

superlopuh:
I implemented the custom parsing syntax for `func.call`, and found that I needed to make some significant changes to the helpers to make it work. The attribute dictionary is not preceded by `attributes` in `func.call`, since there's no region for it to be confused with.

### DIFF
--- a/xdsl/parser/core.py
+++ b/xdsl/parser/core.py
@@ -660,9 +660,23 @@ class Parser(AttrParser):
         dictionary, and usually correspond to the names of the attributes that are
         already passed through the operation custom assembly format.
         """
-        begin_pos = self.lexer.pos
         if self.parse_optional_keyword("attributes") is None:
             return None
+        return self.parse_optional_attr_dict_with_reserved_attr_names(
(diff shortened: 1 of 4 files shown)


xdsl/parser/core.py
```diff
@@ -660,9 +660,23 @@ def parse_optional_attr_dict_with_keyword(
...
+        self, reserved_attr_names: Iterable[str] = ()
+    ) -> DictionaryAttr | None:
+        """
+        Parse a dictionary attribute if present.
+        This is intended to be used in operation custom assembly format.
+        `reserved_attr_names` contains names that should not be present in the attribute
+        dictionary, and usually correspond to the names of the attributes that are
+        already passed through the operation custom assembly format.
+        """
+        begin_pos = self.lexer.pos
```
superlopuh, line 679: slight change in behaviour, but I think it might be a bit more helpful, since it'll point to the dictionary directly

AntonLydike (approved):

PR #1469, opened 2023-08-17
utils: (bits) add utilities for bits cast and comparison

kingiler:
A utility library to cast i8, i16, i32, i64, u8, u16, u32, u64, f16, f32, f64 from and to bytes (i.e. bits).

For all integers from and to are round-tripping ops.
The case for float is tricky and only holds when the value is not `nan` or `inf` as IEE754 spec.


superlopuh:
Thank you for this PR. What is this working towards? I'm not sure that we need a generic bit converting utility file

math-fehr (commented):
Why do we need those functions?
I feel that putting them in `xdsl/utils` adds a bit of unwanted stuff, so maybe another place would be better?
Also, could you maybe reduce the amount of parameters in parametrized tests? I would feel that 256 parameters might be a bit annoying (and makes the UI quite unreadable probably)

tests/utils/test_bits.py
```diff
@@ -0,0 +1,107 @@
...
+)
+
+
+@pytest.mark.parametrize(
+    "i",
+    (randrange(1 << 8) for _ in range(256)),
+)
+def test_8bits(i: int):
+    b = from_u8(i)
+    assert as_str(b).endswith(bin(int(b.hex(), 16)).removeprefix("0b"))
```
AntonLydike, line 39: What's the purpose of this line? Couldn't it just be `as_str(b)[:-8] == bin(i)[:-8]`? Why the hex to int to bin?

AntonLydike (changes requested):
Instead of a thousand test casses for these one-line functions, I would rather like to see one or two well thought out tests that test all the functions. I understand the desire to fuzz these kinds of things, but I'm not sure that's necessary here. Having a few simple tests (that maybe even document how these functions intend to be used) would be much more valuable imo.

tests/utils/test_bits.py
```diff
@@ -0,0 +1,107 @@
...
+    to_u16,
+    to_u32,
+    to_u64,
+)
+
+
+@pytest.mark.parametrize(
+    "i",
+    (randrange(1 << 8) for _ in range(256)),
+)
```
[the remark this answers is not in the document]
kingiler: Notice it is a generator instead of tuple or list. Also it is mimicking QuickCheck in Haskell. 256 is just arbitrary and consistent for all bits (8 / 16/ 32 / 64). I could have 256 for 8bits and 65536 for 16/32/64 bits.

tests/utils/test_bits.py
kingiler: Good question I see what you intend to do. Minor note is that `as_str` always has fixed length (as indicated by the bytes provided) but `bin` has dynamic length (i.e. trailing zeros are omitted) and also leading `0b` prefix.

superlopuh:
My current impression is that we don't have a good use for these helpers for the foreseeable future. Let's close this PR for now and reopen if they become a part of a plan towards a specific project?

superlopuh:
Closing for now

PR #1470, opened 2023-08-17
ci: (ruff) add rule PT017

kingiler:
The rules checks for assert statements in except clauses.


math-fehr (approved):

nazavode (approved):

AntonLydike:
Why are these rules needed in the codebase? I am not sure if I see a point in adding a bunch of ruff rules that don't also change the code? If we don't already break these rules, we may as well not have the rule, or am I wrong? This rule for example feels super niche, and all the noise created by a PR per rule feels incredibly unproductive...

superlopuh (approved):

kingiler:
@AntonLydike  The rule is `PT` which consists of a bunch of rules related to `pytest`. Rules that will create huge diffs are temporarily ignored in the main PR #1466 and subsequently removed from ignore list in order to minimize the diff. Or you prefer adding all rules resulting in a huge diff?

AntonLydike:
> @AntonLydike The rule is `PT` which consists of a bunch of rules related to `pytest`. Sub-rules that will create huge diffs are temporarily ignored in the main PR #1466 and subsequently removed from ignore list in order to minimize the diff. Or you prefer adding all rules resulting in a huge diff?

Hmm, I thought this was the PR that added a ruff rule that had no impact on existing code, but I guess it was the wrong PR. Must have gotten confused in the storm of ruff rule PRs

PR #1472, opened 2023-08-17
ci: (ruff) add rule PT012

kingiler:


math-fehr (approved):

AntonLydike:
Can you please add a description to your PRs? Especially these here are very difficult to understand without a good description where they show off what they change. We even have that in the PR template now.

kingiler:
@AntonLydike it is still a draft PR.

tests/test_op_builder.py
kingiler, line 1: Wrap the builder sequence with a function so `pytest.raise` will only contain that function./

kingiler (commented):
Most changes are moving irrelevant code out of the `pytest.raise`. Some changes are to improve usage.

superlopuh:
Is this transformation done automatically? I'm not sure it's actually helpful... the `build` version is not any more precise and is less readable than the current code

kingiler:
@superlopuh Unfortunately the issue found by PT012 are currently non-fixable, probably due to no general fix method. Most of the tests can be simplified and thus easy to fix. The only one I am not sure is the `Builder` one. Do you have any suggestions?

superlopuh:
Ok that makes sense. I'd probably lean towards not removing this exception in that case. There'll be a bit of friction especially for new contributors, by some rule that's not auto-fixable, I'm not sure it's worth the hassle. I agree that it's nice to have a single expression in the `throws` context but we can do that by hand, I think, without ruff enforcing it.

superlopuh:
I'm still thinking we want to keep this exception in, I don't think having a requirement of a single expression gets us more than the friction it creates.

superlopuh:
I'll close it for now, let's reopen if we change our minds.

PR #1481, opened 2023-08-20
dialects: (func) add custom syntax to func.call

superlopuh:


tobiasgrosser (approved):
Lgtm

xdsl/dialects/func.py
```diff
@@ -322,6 +322,36 @@ def get(
...
+        printer.print_string(")")
+        printer.print_op_attributes(self.attributes, reserved_attr_names=("callee",))
+        printer.print_string(" : ")
+        printer.print_operation_type(self)
+
+    @classmethod
+    def parse(cls, parser: Parser) -> Call:
+        callee = parser.parse_symbol_name()
+        unresolved_arguments = parser.parse_comma_separated_list(
+            Parser.Delimiter.PAREN, lambda: parser.parse_unresolved_operand()
```
math-fehr, line 339:
```suggestion
            Parser.Delimiter.PAREN, parser.parse_unresolved_operand
```
Actually, if I recall, wasn't there an operation that does exactly this?

math-fehr (approved):

xdsl/dialects/func.py
superlopuh: indeed

PR #1483, opened 2023-08-20
transforms: (affine) add basic lower-affine

superlopuh:
This is technically not necessary for the matmul things we're working on, but would let me delete quite a lot of hacky code in Toy, and would let us have a pipeline that's fully in Python and under our control, both of which seem quite nice things to have.


xdsl/transforms/lower_affine.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Inserts operations that evaluate the affine map when given input SSA values.
+    """
+    if map is None:
+        indices = dims
+    else:
+        indices: list[SSAValue] = []
+        for expr in map.data.results:
+            ops, val = affine_expr_ops(expr, dims, [])
+            rewriter.insert_op_before_matched_op(ops)
```
webmiche, line 75: I would personally not pass the rewriter in here and rewrite, but rather return the ops+SSAvalues and then rewrite in the RewritePattern. This way, one could use this function also in a context that does not immediately rewrite (e.g. in testing). Plus, I don't quite like having the rewriting happening somewhere else than in the RewritePattern class, just from a coding style perspective, but that's just a personal preference.

xdsl/transforms/lower_affine.py
```diff
@@ -0,0 +1,148 @@
...
+) -> tuple[list[Operation], list[SSAValue]]:
+    """
+    Returns operations that evaluate the affine map when given input SSA values and the
+    resulting indices.
+    """
+    if map is None:
+        ops = []
+        indices = dims
+    else:
+        ops: list[Operation] = []
```
webmiche, line 73: Why define ops to times and not just do it like this:

```suggestion
     ops: list[Operation] = []
     if map is None:
        indices = dims
    else:
```

superlopuh: In a language that isn't python those might have slightly different semantics but this is not the case here :)

webmiche: I just prefer reading the lower thing, but I also don't have preferences.

I guess I am just more used to block-scopes than function-scopes from reading c++...

superlopuh: Done :)

webmiche (approved):

xdsl/transforms/lower_affine.py
```diff
@@ -0,0 +1,147 @@
...
+            case AffineBinaryOpKind.Mod:
+                op = arith.RemSI(lhs_val, rhs_val)
+            case AffineBinaryOpKind.FloorDiv:
+                op = arith.FloorDivSI(lhs_val, rhs_val)
+            case AffineBinaryOpKind.CeilDiv:
+                op = arith.CeilDivSI(lhs_val, rhs_val)
+
+        return [*lhs_ops, *rhs_ops, op], op.result
+
+    raise ValueError("Unreachable")
```
math-fehr, line 57: I would `assert` here instead.

math-fehr (approved):
Very nice!

PR #1485, opened 2023-08-21
dialects: (riscv_func) add custom syntax to mirror func

superlopuh:
Will make some filecheck files with upcoming changes easier to read


AntonLydike (commented):
Can we abstract the func, return and call custom syntax to a python mixin? That would be super nice to have so we can get the same for llvm syntax, and other func-like ops

superlopuh:
We definitely could, I'm not sure that we necessarily want to do that right now. What do you think about me filing that as an issue for later?

JosseVanDelm (commented):
Really cool! However, I get a bit the feeling that there's quite a bit of code duplication involved for parsing and printing. Is none of this somewhat shareable with func dialect?

superlopuh:
It seems that @AntonLydike and @JosseVanDelm agree, I'm just not sure that this is the PR to do this factoring in? If one more person tells me to do it I'll do it :)

adutilleul (approved):
I also have the same feeling regarding the duplication. Since we're importing more and more from standards dialects into the riscv codebase, maintaining duplicated code on the long term is kinda annoying.
However if you feel that we should move on and open an issue for that, that also seems great to me.

superlopuh:
all done, factored code into xdsl.dialects.utils

nazavode (approved):

xdsl/dialects/utils.py
superlopuh, line 1: pretty much just moved from func.py

AntonLydike (approved):
:heart_eyes:

PR #1486, opened 2023-08-21
backend: (riscv) use canonicalization to optimize riscv

superlopuh:
Also fixes some other issues:

 - redundant mv elimination should only eliminate allocated registers, not unallocated
 - I forgot to register `CanonicalizePass` with xdsl-opt
 - move the file definiing the canonicalization pattern one level up in the hierarchy


AntonLydike (approved):
LGTM

PR #1488, opened 2023-08-21
core: Add properties to Operation

math-fehr:
This PR adds properties to operations.
Properties are inherent attributes that change the semantics of operations, and cannot be discarded.
In practice, most attribute definitions should become properties definition, which will be added in IRDL in a future PR.
Properties are a dictionary of attributes, just like the attribute dictionary, and a property may have the same name as an attribute.


xdsl/dialects/builtin.py
```diff
@@ -1174,14 +1174,21 @@ class UnregisteredOpWithName(UnregisteredOp):
             @classmethod
             def create(
                 cls,
+                *,
```
superlopuh, line 1177: this seems like a great change to separate into its own PR, would make the changes here a little more focused

PapyChacal (approved):
Thanks!

PR #1489, opened 2023-08-21
core: Move Operation create/build/init parameters to named arguments

math-fehr:
This PR moves the arguments from `__init__`, `create`, and `build` to named arguments only.
This means that `MyOperation.create([arg1, arg2])` is not allowed for instance, and should be changed to `MyOperation.create(operands=[arg1, arg2])`.

This both makes it clearer, and avoid easy mistakes.
Note that this doesn't affect custom `__init__` created by users, so things like `Addi(arg1, arg2)` will stay the same for instance.

### DIFF
--- a/docs/tutorial.ipynb
+++ b/docs/tutorial.ipynb
@@ -350,7 +350,7 @@
     "from xdsl.dialects.arith import Constant\n",
     "\n",
     "const_op = Constant.create(\n",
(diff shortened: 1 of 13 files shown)


PapyChacal (approved):

webmiche (approved):
:pray:

PR #1490, opened 2023-08-21
core: Add properties in IRDL

math-fehr:
This PR adds properties in IRDL, through the use of `prop_def` and `opt_prop_def`.

It does not yet switch dialects to properties.
Note that UnregisteredOp had to be changed to not use IRDL, as properties cannot be extended with arbitrary attributes.


tests/test_operation_builder.py
```diff
@@ -40,12 +42,9 @@
 )
 from xdsl.traits import IsTerminator
 
-#  ____                 _ _
-# |  _ \ ___  ___ _   _| | |_
-# | |_) / _ \/ __| | | | | __|
-# |  _ <  __/\__ \ |_| | | |_
-# |_| \_\___||___/\__,_|_|\__|
-#
+################################################################################
```
superlopuh, line 45: should we use `# region Results` with `# endregion` instead? lets us fold the code nicely

PapyChacal: Can we do that and keep the ASCII art stuff? I like ASCII art too much :grin:

PapyChacal (approved):
Thanks!

Just confused about optional properties. I thought they couldn't be optional? Did you find a counter example? Or did you just want to allow it on our side?

tests/test_operation_builder.py
```diff
@@ -186,13 +185,9 @@ def test_var_mixed_builder():
...
-#   ___                                 _
-#  / _ \ _ __   ___ _ __ __ _ _ __   __| |
-# | | | | '_ \ / _ \ '__/ _` | '_ \ / _` |
-# | |_| | |_) |  __/ | | (_| | | | | (_| |
-#  \___/| .__/ \___|_|  \__,_|_| |_|\__,_|
-#       |_|
-#
+################################################################################
+#                                 Operands                                     #
+################################################################################
```
webmiche, line 190: Mimimi unrelated changes mimimi

xdsl/irdl/irdl.py
```diff
@@ -712,28 +718,47 @@ class AttributeDef:
     constr: AttrConstraint
     """The attribute constraint."""
 
-    attr_name: str | None = None
```
webmiche, line 715: I am confused about this change. Why is it necessary? Also, does this break our generic attribute format? Or does this now just live somewhere else?

xdsl/irdl/irdl.py
```diff
@@ -712,28 +718,47 @@ class AttributeDef:
...
         self,
         attr: Attribute | type[Attribute] | AttrConstraint,
-        attr_name: str | None = None,
     ):
-        super().__init__(attr, attr_name=attr_name)
+        super().__init__(attr)
+
+
+@dataclass(init=False)
+class PropertyDef:
```
webmiche, line 740: Properties and Attributes have very similar taste, IMO. Could we not have a class hierarchy that somehow represents that? It currently feels like they are handled quite differently.

webmiche (approved):

tests/test_operation_definition.py
```diff
@@ -195,6 +200,20 @@ def test_constraint_var_fail_not_satisfy_constraint():
...
+class OperationWithoutProperty(IRDLOperation):
+    name = "test.op_without_prop"
+
+    prop1: Attribute = prop_def(Attribute)
+
+
+# Check that an operation cannot accept properties that are not defined
+def test_unknown_property():
+    op = OperationWithoutProperty.create(properties={"prop1": i32, "prop2": i32})
+    with pytest.raises(VerifyException):
```
AntonLydike, line 213: nit: check the returned error message

tests/test_operation_builder.py
math-fehr: These ascii art should be removed sadly, they are a bit useless, and cannot be grepped :/
I'll think of `# region`, maybe we could do a pass in the code to change that everywhere.

tests/test_traits.py
```diff
@@ -374,7 +374,9 @@ class SymbolOp(IRDLOperation):
     ):
         op.verify()
 
-    terminator = test.TestTermOp([[]], [[]], regions=[[]], successors=[[]])
+    terminator = test.TestTermOp(
+        operands=[[]], result_types=[[]], regions=[[]], successors=[[]]
+    )
```
AntonLydike, line 379: :heart_eyes:

xdsl/irdl/irdl.py
math-fehr: We had kind of duplicated data, so we could grab this from somewhere else.
It doesn't change anything otherwise, it lives in another structure already.

AntonLydike (approved):
LGTM to mee (even though Mathieu is yelling it's broken in the background?)

math-fehr:
> Thanks!
> 
> Just confused about optional properties. I thought they couldn't be optional? Did you find a counter example? Or did you just want to allow it on our side?

So I figured out that they can have optional properties.
For instance PDL has optional properties.

xdsl/irdl/irdl.py
math-fehr: I changed it a bit, tell me if that works for you!

PR #1491, opened 2023-08-22
testing: expose and use a predefined command in lit for xDSL roundtrips.

PapyChacal:
Also, switch to custom syntax anywhere applicable.

Goal is to streamline things on that side of the codebase.
Rationale:
- Exposing a predefined command: just a way to centralize this, and say "This is how we test for roundtripping". Will ease discussion/changes in there. I have in mind to try and add a prefix version for generic syntax checking (Allowing `// CHECK:` for cutom syntax and, if desired, `CHECK-GENERIC:` for generic syntax output).
- Using custom syntax everywhere: Basically, I see generic syntax more like an exchange format at this point. I'm not sure we want to test *every dialect*'s generic syntax, and it add noise when doing core changes (like switching to properties properties).This command does use generic syntax printing/parsing as a step of the roundtrip to ensure nothing is lost in that process

 I anyway had to invest some time on those for the switch to MLIR w/ properties, and thought about doing this PR(s) for a while, so sounded like a nice time to get to it!


webmiche:
> I have in mind to try and add a prefix version for generic syntax checking (Allowing `// CHECK:` for cutom syntax and, if desired, `CHECK-GENERIC:` for generic syntax output).

I think that's an interesting idea. Do you already have an idea how to go about it? Feels like you need quite some magic in the lit.cfg, which feels quite hacky and error-prone. Maybe some replace through xdsl-opt?

Another idea would be to extend filecheck itself more for our needs. I am not sure how active the repo still is, but still think we need some more support of actual FileCheck features to properly check our tests.

PapyChacal:
> > I have in mind to try and add a prefix version for generic syntax checking (Allowing `// CHECK:` for cutom syntax and, if desired, `CHECK-GENERIC:` for generic syntax output).
> 
> I think that's an interesting idea. Do you already have an idea how to go about it? Feels like you need quite some magic in the lit.cfg, which feels quite hacky and error-prone. Maybe some replace through xdsl-opt?
> 
> Another idea would be to extend filecheck itself more for our needs. I am not sure how active the repo still is, but still think we need some more support of actual FileCheck features to properly check our tests.

Turns out `tests/filecheck/dialects/riscv/riscv_ops.mlir` was doing exactly this already, so I did it here directly, you can see how in `tests/filecheck/lit.cfg` :slightly_smiling_face: 
Slightly error prone indeed, but centralized at least. I'll add comments at this spot to safeguard a bit!

math-fehr (approved):
Looks nice, thanks!
I would just rename `GENERIC` with `CHECK-GENERIC` if that is allowed by filecheck/lit

webmiche (approved):
LGTM to me too. I like the names, I think they are nicely descriptive :+1:

superlopuh (approved):
Excellent. Is this still [WIP] ?

PapyChacal:
> Excellent. Is this still [WIP] ?

No, edited!

PR #1492, opened 2023-08-22
transforms: (riscv_func) handle non-float arguments for func lowering

superlopuh:
- renames a test file that git doesn't see as a rename for some reason
 - add some debug info to an exception message
 - cannot move floats so explicitly mark those as unsupported for now
 - don't support functions that have more than 8 inputs or outputs


xdsl/dialects/riscv_func.py
```diff
@@ -72,13 +72,13 @@ class CallOp(IRDLOperation):
     name = "riscv_func.call"
     args: VarOperand = var_operand_def(riscv.IntRegisterType)
```
adutilleul, line 73: I don't see why we only change to `riscv.RISCVRegisterType` for results and not args here, am I missing something?

adutilleul (approved):

xdsl/dialects/riscv_func.py
superlopuh: No that's my bad

PR #1493, opened 2023-08-22
misc: remove remove_annotated.sed

AntonLydike:
Remove the migration script from the old irdl syntax to the new one.


AntonLydike:
Will re-open in a month, it's still used as it seems :shrug:

PR #1495, opened 2023-08-23
backend: (riscv) make RdRsOperation generic

superlopuh:


xdsl/dialects/riscv.py
```diff
@@ -867,26 +868,22 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
...
-    rd: OpResult = result_def(IntRegisterType)
-    rs: Operand = operand_def(IntRegisterType)
+    rd: OpResult = result_def(RDInvT)
+    rs: Operand = operand_def(RSInvT)
 
     def __init__(
         self,
         rs: Operation | SSAValue,
         *,
-        rd: IntRegisterType | str | None = None,
```
superlopuh, line 883: Passing `str` or `None` is a casualty of this change, but in my opinion it's not a great loss to lose this helper functionality. It makes the code a bit more explicit at the callsite when constructing this operation individually, and doesn't have a great impact on the lowering passes due to them calling helpers anyway.

JosseVanDelm (approved):
looks good!

PR #1498, opened 2023-08-23
dialects: (riscv) add fmvop

superlopuh:
This doesn't really use it anywhere yet, just provides the ability to express copying a value from one register to another. Hopefully we'll actually manage to optimise out this operation for all the cases we care about, but it's possible that we won't so we'll need to implement a lowering to the rountrip through memory before printing.

### DIFF
--- a/tests/filecheck/dialects/riscv/riscv_ops.mlir
+++ b/tests/filecheck/dialects/riscv/riscv_ops.mlir
@@ -215,6 +215,9 @@
     %f2 = riscv.get_float_register : () -> !riscv.freg<>
     // CHECK-NEXT: %{{.*}} = riscv.get_float_register : () -> !riscv.freg<>
 
+    %fmv = riscv.fmv.s %f0 : (!riscv.freg<>) -> !riscv.freg<>
+    // CHECK: %{{.*}} = riscv.fmv.s %{{.*}} : (!riscv.freg<>) -> !riscv.freg<>
+
     %fmadd_s = riscv.fmadd.s %f0, %f1, %f2 : (!riscv.freg<>, !riscv.freg<>, !riscv.freg<>) -> !riscv.freg<>
     // CHECK-NEXT: %{{.*}} = riscv.fmadd.s %{{.*}}, %{{.*}}, %{{.*}} : (!riscv.freg<>, !riscv.freg<>, !riscv.freg<>) -> !riscv.freg<>
     %fmsub_s = riscv.fmsub.s %f0, %f1, %f2 : (!riscv.freg<>, !riscv.freg<>, !riscv.freg<>) -> !riscv.freg<>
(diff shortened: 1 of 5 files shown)


adutilleul (approved):

JosseVanDelm (commented):
I've added some minor comments. Looks good overall!
Just wondering if this is useful if the performance implications are so high.
Are there multiple lowering paths (some that don't store on the stack?) depending on context maybe?

superlopuh:
Ah fantastic! I'll update to use `fmv.s`

xdsl/backend/riscv/lowering/optimise_riscv.py
```diff
@@ -15,3 +15,14 @@ def match_and_rewrite(self, op: riscv.MVOp, rewriter: PatternRewriter) -> None:
...
         ):
             rewriter.replace_matched_op([], [op.rs])
+
+
+class RemoveRedundantFMv(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: riscv.FMVOp, rewriter: PatternRewriter) -> None:
+        if (
+            op.rd.type == op.rs.type
+            and isinstance(op.rd.type, riscv.RISCVRegisterType)
```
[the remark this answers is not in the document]
superlopuh: Seems that they should be equivalent, conditioned on the fact that riscv.FMVOp is already constrained to be on float types only. I'd rather stick with the most generic interface that provides the `is_allocated` method

nazavode (approved):

tobiasgrosser:
Very cool. Thank you for getting this in.

tests/filecheck/dialects/riscv/riscv_ops.mlir
```diff
@@ -215,6 +215,9 @@
     %f2 = riscv.get_float_register : () -> !riscv.freg<>
     // CHECK-NEXT: %{{.*}} = riscv.get_float_register : () -> !riscv.freg<>
 
+    %fmv = riscv.fmv.s %f0 : (!riscv.freg<>) -> !riscv.freg<>
+    // CHECK: %{{.*}} = riscv.fmv.s %{{.*}} : (!riscv.freg<>) -> !riscv.freg<>
```
kingiler, line 219:
```suggestion
    // CHECK-NEXT: %{{.*}} = riscv.fmv.s %{{.*}} : (!riscv.freg<>) -> !riscv.freg<>
```

xdsl/dialects/riscv.py
```diff
@@ -1531,23 +1531,13 @@ class FMVOp(RdRsOperation[FloatRegisterType, FloatRegisterType]):
     """
     A pseudo instruction to copy contents of one float register to another.
 
-    Equivalent to `fsw rs, 0(sp); flw rd, 0(sp)`.
-
-    This is a much more expensive operation than mv, and does not correspond to an
-    instruction in the RISC-V F extension. Possibly due to this, there is no common
-    assembly pseudo-op for it, so it must be rewritten before assembly printing.
+    Equivalent to `fsw rs, 0(sp); flw rd, 0(sp)` or `fsgnj.s rd, rs, rs`
```
kingiler, line 1534: I'm not sure about the `fsw rs, 0(sp); flw rd, 0(sp)` part especially with the word *equivalent*. `fsgnj.s rd, rs, rs` is enough?

xdsl/dialects/riscv.py
```diff
@@ -1531,23 +1531,13 @@ class FMVOp(RdRsOperation[FloatRegisterType, FloatRegisterType]):
...
     """
 
-    name = "riscv.fmv"
+    name = "riscv.fmv.s"
```
kingiler, line 1537: I know we are not going to handle binary64 for the time being. However, will there be spaces for extension to binary64? Like the class name is `FMVSOp` and `FMVDOp` or something else?

superlopuh: It should be pretty easy to add those extensions IMO, or at least I can't see anything stopping us from doing so. Planning-wise, it would probably need to be as a step towards some specific goal.

xdsl/dialects/riscv.py
superlopuh: Yeah I'll rephrase this.

xdsl/backend/riscv/lowering/optimise_riscv.py
```diff
@@ -15,3 +15,14 @@ def match_and_rewrite(self, op: riscv.MVOp, rewriter: PatternRewriter) -> None:
             and op.rd.type.is_allocated
         ):
             rewriter.replace_matched_op([], [op.rs])
+
+
+class RemoveRedundantFMv(RewritePattern):
```
kingiler, line 20: `RemoveRedundantFMv` looks a little bit weird, expect `RemoveRedundantFmv` or `RemoveRedundantFMV` as class is `FMVOp`.

superlopuh: I see your point, but I'm not sure it's strictly better. The tradeoff is between "the set of all optimisations for a specific operation are easy to find" and "the set of all optimisations for the riscv dialect are easy to find". I think having them all in `optimise_riscv.py` will help us going forward. We can always refactor if we decide that it's getting too noisy.

PR #1500, opened 2023-08-25
backend: (riscv) allocate registers in function bodies only

superlopuh:


nazavode (approved):

xdsl/backend/riscv/register_allocation.py
```diff
@@ -1,25 +1,30 @@
...
 
-from xdsl.dialects import riscv_scf
+from xdsl.dialects import riscv_func, riscv_scf
 from xdsl.dialects.builtin import ModuleOp
 from xdsl.dialects.riscv import FloatRegisterType, IntRegisterType, RISCVOp
 from xdsl.ir import SSAValue
 
 
-class RegisterAllocator(ABC):
+class RegisterAllocator(abc.ABC):
```
AntonLydike, line 9: Wait, isn't this something that we didn't want to do? I remember this being a nono with `dataclasses.dataclass`?

AntonLydike (approved):

xdsl/backend/riscv/register_allocation.py
superlopuh: I have no idea. Maybe there's a ruff rule for this? I would just not think about it unless there's a reason to do it one way or another

kingiler: Excuse me, what's the issue here?

superlopuh: I think `from abc import ABC` vs `import abc`

kingiler: I see. I don't think there is such rules and it is perfectly legit usage?

PR #1503, opened 2023-08-27
backend: (riscv) handle float arguments in func lowering

superlopuh:


tobiasgrosser (approved):
Nice.

tests/filecheck/backend/riscv/convert_func_to_riscv_func.mlir
```diff
@@ -11,6 +11,11 @@ builtin.module {
         %res0, %res1 = "test.op"(%arg0, %arg1) : (i32, i32) -> (i32, i32)
         func.return %res0, %res1 : i32, i32
     }
+
+    func.func @foo_float(%farg0 : f32, %farg1 : f32) -> (f32, f32) {
+        %fres0, %fres1 = "test.op"(%farg0, %farg1) : (f32, f32) -> (f32, f32)
+        func.return %fres0, %fres1 : f32, f32
+    }
 }
```
kingiler, line 19: Could add a function with mixed int and float.
```
    func.func @foo_int_float(%arg0 : i32, %farg0 : f32) -> (i32, f32) {
        %res0, %fres0 = "test.op"(%arg0, %farg0) : (i32, f32) -> (i32, f32)
        func.return %res0, %fres0 : i32, f32
    }
```

kingiler (approved):
LGTM

tests/filecheck/backend/riscv/convert_func_to_riscv_func.mlir
superlopuh: I'm not sure that it would be worth it, let's add it if we get a bug in the future where this would have helped.

PR #1375, opened 2023-07-31
core: (utils) Implement Lazy class

kingiler:

PR #1505, opened 2023-08-28
core: (trait) Implement LazyOpTrait

kingiler:
From #1359, #1375. Implemented `LazyOpTrait` which inherit rom `OpTrait` but the `parameters` are *lazy* (resolve later).

One example usage is the `HasParent` trait is now inheriting from the `LazyOpTrait`.

Main implementation reside in `xdsl/traits.py`.

### DIFF
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -38,7 +38,8 @@ class OpTrait:
 @dataclass(frozen=True)
 class LazyOpTrait(OpTrait):
     """
-    TODO: Add doc
+    A lazy version of OpTrait which postpone the evaluation of parameters.
+    It can used to workaround the circular dependencies in code.
     """
 
     _parameters: Callable[[], Any] | Any = field(default=None)
@@ -48,7 +49,7 @@ class LazyOpTrait(OpTrait):
     def parameters(self) -> Any:
         if self._is_unresolved:
             object.__setattr__(self, "_parameters", self._parameters())
-            object.__setattr__(self, "_unresolved", False)
+            object.__setattr__(self, "_is_unresolved", False)
         return self._parameters
 
     def verify(self, op: Operation) -> None:
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -39,7 +39,7 @@ class OpTrait:
 class LazyOpTrait(OpTrait):
     """
     A lazy version of OpTrait which postpone the evaluation of parameters.
-    It can used to workaround the circular dependencies in code.
+    It can used to work around the circular dependencies in code.
     """
 
     _parameters: Callable[[], Any] | Any = field(default=None)
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+from collections.abc import Callable
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, TypeVar
 
@@ -30,6 +31,25 @@ class OpTrait:
         pass
 
 
+@dataclass(frozen=True)
+class LazyOpTrait(OpTrait):
+    """
+    TODO: Add doc
+    """
+
+    parameters: Callable[[], Any] | Any = field(default=None)
+    _unresolved: bool = field(default=True)
+
+    def resolve(self):
+        if self._unresolved:
+            object.__setattr__(self, "parameters", self.parameters())
+            object.__setattr__(self, "_unresolved", False)
+
+    def verify(self, op: Operation) -> None:
+        """Check that the operation satisfies the trait requirements."""
+        pass
+
+
 OpTraitInvT = TypeVar("OpTraitInvT", bound=OpTrait)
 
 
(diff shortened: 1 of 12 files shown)


xdsl/traits.py
```diff
@@ -30,6 +31,25 @@ def verify(self, op: Operation) -> None:
         pass
 
 
+@dataclass(frozen=True)
+class LazyOpTrait(OpTrait):
+    """
+    TODO: Add doc
```
superlopuh, line 37: Please add doc

xdsl/traits.py
```diff
@@ -38,26 +58,31 @@ class Pure(OpTrait):
...
-        if len(self.parameters) == 1:
-            raise VerifyException(
-                f"'{op.name}' expects parent op '{self.parameters[0].name}'"
-            )
-        names = ", ".join(f"'{p.name}'" for p in self.parameters)
-        raise VerifyException(f"'{op.name}' expects parent op to be one of {names}")
+        self.resolve()
+        match self.parameters:
+            case []:
+                raise ValueError("parameters should not be empty")
```
superlopuh, line 74: This should still be in the init, I think

kingiler: I think it defeat the purpose of laziness?

xdsl/traits.py
kingiler: Added in latest commit.

xdsl/traits.py
```diff
@@ -30,6 +31,25 @@ def verify(self, op: Operation) -> None:
         pass
 
 
+@dataclass(frozen=True)
+class LazyOpTrait(OpTrait):
+    """
+    TODO: Add doc
+    """
+
+    parameters: Callable[[], Any] | Any = field(default=None)
```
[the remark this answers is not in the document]
kingiler: Added in latest commit.

xdsl/traits.py
```diff
@@ -30,6 +31,25 @@ def verify(self, op: Operation) -> None:
...
+    _unresolved: bool = field(default=True)
```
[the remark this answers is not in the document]
kingiler: Addressed in latest commit.

xdsl/traits.py
```diff
@@ -23,7 +24,33 @@ class OpTrait:
...
+    def verify(self, op: Operation) -> None:
+        """Check that the operation satisfies the trait requirements."""
+        pass
+
+
+@dataclass(frozen=True)
+class LazyOpTrait(OpTrait):
+    """
+    A lazy version of OpTrait which postpone the evaluation of parameters.
+    It can used to workaround the circular dependencies in code.
```
superlopuh, line 42:
```suggestion
    It can used to work around the circular dependencies in code.
```

superlopuh (approved):
Great, thank you! Would be nice to have a basic test of the mechanism

xdsl/traits.py
kingiler: Addressed in latest commit.

kingiler:
@superlopuh I have added a test. Is it what you want? Actually I think `HasParent` tests `LazyOpTrait` well already.

superlopuh:
Sorry for long reply, this got lost in my notifications somehow. @math-fehr are you happy with this ?

xdsl/traits.py
```diff
@@ -38,26 +65,30 @@ class Pure(OpTrait):
 
 
 @dataclass(frozen=True)
-class HasParent(OpTrait):
+class HasParent(LazyOpTrait):
     """Constraint the operation to have a specific parent operation."""
 
-    parameters: tuple[type[Operation], ...]
+    _parameters: tuple[type[Operation], ...]
```
AntonLydike, line 71: I feel that this is a very ugly hack that makes pyright happy, but is not actually factually correct, as the `_parameters` attribute can actually be observed as being of type `Callable` at certain times in the lifetime of the object.

This just feels like lying to pyright to me, that may actually make people more confused during debugging.

xdsl/traits.py
```diff
@@ -38,26 +65,30 @@ class Pure(OpTrait):
...
+                if not isinstance(parent, parameter):
+                    raise VerifyException(
+                        f"'{op.name}' expects parent op '{parameter.name}'"
+                    )
+            case parameters:
+                if not isinstance(parent, parameters):
+                    names = ", ".join(f"'{p.name}'" for p in self.parameters)
+                    raise VerifyException(
+                        f"'{op.name}' expects parent op to be one of {names}"
+                    )
```
AntonLydike, line 91: I prefer the old implementation without the match here, as it is smaller, simpler, and defers the differentiation between single and multiple parameters to the less-likely case instead of pattern matching for every `verify` call.

Also, empty parameters should ideally not be detected ad verify time, but at init-time (or as close to init as possible).

AntonLydike (commented):
Thanks for working towards this!

However, I am not sure if I'm a huge fan of this implementation, as it's somewhat rough around the edges and feels a bit broken in a few areas.

xdsl/traits.py
superlopuh: I don't know of a place closer to init that would be better than here. Do you have one in mind?

AntonLydike: I could see a `__lazy_init__` method that is called after the lazy traits have been initialized. That way it's called exactly once, and (at least conceptually) it's closer to the `__init__`...

xdsl/traits.py
superlopuh: Yeah now that I think about it I think we should change it before merging. The first thing that comes to mind is that `OpTrait` doesn't actually need a `parameters` field. I think the code should work just fine if that code is deleted, @kingiler could you please try that out?

xdsl/traits.py
```diff
@@ -23,7 +24,33 @@ class OpTrait:
...
+
+
+@dataclass(frozen=True)
+class LazyOpTrait(OpTrait):
+    """
+    A lazy version of OpTrait which postpone the evaluation of parameters.
+    It can used to work around the circular dependencies in code.
+    """
+
+    _parameters: Callable[[], Any] | Any = field(default=None)
```
math-fehr, line 45: There are no checks here that `_parameters` are callable when creating the object.

math-fehr (commented):
I still feel there are quite some complexity involved just to have the `LazyOpTrait`.
Would there be a reason not to add the `lambda` optionally on the `traits` field directly in an operation? I would feel that this would be simpler overall, and wouldn't require a case-by-case basis for each trait, simplifying user code.

xdsl/traits.py
superlopuh: Who would call it and when?

AntonLydike: It would be called from the lazy trait right after it evaluated the lambda with the arguments being the result of the lambda.

That what makes sense in my mind at least...

xdsl/traits.py
kingiler: I don't think we can just drop `parameters` field?

xdsl/traits.py
kingiler: `TypeError` will be thrown if it is not callable. Or you want to have a more specific error message/check? Also current implementation is lazily checking the type of the parameter given.

math-fehr:
Closing this in favor of #1529

PR #1506, opened 2023-08-28
backend: (riscv) add RegisterQueue to manage available registers

superlopuh:
This factors out a bit of the logic shared between allocation strategies. The J allocator starts with an empty queue and spills to J immediately. The BlockNaive allocator never pushes back to the queue. LivenessBlockNaive pops and pushes.


tests/backend/riscv/test_regalloc.py
superlopuh, line 1: In this and other filecheck files, the order of the registers is changed, but not the fact that they're "real" RISC-V registers from roughly the same class. We don't really have an ability to reason about the difference in ABI specifications for A, S and T registers anyway right now.

superlopuh:
I'll use this register allocation helper in a future PR, so it feals easiest to reopen this one just with the contained changes.

AntonLydike (approved):
LGTM

nazavode (approved):

PR #1507, opened 2023-08-28
transforms: (riscv) partial lowering of memref to riscv

superlopuh:
Only store/load, no alloc/dealloc.

### DIFF
--- a/tests/filecheck/backend/riscv/memref_to_riscv.mlir
+++ b/tests/filecheck/backend/riscv/memref_to_riscv.mlir
@@ -1,15 +1,15 @@
 // RUN: xdsl-opt -p convert-memref-to-riscv  --split-input-file --verify-diagnostics %s | filecheck %s
 
 builtin.module {
-    %v_f32, %v_i32, %r, %c, %m_f32, %m_i32 = "test.op"() : () -> (f32, i32, index, index, memref<3x2xf32>, memref<3x2xi32>)
+    %v_f32, %v_i32, %r, %c, %m_f32, %m_i32 = "test.op"() : () -> (f32, i32, index, index, memref<3x2xf32>, memref<3xi32>)
     "memref.store"(%v_f32, %m_f32, %r, %c) {"nontemporal" = false} : (f32, memref<3x2xf32>, index, index) -> ()
     %x_f32 = "memref.load"(%m_f32, %r, %c) {"nontemporal" = false} : (memref<3x2xf32>, index, index) -> (f32)
-    "memref.store"(%v_i32, %m_i32, %r, %c) {"nontemporal" = false} : (i32, memref<3x2xi32>, index, index) -> ()
-    %x_i32 = "memref.load"(%m_i32, %r, %c) {"nontemporal" = false} : (memref<3x2xi32>, index, index) -> (i32)
+    "memref.store"(%v_i32, %m_i32, %c) {"nontemporal" = false} : (i32, memref<3xi32>, index) -> ()
+    %x_i32 = "memref.load"(%m_i32, %c) {"nontemporal" = false} : (memref<3xi32>, index) -> (i32)
 }
 
 // CHECK:      builtin.module {
-// CHECK-NEXT:   %v_f32, %v_i32, %r, %c, %m_f32, %m_i32 = "test.op"() : () -> (f32, i32, index, index, memref<3x2xf32>, memref<3x2xi32>)
+// CHECK-NEXT:   %v_f32, %v_i32, %r, %c, %m_f32, %m_i32 = "test.op"() : () -> (f32, i32, index, index, memref<3x2xf32>, memref<3xi32>)
 // CHECK-NEXT:   %0 = builtin.unrealized_conversion_cast %v_f32 : f32 to !riscv.freg<>
 // CHECK-NEXT:   %1 = builtin.unrealized_conversion_cast %m_f32 : memref<3x2xf32> to !riscv.reg<>
 // CHECK-NEXT:   %2 = builtin.unrealized_conversion_cast %r : index to !riscv.reg<>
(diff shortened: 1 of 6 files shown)


tests/filecheck/backend/riscv/memref_to_riscv.mlir
```diff
@@ -0,0 +1,48 @@
+// RUN: xdsl-opt -p convert-memref-to-riscv %s --split-input-file --verify-diagnostics | filecheck %s
+
+builtin.module {
+    %v, %r, %c, %m = "test.op"() : () -> (f32, index, index, memref<3x2xf32>)
+    "memref.store"(%v, %m, %r, %c) {"nontemporal" = false} : (f32, memref<3x2xf32>, index, index) -> ()
+    %x = "memref.load"(%m, %r, %c) {"nontemporal" = false} : (memref<3x2xf32>, index, index) -> (f32)
+}
```
JosseVanDelm, line 7: Please add a test for an integer memref

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -0,0 +1,135 @@
...
+            rewriter.insert_op_before_matched_op(
+                [
+                    ptr := riscv.AddOp(mem, idx1),
+                ]
+            )
+        case [idx1, idx2]:
+            rewriter.insert_op_before_matched_op(
+                [
+                    cols := riscv.LiOp(shape[1]),
+                    row_offset := riscv.MulOp(cols, idx1),
```
JosseVanDelm, line 59: note: mul is not an op from the base ISA. We might want to make this opt-in at some point.
In snitch cluster, one M extension ALU is shared by all cores in a compute cluster.

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -0,0 +1,135 @@
...
+                    offset := riscv.AddOp(row_offset, idx2),
+                    offset_bytes := riscv.SlliOp(
+                        offset, 2, comment="mutiply by elm size"
+                    ),
+                    ptr := riscv.AddOp(mem, offset_bytes),
+                ]
+            )
+        case _:
+            raise NotImplementedError(f"Unsupported memref shape {shape}")
```
JosseVanDelm, line 68:
```suggestion
            raise NotImplementedError(f"Unsupported memref shape {shape}, only 1D and 2D memrefs are supported now")
```

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -0,0 +1,135 @@
...
+    shape: Sequence[int],
+    rewriter: PatternRewriter,
+) -> SSAValue:
+    """
+    Returns ssa value representing pointer into the memref at given indices.
+    """
+    assert len(shape) == len(indices)
+
+    match indices:
+        case [idx1]:
```
JosseVanDelm, line 49: I'm assuming you are not allowing N-dimensional cases, since this would incur loops?
Can you add this as a comment maybe?

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -0,0 +1,135 @@
...
+                ]
+            )
+        case [idx1, idx2]:
+            rewriter.insert_op_before_matched_op(
+                [
+                    cols := riscv.LiOp(shape[1]),
+                    row_offset := riscv.MulOp(cols, idx1),
+                    offset := riscv.AddOp(row_offset, idx2),
+                    offset_bytes := riscv.SlliOp(
+                        offset, 2, comment="mutiply by elm size"
```
JosseVanDelm, line 62: I feel like I should know this, but we are always multiplying by 4, right?
It's not dependent on the type, right?

JosseVanDelm (changes requested):
Really cool! Thanks!
Can you add some comments and add a test for integer memrefs?
Thanks!

superlopuh:
I'd rather do ints in a separate PR, if that's ok with you? Just trying to have small and focused PRs, adding ints should be trivial when we need them.

superlopuh:
Hmm actually I guess they're already supported I'll add them to the test

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: hmm 4 because 4 bytes in a word on 32-bit systems but you're right that it's a tricky thing, I'll add a comment

- [ ] add a comment

JosseVanDelm:
@superlopuh yes I think you already supported them hahaha

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: I'm not scared of loops I just don't think it's worth our time to add them right now

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: Interesting, I'm not sure exactly what you mean, but I'm guessing I'll find out soon enough once we try to run this on Snitch and it fails :)

JosseVanDelm: `RV32I`, the most basic ISA to run a RISCV program does not have a multiply instruction. Rather, the compiler unrolls it into a loop with multiple add instructions or converts it into a shift instruction of it's a multiple of 2.

If you have `RV32IM` you have a dedicated instruction and a dedicated multiplier that performs the addition of the partialsums of the multiplication in hardware.

Since snitch is mainly a floating point machine, the integer multiplication unit (the on that provides M) is shared across all compute cores. This should work, but you can not parallelize it across multiple cores, or you will stall certain cores, until the shared multiplier is available. If you just consider single-core it should not be a problem at all. Just note that this will not work on very simple RISCV-cores (ones that don't support M extension).

JosseVanDelm: In principle people could disable the snitch's multiplier.
just a nit.

The ISA of the compute cores is set here: https://github.com/pulp-platform/snitch_cluster/blob/772b86ae84ec0d5a6f1e755cb524ba0aae2cefc3/target/snitch_cluster/cfg/default.hjson#L86

nazavode (approved):

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -0,0 +1,164 @@
...
+    The pointer is byte-indexed, and the indices are strided by element size, so the index
+    into the flat memory buffer needs to be multiplied by the size of the element.
+    """
+
+    assert len(shape) == len(indices)
+
+    # Only handle a small subset of elements
+    # Might be useful as a helper for other passes in the future
+    match element_type:
+        case IntegerType():
```
JosseVanDelm, line 57: I think it's better to constrain this to int32, since using int8 with this method might fail, I think.
If you want to access a int8 integer, you need to access the entire 4-bytes and then shift until you get the right number, I believe, because riscv accesses are four bytes aligned, I think?
Or you should start using other instructions, like SB and LB, I think?
What do you think @superlopuh ?

JosseVanDelm (changes requested):
Hi, sorry to give you extra work for this PR, but it looks to me like we should constrain to i32 types for now to avoid alignment issues (though I'm not very sure about this, and would love your take on this)
Besides that I'm happy to see this merged!

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: Sure sounds good I'll add a special case for i32

JosseVanDelm (approved):
:pray:

adutilleul (approved):

PR #1510, opened 2023-08-30
documentation: (toy) add optimisation steps to Toy

superlopuh:
This is technically in Toy but we're currently using this pipeline for the riscv paper experiments, so would benefit from this change for critical work.

### DIFF
--- a/docs/Toy/toy/compiler.py
+++ b/docs/Toy/toy/compiler.py
@@ -129,6 +129,8 @@ def transform(
     if target == "riscv":
         return
 
+    # Perform optimizations that don't depend on register allocation
+    # e.g. constant folding
     CanonicalizePass().apply(ctx, module_op)
 
     module_op.verify()
@@ -143,6 +145,8 @@ def transform(
     if target == "riscv-regalloc":
         return
 
+    # Perform optimizations that depend on register allocation
+    # e.g. redundant moves
     CanonicalizePass().apply(ctx, module_op)
 
     module_op.verify()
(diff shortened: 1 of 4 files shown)


docs/Toy/toy/compiler.py
```diff
@@ -124,20 +124,32 @@ def transform(
...
     ReconcileUnrealizedCastsPass().apply(ctx, module_op)
 
-    DeadCodeElimination().apply(ctx, module_op)
-
     module_op.verify()
 
     if target == "riscv":
         return
 
+    CanonicalizePass().apply(ctx, module_op)
```
superlopuh, line 132: optimises things that don't need register allocation, like constant folding etc

docs/Toy/toy/compiler.py
```diff
@@ -124,20 +124,32 @@ def transform(
...
+        return
+
     RISCVRegisterAllocation().apply(ctx, module_op)
 
     module_op.verify()
 
     if target == "riscv-regalloc":
         return
 
+    CanonicalizePass().apply(ctx, module_op)
```
superlopuh, line 146: optimises things that need register allocation, like redundant moves etc

nazavode:
```suggestion
    # Perform optimizations that depend on register allocation (e.g.: redundant moves):
    CanonicalizePass().apply(ctx, module_op)
```

docs/Toy/toy/compiler.py
nazavode:
```suggestion
    # Perform optimizations that don't depend on register allocation (e.g.: constant folding):
    CanonicalizePass().apply(ctx, module_op)
```

nazavode (approved):
Nice, left a couple of suggested changes that bring your own comments in source for posterity.

PR #1511, opened 2023-08-30
backend: (riscv) Clean up abi and arch specifications

AntonLydike:
I managed to find a proper document specifying the ABI, and did some exploring in the compiler explorer.

My initial version of the ABI was almost correct, I had just missed a couple of smaller issues (such as stack alignment and float calling conventions).

This should now conform to the original spec.

### DIFF
--- a/xdsl/backend/riscv/targets.py
+++ b/xdsl/backend/riscv/targets.py
@@ -1,10 +1,20 @@
 """
 This file provides tools to describe toolchain targets.
 
-Based on https://github.com/riscv-non-isa/riscv-elf-psabi-doc
+Based on:
 
-And the RISC-V specification
+ [0]: https://github.com/riscv-non-isa/riscv-elf-psabi-doc
+ [1]: `RISC-V ABIs Specification, Document Version 1.0', Editors Kito Cheng and Jessica
+      Clarke, RISC-V International, November 2022.
+      https://github.com/riscv-non-isa/riscv-elf-psabi-doc/releases/tag/v1.0
+ [2]: “The RISC-V Instruction Set Manual, Volume I: User-Level ISA, Document Version
+      20191213”, Editors Andrew Waterman and Krste Asanovi´c, RISC-V Foundation,
+      December 2019.
+      https://github.com/riscv/riscv-isa-manual/releases/download/Ratified-IMAFDQC/riscv-spec-20191213.pdf
+
 """
+
+from __future__ import annotations
 
 import re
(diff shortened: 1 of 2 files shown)


xdsl/backend/riscv/targets.py
```diff
@@ -93,6 +99,17 @@ class ABISpec:
     "elf" is the default for most
     """
 
+    # stack alignment:
+    stack_alignment: int = 128
+    """
+    The stack grows downwards (towards lower addresses) and the stack pointer shall be
+    aligned to a 128-bit boundary upon procedure entry. The first argument passed on the
+    stack is located at offset zero of the stack pointer on function entry; following
+    arguments are stored at correspondingly higher addresses.
```
superlopuh, line 108: lower addresses?

superlopuh: I'm confused, I see that it's verbatim the text form the link but I'm not sure how it can grow downwards and for following arguments being at higher addresses...

superlopuh (approved):

xdsl/backend/riscv/targets.py
AntonLydike: The caller "pushes" arguments to the stack (if they don't fit into registers), once control is given over to the calling function, the stack pointer is pointing to the *end* of the arguments, so that the empty stack lies "below" the stack pointer, and the arguments directly above.

to illustrate:

```
+--------------------------+
|  Calling functions stack |
|                          |
+--------------------------+
|    Padding if needed     |
+--------------------------+
|     stack argument n     |
|     stack argument n-1   |
|            ...           |
|     stack argument 0     |
+--------------------------+ <--- Stack pointer, aligned to 128 bits
|           empty          |
+--------------------------+
```

The fact that stack arguments are passed like this is likely irrelevant for us for now,
but we may need to consider that the `sp` should be 128 bit aligned when calling other
functions, as they likely rely on this.

superlopuh: ah this makes sense, and explains why gcc and llvm allocate more stack space than they need in my explorations, thank you

nazavode: I'm fond of ASCII art in comments, can we add this in source pls?

nazavode (approved):
Beautiful, I love all the references to the spec. Thanks!

xdsl/backend/riscv/targets.py
AntonLydike: Will do!

PR #1514, opened 2023-08-31
backend: (riscv) add riscv-reduce-register-pressure pass

superlopuh:
First rewrite among many that will help reduce the number of registers live at any time. Future work will include things like moving constant definitions closer to their use etc.

### DIFF
--- a/tests/filecheck/backend/riscv_reduce_register_pressure.mlir
+++ b/tests/filecheck/backend/riscv_reduce_register_pressure.mlir
@@ -1,12 +1,15 @@
 // RUN: xdsl-opt -p riscv-reduce-register-pressure %s | filecheck %s
 
 builtin.module {
+  // Replace loading immediate value 0
   %0 = riscv.li 0 : () -> !riscv.reg<>
   %1 = riscv.li 0 : () -> !riscv.reg<a0>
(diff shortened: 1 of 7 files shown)


nazavode (approved):

tests/filecheck/backend/riscv_reduce_register_pressure.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: xdsl-opt -p riscv-reduce-register-pressure %s | filecheck %s
+
+builtin.module {
+  %0 = riscv.li 0 : () -> !riscv.reg<>
+  %1 = riscv.li 0 : () -> !riscv.reg<a0>
+}
```
tobiasgrosser, line 6: I see. I feel this could use some documentation. Just 1-2 sentences stating that a li0 is replaced to a load from a zero register.

AntonLydike:
Nice one! I was a bit concerned how this would affect constant folding, but it should not impact it at all as `add a1, a0, zero` takes the same number of instructions as `addi a1, a0, 0`. And we only eliminate the special case of zero anyways.

AntonLydike (approved):
Very nice!

PR #1516, opened 2023-08-31
backend: (riscv) limit_registers None by default

superlopuh:
We can simulate the J reg allocator with the block naive allocators limited to 0. This PR allows passing 0 as the value, and tests that for code with no loops the result is the same as the J register allocator.


nazavode (approved):

tobiasgrosser (approved):
LGTM

adutilleul (approved):

PR #1517, opened 2023-08-31
backend: (riscv) j reg allocator is block naive with 0 real registers

superlopuh:
A bit of entropy minimisation, with this PR we can allocate for loops also.


nazavode (approved):
Nice convergence of things.

xdsl/backend/riscv/register_allocation.py
```diff
@@ -155,6 +155,97 @@ def __init__(self, limit_registers: int | None = None) -> None:
...
+                self.idx += 1
+            else:
+                reg.type = reg_type(available_regs.pop())
+
+            return True
+
+        return False
+
+    def allocate_for_loop(self, loop: riscv_scf.ForOp) -> None:
+        yield_op = loop.body.block.last_op
```
superlopuh, line 175: this is just copy/pasted from below

superlopuh:
Wait till you see where the register queue comes into all this! ;)

adutilleul (approved):

tobiasgrosser:
@superlopuh, I find this PR really confusing. This is not just deleting the j allocator, but it's pretty much copy-pasting the j-allocator with loop support into the block naive allocator. In which way is the block naive allocator still the block naive allocator?

tobiasgrosser:
Let's maybe discuss this on zulip.

PR #1518, opened 2023-09-01
backend: (riscv) use register_queue for register management

superlopuh:
Decouple the traversal logic from available register logic. This change just uses the helper class instead of having the logic inline, future changes will do a bit more separation.

### DIFF
--- a/xdsl/backend/riscv/register_allocation.py
+++ b/xdsl/backend/riscv/register_allocation.py
@@ -16,7 +16,7 @@ class RegisterAllocator(abc.ABC):
     Base class for register allocation strategies.
     """
 
-    def __init__(self, limit_registers: int | None = None) -> None:
+    def __init__(self) -> None:
         pass
 
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


xdsl/backend/riscv/register_allocation.py
```diff
@@ -15,7 +16,7 @@ class RegisterAllocator(abc.ABC):
     Base class for register allocation strategies.
     """
 
-    def __init__(self, limit_registers: int | None = None) -> None:
+    def __init__(self) -> None:
         pass
```
AntonLydike, line 20: Empty init can be dropped?

AntonLydike (approved):
Minor critic, otherwise this PR :rock:s as well!

xdsl/backend/riscv/register_allocation.py
superlopuh: Probably, I'll take a look

nazavode (approved):

xdsl/backend/riscv/register_allocation.py
superlopuh: I'm not sure exactly what the rules are for abstract base classes and inits. What I'd like to communicate here is that all subclasses should implement an init with no parameters, but I don't think this is actually enforced. I'll delete it for now until I have a better idea.

PR #1521, opened 2023-09-01
backend: (riscv) factor out func body walking and op processing in register allocators

superlopuh:
This more finely controls the walking, and the exact differences in implementation between the two register allocators.

- [ ] allocate unused values to 0 for int registers


nazavode (approved):

xdsl/backend/riscv/register_allocation.py
```diff
@@ -49,6 +49,32 @@ def allocate(self, reg: SSAValue) -> bool:
...
         return False
 
+    def process_operation(self, op: Operation) -> None:
+        match op:
+            case riscv_scf.ForOp():
+                self.allocate_for_loop(op)
+            case RISCVOp():
+                self.process_riscv_op(op)
+            case _:
+                return
```
AntonLydike, line 59: Would it make sense to raise an exception here?

xdsl/backend/riscv/register_allocation.py
```diff
@@ -49,6 +49,32 @@ def allocate(self, reg: SSAValue) -> bool:
...
+        if len(func.body.blocks) != 1:
+            raise NotImplementedError(
+                f"Cannot register allocate func with {len(func.body.blocks)} blocks."
+            )
+
+        for op in func.body.block.ops_reverse:
+            self.process_operation(op)
+
+    @abc.abstractmethod
+    def allocate_for_loop(self, loop: riscv_scf.ForOp) -> None:
```
AntonLydike, line 75: similar here, line out what an implementation of this function is supposed to do.

xdsl/backend/riscv/register_allocation.py
```diff
@@ -49,6 +49,32 @@ def allocate(self, reg: SSAValue) -> bool:
...
+        match op:
+            case riscv_scf.ForOp():
+                self.allocate_for_loop(op)
+            case RISCVOp():
+                self.process_riscv_op(op)
+            case _:
+                return
+
+    @abc.abstractmethod
+    def process_riscv_op(self, op: RISCVOp) -> None:
```
AntonLydike, line 62: I would appreciate some comments on this, e.g. the assumption that this is called from a reverse walk, etc. It would make it much clearer for readers to understand what each implementation of this actually does.

xdsl/backend/riscv/register_allocation.py
```diff
@@ -85,38 +111,41 @@ def _free(self, reg: SSAValue) -> None:
...
-                for op in block.walk_reverse():
-                    # Do not allocate registers on non-RISCV-ops
-                    if not isinstance(op, RISCVOp):
-                        continue
+    def process_riscv_op(self, op: RISCVOp) -> None:
+        for result in op.results:
+            # Allocate registers to result if not already allocated
+            self.allocate(result)
+            # Free the register since the SSA value is created here
+            self._free(result)
```
AntonLydike, line 119: Are you sure this makes sense? Allocating a register and immediately freeing it again? 

I'm not sure I follow what's happening in this function...

AntonLydike (commented):
I'm not sure what exactly this PR does and if it makes sense?

adutilleul (approved):
Nice!

xdsl/backend/riscv/register_allocation.py
superlopuh: I added a comment, I think we just want to keep ignoring non-riscv ops for now

xdsl/backend/riscv/register_allocation.py
superlopuh: #1538

PR #1522, opened 2023-09-01
backend: (riscv) free result registers before allocating

superlopuh:
I'm not sure exactly what the purpose of `to_free` was. @compor @adutilleul am I missing something here? It feels like it just prevents us from reusing the result register for operand register, which seems like generally a good idea.


tobiasgrosser (approved):
This looks great to me.

compor (approved):
LGTM, now that I see it I'd like to have an option for it, but we can do that later maybe.

PR #1524, opened 2023-09-01
backend: (riscv) simplify for loop allocation

superlopuh:
The idea is that the `riscv_scf.for` op already checks that the types are the same, so we can treat it as a precondition of the loop allocation.


nazavode (approved):

AntonLydike (approved):
That's amazing!

PR #1527, opened 2023-09-01
[WIP] For loops with liveness

superlopuh:
My attempt at getting for loops to work with liveness. Everything seems to work, so we just need to agree on how to cleanly merge this, or what changes we need to make.


tobiasgrosser:
Why is `always move zero register` part of this PR? And what is the motivation for this patch? Is moving zero registers somehow reducing a cost. I guess a one-sentence summary of why this patch does what it does would be useful.

tobiasgrosser:
> Sorry this is a bit messy I was just trying to debug things. I'll try to revert the zero change and see if things break

No worries. This is perfect. As part of me understanding a PR I try to understand which parts of it are used how (or if there are unrelated changes). If this is an unrelated change, we can move it out eventually to keep the PR small.

In any case, this already looks quite interesting.

tobiasgrosser:
This makes a lot of sense to me. Beautiful. :heart: 

Also, `RegisterAllocatorBlockNaive` now seems to be a simplified implementation of `RegisterAllocatorLivenessBlockNaive`. Can the 2nd be configured to deliver the 1st?

I feel we might be able to eliminate `RegisterAllocatorBlockNaive` and subsequently get rid of the base class abstracting between the two (in a different PR).

Very cool.

superlopuh:
I'd love to discuss the broader strategic plans for the different allocators. Depending on different objectives it might make sense to make pretty different decisions. This is just one proposal of how they might interact, would be good to discuss tradeoffs of alternatives.

superlopuh:
Decided to add a separate PR for filecheck tests: #1535 once that's merged, I'll pipe it through the tower of Babel of regalloc PRs to keep track of what the effect of each pull request is, and open this PR as a result.

xdsl/backend/riscv/register_allocation.py
```diff
@@ -169,19 +140,70 @@ def allocate_for_loop(self, loop: riscv_scf.ForOp) -> None:
             assert isinstance(yield_operand.type, RISCVRegisterType)
             assert isinstance(op_result.type, RISCVRegisterType)
 
-            if not operand.type.is_allocated:
+            if not op_result.type.is_allocated:
```
superlopuh, line 143: op_result because we are walking backwards, and will already have allocated op_results used in the code below

tobiasgrosser (approved):

adutilleul (approved):

compor (approved):
LGTM

PR #1526, opened 2023-09-01
backend: (riscv) add sym_visibility attr to riscv_func.FuncOp

nazavode:
This PR adds to `riscv_func.FuncOp` the same visibility attribute that `func.FuncOp` has. This is the first step towards emitting correct assembler directives for global symbols.

PR #1528, opened 2023-09-01
backend: (riscv) emit assembler directives for functions

nazavode:
This PR emits all the directives needed to have function labels properly assembled (and hopefully linked) for our current default target (`rv32`). Please note that the emission of directives is done at the last possible moment where we still have structured info (e.g.: we still have `FuncOp` and its attributes like `sym_visibility`). @superlopuh I stashed your version on my laptop, if you think that is more appropriate I will close this and resume that one. Please let me know if something is needed on the `Toy` side.

**Stacked on top of #1526**


tobiasgrosser:
Very nice.

superlopuh:
Is that expected to work, a new `.text` for each function? If so then this looks good to me. It looks like the Toy tests are passing so at least riscemu seems to handle it fine.

superlopuh (approved):

nazavode:
> Is that expected to work, a new `.text` for each function? If so then this looks good to me. It looks like the Toy tests are passing so at least riscemu seems to handle it fine.

This is correct since assembler directives are flat, each section directive tells the assembler to switch in that mode and consider whatever comes next as part of the current section, until another directive makes it switch to another one. This makes sections as regions not really sound right to me (section closing is never seen by the assembler, so whatever we emit after that will still refer to the current section), but anyway it should work fine for now, modulo the linker script we have in the experiments repo. Checking for that when I have internet again.

AntonLydike (approved):
Looks good! Thanks for this!

PR #1529, opened 2023-09-03
core: Add support for lazy traits in IRDL

math-fehr:
This is an alternative implementation to #1505, which implements the laziness at the traits field level, instead of the trait definition level.
Note that we could force all traits definitions to use `traits_def`, while right now operations can both use `traits = frozenset(...)` and `traits = traits_def(frozenset(...))` and `traits = traits_def(lambda: frozenset(...))`.

I believe this implementation reduces the complexity of the traits definitions, as they don't need to specify if they are lazy or not. If a circular dependency arise, this only require a change in the operation definition, and not the trait definition, which I think is better.

### DIFF
--- a/xdsl/dialects/gpu.py
+++ b/xdsl/dialects/gpu.py
@@ -47,6 +47,7 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
+from xdsl.irdl.irdl import traits_def
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
 from xdsl.traits import (
@@ -356,10 +357,7 @@ class MemcpyOp(IRDLOperation):
 class ModuleEndOp(IRDLOperation):
     name = "gpu.module_end"
 
-    # TODO circular dependency disallows this set of traits
-    # tracked by gh issues https://github.com/xdslproject/xdsl/issues/1218
-    # traits = frozenset([HasParent(ModuleOp), IsTerminator()])
-    traits = frozenset([IsTerminator()])
+    traits = traits_def(lambda: frozenset([IsTerminator()]))
 
     def __init__(self):
         return super().__init__()
--- a/xdsl/dialects/gpu.py
+++ b/xdsl/dialects/gpu.py
@@ -45,9 +45,9 @@ from xdsl.irdl import (
     prop_def,
     region_def,
     result_def,
+    traits_def,
     var_operand_def,
 )
-from xdsl.irdl.irdl import traits_def
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
 from xdsl.traits import (
@@ -357,7 +357,7 @@ class MemcpyOp(IRDLOperation):
 class ModuleEndOp(IRDLOperation):
     name = "gpu.module_end"
 
-    traits = traits_def(lambda: frozenset([IsTerminator()]))
+    traits = traits_def(lambda: frozenset([IsTerminator(), HasParent(ModuleOp)]))
 
     def __init__(self):
         return super().__init__()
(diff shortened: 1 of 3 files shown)


math-fehr:
Note that I just added the implementation for a single operation, I'll add the other one if we think this is the right approach

superlopuh:
I see, this does make things cleaner indeed, seems like it's simpler than my suggestion for sure. I say we might as well go with this one.

xdsl/irdl/irdl.py
```diff
@@ -1074,6 +1095,18 @@ class OpDef:
...
+    def traits(self) -> frozenset[OpTrait]:
+        if callable(self._traits):
+            self._traits = self._traits()
+        return self._traits
+
+    @traits.setter
+    def traits(
+        self, traits: frozenset[OpTrait] | Callable[[], frozenset[OpTrait]]
+    ) -> None:
+        self._traits = traits
```
AntonLydike, line 1108: Do we really want a world-accessible setter for traits? I don't think these should be modifiable outside of IRDL, no?

superlopuh: I don't see the disadvantage

xdsl/dialects/gpu.py
```diff
@@ -354,10 +355,7 @@ def verify_(self) -> None:
 class ModuleEndOp(IRDLOperation):
     name = "gpu.module_end"
 
-    # TODO circular dependency disallows this set of traits
-    # tracked by gh issues https://github.com/xdslproject/xdsl/issues/1218
-    # traits = frozenset([HasParent(ModuleOp), IsTerminator()])
-    traits = frozenset([IsTerminator()])
+    traits = traits_def(lambda: frozenset([IsTerminator()]))
```
compor, line 358: I think the `HasParent` trait got removed (maybe accidentally?).
AFAIU it is still [needed](
https://mlir.llvm.org/docs/Dialects/GPU/#gpumodule_end-gpumoduleendop).
BTW, the correct definition was in gh tracking issue comment that was removed.

Unless there is another reason for the commission that I don't see.

compor (approved):
I like this version a lot.
Especially the fact that it accommodates gradual transition wherever necessary.

Thanks a lot!

xdsl/irdl/irdl.py
```diff
@@ -1046,6 +1051,20 @@ def opt_successor_def(
...
+    *,
+    default: None = None,
+    resolver: None = None,
+    init: Literal[False] = False,
+) -> frozenset[OpTrait]:
+    """
+    Defines the traits of an operation.
+    This should only be assigned on the `traits` field of an operation definition.
+    """
+    return cast(frozenset[OpTrait], _TraitsFieldDef(traits))
```
kingiler, line 1065: Should we add a check to see if `traits` is callable?

xdsl/irdl/irdl.py
math-fehr: I believe we need it, mostly for modularity reason.
If someone is defining a new trait, and they want to apply it to existing dialects, they have to override this trait here ;)

superlopuh: Scary business

math-fehr:
I finally found time to reopen this PR, I added tests as well!

superlopuh (approved):

compor:
Good stuff, already have a stacked branch fixing https://github.com/xdslproject/xdsl/issues/1218 on this :-D

math-fehr:
> Good stuff, already have a stacked branch fixing #1218 on this :-D

Nice ;)

PR #1533, opened 2023-09-04
backend: (riscv) add ability to queue j registers

superlopuh:
This lets use reuse j regs in the liveness register allocator. It's useful for tests and potentially if we want to add spilling to the stack we might benefit from something like this.


adutilleul (approved):
LGTM. Do you plan to hide this behind an option to avoid surprises when trying to use it in a real assembler, or do we want to wait further before introducing target specific features?

tobiasgrosser:
I feel `j` register should be reported by any assembler as unsupported, so don't feel we need to hide this behind an option. Happy to have it, but don't think we need this if it adds overhead.

superlopuh:
Yep this is not a short-term concern

PR #1535, opened 2023-09-04
testing: (riscv) add a unified test for register allocation

superlopuh:
Transform the jregs filecheck into a test that verifies how the two remaining allocators behave on limited but representative riscv modules, one with for loops, and one without.


AntonLydike (approved):
This is lit :fire:

PR #1536, opened 2023-09-04
interpreter: Add support for external implementations

AntonLydike:
Improvement upon #1422 

Adds support for registering external functions similar to operation and cast implementations by marking them with `@impl_external(sym_name)`

### DIFF
--- a/xdsl/interpreters/printf.py
+++ b/xdsl/interpreters/printf.py
@@ -1,12 +1,7 @@
 from typing import Any
 
 from xdsl.dialects.printf import PrintFormatOp
-from xdsl.interpreter import (
-    Interpreter,
-    InterpreterFunctions,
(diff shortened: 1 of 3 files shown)


superlopuh (changes requested):
very nice, thank you! Can you please add a pytest that exercises this?

xdsl/interpreters/printf.py
```diff
@@ -1,7 +1,12 @@
 from typing import Any
 
 from xdsl.dialects.printf import PrintFormatOp
-from xdsl.interpreter import Interpreter, InterpreterFunctions, impl, register_impls
```
superlopuh, line 4: nit: might as well undo this diff

superlopuh (approved):
thank you!

PR #1537, opened 2023-09-04
interpreter: (riscv) Add some libc runtime methods like malloc/free

AntonLydike:
Uses the interpreter external functions from #1536 to implement malloc, calloc, free and putchar.


tests/filecheck/runner/riscv.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: xdsl-run %s
```
superlopuh, line 1: can you add the filecheck bit, please?

superlopuh (changes requested):
beautiful, just one minor change remaining

tests/filecheck/runner/riscv.mlir
AntonLydike: That would indeed be quite useful to have

AntonLydike: Fixed!

superlopuh (approved):
thank you

PR #1539, opened 2023-09-05
CI: Bring back JupyterLite page deployment

PapyChacal:
Fixed JupyterLite, visible from my fork at https://papychacal.github.io/xdsl
This bumps some versions and bring back the JupyterLite auto deployment if we want it.
This builds a more minimal JupyterLite site, with no JS kernel for example.

No significant changes to the build time though, and we still have to build a custom pyodide distribution AFAIU.
Here's a cached build run example from my fork : https://github.com/PapyChacal/xdsl/actions/runs/6083243536/job/16503454876


PapyChacal:
Oh, the workflow was still disabled at the repository level. Enabled it again, will trigger a run by updating the branch.

webmiche:
Just to reiterate: We used Jupyterlite to do the browser-embedded notebooks, right? So this makes it such that these are updated with the current xDSL-version again, is that right?

I am not honestly not sure how much I missed Jupyterlite on the CI, I feel it sometimes acted weirdly, so I want to understand if it really makes sense to bring this back.

PapyChacal:
It is unstable indeed, and yes I'm not sure how much we want it back or on which basis, happy to hear everyone's opinion

Mine:
The main annoyance is having it run on every PR and every main commit, at some point the idea was to switch it to run on releases only, I'd be up for this
We can also look into just regular updates (daily, weekly..)

Also no hard feelings if we just want to get rid of it, but we should decide something as we still link to it in many places.

Alternatives I see:

- Just quit and say people we have notebooks in there for them to use as they will
- There's a free solution I don't remember the name of, based on hosting docker containers somewhere to run those notebooks. Less struggles for us, drawbacks are the need for a continuous internet connection, no local work saves, and a wait queue to open them

tobiasgrosser:
What about having nightly jupter-lite runs?

PapyChacal:
Switched the PR to nightly builds at 16:15 UTC.

PapyChacal:
Opened a Zulip thread https://xdsl.zulipchat.com/#narrow/stream/368604-Framework/topic/JupyterLite.20deployment

superlopuh (approved):

webmiche:
nightly builds sounds good to me :+1:

AntonLydike (approved):

PR #1542, opened 2023-09-05
Dialects: finishing Arith custom parse and printing

ajrichins:
adding simplified form parsing and printing for arith operations select, cmpi, cmpf, truncf, negf and extf.

added custom parsing filecheck test: @superlopuh 

Ignore commits previous to the merge, there is no changes there that aren't already in main.  I'll make sure to start fresh next time.

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_ops_custom.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_ops_custom.mlir
@@ -1,0 +1,109 @@
+// RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
+
+"builtin.module"() ({
+  %lhsi1 = "test.op"() : () -> (i1)
+  %lhsi32, %rhsi32 = "test.op"() : () -> (i32, i32)
+  %lhsi64 = "test.op"() : () -> (i64)
+  %lhsf32, %rhsf32 = "test.op"() : () -> (f32, f32)
+  %lhsf64 = "test.op"() : () -> (f64)
+  %lhsvec, %rhsvec = "test.op"() : () -> (vector<4xf32>, vector<4xf32>)
+
+
+  %divsi = arith.divsi %lhsi32, %rhsi32 : i32
+  %divui = arith.divui %lhsi32, %rhsi32 : i32
+
+  // CHECK:      %6 = arith.divsi %1#0, %1#1 : i32
+  // CHECK-NEXT: %7 = arith.divui %1#0, %1#1 : i32
+
+  %ceildivsi = arith.ceildivsi %lhsi32, %rhsi32 : i32
+  %floordivsi = arith.floordivsi %lhsi32, %rhsi32 : i32
+  %ceildivui = arith.ceildivui %lhsi32, %rhsi32 : i32
+
+  // CHECK-NEXT: %8 = arith.ceildivsi %1#0, %1#1 : i32
+  // CHECK-NEXT: %9 = arith.floordivsi %1#0, %1#1 : i32
+  // CHECK-NEXT: %10 = arith.ceildivui %1#0, %1#1 : i32
+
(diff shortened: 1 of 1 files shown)


xdsl/dialects/arith.py
```diff
@@ -418,6 +418,43 @@ def __init__(
...
+        input_type = parser.parse_type()
+        (operand1, operand2) = parser.resolve_operands(
+            [operand1, operand2], 2 * [input_type], parser.pos
+        )
+
+        return cls(operand1, operand2, arg)
+
+    def print(self, printer: Printer):
+        printer.print(" ")
+        cmpi_comparison_operations = [
```
superlopuh, line 438: Can you please move this to `_CMPI_COMPARISON_OPERATIONS` either on the file or module level?

xdsl/dialects/arith.py
```diff
@@ -489,6 +526,49 @@ def __init__(
...
```
superlopuh, line 546: same here

superlopuh (changes requested):
thank you! jsut a minor change would be good

superlopuh:
Looks like you might need to re-run the notebooks, some print MLIR, and the CI tests for the resulting text matching a re-run exactly.

ajrichins:
yeah, i had to run to class right after submitting :( I'll get those done rn

xdsl/dialects/arith.py
ajrichins: put it by other file globals

xdsl/dialects/arith.py
ajrichins: put it by other file globals

docs/database_example.ipynb
```diff
@@ -471,8 +471,16 @@
    "name": "python3"
   },
   "language_info": {
+   "codemirror_mode": {
```
superlopuh, line 474: this is a bit tedious, but could you please modify this notebook by hand to remove this diff?

docs/xdsl-introduction.ipynb
```diff
@@ -843,8 +843,22 @@
   }
  ],
  "metadata": {
+  "kernelspec": {
```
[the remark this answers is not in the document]
ajrichins: should be fixed

docs/database_example.ipynb
ajrichins: should be fixed

superlopuh (approved):
ty

superlopuh:
One thing that's missing is a filecheck for custom syntax MLIR interop. Can you please do something similar to what I do in this PR, which is to duplicate the generic syntax filecheck file, and convert it to custom syntax instead? #1545 Annoyingly, we can't parse MLIR custom syntax directly since MLIR prints `module` and we need `builtin.module`...

ajrichins:
just noticed there is only 3 ops without the custom syntax, truncf, extf and negf so I will try and add those before finishing the PR.

ajrichins:
Can't figure out why the 3.10 build failed the MLIR testing other than it seems to be a network issue?

tobiasgrosser:
This looks all great. Thank you @ajrichins for your contribution.

superlopuh:
@ajrichins I saw that you added a test with XDSL_ROUNDRIP, I'm not sure whether that was to address my comment. Is the MLIR interop test still upcoming?

xdsl/dialects/arith.py
```diff
@@ -42,6 +42,38 @@
...
+    "ole",
+    "one",
+    "ord",
+    "ueq",
+    "ugt",
+    "uge",
+    "ult",
+    "ule",
+    "une",
+    "uno",
```
math-fehr, line 73: Can you make these two constants public (without the `_`)?
In case we ever need to access it from outside.

math-fehr (approved):
Very nice!
I just added one comment, and then this is good to merge for me!

ajrichins:
> 

ahh I did misunderstand, I'll get that added

PR #1543, opened 2023-09-05
dialects: (scf) add custom syntax to scf for and yield

superlopuh:
Also makes a minor change to the parser API, to require the argument label for `expect_type` in parse_argument. Happy to remove it, but seems like a good thing to require the argument label for an ambiguous bool-valued argument such as this.

### DIFF
--- a/tests/filecheck/dialects/stencil/stencil_ops.mlir
+++ b/tests/filecheck/dialects/stencil/stencil_ops.mlir
@@ -63,17 +63,16 @@ builtin.module {
 // CHECK-NEXT:     %time_m = arith.constant 0 : index
 // CHECK-NEXT:     %time_M = arith.constant 1000 : index
 // CHECK-NEXT:     %step = arith.constant 1 : index
-// CHECK-NEXT:     %fnp1, %fn = "scf.for"(%time_m, %time_M, %step, %f0, %f1) ({
-// CHECK-NEXT:     ^0(%time : index, %fi : !stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>, %fip1 : !stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>):
+// CHECK-NEXT:     %fnp1, %fn = scf.for %time = %time_m to %time_M step %step iter_args(%fi = %f0, %fip1 = %f1) -> (!stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>, !stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>) {
 // CHECK-NEXT:       %ti = "stencil.load"(%fi) : (!stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>) -> !stencil.temp<?x?x?xf32>
 // CHECK-NEXT:       %tip1 = "stencil.apply"(%ti) ({
-// CHECK-NEXT:       ^1(%ti_ : !stencil.temp<?x?x?xf32>):
+// CHECK-NEXT:       ^0(%ti_ : !stencil.temp<?x?x?xf32>):
 // CHECK-NEXT:         %v = "stencil.access"(%ti_) {"offset" = #stencil.index<0, 0, 0>} : (!stencil.temp<?x?x?xf32>) -> f32
 // CHECK-NEXT:         "stencil.return"(%v) : (f32) -> ()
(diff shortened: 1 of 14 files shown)


xdsl/parser/core.py
```diff
@@ -515,7 +515,7 @@ def parse_optional_argument(self, expect_type: bool = True) -> Argument | None:
             type = self.parse_type()
         return self.Argument(name_token.span, type)
 
-    def parse_argument(self, expect_type: bool = True) -> Argument:
+    def parse_argument(self, *, expect_type: bool = True) -> Argument:
```
superlopuh, line 518: minor API change, this only causes new pyright errors, and shouldn't break existing code

xdsl/dialects/scf.py
```diff
@@ -171,6 +191,60 @@ def get(
...
+            for iter_arg, iter_arg_operand in parser.parse_comma_separated_list(
+                Parser.Delimiter.PAREN, lambda: parse_assignment(parser)
+            ):
+                iter_args.append(iter_arg)
+                iter_arg_operands.append(iter_arg_operand)
+            parser.parse_characters("->")
+            iter_arg_types = parser.parse_comma_separated_list(
+                Parser.Delimiter.PAREN, parser.parse_attribute
+            )
+            # PR: am I doing this right? Seems like duplicated information
```
superlopuh, line 238: @math-fehr @AntonLydike @PapyChacal do you know what the right move here is? Should I raise a ParseError?

superlopuh:
!! Just saw @compor has a branch with similar changes, but not the filechecks, I hope I just saved you a bunch of time manually updating all the test cases :)

xdsl/parser/core.py
superlopuh: Ah no I was wrong this will break code...

superlopuh: I can revert it or we can keep it, I would rather have the argument label be mandatory, but I don't want to force this in.

compor:
> 

I do, but my plan was to play with adding the terminator op (`yield`) once the criteria are met.
Happy to add this separately once this is in.

tests/filecheck/dialects/stencil/stencil_ops.mlir
```diff
@@ -63,17 +63,16 @@ builtin.module {
 // CHECK-NEXT:     %time_m = arith.constant 0 : index
 // CHECK-NEXT:     %time_M = arith.constant 1000 : index
 // CHECK-NEXT:     %step = arith.constant 1 : index
-// CHECK-NEXT:     %fnp1, %fn = "scf.for"(%time_m, %time_M, %step, %f0, %f1) ({
-// CHECK-NEXT:     ^0(%time : index, %fi : !stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>, %fip1 : !stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>):
+// CHECK-NEXT:     %fnp1, %fn = scf.for %time = %time_m to %time_M step %step iter_args(%fi = %f0, %fip1 = %f1) -> (!stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>, !stencil.field<[-4,54]x[-4,84]x[-4,44]xf32>) {
```
georgebisbas, line 66: Do people consider this syntax better than before here?

compor: it's easier to read but in a WIP branch I have for this [here](https://github.com/xdslproject/xdsl/tree/christos/dialects/scf-custom-format) I follow the MLIR custom format:

```
scf.for %iv_32 = %lb_32 to %ub_32 step %step_32 : i32 {
  ... // body
}
```

which is the most unambiguous to read and understand IMHO.

What do you think @georgebisbas?

compor: I think it's what MLIR uses and it's the most unambiguous and easy to read/understand IMHO.

compor (approved):

superlopuh:
I mean that I spent I think about an hour adjusting the filecheck files by hand for this PR, maybe two, lost track of time...

compor:
> I mean that I spent I think about an hour adjusting the filecheck files by hand for this PR, maybe two, lost track of time...

I see, and that is true :-)
I'll base the implicit addition of `yield` on this one once in.

georgebisbas (approved):

PR #1544, opened 2023-09-05
 dialects: (riscv_scf) add custom syntax to riscv_scf for and yield

superlopuh:


xdsl/dialects/riscv_scf.py
```diff
@@ -112,6 +132,60 @@ def verify_(self):
...
+            for iter_arg, iter_arg_operand in parser.parse_comma_separated_list(
+                Parser.Delimiter.PAREN, lambda: parse_assignment(parser)
+            ):
+                iter_args.append(iter_arg)
+                iter_arg_operands.append(iter_arg_operand)
+            parser.parse_characters("->")
+            iter_arg_types = parser.parse_comma_separated_list(
+                Parser.Delimiter.PAREN, parser.parse_attribute
+            )
+            # PR: am I doing this right? Seems like duplicated information
```
superlopuh, line 179: same Q as other PR

xdsl/dialects/riscv_scf.py
```diff
@@ -112,6 +132,60 @@ def verify_(self):
                         f"variables types."
                     )
 
+    def print(self, printer: Printer):
+        block = self.body.block
```
superlopuh, line 136: another question is whether to factor this into one or multiple helpers since the scf and riscv_scf for loops are supposed to have consistent custom syntax

compor: I cannot forsee anything that would require different custom printing here, but up to you.

compor (approved):

PR #1546, opened 2023-09-06
backend: (riscv) make live-ins iteration stable

nazavode:
This PR makes register allocation stable by using an `OrderedSet` for live-ins computation.

Note: we were observing random failures in `filecheck` in the experiments repo, with correct allocations but with registers changing randomly.


tobiasgrosser:
Thats a great idea.

compor (approved):
I'd like a test here, but not sure how to go about it.
I'd rather have it in than not though.

superlopuh (approved):

nazavode:
> I'd like a test here, but not sure how to go about it.
> I'd rather have it in than not though.

I had to adjust expected values in 665fe7bccabf41b4940bf96bdfaf1af9b43cc2e0 because of `OrderedSet.__eq__` being ordered, having that test break is reassuring imo :)

adutilleul (approved):

PR #1547, opened 2023-09-06
CI: Remove conditions on JupyterLite deploy.

PapyChacal:
Good news: the new daily JupyterLite nighlty workflow seems to work (I was just mistaken and set it to run at 5AM not 5PM. If anybody cares, this PR is a nice time to argue about the exact time!)

I just forgot to remove a now obsolete condition on the deploy job, so it was not triggered. (It only built the page without deploying it) This PR solves this.

First scheduled run: https://github.com/xdslproject/xdsl/actions/runs/6092690186

### DIFF
--- a/.github/workflows/jupyterlite.yml
+++ b/.github/workflows/jupyterlite.yml
@@ -78,7 +78,6 @@ jobs:
 
   deploy:
     needs: build
-    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
     permissions:
       pages: write
       id-token: write


webmiche:
I think the current time is quite fine. Sometime where noone is working, so 5am in the UK or something. I know Mathieu sometimes works until quite late and I usually start at around 8 CEST, so the current time seems a good fit to me.

I also don't think this is insanely important. AFAIU, it will just take the most recent commit, so we could schedule it literally anytime (when main is not broken, which is most of the times).

superlopuh (approved):

compor (approved):

PR #1548, opened 2023-09-07
misc: use isinstance instead of type equality

superlopuh:


compor (approved):

webmiche (approved):

tests/dialects/test_builtin.py
```diff
@@ -36,9 +36,9 @@ def test_DenseIntOrFPElementsAttr_fp_type_conversion():
     value2 = check1.data.data[1].value.data
 
     # Ensure type conversion happened properly during attribute construction.
-    assert type(value1) == float
```
webmiche, line 39: I guess there cannot be subtypes of floats, right? So this is equivalent, i guess.

PR #1550, opened 2023-09-07
backend: (riscv) exclude preallocated registers by default

superlopuh:
Without this our code is somewhat broken, especially in the presence of function calls or function arguments.


compor:
Agreed.

Having it as a default, can you please remove it from the [filecheck invocation](https://github.com/xdslproject/xdsl/blob/3bb698f4b6dc4b5a3cb8871b914f61135b59b323/tests/filecheck/backend/riscv/register_allocation_preallocated.mlir#L1) as well?

compor (approved):

adutilleul (approved):

JosseVanDelm:
Nice!

PR #1552, opened 2023-09-07
backend: (rsicv) unify RdRsRs(Integer|Float)Op

superlopuh:
Similarly to an earlier PR, unifying two abstract operations, in preparation for the addition of some vector ops required for Snitch.

### DIFF
--- a/xdsl/backend/riscv/lowering/convert_arith_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_arith_to_riscv.py
@@ -1,4 +1,4 @@
-from typing import overload
+from dataclasses import dataclass
 
 from xdsl.backend.riscv.lowering.utils import (
     cast_matched_op_results,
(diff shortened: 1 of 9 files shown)


xdsl/backend/riscv/lowering/convert_arith_to_riscv.py
```diff
@@ -93,74 +93,54 @@ def match_and_rewrite(
         )
 
 
-class LowerBinaryOp(RewritePattern):
```
superlopuh, line 96: this was just too clever by half, two separate classes, one for floats and one for ints is less code than this one with all its helpers

xdsl/dialects/riscv.py
```diff
@@ -568,30 +570,28 @@ def riscv_code(module: ModuleOp) -> str:
...
+    rd: OpResult = result_def(RDInvT)
+    rs1: Operand = operand_def(RS1InvT)
+    rs2: Operand = operand_def(RS2InvT)
 
     def __init__(
         self,
         rs1: Operation | SSAValue,
         rs2: Operation | SSAValue,
         *,
-        rd: IntRegisterType | str | None = None,
```
superlopuh, line 588: no dynamic generics, so the type of RD cannot be inferred :(

xdsl/dialects/riscv.py
```diff
@@ -232,6 +232,8 @@ def a_register(cls, index: int) -> Self:
 
 RDInvT = TypeVar("RDInvT", bound=RISCVRegisterType)
 RSInvT = TypeVar("RSInvT", bound=RISCVRegisterType)
+RS1InvT = TypeVar("RS1InvT", bound=RISCVRegisterType)
+RS2InvT = TypeVar("RS2InvT", bound=RISCVRegisterType)
```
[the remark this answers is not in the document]
superlopuh: If we want the ability for RS1 and RS2 to be different then they must be different type variables, since substitution happens at the type variable directly, as opposed to the generic parameter (like in every other generic type system I'm aware of).

xdsl/dialects/riscv.py
superlopuh: I cannot substitute `RDInvT` here since, at the point where the init is called, the information about what the concrete type is that this class has been constrained to has been thrown away.

superlopuh: As far as I know it's not possible to hack it like this: `RD: ClassVar[type[RDInvT]]`, this gets rejected with `"ClassVar" type cannot include type variables`

nazavode (approved):

adutilleul (approved):

xdsl/dialects/riscv.py
```diff
@@ -2755,43 +2755,6 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
         return self.rd, self.rs1, self.rs2, self.rs3
 
 
-class RdRsRsFloatOperation(IRDLOperation, RISCVInstruction, ABC):
```
compor, line 2758: I like this unification a lot!

compor (approved):

PR #1553, opened 2023-09-07
backend: (riscv) print assembly from riscv_func, remove region from label

superlopuh:
I don't think we need the extra step of rewriting to labels, and we benefit from skipping it. For one, fewer rewrites means fewer opportunities for bugs. Also, we get fewer `GetRegisterOp`s, and having 0 of them lets us interpret the code directly. As discussed before, dropping the region from labels also makes the code a bit less weird/confusing.

### DIFF
--- a/docs/Toy/toy/rewrites/setup_riscv_pass.py
+++ b/docs/Toy/toy/rewrites/setup_riscv_pass.py
@@ -26,11 +26,8 @@ class AddSections(RewritePattern):
             ),
         )
         data_section = riscv.AssemblySectionOp(".data", Region(Block()))
-        text_section = riscv.AssemblySectionOp(
-            ".text", rewriter.move_region_contents_to_new_regions(op.regions[0])
-        )
 
-        op.body.add_block(Block([heap_section, data_section, text_section]))
+        rewriter.insert_op_at_start((heap_section, data_section), op.body.block)
 
 
 class SetupRiscvPass(ModulePass):
--- a/tests/dialects/test_riscv.py
+++ b/tests/dialects/test_riscv.py
@@ -1,6 +1,5 @@
 import pytest
 
-from xdsl.builder import Builder
 from xdsl.dialects import riscv
 from xdsl.dialects.builtin import IntegerAttr, ModuleOp, i32
 from xdsl.ir import MLContext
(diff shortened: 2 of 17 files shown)


docs/Toy/toy/rewrites/setup_riscv_pass.py
```diff
@@ -26,11 +26,8 @@ def match_and_rewrite(self, op: ModuleOp, rewriter: PatternRewriter):
             ),
         )
         data_section = riscv.AssemblySectionOp(".data", Region(Block()))
-        text_section = riscv.AssemblySectionOp(
-            ".text", rewriter.move_region_contents_to_new_regions(op.regions[0])
```
superlopuh, line 30: The ".text" assembler directive is now inserted before every function in xDSL so this one is not needed

tests/dialects/test_riscv.py
```diff
@@ -109,22 +108,6 @@ def test_label_op_with_comment():
     assert code == f"{label_str}:                                         # my label\n"
 
 
-def test_label_op_with_region():
```
superlopuh, line 112: labels no longer have regions, so this test is not needed

tests/filecheck/dialects/riscv/riscv_ops.mlir
```diff
@@ -397,6 +397,6 @@
 // CHECK-GENERIC-NEXT:     %fmv_w_x = "riscv.fmv.w.x"(%0) : (!riscv.reg<>) -> !riscv.freg<>
 // CHECK-GENERIC-NEXT:     %flw = "riscv.flw"(%0) {"immediate" = 1 : si12} : (!riscv.reg<>) -> !riscv.freg<>
 // CHECK-GENERIC-NEXT:     "riscv.fsw"(%0, %f0) {"immediate" = 1 : si12} : (!riscv.reg<>, !riscv.freg<>) -> ()
-// CHECK-GENERIC-NEXT:     "riscv.ret"() : () -> ()
-// CHECK-GENERIC-NEXT:   }) {"label" = #riscv.label<"main">} : () -> ()
+// CHECK-GENERIC-NEXT:     "riscv_func.return"() : () -> ()
+// CHECK-GENERIC-NEXT:   }) {"sym_name" = "main", "function_type" = () -> ()} : () -> ()
```
kingiler, line 401: Just curious why we need `"function_type" ` attributes in `FuncOp`.

kingiler (approved):
Thank you! It makes code much easier to read.

tests/filecheck/dialects/riscv/riscv_ops.mlir
superlopuh: We copied this over from `func.func`, my understanding is that it's useful for external function declarations, among other things. At one point we might want to adopt the FunctionOpInterface from MLIR and it's a requirement there AFAIK. https://mlir.llvm.org/docs/Dialects/Func/#funcfunc-funcfuncop

nazavode (approved):
> As discussed before, dropping the region from labels also makes the code a bit less weird/confusing.

Nice, thanks.

adutilleul (approved):

compor (approved):

PR #1558, opened 2023-09-08
transformations: (riscv_scf) add riscv-scf-loop-range-folding like MLIR

superlopuh:
After loweiring memref accesses in loops, we get a lot of multiplications by the same immediate. This transformation is the same as used in LLVM/GCC to hoist the immediate multiplication into the loop parameters.


compor (approved):

compor:
Coverage is lacking a bit, but as this is not critical, we can revisit a bit later

xdsl/transforms/riscv_scf_loop_range_folding.py
```diff
@@ -0,0 +1,95 @@
...
+                ),
+                new_ub := riscv.AddOp(
+                    op.lb, new_distance, rd=riscv.IntRegisterType.unallocated()
+                ),
+                new_ub := riscv.AddOp(
+                    new_lb, new_distance, rd=riscv.IntRegisterType.unallocated()
+                ),
+                new_step := riscv.MulOp(
+                    op.step, factor, rd=riscv.IntRegisterType.unallocated()
+                ),
```
AntonLydike, line 71: I'm afraid I don't quite understand why we need this "dance" here with distances, etc. Shouldn't we just be able to go: `new_x = x * const` for lb, ub and step?

Also, it may or may not be a good idea to do overflow checking?

xdsl/transforms/riscv_scf_loop_range_folding.py
```diff
@@ -0,0 +1,95 @@
...
+                old_distance := riscv.SubOp(
+                    op.ub, op.lb, rd=riscv.IntRegisterType.unallocated()
+                ),
+                new_distance := riscv.MulOp(
+                    old_distance, factor, rd=riscv.IntRegisterType.unallocated()
+                ),
+                new_ub := riscv.AddOp(
+                    op.lb, new_distance, rd=riscv.IntRegisterType.unallocated()
+                ),
```
AntonLydike, line 65: This op seems redundant?

AntonLydike (approved):
This feels super nice to have in! I feel like we are getting much closer to a useful backend very quickly now!

xdsl/transforms/riscv_scf_loop_range_folding.py
superlopuh: Can you come up with the counterexample? ;)

xdsl/transforms/riscv_scf_loop_range_folding.py
superlopuh: Nice spot it gets removed by dce so I never saw it in output

xdsl/transforms/riscv_scf_loop_range_folding.py
superlopuh: I no longer remember why I added it, so I'll remove it

superlopuh: The overflow check sounds like a good idea, not sure if it'll ever be a practical concern

PR #1560, opened 2023-09-08
[WIP] changes necessary for the riscv experiments to work

superlopuh:
This is a collection of changes that are already in flight, I hope this PR will be closed within a few days, just wanted to make the code available


superlopuh:
Looks like there's one change I forgot about, which is to fold the multiplication into the loop, but I'm not 100% sure we even need it any more. In my measurement, it improved the performance of our float kernel by 2.5%, but that's not really in the compilation flow that we care about. I'll close this PR for now.

PR #1561, opened 2023-09-09
Interpreter: Handling attributeOp in pdl rewrites

ajrichins:
added run_attribute to the pdl interpreter.  Not super happy with how the test asserts, but not sure best way to go about it. @superlopuh  @math-fehr

### DIFF
--- a/tests/interpreters/test_pdl_interpreter.py
+++ b/tests/interpreters/test_pdl_interpreter.py
@@ -302,7 +302,6 @@ def test_interpreter_attribute_rewrite():
     interpreter.register_implementations(PDLRewriteFunctions(MLContext()))
 
     input_module = constant_zero()
-    output_module = constant_zero()
     expected_module = constant_one()
     rewrite_module = change_constant_value_pdl()
     rewrite_module.verify()
@@ -321,8 +320,4 @@ def test_interpreter_attribute_rewrite():
         apply_recursively=False,
     ).rewrite_module(input_module)
 
-    assert (
-        not output_module.body.ops.first.attributes["value"]
-        == input_module.body.ops.first.attributes["value"]
-    )
     assert expected_module.is_structurally_equivalent(input_module)
--- a/tests/interpreters/test_pdl_interpreter.py
+++ b/tests/interpreters/test_pdl_interpreter.py
@@ -256,6 +256,15 @@ def constant_zero():
     return ir_module
 
 
+def constant_one():
+    @ModuleOp
+    @Builder.implicit_region
+    def ir_module():
+        arith.Constant.from_int_and_width(1, 32)
+
+    return ir_module
+
+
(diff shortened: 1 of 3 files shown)


xdsl/interpreters/experimental/pdl.py
```diff
@@ -284,6 +284,14 @@ def run_result(
         assert isinstance(parent, Operation)
         return (parent.results[op.index.value.data],)
 
+    @impl(pdl.AttributeOp)
+    def run_attribute(
+        self, interpreter: Interpreter, op: pdl.AttributeOp, args: tuple[Any, ...]
+    ) -> tuple[Any, ...]:
+        op.verify()
```
superlopuh, line 291: why the verify?

tests/interpreters/test_pdl_interpreter.py
```diff
@@ -245,3 +245,81 @@ def test_interpreter_functions():
...
+
+    ctx = MLContext()
+    ctx.register_dialect(arith.Arith)
+
+    PatternRewriteWalker(
+        PDLRewritePattern(pdl_rewrite_op, ctx, file=stream),
+        apply_recursively=False,
+    ).rewrite_module(input_module)
+
+    assert not input_module.is_structurally_equivalent(output_module)
```
superlopuh, line 325: why not? maybe I'm missing something, but wouldn't we want to test that the result specified?

xdsl/interpreters/experimental/pdl.py
ajrichins: Not really sure, just thought it would be useful. I'll take it out

tests/interpreters/test_pdl_interpreter.py
ajrichins: is_structurally_equivalent returns false because I swapped the inputs of the cmpi op, changing the mapping of the dependencies

superlopuh: Can you please change it so that it checks for equality instead? So that we get notified when the behaviour changes

superlopuh (changes requested):
please update the test so that we keep track that the behaviour is the one that we want. Alternatively, replace the test with a simple unit test that checks the behaviour of the op function implementation directly, I'm not sure we need a full integration test for this

tests/interpreters/test_pdl_interpreter.py
ajrichins: gotcha, testing for positive test case now

ajrichins:
I think I created a simpler more direct test that is not as integrated.  Don't know how/unable to resolve most recent request to change.

superlopuh (approved):
Beautiful, thank you. You can't resolve a request to change, the person requesting the change must approve the PR

PR #1564, opened 2023-09-10
interpreter: (riscv) use bytearray for riscv malloc

superlopuh:
This is a change in the infrastructure required for interpreting the vector arithmetic ops in riscv.

In order to interpret storing/loading 64 bit values that are actually two 32 bit floats we're going to have to start dealing with the bytes directly.

### DIFF
--- a/docs/Toy/toy/emulator/toy_accelerator_instruction_functions.py
+++ b/docs/Toy/toy/emulator/toy_accelerator_instruction_functions.py
@@ -1,17 +1,8 @@
-from typing import Generic, TypeVar
-
 from xdsl.dialects import riscv
 from xdsl.dialects.builtin import ModuleOp
 from xdsl.interpreter import Interpreter, PythonValues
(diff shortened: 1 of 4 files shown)


docs/Toy/toy/emulator/toy_accelerator_instruction_functions.py
```diff
@@ -38,37 +29,27 @@ def print_(
...
+    assert len(args) == 3
+    ptr: RawPtr = args[0]
+    rows: int = args[1]
+    cols: int = args[2]
+    shaped_array = ShapedArray(ptr.float32.get_list(rows * cols), [rows, cols])
     interpreter.print(f"{shaped_array}")
     return ()
 
 
-def accelerator_tensor_transpose2d(
```
superlopuh, line 56: forgot to remove this and unused functions below

nazavode (approved):

adutilleul (approved):

PR #1565, opened 2023-09-10
dialects: (riscv) add fsd and fld + interpreter functions

superlopuh:
Unlocks the way for us to load or store a 64-bit region of memory into float registers.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -3283,6 +3283,11 @@ class FSwOp(RsRsImmFloatOperation):
         )
 
 
+# endregion
+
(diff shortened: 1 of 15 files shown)


xdsl/dialects/riscv.py
```diff
@@ -3272,6 +3272,50 @@ def assembly_line(self) -> str | None:
         )
```
adutilleul, line 3274:
```suggestion
# region RV32F: 9 “D” Standard Extension for Double-Precision Floating-Point, Version 2.0
```
This is another extension, so probably best to specify this even if this is only the start of the implementation.

adutilleul (approved):

compor (approved):

PR #1572, opened 2023-09-13
core: Implement properties retrocompatibility in the parser.

PapyChacal:
This implements properties retrocompatibility in the parser as MLIR does, that is, allow to parse a properties-defining operation in generic syntax when it has no property dictionary at all.
In that case, and only that case, it will extract properties from the attribute dictionnary to the properties one.

By *only* that case, I mean that the point is not to relax the cut between properties and attributes: if properties are present on the generic syntax of the operation, we won't be fetching missing ones from the attribute dictionary. This is meant only to parse operations printed before switching to properties.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1144,7 +1144,7 @@ class UnrealizedConversionCastOp(IRDLOperation):
         printer.print_op_attributes(self.attributes)
 
 
-class UnregisteredOp(IRDLOperation, ABC):
+class UnregisteredOp(Operation, ABC):
     """
     An unregistered operation.
 
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -8,6 +8,8 @@ from xdsl.dialects.builtin import (
     Builtin,
     DictionaryAttr,
     IntAttr,
+    IntegerAttr,
+    IntegerType,
     StringAttr,
     SymbolRefAttr,
     i32,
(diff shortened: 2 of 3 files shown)


math-fehr (approved):
:+1:

xdsl/dialects/builtin.py
```diff
@@ -1144,7 +1144,7 @@ def print_fn(operand: SSAValue) -> None:
         printer.print_op_attributes(self.attributes)
 
 
-class UnregisteredOp(IRDLOperation, ABC):
+class UnregisteredOp(Operation, ABC):
```
superlopuh, line 1147: 👍

superlopuh (approved):

compor (approved):

webmiche:
> This is meant only to parse operations printed before switching to properties.
 So we can delete this after the switch? Can we somehow make sure that we actually do? Even though this is new code, maybe add a deprecated warning or whatever? (Not sure if that makes sense, I didn't look at the code in depth).

superlopuh:
I think an issue and a comment would be good

math-fehr:
Yes, we could add a comment and an issue.
Note that this compatibility is still something that exists in MLIR. Though it's better if we can change it yes

PapyChacal:
Hm, I'm actually not sure we want to delete it? What is the issue of keeping it once it's here?

I'm okay with *trying* to get rid at some point, but I'd like to note that it's a really cumbersome and boring task; probably underestimated!
I just went through something similar for the MLIR update; and without a hard reason to do it as here, I'm honestly not planning to do that again anytime soon.

Note that most of my ideas of streamlining / CIing filechecks are really motivated by this. Making those changes less manual and less last-minute-everywhere once needed.

PapyChacal:
Created #1581

PR #1573, opened 2023-09-13
core: Add segment sizes attributes to IRDL operation definition

math-fehr:
Currently, `operand_segment_sizes` is not registered as an attribute in the operation definition class (`OpDef`).
This PR adds this to the operation definition.
This is going to simplify the update to latest MLIR, as these will become properties.


PapyChacal (approved):
Thanks :pray:

compor (approved):

PR #1574, opened 2023-09-13
[WIP] (dialects/transformations) Lowering from linalg to snitch

compor:
This is a WIP PR that has a first attempt at lowering from `linalg` to `snitch` going through an intermediate dialect `snitchstream`.

Example:

```mlir
func.func public @ssum(%X: memref<364xf32>,
                       %Y: memref<364xf32>,
                       %Z: memref<364xf32>) -> () {
  "linalg.generic"(%X, %Y, %Z) ({
  ^0(%x : f32, %y : f32, %z: f32):
    %r0 = arith.addf %x, %y : f32
    "linalg.yield"(%r0) : (f32) -> ()
  }) {"indexing_maps" = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>], "iterator_types" = [#linalg.iterator_type<parallel>], "operand_segment_sizes" = array<i32: 2, 1>} : (memref<364xf32>, memref<364xf32>, memref<364xf32>) -> ()
  func.return
}
```
to

```mlir
func.func public @ssum(%X : memref<364xf32>, %Y : memref<364xf32>, %Z : memref<364xf32>) {
  "snitchstream.stream"(%X, %Y, %Z) ({
  ^0(%0 : f32, %1 : f32, %2 : f32):
    %3 = arith.addf %0, %1 : f32
    "snitchstream.yield"(%3) : (f32) -> ()
  }) {"dimension" = 0 : i32, "bound" = 364 : i32, "stride" = 4 : i32, "operand_segment_sizes" = array<i32: 2, 1>} : (memref<364xf32>, memref<364xf32>, memref<364xf32>) -> ()
  func.return
}
```

The following passes are added to our current lowering pipeline:

```
xdsl-opt -p lower-linalg-to-snitch-stream,convert-arith-to-riscv,lower-snitchstream-to-snitch,lower-snitch, ...
```

- [ ] Decide when to (optionally) emit `frep` from the `snitchstream` lowering
- [ ] Decide how/when to optionally emit vector instructions (e.g., from the `arith` lowering)
- [ ] Determine implicit use of `ftX` registers per SSR stream
- [ ] Tests
   - [ ] Tests for 2D and 3D loops
- [ ] Documentation


superlopuh:
My understanding is that we would want to first transform the linalg loop on scalars into a linalg loop on 2xf32, so the input would actually look different to the one in the description here.

compor:
> My understanding is that we would want to first transform the linalg loop on scalars into a linalg loop on 2xf32, so the input would actually look different to the one in the description here.

That is true, I think this can be accommodated, but I was more interested in seeing what is involved in the actual lowering(s).

PR #1575, opened 2023-09-13
core: add FlatSymbolRefAttr

luisacicolini:


xdsl/dialects/builtin.py
```diff
@@ -218,6 +220,22 @@ def string_value(self):
         return root
 
 
+class EmptyArrayAttrConstraint(AttrConstraint):
```
superlopuh, line 223: Could you please add some documentation strings for this class, and the two values below?

xdsl/dialects/builtin.py
```diff
@@ -232,8 +236,10 @@ def verify(self, attr: Attribute, constraint_vars: dict[str, Attribute]) -> None
 FlatSymbolRefAttrConstraint = ParamAttrConstraint(
     SymbolRefAttr, [AnyAttr(), EmptyArrayAttrConstraint()]
 )
+"""Constrain SymbolRef to be FlatSymbolRef"""
 
 FlatSymbolRefAttr = Annotated[SymbolRefAttr, FlatSymbolRefAttrConstraint]
+"""SymbolRef constrained to Flat"""
```
superlopuh, line 242:
```suggestion
"""SymbolRef constrained to have an empty `nested_references` property."""
```

superlopuh (approved):
Thank you! Can you please expand a bit on one of the documentation strings? Otherwise good to go.

PapyChacal (approved):

compor (approved):

tests/filecheck/dialects/func/func_invalid.mlir
```diff
@@ -31,3 +31,9 @@ builtin.module {
 }
 
 // CHECK: Expected all arguments to be named or all arguments to be unnamed.
+
+// -----
+
+"func.call"() { "callee" = @call::@invalid } : () -> ()
+
+// CHECK:  Operation does not verify: expected empty array, but got ["invalid"]
```
webmiche, line 39: This does not feel like an actionable error message. What is an array in this context? Can we somehow make this understandable?

math-fehr: We have an idea, but that's going to be for another PR.
Essentially we'll add a new constraint that will catch an error, and return it.
For now at least, this error is better than the previous one ;)

superlopuh: #1579

PR #1576, opened 2023-09-13
misc: add filecheck file with assembly-level IR output for riscv paper/presentation

superlopuh:
This PR adds a filechek file to keep track of changes to the target IR that we want to produce when compiling to snitch. Currently has the first two examples, scalar and vectorised, more to follow.


compor (approved):
Thanks for this; I was wondering how we wanted to "bridge" things esp. with the experiments repo, I think this should be a good start.

adutilleul (approved):

PR #1578, opened 2023-09-13
dialects: Add HLS dialect

gabrielrodcanal:
HLS dialect and HLS transformations. The dialect is an abstraction of the typical HLS constructs, giving support for interfaces, task-parallellism and so on. The transformations focus on lowering the stencil dialect to FPGA through a shift buffer based optimisation. Currently we have support for PSyclone implemented. Support for Devito will be added soon.


superlopuh:
I'm surprised this got merged without any reviews/approves?

mesham:
> I'm surprised this got merged without any reviews/approves?

This is in experimental, I thought that was the purpose to enable fast turnaround and experimentation (it was live for 3 months to allow comments)

mesham:
Ok, that's perfectly reasonable - and my fault for telling Gabriel to just merge it in if there weren't any blocker comments as it's independent of anything else. However, it would be good to document all this somewhere (i.e. how many approvals are needed for a merge etc)?

superlopuh:
Yep we definitely have some room for improvement on communication on process, I started a wiki page for regular contributions recently but experimental is especially tricky, would be good to work together on some guidelines

PR #1579, opened 2023-09-14
core: Add a wrapper constraint to substitute error message on failure

superlopuh:


math-fehr (approved):
Really nice! That will be quite useful

webmiche (commented):
Nice, I like the core idea a lot. The one thing I dislike a bit is that this feels like quite a hack. Having a custom error really isn't a constraint on the actual thing. Is there a way to separate the two concepts more?

I do see how this "hack" makes the code simple, so I am fine with living with it in this case.

superlopuh:
One idea we briefly discussed was the ability to inject a custom error message into every constraint but this seems a bit cleaner to me. I'd like to dig into constraints a little more before too long, and make them a bit more declarative, eliminating as much python verification as possible, so this is a step in that direction also.

webmiche:
Okay, sounds good. In any case, this is a step in a good direction, is let's merge it and iterate on it later. Would it make sense to open an issue? Just to have the discussion recorded somewhere.

superlopuh:
What would you name the issue? Could you please file one that you feel would be appropriate?

webmiche:
No idea, i jsut did something. Feel free to expand.

PR #1580, opened 2023-09-14
interpreter: Support properties in PDL interpreter

math-fehr:
This PR adds support for properties in the PDL interpreter.
In particular, this means that matching an operation will match attributes both in the attribute and property dictionary.
Also, creating a new operation with the PDL attribute dictionary will split the dictionary into an attribute and property dictionary, using the IRDL operation specification.

### DIFF
--- a/xdsl/interpreters/experimental/pdl.py
+++ b/xdsl/interpreters/experimental/pdl.py
@@ -7,6 +7,7 @@ from xdsl.dialects import pdl
 from xdsl.dialects.builtin import IntegerAttr, IntegerType, ModuleOp
 from xdsl.interpreter import Interpreter, InterpreterFunctions, impl, register_impls
 from xdsl.ir import Attribute, MLContext, Operation, OpResult, SSAValue, TypeAttribute
+from xdsl.irdl.irdl import IRDLOperation
 from xdsl.pattern_rewriter import PatternRewriter, RewritePattern
 from xdsl.utils.exceptions import InterpretationError
 from xdsl.utils.hints import isa
@@ -130,10 +131,10 @@ class PDLMatcher:
         for avn, av in zip(attribute_value_names, pdl_op.attribute_values):
             assert isinstance(av, OpResult)
             assert isinstance(av.op, pdl.AttributeOp)
-            if avn not in xdsl_op.attributes:
+            if (attr := xdsl_op.get_attr_or_prop(avn)) is None:
                 return False
 
-            if not self.match_attribute(av, av.op, avn, xdsl_op.attributes[avn]):
+            if not self.match_attribute(av, av.op, avn, attr):
                 return False
 
         pdl_operands = pdl_op.operand_values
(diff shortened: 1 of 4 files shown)


xdsl/interpreters/experimental/pdl.py
```diff
@@ -268,10 +269,30 @@ def run_operation(
...
+        attributes = dict[str, Attribute]()
+        properties = dict[str, Attribute]()
+
+        # If the op is an IRDL-defined operation, get the property names.
+        if issubclass(op_type, IRDLOperation):
+            property_names = op_type.irdl_definition.properties.keys()
+        else:
+            property_names = []
+
+        # Move the attributes to the attribute or property dictionary
```
superlopuh, line 281: 😵‍💫

xdsl/ir/core.py
```diff
@@ -769,6 +769,13 @@ def walk_reverse(self) -> Iterator[Operation]:
             yield from region.walk_reverse()
         yield self
 
+    def get_attr_or_prop(self, name: str) -> Attribute | None:
```
superlopuh, line 772: Seems like this isn't used? I feel like we should either remove this diff or use it and test it

superlopuh (approved):

xdsl/ir/core.py
PapyChacal: It is at https://github.com/xdslproject/xdsl/blob/63cd884c75dbd96af3dc7b965ef16e13f3275970/xdsl/interpreters/experimental/pdl.py#L120 :slightly_smiling_face:

webmiche: Could you add a test for the `return None`-path? Also, do we actually want to return None in that case or just throw an error?

webmiche: I see, None is used as the way to identify it's not in there for the interpreter. I personally think it makes more sense to throw en error here. Or maybe you could add a keyword argument to define what should be returned by default? (like list.get does)

superlopuh:
On second thought, this seems weird to me. Does MLIR PDL accept properties in its attribute matching?

xdsl/ir/core.py
superlopuh: I think we need tests for all paths, and a clear documentation for why this is needed and whether it's temporary. It seems pretty weird to me.

math-fehr: I'll add tests, and also doumentation for it yes!
The reason we want this is in order to be compatible with MLIR, we often need to get an attribute that might also be a property at that point.
I feel that this will stay for a while (especially since PDL uses it), but it might be removed once everything move to properties.
Note that MLIR uses *a lot* this method.

math-fehr:
> On second thought, this seems weird to me. Does MLIR PDL accept properties in its attribute matching?

In MLIR PDL, everything use the `get_attr_or_prop` method, and essentially attributes and properties can be seed sharing the same dictionary attribute.
My understanding at least is that this is how this works, since they didn't want to make new problems for downstream users.

superlopuh:
Classic

math-fehr:
I added tests for `get_attr_or_prop` now ;)

PR #1587, opened 2023-09-15
dialects: (riscv) add frep operations to riscv dialect

superlopuh:
This is the first sketch of a frep dialect. I think it would make most sense for frep to take some operands and return some also, something to keep track of the fact that the stream registers are used correctly. Maybe these could be riscv registers, maybe a new kind of streaming register, not sure yet. I think we can do this later as we build up the compilation flow.


xdsl/dialects/riscv.py
```diff
@@ -433,7 +433,7 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
 
 
 AssemblyInstructionArg: TypeAlias = (
-    AnyIntegerAttr | LabelAttr | SSAValue | IntRegisterType | str
+    AnyIntegerAttr | LabelAttr | SSAValue | IntRegisterType | str | int
```
superlopuh, line 436: `int` is already supported, turns out, not sure why it's not an option in this union

xdsl/dialects/riscv.py
```diff
@@ -88,7 +88,7 @@ def parse_parameter(cls, parser: AttrParser) -> str:
         if name is None:
             return ""
         if not name.startswith("j"):
-            assert name in cls.abi_index_by_name()
+            assert name in cls.abi_index_by_name(), f"{name}"
```
superlopuh, line 91: was useful in debugging

webmiche:
Must admit, I was a bit confused when first reading this PR about what `frep` is supposed to mean. Is there a way to name this dialect that is more parsable? Maybe `snitch-frep` (just from glancing over some comments). And I guess there is a good reason why this is its own dialect and not in the snitch file.

superlopuh:
I guess we could move it to snitch also. I thought it looked cute with the `frep.outer` and `frep.inner` which is what the riscv instructions look like in assembly but I'm not attached to it being a separate dialect.

xdsl/dialects/frep.py
```diff
@@ -0,0 +1,139 @@
...
+            if not instruction.has_trait(Pure):
+                raise VerifyException(
+                    "Frep operation body may not contain instructions "
+                    f"with side-effects, found {instruction.name}"
+                )
+
+
+@irdl_op_definition
+class Outer(FRepOperation):
+    name = "frep.outer"
```
JosseVanDelm, line 125: Maybe add the main differences between outer and inner?

tests/filecheck/dialects/frep/assembly_printing.mlir
```diff
@@ -0,0 +1,17 @@
+// RUN: xdsl-opt -t riscv-asm %s | filecheck %s
+
+%i0 = riscv.get_register : () -> !riscv.reg<a0>
+%ft0 = riscv.get_float_register : () -> !riscv.freg<ft0>
+%ft1 = riscv.get_float_register : () -> !riscv.freg<ft1>
+frep.outer %i0, 0, 0 ({
+    %ft2 = riscv.vfadd.s %ft0, %ft1 : (!riscv.freg<ft0>, !riscv.freg<ft1>) -> !riscv.freg<ft2>
```
JosseVanDelm, line 7: Could you change the example to use regular float operation instead of a vector one?
Seems a bit less specfic?

JosseVanDelm (approved):
This is really cool, but I do agree that this is a snitch specific thing (for now).
The problem with the snitch naming is that snitch is just a RISC-V implementation.
The ISA extensions presented are actually `xssr`, and `xfrep`
So maybe they belong in the riscv dialect actually? or a `riscv.xssr` and a `riscv.xfrep` dialect?

adutilleul (approved):

compor (approved):
Thanks for this.

The way I see it now based on this and what I'm currently working on is to have these generated from the "stream" dialect which seems to require a corresponding `snitch` `frep` operation.
Not exactly what goes where precisely ATM, but I agree we can refine it later.

xdsl/dialects/riscv.py
```diff
@@ -2711,6 +2717,168 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
...
+    # Repeat three times, stagger 2, period 2
+    li a0, 3
+    frep.inner a0, 2, 2, 0b0100
+    fadd.d fa0, ft0, ft2
+    fmul.d fa0, ft3, fa0
+    ```
+
+    is equivalent to:
+    ```
+    fadd.d fa0, ft0, ft2
```
JosseVanDelm, line 2867: I'm not sure I fully understand the stagger option yet, but this pattern doesn't seem to be correct?
```suggestion
    fadd.d fa0, ft0, ft3
```

tests/filecheck/dialects/riscv/riscv_assembly_emission.mlir
```diff
@@ -174,6 +174,21 @@
...
 
 
+    // RISC-V Extensions
+
+    riscv.frep_outer %0, 0, 0 ({
+      %add_o = riscv.add %0, %1 : (!riscv.reg<zero>, !riscv.reg<j1>) -> !riscv.reg<j2>
+    }) : (!riscv.reg<zero>) -> ()
+
+    // CHECK:          frep.outer zero, 1, 0, 0
+    // CHECK-NEXT:     add  j2, zero, j1
```
JosseVanDelm, line 184: I don't think this is valid, because the `xfrep` is for floating point instructions only IIUC.
The FPU and the integer ALU are fully decoupled in snitch, and there's only a FPU sequence buffer, no Integer instruction sequence buffer IIUC.
Maybe we should make sure non-floating point instructions don't pass verification?

JosseVanDelm (commented):
Thanks for adding those docs @superlopuh! This looks much better!
I still have an extra suggestion though, IMO the frep block should verify against non floating-point instructions inside it's region. I don't think putting integer operations inside makes a lot of sense, since they are not affected by frep sequencing logic IIUC? What do you think?

tests/filecheck/dialects/riscv/riscv_assembly_emission.mlir
compor: That escaped me as well, the "f" in `frep` stands for floating-point as I've seen it in the paper and I think Verilator complained when I had tried an int instr

https://github.com/pulp-platform/snitch/blob/d026f47843f0ea6c269244c4e6851e0e09141ec3/hw/ip/snitch/src/snitch.sv#L1972C25-L1972C25

xdsl/dialects/riscv.py
compor: I think the mask is for `rs1` only, no?

nazavode (approved):

xdsl/dialects/riscv.py
superlopuh: I copied this from the paper verbatim but I think it's actually wrong there. My instinct is to stick with this until we make it work ourselves and update the doc.

tests/filecheck/dialects/riscv/riscv_assembly_emission.mlir
superlopuh: !! I thought it was "Fixed". We don't currently have a way to check for fp instructions generally, we'll have to make it a set of classes. In any case I'd propose to do this in a downstream PR if that's ok?

JosseVanDelm: Sure @superlopuh !

PR #1589, opened 2023-09-16
dialects: (maths) Using `__init__()` instead of `get()`

ShaolunWang:
Using `__init__()` instead of `get()` for experimental dialect `maths`. This follows the previous changes on dialect. It will also make the tests for the dialect easier to write.

### DIFF
--- a/xdsl/dialects/experimental/math.py
+++ b/xdsl/dialects/experimental/math.py
@@ -31,12 +31,11 @@ class AbsFOp(IRDLOperation):
     operand: Operand = operand_def(AnyFloat)
     result: OpResult = result_def(AnyFloat)
 
-    @staticmethod
-    def get(
-        operand: Operation | SSAValue, fastmath: FastMathFlagsAttr | None = None
-    ) -> AbsFOp:
+    def __init__(
+        self, operand: Operation | SSAValue, fastmath: FastMathFlagsAttr | None = None
+    ):
         operand = SSAValue.get(operand)
-        return AbsFOp.build(
+        return super().__init__(
             attributes={"fastmath": fastmath},
             operands=[operand],
             result_types=[operand.type],
@@ -60,10 +59,9 @@ class AbsIOp(IRDLOperation):
     operand: Operand = operand_def(IntegerType)
     result: OpResult = result_def(IntegerType)
 
-    @staticmethod
-    def get(operand: Operation | SSAValue) -> AbsIOp:
+    def __init__(self, operand: Operation | SSAValue):
         operand = SSAValue.get(operand)
-        return AbsIOp.build(operands=[operand], result_types=[operand.type])
+        return super().__init__(operands=[operand], result_types=[operand.type])
 
 
 @irdl_op_definition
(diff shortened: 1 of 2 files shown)


xdsl/dialects/experimental/math.py
```diff
@@ -95,6 +93,20 @@ class Atan2Op(IRDLOperation):
     rhs: Operand = operand_def(AnyFloat)
     result: OpResult = result_def(AnyFloat)
 
+    # TODO: use __init__
```
superlopuh, line 96: ?

xdsl/dialects/experimental/math.py
```diff
@@ -95,6 +93,20 @@ class Atan2Op(IRDLOperation):
...
+    def __init__(
+        self,
+        lhs: Operation | SSAValue,
+        rhs: Operation | SSAValue,
+        fastmath: FastMathFlagsAttr | None = None,
+    ):
+        lhs = SSAValue.get(lhs)
```
superlopuh, line 103: this isn't necessary `IRDLOperation.__init__` accepts `Operation | SSAValue`

superlopuh (commented):
Thank you for starting this! Can you please convert the rest of the TODOs?

compor (commented):
Updating the rest of `__init__`s would be great. Thanks

xdsl/dialects/experimental/math.py
ShaolunWang: Sorry I forgot to remove the todo after changing that

ShaolunWang:
I just checked my pr and for some reason my script to update this wasn't working on all of them, I will fix them asap.

xdsl/dialects/experimental/math.py
ShaolunWang: Currently I'm doing     
```
return super().__init__(
            attributes=attributes,
            operands=[lhs, rhs],
            result_types=[SSAValue.get(lhs).type],
        )
```
Which is what's currently been used in gpu dialect. I'm not sure whether this is the best way.

ShaolunWang:
> script
Technically it's a set of vim macros that runs on a single file, which can be scripted

ShaolunWang:
Yes, I'm happy to help 👍

webmiche:
(Or upstreaming the macros into some new folder, if that is easy enough).

xdsl/dialects/experimental/math.py
```diff
@@ -589,19 +570,19 @@ class FmaOp(IRDLOperation):
...
         b: Operation | SSAValue,
         c: Operation | SSAValue,
         fastmath: FastMathFlagsAttr | None = None,
-    ) -> FmaOp:
+    ):
         attributes = {"fastmath": fastmath}
 
         a = SSAValue.get(a)
         b = SSAValue.get(b)
         c = SSAValue.get(c)
```
webmiche, line 584: You can remove these two, right?

webmiche: There are even more of those still left.

webmiche (approved):
Minor comments, otherwise LGTM

xdsl/dialects/experimental/math.py
webmiche: There are still 26 of these left.

compor (approved):

webmiche (approved):

PR #1592, opened 2023-09-17
core: Add support for properties in attribute converter

math-fehr:
This adds support for properties in the attribute/type converter.
Note that this also adds properties to the `TestOp` classes. They are required unlike attributes because only defined propreties can be used.

### DIFF
--- a/tests/test_pattern_rewriter.py
+++ b/tests/test_pattern_rewriter.py
@@ -1110,3 +1110,121 @@ def test_type_conversion():
             walk_reverse=True,
         ),
     )
+
+
+def test_type_conversion_property():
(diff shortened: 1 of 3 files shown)


PapyChacal (approved):
Thanks 🙏
Just a concisness note, there were tests for the type conversion pattern already and you could just add properties on some test.op there, not adding more cases and lines;

webmiche (approved):

math-fehr:
> Thanks 🙏 Just a concisness note, there were tests for the type conversion pattern already and you could just add properties on some test.op there, not adding more cases and lines;

Thanks, I changed that ;)

PR #1599, opened 2023-09-19
transformations: add interepreter-based contant folding pass

superlopuh:
Here's a bit of a dopey idea: if we have a way to get the results of operations in Python, and all the inputs are constants, why not call the interpreter during the compilation process? This is done in [Singeli](https://github.com/mlochbaum/Singeli) a part of the BQN interpreter/compiler specialized for high-performance tensor computations. They recently discussed this approach on an episode of the Array Cast and I thought I'd try it out, turns out it took very little work. Would love to hear your feedback.


webmiche (commented):
I think that's a very cool idea!

I would argue that the downside of it is having to set up all the state needed for proper interpretation, that will be unnecessary here as you don't have variables/control flow, but I think in this Python-world we live in, and if we don't care about performance, this is a nice way to save code.

Maybe we could have an "interpreterlite" or something that only works for constants? This way we maybe could have the same code reuse without setting up all the state 🤔

compor:
I like it.
Maybe the constant folding example is not the most indicative of the possibilities (I think that's what @webmiche is pointing at), but I can see applications maybe in partial/lazy evaluation of kernels tailored to specific parameters.

superlopuh:
We could definitely get away with much less than the full interpreter object, I started making those changes but thought I'd concentrate on a PoC first before doing more work. If we decide that it's an interesting pass to merge and keep around it would be good to discuss what the API and expected behaviour might be.

math-fehr:
Looks like a really cool idea yeah!
I would just add a new trait on the operations that can be folded with that technique.
Maybe just the `Pure` trait is enough? But otherwise for non-deterministic operations, this pass wouldn't work I think

superlopuh:
I think Pure + ConstantLike will get us a lot of mileage already. I think the biggest concern is things like which constant op to use to instantiate the value returned by the computation. The constant folding in Toy would expect a `toy.constant` op, so it seems like it would make sense to parametrize the pass in some way to allow for that behaviour.

superlopuh:
I think this is ready to merge as-is, and can be enhanced later, so would like to put it up for review. Are there any more tests that should be added? Anything I missed?

superlopuh:
@math-fehr @webmiche @georgebisbas @PapyChacal @AntonLydike

webmiche (approved):
I think this is a cool idea, so let's get it in!

I am not fully up to speed with all the details anymore, AFAIU, this only applies a single iteration right? So (1 + 2) + 3 would not become 6, right?

superlopuh:
Excellent question, I have no idea, let me add a test for that

superlopuh:
It works, updated the test

webmiche:
Okay, I guess we have fix-point as a default that, did not remember this properly then :+1:

PR #1597, opened 2023-09-18
dialects: add FSM dialect (WIP - refer to #1600)

luisacicolini:
Added the FSM dialect, currently working on file check testing.

PR #1600, opened 2023-09-20
dialects: FSM base operators 

luisacicolini:
Added basic operators, currently working on file check tests integration.


compor:
I think this should target (i.e., be stacked on top of) https://github.com/xdslproject/xdsl/pull/1597. Right?

compor (changes requested):
I'd say a first major improvement would be to increase the test coverage on these operations.
In other words, I think you need to write tests that exercise/trigger all those conditions in the custom verification methods (i.e., `verify_`).

I'd suggest having a look in `test_scf.py` on a way to go about it (you will need to add a `test_fsm.py` as well and proceed from there).

tests/filecheck/dialects/fsm/fsm_op.mlir
superlopuh, line 1: this file appears not to be a test, can you please add the RUN etc?

superlopuh (changes requested):

superlopuh:
One thing that would be great is an MLIR test file with filecheck tests for the error messages. `tests/filecheck/dialects/riscv/frep_verification.mlir` is an example. Would be great to have a test for each VerificationException. I'd also be happy with an issue + pinky promise that this will be done in a future PR.

tests/filecheck/dialects/fsm/fsm_op.mlir
```diff
@@ -143,33 +143,4 @@
...
-      "fsm.output"(%arg0) : (i1) -> ()
-    }, {
-      "fsm.transition"() ({
-      }, {
-      }) {nextState = @B} : () -> ()
-    }) {sym_name = "B"} : () -> ()
-  }) {function_type = (i1) -> i1, initialState = "A", sym_name = "foo"} : () -> ()
-}) : () -> ()
-
+}) : () -> ()
```
superlopuh, line 146:
```suggestion
}) : () -> ()

```
I'd recommend ticking the setting in vscode that adds trailing whitespace. I think you might also need to run `make precommit-install` or `make install-precommit`, I forget which one, on your machine

xdsl/tools/command_line_tool.py
```diff
@@ -21,6 +21,7 @@
 from xdsl.dialects.experimental.fir import FIR
 from xdsl.dialects.experimental.hls import HLS
 from xdsl.dialects.experimental.math import Math
+from xdsl.dialects.fsm import FSM
```
JosseVanDelm, line 24: wondering whether it makes sense to make it like this for clarity, but not sure if necessary:
```suggestion
from xdsl.dialects.circt.fsm import FSM
```

tests/filecheck/dialects/fsm/fsm_op.mlir
```diff
@@ -0,0 +1,146 @@
...
+// CHECK:      "fsm.transition"() ({
+// CHECK:        %1 = "arith.constant"() {value = 0 : i16} : () -> i16
+// CHECK:        %2 = "arith.cmpi"(%0, %1) {predicate = 0 : i64} : (i16, i16) -> i1
+// CHECK:        "fsm.return"(%2) : (i1) -> ()
+// CHECK:      }, {
+// CHECK:      }) {nextState = @IDLE} : () -> ()
+// CHECK:    }) {sym_name = "BUSY"} : () -> ()
+// CHECK:  }) {function_type = (i1) -> i1, initialState = "IDLE", sym_name = "foo"} : () -> ()
+
+
```
JosseVanDelm, line 55: Maybe this is my ignorance, but can you maybe describe in a short sentence what this FSM does?

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+    HasParent,
+    IsTerminator,
+    NoTerminator,
+    SymbolOpInterface,
+    SymbolTable,
+)
+from xdsl.utils.exceptions import VerifyException
+
+signlessIntegerLike = ContainerOf(AnyOf([IntegerType, IndexType]))
+
```
JosseVanDelm, line 46: Maybe it makes sense make clear within this file that you are mirroring the CIRCT FSM ops, so people know where to look for additional documentation, since this is not part of MLIR core?

JosseVanDelm (commented):
Wow, this is quite the dialect.

Nice PR.

Just wondering a few things:
* IIUC The regions don't seem to be created as optional regions? If that is not supported by IRDL right now then maybe you should add a comment saying that it is not supported right now.
*  For documentation purposes, it might not be clear for some people that this is a CIRCT dialect, so maybe you should add this somewhere in the file?
* (As I have no experience with this dialect) I spent quite some time reading through the stuff you made to decode what type of FSM was actually represented, maybe a small comment explaining this would be nice?

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+    of the next state. This op includes an optional `$guard` region with an `fsm.return`
+    as terminator that returns a Boolean value indicating the guard condition of
+    this transition. This op also includes an optional `$action` region that represents
+    the actions to be executed when this transition is taken"""
+
+    name = "fsm.transition"
+
+    guard = region_def()
+
+    action = region_def()
```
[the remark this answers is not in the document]
luisacicolini: As far as I understood talking with more experienced people one such things does not really exists! What we do in these cases is basically having a region and keeping it empty, as a "non region"

tests/filecheck/dialects/fsm/fsm_op.mlir
luisacicolini: sure! great suggestion actually, I will do it

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+from xdsl.traits import (
+    HasParent,
+    IsTerminator,
+    NoTerminator,
+    SymbolOpInterface,
+    SymbolTable,
+)
+from xdsl.utils.exceptions import VerifyException
+
+signlessIntegerLike = ContainerOf(AnyOf([IntegerType, IndexType]))
```
webmiche, line 45: We already have this [here](https://github.com/xdslproject/xdsl/blob/main/xdsl/dialects/arith.py#L40). IMO, it makes more sense to import that than redefining it. What do others think? @math-fehr @superlopuh

xdsl/dialects/fsm.py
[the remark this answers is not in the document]
webmiche: Note that (I think) optional regions were not available in MLIR at some point, so people encoded them as "empty" regions. I think we encountered this when porting the scf dialect for the first time or something. People probably never went back and actually made the regions optional, since that influences quite a lot of the repo, so there are these weird left overs at times.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+        if (
+            self.arg_attrs is not None
+            and self.arg_names is None
+            or self.res_attrs is not None
+            and self.res_names is None
+            or self.arg_attrs is None
+            and self.arg_names is not None
+            or self.res_attrs is None
+            and self.res_names is not None
+        ):
```
webmiche, line 157: It feels like this could be implemented in a way that is more understandable. I am thinking of some sort of list of tuples and then check that for all tuples, either both are None or neither is None. At the very least, please add a comment saying that "if {arg, res}_names is given, {arg, res}_attrs is needed as well" or whatever.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+signlessIntegerLike = ContainerOf(AnyOf([IntegerType, IndexType]))
+
+
+@irdl_op_definition
+class Transition(IRDLOperation):
+    """Represents a transition of a state with a symbol reference
+    of the next state. This op includes an optional `$guard` region with an `fsm.return`
+    as terminator that returns a Boolean value indicating the guard condition of
+    this transition. This op also includes an optional `$action` region that represents
+    the actions to be executed when this transition is taken"""
```
webmiche, line 54:
```suggestion
    """
    Represents a transition of a state with a symbol reference
    of the next state. This op includes an optional `$guard` region with an `fsm.return`
    as terminator that returns a Boolean value indicating the guard condition of
    this transition. This op also includes an optional `$action` region that represents
    the actions to be executed when this transition is taken.
    """
```

Please write block comments like this :) (with a newline after `"""`)

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+            transitions = Region(Block())
+        super().__init__(
+            attributes=attributes,
+            regions=[output, transitions],
+        )
+
+    def verify_(self):
+        parent = self.parent_op()
+
+        while parent is not None:
```
webmiche, line 253: Can there be MachineOps nested inside other MachineOps? If not, you would not need to check all parents, as you are doing here :)

webmiche (commented):
Couple of minor comments on style and stuff :)

I have no idea how this dialect actually works in circt, so I just commented on the things I understand :wink: 

Did you actually try to run the testfiles through circt at some point? I think that would be a valuable test to understand if interoperation actually works properly.

luisacicolini:
> 

Thanks for the suggestion, I have not tried one such thing yet but will do for sure :)

luisacicolini:
> Couple of minor comments on style and stuff :)
> 
> I have no idea how this dialect actually works in circt, so I just commented on the things I understand 😉
> 
> Did you actually try to run the testfiles through circt at some point? I think that would be a valuable test to understand if interoperation actually works properly.

Haven't done that yet but looks absolutely necessary :)

xdsl/tools/command_line_tool.py
compor: Good point, avoids moving/reorganizing things later.

tests/filecheck/dialects/fsm/fsm_op.mlir
```diff
@@ -0,0 +1,146 @@
+// RUN XDSL_ROUNDTRIP
+
+// CHECK:  "fsm.machine"() ({
+// CHECK:  ^bb0(%arg0: i1):
+// CHECK:    "fsm.state"() ({
+// CHECK:    }, {
+// CHECK:    }) {sym_name = "IDLE"} : () -> ()
+// CHECK:  }) {function_type = (i1) -> (), initialState = "IDLE", sym_name = "foo"} : () -> ()
+
+"builtin.module"() ({
```
compor, line 10: You can avoid an indentation level by omitting `builtin.module` if you want; it gets added automatically.

compor (changes requested):

tests/filecheck/dialects/fsm/fsm_op.mlir
superlopuh: You marked this comment as resolved, but I still see the (-) in the diff. I would recommend marking comments as resolved when they have been addressed, this makes it easier to track which changes are still expected.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+            action = Region(Block())
+        if not isinstance(guard, Region):
+            guard = Region(Block())
+        super().__init__(
+            attributes=attributes,
+            regions=[guard, action],
+        )
+
+    def verify_(self):
+        if SymbolTable.lookup_symbol(self, self.nextState) is None or not isinstance(
```
math-fehr, line 87: I would probably create a function for `SymbolTable.lookup_symbol(self, self.nextState)` (since it's going to be useful later on), and I would make this a variable, so you don't lookup the symbol twice.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+            SymbolTable.lookup_symbol(self, self.nextState), StateOp
+        ):
+            raise VerifyException("Can not find next state")
+        if self.guard.blocks and not isinstance(self.guard.block.last_op, ReturnOp):
+            raise VerifyException("Guard region must terminate with ReturnOp")
+        var = self.parent_op()
+        assert isinstance(var, StateOp)
```
math-fehr, line 94: I would either make this an `if` with a `VerifyException`, or I would add the `HasParent([StateOp])` trait to this operation.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+    Represents the outputs of a machine under a specific state. The
+    types of `$operands` should be consistent with the output types of the state
+    machine
+    """
+
+    name = "fsm.output"
+
+    operand = var_operand_def(AnyAttr())
+
+    traits = frozenset([IsTerminator()])
```
math-fehr, line 183: We should add the `HasParent([StateOp])`

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+    def __init__(
+        self,
+        operand: Sequence[SSAValue | Operation],
+    ):
+        super().__init__(
+            operands=[operand],
+        )
+
+    def verify_(self):
+        parent = self.parent_op()
```
math-fehr, line 194: I would add an `assert(isinstance(parent, StateOp))` here, since the operation has the `HasParent` trait (which is checked before the `verify_` method.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+
+    name = "fsm.state"
+
+    output = region_def()
+
+    transitions = region_def()
+
+    sym_name = attr_def(StringAttr)
+
+    traits = frozenset([NoTerminator(), SymbolOpInterface()])
```
math-fehr, line 231: I think you want to add `HasParent([Machine])`

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+        if not isinstance(transitions, Region):
+            transitions = Region(Block())
+        super().__init__(
+            attributes=attributes,
+            regions=[output, transitions],
+        )
+
+    def verify_(self):
+        parent = self.parent_op()
+
```
math-fehr, line 253: I would just add an `assert(isinstance(parent, Machine))`, and remove the `while` loop.
It seems that even though they have the while loop in MLIR (through a function call), this is useless, since the parent is always the `Machine`.

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+        while parent is not None:
+            if (
+                isinstance(parent, MachineOp)
+                and getattr(parent, "res_attrs") is not None
```
math-fehr, line 257:
```suggestion
                and parent.res_attrs is not None
```

math-fehr: And same below

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+    ):
+        super().__init__(
+            operands=[variable, value],
+        )
+
+    def verify_(self) -> None:
+        if not isinstance(self.variable.owner, VariableOp):
+            raise VerifyException("Destination is not a variable operation")
+
+        parent = self.parent_op()
```
math-fehr, line 296: Here same, I would `assert(isinstance(parent, TransitionOp))`

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,379 @@
...
+        while parent is not None:
+            if isinstance(parent, TransitionOp):
+                # walk through the action region
+                found = 0
+                for op in parent.action.walk():
```
math-fehr, line 301:
```suggestion
                for op in parent.action.ops:
```
So it doesn't recursively go into regions (since there are none)

math-fehr (approved):
Nice!
I added some style comments, and some things that are missing, but otherwise that looks really good!

tests/filecheck/dialects/fsm/fsm_op.mlir
superlopuh: Maybe on my end it was just a vscode setting for a while, not precommit?

superlopuh: I still feel like it would be good to add it manually until we amend the precommit hook

xdsl/dialects/fsm.py
compor: The later is better IMHO.

xdsl/dialects/fsm.py
compor: I had a look at the rationale document for FSM and they say it was inspired by UML state machines.
AFAIR, UML has the notion of nesting in state machines, so maybe we need to verify this with a basic nesting example in CIRCT.

Regarding the second point, `StateOp` needs to have a `MachineOp` as a parent since it specifies the `HasParent` trait. The doc mentions only the trait without mention to the operation that needs to fulfill this relationship, but looking at the code we have:

```
def StateOp : FSMOp<"state", [HasParent<"MachineOp">, Symbol, NoTerminator]> {
```

luisacicolini: Thanks a lot! super useful suggestions

xdsl/dialects/fsm.py
luisacicolini: For now I had to leave this one like that because it would raise a "single-block" related exception with the ops operator. maybe we can take a closer look later!

xdsl/dialects/fsm.py
webmiche: I am confused, why is this back? :thinking:

xdsl/dialects/fsm.py
```diff
@@ -0,0 +1,377 @@
...
+        sym_name: str,
+        initial_state: str,
+        function_type: FunctionType | tuple[Sequence[Attribute], Sequence[Attribute]],
+        arg_attrs: ArrayAttr[DictionaryAttr] | None,
+        res_attrs: ArrayAttr[DictionaryAttr] | None,
+        arg_names: ArrayAttr[StringAttr] | None,
+        res_names: ArrayAttr[StringAttr] | None,
+        body: Region | type[Region.DEFAULT] = Region.DEFAULT,
+    ):
+        attributes: dict[str, Attribute | None] = {}
```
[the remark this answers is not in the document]
webmiche: Same here.

xdsl/dialects/fsm.py
webmiche: This happens because the `action` region of the `TransitionOp` is not a SingleBlockRegion. I am not sure about how FSM works internally, but it feels like a lot of these regions might actually be SingleBlockRegions, which would be very useful to specify. If this is not done in the circt dialect, I would not do it here either though :)

luisacicolini: Thanks, I'll double check that :)

xdsl/dialects/fsm.py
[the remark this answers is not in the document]
webmiche: (I think this has been addressed, but doesn't seem to be in the current code)

xdsl/dialects/fsm.py
luisacicolini: Oooh sorry, I tried to make the two PRs more similar as I made some major modifications to more branches yesterday and forgot about some of them! fixing it rn

JosseVanDelm (commented):
Hi @luisacicolini, is it possible you haven't addressed my comments yet? Apart from that this looks great to me and the discussion here is learning me a lot of stuff! Thanks!

luisacicolini:
> Hi @luisacicolini, is it possible you haven't addressed my comments yet? Apart from that this looks great to me and the discussion here is learning me a lot of stuff! Thanks!

Hey! So sorry, I made some major modifications and now the documentation should be okay. For FSMs comments I am waiting a bit because I myself need to understand some things better before explaining them! But that would be definitely super useful so I will do it asap, thanks a lot :9

xdsl/dialects/fsm.py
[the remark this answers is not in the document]
luisacicolini: Should be updated now!

tests/filecheck/dialects/fsm/fsm_op.mlir
compor: @luisacicolini can you please address Josse's feedback with a very brief comment please?

compor (approved):
LGTM minus the commenting of the tests (maybe) and merging in `main` before we merge :-)

PR #1601, opened 2023-09-21
core: add ConstantLike trait from MLIR

superlopuh:


compor (approved):

webmiche (commented):
What is this used for? Could we somehow test it's functionality?

compor:
> I'm not exactly sure how to test it. The functionality is in the `has_trait` method etc, it's like Pure, just a marker for when an operation is somehow like a constant.

I had a look in MLIR and it seems to be a sort of "axiomatic" trait; there is no direct testing of it.
Only indirectly via its uses (e.g., there is a `PureOp` in `test/lib/Dialect/Test/TestOps.td` that is used to check if the operation is always speculatable or not.

webmiche:
Hmm okay. I guess all the lines are covered so I am fine with merging :thinking: 
Must say though, I don't get how this is useful. I guess "constant-like" in this case means "introduces values out of thin air", which I guess might be a useful trait in some case (allocation, memory, something something).

Anyway, let's get it in. If MLIR has it, we can have it too.

PR #1603, opened 2023-09-25
core: Implement AttrSizedSegments option to implement as a property.

PapyChacal:
Extracted (and cleaned up) from the MLIR bump PR, this allows to make "operand_segment_sizes" and friends implemented as property on the Op rather than an attribute.

### DIFF
--- a/xdsl/irdl/irdl.py
+++ b/xdsl/irdl/irdl.py
@@ -548,7 +548,7 @@ class IRDLOption(ABC):
 
 
 @dataclass
-class AttrSizedSegments(IRDLOption):
+class AttrSizedSegments(IRDLOption, ABC):
     """
     Expect an attribute on the operation that contains the segment sizes of the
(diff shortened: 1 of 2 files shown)


xdsl/irdl/irdl.py
```diff
@@ -560,6 +560,7 @@ class AttrSizedSegments(IRDLOption):
     """
 
     attribute_name: ClassVar[str]
+    as_property: bool = False
```
webmiche, line 563: As long as we switch this to default True once the PR is in and some time has passed, I am fine with this :+1:

PapyChacal: If it's dominantly used with as_property=True then, I'm for it too :+1:

PapyChacal: Good point, opened #1606 :slightly_smiling_face:

PR #1582, opened 2023-09-14
misc: Bump MLIR to a3f2751

PapyChacal:
Update MLIR to a3f2751f782f3cdc6ba4790488ec20163a40ac37, as selected on the greenlit list and using properties!

Opening as a draft to get CI running and start looking at it as a PR, seeing if we want to extract some changes in separate PRs.

Things to extract out in individual PRs:

- [x] Type conversion on properties
- [x] Immutable IR with properties
- [x] operand_segment_size as a property opt-in

PR #1605, opened 2023-09-25
testing: remove superfluous test in memref

PapyChacal:
Extracted from #1582.

There is a test in the memref test suite testing the generic constructor, which is already tested in itself and has nothing specifically related to memref. Can we just remove it? It requires updating for the MLIR update anyway.


compor:
No change in coverage and `subview` seems to be tested further down again. LGMT

webmiche (approved):

AntonLydike (approved):
Nice! Thanks!

PR #1611, opened 2023-09-26
core: test and fix IRDLOptions as properties

PapyChacal:
I forgot something while extracting this from #1582 in #1603; This is fixed and tested here.

Basically, by switching some helpers to use the IRDLOptions types rather than placeholder instances, I forgot to actually get the option instance when using this option in the irdl generated builder, causing this to always fall back to the attribute-based behaviour and causing errors in the property-expecting case (hence no regression appearing with no dialect ops using properties yet). Fixed :wrench: 

this is also the last core-preparing PR before #1582 :tada: Opening a Zulip thread in a moment to discuss that.

### DIFF
--- a/tests/test_operation_builder.py
+++ b/tests/test_operation_builder.py
@@ -262,6 +262,16 @@ class TwoVarOperandOp(IRDLOperation):
     irdl_options = [AttrSizedOperandSegments()]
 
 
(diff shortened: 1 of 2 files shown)


webmiche (approved):

math-fehr (approved):
Really nice!

PR #1613, opened 2023-09-26
Switch AttrSizedSegments attribute_names to CamelCase.

### LINKED ISSUE #1612: irdl: Make segment sizes name consistent
Currently, we are using snake_case for xDSL-only attributes (like `region_segment_sizes`).
However, MLIR use camlCase convention, so we should follow that.

PapyChacal:
Companion (but independent) PR to #1582, resolves #1612.


math-fehr:
Perfect!
I'm actually surprised this actually work without changing the tests, but I think this is because I wrote the tests using the `.attribute_name` accessor?

math-fehr (approved):

webmiche (approved):
LGTM, yeah it seems weird that this is already sufficient :thinking:

compor (approved):

PapyChacal:
> Perfect! I'm actually surprised this actually work without changing the tests, but I think this is because I wrote the tests using the `.attribute_name` accessor?

Indeed!

PR #1607, opened 2023-09-25
[WIP] dialects: (scf) Allow implicit terminator (`yield` op) in `scf` operations

compor:
This PR adds:

- Automatic generation of the `scf.yield` terminator in `scf.for` operations that have no loop-carried variables via the `ensure_terminator` method (only for custom printing/parsing).
- Skipping printing of `scf.yield` for `scf.for` operations if it has no loop-carried variables (only for custom printing/parsing).
- Test case for the above (triple `scf.for` loop nest taken from MLIR test suite).

TODO:
- [ ] Open separate PR with corresponding test cases for the `Printer` changes.
- [ ] Apply the above changes to all `scf` operations that require an `scf.yield`.

Once these are done I'll update and remove draft status of this PR, I just wanted to have some eyes on it first.

PR #1617, opened 2023-09-28
core: (printer) Allow optional printing of block terminator operations

compor:
This PR is part of #1607 and adds:

- Ability to optionally skip printing a terminator operation
- Tests of the above

When this is to happen is left on the specific dialect; the immediate application of this is used in custom printing/parsing in order to avoid printing implicit terminators (this is part of #1607).

### DIFF
--- a/tests/test_printer.py
+++ b/tests/test_printer.py
@@ -5,15 +5,14 @@ from io import StringIO
 import pytest
 from conftest import assert_print_op
 
+from xdsl.dialects import test
 from xdsl.dialects.arith import Addi, Arith, Constant
 from xdsl.dialects.builtin import Builtin, IntAttr, IntegerType, UnitAttr, i32
 from xdsl.dialects.func import Func
-from xdsl.dialects.test import Test, TestOp
 from xdsl.ir import (
     Attribute,
     Block,
     MLContext,
-    Operation,
     OpResult,
     ParametrizedAttribute,
     Region,
@@ -56,9 +55,9 @@ def test_simple_forgotten_op():
 def test_print_op_location():
     """Test that an op can be printed with its location."""
     ctx = MLContext()
-    ctx.register_dialect(Test)
+    ctx.register_dialect(test.Test)
 
-    add = TestOp(result_types=[i32])
+    add = test.TestOp(result_types=[i32])
 
     add.verify()
 
@@ -373,7 +372,7 @@ def test_print_block_argument_location():
 def test_print_block():
     """Print a block."""
     block = Block(arg_types=[i32, i32])
-    block.add_op(TestOp(operands=(block.args[1],)))
+    block.add_op(test.TestOp(operands=(block.args[1],)))
 
     # Print block arguments inside the block
     io = StringIO()
@@ -387,7 +386,7 @@ def test_print_block():
 def test_print_block_without_arguments():
     """Print a block and its arguments separately."""
     block = Block(arg_types=[i32, i32])
-    block.add_op(TestOp(operands=(block.args[1],)))
+    block.add_op(test.TestOp(operands=(block.args[1],)))
 
     # Print block arguments separately from the block
     io = StringIO()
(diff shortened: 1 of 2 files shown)


tests/test_printer.py
```diff
@@ -399,10 +398,41 @@ def test_print_block_without_arguments():
...
+    """Print a block and with its terminator."""
+    block = Block(ops=[test.TestOp.create(), test.TestTermOp.create()])
+
+    # Print block ops including block terminator
+    io = StringIO()
+    p = Printer(stream=io)
+    p.print_block(block, print_block_terminator=True)
+    assert (
+        io.getvalue()
+        == """\n^0:\n  "test.op"() : () -> ()\n  "test.termop"() : () -> ()"""
```
superlopuh, line 411:
```suggestion
        == """
        ^0:
          "test.op"() : () -> ()
          "test.termop"() : () -> ()
        """
```
nit but feels like it's a little more readable with the inline newlines

xdsl/printer.py
```diff
@@ -254,8 +261,9 @@ def print_block(self, block: Block, print_block_args: bool = True) -> None:
 
         self._indent += 1
         for op in block.ops:
-            self._print_new_line()
-            self.print_op(op)
+            if not (op.has_trait(IsTerminator) and not print_block_terminator):
+                self._print_new_line()
+                self.print_op(op)
```
superlopuh, line 266: nit:
the multiple negatives is a bit confusing, how about something like this:
```suggestion
            if not print_block_terminator and op.has_trait(IsTerminator):
                # Do not print terminator if `print_block_terminator` is False
                continue
            self._print_new_line()
            self.print_op(op)
```

superlopuh: Also reorders the checks to have the cheaper one first

superlopuh (approved):

tests/test_printer.py
```diff
@@ -399,10 +398,44 @@ def test_print_block_without_arguments():
...
+
+    # Print block ops separately from the block terminator
+    io = StringIO()
+    p = Printer(stream=io)
+    p.print_block(block, print_block_terminator=False)
+    p.print("\n  ")
+    p.print_op(term_op)
+    assert (
+        io.getvalue()
+        == """\n^0:\n  "test.op"() : () -> ()\n  "test.termop"() : () -> ()"""
```
AntonLydike, line 431: Can you make this into a multi line string as well?

AntonLydike: Also, I think you could remove lines 427 and 428 and trim this test a bit to make it clearer what you are actually testing?

compor: I'm trying to print the block without its terminator and the terminator separately, so I have to add the newlines manually; Not sure how else to express this, it is similar to the block args tests here.

AntonLydike: Why do you need to print the terminator separately for the test? Isn't it enough to show that the terminator is not printed when you set the flag?

compor: We don't actually, I was just copying what was happening with the block args tests in this test file, in case I was missing anything.
I'll remove it.

math-fehr (approved):
Nice!
I have to say though, I'm not a big fan of hiding operations like this, even though MLIR does it.
I do understand that we want to handle this on the parsing side, so that we eventually can parse custom MLIR, but I have to say that adding this on the printing side might make things extra confusing (the same way MLIR makes it confusing).
While we can follow what MLIR does (in this PR), we could also try to make something more simple, and less confusing.
Just my two cents, if people think this is worth it, I'm fine merging this, I just want to raise this question ;)

compor:
> Nice! I have to say though, I'm not a big fan of hiding operations like this, even though MLIR does it. I do understand that we want to handle this on the parsing side, so that we eventually can parse custom MLIR, but I have to say that adding this on the printing side might make things extra confusing (the same way MLIR makes it confusing). While we can follow what MLIR does (in this PR), we could also try to make something more simple, and less confusing. Just my two cents, if people think this is worth it, I'm fine merging this, I just want to raise this question ;)

I understand this point and I also find it counterintuitive that while there are traits for implicitly denoting a terminator which is used during custom parsing to ensure the terminator addition, the criteria for not printing that terminator are different (e.g., for the `yield` in `scf.for` is the existence or absence of loop-carried vars).

I like the custom parsing handling on this, but I'm not sure what a simpler approach would be.
Leaving the two "asymmetric" maybe?

superlopuh:
IDK, I think the whole custom parsing and printing is already a symptom of something broken, we might as well lean into the patch over the uglyness of the original syntax while we're at it. And as we know, in printing and parsing, compatibility with MLIR is king.

compor:
Maybe we should open an issue tracking at least the idea of this for potential further exploration.
I feel these discussions are helpful, but they are lost (regardless of whether patch lands or not).

compor:
@AntonLydike if you're happy with this, I'll merge it in.

PR #1620, opened 2023-09-30
misc: (makefile) Fix shell invocation in Makefile for venv creation

compor:
This PR replaces the `source` invocation from the `Makefile` when building a virtual env by using the dot operator.

This fails because `make` seems to search the `PATH` for a program called `source`.
The way to invoke a shell script and run its contents in the current shell within a `Makefile` is by using `.` operator (which is also POSIX-compliant).

I saw this fail on @dshaaban01's MacBook last week and I've been able to replicate it on my Linux box.
Not sure how it was missed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 # set up the venv with all dependencies for development
 venv: requirements-optional.txt requirements.txt
 	python3 -m venv ${VENV_DIR}
-	source ${VENV_DIR}/bin/activate
+	. ${VENV_DIR}/bin/activate
 	python3 -m pip --require-virtualenv install -r requirements-optional.txt -r requirements.txt
 	python3 -m pip --require-virtualenv install -e ".[extras]"
 


superlopuh:
Does it work? I can test it on my computer but I'm aware that the behaviour is weird and different on different machines. The point of using `source` is to activate the shell for the current venv, without it the pip installs on the lines below might install globally

compor:
> Does it work? I can test it on my computer but I'm aware that the behaviour is weird and different on different machines. The point of using `source` is to activate the shell for the current venv, without it the pip installs on the lines below might install globally

It works on every new `venv` installation and with the `VENV_DIR` var which I prefer.

Well `source` is not POSIX-compliant but a `bash` introduction AFAIK, while the `.` operator is.
Hence, any shell that is POSIX-compliant should be OK with it.

As I wrote above the erroneous behaviour was both on Linux and MacOS, but at the end of the day that information doesn't matter much as it was on some version of `bash`.

superlopuh (approved):

compor:
I also found [this](https://stackoverflow.com/questions/44052093/makefile-with-source-get-error-no-such-file-or-directory) relevant SO thread.

AntonLydike (approved):
Awesome!

PR #1626, opened 2023-10-02
dialects: (riscv) Add double-precision FP arithmetic instructions

compor:
This PR adds:

- Double-precision FP arithmetic instructions for RISCV (`fadd.d`, `fsub.d`, `fmul.d`, `fdiv.d`).
- Emission tests for the above


superlopuh (approved):
Great, thank you!

PR #1627, opened 2023-10-02
[RFC] CI: Make ruff flag imports from forbidden modules.

PapyChacal:
RFC Because I'm not sure if we're interested in doing it this way?

We try in reviews to prevent imports from some modules (`xdsl.parser.X`, `xdsl.irdl.irdl` and `xdsl.ir.core`, to my knowledge. please tell me if I missed some.)
It feels to me like something we want a tool to check.

Ruff has a rule for it.

- CON: No auto-fix provided. it is the developer's responsibility to fix it.
- PRO: Well, it automatically finds them. Also allows for a clear to-the-point error message to not make the user scratch his/her head too much.

Also note that for those cases, it's always only about removing the last nested module from the from, it's really not cumbersome, for what it's worth. It then often leads to misformatted imports, but *those* are then auto-fixed.

WDYT?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ typeCheckingMode = "strict"
 profile = "black"
 
 [tool.ruff]
-select = ["E", "F", "W", "I", "UP", "PT"]
+select = ["E", "F", "W", "I", "UP", "PT", "TID251"]
 ignore = [
     "E741",  # https://beta.ruff.rs/docs/rules/ambiguous-variable-name/
     "PT006", # https://beta.ruff.rs/docs/rules/pytest-parametrize-names-wrong-type/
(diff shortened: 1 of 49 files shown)


webmiche:
That sounds like a great opportunity for such a tool, I like it!

webmiche:
What does the error message look like?

PapyChacal:
> What does the error message look like?
```
xdsl/parser/core.py:11:1: TID251 `xdsl.ir.core` is banned: Use xdsl.ir instead.
```
so the line is clickable on VSCode - and I guess most IDEs ? - to go directly to the line and erase the problematic bit ore copy paste the mentionned module. :slightly_smiling_face:

PapyChacal:
> > What does the error message look like?
> 
> ```
> xdsl/parser/core.py:11:1: TID251 `xdsl.ir.core` is banned: Use xdsl.ir instead.
> ```
> 
> so the line is clickable on VSCode - and I guess most IDEs ? - to go directly to the line and erase the problematic bit ore copy paste the mentionned module. 🙂

Nit: It could look better with backquotes on the advised module too?

superlopuh (approved):
👍👍👍

pyproject.toml
```diff
@@ -37,6 +37,14 @@ ignore = [
 line-length = 300
 target-version = "py310"
 
+[tool.ruff.flake8-tidy-imports.banned-api]
+"xdsl.parser.core".msg = "Use xdsl.parser instead."
+"xdsl.parser.attribute_parser".msg = "Use xdsl.parser instead."
+"xdsl.parser.affine_parser".msg = "Use xdsl.parser instead."
+"xdsl.ir.core".msg = "Use xdsl.ir instead."
+"xdsl.irdl.irdl".msg = "Use xdsl.irdl instead"
```
compor, line 45: This is great.

compor (approved):

PR #1628, opened 2023-10-02
interpreter: add mechanism to store runtime data in the interpreter

superlopuh:
This is necessary for some further interpreter changes. The interpreter function implementations are suposed to be stateless, and the riscv interpreter functions currently violate that, by caching the data on the functions object itself. This PR fixes that by introducing a new cache on the interpreter object, which is stateful already.


AntonLydike (approved):
Nice work! Really good to see something like that make it in!

PR #1633, opened 2023-10-02
backend: (riscv) use stack for float constants

superlopuh:
The current use of fcvt relies on an incorrect implementation in riscemu. This patch changes the arith.const lowering to mirror the way llvm and gcc do things, which is to load the bits into a register, store them to memory, and then load from memory again. There's no standard instruction in RISC-V to do this from register to register. Once this PR is merged, we can update riscemu version.


AntonLydike (commented):
If I understand correctly, this is not actually pushing/popping the stack but instead just loading/storing from `sp - 4`? If so, then this will not be able to handle two float values, right?

superlopuh:
What do you mean by two float values?

superlopuh:
Turns out I was wrong, there is an instruction to transfer the bits directly, it's used in the currently open riscemu update PR #1634

PR #1637, opened 2023-10-04
dialects: (builtin) Add bitwidth accessor to float types

compor:
This PR:

- Adds a `get_bitwdith` read-only API method on `FloatType`s similar to MLIR [here](https://github.com/llvm/llvm-project/blob/39d55321bd0b8ce436d6fcd8e5ba51b8bf535559/mlir/lib/IR/BuiltinTypes.cpp#L89)
- Tests of the above
- One use case of this API method

### DIFF
--- a/tests/filecheck/parser-printer/float_parsing.mlir
+++ b/tests/filecheck/parser-printer/float_parsing.mlir
@@ -11,12 +11,12 @@
   "test.op"() {"value" = 34.e0 : f32} : () -> ()
   // CHECK-NEXT: "test.op"() {"value" = 3.400000e+01 : f32} : () -> ()
 
-  "test.op"() {"value" = 34.e-23 : !f32} : () -> ()
+  "test.op"() {"value" = 34.e-23 : f32} : () -> ()
   // CHECK-NEXT: "test.op"() {"value" = 3.400000e-22 : f32} : () -> ()
 
   "test.op"() {"value" = 34.e12 : f32} : () -> ()
   // CHECK-NEXT: "test.op"() {"value" = 3.400000e+13 : f32} : () -> ()
 
-  "test.op"() {"value" = -34.e-12 : !f32} : () -> ()
+  "test.op"() {"value" = -34.e-12 : f32} : () -> ()
   // CHECK-NEXT: "test.op"() {"value" = -3.400000e-11 : f32} : () -> ()
 }) : () -> ()
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -62,7 +62,7 @@ def memref_shape_ops(
                 )
             bytes_per_element = element_type.width.data // 8
         case Float32Type():
-            bytes_per_element = 4
+            bytes_per_element = element_type.get_bitwidth // 8
         case _:
             raise DiagnosticException(
                 f"Unsupported memref element type for riscv lowering: {element_type}"
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -66,7 +66,7 @@ def memref_shape_ops(
         case Float32Type():
             bytes_per_element = element_type.get_bitwidth // 8
         case Float64Type():
-            bytes_per_element = 8
+            bytes_per_element = element_type.get_bitwidth // 8
         case _:
             raise DiagnosticException(
                 f"Unsupported memref element type for riscv lowering: {element_type}"
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -64,9 +64,9 @@ def memref_shape_ops(
                 )
             bytes_per_element = element_type.width.data // 8
         case Float32Type():
-            bytes_per_element = element_type.get_bitwidth // 8
+            bytes_per_element = element_type.width.data // 8
         case Float64Type():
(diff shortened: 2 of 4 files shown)


tests/filecheck/parser-printer/float_parsing.mlir
```diff
@@ -11,12 +11,12 @@
   "test.op"() {"value" = 34.e0 : f32} : () -> ()
   // CHECK-NEXT: "test.op"() {"value" = 3.400000e+01 : f32} : () -> ()
 
-  "test.op"() {"value" = 34.e-23 : !f32} : () -> ()
+  "test.op"() {"value" = 34.e-23 : f32} : () -> ()
```
compor, line 14: I couldn't find any MLIR reference where `!f32` is a valid type for this built-in.
Not sure if I'm overlooking something.

AntonLydike (approved):
Nice one!

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -64,7 +64,7 @@ def memref_shape_ops(
                 )
             bytes_per_element = element_type.width.data // 8
         case Float32Type():
-            bytes_per_element = 4
+            bytes_per_element = element_type.get_bitwidth // 8
```
superlopuh, line 67: probably below also, right?

xdsl/dialects/builtin.py
```diff
@@ -424,34 +424,67 @@ def verify(self) -> None:
 
 
 @irdl_attr_definition
-class BFloat16Type(ParametrizedAttribute, TypeAttribute):
+class _FloatType(ParametrizedAttribute, TypeAttribute, ABC):
+    name = "float_type"
```
superlopuh, line 428: why name here? and why _irdl_attr_definition? is it because it doesn't work the same way if only subclasses of ParametrizedAttribute have the annotation?

superlopuh: I have a feeling that this is a bit of a half-measure, is it that replacing all the uses of the FloatXType in one go is too big a change? It seems worth discussing on how to get rid of FloatXTypes completely, even if it's as follow-up PRs, it would be good to have a plan before we paint ourselves into a corner.

compor: Yeah, I tried that, but not sure how to encode a `BFloat16Type` for example and yes, bigger change.
Not sure if this corners us, currently in MLIR is quite inflexibly limited, so I went with the minimal API change for  now.

compor: > why name here? and why _irdl_attr_definition? is it because it doesn't work the same way if only subclasses of ParametrizedAttribute have the annotation?

Good point, oversight from trying different associations

xdsl/dialects/builtin.py
```diff
@@ -423,35 +423,66 @@ def verify(self) -> None:
...
+class _FloatType(ParametrizedAttribute, TypeAttribute, ABC):
+    width: ParameterDef[IntAttr]
+
+    def __init__(self, width: int | IntAttr) -> None:
+        if isinstance(width, int):
+            width = IntAttr(width)
+        super().__init__([width])
+
+    @property
+    def get_bitwidth(self) -> int:
```
superlopuh, line 435: My personal preference is for fewer helpers, especially in cases where typing out the implementation takes fewer characters than the name of the helper. I would lean towards removing this property, and having clients use `attr.width.data` directly. WDYT?

compor: Can do.

superlopuh (approved):
Thank you!

compor:
@math-fehr quick look?

math-fehr:
I personally dislike using attribute parameters to store this kind of information :/ 
Mostly because we are introducing a parameter that will always be the same.
I somehow feel that interfaces would be more appropriate here, or maybe just a `get_bitwidth` that returns an `int` (and keeping the inheritance)?

superlopuh:
Yeah that makes sense. This is an xDSL invention, the Float types? To me it feels like this PR is a step towards having exactly the same API for floats as for integers, which I think is a good API goal for a number of reasons. Would you rather make the transition in one go?

compor:
> Yeah that makes sense. This is an xDSL invention, the Float types? To me it feels like this PR is a step towards having exactly the same API for floats as for integers, which I think is a good API goal for a number of reasons. Would you rather make the transition in one go?

What do you mean an xDSL invention? There is a `FloatType` in MLIR.
I'd rather do a small step first and we can expand to integers later, but integers do use width as an attribute AFAIU.

I'm happy to simplify.

compor:
Smaller step for now and then unify this API for integers with an interface.

xdsl/dialects/builtin.py
```diff
@@ -422,36 +422,69 @@ def verify(self) -> None:
 
 AnyIntegerAttr: TypeAlias = IntegerAttr[IntegerType | IndexType]
 
+# class _FloatType(ParametrizedAttribute, TypeAttribute, ABC):
```
superlopuh, line 425: commented code

math-fehr:
> Yeah that makes sense. This is an xDSL invention, the Float types? To me it feels like this PR is a step towards having exactly the same API for floats as for integers, which I think is a good API goal for a number of reasons. Would you rather make the transition in one go?

Just as a note, so first MLIR has `FloatType`, but it's using inheritance, and it does not have parameters.
The problem in having parameters is that it doesn't encode the entirety of float semantics. For instance, there are like 4-5 floats with 8 bits if I recall? And there are some bitwidth that do not exist (like 9 bits).
So I would be a bit careful there, especially if we want to follow MLIR dialects.

math-fehr (approved):
Thanks!
At some point, we should probably change `AnyFloat` to be exactly `FloatType`, but that will require some more support in IRDL I think!

PR #1638, opened 2023-10-05
dialects: (riscv) use dict instead of mapping in riscv.py

superlopuh:
Pyright complains otherwise that `|` cannot be used between the arguments in `parse`

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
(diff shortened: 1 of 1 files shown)


compor (approved):

webmiche (commented):
Same here. Why CI and not riscv? :sweat_smile:

superlopuh:
Yeah I guess this one is a bit more of a functional change, I'll make it riscv

AntonLydike (approved):

adutilleul (approved):

PR #1639, opened 2023-10-05
core: make IRNode.parent an abstract property parent_node

superlopuh:
Pyright complains that it's technically mutable with different types, so someone could set `parent` to any `IRNode`.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1159,7 +1159,7 @@ class Block(IRNode):
     _first_op: Operation | None = field(repr=False)
     _last_op: Operation | None = field(repr=False)
 
-    parent: Region | None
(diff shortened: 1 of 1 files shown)


webmiche (commented):
LGTM. Though I wonder, why is this PR concerning the CI? Isn't this a core change?

superlopuh:
I picked CI because it was motivated by CI tools, the functionality barely changes, other than disallowing an undefined behaviour. I'm happy to change it.

webmiche:
I would say it's fixing a type error/inaccuracy in core. Sure we found it through CI, but it does not directly change the CI workflow really. I would argue if we spin that logic further, most testing PRs would have to be CI too....

I think I prefer the core tag over the CI tag here.

PR #1644, opened 2023-10-06
backend: (riscv) Fix snitch scfgw format and add canonicalization

compor:
This PR:

- Fixes `scfgw` to use the R-Type format (`rd` register is always 0)
- Adds canonicalization of `scfgw` to `scfgwi` when a known immediate value is loaded via `li`
- Tests of the above
- Doc string fixes

@superlopuh Can you please double-check the doc string changes, I think we had it wrong, but it's one of those things where when the documentation is wrong one should be cautious.


xdsl/dialects/riscv.py
```diff
@@ -2689,25 +2689,38 @@ class GetFloatRegisterOp(GetAnyRegisterOperation[FloatRegisterType]):
...
+    def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
+        from xdsl.transforms.canonicalization_patterns.riscv import (
+            ScfgwOpUsingImmediate,
+        )
+
+        return (ScfgwOpUsingImmediate(),)
+
+
 @irdl_op_definition
 class ScfgwOp(RsRsIntegerOperation):
```
superlopuh, line 2703:
```suggestion
class ScfgwOp(RdRsRsOperation[IntRegisterType, IntRegisterType, IntRegisterType]):
```

Might as well fix this, there is a result register, it's just constrained to be zero. We should similarly omit the rd when printing assembly as in `scfgwi`

superlopuh:
Yep re-reading the docs our doc string was wrong before, thank you for fixing it!

superlopuh (approved):
LGTM. I'm wondering whether we should add a new __init__ in the subclasses that will always set RD to zero. Seems like a good addition, maybe along with a custom verifier that checks that the type of rd is also zero when parsing/using create or build. That's more of a nit though

KGrykiel (approved):

PR #1652, opened 2023-10-11
core: (traits) Add `insert_or_update` method to SymbolTable

AntonLydike:
This patch adds a `insert_or_update` method to the SymbolTable trait. This introduces a portable way of adding symbols if they don't already exist. As @PapyChacal can confirm, the "if not exists" part is quite underutilized sometimes.

This patch also refactors two lowerings (printf to putchar, mpi to function calls) to use the new mechanic. This also fixes the mpi lowering, which did not previously respect existing symbols with the given name.


xdsl/traits.py
```diff
@@ -264,6 +264,31 @@ def lookup_symbol(
                 return SymbolTable.lookup_symbol(o, nested_name)
         return None
 
+    @staticmethod
+    def insert_symbol_if_not_exists(symbol_table_op: Operation, symbol_op: Operation):
```
compor, line 268: 3 concerns here:
1. MLIR has also a name uniquing functionality here (not very nice IMHO, but depends on the use cases)
2. Based on this, MLIR has a TODO to rename this method to `insertOrUpdate`
3. If we don't do this, maybe just name the method `insert` for now?

compor (approved):

xdsl/traits.py
AntonLydike: I could live with renaming this `insert_or_update` and changing the functionality accordingly. Any concerns?

superlopuh: You're telling me this is the name of the function in MLIR? I'd much rather go with `insert_or_update`. I'd usually call that `upsert` but not everyone is familiar with that term. Also please add a doc comment to explain the logic.

georgebisbas (commented):
What was the mpi issue exactly?

AntonLydike:
> What was the mpi issue exactly?

If multiple instances of `lower-mpi` would be executed in a lowering pipeline, multiple instances of the external function definitions would be inserted to the module. I'm not sure when this would occur, but it was technically a bug

xdsl/traits.py
math-fehr, line 268: Can you please document the function?
It's not obvious to me from the name what the function does, and what are the constraints that need to be fulfilled.

math-fehr: Also, why adding `insert_symbol_if_not_exists`, and not `if_symbol_in_table` for instance?

AntonLydike: I renamed it to `insert_or_update` to follow MLIR, and added a docstring. LMK if there's anything else that needs changing.

PR #1660, opened 2023-10-12
misc (makefile): Mark venv target as phony

compor:
This PR marks the `venv` Make target as phony as it should not depend on whether the directory exists or not.

If the `venv` directory is created and any of the subsequent commands of the target fail, the target is not rerun (i.e., always up to date) unless the directory is removed.
Moreover, this behaviour differs when a different venv dir is selected when `VENV_DIR` is set.


math-fehr (approved):
Nice!

PR #1662, opened 2023-10-12
interpreter: pdl native constraints constant

math-fehr:
Add support for constants in `pdl.apply_native_constraint`.

Something like
```python
    %attr = attribute = 10
    pdl.apply_native_constraint "constraint"(%attr: !pdl.attribute)
```
where `%attr` is not bound

### DIFF
--- a/tests/interpreters/test_pdl_interpreter.py
+++ b/tests/interpreters/test_pdl_interpreter.py
@@ -438,7 +438,7 @@ def test_native_constraint():
     @ModuleOp
     @Builder.implicit_region
     def input_module_false():
-        test.TestOp.create(properties={"attr": StringAttr("fooo")})
+        test.TestOp.create(properties={"attr": StringAttr("baar")})
 
     @ModuleOp
     @Builder.implicit_region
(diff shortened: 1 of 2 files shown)


tests/interpreters/test_pdl_interpreter.py
```diff
@@ -473,3 +473,57 @@ def pdl_module():
...
+
+    @ModuleOp
+    @Builder.implicit_region
+    def input_module_true():
+        test.TestOp.create(properties={"attr": StringAttr("foo")})
+
+    @ModuleOp
+    @Builder.implicit_region
+    def input_module_false():
+        test.TestOp.create(properties={"attr": StringAttr("fooo")})
```
superlopuh, line 492:
```suggestion
        test.TestOp.create(properties={"attr": StringAttr("bar")})
```
🙏

superlopuh (approved):

tests/interpreters/test_pdl_interpreter.py
math-fehr: No, it has to be 4 letters sadly :(
I'll do "baar" if you want

PR #1672, opened 2023-10-17
transforms: (dmp) Add a 3D domain decomposition (with n-d generalization)

AntonLydike:
This adds a `GridSlice3d` decomposition, that used the `_flat_face_exchanges_for_dim(shape, dim)` helper.

This helper generates the exchange declarations for the two opposite flat faces in dim `dim` of any n-dimensional hypercube (n>1 I think?).

I wrote some tests for the `_flat_face_exchanges_for_dim` method, but no filechecks as of yet (they will follow). I would invite @mesham and @mcjamieson to be a second and third brain + pair of eyes, as I am not entirely sure I got everything right yet. [The test](https://github.com/xdslproject/xdsl/pull/1672/files#diff-fbe221f08ca78dc534bcc61aae61ffe1c7f1d88dec29fd56316245dae9976d66) is probably a good point to look at the result, and then you can dive into the code from there. Feel free to drop by my office too, I have some diagrams to support my code on the whiteboard.

@PapyChacal and @georgebisbas, I would love to try and run these with correctness checks, but may need your help with that too.

### DIFF
--- a/tests/dialects/test_dmp.py
+++ b/tests/dialects/test_dmp.py
@@ -1,0 +1,81 @@
+from xdsl.dialects.experimental.dmp import ExchangeDeclarationAttr, ShapeAttr
+from xdsl.transforms.experimental.dmp import decompositions
+
+
+def flat_face_exchanges(
+    shape: ShapeAttr, dim: int
+) -> tuple[ExchangeDeclarationAttr, ExchangeDeclarationAttr]:
+    # we need access to the _flat_face_exchanges_for_dim method in order to test it
+    # since this is a private function, and pyright will yell whenever it's accessed,
+    # we have this wrapper function here that takes care of making the private publicly
+    # accessible in the context of this test.
+    func = (
+        decompositions._flat_face_exchanges_for_dim  # pyright: ignore[reportPrivateUsage]
+    )
+    return func(shape, dim)
+
+
+def test_decomp_flat_face_3d():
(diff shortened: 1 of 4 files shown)


tests/dialects/test_dmp.py
```diff
@@ -0,0 +1,77 @@
+from xdsl.dialects.experimental.dmp import ExchangeDeclarationAttr, ShapeAttr
+from xdsl.transforms.experimental.dmp import decompositions
+
+
+def flat_face_exchanges(
+    shape: ShapeAttr, dim: int
+) -> tuple[ExchangeDeclarationAttr, ExchangeDeclarationAttr]:
+    func = (
+        decompositions._flat_face_exchanges_for_dim  # pyright: ignore[reportPrivateUsage]
```
superlopuh, line 9: Can you please add a small comment saying why you need to use a private function here?

superlopuh: I mean even if it's a test file it might still be a useful thing to add

tobiasgrosser:
@AntonLydike, is there a reason this has not been merged?

georgebisbas:
I think we already used this in some branch, but probably waiting testing from PsyClone?

tobiasgrosser:
I don't think we need large scale testing to merge code. As long as the design is sensible and we have proper in-tree testing and reasonable confidence that this works I am happy to get things in. If further testing exposes bugs, we can always correct things.

PR #1674, opened 2023-10-18
misc: add a launch.json to launch xdsl-opt from vscode

superlopuh:
This should only affect users of VSCode, and not really impact anyone else. By default xdsl-opt launches with the currently open file as argument, parsing and printing it. I use launch configs a lot for debugging and would love to have the file in the repo.


AntonLydike (approved):
:+1:

math-fehr (approved):

PR #1675, opened 2023-10-18
core: (affine) add AffineMap.transpose_map

superlopuh:
I couldn't find an equivalent helper in MLIR Affine, but seems like an appropriate place to put it. Adds a helper to create the affine map from a matrix to its transpose.

### DIFF
--- a/tests/test_affine_builtins.py
+++ b/tests/test_affine_builtins.py
@@ -99,6 +99,12 @@ def test_composition():
     assert map3.eval([5, 6], []) == [-3, -28]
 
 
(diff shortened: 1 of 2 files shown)


superlopuh:
On second thought this helper might be more confusing than it's worth, torch.tensor transpose takes two dimensions to transpose. Also it's not clear from the name alone whether the transpose acts on the domain or codomain. I'll have another go at solving this

compor (approved):

math-fehr (approved):
+1

PR #1641, opened 2023-10-05
[WIP] linalg to snitch

superlopuh:
WIP branch with changes to compile to snitch in Toy.

PRs left to open:
 - [x] pass function arguments to `xdsl-run`
 - [ ] implement interpreter for memref-stream and friends

PR #1679, opened 2023-10-19
dialects: (stream) add stream dialect with basic ops

superlopuh:


adutilleul (approved):

webmiche (commented):
Is this inspired by something? I know some people working in this direction, maybe they would be interested?

superlopuh:
This is a part of the lowering pipeline towards Snitch, which uses streaming registers. We're currently splitting the lowering in two, the first part converts linalg parallelized operations to streams, the second lowers to riscv-specific code. I'm happy to share the slides/have a catch-up call to discuss all this. The WIP total flow is in the #1641 PR, where you can see how this dialect gets used.

PR #1684, opened 2023-10-20
core: add PassPipelinePass with optional callback in between passes

superlopuh:
Also makes the behaviour of `print-between-passes` actually correspond to the behaviour of only printing between passes, not after every pass.

### DIFF
--- a/tests/xdsl_opt/test_xdsl_opt.py
+++ b/tests/xdsl_opt/test_xdsl_opt.py
@@ -146,9 +146,7 @@ def test_print_between_passes():
         opt.run()
 
     output = f.getvalue()
-    assert (
-        len([l for l in output.split("\n") if "builtin.module" in l]) == len(passes) + 1
-    )
+    assert len([l for l in output.split("\n") if "builtin.module" in l]) == len(passes)
 
 
 def test_diagnostic_exception():
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -111,7 +111,7 @@ def _empty_callback(
 
 
 @dataclass
-class PassPipelinePass(ModulePass):
+class PipelinePass(ModulePass):
     passes: list[ModulePass]
     callback: Callable[[ModulePass, builtin.ModuleOp, ModulePass], None] = field(
         default=_empty_callback
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -1,6 +1,7 @@
 import dataclasses
 from abc import ABC, abstractmethod
-from dataclasses import Field, dataclass
+from collections.abc import Callable
+from dataclasses import Field, dataclass, field
 from types import NoneType, UnionType
 from typing import Any, ClassVar, TypeVar, Union, get_args, get_origin
 
@@ -72,21 +73,21 @@ class ModulePass(ABC):
         arg_dict = dict[str, PassArgListType | PassArgElementType | None]()
 
         # iterate over all fields of the dataclass
-        for field in fields:
+        for op_field in fields:
             # ignore the name field and everything that's not used by __init__
-            if field.name == "name" or not field.init:
+            if op_field.name == "name" or not op_field.init:
                 continue
             # check that non-optional fields are present
-            if field.name not in spec.args:
-                if _is_optional(field):
-                    arg_dict[field.name] = _get_default(field)
+            if op_field.name not in spec.args:
+                if _is_optional(op_field):
+                    arg_dict[op_field.name] = _get_default(op_field)
                     continue
-                raise ValueError(f'Pass {cls.name} requires argument "{field.name}"')
+                raise ValueError(f'Pass {cls.name} requires argument "{op_field.name}"')
 
             # convert pass arg to the correct type:
-            arg_dict[field.name] = _convert_pass_arg_to_type(
-                spec.args.pop(field.name),
-                field.type,
+            arg_dict[op_field.name] = _convert_pass_arg_to_type(
+                spec.args.pop(op_field.name),
+                op_field.type,
             )
             # we use .pop here to also remove the arg from the dict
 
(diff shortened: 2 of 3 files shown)


tests/xdsl_opt/test_xdsl_opt.py
```diff
@@ -146,9 +146,7 @@ def test_print_between_passes():
         opt.run()
 
     output = f.getvalue()
-    assert (
-        len([l for l in output.split("\n") if "builtin.module" in l]) == len(passes) + 1
-    )
+    assert len([l for l in output.split("\n") if "builtin.module" in l]) == len(passes)
```
math-fehr, line 149: What changed here?
Were we printing too much at one point?

xdsl/passes.py
```diff
@@ -103,6 +104,35 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
...
+
+@dataclass
+class PassPipelinePass(ModulePass):
+    passes: list[ModulePass]
+    callback: Callable[[ModulePass, builtin.ModuleOp, ModulePass], None] = field(
+        default=_empty_callback
+    )
+    """
+    Function called in between every pass, taking the pass that just ran, the module, and
+    the next pass.
```
math-fehr, line 121: Should this be run after every pass, or in between passes?
Should we have two options? Not sure. 
I feel this kind of complexity may not make this worth it?

xdsl/xdsl_opt_main.py
```diff
@@ -212,9 +212,21 @@ def setup_pipeline(self):
...
                 raise Exception(f"Unrecognized pass: {p.name}")
 
-        self.pipeline = [
-            self.available_passes[p.name].from_pass_spec(p) for p in pipeline
-        ]
+        def callback(
+            previous_pass: ModulePass, module: ModuleOp, next_pass: ModulePass
+        ) -> None:
+            if not self.args.disable_verify:
+                module.verify()
```
math-fehr, line 219: Here, we don't verify after the last pass anymore?

math-fehr (commented):
What's your current use case for this?
Because since we could have the same for after/before each pass, I'm not really sure that we want something like this by default.

tests/xdsl_opt/test_xdsl_opt.py
superlopuh: Yeah, the print-between-passes also printed after the last pass, which seems wrong to me

xdsl/xdsl_opt_main.py
superlopuh: Ah yes that's true, I'll add the verification back

xdsl/passes.py
superlopuh: I think in between passes is the only thing that this class needs to provide, the client can call their own code before or after executing the pipeline

superlopuh:
My use-case is to provide more functionality for people building a compiler using the Python API, as opposed to using the command-line tool. In Toy, I found that this simplified my code quite a bit, in my Snitch branch. I think that clients of xDSL would benefit from this helper.

xdsl/xdsl_opt_main.py
superlopuh: Done

xdsl/passes.py
webmiche: After executing the pipeline, a lot of state will be deleted, right? So you might want to have several options of callbacks 🤔

xdsl/xdsl_opt_main.py
```diff
@@ -26,7 +26,7 @@ class xDSLOptMain(CommandLineTool):
     stream.
     """
 
-    pipeline: list[ModulePass]
+    pipeline: PassPipelinePass
```
webmiche, line 29: While I agree that several passes can be viewed as one pass, I think the name `PassPipelinePass` will be quite confusing for people. The previous type annotation was more intuitive, I think.

Maybe we could rename this? Isn't this just essentially a class capturing a pass pipeline? So why not just `PassPipeline`?

webmiche (commented):
I think this opens the door for in-flight parametrized passes, right? You could run one pass, that produces some result, which can then be fed into the next pass. That sounds quite cool tbh


What is the exact use case you were thinking about? After fighting with LLVM for a half a year, I think extending this idea and making it very generic would be an extremely powerful debugging tool.

xdsl/passes.py
```diff
@@ -103,6 +104,35 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
...
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        if not self.passes:
+            # Early exit to avoid fetching a non-existing last pass.
+            return
+
+        for prev, next in zip(self.passes[:-1], self.passes[1:]):
+            prev.apply(ctx, op)
+            self.callback(prev, op, next)
+
+        self.passes[-1].apply(ctx, op)
```
compor, line 133: A pipeline by definition is a serial processing regime.
Why do we need to know the next pass?
Can't the callback points only be modelled around a single pass' behaviour?
For example,
- before execution
- after execution
- when the IR changes? (not sure how viable this is)

xdsl/xdsl_opt_main.py
compor: I agree, weirdly recurring, although I think `PipelinePass` might encapsulate our current naming convention better (A pipeline is a series of passes, but itself is also a pass).

compor (approved):
I like this.
It seems like an opportunity to make the pass infrastructure a bit more supple.
It's a bit reminiscent of how LLVM transitioned from the old to the new-style pass manager a few years back, but Python can offer an even more flexible perspective.

xdsl/passes.py
```diff
@@ -72,21 +73,21 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
         arg_dict = dict[str, PassArgListType | PassArgElementType | None]()
 
         # iterate over all fields of the dataclass
-        for field in fields:
+        for op_field in fields:
```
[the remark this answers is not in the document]
superlopuh: Ruff complained about a loop variable shadowing an import, which seems sensible. So it's kinda related to the import of `field` from `dataclasses`

xdsl/passes.py
superlopuh: My main idea was that the API guarantees that it's in between passes and not before the first one or after the last one. If you have both then it's unambiguous. The other thing is that it may be useful to execute something before a certain pass, maybe as a debug step, independent of the passes before. There may be other use-cases that I'm not thinking of.

xdsl/xdsl_opt_main.py
superlopuh: Yeah I like `PipelinePass` more than `PassPipeline` since I think it communicates more clearly that it's also a pass itself. I'll rename it.

xdsl/passes.py
```diff
@@ -103,6 +104,35 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
...
+        default=_empty_callback
+    )
+    """
+    Function called in between every pass, taking the pass that just ran, the module, and
+    the next pass.
+    """
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        if not self.passes:
+            # Early exit to avoid fetching a non-existing last pass.
```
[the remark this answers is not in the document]
superlopuh: I couldn't find that spelling... https://www.merriam-webster.com/dictionary/existant I think the current phrasing is ok, but maybe we could change it another way if you think it needs work?

xdsl/passes.py
superlopuh: Not sure exactly what you mean

xdsl/passes.py
compor: Thanks for the clarification; if the handling is at the "pipeline level", this makes sense.

PR #1686, opened 2023-10-22
core: (affine_map) add a convenience affine map constructor with lambda

superlopuh:
I think we'd benefit from a little helper here

### DIFF
--- a/xdsl/ir/affine/affine_map.py
+++ b/xdsl/ir/affine/affine_map.py
@@ -31,6 +31,15 @@ class AffineMap:
         )
 
     @staticmethod
+    def transpose(rank: int) -> AffineMap:
+        """
+        Reverses the indices.
+        """
+        return AffineMap(
+            rank, 0, tuple(AffineExpr.dimension(dim) for dim in reversed(range(rank)))
+        )
+
+    @staticmethod
     def empty() -> AffineMap:
         return AffineMap(0, 0, ())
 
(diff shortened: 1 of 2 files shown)


xdsl/ir/affine/affine_map.py
```diff
@@ -42,6 +55,43 @@ def transpose_map() -> AffineMap:
...
         return AffineMap(0, 0, ())
 
+    @staticmethod
+    def from_callable(
+        func: AffineMapBuilderT, *, dim_symbol_split: tuple[int, int] | None = None
+    ) -> AffineMap:
+        """
+        Creates an `AffineMap` by calling the function provided. If `dim_symbol_split` is
+        not provided or `None`, then all parameters are treated as dimension expressions.
+        If `dim_symbol_split` is provided, `func` is expected to have the same number of
```
compor, line 65: Does this condition/expectation need to be tested within this or do we just let all error handling/checking trickle down to `AffineMap`?

compor (approved):

xdsl/ir/affine/affine_map.py
superlopuh: I was thinking about that. It would be really easy to test in the code, and I didn't do it because it feels like it'll be slow but, thinking about it some more, if that's a problem then the user can use a different initializer, so it shouldn't really be a concern. I'll add the check.

math-fehr (approved):
Looks really cool!

PR #1687, opened 2023-10-23
backend: (riscv) remove lower-snrt-to-func pass

superlopuh:
It turns out that the snitch runtime C library isn't designed to be linked against, and that all the implementations need to be inlined for it to work properly. We can't really do that with our setup, so we might as well delete this code for now.


compor (approved):

JosseVanDelm (approved):
:smiling_face_with_tear:

PR #1692, opened 2023-10-24
transformations: Rename dmp-decompose-2d to dmp-decompose

PapyChacal:
(This is targetting the devito backport branch, not main)

The actual domain (and grid) dimensionality of dmp-decompose-2d is now parameterized, making the old-name pretty confusing in user-code. Simple rename to make it clear the dimensionality is not fixed by the pass :slightly_smiling_face:


AntonLydike (approved):
Nice!

PR #1694, opened 2023-10-24
dialects: (gpu) Elide returns in __init__ methods

compor:
This PR:

- Elides the explicit `return` statements in `__init__` methods of the `gpu` dialect.

This must have been a leftover from when `.get()` constructor methods were used.
I had in the backlog and my linter complained since I happened to open the dialect source file recently.


superlopuh (approved):

math-fehr (approved):
Oh yes, thanks!

PR #1695, opened 2023-10-24
backend: (riscv) Add canonicalization pattern for AddImmediateConstant

superlopuh:
Also throw in some Pure() annotations

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/riscv.py
+++ b/xdsl/transforms/canonicalization_patterns/riscv.py
@@ -129,6 +129,24 @@ class AddImmediateZero(RewritePattern):
             rewriter.replace_matched_op(riscv.MVOp(op.rs1, rd=rd))
 
 
+class AddImmediateConstant(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: riscv.AddiOp, rewriter: PatternRewriter) -> None:
+        if (
+            isinstance(li := op.rs1.owner, riscv.LiOp)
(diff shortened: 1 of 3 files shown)


xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -129,6 +129,24 @@ def match_and_rewrite(self, op: riscv.AddiOp, rewriter: PatternRewriter) -> None
             rewriter.replace_matched_op(riscv.MVOp(op.rs1, rd=rd))
 
 
+class AddImmediateConstant(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: riscv.AddiOp, rewriter: PatternRewriter) -> None:
+        if (
+            isinstance(li := op.rs1.owner, riscv.LiOp)
```
compor, line 136: Please remind me why do we do `.owner` here instead of checking if it's an instance of `OpResult` (like in other canonicalizations here)?

superlopuh: It avoids an extra checks, I think the other rewrites should actually be changed, but it's not a high priority.

compor: Yeah, OK, apart from saving a check, what is the difference?

math-fehr (approved):
Nice!
I'm not sure if the `Pure` trait was meant for that PR? I'm fine if it is though.

superlopuh:
It kind of isn't but it feels like a fairly minor change to open another PR for.

PR #1698, opened 2023-10-25
interpreter: (riscv) gather data from all .data sections

superlopuh:
@xdslproject/risc-v-backend

### DIFF
--- a/xdsl/interpreters/riscv.py
+++ b/xdsl/interpreters/riscv.py
@@ -262,33 +262,43 @@ class RiscvFunctions(InterpreterFunctions):
 
     @staticmethod
     def get_data(module_op: ModuleOp) -> dict[str, RawPtr]:
-        data: dict[str, RawPtr] | None = None
+        data: dict[str, RawPtr] = {}
         for op in module_op.ops:
             if isinstance(op, riscv.AssemblySectionOp):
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/riscv.py
```diff
@@ -281,9 +282,9 @@ def get_data(module_op: ModuleOp) -> dict[str, RawPtr]:
                                 assert (
                                     False
                                 ), f"Unexpected directive {data_op.directive.data}"
-                    return data
-        else:
+        if data is None:
```
compor, line 285:
```suggestion
       assert data is not None, "Could not find data section"
```

compor: I'm a bit confused by all the `if` checks and `assert` statements.

compor (approved):

xdsl/interpreters/riscv.py
superlopuh: They're markers of me being lazy, really they should be raising better errors, I'd rather either change them to errors now, or keep the weird if data is None structure.

superlopuh: Swapped asserts

compor (approved):

adutilleul (approved):

PR #1702, opened 2023-10-26
backend: (riscv) make dimension an IntAttr in snitch ops

superlopuh:
I'm not sure what exactly the tradeoff is between making things a plain intattr or the integer attr with signedness, but it feels like if we're operating on things that are actually 5 or 7 bits wide and keeping them in i32 for some reason then it's a bit more confusing. I'm also not sure we want to take the time now to ensure that the values fit in those widths, that'll be true for our lowerings by construction.

@xdslproject/risc-v-backend

### DIFF
--- a/tests/dialects/test_snitch.py
+++ b/tests/dialects/test_snitch.py
@@ -1,7 +1,7 @@
 import pytest
 
 from xdsl.dialects import riscv, snitch
-from xdsl.dialects.builtin import IntegerAttr, i32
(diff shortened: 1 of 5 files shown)


nazavode (approved):

compor (approved):

compor:
Maybe track the check for correct bitwidth with an issue.

adutilleul (approved):

PR #1703, opened 2023-10-26
backend: (riscv) SsrSetStreamConfigOperation has dm attr instead of stream operand

superlopuh:
Two main changes:
 `stream` -> `dm` in snitch, since those ops correspond more to the assembly level than the stream level, and having naming consistent with the docs and snitch runtime library seems advantageous.
  operand -> attribute, we always have that data at compile-time, so we might as well avoid the dance of `li` + `addi` canonicalisation, and just create the `li` at the end.

@xdslproject/risc-v-backend

### DIFF
--- a/tests/dialects/test_snitch.py
+++ b/tests/dialects/test_snitch.py
@@ -7,30 +7,30 @@ from xdsl.utils.test_value import TestSSAValue
 
 
 def test_csr_op():
-    stream = TestSSAValue(riscv.Registers.A1)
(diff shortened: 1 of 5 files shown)


nazavode (approved):
Much more streamlined, thanks.

compor (approved):

compor:
Thanks for the name change; I've been wanting to do that for a while now.

adutilleul (approved):

PR #1704, opened 2023-10-26
[WIP] dialects: (hw) Implement HW inner symbols

lucjaulmes:
This seems to be the minimal subset of the `HW` dialect to implement inner symbols, which are required by the `seq` dialect. Essentially it is a symbol lookup system parallel to `mlir.SymbolTable` (here `hw.InnerSymbolTable`) to be able to lookup symbols not in the direct parent but beyond that, specifying the symbol location (generally a module) by name.

I’ve gathered what I could from upstream and pythonized/xdsl’d it as much as possible, but I’m now left with a lot of interfaces/traits and a few attributes (the inner refs themselves) -- so mostly abstract things, and I’m not entirely sure how to proceed to writing tests.

I would like to avoid implementing all the operations/types etc. inheriting the traits/interfaces, to be able to split the contribution into smaller, more manageable bits for review. Keeping the commit split with 1-4 classes and ≤150 lines each would be nice (even though the first commit lacks some coherence without context ; hopefully this PR gives that context).

Any pointers on the best strategy for this are welcome, it’s kind of hard to grep for in the codebase. There aren’t any tests in CIRCT that don’t rely on concrete operations either. For example, how do you test the parsing/printing of attributes that are not used by any operations (yet)?

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -947,6 +947,17 @@ class Operation(IRNode):
         """
         return [t for t in cls.traits if isinstance(t, trait_type)]
 
+    def get_parent_with_trait(
+        self, trait: type[OpTraitInvT], parameters: Any = None
+    ) -> Operation | None:
+        """
+        Returns the closest surrounding parent operation with given trait.
+        """
+        op: Operation | None = self
+        while op is not None and not op.has_trait(trait, parameters):
+            op = op.parent_op()
+        return op
+
     def erase(self, safe_erase: bool = True, drop_references: bool = True) -> None:
         """
         Erase the operation, and remove all its references to other operations.
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -5,10 +5,13 @@ It currently implements minimal types and operations used by other dialects.
 [1] https://circt.llvm.org/docs/Dialects/HW/
 """
 
-from collections.abc import Mapping
+from collections.abc import Iterable, Iterator, Mapping, Sequence
+from typing import cast, overload
 
 from xdsl.dialects.builtin import (
+    ArrayAttr,
     FlatSymbolRefAttr,
+    IntAttr,
     ParameterDef,
     StringAttr,
 )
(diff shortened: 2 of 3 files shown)


superlopuh:
We have test operations that can create and consume any number of values of arbitrary types, it's probably worth searching for `"test.op"` in the filechecks

xdsl/ir/core.py
```diff
@@ -947,6 +947,17 @@ def get_traits_of_type(cls, trait_type: type[OpTraitInvT]) -> list[OpTraitInvT]:
         """
         return [t for t in cls.traits if isinstance(t, trait_type)]
 
+    def get_parent_with_trait(
```
superlopuh, line 950: I would want to call this `get_ancestor_with_trait`, but I'm not sure if we use `parent` elsewhere for the same concept. Have you found another funciton in MLIR or xDSL where parent is not the direct parent op?

PapyChacal: +1

lucjaulmes: The name is copied from the MLIR codebase. There’s `getParentOfType()` and `getParentWithTrait()` that go up symbols using `getParentOp()`. Then there are `isAncestor()` and `isProperAncestor()` that go up symbols using `getParentRegion()`.

superlopuh: Ok when in doubt we stick with MLIR conventions even when they don't make a lot of sense :)

tobiasgrosser:
Is this something we can/want to merge? I have not seen a lot of activity here over the last two weeks.

lucjaulmes:
Not something we want to merge no − it’s an overview to give context for smaller PRs and to allow discussion early on.

Essentially the same contribution is decomposed in linked #1708 #1758 and #1780 (and maybe 1 or 2 more). I was planning on closing it when all the parts are merged.

tobiasgrosser:
@lucjaulmes, can this meanwhile be merged?

lucjaulmes:
I don’t think there’s much of a point to keep this open as the individual PRs this was broken into have had to evolve a lot. Work is still ongoing though.

PR #1710, opened 2023-10-27
interpreter: (snitch_stream) add SnitchStreamInterpreterFunctions

superlopuh:
CC @xdslproject/risc-v-backend

### DIFF
--- a/xdsl/interpreters/snitch_stream.py
+++ b/xdsl/interpreters/snitch_stream.py
@@ -1,0 +1,133 @@
+from collections.abc import Iterator
+from dataclasses import dataclass
+from itertools import product
+from typing import Any
+
+from xdsl.dialects import snitch_stream
+from xdsl.interpreter import (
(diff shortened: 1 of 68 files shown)


compor (approved):

xdsl/interpreters/snitch_stream.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+
+
+@dataclass
+class StridePattern:
+    ub: list[int]
+    strides: list[int]
+
+    @property
+    def offset_expr(self) -> AffineExpr:
```
compor, line 63: Sorry, can you add some basic doc on this method, and the indexing and offset `map_from` methods above?

Maybe just adding comments to the test would work best? Not sure, but I feel there are a lot of abstraction levels "meeting" here.

compor (changes requested):

PR #1711, opened 2023-10-29
core: Name dialects and make then citizens of MLContext.

PapyChacal:
This is to

1. Get closer to MLIR
2. Specifically, allow to handle opaque dialect attributes properly (as defined here https://mlir.llvm.org/docs/LangRef/#dialect-attribute-values; "opaque" seems overloaded on MLIR's side.). Stacked #1712

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -35,6 +35,8 @@ OpT = TypeVar("OpT", bound="Operation")
 class Dialect:
     """Contains the operations and attributes of a specific dialect"""
 
+    _name: str
+
     _operations: list[type[Operation]] = field(
         default_factory=list, init=True, repr=True
     )
@@ -50,6 +52,10 @@ class Dialect:
     def attributes(self) -> Iterator[type[Attribute]]:
         return iter(self._attributes)
 
+    @property
+    def name(self) -> str:
+        return self._name
+
 
 @dataclass
 class MLContext:
@@ -57,6 +63,7 @@ class MLContext:
 
     allow_unregistered: bool = field(default=False)
 
+    _loaded_dialects: dict[str, Dialect] = field(init=False, default_factory=dict)
     _loaded_ops: dict[str, type[Operation]] = field(init=False, default_factory=dict)
     _loaded_attrs: dict[str, type[Attribute]] = field(init=False, default_factory=dict)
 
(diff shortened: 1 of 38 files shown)


math-fehr (approved):
Nice!
That should also help me for the lazy registration of dialects ;)

webmiche (commented):
LGTM, only a minor comment.

In general, we usually define our ops and attrs a "{dialect.name}.{op/attr.name}". We could move this to "{op/attr.name}" after this change. What do you guys think?

xdsl/ir/core.py
```diff
@@ -50,13 +52,18 @@ def operations(self) -> Iterator[type[Operation]]:
...
 
 @dataclass
 class MLContext:
     """Contains structures for operations/attributes registration."""
 
     allow_unregistered: bool = field(default=False)
 
+    _loaded_dialects: dict[str, Dialect] = field(init=False, default_factory=dict)
     _loaded_ops: dict[str, type[Operation]] = field(init=False, default_factory=dict)
     _loaded_attrs: dict[str, type[Attribute]] = field(init=False, default_factory=dict)
```
webmiche, line 68: We could also not load the operations/attrs individually and go through all the dialects when looking up. What do you think?

PapyChacal:
I had some thoughts in that area, and I'd be happy to discuss them separately, as this is the minimal enabling change

My personal opinion at this point is that it's very handy to have the fully specified name in the Operations and Attributes, I would leave them there and maybe enforce the prefix when creating or registering the Dialect, WDYT? Otherwise, for usability we would at least need a reference to the Dialect from Operations and it starts complexifying 🫤

xdsl/ir/core.py
PapyChacal: I like it in general, but it requires some design and thoughts, especially wrt to unregistered attributes and operations

Would you be comfortable merging as is, and starting a discussion over Zulip? I do think it's valuable to think about it, and at the same time I'd like to enable whatever just naming dialects enables first 🙂 (opaque dialect attributes being my first target)

webmiche: Sure, this is essentially a first "backwards compatible" version that saves us from having to rename all oeprations, so I am fine with first merging this.

webmiche:
Let's discuss this further on Zulip. It might also be interesting to check what MLIR does exactly. Maybe we can even make operation names a bit declarative and add the dialect name on the fly :thinking:

superlopuh (approved):

xdsl/ir/core.py
PapyChacal: Exactly, cool :slightly_smiling_face:

PR #1712, opened 2023-10-29
[WIP/RFC] core: Add OpaqueAttribute

PapyChacal:
This is WIP for implementing OpaqueAttribute, as defined at https://mlir.llvm.org/docs/LangRef/#dialect-attribute-values. (The term "opaque" seems overloaded)
It uses it on my early hack for the `gpu.dim` and `gpu.all_reduce_op` attributes, as an example and sanitychecker; happy to split the PR once we agree on the result!

Opening as a draft to have feedback before overengineering the wrong way!
This on implements it as a syntax tweak, similarly to `TypeAttribute`, specifically it only implements attributes having the `$dialect<$name body>` syntax, not the fully generic `$dialect<whateverbody>` one. Started with that and opening discussion because I fell I'll easily get lost otherwise:

1. I didn't encounter any opaque attribute not using that one (ruling out builtin ones, cooked in the parser logic itself either way!). But it might be dangerous to not care; WDYT?
2. Thinking about it, I think worst case we can use this one for matching attributes, and if counterexamples showup, we could factor out the more generic logic and keep this one as a common helper?

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -520,6 +520,7 @@ class CastOp(IRDLOperation):
 
 
 Toy = Dialect(
+    "toy",
     [
         ConstantOp,
         AddOp,
--- a/tests/dialects/test_gpu.py
+++ b/tests/dialects/test_gpu.py
@@ -3,7 +3,7 @@ from xdsl.dialects import arith, builtin, memref
 from xdsl.dialects.gpu import (
     AllocOp,
     AllReduceOp,
-    AllReduceOperationAttr,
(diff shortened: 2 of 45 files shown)


superlopuh (changes requested):
Thank you for this! Looking great, just a couple of things that would be nice to explore before merging

PapyChacal:
@superlopuh Thanks for the review; I understood a couple more things meanwhile thanks to our current weird syntax specialist @math-fehr though and will have completely another go at this though!

PapyChacal:
Closing in favor of a new WIP, more compliant with MLIR

PR #1716, opened 2023-10-30
core: Remove duplicated methods

math-fehr:
This removes methods in `Parser` that are already duplicated in `AttrParser`.


PapyChacal (approved):
:heart:

compor (approved):

PR #1643, opened 2023-10-06
backend: (riscv) Add an initial lowering from snitch_runtime to snitch

AntonLydike:
Add lowering from `snitch_runtime` to `snitch`.

PR #1718, opened 2023-10-30
backend: (riscv) add convert-snitch-stream-to-snitch

superlopuh:
Thank you @AntonLydike for the intitial implementation in #1643! This more or less copies those changes and lowers from snitch-stream instead, since we no longer want to use the snitch runtime.

CC @xdslproject/risc-v-backend


compor (approved):
Thanks!

PR #1723, opened 2023-10-31
core: add AbstractYieldOperation and use in riscv-scf

superlopuh:
Just a little helper that should remove some boilerplate code as we add new dialects with a yield-like operation. I only used it in riscv-scf for now but am happy to migrate the rest of them in this PR.


compor (approved):

AntonLydike (approved):
Nice!

PR #1724, opened 2023-10-31
core: Remove hardcoded angle brackets from Data printing/parsing.

PapyChacal:
More MLIR compliancy! This is a step of my proper opqaue syntax implementation quest.
Basically, the angle brackets we use in Data attributes are not supposed to be part of the general syntax, but is just a convention adopted by most attributes.
This reverts this hardcoding, moving those brackets in every Data Attribute's custom parsing/printing code.
It already implements helpers for this; both for my laziness and for next users' sanity :slightly_smiling_face:


superlopuh (approved):
beautiful

compor (approved):

PR #1725, opened 2023-10-31
dialects: (scf) Revamp `scf.for` verification logic and messages

compor:
This PR:

- Separates the region and body block verification checks for `scf.for`
- Updates the custom verification messages with a bit more information (i.e., operand position)
- Renames the uses of `index`, `idx` to `indvar` and `i` respectively.
  I found the mixed use of `IndexType`, `index` and `idx` rather confusing.
- Updates the filecheck verification messages for the above.

This was en route to allow signless integers as types for induction variables (I'll open a separate PR for that).


tests/filecheck/dialects/scf/for_iv.mlir
```diff
@@ -3,7 +3,7 @@
   %lb = "arith.constant"() {"value" = 0 : index} : () -> index
   %ub = "arith.constant"() {"value" = 42 : index} : () -> index
   %step = "arith.constant"() {"value" = 7 : index} : () -> index
-// CHECK: Wrong number of block arguments, expected 1, got 0. The body must have the induction variable and loop-carried variables as arguments.
+// CHECK: Body block must at least have induction var as block arg
```
superlopuh, line 6: this reads weird

superlopuh: How about this:

```suggestion
// CHECK: Body block must have induction var as first block arg
```

superlopuh (approved):

tests/filecheck/dialects/scf/for_iv.mlir
compor: It was what `mlir-opt` was printing as error message; not married to it :)

superlopuh: Ah maybe in that case we can stick with mlir-opt, your call

xdsl/dialects/scf.py
```diff
@@ -189,44 +189,61 @@ def get(
         return For(lb, ub, step, iter_args, body)
 
     def verify_(self):
-        if (len(self.iter_args) + 1) != len(self.body.block.args):
+        # op region verification
+        for i, opnd in enumerate(self.operands[:3]):
+            if not isinstance(opnd.type, IndexType):
```
compor, line 194: This check is not triggered since this is caught by IRDL (hence the drop in coverage for this PR), but it makes sense in the context of #1727 and my question there regarding `signlessIntegerLike`. I'll probably remove this when that question is answered.

compor:
A quick look guys?

xdsl/dialects/scf.py
```diff
@@ -169,44 +169,54 @@ def get(
         return For(lb, ub, step, iter_args, body)
 
     def verify_(self):
-        if (len(self.iter_args) + 1) != len(self.body.block.args):
+        # body block verification
+        if len(self.body.block.args) == 0:
```
superlopuh, line 173:
```suggestion
        if not self.body.block.args:
```

xdsl/dialects/scf.py
```diff
@@ -169,44 +169,54 @@ def get(
...
             )
-        if self.body.block.args and (iter_var := self.body.block.args[0]):
-            if not isinstance(iter_var.type, IndexType):
+
+        indvar, *block_iter_args = self.body.block.args
+        block_iter_args_num = len(block_iter_args)
+        iter_args = self.iter_args
+        iter_args_num = len(self.iter_args)
+
+        for i, opnd in enumerate(self.operands[:3]):
```
superlopuh, line 183: the :3 is a little scary, maybe something like this?
```suggestion
        for opnd in (self.lb, self.ub, self.step):
```
i also seems not to be used

compor: That's there because the current `signlessInteger` constrained type doesn't catch the case when all signed integers are used, which is a bug and I'm trying to resolve it in #1727.

Not sure what the best course of action is here.

compor: Move the test from #1727 that should fail here?

webmiche (approved):

xdsl/dialects/scf.py
compor: maybe an expectedly failed test?

compor: I made a mistake here; This actually used and exercised by the currently added test cases.

Issue #1118, opened 2023-06-11
core: Allow to use the "generic" dialect attributes syntax


math-fehr:
For instance, `gpu<"dim">` is an allowed syntax for `gpu.dim`.
However, MLIR sometimes only outputs the first version (I am not sure why).
Currently, to make this work, we are using a hack (See `_GpuAttr` attribute)

PR #1726, opened 2023-10-31
core: Implement dialect attribute opaque syntax.

PapyChacal:
I expected it to be bigger!
Do people feel like we should split this? I think the main splittable bit is renaming attributes.

Implement the dialect attribute opaque syntax, i.e., `#arith.fastmath<none>` and `#arith<fastmath<none>>` are the same thing :upside_down_face: 
Adds an OpaqueSyntaxAttribute simple class to inherit from on Attributes meant to be printed using that syntax, just like TypeAttribute for using `!` instead of `#`
More MLIR compliancy!
Leverage that to unhack GPU attributes; Probably some others could be unhacked? More importantly, it's now easier to implement enum attributes :slightly_smiling_face:


xdsl/dialects/builtin.py
```diff
@@ -1116,7 +1116,7 @@ def from_strings(name: str, value: str, type: Attribute = NoneAttr()) -> OpaqueA
 class StridedLayoutAttr(ParametrizedAttribute):
     """
     An attribute representing a strided layout of a shaped type.
-    See https://mlir.llvm.org/docs/Dialects/Builtin/#stridedlayoutattr
+    See https://mlir.llvm.org/docs/Dialects/Builtin/#test.stridedlayoutattr
```
webmiche, line 1119: :thinking:

xdsl/dialects/llvm.py
```diff
@@ -69,7 +69,7 @@
 @irdl_attr_definition
 class LLVMStructType(ParametrizedAttribute, TypeAttribute):
     """
-    https://mlir.llvm.org/docs/Dialects/LLVM/#structure-types
+    https://mlir.llvm.org/docs/Dialects/LLVM/#test.structure-types
```
webmiche, line 72: :thinking:

xdsl/interpreters/comparisons.py
```diff
@@ -19,7 +19,7 @@
 
     We follow LLVM and MLIR in having a concept of signless integers:
 
-    https://mlir.llvm.org/docs/Rationale/Rationale/#integer-signedness-semantics
+    https://mlir.llvm.org/docs/Rationale/Rationale/#builtin.integer-signedness-semantics
```
webmiche, line 22: :thinking:

xdsl/dialects/gpu.py
```diff
@@ -730,7 +690,7 @@ def verify_(self) -> None:
                 )
 
 
-# _GPUAttr has to be registered instead of DimensionAttr and AllReduceOperationAttr here.
+# _GPUAttr has to be registered instead of DimensionAttr and AllReduceOpAttr here.
```
compor, line 693: This comment doesn't seem relevant anymore, right?

xdsl/parser/attribute_parser.py
```diff
@@ -184,24 +184,40 @@ def _parse_dialect_type_or_attribute_inner(
         The contents will be parsed by a user-defined parser, or by a generic parser
         if the dialect attribute/type is not registered.
         """
+        pretty = "." in attr_name
```
compor, line 187: Can you please update the comment in the docstring above based on what is actually parsed now?

compor, line 187: is this format also called "pretty" in MLIR? Just making sure that there is no confusion in terminology between pretty/custom/etc.

xdsl/dialects/gpu.py
```diff
@@ -66,85 +68,43 @@ class AsyncTokenType(ParametrizedAttribute, TypeAttribute):
...
-            vtok = parser.parse_optional_identifier()
-            if vtok not in ["x", "y", "z"]:
-                parser.raise_error(
-                    f"Unexpected dim {vtok}. A gpu dim can only be x, y, or z",
-                )
+    def parse_parameter(
+        cls, parser: AttrParser
+    ) -> Literal["add", "and", "max", "min", "mul", "or", "xor"]:
+        for kw in ("add", "and", "max", "min", "mul", "or", "xor"):
+            val = parser.parse_optional_keyword(kw)
```
superlopuh, line 86: same comment as previous PR, can we peek ahead here instead of parsing k times in a loop?

xdsl/ir/core.py
```diff
@@ -407,6 +392,20 @@ def __str__(self) -> str:
         return res.getvalue()
 
 
+class TypeAttribute(Attribute):
+    """
+    This class should only be inherited by classes inheriting Attribute.
```
compor, line 397: How is this enforced now? It used to be a `__post_init__` check.

superlopuh (changes requested):

tests/filecheck/dialects/builtin/attrs.mlir
```diff
@@ -19,12 +19,12 @@
     %x7 = "arith.constant"() {"value" = 0 : i64, "test" = array<i32: 2, 3, 4>} : () -> i64
     // CHECK:  "test" = array<f32: 2.1, 3.2, 4.3>
     %x8 = "arith.constant"() {"value" = 0 : i64, "test" = array<f32: 2.1, 3.2, 4.3>} : () -> i64
-    // CHECK:  "test" = #signedness<signless>
-    %x9 = "arith.constant"() {"value" = 0 : i64, "test" = #signedness<signless>} : () -> i64
-    // CHECK:  "test" = #signedness<signed>
-    %x10 = "arith.constant"() {"value" = 0 : i64, "test" = #signedness<signed>} : () -> i64
-    // CHECK:  "test" = #signedness<unsigned>
-    %x11 = "arith.constant"() {"value" = 0 : i64, "test" = #signedness<unsigned>} : () -> i64
+    // CHECK:  "test" = #builtin.signedness<signless>
```
compor, line 22: This syntax makes sense to me somehow :)

xdsl/dialects/gpu.py
compor: Moreover, if the implementation could be the same for the parsing of both attributes, it'd be even better.

compor (changes requested):
LGTM and I don't have an issue applying this wholesale as the changes are coherent and relatively small despite affecting so many files.

Lastly, I think this would resolve #1118 if I'm not mistaken, right?

xdsl/ir/core.py
superlopuh: I think just by subclassing Attribute

xdsl/dialects/builtin.py
PapyChacal: Oopsie 😅

PapyChacal:
Let's check with @math-fehr, as I'm not sure about the quotes there?

georgebisbas (approved):

xdsl/dialects/gpu.py
math-fehr: We can just "parse_bare_ident", and then compare it to the elements in the list.

xdsl/dialects/gpu.py
```diff
@@ -66,85 +68,43 @@ class AsyncTokenType(ParametrizedAttribute, TypeAttribute):
...
+            if val in ("x", "y", "z"):
+                return val
+        parser.raise_error("Expected x, y or z.")
 
 
-DimensionAttr = _GPUAttr[_DimensionAttr]
-AllReduceOperationAttr = _GPUAttr[_AllReduceOperationAttr]
+_Element = TypeVar("_Element", bound=Attribute, covariant=True)
 
 _Element = TypeVar("_Element", bound=Attribute, covariant=True)
```
math-fehr, line 109: Duplicated line here?

xdsl/ir/core.py
math-fehr: This is not really enforced anymore after this change I believe :/
The main thing is that `Attribute` classes should either be `Data` or `ParametrizedAttribute`, and we should probably just enforce that in general.
I'm fine removing the check here, and having a check somewhere else at some point.

xdsl/ir/core.py
```diff
@@ -407,6 +392,20 @@ def __str__(self) -> str:
...
+    This class is only used for printing attributes in the MLIR format,
+    inheriting this class prefix the attribute by `!` instead of `#`.
+    """
+
+    pass
+
+
+class OpaqueSyntaxAttribute(Attribute):
+    pass
```
math-fehr, line 406: Missing comment here ;)

xdsl/parser/attribute_parser.py
math-fehr: I would probably change the name of the function as well.
Probably having both a `parse_pretty_dialect_type_inner`, and a `parse_opaque_dialect_type_inner`?
That should make things clearer.

math-fehr: Yep!

math-fehr: You probably want a function like `_parrse_attribute_parameters` that will dispatch to the right function as well, that should make things clearer.

xdsl/parser/attribute_parser.py
```diff
@@ -184,24 +184,40 @@ def _parse_dialect_type_or_attribute_inner(
...
             attr_name,
             create_unregistered_as_type=is_type,
         )
         if attr_def is None:
             self.raise_error(f"'{attr_name}' is not registered")
 
+        attr = None
         # Pass the task of parsing parameters on to the attribute/type definition
         if issubclass(attr_def, UnregisteredAttr):
             body = self._parse_unregistered_attr_body()
```
math-fehr, line 206: Will this work when we print back the opaque unregistered attribute?
How do we know how to print it back with the opaque form?

math-fehr (approved):
Nice!
I'm just a bit worried on how the opaque syntax interact with the unregistered attribute, but that we can figure out later I think!

xdsl/ir/core.py
compor: Maybe track this with a gh issue?

xdsl/dialects/gpu.py
PapyChacal: @compor: The is definitely the next step ;) I want to implement a litlle helper for enum attributes, and switch the GPU ones to this.

xdsl/ir/core.py
PapyChacal: Right @math-fehr, you mentionned that already and it slipped through, my bad.
I think I would prefer the same check but on Attribute himself then? I'll propose something

PapyChacal: Oh but wait; how was this check enforcing this?
All it did was checking that it was an `Attribute` instance, not specifically a `Data` or `ParameterizedAttribute`.
But, same answer; I'll propose something to check this specifically, at the `Attribute` level.

xdsl/parser/attribute_parser.py
PapyChacal: Good point! I'll add some logic for that.

xdsl/parser/attribute_parser.py
PapyChacal: Happy to do a small refactor!

xdsl/ir/core.py
PapyChacal: Okay, this is already enforced at the IRDL level; just adding a test for it

PapyChacal: Do we have non-IRDL attributes like we have non-IRDL Operations?

PapyChacal: Pushed a simple test for the existing IRDL enforcment. Not sure if it would make more sense at the Attribute level?

xdsl/parser/attribute_parser.py
```diff
@@ -185,37 +206,22 @@ def _parse_dialect_type_or_attribute_inner(
...
-            attr = cast(Data[Any], attr_def(param))
 
-        assert attr is not None, "Attributes are either ParametrizedAttribute or Data."
+        attr = self._parse_dialect_type_or_attribute_body(
+            attr_or_dialect_name, is_type, is_opaque
+        )
 
-        if not pretty:
+        if is_opaque:
+            # print("parsed opaque ", attr)
```
superlopuh, line 224: nit: commented code

superlopuh (approved):

compor (approved):

PR #1727, opened 2023-10-31
dialects: (scf) Allow signless integers as induction variable types for `scf.for`

### LINKED ISSUE #1304: Scf.for does not verify for integer operands yet
It looks like previously (e.g. MLIR 16.0) only index types were supported as lower/upper bound and step variables for scf.for loops.
As per the official documentation https://mlir.llvm.org/docs/Dialects/SCFDialect/#scffor-scfforop it seems that now also integers are supported:
```mlir
// Index case.
scf.for %iv = %lb to %ub step %step {
  ... // body
}
...
// Integer case.
scf.for %iv_32 = %lb_32 to %ub_32 step %step_32 : i32 {
  ... // body
}
```
xdsl does not yet verify for these cases, only index types seem to pass verification

compor:
This PR adds:

- Support for signless integers as induction variable types for `scf.for` (also supported in MLIR)
- Custom parsing and printing for the above
- Tests (including MLIR interoperability tests)

Resolves #1304


xdsl/dialects/builtin.py
```diff
@@ -334,6 +334,12 @@ def __init__(
 i1 = IntegerType(1)
 
 
+SignlessIntegerConstraint = ParamAttrConstraint(
```
compor, line 337: Uncertain about the name and placement of this.
Thanks to @PapyChacal who showed me a similar way to do this.

superlopuh (approved):

xdsl/dialects/scf.py
```diff
@@ -148,9 +155,13 @@ def get(
...
-    lb: Operand = operand_def(IndexType)
-    ub: Operand = operand_def(IndexType)
-    step: Operand = operand_def(IndexType)
+    T = Annotated[
+        Attribute, AnyOf([IndexType, SignlessIntegerConstraint]), ConstraintVar("T")
+    ]
+
+    lb: Operand = operand_def(T)
+    ub: Operand = operand_def(T)
+    step: Operand = operand_def(T)
```
compor, line 164: I initially wanted to do

```suggestion
    step: Operand = operand_def(signlessIntegerLike)
```

from [here](https://github.com/xdslproject/xdsl/blob/79bbb8b2ca2b23c03636bb216206b2193fd74635/xdsl/dialects/arith.py#L39), but that allowed signed ints.
So, now I'm also questioning the intention behind that type alias.

compor: @math-fehr ?

webmiche: I am not quite sure I am getting the question, but your change would make it such that the operands could have different types, whereas the current version enforces them to all be the same (at least AFAIU).

compor: @webmiche my bad, I didn't select the right range of LoCs (updated above); I meant if I change all three to be restricted to `signlessIntegerLike`, signed integers (i.e., `si32`) are allowed.

xdsl/dialects/builtin.py
math-fehr: I would also add
```
SignlessIntegerType: TypeAlias = Annotated[IntegerType, SignlessIntegerConstraint]
```
Then you can use `SignlessIntegerType` in place of `IntegerType` when you need python types instead of python objects.

math-fehr: The placement and name are good I think.

xdsl/dialects/scf.py
```diff
@@ -148,9 +155,13 @@ def get(
 class For(IRDLOperation):
     name = "scf.for"
 
-    lb: Operand = operand_def(IndexType)
-    ub: Operand = operand_def(IndexType)
-    step: Operand = operand_def(IndexType)
+    T = Annotated[
+        Attribute, AnyOf([IndexType, SignlessIntegerConstraint]), ConstraintVar("T")
+    ]
```
math-fehr, line 160: That's where you could write `Annotated[IndexType | SignlessIntegerType, ConstraintVar("T")]`

xdsl/dialects/scf.py
math-fehr: I'm not sure what you mean, but with your current change, you should get what you want.
Which is three operands that have the same type, and that type being either an index type, or a signless integer.

math-fehr (approved):
Nice!

xdsl/dialects/scf.py
webmiche: Hmm, I am not 100% sure anymore about how PyRDL will handle this, so maybe I am just wrong here :sweat_smile: 
But intuitively, changing all of them to `singlessIntegerLike` might make it possible to instantiate one with `i32` and another one with `i64`, which would not make sense to me. Having the type alias to me looks like all of them should be instantiated to the same type.
I might be completely wrong here, I am not up to speed with these things anymore, so feel free to point it out :)

Also, if this change really makes all the types the same, how would you then encode it if you have multiple operands with potentially different types?

xdsl/dialects/scf.py
compor: I tried this and other variations but I get:

```py
TypeError: unhashable type: 'ParamAttrConstraint'
```

from the `typing` module.

compor: Alternatively, I'd move this in `builtin` and name it `AnySignlessIntegerOrIndexType` like in MLIR along with naming the above to `AnySignlessIntegerType` instead of `SignlessIntegerType`.

WDYT @math-fehr

compor: @math-fehr I've done it the way I wrote above.

JosseVanDelm (approved):
Neat!

compor:
I'll merge for now and I can always amend later.

PR #1729, opened 2023-10-31
installation: move non-versioneer settings to pyproject.toml

superlopuh:


compor (approved):

math-fehr (approved):
Nice!

setup.cfg
```diff
@@ -1,25 +1,7 @@
...
 
 [versioneer]
 VCS = git
 style = pep440
 versionfile_source = xdsl/_version.py
 versionfile_build = xdsl/_version.py
 tag_prefix = v
-
-[aliases]
-test = pytest -v
```
webmiche, line 25: Where has this gone?

superlopuh: Aliases in setup.cfg are actually aliases to pytest, not geral commands. So this alias is invoked with `python setup.py test` and is equivalent to `python setup.py pytest -v`, which I don't think actually does anything. I'm not sure why it was added in the first place.

https://stackoverflow.com/questions/63672256/adding-aliases-fails-with-error-invalid-command-xyz

PR #1730, opened 2023-10-31
installation: move versioneer cfg to pyproject.toml

superlopuh:
My understanding is that setup.cfg is borderline legacy, and that we're better off having fewer config files if possible.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [build-system]
-requires = ["setuptools>=42", "wheel"]
+requires = ["setuptools>=42", "wheel", "versioneer[toml]"]
+build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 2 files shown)


math-fehr (approved):
Yes I agree.
We probably also want to move our dependencies here at some point: https://github.com/opencompl/xdsl-smt/blob/main/pyproject.toml

superlopuh:
That's in progress :)

AntonLydike:
We could get rid of all of `setup.py` by switching to poetry. All that would change is that instead of setup being `make venv; source venv/bin/activate`, it now becomes `poetry install; poetry shell`.

compor (approved):

AntonLydike (approved):

webmiche (approved):

PR #1735, opened 2023-11-01
dialects: (linalg) Use abstract yield operation in `linalg.yield`

compor:
This PR:

- Uses the abstract base implementation of Yield in `linalg.yield`
- Rename class from `Yield` to `YieldOp` (white we are at it)


superlopuh (approved):
ty

math-fehr (approved):
Very nice!

webmiche (approved):

PR #1737, opened 2023-11-01
dialects: (scf) Use abstract yield operation in `scf.yield`

compor:
This PR:

- Uses the abstract base implementation of Yield in `scf.yield`

Similar to #1735.


superlopuh (approved):

math-fehr (approved):
Very nice!

webmiche (approved):

PR #1740, opened 2023-11-02
[WIP] interactive: App Draft

dshaaban01:
Using Textual Python, I have created a GUI/app.This app allows you to paste xDSL IR into the "Input" TextArea, select a pass (or multiple passes) to be applied on the input IR and subsequently the IR generated by the application of the selected pass(es) is displayed in the "Output" TextArea. The selected passes are displayed in the top right "Selected passes/query" box. The "Condense" button filters the pass selection list to display passes that change the IR, or require arguments to be executed (and thus "may" change the IR). The "Uncondense" button returns the selection list to contain all the passes. The "Clear Passes" Button removes the application of the selected passes. The "Copy Query" Button allows the user to copy the selected passes/query that they have so far selected (i.e. copy the top right box).

This app is still under construction.

Immediate To Do's:
1. Add the ability to add arguments to the passes that need them 
2. Extend "Condense Button" functionality to analyze current state (i.e. output - passes applied) and not just the static input IR
3. Make all buttons visible (currently need to expand terminal to see all the buttons)

image

---

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -96,10 +96,7 @@ class InputApp(App[None]):
 
         my_theme = TextAreaTheme(
             name="my_cool_theme",
-            # Basic styles such as background, cursor, selection, gutter, etc...
             base_style=Style(bgcolor="white"),
-            # `syntax_styles` is for syntax highlighting.
-            # It maps tokens parsed from the document to Rich styles.
             syntax_styles={
                 "string": Style(color="red"),
                 "comment": Style(color="magenta"),


xdsl/ir/core.py
superlopuh, line 1: can you please open a separate PR for this?

requirements-optional.txt
superlopuh, line 1: can you please add textual and textual-dev to this file?

xdsl/interactive/app.py
```diff
@@ -0,0 +1,256 @@
+from collections.abc import Callable
+from io import StringIO
+
+# pyright: ignore[reportMissingTypeStubs, reportGeneralTypeIssues]
+from pyclip import copy as pyclip_copy
```
superlopuh, line 5: Please put the comment back on the same line, the CI is not picking this up

xdsl/interactive/app.py
```diff
@@ -0,0 +1,256 @@
...
+from textual.widgets import Button, Label, SelectionList, TextArea
+
+from xdsl.dialects import builtin
+from xdsl.ir import MLContext
+from xdsl.parser import Parser
+from xdsl.passes import ModulePass, PipelinePass
+from xdsl.printer import Printer
+from xdsl.tools.command_line_tool import get_all_dialects, get_all_passes
+
+"""
```
superlopuh, line 18: Please put the doc string to the top of the file

xdsl/interactive/app.py
```diff
@@ -0,0 +1,256 @@
...
+returns the selection list to contain all the passes. The "Clear Passes" Button removes the application of the selected passes.
+The "Copy Query" Button allows the user to copy the selected passes/query that they have so far selected (i.e. copy the top right
+box).
+
+This app is still under construction.
+"""
+pyclip_copy: Callable[[str], None] = pyclip_copy
+
+
+# Function that takes the input IR, the list of passes to be applied, and applies the passes to the IR.
```
superlopuh, line 32: Python doc strings are defined by putting a string literal as the first expression in the file, could you please move this comment down into the body?

superlopuh: Similarly below

xdsl/interactive/app.py
```diff
@@ -0,0 +1,256 @@
...
+    def on_undo_condense_button_pressed(self, event: Button.Pressed) -> None:
+        self.module_passes_selection_list.clear_options()
+        passes = get_all_passes()
+        selections = [(value.name, value) for value in passes]
+        selections.sort()
+        self.module_passes_selection_list.add_options(selections)
+
+
+if __name__ == "__main__":
+    path = "tests/filecheck/backend/riscv/canonicalize.mlir"
```
superlopuh, line 252: this hardcoded path should be removed

xdsl/interactive/app.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+# Used to prevent users from being able to change/alter the Output TextArea
+class OutputTextArea(TextArea):
+    async def _on_key(self, event: events.Key) -> None:
+        event.prevent_default()
+
+
+# Class buildling the Experimental Interactive Compilation App.
+class InputApp(App[None]):
```
superlopuh, line 67: Textual encourages state to be held in reactive variables on the App and Widget classes, and the UI to be derived from those values as they change. Could you please transition to that system?

superlopuh: This should also make it easier to evolve the app over time

superlopuh (changes requested):
I'm really excited about this, but would love for it to be merged cleanly, as a fully-tested component of xDSL. I don't want that to block progress on your thesis project, but there's a number of changes that I think would be important to change in order to get it into the `main` branch. It would be good to discuss what the minimal version of this app is that we want to merge, and work towards covering all the usual things we do when merging code into xDSL. My main worry is that it'll be much more difficult to change things after merging than in small, incremental PRs.

I left some comments on this PR, and those would be great to address, but also let's chat about the broader points on the Zulip.

webmiche (commented):
Cool, great to see this PR up and the project making progress!

I absolutely agree with Sasha here, we should not block you, but still upstream this with our normal quality checks. So, do we want to have this discussion here or on Zulip? I have several ideas how to improve this in terms of software engineering, but I don't want to make this a case of PR creep :wink:

xdsl/interactive/app.py
```diff
@@ -0,0 +1,250 @@
...
+from xdsl.printer import Printer
+from xdsl.tools.command_line_tool import get_all_dialects, get_all_passes
+
+pyclip_copy: Callable[[str], None] = pyclip_copy
+
+
+def transform_input(
+    input_text: str, passes: list[type[ModulePass]]
+) -> builtin.ModuleOp:
+    # Function that takes the input IR, the list of passes to be applied, and applies the passes to the IR.
```
compor, line 36: Maybe convert to docstrings?

xdsl/interactive/app.py
```diff
@@ -0,0 +1,250 @@
...
+from xdsl.ir import MLContext
+from xdsl.parser import Parser
+from xdsl.passes import ModulePass, PipelinePass
+from xdsl.printer import Printer
+from xdsl.tools.command_line_tool import get_all_dialects, get_all_passes
+
+pyclip_copy: Callable[[str], None] = pyclip_copy
+
+
+def transform_input(
```
compor, line 33: I don't want to suggest MVC and the likes but I feel it would make sense to have those calls organized separately in order for people to see more easily what is required by xDSL "core" to drive this.

I'm not sure if this separation would help plug this system later on with a browser or something else.
But maybe something to consider without derailing current progress.

xdsl/interactive/app.py
```diff
@@ -0,0 +1,250 @@
...
+
+    return module
+
+
+def check_if_pass_changes_IR(
+    input: builtin.ModuleOp, ctx: MLContext, one_pass: ModulePass
+) -> bool:
+    # Used to check if a pass has had an effect on the IR (i.e. has changed the IR)
+    cloned_module = input.clone()
+    cloned_ctx = ctx.clone()
```
compor, line 56: Why do we need to clone here? That is also wrt #1743.

xdsl/interactive/app.py
```diff
@@ -0,0 +1,250 @@
...
+    parser = Parser(ctx, input_text)
+    module = parser.parse_module()
+
+    pipeline = PipelinePass([p() for p in passes])
+    pipeline.apply(ctx, module)
+
+    return module
+
+
+def check_if_pass_changes_IR(
```
compor, line 51: Would it make sense to use the `PipelinePass` for this and use its callback to perform this check here?

compor (commented):
I like the direction!
I left some comments, but again at this point are broad things to consider, so maybe not very actionable till you move further with this prototype!

xdsl/interactive/app.py
superlopuh: The passes can (and should, and often do in MLIR) modify the context, so it's another bit of state that should not change when tentatively trying passes

xdsl/interactive/app.py
superlopuh: What di you mean by "core"?

xdsl/interactive/app.py
compor: Thanks, I thought so (I couldn't find a use case in our codebase), I'll have a look in MLIR sometime later.

tobiasgrosser:
It might also be a good idea to have a work-in-progress PR where dalia just pushes without ever having an intention to upstream things. And then spin of feature PRs that bring things step-by-step into xDSL. As this project matures and the upstreaming goes faster and faster, I would hope that the actual delta of these branches shrinks and @dshaaban01 can then work mostly on upstream.

PR #1743, opened 2023-11-02
api: added clone function in core.py

dshaaban01:
Opened a new branch for the clone() function added in core.py - originally in the interactive app PR (as requested by @superlopuh). To-do: add tests.


xdsl/ir/core.py
```diff
@@ -63,9 +63,17 @@ class MLContext:
...
+    _loaded_dialects: dict[str, Dialect] = field(default_factory=dict)
+    _loaded_ops: dict[str, type[Operation]] = field(default_factory=dict)
+    _loaded_attrs: dict[str, type[Attribute]] = field(default_factory=dict)
+
+    def clone(self) -> MLContext:
+        return MLContext(
+            self.allow_unregistered,
+            self._loaded_dialects,
+            self._loaded_ops,
+            self._loaded_attrs,
```
webmiche, line 75: Isn't this a classic case of capturing? So, if I were to load more operations in the cloned context, I would also modify the original. I don't think we should implement it like this. I would prefer a deep clone in this first instance, and maybe later a shallow clone, if we feel this is needed.

webmiche (dismissed):

xdsl/ir/core.py
superlopuh: Good point, I think we can get away with just creating a new list for each of these.

webmiche: I guess a couple of `.copy()`s might do the trick?

superlopuh: sure, not sure what the difference would be between `list(self._loaded_dialects)` and `self._loaded_dialects.copy()` would be in this case... I guess if the type changes from list to something else then the copy would still work but the list would fail, so copy is better.

compor: I agree with @webmiche. Maybe it's too early to see what is required by "core".
If it is something to consider for the interactive project, maybe tracking these TODOs/FIXMEs/etc in a gh project? Not sure what to suggest without impeding progress for @dshaaban01

superlopuh: I would say that this is a useful thing to have independently of the interactive stuff, we should be able to copy an MLContext IMO

xdsl/ir/core.py
```diff
@@ -559,7 +567,8 @@ def __setitem__(self, idx: int, operand: SSAValue) -> None:
         operands[idx].remove_use(Use(self._op, idx))
         operand.add_use(Use(self._op, idx))
         new_operands = (*operands[:idx], operand, *operands[idx + 1 :])
-        self._op._operands = new_operands  # pyright: ignore[reportPrivateUsage]
+        # pyright: ignore[reportPrivateUsage]
```
superlopuh, line 570: can you please undo this change?

superlopuh (changes requested):

superlopuh (approved):

superlopuh:
@webmiche is OOA for a while, would be good to get a second tick before you merge, maybe @PapyChacal, @AntonLydike or @math-fehr will get a chance to review

compor (approved):

compor (approved):

math-fehr:
I think I would need context on why this is actually needed?
Why do you want to clone an `MLContext`? They don't store any information besides the loaded dialects, so why do you want to copy them?

superlopuh:
This is for filtering passes that modify the IR in the gui. Passes can modify the context, so it needs to be cloned for each pass that is tried, in case one of the passes changes the ctx in an incompatible way with another one.

math-fehr:
> This is for filtering passes that modify the IR in the gui. Passes can modify the context, so it needs to be cloned for each pass that is tried, in case one of the passes changes the ctx in an incompatible way with another one.

Okay, I don't think any pass should make a change in the context, but in your case let's add that in case in the future we would allow this.

math-fehr:
Also, just so you know, this will lead to very weird behaviors with unregistered attributes/ops.
The class for an unregistered op will be different for two different `MLContext`.

math-fehr (approved):

PR #1744, opened 2023-11-02
misc: move dependencies out of requirements-optional

superlopuh:
After reading up a bit on the recommended uses of `requirements.txt`, `pyproject.toml`, and `setup.py`, I have come to this understanding:

`requirements.txt` is supposed to be used only for local development. The understanding is that it is a document that is pinned to the local venv of the developers of the project. It is supposed to be more strict and specific than the packaging for pypi publishing, which is supposed to be more permissive.

`pyproject.toml`/`setup.py` are supposed to specify the ranges of version of dependencies for clients of xdsl that `pip install xdsl` from their own projects.

It is not recommended to couple the two concepts to each other.

This PR shuffles the same set of dependencies around between files. `requirements.txt` now holds all the developer dependencies for testing and linting. People who `pip install xdsl` should not care that their version of pytest is not the same as ours. `pyproject.toml` now holds the info of the required dependencies (immutabledict, typing-extensions, ordered-set), and "extras" dependencies (riscemu, wgpu).

The new command to install xdsl for local development is `pip install -r requirements.txt` (or `make venv`).

I've tested that `pip install -e .` produces exactly the same output from `pip list` before and after, and the command above installs exactly the same packages as `pip install -e ".[extras]"` before. Now the same command skips the testing projects, which I think is good.

Here are some links for future reference:

https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
https://peps.python.org/pep-0508/
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html


requirements.txt
```diff
@@ -1,4 +1,19 @@
...
+lit<18.0.0
+pre-commit==3.5.0
+ruff==0.1.3
+asv<0.7
+isort==5.12.0
+nbconvert>=7.7.2,<8.0.0
+# pyright version has to be fixed with `==`. The CI parses this file
+# and installs the according version for typechecking.
+pyright==1.1.333
+-e .[extras]
```
webmiche, line 19: I am surprised, does this actually do what I think it does? Man, I hate python at times xD

.github/workflows/ci-pyright-fails.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         run: pip install -e ".[extras]"
```
webmiche, line 34: I think you forgot a line above this, right? That's why it did not find the pytest dependency.

requirements.txt
```diff
@@ -1,4 +1,19 @@
...
+filecheck<0.0.24
+lit<18.0.0
+pre-commit==3.5.0
+ruff==0.1.3
+asv<0.7
+isort==5.12.0
+nbconvert>=7.7.2,<8.0.0
+# pyright version has to be fixed with `==`. The CI parses this file
+# and installs the according version for typechecking.
+pyright==1.1.333
```
webmiche, line 18: Where is the wgpu dependency? :thinking:

webmiche (changes requested):
That sounds a lot cleaner, thanks for changing this!

requirements.txt
superlopuh: It's in the `pyproject.toml`, along with riscemu

requirements.txt
superlopuh: This time it's `pip` not `python` that you should be unhappy with :)

requirements.txt
superlopuh: in `[project.optional-dependencies]`

webmiche (approved):

compor (approved):

compor:
Thanks for the writeup.

PapyChacal (approved):
Lovely, thanks for inversting time in there!

setup.py
```diff
@@ -1,66 +1,12 @@
-import re
 from collections.abc import Mapping
-from pathlib import Path
 from typing import cast
 
 from setuptools import Command, find_packages, setup
 
 import versioneer
 
-# Add README.md as long description
```
georgebisbas, line 10: Will these changes drop the description etc from pypi?

georgebisbas:
I am not sure whether the project is in a state to distinguish between users-only/customers/developers.
To my understanding, most of the current users, contribute a few things as well.

math-fehr (approved):

setup.py
superlopuh: I actually moved the descriptions in a different PR. They're now specified by the following two lines in pyproject.toml:
``
description = "xDSL"
readme = "README.md"
```
This diff just removes some dead code that I forgot to remove in the previous PR.

superlopuh:
That's fair, here I'm using a bit of a weird definition of "users" and "contributors" as non-exclusive. "user" is someone who installs without cloning the repo, and "contributor" is someone who installs by cloning the repo.

PR #1748, opened 2023-11-04
misc: improve get_type_var_mapping.

PapyChacal:
Some experiments on improving generic helpers we offer. Still not crystal clear how I plan to use them, I have something in mind, will stack if it works!

For example, currently `get_type_var_from_generic_class` (used in `get_type_var_mapping`) is - to me - overprotective, and does not allow any Generic specialization not done directly at the Generic child level; i.e: (see added test)
```python
T = TypeVar("T")
R = TypeVar("R")

class GenericBase(Generic[T]):
    pass

class GenericSubclassOfGenericBase(GenericBase[R]):
    pass

class SpecificSubclassOfGenericSubclass(GenericSubclassOfGenericBase[int]):
    pass
```
```
E       ValueError: <class 'test_hints.SpecificSubclassOfGenericSubclass'> class does not have a `Generic` parent.
```
This PR adds some logic to return
```python
(SpecificSubclassOfGenericSubclass, {R: int})
```
in that case, which feels closer to the expected behavior to me?
I add support for GenericAliases too.

I'd love to get the information `T : int` out of it too.

### DIFF
--- a/tests/utils/test_hints.py
+++ b/tests/utils/test_hints.py
@@ -5,13 +5,14 @@ import pytest
 from xdsl.utils.hints import get_type_var_mapping
 
 T = TypeVar("T")
+R = TypeVar("R")
 
 
 class GenericBase(Generic[T]):
     pass
 
 
-class GenericSubclassOfGenericBase(GenericBase[T]):
+class GenericSubclassOfGenericBase(GenericBase[R]):
     pass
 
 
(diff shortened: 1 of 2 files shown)


tests/utils/test_hints.py
```diff
@@ -0,0 +1,36 @@
...
+class GenThingy(GenThing[T]):
+    pass
+
+
+class IntThing(GenThing[int]):
+    pass
+
+
+class IntThingy(GenThingy[int]):
+    pass
```
superlopuh, line 23: Can you please add more descriptive names to these classes, so it's clear what the purpose is of each part of the generics hierarchy? I'll have a go at naming them to show my current understanding, please correct me if I'm wrong:

```suggestion
class GenericBase(Generic[T]):
    pass


class GenericSubclassOfGenericBase(GenericBase[T]):
    pass


class SpecificSubclassOfGenericBase(GenericBase[int]):
    pass


class SpecificSubclassOfGenericSubclass(GenericSubclassOfGenericBase[int]):
    pass
```

superlopuh:
What is it used for? Why would you like to change it?

PapyChacal:
Was an interesting rabbit hole, but it wouldn't be pragmatic to keep improving that for my more immediate goals, closing for now.
Interested to chat about it if someone face issues that this would help with! Or juts out of curiosity, of course :slightly_smiling_face:

PR #1749, opened 2023-11-04
core: implement EnumAttribute

PapyChacal:
Core helper for Enum Attributes. Takes a StrEnum type parameter, and defines parsing/printing automatically from its values, restricted to be parsable as identifiers.
example:
```python
class MyEnum(StrEnum):
    First = auto()
    Second = auto()

class MyEnumAttribute(EnumAttribute[MyEnum], OpaqueSyntaxAttribute):
    name = "example.my_enum"
```
Is enough to use this attribute, having a textual representation of `example<my_enum first>` and `example<my_enum second>`.

The main implementation trick is `__init_subclass__`:
This hook first checks two constraints, enforced to keep the implementation reasonable, until more complex usecases appears. It then store the Enum type used by the subclass to use in parsing/printing.
The constraints are:
- Only direct, specialized inheritance is allowed. That is, using a subclass of EnumAttribute as a base class is *not supported*.
  This simplifies type-hacking code and I don't see it being too restrictive anytime soon?
- The StrEnum values must all be parsable as identifier. This is to keep the parsing code simple and efficient. This restriction seems easier to lift, but I didn't yet meet an example use case where it matters, so I'm keeping it simple.

### DIFF
--- a/tests/test_attribute_definition.py
+++ b/tests/test_attribute_definition.py
@@ -148,6 +148,12 @@ class TestEnum(StrEnum):
 
 
 class TestNonIdentifierEnum(StrEnum):
+    """
+    The value defined by this StrEnum is not parsable as an identifier, because of the
+    contained space.
+    While valid as a StrEnum, it is thus invalid to use it in an EnumAttribute.
+    """
+
     Spaced = "left right"
 
 
--- a/tests/test_attribute_definition.py
+++ b/tests/test_attribute_definition.py
@@ -5,7 +5,7 @@ Test the definition of attributes and their constraints.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from enum import StrEnum, auto
+from enum import auto
 from io import StringIO
 from typing import Annotated, Any, Generic, TypeAlias, TypeVar, cast
 
@@ -18,6 +18,7 @@ from xdsl.ir import (
     EnumAttribute,
     OpaqueSyntaxAttribute,
     ParametrizedAttribute,
+    StrEnum,
 )
 from xdsl.irdl import (
     AnyAttr,
--- a/tests/test_attribute_definition.py
+++ b/tests/test_attribute_definition.py
@@ -5,13 +5,20 @@ Test the definition of attributes and their constraints.
 from __future__ import annotations
 
 from dataclasses import dataclass
+from enum import StrEnum, auto
 from io import StringIO
 from typing import Annotated, Any, Generic, TypeAlias, TypeVar, cast
 
 import pytest
 
 from xdsl.dialects.builtin import IndexType, IntegerAttr, IntegerType, Signedness
-from xdsl.ir import Attribute, Data, ParametrizedAttribute
+from xdsl.ir import (
+    Attribute,
+    Data,
+    EnumAttribute,
+    OpaqueSyntaxAttribute,
+    ParametrizedAttribute,
+)
 from xdsl.irdl import (
     AnyAttr,
     AttrConstraint,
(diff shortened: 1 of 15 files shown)


PapyChacal:
Didn't realize that `StrEnum` is Python 3.11 :smiling_face_with_tear: For that use case, should be only a few more lines to do with bare Enum..

xdsl/ir/core.py
```diff
@@ -456,6 +462,96 @@ def print_parameter(self, printer: Printer) -> None:
...
+        printer.print(" ", self.data.value)
+
+    @final
+    @classmethod
+    def parse_parameter(cls, parser: AttrParser) -> EnumType:
+        enum_type = cls.enum_type
+
+        val = parser.parse_identifier()
+        if val in enum_type.__members__.values():
+            return cast(EnumType, enum_type(val))
```
superlopuh, line 546: super nit: it feels a little weird to have the success condition not be at the bottom of the function. If you feel like doing it, I feel that it would be a little clearer to invert the if condition and raise the error in the if block. But not a big deal

superlopuh (approved):
very nice

xdsl/ir/core.py
PapyChacal: Agreed and I don't think it's a big deal, will do 🙂

xdsl/ir/core.py
```diff
@@ -456,6 +462,96 @@ def print_parameter(self, printer: Printer) -> None:
...
+
+    @staticmethod
+    def _generate_next_value_(
+        name: str, start: int, count: int, last_values: list[Any]
+    ):
+        return name.lower()
+
+    def __str__(self) -> str:
+        return self.value
+
```
math-fehr, line 478: Could we move that to xdsl.utils?

math-fehr (approved):
Very very cool!

tests/test_attribute_definition.py
```diff
@@ -134,6 +142,53 @@ def test_non_class_data():
...
+        class IndirectEnumData(  # pyright: ignore[reportUnusedClass]
+            EnumAttribute[EnumType]
+        ):
+            name = "test.indirect_enum"
+
+
+def test_identifier_enum_guard():
+    with pytest.raises(
+        ValueError,
+        match="All StrEnum values of an EnumAttribute must be parsable as an identifer.",
```
webmiche, line 183: Is this wrong because of the space in between? Could you maybe point this out as a comment on this enum?

webmiche (approved):
That's a very cool feature, thanks!

xdsl/ir/core.py
```diff
@@ -456,6 +462,96 @@ def print_parameter(self, printer: Printer) -> None:
         """Print the attribute parameter."""
 
 
+class StrEnum(str, Enum):
+    """
+    Homemade StrEnum. StrEnum is standard in Python>=3.11.
```
compor, line 467: Can we add a deprecation warning if Python 3.11 and higher is used? It will help with version transition when we actually do so.

xdsl/ir/core.py
```diff
@@ -456,6 +462,96 @@ def print_parameter(self, printer: Printer) -> None:
...
+    ```python
+    class MyEnum(StrEnum):
+        First = auto()
+        Second = auto()
+
+    class MyEnumAttribute(EnumAttribute[MyEnum], OpaqueSyntaxAttribute):
+        name = "example.my_enum"
+    ```
+    Is enough to use this attribute, having a textual representation
+    of `example<my_enum first>` and ``example<my_enum second>``
```
compor, line 499: This reads weird:
```suggestion
    To use this attribute suffices to have a textual representation
    of `example<my_enum first>` and ``example<my_enum second>``
```

xdsl/ir/core.py
```diff
@@ -456,6 +462,96 @@ def print_parameter(self, printer: Printer) -> None:
...
+        The constraints are:
+
+        - Only direct, specialized inheritance is allowed. That is, using a subclass
+        of EnumAttribute as a base class is *not supported*.
+          This simplifies type-hacking code and I don't see it being too restrictive
+          anytime soon.
+        - The StrEnum values must all be parsable as identifier. This is to keep the
+        parsing code simple and efficient. This restriction is easier to lift, but I
+        didn't yet meet an example use case where it matters, so I'm keeping it simple.
+        """
```
compor, line 520:
```suggestion
        This hook first checks two constraints, enforced to keep the implementation
        reasonable, until more complex use cases appear. It then stores the Enum type
        used by the subclass to use in parsing/printing.

        The constraints are:

        - Only direct, specialized inheritance is allowed. That is, using a subclass
        of EnumAttribute as a base class is *not supported*.
          This simplifies type-hacking code and I don't see it being too restrictive
          anytime soon.
        - The StrEnum values must all be parsable as identifiers. This is to keep the
        parsing code simple and efficient. This restriction is easier to lift, but I
        haven't yet met an example use case where it matters, so I'm keeping it simple.
        """
```

compor (approved):
great refactoring!

xdsl/ir/core.py
superlopuh: I don't think that'll work, my current understanding is that we support a range of versions, including 3.10 and 3.11, and would like the deprecation warning to show in neither of them. I don't know if there's a way to warn only when we drop support for 3.10. I expect we'll do an audit of the codebase when that happens (in a year or two).

compor: OK then!

PR #1752, opened 2023-11-05
dialects: implement affine.apply

PapyChacal:
Mirror affine.apply from MLIR

### DIFF
--- a/xdsl/dialects/affine.py
+++ b/xdsl/dialects/affine.py
@@ -31,6 +31,24 @@ from xdsl.irdl import (
     var_result_def,
 )
 from xdsl.traits import IsTerminator
+from xdsl.utils.exceptions import VerifyException
+
+
+@irdl_op_definition
+class Apply(IRDLOperation):
+    name = "affine.apply"
+
+    mapOperands = var_operand_def(IndexType)
+    map = attr_def(AffineMapAttr)
+    result = result_def(IndexType)
+
+    def verify_(self) -> None:
+        if len(self.mapOperands) != self.map.data.num_dims + self.map.data.num_symbols:
+            raise VerifyException(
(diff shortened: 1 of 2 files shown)


superlopuh (approved):

math-fehr (approved):

georgebisbas (approved):

xdsl/dialects/affine.py
```diff
@@ -25,12 +25,31 @@
...
     var_operand_def,
     var_result_def,
 )
 from xdsl.traits import IsTerminator
+from xdsl.utils.exceptions import VerifyException
+
+
+@irdl_op_definition
+class Apply(IRDLOperation):
+    name = "affine.apply"
```
compor, line 40:
```suggestion
class ApplyOp(IRDLOperation):
    name = "affine.apply"
```

xdsl/dialects/affine.py
```diff
@@ -25,12 +25,31 @@
...
+
+    mapOperands = var_operand_def(IndexType)
+    map = prop_def(AffineMapAttr)
+    result = result_def(IndexType)
+
+    def verify_(self) -> None:
+        if len(self.mapOperands) != self.map.data.num_dims + self.map.data.num_symbols:
+            raise VerifyException(
+                "affine.apply expects as many operands as its map's dimensions and symbols."
```
compor, line 49: Can we make this error message more descriptive? Something like (adjust accordingly):

```suggestion
                f"{self.name} expects {self.map.data.num_dims + self.map.data.num_symbols} operands, but got len(self.mapOperands). The number of map operands must match the sum of the dimensions and symbols of its map."
```

compor: Same for below

compor (approved):

PR #1756, opened 2023-11-05
dialects: Implement memref.alloca_scope and memref.alloca_scope.return.

PapyChacal:
Mirror `memref.alloca_scope` and `memref.alloca_scope.return` from MLIR.


superlopuh (approved):
please add interop test

math-fehr (approved):

georgebisbas (approved):

PR #1759, opened 2023-11-07
interactive: App (1/7)

dshaaban01:
Using Textual Python, I have created a GUI/app.This app allows you to write/paste3 xDSL IR into the "Input" TextArea, which is mirrored in the "Output" TextArea.

This app is still under construction, this is the first step towards building the full tool.

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -114,10 +114,10 @@ class InputApp(App[None]):
         """Configure widgets in this application before it is first shown."""
 
         # register's the theme for the Input/Output TextArea's
-        self.input_text_area.theme = "my_theme_design"
-        self.output_text_area.theme = "my_theme_design"
         self.input_text_area.register_theme(InputApp._DEFAULT_THEME)
         self.output_text_area.register_theme(InputApp._DEFAULT_THEME)
+        self.input_text_area.theme = "my_theme_design"
+        self.output_text_area.theme = "my_theme_design"
 
         self.query_one("#input_container").border_title = "Input xDSL IR"
         self.query_one("#output_container").border_title = "Output xDSL IR"


compor:
how does this PR relate to #1740 ?

xdsl/ir/core.py
superlopuh, line 1: please remove this diff

xdsl/interactive/app.tcss
```diff
@@ -0,0 +1,40 @@
...
+    # overflow: auto auto;
+
+}
+
+# Horizontal(Vertical(OutputTextArea), Vertical(OutputTextArea))
+#input_output{
+}
+
+Screen {
+    # layout: grid;
```
superlopuh, line 37: please remove all commented lines and unused css classes

xdsl/interactive/app.tcss
```diff
@@ -0,0 +1,40 @@
+
+# TextArea
+#input{
```
superlopuh, line 3:
```suggestion
#input {
```
Please add a space before the curly brackets, makes it a bit easier to read

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
...
+        """An action to toggle dark mode."""
+        self.dark = not self.dark
+
+    def action_quit_app(self) -> None:
+        """An action to quit the app."""
+        self.exit()
+
+
+if __name__ == "__main__":
+    app = InputApp(None)
```
superlopuh, line 139: Why the None here?

.vscode/launch.json
```diff
@@ -24,5 +24,12 @@
...
+        {
+            "name": "interactive",
+            "type": "python",
+            "request": "launch",
+            "module": "xdsl.interactive.app",
+            "justMyCode": false,
+        },
     ]
-}
+}
```
superlopuh, line 35:
```suggestion
}

```

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
+"""
+This app allows you to paste xDSL IR into the Input TextArea, select a pass (or multiple passes) to be applied on the input IR
+and subsequently the IR generated by the application of the selected pass(es) is displayed in the Output TextArea. The selected
+passes are displayed in the top right "Selecred passes/query" box. The "Condense" button filters the pass selection list to
+display passes that change the IR, or require arguments to be executed (and thus "may" change the IR). The "Uncondense" button
+returns the selection list to contain all the passes. The "Clear Passes" Button removes the application of the selected passes.
+The "Copy Query" Button allows the user to copy the selected passes/query that they have so far selected (i.e. copy the top right
+box).
+
+This app is still under construction.
```
superlopuh, line 10: `xDSL` is under construction, I think this file inherits that property :)

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
+"""
+This app allows you to paste xDSL IR into the Input TextArea, select a pass (or multiple passes) to be applied on the input IR
```
superlopuh, line 2: This is a great description of the current behaviour in the other PR, but may quickly become outdated as the code is changed. I would recommend describing the purpose/overall description of the contents of this file, rather than a description of the behaviour. Maybe something along these lines:

```
An interactive command-line tool to explore compilation pipeline construction.

Run `terminal -m xdsl.interactive.app:InputApp --def` to run in development mode. Please be sure to install `textual-dev` to run this command.
```

Another thing I forgot to suggest is to install this as a script along with `xdsl-opt` and `xdsl-run`, which we may discuss in this PR or an upcoming one. Maybe best left for later.

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
...
+
+    async def _on_key(self, event: events.Key) -> None:
+        event.prevent_default()
+
+
+class InputApp(App[None]):
+    """
+    Class buildling the Interactive Compilation App. Uses Textual Python to construct reactive variables on the App, Event,
+    Widget classes etc. The UI is derived from those values as they change.
+    """
```
superlopuh, line 41:
```suggestion
    """
    Interactive application for constructing compiler pipelines.
    """
```

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def compose(self) -> ComposeResult:
+        """
+        Creates the required widgets, events, etc.
+        Get the list of xDSL passes, add them to an array in "Selection" format (so it can be added to a Selection List)
+        and sort the list in alphabetical order.
+        """
+
+        # defines a theme for the Input/Output TextArea's
+        my_theme = TextAreaTheme(
```
superlopuh, line 66: This constant makes sense at the top level. In python, the convension is `UPPER_SNAKE_CASE` for constants. I would recommend calling it `_DEFAULT_TEXT_AREA_THEME` at the top level.

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
...
+            name="my_theme_design",
+            base_style=Style(bgcolor="white"),
+            syntax_styles={
+                "string": Style(color="red"),
+                "comment": Style(color="magenta"),
+            },
+        )
+
+        # register's the theme for the Input/Output TextArea's
+        self.input_text_area.register_theme(my_theme)
```
superlopuh, line 76: please move this to `on_mount`

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
...
+                output_text = output_stream.getvalue()
+            case ModuleOp():
+                output_stream = StringIO()
+                Printer(output_stream).print(self.current_module)
+                output_text = output_stream.getvalue()
+
+        self.output_text_area.load_text(output_text)
+
+    def on_mount(self) -> None:
+        """On App Mount, add titles"""
```
superlopuh, line 125:
```suggestion
        """Configure widgets in this application before it is first shown."""
```

superlopuh (changes requested):
Looking great! A few changes to make.

PapyChacal (approved):
Looking forward for the rest of the application!

xdsl/interactive/app.py
superlopuh: I agree, the top-level command is best left for later, but it would be good for this documentation to already be a little more instructive to people who come across this file without seeing this PR.

xdsl/interactive/app.tcss
```diff
@@ -0,0 +1,16 @@
...
+    border-title-align: center;
+}
+
+# Vertical(OutputTextArea)
+#output_container {
+    margin: 1;
+    border: heavy $warning-lighten-1;
+    border-title-color: $primary;
+    border-title-align: center;
+}
```
superlopuh, line 16:
```suggestion
}

```

xdsl/interactive/app.py
```diff
@@ -0,0 +1,140 @@
...
+        self.output_text_area.load_text(output_text)
+
+    def on_mount(self) -> None:
+        # register's the theme for the Input/Output TextArea's
+        self.input_text_area.register_theme(InputApp._DEFAULT_THEME)
+        self.output_text_area.register_theme(InputApp._DEFAULT_THEME)
+        self.input_text_area.theme = "my_theme_design"
+        self.output_text_area.theme = "my_theme_design"
+
+        """Configure widgets in this application before it is first shown."""
```
superlopuh, line 125: Please move this to the top of the function

superlopuh (approved):
🚀🚀🚀🚢

math-fehr (approved):

xdsl/interactive/app.py
```diff
@@ -0,0 +1,139 @@
...
+        input_text = self.input_text_area.text
+        try:
+            ctx = MLContext(True)
+            for dialect in get_all_dialects():
+                ctx.load_dialect(dialect)
+            parser = Parser(ctx, input_text)
+            module = parser.parse_module()
+            self.current_module = module
+        except Exception as e:
+            self.current_module = e
```
webmiche, line 92: I think this function should be in an individual class that implements the logic behind the tool.

webmiche (commented):
To me, the current version of the tool combines the logic and the visualization in one tool. While I think that might be okay and easy enough to understand for the current version, I think this is not a good idea in the long run. So, I suggest splitting this into two parts: the visualization and the logic. In particular, the function I mention below should be in the logic to me, IMO.

I think this split does make sense especially if we want to start comparing/executing several decision paths, as we only want the logic there and the GUI stuff will be in the way and make things more complex (and inefficient).

webmiche:
tagging @superlopuh @math-fehr  and @dshaaban01 to make sure you see this :)

superlopuh:
I broadly agree, I didn't want to mention it upfront, because it is usually more difficult to iterate once the logic is split. I would say let's keep things in one place until that gets in the way, and then we can move things out. If you feel strongly about this we can move things out earlier too.

webmiche:
I prefer splitting as early as possible. The later you split, the more painful it gets and the more you need to redesign to understand how you want to interact among the two.

But, I also think you are more experienced than I am, so I am fine following your call.

superlopuh:
Let's see where we are in a couple of weeks and decide?

PR #1763, opened 2023-11-08
dialects: (aie) Add aie dialect

gabrielrodcanal:
Port of the AMD Xilinx AIE dialect for programming the AIEs on the AMD Xilinx Versal FPGA architecture. AIE is a hardened systolic array present in the Versal devices. The dialect describes netlists of AIE components and it can be lowered to the processor's assembly using the vendor's compiler. A description of the original dialect can be found here https://xilinx.github.io/mlir-aie/AIEDialect.html

Other than obvious improvements to the code the reviewers might spot, I would be interested in receiving feedback on the use of `SymbolTable` for `DeviceOp` and the reference to `createObjectFifo` from `ObjectFifoAcquireOp`. My main concern is that the user has to make sure the `createObjectFifo` operation is assigned to a `DeviceOp` before creating an `ObjectFifoAcquireOp` operation that refers to it. Do you see this as an acceptable restriction? If so, should we enforce this through the verifier? Find an example of what I am talking about in https://github.com/xdslproject/mlir-aie/blob/xdsl_tutorials/tutorials/tutorial-4/xdsl-tutorial-4.ipynb. In particular:
```
    object_fifo = aie.createObjectFifo(elem_number, tile14, tile34, i32, [256], "of")
    ...

    @Builder.region
    def region0(builder: Builder):
        ...
        builder.insert(object_fifo)
        ...
        
    device_val = IntegerAttr.from_int_and_width(0, i32)
    device = aie.DeviceOp(device_val, region0)

    acquire_fifo = aie.ObjectFifoAcquireOp(I32Attr(aie.PRODUCE_PORT), I32Attr(1), object_fifo.sym_name, device)
```

This forces the user to add new operations to the `DeviceOp` region after its creation instead of being able to insert all the operations with the `builder`. It looks a bit nasty to me. Example:

```
region0.block.add_op(core14)
```

I would like to request a  review from @mesham, @superlopuh, @math-fehr, @AntonLydike and @PapyChacal.

BTW this doesn't pass Pyright yet, but I'd like to get some feedback before I work on fixing this.


xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+    IntegerType,
+    NoneAttr,
+    Region,
+    Signedness,
+    StringAttr,
+    SymbolRefAttr,
+    i32,
+    i64,
+)
+from xdsl.dialects.memref import _MemRefTypeElement
```
superlopuh, line 25: here I would just copy/paste the definition, and also make it private to your file, clients shouldn't need to import TypeVars

superlopuh: The alternative is to import `AttributeInvT`

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+    @classmethod
+    def parse_parameter(cls, parser: AttrParser) -> str:
+        return parser.parse_str_literal()
+
+    def print_parameter(self, printer: Printer) -> None:
+        printer.print_string(f'"{self.data}"')
+
+
+@irdl_attr_definition
+class ObjectFIFO(ParametrizedAttribute):
```
superlopuh, line 82:
```suggestion
class ObjectFIFO(Generic[AttributeInvT], ParametrizedAttribute):
```

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
+from __future__ import annotations
```
superlopuh, line 1: would be great to have your PR description as a docstring at the top of this file, I'm sure future readers will find it useful, especially external links

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+        if self.link_with is not None:
+            printer.print(' { link_with="', self.link_with, '" }')
+
+
+@irdl_op_definition
+class DMABDOp(IRDLOperation):
+    name = "AIE.dmaBd"
+    offset: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
+    length: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
+    AB: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
```
superlopuh, line 252: nit, we've been trying to stick with lowercase names for our python classes, even if that disagrees with the way that's printed in the IR:
```suggestion
    ab: IntegerAttr[i32] = attr_def(IntegerAttr[i32], attr_name="AB")
```

superlopuh: Also, the type annotation isn't actually necessary, I'm guilty of having started this before realising that only the thing after the `=` is actually needed
```suggestion
    ab = attr_def(IntegerAttr[i32], attr_name="AB")
```

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+                memref.MemRefType.from_element_type_and_shape(element_type, shape)
+            ],
+        )
+
+
+@irdl_op_definition
+class FlowOp(IRDLOperation):
+    name = "AIE.flow"
+
+    sourceBundle: WireBundleAttr = attr_def(WireBundleAttr)
```
superlopuh, line 411: similarly here for attribute names and python vs C++ conventions, I'd recommend making python pretty and adding the translation hook thing. I'm not sure if everyone else will agree with this though, definitely a matter of personal preference

superlopuh (commented):
The symbol things look good to me. Is it the same as the aie dialect? Would it make sense to add some interop tests to check that this really is a reimplementation of their dialect exactly?

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+    result: OpResult = result_def(IndexType())
+
+    def __init__(self, col: IntegerAttr[i32], row: IntegerAttr[i32]):
+        super().__init__(
+            attributes={"col": col, "row": row}, result_types=[IndexType()]
+        )
+
+
+@irdl_op_definition
+class ConnectOp(IRDLOperation):
```
compor, line 178: missing `HasParent` trait

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+    name = "AIE.buffer"
+    tile: Operand = operand_def(IndexType())
+    shape: ArrayAttr[i32] = attr_def(ArrayAttr[i32])
+    element_type: AnyAttr = attr_def(AnyAttr())
+    buffer: OpResult = result_def(memref.MemRefType)
+    sym_name: StringAttr = attr_def(StringAttr)
+
+    def __init__(
+        self,
+        tile: IndexType(),
```
compor, line 151: Would it make sense for this to be 

```suggestion
        tile: SSAValue | Operation,
```

compor: Similarly in other places.

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+        element_type: AnyAttr,
+        shape: ArrayAttr[i32],
```
compor, line 153: Would it make sense to mimick the type annotations from the `MemRefType.from_element_type_and_shape` from `memref`?

compor (changes requested):
This is a massive PR and I'd love to have some test cases that exercise even a few ops.

I left some comments, which I think is a good start, but more iterations to come.

compor (commented):
I had a look wrt your `ObjectFifoAcquireOp` and `ObjectFifo` in Xilinx's `mlir-aie`project and to my understanding, there is no reason to expect anything to be "in place" during construction for one to lookup the symbol table.

First, if you want to query `ModuleOp`'s symbol table via the containing `DeviceOp`, there is no guarantee or easy/elegant way to restrict the construction order.
Second, the `ObjectFifoAcquireOp` seems to just store the `FlatSymbolRefAttr` in the TableGen [source][1]. Moreover, it is a flat symbol reference, so no nested lookups seem to be considered.

Thus, I'd do what you suggested and simply store it in `__init__` and use a custom `verify_` to check all the required conditions.
This seems to be what happens in `mlir-aie`, where the `getObjFifoName` is called in the verifier of `ObjectFifoAcquireOp` ([here][2]).

This should simplify construction in your notebook example.
Also, note that you can nest builders in xDSL.

[1]: https://github.com/Xilinx/mlir-aie/blob/76acfeef8faa929b09e9b19d34a9f26834f488e1/include/aie/Dialect/AIE/IR/AIEOps.td#L1608
[2]: https://github.com/Xilinx/mlir-aie/blob/76acfeef8faa929b09e9b19d34a9f26834f488e1/lib/Dialect/AIE/IR/AIEDialect.cpp#L712

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+            printer.print(' { link_with="', self.link_with, '" }')
+
+
+@irdl_op_definition
+class DMABDOp(IRDLOperation):
+    name = "AIE.dmaBd"
+    offset: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
+    length: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
+    AB: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
+    buffer: Operand = operand_def(memref.MemRefType)
```
gabrielrodcanal, line 253: @superlopuh does omiting the type annotation apply to every type of class member? For example for operands do we prefer 
```suggestion
    buffer = operand_def(memref.MemRefType)
```
?

xdsl/dialects/experimental/aie.py
gabrielrodcanal: What would be your suggestion for the name of this attribute?

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1056 @@
...
+                    object_fifo.buffer.memory_space,
+                )
+            ]
+        )
+
+
+@irdl_op_definition
+class AMSelOp(IRDLOperation):
+    name = "AIE.amsel"
+    arbiterID: IntegerAttr[i32] = attr_def(IntegerAttr[i32])
```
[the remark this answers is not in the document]
gabrielrodcanal: I've tried this but the verifier doesn't seem to be called upon creation of the operation?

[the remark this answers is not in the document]
compor: hmm, weird. If you have that implementation, please push it and we can have a look.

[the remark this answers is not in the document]
gabrielrodcanal: Just pushed it. You can try changing the following in this tutorial https://github.com/xdslproject/mlir-aie/blob/xdsl_tutorials/tutorials/tutorial-4/switchbox/xdsl-tutorial-4-switchbox.ipynb
`connect14 = aie.ConnectOp(aie.WireBundleAttr("DMA"), I32Attr(0), aie.WireBundleAttr("East"), I32Attr(1))`
by
`connect14 = aie.ConnectOp(aie.WireBundleAttr("DMA"), I32Attr(-1), aie.WireBundleAttr("East"), I32Attr(1))`

By changing the `arbiterID` the verifier should be trigger, but that doesn't happen unless `connect14.verify()` is called.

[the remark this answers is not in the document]
AntonLydike: Operation verifiers are not run automatically, they happen at precise points like:
- after input IR parsing is complete
- between passes

If you want to verify IR at other points, you should manually call `verify()` on the module op to verify the whole module, or on any sub-operation you want to verify individually.

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+            "<",
+            self.sourceBundle.data,
+            ": ",
+            self.sourceChannel.value.data,
+            ", ",
+            self.destBundle.data,
+            ": ",
+            self.destChannel.value.data,
+            ">",
+        )
```
superlopuh, line 234: This is missing a corresponding `parse` function. But also, it feels like this could leverage the format description API instead, like so:
```suggestion
    def print(self, printer: Printer):
        printer.print(
            "<",
            self.sourceBundle.data,
            ": ",
            self.sourceChannel.value.data,
            ", ",
            self.destBundle.data,
            ": ",
            self.destChannel.value.data,
            ">",
        )
        assembly_format = "`<` $sourceBundle `: ` $sourceChannel `, ` $destBundle `: ` destChannel `>`  attr-dict"
```

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+
+    def verify_(self) -> None:
+        print(self.sourceChannel.value.data)
```
superlopuh, line 237: please remove this print

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+            ": ",
+            self.buffer.type,
+            ", ",
+            self.offset.value.data,
+            ", ",
+            self.length.value.data,
+            ">, ",
+            self.ab.value.data,
+            ")",
+        )
```
superlopuh, line 338: looks like this is missing a parse as well

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+        printer.print(
+            "(",
+            f"0x{self.packet_type.value.data:X}",
+            ", ",
+            f"0x{self.packet_id.value.data:X}",
+            ")",
+        )
+
+
+# TODO: add successor basic blocks
```
superlopuh, line 364: what do you mean by this?

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+
+    device = attr_def(IntegerAttr[Annotated[IntegerType, i32]])
+    traits = frozenset([SymbolTable()])
+
+    def __init__(self, device: IntegerAttr[IntegerType], region: Region):
+        super().__init__(attributes={"device": device}, regions=[region])
+
+    def print(self, printer: Printer):
+        printer.print("(")
+        device_str = "xcvc1902" if self.device.value.data == 0 else ""
```
superlopuh, line 419: Something about this feels off to me, I'd expect the textual IR to reflect the value in the attribute. I have a vague recollection of @PapyChacal doing something conceptually related with enums but I might be wrong

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+
+    region = region_def()
+
+    device = attr_def(IntegerAttr[Annotated[IntegerType, i32]])
+    traits = frozenset([SymbolTable()])
+
+    def __init__(self, device: IntegerAttr[IntegerType], region: Region):
+        super().__init__(attributes={"device": device}, regions=[region])
+
+    def print(self, printer: Printer):
```
superlopuh, line 417: also missing a `parse` function

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+        sym_name: str,
+        shape: ArrayAttr[AnyIntegerAttr],
+        element_type: Attribute,
+    ):
+        super().__init__(
+            attributes={"sym_name": StringAttr(sym_name)},
+            result_types=[
+                memref.MemRefType.from_element_type_and_shape(element_type, shape)
+            ],
+        )
```
superlopuh, line 443: I would recommend keeping the inits as simple as possible, and letting the client deal with constructing the values.
```suggestion
    def __init__(
        self,
        sym_name: str,
        buffer: memref.MemRefType[Attribute],
    ):
        super().__init__(
            attributes={"sym_name": StringAttr(sym_name)},
            result_types=[buffer],
        )
```

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+
+    def print(self, printer: Printer):
+        printer.print(
+            "<", self.tile, ", ", self.bundle.data, " : ", self.channel.value.data, ">"
+        )
+
+
+@irdl_op_definition
+class putCascade(IRDLOperation):
```
superlopuh, line 901:
```suggestion
class PutCascade(IRDLOperation):
```
Please make all the classes PascalCase

xdsl/dialects/experimental/aie.py
```diff
@@ -0,0 +1,1129 @@
...
+        putStream,
+        ShimDMAOp,
+        ShimMuxOp,
+        ShimSwitchBoxOp,
+        SwitchboxOp,
+        UseLockOp,
+        WireOp,
+        EndOp,
+    ],
+    [WireBundleAttr, ObjectFIFO, ObjectFIFOSubview],
```
superlopuh, line 1128: nit:
```suggestion
    [WireBundleAttr, ObjectFIFO, ObjectFIFOSubview,],
```
Trailing commas makes it a little easier on git in the future

superlopuh (changes requested):
Looking good! I forget exactly what the agreement that we came up with wrt experimental was. Do we not want any tests? There are a number of operations with a print function but no parse function, so they won't roundtrip, is that a problem? If not, then I guess we could merge once the style comments are addressed

gabrielrodcanal:
> Looking good! I forget exactly what the agreement that we came up with wrt experimental was. Do we not want any tests? There are a number of operations with a print function but no parse function, so they won't roundtrip, is that a problem? If not, then I guess we could merge once the style comments are addressed

If things have not changed since the last conversation we had about experimental we don't want tests. Someone correct me if I'm wrong!

Oh, I think that might be a problem, now that you mention it. I'll address the issue in a following commit before merging.

xdsl/dialects/experimental/aie.py
gabrielrodcanal: I'm getting an error when using the format description API. It's my first time using it, but from what I've seen in other dialects just by removing the print method and having the `assembly_format` definition you provide it should be fine. I get an error during parser though: `'expected variable to refer to an operand, attribute, region, result, or successor'`. It fails with the first variable already, `sourceBundle`.

superlopuh: Can I have a go at this and push to your PR?

superlopuh: Ah I spoke with @math-fehr, and it turns out that there isn't support for attributes right now. He says that maybe he can work on it today or tomorrow, not sure if it's worth blocking the PR on this, even if it would make implementing the printers and parsers easier.

gabrielrodcanal: If it just has to wait one day maybe I'd say it's worth the wait. I have to work on some details of the execution of the code on actual hardware anyways, as it needs some tweaks with respect to the simulator.

superlopuh: Let's see what @math-fehr says

superlopuh: #2028

gabrielrodcanal: Just merged main with this branch and I keep getting the same error.

superlopuh: Can you please push the broken version?

xdsl/dialects/experimental/aie.py
PapyChacal: Completely missed that ping!
I implemented EnumAttribute, but it wouldn't help for printing a keyword-prefixed integer?
The declarative assembly format soon should have support for that.
That's the job of optional groups, that would be written 
```
( `xcvc1902` $device^ )?
```
And attribute directives and optional groups are both open on separate PRs :slightly_smiling_face:

tobiasgrosser:
> > Looking good! I forget exactly what the agreement that we came up with wrt experimental was. Do we not want any tests? There are a number of operations with a print function but no parse function, so they won't roundtrip, is that a problem? If not, then I guess we could merge once the style comments are addressed
> 
> If things have not changed since the last conversation we had about experimental we don't want tests. Someone correct me if I'm wrong!
> 
> Oh, I think that might be a problem, now that you mention it. I'll address the issue in a following commit before merging.

We should not unnecessarily slow down things in experimental. On the other side, I really wonder if test cases indeed slow down development so much. If they have a low cost, I would invite test cases but would maybe not insist on them,

xdsl/dialects/experimental/aie.py
gabrielrodcanal: @superlopuh sorry, didn't get a notification of this! The broken version is already pushed.

superlopuh: It looks like all the tests pass, there's no error. Could you please push the filecheck also?

gabrielrodcanal: Sorry, I didn't commit the broken version in a "broken" state, but with the custom print method and I thought you would replace it with the API line on your end as I was doing! Just comitted the broken version and now you can see how a test fails and gives the same error I get.

superlopuh: Thank you, I merged main and fixed a few issues. I also added a super basic filecheck test. It feels like it would be worth the effort to add the other ops for which you'd like to have custom syntax to the filecheck, just to make sure that they roundtrip ok

superlopuh (approved):
Great! I think it might be worth having a go at replacing all the `print` without a corresponding `parse` with the assembly_format, but don't want to slow you down from now :)

gabrielrodcanal:
> Great! I think it might be worth having a go at replacing all the `print` without a corresponding `parse` with the assembly_format, but don't want to slow you down from now :)

Yes! I'm working on it :)

xdsl/dialects/experimental/aie.py
gabrielrodcanal: That statement makes use of an EnumAttribute for the type of device. I'm afraid the declarative assembly format is not an option here because this operation has a region. Unless this capability has been added over the last week.

compor (commented):
I understand that this goes to `experimental` but I think it might worth revisting how we do things in general regarding this.
What would constitute "graduating" from `experimental`?

Considering how long this took to "finalize", I am not sure writing tests would slow down development that much.

On the flip side, we are called to review a massive PR and regain context after weeks of giving an initial review.

Considering this, I don't feel confident approving this, but I don't want to block it.
So, please go ahead without my explicit approval.

mesham (approved):
I think this looks good - it's adding the AMD AIE dialect into xDSL via experimental and as Gabriel has actioned the specific suggestions and comments on this I think it's good to get it checked in and usable more widely. There are some actions to write tests around this in the future, but many things in experimental lack tests (and indeed the point of experimental is to get less mature things in and available sooner) so it won't be unique in this regard. As experimental is explicitly "use at your own risk, this is less mature", then I think this is fine and nice to expand the available dialects.

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -313,7 +315,9 @@ def parse_optional_variable(
 
         self.raise_error(
             "expected variable to refer to an operand, "
-            "attribute, region, result, or successor"
+            "attribute, region, result, or successor",
+            at_position=start_pos,
+            end_position=end_pos,
         )
 
     def parse_type_directive(self) -> FormatDirective:
```
tobiasgrosser, line 323: I would suggest submitting this through a separate PR, as this is indeed not in the experimental folder or taking this out of this PR.

tobiasgrosser (approved):
I am fine adding this for now. Then we can go with smaller PRs that incrementally improve what we have.

xdsl/irdl/declarative_assembly_format_parser.py
gabrielrodcanal: I think this is a change from main, if I click on 'View file' it takes me to the merge @superlopuh did from main, so I guess there's no need for a new PR?

tobiasgrosser: Can you merge latest main and make sure this diff disappears?

tobiasgrosser:
OK, that is good to go for me. I personally would add tests but that's up to you, I guess.

gabrielrodcanal:
> OK, that is good to go for me. I personally would add tests but that's up to you, I guess.

I will add tests in the near future, once I'm done with some deadlines!

PR #1768, opened 2023-11-09
interactive: add clear input button and tests (2/7)

dshaaban01:
added clear input button and tests

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -93,6 +93,5 @@ async def test_input_and_buttons():
         """
         )
         await pilot.click("#clear_input_button")
-        # assert that the curent_module and test_module's are structurally equivalent
         await pilot.pause()
         assert app.input_text_area.text == ""


tests/interactive/test_app.py
```diff
@@ -66,3 +71,26 @@ async def test_input():
...
+
+@pytest.mark.asyncio()
+async def test_buttons():
+    """Test pressing keys has the desired result."""
+    async with InputApp().run_test() as pilot:
+        pilot: Pilot[None] = pilot
+        app = cast(InputApp, pilot.app)
+
+        # Test clicking the "clear input" button
+        app.input_text_area.insert(
```
superlopuh, line 84: can you please check that the input is not empty after this text is inserted?

xdsl/interactive/app.tcss
```diff
@@ -1,5 +1,11 @@
 
-# Vertical(TextArea)
+# Horizontal(Button)
+#clear_input{
+    height: 3;
```
superlopuh, line 4: is the height necessary? I'm not sure we need either of these css rules, can you please check if it looks any different without this?

superlopuh (changes requested):
thank you for splitting this change out! a couple of small changes

superlopuh (approved):
looks great, thank you

xdsl/interactive/app.py
```diff
@@ -130,6 +132,11 @@ def action_quit_app(self) -> None:
         """An action to quit the app."""
         self.exit()
 
+    @on(Button.Pressed, "#clear_input_button")
+    def on_clear_input_button_pressed(self, event: Button.Pressed) -> None:
+        """When the "Clear Input" button is pressed, the input IR TextArea is cleared and the current_module is updated"""
```
compor, line 137:
```suggestion
```

I think documentation is excessive here, if not inaccurate.
First, the `@on` decorator makes it clear which event and widget are handled here; no need to comment IMHO.
Second, "and the current_module is updated." part is wrong; that is done by `update_current_module`, right?

xdsl/interactive/app.tcss
compor: In the future, I'd suggest keeping these UI tweaks in separate PR's or add screenshots with the before/after maybe as @superlopuh has also requested.

compor (approved):
Thanks for keeping this minimal!

xdsl/interactive/app.py
superlopuh: Oh BTW I don't think the `@on` decorator is needed, the name of the function should already be enough to specify the behaviour

superlopuh: Ah no it would have been if the class name was custom, but it's not in our case. In that case, I'd recommend renaming the function `clear_input` to make the role of the function clearer, and keep the `@on` decorator.

xdsl/interactive/app.py
```diff
@@ -3,7 +3,8 @@
 
 Execute `xdsl-gui` in your terminal to run it.
 
-Run `terminal -m xdsl.interactive.app:InputApp --def` to run in development mode. Please be sure to install `textual-dev` to run this command.
+Run `terminal -m xdsl.interactive.app:InputApp --def` to run in development mode. Please
+be sure to install `textual-dev` to run this command.
```
compor, line 7: 🙏🏽 📜

superlopuh (approved):

tests/interactive/test_app.py
```diff
@@ -66,3 +69,27 @@ async def test_input():
...
+            """
+        func.func @hello(%n : index) -> index {
+          %two = arith.constant 2 : index
+          %res = arith.muli %n, %two : index
+          func.return %res : index
+        }
+        """
+        )
+        await pilot.pause()
+        assert app.input_text_area.text != ""
```
superlopuh, line 91: actually can you please test for the exact expected value?

superlopuh (changes requested):

tests/interactive/test_app.py
superlopuh: then we're gtg

superlopuh: I'm not sure I get what the shared mutable state is?

superlopuh: And how that's relevant to this != test

superlopuh: Ah I see what happened, not sure why you call it shared mutable state, there's something I'm not getting still, but we can discuss that later.

superlopuh (approved):

tests/interactive/test_app.py
```diff
@@ -66,3 +69,30 @@ async def test_input():
...
+            == """
+        func.func @hello(%n : index) -> index {
+          %two = arith.constant 2 : index
+          %res = arith.muli %n, %two : index
+          func.return %res : index
+        }
+        """
+        )
+        await pilot.click("#clear_input_button")
+        # assert that the curent_module and test_module's are structurally equivalent
```
superlopuh, line 96: This comment doesn't seem right

PR #1770, opened 2023-11-09
interactive: add copy output button + made changes related to textual update affecting testing (3/7)

dshaaban01:
added copy output button and some minor changes to test due to textual update. No tests needed as it is just a copy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,16 +101,6 @@ exclude = """
 |xdsl/_version.py$
 """
 
-[tool.flake8]
-max-line-length = 90
-ignore = "F403,E226,E731,E275,W503,F405,E722,E741,W504,W605,E402"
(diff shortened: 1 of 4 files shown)


xdsl/interactive/app.py
```diff
@@ -23,13 +17,12 @@
 from textual.widgets.text_area import TextAreaTheme
 
 from xdsl.dialects.builtin import ModuleOp
+from xdsl.interactive._pasteboard import pyclip_copy
```
superlopuh, line 20: Maybe this will work? I think it's unhappy with the underscore, which implies that it should not be imported from outside this scope.

```suggestion
from ._pasteboard import pyclip_copy
```

superlopuh (approved):

PR #1771, opened 2023-11-10
transformations: Implement convert-scf-to-openmp

PapyChacal:
Implement an xDSL `convert-scf-to-openmp`. While not -yet- supporting reductions (None of my usecases require it at the moment. I don't think it will be hard to add.) It does support arguments suited to more closely drive the parallelization.

Here's the included docstring:
Convert `scf.parallel` loops to `omp.wsloop` constructs for parallel execution.
Arguments (all optional):
- collapse : int: specify a positive number of loops to collapse. By default, the full dimensionality of converted parallel loops is collapsed. This argument allows to take a 2D loop and only OMPize the first dimension, for example.
- nested: bool: Set this to true to convert nested parallel loops. This is rarely a good idea, and is disabled by default. Note that setting it to true mimics MLIR's convert-scf-to-openmp.
- schedule: {"static", "dynamic", "auto"}: Set the schedule used by the OMP loop. By default, none is set, leaving the decision to MLIR's omp lowering. At the time of writing, this means static.
- chunk: int: Set the chunk size used by the OMP loop. By default, none is set. Note that the OMP dialect does not support setting a chunk size without a schedule; Thus selecting a chunk size without a schedule will use the static schedule.


xdsl/transforms/convert_scf_to_openmp.py
```diff
@@ -0,0 +1,136 @@
...
+class ConvertParallel(RewritePattern):
+    collapse: int | None
+    nested: bool
+    schedule: Literal["static", "dynamic", "auto"] | None
+    chunk: int | None
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, loop: scf.ParallelOp, rewriter: PatternRewriter, /):
+        if len(loop.initVals) > 0:
+            # TODO
```
superlopuh, line 28: issue?

tests/filecheck/transforms/convert-scf-to-openmp.mlir
superlopuh, line 1: it's not super clear what corresponds to what in this file, is there a way to make that a little clearer? for example, it seems that the check lines are below the inputs, but the last input is not followed by check lines at all

xdsl/transforms/convert_scf_to_openmp.py
PapyChacal: Sure why not

tests/filecheck/transforms/convert-scf-to-openmp.mlir
PapyChacal: I can add some sentences explaining what is checked; would that address your concern?

PapyChacal: Added some comments

xdsl/transforms/convert_scf_to_openmp.py
superlopuh: :pray:

superlopuh (approved):

xdsl/ir/core.py
```diff
@@ -1087,6 +1087,7 @@ def detach(self):
         if self.parent is None:
             raise Exception("Cannot detach a toplevel operation.")
         self.parent.detach_op(self)
+        return self
```
webmiche, line 1090: Why do you need this? You have the object when calling `detach()` on it, right? If you just need it like this in this specific lowering, I would prefer wrapping this into a local lambda rather than changing core. Otherwise, this should probably be PRed (and tested) individually, IMO.

PR #1772, opened 2023-11-10
interactive: app - ability to choose and apply passes onto IR (4/7)

dshaaban01:
Have added the ability to apply passes to the IR via the selection list. the output IR text area contains the changed IR. Tests have been added to test that pass application works (added pass tests to the previous test buttons function (i.e. merged under one async test) due to shared global variable negative affects failing multiple async tests.

image

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -22,12 +22,9 @@ async def test_inputs():
     async with InputApp().run_test() as pilot:
         app = cast(InputApp, pilot.app)
 
-        # Due to Python Static Analysis weirdness
-        curr_module = app.current_module
-
         # Test no input
         assert app.output_text_area.text == "No input"
-        assert curr_module is None
+        assert app.current_module is None
 
         # Test inccorect input
         app.input_text_area.insert("dkjfd")
@@ -37,11 +34,9 @@ async def test_inputs():
             == "(Span[5:6](text=''), 'Operation builtin.unregistered does not have a custom format.')"
         )
 
-        # Due to Python Static Analysis weirdness
-        curr_module = app.current_module
-        assert isinstance(curr_module, ParseError)
+        assert isinstance(app.current_module, ParseError)
         assert (
-            str(curr_module)
+            str(app.current_module)
             == "(Span[5:6](text=''), 'Operation builtin.unregistered does not have a custom format.')"
         )
 
@@ -79,10 +74,8 @@ async def test_inputs():
                 res = arith.Muli(n, two)
                 func.Return(res)
 
-        # Due to Python Static Analysis weirdness
-        curr_module = app.current_module
-        assert isinstance(curr_module, ModuleOp)
-        assert curr_module.is_structurally_equivalent(expected_module)
+        assert isinstance(app.current_module, ModuleOp)
+        assert app.current_module.is_structurally_equivalent(expected_module)
 
 
 @pytest.mark.asyncio()
@@ -115,7 +108,6 @@ async def test_passes():
     async with InputApp().run_test() as pilot:
         app = cast(InputApp, pilot.app)
         # Testing a pass
-        app.input_text_area.clear()
         app.input_text_area.insert(
             """
         func.func @hello(%n : index) -> index {
@@ -190,8 +182,6 @@ async def test_passes():
                     two_two = riscv.MVOp(one, rd=riscv.Registers.A0)
                     riscv_func.ReturnOp(two_two)
 
-        # Due to Python Static Analysis weirdness
-        curr_module = app.current_module
-        assert isinstance(curr_module, ModuleOp)
+        assert isinstance(app.current_module, ModuleOp)
         # Assert that the current module has been changed accordingly
-        assert curr_module.is_structurally_equivalent(expected_module)
+        assert app.current_module.is_structurally_equivalent(expected_module)
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -73,7 +73,6 @@ class InputApp(App[None]):
     def __init__(self):
         self.input_text_area = TextArea(id="input")
         self.output_text_area = OutputTextArea(id="output")
-
         self.passes_selection_list = SelectionList(id="passes_selection_list")
         super().__init__()
 


tests/interactive/test_app.py
```diff
@@ -69,12 +70,13 @@ async def test_input():
                 res = arith.Muli(n, two)
                 func.Return(res)
 
+        await pilot.pause()
```
superlopuh, line 73: why is this necessary? we're not simulating any user actions between this and the previous pause, I'd expect the test to work without it

tests/interactive/test_app.py
```diff
@@ -12,7 +13,7 @@
 
 
 @pytest.mark.asyncio()
-async def test_input():
+async def tests_input():
```
superlopuh, line 16: why this change?

tests/interactive/test_app.py
```diff
@@ -90,7 +92,51 @@ async def test_buttons():
...
+        assert app.input_text_area != ""
+
+        # Select a pass
+        app.passes_selection_list.select(
+            convert_func_to_riscv_func.ConvertFuncToRiscvFuncPass
+        )
+
+        await pilot.pause()
+        assert app.output_text_area != app.input_text_area
+        assert str(app.current_module) != "No input"
```
superlopuh, line 122: can you please make this test for the expected value, not unexpected string?

xdsl/interactive/app.py
```diff
@@ -61,13 +62,22 @@ class InputApp(App[None]):
     input_text_area = TextArea(id="input")
     output_text_area = OutputTextArea(id="output")
 
+    list_of_passes = get_all_passes()
```
superlopuh, line 65: this doesn't look right, if it's a constant, then it should be at the top level of the file, and probably a tuple instead of a string to enforce immutability

xdsl/interactive/app.py
```diff
@@ -79,23 +89,38 @@ def compose(self) -> ComposeResult:
...
 
+    @on(SelectionList.SelectedChanged)
+    def update_selected_view(self) -> None:
+        """
+        When the SelectionList (pass options) changes (i.e. a pass was selected or deselected), update the label to show
+        the query, and then call the update_current_module() function, which applies the selected passes to the input and displays the output
+        """
+        self.update_current_module()
+
     @on(TextArea.Changed, "#input")
```
superlopuh, line 100:
```suggestion
    @on(SelectionList.SelectedChanged)
    @on(TextArea.Changed, "#input")
```

You can stack the `@on` annotations it just means that when either of those events are received this method will be called

xdsl/interactive/app.py
```diff
@@ -79,23 +89,38 @@ def compose(self) -> ComposeResult:
...
                 ctx.load_dialect(dialect)
             parser = Parser(ctx, input_text)
             module = parser.parse_module()
+            pipeline = PipelinePass([p() for p in selected_passes])
+            pipeline.apply(ctx, module)
             self.current_module = module
         except Exception as e:
             self.current_module = e
 
+        self.watch_current_module()
```
superlopuh, line 122: this isn't right, the watch_current_module should be called automatically when it's updated, if it isn't then we have a problem that we should fix

xdsl/interactive/app.py
```diff
@@ -126,6 +151,16 @@ def on_mount(self) -> None:
 
         self.query_one("#input_container").border_title = "Input xDSL IR"
         self.query_one("#output_container").border_title = "Output xDSL IR"
+        self.query_one(
+            "#passes_selection_list"
+        ).border_title = "Choose a pass or multiplepasses to be applied."
+
+        # aids in the construction of the seleciton list containing all the passes
+        selections = [(value.name, value) for value in self.list_of_passes]
+        selections.sort()
```
superlopuh, line 160:
```suggestion
        selections = sorted((value.name, value) for value in self.list_of_passes)
```
If we're not providing a way for clients to specify the list of available passes for now, we can just reuse the function that generates them, without a class variable.

xdsl/interactive/app.py
```diff
@@ -126,6 +151,16 @@ def on_mount(self) -> None:
...
+        self.passes_selection_list.add_options(  # pyright: ignore[reportUnknownMemberType]
```
superlopuh, line 161: why the pyright ignore? we generally try to avoid them, especially things like `reportUnknownMemberType`, which might be hiding errors other than those we want to ignore.

xdsl/interactive/app.tcss
```diff
@@ -1,3 +1,14 @@
+# SelectionList
+#passes_selection_list {
+    margin: 1 1 1 1;
+    border: heavy $warning-lighten-1;
+    # border: heavy $accent;
```
superlopuh, line 5: please remove commented line

superlopuh (changes requested):
looking good, a few things to look into

xdsl/interactive/app.py
```diff
@@ -174,6 +174,8 @@ def action_quit_app(self) -> None:
     def on_clear_input_button_pressed(self, event: Button.Pressed) -> None:
         """When the "Clear Input" button is pressed, the input IR TextArea is cleared and the current_module is updated"""
         self.input_text_area.clear()
+        # when input is cleared, deselect the passes
+        self.passes_selection_list.deselect_all()
```
superlopuh, line 178: that's pretty unexpected behaviour, why should the selected passes change?

tests/interactive/test_app.py
superlopuh: pytest works via string matching, so `test_` functions are matched specifically. I think we allow for `tests_` also but it threw me off, I'd recommend changing it back

xdsl/interactive/app.py
superlopuh: If you were using this app, not having built it, would you expect a button that says "Clear Input" to do something other than clearing the input? But in general it is common to want to apply a certain pass pipeline to different inputs

xdsl/interactive/app.py
```diff
@@ -130,6 +146,15 @@ def on_mount(self) -> None:
 
         self.query_one("#input_container").border_title = "Input xDSL IR"
         self.query_one("#output_container").border_title = "Output xDSL IR"
+        self.query_one(
+            "#passes_selection_list"
+        ).border_title = "Choose a pass or multiplepasses to be applied."
```
superlopuh, line 151:
```suggestion
        ).border_title = "Choose a pass or multiple passes to be applied."
```

xdsl/interactive/app.py
superlopuh: Ah I see another Textual bug... Could you please file an issue on me to investigate this later? And add a comment above this line? I'll have a look when the next version of Textual comes out.

superlopuh (changes requested):

xdsl/interactive/app.py
superlopuh: Thank you!

xdsl/interactive/app.py
```diff
@@ -3,7 +3,8 @@
 
 Execute `xdsl-gui` in your terminal to run it.
 
-Run `terminal -m xdsl.interactive.app:InputApp --def` to run in development mode. Please be sure to install `textual-dev` to run this command.
+Run `terminal -m xdsl.interactive.app:InputApp --def` to run in development mode. Please
```
superlopuh, line 6:
```suggestion
Run `terminal -m xdsl.interactive.app:InputApp --dev` to run in development mode. Please
```

xdsl/interactive/app.py
```diff
@@ -13,14 +14,20 @@
...
 from xdsl.ir import MLContext
 from xdsl.parser import Parser
+from xdsl.passes import ModulePass, PipelinePass
 from xdsl.printer import Printer
-from xdsl.tools.command_line_tool import get_all_dialects
+from xdsl.tools.command_line_tool import get_all_dialects, get_all_passes
+
+from ._pasteboard import pyclip_copy
+
+available_passes = tuple(get_all_passes())
```
superlopuh, line 29:
```suggestion
ALL_PASSES = tuple(get_all_passes())
```

superlopuh (changes requested):
Looking great, a couple of things remaining and we're GTG

tests/interactive/test_app.py
```diff
@@ -81,17 +82,65 @@ async def test_input_and_buttons():
...
         func.func @hello(%n : index) -> index {
           %two = arith.constant 2 : index
           %res = arith.muli %n, %two : index
           func.return %res : index
         }
         """
         )
-        await pilot.click("#clear_input_button")
+
+        # Select a pass
```
superlopuh, line 104: This should be in a separate test. What exactly breaks when you split them out? Let's figure this out before merging, please let me know what you've tried and what breaks exactly

tests/interactive/test_app.py
```diff
@@ -81,17 +82,65 @@ async def test_input_and_buttons():
...
+        app.passes_selection_list.select(
+            convert_func_to_riscv_func.ConvertFuncToRiscvFuncPass
+        )
+
         await pilot.pause()
-        assert app.input_text_area.text == ""
+        assert (
+            str(app.current_module)
```
superlopuh, line 111: this should be using structural equality

tests/interactive/test_app.py
superlopuh: ok that's the whole point of cloning them, let's make them not shared

tests/interactive/test_app.py
superlopuh: There's an example of this just above BTW, there's a bit more code to write for this module but it should be pretty similar. I'd probably recommend changing the example being tested to something like `dce` on arith. The riscv lowering representation is likely to change, and it's a nice general principle to make unit tests be as specific as possible to avoid breaking tests that are testing something unrelated

superlopuh (approved):
glad we took the time to iron this out, looking great now, modulo test case

PR #1775, opened 2023-11-10
interactive: app, add "clear passes" button (7/7)

dshaaban01:
add "clear passes" query - removes the effect of passes from the input IR (i.e. input and output textarea are the same), deselects all passes from the selection list and removes the pass names from the query label. 

image

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -117,12 +117,13 @@ class InferAddOpShapeTrait(ToyShapeInferenceTrait):
         if not isinstance(op, AddOp):
             raise TypeError
         if not (
-            isinstance(op.lhs.type, TensorType) and isinstance(op.rhs.type, TensorType)
+            isinstance(op_lhs_type := op.lhs.type, TensorType)
+            and isinstance(op_rhs_type := op.rhs.type, TensorType)
         ):
             return
-        assert op.lhs.type.get_shape() == op.rhs.type.get_shape()
-        if isinstance(op.res.type, TensorType):
-            assert op.lhs.type.get_shape() == op.res.type.get_shape()
+        assert op_lhs_type.get_shape() == op_rhs_type.get_shape()
+        if isinstance(op_res_type := op.res.type, TensorType):
+            assert op_lhs_type.get_shape() == op_res_type.get_shape()
         else:
             op.res.type = op.lhs.type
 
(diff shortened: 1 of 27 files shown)


tests/interactive/test_app.py
```diff
@@ -141,5 +141,32 @@ async def test_input_and_buttons():
...
+"""
+        )
+
+        # press clear input button
+        await pilot.click("#clear_selection_list_button")
+
+        await pilot.pause()
+        assert app.passes_selection_list.selected == []
+        assert (
+            str(app.current_module)
```
superlopuh, line 153: please switch this to comparing for structural equality

superlopuh (changes requested):

xdsl/interactive/app.py
```diff
@@ -102,6 +105,18 @@ def compose(self) -> ComposeResult:
...
+    def update_selected_view(self) -> None:
+        """
+        When the SelectionList (pass options) changes (i.e. a pass was selected or deselected), update the label to show
+        the query, and then call the update_current_module() function, which applies the selected passes to the input and displays the output
+        """
+        new_passes = "\n" + (", " + "\n").join(
+            p.name for p in self.passes_selection_list.selected
+        )
+        new_label = f"xdsl-opt -p {new_passes}"
+        self.query_one(Label).update(new_label)
```
webmiche, line 118: I am confused. Why is this now needed? Also, why are you saying that you are calling the `update_current_module`-function in the docu, but don't in the actual code?

tests/interactive/test_app.py
webmiche, line 1: Maybe it makes sense in this case (since you need to spawn an instance of the tool for everything), but I generally don't like having only one test that tests everything. Having unittests for individual parts makes it much easier to find bugs.

Also, testing would be simplified quite a bit too by splitting the logic and the UI :wink:

webmiche: You mean the `asyc ... wait`? This is due textual, right? If we had a split, we don't even need to have textual tests, IMO.

Let's keep it as-is for the moment until we have all your current work upstream and then think about the split again :)

superlopuh: I disagree, if we split things up we'd need two tests, one for the model, and one for the app, so this would still be unavoidable. Will be a bit easier to discuss when the current batch of PRs is upstreamed

superlopuh (approved):
great job Dalia

xdsl/interactive/app.py
compor: I agree here, that the documentation is confusing; and I think I've commented this on a previous PR.

If it is to use Textual's update model, I'd suggest avoiding documenting it as this is how the framework works and saves you from having to update 2 places in the source (functionality and documentation) if things change.

compor (commented):
A wee post-merge review.
LGTM with a few comments.

I guess my recurring comment would be to tone down the very verbose docstrings a bit.

PR #1780, opened 2023-11-13
dialects: (hw) Add InnerSymbolTable and InnerRefNamespace to HW dialect

lucjaulmes:
This is another standalone part of #1704 (which you can refer to for broader context). It introduces 2 main classes, `InnerSymbolTable` and `InnerRefNamespace` (which are too interconnected to add separately), as well as 2 smaller (and similarly intricated) `InnerSymbolTableCollection` and `InnerRefUserOpInterface`.

Note that I’ve tried reducing as much as I could but this is as small as I can make this contribution (~150 loc / 4 classes + tests).

The PR introduces a verification step, called `verifyInnerRefs`:
- I think the most explicit description is from the [Symbol and Inner Symbol Rationale document](https://circt.llvm.org/docs/RationaleSymbols/#innerrefnamespace):

  > **InnerRefNamespace**
  > 
  > [As class:] Combines InnerSymbolTableCollection with a SymbolTable for resolution of InnerRefAttrs, primarily used during verification as argument to the verifyInnerRefs hook which operations may use for more efficient checking of InnerRefAttrs.
  > 
  > [As trait:] InnerRefNamespace is used by Operations to define a new scope for InnerRef’s. Operations with this trait must also be a SymbolTable. Presently the only user is CircuitOp.
  > 
  > Inner symbols are more costly than normal symbols, precisely from the relaxation of MLIR symbol constraints. Since nested regions are allowed, finding all operations defining an inner_sym requires a recursive IR scan.
  > Verification is likewise trickier, partly due the significant increase in non-local references.
  > 
  > For this reason, verification is driven as a trait verifier on InnerRefNamespace which constructs and verifies InnerSymbolTables in parallel, and uses these to conduct a per-InnerSymbolTable parallelized walk to verify references by calling the verifyInnerRefs hook on InnerRefUserOpInterface operations.

  I’ve tried to summarise this in the `InnerRefNamespace` docstring.
- I still have a couple of doubts on how/where this `verify_inner_refs()` should be called. It seems to be called primarily from `verifyRegionTrait`, though I couldn’t find what that corresponds to in xdsl terminology. Just a `verify()` or `verify_()`?  
- <details><summary><code>verify_inner_refs</code> is additionally called in a <code>VerifyInnerRefNamespacePass</code> HW transform, but I left that out for now as it seems we can do without it (at least for now).</summary> 
  
  - The doc for `VerifyInnerRefNamespacePass` says it’s used to “_to drive verification of operations that want to be InnerRefNamespace's but don't have the trait_” using `InnerRefNamespaceLike` to find applicable operations.  
  - `InnerRefNamespaceLike` interface in turn says “_Classify operations that are InnerRefNamespace-like, until structure is in place to do this via Traits. […] Prefer putting the trait on operations here or downstream._”.
  </details>

@PapyChacal since AFAIU you worked on the main symbol table, can you try and see if I tested this right, and find what more could be tested in here? Especially I think my attempts at creating synthetic ops using the interfaces in the python tests might be clunky/simplifiable.

### DIFF
--- a/tests/dialects/test_hw.py
+++ b/tests/dialects/test_hw.py
@@ -2,11 +2,36 @@
 Test HW traits and interfaces.
 """
 
+from unittest.mock import ANY, patch
 
+import pytest
+
+from xdsl.dialects.builtin import StringAttr
 from xdsl.dialects.hw import (
+    InnerRefAttr,
+    InnerRefNamespaceTrait,
+    InnerRefUserOpInterfaceTrait,
+    InnerSymbolTableCollection,
+    InnerSymbolTableTrait,
     InnerSymTarget,
 )
 from xdsl.dialects.test import TestOp
+from xdsl.irdl import (
+    IRDLOperation,
+    Region,
+    attr_def,
+    irdl_op_definition,
+    opt_region_def,
+    region_def,
+)
+from xdsl.traits import (
+    IsTerminator,
+    SingleBlockImplicitTerminator,
+    SymbolOpInterface,
+    SymbolTable,
+    ensure_terminator,
+)
+from xdsl.utils.exceptions import VerifyException
 
 
 def test_inner_sym_target():
@@ -25,3 +50,221 @@ def test_inner_sym_target():
     assert sub_target
     assert not sub_target.is_op_only()
     assert sub_target.is_field()
+
+
+# Module / Circuit / Wire / Output naming taken from FIRRTL classes using these interfaces.
+@irdl_op_definition
+class ModuleOp(IRDLOperation):
+    name = "module"
+    region = region_def()
+    sym_name = attr_def(StringAttr)
+    traits = frozenset({InnerSymbolTableTrait(), SymbolOpInterface()})
+
+
+@irdl_op_definition
+class OutputOp(IRDLOperation):
+    name = "output"
+    traits = frozenset({IsTerminator()})
+
+
+@irdl_op_definition
+class CircuitOp(IRDLOperation):
+    name = "circuit"
+    region: Region | None = opt_region_def()
+    sym_name = attr_def(StringAttr)
+    traits = frozenset(
+        {
+            InnerRefNamespaceTrait(),
+            SymbolTable(),
+            SingleBlockImplicitTerminator(OutputOp),
+        }
+    )
+
+    def __post_init__(self):
+        for trait in self.get_traits_of_type(SingleBlockImplicitTerminator):
+            ensure_terminator(self, trait)
+
+
+@irdl_op_definition
+class WireOp(IRDLOperation):
+    name = "wire"
+    sym_name = attr_def(StringAttr)
+    traits = frozenset({InnerRefUserOpInterfaceTrait()})
+
+
+def test_inner_symbol_table_interface():
+    """
+    Test operations that conform to InnerSymbolTableTrait
+    """
+    mod = ModuleOp(
+        attributes={"sym_name": StringAttr("symbol_name")}, regions=[[OutputOp()]]
+    )
+
+    circ = CircuitOp(attributes={"sym_name": StringAttr("other_name")}, regions=[[mod]])
+    circ.verify()
+
+    mod_no_parent = ModuleOp(
+        attributes={"sym_name": StringAttr("symbol_name")}, regions=[[OutputOp()]]
+    )
+    with pytest.raises(
+        VerifyException,
+        match="Operation module with trait InnerSymbolTableTrait must have a parent with trait SymbolOpInterface",
+    ):
+        mod_no_parent.verify()
+
+    mod_no_trait_circ = ModuleOp(
+        attributes={"sym_name": StringAttr("symbol_name")}, regions=[[OutputOp()]]
+    )
+    no_trait_circ = TestOp(regions=[[mod_no_trait_circ, OutputOp()]])
+    with pytest.raises(
+        VerifyException,
+        match="Operation module with trait InnerSymbolTableTrait must have a parent with trait InnerRefNamespaceTrait",
+    ):
+        mod_no_trait_circ.verify()
+    with pytest.raises(
+        VerifyException,
+        match="Operation module with trait InnerSymbolTableTrait must have a parent with trait InnerRefNamespaceTrait",
+    ):
+        no_trait_circ.verify()
+
+    @irdl_op_definition
+    class MissingTraitModuleOp(IRDLOperation):
+        name = "module"
+        region = region_def()
+        sym_name = attr_def(StringAttr)
+        traits = frozenset({InnerSymbolTableTrait()})
+
+    mod_missing_trait = MissingTraitModuleOp(
+        attributes={"sym_name": StringAttr("symbol_name")}, regions=[[OutputOp()]]
+    )
+    circ_mod_missing_trait = CircuitOp(regions=[[mod_missing_trait, OutputOp()]])
+    with pytest.raises(
+        VerifyException, match="Operation module must have trait SymbolOpInterface"
+    ):
+        mod_missing_trait.verify()
+    with pytest.raises(
+        VerifyException, match="Operation module must have trait SymbolOpInterface"
+    ):
+        circ_mod_missing_trait.verify()
+
+    @irdl_op_definition
+    class MissingAttrModuleOp(IRDLOperation):
+        name = "module"
+        region = region_def()
+        traits = frozenset({InnerSymbolTableTrait(), SymbolOpInterface()})
+
+    mod_missing_trait_parent = ModuleOp(regions=[[OutputOp()]])
+    MissingAttrModuleOp(regions=[[mod_missing_trait_parent, OutputOp()]])
+    with pytest.raises(
+        VerifyException,
+        match="attribute sym_name expected",
+    ):
+        mod_missing_trait_parent.verify()
+
+
+def test_inner_ref_namespace_interface():
+    """
+    Test operations that conform to InnerRefNamespaceTrait
+    """
+
+    @irdl_op_definition
+    class MissingTraitCircuitOp(IRDLOperation):
+        name = "circuit"
+        region: Region | None = opt_region_def()
+        sym_name = attr_def(StringAttr)
+        traits = frozenset(
+            {InnerRefNamespaceTrait(), SingleBlockImplicitTerminator(OutputOp)}
+        )
+
+    wire0 = WireOp(attributes={"sym_name": StringAttr("wire0")})
+    mod0 = MissingTraitCircuitOp(
+        attributes={"sym_name": StringAttr("mod0")}, regions=[[wire0, OutputOp()]]
+    )
+    circuit_no_symboltable = CircuitOp(
+        attributes={"sym_name": StringAttr("circuit")}, regions=[[mod0]]
+    )
+
+    with pytest.raises(
+        VerifyException, match="Operation circuit must have trait SymbolTable"
+    ):
+        circuit_no_symboltable.verify()
+
+    wire1 = WireOp(attributes={"sym_name": StringAttr("wire1")})
+    wire2 = WireOp(attributes={"sym_name": StringAttr("wire2")})
+    mod1 = ModuleOp(
+        attributes={"sym_name": StringAttr("mod1")}, regions=[[wire1, OutputOp()]]
+    )
+    mod2 = ModuleOp(
+        attributes={"sym_name": StringAttr("mod2")}, regions=[[wire2, OutputOp()]]
+    )
+    circuit = CircuitOp(
+        attributes={"sym_name": StringAttr("circuit")}, regions=[[mod1, mod2]]
+    )
+
+    # InnerRefUserOpInterfaceTrait.verify_inner_refs() does not do anything, so just mock
+    # to check it is called
+    with patch.object(
+        InnerRefUserOpInterfaceTrait, "verify_inner_refs"
+    ) as inner_ref_verif:
+        circuit.verify()
+
+    inner_ref_verif.assert_any_call(wire1, ANY)
+    inner_ref_verif.assert_any_call(wire2, ANY)
+    assert inner_ref_verif.call_count == 2
+
+
+def test_inner_symbol_table_collection():
+    """
+    Test operations that pertain to InnerSymbolTableCollection
+    """
+    wire1 = WireOp(attributes={"sym_name": StringAttr("wire1")})
+    wire2 = WireOp(attributes={"sym_name": StringAttr("wire2")})
+    mod1 = ModuleOp(
+        attributes={"sym_name": StringAttr("mod1")}, regions=[[wire1, OutputOp()]]
+    )
+    mod2 = ModuleOp(
+        attributes={"sym_name": StringAttr("mod2")}, regions=[[wire2, OutputOp()]]
+    )
+    circuit = CircuitOp(
+        attributes={"sym_name": StringAttr("circuit")}, regions=[[mod1, mod2]]
+    )
+
+    inner_sym_tables = InnerSymbolTableCollection(op=circuit)
+
+    with pytest.raises(
+        VerifyException, match="Operation wire should have InnerSymbolTableTrait trait"
+    ):
+        inner_sym_tables.get_inner_symbol_table(wire1)
+
+    sym_table1 = inner_sym_tables.get_inner_symbol_table(mod1)
+    sym_table2 = inner_sym_tables.get_inner_symbol_table(mod2)
+    assert (
+        sym_table1 is not sym_table2
+    ), "Different InnerSymbolTableTrait objects must return different instances of inner symbol tables"
+
+    unpopulated_inner_sym_tables = InnerSymbolTableCollection()
+    sym_table3 = unpopulated_inner_sym_tables.get_inner_symbol_table(mod1)
+    sym_table4 = unpopulated_inner_sym_tables.get_inner_symbol_table(mod2)
+    assert (
+        sym_table3 is not sym_table4
+    ), "InnerSymbolTableTrait still behave as expected when created on the fly"
+
+
+def test_inner_ref_attr():
+    """
+    Test inner reference attributes
+    """
+    wire1 = WireOp(attributes={"sym_name": StringAttr("wire1")})
+    wire2 = WireOp(attributes={"sym_name": StringAttr("wire2")})
+    mod1 = ModuleOp(
+        attributes={"sym_name": StringAttr("mod1")}, regions=[[wire1, OutputOp()]]
+    )
+    mod2 = ModuleOp(
+        attributes={"sym_name": StringAttr("mod2")}, regions=[[wire2, OutputOp()]]
+    )
+    CircuitOp(attributes={"sym_name": StringAttr("circuit")}, regions=[[mod1, mod2]])
+
+    ref = InnerRefAttr("mod2", "wire2")
+    assert (
+        ref.get_module().data == "mod2"
+    ), "Name of the referenced module should be returned correctly"
--- a/tests/dialects/test_hw.py
+++ b/tests/dialects/test_hw.py
@@ -229,7 +229,17 @@ def test_inner_symbol_table_collection():
         attributes={"sym_name": StringAttr("circuit")}, regions=[[mod1, mod2]]
     )
 
-    inner_sym_tables = InnerSymbolTableCollection(op=circuit)
+    with pytest.raises(
+        VerifyException, match="Operation wire should have InnerRefNamespaceTrait trait"
+    ):
+        inner_sym_tables = InnerSymbolTableCollection(wire1)
+
+    inner_sym_tables = InnerSymbolTableCollection(circuit)
+
+    with pytest.raises(
+        VerifyException, match=r"Trying to insert the same op twice in symbol tables:"
+    ):
+        inner_sym_tables.populate_and_verify_tables(circuit)
 
     with pytest.raises(
         VerifyException, match="Operation wire should have InnerSymbolTableTrait trait"
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -194,7 +194,7 @@ class InnerSymbolTableCollection:
         for op in inner_sym_table_ops:
             if op in self.symbol_tables:
                 raise VerifyException(
-                    f"Trying to insert the same op {op} twice in symbol tables"
+                    f"Trying to insert the same op twice in symbol tables: {op}"
                 )
             self.symbol_tables[op] = InnerSymbolTable(op)
 
@@ -220,11 +220,8 @@ class InnerRefNamespaceTrait(OpTrait):
                 f"Operation {op.name} must have trait {trait.__name__}"
             )
 
-        # These 2 checks are in fact redundant, as already checked as a SymbolTable
-        if len(op.regions) != 1:
-            raise VerifyException(f"Operation {op.name} must have a single region")
-        if len(op.regions[0].blocks) != 1:
-            raise VerifyException(f"Operation {op.name} must have a single block")
+        # Upstreams verifies that len(op.regions) == 1 and len(op.regions[0].blocks) == 1
+        # however this is already checked as part of SymbolTable, so would be redundant to re-check here
 
         namespace = InnerRefNamespace(op)
 
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -1,11 +1,12 @@
 """
 This is a stub of CIRCT’s hw dialect.
-It currently implements minimal types and operations used by other dialects.
+It currently implements minimal types and operations for the symbols and inner symbols used by other dialects.
 
 [1] https://circt.llvm.org/docs/Dialects/HW/
+[2] https://circt.llvm.org/docs/RationaleSymbols/
 """
 
-from dataclasses import dataclass
+from dataclasses import InitVar, dataclass, field
 
 from xdsl.dialects.builtin import (
     FlatSymbolRefAttr,
@@ -24,6 +25,12 @@ from xdsl.irdl import (
 )
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
+from xdsl.traits import (
+    OpTrait,
+    SymbolOpInterface,
+    SymbolTable,
+)
+from xdsl.utils.exceptions import VerifyException
 
 
 @dataclass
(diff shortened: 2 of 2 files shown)


superlopuh (approved):
not familiar with the dialect, but code looks reasonable

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
...
+    assert iface3 is not None
+    iface3.verify_region_trait(op4)
+
+
+@irdl_op_definition
+class InnerRefUserOp(IRDLOperation):
+    name = "ref"
+    sym_name = attr_def(StringAttr)
+    traits = frozenset({InnerRefUserOpInterface()})
+
```
PapyChacal, line 150: I think you can remove this one :slightly_smiling_face:

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
...
+    Test that operations that conform to SymbolOpInterface have necessary attributes.
+    """
+
+    @irdl_op_definition
+    class MissingTraitModuleOp(IRDLOperation):
+        name = "module"
+        traits = frozenset({InnerSymbolTable()})
+
+    op0 = MissingTraitModuleOp()
+    iface0 = op0.get_trait(InnerSymbolTable)
```
PapyChacal, line 90: AFAIU you could hoist it out of the test function

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
...
+    ):
+        # NB. seems to be called in a separate pass
+        iface0.verify_region_trait(op0)
+
+    @irdl_op_definition
+    class MissingAttrModuleOp(IRDLOperation):
+        name = "module"
+        traits = frozenset({InnerSymbolTable(), SymbolOpInterface()})
+
+    op1 = MissingAttrModuleOp()
```
PapyChacal, line 104: Idem, AFAIU you could hoist it out of the test function

PapyChacal (approved):
Left some testing comments as asked for! Looks good overall

You have some untested lines, codecov report them nicely, you can click on the filenames on its PR comment to see which ones are not tested.

Note that I'm giving both those comments as you requested specific feedback; In my opinion, as my approval says, it does look way good enough to move on for me.
This is in a pretty clean state already, I would say you can merge as is and you can start using the dialect and fix anything that would show up on the go.

Now, I won't block from trying to merge it directly in a perfect state either :wink:

tests/dialects/test_hw.py
lucjaulmes: I was trying to scope the failing definitions to their test functions and only keeping the ones that are reused for several tests outside − not a pertinent distinction then I’m guessing?

tests/dialects/test_hw.py
lucjaulmes: Ah yes I wanted to assert the verify trait function got called… that slipped through the cracks apparently

tests/dialects/test_hw.py
superlopuh: I think it's great as it is

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
     assert sub_target
     assert not sub_target.is_op_only()
     assert sub_target.is_field()
+
+
+@irdl_op_definition
+class ModuleOp(IRDLOperation):
+    name = "module"
+    region: Region = region_def()
```
compor, line 54:
```suggestion
    region = region_def()
```

Don't feel obliged to change this, just easier to read. Similarly below.

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
...
+
+@irdl_op_definition
+class OutputOp(IRDLOperation):
+    name = "output"
+    traits = frozenset({IsTerminator()})
+
+
+@irdl_op_definition
+class CircuitOp(IRDLOperation):
+    name = "circuit"
```
compor, line 67: Are these mimicking CIRCT operations? Is this from FIRRTL for example?
If so, do you mind adding a comment pointing that out?

Not sure if the plan is to actually implement them in the end in their respective dialects, but it might be helpful as placeholder comment.

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
...
+    )
+    op2.verify()
+    iface2 = op2.get_trait(InnerSymbolTable)
+
+    @irdl_op_definition
+    class ParentOp(IRDLOperation):
+        name = "parent"
+        region: Region = region_def()
+
+    ParentOp(regions=[[op2]])
```
compor, line 124:
```suggestion
```

This seems to be not used, correct?

tests/dialects/test_hw.py
```diff
@@ -25,3 +46,132 @@ def test_inner_sym_target():
...
+
+@irdl_op_definition
+class InnerRefUserOp(IRDLOperation):
+    name = "ref"
+    sym_name = attr_def(StringAttr)
+    traits = frozenset({InnerRefUserOpInterface()})
+
+
+@irdl_op_definition
+class WireOp(IRDLOperation):
```
compor, line 153: Maybe move this up in the file to be together with the rest of the "mock" classes?

xdsl/dialects/hw.py
```diff
@@ -112,6 +120,173 @@ def print_parameters(self, printer: Printer) -> None:
...
+        if parent is None or len(parent.get_traits_of_type(InnerRefNamespace)) != 1:
+            raise VerifyException(
+                f"Operation {op.name} with trait {cls.__name__} must have a parent with trait InnerRefNamespace"
+            )
+
+    def lookup(
+        self, inner_symbol_table_op: Operation, name: StringAttr
+    ) -> InnerSymTarget:
+        """Look up a symbol with the specified name, returning empty InnerSymTarget if
+        no such name exists. Names never include the @ on them."""
```
compor, line 157:
```suggestion
        """Look up a symbol with the specified name, returning an empty InnerSymTarget if
        no such name exists. Names never include the `@` sign."""
```

compor (approved):

tests/dialects/test_hw.py
compor: Thinking again about this, I'm not sure what the intention is here.

Maybe clarify with a comment?
Also, if this is testing that this creation/nesting is valid maybe put some `assert`s of what is expected and then you won't need a comment.

Lastly, do you need a `ParentOp`? Could you get away with a `TestOp` having variadic regions?

lucjaulmes: As to what it’s testing: the verification checks that the parent of an `InnerSymbolTable` has the `InnerRefNamespace` trait. `ParentOp` is just a mock that doesn’t have that trait.

compor: It's a keeper then!

lucjaulmes:
Any takes on whether `verify_region_trait()` should stay as is or be changed to just `verify()`? @compor maybe?

xdsl/dialects/hw.py
```diff
@@ -112,6 +120,173 @@ def print_parameters(self, printer: Printer) -> None:
...
+        # essential to how InnerRef's work.
+        if not op.has_trait(SymbolOpInterface):
+            raise VerifyException(
+                f"Operation {op.name} must have trait SymbolOpInterface"
+            )
+
+        # InnerSymbolTable's must be directly nested within an InnerRefNamespace,
+        # however don’t test InnerRefNamespace’s symbol lookups
+        parent = op.parent_op()
+        if parent is None or len(parent.get_traits_of_type(InnerRefNamespace)) != 1:
```
compor, line 148:
```suggestion
        parent = op.parent_op()
        if (parent := op.parent_op()) is None or len(parent.get_traits_of_type(InnerRefNamespace)) != 1:
```

compor:
> Any takes on whether `verify_region_trait()` should stay as is or be changed to just `verify()`? @compor maybe?

Good point, I missed this. I'd say to rename to `verify` since this is both a class and a trait in CIRCT.

xdsl/dialects/hw.py
```diff
@@ -112,6 +120,173 @@ def print_parameters(self, printer: Printer) -> None:
...
+        super().__init__()
+
+    @classmethod
+    def verify_region_trait(cls, op: Operation):
+        # Insist that ops with InnerSymbolTable's provide a Symbol, this is
+        # essential to how InnerRef's work.
+        if not op.has_trait(SymbolOpInterface):
+            raise VerifyException(
+                f"Operation {op.name} must have trait SymbolOpInterface"
+            )
```
compor, line 143:
```suggestion
        if not op.has_trait(trait := SymbolOpInterface):
            raise VerifyException(
                f"Operation {op.name} must have trait {trait.__name__}"
            )
```

Can you please do this where appropriate throughout this PR?

lucjaulmes:
It’s mostly done now, just a couple of tests to add on the lookup functions.

Unfortunately there are some exceptions I can’t manage to trigger as they are redundant with other checks elsewhere. Should I keep these checks or remove them?

One example is checking `InnerRefNamespace` operations only have a single region/block, which is redundant with `SymbolTable` checks (and `InnerRefNamespace` enforces the operation is also a `SymbolTable`).

lucjaulmes:
Sorry this is taking longer than planned. I had a couple of other things on my plate but also hit a bit of a wall.

AFAIU the whole point of the `InnerSymbolTable` and associated classes (namespace etc.) is according to the docs to cache the symbol lookup, to avoid repeatedly doing recursive IR scans for every inner symbol lookup (remember inner symbols can reach beyond classic MLIR symbol visibility, that’s their whole point).

This sort of clashes with the xdsl merging of traits and interfaces, as the traits now need to hold mutable state: the lookup tables – essentially a mapping structure that is filled as child nodes are added or with `verify()` / `lookup()` calls. That doesn’t work with traits being frozen dataclasses.

Additionally these lookups are distinct for every inner symbol table and contain their respective descendants, so we would also need to diverge from the classic OpTrait lookup, where all operations share the same trait instance. I think this point is minor compared to the previous one but still bring it up as it likely has the same solution.

I only see 2 possible solutions:
1. drop the caching for now, i.e. have the lookup functions just walk the inner symbol table nodes to find references.  
   The main downside is potentially terrible performance when writing things of significant (or even medium?) size that use inner references. Essentially just kicking the can down the road – only fix it when it becomes an issue.
2. find somewhere else where that cached lookup data can live.  
   From scouring the code the closest I haven’t seen any similar pattern in the existing traits.

I’m not sure if I’m overthinking this or missing something obvious? Could use some input.

lucjaulmes:
- Found a solution for my problem (writing it out seemed to help?), which is to use what upstream does: `InnerSymbolTableCollection` maps from operations to `InnerSymbolTable` instances, instead of just being a set of operations and using trait lookup.
- This has somewhat complicated implications for the untested bits (lookup functions) so I removed those and will add them in an upcoming PR.
- Rebased on master

lucjaulmes:
This is ready for merging - remaining non-covered lines are either just `...` statements or checks that can only be true because they’re already enforced somewhere else as discussed [previously](#issuecomment-1821048651). I’ve left them both for their documentation value and because if the tests that currently enforce them change (say we’re using a subclass of `SymbolTable` with some relaxations?) then it’s still clear they’re needed here.

tobiasgrosser:
@luisacicolini, could you do a first pass on this PR.

xdsl/dialects/hw.py
```diff
@@ -112,6 +121,142 @@ def print_parameters(self, printer: Printer) -> None:
...
+    def __init__(self) -> None:
+        # Create an empty table, not used for lookups
+        ...
+
+    @overload
+    def __init__(self, op: Operation) -> None:
+        ...
+
+    def __init__(self, op: Operation | None = None) -> None:
+        self.symbol_table: Mapping[StringAttr, InnerSymTarget] = dict()
```
superlopuh, line 139: Can you please move the property declaration to the class level? I'm surprised we don't have a ruff rule for this, but we've converged on pretty much always declaring the full list of properties before the inits.

xdsl/dialects/hw.py
```diff
@@ -112,6 +121,142 @@ def print_parameters(self, printer: Printer) -> None:
...
+    """
+
+    @overload
+    def __init__(self) -> None:
+        # Create an empty table, not used for lookups
+        ...
+
+    @overload
+    def __init__(self, op: Operation) -> None:
+        ...
```
superlopuh, line 136: I'm not sure that these give any functional difference to the non-overloaded version. What changes if they are removed?

xdsl/dialects/hw.py
```diff
@@ -112,6 +121,142 @@ def print_parameters(self, printer: Printer) -> None:
...
+            raise VerifyException(
+                f"Operation {op.name} with trait {type(self).__name__} must have a parent with trait {trait.__name__}"
+            )
+
+
+class InnerSymbolTableCollection:
+    """This class represents an InnerSymbolTable collection."""
+
+    def __init__(self, inner_ref_ns_op: Operation | None = None) -> None:
+        self.symbol_tables: dict[Operation, InnerSymbolTable] = {}
```
superlopuh, line 166: same here, could you please move the property declaration out?

xdsl/dialects/hw.py
```diff
@@ -112,6 +121,142 @@ def print_parameters(self, printer: Printer) -> None:
...
+
+class InnerRefUserOpInterface(OpTrait):
+    """This interface describes an operation that may use a `InnerRef`. This
+    interface allows for users of inner symbols to hook into verification and
+    other inner symbol related utilities that are either costly or otherwise
+    disallowed within a traditional operation."""
+
+    def verify_inner_refs(self, op: Operation, namespace: "InnerRefNamespace"):
+        """Verify the inner ref uses held by this operation."""
+        ...
```
superlopuh, line 204: Is this abstract?

xdsl/dialects/hw.py
```diff
@@ -112,6 +121,142 @@ def print_parameters(self, printer: Printer) -> None:
...
+            inner_sym_tables = InnerSymbolTableCollection()
+        super().__init__((symbol_table, inner_sym_tables))
+
+    @property
+    def symbol_table(self) -> SymbolTable:
+        return self.parameters[0]
+
+    @property
+    def inner_sym_tables(self) -> InnerSymbolTableCollection:
+        return self.parameters[1]
```
PapyChacal, line 236: I think you could just declare `symbol_table` and `inner_sym_tables` at the class level and use dataclass's field(factory=...) feature for the default value. You would get all this code from dataclass.
Or am I missing a reason to use this specific tuple/parameters style?

luisacicolini (approved):
looks good to me, as far as I could understand :)

lucjaulmes:
Great feedback, thanks everyone. Regarding @PapyChacal and @superlopuh I have ended up with convoluted initialisations indeed. The issue is essentially the dual use of traits in xdsl.

- As traits, they need to be hashable to be used in a `frozenset()`
- As classes, they need to hold lookup mappings, so `dict`, which are mutable and therefore not hashable
- If initialising outside of `__init__()` using `@dataclass`, this implies `dataclass(frozen=True)` so I can’t use these both ways because now `__setitem__` is removed and I can’t set the `dict` on instances that aren’t used for traits.

If you can suggest a for solution to this I’d much appreciate it.

lucjaulmes:
I think what’s getting me confused is this:

https://github.com/xdslproject/xdsl/blob/c174eb6e596e8c38f1ad1fb75218d7acdb79b5b7/xdsl/traits.py#L23

Taking the example of `InnerRefNamespace`, the trait is just saying `verify()` essentially:

<details><summary>
<a href="https://github.com/llvm/circt/blob/57372957e8365b34ca469299b8c864d830e836a1/include/circt/Dialect/HW/HWOpInterfaces.h#L56-L74">llvm/circt <code>include/circt/Dialect/HW/HWOpInterfaces.h</code> lines 56-74</a>
</summary>

```c++
/// This trait is for operations that define a scope for resolving InnerRef's,
/// and provides verification for InnerRef users (via InnerRefUserOpInterface).
template <typename ConcreteType>
class InnerRefNamespace : public TraitBase<ConcreteType, InnerRefNamespace> {
public:
  static LogicalResult verifyRegionTrait(Operation *op) {
    static_assert(
        ConcreteType::template hasTrait<::mlir::OpTrait::SymbolTable>(),
        "expected operation to be a SymbolTable");


    if (op->getNumRegions() != 1)
      return op->emitError("expected operation to have a single region");
    if (!op->getRegion(0).hasOneBlock())
      return op->emitError("expected operation to have a single block");


    // Verify all InnerSymbolTable's and InnerRef users.
    return ::circt::hw::detail::verifyInnerRefNamespace(op);
  }
};
```
</details>

The interface on the other hand defines a bunch of lookups:

<details><summary>
<a href="https://github.com/llvm/circt/blob/57372957e8365b34ca469299b8c864d830e836a1/include/circt/Dialect/HW/InnerSymbolTable.h#L226-L243">llvm/circt <code>include/circt/Dialect/HW/InnerSymbolTable.h</code> lines 226-243</a>
</summary>

```c++
/// This class represents the namespace in which InnerRef's can be resolved.
struct InnerRefNamespace {
  SymbolTable &symTable;
  InnerSymbolTableCollection &innerSymTables;


  /// Resolve the InnerRef to its target within this namespace, returning empty
  /// target if no such name exists.
  InnerSymTarget lookup(hw::InnerRefAttr inner);


  /// Resolve the InnerRef to its target within this namespace, returning
  /// empty target if no such name exists or it's not an operation.
  /// Template type can be used to limit results to specified op type.
  Operation *lookupOp(hw::InnerRefAttr inner);
  template <typename T>
  T lookupOp(hw::InnerRefAttr inner) {
    return dyn_cast_or_null<T>(lookupOp(inner));
  }
};
```
</details>

I end up doing quite contrived things to make the duality hold. Sometimes the class is used as a trait, and it’s non-comparable attributes are not used. Other times the class is instantiated for lookups, using `__post_init__()` and a `dataclasses.InitVar`.

Alternately, I could just define 2 different classes, one for the trait and one for the lookups. That would be simpler but break the larger pattern. @PapyChacal advocates for this and I think it is sensible. But are there reasons to not do it?

xdsl/dialects/hw.py
lucjaulmes: Just virtual AFAICT. It allows users of inner symbols a way to hook into verifications but there’s no obligation to actually do anything.

PapyChacal:
> Alternately, I could just define 2 different classes, one for the trait and one for the lookups. That would be simpler but break the larger pattern. @PapyChacal advocates for this and I think it is sensible. But are there reasons to not do it?

This is not what I had in mind, but I don't see any issue with this if it helps :slightly_smiling_face: Honestly, do however you feel is most manageable for now. Our Traits/Interfaces implementation is far from optimal, and after having close looks with you on this, I believe your work is suffering from its flaws more than any so far. (as-in, what you're trying to do, not the way you do it; I do not see obvious answers with this state of things TBH).

To repeat it on the record, One thought I have is that you might want to avoid having actual state in your table, and just do naive lookup on the IR. Building the state is okay, but updating it with IR changes is another problem, for which xDSL lacks infrastructure so far. And to repeat my point too, I believe this can be okay for a while, as we usually don't work with heavy loads in xDSL. This will depends on the planned usage of this dialect I guess :slightly_smiling_face:

lucjaulmes:
The lookups are less long-lived than what we initially discussed, so I think updating isn’t too much of an issue here.

@math-fehr do you have insight in design decisions to weigh in on this? If not, can you suggest someone else?

To summarize the issue: the interfaces I’m trying to copy from circt hold state (lookup tables) which doesn’t seem possible with current xDSL implementation (due to uniqueness and frozen dataclass semantics). This implementation uses a `field(compare=False, …)` which isn’t used when the class is instantiated as a trait, and is used when the class is instantiated for lookups.

It’s what fits with the code restrictions but I’m a little uneasy introducing this double-use of a single class without anyone properly saying it’s the best way forward.

math-fehr:
@lucjaulmes Can you show me where an interface is keeping state? This is something that is not allowed in MLIR AFAIK.
In C++, the interfaces are defined on the Operation classes, not on the operations themselves, so I'm not even sure how would that be possible.

lucjaulmes:
The `SymbolTable` and `InnerSymbolTableCollection` contain the lookup info themselves @math-fehr , e.g. [for the former](https://www.github.com/llvm/circt/blob/57372957e8365b34ca469299b8c864d830e836a1/include/circt/Dialect/HW/InnerSymbolTable.h#L184-L196):

```
private:
  using TableTy = DenseMap<StringAttr, InnerSymTarget>;
  /// Construct an inner symbol table for the given operation,
  /// with pre-populated table contents.
  explicit InnerSymbolTable(Operation *op, TableTy &&table)
      : innerSymTblOp(op), symbolTable(table){};


  /// This is the operation this table is constructed for, which must have the
  /// InnerSymbolTable trait.
  Operation *innerSymTblOp;


  /// This maps inner symbol names to their targets.
  TableTy symbolTable;
```

And [for the latter](https://www.github.com/llvm/circt/blob/57372957e8365b34ca469299b8c864d830e836a1/include/circt/Dialect/HW/InnerSymbolTable.h#L221-L223):
```
private:
  /// This maps Operations to their InnnerSymbolTable's.
  DenseMap<Operation *, std::unique_ptr<InnerSymbolTable>> symbolTables;
```

Some more relevant code is linked / quoted in [this comment above](https://github.com/xdslproject/xdsl/pull/1780#issuecomment-1858221569)n

math-fehr:
> The `SymbolTable` and `InnerSymbolTableCollection` contain the lookup info themselves @math-fehr , e.g. [for the former](https://www.github.com/llvm/circt/blob/57372957e8365b34ca469299b8c864d830e836a1/include/circt/Dialect/HW/InnerSymbolTable.h#L184-L196):
> 
> ```c++
> private:
>   using TableTy = DenseMap<StringAttr, InnerSymTarget>;
>   /// Construct an inner symbol table for the given operation,
>   /// with pre-populated table contents.
>   explicit InnerSymbolTable(Operation *op, TableTy &&table)
>       : innerSymTblOp(op), symbolTable(table){};
> 
> 
>   /// This is the operation this table is constructed for, which must have the
>   /// InnerSymbolTable trait.
>   Operation *innerSymTblOp;
> 
> 
>   /// This maps inner symbol names to their targets.
>   TableTy symbolTable;
> ```
> 
> And [for the latter](https://www.github.com/llvm/circt/blob/57372957e8365b34ca469299b8c864d830e836a1/include/circt/Dialect/HW/InnerSymbolTable.h#L221-L223):
> 
> ```c++
> private:
>   /// This maps Operations to their InnnerSymbolTable's.
>   DenseMap<Operation *, std::unique_ptr<InnerSymbolTable>> symbolTables;
> ```
> 
> Some more relevant code is linked / quoted in [this comment above](https://github.com/xdslproject/xdsl/pull/1780#issuecomment-1858221569)n

I see, but they are not used as an interface right? My understanding is that they are returned by the interface right?
SO it would be having an interface that returns this structure

math-fehr:
For instance, to me, these structures are populated here:
https://github.com/llvm/circt/blob/27f35ba7c569dea50467c24427cfb430d3bb45bf/lib/Dialect/HW/InnerSymbolTable.cpp#L156

They are not stored anywhere in the interface AFAIK?

lucjaulmes:
It may be part of the trait more than the interface in itself, but AFAICT those 2 concepts are merged in xDSL. Or I may be confused and missing something?

math-fehr:
So the difference between a trait and an interface is that interfaces have "virtual" methods, while trait do not.
In xDSL I merged both because that just mean that traits are interfaces that have no "virtual" methods.

So to me, I think this is a different problem?

lucjaulmes:
My problem is I can’t hold state in xDSL traits because they’re frozen singletons @math-fehr, and they seem to be the logical place for this to go due to how MLIR concepts are mapped to xDSL (in my understanding at least?)

Are you saying there’s some place I missed where I this lookup info could logically go? Should I use the same class for both usages (current proposal)? Or should I split MLIR (rather, CIRCT) traits into xDSL traits and other “friend” classes? Or something else entirely?

math-fehr:
> My problem is I can’t hold state in xDSL traits because they’re frozen singletons @math-fehr, and they seem to be the logical place for this to go due to how MLIR concepts are mapped to xDSL (in my understanding at least?)
> 
> Are you saying there’s some place I missed where I this lookup info could logically go? Should I use the same class for both usages (current proposal)? Or should I split MLIR (rather, CIRCT) traits into xDSL traits and other “friend” classes? Or something else entirely?

So my understanding is that MLIR has the exact same concept, traits are frozen (and so are interfaces).

My understanding is that the classes you sent me are like "friend" classes that are used by the actual interface?
Or that this is not an interface at all, but something else.

math-fehr:
Sorry about that, missclick.

I'll try to read that MLIR code to really understand what's going on!

math-fehr:
Okay, I think I got the confusion now!

There are actually two `InnerSymbolTrait` classes. One is in the `OpTrait` namespace, and one is not.
The [one you sent](https://github.com/llvm/circt/blob/6212d2ed21c4e06a982d9a253a6b60d24efbaa7f/include/circt/Dialect/HW/InnerSymbolTable.h#L104) is the non-trait one, and is the one with a state.
The trait one [is here](https://github.com/llvm/circt/blob/6212d2ed21c4e06a982d9a253a6b60d24efbaa7f/include/circt/Dialect/HW/HWOpInterfaces.h#L78). The trait has no state here.

If you look at the [non-trait constructor](https://github.com/llvm/circt/blob/27f35ba7c569dea50467c24427cfb430d3bb45bf/lib/Dialect/HW/InnerSymbolTable.cpp#L27), you can see that the `InnerSymbolTable` class (the one with the state) is created at an operation location.

So you should have two structures. One xDSL trait that has no state, and which can be used to create the "state" class

tobiasgrosser:
@lucjaulmes, any update on this?

lucjaulmes:
Rebased on master and squashed in a8ddbbd, should be good to merge.

(Incremental changes with last version of PR can be viewed in e3eec8b.)

superlopuh (approved):
This looks great! I'm not sure how exhaustively we want to check the functionality added here, I would say that it's pretty well tested already, I just thought I'd suggest looking at [the codecov](https://app.codecov.io/gh/xdslproject/xdsl/pull/1780?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=xdslproject) to be sure that the verification errors that aren't raised in the tests aren't as critical as the ones covered in the test already.

xdsl/dialects/hw.py
```diff
@@ -112,6 +119,138 @@ def print_parameters(self, printer: Printer) -> None:
...
+        if not op.has_trait(trait := SymbolTable):
+            raise VerifyException(
+                f"Operation {op.name} must have trait {trait.__name__}"
+            )
+
+        # These 2 checks are in fact redundant, as already checked as a SymbolTable
+        if len(op.regions) != 1:
+            raise VerifyException(f"Operation {op.name} must have a single region")
+        if len(op.regions[0].blocks) != 1:
+            raise VerifyException(f"Operation {op.name} must have a single block")
```
PapyChacal, line 227:
```suggestion
```

PR #1782, opened 2023-11-14
dialects: Add memref.memory_space_cast

jorendumoulin:
This PR adds the `memref.memory_space_cast` op to the `memref` dialect.
We intend to use this to properly handle `memref` transfers between the different memory spaces in a snitch cluster.

### DIFF
--- a/tests/dialects/test_memref.py
+++ b/tests/dialects/test_memref.py
@@ -22,8 +22,8 @@ from xdsl.dialects.memref import (
     DmaWaitOp,
     ExtractAlignedPointerAsIndexOp,
     Load,
-    MemRefType,
     MemorySpaceCast,
+    MemRefType,
     Store,
     Subview,
     UnrankedMemrefType,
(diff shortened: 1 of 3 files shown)


xdsl/dialects/memref.py
```diff
@@ -587,6 +587,33 @@ def get(
...
+
+    source: Operand = operand_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
+    dest: OpResult = result_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
+
+    @staticmethod
+    def get(
+        source: SSAValue | Operation,
+        type: MemRefType[Attribute] | UnrankedMemrefType[Attribute],
+    ):
+        return MemorySpaceCast.build(operands=[source], result_types=[type])
```
JosseVanDelm, line 602: This seems to be using the old style of creating new operations with `get` and `build`.
I believe new operations are now added with an `__init__` instead e.g.: 
https://github.com/xdslproject/xdsl/blob/a030c55d5e6edee72c0a334618290a9bc03689f7/xdsl/dialects/riscv_scf.py#L47_78

tests/dialects/test_memref.py
```diff
@@ -259,6 +260,39 @@ def test_memref_cast():
...
+    memory_space_cast = MemorySpaceCast.get(memref_ssa_value, res_type)
+
+    assert memory_space_cast.source is memref_ssa_value
+    assert memory_space_cast.dest.type is res_type
+
+    with pytest.raises(
+        VerifyException,
+        match="Expected source and destination to have the same element type.",
+    ):
+        MemorySpaceCast.get(memref_ssa_value, res_type_wrong_type).verify()
```
JosseVanDelm, line 288: see other comment on using `__init__` style constructors

JosseVanDelm (changes requested):
@jorendumoulin nice PR!
I think xdsl has moved on to a different style of constructing ops though, so I think it makes sense to move to this new style.
Wondering what @PapyChacal and/or @AntonLydike think about this?

xdsl/dialects/memref.py
jorendumoulin: Updated this to use new constructor style

tests/dialects/test_memref.py
jorendumoulin: now using new constructor

superlopuh (approved):

xdsl/dialects/memref.py
```diff
@@ -587,6 +587,33 @@ def get(
         return Cast.build(operands=[source], result_types=[type])
 
 
+@irdl_op_definition
+class MemorySpaceCast(IRDLOperation):
+    name = "memref.memory_space_cast"
+
+    source: Operand = operand_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
+    dest: OpResult = result_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
```
compor, line 595:
```suggestion
    source = operand_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
    dest = result_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
```

compor (approved):

xdsl/dialects/memref.py
```diff
@@ -587,6 +587,33 @@ def get(
...
+
+    def __init__(
+        self,
+        source: SSAValue | Operation,
+        dest: MemRefType[Attribute] | UnrankedMemrefType[Attribute],
+    ):
+        super().__init__(operands=[source], result_types=[dest])
```
PapyChacal, line 602: If I understand correctly, the only thing that could change in the result type is the memory space, right?
If so, I would make the constructor just take the new memory space and construct the corresponding MemrefType from the input one, for user-friendliness

(Just an idea, not a requirement for merging at all :slightly_smiling_face: )

PapyChacal (approved):
Welcome and thanks!! Super clean PR, I only left a minor API comment, to only apply if you like it. This is ready to go either way IMO :slightly_smiling_face:

xdsl/dialects/memref.py
superlopuh: Sorry to give conflicting advice, but I have a feeling that there was a benefit of the previous version, in that it seems more flexible. The new initialiser might not be usable in the context of a rewrite, where the type of the ssa value being passed might be invalid, and fixed later on in the compiler pass. I would recommend reverting the latest change, and adding the new function either as a helper, or as an overload to the init, like in `arith.constant`

xdsl/dialects/memref.py
```diff
@@ -587,6 +587,32 @@ def get(
...
+    dest = result_def(MemRefType[Attribute] | UnrankedMemrefType[Attribute])
+
+    def __init__(
+        self,
+        source: SSAValue | Operation,
+        dest_space: Attribute,
+    ):
+        source = SSAValue.get(source)
+
+        assert isinstance(source.type, MemRefType)
```
superlopuh, line 604: this is not true in the general case, as during rewrites the types might not be the expected ones

xdsl/dialects/memref.py
superlopuh: this is great, init + constructor seems like the best way to go for now

JosseVanDelm (approved):
B e a utiful

PR #1786, opened 2023-11-15
backend: (riscv) Exclude FP registers from RISC-V regalloc in the presence of `snitch_stream` IO ops

compor:
This PR adds:

- Exclusion of FP registers (currently `ft0-ft2`) from RISC-V register allocation in the presence of `snitch_stream` IO ops (`strided_read` and `strided_write`).
- Tests of the above functionality.

Resolves https://github.com/opencompl/riscv-paper-experiments/issues/63

### DIFF
--- a/xdsl/backend/riscv/register_allocation.py
+++ b/xdsl/backend/riscv/register_allocation.py
@@ -35,11 +35,10 @@ def gather_allocated(func: riscv_func.FuncOp) -> set[RISCVRegisterType]:
 def _uses_snitch_stream(func: riscv_func.FuncOp) -> bool:
     """Utility method to detect use of read/write ops of the `snitch_stream` dialect."""
 
-    for op in func.walk():
-        if isinstance(op, snitch_stream.StridedReadOp | snitch_stream.StridedWriteOp):
-            return True
-
-    return False
+    return any(
+        isinstance(op, snitch_stream.StridedReadOp | snitch_stream.StridedWriteOp)
+        for op in func.walk()
+    )
 
 
 class RegisterAllocator(abc.ABC):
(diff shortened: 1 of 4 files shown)


adutilleul (approved):

xdsl/transforms/snitch_register_allocation.py
```diff
@@ -13,6 +13,18 @@
 )
 
 
+def get_snitch_reserved() -> set[riscv.FloatRegisterType]:
+    """Utility method to make explicit the Snitch ISA assumptions wrt the
+    floating-point registers that are considered reserved.
+    Currently, the first 3 floating-point registers are reserved.
+    """
```
superlopuh, line 20:
```suggestion
    """
    Utility method to make explicit the Snitch ISA assumptions wrt the
    floating-point registers that are considered reserved.
    Currently, the first 3 floating-point registers are reserved.
    """
```

xdsl/backend/riscv/register_allocation.py
```diff
@@ -31,6 +32,16 @@ def gather_allocated(func: riscv_func.FuncOp) -> set[RISCVRegisterType]:
...
 
 
+def _uses_snitch_stream(func: riscv_func.FuncOp) -> bool:
+    """Utility method to detect use of read/write ops of the `snitch_stream` dialect."""
+
+    for op in func.walk():
+        if isinstance(op, snitch_stream.StridedReadOp | snitch_stream.StridedWriteOp):
+            return True
+
+    return False
```
superlopuh, line 42: supernit:
```suggestion
    return any(isinstance(op, snitch_stream.StridedReadOp | snitch_stream.StridedWriteOp) for op in func.walk())
```

superlopuh (approved):
Looks great

xdsl/backend/riscv/register_allocation.py
compor: much better, thanks!

PR #1788, opened 2023-11-15
interactive: change to ListView (1/7)

dshaaban01:
Instead of doing a selection list, I have changed it to a list view. This allows us to apply the SAME pass multiple times instead of only being allowed to apply one pass type one time. This is also useful to introduce the condense button functionality (which was difficult to cleanly implement under the SelectionList set up).

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ tests-toy: filecheck-toy pytest-toy
 
 # run all tests
 tests: pytest tests-toy filecheck pytest-nb pyright
-	@echo test
+	@echo All tests done.
 
 # re-generate the output from all jupyter notebooks in the docs directory
 rerun-notebooks:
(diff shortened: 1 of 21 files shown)


compor:
Screenshot or it didn't happen :)

tests/interactive/test_app.py
```diff
@@ -76,169 +74,3 @@ async def test_inputs():
 
         assert isinstance(app.current_module, ModuleOp)
         assert app.current_module.is_structurally_equivalent(expected_module)
-
```
superlopuh, line 79: feels like a shame to remove the test entirely. Let's keep as much of the functionality covered as makes sense. For example, we can still test that the outputs are updated in the way that we expect when the model object changes, even if we don't test the user action for now.

superlopuh (changes requested):

superlopuh (approved):

superlopuh:
Looks like a bug in the CI, makes sense since PRs on top of this one pass, it happens sometimes, I just restarted it

PR #1792, opened 2023-11-15
dialects: (riscv_snitch) add FrepYieldOp to dialect definition

superlopuh:
A few things going on here:
1. turns out FrepYieldOp was never added to the dialect definition, oops
2. I made it an AbstractYieldOp but that broke irdl_op_def for some reason, @math-fehr will need your ok for my fix
3. Feels like it should be implicit, frep almost never yields things, so might as well leave it out by default

### DIFF
--- a/tests/filecheck/dialects/riscv/frep_verification.mlir
+++ b/tests/filecheck/dialects/riscv/frep_verification.mlir
@@ -36,3 +36,13 @@ riscv_snitch.frep_outer %i0 {
 }
 
 // CHECK: Operation does not verify: Frep operation body may not contain instructions with side-effects, found riscv.sw
+
+// -----
+
+%0 = "test.op"(): () -> !riscv.reg<>
+
+"riscv_snitch.frep_outer"(%0) ({
+^bb0:
+}) {"stagger_mask" = #builtin.int<0>, "stagger_count" = #builtin.int<0>} : (!riscv.reg<>) -> ()
+
+// CHECK: Operation riscv_snitch.frep_outer contains empty block in single-block region that expects at least a terminator
--- a/xdsl/dialects/riscv_snitch.py
+++ b/xdsl/dialects/riscv_snitch.py
@@ -17,7 +17,9 @@ from xdsl.dialects.riscv import (
     RISCVInstruction,
     RISCVOp,
 )
+from xdsl.dialects.utils import AbstractYieldOperation
 from xdsl.ir import (
+    Attribute,
     Block,
     Dialect,
     Operation,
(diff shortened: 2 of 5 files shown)


math-fehr:
Yes, this will be fixed (by your fix) in #1794.
Thanks for bringing this up

xdsl/utils/hints.py
```diff
@@ -141,7 +141,9 @@ def get_type_var_mapping(
     # Get the generic parent
     orig_bases: Iterable[Any] = getattr(cls, "__orig_bases__")
     orig_bases = [
-        orig_base for orig_base in orig_bases if get_origin(orig_base) is not Generic
+        orig_base
```
superlopuh, line 144: now that Mathieu has fixed this in #1794, please ignore changes in this file, I'll merge this PR only after that one

xdsl/dialects/riscv_snitch.py
```diff
@@ -129,8 +130,6 @@ class FRepOperation(IRDLOperation, RISCVInstruction):
     wraps again (up to 23 = 8).
     """
 
-    traits = frozenset((NoTerminator(),))
-
```
compor, line 133: I'm a bit confused by point 3 in your PR message;
why not use the `SingleBlockImplicitTerminator` trait like in `scf.for` and call `ensure_terminator` below?

xdsl/dialects/riscv_snitch.py
```diff
@@ -177,6 +176,11 @@ def parse(cls, parser: Parser) -> Self:
 
         body = parser.parse_region()
 
+        if body.blocks and (
```
compor, line 179: Call `ensure_terminator` here?

xdsl/dialects/riscv_snitch.py
```diff
@@ -212,6 +221,18 @@ def verify_(self) -> None:
                     f"with side-effects, found {instruction.name}"
                 )
 
+        block = self.body.blocks[0]
```
compor, line 224: Using the trait, you won't need all these checks

compor (commented):
A bit confused, left some comments for this.

xdsl/dialects/riscv_snitch.py
superlopuh: I somehow missed that trait, will add it now

tests/filecheck/dialects/riscv/frep_verification.mlir
```diff
@@ -36,3 +35,40 @@ riscv_snitch.frep_outer %i0 {
 }
 
 // CHECK: Operation does not verify: Frep operation body may not contain instructions with side-effects, found riscv.sw
+
```
compor, line 38: I don't think we need any of these tests; they just replicate the trait's expected behaviour which has already been tested.

compor (approved):

tests/filecheck/dialects/riscv/frep_verification.mlir
superlopuh: I've been thinking about this, and I'm not sure still one way or the other. The behaviour of the op is distinct from the implementation, so they are decouple from whether a trait is used or custom code like I had in an earlier state of this PR. I think that it makes sense for something to fail if we remove the implicit terminator trait from `FrepOperation`. But then maybe that's overly defensive.

compor: Yeah, I don't know, we haven't been doing in general AFAIU.
Happy to keep it if your think the API is still in flux here.

PR #1793, opened 2023-11-15
interactive: condense feature added and condense_pass list calculated dynamically! (3/7)

dshaaban01:
Condense feature has been added. It is computed based on the current_module (i.e. the output) and is recomputed every time a pass is applied (i.e. the dynamic update we discussed).

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -9,13 +9,11 @@ be sure to install `textual-dev` to run this command.
 
 from io import StringIO
 
-from rich.style import Style
 from textual import events, on
 from textual.app import App, ComposeResult
 from textual.containers import Horizontal, ScrollableContainer, Vertical
 from textual.reactive import reactive
 from textual.widgets import Button, Footer, Label, ListItem, ListView, TextArea
-from textual.widgets.text_area import TextAreaTheme
 
 from xdsl.dialects import builtin
 from xdsl.dialects.builtin import ModuleOp
@@ -74,16 +72,6 @@ class InputApp(App[None]):
         ("q", "quit_app", "Quit"),
     ]
 
-    # defines a theme for the Input/Output TextArea
-    _DEFAULT_THEME = TextAreaTheme(
-        name="my_theme_design",
-        base_style=Style(bgcolor="white"),
-        syntax_styles={
-            "string": Style(color="red"),
-            "comment": Style(color="magenta"),
-        },
-    )
-
     current_module = reactive[ModuleOp | Exception | None](None)
     """
     Reactive variable used to save the current state of the modified Input TextArea
@@ -140,11 +128,9 @@ class InputApp(App[None]):
     def on_mount(self) -> None:
         """Configure widgets in this application before it is first shown."""
 
-        # register the theme for the Input/Output TextAreas
-        self.input_text_area.register_theme(InputApp._DEFAULT_THEME)
-        self.output_text_area.register_theme(InputApp._DEFAULT_THEME)
-        self.input_text_area.theme = "my_theme_design"
-        self.output_text_area.theme = "my_theme_design"
+        # register's the theme for the Input/Output TextArea's
+        self.input_text_area.theme = "vscode_dark"
+        self.output_text_area.theme = "vscode_dark"
 
         self.query_one("#input_container").border_title = "Input xDSL IR"
         self.query_one("#output_container").border_title = "Output xDSL IR"
--- a/xdsl/interactive/app.tcss
+++ b/xdsl/interactive/app.tcss
@@ -1,7 +1,7 @@
 # ListView
 #passes_list_view{
     border: heavy $accent-darken-1;
-    border-title-color: $warning-lighten-1;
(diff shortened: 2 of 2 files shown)


xdsl/interactive/app.py
```diff
@@ -147,8 +178,24 @@ def update_current_module(self) -> None:
...
             pipeline.apply(ctx, module)
             self.current_module = module
+            # for current_module's that are valid ModuleOp's find the condensed pass list
+            self.current_condensed_pass_list = condensed_pass_list(self.current_module)
+
+            # trigger watch_condense_mode function every time current_module is updated to recalculate/reupdate condensed_pass_list
+            if self.condense_mode is True:
+                self.condense_mode = True
+            else:
+                self.condense_mode = False
```
webmiche, line 188: This kinds of code feels very weird. I know, it is supposed to trigger the update-function on a write to the variable, but this kind of "implicit function calling" is not very understandable. Can you not just call the update function directly?

xdsl/interactive/app.py
```diff
@@ -30,6 +31,32 @@
 """Contains the list of xDSL passes."""
 
 
+def condensed_pass_list(input: builtin.ModuleOp) -> tuple[type[ModulePass], ...]:
+    """Used to check if a pass has had an effect on the IR (i.e. has changed the IR)."""
```
compor, line 35:
```suggestion
    """Returns a tuple of passes (pass name and pass instance) that modify the IR."""
```

xdsl/interactive/app.py
```diff
@@ -147,8 +178,24 @@ def update_current_module(self) -> None:
             pipeline = PipelinePass([p() for p in self.current_selected_pass_list])
             pipeline.apply(ctx, module)
             self.current_module = module
+            # for current_module's that are valid ModuleOp's find the condensed pass list
```
compor, line 181:
```suggestion
```

I think the comment is a bit disorienting; since we are in the `try` of a `try/catch` block, let's just assume `module` has been parsed successfully and just follow the happy path scenario.

xdsl/interactive/app.py
```diff
@@ -147,8 +178,24 @@ def update_current_module(self) -> None:
...
+            self.current_condensed_pass_list = condensed_pass_list(self.current_module)
+
+            # trigger watch_condense_mode function every time current_module is updated to recalculate/reupdate condensed_pass_list
+            if self.condense_mode is True:
+                self.condense_mode = True
+            else:
+                self.condense_mode = False
         except Exception as e:
             self.current_module = e
+            # for current_module's that are exceptions, the condensed pass list is empty
```
compor, line 191:
```suggestion
```

compor (changes requested):
I left some minor comments but I think @webmiche concern is the main item to be addressed in this PR.

xdsl/interactive/app.py
superlopuh: Textual recommends doing that as little as possible, instead treating UI as derived from the model, and adds machinery to recompute the UI when the model changes. So I would say we want to stay away from calling the updarte function directly as much as possible. I agree with you that this particular bit of code is confusing, which I think is more related to the fact that there hook to update the available list of passes is missing. I think this PR would benefit from a refactor to set up the reactive hook to recompute the list of available passes when the current module changes.

compor (approved):

xdsl/interactive/app.py
```diff
@@ -167,24 +222,31 @@ def watch_current_module(self):
...
-        self.query_one("#input_container").border_title = "Input xDSL IR"
-        self.query_one("#output_container").border_title = "Output xDSL IR"
-        self.query_one(
-            "#passes_list_view"
-        ).border_title = "Choose a pass or multiple passes to be applied."
-        self.query_one("#selected_passes").border_title = "Selected passes/query"
-
-        for n, _ in ALL_PASSES:
-            self.passes_list_view.append(ListItem(Label(n), name=n))
+    def trigger(self):
```
superlopuh, line 225: My high-level model of how this works is "when either the current module or condense mode changes, update the current available passes". My understanding of textual is that the idiomatic way to do this is with a `compute_current_condensed_pass_list`, which would call the helper function above to update the list. Could you please give that a go?

superlopuh (changes requested):

xdsl/interactive/app.py
```diff
@@ -106,6 +137,51 @@ def compose(self) -> ComposeResult:
...
+        self.query_one("#output_container").border_title = "Output xDSL IR"
+        self.query_one(
+            "#passes_list_view"
+        ).border_title = "Choose a pass or multiple passes to be applied."
+        self.query_one("#selected_passes").border_title = "Selected passes/query"
+
+        for n, _ in ALL_PASSES:
+            self.passes_list_view.append(ListItem(Label(n), name=n))
+
+        self.condense_mode = False
```
superlopuh, line 159: We might as well make this the initial value and remove this line

superlopuh (changes requested):
Much clearer, thank you

superlopuh (approved):

PR #1794, opened 2023-11-16
core: Allow multiple inheritance when one parent is generic

math-fehr:
This allows Operation classes to inherit from multiple classes, when one of the classes is a generic.
Before that, this was not allowed.


superlopuh (approved):
Thank you!

webmiche (approved):

PR #1799, opened 2023-11-16
interactive: toggle condense/uncondense button (4/7)

dshaaban01:
Added uncondense button functionality.

image


superlopuh:
Why two buttons? Could you make the condense button say "Uncondense" when it's condensed? There's an example of how to do this with tcss and hiding/showing buttons based on a boolean state in the tutorial

compor (commented):
A "toggle" button would work best as @superlopuh suggested

superlopuh:
The button code looks great but it feels like there might be some unrelated code in here also?

superlopuh:
I don't get where those errors are introduced, seems like something's off with the stacked branches

PR #1801, opened 2023-11-16
interactive: preload initial example(6/7)

dshaaban01:
preload initial example into gui so users can gain a quicker understanding + theme changes to make sure that text area is visible to everyone (@tobiasgrosser had some issues seeing the text in the text area as the background colour+text colour of TextArea's were similar ). 

Light Mode: 
image

Dark Mode:
image

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -192,18 +192,12 @@ async def test_buttons():
         )
 
         # press "Remove Last Pass" button
+        current_pipeline = app.pass_pipeline
         await pilot.click("#remove_last_pass_button")
         await pilot.pause()
 
         # assert that the pass pipeline and Output Text Area change accordingly
-        expected_pass_pipeline = ()
-        expected_pass_pipeline = tuple(
-            (
-                *expected_pass_pipeline,
-                convert_func_to_riscv_func.ConvertFuncToRiscvFuncPass,
-            )
-        )
-        assert app.pass_pipeline == expected_pass_pipeline
+        assert app.pass_pipeline == current_pipeline[:-1]
         assert (
             app.output_text_area.text
             == """builtin.module {


compor (approved):
LGTM, an external config file for the preloaded text, the themes, etc might be something to consider in the future.

xdsl/interactive/app.py
```diff
@@ -140,12 +131,8 @@ def compose(self) -> ComposeResult:
 
     def on_mount(self) -> None:
         """Configure widgets in this application before it is first shown."""
-
-        # register's the theme for the Input/Output TextArea's
-        self.input_text_area.register_theme(InputApp._DEFAULT_THEME)
-        self.output_text_area.register_theme(InputApp._DEFAULT_THEME)
-        self.input_text_area.theme = "my_theme_design"
-        self.output_text_area.theme = "my_theme_design"
+        self.input_text_area.theme = "vscode_dark"
```
superlopuh, line 134: This is unrelated to the preloading of the text, could you please split out the two PRs? They also don't seem like they would need to be in a stack as they don't functionally depend on anything not in main

superlopuh (changes requested):

xdsl/interactive/app.py
superlopuh: Yes

xdsl/interactive/app.py
```diff
@@ -161,6 +164,19 @@ def on_mount(self) -> None:
         # initialize gui to inital state - uncondensed mode.
         self.condense_mode = False
 
+        # initialize GUI with an interesting input IR and pass application
+        self.input_text_area.load_text(
+            """func.func @hello(%n : index) -> index {
```
superlopuh, line 169: Can you please move this out to an UPPER_CAMEL_CASE constant at the level of the file? Maybe called `INITIAL_IR_TEXT`, or something like that?

superlopuh (changes requested):

webmiche (approved):

PR #1802, opened 2023-11-16
[wip] interactive: load files into gui (7/7)

dshaaban01:
Goal of this PR:
1. Add ability for users to run `xdsl-gui <filename>` command to preload a file into gui input textarea
3. Provide ability _inside_ gui to do this: i.e add textarea where users type file name, add press "load file" and have the specified file loaded in Input TextArea

So far: text area and buttons widgets added for goal 2) but have not made it work yet. Advice/thoughts on this functionality welcome.

image

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -9,11 +9,13 @@ be sure to install `textual-dev` to run this command.
 
 from io import StringIO
 
+from rich.style import Style
 from textual import events, on
 from textual.app import App, ComposeResult
 from textual.containers import Horizontal, ScrollableContainer, Vertical
 from textual.reactive import reactive
 from textual.widgets import Button, Footer, Label, ListItem, ListView, TextArea
+from textual.widgets.text_area import TextAreaTheme
 
 from xdsl.backend.riscv.lowering import (
     convert_func_to_riscv_func,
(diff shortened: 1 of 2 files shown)


xdsl/interactive/app.py
```diff
@@ -301,6 +305,10 @@ def uncondense(self, event: Button.Pressed) -> None:
     def remove_last_pass(self, event: Button.Pressed) -> None:
         self.pass_pipeline = self.pass_pipeline[:-1]
 
+    @on(Button.Pressed, "#load_file_button")
+    def load_file(self, event: Button.Pressed) -> None:
```
compor, line 309: Does nothing?

compor (commented):
Not sure where this is going, but I would have preferred a file browser to pop up here.
I know this is a bigger feature, but now I have to move in and out of the application.
Also, I have no hints on what is the default path for the file I want to load.

Just a comment, so I'm not going to block this.

superlopuh:
Could you please make this a draft PR until it's ready to review? Makes it a bit easier to prioritise PRs

webmiche:
Personally, I would first focus on only the first part before trying to load files while the gui is running. This should be relatively simple by changing the `INITIAL_IR_TEXT` variable.

PR #1808, opened 2023-11-20
interactive: preload initial example(6/7)

dshaaban01:
preload initial example

made a second version of this PR as the first one had bizarre failing tests and this was easier.

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -15,9 +15,6 @@ from textual.containers import Horizontal, ScrollableContainer, Vertical
 from textual.reactive import reactive
 from textual.widgets import Button, Footer, Label, ListItem, ListView, TextArea
 
-from xdsl.backend.riscv.lowering import (
-    convert_func_to_riscv_func,
-)
 from xdsl.dialects import builtin
 from xdsl.dialects.builtin import ModuleOp
 from xdsl.ir import MLContext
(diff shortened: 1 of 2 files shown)


superlopuh:
can you retarget this on the other branch, please? Or remove the changes from the remove last pass button PR?

tobiasgrosser:
> can you retarget this on the other branch, please? Or remove the changes from the remove last pass button PR?

I merged master, so this should be fixed.

@dshaaban01, in a perfect world independent PRs do not spill into each other.

xdsl/interactive/app.py
```diff
@@ -143,6 +154,12 @@ def on_mount(self) -> None:
         for n, _ in ALL_PASSES:
             self.passes_list_view.append(ListItem(Label(n), name=n))
 
+        # initialize GUI with an interesting input IR and pass application
+        self.input_text_area.load_text(self.INITIAL_IR_TEXT)
```
superlopuh, line 158: nit: 
```suggestion
        self.input_text_area.load_text(InputApp.INITIAL_IR_TEXT)
```

xdsl/interactive/app.py
```diff
@@ -143,6 +154,12 @@ def on_mount(self) -> None:
...
+        self.pass_pipeline = tuple(
+            (*self.pass_pipeline, convert_func_to_riscv_func.ConvertFuncToRiscvFuncPass)
```
superlopuh, line 160: I would recommend going with something that will be relevant to any project rather than just the riscv thing, it seems like `dce` or `canonicalize` might be more universally applicable. I also think there's a case to be made for no passes being interesting enough, just to see how it parses and re-prints the code. There's also fewer actions to make to remove the preloaded pass in that case.

superlopuh (approved):

PR #1809, opened 2023-11-20
interactive: add button "load file" which opens directory tree allowing file selection +load into text area (7/7) 

dshaaban01:
add button "load file" which opens directory tree. select the file you want to open, click on it, and the file is loaded into the input text area, where you can now start working on pass application and visualisation. 

image

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -172,7 +172,6 @@ class InputApp(App[None]):
         # initialize GUI with an interesting input IR and pass application
         self.input_text_area.load_text(InputApp.INITIAL_IR_TEXT)
 
-
     def compute_available_pass_list(self) -> tuple[type[ModulePass], ...]:
         match self.current_module:
             case None:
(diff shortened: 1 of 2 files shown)


xdsl/interactive/app.tcss
```diff
@@ -62,3 +79,35 @@ Screen {
...
+# LoadFile Screen
+LoadFile{
+    background: rgba(100,100,100,0.5);
+    align: center middle;
+    layout: vertical;
+}
+
+# DirectoryTree
+#directory_tree{
+    background: white;
```
superlopuh, line 92: does this work in both light and dark mode?

xdsl/interactive/app.py
```diff
@@ -52,6 +67,35 @@ def condensed_pass_list(input: builtin.ModuleOp) -> tuple[type[ModulePass], ...]
     return selections
 
 
+class LoadFile(Screen[str]):
```
superlopuh, line 70: This screen seems like a good piece of code to factor out into its own file, as well as its corresponding tcss file.

PR #1811, opened 2023-11-21
[wip] interactive: add ability to input parameters on mlir opt arguments (2/2)

dshaaban01:
When mlir-opt pass is selected, a text area pops up where the user is able to write a list of mlir passes they would like to apply as arguments to the mlir pass. 
image

after pressing the enter button, the query label is updated and the output should change accordingly. unfortunately no matter what mlir passes I write, I get the same error. 
image

@superlopuh @math-fehr I'm getting an "mliropt pass is not callable error" and idk what that means? is it that the input IR does not suit the mlir-opt pass? I'm a bit confused.

BTW: This is the final PR for this tool (for now - open for reevaluation later) as I think I have achieved all of the basic goals that we wanted the tool to accomplish and I think it is time to move on to peephole rewrites.

### DIFF
--- a/xdsl/interactive/mlir_opt_pass_screen.py
+++ b/xdsl/interactive/mlir_opt_pass_screen.py
@@ -1,0 +1,44 @@
+from textual import on
+from textual.app import ComposeResult
+from textual.containers import Horizontal, ScrollableContainer
+from textual.screen import Screen
+from textual.widgets import Button, TextArea
+
+
+class MlirOptPassScreen(Screen[str]):
+    CSS_PATH = "mlir_opt_pass_screen.tcss"
+
+    mlir_opt_argument_text_area: TextArea
+
+    def __init__(self):
+        self.mlir_opt_argument_text_area = TextArea(
+            "", id="mlir_opt_argument_text_area"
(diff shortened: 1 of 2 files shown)


xdsl/interactive/mlir_opt_pass_screen.py
```diff
@@ -0,0 +1,44 @@
+from textual import on
+from textual.app import ComposeResult
+from textual.containers import Horizontal, ScrollableContainer
+from textual.screen import Screen
+from textual.widgets import Button, TextArea
+
+
+class MlirOptPassScreen(Screen[str]):
```
superlopuh, line 8: Why the separate screen? Multiple of the passes accept arguments, it would be good to handle all of them, and ones that xDSL doesn't know about when used as a dependency.

superlopuh: I think that sort of makes sense as a first step, it would be nice if the UI could actually present the user with the list of available arguments, and default values, but we don't have APIs for that yet. I think to start with it could just take a string and it would be up to the user to make sure that it works.

PR #1821, opened 2023-12-01
interactive: add comments + docstrings (1/4)

dshaaban01:
Add comments and docstrings

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -106,7 +106,10 @@ class InputApp(App[None]):
     condense_mode = reactive(False, always_update=True)
     """Reactive boolean."""
     available_pass_list = reactive(tuple[type[ModulePass], ...])
-    """Reactive variable that saves the list of passes that have an effect on current_module."""
+    """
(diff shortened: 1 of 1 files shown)


xdsl/interactive/app.py
```diff
@@ -118,6 +121,10 @@ class InputApp(App[None]):
     """ListView displaying the passes available to apply."""
 
     def __init__(self):
+        """
```
superlopuh, line 124: Not sure this one is informative enough to be useful, let's drop it

xdsl/interactive/app.py
```diff
@@ -154,25 +161,30 @@ def compose(self) -> ComposeResult:
 
     def on_mount(self) -> None:
         """Configure widgets in this application before it is first shown."""
-        # register's the theme for the Input/Output TextArea's
+        # register's the theme for the Input/Output TextAreas
```
superlopuh, line 164:
```suggestion
        # Registers the theme for the Input/Output TextAreas
```

PR #1824, opened 2023-12-01
misc: Bump MLIR to 98e674c9f16d677d95c67bc130e267fae331e43c.

PapyChacal:
Hey, it's MLIR bumping time again :tada: 
For the record, my motivation is a fix: the current commit can't build with CUDA support against CUDA<12, and that has been fixed meanwhile.

This one should be relatively easy from what I see:
- `llvm.ptr` is now opaque, so it has no information on the pointee's type but only an adress space. Any pointee type information needed must then be recovered somehow, this has implications on some lowerings.
- Bye bye `arith.maxf` :cry: Hello `arith.maxinumf` and `arith.maxnumf`! https://mlir.llvm.org/docs/Dialects/ArithOps/#arithmaximumf-arithmaximumfop
- `llvm.call` now requires CConv (calling convention) and callee_type properties.
- `affine` switched from attributes to properties

This `llvm.ptr` has non-trivial consequences on @gabrielrodcanal's HLS passes. Because they are untested and in experimental, I left them unchanged and ignored the pyright error here. Let's fix those asynchronously!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
 dynamic = ["version"]
 
 [project.optional-dependencies]
-extras = ["riscemu==2.2.5", "wgpu==0.13.0", "textual==0.43.2", "pyclip==0.7"]
+extras = ["riscemu==2.2.5", "wgpu==0.13.0", "textual==0.44.0", "pyclip==0.7"]
 
 [project.urls]
 Homepage = "https://xdsl.dev/"
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ asv<0.7
 isort==5.12.0
 nbconvert>=7.7.2,<8.0.0
 textual-dev==1.2.1
-pytest-asyncio==0.21.1
+pytest-asyncio==0.23.1
 # pyright version has to be fixed with `==`. The CI parses this file
 # and installs the according version for typechecking.
 pyright==1.1.337
--- a/tests/filecheck/dialects/riscv/frep_verification.mlir
+++ b/tests/filecheck/dialects/riscv/frep_verification.mlir
@@ -1,9 +1,8 @@
 // RUN: xdsl-opt --split-input-file --verify-diagnostics %s | filecheck %s
 
-%i0 = "test.op"() : () -> !riscv.reg<a0>
-%ft0, %ft1 = "test.op"() : () -> (!riscv.freg<ft0>, !riscv.freg<ft1>)
-riscv_snitch.frep_outer %i0, 1, 0 {
-}
+%i0 = "test.op"() : () -> !riscv.reg<>
+"riscv_snitch.frep_outer"(%i0) ({
+}) {"stagger_mask" = #builtin.int<0>, "stagger_count" = #builtin.int<0>} : (!riscv.reg<>) -> ()
 
 // CHECK: expected a single block, but got 0 blocks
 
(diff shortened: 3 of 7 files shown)


PapyChacal:
@math-fehr The only remaining problem is parsing `!llvm.func<i32 (ptr, ...)>`; xDSL doesn't like `ptr`, and only knows `!llvm.ptr`
Given you worked on the attributes/types names area of MLIR, I'd appreciate your input!
I checked, and (this) MLIR is happy to parse `%cst = arith.constant {"attr" = !llvm.func<ptr (i64, ...)>} 10 : i32`
Giving me two interpretations:
- The parameters of !llvm.func are custom parsed in such a way that we can elide all llvm prefixes
- In general, attribute parameters have scope-sensible parsing wrt to the containing attribute, like operations do wrt to their parent operation.

Would you know which it is if any?

math-fehr:
> @math-fehr The only remaining problem is parsing `!llvm.func<i32 (ptr, ...)>`; xDSL doesn't like `ptr`, and only knows `!llvm.ptr` Given you worked on the attributes/types names area of MLIR, I'd appreciate your input! I checked, and (this) MLIR is happy to parse `%cst = arith.constant {"attr" = !llvm.func<ptr (i64, ...)>} 10 : i32` Giving me two interpretations:
> 
>     * The parameters of !llvm.func are custom parsed in such a way that we can elide all llvm prefixes
> 
>     * In general, attribute parameters have scope-sensible parsing wrt to the containing attribute, like operations do wrt to their parent operation.
> 
> 
> Would you know which it is if any?
> 
> EDIT: I see it's at least a common pattern, I think it comes from their declarative assembly format : https://discourse.llvm.org/t/rfc-elide-type-attribute-prefixes-when-using-declarative-assembly/4759/11 We could add some state to the parser to implement this in general? Like a stack oialects, pushing to it the dialect of the attribute we're currently parsing parameters for. EDIT2: I also see some custom fixing for void already in there. I'll try just adding `ptr` there for now
> 
> EDIT3: **Yep nevermind**, I fixed it in the custom printer/parser. If there isn't any, it would be nice to make an issue regarding this and try to implement a general solution 🤔

Yes, sorry for the delay!
`llvm` is implementing a custom solution to not have to print `!llvm`. There are currently no generic solution, because MLIR types don't have names (or at least didn't until I added them 2 days ago).
So there are currently no generic solution (the LLVM one is hardcoded), but I'm working to make this a generic solution in MLIR, so then we could do the exact same here!

math-fehr (approved):
Nice!

gabrielrodcanal:
Opaque pointers were always a problem with FPGA. In fact, the Xilinx's compiler is built in LLVM v7 and I don't think they'll ever upgrade it to the newer versions for this reason. I guess it is possible to carry the type information in the operation attributes. However, does anyone know if it's still possible to generate LLVM IR with typed pointers? From what I know, the support has been completely removed. We cannot have FPGA compatible IR without this.

superlopuh (approved):
Great timing, the experiments for the paper we just submitted rely on non-opaque pointers also

PapyChacal:
So, after some discussion and feedback, I will allow typed pointers still in xDSL, despite being rejected by the targetted MLIR version, so as to stay compatible with more versions.
If mixing those becomes a problem to target this new version, implementing a pass to erase pointer types is trivial!

superlopuh:
This feels like a change that is not directly CI related, now that I think about it

PapyChacal:
Indeed,  this MLIR version refuses `!llvm.ptr<type>` (It's actually a parsing error!)
Sure, let's implement it quickly. With the rough type conversion pattern I added recently, it should be a handful of lines!

PR #1825, opened 2023-12-02
dialects: Use IntAttr in Vector/Tensor/MemRef types

math-fehr:
We currently use `IntegerAttr` for shapes in vector, tensor, and memref types.
They were also inconsistent, since some of them allowed index types, while other did not.
In MLIR, these dimensions are not typed anyway, so that makes sense.

I also removed the "get" constructors in favor of a single `__init__` one. It is not possible to deprecate
these constructors, as their type changed anyway.

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -409,7 +409,7 @@ class ReshapeOp(IRDLOperation):
                 " {AnyTensorTypeF64}"
             )
         element_type = arg.type.element_type
-        t = TensorTypeF64.from_type_and_list(element_type, shape)
+        t = TensorTypeF64(element_type, shape)
         return super().__init__(result_types=[t], operands=[arg])
 
     @staticmethod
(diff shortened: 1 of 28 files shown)


superlopuh (approved):
beautiful

PapyChacal (approved):
:heart:

AntonLydike (approved):
:pray: Maybe we should mention somewhere that this breaks xDSL API? like, moving on to `__init__` from the old style of constructors...

Otherwise, thanks for doing this!

math-fehr:
> 🙏 Maybe we should mention somewhere that this breaks xDSL API? like, moving on to `__init__` from the old style of constructors...
> 
> Otherwise, thanks for doing this!

Where would be that somewhere? I can write a PSA on zulip, but not sure where I could write that otherwise. `@deprecated` makes no sense here since we are changing the types, meaning they can't be used anyway.

superlopuh:
Maybe in the commit message? Maybe we need some sort of breaking change tag?

PR #1830, opened 2023-12-04
dialects: Add missing verifiers for PDL operations

math-fehr:
This adds some verifiers that were missing in the PDL dialect.


math-fehr:
A bunch of generic syntax tests were now not working anymore, as they are not valid pattern, so they don't verify.
I don't think at that point that it's worth keeping them, since anyway we are roundtripping with generic syntax anyway AFAIK.
We should also probably still at some point decide what tests we want for each dialect.

tests/filecheck/dialects/pdl/verifiers.mlir
```diff
@@ -0,0 +1,86 @@
...
+pdl.pattern @UnusedOperand : benefit(0) {
+    %type = pdl.type : i32
+    %operand = pdl.operand : %type
+    %op = pdl.operation "test.op"
+    pdl.rewrite %op {
+        pdl.erase %op
+    }
+}
+
+// CHECK: expected a bindable user when defined in the matcher body of a `pdl.pattern`
```
superlopuh, line 73: My understanding is that tests with the same string are inherently not testing what you think they are, as the filecheck checks within the whole file, not the section within `// -----` delimiters. So actually three of the tests could "fail" but the filecheck would still pass. It would be good to add some identifiable information to these four tests that would ensure that they actually would signal failure if behaviour changed.

math-fehr: Hmm, aren't they supposed to be printed only once?
Two `CHECK:` should not match the same line?

martin-luecke: My understanding is that the next `CHECK` can not be in the same line. I think in the current state a test could fail and it would not be apparent which one. 
If someone adds more tests to this in the future this might become a graver problem.
As @superlopuh suggested, we should add a `CHECK` line for each test for anchoring. e.g. 
`// CHECK: pdl.pattern @NonConstantAttrInRewrite` for the first. (I am not sure this is actually printed with the verification failure)

martin-luecke (commented):
> A bunch of generic syntax tests were now not working anymore, as they are not valid pattern, so they don't verify.
I don't think at that point that it's worth keeping them, since anyway we are roundtripping with generic syntax anyway AFAIK.

I am not quite sure I understand what you are saying here. Do the verifiers you wrote not work for the generic syntax? Why would that be the case, I expect that after parsing the in memory representation of the op should be similar, no matter if generic syntax or not.

Or were the tests actually written wrongly in the first place?

I think we should be able to roundtrip the generic and the non-generic syntax, no?

math-fehr:
> > A bunch of generic syntax tests were now not working anymore, as they are not valid pattern, so they don't verify.
> > I don't think at that point that it's worth keeping them, since anyway we are roundtripping with generic syntax anyway AFAIK.
> 
> I am not quite sure I understand what you are saying here. Do the verifiers you wrote not work for the generic syntax? Why would that be the case, I expect that after parsing the in memory representation of the op should be similar, no matter if generic syntax or not.
> 
> Or were the tests actually written wrongly in the first place?
> 
> I think we should be able to roundtrip the generic and the non-generic syntax, no?

The tests are not passing the verifier, because the PDL verifier checks that each type/attribute/definition is actually used.
I'll try to add these tests again in a "verifying" form today/tomorrow.

martin-luecke:
> The tests are not passing the verifier, because the PDL verifier checks that each type/attribute/definition is actually used. I'll try to add these tests again in a "verifying" form today/tomorrow.

Does it make sense to simply reuse the IR tests of mlir upstream for this?

tests/filecheck/dialects/pdl/pdl_operand.mlir
```diff
@@ -0,0 +1,47 @@
...
+  // Operands with known types
+  %types = pdl.types : [i32, i64, i1]
+  %operands = pdl.operands : %types
+  
+  %root = pdl.operation(%operands : !pdl.range<!pdl.value>)
+  pdl.rewrite %root with "test_rewriter"(%root : !pdl.operation)
+}
+
+// CHECK: @boundedOperands
+// CHECK: %{{.*}} = pdl.operands : %{{.*}}
```
superlopuh, line 47:
```suggestion
// CHECK: %{{.*}} = pdl.operands : %{{.*}}

```

PapyChacal (approved):
Code looks good, not sure about the semantics!

tests/filecheck/dialects/pdl/pdl_result.mlir
```diff
@@ -0,0 +1,34 @@
...
+pdl.pattern @extractOneResultRange : benefit(1) {
+  %types = pdl.types
+  %root = pdl.operation -> (%types : !pdl.range<!pdl.type>)
+  %result = pdl.results 1 of %root -> !pdl.range<!pdl.value>
+
+  pdl.rewrite %root with "test_rewriter"
+}
+
+// CHECK: @extractOneResultRange
+// CHECK: %{{.*}} = pdl.results 1 of %{{.*}}
```
superlopuh, line 34:
```suggestion
// CHECK: %{{.*}} = pdl.results 1 of %{{.*}}

```

superlopuh (approved):
Would have been good to split this one for sure, it feels like there are some natural splitting points that would have helped review this PR. Changes seem good overall.

martin-luecke (approved):

PR #1834, opened 2023-12-05
interactive: adding function that takes a pipelinepasspsec and returns its strings+ tests 

dshaaban01:
Added a function that takes a pipelinepasspsec and returns its string version. this helps with being able to print out directly what should be used in the terminal (after the xdsl-opt command)

### DIFF
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -103,33 +103,6 @@ class ModulePass(ABC):
         # instantiate the dataclass using kwargs
         return cls(**arg_dict)
 
-    def get_pass_argument_names_and_types(self) -> str:
-        """
-        This method takes a ModulePassT and outputs a string containing the names of the
-        pass arguments and their types. If an argument has a default value, it is not
-        added to the string.
-        """
-        try:
-            return "\n".join(
-                [
-                    f"{field.name}={field.type}"
-                    for field in dataclasses.fields(self)
-                    if not hasattr(self, field.name)
-                ]
-            )
-        except Exception as e:
-            raise e
-
-    def from_pass_to_spec(self) -> PipelinePassSpec:
-        """
-        This function takes a ModulePassT and returns a PipelinePassSpec.
-        """
-        res = PipelinePassSpec(self.name, dict())
-        for f in dataclasses.fields(self):
-            res.args.update(getattr(self, f.name))
-
-        return res
-
 
 def _empty_callback(
     previous_pass: ModulePass, module: builtin.ModuleOp, next_pass: ModulePass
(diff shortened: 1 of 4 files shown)


xdsl/passes.py
```diff
@@ -103,6 +103,24 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
         # instantiate the dataclass using kwargs
         return cls(**arg_dict)
 
+    @classmethod
+    def get_pass_argument_names_and_types(cls: type[ModulePassT]) -> str:
```
superlopuh, line 107: I'm not sure I get the purpose of this string. My understanding of the goal is the following: for a ModulePass instance, we want the text that we could pass in the command line to get back the same result exactly. So the tests for this might be to create a pass programmatically, get its string, parse it using `from_pass_spec`, and check for equality.

superlopuh: or maybe actually getting the spec instead of the string, that way it's structured

xdsl/utils/parse_pipeline.py
```diff
@@ -123,6 +123,27 @@ def normalize_arg_names(self):
...
+        query = f"\n{self.name}"
+        arguments_pipeline = (
+            ", ".join(
+                f"{arg_name}={','.join(map(str.lower, map(str, arg_val)))}"
+                if isinstance(arg_val[0], bool)
+                else f"{arg_name}={','.join(map(str, arg_val))}"
+                for arg_name, arg_val in self.args.items()
+            )
+            if self.args
+            else ""
```
superlopuh, line 140:
```
>>> ", ".join(())
''
```
seems like these two lines are unnecessary, joining an empty collection returns an empty string

xdsl/utils/parse_pipeline.py
```diff
@@ -106,6 +106,22 @@ def peek(self) -> Token:
...
+        else:
+            return str(arg).lower()
+    elif isinstance(arg, str):
+        return f'"{arg}"'
+    else:
+        return str(arg)
+
+
+def _pass_arg_list_type_str(arg: PassArgListType) -> str:
+    return ",".join(_pass_arg_element_type_str(val) for val in arg)
```
superlopuh, line 122: Could you please add the name logic in here? Will make the code below a little simpler
```suggestion
def _named_pass_arg_list_type_str(name: str, arg: PassArgListType) -> str:
    if arg:
        return f'{name}={",".join(_pass_arg_element_type_str(val) for val in arg)}"
    else:
        return name
```

superlopuh (changes requested):
Looking great! I'm being a little pedantic here, because the code is already pretty good, but there are a few small changes that would be good to make, like not having commented code, and checking for truthiness of collections as opposed to their lengths

superlopuh (approved):

tobiasgrosser:
Lovely!! :ship:

PR #1836, opened 2023-12-06
backend: (riscv) support any memref shape for access lowering

superlopuh:
We only handled 1d and 2d now we handle arbitrary ranks of memrefs with known dimensions.

@xdslproject/risc-v-backend


adutilleul (approved):

compor (commented):
LGTM

PR #1837, opened 2023-12-06
backend: (riscv) fix function lowering with mixed floats and ints

superlopuh:
I was under the impression that, in the presence of mixed x and f registers, the ABI was to have them be numbered according to their index in the arg list, so for (int, float, int) -> (float, int) it would be (a0, fa1, a2) -> (fa0, a1). This is not right, the ABI specifies (a0, fa0, a1) -> (fa0, a0) for the same example. This PR fixes this.

@xdslproject/risc-v-backend


adutilleul (approved):

tobiasgrosser:
@superlopuh, this change is approved. Any reason this is not merged?

superlopuh:
Same here, I usually wait for two ticks before merging

superlopuh:
That and time to give other people on the team to catch up with the context

compor (approved):
LGTM, plus it would be good if this could be briefly documented in `move_to_a_regs` at least till (if at all) we use the ABI mechanism.

PR #1839, opened 2023-12-06
backend: (riscv) add mechansim to reserve and unreserve registers

superlopuh:
Turns out there's a bug with our handling of iteration arguments in the register allocator. This is PR 1/2 to fix this issue (I'll open the other ones in order to avoid stacking), it just abstracts reserving and unreserving registers in the register queue class.

@xdslproject/risc-v-backend

### DIFF
--- a/tests/backend/riscv/test_register_queue.py
+++ b/tests/backend/riscv/test_register_queue.py
@@ -38,6 +38,24 @@ def test_push_register():
     assert riscv.Registers.FA0 == register_queue.available_float_registers[-1]
 
 
+def test_reserve_register():
+    register_queue = RegisterQueue()
+
+    register_queue.reserve_register(riscv.IntRegisterType("j0"))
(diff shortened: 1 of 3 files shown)


adutilleul (approved):

tobiasgrosser:
@superlopuh, this change is approved. Any reason this is not merged?

superlopuh:
I generally wait for two ticks before merging, do you think that one is enough always?

tobiasgrosser:
That's for you and the reviewers to judge. Depending on the complexity of the PR one may be good enough. It is actually a good practice for reviewers to comment if they believe input from someone else is needed or if the PR can already go in.

I am not sure  what you mean to give people time to catch up with context?

AntonLydike (approved):
Very nice! Thanks a lot!

compor (approved):

PR #1840, opened 2023-12-06
core: add ability to parse MLIR IR that drops dialect names from operations

superlopuh:
Only parsing, no printing. This unlocks us being able to parse MLIR IR directly output from the `mlir-opt` command, without generic syntax or "local scope". I needed this to unlock our experiments for the linalg to frep project, since we use MLIR as the input, and our version of MLIR had different names for some properties, which messed up the generic syntax. I think that the main xdsl project would benefit from these changes also.

In my understanding, this is more permissive than MLIR itself, that has specific rules for which ops can drop names and which ones cannot. I'm not sure whether we want to be as strict as them on this.

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/scope.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/scope.mlir
@@ -1,0 +1,13 @@
+// RUN: mlir-opt %s | xdsl-opt | filecheck %s
+
+module {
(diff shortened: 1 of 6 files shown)


PapyChacal (approved):
Yay thanks 😁

math-fehr (approved):
Nice!

compor (commented):
LGTM thanks

PR #1841, opened 2023-12-06
interactive: adding a function that takes a module pass and returns a pipelinepassspec + tests

dshaaban01:
This function takes a ModulePassT and returns a PipelinePassSpec. This is useful for interactive gui app, in order to have direct access to arguments and their values (helps for printing out the terminal command etc.).


tests/test_pass_to_spec.py
```diff
@@ -0,0 +1,88 @@
...
+    (
+        (
+            CustomPass(3, [1, 2], None, ["clown", "season"]),
+            PipelinePassSpec(
+                "custom-pass",
+                {
+                    "number": [3],
+                    "int_list": [1, 2],
+                    "str_thing": [],
+                    "list_str": ["clown", "season"],
```
superlopuh, line 64: ❤️

tests/test_pass_to_spec.py
```diff
@@ -0,0 +1,88 @@
...
+        (EmptyPass(), PipelinePassSpec("empty", {})),
+        (
+            SimplePass([3.14, 2.13], 2),
+            PipelinePassSpec("simple", {"a": [3.14, 2.13], "b": [2]}),
+        ),
+    ),
+)
+def test_pass_to_spec_equality(test_pass: ModulePass, test_spec: PipelinePassSpec):
+    assert test_pass.from_pass_to_spec().name == test_spec.name
+    for f in dataclasses.fields(test_pass):
```
superlopuh, line 79: Using dataclasses.fields in this test feels a little off to me, as it feels like an implementation detail of the class. Ideally we'd want this function to only test the behaviour of `from_pass_to_spec()`, and not the internals of passes or specs. What breaks if we just test for equality here with `==`?

superlopuh (changes requested):

tests/test_pass_to_spec.py
dshaaban01: You are totally right. I spazzed. Fixed it.

tests/test_pass_to_spec.py
```diff
@@ -0,0 +1,78 @@
...
+        (
+            CustomPass(3, [1, 2], None, ["clown", "season"]),
+            PipelinePassSpec(
+                "custom-pass",
+                {
+                    "number": [3],
+                    "int_list": [1, 2],
+                    "str_thing": [],
+                    "list_str": ["clown", "season"],
+                    # "literal": ["no"],
```
superlopuh, line 64: commented code

tests/test_pass_to_spec.py
```diff
@@ -0,0 +1,78 @@
...
+        (EmptyPass(), PipelinePassSpec("empty", {})),
+        (
+            SimplePass([3.14, 2.13], 2),
+            PipelinePassSpec("simple", {"a": [3.14, 2.13], "b": [2]}),
+        ),
+    ),
+)
+def test_pass_to_spec_equality(test_pass: ModulePass, test_spec: PipelinePassSpec):
+    assert test_pass.from_pass_to_spec() == test_spec
+    assert str(test_pass.from_pass_to_spec()) == str(test_spec)
```
superlopuh, line 78: what's the benefit from this if the line above passes?

xdsl/passes.py
```diff
@@ -103,6 +103,33 @@ def from_pass_spec(cls: type[ModulePassT], spec: PipelinePassSpec) -> ModulePass
         # instantiate the dataclass using kwargs
         return cls(**arg_dict)
 
+    def from_pass_to_spec(self) -> PipelinePassSpec:
```
superlopuh, line 106:
```suggestion
    def pipeline_pass_spec(self) -> PipelinePassSpec:
```

superlopuh: Or something like this, "from_pass" seems a little off

superlopuh (changes requested):
Looking good, a few minor changes seem like a good idea

superlopuh (approved):
Looking great, thank you

PR #1843, opened 2023-12-06
dialects: (linalg) add custom format to linalg.generic

superlopuh:
- [x] add optional attributes dictionary


PapyChacal (commented):
I think your custom syntax is missing the optional attributes dictionary? Right before the region, according to MLIR's doc.

PapyChacal (approved):
Thanks :rocket:

PR #1846, opened 2023-12-10
transforms: implement gpu-map-parallel-loops

PapyChacal:
Implement `gpu-map-parallel-loops`, and the corresponding attribute `gpu.loop_dim_map`.

Dimension wise, this is doing the reverse of MLIR's `gpu-map-parallel-loops`, as per their own [TODO](https://github.com/llvm/llvm-project/blob/9930f3e2982cd590b75f1252ea5253e53401b605/mlir/lib/Dialect/GPU/Transforms/ParallelLoopMapper.cpp#L70), which maps `scf.parallel`'s faster varying iterator to `gpu.launch` innermost dimension, to have performing codegen for both GPU and CPU cases without specializing the `scf.parallel`. 

Tests are exactly the ones from MLIR, with the said mapped dimensions swapped.

### DIFF
--- a/xdsl/tools/command_line_tool.py
+++ b/xdsl/tools/command_line_tool.py
@@ -60,6 +60,7 @@ from xdsl.transforms import (
     constant_fold_interp,
     convert_scf_to_openmp,
     dead_code_elimination,
+    gpu_map_parallel_loops,
     lower_affine,
     lower_mpi,
     lower_riscv_func,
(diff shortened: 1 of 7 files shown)


xdsl/tools/command_line_tool.py
```diff
@@ -130,6 +131,7 @@ def get_all_dialects() -> list[Dialect]:
 def get_all_passes() -> list[type[ModulePass]]:
     """Return the list of all available passes."""
     return [
+        gpu_map_parallel_loops.GpuMapParallelLoopsPass,
```
superlopuh, line 134: can you put it lower on the list, so it stays alphabetical, please?

superlopuh (approved):

xdsl/tools/command_line_tool.py
PapyChacal: AFAIU, it is not; I'll put at a spot where it is somehow locally in order, sure

superlopuh: thank you. At some point I'll open a PR to sort it again :)

PR #1847, opened 2023-12-10
dialects: Implement affine.min.

PapyChacal:


xdsl/dialects/affine.py
```diff
@@ -222,6 +222,23 @@ def __init__(
...
+
+    map = prop_def(AffineMapAttr)
+
+    def verify_(self) -> None:
+        if len(self.operands) != self.map.data.num_dims + self.map.data.num_symbols:
+            raise VerifyException(
+                f"{self.name} expects {self.map.data.num_dims + self.map.data.num_symbols} operands, but got {len(self.operands)}. The number of map operands must match the sum of the dimensions and symbols of its map."
+            )
+        if len(self.results) != 1:
+            raise VerifyException("affine.min expects a unidimensional map.")
```
superlopuh, line 239: hmm, does `verify` not check this automatically? Is this for parsing error checking? Because it feels like with a custom init this would be impossible to construct.

PapyChacal: I'm discovering Copilot and forgot to clean up after it :)

superlopuh: 👶🤖

georgebisbas (approved):

PR #1848, opened 2023-12-10
transforms: Implement scf-parallel-loop-tiling

PapyChacal:
Implement `scf-parallel-loop-tiling`.

AFAIK, there is no way upstream to tile a `scf.parallel` on a subset of its dimension, e.g.:

```
for x
  for y
     for z
```
to
```
for bx
  for by
    for x
      for y
        for z
```

The upstream `scf-parallel-loop-tiling` also considers a tile size of 0 to be an error.
So here's an implementation of the pass, modulo a tile size of 0 along a dimension here means "don't tile this one".

Stacked on #1847

NB: this does not implement no-min-max-bounds though.


tests/filecheck/transforms/scf-parallel-loop-tiling-partial.mlir
```diff
@@ -0,0 +1,86 @@
...
+// CHECK-LAST-NEXT:     "scf.parallel"(%{{.*}}, %{{.*}}, %zero, %{{.*}}, %{{.*}}, %size, %{{.*}}, %{{.*}}, %one) <{"operandSegmentSizes" = array<i32: 3, 3, 3, 0>}> ({
+// CHECK-LAST-NEXT:     ^{{.*}}(%{{.*}} : index, %{{.*}} : index, %{{.*}} : index):
+// CHECK-LAST-NEXT:       %{{.*}} = arith.addi %{{.*}}, %{{.*}} : index
+// CHECK-LAST-NEXT:       %{{.*}} = arith.addi %{{.*}}, %{{.*}} : index
+// CHECK-LAST-NEXT:       scf.yield
+// CHECK-LAST-NEXT:     }) : (index, index, index, index, index, index, index, index, index) -> ()
+// CHECK-LAST-NEXT:     scf.yield
+// CHECK-LAST-NEXT:   }) : (index, index, index, index, index, index) -> ()
+// CHECK-LAST-NEXT:   func.return
+// CHECK-LAST-NEXT: }
```
superlopuh, line 86:
```suggestion
// CHECK-LAST-NEXT: }

```

xdsl/transforms/scf_parallel_loop_tiling.py
```diff
@@ -0,0 +1,156 @@
...
+        for _ in range(len(tile_sizes_v), len(lower)):
+            tile_sizes_v.append(1)
+
+        # Only handle constant loop bounds for now
+        # if not (
+        #     isa(lower, tuple[arith.Constant, ...])
+        #     and isa(upper, tuple[arith.Constant, ...])
+        #     and isa(step, tuple[arith.Constant, ...])
+        # ):
+        #     return
```
superlopuh, line 42: Am I right in understanding that the comment is about the commented code? It feels ambiguous, hard to tell whether the comment is related to the commented code or just there... It might be worth filing an issue to resolve this, or add a comment with something like "code commented because ..."

xdsl/transforms/scf_parallel_loop_tiling.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+@dataclass
+class ScfParallelLoopTilingPattern(RewritePattern):
+    tile_sizes: list[int]
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: ParallelOp, rewriter: PatternRewriter, /):
+        # Only tile the innermost parallel loop
+        if any(isinstance(o, ParallelOp) for o in op.body.walk()):
```
superlopuh, line 25: do you need to walk the body? would this get different results?
```suggestion
        if any(isinstance(o, ParallelOp) for o in op.body.block.ops):
```

xdsl/transforms/scf_parallel_loop_tiling.py
```diff
@@ -0,0 +1,156 @@
...
+            return
+        lower = op.lowerBound
+        upper = op.upperBound
+        step = op.step
+
+        # fill the tile sizes with ones
+        tile_sizes_v = self.tile_sizes
```
superlopuh, line 32: this is a little scary, why the mutable state?

xdsl/transforms/scf_parallel_loop_tiling.py
```diff
@@ -0,0 +1,156 @@
...
+                assert inner_loop.body.block.first_op is not None
+                inner_loop.body.block.insert_op_before(
+                    iv, inner_loop.body.block.first_op
+                )
+                for use in tuple(arg.uses):
+                    if use.operation is iv:
+                        continue
+                    use.operation.operands[use.index] = iv.result
+        outter_loop.body.block.insert_ops_before([*minops, inner_loop], outter_yield)
+        # print("tiled")
```
superlopuh, line 136: commented code

xdsl/transforms/scf_parallel_loop_tiling.py
PapyChacal: Just forgotten commented code, will clean!

xdsl/transforms/scf_parallel_loop_tiling.py
PapyChacal: Yep, could be a parallel nested in something else, I want to catch those

xdsl/transforms/scf_parallel_loop_tiling.py
PapyChacal: I'm filling it with ones if needed, which is safe for this specific pass logic (Either I had enough values, either I'd add ones anyway)
Now that you mention it, I get the cringe though. Will switch to a copy for safer future changes

georgebisbas:
> NB: this does not implement no-min-max-bounds though.

wdym on this?

PapyChacal:
> > NB: this does not implement no-min-max-bounds though.
> 
> wdym on this?

`no-min-max-bounds` is a flag of MLIR's `scf-parallel-loop-tiling`
It makes it generate conditions inside the tiles instead of computing the minimum before the tile.
It's intended for GPU codegen, where you need perfectly nested loops.
Which is okay for the Devito usecase, where MLIR's `scf-parallel-loop-tiling` is good enough for GPU :slightly_smiling_face:

xdsl/transforms/scf_parallel_loop_tiling.py
```diff
@@ -0,0 +1,147 @@
...
+        rewriter.replace_matched_op(
+            [zero, *tile_sizes.values(), *outter_step, outter_loop]
+        )
+
+
+@dataclass
+class ScfParallelLoopTilingPass(ModulePass):
+    name = "scf-parallel-loop-tiling"
+
+    parallel_loop_tile_sizes: list[int]
```
superlopuh, line 137: would this make sense to be a tuple? do we want it to be mutable? similarly above

superlopuh (approved):

xdsl/transforms/scf_parallel_loop_tiling.py
PapyChacal: That's our Pass parameters implementation. Using tuple errors out as the infrastructure expects lists; I just tried and just switching to a tuple here yields an error; I'll let you open an issue to change this if you want?

superlopuh: ah yes of course my bad

superlopuh: but yes I think that's worth an issue

superlopuh: #1858

superlopuh: How about for the rewrite? Seems like that could be a tuple today

PapyChacal: Sure

PR #1853, opened 2023-12-11
transformations: Implement stencil unrolling

PapyChacal:
Implement `stencil-unroll`, unrolling a stencil by repeating and offsetting computation in its body and marking it for unrolled lowering.


xdsl/transforms/stencil_unroll.py
```diff
@@ -0,0 +1,127 @@
...
+        return_op = op.region.block.last_op
+        # ReturnOp is ApplyOp's terminator
+        assert isinstance(return_op, ReturnOp)
+
+        # Don't unroll already unrolled stencils.
+        if return_op.unroll is not None:
+            return
+
+        # Don't work on degenerate apply with no result
+        if len(op.results) == 0:
```
superlopuh, line 70:
```suggestion
        if not op.results:
```

xdsl/transforms/stencil_unroll.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # If unroll factors list is shorter than the dim, fill with ones from the front
+        unroll = self.unroll_factor.copy()
+        for _ in range(len(unroll), dim):
+            unroll.insert(0, 1)
+
+        # If unroll factors list is longer than the dim, pop from the front to keep
+        # similar semantics
+        for _ in range(dim, len(unroll)):
+            unroll.pop(0)
```
superlopuh, line 86: Popping or inserting from the front feels off to me, this seems equivalent and a little less spidey-sense-triggering:
```suggestion
        unroll = self.unroll_factor
        if len(unroll) < dim:
            # If unroll factors list is shorter than the dim, fill with ones from the front
            unroll = [1] * (dim - len(unroll)) + unroll
        elif len(unroll) > dim:
            # If unroll factors list is longer than the dim, pop from the front to keep
            # similar semantics
            unroll = unroll[-dim:]
```

PapyChacal: Sure

docs/Toy/Toy_Ch0.ipynb
AntonLydike, line 1: Hmm, we seem to have weird formatting changes in these notebooks. Any idea why?

AntonLydike (approved):
:yum:

PR #1854, opened 2023-12-11
interactive: add ability to input parameters on passes that have arguments

dshaaban01:
add ability to input parameters on passes that have arguments using recently merged functions. Commented code left as I am debating with @superlopuh on how to declare the pass_pipeline variable. Will delete when we have discussed.


PapyChacal (commented):
Nice!!
I tried it, and validating invalid parameters makes the app crash at this point though;
To reproduce, just select `convert-scf-to-openmp` and hit enter twice!
Otherwise looks good :slightly_smiling_face:

xdsl/interactive/app.py
```diff
@@ -100,9 +107,10 @@ class InputApp(App[None]):
     Reactive variable used to save the current state of the modified Input TextArea
     (i.e. is the Output TextArea).
     """
-    pass_pipeline = reactive(tuple[type[ModulePass], ...])
-    """Reactive variable that saves the list of selected passes."""
+    pass_pipeline = reactive(tuple[ModulePass, ...])
+    # pass_pipeline = reactive(tuple[dict[type[ModulePass], PipelinePassSpec], ...])
```
superlopuh, line 111: as discussed over zulip, it feels like this is the way to go, since ModulePass is not frozen, and we can't guarantee that they won't be in a weird state after being run once

dshaaban01: Done.

dshaaban01:
> Nice!! I tried it, and validating invalid parameters makes the app crash at this point though; To reproduce, just select `convert-scf-to-openmp` and hit enter twice! Otherwise looks good 🙂

Yay!!! Thanks so much for pointing out the error - will try it now and do my best to fix it.

PapyChacal:
> > Nice!! I tried it, and validating invalid parameters makes the app crash at this point though; To reproduce, just select `convert-scf-to-openmp` and hit enter twice! Otherwise looks good 🙂
> 
> Yay!!! Thanks so much for pointing out the error - will try it now and do my best to fix it.

Can't reproduce now, it seems the meanwhile commit did make that better!

xdsl/interactive/app.py
```diff
@@ -213,17 +220,66 @@ def watch_available_pass_list(
...
         """
         selected_pass = event.item.name
-        for name, value in ALL_PASSES:
-            if name == selected_pass:
-                self.pass_pipeline = tuple((*self.pass_pipeline, value))
-                return
+        for name_pass, value_pass in ALL_PASSES:
+            if name_pass == selected_pass:
+                # check if pass has concatenated_arg_val
+                if len(fields(value_pass)) != 0:
```
superlopuh, line 273:
```suggestion
                if fields(value_pass):
```

xdsl/interactive/app.py
```diff
@@ -213,17 +220,66 @@ def watch_available_pass_list(
...
         """
         When a new selection is made, the reactive variable storing the list of selected
         passes is updated.
         """
         selected_pass = event.item.name
-        for name, value in ALL_PASSES:
-            if name == selected_pass:
-                self.pass_pipeline = tuple((*self.pass_pipeline, value))
-                return
+        for name_pass, value_pass in ALL_PASSES:
```
superlopuh, line 270:
```suggestion
        for pass_name, pass_value in ALL_PASSES:
```

"Pass name" reads a little better to "name pass" to me, could you pease change them?

xdsl/interactive/app.py
```diff
@@ -213,17 +220,66 @@ def watch_available_pass_list(
...
+                # check if pass has concatenated_arg_val
+                if len(fields(value_pass)) != 0:
+                    # call function that handles adding passes with concatenated_arg_val to the pass_pipeline
+                    self.get_pass_arguments(value_pass)
+                else:
+                    # add the selected pass to pass_pipeline
+                    self.pass_pipeline = (
+                        *self.pass_pipeline,
+                        (value_pass, value_pass().pipeline_pass_spec()),
+                    )
```
superlopuh, line 281: I would recommend folding this logic into `get_pass_arguments`, it feels like that function should be deciding whether to show a screen or not

xdsl/interactive/app.py
```diff
@@ -213,17 +220,66 @@ def watch_available_pass_list(
...
+                )[0]
+                self.pass_pipeline = (
+                    *self.pass_pipeline,
+                    (selected_pass_value, new_pass_with_arguments),
+                )
+
+            except PassPipelineParseError as e:
+                res = f"PassPipelineParseError: {e}"
+                screen = AddArguments(TextArea(res, id="argument_text_area"))
+                self.push_screen(screen, add_pass_with_arguments_to_pass_pipeline)
```
superlopuh, line 250: could you please make this a method?

superlopuh: and then pass it as a lambda to `self.push_screen`

dshaaban01: I'm not sure that this is possible. This is a specific thing in textual where when a screen is dismissed, it somehow automatically returns the string we want in the form of the variable (`concatenated_arg_val : str`) to the`add_pass_with_arguments_to_pass_pipeline` function that is declared in the push_screen method `self.push_screen(screen, add_pass_with_arguments_to_pass_pipeline)`. It's all connected and cannot be made as a method I think.

See 'Returning Data from Screens' Section here: 
https://textual.textualize.io/guide/screens/#returning-data-from-screens

georgebisbas:
Hi there, I am not involved in this project, but would like to ask since I get all these emails from commits,
is it really needed to have all these short commits happening towards merging a PR?

georgebisbas:
Maybe some commit squashing would be useful even though they are all squashed when merging.
This resource could also be useful: https://github.com/xdslproject/xdsl/wiki/Tags-for-commit-messages-and-PR-titles

tobiasgrosser:
> Hi there, I am not involved in this project, but would like to ask since I get all these emails from commits, is it really needed to have all these short commits happening towards merging a PR?

Is this something that can be resolved with email filters? I am a little concerned to increase the cost of upstreaming PRs.

georgebisbas:
> > Hi there, I am not involved in this project, but would like to ask since I get all these emails from commits, is it really needed to have all these short commits happening towards merging a PR?
> 
> Is this something that can be resolved with email filters? I am a little concerned to increase the cost of upstreaming PRs.

Sure, can add some filter

dshaaban01:
> > > Hi there, I am not involved in this project, but would like to ask since I get all these emails from commits, is it really needed to have all these short commits happening towards merging a PR?
> 
> > 
> 
> > Is this something that can be resolved with email filters? I am a little concerned to increase the cost of upstreaming PRs.
> 
> 
> 
> Sure, can add some filter

I will do my best to minimize commits levels as much as possible. Thank you for looking into the email filters, I appreciate it!

PR #1856, opened 2023-12-12
dialects: (memref) add alloc(a) dynamic dimension support

jorendumoulin:
This PR adds support for dynamic dimensions to the `memref.alloc` and `memref.alloca` operations.

The operands of the `alloc` operation (`%d` of type `index`) indicate the size of the dynamic dimension in the result:
```mlir
%0 = memref.alloc(%d) : memref<8x?xf32, 1>
```

There is also a verify method to check if the amount of operands is equal to the amount of dynamic dimensions in the memref.

Additionally, I also aligned the implementation of `alloc` and `alloca` to be identical.

@JosseVanDelm

### DIFF
--- a/xdsl/dialects/memref.py
+++ b/xdsl/dialects/memref.py
@@ -342,10 +342,12 @@ class Alloc(IRDLOperation):
         )
 
     def verify_(self) -> None:
-        if not isinstance(self.results[0].type, MemRefType):
+        memref_type = self.memref.type
+        if not isinstance(memref_type, MemRefType):
             raise VerifyException("expected result to be a memref")
+        memref_type = cast(MemRefType[Attribute], memref_type)
 
-        dyn_dims = [x for x in self.results[0].type.shape.data if x.value.data == -1]
+        dyn_dims = [x for x in memref_type.shape.data if x.value.data == -1]
         if len(dyn_dims) != len(self.dynamic_sizes):
             raise VerifyException(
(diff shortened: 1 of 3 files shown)


xdsl/dialects/memref.py
```diff
@@ -314,27 +314,43 @@ class Alloc(IRDLOperation):
...
+            properties={
+                "alignment": alignment,
             },
         )
 
+    def verify_(self) -> None:
+        if not isinstance(self.results[0].type, MemRefType):
+            raise VerifyException("expected result to be a memref")
+
+        dyn_dims = [x for x in self.results[0].type.shape.data if x.value.data == -1]
```
superlopuh, line 348:
```suggestion
        memref_type = self.memref.type
        if not isinstance(memref_type, MemRefType):
            raise VerifyException("expected result to be a memref")
        memref_type = cast(MemRefType[Any], memref_type)

        dyn_dims = [x for x in memref_type.shape.data if x.value.data == -1]
```

superlopuh: Similarly below, should help get rid of the pyright errors

JosseVanDelm (changes requested):
Very nice! 
Please solve the pyright issues as per @superlopuh's suggestion.
Can you also add an extra filecheck testcase for the case you are describing?

superlopuh (approved):

JosseVanDelm (approved):
hell yeah

PR #1857, opened 2023-12-12
interactive: remove flashing error from mlir-opt

superlopuh:


.vscode/launch.json
```diff
@@ -29,7 +29,7 @@
             "type": "python",
             "request": "launch",
             "module": "xdsl.interactive.app",
-            "justMyCode": false,
+            "justMyCode": true,
```
superlopuh, line 32: this change helped debug this error, and seems like a safer default

dshaaban01 (approved):
Looks great! Consider changing selections to a list and then back to a tuple to avoid costly operations

math-fehr (approved):

PR #1859, opened 2023-12-12
interactive: change normalize_arg_names() in parse_pipeleline.py to return a tuple instead of dict

dshaaban01:
The `from_pass_spec()` function in passes.py is used in my interactive app. Unfortunately, the way it is implemented is that it mutates the PassPipelineSpec and empties out its arguments dictionary. This causes some drama as I still need to reuse the spec in my app, and do not want the arguments to be altered/mutated.

After discussions with @superlopuh , we originally had the idea of altering the below code: 

```
@dataclass(eq=True, frozen=True)
class PipelinePassSpec:
    """
    A pass name and its arguments.
    """

    name: str
    args: dict[str, PassArgListType]
```

and instead replacing `args` with:

`args: tuple(tuple(str,PassArgListType))`

This will come in a future PR or issue, tbd.

And so, for now, I have changed the normalize_arg_names() to return a new `PassPipelineSpec` altogether in order to mitigate the mutation situation.

### DIFF
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -64,10 +64,9 @@ class ModulePass(ABC):
             )
 
         # normalize spec arg names:
-        # spec.normalize_arg_names()
-        spec_arguments_dict: dict[str, PassArgListType] = dict()
-        for k, v in list(spec.normalize_arg_names()):
-            spec_arguments_dict[k] = v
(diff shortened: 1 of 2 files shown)


superlopuh (approved):
LGTM

PR #1861, opened 2023-12-13
interactive: add function that counts number of operations in a module

dshaaban01:
Add function that takes a ModuleOp and returns a dictionary containing the number of occurences of each Operation in the ModuleOp. First step to provide users with feedback on whether the passes/ optimisations they have chosen are "good" or "bad".


dshaaban01:
@webmiche What do you think? We briefly discussed smth like this in our last meeting. Thoughts?

xdsl/interactive/pass_metrics.py
```diff
@@ -0,0 +1,25 @@
...
+
+    # Apply .name to each element in the list using a list comprehension
+    names_list = [op.name for op in all_operations]
+
+    # Count occurrences of each unique Operation
+    operation_counter = Counter(names_list)
+
+    res_operations_count: dict[str, int] = dict(operation_counter.items())
+
+    return res_operations_count
```
superlopuh, line 25:
```suggestion
    return Counter(op.name for op in module.walk())
```

superlopuh: Might need to change the return type to `Mapping[str, int]`

superlopuh: But probably not since `Counter` subclasses `dict`

dshaaban01: Wow. You really made 13 lines of code redundant. That's optimization if I ever saw it.

superlopuh: The power of Python :)

superlopuh (approved):
Looks great!

webmiche (commented):
Nice, LGTM.

PR #1864, opened 2023-12-13
interactive: add ability to input parameters on passes that have arguments

dshaaban01:
add ability to input parameters on passes that have arguments using recently merged functions.


tests/interactive/test_app.py
```diff
@@ -152,12 +153,16 @@ async def test_buttons():
         )
 
         # Select two passes
-        app.pass_pipeline = tuple(
+        app.pass_pipeline += (
+            *app.pass_pipeline,
```
superlopuh, line 157: this feels off, it adds it to itself?

dshaaban01: Will fix now.

xdsl/interactive/app.py
```diff
@@ -213,17 +221,66 @@ def watch_available_pass_list(
...
                 )
 
+    def get_pass_arguments(self, selected_pass_value: type[ModulePass]) -> None:
+        """
+        This function facilitates user input of pass concatenated_arg_val by navigating
+        to the AddArguments screen, and subsequently parses the returned string upon
+        screen dismissal and appends the pass to the pass_pipeline variable.
+        """
+
+        def add_pass_with_arguments_to_pass_pipeline(concatenated_arg_val: str) -> None:
```
superlopuh, line 231: can you please make this a method, and pass in a lambda for the callback below?

dshaaban01: I'm not sure that this is possible. This is a specific thing in textual where when a screen is dismissed, it somehow automatically returns the string we want in the form of the variable (concatenated_arg_val : str) to theadd_pass_with_arguments_to_pass_pipeline function that is declared in the push_screen method self.push_screen(screen, add_pass_with_arguments_to_pass_pipeline). It's all connected and cannot be made as a method I think.

See 'Returning Data from Screens' Section here:
https://textual.textualize.io/guide/screens/#returning-data-from-screens

superlopuh: Sure that makes sense, it just feels like a part of the logic might be better in a method. Re-reading it some more, I'm not sure exactly how I would do this myself, so maybe it's better to merge as-is and see how it evolves later.

superlopuh (approved):

xdsl/interactive/app.py
dshaaban01: Sounds great - we can discuss further in future.

PR #1865, opened 2023-12-13
interactive: add number of operations to visual and toggle between 2 modes

dshaaban01:
Number of operations count DataTable added so users can see a difference. Added Button to toggle between normal view (without the DataTable) and this view.

Normal View:
image

Show Operation Count View:
image


superlopuh:
Could you please add a button to hide and show this? Feels like it's not always useful to have this information

PapyChacal:
Amazing! Feature request here: could we get a display of the difference? I.e. If I now have 3 `arith.addi`  and used to have 8, display a neat `-5` somewhere?
That's a use I had recently, wondering how many memory accesses a pipeline removed on an example.

dshaaban01:
> Could you please add a button to hide and show this? Feels like it's not always useful to have this information

Was going to do this in my next PR. Should I do this on this one or will it be too big?

dshaaban01:
> Amazing! Feature request here: could we get a display of the difference? I.e. If I now have 3 `arith.addi` and used to have 8, display a neat `-5` somewhere? That's a use I had recently, wondering how many memory accesses a pipeline removed on an example.

That was going to be my text step :)) Actually I was thinking of changing the background of the RHS DataTable (i.e. based on the Output IR) to be green or red or neutral (to symbolise +, - or staying the same) but your idea is honestly better! Will do this :)

webmiche (approved):
LGTM, I would address the feature request in a follow up :wink:

xdsl/interactive/app.py
```diff
@@ -314,8 +341,10 @@ def update_current_module(self) -> None:
             )
             pipeline.apply(ctx, module)
             self.current_module = module
+            self.update_number_of_input_ops(input_text)
```
superlopuh, line 344: it's a shame to reparse and compute everything when we already have the input module a few lines above, could you please change the code so that it doesn't redo this computation?

xdsl/interactive/app.py
```diff
@@ -347,6 +377,32 @@ def get_query_string(self) -> str:
...
+        else:
+            ctx = MLContext(True)
+            for dialect in get_all_dialects():
+                ctx.load_dialect(dialect)
+            module = Parser(ctx, input_text).parse_module()
+            input_ops = count_number_of_operations(module)
+
+            self.input_number_of_ops.clear()
+            for k, v in input_ops.items():
+                self.input_number_of_ops.add_row(k, v)
```
superlopuh, line 392: To me it feels like the op count data is state inherent to the app, and would benefit from being in a reactive property. Could you please add properties (probably tuples of key-value pairs) and watch methods to update the UI when the data changes?

superlopuh (changes requested):

xdsl/interactive/app.py
dshaaban01: I think that the input_number_of_ops variable does not need to be a reactive property since the input text does not change frequently and when it does - it can just call the update_number_of_input ops function. 
I considered making the output_number_of_ops variable reactive (and adding watch/compute methods) but came to the conclusion that it was unnecessary.

Saw your comment in the other PR - is it okay if it stays just like that PR (except where I swap out dict for tuples) and not making those two variables reactive? what do u think? I think reactivity here would be overkill.

superlopuh: I think it's worth doing, mostly because state that is explicit is better, that way we can easily introspect, debug, and extend it later, as opposed to being made of these temporary side-effecty things. Especially if you've already added it in the other PR, might as well bring it over here.

xdsl/interactive/app.py
```diff
@@ -347,6 +389,49 @@ def get_query_string(self) -> str:
...
+            for dialect in get_all_dialects():
+                ctx.load_dialect(dialect)
+            module = Parser(ctx, input_text).parse_module()
+            self.input_operation_count_tuple = tuple(
+                count_number_of_operations(module).items()
+            )
+
+            self.input_operation_count_datatable.clear()
+            for k, v in self.input_operation_count_tuple:
+                self.input_operation_count_datatable.add_row(k, v)
```
superlopuh, line 410: feels like this should be a "watch" method, similarly below

xdsl/interactive/app.py
```diff
@@ -347,6 +389,49 @@ def get_query_string(self) -> str:
         )
         return f"xdsl-opt -p {query}"
 
+    def update_input_operation_count_tuple(self, input_text: str) -> None:
```
superlopuh, line 392: can you change this to take a module instead, please?

xdsl/interactive/app.py
superlopuh: Ah only saw later that you already made the change, thank you

xdsl/interactive/app.py
dshaaban01: but wait - do we want these variables to be reactive or not? I still don't understand. Right now they aren't reactive, but I suppose you want them to be? I explained in my comment above that I think making them reactive in this instance is a bit overkill, but okay I can do it.

xdsl/interactive/app.py
dshaaban01: ok

superlopuh (approved):
Looks great!

PR #1870, opened 2023-12-14
interactive: add diff to operation count table

dshaaban01:
image

Added the diff as per @PapyChacal 's request :)


PapyChacal (approved):
Thanks 😁

xdsl/interactive/app.py
```diff
@@ -135,16 +135,27 @@ class InputApp(App[None]):
     passes_list_view: ListView
     """ListView displaying the passes available to apply."""
 
-    input_number_of_ops: DataTable[str | int]
-    output_number_of_ops: DataTable[str | int]
+    input_operation_count_dict: dict[str, int] = dict()
+    """Saves the operation name and count of the input text area in a dictionary."""
+    output_operation_count_dict: dict[str, int] = dict()
+    """Saves the operation name and count of the output text area in a dictionary."""
```
superlopuh, line 141: ah, this is exactly what I was hoping for in my comment in the other PR, could you please move this change there? One bug here is that dict is mutable, so changes that mutate it directly won't be picked up by the reactive system, could you please make it an n-tuple of pairs, to make sure that if it's mutated the system propagates the changes?

dshaaban01: Yes will do.

xdsl/interactive/app.py
```diff
@@ -432,6 +432,20 @@ def update_output_operation_count_tuple(self) -> None:
...
+            else:
+                self.output_operation_count_datatable.add_row(k, v, "-")
+
+            # for k_input, v_input in self.input_operation_count_tuple:
+            #     if k == k_input:
+            #         diff = v - v_input
+            #         self.output_operation_count_datatable.add_row(k, v, diff)
+            #     else:
+            #         self.output_operation_count_datatable.add_row(k, v, "-")
+
```
dshaaban01, line 448: Have provided 2 different implementations - to loop through tuple to check if k is in the tuple (double for loop seemed a bit costly though...) so I thought might be cheaper to convert to dictionary and then do dictionary lookup? lmk which is preferred.

superlopuh: I would recommend a third option, adding the diff to the data source tuple, since we have both the output and input counts at that stage, so the view calculation is a dumb display of data

dshaaban01: I don't quite understand what you mean. What is the data source tuple? We have the input count tuple and the output count tuple. Which one are you referring to?

superlopuh: I mean the output count tuple

xdsl/interactive/app.py
```diff
@@ -389,24 +403,60 @@ def get_query_string(self) -> str:
...
+        # iterate through input_operation_count_tuple
+        for k, v in input_op_count_dict.items():
+            if k not in output_op_count_dict:
+                diff = -input_op_count_dict[k]
+                res = (
+                    *res,
+                    (k, 0, diff),
+                )
+
+        return res
```
superlopuh, line 438: I'd recommend something like this, to create a set of all keys, and iterate over that once, and accumulate into a dict also.

```suggestion
        input_op_count_dict = dict(self.input_operation_count_tuple)
        output_op_count_dict = dict(self.output_operation_count_tuple)
        all_keys = {*input_op_count_dict, *output_op_count_dict}
        
        res: dict[str, tuple[int, int]] = {}
        for k in all_keys:
            input_count = input_op_count_dict.get(k, 0)
            output_count = output_op_count_dict.get(k, 0)
            res[k] = (output_count, output_count - input_count)

        return tuple((k, v0, v1) for (k, (v0, v1)) in res.items().sorted())
```

superlopuh: I would also recommend putting this logic into a function near the one that computes the dicts and to test the logic separately from the gui tool

xdsl/interactive/app.py
```diff
@@ -421,18 +471,33 @@ def update_output_operation_count_tuple(self) -> None:
             case Exception():
                 self.output_operation_count_tuple = ()
             case ModuleOp():
-                self.output_operation_count_tuple = tuple(
-                    count_number_of_operations(self.current_module).items()
+                # sort tuples by count in ascending order
```
superlopuh, line 474: I'd lean towards sorting them alphabetically always so that the user learns where to look for which op over time. It also means that the lists are grouped by dialect, which seems useful at first glance. I don't feel very strongly about this

dshaaban01: Okay.

tests/interactive/test_pass_metrics.py
```diff
@@ -54,12 +57,64 @@ def test_operation_counter_with_parsing_text():
...
+  }
+}
+"""
+    parser = Parser(ctx.clone(), output_text)
+    output_module = parser.parse_module()
+
+    expected_diff_res: tuple[tuple[str, int, str], ...] = ()
+    expected_diff_res = (
+        *expected_diff_res,
+        ("arith.constant", 0, "-1"),
```
superlopuh, line 107: Why the split? Does this not work? How about with no type annotation?
```suggestion
    expected_diff_res: tuple[tuple[str, int, str], ...] = (
        ("arith.constant", 0, "-1"),
```

xdsl/interactive/app.py
```diff
@@ -421,18 +439,32 @@ def update_output_operation_count_tuple(self) -> None:
...
+                    (k, v)
+                    for (k, v) in sorted(
+                        count_number_of_operations(self.current_module).items()
+                    )
                 )
 
-    def watch_output_operation_count_tuple(self) -> None:
+        self.update_operation_count_diff_tuple()
+
+    def update_operation_count_diff_tuple(self) -> None:
```
superlopuh, line 452: Can we rely on the reactive infrastructure and have this be called automatically?
```suggestion
    def compute_operation_count_diff_tuple(self) -> None:
```

superlopuh: There's a risk of it being called twice, once when the input updates, and again when the output updates, it would be good to figure out whether textual is smart enough not to do that

xdsl/interactive/app.py
```diff
@@ -136,14 +139,28 @@ class InputApp(App[None]):
...
+    Saves the operation name and count of the input text area in a reactive tuple of
+    tuples.
+    """
     output_operation_count_tuple = reactive(tuple[tuple[str, int], ...])
-    """Saves the operation name and count of the output text area in a dictionary."""
+    """
+    Saves the operation name and count of the output text area in a reactive tuple of
+    tuples.
+    """
+    diff_operation_count_tuple = reactive(tuple[tuple[str, int, str], ...])
```
superlopuh, line 151: Why have both the diff and the output? Could we compute the diff directly?

xdsl/interactive/app.py
dshaaban01: I don't want it to be called unnecessarily though... compute functions fire whenever ANY reactive variable is fired.. should I still do it

superlopuh: Good to know, thank you

xdsl/interactive/app.py
```diff
@@ -394,45 +406,55 @@ def update_input_operation_count_tuple(self, input_module: ModuleOp) -> None:
         Function that updates the input_operation_datatable to display the operation
         names and counts in the input text area.
         """
+        # sort tuples alphabetically by operation name
         self.input_operation_count_tuple = tuple(
-            count_number_of_operations(input_module).items()
+            (k, v)
+            for (k, v) in sorted(count_number_of_operations(input_module).items())
```
superlopuh, line 412:
```suggestion
            sorted(count_number_of_operations(input_module).items())
```

xdsl/interactive/app.py
```diff
@@ -394,45 +406,55 @@ def update_input_operation_count_tuple(self, input_module: ModuleOp) -> None:
...
+                # sort tuples alphabetically by operation name
+                output_operation_count_tuple = tuple(
+                    (k, v)
+                    for (k, v) in sorted(
+                        count_number_of_operations(self.current_module).items()
+                    )
+                )
+                self.diff_operation_count_tuple = get_diff_operation_count(
+                    self.input_operation_count_tuple, output_operation_count_tuple
                 )
```
superlopuh, line 449: I'd recommend setting the output_operation_count_tuple inside the switch, to an empty tuple if necessary, and using that value after the switch.
```suggestion
                # sort tuples alphabetically by operation name
                output_operation_count_tuple = tuple(
                    (k, v)
                    for (k, v) in sorted(
                        count_number_of_operations(self.current_module).items()
                    )
                )
            self.diff_operation_count_tuple = get_diff_operation_count(
                self.input_operation_count_tuple, output_operation_count_tuple
            )
```

superlopuh (approved):
looking great, ty, a couple of small suggestions remaining

PR #1872, opened 2023-12-15
core: correct `__str__`/`__repr__` usage in ParseError

PapyChacal:
Hey, I'm trying to have fun with MLIROptPass as part of our Python DSL integration.

Sadly, when chaining a ParserError (such as when MLIROptPass fails because xDSL can't parse back MLIR's output), it looks like this:
```
>       raise ParseError(at_position, msg)
E       xdsl.utils.exceptions.ParseError: (Span[1151:1154](text='ptr'), 'type expected')
```
Which isn't very helpful.
This PR offers to force it to display nicely in this case too, yielding instead:
```
    raise ParseError(at_position, msg)
xdsl.utils.exceptions.ParseError: <unknown>:23:52
    %19 = "llvm.mlir.undef"() : () -> !llvm.struct<(ptr, ptr, i64, array<3 x i64>, array<3 x i64>)>
                                                    ^^^
                                                    type expected
```
Which is more actionnable.

I'm not sure overriding `__str__` is the only nor the best solution though. WDYT?

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -43,13 +43,13 @@ async def test_inputs():
         await pilot.pause()
         assert (
             app.output_text_area.text
-            == "(Span[5:6](text=''), 'Operation builtin.unregistered does not have a custom format.')"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
         )
 
         assert isinstance(app.current_module, ParseError)
         assert (
             str(app.current_module)
-            == "(Span[5:6](text=''), 'Operation builtin.unregistered does not have a custom format.')"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
         )
 
         # Test corect input
--- a/xdsl/utils/exceptions.py
+++ b/xdsl/utils/exceptions.py
@@ -68,6 +68,9 @@ class ParseError(Exception):
     span: Span
     msg: str
 
+    def __str__(self) -> str:
+        return self.__repr__()
+
     def __repr__(self) -> str:
         return self.span.print_with_context(self.msg)
 
--- a/xdsl/utils/exceptions.py
+++ b/xdsl/utils/exceptions.py
@@ -69,9 +69,6 @@ class ParseError(Exception):
     msg: str
 
     def __str__(self) -> str:
-        return self.__repr__()
-
-    def __repr__(self) -> str:
         return self.span.print_with_context(self.msg)
 
     def with_context(self) -> str:


webmiche (commented):
Hmm, I think I would prefer to handle this special case in the MLIROpt pass itself rather than changing core classes like the ParseError. I would assume that the MLIROpt pass, as a pass that runs something on the cmdline and reparses and stuff, is more of a unicorn than a normal pass.

Also, I am surprised that no other tests actually fail, but I guess we did not test parsing errors very thoroughly :thinking:

webmiche:
Also, connecting `__repr__` and `__str__` in this way feels not very "pythonic" given that the two usually have different focus (AFAIU).

PapyChacal:
> Hmm, I think I would prefer to handle this special case in the MLIROpt pass itself rather than changing core classes like the ParseError. I would assume that the MLIROpt pass, as a pass that runs something on the cmdline and reparses and stuff, is more of a unicorn than a normal pass.
> 
> Also, I am surprised that no other tests actually fail, but I guess we did not test parsing errors very thoroughly 🤔

Makes sense, I'm okay with that as it's enough for my usecase!
Though, the failing GUI test made me think it might help there too, hence proposing the core change! But I'll move this to MLIROptPass first :slightly_smiling_face:

PapyChacal:
Mh; I'm actually not sure how to do this without going far with the exception handling.
It seems that using `__str__` is just the default behaviour of chained exceptions.
Also, doesn't this user-friendly make more sense as `__str__` than `__repr__`, in the Pythonic mindset?

AntonLydike:
I guess overwriting `__str__` is the easy way out, but eventually we would want a global exception handler in xdsl-opt-main or something so that we can catch and pretty print these errors (and verification errors, etc.)

For now I'm very happy with this change, I think it's overall positive.

PapyChacal:
So I just pushed something else, basically just replacing the `__repr__` by the `__str__`.
This is the intended usage of the API AFAIU? `__str__` is the nice, user-friendly representation, and `__repr__` is the strict, dataclass-generated, evaluatable representation.
WDYT @webmiche?

PapyChacal:
> I guess overwriting `__str__` is the easy way out, but eventually we would want a global exception handler in xdsl-opt-main or something so that we can catch and pretty print these errors (and verification errors, etc.)
> 
> For now I'm very happy with this change, I think it's overall positive.



> I guess overwriting `__str__` is the easy way out, but eventually we would want a global exception handler in xdsl-opt-main or something so that we can catch and pretty print these errors (and verification errors, etc.)
> 
> For now I'm very happy with this change, I think it's overall positive.

I think that's just the role of `__str__` though? Why would we need to reimplement an exception handler for pretty printing?

math-fehr:
Yeah, looks to me that this should have always been __str__ instead of __repr__!

webmiche (approved):
Yes, I agree. Thinking about it more, this should have always been `__str__` :+1:

dshaaban01 (commented):
Look's great! Thank you for adding this!

PR #1873, opened 2023-12-15
dialects: Refactor and improve llvm type parsing.

PapyChacal:
Before it all makes sense in MLIR someday, here's a small improvement on `llvm` types parsing from xDSL!
We had some special cases scattered over the place, I just refactored it in top-level functions.
Then, I implement the `array<>` special case too.


AntonLydike (approved):
:smiley: Very nice to see this get some love!

PR #1875, opened 2023-12-16
backend: (riscv) reserve iter arg from allocation inside of riscv_scf body

superlopuh:
The iteration arguments live in registers that currently are not protected from being clobbered by the register allocator. This PR fixes this by reserving them form being reused inside the body, by using the newly introduced reserve/unreserve mechanism.


AntonLydike (approved):
Very nice stuff! :+1:

PR #1882, opened 2023-12-17
dialects: LLVM updates.

PapyChacal:
Some updates to the LLVM dialect, everything is retro-compatible:
- `llvm.call` can return something.
- `llvm.func` has a sym_visibility optional property.


xdsl/dialects/llvm.py
```diff
@@ -1202,9 +1207,10 @@ class CallOp(IRDLOperation):
     args: VarOperand = var_operand_def()
 
     callee: SymbolRefAttr = prop_def(SymbolRefAttr)
-    callee_type: LLVMFunctionType = prop_def(LLVMFunctionType)
+    callee_type: LLVMFunctionType | None = opt_prop_def(LLVMFunctionType)
```
webmiche, line 1210: Why is the function type itself optional? That sounds very weird :thinking: 
Or is it just a shorthand because given the args and the result types the function type is trivial?

superlopuh (approved):

xdsl/dialects/llvm.py
PapyChacal: You got it :slightly_smiling_face: Turns out it's mandatory only in the variadic case; https://mlir.llvm.org/docs/Dialects/LLVM/#llvmcall-llvmcallop :
```quote
If the callee is a variadic function, then the callee_type attribute must carry the function type.
```
And the constructor I adapted already build this callee type in both cases either way! This was only problematic to parse some output from MLIR's `convert-func-to-llvm`, where it wasn't adding this callee_type.

webmiche: Ah, nice. Thanks :+1:

webmiche (approved):

PR #1888, opened 2023-12-19
core: Expose InsertPoint class (NFC)

math-fehr:
This PR creates the InsertPoint class, and use it in the builder.
This has no functional changes otherwise

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -30,29 +30,29 @@ class InsertPoint:
     """
 
     @overload
-    def __init__(self, point: Operation) -> None:
+    def __init__(self, op_before_or_block: Operation) -> None:
         ...
 
     @overload
-    def __init__(self, point: Block) -> None:
+    def __init__(self, op_before_or_block: Block) -> None:
         ...
 
     @overload
-    def __init__(self, point: Block, point2: Operation | None) -> None:
+    def __init__(self, op_before_or_block: Block, op_before: Operation | None) -> None:
         ...
 
     def __init__(
-        self, point: Block | Operation, point2: Operation | None = None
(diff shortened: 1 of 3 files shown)


xdsl/builder.py
```diff
@@ -12,47 +12,92 @@
...
+    def __init__(self, point: Block) -> None:
+        ...
+
+    @overload
+    def __init__(self, point: Block, point2: Operation | None) -> None:
+        ...
+
+    def __init__(
+        self, point: Block | Operation, point2: Operation | None = None
+    ) -> None:
```
webmiche, line 46: Could this be `{inner,outer}_point` or something? `point{1,2}` seems like suboptimal naming :wink:

webmiche (approved):

xdsl/builder.py
superlopuh: I have a feeling that it can't be due to Python's naming things, but I would suggest `op_before_or_block, op_before`

PapyChacal (approved):
Thanks! I love this but I'm not sure why, given it's a NFC. OOC, can I ask about your concrete motivation for implementing it, if any?

math-fehr:
> Thanks! I love this but I'm not sure why, given it's a NFC. OOC, can I ask about your concrete motivation for implementing it, if any?

I'm currently thinking of changing the rewriter, for mostly two reasons:
* One is that Dalia would need one that works with a worklist, and currently we only support rewrites that operate in order in the IR.
* The other is that I'm tired of writing `rewriter.insert_op_before_matched_op`. I just want to use the builder, so I first need to make `RewritePattern` inherit from `Builder`. But while looking at it I felt that `InsertPoint` should be its own class. I might never use it anywhere else, but I felt it's better that way.

PR #1889, opened 2023-12-19
CI: Fix user name for api key

math-fehr:
User should be `__token__` instead of `api` for api keys.


webmiche (approved):
If you say so, oh almighty Mathieu :pray:

tobiasgrosser:
> If you say so, oh almighty Mathieu 🙏

Did turn xdsl into a religion? 😂

math-fehr:
> > If you say so, oh almighty Mathieu 🙏
> 
> Did turn xdsl into a religion? 😂

Always has been :unicorn:

PR #1890, opened 2023-12-19
core: Add constructors to `Builder` and `InsertPoint`

math-fehr:
The constructors for `Builder` and `InsertPoint` were too ambiguous, so they are removed in favor of named constructors `at_begin`, `at_end`, `before` and `after`.
Some naming changed as well in `InsertPoint`, and documentation improved.

This should hopefully make the code more readable

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -23,36 +23,36 @@ class InsertPoint:
     block: Block
     """The block where the insertion point is in."""
 
-    op_before: Operation | None = field(default=None)
+    insert_before: Operation | None = field(default=None)
     """
-    The operation right before the insertion point.
-    None if the insertion point is at the end of the block.
+    The insertion point is right before this operation.
+    If the operation is None, the insertion point is at the end of the block.
     """
 
-    @overload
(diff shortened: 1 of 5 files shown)


xdsl/builder.py
```diff
@@ -23,45 +23,45 @@ class InsertPoint:
...
         """
         Check that the insertion point is valid.
         An insertion point can only be invalid if `op_before` is an `Operation`, and its
         parent is not `block`.
         """
-        if self.op_before is not None:
-            if self.op_before.parent is not self.block:
+        if self.insert_before is not None:
+            if self.insert_before.parent is not self.block:
                 raise ValueError("Insertion point must be in the builder's `block`")
```
superlopuh, line 65: My intuition would be to make this class frozen, and move the verification to the init. The builder can make a new insertion point instead of mutating the one it has

superlopuh (approved):

xdsl/builder.py
math-fehr: Fair, makes sense!

PR #1892, opened 2023-12-20
core: Add callbacks to the `Builder`

math-fehr:
These callbacks are called whenever an operation or block is created and inserted.
They are used in the pattern matching engine to keep the worklist updated.

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -76,6 +76,12 @@ class Builder:
     insertion_point: InsertPoint
     """Operations will be inserted at this location."""
 
+    notify_operation_inserted: Callable[[Operation], None] | None = field(default=None)
+    """A callback that is called when an operation is inserted by the builder."""
+
+    notify_block_created: Callable[[Block], None] | None = field(default=None)
+    """A callback that is called when a block is created by the builder."""
+
     def __init__(self, insert_point: InsertPoint) -> None:
         self.insertion_point = insert_point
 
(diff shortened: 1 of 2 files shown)


xdsl/builder.py
```diff
@@ -78,6 +78,12 @@ class Builder:
     _insertion_point: InsertPoint
     """Operations will be inserted at this location."""
 
+    notify_operation_inserted: Callable[[Operation], None] | None = field(default=None)
```
superlopuh, line 81: Is there a reason you chose this name? I would have expected something like `on_operation_inserted`, or `operation_inserted_handler`. Are these the names from the cpp builder?

superlopuh (approved):
Code looks good small nit on naming

xdsl/builder.py
PapyChacal: +1 for handler on my side
Also, why not handlers plural, with a list of callable ?

PapyChacal (approved):

xdsl/builder.py
math-fehr: Yeah, the name come from the C++. I can change it to handler though, it's not that important to keep the same names here!
We *could* do a list, but I'm not sure if it's worth compare to nesting `Callable` together. In general, I'm never sure if callbacks should be single `Callable` or a list of `Callable`. In that particular case, I don't think we will ever remove callbacks, and there is none by default, so I would say that a single one makes sense? If you want, I could add a method `add_handler` or something, that will take care of adding one and nesting it nicely!

math-fehr:
I fixed the naming, and replaced `callback | None` with `list[callback]`.
If that's good for you, I'll merge it!

webmiche (approved):
LGTM :+1:

superlopuh (approved):

PR #1901, opened 2023-12-20
core: fix string literal escaping

PapyChacal:
Fix string litteral escaping in parsing and printing (as in, make it MLIR-compatible)
This enables parsing and printing CUDA binaries as serialized by MLIR, for example.


AntonLydike:
`ast.literal_eval(self.text)` is used to evaluate literals such as `1`, `14.333e42` and `"\u0041\u0042\u0043"`. Especially the last one is by far the hardest to get right.

Anyhow, it seems that you get an unquoted string literal? That's pretty weird indeed. Do you have a testcase for this?

AntonLydike:
> Anyhow, it seems that you get an unquoted string literal? That's pretty weird indeed. Do you have a testcase for this?

Ah, I see the problem in the PR. Yikes.

PapyChacal:
I did not measure this myself, but FWIW, the CI run I watched didn't feel any slower.

@math-fehr (who made the fix, thanks! 🙏) measured it with a benchmark from the paper and told me it went from 700  to 900 ms. Definitely an acceptable price for correctness as far as I'm concerned!
There are also easy leads for optimization, but we'd be happy to merge the fix first and optimize later!

xdsl/printer.py
```diff
@@ -349,7 +348,27 @@ def print_paramattr_parameters(
...
+        for char in string:
+            o = ord(char)
+            if char == "\\":
+                self.print("\\\\")
+            elif 0x20 <= o <= 0x7E and char != '"':
+                self.print(char)
+            else:
+                self.print("\\")
+                self.print(hexdigits[o >> 4])
+                self.print(hexdigits[o & 0x0F])
```
AntonLydike, line 370: Could we `f"{o:x}"` instead? This code here fails for larger unicode codepoints

xdsl/printer.py
```diff
@@ -349,7 +348,27 @@ def print_paramattr_parameters(
...
+            (0x20 <= ord(char) <= 0x7E) and char not in ('"', "\\") for char in string
+        ):
+            self.print(f'"{string}"')
+            return
+        hexdigits = "0123456789ABCDEF"
+        self.print('"')
+        for char in string:
+            o = ord(char)
+            if char == "\\":
+                self.print("\\\\")
```
AntonLydike, line 364: I think it makes it easier to look at non-escaped strings to use `r"\\"` here (the `r` removes all escape sequences, it's usually used for writing regexes).

But opinions may vary here, and pyright, ruff, or whatever else we may use might disagree.

xdsl/utils/lexer.py
```diff
@@ -173,8 +172,42 @@ def from_span(cls, span: Span | None) -> StringLiteral | None:
 
     @property
     def string_contents(self):
-        # TODO: is this a hack-job?
-        return ast.literal_eval(self.text)
+        # Common case, so we optimize for it
+        if "\\" not in self.text:
+            return self.text[1:-1]
+        res = list[str]()
```
AntonLydike, line 178: Is there a reason why `res` is not a string that we append to? Is it so much slower in python?

xdsl/utils/lexer.py
```diff
@@ -173,8 +172,42 @@ def from_span(cls, span: Span | None) -> StringLiteral | None:
...
+                case '"':
+                    res.append('"')
+                case _ if c1 in hexdigits:
+                    if i >= len(text):
+                        raise ValueError(self, "invalid string literal")
+                    c2 = text[i]
+                    i += 1
+                    if c2 not in hexdigits:
+                        raise ValueError(self, "invalid string literal")
+                    res.append(chr(int(c1 + c2, 16)))
```
AntonLydike, line 207: Similarly to above, this does not handle larger unicode codepoints.

I think there might be a better approach than iterating over the whole input one-by-one (makes longer hex escapes easier):

 - use `split(r"\")` to get parts
 - look at the start of each part in `parts[1:]` to identify escape sequences (can be e.g. regex)
 - convert escape sequences and leave the other string parts in tact for printing.

xdsl/utils/lexer.py
```diff
@@ -595,14 +628,26 @@ def _lex_string_literal(self, start_pos: Position) -> Token:
...
+                # Allowed escape characters
+                if escaped_char in ['"', "\\", "n", "t"]:
+                    continue
+                # Hexadecimal escape
+                if escaped_char in hexdigits:
+                    second_hexdigit = self._get_chars()
+                    if second_hexdigit is None:
+                        break
+                    if second_hexdigit in hexdigits:
+                        continue
```
AntonLydike, line 646: Do we need to have this check here?

AntonLydike (commented):
This looks great already! Thanks @PapyChacal and @math-fehr for this! I just have some nitpicks, especially for larger unicode codepoints.

xdsl/utils/lexer.py
superlopuh: That was my first instinct, I was thinking maybe a mutablebytearray or something

xdsl/utils/lexer.py
superlopuh: the most expensive case is definitely the hex codes, and I have a suspicion that it's not very common if we only hit them now, so it may be worth adding a separate regex to escape `\n` et al, and do another `'\' in res` check to bail early if we don't need to invoke the state machine

xdsl/printer.py
PapyChacal: Definitely agree, whatever removes some escape hell layers, thanks for the idea!

xdsl/utils/lexer.py
PapyChacal: Yeah I think that was just some first performance concern. Either way, plan is to have cryptic optimized code here in the end, because it's perf-critical! Just keeping whatever works for a first fix

xdsl/printer.py
PapyChacal: I disagree on not handling larger codepoints, MLIR does in a way
{o:x} prints the value in hexadecimal, but not the (MLIR) way we need to

AntonLydike:
Maybe stupid idea, but what if we instead change the `StringAttr` API to handle both binary "string" data and decoded string data? Then the parser wouldn't need to encode/decode, and we could instead handle this in the attribute. The `attr` could then cache the decoding, so worst case perf should be about the same.

math-fehr:
That is 100% my current plan yes!

PR #1903, opened 2023-12-21
core: Add a worklist for pattern rewriting

math-fehr:
This PR adds a `Worklist` class that will be used for pattern rewriting.
The objective is to use that worklist to check which operations should be considered next by the rewriter.

For now, the class in not used, just tested.


georgebisbas (approved):

xdsl/pattern_rewriter.py
```diff
@@ -609,6 +609,51 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter) -> None:
         return
 
 
+@dataclass(eq=False)
+class Worklist:
+    _lifo: list[Operation | None] = field(default_factory=list, init=False)
+    """
+    The list of operations to iterate over, used as a lifo stack.
```
webmiche, line 616: Not sure what you guys' think, but in terms of terminology, I would always call a stack LIFO and a queue FIFO. So, why not just call this the `op_stack` or whatever?

webmiche (approved):
I am not sure what kind of design you are thinking of, but in particular when it comes to reordering rewrite applications, wouldn't something that looks more like an actual list instead of like a stack be easier to use?

math-fehr:
> I am not sure what kind of design you are thinking of, but in particular when it comes to reordering rewrite applications, wouldn't something that looks more like an actual list instead of like a stack be easier to use?

What do you mean by that?
The reason I'm using a stack is because that's exactly how MLIR works, and that's how you want it to be in the rewrite applicator. So when you replace an operation, you want to directly rewrite the next operations that are on the stack.

webmiche:
I think I am a bit confused on how exactly this will be used. Is this a stack of operations that you still need to progress (i.e. check whether any rewrite matches and then apply it)? And once something is applied, you just add the new operations to the stack?

I was thinking of reordering and doing things interactvely, so something that is probably not directly related to this, right?

math-fehr:
Yeah, this is how this is supposed to be use. You have a stack of operations, that you first fill by walking the IR.
Then you try rewriting on each operation one by one, and whenever you have a rewrite, you add the added and modified operations in the worklist again.

For the interactive part, this worklist won't be used, and instead we will use another data structure, so that won't be used for that!

webmiche:
Okay, then this LGTM :+1:

xdsl/pattern_rewriter.py
```diff
@@ -609,6 +609,51 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter) -> None:
...
+    Operations are added and removed at the end of the list.
+    Operation that are `None` are meant to be discarded, and are used to
+    keep removal of operations O(1).
+    """
+
+    _map: dict[Operation, int] = field(default_factory=dict, init=False)
+    """
+    The map of operations to their index in the lifo.
```
webmiche, line 624: You missed one :wink:

tests/pattern_rewriter/test_worklist.py
```diff
@@ -0,0 +1,65 @@
+from xdsl.dialects import test
+from xdsl.pattern_rewriter import Worklist
+
+
+def test_worklist_posh_pop():
```
superlopuh, line 5:
```suggestion
def test_worklist_push_pop():
```
👸

xdsl/pattern_rewriter.py
```diff
@@ -609,6 +609,52 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter) -> None:
...
+    _map: dict[Operation, int] = field(default_factory=dict, init=False)
+    """
+    The map of operations to their index in the stack.
+    It is used to check if an operation is already in the stack, and to
+    remove it in O(1).
+    """
+
+    def is_empty(self) -> bool:
+        """Check if the worklist is empty."""
+        return bool(self._op_stack)
```
superlopuh, line 631:
```suggestion
        return not bool(self._op_stack)
```

superlopuh: Would be good to add tests for this

xdsl/pattern_rewriter.py
```diff
@@ -609,6 +609,52 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter) -> None:
...
+            self._op_stack.append(op)
+
+    def pop(self) -> Operation | None:
+        """Pop the operation at the end of the worklist."""
+        while self._op_stack:
+            op = self._op_stack.pop()
+            if op is not None:
+                del self._map[op]
+                return op
+        return None
```
superlopuh, line 648: It feels weird for the stack to be empty twice, effectively, first explicitly, and then by checking that the op is not `None`. Maybe something like this?
```suggestion
        if not self._op_stack:
            return None
        op = self._op_stack[-1]
        del self._op_stack[-1]
        del self._map[op]
        return op
```

superlopuh (changes requested):

xdsl/pattern_rewriter.py
math-fehr: I'm not sure I understand your comment here.
The reason why there is this while loop is because any `None` value in our `_op_stack` is a value that was deleted.
Instead of removing it from the stack, we rename it to `None`, so we can essentially remove them in `O(1)` instead of `O(n)`.
So this removes all values that are `None` at the end.

xdsl/pattern_rewriter.py
math-fehr: Done!

xdsl/pattern_rewriter.py
superlopuh: Ah, could you please make that a little clearer in the docstrings? I completely missed the whole `None` mechanism. It feels like a good idea to first find the index of the last non-`None` value in the list, remove all the values after that index, and then pop, instead of looping.

math-fehr: It is in the docstring, but in the `_op_stack` one ;)
I'll add a comment on top of this part to make it clearer, and indeed should first look at it then delete it!

math-fehr: Actually, I'm not sure that I should first look for the last non-`None`, then remove all at once.

Since most of the time, there are no `None`. This only happens when you remove an operation that is not the matched one, so that's pretty rare.

First looking at the last non-`None` value will be slower, as we currently do only one `pop` and one `if isinstance(.., None)` in the "normal" case, while checking for `None` might do multiple checks. 

Also, I would argue that the current version is clearer and less error-prone?

superlopuh: I see what you mean. Sounds good :)

superlopuh: Would still be good to have more doc strings and comments

superlopuh (approved):

xdsl/pattern_rewriter.py
math-fehr: Just to be sure, what other doc strings are you referring to?
I'm adding comments there, but I'm not sure what doc strings should be added?

xdsl/pattern_rewriter.py
```diff
@@ -609,6 +609,58 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter) -> None:
...
+        """
+        Push an operation to the end of the worklist, if it is not already in it.
+        """
+        if op not in self._map:
+            self._map[op] = len(self._op_stack)
+            self._op_stack.append(op)
+
+    def pop(self) -> Operation | None:
+        """Pop the operation at the end of the worklist."""
+        # All `None` operations at the end of the stack are distarded,
```
superlopuh, line 645:
```suggestion
        # All `None` operations at the end of the stack are discarded,
```

xdsl/pattern_rewriter.py
superlopuh: The comment you added explains it well, thank you for adding it

PR #1904, opened 2023-12-21
core: Add options to `walk` to control traversal order

math-fehr:
This PR allows to change the traversal order of `walk` through the `region_first` option and `reverse` option.

* `reverse` will reverse the order of traversal of blocks, regions, and operation lists.
* `region_first` will make the traversal first traverse the regions, then the current operation. 

This also deprecates `walk_reverse`, which should be replaced by `walk(region_first=True, reverse=True)`.

This is stack on top of an unrelated PR because GitHub doesn't allow me to stack a PR on top of two unrelated PRs (which I'll do on the next PR).

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -25,6 +25,7 @@ from typing_extensions import Self
 
 from xdsl.traits import IsTerminator, NoTerminator, OpTrait, OpTraitInvT
 from xdsl.utils import lexer
+from xdsl.utils.deprecation import deprecated
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.str_enum import StrEnum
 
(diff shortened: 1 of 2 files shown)


xdsl/ir/core.py
```diff
@@ -888,14 +889,23 @@ def drop_all_references(self) -> None:
...
+        If region_first is set, then the operation regions are iterated before the
+        operation. If reverse is set, then the region, block, and operation lists are
+        iterated in reverse order.
         """
-        yield self
-        for region in self.regions:
-            yield from region.walk()
+        if not region_first:
+            yield self
+        for region in self.regions[::-1] if reverse else self.regions:
```
superlopuh, line 903:
```suggestion
        for region in reversed(self.regions) if reverse else self.regions:
```
Would this work? `[::-1]` creates a new list

math-fehr: Oh yes, thanks!

superlopuh (approved):

PR #1908, opened 2023-12-22
core: use worklist in pattern matching algorithm

math-fehr:
Use a worklist for pattern rewriting.

This means that every time an operation is modified, inserted, removed, the worklist gets updated, and only the operations in the worklist are rewritten.
This will allow us to remove restriction in our pattern matches, such that only operations right before, after, or nested in the matched operation can be rewritten.
These restrictions will be removed later on.

### DIFF
--- a/docs/Toy/toy/rewrites/setup_riscv_pass.py
+++ b/docs/Toy/toy/rewrites/setup_riscv_pass.py
@@ -34,4 +34,4 @@ class SetupRiscvPass(ModulePass):
     name = "setup-lowering-to-riscv"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
-        PatternRewriteWalker(AddSections()).rewrite_module(op)
+        PatternRewriteWalker(AddSections(), apply_recursively=False).rewrite_module(op)
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -39,8 +39,7 @@ class PatternRewriterListener(BuilderListener):
     """Callbacks that are called when an operation is removed."""
 
     operation_modification_handler: list[Callable[[Operation], None]] = field(
-        default_factory=list,
-        kw_only=True,
+        default_factory=list, kw_only=True
     )
     """Callbacks that are called when an operation is modified."""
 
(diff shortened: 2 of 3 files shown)


math-fehr:
In terms of slowdown, it seems that this approach makes my benchmark go from 0.85s to 1.01s, which is acceptable I would say, for the win we would get by being able to modify whatever we want (and not only operations next to the matched operation).

xdsl/pattern_rewriter.py
```diff
@@ -786,68 +785,128 @@ class PatternRewriteWalker:
...
+
+    def rewrite_op(self, op: Operation) -> None:
+        """
+        Rewrite operations nested in the given operation by repeatedly applying the
+        pattern.
+        """
+        pattern_listener = self._get_rewriter_listener()
 
-    def _rewrite_op(self, op: Operation) -> Operation | None:
+        while True:
```
superlopuh, line 866: meganit:
```suggestion
        self._populate_worklist(op)
        should_continue = self._process_worklist(pattern_listener)
        if not self.apply_recursively:
            return
        while should_continue:
```
The `while True` here seems a little off to me, and checking for a constant value in a loop

superlopuh: It's an arbitrary thing, feel free to ignore

xdsl/pattern_rewriter.py
```diff
@@ -786,68 +785,128 @@ class PatternRewriteWalker:
...
-        new regions.
-        """
-        for region in op.regions:
-            blocks = reversed(region.blocks) if self.walk_reverse else region.blocks
-            for block in blocks:
-                iter_op = block.last_op if self.walk_reverse else block.first_op
-                while iter_op is not None:
-                    iter_op = self._rewrite_op(iter_op)
+
+        while True:
```
superlopuh, line 900: Here the `while True` feels a little off also, since it's clearly one loop per operation, maybe it would be worth refactoring to something like this?
```suggestion
        while (op := self._worklist.pop()) is not None:
```

xdsl/pattern_rewriter.py
math-fehr: Yeah I agree, is this better?
Btw, I was really frustrated to learn that there was no `do/while` in Python :(

xdsl/pattern_rewriter.py
math-fehr: Can't do really much more here sadly.
The reason why we cannot have this check at the beginning, is because we need to start with an `op` already, because the `PatternRewriter` should start with an `op` (I don't want to allow ill-formed `PatternRewriter`).
Main issue here to be honest, is that there are no do/while loop.
I changed it to make the structure more obvious, thoughts?

xdsl/pattern_rewriter.py
superlopuh: Yeah do while is useful sometimes

superlopuh (approved):
Looks great, thank you

math-fehr:
I'll wait after the holidays to merge that, as I first want to check that everything out-of-tree that people are using is still working with this patch!

xdsl/pattern_rewriter.py
```diff
@@ -793,68 +792,129 @@ class PatternRewriteWalker:
...
+
+        while op_was_modified:
+            self._populate_worklist(op)
+            op_was_modified = self._process_worklist(pattern_listener)
+
+    def _populate_worklist(self, op: Operation) -> None:
+        """Populate the worklist with all nested operations."""
+        # We walk in reverse order since we use a stack for our worklist.
+        for sub_op in op.walk(
+            reverse=not self.walk_reverse, region_first=not self.walk_regions_first
```
webmiche, line 887: That looks like we are not quite consistent with naming. How can it be that we need a not here?

xdsl/pattern_rewriter.py
```diff
@@ -39,8 +39,7 @@ class PatternRewriterListener(BuilderListener):
     """Callbacks that are called when an operation is removed."""
 
     operation_modification_handler: list[Callable[[Operation], None]] = field(
-        default_factory=list,
-        kw_only=True,
+        default_factory=list, kw_only=True
     )
```
webmiche, line 43: This change looks weird :thinking: 
Is this black acting up?

xdsl/pattern_rewriter.py
webmiche: I guess we could just hoist the first iteration of the loop and the repeat the loop body, but that does not sound a lot cleaner.

The other issue that makes this impossible is that we need to create the rewriter in the first place. So you could probably also implement something that really only happens in the first iteration. like:

```
rewriter = None
while (op := self._worklist.pop()) is not None:
  if not rewriter:
    rewriter = PatternRewriter(op)
    rewriter.extend_from_listener(listener)
```

But that is even worse, IMO. I don't have a very strong opinion on this. I think this and hoisting the first iteration are both fine solutions.

webmiche (approved):
LGTM, only minor comments.

docs/Toy/toy/rewrites/setup_riscv_pass.py
```diff
@@ -34,4 +34,4 @@ class SetupRiscvPass(ModulePass):
     name = "setup-lowering-to-riscv"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
-        PatternRewriteWalker(AddSections()).rewrite_module(op)
+        PatternRewriteWalker(AddSections(), apply_recursively=False).rewrite_module(op)
```
AntonLydike, line 37: What's the change that requires us to specify non-recursive application here?

xdsl/pattern_rewriter.py
```diff
@@ -793,68 +792,129 @@ class PatternRewriteWalker:
...
+        listener passed as configuration to the walker.
+        """
+        return PatternRewriterListener(
+            operation_insertion_handler=(
+                self.listener.operation_insertion_handler
+                + [self._handle_operation_insertion]
+            ),
+            operation_removal_handler=(
+                self.listener.operation_removal_handler
+                + [self._handle_operation_removal]
```
AntonLydike, line 846: dito for this and the next few

AntonLydike (approved):
I like the overall direction this is going, just have some minor things that we may want to address.

xdsl/pattern_rewriter.py
math-fehr: Nope, because we need an `op` to create the `rewriter`.
To be fair, I feel that the current version is fine enough, with the current comments.

xdsl/pattern_rewriter.py
math-fehr: The comment is there for that ;)
You need to walk in reverse order because you have a stack (LIFO). The reason we have a stack is to mirror MLIR, which I believe they made the right choice to be honest.
Naming wise, it does make sense, since `walk` can be omitted in the `walk` method IMO.

xdsl/pattern_rewriter.py
math-fehr: It was me, because it wasn't consistent with the field right on top of it.
To be fair, I'd rather add this in this PR than adding a completely separate PR just for this typo, as you prefer.

docs/Toy/toy/rewrites/setup_riscv_pass.py
math-fehr: Here, `AddSections()` add an operation to `ModuleOp`.
The old rewriter would not apply this again, as you wouldn't look again at `ModuleOp`, because we didn't marked the operation as changed.
Now since the rewriter look at the fixpoint, this will loop forever as it will add operations continuously.

xdsl/pattern_rewriter.py
```diff
@@ -793,68 +792,129 @@ class PatternRewriteWalker:
...
     """The listener that will be called when an operation or block is modified."""
 
-    def rewrite_module(self, op: ModuleOp):
-        """Rewrite an entire module operation."""
-        self._rewrite_op(op)
+    _worklist: Worklist = field(default_factory=Worklist, init=False)
+    """The worklist of operations to walk over."""
 
-    def _rewrite_op(self, op: Operation) -> Operation | None:
+    def _add_operands_to_worklist(self, operands: Iterable[SSAValue]) -> None:
```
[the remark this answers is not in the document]
math-fehr: Yeah, I just took the MLIR name and didn't have better ideas.
Do you have a better name in mind? I couldn't find anything I found satisfactory

PR #1909, opened 2023-12-22
fix attribute docstring

jorendumoulin:
This PR fixes a very small mistake in the docstring of the attribute parser.
The code parses a `#` symbol (which is correct), while the docstring mentions `!`


math-fehr (approved):
Oh yes, thanks a lot!

PR #1910, opened 2023-12-22
core: Simplify the pattern rewriter

math-fehr:
Now that the pattern rewriter use the worklist, it is not necessary to keep track of the inserted operations before/after
the matched operation.
We can thus remove all these checks.


superlopuh (approved):
beautiful

webmiche:
Could you rebase this? I think that would make it more reviewable :)

webmiche (approved):
Now that actually looks like a simplification :wink: 

I still wonder if no downstream users are affected (in particular in the Chocopy world), but if we can make sure it's fine, I think this is a good change :+1:

PR #1911, opened 2023-12-23
core: Register passes lazily

math-fehr:
This means we don't load a pass file unless we are using it.
This will save us a lot of startup time when we do the same for operations.

This already put us from 500ms (!!) to 400ms

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,6 @@
+from xdsl.tools.command_line_tool import get_all_passes
+
+
+def test_get_all_passes_names():
+    for name, pass_factory in get_all_passes():
+        assert name == pass_factory().name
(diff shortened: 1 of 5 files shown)


superlopuh:
Looks good! Have you solved the trace breaking?

math-fehr:
> Looks good! Have you solved the trace breaking?

What trace breaking? Do you have an example?

superlopuh:
If you remember the last time I tried this, I ran `xdslopt` with and without this diff, and for some reason tracing broke when the passes were made lazy. I'll try to reproduce this now

superlopuh:
Seems to work fine now, I got to traces, and can confirm that it's 100 ms faster

superlopuh (approved):

superlopuh:
Ah also I think I got confused with the lazy dialect loading

math-fehr:
Hmm, I'm wondering if what you saw was actually a "compile-time" error triggered when loading a dialect file?
But nice!

math-fehr:
Couldn't reproduce it as well, so I guess that's fine!

PR #1912, opened 2023-12-23
transforms: Canonicalize without the MLContext

math-fehr:
Do not rely on the loaded operations in the canonicalization pass.

This change is necessary to let dialects be loaded only during parsing. An alternative solution would be to load all dialects before doing the rewrites, or to load dialects automatically during imports.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,5 +17,5 @@ textual-dev==1.3.0
 pytest-asyncio==0.23.2
 # pyright version has to be fixed with `==`. The CI parses this file
 # and installs the according version for typechecking.
-pyright==1.1.342
+pyright==1.1.343
 -e .[extras]
--- a/tests/filecheck/dialects/pdl/verifiers.mlir
+++ b/tests/filecheck/dialects/pdl/verifiers.mlir
@@ -10,6 +10,7 @@ pdl.pattern @NonConstantAttrInRewrite : benefit(0) {
 }
 
 // CHECK: expected constant value when specified within a `pdl.rewrite`
+// CHECK: NonConstantAttrInRewrite
 
 // -----
 
@@ -22,6 +23,7 @@ pdl.pattern @UnnamedOpInRewrite : benefit(0) {
 }
 
 // CHECK: must have an operation name when nested within a `pdl.rewrite`
+// CHECK: UnnamedOpInRewrite
 
 // -----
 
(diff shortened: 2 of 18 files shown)


superlopuh:
I'm not sure I get it, what's the benefit of this? Or rather, what breaks if this change isn't made?

math-fehr:
The main issue is coming from my "lazy dialects" change.

The overall goal is to not load files that do not need to be loaded, since we use a non-trivial amount of time loading Python files. My current solution is to never load dialect files, and instead load Python files dynamically during parsing, or during a pass call. However, that means in particular that we cannot get all dialects without loading all of them anymore, which is becoming costly.

In short, the `MLContext` will contain load the dialects that have been parsed (I'm going to remove most of `MLContext` uses btw). This means that we can't rely on it anymore for that pass (and it is the only pass requiering this). It will definitely have a cost for that pass, but there is probably room of improvement (like walking the pass first, to build the rewritepattern).

superlopuh:
I'm still not sure that I see the issue. If the ops are in a module with a context, then it should be safe to assume that their dialect is loaded in the context, also, no?

math-fehr:
In MLIR, this is done by loading dialects before each pass, and noting in each pass which dialects are dependent.
Since we don't do that, we don't have at all that assumption. Also, I'm not sure we actually want to have that, since besides this pass, we have absolutely no need for that context.

The dialect Python file will be loaded though, but the dialect won't necessary be loaded in the context.

superlopuh:
Ah. I would really like for us to have that mechanism in all our passes, though. I think there's quite a lot of benefit of having the context represent the set of legal operations at a point in the compilation pipeline. I can see a few ways it could be used for optimisation and for raising errors if something weird happens. In any case, I can see how that is scope creep wrt this PR.

superlopuh:
Could we not register the dialects of the ops in the current IR? This would be a smaller change wrt the current mechanism, and should not affect the python import stuff since the dialects must already be in the interpreter if the ops have been constructed

math-fehr:
> Could we not register the dialects of the ops in the current IR? This would be a smaller change wrt the current mechanism, and should not affect the python import stuff since the dialects must already be in the interpreter if the ops have been constructed

But that would be much slower right?
The cost of constructing the RewritePatterns will anyway be larger, because here we don't really do anything.
However, in the previous version, the fastest non-match is `O(pattern_size)`, since we do an `isinstance` for each rewrite pattern we have (as we don't optimize them at all). Here, it would be `O(op_trait_size)`, which scales better?

superlopuh (approved):
Good point, you've convinced me

PR #1913, opened 2023-12-24
dialects (pdl): Add better verifiers to `pdl.pattern`

math-fehr:
Add missing patterns to `pdl.pattern`, and add tests for it.

I also realized that I merged the last PDL PR without addressing one review about the `split-input-file` tests, really sorry for that. This PR is fixing it at the same time (it's just 5-6 line changes)


superlopuh (approved):

martin-luecke (approved):
LGTM

PR #1914, opened 2023-12-26
dialects (onnx): Initial onnx dialect support

### LINKED ISSUE #1894: add onnx.Add op

brnorris03:
onnx dialect initial support, as discussed on zulip. 

* Add elementwise binary operations, `onnx.Add`, `onnx.Sub`, `onnx.Mul`, `onnx.Div` (see [doc](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Add) for add semantics; the others are equivalent). 
* Add verification of broadcastable arguments
* Add lit tests

Resolves #1894

### DIFF
--- a/tests/filecheck/dialects/onnx/onnx_ops.mlir
+++ b/tests/filecheck/dialects/onnx/onnx_ops.mlir
@@ -1,10 +1,18 @@
 // RUN: XDSL_ROUNDTRIP
 
-func.func @add(%arg0: tensor<1x2x6xf32>, %arg1: tensor<1x2x6xf32>) ->  tensor<1x2x6xf32> {
-  %0 = "onnx.Add"(%arg0, %arg1) {onnx_node_name = "/model.1/Add"} : (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
-  return %0 : tensor<1x2x6xf32>
-}
+%t0, %t1 = "test.op"(): () -> (tensor<1x2x6xf32>, tensor<1x2x6xf32>)
+%t2, %t3 = "test.op"(): () -> (tensor<3x2xf32>, tensor<1x2xf32>)
+%t4, %t5 = "test.op"(): () -> (tensor<3x1x2xf32>, tensor<3x4x1xf32>)
+%t6, %t7 = "test.op"(): () -> (tensor<1x5x1x3xf32>, tensor<2x1x6x3xf32>)
 
-// CHECK-LABEL: func.func @add
-// CHECK-SAME:      (%arg0 : tensor<1x2x6xf32>, %arg1 : tensor<1x2x6xf32>) -> tensor<1x2x6xf32>       
-// CHECK:   %0 = onnx.Add(%arg0, %arg1) {"onnx_node_name" = "/model.1/Add"} : (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
+%res_add = "onnx.Add"(%t0, %t1) {onnx_node_name = "/Add"} : (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
+// CHECK: %res_add = onnx.Add(%t0, %t1) {"onnx_node_name" = "/Add"}: (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
+
+%res_sub = "onnx.Sub"(%t2, %t3) {onnx_node_name = "/Sub"}: (tensor<3x2xf32>, tensor<1x2xf32>) -> tensor<3x2xf32>
+// CHECK: %res_sub = onnx.Sub(%t2, %t3) {"onnx_node_name" = "/Sub"}: (tensor<3x2xf32>, tensor<1x2xf32>) -> tensor<3x2xf32>
+
+%res_mul = "onnx.Mul"(%t4, %t5) {onnx_node_name = "/Mul"}: (tensor<3x1x2xf32>, tensor<3x4x1xf32>) -> tensor<3x4x2xf32>
+// CHECK: %res_mul = onnx.Mul(%t4, %t5) {"onnx_node_name" = "/Mul"}: (tensor<3x1x2xf32>, tensor<3x4x1xf32>) -> tensor<3x4x2xf32>
+
+%res_div = "onnx.Div"(%t6, %t7) {onnx_node_name = "/Div"}: (tensor<1x5x1x3xf32>, tensor<2x1x6x3xf32>) -> tensor<2x5x6x3xf32>
+// CHECK: %res_div = onnx.Div(%t6, %t7) {"onnx_node_name" = "/Div"}: (tensor<1x5x1x3xf32>, tensor<2x1x6x3xf32>) -> tensor<2x5x6x3xf32>
--- a/tests/filecheck/dialects/onnx/onnx_ops.mlir
+++ b/tests/filecheck/dialects/onnx/onnx_ops.mlir
@@ -1,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+func.func @add(%arg0: tensor<1x2x6xf32>, %arg1: tensor<1x2x6xf32>) ->  tensor<1x2x6xf32> {
+  %0 = "onnx.Add"(%arg0, %arg1) {onnx_node_name = "/model.1/Add"} : (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
+  return %0 : tensor<1x2x6xf32>
+}
+
+// CHECK-LABEL: func.func @add
+// CHECK-SAME:      (%arg0 : tensor<1x2x6xf32>, %arg1 : tensor<1x2x6xf32>) -> tensor<1x2x6xf32>       
+// CHECK:   %0 = onnx.Add(%arg0, %arg1) {"onnx_node_name" = "/model.1/Add"} : (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from abc import ABC
-from typing import TypeAlias, TypeVar
 
 from xdsl.dialects.builtin import AnyTensorType, SSAValue
 from xdsl.ir import (
@@ -10,25 +9,21 @@ from xdsl.ir import (
     OpResult,
 )
 from xdsl.irdl import (
-    Attribute,
     IRDLOperation,
     irdl_op_definition,
     operand_def,
     result_def,
 )
-from xdsl.parser import Parser
-from xdsl.printer import Printer
-
-ArgT = TypeVar("ArgT", bound=Attribute)
-Operand: TypeAlias = SSAValue
+from xdsl.utils.exceptions import VerifyException
 
 
 class ElementwiseBinOpBase(IRDLOperation, ABC):
     """Base class for element-wise binary operations on tensors with Numpy-style broadcasting."""
 
-    lhs: Operand = operand_def()
-    rhs: Operand = operand_def()
+    lhs = operand_def()
+    rhs = operand_def()
     res: OpResult = result_def(AnyTensorType)
+    assembly_format = "`(` $lhs `,` $rhs `)` attr-dict `:` `(` type($lhs) `,` type($rhs) `)` `->` type($res)"
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,91 @@
...
+ArgT = TypeVar("ArgT", bound=Attribute)
+Operand: TypeAlias = SSAValue
+
+
+class ElementwiseBinOpBase(IRDLOperation, ABC):
+    """Base class for element-wise binary operations on tensors with Numpy-style broadcasting."""
+
+    lhs: Operand = operand_def()
+    rhs: Operand = operand_def()
+    res: OpResult = result_def(AnyTensorType)
```
superlopuh, line 31:
```suggestion
    lhs = operand_def()
    rhs = operand_def()
    res = result_def(AnyTensorType)
```

tests/filecheck/dialects/onnx/onnx_ops.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+func.func @add(%arg0: tensor<1x2x6xf32>, %arg1: tensor<1x2x6xf32>) ->  tensor<1x2x6xf32> {
+  %0 = "onnx.Add"(%arg0, %arg1) {onnx_node_name = "/model.1/Add"} : (tensor<1x2x6xf32>, tensor<1x2x6xf32>) -> tensor<1x2x6xf32>
```
superlopuh, line 4: Could you please add tests for the other ops, also?

superlopuh (commented):
Looking great, thank you!

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,91 @@
...
+        for x in [
+            " : ",
+            " (",
+            self.lhs.type,
+            ", ",
+            self.rhs.type,
+            ") -> ",
+            self.res.type,
+        ]:
+            printer.print(x)
```
math-fehr, line 68:
```suggestion
    assembly_format = "`(` $lhs `,` $rhs `)` attr-dict `:` `(` type($lhs) `,` type($rhs) `)` `->` type($res)"
```
Sadly I realized that I never implemented the space directive \` \` so the `:` will miss the space before.
I'm adding a fix for it right now so you could keep the same syntax!

tests/filecheck/dialects/onnx/onnx_ops.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+func.func @add(%arg0: tensor<1x2x6xf32>, %arg1: tensor<1x2x6xf32>) ->  tensor<1x2x6xf32> {
```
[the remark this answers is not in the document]
brnorris03: Very useful, thanks! Updated

tests/filecheck/dialects/onnx/onnx_ops.mlir
brnorris03: Done

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,91 @@
...
+    Attribute,
+    IRDLOperation,
+    irdl_op_definition,
+    operand_def,
+    result_def,
+)
+from xdsl.parser import Parser
+from xdsl.printer import Printer
+
+ArgT = TypeVar("ArgT", bound=Attribute)
```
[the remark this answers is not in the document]
brnorris03: Good to know, thanks. I actually wasn't using this so removed for now but will use the typevars as needed

xdsl/dialects/onnx.py
brnorris03: I'm curious -- when do you prefer not to have type hints?

xdsl/dialects/onnx.py
brnorris03: Ah, but it's great you support assembly_format, so much more concise, thanks!

brnorris03 (commented):
Thanks for the helpful suggestions! I added some minimal validation and more tests.

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,93 @@
...
+            operands=[lhs, rhs],
+            attributes=attributes,
+            result_types=[[]],
+        )
+
+    def verify_(self) -> None:
+        # Check that the arguments are broadcastable (using Numpy semantics) and that the result type is correct.
+        res_shape: list[int] = []
+        # Iterate over the shapes in reverse order and compute the result shape.
+        lhs_shape: list[int] = [x.data for x in self.lhs.type.shape]
```
brnorris03, line 44: Is there a better way to get the shapes of tensor types (I hope so)?

superlopuh: Unfortunately not, we have to use isinstance checks for this
```suggestion
            if not isinstance(lhs_type := self.lhs.type, ShapedType):
                raise ValueError(
                    "onnx binary operation operands must be of type ShapedType"
                )
        lhs_shape: list[int] = [x.data for x in lhs_type.shape]
```

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,93 @@
...
+    lhs = operand_def()
+    rhs = operand_def()
+    res: OpResult = result_def(AnyTensorType)
+    assembly_format = "`(` $lhs `,` $rhs `)` attr-dict `:` `(` type($lhs) `,` type($rhs) `)` `->` type($res)"
+
+    def __init__(
+        self,
+        lhs: SSAValue,
+        rhs: SSAValue,
+        attributes: dict[str, Attribute] = {},
```
superlopuh, line 32: Just spotted this, could you please remove argument? We try to keep our APIs fully typed, and then add dynamic attributes separately. Now that I think about it, adding an attributes argument might be a good idea in general, but I'd rather keep the API consistent for now. For the custom parser, we have a number of examples of adding attributes after the operation has been created in the codebase

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,93 @@
...
+            if d2 == 1:
+                res_shape.append(d1)
+                continue
+            raise VerifyException(
+                f"operands have incompatible shapes: {lhs_shape} and {rhs_shape}"
+            )
+        # Reverse the result shape and check that it matches the result type.
+        res_shape.reverse()
+        if not all(
+            x == y for x, y in zip(res_shape, [v.data for v in self.res.type.shape])
```
superlopuh, line 66:
```suggestion
            x == y for x, y in zip(res_shape, (v.data for v in self.res.type.shape), strict=True)
```
This avoids the temporary list creation, and throws an error if the length of the iterators is not the same. BTW it's probably a good idea to add a separate length check in the verifier with a pretty error.

xdsl/dialects/onnx.py
superlopuh: We don't really have a strict style guide. When we migrated the code over to the new API I, regrettably, wrote a sed script that added the type hints. I now think they're just a bit noisy and not super helpful. The fact that you reached for a typealias to make it more explicit that the type corresponded to an operand is probably a good signal that they're confusing. I'd recommend against them whenever we use the `_def` things, except for existing files where we already have the type hints, where it might be worth keeping them for consistency.

xdsl/dialects/onnx.py
superlopuh: Ah nice, I forgot about that, much cleaner. Thanks @math-fehr

xdsl/dialects/onnx.py
math-fehr: Actually, you should be able to just write:
`lhs_shape = lhs_type.get_shape()`

It seems that `lhs` and `rhs` are not constrained to be a `Tensor` or a `ShapedType` in their definitions, is that expected?
Actually, are all operands and results supposed to be the same type? Here is an example of how we do it in xDSL: https://github.com/xdslproject/xdsl/blob/2ccde61cb96a8e8ea14c1313ebd6e5df514d84c0/xdsl/dialects/arith.py#L173

brnorris03: The `lhs` and `rhs` do not have to be the same type, however, they should have the same element type if they are tensors (e.g., all have to have `int32` elements, or be an `int32` scalar) and the shapes must be broadcastable (what the current validation checks). 

I will add a check for matching element types, which is easy enough; but I don't know how to create them in a way that allows scalar or any shaped type (including tensor).

math-fehr: Here is an example on how you could make this work:

```
class BinaryOperation(IRDLOperation):
  T = ConstraintVar("T", Attribute)
  lhs = operand_def(T | VectorOrTensorOf[T])
  rhs = operand_def(T | VectorOrTensorOf[T])
  res = result_def(T | VectorOrTensorOf[T])
```
This will ensure that `lhs`, `rhs`, and `res` are either a `Vector`, a ranked or unranked `Tensor`, and that they all have the same element type.

Making this work for `ShapedType` is a bit more complex, because we are not handling `ShapedType` well currently :/

brnorris03: The onnx spec itself actually doesn't allow scalars (I double checked), just ranked tensors, so perhaps something like this is more appropriate I think (similar to `arith`), can you confirm?
```python
_T = TypeVar("_T", bound=Attribute)

class ElementwiseBinOpBase(IRDLOperation):
    """Base class for element-wise binary operations on tensors with Numpy-style broadcasting."""

    T = Annotated[Attribute, ConstraintVar("T"), _T]
    lhs = operand_def(TensorType[T])
    rhs = operand_def(TensorType[T])
    res = result_def(TensorType[T])
  ```
  
In `verify_` the type of `lhs.type` is `xdsl.dialects.builtin.TensorType`, however vscode (and pyright) are not able to statically determine that, and so I still get errors when trying to access and use the shapes. Not sure how to fix those, type hints (e.g., `lhs_shape: tuple[int] = self.lhs.type.get_shape()`) are not enough. So, how can I get the shapes of the operands and result without pyright errors?

Comment (superlopuh) [reply xdsl/dialects/onnx.py:44]: I'm afraid you still need to add the snippet I suggested above, or something like it

Comment (brnorris03) [reply xdsl/dialects/onnx.py:44]: I did add the snippet above (and variations) but cannot get types resolved statically, sorry if this is basic, but am kind of out of ideas of what to try to make pyright happy. I can push the latest if it helps.

Comment (brnorris03) [reply xdsl/dialects/onnx.py]: Actually, never mind, I had a silly mistake. I was able to get rid of the pyright errors.

Comment (math-fehr) [reply xdsl/dialects/onnx.py]: Yes, your snippet seems correct!
The main issue here is that you cannot really statically guarantee that `self.lhs` is a `TensorType`, because if you are in the middle of a rewrite, then you cannot guarantee that your operation verifies.
I'm not completely sure how to fix that currently, to be honest, MLIR has now the concept of `TypedValue` that we could use, but I don't have an easy way of integrating that yet without making it a hassle to use :/

File: xdsl/dialects/onnx.py [hunk 9]
```diff
@@ -0,0 +1,105 @@
...
+from xdsl.irdl import (
+    ConstraintVar,
+    IRDLOperation,
+    irdl_op_definition,
+    operand_def,
+    result_def,
+)
+from xdsl.utils.exceptions import VerifyException
+
+_T = TypeVar("_T", bound=Attribute)
```
math-fehr, line 22: I think you can remove this now?

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class ElementwiseBinOpBase(IRDLOperation):
+    """Base class for element-wise binary operations on tensors with Numpy-style broadcasting."""
+
+    T = Annotated[Attribute, ConstraintVar("T"), _T]
```
math-fehr, line 28:
```suggestion
    T = Annotated[Attribute, ConstraintVar("T")]
```
Since you don't use `_T` in `ElementwiseBinOpBase`, you don't really need this.

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,105 @@
...
+    irdl_op_definition,
+    operand_def,
+    result_def,
+)
+from xdsl.utils.exceptions import VerifyException
+
+_T = TypeVar("_T", bound=Attribute)
+
+
+class ElementwiseBinOpBase(IRDLOperation):
```
math-fehr, line 25:
```suggestion
class ElementwiseBinOpBase(IRDLOperation, ABC):
```
To be sure we don't make the mistake of instantiating it directly.

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(
+        self,
+        lhs: SSAValue,
+        rhs: SSAValue,
+    ):
+        super().__init__(
+            operands=[lhs, rhs],
+            result_types=[[]],
+        )
```
math-fehr, line 42:
```suggestion
    def __init__(
        self,
        lhs: SSAValue,
        rhs: SSAValue,
        res_type: Attribute | None
    ):
        if res_type is None:
          res_type = lhs.type
        super().__init__(
            operands=[lhs, rhs],
            result_types=[res_type],
        )
```
The `res_type` argument is quite useful in the middle of a rewrite, when the operation would not verify.
Making it optional is quite useful as well, if by default you take the type of the first operand by example.

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Iterate over the shapes in reverse order and compute the result shape.
+        if (
+            not isinstance(lhs_type := self.lhs.type, TensorType)
+            or not isinstance(rhs_type := self.rhs.type, TensorType)
+            or not isinstance(res_type := self.res.type, TensorType)
+        ):
+            raise ValueError(
+                "onnx elementwise binary operation operands and result must be of type TensorType"
+            )
```
math-fehr, line 56: I would make an `assert` for this here, as the `verify_` function is always called after the IRDL verifier, which will always ensure that these are `TensorType`.
The `assert` makes it clearer that this is for pyright only.

math-fehr (approved):
Nice!
I just made some small style comments, but otherwise that looks good to me!
I would just wait for Sasha's approval.

xdsl/dialects/onnx.py
```diff
@@ -0,0 +1,105 @@
...
+class Mul(ElementwiseBinOpBase):
+    name = "onnx.Mul"
+
+
+@irdl_op_definition
+class Div(ElementwiseBinOpBase):
+    name = "onnx.Div"
+
+
+ONNX = Dialect("onnx", [Add, Sub, Mul, Div])
```
superlopuh, line 105: nit:
```suggestion
ONNX = Dialect("onnx", [Add, Sub, Mul, Div,])
```
Adding a comma here will make the formatter move the ops to their own lines, which will minimise diffs in the future. A really minor nit, feel free to ignore :)

xdsl/dialects/onnx.py
brnorris03: Makes sense (my initial attempt was based on examples, which may not have been the most appropriate). I think by default the result type should be computed based on both operands, but in any case, the suggested modification is better.

brnorris03: I think intiially, at least, I would not include `None` as an option to the result type because I think it's not essential to be able to omit.

xdsl/dialects/onnx.py
brnorris03: Related to this, I tried to figure out how to specify that the element type should be any of a specific subset of builtin types -- is there a good way to do this?

xdsl/dialects/onnx.py
math-fehr: Sure, works as well!

xdsl/dialects/onnx.py
math-fehr: Yes, just replace the `Attribute` with the type you want, something like:
```
T = Annotated[Float32Type | IntegerType, ConstraintVar("T")]
```

brnorris03: Thanks! Changed to `T = Annotated[AnyFloat | IntegerType, ConstraintVar("T")]`, which matches the spec more closely.

georgebisbas (approved):
Looks good!

superlopuh (approved):
Perfect, thank you!

PR #1917, opened 2023-12-27
core: Add support for lazily loading dialects

math-fehr:
This PR adds support for loading dialects only when necessary, which means we only load the associated Python files when necessary.

This makes my `lit tests/filecheck/ -v` go from 52s to 32s! Big win!


superlopuh:
Just tested this locally, `xdsl/tools/xdsl_opt.py tests/filecheck/backend/riscv/convert_func_to_riscv_func.mlir` takes 212ms instead of 330 on my laptop, and tracing works fine. Not sure what the deal was with the previous attempt but this seems to work great on my very limited testing :)

tests/test_mlcontext.py
```diff
@@ -1,39 +1,45 @@
...
 
 
 @irdl_attr_definition
 class DummyAttr2(ParametrizedAttribute):
-    name = "dummy_attr2"
+    name = "test.dummy_attr2"
+
+
+TestDialect = Dialect("test", [DummyOp], [DummyAttr])
+TestDialect2 = Dialect("test", [DummyOp2], [DummyAttr2])
```
superlopuh, line 29: I know this is usually the case we use for dialects but it reads a bit weird here. Could you please change them to `TEST_DIALECT` or `testDialect` for these tests?

xdsl/dialects/snitch_stream.py
```diff
@@ -204,7 +204,7 @@ def __init__(
 
 
 SnitchStream = Dialect(
-    "snitch-stream",
+    "snitch_stream",
```
superlopuh, line 207: 🙏

superlopuh (approved):

xdsl/ir/core.py
```diff
@@ -73,13 +73,19 @@ class MLContext:
     _loaded_dialects: dict[str, Dialect] = field(default_factory=dict)
     _loaded_ops: dict[str, type[Operation]] = field(default_factory=dict)
     _loaded_attrs: dict[str, type[Attribute]] = field(default_factory=dict)
+    _registered_dialects: dict[str, Callable[[], Dialect]] = field(default_factory=dict)
+    """
+    A dictionary of all registered dialects that are not yet loaded. This is used to
+    only load the respective Python files when the dialect is actually used.
+    """
```
webmiche, line 80: To me, it is not quite intuitive that a dialect is not registered anymore after loading it. I guess it makes some of the implementation easier, but we can just compute the information (registered dialects - loaded dialects) when we need it.

xdsl/tools/command_line_tool.py
```diff
@@ -4,89 +4,239 @@
...
-from xdsl.frontend.symref import Symref
+from xdsl.dialects.builtin import ModuleOp
 from xdsl.ir import Dialect, MLContext
 from xdsl.parser import Parser
 from xdsl.passes import ModulePass
 from xdsl.utils.exceptions import ParseError
 
 
-def get_all_dialects() -> list[Dialect]:
+def get_all_dialects() -> list[tuple[str, Callable[[], Dialect]]]:
```
webmiche, line 14: Isn't the returned thing here pretty much exactly a `dict[str, Callable[...]]`? Why don't we just make it one?

I think loading lazily in a method called `get_all_dialects()` makes the interface of the cli-tool not quite as intuitive anymore. Couldn't we just add another method? `get_all_dialects_lazy()` or whatever?

webmiche (commented):
Really nice to get this in. I think we need to figure out how exactly we want certain APIs to look like and behave to make this not too much of a burden on downstream users, but overall certainly a great direction to move in!

xdsl/tools/command_line_tool.py
superlopuh: This is ordered, which I think is a good thing

xdsl/ir/core.py
math-fehr: You would prefer not removing them? That's fine for me too!

xdsl/tools/command_line_tool.py
math-fehr: Hmm, I'm wondering if `dict` is indeed better. Right now, we have an order, but it's ordered semi-randomly.
We don't necessarily want people to rely on that order, so I would argue `dict` is better.
And I'm fairly against `get_all_dialects` in a non-lazy way. Having an easy API to load all dialects at once is a bit dangerous because people might use that directly without knowing it's better in a lazy_way.
Also, that's just simplifying the API for something that is likely called only once per project.

webmiche: I guess there exists OrderedDict if we care about the order? Just feels weird to me to see something used like a map/dict but not being called a dict :sweat_smile: 

I guess you are right on the get_all_dialects front :+1:

xdsl/ir/core.py
webmiche: Yes, I think it makes the name align with the actual behavior (in a registered dialect is one that previously has been registered).

xdsl/tools/command_line_tool.py
superlopuh: I guess the dict guarantees that the key is unique and that the order is undefined, so it would probably be a slight improvement

math-fehr:
@webmiche if this looks good for you, I'm good to merge it soon

webmiche (approved):
LGTM now! Cool to see this in!

PR #1923, opened 2023-12-31
interpreter: (riscv_scf) add interpreter function for while op

superlopuh:
Adds the necessary functions to the interpreter to interpret riscv_scf.while

### DIFF
--- a/tests/interpreters/test_riscv_scf_interpreter.py
+++ b/tests/interpreters/test_riscv_scf_interpreter.py
@@ -43,6 +43,50 @@ def sum_to_for_op():
         func.Return(result)
 
 
+def sum_to_while_fn(n: int) -> int:
+    """
+    Python implementation of sum_to_while_op
+    """
+    result = 0
+    i = 0
+    while i < n:
+        result += 1
+        i += 1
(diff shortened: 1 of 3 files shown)


tests/interpreters/test_riscv_scf_interpreter.py
```diff
@@ -43,6 +43,50 @@ def for_loop_region(args: tuple[BlockArgument, ...]):
...
+def sum_to_while_fn(n: int) -> int:
+    """
+    Python implementation of sum_to_while_op
+    """
+    result = 0
+    i = 0
+    while i < n:
+        result += 1
+        i += 1
+    return result
```
math-fehr, line 55: Should we comment this if this isn't called?

math-fehr (approved):
:+1: besides one small comment

tests/interpreters/test_riscv_scf_interpreter.py
superlopuh: Ah yeah I mean it's basically documentation becvause python is more readable than IR with builders, I'll comment it out.

PR #1924, opened 2023-12-31
dialects: (scf) add custom syntax for scf.condition

superlopuh:
Same as MLIR: https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfcondition-scfconditionop

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -556,14 +556,14 @@ class Condition(IRDLOperation):
         cond = parser.resolve_operand(unresolved_cond, IntegerType(1))
         attrs = parser.parse_optional_attr_dict()
 
-        # The optional list of ssa values can be confused with the list of results of the
-        # next operation. We save the position to backtrack in case we can't parse the
-        # types.
+        # scf.condition is a terminator, so the list of arguments cannot be confused with
+        # the results of a hypothetical operation on the next line.
         pos = parser.pos
         unresolved_arguments = parser.parse_optional_undelimited_comma_separated_list(
             parser.parse_optional_unresolved_operand, parser.parse_unresolved_operand
         )
-        if unresolved_arguments is not None and parser.parse_optional_punctuation(":"):
+        if unresolved_arguments is not None:
+            parser.parse_punctuation(":")
             types = parser.parse_comma_separated_list(
                 parser.Delimiter.NONE, parser.parse_type
             )
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -536,6 +536,45 @@ class Condition(IRDLOperation):
     def get(cond: SSAValue | Operation, *output_ops: SSAValue | Operation) -> Condition:
         return Condition(cond, *output_ops)
 
+    def print(self, printer: Printer):
+        printer.print("(", self.cond, ") ")
+        if self.attributes:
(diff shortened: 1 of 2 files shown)


xdsl/dialects/scf.py
```diff
@@ -536,6 +536,45 @@ def __init__(
     def get(cond: SSAValue | Operation, *output_ops: SSAValue | Operation) -> Condition:
         return Condition(cond, *output_ops)
 
+    def print(self, printer: Printer):
+        printer.print("(", self.cond, ") ")
+        if self.attributes:
+            printer.print_op_attributes(self.attributes)
+            printer.print(" ")
```
math-fehr, line 543: I'm pretty sure that this will print two spaces when printing the op attributes.
Also, it is missing a test I think?

xdsl/dialects/scf.py
```diff
@@ -536,6 +536,45 @@ def __init__(
...
+    def parse(cls, parser: Parser) -> Self:
+        parser.parse_punctuation("(")
+        unresolved_cond = parser.parse_unresolved_operand("cond expected")
+        parser.parse_punctuation(")")
+        cond = parser.resolve_operand(unresolved_cond, IntegerType(1))
+        attrs = parser.parse_optional_attr_dict()
+
+        # The optional list of ssa values can be confused with the list of results of the
+        # next operation. We save the position to backtrack in case we can't parse the
+        # types.
```
math-fehr, line 561: This is so wrong at so many levels, but I'm starting to understand why this works in MLIR.
You are actually not allowed to backtrack, so that's not how this is parsed. This is just parsed greedily, and if there is an SSAValue then it will try to parse everything. However, since `scf.condition` is technically a terminator, you will never be able to print custom format with an `scf.condition` followed by an operation.
This is a bit broken, but it's definitely not my priority to fix that in MLIR.

math-fehr: But that's what you implemented anyway, so I would just change the comment.

xdsl/dialects/scf.py
```diff
@@ -536,6 +536,45 @@ def __init__(
...
+        pos = parser.pos
+        unresolved_arguments = parser.parse_optional_undelimited_comma_separated_list(
+            parser.parse_optional_unresolved_operand, parser.parse_unresolved_operand
+        )
+        if unresolved_arguments is not None and parser.parse_optional_punctuation(":"):
```
math-fehr, line 566: This will parse `%arg0, %arg1` and discard it if it doesn't find the `:`.
Instead, you should just have a `parse_punctuation(':')` in the `if` directly, and just do the condition on `unresolved_arguments`.

math-fehr (commented):
I think one case is just broken, but otherwise it should be fine!
This parsing operation is definitely broken in MLIR though

xdsl/dialects/scf.py
superlopuh: it was printing an extra space before, not after, I updated the method

webmiche (commented):
Interop tests? :stuck_out_tongue_winking_eye:

superlopuh:
@webmiche I added one in #1927, since scf.condition has to be in a while op and that PR adds the custom syntax for scf.while

math-fehr (approved):
:+1:

webmiche (approved):

PR #1925, opened 2023-12-31
dialects: (riscv_scf) add custom syntax for riscv_scf.condition

superlopuh:
Follows scf syntax except for the cond itself, which now must have an explicit type to account for unknown register.


xdsl/dialects/riscv_scf.py
```diff
@@ -252,6 +252,47 @@ class ConditionOp(IRDLOperation):
     def __init__(self, cond: SSAValue | Operation, *output_ops: SSAValue | Operation):
         super().__init__(operands=[cond, output_ops])
 
+    def print(self, printer: Printer):
+        printer.print("(", self.cond, " : ", self.cond.type, ") ")
+        if self.attributes:
+            printer.print_op_attributes(self.attributes)
```
math-fehr, line 258: Same comment here than before, there is probably a space too much here

xdsl/dialects/riscv_scf.py
```diff
@@ -252,6 +252,47 @@ class ConditionOp(IRDLOperation):
...
+        unresolved_cond = parser.parse_unresolved_operand("cond expected")
+        parser.parse_punctuation(":")
+        cond_type = parser.parse_type()
+        parser.parse_punctuation(")")
+        cond = parser.resolve_operand(unresolved_cond, cond_type)
+        attrs = parser.parse_optional_attr_dict()
+
+        # The optional list of ssa values can be confused with the list of results of the
+        # next operation. We save the position to backtrack in case we can't parse the
+        # types.
```
math-fehr, line 279: Same comment here as before

math-fehr (approved):

PR #1926, opened 2023-12-31
core: add helper for function type printing

superlopuh:
Seems like it pops up a few times for custom syntax, I'd like to use this helper for the `scf` and `riscv_scf` `while` custom syntaxes.

### DIFF
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -733,29 +733,50 @@ class Printer:
     def print_function_type(
         self, input_types: Iterable[Attribute], output_types: Iterable[Attribute]
     ):
+        """
+        Prints a function type like `(i32, i64) -> (f32, f64)` with the following
+        format:
+
+        The inputs are always a comma-separated list in parentheses.
+        If the output has a single element, the parentheses are dropped, except when the
+        only return type is a function type, in which case they are kept.
+
+        ```
+        () -> ()                 # no inputs, no outputs
+        (i32) -> ()              # one input, no outputs
+        (i32) -> i32             # one input, one output
+        (i32) -> (i32, i32)      # one input, two outputs
+        (i32) -> ((i32) -> i32)  # one input, one function type output
+        ```
+        """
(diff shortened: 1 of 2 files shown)


superlopuh:
Ok turns out that the scf functions types in the custom syntax are exactly the op types anyway, so this separate method isn't necessary quite yet. I still think it might be a good idea to add it. Happy to add tests for it if anyone things that it would be a useful thing to factor out cc @math-fehr

xdsl/printer.py
```diff
@@ -729,25 +730,38 @@ def print_op_with_default_format(self, op: Operation) -> None:
...
+                self.print("(", first_type, ")")
             else:
-                self.print(res_type)
-        else:
-            self.print("(")
-            self.print_list(
-                op.results, lambda result: self.print_attribute(result.type)
-            )
-            self.print(")")
+                self.print(first_type)
```
math-fehr, line 754: I'm not a huge huge fan of this style, especially since it's not in a hot loop or anything.
Do you think we could either make the requirement a `Sequence`, or cast this to a list to then use a match/case that is much more readable?

math-fehr (commented):
Looks okay to me to have this function.
I would however add a comment on it with the syntax (it was missing in print_operation_type too).
Also I would add one test for it at least.

xdsl/printer.py
superlopuh: It feels like the most idiomatic way of doing things to me, and it doesn't feel particularly weird. It's a clear reflection of the logic to me, so it's not optimisation for optimisation's sake, but as a reflection of the behaviour, which seems like a good idea in general. I'd rather keep it as-is, unless you really have a strong preference to refactor it. I'm happy to add comments to make it clearer if you'd like.

math-fehr: Okay, I'm fine with it if you feel more strong towards it!

math-fehr (approved):
:+1:

PR #1927, opened 2023-12-31
dialects: (scf) add custom syntax for scf while

superlopuh:
Same as MLIR: https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -536,6 +536,45 @@ class Condition(IRDLOperation):
     def get(cond: SSAValue | Operation, *output_ops: SSAValue | Operation) -> Condition:
         return Condition(cond, *output_ops)
 
+    def print(self, printer: Printer):
+        printer.print("(", self.cond, ") ")
+        if self.attributes:
+            printer.print_op_attributes(self.attributes)
+            printer.print(" ")
+        if self.arguments:
+            printer.print_list(self.arguments, printer.print_ssa_value)
+            printer.print_string(" : ")
+            printer.print_list(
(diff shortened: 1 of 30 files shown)


xdsl/dialects/scf.py
```diff
@@ -81,6 +81,67 @@ def verify_(self):
...
+        function_type = parser.parse_function_type()
+
+        def resolve_argument(arg: parser.Argument, type: Attribute):
+            arg.type = type
+            return arg
+
+        block_args = tuple(
+            resolve_argument(block_arg, t)
+            for ((block_arg, _), t) in zip(
+                tuples, function_type.inputs.data, strict=True
```
math-fehr, line 123: You should return a Parsing Error if the function types and the inputs are not matching.

xdsl/dialects/scf.py
```diff
@@ -81,6 +81,67 @@ def verify_(self):
...
+            )
+        )
+
+        arguments = tuple(
+            parser.resolve_operand(operand, t)
+            for ((_, operand), t) in zip(tuples, function_type.inputs.data, strict=True)
```
math-fehr, line 129: Same here

math-fehr (approved):
Just one error that I think should be catched, and looks good to me!

xdsl/dialects/scf.py
superlopuh: it's actually the same value error we just iterate twice

webmiche (commented):
Interop tests? :upside_down_face:

webmiche (approved):

PR #1930, opened 2024-01-01
dialects: (arith) make binary ops pure

superlopuh:


tobiasgrosser:
Why is there `riscv` in the title? This seems to all be about the `arith` dialect, no?

tobiasgrosser (approved):

PR #1932, opened 2024-01-01
interpreter: (linalg) allow scalar params for linalg generic

superlopuh:


tests/interpreters/test_linalg_interpreter.py
```diff
@@ -91,3 +91,49 @@ def test_linalg_generic():
     interpreter.run_op(op, (a, b, c))
 
     assert c.data == [1, 4, 9, 16, 25, 36]
+
+
+def test_linalg_generic_scalar():
```
math-fehr, line 96: Out of curiosity, why doing the test here rather than in an MLIR file?
I find MLIR files overall more readable, and making more sense since they test exactly what we want.

superlopuh: TBH mostly because I already had this file lying around. I'll take a look into converting this to filecheck.

superlopuh: I usually add python files for individual ops but the generic one is a bit awkward to test for individually, easier to set up more of an integration style test

superlopuh: Actually let's merge this and I create a task to do this later? Feels less time-sensitive than some other things I'd like to move on to in the short term.

AntonLydike (approved):
Would love to see a description of what this is doing in the PR, makes it easier to review the code! LGTM I think!

PR #1936, opened 2024-01-02
core: add method for splitting block before a given operation

superlopuh:
This is useful for things like scf to cf rewriting.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1478,6 +1478,65 @@ class Block(IRNode):
 
             existing_op = op
 
+    def split_before(self, b_first: Operation) -> Block:
+        """
+        Split the block into two blocks before the specified operation or iterator.
+
+        Note that all operations before the specified iterator stay as part of the
+        original basic block, and the rest of the operations in the original block are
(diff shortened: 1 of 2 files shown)


xdsl/ir/core.py
```diff
@@ -1478,6 +1478,65 @@ def insert_ops_after(
...
+    def split_before(self, b_first: Operation) -> Block:
+        """
+        Split the block into two blocks before the specified operation.
+
+        Note that all operations before the one given stay as part of the original basic
+        block, and the rest of the operations in the original block are moved to the new
+        block, including the old terminator.
+        The original block is left without a terminator.
+        The newly formed Block is inserted into the parent region immediately after `self`
+        and returned.
```
webmiche, line 1490:
```suggestion
        The newly formed block is inserted into the parent region immediately after `self`
        and returned.
```

webmiche (approved):

superlopuh:
@math-fehr I'm merging this to lower the number of open PRs but please have a look in case we missed anything

math-fehr:
Looking good! Nothing to add, thanks for the ping

Issue #1921, opened 2023-12-28
add onnx.Relu op


brnorris03:

kayode-gif:
id like to work on this

PR #1939, opened 2024-01-04
dialects: (onnx) add onnx.Relu

kayode-gif:
resoves #1921


tests/filecheck/dialects/onnx/onnx_ops.mlir
```diff
@@ -16,3 +17,6 @@
 
 %res_div = "onnx.Div"(%t6, %t7) {onnx_node_name = "/Div"}: (tensor<1x5x1x3xf32>, tensor<2x1x6x3xf32>) -> tensor<2x5x6x3xf32>
 // CHECK: %res_div = onnx.Div(%t6, %t7) {"onnx_node_name" = "/Div"}: (tensor<1x5x1x3xf32>, tensor<2x1x6x3xf32>) -> tensor<2x5x6x3xf32>
+
+%res_relu = "onnx.Relu"(%t8) {onnx_node_name = "/Relu"}: (tensor<3x4xf32>) -> tensor<3x4xf32>
+// CHECK: %res_relu = onnx.Relu(%t8) {"onnx_node_name" = "/Relu"}: (tensor<3x4xf32>) -> tensor<3x4xf32>
```
superlopuh, line 22:
```suggestion
// CHECK: %res_relu = onnx.Relu(%t8) {"onnx_node_name" = "/Relu"}: (tensor<3x4xf32>) -> tensor<3x4xf32>

```
We try to add newlines to the end of all the files, could you add one here also, please? one of the things this helps with is git, which considers the newline a part of a line.

kayode-gif: can i just commit your suggestion?

tobiasgrosser: Yes

xdsl/dialects/onnx.py
```diff
@@ -100,12 +100,43 @@ class Div(ElementwiseBinOpBase):
...
+
+    def verify_(self) -> None:
+        if (
+                not isinstance(self.operand.type, TensorType)
+                or not isinstance(self.res.type, TensorType)
+                or self.operand.type != self.res.type
+        ):
+            assert (
+                False
+            ), "onnx elementwise operation operand and result must be of type TensorType"
```
superlopuh, line 131: pyright is a bit pedantic here, because the types are generic, and the generic parameter is unknown. We can actually avoid the whole verification thing with a ConstraintVar

superlopuh: Oh no actually we can't do it with a constraint var, I'd recommend something like this:

```suggestion
        if (
                not isinstance(operand_type := self.operand.type, TensorType)
                or not isinstance(res_type := self.res.type, TensorType)
        ):
            assert (
                False
            ), "onnx elementwise operation operand and result must be of type TensorType"
            operand_type = cast(TensorType[Attribute], operand_type)
            res_type = cast(TensorType[Attribute], res_type)
            
            if operand_type != res_type:
                raise VerificationException("Mismatch between operand type and res type")
```

superlopuh:
For the code formatting, can you please follow the instructions in the README?

superlopuh (approved):
Great! Thank you

xdsl/dialects/onnx.py
```diff
@@ -100,12 +100,47 @@ class Div(ElementwiseBinOpBase):
...
+            result_types=[operand.type],
+        )
+
+    def verify_(self) -> None:
+        if not isinstance(
+            operand_type := self.operand.type, TensorType
+        ) or not isinstance(res_type := self.res.type, TensorType):
+            assert (
+                False
+            ), "onnx elementwise operation operand and result must be of type TensorType"
```
brnorris03, line 129: Minor -- the assert message can be Relu-specific, but it's also fine as is.

brnorris03 (approved):
Looks good to me!

PR #1942, opened 2024-01-05
misc: (Toy) add toy run action for VSCode

superlopuh:
Allows us to press play and attach to the Toy compiler, helps for debugging.


AntonLydike (approved):
Very nice!

PR #1943, opened 2024-01-05
core: Add region inlining helpers to rewriter (before/after/at_start/at_end)

superlopuh:


xdsl/rewriter.py
```diff
@@ -186,3 +186,15 @@ def move_region_contents_to_new_regions(region: Region) -> Region:
...
+    def inline_region_before(region: Region, target: Block) -> None:
+        """Move the region blocks to an existing region."""
+        parent_region = target.parent
+        if parent_region is None:
+            raise ValueError("Cannot inline region before a block with no parent")
+        for block in region.blocks:
+            block.parent = None
+        pos = parent_region.get_block_index(target)
+        parent_region.insert_block(region.blocks, pos)
+        region.blocks = []
```
math-fehr, line 200: Can you also add the `inline_region_after`, `inline_region_at_end`, and `inline_region_at_start`?
That way, we are sure we don't forget about them.

xdsl/rewriter.py
```diff
@@ -186,3 +186,15 @@ def move_region_contents_to_new_regions(region: Region) -> Region:
             new_region.add_block(block)
         region.blocks = []
         return new_region
+
+    @staticmethod
+    def inline_region_before(region: Region, target: Block) -> None:
+        """Move the region blocks to an existing region."""
+        parent_region = target.parent
+        if parent_region is None:
```
math-fehr, line 194: You also need to check that `target.parent` is not `region`, otherwise this will break.

math-fehr (commented):
Nice!
There is just one small bug, otherwise I just think that we should the other similar function.
That way, we don't forget about them, and I don't really think it complexifies the PR process (I actually think it would make it faster by reviewing everything at the same time).

xdsl/pattern_rewriter.py
```diff
@@ -126,6 +126,8 @@ def _can_modify_region(self, region: Region) -> bool:
         """Check if the region and its children can be modified by this rewriter."""
         if region.parent is None:
             return True  # Toplevel operation of current_operation is always a ModuleOp
+        if region is self.current_operation.parent_region():
+            return True
```
superlopuh, line 130: @math-fehr, to make the test pass, I had to add this line. I feel that it makes sense, we should be able to modify the regions that are the direct parents of the matched operation.

AntonLydike (approved):
Very nice work! Thanks for this!

xdsl/pattern_rewriter.py
```diff
@@ -471,6 +473,58 @@ def move_region_contents_to_new_regions(self, region: Region) -> Region:
             )
         return Rewriter.move_region_contents_to_new_regions(region)
 
+    def inline_region_before(self, region: Region, target: Block) -> None:
+        """Move the region blocks to an existing region."""
+        self.has_done_action = True
+        if not self._can_modify_region(region):
+            raise Exception(
+                "Cannot move regions that are not children of the matched operation"
+            )
```
webmiche, line 482: I guess we could get rid of some code duplication here by making `_can_modify_region` something like `_assert_can_modify_region` and make it raise the Exception itself (the exception seems to be a pretty generic string in all cases anyways).

webmiche (approved):
LGTM

xdsl/pattern_rewriter.py
superlopuh: Yeah that's a good idea, could you please file an issue? Seems like a good starter bug for someone

math-fehr: Just so you know, these restrictions will definitely be removed in like 1-2 PRs, so we won't need to change that at all!

math-fehr (approved):
Nice!

PR #1946, opened 2024-01-06
dialects: (riscv_cf) add riscv_cf dialect

superlopuh:
This adds the riscv_cf dialect, duplicating some instructions in the riscv dialect, but with dataflow semantics with successors etc. This lets us reason about dataflow using xDSL IR constructs, and is a step towards being able to interpret exactly the same IR that we use to print assembly.

I forgot to add two things, will do them tomorrow:
 - [x] attributes
 - [x] comments in assembly

### DIFF
--- a/xdsl/dialects/riscv_cf.py
+++ b/xdsl/dialects/riscv_cf.py
@@ -146,6 +146,12 @@ class ConditionalBranchOperation(IRDLOperation, RISCVInstruction, ABC):
             self.else_arguments, lambda val: _print_type_pair(printer, val)
         )
         printer.print_string(")")
+        if self.attributes:
+            printer.print_op_attributes(
+                self.attributes,
+                reserved_attr_names="operandSegmentSizes",
+                print_keyword=True,
+            )
 
     @classmethod
     def parse(cls, parser: Parser) -> Self:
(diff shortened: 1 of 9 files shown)


xdsl/dialects/riscv_cf.py
```diff
@@ -0,0 +1,406 @@
...
+    if (x[rs1] >=u x[rs2]) pc += sext(offset)
+
+    https://msyksphinz-self.github.io/riscv-isadoc/html/rvi.html#bgeu
+    """
+
+    name = "riscv_cf.bgeu"
+
+
+@irdl_op_definition
+class BranchOp(IRDLOperation, riscv.RISCVOp):
```
superlopuh, line 247: Taking suggestions for names for this operation, which is basically a no-op. Its use is to fall into a block that can be jumped to as well.

AntonLydike: Hmmm, how about `yield_through`?

AntonLydike (approved):
Nice work!

xdsl/dialects/riscv_cf.py
superlopuh: Oops, forgot to do this, let's rename later if we feel like it

PR #1951, opened 2024-01-08
dialects: (onnx) add onnx.Reshape

### LINKED ISSUE #1922: add onnx.Reshape op

kayode-gif:
resolves #1922

### DIFF
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -376,7 +376,7 @@ class Reshape(IRDLOperation):
             raise VerifyException("Shape tensor rank must not be equal to -1")
 
         # There is currently only support for rank one shape tensors in onnx-mlir
-        # Shape tensor must have a constant shape
+        # Shape tensor must have a constant shape.
         if len(shape_type) != 1:
             raise VerifyException("Shape tensor must have a rank one")
 
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -337,19 +337,18 @@ class Reshape(IRDLOperation):
             missing_dim = total_elements // specified_dim
             new_shape[index_of_minus_one] = missing_dim
 
-        # Handle case where dimension is zero (TODO)
-        shape_type_list = list(shape_type)
-        for i, dim in enumerate(shape_type_list):
+        # Handle case where dimension is zero
+        for i, dim in enumerate(new_shape):
             if dim == 0:
                 if self.allow_zero:
                     # If allow_zero is set, explicitly set the dimension to zero  (i.e. not taken from input tensor)
-                    shape_type_list[i] = 0
+                    new_shape[i] = 0
                 else:
                     # dimension is 0, leave it unchanged  (i.e. taken from the input tensor).
-                    shape_type_list[i] = data_type[i]
-        # # Shape (second input) could be an empty shape, which means converting to a scalar.
-        if not shape_type_list:
-            shape_type_list = [IntegerType(64)]
+                    new_shape[i] = data_type[i]
+        # Shape (second input) could be an empty shape, which means converting to a scalar.
+        if len(shape_type) == 0:
+            shape_type = IntegerType(64)
 
 
 ONNX = Dialect(
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -371,7 +371,7 @@ class Reshape(IRDLOperation):
         shape_type = shape_type.get_shape()
         reshaped_type = reshaped_type.get_shape()
 
-        # Shape tensor rank can't be -1
+        # Shape tensor rank can't be -1.
         if shape_type[0] == -1:
             raise VerifyException("Shape tensor rank must not be equal to -1")
 
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -338,18 +338,18 @@ class Reshape(IRDLOperation):
             new_shape[index_of_minus_one] = missing_dim
 
         # Handle case where dimension is zero
-        # shape_type_list = list(shape_type)
-        # for i, dim in enumerate(shape_type_list):
-        #     if dim == 0:
-        #         if self.allow_zero:
-        #             # If allow_zero is set, explicitly set the dimension to zero  (i.e. not taken from input tensor)
-        #             shape_type_list[i] = 0
-        #         else:
-        #             # dimension is 0, leave it unchanged  (i.e. taken from the input tensor).
-        #             shape_type_list[i] = data_type[i]
+        shape_type_list = list(shape_type)
+        for i, dim in enumerate(shape_type_list):
+            if dim == 0:
+                if self.allow_zero:
+                    # If allow_zero is set, explicitly set the dimension to zero  (i.e. not taken from input tensor)
+                    shape_type_list[i] = 0
+                else:
+                    # dimension is 0, leave it unchanged  (i.e. taken from the input tensor).
+                    shape_type_list[i] = data_type[i]
         # # Shape (second input) could be an empty shape, which means converting to a scalar.
-        # if not shape_type_list:
-        #     shape_type_list = [IntegerType(64)]
+        if not shape_type_list:
+            shape_type_list = [IntegerType(64)]
 
 
 ONNX = Dialect(
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -312,11 +312,12 @@ class Reshape(IRDLOperation):
         shape_type = shape_type.get_shape()
         reshaped_type = reshaped_type.get_shape()
 
-        # Shape tensor rank can't be -1 and it must have a constant shape
+        # Shape tensor rank can't be -1
         if shape_type[0] == -1:
             raise VerifyException("Shape tensor rank must not be equal to -1")
 
         # There is currently only support for rank one shape tensors in onnx-mlir
+        # Shape tensor must have a constant shape
         if len(shape_type) != 1:
             raise VerifyException("Shape tensor must have a rank one")
 
@@ -346,6 +347,7 @@ class Reshape(IRDLOperation):
                 else:
                     # dimension is 0, leave it unchanged  (i.e. taken from the input tensor).
                     new_shape[i] = data_type[i]
+
         # Shape (second input) could be an empty shape, which means converting to a scalar.
         if len(shape_type) == 0:
             shape_type = IntegerType(64)
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -254,6 +254,104 @@ class Gemm(IRDLOperation):
             )
 
 
+@irdl_op_definition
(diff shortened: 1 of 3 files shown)


xdsl/dialects/onnx.py
```diff
@@ -255,6 +255,83 @@ def verify_(self) -> None:
             )
 
 
+@irdl_op_definition
+class Reshape(IRDLOperation):
```
superlopuh, line 259: is there any documentation we could add here as a doc comment?

xdsl/dialects/onnx.py
```diff
@@ -255,6 +255,83 @@ def verify_(self) -> None:
...
+    name = "onnx.Reshape"
+    T = Annotated[AnyFloat | IntegerType, ConstraintVar("T")]
+    data = operand_def(TensorType[T])
+    shape = operand_def(TensorType[IntegerType(64)])
```
superlopuh, line 263: I have a feeling we can't specify the bitwidth here, will have to do that by hand in the `verify_` method

xdsl/dialects/onnx.py
kayode-gif: ill add the doc in now

xdsl/dialects/onnx.py
compor: yeah, I think so, @kayode-gif a usage of the `width` field can be found here: https://github.com/xdslproject/xdsl/blob/fbc1a7996ff6e433e9a9b24a679b7165d00b7142/xdsl/dialects/builtin.py#L425
and an example of a custom `verify_` method here: https://github.com/xdslproject/xdsl/blob/fbc1a7996ff6e433e9a9b24a679b7165d00b7142/xdsl/dialects/seq.py#L58

kayode-gif: i've added that verification check in the `verify_` method now, i don't think i have to do any other check with the bitwidth?

xdsl/dialects/onnx.py
```diff
@@ -256,6 +256,98 @@ def verify_(self) -> None:
...
+            data_type = cast(TensorType[Attribute], data_type)
+            reshaped_type = cast(TensorType[Attribute], reshaped_type)
+            shape_type = cast(TensorType[Attribute], shape_type)
+
+            if data_type != reshaped_type:
+                raise VerifyException("Mismatch between operand type and res type")
+            if shape_type.element_type != IntegerType(64):
+                raise VerifyException(
+                    "shape element type has to be a 64-bit signless integer"
+                )
```
compor, line 307: these seem to have the wrong indentation.

xdsl/dialects/onnx.py
```diff
@@ -256,6 +256,98 @@ def verify_(self) -> None:
...
+                False
+            ), "onnx elementwise operation operands and result must be of type TensorType"
+
+            data_type = cast(TensorType[Attribute], data_type)
+            reshaped_type = cast(TensorType[Attribute], reshaped_type)
+            shape_type = cast(TensorType[Attribute], shape_type)
+
+            if data_type != reshaped_type:
+                raise VerifyException("Mismatch between operand type and res type")
+            if shape_type.element_type != IntegerType(64):
```
compor, line 304: this is not exercised by a test case, can you please add one?

Moreover, on the same note, can you please check the code coverage report and increase test coverage as much as possible?

xdsl/dialects/onnx.py
```diff
@@ -256,6 +256,98 @@ def verify_(self) -> None:
...
+            and shape_type_list.count(0) > 0
+            and shape_type_list.count(-1) > 0
+        ):
+            raise VerifyException(
+                "Invalid shape: Both 0 and -1 are present, making the dimension ambiguous."
+            )
+        # At most one dimension of the new shape can be -1.
+        # In this case, the value is inferred from the size of the tensor and the remaining dimensions.
+        count_minus_one = shape_type_list.count(-1)
+        if count_minus_one == 1:
```
compor, line 325: what happens if there are more than `-1` specified dimension sizes? does mlir-onnx handle this (see the `ShapeHelper.cpp` if necessary)?

compor (changes requested):
I left some initial comments. Please remove the draft status if you feel this is closer to being merged.

xdsl/dialects/onnx.py
compor: @kayode-gif I've added this in my initial comments below

xdsl/dialects/onnx.py
```diff
@@ -256,6 +256,98 @@ def verify_(self) -> None:
...
+    inferred from the size of the tensor and the remaining dimensions. A dimension
+    could also be 0, in which case the actual dimension value is unchanged (i.e. taken
+    from the input tensor). Shape (second input) could be an empty shape, which means converting to a scalar.
+    The input tensor's shape and the output tensor's shape are required to have the same number of elements.
+    """
+
+    name = "onnx.Reshape"
+    T = Annotated[AnyFloat | IntegerType, ConstraintVar("T")]
+    data = operand_def(TensorType[T])
+    shape = operand_def(TensorType)
```
brnorris03, line 274: The shape in onnx is a tensor of 64-bit signless integer values, so I think constraining it at least to an integer here is better than leaving it as a tensor with arbitrary element types.

compor: This is checked in `verify_` as @kayode-gif has implemented ATM. We are just missing a test case in invalid ops for lit.

xdsl/dialects/onnx.py
kayode-gif: seems onnx-mlir doesnt handle this, i looked at the helper and it seems it only supports shapes of rank one, so i dont think it has to be handled

compor: Yeah, I glanced at it again earlier and nothing obvious, I just find it weird that it is explicitly mentioned as a constraint in the documentation. I wonder what it does if given more than one.

xdsl/dialects/onnx.py
brnorris03: Ok. Based on previous reviews, it's still unclear (to me) when the actual definition should be annotated with types and when most checking should be deferred to `verify_`. I would just like to understand better why it's preferable not to use a type constraint for the shape here (and instead check in `verify_`).

compor: My understanding is that the bitwidth is not part of the type in Python "typeland" and in the previous formulation in this PR it was just a Python instance; Hence cannot be checked as part of the type during type inference.
I vaguely remember having trouble with this when I needed some similar restriction on types in the past.

@superlopuh or @math-fehr might be able to explain this better.

superlopuh:
It seems that this is pretty much ready to review, but I would say that the two new issues in the project would be good to merge before adding more ops. Could you please have a go at them?

xdsl/dialects/onnx.py
compor: @kayode-gif can you please resolve review issues that are simple enough and maybe ping again for a review of features that you might be a bit more certain about?
It makes things easier when revisiting PRs, thanks!

For example, the above nit is simple enough and I trust that you can handle it well.

xdsl/dialects/onnx.py
compor: After looking at the docs, this refers to the contents of the 1-d shape tensor, so something like this:

```mlir
 %shape = onnx.Constant dense<[-1, -1, 2]> : tensor<3xi64>
```

would cause a runtime error, so there is nothing for this patch to act on.

xdsl/dialects/onnx.py
kayode-gif: okay, i have now handled that indentation i believe and the couple other issues

xdsl/dialects/onnx.py
superlopuh: I agree, with @brnorris03, constraining it to an IntegerAttr seems like a good idea, and that should be supplemented with a check for 64-bit signless integer in the `verify_` method

xdsl/dialects/onnx.py
```diff
@@ -254,6 +254,105 @@ def verify_(self) -> None:
...
+                if self.allow_zero:
+                    # If allow_zero is set, explicitly set the dimension to zero  (i.e. not taken from input tensor)
+                    new_shape[i] = 0
+                else:
+                    # dimension is 0, leave it unchanged  (i.e. taken from the input tensor).
+                    new_shape[i] = data_type[i]
+
+        # Shape (second input) could be an empty shape, which means converting to a scalar.
+        if len(shape_type) == 0:
+            shape_type = IntegerType(64)
```
superlopuh, line 353: what does this do?

kayode-gif: i guess this is code that isn't exactly supposed to be in `verify_`, but the first bit checks if atleast one dim of the reshaped tensor (new shape) is -1, then it infers the actual dimension, the next bit (for loop section) checks the dimension is zero in the new shape according the the conditions in the doc

kayode-gif: i guess i could put these in helper functions actually

superlopuh: it's just that it looks like dead code, this has no effect

superlopuh:
Looks like a merge commit went wrong

kayode-gif:
> Looks like a merge commit went wrong

yup, it did, im trying to sort it out

xdsl/dialects/onnx.py
kayode-gif: okay, is it because im not setting it to the inital right

superlopuh: this is setting a local variable, not `self`. I would say that `verify_` should never mutate state, this sort of inference may be better done in a rewrite pass.

kayode-gif: @superlopuh , i've been able to sort the file changes out, but i've been trying to understand why code is dead code in the block you highlighted, if i want to add what i said is that even needed now?

superlopuh: I would recommend removing this part of the logic, and then I think it should be ready to merge

kayode-gif: okay i will remove it, id appreciate (if you have time later) to give me a bit more of an understanding on these sort of inferences, or point me to somewhere to read more on it

superlopuh (approved):
LGTM!

xdsl/dialects/onnx.py
superlopuh: Sure! Chapter 4 of the MLIR Toy tutprial talks about it a little bit, although it's made a little more complicated than it needs to be by using a lot of abstractions for the implementation of the shape inference: https://mlir.llvm.org/docs/Tutorials/Toy/Ch-4/

The main idea is to be able to represent the IR at all, even with casts and reshapes, and then have a dedicated rewrite step to rewrite it into the cleaner representation once the shape information is available. The ONNX operations we probably want to represent as close to the input as possible, and then use compiler transformations to mutate them, to leverage all the infrastructure we've built for that.

kayode-gif:
@compor i believe its ready to look at

compor (approved):

PR #1952, opened 2024-01-08
core: (parser) add UnresolvedArgument with no type

superlopuh:
Argument will have a type

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -200,7 +200,9 @@ def test_parse_argument(
     if expected is not None:
         assert res is not None
         assert res.name.text[1:] == expected[0]
-        assert res.type == expected[1]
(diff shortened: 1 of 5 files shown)


tests/test_parser.py
```diff
@@ -200,7 +200,9 @@ def test_parse_argument(
     if expected is not None:
         assert res is not None
         assert res.name.text[1:] == expected[0]
-        assert res.type == expected[1]
+        if expected[1] is not None:
+            assert isinstance(res, Parser.Argument)
+            assert res.type == expected[1]
```
math-fehr, line 205: Add an else case to check that we have a `Parser.UnresolvedArgument`?

math-fehr (approved):
Very nice :rocket:

PR #1954, opened 2024-01-09
core: (printer) use the same code to print op attrs, properties, and DictionaryAttr

superlopuh:
We currently are not consistent between the two, which is a problem when printing dictionaries that have UnitAttr as a value, which is supposed to be printed without the `=`


math-fehr (approved):
:+1:

PR #1958, opened 2024-01-09
backend: (riscv_cf) add convert-riscv-scf-to-riscv-cf

superlopuh:


tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
```diff
@@ -0,0 +1,84 @@
+// RUN: xdsl-opt -p convert-riscv-scf-to-riscv-cf --split-input-file %s | filecheck %s
+
+// sum(range(arg0, arg1))
+
+builtin.module {
```
compor, line 5: This is great, but I'd argue the simplest example to see the lowering (especially the labels, etc) would be to reduce the "noise" from loop-carried args and other statements.

Could you add something like this, please?

```mlir
riscv_func.func @foo(%0 : !riscv.reg<a0>, %1 : !riscv.reg<a1>) {
  %2 = riscv.li 1 : () -> !riscv.reg<a2>
  %3 = riscv.li 0 : () -> !riscv.reg<a3>
  riscv_scf.for %4 : !riscv.reg<a4> = %0 to %1 step %2 {
      %5 = riscv.add %2, %3 : (!riscv.reg<a2>, !riscv.reg<a3>) -> !riscv.reg<a3>
      riscv_scf.yield
  }
  riscv_func.return
}
```

You can use this as is, I've tried it.

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
```diff
@@ -0,0 +1,166 @@
...
+    op_type_rewrite_pattern,
+)
+
+
+class LowerRiscvScfForPattern(RewritePattern):
+    """
+    Create a CFG subgraph for the loop around its body blocks (if the body
+    contained other loops, they have been already lowered to a flow of blocks).
+    Maintain the invariants that a CFG subgraph created for any loop has a single
+    entry and a single exit, and that the entry/exit blocks are respectively
```
compor, line 17: I'm a bit confused here, because from what I understand we do not do anything specifically to maintain the single entry-exit invariant, right? we get this for free coming from `scf`.
The latter (i.e., placement of blocks) is fine.

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
```diff
@@ -0,0 +1,166 @@
...
+    first/last blocks in the parent region.  The original loop operation is
+    replaced by the initialization operations that set up the initial value of
+    the loop induction variable (%iv) and computes the loop bounds that are loop-
+    invariant for affine loops.  The operations following the original scf.for
+    are split out into a separate continuation (exit) block. A condition block is
```
compor, line 22: [can you please follow standard loop terminology? you say `continuation` while in the diagram below and the code you label that as `end`. The "conventional" name for that block is `exit` (at least in LLVM) and I'd use that throughout (including generated labels)

More info here: https://llvm.org/docs/LoopTerminology.html

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
```diff
@@ -0,0 +1,166 @@
...
+    those blocks. Induction variable modification is appended to the last block
+    of the body (which is the exit block from the body subgraph thanks to the
+    invariant we maintain) along with a branch that loops back to the condition
+    block. Loop-carried values are the loop terminator operands, which are
+    forwarded to the branch.
+
+    ```
+
+         +--------------------------------------------------------------+
+         |   <code before the ForOp>                                    |
```
compor, line 37:
```suggestion                              |
```

I'd remove this as this happens anyway. Same for the exit block below.

compor (approved):
LGTM left some minor comments.

tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
```diff
@@ -0,0 +1,84 @@
+// RUN: xdsl-opt -p convert-riscv-scf-to-riscv-cf --split-input-file %s | filecheck %s
+
+// sum(range(arg0, arg1))
```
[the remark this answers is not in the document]
superlopuh: This is the python version of foo below, it computes the sum of the integers from arg0 to arg1

tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
superlopuh: I agree that it would be good to have a loop with no iter args, I'll add one.

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
superlopuh: I think it's a helpful visualisation of the fact that the code before the op is in the same block as the initial %iv value and bge op, similarly for the exit block.

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
superlopuh: This is exactly from [llvm](https://github.com/llvm/llvm-project/blob/aec73eade7af0e22c944714bec31570181bc1ad4/mlir/lib/Conversion/SCFToControlFlow/SCFToControlFlow.cpp#L50), but I'm happy to reword it to whatever you suggest :)

superlopuh: Similarly for above, I didn't feel like rewriting this thing but it could definitely be made more approachable.

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
compor: OK, no strong feelings about it

xdsl/backend/riscv/lowering/convert_riscv_scf_to_riscv_cf.py
compor: I meant llvm "proper" not mlir; including a lot of compiler bibliography/projects and the LLVM doc I linked reflecting what llvm IR gets generated.
It might be approachable to a broader audience, that's my point.

I'll leave it up to you.

PR #1960, opened 2024-01-09
backend: (riscv) remove lowering to labels

superlopuh:


superlopuh:
@tobiasgrosser suggested in persont not to merge this, and to keep the labels lowering around as an alternative. I would prefer to have one recommended lowering pipeline but I guess there's not much cost to having alternatives. I'd be interested in other reviewers' opinions on this!

compor:
Although I was interested from the start to compare both ways, the comparison never materialized.
Maybe this is a second chance to compare.

This time I agree with @tobiasgrosser but mainly because of the above reason, plus it might indeed be premature to judge the utility (or lack of) and keeping it around doesn't really hurt ATM.

superlopuh:
Ok I'll just merge the updated unit test for riscmu emulation in a new PR

PR #1963, opened 2024-01-09
backend: (snitch_stream) add rank to stride pattern type

superlopuh:
This helps in some lowerings, I'll open the PRs for those later.


compor (approved):
This is simple enough, but I have difficulty with the use case.
Could you please provide a quick example here (even pseudo-code)?
Does this relate to any task/issue?

superlopuh:
It's more relating to lowering from snitch_stream, here's one line where it's used: https://github.com/xdslproject/xdsl/blob/2d2a8318a2dde2c4029b1655dfbab73f14ee718d/xdsl/backend/riscv/lowering/convert_snitch_stream_to_snitch.py#L145

PR #1966, opened 2024-01-11
dialects: (onnx) lint onnx.py

superlopuh:
Two small nits highligthed by VSCode that I missed in my review:

 - FloatAttr and IntegerAttr are generic, and Pylance isn't happy when the generic parameter is not specified. AnyFloatAttr and AnyIntegerAttr is probably what we want if we want to allow any bitwidth in the types. Pylance is the version of Pyright built into VSCode, not sure why it disagrees with the one we have in the CI.
 - There was dead code after the `assert False`, not sure why Pyright didn't bring this up in the CI either, but we should test for it also.

### DIFF
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -5,8 +5,8 @@ from typing import Annotated, cast
 
 from xdsl.dialects.builtin import (
     AnyFloat,
+    AnyIntegerAttr,
     FloatAttr,
-    IntegerAttr,
     IntegerType,
     SSAValue,
     TensorType,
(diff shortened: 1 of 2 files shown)


brnorris03 (approved):
Looks good to me, thank you!

xdsl/dialects/onnx.py
```diff
@@ -124,17 +124,15 @@ def __init__(self, operand: SSAValue):
...
-        if not isinstance(
-            operand_type := self.operand.type, TensorType
-        ) or not isinstance(res_type := self.res.type, TensorType):
-            assert (
-                False
-            ), "onnx elementwise operation operand and result must be of type TensorType"
-            operand_type = cast(TensorType[Attribute], operand_type)
-            res_type = cast(TensorType[Attribute], res_type)
+        assert isinstance(operand_type := self.operand.type, TensorType)
+        assert isinstance(res_type := self.res.type, TensorType)
```
brnorris03, line 128: Is this really necessary, given the way `operand_type` and `res_type` are defined?

kayode-gif (approved):
Changes look fine to me, thanks

xdsl/dialects/onnx.py
superlopuh: you mean the type constraint? We have a few experimental branches that makes this unnecessary but currently SSAValue doesn't have the information about the type of the values it can take at runtime in the type system, so we have to check them every time. We can assert here because the built-in verifier by irdl is called before the custom `verify_` method

PR #1967, opened 2024-01-11
misc: sort passes dict literal alphabetically [NFC]

superlopuh:
Should minimise conflicts in the future


webmiche (approved):
Looks like a good idea to me. Is there some code refactoring tool we could use to do this automatically? Something similar to how isort orders imports :thinking:

superlopuh:
I'd say it's a low enough overhead not to worry about it too much and just fix it up occasionally

PR #1968, opened 2024-01-11
dialects: (onnx) factor out broadcastability helper in onnx dialect

### LINKED ISSUE #1965: factor out broadcastability helper in onnx dialect

kayode-gif:
resolves #1965


superlopuh:
It looks like there are a bunch of unwanted changes, could you please merge main or make a new branch off main?

kayode-gif:
okay, if i make a new branch how can i push it so its on this pr?

kayode-gif:
or would i have to close this pr and then do it again?

superlopuh:
If you'd like we can call and do this over video chat, might be easier

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,54 @@
 from xdsl.utils.exceptions import VerifyException
 
 
+def verify_shapes(
+    lhs_type: TensorType[Attribute] | list[int],
+    rhs_type: TensorType[Attribute] | list[int],
+    res_type: TensorType[Attribute],
+) -> None:
+    """
+    https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md
```
compor, line 35: Thanks for linking the doc, but could you also add a little blob describing what this method does and leave the URL at the bottom of the docstring for anyone who wants to have a further look?

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,54 @@
...
+    """
+
+    # Check that the arguments are broadcastable (using Numpy semantics) and that the result type is correct.
+    res_shape: list[int] = []
```
compor, line 39: Can you move this closer to its first use?

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,54 @@
...
+    if isinstance(lhs_type, TensorType):
+        lhs_shape = lhs_type.get_shape()
+    else:
+        lhs_shape = lhs_type
```
compor, line 43:
```suggestion
    lhs_shape = lhs_type
    if isinstance(lhs_type, TensorType):
        lhs_shape = lhs_type.get_shape()
```

or even see if a ternary operator would make this more readable/fewer lines?

Same below.

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,54 @@
...
+
+    if isinstance(rhs_type, TensorType):
+        rhs_shape = rhs_type.get_shape()
+    else:
+        rhs_shape = rhs_type
+    # Iterate over the shapes in reverse order and compute the result shape.
+    i = max(len(lhs_shape), len(rhs_shape))
+    while i > 0:
```
compor, line 51: Can this become a bit more Pythonic? Not fond of `while` and `continue` statements if they can be avoided.

xdsl/dialects/onnx.py
```diff
@@ -227,41 +251,19 @@ def verify_(self) -> None:
...
 ONNX = Dialect(
     "onnx",
     [
         Add,
-        Sub,
-        Mul,
         Div,
-        Relu,
         Gemm,
+        Mul,
```
compor, line 265: This seems to be an unrelated change.

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,54 @@
 from xdsl.utils.exceptions import VerifyException
 
 
+def verify_shapes(
+    lhs_type: TensorType[Attribute] | list[int],
+    rhs_type: TensorType[Attribute] | list[int],
+    res_type: TensorType[Attribute],
+) -> None:
```
superlopuh, line 33: I would recommend the following refactor:
```suggestion
def broadcast_shape(
    lhs: Sequence[int],
    rhs: Sequence[int],
) -> list[int] | None:
```
This will make it easier to test, it would be great to add some python unit tests for this helper in `tests/dialects/onnx/test_broadcast.py`

I would then recommend the following approach:
1. figure out which list is longer than the other
2. the prefix of the result is the prefix of the longer list
3. zip the shorter list with the suffix of the longer list
4. if the dimension is the same, then that's the dimension in the result, if one of them is 0, then take the other one, if they're not equal return `None`

The client of this function should extract the shapes from the types, check for `None`, and compare against the result type itself. Although maybe we may want a second helper to factor out that boilerplate too, but that helper itself would need no unit tests specifically.

compor (changes requested):
Minor things, if you want to refactor in a separate PR, I'm fine with it.

Maybe this is a good opportunity to add a Pytest for the factored-out method?
I haven't checked how well-covered this is ATM.

Regarding unwanted changes, let me know if you need help I know @superlopuh offered assistance, but ping me as well if you're around.

superlopuh (changes requested):
Thank you for this PR! The logic looks good, I'd recommend a refactor for testability

superlopuh:
Now that I think about it, let's add both the unidirectional and multidirectional shape computation in this PR? https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md

xdsl/dialects/onnx.py
compor: Pull from `main` should resolve this now that #1969 is merged.

xdsl/dialects/onnx.py
superlopuh: Helpfully, the docs from onnx provide a minimal set of test cases for the two broadcast shape calculation helpers

xdsl/dialects/onnx.py
compor: see @superlopuh's suggestion below

xdsl/dialects/onnx.py
kayode-gif: i've attempted refactoring again, however here are my concerns:
1. i can't seem to figure out why onnx_invalid_mlir won't pass, the other one works fine
2. need help setting up pytest, i've checked the test cases in another editor and they all pass fine so thats not an issue, (i was getting the error that it cant recognise the tests)
3.  multidirectional broadcasting and unidirectional broadcasting seem to be the same thing to me? i ran the test cases for unidirectional broadcasting in the multidirectional broadcasting and the outputs work, so im not sure why there is a need for another computation.

I will paste my code here, if i am doing anything stupid or been missing something stupid, id appreciate the spot:

This method extracts the types
`def extract_shape_from_type(shape_type: [Union[Sequence[int], TensorType]]) -> Sequence[int] | None:
    if shape_type is None:
        return None
    elif isinstance(shape_type, TensorType):
        return shape_type.get_shape()
    else:
        return shape_type
`
This is the redone broadcast method (not sure if this is exacly how you wanted it done?):
`def broadcast_shape(lhs: Sequence[int], rhs: Sequence[int]) -> list[int] | None:
    """
    https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md
    """
    lhs_shape = list(extract_shape_from_type(lhs))
    rhs_shape = list(extract_shape_from_type(rhs))

    # figure out which list is longer than the other
    if len(lhs_shape) > len(rhs_shape):
        longer_shape, shorter_shape = lhs_shape, rhs_shape
    else:
        longer_shape, shorter_shape = rhs_shape, lhs_shape
    # store the difference of shape dimensions between the shapes
    shape_dimension_diffs = len(longer_shape) - len(shorter_shape)
    # checking which side needs to be padded to the longer side
    if len(lhs_shape) != rhs_shape:
        shorter_shape = [1] * shape_dimension_diffs + list(shorter_shape)
    # Iterate over the shape dimensions and verify the broadcasting rules
    res_shape: list[int] = []
    for d1, d2 in zip(longer_shape, shorter_shape):
        if d1 == d2:
            res_shape.append(d1)
        elif d1 == 1:
            res_shape.append(d2)
        elif d2 == 1:
            res_shape.append(d1)
        else:
            raise VerifyException(
                f"operands have incompatible shapes: {shorter_shape} and {longer_shape}"
            )
    return res_shape`
(I was thinking to add the other check of result shape in the `verify_` of the opps,
All else i do is call the `broadcast_shape` method in class  method

I appreciate the feedback and help, @compor @superlopuh

superlopuh: 1. What happens when you run `make tests`?
2. I think I may have given you bad advice, or at least I got a better idea, it's probably best to copy the longer sequence into a new list, and then swap out the 1s with the other element or return None in a for loop.
3. It definitely seems to me like the two broadcast functions are different, Multidirectional seems symmetric, whereas Unidirectional is not, but it also seems like one can call the other after it determines which shape is shorter.

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,82 @@
...
+
+    https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md
+    """
+    # Check if Tensor A and B both have exactly the same shape
+    if lhs == rhs:
+        return lhs
+
+    # Check if Tensor A and B have the same number of dimensions and the length of each dimensions is either a common
+    # length or B's length is 1.
+    res_shape: list[int] = []
```
superlopuh, line 51:
```suggestion
    lhs_len = len(lhs)
    rhs_len = len(rhs)
    prefix_len = lhs_len - rhs_len
    if prefix_len < 0:
        # lhs must not be shorter than rhs
        return None

    res_shape = lhs[:prefix_len]
    
    for dl, dr in zip(lhs[prefix_len:], rhs):
        ...
```

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,82 @@
...
+    # Checking shape broadcasting compatibility
+    for d1, d2 in zip(longer_shape, shorter_shape):
+        if d1 == d2 or d1 == 1 or d2 == 1:
+            res_shape.append(max(d1, d2))
+        else:
+            raise VerifyException(
+                f"operands have incompatible shapes: {tuple(shorter_shape)} and {tuple(longer_shape)}"
+            )
+    return res_shape
```
superlopuh, line 102: Does this work?
```suggestion
    if len(lhs) > len(rhs):
        return unidirectional_broadcast_shape(rhs, lhs)
    else:
        return unidirectional_broadcast_shape(lhs, rhs)
```

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,82 @@
 from xdsl.utils.exceptions import VerifyException
 
 
+def unidirectional_broadcast_shape(lhs: list[int], rhs: list[int]) -> list[int]:
```
superlopuh, line 29:
```suggestion
def unidirectional_broadcast_shape(lhs: list[int], rhs: list[int]) -> list[int] | None:
```
It feels like mixing this "business" function with the verification logic is a little confusing, the operation itself can raise the exception if the result is None

xdsl/dialects/onnx.py
kayode-gif: this did  work well, thanks

superlopuh (changes requested):
Looking great! Could you please add some pytests for failing cases? Also, it feels like we might want to add two more helpers that take tensor types and return tensor types, that internally call the functions you've added that operate on Python lists

kayode-gif:
> Looking great! Could you please add some pytests for failing cases? Also, it feels like we might want to add two more helpers that take tensor types and return tensor types, that internally call the functions you've added that operate on Python lists

okay ive added the tests, i am however confused why do we need those tensor type helpers for the broadcast functions, what is the use?.

superlopuh:
It makes it a little easier to have consistent error messages in the ops, and removes a little bit of boilerplate code. It's not really a big deal, but seems like we'll make use of it

tests/dialects/onnx/test_broadcast.py
```diff
@@ -0,0 +1,76 @@
...
+    (([2, 3, 4, 5], [1, 3, 1, 5]), [2, 3, 4, 5]),
+]
+
+m_fail_test_cases: list[tuple[tuple[list[int], list[int]], list[int]]] = [
+    (([2, 3, 4, 5], []), []),
+    (([2, 3, 4, 5], [5]), [2, 3, 4, 5]),
+    (([4, 5], [2, 3, 4, 5]), [2, 3, 4, 5]),
+    (([1, 4, 5], [2, 3, 1, 5]), [2, 4, 5, 5]),
+    (([3, 4, 5], [2, 1, 1, 1]), [3, 3, 4, 5]),
+]
```
superlopuh, line 29: Could you please add these inputs to the test cases above with `None`? It's pretty much always better to unit test for expected result, as opposed to `!=`

xdsl/dialects/onnx.py
```diff
@@ -201,6 +239,7 @@ def verify_(self) -> None:
         # check shape compatibility
         tensor_a_shape = tensor_a_type.get_shape()
         tensor_b_shape = tensor_b_type.get_shape()
+        tensor_c_shape = list(tensor_c_type.get_shape())
```
superlopuh, line 242: why list?

kayode-gif: i had this as list, so it match the type of the broadcast function (as its an input for it), but this is now changed

tests/dialects/onnx/test_broadcast.py
```diff
@@ -5,30 +5,46 @@
...
 )
 
-m_test_cases: list[tuple[tuple[list[int], list[int]], list[int]]] = [
-    (([2, 3, 4, 5], []), [2, 3, 4, 5]),
+m_test_cases: list[
+    tuple[
+        tuple[list[int | None] | None, list[int | None] | None], list[int | None] | None
+    ]
+] = [
+    (([2, 3, 4, 5], [None]), [2, 3, 4, 5]),
```
superlopuh, line 13: I'm a little confused, why the `[None]`?

superlopuh: I meant to add the failure cases with `None` as the result below the successful cases

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,104 @@
...
+def tensor_unidirectional_broadcast_shape(
+    lhs: TensorType[Attribute], rhs: TensorType[Attribute], res: TensorType[Attribute]
+) -> TensorType[Attribute]:
+    """
+    Returns the tensor format of an unidirectional broadcastable shape
+    """
+    lhs_shape = lhs.get_shape()
+    rhs_shape = rhs.get_shape()
+    res_shape = unidirectional_broadcast_shape(list(lhs_shape), list(rhs_shape))
+    res_shape_tensor = cast(TensorType[Attribute], res_shape)
```
superlopuh, line 38: Doesn't this always fail? `res_shape` should be `None` or `list[int]`, and never a tensor type

kayode-gif: okay, i delete this and change the return type of the function to `list[int] | None`, but the return type wont be a tensor then right

superlopuh: That seems like the right move to me, and then we could use terms like `expected_shape` for whatever is returned from the helper function, and `res_shape` for the shape of the `res` tensor passed in

tests/filecheck/dialects/onnx/onnx_invalid.mlir
superlopuh, line 1: Ideally this PR wouldn't change any test files

kayode-gif: sorry about that change, i think that test before wasn't accurate review

superlopuh: TBH neither the old nor the new case make perfect sense to me, but the old one does a little more than the new, or at least they seem to be testing different kinds of errors. The first one checks for incompatible shapes, the second one checks for shapes that are compatible but are not exactly correct. I'd say let's dive deeper into this in another PR, and try to keep the tests as they are for the time being, if possible

kayode-gif: yup, ill change it back to the original

tests/dialects/onnx/test_broadcast.py
```diff
@@ -5,50 +5,34 @@
...
-    ]
-] = [
-    (([2, 3, 4, 5], [None]), [2, 3, 4, 5]),
+m_test_cases: list[tuple[tuple[list[int], list[int]], list[int]]] = [
+    (([2, 3, 4, 5], []), [2, 3, 4, 5]),
     (([2, 3, 4, 5], [5]), [2, 3, 4, 5]),
     (([4, 5], [2, 3, 4, 5]), [2, 3, 4, 5]),
     (([1, 4, 5], [2, 3, 1, 5]), [2, 3, 4, 5]),
     (([3, 4, 5], [2, 1, 1, 1]), [2, 3, 4, 5]),
 ]
```
superlopuh:
```suggestion
m_test_cases: list[tuple[tuple[list[int], list[int]], list[int] | None]] = [
    (([2, 3, 4, 5], []), [2, 3, 4, 5]),
    (([2, 3, 4, 5], [5]), [2, 3, 4, 5]),
    (([4, 5], [2, 3, 4, 5]), [2, 3, 4, 5]),
    (([1, 4, 5], [2, 3, 1, 5]), [2, 3, 4, 5]),
    (([3, 4, 5], [2, 1, 1, 1]), [2, 3, 4, 5]),
    (([2, 3, 4, 5], [6, 1]), None),
    (([1, 3, 4, 1], [5]), None),
    (([4, 5, 6], [2, 3, 4, 5]), None),
    (([1, 4, 5], [2, 3, 1, 5]), None),
    (([3, 4, 5], [2, 1, 1, 1]), None),
]
```

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,102 @@
 from xdsl.utils.exceptions import VerifyException
 
 
+def tensor_unidirectional_broadcast_shape(
+    lhs: TensorType[Attribute], rhs: TensorType[Attribute], res: TensorType[Attribute]
+) -> list[int] | None:
```
superlopuh, line 31:
```suggestion
def verify_unidirectional_broadcast_shape(
    lhs: TensorType[Attribute], rhs: TensorType[Attribute], res: TensorType[Attribute]
) -> None:
```

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,102 @@
...
+    """
+    Returns a unidirectional broadcastable shape
+    """
+    lhs_shape = lhs.get_shape()
+    rhs_shape = rhs.get_shape()
+    res_shape = unidirectional_broadcast_shape(list(lhs_shape), list(rhs_shape))
```
superlopuh, line 37:
```suggestion
    expected_shape = unidirectional_broadcast_shape(list(lhs_shape), list(rhs_shape))
```

tests/dialects/onnx/test_broadcast.py
```diff
@@ -0,0 +1,50 @@
...
+
+u_test_cases: list[tuple[tuple[list[int], list[int]], list[int | None]]] = [
+    (([2, 3, 4, 5], []), [2, 3, 4, 5]),
+    (([2, 3, 4, 5], [5]), [2, 3, 4, 5]),
+    (([2, 3, 4, 5], [2, 1, 1, 5]), [2, 3, 4, 5]),
+    (([2, 3, 4, 5], [1, 3, 1, 5]), [2, 3, 4, 5]),
+    (([], [2, 3, 4, 5]), [None]),
+    (([5], [2, 3, 4, 5]), [None]),
+    (([2, 1, 1, 5], [2, 3, 4, 5]), [None]),
+    (([2, 3, 5], [1, 3, 1, 5]), [None]),
```
superlopuh, line 29: I'm surprised this works with the [None] in brackets

superlopuh: Is it pytest magic list equality?

superlopuh: Thank you

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,100 @@
...
+def verify_unidirectional_broadcast_shape(
+    lhs: TensorType[Attribute], rhs: TensorType[Attribute], res: TensorType[Attribute]
+) -> None:
+    """
+    Returns a unidirectional broadcastable shape
+    """
+    lhs_shape = lhs.get_shape()
+    rhs_shape = rhs.get_shape()
+    expected_shape = unidirectional_broadcast_shape(list(lhs_shape), list(rhs_shape))
+    if not expected_shape:
```
superlopuh, line 38:
```suggestion
    if expected_shape is not None:
```

xdsl/dialects/onnx.py
```diff
@@ -26,6 +26,100 @@
...
+        raise VerifyException(
+            f"operands have incompatible shapes: {lhs_shape} and {rhs_shape}"
+        )
+    res_type_shape = list(res.get_shape())
```
superlopuh, line 42: what happens when you make this change? I'm surprised you need `list` here and not above
```suggestion
    res_type_shape = res.get_shape()
```

superlopuh (approved):
Looking great!

superlopuh:
CC @compor

xdsl/dialects/onnx.py
kayode-gif: you mean is `None` right? because that would fail tests?

xdsl/dialects/onnx.py
kayode-gif: erm i believe since in the tests since its a list '[]`, it would fail, because res.get_shape() is a `tuple[int]' , thats what i believe it returns when i looked at the shaped class in the builtin

kayode-gif: correction: i mean when i looked at the tensortype class, that .get_shape() is a `tuple[int]` so i need the `list` so it matches

superlopuh: Ah, I see, in that case I'd convert the list to a tuple in the check on the line below, but it's a nit at this point

compor (approved):

Issue #1971, opened 2024-01-11
xDSL can not parse a DenseAttr with Hex values


JosseVanDelm:
Hi!

I'm trying to compile some of the MLPerf Tiny benchmarks.
During this, at some point in compilation, the following IR is generated for the weight values, which tend to be quite a bit larger than what I'm mentioning here:
```mlir
%0 = "arith.constant"() <{value = dense<"0x82F5AB00"> : tensor<1xi32>}> : () -> tensor<1xi32>
```
`mlir-opt` is able to parse this just fine:
```mlir
module {
  %cst = arith.constant dense<11269506> : tensor<1xi32>
}
```
But `xdsl-opt` is a bit confused :cry: 
<details>

```
Traceback (most recent call last):
  File "/home/josse/phd/xdsl/xdsl/tools/command_line_tool.py", line 569, in parse_chunk
    return self.available_frontends[file_extension](chunk)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josse/phd/xdsl/xdsl/tools/command_line_tool.py", line 557, in parse_mlir
    ).parse_module(not self.args.no_implicit_module)
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josse/phd/xdsl/xdsl/parser/core.py", line 127, in parse_module
    if (parsed_op := self.parse_optional_operation()) is None:
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josse/phd/xdsl/xdsl/parser/core.py", line 658, in parse_optional_operation
    return self.parse_operation()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josse/phd/xdsl/xdsl/parser/core.py", line 695, in parse_operation
    op = self._parse_generic_operation(op_type)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josse/phd/xdsl/xdsl/parser/core.py", line 837, in _parse_generic_operation
    properties = self.parse_optional_properties_dict()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/josse/phd/xdsl/xdsl/parser/core.py", line 777, in parse_optional_properties_dict
    entries = self.parse_comma_separated_list(
[...]

PR #1974, opened 2024-01-12
core: Support hex denseAttr parsing for integers, f32, f64

JosseVanDelm:
Attempt at solving #1971

This was a bit more tricky than I initially expected due to endianess issues.

Some things I noticed going back and forth between `xdsl-opt` and `mlir-opt`:
* `mlir-opt` has the tendency to print signless values as signed, whereas xdsl seems to print them as unsigned.
   here i'm adding the integer values by their signed value, is that okay?
* `mlir-opt` prints floats slightly different from xdsl, i made sure the values are same-ish manually

Currently still draft:
- [x] improve negative tests
- [x] improve shape testing/verifying?
- [x] Can probably be made a bit prettier, suggetions welcome!

I'm leaving other float types for future work, these seem not possible to be unpacked with python's `struct` package.


xdsl/parser/attribute_parser.py
```diff
@@ -653,10 +655,17 @@ def _parse_builtin_dense_attr(self, _name: Span) -> DenseIntOrFPElementsAttr:
...
             values, shape = [], None
         else:
-            values, shape = self._parse_tensor_literal()
+            # Parse hex string e.g. dense<"0x82F5AB00">
+            hex_string = self.parse_optional_str_literal()
+            if hex_string is not None:
+                values, shape = [], None
+            else:
+                # Expect a tensor literal instead
+                values, shape = self._parse_tensor_literal()
```
superlopuh, line 668: feels like we would benefit from a dedicated helper to parse binary data encoded in a string literal, would make it easier to test, also

xdsl/parser/attribute_parser.py
```diff
@@ -643,20 +645,98 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
+            for i in range(num_chunks):
+                parsed_int = int.from_bytes(
+                    byte_list[i * chunk_size : (i + 1) * chunk_size],
+                    sys.byteorder,
+                    signed=True,
+                )
+                data_values.append(parsed_int)
+        # Splat attribute case, same value everywhere
+        if chunk_size == len(byte_list) and type.get_shape() != (1,):
+            return data_values, []
```
JosseVanDelm, line 716: @math-fehr @superlopuh @compor this might not be a sufficient check here. Any suggestions?

xdsl/parser/attribute_parser.py
```diff
@@ -643,20 +645,98 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
             values, shape = [], None
         else:
-            values, shape = self._parse_tensor_literal()
+            hex_string = self.parse_optional_str_literal()
+            if hex_string is not None:
+                # Can not determine values without type yet
+                values, shape = [], None
+            else:
+                # Expect a tensor literal instead
+                values, shape = self._parse_tensor_literal()
```
compor, line 739: I have trouble understanding what is expected to be at each line here.

Can this be simplified for the case where `values, shape = [], None`?

Can this be combined with the `_parse_builtin_dense_attr_hex` below?
You seem to assume this is a hex, but you don't really know that yet, wouldn't determining this as fast as possible make sense and potentially elide the extra check `if hex_string is not None:` below?

Ideally, I'd like to come out of the block knowing that if hex exists is valid.

xdsl/parser/attribute_parser.py
```diff
@@ -643,20 +645,98 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
+                self.raise_error(
+                    "Hex strings for dense literals are only supported for int, f32 and f64 types"
+                )
+        num_chunks = len(byte_list) // chunk_size
+
+        data_values: list[int] | list[float] = []
+
+        if isa(element_type, Float32Type | Float64Type):
+            for i in range(num_chunks):
+                parsed_float = struct.unpack(
```
superlopuh, line 701: Should we use `unpack_from` or `iter_unpack`? feels like we could avoid the iteration in Python, and instead check that the number was what we expected

xdsl/parser/attribute_parser.py
superlopuh, line 739: Why not put the str literal logic inside the `_parse_tensor_literal` code? Do you know how MLIR structures this parsing logic?

xdsl/parser/attribute_parser.py
```diff
@@ -643,20 +645,98 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
+    ) -> tuple[list[int] | list[float], list[int]]:
+        """
+        Parse a hex string literal e.g. dense<"0x82F5AB00">, and returns its flattened data
+        and it's flattened shape, based on the parsed type.
+        Parse a hex string tensor literal, and returns its flatten data and its shape.
+
+        For instance, a dense<"0x82F5AB0182F5AB00"> attribute will return [28046722, 11269506]
+        for a tensor<2xi32> type.
+
+        only supports integer types that are multiple of 8, f32 and f64
```
compor, line 663:
```suggestion
        Parse a hex string literal e.g. dense<"0x82F5AB00">, and returns its flattened data
        and its flattened shape, based on the parsed type.
        Parse a hex string tensor literal, and return its flattened data and its shape.

        For instance, a dense<"0x82F5AB0182F5AB00"> attribute will return [28046722, 11269506]
        for a tensor<2xi32> type.

        Only supports integer types that are multiple of 8, f32 and f64.
```

typo nits

xdsl/parser/attribute_parser.py
JosseVanDelm: I haven't looked at the parser code of MLIR tbh, I just parsed some denseAttrs in compiler explorer and took a look at what was produced. The MLIR codebase scares me a bit haha, but it seems that they put it inside the TensorLiteralParser?
https://github.com/llvm/llvm-project/blob/d85df3f2d6e8687c44e6802dcc0e59c14ff32c9b/mlir/lib/AsmParser/AttributeParser.cpp#L456

xdsl/parser/attribute_parser.py
JosseVanDelm: Changed to using `unpack_from`, thanks for the suggestion!

xdsl/parser/attribute_parser.py
JosseVanDelm: Hi @compor I agree, that this is a bit of an ugly way of going about this.

I think it looks ugly because of the following problems I had:

* The problem is that I can not really extract the shape information out of the hex_string since the information is already flattened inside the hex attribute, so there are a few checks you can not perform on hex values.
* The added code kind of breaks the previous flow because you already need your type info to be able to parse the hex string in the first place. In the previous code this was not a problem, because you can already parse the ints and the floats purely on the format of the numbers
* another problem is that I was afraid of emitting `_TensorLiteralElement`s because they seem to require a span, and I was unsure on how this works. @superlopuh would this solve your question as well?

Does this make sense? Thanks for your help!

JosseVanDelm: To answer your actual question @compor:

Originally:
`shape = []` for splat attribute, list of ints for bigger attributes, e.g. [[2,3],[3,2]] would give a `shape = [2,2] `
`values` = a list of `_TensorLiteralElement`s which contains parsed `TensorLiteral` elements

Now:
`hex_string` The value of the hex string, if present, `None` if not present
`shape = []` for splat attribute, list of ints for bigger attributes, e.g. [[2,3],[3,2]] would give a `shape = [2,2] `.
For hex attributes I can only give the flattened shape as a single-element list (e.g. `shape = [4]`) because of a lack of information.
`values` = a list of `_TensorLiteralElement`s which contains parsed `TensorLiteral` elements
In the case of a hex attribute, the `values` are already in their python counterparts, so values is already of type `list[int]` or `list[float]` and doesn't need to be converted from TensorLiteralElement anymore.

I agree that this is a lot of info necessary for just 3 variables :sweat_smile: , any suggestions very much welcome!

JosseVanDelm:
Okay it seems there's some further work necessary beyond your comments.
I got this error on a 2D shape:
```
Shape mismatch in dense literal. Expected [640, 128] shape from the type, but got [81920] shape.
```

xdsl/parser/attribute_parser.py
superlopuh: I had a similar issue with this BTW, when I tried to implement `memref.global` custom syntax: https://xdsl.zulipchat.com/#narrow/stream/368604-Framework/topic/printing.20and.20parsing.20attributes/near/410844828

xdsl/parser/attribute_parser.py
math-fehr: `if len(data_values) == 1` should be a check that is enough?
Any `1x...` dense attribute is a splat anyway

xdsl/parser/attribute_parser.py
math-fehr: We could just put the entire span for the string, so we don't have `list[int]` or `list[float]` and have the list of literal elements, I would be fine with that!
We won't use these errors anyway, so there is no problems.

math-fehr: To simplify the logic actually, can we just split the following code in two functions? One function to deal with the hex, and the other to deal with the lists?
Because currently the logic gets quite complex

math-fehr (approved):
Looks like it's going to the right direction, the only problem is the code getting complex now

xdsl/parser/attribute_parser.py
JosseVanDelm: Thanks @math-fehr I'll try to refactor to using the literals now and I'll see what I'm left with :smile:

xdsl/parser/attribute_parser.py
```diff
@@ -701,15 +778,33 @@ def _parse_builtin_dense_attr(self, _name: Span) -> DenseIntOrFPElementsAttr:
                 "Expected vector or tensor type of " "integer, index, or float type"
             )
 
+        # Get values and shape in case of hex_string (requires parsed type)
+        if hex_string is not None:
+            values, shape = self._parse_builtin_dense_attr_hex(hex_string, type)
```
superlopuh, line 783: I might be missing something, but what changes if this line is inlined above, and the `hex_string` doesn't escape the `else:` branch above? I have a feeling that this method would be easier to understand with an `is_hex_string` `bool` value instead of the string, just to communicate that it's not used for anything once its parsed

xdsl/parser/attribute_parser.py
```diff
@@ -718,13 +813,21 @@ def _parse_builtin_dense_attr(self, _name: Span) -> DenseIntOrFPElementsAttr:
             self.raise_error("Dense literal attribute should have a static shape.")
 
         element_type = type.element_type
+        data_values: list[int] | list[float] = []
         # Convert list of elements to a list of values.
-        if shape != []:
-            data_values = [value.to_type(self, element_type) for value in values]
+        # For hex string, the data values are already constructed
```
superlopuh, line 818: Is this true? Isn't one of your examples a `2xi32` array where there's a single repeated value decoded?

xdsl/parser/attribute_parser.py
JosseVanDelm: On second thought @math-fehr it doesn't make a lot of sense to make the _TensorLiteralElements, because then you have to convert them to a list of ints going out of the function anyways :(

xdsl/parser/attribute_parser.py
```diff
@@ -670,61 +672,174 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
-
-        # Check that the shape matches the data when given a shaped data.
+        type = _parse_dense_type()
         type_shape = list(type.get_shape())
-        num_values = math.prod(type_shape)
-
-        if shape is None and num_values != 0:
-            self.raise_error(
-                "Expected at least one element in the " "dense literal, but got None"
-            )
```
JosseVanDelm, line 711: upon deeper inspection, this check makes sense (if you encounter a `dense<>` there can not be any shape in your type (when does this happen??) ), but the message really doesn't.
Should I change it?

xdsl/parser/attribute_parser.py
```diff
@@ -670,61 +672,174 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
+                    f"Shape mismatch in dense literal. Expected {type_num_values} "
+                    f"elements from the type, but got {shape[0]} elements."
+                )
+
+        def _check_empty_dense_attr(type_num_values: int):
+            if type_num_values != 0:
+                self.raise_error(
+                    "Expected at least one element in the "
+                    "dense literal, but got None"
+                )
```
superlopuh, line 796: What's the purpose of these helpers? I see that they factor logic but also obscure the flow a little, and make the diff slightly messier. Do you think they'll be called more than once in the future? My first instinct is to inline them again.

math-fehr: I disagree though. The code before was really long and that makes things harder to read.
With these functions, it's easier to understand why this code works tbh.
The thing I would change though is maybe make a big if/else in the code for the hex case, since we can use the `parse_dense_type` can be reused

math-fehr (commented):
To me, that definitely looks better!

xdsl/parser/attribute_parser.py
```diff
@@ -670,61 +672,174 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
...
+        """
+        Parse a hex string literal e.g. dense<"0x82F5AB00">, and returns its flattened data
+        and its flattened shape, based on the parsed type.
+        Parse a hex string tensor literal, and return its flattened data and its shape.
+
+        For instance, a dense<"0x82F5AB0182F5AB00"> attribute will return [28046722, 11269506]
+        for a tensor<2xi32> type.
+
+        Only supports integer types that are multiple of 8, f32 and f64.
+        """
```
superlopuh, line 691: The first two sentenses seem to be saying the same thing

xdsl/parser/attribute_parser.py
superlopuh: But that's because the logic of the function is too convoluted. @JosseVanDelm, do you mind if I try to refactor it a little bit? I think it could benefit from an early exit in the empty case, and from encapsulating the hex parsing logic to a single section of code.

superlopuh: I played with this locally, and it looks like my initial intuition about early exit was wrong, but I found another way to refactor this little function that I feel is beneficial, not sure how best to share the snippet for your consideration.

JosseVanDelm: @superlopuh let me try to add @math-fehr's suggestion first

JosseVanDelm:
@superlopuh I think the last 4 commits fully unspaghetti'd it (I think that is what @math-fehr was suggesting?)
What do you think?

superlopuh:
Definitely looks much neater, thank you!

superlopuh:
Here's what I came to after a bit of fiddling, it's not exactly what I recommended before, but also seems like a direction to look into: `sasha/add-hex-parser`

superlopuh:
pushed! @math-fehr, can you please review my latest patch?

superlopuh:
Got the ok from Mathieu, thank you very much @JosseVanDelm !

PR #1976, opened 2024-01-15
dialects: Add optional alignment property to memref.global

JosseVanDelm:
Hi!

I was trying to parse a compiled neural network from tensorflow when I hit a memref.global op that did not have an alignment property/attribute (?) defined.
```
  ^^^^^^^^^^^^^^^-----------------------------------------------------------------------------------------------------------------------------------------------
  | Operation does not verify: property 'alignment' is not defined by the operation. Use the dictionary attribute to add arbitrary information to the operation.
  --------------------------------------------------------------------------------------------------------------------------------------------------------------
```
According to:
https://mlir.llvm.org/docs/Dialects/MemRef/#memrefglobal-memrefglobalop

This is supposed to be a i64 value that has to be a power of two.

### DIFF
--- a/tests/dialects/test_memref.py
+++ b/tests/dialects/test_memref.py
@@ -4,11 +4,13 @@ from xdsl.builder import Builder
 from xdsl.dialects import arith, builtin, func, memref, scf
 from xdsl.dialects.builtin import (
     ArrayAttr,
+    DenseIntOrFPElementsAttr,
     FloatAttr,
     IndexType,
     IntAttr,
     IntegerType,
     StridedLayoutAttr,
+    StringAttr,
     i32,
     i64,
 )
@@ -138,6 +140,24 @@ def test_memref_alloc():
         match="op dimension operand count does not equal memref dynamic dimension count",
     ):
         alloc6.verify()
+
+
+def test_memref_global():
+    memref_type = memref.MemRefType(i32, (2, 2))
+    tensor_type = builtin.TensorType(i32, (2, 2))
+    alignment = 65
+    global1 = memref.Global.get(
+        StringAttr("my_global"),
+        memref_type,
+        DenseIntOrFPElementsAttr.from_list(tensor_type, [1, 2, 3, 4]),
+        sym_visibility=StringAttr("public"),
+        alignment=builtin.IntegerAttr(alignment, 64),
+    )
+    with pytest.raises(
+        VerifyException,
+        match=f"Alignment attribute {alignment} is not a power of 2",
+    ):
+        global1.verify()
 
 
 def test_memref_alloca():
--- a/tests/dialects/test_memref.py
+++ b/tests/dialects/test_memref.py
@@ -153,11 +153,19 @@ def test_memref_global():
         sym_visibility=StringAttr("public"),
         alignment=builtin.IntegerAttr(alignment, 64),
     )
+    global2 = memref.Global.get(
+        StringAttr("my_valid_global"),
+        memref_type,
+        DenseIntOrFPElementsAttr.from_list(tensor_type, [1, 2, 3, 4]),
+        sym_visibility=StringAttr("public"),
+        alignment=64,
+    )
     with pytest.raises(
         VerifyException,
         match=f"Alignment attribute {alignment} is not a power of 2",
     ):
         global1.verify()
+    global2.verify()
 
 
 def test_memref_alloca():
--- a/tests/dialects/test_memref.py
+++ b/tests/dialects/test_memref.py
@@ -4,13 +4,11 @@ from xdsl.builder import Builder
 from xdsl.dialects import arith, builtin, func, memref, scf
 from xdsl.dialects.builtin import (
     ArrayAttr,
-    DenseIntOrFPElementsAttr,
     FloatAttr,
     IndexType,
     IntAttr,
     IntegerType,
     StridedLayoutAttr,
-    StringAttr,
     i32,
     i64,
 )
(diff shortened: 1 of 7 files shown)


tests/dialects/test_memref.py
```diff
@@ -140,6 +142,32 @@ def test_memref_alloc():
...
+        DenseIntOrFPElementsAttr.from_list(tensor_type, [1, 2, 3, 4]),
+        sym_visibility=StringAttr("public"),
+        alignment=64,
+    )
+    with pytest.raises(
+        VerifyException,
+        match=f"Alignment attribute {alignment} is not a power of 2",
+    ):
+        global1.verify()
+    global2.verify()
```
superlopuh, line 168: Can you please make the failing version of this a failing filecheck test instead? And I think that the memref_ops filecheck already convers the non-failing case, so we should be ok

xdsl/dialects/memref.py
```diff
@@ -484,20 +485,33 @@ def verify_(self) -> None:
                 "Global initial value is expected to be a "
                 "dense type or an unit attribute"
             )
+        if self.alignment is not None:
+            assert isa(self.alignment, IntegerAttr)
```
superlopuh, line 489: Do we need the isa here? can we use isinstance?

xdsl/dialects/memref.py
```diff
@@ -484,20 +485,33 @@ def verify_(self) -> None:
...
+            test = self.alignment.value.data
+            # test if is power of 2 with binary magic
+            if not ((test & (test - 1) == 0) and test != 0):
```
superlopuh, line 492: Would you mind factoring this out into its own function called `is_power_of_two` or something? Also, I have a feeling that testing for `test.bit_count() == 1` might be marginally better but it's definitely a matter of style.

xdsl/dialects/memref.py
```diff
@@ -484,20 +485,33 @@ def verify_(self) -> None:
...
+                    f"Alignment attribute {self.alignment.value.data} is not a power of 2"
+                )
 
     @staticmethod
     def get(
         sym_name: StringAttr,
         sym_type: Attribute,
         initial_value: Attribute,
         sym_visibility: StringAttr = StringAttr("private"),
+        alignment: int | AnyIntegerAttr | None = None,
```
superlopuh, line 503: If it has to be i64 then we should probably use `IntegerAttr[IntegerType]`

xdsl/dialects/memref.py
```diff
@@ -472,6 +472,7 @@ class Global(IRDLOperation):
     sym_visibility: StringAttr = prop_def(StringAttr)
     type: Attribute = prop_def(Attribute)
     initial_value: Attribute = prop_def(Attribute)
+    alignment: AnyIntegerAttr | None = opt_prop_def(AnyIntegerAttr)
```
superlopuh, line 475:
```suggestion
    alignment = opt_prop_def(Annotated[IntegerAttr[IntegerType], i64])
```
Does this work by any chance?

tests/dialects/test_memref.py
JosseVanDelm: Sure, why a filecheck instead of a pytest test?

xdsl/dialects/memref.py
JosseVanDelm: seems to work!

xdsl/dialects/memref.py
JosseVanDelm: yup!

xdsl/dialects/memref.py
JosseVanDelm: I like your proposal better!

xdsl/dialects/memref.py
JosseVanDelm: That makes more sense! thanks!

tests/dialects/test_memref.py
superlopuh: It's just the thing we've been going for in the project. I usually prefer pytest tests but it's probably worth doing for consistency. It also couples the test less with the Python API.

xdsl/utils/bitwise_casts.py
```diff
@@ -22,3 +22,11 @@ def convert_u32_to_f32(value: int) -> float:
...
     raw_float = ctypes.c_float.from_address(ctypes.addressof(raw_int)).value
     return raw_float
+
+
+def is_power_of_two(value: int) -> bool:
+    """
+    Return True if an integer is a power of two.
+    Powers of two have only one bit set to one
+    """
+    return (value.bit_count() - 1 == 0) and value != 0
```
superlopuh, line 32: Could you please add some pytest tests for this? I'd just do `-2 to 5` as a range of tests.

Also, doesn't this give the same result?
```suggestion
def is_power_of_two(value: int) -> bool:
    """
    Return True if an integer is a power of two.
    Powers of two have only one bit set to one
    """
    return (value > 0) and (value.bit_count() == 1)
```

JosseVanDelm: That looks even prettier indeed haha

JosseVanDelm: will do

PapyChacal (approved):
Thanks!
Adding interoperability test for those case would be even better, but not a strong requirement for me
But it helps catching this kind of thing quicker when we move the targetted MLIR version :)

tests/utils/test_bitwise_casts.py
```diff
@@ -18,3 +22,9 @@
 def test_float_bitwise_casts(i: int, f: float):
     assert convert_f32_to_u32(f) == i
     assert struct.pack(">f", convert_u32_to_f32(i)) == struct.pack(">f", f)
+
+
+def test_is_power_of_two():
+    powers = [is_power_of_two(i) for i in range(-2, 6)]
+    # [-2, -1, 0, 1, 2, 3, 4, 5]
+    assert powers == [False, False, False, True, True, False, True, False]
```
superlopuh, line 30:
```suggestion
@pytest.mark.parametrize(
    "i, p",
    [
        (-2, False),
        (-1, False),
        (0, False),
        (1, True),
        (2, True),
        (3, False),
        (4, True),
        (5, False),
    ],
)
def test_is_power_of_two(i: int, p: bool):
    assert is_power_of_two(i) == p
```

superlopuh (approved):

tests/utils/test_bitwise_casts.py
JosseVanDelm: :open_mouth:

JosseVanDelm: yes

PR #1979, opened 2024-01-16
dialects: Implement affine.parallel

PapyChacal:


xdsl/dialects/affine.py
```diff
@@ -142,6 +146,41 @@ class If(IRDLOperation):
     else_region = region_def()
 
 
+@irdl_op_definition
+class ParallelOp(IRDLOperation):
```
superlopuh, line 150: nit: link to doc in MLIR?

superlopuh (commented):
Could you add an interop test, please?

xdsl/dialects/affine.py
PapyChacal: Is that a new convention we have? Otherwise, why specifically here?

PapyChacal:
> Could you add an interop test, please?

#1981 ! Will rebase this one, nothing urgent

xdsl/dialects/affine.py
superlopuh: No, I just felt like it would be useful here specifically, don't see a downside to it also

superlopuh (approved):
🚀

PR #1981, opened 2024-01-16
dialects: Refresh Affine, add interoperability test.

PapyChacal:
Refresh affine:
- `affine.for`:
  - Rename (lower|upper)_bound -> (lower|upper)BoundMap
  - Split variadics arguments in 3 variadic arguments :tada: 
- `affine.if`:
  For some reason `affine.if`'s condition is not a property.

Add interoperability test.


superlopuh (approved):
Thank you!

PR #1982, opened 2024-01-16
backend: (snitch) add iteration arguments to frep

superlopuh:
Adds iteration arguments to frep inner and outer, this lets us express things like dot products in frep. This PR also updates the reigster allocator to take these into account, and to treat them similarly to the carried variables in scf for loops.


compor (approved):
Rather on the big side and lots of different little things (even adding missing verification steps).
Let's get it in and worry later.

superlopuh:
Yeah apologies for the big diff, trying to decouple from an even bigger diff makes it a little difficult, I'll try to make future ones smaller.

PR #1983, opened 2024-01-16
dialects: Add constant unitattr to global

JosseVanDelm:
Adds an optional "constant" unit attribute to the memref global, which is emitted by upstream MLIR, but not yet parsed by xDSL :(

### DIFF
--- a/xdsl/dialects/memref.py
+++ b/xdsl/dialects/memref.py
@@ -516,7 +516,7 @@ class Global(IRDLOperation):
         }
         if constant is not None and constant:
             props["constant"] = UnitAttr()
-        return Global.build(*props)
+        return Global.build(properties=props)
 
 
 @irdl_op_definition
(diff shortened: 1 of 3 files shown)


xdsl/dialects/memref.py
```diff
@@ -501,20 +502,21 @@ def get(
         sym_type: Attribute,
         initial_value: Attribute,
         sym_visibility: StringAttr = StringAttr("private"),
+        constant: bool | None = None,
```
superlopuh, line 505:
```suggestion
        constant: UnitAttr | None = None,
```

xdsl/dialects/memref.py
```diff
@@ -501,20 +502,21 @@ def get(
...
-        )
+        props: dict[str, Attribute | None] = {
+            "sym_name": sym_name,
+            "type": sym_type,
+            "initial_value": initial_value,
+            "sym_visibility": sym_visibility,
+            "alignment": alignment,
+        }
+        if constant is not None and constant:
+            props["constant"] = UnitAttr()
```
superlopuh, line 518: It's probably worth changing the `constant` to be an optional UnitAttr, so this logic can be replaced by just `"constant": constant`, so the dictionary literal can move back into the `.build` helper

superlopuh (changes requested):
Thank you for this!

superlopuh (approved):

PR #1984, opened 2024-01-16
core: Move to assertions for modification permission checks 

### LINKED ISSUE #1950: core: Move to assertions for modification permission checks
As discussed in #1943, there is a lot of code duplication in the pattern rewriter around the `_can_modify_{op,region,block,op_in_block}` methods. In particular, these are mostly used as a check before raising an exception with fairly generic error messages. These checks could be moved to an `_assert_can_modify_{op,region,block,op_in_block}` to remove some of this duplication.

tarinduj:
Fixes #1950

### DIFF
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -130,6 +130,26 @@ class PatternRewriter(PatternRewriterListener):
             return True
         return self._can_modify_op(region.parent)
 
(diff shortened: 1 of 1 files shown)


superlopuh (approved):
Perfect, thank you!

math-fehr (approved):
Looks good, thanks a lot!

PR #1986, opened 2024-01-17
dialects: (snitch_stream) replace snitch_stream generic with streaming region

superlopuh:
Apologies for the large diff, I'm not sure how to reduce it.

This change simplifies the dialect quite a bit, by decoupling the iteration logic from the streaming logic. Instead of having a "generic" op at the snitch_stream level, it's expected that the streaming region contains all the iteration logic directly, and the only semantic support it provides is a high-level representation of a chunk of code where certain streams can be read from or written to.

### DIFF
--- a/xdsl/dialects/snitch_stream.py
+++ b/xdsl/dialects/snitch_stream.py
@@ -1,3 +1,23 @@
+"""
+A dialect that represents at the highest level of abstraction the capabilities of the
+[Snitch](https://github.com/pulp-platform/snitch_cluster) accelerator core, as used in
+[Occamy](https://github.com/pulp-platform/occamy) and others.
+
+The core aims to optimise for performance per watt, by replacing caches and branch
+prediction logic with streaming registers and fixed-repetition loops. This dialect models
+the streaming functionality of the Snitch core.
+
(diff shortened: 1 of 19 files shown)


xdsl/dialects/snitch_stream.py
```diff
@@ -55,58 +50,37 @@ def print_parameter(self, printer: Printer):
 
 
 @irdl_op_definition
-class GenericOp(IRDLOperation):
+class StreamingRegionOp(IRDLOperation):
```
compor, line 53: Can this be documented a bit better? Specifically the `inputs` and `outputs`.
At this point in the dialect we are pretty much expecting integer registers holding a memory location pointer, right?

Moreover, can you please state this separation of responsibilities (iteration and streaming) and where it happens as you have so nicely written in the PR message?

This is minor, but it made me think and had to look at the lowering to make sure that this is what is happening.

compor (approved):

xdsl/dialects/snitch_stream.py
superlopuh: Very good points, I'll add some documentation

PR #1987, opened 2024-01-18
dialects: Add result arguments to func.func

JosseVanDelm:
This is also part of my crusade to get some tflite models to be parseable by xDSL:

Adds an optional result arguments dictionary attribute to func.func.

This PR does not add parsing/printing support yet in its current form, could someone help me with that please!

Thanks!


superlopuh (approved):
LGTM, can you please file an issue to add result attributes to func.func custom format, and add it in a comment? Ideally the issue would have a snippet that we should be able to rountrip.

JosseVanDelm:
Thanks @superlopuh, I made to track progress on the printing and parsing #1988

PR #1989, opened 2024-01-19
dialects: Update pdl dialect to leverage properties

martin-luecke:
This updates all operations of the PDL dialect to use properties rather than attributes. 
Correspondingly, the tests that relied on this were updated.
This also adds tests for `pdl.PatternOp`, which was untested before.


tests/dialects/test_pdl.py
```diff
@@ -94,6 +97,49 @@ def test_build_operation_replace():
...
+
+        @Builder.implicit_region
+        def body() -> None:
+            pdl.RewriteOp(root.results[0])
+
+        pattern = pdl.PatternOp(1, "pattern", body)
+        pattern.verify()
+    assert "the pattern must contain at least one `pdl.operation`" in str(
+        verify_exception.value
+    )
```
superlopuh, line 140:
```suggestion
    with pytest.raises(VerifyException, match="the pattern must contain at least one `pdl.operation`"):
        root = pdl.OperationOp("test.op")

        @Builder.implicit_region
        def body() -> None:
            pdl.RewriteOp(root.results[0])

        pattern = pdl.PatternOp(1, "pattern", body)
        pattern.verify()
```

and above

superlopuh (approved):
Thank you! Can you please make a couple of small changes to the tests?

tests/dialects/test_pdl.py
martin-luecke: I wondered if there was a better way, thanks!

math-fehr (approved):
Very nice!

PR #1990, opened 2024-01-21
dialects: (riscv_debug) add RISCV_Debug dialect with printf op

superlopuh:
This op corresponds to the printf op in riscemu, and should make it easier for us to debug values in the future.


compor (approved):

xdsl/dialects/riscv_debug.py
```diff
@@ -0,0 +1,76 @@
+from collections.abc import Sequence, Set
```
AntonLydike, line 1: I think a short comment that this is a specific extension to `riscemu` would be reasonable here, just so that people don't mistake this for a standard riscv debugging extension.

AntonLydike: Ah nevermind, it's just below here^^

xdsl/dialects/riscv_debug.py
```diff
@@ -0,0 +1,76 @@
...
+    and xDSL's interpreter.
+
+    During assembly emission, the results are printed before the operands:
+
+    ``` python
+    s0 = riscv.GetRegisterOp(Registers.s0).res
+    s1 = riscv.GetRegisterOp(Registers.s1).res
+    op = PrintfOp("s0: {}, s1: {}", (s0, s1))
+
+    op.assembly_line()   # "printf "s0: {}, s1: {}", s0, s1"
```
AntonLydike, line 26: This line seems to have a few too many quotation marks?

AntonLydike (approved):
Nice stuff!

xdsl/dialects/riscv_debug.py
superlopuh: ah it's just the wrong quotation marks

PR #1996, opened 2024-01-22
core: Implement basic ConstraintVar resolution for assembly format

PapyChacal:


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -119,6 +134,32 @@ def parse(
             result_types=result_types, operands=operands, attributes=state.attributes
         )
 
+    def resolve_operand_types(self, state: ParsingState) -> None:
+        for i, operand_type in enumerate(state.operand_types):
+            if operand_type is None:
+                state.operand_types[i] = self._resolve_type(
```
superlopuh, line 140: is this legal? the dict is being modified while iterated, I don't think the keys are copied by default

PapyChacal: state.operand_types is a list, and I'm only updating existing elements

superlopuh: Ah, my bad, I think this is still not allowed, I'd recommend iterating over the indices and getting the element in the loop

superlopuh: Looks like it is allowed all good

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -119,6 +134,32 @@ def parse(
...
+
+    def _resolve_type(self, state: ParsingState, construct: VarIRConstruct, index: int):
+        resolve, (construct, idx) = self.type_resolutions[construct, index]
+        match construct:
+            case VarIRConstruct.OPERAND:
+                input_type = state.operand_types[idx]
+            case VarIRConstruct.RESULT:
+                input_type = state.result_types[idx]
+            case _:
+                raise ValueError()
```
superlopuh, line 159: Would be nice to add a string here and a test that reaches it

PapyChacal: Not done :+1: 
Reworked the typing to let pyright know that the two handle cases are the only possible ones :yum:

xdsl/irdl/declarative_assembly_format.py
PapyChacal: 🫡🚔

superlopuh (approved):

math-fehr (approved):

PR #1997, opened 2024-01-22
dialects: Move MemRefType et al. to builtin.

PapyChacal:
(Unranked)MemRefType are builtin attributes, at least in the sense that they have special handling in the core parser/printer, and as such should live in the `builtin` dialect, to avoid spreading dependencies. (This is the MLIR case too)
I am working towards having our assembly_format system ready enough for memeref, and one of the issues is that using it triggers a circular dependency if we leave those types in `memref` rather than `builtin` :slightly_smiling_face:

### DIFF
--- a/docs/irdl.ipynb
+++ b/docs/irdl.ipynb
@@ -1204,9 +1204,7 @@
     "add_op2 = AddVariadicOp2.build(\n",
     "    operands=[[i32_ssa_var] * 2, [i32_ssa_var]],\n",
     "    result_types=[i32],\n",
-    "    attributes={\n",
-    "        \"operandSegmentSizes\": VectorType(i32, [2, 1])\n",
-    "    },\n",
+    "    attributes={\"operandSegmentSizes\": VectorType(i32, [2, 1])},\n",
     ")\n",
     "print(\"Length of add_op2.ops1:\", len(add_op2.ops1))\n",
     "print(\"Length of add_op2.ops2:\", len(add_op2.ops2))"
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1472,6 +1472,123 @@ f80 = Float64Type()
 f128 = Float64Type()
 
 
+_MemRefTypeElement = TypeVar("_MemRefTypeElement", bound=Attribute)
+_UnrankedMemrefTypeElems = TypeVar(
+    "_UnrankedMemrefTypeElems", bound=Attribute, covariant=True
+)
+_UnrankedMemrefTypeElemsInit = TypeVar("_UnrankedMemrefTypeElemsInit", bound=Attribute)
+
+
+@irdl_attr_definition
+class MemRefType(
+    Generic[_MemRefTypeElement],
+    ParametrizedAttribute,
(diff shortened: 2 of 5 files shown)


compor (approved):

webmiche (commented):
First things first, I don't really know your usecase, so maybe this is really the only way. If so, I am fine with doing this.

However, from a conceptual standpoint, I don't like this very much. To me, memrefs very much are attributes that "belong" to the memref dialect, I think this should be reflected in the repo structure. I know that MLIR has them in builtin and that they share aspects (most implementation-specific ones) that makes them fit there, but I strongly suggest to consider whether we don't have other options that can achieve the same result. My main argument would be that I really want this to be as extensible as possible, especially with "out-of-tree" dialects, which should not have to move their attributes to "builtin" to get parts of the functionality.

In order to circumvent circular imports, how about having a memref folder with a `memref_ops` and a `memref_attrs`-file? Would that be a solution?

math-fehr:
To be honest, moving memref would both solve some of these issues, but also would make sense relative to MLIR.

It's always a bit weird that `memref` is outside of the `builtin` dialect in xDSL, while it is in `builtin` in MLIR.
Also, it is a bit inconsistent right now that `memref` is in `memref`, while `vector` or `tensor` is in `builtin` in xDSL.

So I would be in favor of moving it to `builtin` to be fair

math-fehr:
The real design flaw is that `memref` is required by the core (the parser in this case).
As long as this is the case, we should keep `memref` in `builtin` IMO.
But that's a fight I'm currently fighting in MLIR itself, my hope is that this year we can move it to `memref` dialect both in MLIR and xDSL, and not have this special case in the parser.

webmiche:
Okay, how about we move it there for now and open an issue to later move it back?

PapyChacal:
> Well, I would argue that this is a design flaw in MLIR. Not sure we want to just copy it.

Definitely! But AFAIK, so far parsing memref as a builtin type is the only way; so for compatibility, we have to special-case parse it, which means core parser dependency on the type! That's a technical need for some attributes to be in builtin. We were just lucky to get around it so far with Python, but using assemby_format triggers it. I am happy to hear concrete workarounds ideas, but:
- I don't think splitting the file would do, as the cycle is directly from those types to the parser and back.
- I won't invest energy in finding one myself; while it is a minor conceptual frustration, I think there are more profitable can of worms to open :yum:

To add to the "doing like MLIR" rationale: right now, the actual types syntax is the same precisely because of its custom handling in the parser. If we were to make it compatible with the more principled syntax/naming (i.e. `!dialect.memref`), it would be by prefixing the dialect's name; just stucking us to use the same than MLIR for compatibility again.

PapyChacal:
> Okay, how about we move it there for now and open an issue to later move it back?

Oh, missed it while writing my last comment. I'm fine with the idea, opening one! Thanks for the compromise :wink:

PapyChacal:
Opened #2004 just now; not sure about the writing, feel free to message me or edit it for clarity :slightly_smiling_face:

math-fehr (approved):

PR #1142, opened 2023-06-15
transforms: (print) Add a lowering to printf

AntonLydike:
This is a working, minimal lowering to `printf`!

I have some hacky stuff happening in the ModulePass, let me know if this is undesireable!

PR #2002, opened 2024-01-22
documentation: (Toy) add instruction on how to jit Toy using lli

superlopuh:
Now that Toy lowers to print_format, we can compile it to LLVM, leveraging the printf flow added in #1142 .

### DIFF
--- a/docs/Toy/Toy_Ch0.ipynb
+++ b/docs/Toy/Toy_Ch0.ipynb
@@ -74,18 +74,10 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": 2,
    "id": "3af9c75a",
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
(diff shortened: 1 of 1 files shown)


JosseVanDelm:
Why not use `mlir-cpu-runner` instead of `mlir-translate | lli`?

superlopuh:
Because I found out about this via this PR :)

webmiche (approved):
:heart_eyes:

superlopuh:
```
!python -m toy examples/interpret.toy --emit=scf --ir | xdsl-opt -p printf-to-llvm | mlir-opt --test-lower-to-llvm | mlir-cpu-runner
Error: only single f32 function result supported
```
Not sure what's going wrong here...

compor (approved):

JosseVanDelm:
@superlopuh I think you have to use `mlir-cpu-runner --entry-point-result=void`:
```
  --entry-point-result=<f32 | i32 | i64 | void>      - Textual description of the function type to be called
```

PR #2005, opened 2024-01-23
core: Add support for nested inferrence in IRDL

math-fehr:
This PR allows us to infer the types of some operands and results from the types of other operands and results.

It does so by adding three functions in `AttrConstraint`:
* `get_resolved_variables`: This function returns the set of `ConstraintVar` that will be set during verification.
* `can_infer`: Returns `True` if we can infer a unique attribute represented by the `AttrConstraint` from the assignment of `ConstraintVar`.
* `infer`: Returns the only attribute that satisfy this constraint, given the constraint variables.

We use these methods to get all `ConstraintVar`  from the parsed operand/result types present in the syntax, then infer the constraints on the types that are not in the syntax.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -116,6 +116,39 @@ class FormatProgram:
             stmt.parse(parser, state)
 
         # Get constraint variables from the parsed operand and result types
+        self.assign_constraint_variables(parser, state, op_def)
+
+        # Infer operand types that should be inferred
+        unresolved_operands = state.operands
+        assert isa(unresolved_operands, list[UnresolvedOperand])
+        self.resolve_operand_types(state, op_def)
+        operand_types = state.operand_types
+        assert isa(operand_types, list[Attribute])
+
+        # Infer result types that should be inferred
+        self.resolve_result_types(state, op_def)
+        result_types = state.result_types
+        assert isa(result_types, list[Attribute])
+
+        # Resolve all operands
+        operands = parser.resolve_operands(
+            unresolved_operands, operand_types, parser.pos
+        )
+        properties = op_def.split_properties(state.attributes)
+        return op_type.build(
+            result_types=result_types,
+            operands=operands,
+            attributes=state.attributes,
+            properties=properties,
+        )
+
+    def assign_constraint_variables(
+        self, parser: Parser, state: ParsingState, op_def: OpDef
+    ):
+        """
+        Assign constraint variables with values got from the
+        parsed operand and result types.
+        """
         if any(type is None for type in (*state.operand_types, *state.result_types)):
             try:
                 for (_, operand_def), operand_type in zip(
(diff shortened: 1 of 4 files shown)


PapyChacal (approved):
Yaaay :rocket:

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -136,44 +136,60 @@ def parse(
...
+    ):
+        """
+        Assign constraint variables with values got from the
+        parsed operand and result types.
+        """
+        if any(type is None for type in (*state.operand_types, *state.result_types)):
+            try:
+                for (_, operand_def), operand_type in zip(
+                    op_def.operands, state.operand_types
+                ):
```
superlopuh, line 150: Should this be strict?

superlopuh: And below

xdsl/irdl/irdl.py
```diff
@@ -268,6 +333,13 @@ def verify(self, attr: Attribute, constraint_vars: dict[str, Attribute]) -> None
         for idx, param_constr in enumerate(self.param_constrs):
             param_constr.verify(attr.parameters[idx], constraint_vars)
 
+    def get_resolved_variables(self) -> set[str]:
+        if len(self.param_constrs) == 0:
+            return set()
```
superlopuh, line 338:
```suggestion
        if not self.param_constrs:
            return set()
```

xdsl/irdl/irdl.py
```diff
@@ -268,6 +333,13 @@ def verify(self, attr: Attribute, constraint_vars: dict[str, Attribute]) -> None
...
+        return set[str].union(
+            *[constr.get_resolved_variables() for constr in self.param_constrs]
+        )
```
superlopuh, line 341: Does this work?
```suggestion
        return {var for for constr in self.param_constrs for var in constr.get_resolved_variables()}
```

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -114,95 +114,62 @@ def parse_format(self) -> FormatProgram:
...
         """
-        for i, (operand, operand_type, (operand_name, _)) in enumerate(
-            zip(self.seen_operands, self.seen_operand_types, self.op_def.operands)
-        ):
-            if not operand:
+        for (
+            seen_operand,
+            seen_operand_type,
+            (operand_name, operand_def),
+        ) in zip(self.seen_operands, self.seen_operand_types, self.op_def.operands):
```
superlopuh, line 145: nit: strict here also?

xdsl/irdl/irdl.py
```diff
@@ -209,6 +251,13 @@ def verify(self, attr: Attribute, constraint_vars: dict[str, Attribute]) -> None
                 pass
         raise VerifyException(f"Unexpected attribute {attr}")
 
+    def get_resolved_variables(self) -> set[str]:
+        if len(self.attr_constrs) == 0:
+            return set()
+        return set[str].intersection(
+            *[constr.get_resolved_variables() for constr in self.attr_constrs]
```
superlopuh, line 258: nit:
```suggestion
            *(constr.get_resolved_variables() for constr in self.attr_constrs)
```

superlopuh (approved):

xdsl/irdl/irdl.py
PapyChacal: Good catch thanks :)

PR #2011, opened 2024-01-23
core: Support variadic operands and results in declarative assembly format.

PapyChacal:


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -35,24 +35,21 @@ class ParsingState:
...
+    result_types: list[Attribute | None | list[Attribute | None]]
     attributes: dict[str, Attribute]
     constraint_variables: dict[str, Attribute]
 
     def __init__(self, op_def: OpDef):
-        if op_def.attributes or op_def.regions or op_def.successors:
+        attributes = bool(op_def.attributes) and list(op_def.attributes.keys()) != [
+            "operandSegmentSizes"
+        ]
+        if attributes or op_def.regions or op_def.successors:
```
superlopuh, line 48: `attributes` threw me off here, I feel like `has_attributes` communicates that it's a `bool`
```suggestion
        has_attributes = bool(op_def.attributes) and list(op_def.attributes.keys()) != [
            "operandSegmentSizes"
        ]
        has_if attributes or op_def.regions or op_def.successors:
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -118,20 +115,27 @@ def parse(
...
         result_types = state.result_types
-        assert isa(result_types, list[Attribute])
+        assert isa(result_types, list[Attribute | list[Attribute]])
 
         # Resolve all operands
-        operands = parser.resolve_operands(
-            unresolved_operands, operand_types, parser.pos
-        )
+        operands: Sequence[SSAValue | Sequence[SSAValue]] = []
+        for uo, ot in zip(unresolved_operands, operand_types):
```
superlopuh, line 132:
```suggestion
        for uo, ot in zip(unresolved_operands, operand_types, strict=True):
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -118,20 +115,27 @@ def parse(
...
+            if isinstance(uo, list):
+                assert isinstance(ot, list), ot
```
superlopuh, line 134: feels like these should be parsing errors but I'm not sure how important that is

superlopuh (approved):
Looking forward to using this

xdsl/irdl/declarative_assembly_format.py
PapyChacal: This holds by construction, it's pyright handshaking; it's one of those cases where having this being false is not an input case, but properly a bug of the preceding steps.
I feel like throwing errors for the sake of prettifying some code present for static analysis might be sending the wrong message to later code readers :thinking: 
I guess a silent assert is not saying much more though. WDYT, should it just be a cast in that case? Then catching the bug is confusing :smiling_face_with_tear:

superlopuh: I think it's good, maybe worth adding this explanation in the string next to the assert, instead of `ot`?

superlopuh: At this point it's more of a nit

PapyChacal: Oh right, that much is just uncleaned assert debugging ahah. Thanks for the catch!

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -113,96 +117,83 @@ def parse_format(self) -> FormatProgram:
...
+                and last.punctuation == ","
+                and isinstance(
+                    elements[-2],
+                    VariadicOperandTypeDirective
+                    | VariadicResultTypeDirective
+                    | VariadicOperandVariable,
+                )
+            ):
+                self.raise_error(
+                    "A variadic directive cannot be followed by a comma litteral."
```
compor, line 133:
```suggestion
                    "A variadic directive cannot be followed by a comma literal."
```

nit

compor (approved):

PR #2013, opened 2024-01-23
core: Make assembly format use attr-dict for properties

PapyChacal:
This is MLIR's behaviour. Probably only when there is no specified prop-dict in the format, but on the current commit, this prop-dict is only used in their tests, so I'm not rushing implementation :)
Some secondary changes:
- Factor out the property retro-compatibility from the core Parser to OpDef
- Don't forbid Operations with defined attributes. Why was it forbidden? @math-fehr They are already usable through the attr-dict, simply not in directives, am I missing something?

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -230,7 +230,6 @@ class AttrDictDirective(FormatDirective):
         state.attributes = res
 
     def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> None:
-        # assert False, (op.attributes | op.properties, op)
         if not op.attributes and not op.properties:
             return
         if self.with_keyword:
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -234,6 +234,11 @@ class AttrDictDirective(FormatDirective):
             return
         if self.with_keyword:
             printer.print(" attributes")
+        if any(name in op.attributes for name in op.properties):
+            raise ValueError(
+                "Cannot print attributes and properties with the same name"
+                "in a signle dictionary"
+            )
         printer.print_op_attributes(op.attributes | op.properties)
         state.last_was_punctuation = False
         state.should_emit_space = False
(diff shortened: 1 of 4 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -223,11 +230,12 @@ def parse(self, parser: Parser, state: ParsingState) -> None:
         state.attributes = res
 
     def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> None:
-        if not op.attributes:
+        # assert False, (op.attributes | op.properties, op)
```
math-fehr, line 233:
```suggestion
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -223,11 +230,12 @@ def parse(self, parser: Parser, state: ParsingState) -> None:
...
+        if not op.attributes and not op.properties:
             return
         if self.with_keyword:
             printer.print(" attributes")
-        printer.print_op_attributes(op.attributes)
+        printer.print_op_attributes(op.attributes | op.properties)
```
math-fehr, line 238: What happens in MLIR if there are both properties and attributes with the same name?
I feel that we should raise an error, whatever MLIR does.

xdsl/irdl/irdl.py
```diff
@@ -1369,6 +1369,17 @@ def verify(self, op: Operation):
         for trait in self.traits:
             trait.verify(op)
 
+    def split_properties(self, attr_dict: dict[str, Attribute]):
```
math-fehr, line 1372:
```suggestion
    def split_properties(self, attr_dict: dict[str, Attribute]) -> dict[str, Attribute]:
```

xdsl/parser/core.py
```diff
@@ -885,9 +885,7 @@ def _parse_generic_operation(self, op_type: type[Operation]) -> Operation:
         # Properties retrocompatibility : if no properties dictionary was present at all,
         # We extract them from the attribute dictionary by name.
         if issubclass(op_type, IRDLOperation) and not properties:
-            for property_name in op_type.get_irdl_definition().properties.keys():
```
math-fehr, line 888: Nice

math-fehr (approved):

xdsl/irdl/declarative_assembly_format.py
PapyChacal: MLIR silently overrides the extra attributes by the expected property in that case.
I'm okay with making it an error on our side

superlopuh (approved):

compor (approved):

PR #2016, opened 2024-01-24
dialects: (builtin) add `NoneType`

kayode-gif:
adds `NoneType` to `builtin` dialect


PapyChacal:
Hello, thank you for opening this! Out of curiosity, do you have a usecase in mind?

In general, when adding an attribute or operation, you should add filecheck tests to check that it parses and prints as you intended. I would add this one in `tests/filecheck/dialects/builtin/attrs.mlir`

Because this one is a builtin attribute, it needs to be handled specially in the core printer and parser. This is more complex in itself, but it's a simple example! I would look into `xdsl/parser/attribute_parser.py`, method `try_parse_builtin_boolean_attr` for the `bool`/`i1` example. Similarly for the printer, look in `/home/papychacal/xdsl/xdsl/printer.py`, method `print_attribute` : implemented builtin attributes are handled there, and this one should be too!

kayode-gif:
> Hello, thank you for opening this! Out of curiosity, do you have a usecase in mind?
> 
> In general, when adding an attribute or operation, you should add filecheck tests to check that it parses and prints as you intended. I would add this one in `tests/filecheck/dialects/builtin/attrs.mlir`
> 
> Because this one is a builtin attribute, it needs to be handled specially in the core printer and parser. This is more complex in itself, but it's a simple example! I would look into `xdsl/parser/attribute_parser.py`, method `try_parse_builtin_boolean_attr` for the `bool`/`i1` example. Similarly for the printer, look in `/home/papychacal/xdsl/xdsl/printer.py`, method `print_attribute` : implemented builtin attributes are handled there, and this one should be too!

thanks for the reply, the usecase is that i need this to represent an absent value for an onnx operation (```onnx.conv```), which i can't do atm as nonetype isn't a type in xdsl rn

xdsl/parser/attribute_parser.py
```diff
@@ -1124,6 +1125,15 @@ def try_parse_builtin_boolean_attr(
             return IntegerAttr(1 if value else 0, IntegerType(1))
         return None
 
+    def try_parse_builtin_none_type(self) -> NoneType | None:
+        """
+        Parse a none type, if present
+        none-type ::= `none`
+        """
+        if (value := self.parse_optional_type()) is not None:
+            return NoneType(value)
```
superlopuh, line 1134: I would recommend looking into how `IndexType` is parsed, and adding similar logic there. This won't work because parse_optional_type should itself be able to pase `none` and return `NoneType`

xdsl/printer.py
```diff
@@ -666,6 +667,11 @@ def print_int_or_question(value: IntAttr | NoneAttr) -> None:
 
         if isinstance(attribute, OpaqueSyntaxAttribute):
             self.print(">")
+        # add none-type to printer
```
superlopuh, line 670:
```suggestion

```

xdsl/parser/attribute_parser.py
```diff
@@ -1124,6 +1125,17 @@ def try_parse_builtin_boolean_attr(
             return IntegerAttr(1 if value else 0, IntegerType(1))
         return None
 
+    def try_parse_builtin_none_type(self) -> NoneType | None:
```
superlopuh, line 1128:
```suggestion
    def _parse_optional_none_type(self) -> NoneType | None:
```
Can you please add a call to this in `_parse_optional_builtin_type`?

kayode-gif:
> Can you please add `none` testing to two filechecks, one in `tests/filecheck/dialects/builtin/attrs.mlir`, and in a new one in the same folder as `unrealized_conversion_cast.mlir`? The first one would check that xDSL can parse and print the new type, the second that `mlir-opt` can parse what xDSL can print

okay, im having issues with this bit, is it something im missing in the builtin? 
```    %x16 = "arith.constant"() <{"value" = 0 : i64}> {"test" = !none_typenone} : () -> i64
                                                                            ^
                                                                            undefined symbol alias '!none_typenone'
```

kayode-gif:
%x16 = "arith.constant"() <{"value" = 0 : i64}> {"test" = !none_typenone} : () -> i64
                                                                            ^
                                                                            undefined symbol alias '!none_typenone'
this is what i mean

kayode-gif:
> %x16 = "arith.constant"() <{"value" = 0 : i64}> {"test" = !none_typenone} : () -> i64 ^ undefined symbol alias '!none_typenone' this is what i mean

this is for the attrs.mlir file btw

superlopuh:
I'm a little confused about the example you give, I'd expect something like this:
```
%none = "test.op"() : () -> none
"test.op"(%none) : none -> ()
```

superlopuh:
For the MLIR interop, we'd have to explicitly say "--allow-unregistered-dialect" so that it's happy with our test op

kayode-gif:
> I'm a little confused about the example you give, I'd expect something like this:
> 
> ```
> %none = "test.op"() : () -> none
> "test.op"(%none) : none -> ()
> ```

i mean i receive this error:   undefined symbol alias '!none_typenone', when adding a test in

kayode-gif:
> For the MLIR interop, we'd have to explicitly say "--allow-unregistered-dialect" so that it's happy with our test op

sorry, where would this go, inside the attr-dict, i am a bit unsure with adding tests in tbf

xdsl/printer.py
```diff
@@ -666,6 +667,11 @@ def print_int_or_question(value: IntAttr | NoneAttr) -> None:
 
         if isinstance(attribute, OpaqueSyntaxAttribute):
             self.print(">")
+
+        if isinstance(attribute, NoneType):
+            self.print("none")
+            return
+
```
superlopuh, line 674: this needs to be higher

PapyChacal: Yep, More specifically anywhere before TypeAttribute: TrypeAttribute and below cases are more generic ones that will be applied to any attribute not falling in the earlier special cases.

PapyChacal: NB to the team: this deserve a comment ;)

kayode-gif: thanks for the spot :)

superlopuh (approved):

webmiche:
> thanks for the reply, the usecase is that i need this to represent an absent value for an onnx operation (`onnx.conv`), which i can't do atm as nonetype isn't a type in xdsl rn

Could you not model this with an optional attribute?

math-fehr:
> > thanks for the reply, the usecase is that i need this to represent an absent value for an onnx operation (`onnx.conv`), which i can't do atm as nonetype isn't a type in xdsl rn
> 
> Could you not model this with an optional attribute?

They do not: 
image

xdsl/parser/attribute_parser.py
```diff
@@ -1124,6 +1125,17 @@ def try_parse_builtin_boolean_attr(
...
+    def _parse_optional_none_type(self) -> NoneType | None:
+        """
+        Parse a none type, if present
+        none-type ::= `none`
+        """
+        name = self._current_token.text
+        if name == "none":
+            self._consume_token()
+            return NoneType()
+        return None
```
math-fehr, line 1137:
```suggestion
        if self.parse_optional_keyword("none"):
            return NoneType()
        return None
```

math-fehr (approved):
Nice!

webmiche:
> They do not: image

 Ah, I see. This is some ONNX-internal typing stuff. Next question: should this be in builtin or should this just be in the ONNX dialect?

math-fehr:
> > They do not: image
> 
> Ah, I see. This is some ONNX-internal typing stuff. Next question: should this be in builtin or should this just be in the ONNX dialect?

This is in builtin: https://mlir.llvm.org/docs/Dialects/Builtin/#nonetype
So to me the PR is good!

webmiche:
> > Ah, I see. This is some ONNX-internal typing stuff. Next question: should this be in builtin or should this just be in the ONNX dialect?
> 
> This is in builtin: https://mlir.llvm.org/docs/Dialects/Builtin/#nonetype So to me the PR is good!

Ah, I see again xD
We should add a PR template for porting MLR stuff at some point. There, we could require people to give a link to the MLIR doc. This LGTM then.

compor (approved):

PR #2018, opened 2024-01-24
Add pass to set `fastmath` flags of FP binary operations from `arith`

### LINKED ISSUE #2017: Add pass to manipulate `fastmath` flags of `arith` operations

compor:
This PR:

- Adds a rewrite pass to set the `fastmath` flags of FP binary `arith` operations.
- Tests of the above.

Resolves #2017

### DIFF
--- a/xdsl/transforms/arith_add_fastmath.py
+++ b/xdsl/transforms/arith_add_fastmath.py
@@ -1,4 +1,5 @@
-from dataclasses import dataclass, field
+from dataclasses import dataclass
+from typing import Literal
 
 from xdsl.dialects import arith, builtin, llvm
 from xdsl.ir import Operation
@@ -28,6 +29,8 @@ def _get_flag_list(flags: list[str], enum_type: type[llvm.FastMathFlag]):
 
 @dataclass
 class AddArithFastMathFlags(RewritePattern):
+    """Adds fastmath flags to FP binary operations from arith dialect."""
+
     arith_op_cls: type[arith.FloatingPointLikeBinaryOp]
     fastmath_op_attr: arith.FastMathFlagsAttr
 
@@ -40,22 +43,26 @@ class AddArithFastMathFlags(RewritePattern):
 
 @dataclass
 class AddArithFastMathFlagsPass(ModulePass):
+    """Module pass that adds fastmath flags to FP binary operations from arith dialect.
+    It currently does not preserve any existing fastmath flags that may already be part
+    of the operation.
+    By default (no arguments) it adds the "fast" flag.
+
+    Arguments (all optional):
+
+    - flags: {"fast", "none"} | list[str]: Set specific fastmath flags. Using "fast" or
+      "none" enables or disables all flags, respectively.
+    """
+
     name = "arith-add-fastmath"
 
-    flags: list[str] = field(default_factory=lambda: ["fast"])
+    flags: Literal["fast", "none"] | list[str] = "fast"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         fm_flags = arith.FastMathFlagsAttr("fast")
 
-        if len(self.flags) == 1:
-            if "none" in self.flags:
-                fm_flags = arith.FastMathFlagsAttr("none")
-            elif "fast" in self.flags:
-                fm_flags = arith.FastMathFlagsAttr("fast")
-            else:
-                fm_flags = arith.FastMathFlagsAttr(
-                    _get_flag_list(self.flags, llvm.FastMathFlag)
-                )
+        if isinstance(self.flags, str):
+            fm_flags = arith.FastMathFlagsAttr(self.flags)
         else:
             if "none" in self.flags or "fast" in self.flags:
                 raise ValueError(
--- a/xdsl/transforms/arith_add_fastmath.py
+++ b/xdsl/transforms/arith_add_fastmath.py
@@ -13,20 +13,14 @@ from xdsl.pattern_rewriter import (
     RewritePattern,
 )
 
-
-def _match_string_to_enum(flag: str, enum_type: type[llvm.FastMathFlag]):
-    for member in enum_type:
-        if flag == str(member.value):
-            return member
-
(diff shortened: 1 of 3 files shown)


xdsl/transforms/arith_add_fastmath.py
```diff
@@ -0,0 +1,90 @@
...
+    flags: Literal["fast", "none"] | list[str] = "fast"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        fm_flags = arith.FastMathFlagsAttr("fast")
+
+        if isinstance(self.flags, str):
+            fm_flags = arith.FastMathFlagsAttr(self.flags)
+        else:
+            if "none" in self.flags or "fast" in self.flags:
+                raise ValueError(
```
compor, line 70: I'm not sure if this needs to be tested; I looked at other passes which raise exceptions and we don't do it.
Not sure why.

xdsl/transforms/arith_add_fastmath.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@dataclass
+class AddArithFastMathFlags(RewritePattern):
+    """Adds fastmath flags to FP binary operations from arith dialect."""
+
+    arith_op_cls: type[arith.FloatingPointLikeBinaryOp]
+    fastmath_op_attr: arith.FastMathFlagsAttr
+
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter) -> None:
```
PapyChacal, line 39: Wouldn't
```suggestion
    @op_type_rewrite_pattern
    def match_and_rewrite(self, op: arith.Addf | arith.Subf | arith.Mulf | arith.Divf | arith.Minimumf | arith.Maximumf, rewriter: PatternRewriter) -> None:
```
achieve what you hardcode at the Pass level either way? Now, maybe I'm missing a rationale to not use it

xdsl/transforms/arith_add_fastmath.py
PapyChacal: I guess because most passes we wrote were written in a paper rush :thinking: 
I feel that this logic should rather live in `FastMathFlagsAttr` either way; and would be more sensible to test there.

xdsl/transforms/arith_add_fastmath.py
```diff
@@ -0,0 +1,90 @@
...
+        if not isinstance(op, self.arith_op_cls):
+            return
+
+        op.fastmath = self.fastmath_op_attr
```
PapyChacal, line 43: Wouldn't
```suggestion
    @op_type_rewrite_pattern
    def match_and_rewrite(self, op: arith.Addf | arith.Subf | arith.Mulf | arith.Divf | arith.Minimumf | arith.Maximumf, rewriter: PatternRewriter) -> None:
        op.fastmath = self.fastmath_op_attr
```
achieve what you hardcode at the Pass level either way? Now, maybe I'm missing a rationale to not use it.

PapyChacal (approved):
Some discussion but probably more like potential improvements than anything else!
Thanks

xdsl/transforms/arith_add_fastmath.py
compor: Yeah, agreed, not sure what's best.

We do what I'm doing at the pass level at the `arith` lowering to `riscv` [here](https://github.com/xdslproject/xdsl/blob/b9b6223fac1528b94116057891cfc1f7289b28cb/xdsl/backend/riscv/lowering/convert_arith_to_riscv.py#L172).

Not sure what are the pros and cons for each, since we need to spell out explicitly the related operations at some point.

xdsl/transforms/arith_add_fastmath.py
compor: I'll see if this warrants a separate PR, thanks!

xdsl/transforms/arith_add_fastmath.py
PapyChacal: Interesting! I'd say it makes more sense in that lowering case, where you map each type to another. So having the decorator do the check to then map the actual type would be paying overhead, where here I don't think you ever want to do more than skip non-listed types?

But anyway, I guess if we want to be perfectionists about it, the right way is a Trait/Interface: MLIR defines `ArithFastMathInterface` for this! (Used in math et al. too)

So I'd say let's just merge this as a simple helper, and see when we want a properly generic one :slightly_smiling_face:

xdsl/transforms/arith_add_fastmath.py
```diff
@@ -0,0 +1,90 @@
...
+            return member
+
+    raise ValueError(f"{flag} is not a valid fastmath flag.")
+
+
+def _get_flag_list(flags: list[str], enum_type: type[llvm.FastMathFlag]):
+    list_flags: list[enum_type] = []
+    for flag in flags:
+        list_flags.append(_match_string_to_enum(flag, llvm.FastMathFlag))
+    return list_flags
```
superlopuh, line 29:
```suggestion
_STRING_TO_ENUM = {str(member.value): member for member in llvm.FastMathFlag}

def _get_flag_list(flags: list[str]):
    try:
        return [_STRING_TO_ENUM[flag] for flag in flags] 
    except KeyError:
        raise ValueError(f"{flag} is not a valid fastmath flag.")    
```

superlopuh: Feels worth caching, and it seems like the enum_type parameter won't ever be different from `llvm.FastMathFlag`

superlopuh (approved):

xdsl/transforms/arith_add_fastmath.py
compor: What decorator overhead? I'm not using one.

xdsl/transforms/arith_add_fastmath.py
compor: I had a look and the definition of `__init__` restricts the choices well. Moreover, any pass exceptions are not tested anywhere, so I'll just leave it as is for now.

This is the only point where string values coming from a user need to match to this enum.

PR #2019, opened 2024-01-24
bug: fix AffineMap.inverse_permutation

superlopuh:
Currently, loop bounds are not computed correctly for linalg.generic lowering

### DIFF
--- a/xdsl/ir/affine/affine_map.py
+++ b/xdsl/ir/affine/affine_map.py
@@ -190,25 +190,26 @@ class AffineMap:
         for i, expr in enumerate(self.results):
             match expr:
                 case AffineDimExpr():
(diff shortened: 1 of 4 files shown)


xdsl/ir/affine/affine_map.py
```diff
@@ -190,25 +190,26 @@ def inverse_permutation(self) -> AffineMap | None:
         for i, expr in enumerate(self.results):
             match expr:
                 case AffineDimExpr():
-                    found_dims[expr.position] = i
+                    if found_dims[expr.position] == -1:
```
compor, line 193: good catch

compor (approved):

math-fehr (approved):

PR #2023, opened 2024-01-25
dialects: Switch memref accesses to assembly format

PapyChacal:
To switch to the assembly format, we need to use a ConstraintVar for the element type, which also gets rid of redundant verification code :)


math-fehr (approved):
<3

superlopuh (approved):
👨‍🍳👌

compor (approved):
less is more :)

webmiche (approved):

PR #2028, opened 2024-01-26
core: Add attribute variables to the declarative assembly format.

math-fehr:
Add attribute variables to the declarative assembly format.
This does not support properties yet, this will be added in a following PR.

The attributes that are parsed by the declarative assembly format are not
parsed/printed back in the attr-dict directive.

### DIFF
--- a/tests/test_declarative_assembly_format.py
+++ b/tests/test_declarative_assembly_format.py
@@ -26,6 +26,7 @@ from xdsl.irdl import (
     ParameterDef,
     VarOperand,
     VarOpResult,
+    attr_def,
     irdl_attr_definition,
     irdl_op_definition,
     operand_def,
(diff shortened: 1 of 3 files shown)


superlopuh (approved):

tests/test_declarative_assembly_format.py
```diff
@@ -245,6 +246,49 @@ class PropOp(IRDLOperation):
...
+def test_attr_variable_shadowed():
+    ctx = MLContext()
+    ctx.load_op(OpWithAttr)
+
+    parser = Parser(ctx, "test.one_attr i32 {attr = i64}")
+    with pytest.raises(
+        ParseError,
+        match="attributes attr are defined in other parts",
+    ):
+        parser.parse_operation()
```
PapyChacal, line 289: This is where MLIR silently does something instead of erroring out, right?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -442,7 +453,7 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
 class ResultTypeDirective(FormatDirective):
     """
     A result variable type directive, with the following format:
-      result-type-directive ::= type(percent-ident)
```
PapyChacal, line 445: :hand_over_mouth:

PapyChacal (approved):
:rocket:

tests/test_declarative_assembly_format.py
math-fehr: Yep, MLIR will override one of them!

PapyChacal: Alright! Let's hope we can peacefully enforce that then ahah

PR #2031, opened 2024-01-26
core: properties in declarative format

math-fehr:
This PR enables properties in the declarative format.

By default, all properties should be specified in the format.
However, with the `ParsePropInAttrDict` IRDL option, properties are instead parsed by the attribute dictionary, to keep compatibility with MLIR.


superlopuh:
should this be based on the other PR?

math-fehr:
Good point, done now

xdsl/irdl/irdl.py
```diff
@@ -1987,9 +1997,7 @@ def irdl_op_init(
                             f"Unexpected option {option} in operation definition {op_def}."
                         )
             case _:
-                raise ValueError(
-                    f"Unexpected option {option} in operation definition {op_def}."
-                )
```
superlopuh, line 1992: Why this change?

PapyChacal (approved):

PR #2032, opened 2024-01-27
core: Implement BytesAttr

PapyChacal:
Here's a non-intrusive variant to be able to roundtrip more general string literals as MLIR does.
The idea is to have another bytes literal, and an extra BytesAttr builtin attribute, to handle cases breaking Python's `str`.

It parses all string literals, escaping those binary escape sequences as MLIR does; if the result is Unicode-decodable, then it's a string literal! (@AntonLydike, yup, this thing handles escaped emojis :wink:)
If it is *not* Unicode-decodable, then it's a BytesAttr! This is good enough for my needs so far, and doesn't make a mess across the whole codebase :slightly_smiling_face:.

At some point, the question might arise for an operation expecting a general BytesAttr but receiving something Unicode-decodable... But I say let's worry about that when it shows up? It might just never occur :thinking:


xdsl/utils/lexer.py
```diff
@@ -174,7 +173,31 @@ def from_span(cls, span: Span | None) -> StringLiteral | None:
     @property
     def string_contents(self):
         # TODO: is this a hack-job?
```
superlopuh, line 175: No more :)

superlopuh (approved):

xdsl/utils/lexer.py
```diff
@@ -201,6 +223,7 @@ class Kind(Enum):
         FLOAT_LIT = object()  # 1.0
         INTEGER_LIT = object()  # 1
         STRING_LIT = object()  # "foo"
+        BYTES_LIT = object()  # "foo"
```
AntonLydike, line 226: I'd put something like `"foo\00\00" in the comment?

AntonLydike (approved):
Nice stuff!

PR #2036, opened 2024-01-29
dialects: Add memref.expand_shape and memref.collapse_shape ops

JosseVanDelm:
Simple PR to add support for:
* https://mlir.llvm.org/docs/Dialects/MemRef/#memrefexpand_shape-memrefexpandshapeop
* https://mlir.llvm.org/docs/Dialects/MemRef/#memrefcollapse_shape-memrefcollapseshapeop

Note that these two ops have quite a bit of verification happening in upstream, which this PR does not implement.

### DIFF
--- a/xdsl/dialects/memref.py
+++ b/xdsl/dialects/memref.py
@@ -379,28 +379,31 @@ ReassociationAttr = ArrayAttr[
 ]
 
 
+class AlterShapeOp(IRDLOperation):
+    src: Operand = operand_def(MemRefType)
+    result: OpResult = result_def(MemRefType)
+    reassociation = prop_def(ReassociationAttr)
+    assembly_format = (
+        "$src $reassociation attr-dict `:` type($src) `into` type($result)"
+    )
+
+
 @irdl_op_definition
-class CollapseShapeOp(IRDLOperation):
+class CollapseShapeOp(AlterShapeOp):
     """
     https://mlir.llvm.org/docs/Dialects/MemRef/#memrefcollapse_shape-memrefcollapseshapeop
     """
 
     name = "memref.collapse_shape"
-    src: Operand = operand_def(MemRefType)
-    result: OpResult = result_def(MemRefType)
-    reassociation = prop_def(ReassociationAttr)
 
 
 @irdl_op_definition
-class ExpandShapeOp(IRDLOperation):
+class ExpandShapeOp(AlterShapeOp):
     """
     https://mlir.llvm.org/docs/Dialects/MemRef/#memrefexpand_shape-memrefexpandshapeop
     """
 
     name = "memref.expand_shape"
-    src: Operand = operand_def(MemRefType)
-    result: OpResult = result_def(MemRefType)
-    reassociation = prop_def(ReassociationAttr)
 
 
 @irdl_op_definition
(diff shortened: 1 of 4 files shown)


xdsl/dialects/memref.py
```diff
@@ -371,6 +372,22 @@ def from_memref(memref: Operation | SSAValue):
         return Rank.build(operands=[memref], result_types=[IndexType()])
 
 
+@irdl_op_definition
+class CollapseShapeOp(IRDLOperation):
+    name = "memref.collapse_shape"
+    src: Operand = operand_def(MemRefType)
+    result: OpResult = result_def(MemRefType)
+    reassociation: AnyArrayAttr = prop_def(AnyArrayAttr)
```
JosseVanDelm, line 380: @superlopuh How can I restrict this to check only for i64 `ArrayAttr`s?

superlopuh: I think the best we can do is this:
```suggestion
    reassociation = prop_def(ArrayAttr[IntegerAttr[IntegerType]])
```
And then still have a manual verify impl that checks for 64 bits

superlopuh (commented):
Could you please add these to the MLIR interop tests, and try adding the custom format? I think with the new format string features that should be pretty easy

xdsl/dialects/memref.py
PapyChacal: I think you can replace IntegerAttr by Annotated[IntegerAttr, IntegerAttr(i64)]

JosseVanDelm:
>  think with the new format string features that should be pretty easy
What new format string features @superlopuh ?

superlopuh:
We've added some features to the assembly_format spec. I can quickly migrate Toy to this to have a range of examples

superlopuh:
Ah no, Toy seems a little trickier. Maybe the tests are the best reference for now, can you search for `assembly_format = "`?

superlopuh:
Could you please add the links to the memref ops to the doc strings of the ops?

xdsl/dialects/memref.py
JosseVanDelm:
```suggestion
     reassociation = prop_def(ArrayAttr[Annotated[IntegerType, IntegerType(64)]])
```
Seems to be doing the trick! Thanks for the suggestions!

JosseVanDelm: Okay nope it didn't anyway, I think I fixed it now?

JosseVanDelm:
Awaiting #2031

JosseVanDelm:
@superlopuh Thanks for giving this tip. 
Super easy to add custom printing (and parsing!) this way!
I did the following:
* Added custom parsing/printing
* Changed the type to `ArrayAttr[ArrayAttr[IntegerAttr[Annotated[IntegerType, IntegerType(64)]]]` because this always seems (I'm not entirely sure!) to be a 2 level nested array? 
* Made a base class AlterShapeOp (might be overkill though :D, let me know what you think!)

superlopuh (approved):
Beautiful

compor:
Having a look in a bit! Thanks for this!

compor (approved):
LGTM, but can you please add a tracking issue regarding the lack of verification?
I find it hard to keep track of "completeness" with MLIR and this makes it slightly more obvious.

Thanks again!

JosseVanDelm:
Sure thing @compor !

PR #2038, opened 2024-01-29
transformations: add lower-linalg-to-loops to xDSL

### LINKED ISSUE #2010: add lowering from linalg.generic to perfectly nested scf loops

superlopuh:
This should have the same behaviour as the MLIR version of this pass. We would like to copy this implementation to have a foundation on which to build slightly modified lowerings, such as lowering to "rof" loops that count down, or to imperfectly nested loops that do not repeatedly write to and read from values that are reduced over.

resolves #2010


AntonLydike:
I don't know enough about the ins and outs of the affine stuff, I think @PapyChacal is a much better candidate for reviewing this

compor:
Just a heads up, this will take me a bit to review and I want to see it a bit more in detail.
Having a cursory look seems OK.

Moreover, I think the wider audience might benefit from a bit clearer PR message with what you/we have in mind by the "slightly modified" lowerings, even if nothing is concrete ATM.
Regardless, this might be a good place to discuss this, without this context, we are simply copying MLIR behaviour and there's nothing to tip the scale on going with this or not.

PapyChacal (approved):
I am not familiar with all the corner cases of linalg.generic itself; but this looks sound for what you're trying to do, and I think it's okay to just move on with this for now, find out later if this is wrong in cases we're not currently interested in either way :slightly_smiling_face: 
Had a call with Sasha on this, my suggestion would be to check the corresponding cases against mlir-opt directly, and probably to test helpers like `get_static_loop_ranges` more extensively before moving one to more complex cases, as I think they are the error-prone bits.

xdsl/transforms/convert_linalg_to_loops.py
```diff
@@ -0,0 +1,152 @@
...
+from xdsl.pattern_rewriter import (
+    GreedyRewritePatternApplier,
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+def indices_for_map(
```
compor, line 23: What does this method do? Some documentation would help; specifically on the `else` bit where unused dimensions are filtered out. I guess this dictates if a load/store ends up being indexed (i.e., being dependent on) by one of the resulting loops.

xdsl/transforms/convert_linalg_to_loops.py
```diff
@@ -0,0 +1,152 @@
...
+
+class LowerGenericOpPattern(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: linalg.Generic, rewriter: PatternRewriter) -> None:
+        if op.res:
+            raise NotImplementedError(
+                "lowering for linalg.generic with results not yet supported"
+            )
+
+        # Create loop nest
```
compor, line 70: I don't think this is where the loop nest is created, more like the bounds for it, right?

xdsl/transforms/convert_linalg_to_loops.py
```diff
@@ -0,0 +1,152 @@
...
+            rewriter.insert_op_before_matched_op((zero_op, one_op))
+
+        index = IndexType()
+
+        # Insert loop nest
+
+        loop_args: list[BlockArgument] = []
+        insertion_target: Operation = op
+
+        for ub in bound_constant_values:
```
compor, line 94: Can you document in which direction the loop nest is built? It seems outermost to innermost to me

xdsl/transforms/convert_linalg_to_loops.py
```diff
@@ -0,0 +1,152 @@
...
+            output_indexing_maps, linalg_yield_op.operands, output_operands, strict=True
+        ):
+            affine_map = affine_map_attr.data
+            indices = indices_for_map(rewriter, insertion_target, affine_map, loop_args)
+            store_op = memref.Store.get(yield_value, ref, indices)
+            rewriter.insert_op_before(store_op, linalg_yield_op)
+
+        rewriter.erase_op(linalg_yield_op)
+
+        # Inline generic body
```
compor, line 136: not fond of comments that explain obvious oneliners.

compor (approved):

PR #2045, opened 2024-01-29
interactive : command line additions

dshaaban01:
with this PR, it will be possible to call `xdsl-gui file_path` in the command line, and it should pre-load the file into the input_text area. 

Next PR: `xdsl-gui file_path -p=pass(es)` will be possible, i.e. passes will be pre-loaded into the gui.


xdsl/interactive/app.py
```diff
@@ -157,7 +158,15 @@ class InputApp(App[None]):
...
 
-    def __init__(self):
+    pre_loaded_file_path_option: str | None = None
+    """
+    Saves the path name of the file called on terminal to be pre-loaded into the gui.
+    """
+
+    def __init__(
+        self,
+        pre_load_file_path: str | None,
```
superlopuh, line 168:
```suggestion
        input_text: str | None = "",
```
I would recommend having the file reading logic outside of the app for now. If there's an error, might as well not launch the app at all

xdsl/interactive/app.py
```diff
@@ -157,7 +158,15 @@ class InputApp(App[None]):
     text areas.
     """
 
-    def __init__(self):
+    pre_loaded_file_path_option: str | None = None
```
superlopuh, line 161:
```suggestion
    input_text: str
```
In the init, this should probably be initialised to `INITIAL_IR_TEXT if self.input_text is None else self.input_text`

superlopuh (changes requested):
Thank you! Couple of small changes

superlopuh (approved):
LGTM

Issue #2040, opened 2024-01-29
add riscv_scf.rof


superlopuh:

AntonLydike:
Completed by the merge of #2048

PR #2048, opened 2024-01-30
dialects: (riscv_scv) Add reverse order for (rof) op

AntonLydike:
Adds the `riscv_scf.rof` (reverse order for) operation.

MLIRs for loops have the implicit constraint that `lb < ub`, and `step > 0`.

In order to express loops that count down in higher-level IR, the rof op
is needed.

ROF has the semantics of going from ub to lb, decrementing by step each time.

In order to convert a for to a rof, one needs to switch lb and ub.
(for the normalized case that `(ub - lb) % step == 0`)

### DIFF
--- a/xdsl/dialects/riscv_scf.py
+++ b/xdsl/dialects/riscv_scf.py
@@ -40,13 +40,10 @@ from xdsl.utils.exceptions import VerifyException
 class YieldOp(AbstractYieldOperation[RISCVRegisterType]):
     name = "riscv_scf.yield"
 
-    traits = traits_def(lambda: frozenset([IsTerminator(), HasParent(ForOp, WhileOp)]))
+    traits = traits_def(lambda: frozenset([IsTerminator(), HasParent(WhileOp, ForRofBaseOp)]))
 
 
-@irdl_op_definition
-class ForOp(IRDLOperation):
-    name = "riscv_scf.for"
-
+class ForRofBaseOp(IRDLOperation):
     lb: Operand = operand_def(IntRegisterType)
     ub: Operand = operand_def(IntRegisterType)
     step: Operand = operand_def(IntRegisterType)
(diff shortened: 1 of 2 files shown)


xdsl/dialects/riscv_scf.py
```diff
@@ -40,13 +40,10 @@
...
 
-    traits = traits_def(lambda: frozenset([IsTerminator(), HasParent(ForOp, WhileOp)]))
+    traits = traits_def(lambda: frozenset([IsTerminator(), HasParent(WhileOp, ForRofBaseOp)]))
 
 
-@irdl_op_definition
-class ForOp(IRDLOperation):
-    name = "riscv_scf.for"
-
+class ForRofBaseOp(IRDLOperation):
```
superlopuh, line 48:
```suggestion
class ForRofOperation(IRDLOperation, abc.ABC):
```
We usually use Operation for abstract superclasses

xdsl/dialects/riscv_scf.py
```diff
@@ -189,6 +186,32 @@ def parse(cls, parser: Parser) -> Self:
...
+
+    In order to express loops that count down in higher-level IR, the rof op
+    is needed.
+
+    ROF has the semantics of going from lb to ub, decrementing by step each time.
+
+    In order to convert a for to a rof, one needs to switch lb and ub.
+    (for the normalized case that (ub - lb) % step == 0)
+    """
+    name = "riscv_scf.rof"
```
superlopuh: I'm not sure I like this, I think for the custom syntax especially it's confusing. I'd say we want the new custom syntax to be 
`riscv_scf.rof %i : !riscv.reg<> = %ub down to %lb by %step ... {`
The `down to` and switching of ub and lb will hopefully be a little clearer to the reader. I would say that the step still needs to be >0 for both loops.

superlopuh (changes requested):

xdsl/dialects/riscv_scf.py
```diff
@@ -189,6 +186,32 @@ def parse(cls, parser: Parser) -> Self:
...
+
+    MLIRs for loops have the implicit constraint that lb < ub, and step > 0.
+
+    In order to express loops that count down in higher-level IR, the rof op
+    is needed.
+
+    ROF has the semantics of going from lb to ub, decrementing by step each time.
+
+    In order to convert a for to a rof, one needs to switch lb and ub.
+    (for the normalized case that (ub - lb) % step == 0)
```
compor, line 213: Just noticed that we do not do any sanity checking for the bounds and step, or at least not as extensive as in `scf.for` and definitely lacking overall against MLIR.

Just noting this for now, I'm OK to go as we are!

compor (approved):

xdsl/dialects/riscv_scf.py
AntonLydike: Makes sense!

AntonLydike: But that's just commentary on the custom format? Are the semantics okay? I'll clarify a bit, but it's essentially the same semantics as `for`, so `lb < ub` and `step > 0`.

xdsl/dialects/riscv_scf.py
AntonLydike: Not sure how well that works, as this may not be compile time information?

AntonLydike:
Fixes #2040

superlopuh (approved):
Perfect, thank you! Maybe could clarify the doc string a little, but I don't think that's critical right now

xdsl/dialects/riscv_scf.py
compor: It doesn't matter much ATM, there are other inconsistencies between the various flavours of our `for`s, but this is a good start; that's where I'm getting at.

PR #2049, opened 2024-01-30
transforms: (riscv_cf) Add canonicalization for branch ops with constant inputs

AntonLydike:
This patch:

- Adds a small bit of machinery to constant-evaluate a set of SSA values using the interpreter
- Adds a pattern that replaces `riscv_cf` conditional branches by unconditional ones if they can be compile time evaluated
- Gives `riscv.li` the `ConstantLike` trait
- Adds constant "is branch taken" evaluation methods to conditional branch ops in `riscv_cf` dialect (this can probably be done better. Maybe a trait?)

I think we want a few more filechecks for this, but otherwise it should be good?


AntonLydike:
Very interesting, I thought I had fixed the filecheck before pushing lol. Should work now.

xdsl/transforms/canonicalization_patterns/riscv_cf.py
```diff
@@ -0,0 +1,114 @@
...
+from xdsl.dialects.riscv_cf import BranchOp, ConditionalBranchOperation, JOp
+from xdsl.interpreter import Interpreter
+from xdsl.interpreters import arith, riscv, riscv_debug
+from xdsl.ir import Operation, SSAValue
+from xdsl.pattern_rewriter import PatternRewriter, RewritePattern
+from xdsl.traits import ConstantLike, Pure
+from xdsl.utils.exceptions import InterpretationError
+
+
+class ElideConstantBranches(RewritePattern):
```
superlopuh, line 14: While this works, I feel like it's adding too much infrastructure for what we need. I also don't feel comfortable adding the interpreter to this lowering. Could you please try with an approach more like in the riscv canonicalization patterns, that checks for the inputs being immediates, and matching on the type of the op?

superlopuh (changes requested):

xdsl/transforms/canonicalization_patterns/riscv_cf.py
compor: I'll have to agree here; while this is a cool use, going for the usual approach might even make this PR shorter (?)

AntonLydike: But that doesn't work for the provided filecheck (because the value is "moved"). Implementing this in a manual way would either duplicate a lot of code, or result in inflexible, mess and buggy code imo.

AntonLydike:
I gave it a go, switching away from the interpreter saved 15 lines, at the cost of losing all of the arithmetic abilities. I don't think that's worth it imo.

(PR is updated)

xdsl/transforms/canonicalization_patterns/riscv_cf.py
```diff
@@ -0,0 +1,99 @@
...
+            # check that the evaluation was successful
+            if not evaluated_inputs.success:
+                return ConstEvalResult(None)
+            assert evaluated_inputs.value is not None
+            results.extend(evaluated_inputs.value)
+        else:
+            # if op is neither ConstantLike nor Pure, fail
+            return ConstEvalResult(None)
+
+    return ConstEvalResult(tuple(results))
```
superlopuh, line 99: It feels like this makes things more complicated than necessary. We might as well add the MVOp canonicalization pattern, and only check for LiOp directly in the rewrite, I don't think we benefit from ConstEvalResult either

AntonLydike: If we don't need to check for `mv` ops, then we should remove that from the filecheck. I thought following through moves was required, as it was part of the provided filecheck case.

AntonLydike:
I cleaned this up further. I am hesitating to remove handling of the `mv` operation, as it would break the filecheck.

xdsl/transforms/canonicalization_patterns/riscv_cf.py
superlopuh: The filecheck is our exact use case, so we need to make it work, although we likely should add more, and better targeted ops.

The way I see to make mv work is to add a canonicalisation pattern to it, to convert it to a li. I don't see the downside of doing this, and it would improve all the other existing canonicalisation patterns.

AntonLydike: We have a canonicalization pattern for `mv`, but that only removes allocated moves from the the same named registers. I don't know if we can just remove mvs if they are unallocated. That feels risky and out of the scope of this PR. Let's discuss on zulip.

xdsl/dialects/riscv_cf.py
```diff
@@ -240,6 +270,10 @@ class BltuOp(ConditionalBranchOperation):
 
     name = "riscv_cf.bltu"
 
+    def const_evaluate(self, rs1: int, rs2: int) -> bool:
+        # TODO: use proper bitwidth here?
+        return to_unsigned(rs1, 32) < to_unsigned(rs2, 32)
```
superlopuh, line 275: This is what makes this complicated, we need to add an xlen to the rewrite, meaning also to the canonicalization infrastructure.

superlopuh: TBH I think the rewritePattern base class should have an indexwidth property

superlopuh (approved):

xdsl/dialects/riscv_cf.py
AntonLydike: Interesting thought, let's put an issue up for this?

superlopuh: #2060

PR #2053, opened 2024-01-30
interactive: add file name to query

dshaaban01:
add file name to query so that when user copies the query, the file name is automatically included.


xdsl/interactive/app.py
```diff
@@ -159,10 +159,11 @@ class InputApp(App[None]):
     """
 
     pre_loaded_input_text: str
-    pre_loaded_pass_pipeline: tuple[tuple[type[ModulePass], PipelinePassSpec], ...]
```
superlopuh, line 162: doesn't seem right

xdsl/interactive/app.py
```diff
@@ -159,10 +159,11 @@ class InputApp(App[None]):
...
+    pre_loaded_file_path: str
```
superlopuh, line 162: I would say that when the user selects a file from within the app, this should also be updated, so this should be more like `file_path: str | None`, and we should have appropriate logic to set it to None when the user modifies the text in the input, until we have a way to save the file

xdsl/interactive/app.py
```diff
@@ -413,14 +419,18 @@ def get_query_string(self) -> str:
         Function returning a string containing the textual description of the pass
         pipeline generated thus far.
         """
-        query = ""
+        if self.pre_loaded_file_path == "":
+            query = "-p "
+        else:
+            query = self.pre_loaded_file_path + " -p "
+
         if self.pass_pipeline != ():
```
superlopuh, line 427:
```suggestion
        if self.pass_pipeline:
```

xdsl/interactive/app.py
superlopuh: Actually let's skip the logic for setting it to None, but it would be good to update the file name in the gui

xdsl/interactive/app.py
dshaaban01: It does seem right to me?

xdsl/interactive/app.py
```diff
@@ -554,6 +565,8 @@ def check_load_file(file_path: str) -> None:
                     with open(file_path) as file:
                         file_contents = file.read()
                         self.input_text_area.load_text(file_contents)
+                    self.pre_loaded_file_path = file_path
```
superlopuh, line 568:
```suggestion
                    self.current_file_path = file_path
```
It feels like the name is a bit misleading, if we update it when the user interacts with the app, then `pre_loaded` no longer applies.

compor (approved):
LGTM module pending comments

PR #2054, opened 2024-01-30
core: make ParametrizedAttribute parameters a tuple

superlopuh:
Two problems with the current code:
1. the class is declared as frozen, but a list makes that false, it's mutable in a hidden way
2. the fact that it's frozen means a hash function is generated, but python crashes when it tries to hash a list

### DIFF
--- a/tests/test_attribute_definition.py
+++ b/tests/test_attribute_definition.py
@@ -249,7 +249,7 @@ class BoolWrapperAttr(ParametrizedAttribute):
 
 def test_bose_constraint():
     """Test the verifier of a base attribute type constraint."""
-    attr = BoolWrapperAttr([BoolData(True)])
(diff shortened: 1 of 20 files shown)


compor (approved):

AntonLydike (approved):
Nice work!

superlopuh:
After some discussions with @math-fehr, I rolled back a fair number of changes, to keep the current API working. It means we incur the cost of creating a new tuple with every init, but I'm hoping to change the way this class works in the future, so would make sense to keep the number of API changes to 1.

tests/test_attribute_definition.py
superlopuh, line 1: some of the changes in this file were necessary due to pyright, so I migrated the whole thing, I hope that's ok

math-fehr (approved):
Very nice!

PR #2055, opened 2024-01-30
dialects: (riscv) remove SImm and replace with annotated IntegerAttr

superlopuh:
@math-fehr is planning to make all Attributes final, so our subclassing of IntegerAttr needs changing. One of the things that no longer works is the unsigned interpretations of the signed bitwidth values. I'm not sure we really want it TBH, as we can't know whether we're dealing with 32-bit or 64-bit code.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -643,17 +643,17 @@ class RdImmJumpOperation(IRDLOperation, RISCVInstruction, ABC):
     The rd register here is not a register storing the result, rather the register where
     the program counter is stored before jumping.
     """
-    immediate: AnyIntegerAttr | LabelAttr = attr_def(AnyIntegerAttr | LabelAttr)
+    immediate = attr_def(SImm20Attr | LabelAttr)
 
     def __init__(
         self,
-        immediate: int | AnyIntegerAttr | str | LabelAttr,
+        immediate: int | SImm20Attr | str | LabelAttr,
         *,
         rd: IntRegisterType | str | None = None,
         comment: str | StringAttr | None = None,
     ):
         if isinstance(immediate, int):
-            immediate = IntegerAttr(immediate, IntegerType(20, Signedness.SIGNED))
+            immediate = IntegerAttr(immediate, si20)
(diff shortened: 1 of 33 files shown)


compor:
Sorry, but when was this discussed? Is there a timeline of features? I don't recall being discussed on Zulip or the biweekly meeting.

compor (approved):

AntonLydike (approved):
Good stuff!

xdsl/dialects/riscv.py
```diff
@@ -729,20 +707,20 @@ class RdRsImmIntegerOperation(IRDLOperation, RISCVInstruction, ABC):
     This is called I-Type in the RISC-V specification.
     """
 
-    rd: OpResult = result_def(IntRegisterType)
-    rs1: Operand = operand_def(IntRegisterType)
-    immediate: AnyIntegerAttr | LabelAttr = attr_def(AnyIntegerAttr | LabelAttr)
+    rd = result_def(IntRegisterType)
+    rs1 = operand_def(IntRegisterType)
+    immediate = attr_def(IntegerAttr[IntegerType] | LabelAttr)
```
superlopuh, line 712: I just realised that this doesn't actually check the bitwidth, I'll leave it as-is and try and fix in a follow-up PR.

superlopuh:
> Sorry, but when was this discussed? Is there a timeline of features? I don't recall being discussed on Zulip or the biweekly meeting.

Sorry for the confusion, I just saw @math-fehr touching this part of the code as part of his work on attributes, and thought I'd help him out to clean up this dialect I'm more familiar with. It's not related to the rest of the RISC-V work.

math-fehr:
@compor, sorry about that, I probably should have announced the change.
This is like 4 layers of indirections, but the idea is that to enable attributes/properties in the declarative format, we need to make attributes final (which to my opinion they always were, but we never enforced it).
I looked at changing this yesterday, and I found 2 attributes that were actually using inheritance (one is the fastmathflags, the other one is this SImm32).|

math-fehr:
Looks good to me, though it's harder to say now that it sholud be rebased

PR #2057, opened 2024-01-31
transforms: (riscv) Add pattern to convert move of immediate value to immediate instantiation

AntonLydike:
Unblocks #2049 

Expands the `mv` op canonicalization pattern to convert a move of an immediate value to a new instantiation.

This will help simplify other passes as well, and makes the dependency graph simpler.


xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -21,6 +21,17 @@ def match_and_rewrite(self, op: riscv.MVOp, rewriter: PatternRewriter) -> None:
             rewriter.replace_matched_op([], [op.rs])
 
 
+class ImmediateMoveToCopy(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: riscv.MVOp, rewriter: PatternRewriter, /):
+        if (
+            isinstance(op.rd.type, riscv.RISCVRegisterType)
+            and not op.rd.type.is_allocated
```
superlopuh, line 29: I think this should work fine whether or not the result is allocated, as long as the move is to the same register

superlopuh (approved):
TY!

PR #2058, opened 2024-01-31
Add fastmath flags to riscv dialect

compor:
This PR:

- Adds `fastmath` flags to the single- and double-precision variants of `fsub`, `fadd`, `fmul` and `fdiv`.
- Modifies filechecks for the above

In a follow-up patch, we plan to use these flags to enable contraction (fusion) between ops via canonicalization.
This will allow our backend to follow the same conventions as other compilers when producing fused ops (i.e., `fmadd`, etc).

Related to: https://github.com/opencompl/riscv-paper-experiments/issues/137

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -17,6 +17,7 @@ from xdsl.dialects.builtin import (
     StringAttr,
     UnitAttr,
 )
+from xdsl.dialects.llvm import FastMathAttr as LLVMFastMathAttr
 from xdsl.ir import (
     Attribute,
     Block,
@@ -55,6 +56,14 @@ from xdsl.traits import (
 )
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import isa
+
+
+class FastMathFlagsAttr(LLVMFastMathAttr):
+    """
+    riscv.fastmath is a mirror of LLVMs fastmath flags.
+    """
+
+    name = "riscv.fastmath"
 
 
 class RISCVRegisterType(Data[str], TypeAttribute, ABC):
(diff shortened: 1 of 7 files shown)


xdsl/dialects/riscv.py
```diff
@@ -57,6 +58,14 @@
 from xdsl.utils.hints import isa
 
 
+class FastMathFlagsAttr(LLVMFastMathAttr):
```
superlopuh, line 61: My understanding is that this will be disallowed in the near future, and that all attributes will be marked as final. (maybe operations too?) CC @math-fehr. I think we'll need a shared base class for fastmath attributes in some neutral location to get around this.

xdsl/dialects/riscv.py
```diff
@@ -600,6 +609,59 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
         return self.rd, self.rs1, self.rs2
 
 
+class RdRsRsFloatOperationWithFastMath(IRDLOperation, RISCVInstruction, ABC):
```
superlopuh, line 612: Could this subclass `RdRsRsOperation[FloatRegisterType, FloatRegisterType, FloatRegisterType]`? Is there a problem with the attributes being passed in?

compor: OK, I'll change it and see if it complaints

compor:
I also forgot the lowering via arith in another branch, so moving that over too.

xdsl/dialects/riscv.py
superlopuh: I remember it being annoying about something similar, but I'm not sure exactly what the condition was for it to fail. It might be easier to duplicate the classes for now, rather than deal with irdl logic

compor: With

```py
class RdRsRsFloatOperationWithFastMath(RdRsRsOperation[FloatRegisterType, FloatRegisterType, FloatRegisterType]):
```

the same `super().__init__` call complains that arguments are missing for `rs1`, `rs2` and `rd`.

Not sure what irks it.

xdsl/dialects/riscv.py
math-fehr: I'll fix it after this is merge if you prefer, but yes the idea is to have a `FastMathAttrBase` that these operations will inherit!
Otherwise, with this current implementation, you could use the `riscv` `FastMathFlagsAttr` in LLVM and that would still work

superlopuh (approved):

xdsl/dialects/riscv.py
compor: I can have a look after this is merged, I'll open an issue to have anyone pick it up whenever.

compor: #2064

PR #2061, opened 2024-01-31
core: add replacement arg_values to inline_block_before

superlopuh:
Slightly modifies the semantics of inlining, bringing it in line with MLIR's behaviour, in my understanding. Currently, xDSL empties the block when inlining it to a different block, but does not erase it. With this PR, this behaviour is changed to erasing the block, potentially leaving an empty region.

### DIFF
--- a/tests/pattern_rewriter/test_pattern_rewriter.py
+++ b/tests/pattern_rewriter/test_pattern_rewriter.py
@@ -766,9 +766,8 @@ def test_inline_block_before_matched_op():
   %0 = "test.op"() : () -> !test.type<"int">
   %1 = "test.op"() : () -> !test.type<"int">
   %2 = "test.op"() ({
-  ^0:
   }, {
-  ^1:
+  ^0:
   }) : () -> !test.type<"int">
 }) : () -> ()
 """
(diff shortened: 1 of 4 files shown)


compor:
Is this ready? I see commented code and TODOs.
Can you also please link where that behaviour occurs in MLIR? It saves a bit of time in regexing the codebase?

Thanks

superlopuh:
This is ready, there's just behaviour that we can currently not replicate due to lack of successor/predecessor accessors on Blocks in xDSL. I can add more comments to the commented code? I'll add the link to MLIR implementation.

compor:
> This is ready, there's just behaviour that we can currently not replicate due to lack of successor/predecessor accessors on Blocks in xDSL. I can add more comments to the commented code? I'll add the link to MLIR implementation.

OK, I'll have a look then, is it worth tracking the TODOs as issues?

compor (approved):

math-fehr (approved):
Looks good to me, but I have a question about it.
Why do we want to check for the correctness of successors after inlining?
If we do multiple rewrites at once, I would expect to only verify the correctness of successors at the end of the sequence of rewrites?
Is it what MLIR does?

superlopuh:
Yeah that's fully lifted from the MLIR code base. I'm not 100% sure what all the reasons behind it are, but it feels reasonable for me to leave the commented stubs in so that we have something to go from when we hit a bug in a rewrite that leverages this.

PR #2062, opened 2024-02-01
transformations: (scf) add SimplifyTrivialLoops pattern

### LINKED ISSUE #2008: inline riscv_scf.for loops with single iteration count

superlopuh:
Note stacked PR

Mostly lifted from MLIR

Fixes #2008 (at scf level instead of riscv_scf level, but that's actually better for us)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,8 +12,6 @@
             "justMyCode": true,
             "args": [
                 "${file}",
-                "-p",
-                "canonicalize"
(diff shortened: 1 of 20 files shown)


compor (approved):

math-fehr:
Good for me, but is there any reasons we are reimplementing this rather than using the MLIR one?

superlopuh:
Because we need it to work on both for and rof :)

PR #2073, opened 2024-02-01
testing: add kernels from our paper with verified assembly

superlopuh:
The idea is to grow the compiler to snitch_stream bottom-up, meaning that the target output is fixed, and the input rises in abstraction level over time. This PR introduces both the kernels for which we know exactly what the assembly should be, and a test pass that fixes our compilation pipeline, currently copied from the Makefile.rules in the other repo

### DIFF
--- a/xdsl/transforms/test_lower_linalg_to_snitch.py
+++ b/xdsl/transforms/test_lower_linalg_to_snitch.py
@@ -1,17 +1,3 @@
-# XDSLOPT_PASSES += snitch-allocate-registers
-# XDSLOPT_PASSES += convert-snitch-stream-to-snitch
-# XDSLOPT_PASSES += lower-snitch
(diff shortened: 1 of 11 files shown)


xdsl/transforms/test_lower_linalg_to_snitch.py
```diff
@@ -0,0 +1,45 @@
...
+class TestLowerLinalgToSnitchPass(ModulePass):
+    """
+    A compiler pass used for testing of the lowering from ML kernels defined as
+    linalg.generic operations to riscv-assemby leveraging Snitch cores.
+    """
+
+    name = "test-lower-linalg-to-snitch"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PipelinePass(
```
compor, line 30: :heart:

compor (approved):

PR #2077, opened 2024-02-01
dialects: (riscv) Canonicalize the loading 0 to `li zero, 0`

superlopuh:
Note stacked PR

We need to canonicalize getting the 0 value, so that the other transforms can have simple logic for checking for constant values. Adding this canonicalization pattern by itself messed with the assembly generated downstream, hence the PR this one points to, to get the assembly back to normal. I have run the verilator based on this PR, all the tests pass, most have a lower cycle count. One has a higher cycle count, which makes no sense to me, as the only change is skipping the branch.

### DIFF
--- a/tests/filecheck/backend/riscv/canonicalize.mlir
+++ b/tests/filecheck/backend/riscv/canonicalize.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt -p canonicalize %s | filecheck %s
+// RUN: xdsl-opt --split-input-file -p canonicalize %s | filecheck %s
 
 builtin.module {
   %i0, %i1, %i2 = "test.op"() : () -> (!riscv.reg<a0>, !riscv.reg<a1>, !riscv.reg<>)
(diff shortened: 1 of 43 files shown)


compor (approved):

tests/filecheck/backend/riscv/canonicalize.mlir
```diff
@@ -125,11 +125,11 @@ builtin.module {
 // CHECK-NEXT:   %{{.*}} = riscv.fmv.d %{{.*}} : (!riscv.freg<>) -> !riscv.freg<>
 // CHECK-NEXT:   "test.op"(%f0, %fo1, %fo2, %f0, %fo4, %fo5) : (!riscv.freg<fa0>, !riscv.freg<fa2>, !riscv.freg<>, !riscv.freg<fa0>, !riscv.freg<fa2>, !riscv.freg<>) -> ()
 
-// CHECK-NEXT:   %0 = riscv.li 0 : () -> !riscv.reg<>
+// CHECK-NEXT:   %0 = riscv.li 0 : () -> !riscv.reg<zero>
```
AntonLydike, line 128: We should be able to elide this completely, no?

xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -471,3 +471,39 @@ def match_and_rewrite(
...
                 ),
             )
+
+
+class GetZeroRegister(RewritePattern):
+    """
+    The canonical form of an operation that creates a zero register is `li 0`, so that the
+    downstream patterns can assume that any constant is created by `li`, and don't need
+    more sophisticated checks.
+    To revert to getting the zero register, use `riscv-get-zero-register`
```
AntonLydike, line 481: Hmmmmm, this feels like the wrong approach to me. I'd rather factor out the code that does the cnostant inferrence to handle `zero` registers.

AntonLydike (commented):
I don't think this is the right approach to take here. This feels like adding passes to fix a compile time inference problem...

xdsl/transforms/canonicalization_patterns/riscv.py
superlopuh: I'm not sure that any approach is the wrong aproach, TBH. My understanding of canonicalization is that it's supposed to make things more uniform, as opposed to more efficient. This pass satisfies this quality, in my understanding, making the statement "if it's constant, it's LiOp` true. This feels conceptually simpler than a helper that gets a constant value from a possible range of operations. Ultimately, I'm not sure that either approach is wrong, I think either could work. Am I missing something?

superlopuh:
https://github.com/xdslproject/xdsl/issues/2087

PR #2079, opened 2024-02-01
backend (riscv): Add canonicalization pattern to fuse `fmul.d` and `fadd.d` to `fmadd.d`

compor:
This PR:

- Adds a canonicalization for fusing `fmul.d` with `fadd.d` to `fmadd.d`.
- Tests of the above

Essentially, `c <- a * b` and `d <- e + c` to `d <- e + a * b`.
`c` should not be used anywhere else and both operations must have the `contract` fastmath flag set.

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/riscv.py
+++ b/xdsl/transforms/canonicalization_patterns/riscv.py
@@ -392,11 +392,11 @@ class FuseMultiplyAddD(RewritePattern):
         elif isinstance(op.rs2.type, riscv.RISCVRegisterType):
             addend = op.rs2
             potential_multiplicand1 = op.rs1
+        else:
+            return
 
         if (
-            addend is not None
-            and potential_multiplicand1 is not None
-            and isinstance(mulop := potential_multiplicand1.owner, riscv.FMulDOp)
+            isinstance(mulop := potential_multiplicand1.owner, riscv.FMulDOp)
             and len(mulop.rd.uses) == 1
             and (
                 mulop.fastmath is not None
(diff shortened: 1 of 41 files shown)


xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -376,6 +376,53 @@ def match_and_rewrite(self, op: riscv.AddOp, rewriter: PatternRewriter) -> None:
...
+                and llvm.FastMathFlag.ALLOW_CONTRACT in mulop.fastmath.flags
+            )
+            and isinstance(mulop.rd.type, riscv.RISCVRegisterType)
+        ):
+            rewriter.replace_matched_op(
+                riscv.FMAddDOp(
+                    mulop.rs1,
+                    mulop.rs2,
+                    addend,
+                    comment=op.comment,
```
superlopuh, line 421:
```suggestion
                riscv.FMAddDOp(
                    mulop.rs1,
                    mulop.rs2,
                    addend,
                    rd=op.rd.type,
                    comment=op.comment,
```

superlopuh: This might need another cast/isinstance check. It would be good to have tests for this also.

compor: Ah just saw this, I was working on the tests for the commutative cases.
Is this to transfer the register if we use this after register allocation?

superlopuh: Yeah, there's a possibility that that happens, and then there's the danger of messing with the existing register allocation of the result value

compor: yeah, I was just planning the use of this prior to regalloc, but you're right, an oversight.

superlopuh (approved):

PR #2085, opened 2024-02-04
transformations (riscv_scf): Change forwarded block args of `riscv_cf.blt` to use `riscv_scf.yield` operands when lowering from `riscv_scf` to `riscv_cf`

compor:
Lowering from `riscv_scf.for` to `riscv_cf.blt` seems to use the `iter_args` operands instead of the yielded values.
This seems to render the result from the `for` not used and DCE eliminates the main calculation result of the loop.

This can be reproduced with:

```mlir
builtin.module {
    riscv_func.func @sum_range(%0 : !riscv.reg<a0>, %1 : !riscv.reg<a1>) {
        %2 = riscv.li 1 : () -> !riscv.reg<>
        %3 = riscv.li 0 : () -> !riscv.reg<>
        %4 = riscv_scf.for %5 : !riscv.reg<> = %0 to %1 step %2 iter_args(%6 = %3) -> (!riscv.reg<>) {
            %7 = riscv.add %5, %6 : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
            riscv_scf.yield %7 : !riscv.reg<>
        }
        %8 = riscv.mv %4 : (!riscv.reg<>) -> !riscv.reg<>
        riscv_func.return %8 : !riscv.reg<>
    }
}
```

lowered with `xdsl-opt -p riscv-allocate-registers,convert-riscv-scf-to-riscv-cf,canonicalize`.

I think the issue was not triggered because the filechecks for this lowering were manually regalloc'd (but not 100% sure).
I've tested this on the riscv experiments repo and seems to work.

### DIFF
--- a/tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
+++ b/tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
@@ -66,7 +66,7 @@ builtin.module {
 // CHECK-NEXT:      riscv.label "scf_body_0_for"
 // CHECK-NEXT:      %7 = riscv.add %5, %6 : (!riscv.reg<a4>, !riscv.reg<a3>) -> !riscv.reg<a3>
 // CHECK-NEXT:      %8 = riscv.add %5, %2 : (!riscv.reg<a4>, !riscv.reg<a2>) -> !riscv.reg<a4>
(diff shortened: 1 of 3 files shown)


superlopuh (approved):
thank you

PR #2092, opened 2024-02-05
dialects: (riscv_scf) make riscv_scf.yield implicit

superlopuh:


compor (approved):
Thanks, been wanting to do this for a while.

PR #2093, opened 2024-02-05
interactive: adding an IndividualRewrite pass

dshaaban01:
PR 1 out of 4 into getting peephole rewrites to work. 

The way we will do it is by creating a pass that represents an individual peephole rewrite. This enables us to reuse the app's infrastructure (which is pass-based).

### DIFF
--- a/tests/filecheck/transforms/individual_rewrite.mlir
+++ b/tests/filecheck/transforms/individual_rewrite.mlir
@@ -1,0 +1,19 @@
+// RUN: xdsl-opt -p 'apply-individual-rewrite{matched_operation_index=4 operation_name="arith.addi" pattern_name="AddImmediateZero"}'| filecheck %s
+
+func.func @hello(%n : i32) -> i32 {
+  %two = arith.constant 0 : i32
+  %three = arith.constant 0 : i32
+  %res = arith.addi %two, %n : i32
+  %res2 = arith.addi %three, %res : i32
+  func.return %res : i32
+}
+
+
+//CHECK:         builtin.module {
+// CHECK-NEXT:     func.func @hello(%n : i32) -> i32 {
+// CHECK-NEXT:       %two = arith.constant 0 : i32
+// CHECK-NEXT:       %three = arith.constant 0 : i32
+// CHECK-NEXT:       %res2 = arith.addi %three, %n : i32
+// CHECK-NEXT:       func.return %res2 : i32
+// CHECK-NEXT:     }
+// CHECK-NEXT:  }
--- a/tests/filecheck/transforms/individual_rewrite.mlir
+++ b/tests/filecheck/transforms/individual_rewrite.mlir
@@ -1,19 +1,61 @@
-// RUN: xdsl-opt -p 'apply-individual-rewrite{matched_operation_index=4 operation_name="arith.addi" pattern_name="AddImmediateZero"}'| filecheck %s
+// RUN:xdsl-opt %s -p 'apply-individual-rewrite{matched_operation_index=8 operation_name="riscv.add" pattern_name="AddImmediates"}'| filecheck %s
 
-func.func @hello(%n : i32) -> i32 {
-  %two = arith.constant 0 : i32
-  %three = arith.constant 0 : i32
-  %res = arith.addi %two, %n : i32
-  %res2 = arith.addi %three, %res : i32
-  func.return %res : i32
+riscv_func.func @at_least_once() {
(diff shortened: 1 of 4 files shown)


xdsl/transforms/individual_rewrite.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass
+
+from xdsl.dialects.builtin import ModuleOp
+from xdsl.ir import MLContext
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import PatternRewriter, RewritePattern
+from xdsl.tools.command_line_tool import get_all_dialects
+from xdsl.traits import HasCanonicalisationPatternsTrait
+
+REWRITE_BY_NAMES: dict[str, dict[str, RewritePattern]] = {
```
compor, line 10: hmm, this feels like the optimization machinery within xDSL should be a bit more introspectable by construction.
I understand this is the minimum effort that can be done ATM without overhauling that part of the framework.

Maybe it's worth having it as a discussion some time soon.

superlopuh, line 10: Could you please add a docstring?

superlopuh (changes requested):
Can you please add two filecheck tests to exercise this? One that succeeds, and one that fails?

xdsl/transforms/individual_rewrite.py
```diff
@@ -0,0 +1,37 @@
...
+    }
+    for dialect in get_all_dialects().values()
+    for op in dialect().operations
+    if (trait := op.get_trait(HasCanonicalisationPatternsTrait)) is not None
+}
+
+
+@dataclass
+class IndividualRewrite(ModulePass):
+    name = "individual-rewrite"
```
math-fehr, line 23: I'm not a big fan of the name, usually we have verbs to express what a pass is doing?
Like `apply-individual-rewrite`?

xdsl/transforms/individual_rewrite.py
```diff
@@ -0,0 +1,37 @@
...
+    matched_operation_index: int | None = None
+    operation_name: str | None = None
+    pattern_name: str | None = None
+
+    def apply(self, ctx: MLContext, op: ModuleOp) -> None:
+        assert self.matched_operation_index is not None
+        assert self.operation_name is not None
+        assert self.pattern_name is not None
+
+        matched_operation = list(op.walk())[self.matched_operation_index]
```
math-fehr, line 34: Can you check that the index exists, and return a proper error otherwise?

xdsl/transforms/individual_rewrite.py
```diff
@@ -0,0 +1,37 @@
...
+        rewriter = PatternRewriter(matched_operation)
+        pattern = REWRITE_BY_NAMES[self.operation_name][self.pattern_name]
```
math-fehr, line 36: Same here with the rewrite name

math-fehr (commented):
Nice!

xdsl/transforms/individual_rewrite.py
dshaaban01: Agree. Would you like to join this Thursday's weekly interactive meeting to talk about it? :) It's at 14:00 UK time.

math-fehr:
> Can you please add two filecheck tests to exercise this? One that succeeds, and one that fails?

I actually don't think we can have a "failing" case here.
At least, there is no way of using filecheck with exceptions that are not `VerifyException` :/
Should we skip those tests for now, or what should we do for that? I don't have time until the end of the week to add support for other exceptions at least

math-fehr (approved):

tests/filecheck/transforms/individual_rewrite.mlir
superlopuh, line 1: could you please change it to something simpler? maybe something like this:
```
%a = riscv.li 1 : () -> !riscv.reg<>
%b = riscv.li 2 : () -> !riscv.reg<>
%c = riscv.li 3 : () -> !riscv.reg<>
%d = riscv.add %a, %b : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
%e = riscv.add %b, %c : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
```
and apply addimmediates to %e (which I think should be op number 6

This filecheck has quite a lot of code so it's difficult to see what's happening

dshaaban01: Done. (changed %d though, not e but I think its okay?)

superlopuh (approved):

PR #2094, opened 2024-02-05
dialects: (arith) adding an arith canonicalisation pattern 

dshaaban01:
PR 2 out of 4. 

Added an arith canonicalisation pattern to use as an example/ starting point to extending arith canonicalisation patterns.


superlopuh (changes requested):
can you please add a filecheck test for this?

tests/filecheck/transforms/arith-add-immediate-zero.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: xdsl-opt %s -p 'apply-individual-rewrite{matched_operation_index=5 operation_name="arith.addi" pattern_name="AddImmediateZero"}'| filecheck %s
```
math-fehr, line 1:
```suggestion
// RUN: xdsl-opt %s -p canonicalization | filecheck %s
```

math-fehr (approved):

superlopuh (approved):

xdsl/transforms/individual_rewrite.py
```diff
@@ -0,0 +1,68 @@
...
+
+        pattern = rewrite_dictionary.get(self.pattern_name)
+        if pattern is None:
+            raise ValueError(
+                f"Pattern name {self.pattern_name} not found for the provided operation name."
+            )
+
+        pattern.match_and_rewrite(matched_operation, rewriter)
+        if not rewriter.has_done_action:
+            raise ValueError("Invalid rewrite at current location.")
```
webmiche, line 68: To me, this feels like a case that should have a warning instead of an exception. It feels like this might make it hard to interact with this pass through scripts and stuff.
Though, I guess people consider exceptional control flow for basic stuff acceptable in Python, so this is just a side comment.

PR #2096, opened 2024-02-05
interactive: add functions that make xdsl-gui add rewrite functionality to tool

dshaaban01:
PR 4 out of 4. 

In this final PR, we add functions that compute the list of "current_possible_rewrites" (i.e. rewrites that apply to our input text). these rewrites are added to the "list view", and we adapt the function to correctly update the pass_pipeline if a rewrite is selected to applied in a way that leverages the "Individual Rewrite" pass that we created previously. 

We have thus been able to reuse the pass infrastructure for rewrites. :)) huge win.


xdsl/interactive/app.py
```diff
@@ -165,14 +165,19 @@ class InputApp(App[None]):
...
-    Reactive variable that saves the list of passes that have an effect on
-    current_module.
+    Reactive variable that saves the passes that have an effect on current_module.
+    """
+    current_possible_rewrites = reactive(
+        tuple[tuple[int, tuple[str, str, RewritePattern], ModuleOp]]
+    )
+    """
+    Reactive variable that saves the rewrite patterns that have an effect on current_module.
     """
```
superlopuh, line 181: Conceptually, the new rewrites are not any different to the passes, both because they end up in the pass pipeline in the top right quarter, but also because they'll all be represented as a pass in the `pass_pipeline`. I feel like it would be worth spending a bit of time now to make them more uniform at the input stage. I would propose something like this:
```
    condense_mode = reactive(False, always_update=True)
    show_individual_rewrites_mode = reactive(False, always_update=True)
    available_pass_list = reactive(tuple[tuple[type[ModulePass], PipelinePassSpec|None], ...])
```
Where the available pass list contains both the passes (with None instead of a PipelinePassSpec, opening the window if pass parameters are required) and the individual rewrites, where the rewrite info is stored in the PipelinePassSpec in the same element of the tuple.

superlopuh: It would probably make sense to make a separate PR to prepare for this change, and rebase this one on top of that

dshaaban01: I can do a separate PR in future to add a button that does a "rewrite mode" so to speak.

xdsl/interactive/app.py
```diff
@@ -165,14 +170,15 @@ class InputApp(App[None]):
...
     pass_pipeline = reactive(tuple[tuple[type[ModulePass], PipelinePassSpec], ...])
-    """Reactive variable that saves the list of selected passes."""
+    """Reactive variable that saves the selected passes."""
 
     condense_mode = reactive(False, always_update=True)
     """Reactive boolean."""
-    available_pass_list = reactive(tuple[type[ModulePass], ...])
+    available_pass_list = reactive(
+        tuple[tuple[str, type[ModulePass], PipelinePassSpec | None], ...]
```
superlopuh, line 178: Can you please factor out this tuple?
```suggestion
        tuple[AvailablePass, ...]
```

``` python

class AvailablePass(NamedTuple):
    display_name: str
    module_pass: type[ModulePass]
    pass_spec: PipelinePassSpec | None
```

xdsl/interactive/app.py
```diff
@@ -97,19 +97,24 @@ def get_all_possible_rewrites(
...
+) -> tuple[tuple[str, type[ModulePass], PipelinePassSpec | None], ...]:
     """Returns a tuple of passes (pass name and pass instance) that modify the IR."""
 
     ctx = MLContext(True)
 
     for dialect_name, dialect_factory in get_all_dialects().items():
         ctx.register_dialect(dialect_name, dialect_factory)
 
-    selections: list[type[ModulePass]] = []
+    selections: tuple[tuple[str, type[ModulePass], PipelinePassSpec | None], ...] = ()
```
superlopuh, line 110: adding to a list in a for loop is much more efficient than appending to a tuple, could you factor this back, please?

xdsl/interactive/app.py
```diff
@@ -311,39 +317,74 @@ def on_mount(self) -> None:
...
         """
         match self.current_module:
             case None:
-                return tuple(p for _, p in ALL_PASSES)
+                return tuple((p.name, p, None) for _, p in ALL_PASSES)
             case Exception():
                 return ()
             case ModuleOp():
+                # transform rewrites into passes
+                rewrites = get_all_possible_rewrites(ALL_PATTERNS, self.current_module)
```
superlopuh, line 334: Would it make sense to have this function return a tuple of `AvailablePass`?

PR #2102, opened 2024-02-06
testing: add a timeout for `make filecheck`

superlopuh:
3s seems safe enough

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ clean:
 
 # run filecheck tests
 filecheck:
-	lit -vv tests/filecheck --order=smart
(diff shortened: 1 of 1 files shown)


math-fehr (approved):
Yes, got bitten by this before!
For context to the others, if you don't add this timeout and you have an infinite loop, the processes will stay alive forever, and that's really annoying

webmiche (approved):
Did you have problems? I am fine with this, though I wonder if this might lead to crashes for people with very slow computers...

superlopuh:
We can always increase it. Without it, my problem was Python allocating 320GB of memory :)

PR #2103, opened 2024-02-06
backend: (riscv) allocate a constant value of 0 to ZERO register

superlopuh:
Note stacked PR

If we know that a value is always zero, then we might as well allocate the dedicated register. The canonicalization pass will then elide the li and replace with the GetRegisterOp


xdsl/backend/riscv/register_allocation.py
```diff
@@ -96,7 +98,10 @@ def allocate(self, reg: SSAValue) -> bool:
             isinstance(reg.type, IntRegisterType | FloatRegisterType)
             and not reg.type.is_allocated
         ):
-            reg.type = self.available_registers.pop(type(reg.type))
+            if (val := get_constant_value(reg)) is not None and val.value.data == 0:
```
compor, line 101: Not really fond of this, but let's revisit later.

compor (approved):

PR #2106, opened 2024-02-06
core: Add checks on chaining variadics in declarative assembly format.

PapyChacal:
Just adding more checks and tests for variadics in the declarative assembly format; for example, don't chain multiple variadic type directives; they need a separator for the parser to know when the first ends.

PR #2107, opened 2024-02-06
Implement and test optional operand and result variables.

PapyChacal:
Stacked on #2106.

Add optional operands and type directives; very similar to the variadic ones. Adds the notion of "variadic-like" to relate to checks added in #2106


math-fehr (approved):
Very very nice, thanks a lot <3

PR #2111, opened 2024-02-07
dialects: (memref_stream) add memref_stream dialect with read and write ops

superlopuh:
Note stacked PR

This is leading towards a target-independent representation of Snitch streams.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
     "configurations": [
         {
             "name": "xdsl-opt",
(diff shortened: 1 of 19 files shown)


nazavode (approved):
I like the idea of having our own memory accesses.

math-fehr (approved):
Nice, can you just add a small documentation paragraph on top of the dialect file to just describe what it is supposed to represent?

compor (approved):

PR #2112, opened 2024-02-07
backend: (riscv) add convert-memref-stream-to-snitch pass

superlopuh:
Note stacked PR

Add the lowering for the first two ops in the dialect


tests/filecheck/transforms/convert_memref_stream_to_snitch.mlir
```diff
@@ -0,0 +1,16 @@
+// RUN: xdsl-opt -p convert-memref-stream-to-snitch %s | filecheck %s
+
+%readable, %writable = "test.op"() : () -> (!stream.readable<f32>, !stream.writable<f32>)
+
+%val = memref_stream.read from %readable : f32
+memref_stream.write %val to %writable : f32
+
+// CHECK:       builtin.module {
+// CHECK-NEXT:    %readable, %writable = "test.op"() : () -> (!stream.readable<f32>, !stream.writable<f32>)
+// CHECK-NEXT:    %val = builtin.unrealized_conversion_cast %readable : !stream.readable<f32> to !stream.readable<!riscv.freg<>>
```
nazavode, line 10: I don't know whether we want to take into account the stream transfer size fixed to `FLEN` at this point in the lowering, e.g.: the `!stream.readable<f32>` is in fact fixed by the target (where `FLEN == 64`) to `!stream.readable<vector<2xf32>>`. Probably some trick involving additional registers (e.g. like extracting the lower/upper `f32` parts from a `!stream.readable<vector<2xf32>>`) would be needed when lowering the following at a later stage:

> %{{.*}} = builtin.unrealized_conversion_cast %{{.*}} : !riscv.freg<> to f32

The annoying part is that with `!stream.readable<T>` with `sizeof(T) == FLEN` you have scalar code, while with `sizeof(T) < FLEN` you have vector code.

superlopuh: That's a good point. We'll need both xlen and flen for all the riscv lowerings. I can file an issue so we can revisit once we start dealing with data on f32?

nazavode: Cool, my doubt was in fact whether taking that stuff into account right now would have been useful vs noise.

nazavode (approved):

compor (approved):

PR #2113, opened 2024-02-07
core: Add accessor to basic block

compor:
This PR:

- Adds a block accessor to retrieve operations from within a block using their position index (`get_operation_by_index`)
- Tests of the above

I recently needed to patch cloned IR instructions between the source and the cloned block and couldn't find a better way to do that. `value_mapper` didn't make much sense since I knew exactly which operations I needed to work on from prior analyses.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1615,10 +1615,21 @@ class Block(IRNode):
     def get_operation_index(self, op: Operation) -> int:
         """Get the operation position in a block."""
         if op.parent is not self:
-            raise Exception("Operation is not a children of the block.")
+            raise Exception("Operation is not a child of the block.")
         for idx, block_op in enumerate(self.ops):
             if block_op is op:
                 return idx
+        assert False, "Unexpected xdsl error"
+
+    def get_operation_by_index(self, idx: int) -> Operation:
+        """Get the operation by its position in its parent block."""
+        if idx not in range(0, len(self.ops)):
+            raise Exception(
+                f"Cannot get operation by out-of-bounds index {idx} in its parent block."
+            )
+        for _idx, block_op in enumerate(self.ops):
+            if idx == _idx:
+                return block_op
         assert False, "Unexpected xdsl error"
 
     def detach_op(self, op: Operation) -> Operation:
--- a/tests/test_ir.py
+++ b/tests/test_ir.py
@@ -432,6 +432,42 @@ def test_split_block_args():
     assert arg_types == [i32, i64]
 
 
+def test_block_get_op_by_index():
+    # test block with ops
+    bb0 = Block((op0 := test.TestOp(), op1 := test.TestOp(), op2 := test.TestOp()))
+
+    assert bb0.get_operation_by_index(0) is op0
+    assert bb0.get_operation_by_index(1) is op1
+    assert bb0.get_operation_by_index(2) is op2
+
+    with pytest.raises(
+        Exception,
+        match="Cannot get operation by out-of-bounds index -1 in its parent block.",
+    ):
+        bb0.get_operation_by_index(-1)
+
+    with pytest.raises(
+        Exception,
+        match=f"Cannot get operation by out-of-bounds index {len(bb0.ops)} in its parent block.",
+    ):
+        bb0.get_operation_by_index(len(bb0.ops))
+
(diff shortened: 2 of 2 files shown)


xdsl/ir/core.py
```diff
@@ -1615,12 +1615,23 @@ def split_before(
     def get_operation_index(self, op: Operation) -> int:
         """Get the operation position in a block."""
         if op.parent is not self:
-            raise Exception("Operation is not a children of the block.")
+            raise Exception("Operation is not a child of the block.")
```
compor, line 1618: Couldn't resist fixing a grammar mistake!

xdsl/ir/core.py
```diff
@@ -1615,12 +1615,23 @@ def split_before(
...
         for idx, block_op in enumerate(self.ops):
             if block_op is op:
                 return idx
         assert False, "Unexpected xdsl error"
 
+    def get_operation_by_index(self, idx: int) -> Operation:
+        """Get the operation by its position in its parent block."""
+        if idx not in range(0, len(self.ops)):
```
superlopuh, line 1626: this is O(n), I'd recommend throwing the error instead of asserting False instead

xdsl/ir/core.py
```diff
@@ -1615,12 +1615,23 @@ def split_before(
...
         """Get the operation position in a block."""
         if op.parent is not self:
-            raise Exception("Operation is not a children of the block.")
+            raise Exception("Operation is not a child of the block.")
         for idx, block_op in enumerate(self.ops):
             if block_op is op:
                 return idx
         assert False, "Unexpected xdsl error"
 
+    def get_operation_by_index(self, idx: int) -> Operation:
```
superlopuh, line 1624: nit:
```suggestion
    def get_operation_at_index(self, index: int) -> Operation:
```
I can't formulate it with words but for some reason at index seems more natural than by index

compor: yeah, when in a hurry, prepositions are not my strong suit.

xdsl/ir/core.py
```diff
@@ -1615,12 +1615,22 @@ def split_before(
...
                 return idx
         assert False, "Unexpected xdsl error"
 
+    def get_operation_at_index(self, idx: int) -> Operation:
+        """Get the operation by its position in its parent block."""
+        for _idx, block_op in enumerate(self.ops):
+            if idx == _idx:
+                return block_op
+
+        raise Exception(
```
math-fehr, line 1630:
```suggestion
        raise ValueError(
```

xdsl/ir/core.py
```diff
@@ -1615,12 +1615,22 @@ def split_before(
...
         if op.parent is not self:
-            raise Exception("Operation is not a children of the block.")
+            raise Exception("Operation is not a child of the block.")
         for idx, block_op in enumerate(self.ops):
             if block_op is op:
                 return idx
         assert False, "Unexpected xdsl error"
 
+    def get_operation_at_index(self, idx: int) -> Operation:
+        """Get the operation by its position in its parent block."""
```
math-fehr, line 1625: Can you add that the complexity is in O(n)? Just as a warning to people to not use this too much.

math-fehr (approved):
Nice!
I'm wondering why you need this function? Overall I would prefer not adding it (because of its cost, so making the API visible is dangerous), but if you absolutely need it then let's add it.

math-fehr:
> Nice! I'm wondering why you need this function? Overall I would prefer not adding it (because of its cost, so making the API visible is dangerous), but if you absolutely need it then let's add it.

By that, I mean, why do you have the index of operations but not a pointer to the operation themselves.

compor:
> > Nice! I'm wondering why you need this function? Overall I would prefer not adding it (because of its cost, so making the API visible is dangerous), but if you absolutely need it then let's add it.
> 
> By that, I mean, why do you have the index of operations but not a pointer to the operation themselves.

Yeah, gotcha. I'm hoisting loop-independent memrefs outside of a loop nest.
For this, I clone a region and there is no way I could find that keeps track of original block ops to the cloned ones.
Ideally, I'd like something like `value_mapper`, but that keeps track only of block args AFAICT.

I'm happy to close this PR and limit that functionality in my transform pass, properly do a `value_mapper` like approach a bit later.

PR #2123, opened 2024-02-08
core: add custom attr and prop name support to assembly format

superlopuh:
We can either use the python or the irdl name in the custom assembly format, it make sense to me to stick to the IRDL name of things to be portable. Not sure if there's an aspect that I'm missing here.


math-fehr (approved):
Nice! I agree that the IRDL name should be the one used for portability!

PR #2126, opened 2024-02-08
dialects: (snitch_stream) remove stride pattern operand from strided read and write

superlopuh:


nazavode (approved):

xdsl/dialects/snitch_stream.py
```diff
@@ -200,22 +202,23 @@ class StridedWriteOp(IRDLOperation):
     name = "snitch_stream.strided_write"
 
     pointer = operand_def(riscv.IntRegisterType)
-    pattern = operand_def(StridePatternType)
     stream = result_def(WritableStreamType[riscv.FloatRegisterType])
     dm = attr_def(IntAttr)
+    rank = attr_def(IntAttr)
```
compor, line 207: Can you please document very briefly `rank` and while you're at it `dm`?

compor (approved):

compor:
I think giving context for the design choice in the PR (if not within the dialect documentation) might be a good idea.
It would make obvious where this is going and why, which is not visible with piecewise/minimal PRs.

superlopuh:
Sorry, I should have been clearer, the main idea for this PR was to make the strided read and strided write operations more like the ones in the `snitch` dialect, to make the rewrite change in the next PR conceptually smaller. I'm not sure it makes sense to add documentation to ops that are deleted in the next PR, but I'll do a pass on the remaining ops in the upcoming ones!

PR #2130, opened 2024-02-08
interactive: add function that returns all the possible rewrites in a given module (3/4)

dshaaban01:
PR 3 out of 4. I broke the first one, had to resubmit PR. 

Adds a function called get_all_possible_rewrites() given possible patterns and a given module (essentially returns all the possible rewrites in a given module).

ALL_PATTERNS is a global variable with contains all the rewrites that we have (hopefully we will add more). Equivalent to rewrite version of "get_all_passes()" (list of all the possible passes).


dshaaban01:
TODO - add tests.

dshaaban01:
> TODO - add tests.

@superlopuh  I added the test in the 4th PR (https://github.com/xdslproject/xdsl/pull/2131) as I thought this made more sense since we introduce gui functionality in that PR. Let me know what you think.

xdsl/interactive/app.py
```diff
@@ -40,17 +40,83 @@
...
+    """
+
+    operation_index: int
+    rewrite: IndividualRewrite
+
+
 ALL_PASSES = tuple(sorted((p_name, p()) for (p_name, p) in get_all_passes().items()))
 """Contains the list of xDSL passes."""
 
+ALL_PATTERNS: tuple[IndividualRewrite, ...] = tuple(
```
superlopuh, line 75: this seems not to be used

xdsl/interactive/app.py
```diff
@@ -40,17 +40,83 @@
...
+    IndividualRewrite(op_name, pattern_name)
+    for (op_name, pattern_by_name) in individual_rewrite.REWRITE_BY_NAMES.items()
+    for (pattern_name, _) in pattern_by_name.items()
+)
+"""Contains all the rewrite patterns."""
+
+
+def get_all_possible_rewrites(
+    patterns: tuple[IndividualRewrite, ...], op: ModuleOp
+) -> tuple[IndexedIndividualRewrite, ...]:
```
superlopuh, line 85: can you please pass the REWRITE_BY_NAMES as an argument? that way the function doesn't depend on external state, and can be tested individually

superlopuh:
I think it would still be worth it to add a few tests for this function. I'd probabyl also recommend moving it into its own file in the interactive directory, since it doesn't depend on any other interactive things, and app.py is getting quite large

xdsl/interactive/app.py
dshaaban01: its used in the next PR for the gui tool.

math-fehr (approved):

xdsl/interactive/app.py
superlopuh: Then can you please add it in the next PR?

dshaaban01: yes sure

dshaaban01:
> I think it would still be worth it to add a few tests for this function. I'd probabyl also recommend moving it into its own file in the interactive directory, since it doesn't depend on any other interactive things, and app.py is getting quite large

I created a new file where I put in the "get all rewrites" and "condensed pass list functions" as well as the global ALL_PASSES variable (in the next PR will add the ALL_PATTERNS variable. it also includes the Named Tuples. I thought it would make sense to have all of the passes and rewrites specific functions, variables, typings etc. in one file. what do u think?

dshaaban01:
> I think it would still be worth it to add a few tests for this function. I'd probabyl also recommend moving it into its own file in the interactive directory, since it doesn't depend on any other interactive things, and app.py is getting quite large

can I please add these tests in PR 4? turns out ALL_PATTERNS is needed (which I have moved back and forth from PR 3/4 only to find out I also need the AvailablePass Named Tuple we creat in PR4. I will add the tests to PR4 bc its getting really confusing...

dshaaban01:
> > I think it would still be worth it to add a few tests for this function. I'd probabyl also recommend moving it into its own file in the interactive directory, since it doesn't depend on any other interactive things, and app.py is getting quite large
> 
> can I please add these tests in PR 4? turns out ALL_PATTERNS is needed (which I have moved back and forth from PR 3/4 only to find out I also need the AvailablePass Named Tuple we creat in PR4. I will add the tests to PR4 bc its getting really confusing...

right - so ive moved the ALL_PATTERNS variable into PR 4 as well as the test into PR 4. I hope the way that I've done it is fine and that you are okay with it... @superlopuh

xdsl/interactive/get_rewrites_and_passes.py
```diff
@@ -0,0 +1,96 @@
...
+    patterns: tuple[IndividualRewrite, ...],
+    op: ModuleOp,
+    rewrite_by_name: dict[str, dict[str, RewritePattern]],
+) -> tuple[IndexedIndividualRewrite, ...]:
+    """
+    Function that takes a sequence of IndividualRewrite Patterns and a ModuleOp, and
+    returns the possible rewrites.
+    """
+    old_module = op.clone()
+    num_ops = len(list(old_module.walk()))
```
superlopuh, line 44:
```suggestion
    ops = list(old_module.walk())
    num_ops = len(ops)
```

xdsl/interactive/get_rewrites_and_passes.py
```diff
@@ -0,0 +1,98 @@
...
+
+    current_module = old_module.clone()
+
+    res: tuple[IndexedIndividualRewrite, ...] = ()
+
+    for op_idx in range(num_ops):
+        matched_op = list(current_module.walk())[op_idx]
+        if matched_op.name not in individual_rewrite.REWRITE_BY_NAMES:
```
superlopuh, line 52: this is the global variable, not the one passed into the function

xdsl/interactive/get_rewrites_and_passes.py
```diff
@@ -0,0 +1,98 @@
...
+                        )
+                    ),
+                )
+                current_module = old_module.clone()
+                matched_op = list(current_module.walk())[op_idx]
+
+    return res
+
+
+def condensed_pass_list(input: builtin.ModuleOp) -> tuple[type[ModulePass], ...]:
```
superlopuh, line 74: why move this?

xdsl/interactive/get_rewrites_and_passes.py
```diff
@@ -0,0 +1,98 @@
...
+                res = (
+                    *res,
+                    (
+                        IndexedIndividualRewrite(
+                            op_idx, IndividualRewrite(matched_op.name, pattern_name)
+                        )
+                    ),
+                )
+                current_module = old_module.clone()
+                matched_op = list(current_module.walk())[op_idx]
```
superlopuh, line 69: why this line?

xdsl/interactive/get_rewrites_and_passes.py
superlopuh, line 1: can you please make this file contain only the function that gets all applicable rewrites given a module and a dictionary of rewrites to choose from? Could you also please add a test that exercises that function with a rewrite defined by the test, so that it doesn't depend on the rest of the framework? Could you please also remove all unrelated changes from this PR?

superlopuh (changes requested):
Please add a self-contained test and remove unnecessary changes from this PR

xdsl/interactive/get_rewrites_and_passes.py
dshaaban01: I thought it might make sense to put the get rewrites and passes functions in one file to condense app.py. Should i move it back..?

xdsl/interactive/get_rewrites_and_passes.py
dshaaban01: I don't understand what the requested change is.

superlopuh:
```suggestion
        if matched_op.name not in rewrite_by_name:
```

xdsl/interactive/get_rewrites_and_passes.py
superlopuh: I'd recommend moving it back in this PR, even if it might make sense to move it to a different file in another future PR

dshaaban01:
> Please add a self-contained test and remove unnecessary changes from this PR

I've done as you have requested. Please let me know if there are any further changes to be made.

tests/interactive/test_get_all_possible_rewrites.py
```diff
@@ -0,0 +1,57 @@
...
+    expected_res = (
+        (
+            IndexedIndividualRewrite(
+                3, IndividualRewrite(operation="arith.addi", pattern="AddImmediateZero")
+            )
+        ),
+    )
+
+    res = get_all_possible_rewrites(module, individual_rewrite.REWRITE_BY_NAMES)
+    assert res == expected_res
```
superlopuh, line 37: Could you please add a custom rewrite defined in this file? So the module could be like this:
```
module {
    "test.op"() {"label" = "a"}
    "test.op"() {"label" = "a"}
    "test.op"() {"label" = "b"}
}
```
And the rewrites like this:
``` python
@dataclass

    label: str

    @op_type_rewrite_pattern
    def match_and_rewrite(self, op: TestOp, rewriter: PatternRewriter):
        if op.attributes["label"] != StringAttr(self.label):
            return
        rewriter.replace(TestOp(attributes={"label": StringAttr("c")}))
```
The goal of this would be to have a test that is self-contained, and doesn't depend on dialects and rewrites that may change

superlopuh (approved):

PR #2133, opened 2024-02-08
dialects: (tensor) add tensor.Empty

### LINKED ISSUE #2117: add tensor.empty op

kayode-gif:
resolves #2117 
Note: Merging this branch into linalg-add branch so that it can be merged in one

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -4,7 +4,7 @@
 %t1 = tensor.empty() : () -> tensor<2x3xf32>
 %t2 = tensor.empty() : () -> tensor<2xf32>
 %i1 = "test.op"() : () -> index
-%t3 = tensor.empty(%i1) : (index) -> tensor<3xf32>
+%t3 = tensor.empty(%i1 : index) : (index) -> tensor<3xf32>
 
 
 
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -1,14 +1,14 @@
 // RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
 
-
 %t1 = tensor.empty() : tensor<2x3xf32>
 %t2 = tensor.empty() : tensor<2xf32>
 %i1 = "test.op"() : () -> (index)
-
+%t3 = tensor.empty(%i1) : tensor<3xf32>
 
 
 // CHECK:       module {
 // CHECK-NEXT:  %0 = tensor.empty() :  tensor<2x3xf32>
 // CHECK-NEXT:  %1 = tensor.empty() : tensor<2xf32>
 // CHECK-NEXT:  %2 = "test.op"() : () -> index
+// CHECK-NEXT:  %3 =  tensor.empty(%2) : tensor<3xf32>
 // CHECK-NEXT: }
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -1,16 +1,16 @@
 // RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
 
 
-%t1 = tensor.empty() : () -> tensor<2x3xf32>
-%t2 = tensor.empty() : () -> tensor<2xf32>
+%t1 = tensor.empty() : tensor<2x3xf32>
+%t2 = tensor.empty() : tensor<2xf32>
 %i1 = "test.op"() : () -> index
-%t3 = tensor.empty(%i1 : index) : (index) -> tensor<3xf32>
+%t3 = tensor.empty(%i1 : index) : tensor<3xf32>
 
 
 
 // CHECK:       module {
-// CHECK-NEXT:  %0 = tensor.empty() : () -> tensor<2x3xf32>
-// CHECK-NEXT:  %1 = tensor.empty() : () -> tensor<2xf32>
+// CHECK-NEXT:  %0 = tensor.empty() :  tensor<2x3xf32>
+// CHECK-NEXT:  %1 = tensor.empty() : tensor<2xf32>
 // CHECK-NEXT:  %2 = "test.op"() : () -> index
-// CHECK-NEXT:  %3 = tensor.empty(%i1 : index) : (index) -> tensor<3xf32>
+// CHECK-NEXT:  %3 = tensor.empty(%i1 : index) : tensor<3xf32>
 // CHECK-NEXT: }
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -6,6 +6,6 @@
 
 
 // CHECK:       module {
-// CHECK-NEXT:  %0 = tensor.empty() : () -> tensor<2x3xf32>
-// CHECK-NEXT:  %1 = tensor.empty() : () -> tensor<2xf32>
+// CHECK-NEXT:  %0 = "tensor.empty"() : () -> tensor<2x3xf32>
+// CHECK-NEXT:  %1 = "tensor.empty"() : () -> tensor<2xf32>
 // CHECK-NEXT: }
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -4,5 +4,5 @@
 %t1 = tensor.empty() : tensor<2x3xf32>
 
 // CHECK:       module {
-// CHECK-NEXT:   %1 = tensor.empty() : tensor<2x3xf32>
+// CHECK-NEXT:   %0 = tensor.empty() : tensor<2x3xf32>
 // CHECK-NEXT: }
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -1,5 +1,4 @@
-// RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
+// RUN: XDSL_ROUNDTRIP
 
-%t1 = tensor.empty() : tensor<2x3xf32>
-
-// CHECK:   %1 = tensor.empty() : tensor<2x3xf32>
+%t1 = "tensor.empty"() : tensor<2x3xf32>
+// CHECK:   %t1 = tensor.empty() : tensor<2x3xf32>
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
@@ -4,7 +4,6 @@
 %t1 = tensor.empty() : tensor<2x3xf32>
 %t2 = tensor.empty() : tensor<2xf32>
 %i1 = "test.op"() : () -> (index)
-%t3 = tensor.empty(%i1 : index) : tensor<3xf32>
 
 
 
@@ -12,5 +11,4 @@
 // CHECK-NEXT:  %0 = tensor.empty() :  tensor<2x3xf32>
 // CHECK-NEXT:  %1 = tensor.empty() : tensor<2xf32>
 // CHECK-NEXT:  %2 = "test.op"() : () -> index
-// CHECK-NEXT:  %3 = tensor.empty(%i1 : index) : tensor<3xf32>
 // CHECK-NEXT: }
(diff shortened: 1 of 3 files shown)


xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,35 @@
...
+
+@irdl_op_definition
+class EmptyOp(IRDLOperation):
+    name = "tensor.empty"
+
+    dynamic_sizes = var_operand_def(IndexType)
+
+    tensor = result_def(TensorType[Attribute])
+
+    assembly_format = (
```
superlopuh, line 18: So this code was a bit of a hack, that's not quite valid, unfortunately. Here's the link to the op definition in MLIR:

https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorempty-tensoremptyop

The types of the dynamic sizes are actually inside the brackets also I don't know how MLIR handles this in its assembly format (CC @PapyChacal @math-fehr) but, for the time being, we might have to go with a custom parser/printer pair

kayode-gif: okay, after reading the docs, the only issues i see is with the assembly format and perhaps also needing a check for static and dynamic sizing (i'm not sure if there a functions in xdsl that deal with that perhaps or if it is that important as of rn), other than that i don't see anything else wrong right?

kayode-gif: i can't also seem to find how mlir handles this too, also what would be the issue with not using the assembly custom format, would that provide issues in parsing and printing this or?

superlopuh: that's right, it's just about the dynamic sizes not being empty, and this format definition not being compatible with MLIR's. We probably want to add another test case with a couple of variables in the dynamic sizes variadic operand to check that we can print and parse in the same way as MLIR

compor (commented):
Why does this PR need to be stacked?

kayode-gif:
> Why does this PR need to be stacked?

I stacked this as the other pr wasn't merged yet, i shall change the target branch to main then, guess that is better

compor:
> > Why does this PR need to be stacked?
> 
> I stacked this as the other pr wasn't merged yet, i shall change the target branch to main then, guess that is better

I'm not sure if you were working on a lowering or similar, but I still don't see the connection.
Anyhow, thanks for fixing it.

xdsl/dialects/tensor.py
kayode-gif: i've been playing around locally to see if tensor.empty can take the input, but it doesn't like it, as it return "expected 0 operand types but had x", not sure where to go from here, still can't find nothing on how mlir handles this too

superlopuh: This is an error from xDSL or MLIR? I would first create an example that works with MLIR, and then implement whatever is needed to make it work in xDSL. I have a feeling that a custom print and parse function will be necessary.

kayode-gif: this is an error within .mlir, so yh mlir, here is a snippet:
```
%t1 = "tensor.empty"() : () -> tensor<2x3xf32>
%i0 = "test.op"() : () -> index
%t2 = tensor.empty(%i0) : () -> tensor<2xf32>

// CHECK:       module {
// CHECK-NEXT:  %0 = tensor.empty() : tensor<2x3xf32>
// CHECK-NEXT   %1 = test.op() : () -> index
// CHECK-NEXT:  %1 = tensor.empty(%i0) : tensor<2xf32>
// CHECK-NEXT:  }
```
obviously xdsl can take the input but it seems not recognised in mlir

compor: Two things:

- MLIR is both the framework and the format. Usually the format manifests itself in textual form in `.mlir` files.
- Both xDSL and MLIR (`xdsl-opt` and `mlir-opt`) are MLIR frameworks, it's just that xDSL is a "deep copy" of the MLIR framework

So, are you getting this error with `mlir-opt` or `xdsl-opt`?

I think `mlir-opt` is right because your operation signature here is different from how you're using it.

```mlir
%t2 = tensor.empty(%i0) : () -> tensor<2xf32>
```

this bit `() -> tensor<2xf32>` says it accepts no args and returns a `tensor<2xf32>`, but you seem to actually pass argument `i0` there.

compor: also is that filecheck mixing generic and custom formats for a reason? I didn't check the whole discussion above very closely.

kayode-gif: oh okay, that makes sense, the error is with ```xdsl-opt```

kayode-gif:
currently facing issues with printing/parsing, specifically with the dynamic size types, ``` parse_optional_type``` and ```parse_optonal_resolved_operand``` doesn't behave as i thought it would be. some issues are that it is asking for an operand being expected also or type. thanks @compor @superlopuh

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,75 @@
...
+            result_types=(tensor_type,),
+        )
+
+    def print(self, printer: Printer):
+        if self.dynamic_sizes:
+            printer.print_string("(")
+            printer.print_list(self.dynamic_sizes, printer.print_ssa_value)
+            printer.print_string(")")
+
+        if self.tensor:
```
superlopuh, line 34: the tensor is always present, so there's no need to check here.

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, dynamic_sizes: Sequence[SSAValue], tensor_type: Attribute):
+        super().__init__(
+            operands=(dynamic_sizes,),
+            result_types=(tensor_type,),
+        )
+
+    def print(self, printer: Printer):
+        if self.dynamic_sizes:
+            printer.print_string("(")
+            printer.print_list(self.dynamic_sizes, printer.print_ssa_value)
```
superlopuh, line 31: in my understanding, the format is `tensor.empty(*list of ssa values* : *list of types of the ssa values*) : () -> *tensor type*`
This means that there should be two print lists here, one to print the values, then another to print the types of the values. Have a look at how the return or yield ops are printed, they do the same thing

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,75 @@
...
+            printer.print_list(
+                self.tensor, lambda tensor: printer.print_attribute(tensor.type)
+            )
+
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        pos = parser.pos
+        if parser.parse_optional_punctuation("("):
+            unresolved_dynamic_sizes = parser.parse_comma_separated_list(
+                Parser.Delimiter.NONE, parser.parse_optional_unresolved_operand
```
superlopuh, line 45:
```suggestion
                Parser.Delimiter.NONE, parser.parse_unresolved_operand
```
The strategy here should be to check if there's a closing paren. If so, then there are no operands. If not, then we can be sure that there's at least one element, and use the non-optional method to parse each operand. Similarly below.

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,75 @@
...
+                Parser.Delimiter.NONE, parser.parse_optional_type
+            )
+            parser.parse_punctuation(")")
+            dynamic_sizes = parser.resolve_operands(
+                unresolved_dynamic_sizes, dynamic_sizes_types, pos
+            )
+        else:
+            dynamic_sizes = ()
+
+        if parser.parse_optional_punctuation(" : "):
```
superlopuh, line 57: not optional

kayode-gif:
also i may need help on doing a test with an arg too

tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
+
+
+%t1 = "tensor.empty"() : () -> tensor<2x3xf32>
+%t2 = "tensor.empty"() : () -> tensor<2xf32>
```
superlopuh, line 5: can you please add a tensor with arguments?

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,90 @@
...
+                parser.parse_punctuation(")")
+            dynamic_sizes = parser.resolve_operands(
+                unresolved_dynamic_sizes, unresolved_types, pos
+            )
+        else:
+            dynamic_sizes = ()
+
+        parser.parse_punctuation(":")
+
+        if parser.parse_punctuation("->"):
```
superlopuh, line 72: not optional, so if does nothing

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,90 @@
...
+            tensor = parser.parse_comma_separated_list(
```
superlopuh, line 73: the tensor isn't a list, it's an Attribute, so `parser.parse_attribute` should do the trick

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,89 @@
...
+                unresolved_dynamic_sizes, unresolved_types, pos
+            )
+        else:
+            dynamic_sizes = ()
+
+        parser.parse_punctuation(":")
+        parser.parse_punctuation("(")
+        parser.parse_punctuation(")")
+        parser.parse_punctuation("->")
+        tensor_types = parser.parse_attribute
```
superlopuh, line 76:
```suggestion
        tensor_types = parser.parse_attribute()
```

xdsl/dialects/tensor.py
```diff
@@ -62,18 +62,19 @@ def parse(cls, parser: Parser) -> Self:
                 unresolved_types = parser.parse_comma_separated_list(
                     Parser.Delimiter.NONE, parser.parse_type
                 )
-                parser.parse_punctuation(")")
+                parser.parse_optional_punctuation(")")
```
superlopuh, line 65:
```suggestion
                parser.parse_punctuation(")")
```

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,90 @@
...
+                    Parser.Delimiter.NONE, parser.parse_type
+                )
+                parser.parse_optional_punctuation(")")
+            dynamic_sizes = parser.resolve_operands(
+                unresolved_dynamic_sizes, unresolved_types, pos
+            )
+        else:
+            dynamic_sizes = ()
+
+        parser.parse_optional_punctuation(":")
```
superlopuh, line 72:
```suggestion
        parser.parse_punctuation(":")
```

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,90 @@
...
+        parser.parse_punctuation("(")
+        parser.parse_optional_type()
+        parser.parse_punctuation(")")
```
superlopuh, line 75: I don't think this is in the MLIR syntax

tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir
```diff
@@ -0,0 +1,16 @@
+// RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
+
+
+%t1 = tensor.empty() : () -> tensor<2x3xf32>
+%t2 = tensor.empty() : () -> tensor<2xf32>
+%i1 = "test.op"() : () -> index
+%t3 = tensor.empty(%i1 : index) : (index) -> tensor<3xf32>
```
superlopuh, line 7:
```suggestion
%t1 = tensor.empty() : tensor<2x3xf32>
%t2 = tensor.empty() : tensor<2xf32>
%i1 = "test.op"() : index
%t3 = tensor.empty(%i1 : index) : tensor<3xf32>
```

xdsl/dialects/tensor.py
superlopuh: oh actually no

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,90 @@
...
+        parser.parse_punctuation("->")
```
superlopuh, line 76:
```suggestion
```

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,90 @@
...
+        tensor_types = parser.parse_attribute()
```
superlopuh, line 77:
```suggestion
        result_type = parser.parse_attribute()
```

kayode-gif:
@superlopuh very unsure why mlir based testing is till failing, to me it matches when i run the opt

kayode-gif:
xdsl opt to note

kayode-gif:
okay so it doesn't like this one: ```%t3 = tensor.empty(%i1 : index) : tensor<3xf32>```, i don't think the types are defined inside the () in mlir

superlopuh:
Can you please find out which syntax mlir uses and ping on the Zulip?

xdsl/dialects/tensor.py
superlopuh: My understanding was incorrect, thank you for investigating

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,73 @@
...
+            printer.print_string("(")
+            printer.print_string(")")
+
+        printer.print_string(" : ")
+        printer.print_attribute(self.tensor.type)
+
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        pos = parser.pos
+        if parser.parse_punctuation("("):
```
superlopuh, line 43: this will always succeed, so might as well inline the body here

xdsl/dialects/tensor.py
```diff
@@ -0,0 +1,73 @@
...
+            if parser.parse_optional_punctuation(")"):
+                unresolved_dynamic_sizes = ()
+                unresolved_types = ()
```
superlopuh, line 46:
```suggestion
                dynamic_sizes = ()
```

superlopuh (approved):

PR #2134, opened 2024-02-08
frontend: (onnx) add get_shape and get_tensor_type

superlopuh:
Also adds onnx as an optional dependency, that doesn't get installed by default even with `-r requirements.txt`, and adds a unit test that gets skipped if onnx is not installed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,9 +45,10 @@ typeCheckingMode = "strict"
 "exclude" = [
     "tests/test_frontend_op_resolver.py",
     "tests/test_frontend_python_code_check.py",
+    "tests/frontend/onnx/test_tensor_type.py",
     "tests/dialects/test_memref.py",
     "xdsl/interpreters/experimental/wgpu.py",
-    "xdsl/frontend/onnx.py",
+    "xdsl/frontend/onnx/shape_type.py",
 ]
 "ignore" = [
     "tests/filecheck/frontend/dialects/builtin.py",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 extras = ["riscemu==2.2.5", "wgpu==0.13.2", "textual==0.47.1", "pyclip==0.7"]
+onnx = ["onnx-weekly"]
 
 [project.urls]
 Homepage = "https://xdsl.dev/"
@@ -46,6 +47,7 @@ typeCheckingMode = "strict"
     "tests/test_frontend_python_code_check.py",
     "tests/dialects/test_memref.py",
     "xdsl/interpreters/experimental/wgpu.py",
+    "xdsl/frontend/onnx.py",
 ]
 "ignore" = [
     "tests/filecheck/frontend/dialects/builtin.py",
--- a/tests/frontend/onnx/test_tensor_type.py
+++ b/tests/frontend/onnx/test_tensor_type.py
@@ -1,0 +1,49 @@
+import pytest
+
+from xdsl.dialects.builtin import TensorType, f32, f64
+
+pytest.importorskip("onnx", reason="riscemu is an optional dependency")
+
+from onnx import TensorShapeProto, TypeProto  # noqa: E402
+
+from xdsl.frontend.onnx.shape_type import get_shape, get_tensor_type  # noqa: E402
+
+
+def test_get_shape():
+    assert get_shape(TensorShapeProto()) == ()
+    assert get_shape(
+        TensorShapeProto(dim=(TensorShapeProto.Dimension(dim_value=1),))
+    ) == (1,)
+    assert get_shape(
+        TensorShapeProto(
+            dim=(
+                TensorShapeProto.Dimension(dim_value=1),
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -15,6 +15,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 extras = ["riscemu==2.2.5", "wgpu==0.13.2", "textual==0.47.1", "pyclip==0.7"]
+onnx = ["onnx-weekly"]
```
superlopuh, line 18: weekly because 1.16 will get typing support, but 1.15 is not pyright-friendly

pyproject.toml
```diff
@@ -46,6 +47,7 @@ typeCheckingMode = "strict"
     "tests/test_frontend_python_code_check.py",
     "tests/dialects/test_memref.py",
     "xdsl/interpreters/experimental/wgpu.py",
+    "xdsl/frontend/onnx.py",
```
superlopuh, line 50: don't show pyright errors for this file if you don't have it open

superlopuh:
I'm not sure why the CI doesn't work... At least it doesn't fail, but it would be good to keep testing the onnx things continuously. I'll take another look tomorrow.

webmiche (commented):
(on the dependency part) LGTM. I was wondering how you manage different workflows and I think it's a reasonable choice to add this in the MLIR workflow.

In a perfect world, I think we would have CI flows with different configurations, but I think this is probably not worth the effort of configuring it and of running the CI several times on the same commit/PR.

What I don't quite get is the weekly thing and how you handle versions. I think, one of the cool parts with the requirements file (and with our optional dependencies) is that the version is fixed and we explicitely need to change it. How does this work here? In essence, are we safe from `onnx-weekly` versions updates that break something that then breaks all our CI runs?

webmiche:
> I'm not sure why the CI doesn't work... At least it doesn't fail, but it would be good to keep testing the onnx things continuously. I'll take another look tomorrow.

I am confused. I see all checks passing. Have you already changed something that is not reflected in this PR?

compor:
> > I'm not sure why the CI doesn't work... At least it doesn't fail, but it would be good to keep testing the onnx things continuously. I'll take another look tomorrow.
> 
> I am confused. I see all checks passing. Have you already changed something that is not reflected in this PR?

I checked the workflow step and I didn't see `onnx` getting installed as per the new YAML CI file.

webmiche:
> I checked the workflow step and I didn't see `onnx` getting installed as per the new YAML CI file.

Okay, I do see it, for both python 3.10 and python 3.11. For both the step `Install the package locally and nbval` finishes with 

```
Successfully installed numpy-1.26.4 onnx-weekly-1.16.0.dev20240205 protobuf-4.25.2 xdsl-0+untagged.1.g43ad93f
```

I also see that the test that is supposed to skip has successfully run. So this is fixed, right?

compor:
> > I checked the workflow step and I didn't see `onnx` getting installed as per the new YAML CI file.
> 
> Okay, I do see it, for both python 3.10 and python 3.11. For both the step `Install the package locally and nbval` finishes with
> 
> ```
> Successfully installed numpy-1.26.4 onnx-weekly-1.16.0.dev20240205 protobuf-4.25.2 xdsl-0+untagged.1.g43ad93f
> ```
> 
> I also see that the test that is supposed to skip has successfully run. So this is fixed, right?

It seems so, not sure why it was missed in the coverage report.

webmiche:
Oh no, not codecov again. Taking a look :)

webmiche:
Okay, it does not seem to be codecovs fault this time. In the `Combine coverage data` step, one can see how the coverage looks like that was uploaded and it indeed has a line.

```tests/frontend/onnx/test_tensor_type.py                                      12      9      0      0    25%```

I would suggest trying this out locally and see whether you arrive at the same coverage data. However, to me, this is not a blocker, I am fine with merging this. Maybe, if you can't figure it out, open an issue, but that's also not extremely important to me.

**Review by webmiche — APPROVED** [review]

File: tests/frontend/onnx/test_tensor_type.py [hunk 3]
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+from xdsl.dialects.builtin import TensorType, f32, f64
+
+pytest.importorskip("onnx-weekly", reason="onnx is an optional dependency")
```
webmiche, line 5: Maybe I am trippin, but this kind of does not make 100% sense to me. You are including from onnx in the line below, but here you are checking whether you can include onnx-weekly? (This might explain the coverage thingi)

superlopuh: Yeah this was super confusing to me. It outright fails for `"onnx"`, and skips successfully with `"onnx-weekly"`. Maybe I'm not getting something, but I don't understand how it can fail to not find a package that's not there...

superlopuh: I'll try again now with a new venv

superlopuh:
Replacing importorskip seems to have worked. Hopefully this is a temporary workaround, until they release ONNX 1.16

kayode-gif (approved):

PR #2135, opened 2024-02-10
transformations: Hoist pairs of `memref` loads and stores out of an `scf.for` loop

compor:
This PR:

- Adds a loop hoisting transformation to move `memref` loads and stores using the same memory location out of an `scf.for` if the indexing in that `memref` is loop-independent.
- Adds of the above.

The functionality is intentionally very restricted and does not handle:
- Multiple loads from the same location
- Multiple stores of the same value and/or to same location
- Loads must precede stores (i.e., WAR-only dependence)
- Does not handle only loads or only stores

This helps us improve the execution performance when lowering `linalg` kernels to `scf` and down to our vanilla `riscv` code generation passes.


tests/transforms/test_transform_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    op3 = test.TestOp([op2, op1], result_types=[i32])
+    op4 = test.TestOp([bb0.args[0], op2], result_types=[i32])
+    bb0.add_ops([op1, op2, op3, op4])
+
+    for_op = For(lb, ub, step, [], bb0)
+
+    assert is_loop_dependent(op1.results[0], for_op) is False
+    assert is_loop_dependent(op2.results[0], for_op) is False
+    assert is_loop_dependent(op3.results[0], for_op) is False
+    assert is_loop_dependent(op4.results[0], for_op) is True
```
superlopuh, line 134: nit:
```suggestion
    assert not is_loop_dependent(op1.results[0], for_op)
    assert not is_loop_dependent(op2.results[0], for_op)
    assert not is_loop_dependent(op3.results[0], for_op)
    assert is_loop_dependent(op4.results[0], for_op)
```
and elsewhere, is True and is False is throwing me off a bit, I guess not a big deal

superlopuh:
I've been looking at the way MLIR does this, and I'm wondering if we want to mirror their logic a bit more, especially to be able to do this kind of operation both at the scf/memref level and and at the riscv_scf/pointer level. It feels like it might be a good idea to at least attempt something in that direction, to factor out some of the functionality into traits, (LoopLike and maybe others), and start with things like hoisting arithmetic, and then hoisting memory accesses. Do you think that would be too much overhead? I feel like we can't really avoid the optimisations at the riscv level, due to the creation of arithmetic operations when lowering memref stores

compor:
> I've been looking at the way MLIR does this, and I'm wondering if we want to mirror their logic a bit more, especially to be able to do this kind of operation both at the scf/memref level and and at the riscv_scf/pointer level. It feels like it might be a good idea to at least attempt something in that direction, to factor out some of the functionality into traits, (LoopLike and maybe others), and start with things like hoisting arithmetic, and then hoisting memory accesses. Do you think that would be too much overhead? I feel like we can't really avoid the optimisations at the riscv level, due to the creation of arithmetic operations when lowering memref stores

I agree, actually implementing the LoopLike interface is on my to-do list for after the rebuttal.
This is very constrained but still can be reapplied to different abstraction levels given there are traits/interface support.

I couldn't find anything specific in MLIR, and this is very LLVM-like.
I would also like to explore doing this at a higher level.

xdsl/transforms/loop_hoist_memref.py
```diff
@@ -0,0 +1,144 @@
...
+        loads: list[memref.Load] = []
+        for op in for_op.body.ops:
+            if isinstance(op, memref.Load):
+                loads.append(op)
+
+        if len(loads) == 0:
+            return
+
+        parent_block = loads[0].parent_block()
+        assert parent_block is not None
```
superlopuh, line 43:
```suggestion
        parent_block = for_op.body.block
        
        loads = [op for op in parent_block.ops if isinstance(op, memref.Load)]

        if not loads:
            return
```

xdsl/transforms/loop_hoist_memref.py
```diff
@@ -0,0 +1,144 @@
...
+        store_vals = [store.value for store in load_store_pairs.values()]
+        dup_store_vals = [val for val in store_vals if store_vals.count(val) > 1]
+        load_store_pairs = {
+            load: store
+            for load, store in load_store_pairs.items()
+            if store.value not in dup_store_vals
+        }
+
+        if len(load_store_pairs.items()) == 0:
+            return
```
superlopuh, line 71:
```suggestion
        if not load_store_pairs:
            return
```

xdsl/transforms/utils.py
```diff
@@ -0,0 +1,58 @@
...
+        ):
+            if found_op is None:
+                found_op = op
+            else:
+                return None
+
+    return found_op
+
+
+def is_loop_dependent(val: SSAValue, loop: scf.For):
```
superlopuh, line 40: can you please add a doc comment explaining what this does?

xdsl/transforms/utils.py
superlopuh, line 1: I feel a little queasy about these utils, and would feel a little more comfortable with them in the same file as the rewrite. I'm not sure that now is the time to make them more general, but it feels like we would want to take the time to add both of these functions slowly, making sure to find the most efficient implementation. For the second function, I'm a little wary of computing the entire use-def chain of an SSAValue to determine that it doesn't depend on another one, I have a feeling that it should be possible to add some shortcuts, and it would be nice to take the time to consider our options when we're in less of a rush to add optimisations influenced by a paper schedule. We can always move the functions to a utils file in the future.

xdsl/transforms/loop_hoist_memref.py
```diff
@@ -0,0 +1,144 @@
...
+        new_loads = [load.clone() for load in load_store_pairs.keys()]
+        rewriter.insert_op_before(new_loads, for_op)
+
+        ld_indices = [
+            parent_block.get_operation_index(load) for load in load_store_pairs.keys()
+        ]
+        st_indices = [
+            parent_block.get_operation_index(store)
+            for store in load_store_pairs.values()
+        ]
```
superlopuh, line 83: why go through the indices and not the ops directly?

compor: I want a map between the original to cloned block ops, and not all ops, just load and stores.
Till the `clone_into` offers that, this is the simplest way to filter them once we have the targets to operate on.

xdsl/transforms/utils.py
compor: I wanted to test them, so doing it as private methods wasn't really an option, unless I'm missing something.
That's standard dependence in the SSA, you should check out how LLVM does it, it even has a number of operations threshold to check.
I'm not sure what kind of shortcuts you foresee, but for being absolutely definitive I don't see a way around.
Lastly, it's not that we are handling hundreds of line of code and `visited` ensures things are not checked twice.

xdsl/transforms/loop_hoist_memref.py
compor: I'll see if I can change it closer to the proposed API change.

xdsl/transforms/utils.py
compor: Should I move them back in the hoist module?

superlopuh: I would recommend moving these functions to the file for now

superlopuh (approved):

xdsl/transforms/utils.py
```diff
@@ -0,0 +1,58 @@
...
+    for _idx, block_op in enumerate(block.ops):
+        if idx == _idx:
+            return block_op
+
+    raise ValueError(
+        f"Cannot get operation by out-of-bounds index {idx} in its parent block."
+    )
+
+
+def find_same_target_store(load: memref.Load):
```
JosseVanDelm, line 18: Possibly very dumb question, but there's no guarantee that the result of one memref.load operation is used by just one store op, is it? Shouldn't this function be returning a list of `found_ops`?

JosseVanDelm: Oh woops I see that this is in the comment of the PR, maybe also put these limitations in comments?

compor: That's why it returns `None` when there's more than one :)

xdsl/transforms/loop_hoist_memref.py
```diff
@@ -0,0 +1,144 @@
...
+from xdsl.transforms.utils import (
+    find_same_target_store,
+    get_operation_at_index,
+    is_loop_dependent,
+)
+
+
+@dataclass
+class LoopHoistMemref(RewritePattern):
+    """Hoist pairs of memref.loads and memref.stores out of a loop."""
```
nazavode, line 80: Trying to add some context to the rewrite as an excuse to add some familiar words (e.g.: redundant):

```suggestion
    """Hoist pairs of memref.loads and memref.stores out of a loop.
    
    This rewrite hoist pairs of redundant memref.load and memref.store operations outside of their enclosing scf loop nest. The memref operation pair is considered for this rewrite if their memref target location is the same and it is constant with respect to the containing loop's induction variable.
    """
```

nazavode (changes requested):

webmiche (commented):
(abstract question) Do you need to handle aliasing of memrefs? Or do memrefs trivially not alias in MLIR?

If you assume aliasing away, you should probably mention it in the description of the pattern.

xdsl/transforms/loop_hoist_memref.py
compor: `redundant` is a bit vague, redundant stuff can get DCE'd in my mind. But I'll add some more "words"

compor:
> (abstract question) Do you need to handle aliasing of memrefs? Or do memrefs trivially not alias in MLIR?
> 
> If you assume aliasing away, you should probably mention it in the description of the pattern.

Good point, I have ignored this on purpose.
My understanding is that there is aliasing, but the analysis here is very basic, relying on the use of the same `memref`.
If there is aliasing, it is not considered, so we are conservatively safe.

I'll add a point to the doc regarding this. Thanks!

nazavode (approved):

PR #2136, opened 2024-02-11
interactive: change input text so you can see example of a rewrite on gui launch 

dshaaban01:
Changed input text to be an example where you can see an example of rewrite on gui launch. A canonicalizaiton pattern for Add Immediate Zero was pushed, and thats why this example will show a rewrite and the other one did not (no rewrite defined yet). Still need to add more rewrite patterns --> next PR's.

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -107,10 +107,10 @@ class InputApp(App[None]):
     """
 
     INITIAL_IR_TEXT = """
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
Maybe explain in the commit message why the previous example did not show a rewrite.

superlopuh (approved):

PR #2146, opened 2024-02-13
dialects: (memref_stream) add streaming region

superlopuh:
Note stacked PR

Add a higher-level representation of a streaming region, operating on memrefs and affine maps.

### DIFF
--- a/xdsl/dialects/memref_stream.py
+++ b/xdsl/dialects/memref_stream.py
@@ -1,20 +1,31 @@
 """
-This work-in-progress dialect represents a target-independent representation of streams of
-registers over time. It's missing an operation and an attribute:
+A target-independent representation of streams of buffers over time.
 
- - memref_stream.stride_pattern:
-    attribute specifying the order in which elements are accessed
- - memref_stream.streaming_region:
-    taking streams and stride patterns, and providing the stream values to be read from or written to
-
-These, along with more detailed documentation, will be added over the next week.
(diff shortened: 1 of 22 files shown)


xdsl/dialects/memref_stream.py
```diff
@@ -27,11 +38,166 @@ class WriteOp(stream.WriteOperation):
     name = "memref_stream.write"
 
 
+@irdl_op_definition
+class StreamingRegionOp(IRDLOperation):
+    """
+    An operation that creates streams from access patterns, which are only available to
+    read from and write to within the body of the operation.
+
+    Withing the loop body, memrefs that are streamed must not be otherwise accessed
```
nazavode, line 47:
```suggestion
    Within the loop body, memrefs that are streamed must not be otherwise accessed
```

xdsl/dialects/memref_stream.py
```diff
@@ -27,11 +38,166 @@ class WriteOp(stream.WriteOperation):
...
+    through loads and stores.
```
nazavode, line 48:
```suggestion
    via memref.load, memref.store or any other access mean, including extraction (e.g.: memref.view).
```

nazavode (changes requested):
Looks good, thanks! Addressed a typo and tried to add details to the docstrings (that were very good already).

compor (approved):

nazavode (approved):

PR #2147, opened 2024-02-13
transformations: (memref_stream) add streaming_region lowering

superlopuh:
Note stacked PR

Add a rewrite pattern from memref_stream.streaming_region to snitch_stream.streaming_region

### DIFF
--- a/xdsl/transforms/convert_memref_stream_to_snitch_stream.py
+++ b/xdsl/transforms/convert_memref_stream_to_snitch_stream.py
@@ -1,14 +1,29 @@
+import operator
+from collections.abc import Sequence
+from functools import reduce
+from itertools import accumulate
 from typing import cast
 
 from xdsl.backend.riscv.lowering.utils import (
+    cast_operands_to_regs,
     register_type_for_type,
 )
-from xdsl.dialects import memref_stream, riscv_snitch, stream
+from xdsl.dialects import (
+    builtin,
+    memref,
+    memref_stream,
+    riscv,
+    riscv_snitch,
+    snitch_stream,
+    stream,
+)
 from xdsl.dialects.builtin import (
+    ArrayAttr,
     ModuleOp,
     UnrealizedConversionCastOp,
 )
 from xdsl.ir import Attribute, MLContext
+from xdsl.ir.affine import AffineExpr, AffineMap
 from xdsl.passes import ModulePass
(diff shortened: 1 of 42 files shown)


xdsl/transforms/convert_memref_stream_to_snitch_stream.py
```diff
@@ -70,6 +85,133 @@ def match_and_rewrite(
...
+            ),
+        ),
+    )
+
+
+def strides_for_affine_map(
+    affine_map: AffineMap, shape: Sequence[int], factor: int
+) -> list[int]:
+    """
+    Given an affine map mapping an iteration space to a domain, and a shape
```
nazavode, line 196:
```suggestion
    Given an iteration space represented as an affine map (for indexing) and a shape (for bounds), returns the corresponding iteration strides for each dimension.
    
    The affine map must not have symbols.
```

nazavode (changes requested):
LGTM, just completed a docstring that seemed an oversight.

xdsl/transforms/convert_memref_stream_to_snitch_stream.py
```diff
@@ -70,6 +85,133 @@ def match_and_rewrite(
...
+def offset_map_from_shape(shape: Sequence[int], factor: int) -> AffineMap:
+    """
+    Given a list of lengths for each dimension of a memref, and the number of bytes per
+    element, returns the map from indices to an offset in bytes in memory. The resulting
+    map has one result expression.
+
+    e.g.:
+    ```
+    my_list = [1, 2, 3, 4, 5, 6]
+    shape = [2, 3]
```
compor, line 153: is the shape `[2, 3]` or `[3, 2]`? Sorry, but the reversing somehow trips me up on what I expect.
Not sure what's actionable here.

compor (approved):

xdsl/transforms/convert_memref_stream_to_snitch_stream.py
compor: I agree, keeping it as it is in the other dialects (incl. upstream) would help.

superlopuh: The good news is that as long as we get it wrong an even number of times then we'll get the right result in the end :)

nazavode (approved):

PR #2148, opened 2024-02-13
dialects: (memref_stream) add memref_stream.generic

superlopuh:
Note stacked PR

This PR introduces the memref_stream equivalent of the linalg generic operation. The main differences are that this one has an extra property `bounds`, which contains the bounds of the loop nest, and that the operands may be streams, in which case they'll be lowered to stream reads and writes instead of loads and stores.


compor (approved):

xdsl/dialects/experimental/aie.py
nazavode, line 1: I'm a bit confused by this whole new dialect...

superlopuh: It's a main merge conflict, I'll resolve later today

nazavode (approved):

PR #2150, opened 2024-02-13
transformations: add memref-streamify

superlopuh:
Note stacked PR

This PR introduces a rewrite that creates streams per memref_stream.generic operation.

### DIFF
--- a/xdsl/transforms/memref_streamify.py
+++ b/xdsl/transforms/memref_streamify.py
@@ -1,0 +1,104 @@
+from dataclasses import dataclass, field
+from typing import cast
+
+from xdsl.dialects import memref, memref_stream, stream
+from xdsl.dialects.builtin import (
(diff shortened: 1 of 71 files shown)


xdsl/transforms/memref_streamify.py
```diff
@@ -0,0 +1,104 @@
...
+            ),
+            new_body,
+        )
+        rewriter.erase_matched_op()
+
+
+@dataclass
+class MemrefStreamifyPass(ModulePass):
+    """
+    Converts a linalg generic to perfectly nested loops.
```
compor, line 93:
```suggestion
    Converts a linalg generic to perfectly nested loops.
```

is this right?

superlopuh: It is not! Good catch

compor (approved):

PR #2151, opened 2024-02-13
transformations: add lowering from memref_stream generic to loops

superlopuh:
Note stacked PR

This PR refactors the linalg generic lowering to be usable for the memref_stream generic also, with a custom function to lower sources and destinations when given indices


compor (approved):
I've lost the ball around here, but I don't want to block things.

nazavode (approved):

PR #2158, opened 2024-02-13
transformation: (convert_onnx_to_linalg) add lowering from onnx.Add to linalg.add

### LINKED ISSUE #2119: add lowering from onnx.Add to linalg.add
Blocked by #2116

kayode-gif:
resolves #2119

### DIFF
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -15,3 +15,4 @@ func.func @main_graph(%0 : tensor<3x2xf32>, %1 : tensor<3x2xf32>) -> tensor<3x2x
 // CHECK-NEXT:     func.return %3 : tensor<3x2xf32>
 // CHECK-NEXT:   }
 // CHECK-NEXT: }
+
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -1,18 +1,12 @@
 // RUN: xdsl-opt -p convert-onnx-to-linalg %s | filecheck %s
 
-
-func.func @main_graph(%0 : tensor<3x2xf32>, %1 : tensor<3x2xf32>) -> tensor<3x2xf32> {
-     %2 = tensor.empty() : tensor<3x2xf32>
-     %3 = linalg.add ins(%0, %1 : tensor<3x2xf32>, tensor<3x2xf32>) outs(%2 : tensor<3x2xf32>) -> tensor<3x2xf32>
-     func.return %3 : tensor<3x2xf32>
-   }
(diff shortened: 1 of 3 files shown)


tests/filecheck/transforms/convert_onnx_to_linalg.mlir
superlopuh, line 1: this doesn't seem to exercise the lowering

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: xdsl-opt -p convert-onnx-to-linalg %s | filecheck %s
+
+
+func.func @main_graph(%0 : tensor<3x2xf32>, %1 : tensor<3x2xf32>) -> tensor<3x2xf32> {
```
superlopuh, line 4: I would recommend not using func.func in this test, replacing it with test.op

superlopuh (commented):
This looks like a great start, although it feels like we might as well start on the logic to convert between broadcasting rules in ONNX and MLIR/linalg. Namely, it looks like ONNX denotes broadcastable dimensions with `1`, whereas MLIR does with `?`. One approach would be to check whether any of the dimensions of the tensors being added are `1`, and raise `NotImplementedError`, and tackle this in a follow-up PR. The other is to add the helper in this PR directly. I'd maybe lean towards the first option, but the lowering itself is so tiiny that I'd be happy to do it all in one go. Your call.

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
superlopuh: yep exactly

superlopuh (approved):

superlopuh:
Lowering looks great! I'm still concerned about broadcasting. Could you please add a check to the lowering to make sure that we raise `NotImplementedError` when the shape is dynamic?

superlopuh:
And also file an issue to handle dynamic shapes in the ONNX to linalg conversion

kayode-gif:
> And also file an issue to handle dynamic shapes in the ONNX to linalg conversion

xdsl/transforms/convert_onnx_to_linalg.py
```diff
@@ -0,0 +1,46 @@
...
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+@dataclass
+class AddOpLowering(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, add: onnx.Add, rewriter: PatternRewriter, /):
+        lhs_shape = add.lhs.type.get_shape()
```
superlopuh, line 20: just needs an isinstance check

superlopuh (approved):
🚀

compor (approved):
Beautiful, thank you!

Issue #2064, opened 2024-02-01
dialects (llvm,arith,riscv): Factor out common implementation of `FastMathFlagsAttr`


compor:
Currently, `FastFlagsAttr` is implemented in the `llvm` dialect and reused in `arith`, `math` and `riscv` dialects.

It would make sense to factor this out to a `FastMathAttrBase` and reused in the dialects when needed.

Relevant discussion: https://github.com/xdslproject/xdsl/pull/2058#discussion_r1473614224

PR #2159, opened 2024-02-14
core: Make IRDL operations and attributes final

math-fehr:
It was previously possible to inherit an operation or attribute.
While this may be useful, it is also pretty dangerous, as now a constraint
such as `operand_def(MyAttribute)` will also accept other attributes such as 
`MyInheritedAttribute`, which likely shouldn't be accepted.

For instance, it was previously possible to use `riscv.fastmathflags` in the LLVM
dialect, as the attribute was inheriting `llvm.fastmathflags`.

Moreover, it is necessary for the declarative format to know that a specific attribute
type is expected, and this isn't possible if the attribute might be inherited.

The solution I'm proposing here is to use a new `final` decorator inside `irdl_attr_definition`
and `irdl_op_definition` to avoid these inheritances. Note that this `final` decorator will
trigger an error if anyone tries to subclass the final class.

Resolves: #2064


xdsl/dialects/llvm.py
```diff
@@ -1170,6 +1171,14 @@ def print_parameter(self, printer: Printer):
                 )
 
 
+@irdl_attr_definition
+class FastMathAttr(FastMathAttrBase):
+    name = "llvm.fastmath"
+
+    def __init__(self, flags: None | Sequence[FastMathFlag] | Literal["none", "fast"]):
+        super().__init__(flags)
+
```
PapyChacal, line 1180: I think you can just remove it then?

xdsl/dialects/riscv.py
```diff
@@ -60,13 +60,17 @@
...
+class FastMathFlagsAttr(FastMathAttrBase):
     """
     riscv.fastmath is a mirror of LLVMs fastmath flags.
     """
 
     name = "riscv.fastmath"
 
+    def __init__(self, flags: None | Sequence[FastMathFlag] | Literal["none", "fast"]):
+        super().__init__(flags)
+
```
PapyChacal, line 73: Idem

PapyChacal:
Why reimplement `@final`?

xdsl/dialects/llvm.py
math-fehr: Actually, no, but I don't know why ^^'
My guess is that I'm doing something wrong in the `irdl_op_definition`, but I'm not really sure why?
I'm adding an issue on this though, as this *should* work

math-fehr:
I renamed it to `runtime_final`.
The idea is that `final` is just type annotations in Python, and do not check anything at runtime.
This is a problem for us though, as we do want to have that check at runtime, as we cannot force the typing to mark `irdl_op_definition` finals.

xdsl/dialects/llvm.py
superlopuh: I left a comment on the issue, I think `Data.new` always calls init on `Data`, not on the class

superlopuh: Feels like we might as well fix that before merging this

math-fehr: I would suggest to not PR creep this here.
The solution is a bit more complex than expected.

The issue is that irdl_op_definition generates the `__init__`, as someone might inherit an `ABC` `Attribute` and then add more fields, so we cannot use the `super().__init__` by default. We would need to check that there is a previous `__init__`, and that no more fields are added in the child class.
So I would like to not do that here, and do this later when this become relevant (especially if we change the way we handle attributes)

superlopuh: fair enough

superlopuh: We should probably try and add some docs to explain the workaround

superlopuh (approved):

PapyChacal (approved):

PR #2161, opened 2024-02-14
core: Allow unregistered citizens in a registered dialects

PapyChacal:
Currently, when looking for an operation/attribute, even in "allow unregistered dialect" mode, we aggressively error out on an unregistered operation/attribute, if it so happens to have a dialect name of a registered dialect. This PR relaxes this.


xdsl/ir/core.py
```diff
@@ -225,10 +226,9 @@ def get_optional_attr(
         # Otherwise, check if the attribute dialect is registered.
         dialect_name, _ = name.split(".", 1)
         if dialect_name in self._registered_dialects:
-            if dialect_name in self._loaded_dialects:
-                return None
-            self.load_registered_dialect(dialect_name)
-            return self.get_optional_attr(name)
+            if dialect_name not in self._loaded_dialects:
+                self.load_registered_dialect(dialect_name)
+                return self.get_optional_attr(name)
```
superlopuh, line 231: nit: can you add an `and` and unnest like above?

superlopuh (approved):

webmiche (commented):
Is there a use case for this? Could you add a test?

webmiche:
Ah, I see you link from the devito stuff. I guess that's the usecase :sweat_smile:

PapyChacal:
More generally, without this, it means that any dialect implementing any op or attr, would require any other op or attr to be actually implemented for unregistered interoperability with MLIR. I don't see any bright future for xDSL as a prototyping platform with such a constraint!
Will add a test clarifying this :)

xdsl/ir/core.py
PapyChacal: Right!

math-fehr (approved):
Nice!

compor (approved):

PR #2164, opened 2024-02-14
transformations: Make mlir-opt pass take the top-level module attributes into account.

PapyChacal:
Phew this one got me running in circles!
Had a case where using our `mlir-opt` pass was arong, because it currently just discards any attributes on the parsed top-level module.


math-fehr (approved):

PapyChacal:
:laughing:

PR #2165, opened 2024-02-14
[RFC] Add [] mlir-opt sugar in our pipeline syntax.

PapyChacal:
Add some sugar in our textual pipeline syntax for MLIR interoperation: `[some-mlir-pipeline]` is equivalent to `mlir-opt{arguments="--mlir-print-op-generic","--allow-unregistered-dialect","-p","builtin.module(some-mlir-pipeline)"}`.
e.g., this allows to type: `xdsl-opt <input> -p [canonicalize],lower-linalg-to-loops,[cse]` to use MLIR's canonicalization and common subexpression elimination with less struggle, and should be equivalent to:
`
mlir-opt <input> --mlir-print-op-generic --allow-unregistered-dialect -p builtin.module(canonicalize) | xdsl-opt -p lower-linalg-to-loops | mlir-opt --mlir-print-op-generic --allow-unregistered-dialect -p builtin.module(cse) | xdsl-opt
`
 PS: I used `[...]` arbitrarily, and this is obviously not urgent; bikeshedding on the actual syntax is very welcome!


superlopuh:
cute

webmiche:
> PS: I used `[...]` arbitrarily, and this is obviously not urgent; bikeshedding on the actual syntax is very welcome!

I was actually just about to xD You caught me.

> I sort of like it. is there a better way to signal the use of `mlir-opt`? Something like `mlir-opt[...]`? Also not sure about the actual syntax here.

I also prefer to add something that says mlir in some way to this. We should probably also be careful around how MLIR passes passarguments on the cli, not to run into ambiguity issues.
So, if you tell me `mlir-opt[...]` will work, then I would be in favor of it :+1:

PapyChacal:
@webmiche 
> I also prefer to add something that says mlir in some way to this. We should probably also be careful around how MLIR passes passarguments on the cli, not to run into ambiguity issues. So, if you tell me `mlir-opt[...]` will work, then I would be in favor of it 👍

Yeah I tested locally for MLIR pass arguments and it's all fine; will add more tests! It's basically parsed similarly to a string literal, so whatever is in `[...]` ends up verbatim in the pass pipeline. The only no-go is, well, `]`, which as far as I know is not used in MLIR's pipeline syntax?
Either way, I'm thinking worst case it's just sugar, nothing critical either, so we can always hop to the next unused delimiter if it evolves on their side.
@compor 
> I sort of like it. is there a better way to signal the use of `mlir-opt`? Something like `mlir-opt[...]`?

I like it but I'm trying to optimize keystrokes here :yum: How do you both feel about just `mlir[...]`?
Now that you mention it, it's really nice :thinking: 'cause now we can basically do the same for IREE, XLA and all the MLIR-based family :grin:

webmiche:
I like `mlir[...]` as well. If you find a shorter word describing mlir, I would also be fine with that, but I honestly can't imagine one.
I guess one of the downsides of just `mlir` would be that if we add support for `mlir-translate` or `mlir-reduce` or `mlir-diff` or whatever might exist in the future, we cannot distinguish. Though, I would argue that most people want to use mlir-opt when using the functionality, so mlir-opt as the "default case" of mlir would make sense too.
Just dumping thoughts, what do others think?

PapyChacal:
Agreed on all points 🙂
I do think it only makes sense for mlir-opt and not mlir-translate et al tbh: this is still supposed to be a pass, i.e., MLIR-IR gets in, MLIR-IR gets out.
Integrating say mlir-translate in there would be a whole different story, way beyond the scope of this sugar-only proposal!

Now don't take me wrong, this would be nice sugar too, but I don't think it would belong in the pass pipeline itself

webmiche:
That's fair I'd say. Though we aren't guaranteed that people won't invent something new that also has IR in and IR out, but lives in a different binary. But, I guess we just have to revise this design when this actually happens and not worry about it now already.

compor:
> @webmiche
> 
> > I also prefer to add something that says mlir in some way to this. We should probably also be careful around how MLIR passes passarguments on the cli, not to run into ambiguity issues. So, if you tell me `mlir-opt[...]` will work, then I would be in favor of it 👍
> 
> Yeah I tested locally for MLIR pass arguments and it's all fine; will add more tests! It's basically parsed similarly to a string literal, so whatever is in `[...]` ends up verbatim in the pass pipeline. The only no-go is, well, `]`, which as far as I know is not used in MLIR's pipeline syntax? Either way, I'm thinking worst case it's just sugar, nothing critical either, so we can always hop to the next unused delimiter if it evolves on their side. @compor
> 
> > I sort of like it. is there a better way to signal the use of `mlir-opt`? Something like `mlir-opt[...]`?
> 
> I like it but I'm trying to optimize keystrokes here 😋 How do you both feel about just `mlir[...]`? Now that you mention it, it's really nice 🤔 'cause now we can basically do the same for IREE, XLA and all the MLIR-based family 😁

I understand, I'm in favour! Not too picky!

tests/filecheck/mlir-conversion/with-mlir/mlir_opt.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: xdsl-opt %s -p mlir-opt{arguments='--cse','--mlir-print-op-generic'} --print-op-generic | filecheck %s
+// RUN: xdsl-opt %s -p mlir[cse] --print-op-generic | filecheck %s
```
webmiche, line 2: This will run both tests, right? I think it's good to run and test this mlir-shorthand on the CI, but I think running twice here does not make sense. I would prefer to delete the upper line since the short-hand is now our way of interacting with mlir. What do you guys think?

PapyChacal: I wouldn't overwrite the first line, as my addition is more like sugar on top of it than intended as a replacement (it does not have the full flexibility of the underlying pass).
I get that this part is confusing; it's just me being lazy to be fair; I could split it in two filechecks; would that make sense to you?

PapyChacal: Oh, or you meant that it's redundant to test and the sugar and the underlying pass in a filecheck? I can agree with that and remove the second line; the sugar syntax is tested in the pipeline parser tests already.

webmiche: My point is that people will be looking at the tests to figure out how to use certain infrastructure. I think it makes more sense here to only have the new syntax because this is how we want people to interact with the infrastructure.
And given that both lines test the exact same thing (plus the conversion that is tested somewhere else), I think it does not matter which we choose from a testing perspective. So, I would go with the "user will look for examples" argument and use the latter.
This is also really only a minor comment, I don't mind running twice or having either of the two lines. This is just my feeling of what I think would be the best of all the choices :)

PapyChacal: Ah, I see;
I would still keep an example with the underlying pass, as it is still expected to be exposed. But I can switch that case to use it in a way that the `mlir[]` does not allow and add comments!

superlopuh (approved):

PR #2166, opened 2024-02-14
dialects: add arith canoncialization pattern "AdditionOfSameVariablesToMultiplybyTwo"

dshaaban01:
This arith rewrite pattern does the following:

`a + a -> a * 2`

Update: @math-fehr says we may not want to merge "bad" rewrite patterns. Please see [this PR](https://github.com/xdslproject/xdsl/pull/2167) as a potential solution.

### DIFF
--- a/tests/filecheck/transforms/arith-addition-of-same-variables-to-multiply-by-two.mlir
+++ b/tests/filecheck/transforms/arith-addition-of-same-variables-to-multiply-by-two.mlir
@@ -1,17 +1,17 @@
 // RUN: xdsl-opt %s -p canonicalize | filecheck %s
 
 func.func @hello(%n : i32) -> i32 {
-  %one = arith.constant 2 : i32
-  %res = arith.addi %one, %one : i32
+  %two = arith.constant 3 : i32
+  %res = arith.addi %two, %two : i32
   func.return %res : i32
 }
 
-
 //CHECK:         builtin.module {
 // CHECK-NEXT:     func.func @hello(%n : i32) -> i32 {
-// CHECK-NEXT:       %one = arith.constant 2 : i32
-// CHECK-NEXT:       %res = arith.muli %one, 2 : i32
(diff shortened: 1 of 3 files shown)


tests/filecheck/transforms/arith-addition-of-same-variables-to-multiply-by-two.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: xdsl-opt %s -p canonicalize | filecheck %s
+
+func.func @hello(%n : i32) -> i32 {
+  %two = arith.constant 3 : i32
```
superlopuh, line 4: 🤔

tests/filecheck/transforms/arith-addition-of-same-variables-to-multiply-by-two.mlir
superlopuh, line 1: Can you please put all canonicalization tests in the same file, like for the riscv canonicalization?

webmiche:
Why not `a + a -> a << 1`? I guess a shift would even be faster than an add usually.

superlopuh:
Is this related to the other PR? I would also prefer shifting, but overall it feels like this may be better in an xdsl-gui-specific file somewhere, unless MLIR also has the same canonicalization. The last bit should be easy to test, just run `mlir-opt --canonicalize` on the test file.

webmiche:
> Is this related to the other PR? 

My comment? Not really, just bikeshedding ideas. Are we reimplementing mlir canonicalizations on arith here? Or general ones? I am confused :sweat_smile: 

I guess we might only want to do the same canonicalizations as MLIR, out-of-the box at least.

superlopuh:
No I meant to direct the question to Dalia

superlopuh:
I think we're on the same page, I wouldn't be comfortable adding a canonicalization that's not in MLIR

PR #2167, opened 2024-02-14
interactive : add experimental interactive rewrite framework to support rewrite patterns that we do not want to be registered

dshaaban01:
I had started [opening PR's on arith rewrite patterns ](https://github.com/xdslproject/xdsl/pull/2166) when @math-fehr told me that we may not want all rewrite patterns to be merged into xdsl, as some rewrite patterns may negatively impact performance, and are only of use to the xdsl-gui. 

I have therefore drafted this PR and wanted to get feedback on if this should be merged, or should just be kept for myself/experimental. I propose the following: 

- I created a "canonicalisation_patterns" folder in interactive (for now, just arith.py is the only file but in future may expand) where I would define all the arith rewrite patterns, just relevant for the gui tool there, WITHOUT registering them as a canonicalisation pattern in the arith dialect folder. In this folder, I would also include a variable that maps operations to rewrite patterns (to somehow reproduce the "HasCanonicalizationTrait" mechanism.
- I then created a new pass called `InteractiveIndividualRewrite`  which is a subclass of `IndividualRewrite`. This pass would just be used internally in xdsl-gui. In this folder, I similarly define an `INTERACTIVE_REWRITES_BY_NAME` variable, which is the same concept as `REWRITES_BY_NAME`, except it collects our interactive/experimental rewrites. 
- `ALL_REWRITES` concatenates REWRITES_BY_NAME and INTERACTIVE_REWRITES_BY_NAME ==>  to be used in the `apply` function of `InteractiveIndividualRewrite` . the apply function is identical to the previous one, except now `ALL_REWRITES` is used instead of `REWRITES_BY_NAME` (so that we can include the experimental rewrites). 

I hope this makes sense. I thought it would be cleaner to do it like this, than to hard code it and potentially destroy the functionality we built (wrapping rewrites in passes).


webmiche:
I would be fine with the rewrites in core xdsl, but not being executed by default. However, I can see how that might become confusing for people. So, having them in a bit more separated place sounds like a good idea. I would prefer though to have them in a way that you could just copy paste them around to the arith rewrites and it still works. That makes it easier for us to pivot if we figure out that this was a bad decision.
(Honestly, I did not get all the details of your description, so it might very well be the case that your current design already achieves easy refactoring. If so, perfect. If not, please take a moment to think about it and then ignore my comment if it's too much pain)

xdsl/interactive/transforms/experimental/individual_rewrite_interactive.py
```diff
@@ -0,0 +1,78 @@
...
+    for op_name in set(individual_rewrite.REWRITE_BY_NAMES)
+    | set(INTERACTIVE_REWRITE_BY_NAMES)
+}
+"""
+Concatenates REWRITE_BY_NAMES and INTERACTIVE_REWRITE_BY_NAMES.
+"""
+
+
+@dataclass
+class IndividualRewriteInteractive(individual_rewrite.IndividualRewrite):
```
superlopuh, line 42: It feels a little off to subclass this rewrite that's already pretty specific to the interactive compilation stack, we might as well add the custom arith patterns to the dictionary it uses directly

xdsl/interactive/transforms/canonicalization_patterns/arith.py
superlopuh, line 1: I would say that these aren't canonicalization patterns, since canonicalization has some specific rules, like moving constants to the rhs of binary ops, etc, that we don't want to respect with these rewrites. I'd say that they belong in the same file as the individual rewrite pass, since that's the only pass that'll be able to exercise them anyway

PR #2168, opened 2024-02-15
backend: (riscv) fix scf to riscv scf lowering

superlopuh:
I accidentally used the new register types, as opposed to the old fp types, which prevents us from picking the correct move instruction, I added this to the lowering test.


nazavode (approved):

compor:
LGTM

PR #2171, opened 2024-02-15
interpreter: add infrastructure to collect interpreter traces

superlopuh:
Adds a no-op base observer class, and a subclass to count the number of ops executed during an interpreter run. This will let us gather basic statistics about some code on given inputs, hopefully giving us a hint at the performance.

### DIFF
--- a/xdsl/interpreter.py
+++ b/xdsl/interpreter.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections import Counter
 from collections.abc import Callable, Generator, Iterable
 from dataclasses import dataclass, field
 from typing import (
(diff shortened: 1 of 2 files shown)


xdsl/interpreter.py
```diff
@@ -482,15 +495,20 @@ def register_implementations(
         """
         self._impls.register_from(impls, override=override)
 
+    def _run_op(self, op: Operation, inputs: PythonValues) -> OpImplResult:
```
superlopuh, line 498: centralise the logic for running the impl for an operation

math-fehr:
Oh that's a good idea!
Naming wise, do you think we could rename this to `Listener`, as this is overall the naming we use in the other parts of xDSL (and MLIR)?

nazavode (approved):

math-fehr (approved):
Thanks!

compor (approved):

PR #2179, opened 2024-02-16
installation: include missing stylesheets for gui

miccio-dk:


superlopuh:
Do you know what the tradeoff is between using the Manifest vs pyproject.toml?

superlopuh:
I just did a bit of googling, it looks like we might as well replace our Manifest.in with pyproject.toml

superlopuh:
I'll do that in a separate PR, thank you for the bug fix!

PR #2182, opened 2024-02-17
(transformations): lower onnx.Relu to linalg

kayode-gif:
wip on #2181 

dealing with following issues:
1. new ssa value is created rather than one i'm directly trying to reference from the block
2. format of ir in my xdsl-opt mlir doesn't entirely match my construction

i think i have all the "parts" but i need to refine some stuff

### DIFF
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -2,11 +2,19 @@
 
 %t0, %t1 = "test.op"() : () -> (tensor<3x2xf32>, tensor<3x2xf32>)
 %res_add = onnx.Add(%t0, %t1) {onnx_node_name = "/Add"} : (tensor<3x2xf32>, tensor<3x2xf32>) -> tensor<3x2xf32>
+%t2 = "test.op"() : () -> (tensor<3x4xf32>)
+%res_relu = onnx.Relu(%t2) {onnx_node_name = "/Relu"}: (tensor<3x4xf32>) -> tensor<3x4xf32>
 
 
 // CHECK:       builtin.module {
 // CHECK-NEXT:     %t0, %t1 = "test.op"() : () -> (tensor<3x2xf32>, tensor<3x2xf32>)
 // CHECK-NEXT:     %res_add = tensor.empty() : tensor<3x2xf32>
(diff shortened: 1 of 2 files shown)


xdsl/transforms/convert_onnx_to_linalg.py
```diff
@@ -34,6 +36,37 @@ def match_and_rewrite(self, add: onnx.Add, rewriter: PatternRewriter, /):
...
+                    ),
+                    (
+                        linalg.IteratorTypeAttr.parallel(),
+                        linalg.IteratorTypeAttr.parallel(),
+                    ),
+                    (relu.res.type,),
+                ),
+            )
+        )
+
```
superlopuh, line 68:
```suggestion
        body = Region(Block(arg_types=(f64, f64)))
        affine_map = AffineMapAttr(AffineMap.from_callable(lambda d0, d1: (d0, d1)))
        rewriter.replace_matched_op(
            (
                empty := tensor.EmptyOp((), relu.res.type),
                zero := arith.Constant(FloatAttr(0, f64))
                generic := linalg.Generic(
                    (relu.operand,),
                    (empty.tensor,),
                    body,
                    (
                        affine_map,
                        affine_map,
                    ),
                    (
                        linalg.IteratorTypeAttr.parallel(),
                        linalg.IteratorTypeAttr.parallel(),
                    ),
                    (relu.res.type,),
                ),
            )
        )
        with ImplicitBuilder(body) as (in, out):
            max_op = arith.Maximumf(in, zero.result)
            linalg.YieldOp(max_op.result)
```
The zero op was never inserted, also there's a convenience with the ImplicitBuilder of getting the block arguments with the `with` syntax.

kayode-gif: :)

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -2,11 +2,23 @@
 
 %t0, %t1 = "test.op"() : () -> (tensor<3x2xf32>, tensor<3x2xf32>)
 %res_add = onnx.Add(%t0, %t1) {onnx_node_name = "/Add"} : (tensor<3x2xf32>, tensor<3x2xf32>) -> tensor<3x2xf32>
+%t2 = "test.op"() : () -> (tensor<3x4xf64>)
```
superlopuh, line 5: nit: I would bring this down just for the sake of the diff

superlopuh (approved):

PR #2184, opened 2024-02-18
transformations: add riscv-cse

superlopuh:
I'm not sure if we're going to need it, but I thought I'd implement a basic RISC-V-aware common subexpression elimination pass. I'm still not sure how to get partially register-allocated values play nice with MLIR, we can't arbitrarily fold constant values with the same types, since the register might have been written to between the two definitions, so cse is only valid on unallocated regs.


tests/filecheck/transforms/riscv_cse.mlir
```diff
@@ -0,0 +1,51 @@
...
+riscv_scf.for %arg3 : !riscv.reg<> = %c0 to %c8 step %c1 {
+    riscv_scf.for %arg4 : !riscv.reg<> = %c0 to %c8 step %c1 {
+        %3 = riscv.li 8 : () -> !riscv.reg<>
+        %4 = riscv.mul %3, %arg3 : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
+        %5 = riscv.add %4, %arg5 : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
+        riscv_scf.for %arg5 : !riscv.reg<> = %c0 to %c8 step %c1 {
+            %6 = riscv.li 8 : () -> !riscv.reg<>
+            %7 = riscv.mul %3, %arg3 : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
+            %8 = riscv.add %4, %arg5 : (!riscv.reg<>, !riscv.reg<>) -> !riscv.reg<>
+            %9 = riscv.li 8 : () -> !riscv.reg<>
```
compor, line 18: I'm not sure I understand what happens with this specific `li` operation.
Does it get `cse`'d or `dce`'d?

xdsl/transforms/riscv_cse.py
```diff
@@ -0,0 +1,57 @@
...
+            if operands in op_by_operands:
+                rewriter.replace_op(op, (), op_by_operands[operands].results)
+            else:
+                op_by_operands[operands] = op
+
+
+@dataclass
+class RiscvCommonSubexpressionElimination(ModulePass):
+    """
+    Eliminates common sub-expressions on unallocated registers.
```
compor, line 44: CSE can be very complicated (e.g., using advanced techniques like value numbering).

Can you please clearly highlight what this version operates on, e.g. in a sort of list of operations and conditions?

```
Reusing:
- `li` operations with the same attribute
- `Pure` operations with the same operands
```

compor (approved):

tests/filecheck/transforms/riscv_cse.mlir
superlopuh: Good catch! I think it gets csed, since there's another `li 8` in the same block

xdsl/transforms/riscv_cse.py
superlopuh: I'm not sure I understand the difference between your comment and the two lines below

compor: See below.

PR #2187, opened 2024-02-19
interactive: moves the rewrite to pass conversion from app.py into get_all_possibl…

dshaaban01:
moves the rewrite to pass conversion from app.py into get_all_possible_rewrites


xdsl/interactive/get_all_possible_rewrites.py
```diff
@@ -22,6 +25,31 @@ class IndexedIndividualRewrite(NamedTuple):
     rewrite: IndividualRewrite
 
 
+def convert_indexed_individual_rewrites_to_available_pass(
```
superlopuh, line 28: now that this is in a file that doesn't depend on the UI, could you please add a test that exercises this function, and doesn't depend on any passes in xDSL?

tests/interactive/test_rewrites.py
```diff
@@ -0,0 +1,120 @@
...
+                operation_index=2,
+                rewrite=IndividualRewrite(operation="test.op", pattern="TestRewrite"),
+            )
+        ),
+    )
+
+    res = get_all_possible_rewrites(module, {"test.op": {"TestRewrite": Rewrite()}})
+    assert res == expected_res
+
+
```
dshaaban01, line 65: everything after line 66 is new, everything before was in a previously named file test_get_all_possible_rewrites(). I renamed the file to test_rewrites().

superlopuh (approved):

PR #2189, opened 2024-02-19
dialects: add linalg.fill 

kayode-gif:


xdsl/dialects/linalg.py
```diff
@@ -416,12 +416,55 @@ def __init__(
...
+        else:
+            result_types = res
+
+        super().__init__(
+            operands=(inputs, outputs),
+            result_types=result_types,
+        )
+        # Numeric casting is performed on the value operand, promoting it to the same data type as the output.
+        if outputs:
+            self.inputs = [cast(outputs[0].type, operand) for operand in inputs]
```
superlopuh, line 458: this is not needed, the casting is at runtime

superlopuh: The meaning is that you can fill a tensor of f64 with inputs of type f32, these types don't need to match. I don't know what kind of validation MLIR does on compatibility of types.

superlopuh (approved):

superlopuh:
As you can see from the CI output, the fill operation in MLIR expects a scalar input, not a tensor. We might want to implement the same verification. The best thing to do would be to look at the MLIR source code to find the verifier for the fill op

xdsl/dialects/linalg.py
kayode-gif: ahh that makes sense, thanks

tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/invalid_ops.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
```
superlopuh, line 1:
```suggestion
// RUN: xdsl-opt --verify-diagnostics %s | filecheck %s
```
This is to test our own verifier, right? Not mlir-opt

tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/ops.mlir
```diff
@@ -38,9 +45,11 @@ linalg.generic {indexing_maps = [affine_map<(d0, d1) -> ()>, affine_map<(d0, d1)
...
 // CHECK-NEXT:    %1:2 = "test.op"() : () -> (tensor<2x3xf32>, tensor<2x3xf32>)
 // CHECK-NEXT:    %2 = linalg.add ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#1 : tensor<2x3xf32>) -> tensor<2x3xf32>
-// CHECK-NEXT:    %3 = linalg.generic {indexing_maps = [#map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#0 : tensor<2x3xf32>) {
+// CHECK-NEXT:    %cst = arith.constant 0.000000e+00 : f32
+// CHECK-NEXT:    %3 = linalg.fill ins(%cst : f32) outs(%1#0 : tensor<2x3xf32>) -> tensor<2x3xf32>
+// CHECK-NEXT:    %4 = linalg.generic {indexing_maps = [#map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#0 : tensor<2x3xf32>) {
 // CHECK-NEXT:    ^bb0(%in: f32, %in_0: f32, %out: f32):
-// CHECK-NEXT:      %4 = arith.addf %in, %in_0 : f32
-// CHECK-NEXT:      linalg.yield %4 : f32
+// CHECK-NEXT:      %5 = arith.addf %in, %in_0 : f32
```
superlopuh, line 52: this sort of diff is why I usually append to the end of the test file

superlopuh, line 52:
```suggestion
// CHECK-NEXT:      %{{.*}} = arith.addf %in, %in_0 : f32
```
But if we want the generic to be at the end of the test file then let's fix this now

tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/ops.mlir
```diff
@@ -38,9 +45,11 @@ linalg.generic {indexing_maps = [affine_map<(d0, d1) -> ()>, affine_map<(d0, d1)
 // CHECK-NEXT:    }
 // CHECK-NEXT:    %1:2 = "test.op"() : () -> (tensor<2x3xf32>, tensor<2x3xf32>)
 // CHECK-NEXT:    %2 = linalg.add ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#1 : tensor<2x3xf32>) -> tensor<2x3xf32>
-// CHECK-NEXT:    %3 = linalg.generic {indexing_maps = [#map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#0 : tensor<2x3xf32>) {
+// CHECK-NEXT:    %cst = arith.constant 0.000000e+00 : f32
+// CHECK-NEXT:    %3 = linalg.fill ins(%cst : f32) outs(%1#0 : tensor<2x3xf32>) -> tensor<2x3xf32>
+// CHECK-NEXT:    %4 = linalg.generic {indexing_maps = [#map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#0 : tensor<2x3xf32>) {
```
superlopuh, line 50: Sorry, this one also would be great
```suggestion
// CHECK-NEXT:    %{{.*}} = linalg.generic {indexing_maps = [#map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%1#0, %1#0 : tensor<2x3xf32>, tensor<2x3xf32>) outs(%1#0 : tensor<2x3xf32>) {
```

superlopuh: That way we eliminate future spurious diffs

gabrielrodcanal:
Haven't really checked the custom format in the operation definition in MLIR, but I've found an example in which `-> type($res)` is optional (I'm referring to https://github.com/xdslproject/xdsl/blob/ae2054d727e261e8873115e182d8bb53bc7ca0f9/xdsl/dialects/linalg.py#L439). See the example here: https://github.com/Xilinx/mlir-air/blob/82d23361cb3843e5d8fc01d746d513f25a224125/test/airhost/51_air_mmult_2x2_channel_broadcast/air.mlir#L30

I'm not sure if this was a mistake or it's just provisional until the cutom format API supports optional groups, but just in case I thought I'd have you know!

superlopuh:
Oh interesting, should be easy to fix, can you please file an issue with this info?

gabrielrodcanal:
See #2228.

PR #2191, opened 2024-02-19
dialects: (memref) add custom syntax to memref.alloc

superlopuh:


xdsl/dialects/memref.py
```diff
@@ -147,15 +149,29 @@ class Alloc(IRDLOperation):
...
+        result_type: Attribute,
+        alignment: Attribute | None = None,
+    ):
+        super().__init__(
+            operands=(dynamic_sizes, symbol_operands),
+            result_types=(result_type,),
+            properties={"alignment": alignment},
+        )
+
+    @classmethod
```
superlopuh, line 165: can deprecate this in the future

JosseVanDelm: Dumb question: Why is this suddenly a classmethod?

xdsl/dialects/memref.py
```diff
@@ -185,6 +200,67 @@ def verify_(self) -> None:
                 "op dimension operand count does not equal memref dynamic dimension count."
             )
 
+    def print(self, printer: Printer):
+        printer.print_string("(")
+        printer.print_list(self.dynamic_sizes, printer.print_ssa_value)
+        printer.print_string(")")
```
JosseVanDelm, line 206: Another dumb question, is this not possible with custom-format printing/parsing?

JosseVanDelm (approved):
Looks good, I just have some questions

xdsl/dialects/memref.py
superlopuh: I don't think we handle not printing the delimiters of an optional list if it's empty. CC @math-fehr @PapyChacal is this correct?

xdsl/dialects/memref.py
superlopuh: That was a remainder of a WIP change, I can change it back now, I just felt like it didn't hurt.

xdsl/dialects/memref.py
math-fehr: Soon TM, for now we can't do that :(
But @PapyChacal is working on this AFAIK

PR #2193, opened 2024-02-19
transformations: add "generic" parameter to MLIROptPass

### LINKED ISSUE #2120: enhancement: add option to MlirOptPass to print in custom or generic format
Currently only prints in generic format, but sometimes custom is preferred.

superlopuh:
Fixes #2120

I'm not sure how to test this, we need to somehow intercept that the syntax passed to mlir-opt is in a certain format, but we pass it along directly... I think just testing that it works with true, false, and nothing is enough for now.


PapyChacal (commented):
Why not always use generic syntax in this pass instead?

superlopuh:
Sometimes we have ops it doesn't know about

superlopuh:
It was an op I added in a WIP branch, now merged (tensor.empty). I found that the quickest way to get what I wanted to work to work was to use custom syntax, so I'd still like to add this functionality for the future.

PapyChacal:
Sure why not, I sounded a bit assertive, apologies for that!

I'm just concerned about pushing too much on the custom syntax and willing to prioritize a bullet-proof generic syntax interoperability; I'm all for improving custom syntax and having that kind of flexibility too though 🙂

Thanks for the discussion!

PR #2195, opened 2024-02-19
(transformations): lower onnx.Constant to linalg

kayode-gif:
wip #2194

unsure on how to deal with the attributes of onnx.constant, also constant takes no input 
so not sure whether to just pass in an zero constant

### DIFF
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -22,10 +22,13 @@
 // CHECK-NEXT:    } -> tensor<3x4xf64>
 
 %res_constant = "onnx.Constant"() {onnx_node_name = "/Constant", "value" = dense<1> : tensor<1xi64>}: () -> tensor<1xi64>
-// CHECK-NEXT:  %res_constant = ml_program.global_load_const @global_constant : tensor<1xi64>
-
 %res_constant_2 = "onnx.Constant"() {onnx_node_name = "/Constant", "value" = dense<2.0> : tensor<1x5xf32>} : () -> tensor<1x5xf32>
-// CHECK-NEXT:    %res_constant_2 = ml_program.global_load_const @global_constant : tensor<1x5xf32>
-// CHECK-NEXT:    ml_program.global private @global_constant(dense<2.000000e+00> : tensor<1x5xf32>) : tensor<1x5xf32>
+
+// CHECK-NEXT: %res_constant = ml_program.global_load_const @onnx_constant_1 : tensor<1xi64>
+// CHECK-NEXT: %res_constant_2 = ml_program.global_load_const @onnx_constant_2 : tensor<1x5xf32>
+// CHECK-NEXT: ml_program.global private @onnx_constant_1(dense<1> : tensor<1xi64>) : tensor<1xi64>
+// CHECK-NEXT: ml_program.global private @onnx_constant_2(dense<2.000000e+00> : tensor<1x5xf32>) : tensor<1x5xf32>
+
+
 // CHECK-NEXT:  }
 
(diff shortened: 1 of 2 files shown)


superlopuh:
This is one of the weirder ones, the best way might be to port the global op from here: https://mlir.llvm.org/docs/Dialects/MLProgramOps/#ml_programglobal-ml_programglobalop

kayode-gif:
> This is one of the weirder ones, the best way might be to port the global op from here: https://mlir.llvm.org/docs/Dialects/MLProgramOps/#ml_programglobal-ml_programglobalop

this isn't within xdsl right? and by port you mean use this to be a placeholder for my input?

superlopuh:
Right, it would involve adding a new dialect and op as a prerequisite to this lowerint

xdsl/transforms/convert_onnx_to_linalg.py
```diff
@@ -67,6 +75,28 @@ def match_and_rewrite(self, relu: onnx.Relu, rewriter: PatternRewriter, /):
             linalg.YieldOp(max_op.result)
 
 
+@dataclass
+class ConstantOpLowering(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, constant: onnx.Constant, rewriter: PatternRewriter, /):
+        attr_value = list(constant.attributes.values())[1]
+        global_op = ml_program.Global(
```
superlopuh, line 83: Need to insert this also, something like this:

```
SymbolTable.insert_or_update(module_op, func_op)
```

We don't have a best way to get the module op, probably just find it by getting parent_op until there are no more

kayode-gif: i don't understand why this is needed, what does it do?
q1. would it be something like: `SymbolTable.insert_or_update(constant, global_op)`
since constant is the "module op",  
q1b -  is this added to the apply bit instead, since i see that does moduleop
q2. "probably just find it by getting parent_op until there are no more" - recursive helper function right?
thanks

superlopuh: q1: no, the first argument is the op in whose region the second argument will be inserted. Sorry, I changed the `op` in my copied code to module_op, but not the second arg, which should be `global_op`.
q1b: in the apply bit, we no longer have the data to insert, so it has to be done inside the pattern
q2: yeah recursive helper function is probably best. @math-fehr, do you have any ideas/advice for this or is it best to just manually roll one?

superlopuh: @kayode-gif as a first step, just make it work, then we can make the code pretty :)

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -22,8 +22,8 @@
 // CHECK-NEXT:    } -> tensor<3x4xf64>
 
 %res_constant = "onnx.Constant"() {onnx_node_name = "/Constant", "value" = dense<1> : tensor<1xi64>}: () -> tensor<1xi64>
```
superlopuh, line 24: can you please add a second constant with a different name and shape to the test?

superlopuh: We need the constants to have different names, since they'll have different values. Can you please modify the lowering to make sure that the names are unique? My first thought is to reuse the onnx_node_name, but I'm not sure how that plays with the restriction on symbol names in MLIR. Another possible solution is `onnx_constant_1`, 2, 3, etc

kayode-gif: ahhh, my bad, i misunderstood, it should be sorted now :)

superlopuh (approved):

PR #2202, opened 2024-02-21
transformations: Remove target argument from stencil conversion

PapyChacal:
No need to specialize.
This was done solely to map the fastest varying dimension to the right CUDA dimension; now done by xDSL's inhouse `gpu-map-parallel-loops` in a simpler and independent manner :slightly_smiling_face:


tests/filecheck/dialects/stencil/hdiff_gpu.mlir
georgebisbas, line 1: Why is this totally dropped?

PapyChacal: It was testing the gpu lowering; this removes the GPU lowering

PapyChacal:
> Looks good, do we expect any perf diff here?

No, that's why I implemt `gpu-map-parallel-loops` in xDSL.
This is sufficient to end up generating the same CUDA code.

AntonLydike (approved):
Noice!

PR #2203, opened 2024-02-21
dialects: Add LLVM InlineAsm op

JosseVanDelm:
This PR attempts to add the LLVM InlineAsm op, which allows to bake in inline assembly from within MLIR.

Some things I noted while making this:
* The docs mention this:
   > The InlineAsmOp mirrors the underlying LLVM semantics with a notable exception: the embedded asm_string is not    allowed to define or reference any symbol or any global variable: only the operands of the op may be read, written, or    referenced. Attempting to define or reference any symbol or any global behavior is considered undefined behavior at    this time.
   
   I'm assuming this means that this constraint is never actually checked/verified :sweat_smile: ?
* There's an additional attribute to select the inline assembly dialect (not dialect in context of MLIR, choices are AT&T or Intel). I just implemented this as an `IntegerAttr`, is that okay? It looked like `Linkage` (prior to this PR) was/is implemented differently from upstream in a similar way?

Currently I have only implemented round-trip tests.

Happy to add any other tests if required.

CC @jorendumoulin

### DIFF
--- a/xdsl/dialects/llvm.py
+++ b/xdsl/dialects/llvm.py
@@ -723,7 +723,6 @@ class InlineAsmOp(IRDLOperation):
 
     name = "llvm.inline_asm"
 
-    # operands_, because is already defined?
     operands_: VarOperand = var_operand_def()
 
     res: OptOpResult = opt_result_def()
@@ -733,7 +732,7 @@ class InlineAsmOp(IRDLOperation):
     # 0 for AT&T inline assembly dialect
     # 1 for Intel inline assembly dialect
     # In this context dialect does not refer to an MLIR dialect
-    asm_dialect = opt_prop_def(IntegerAttr[IntegerType])
+    asm_dialect = opt_prop_def(IntegerAttr[Annotated[IntegerType, IntegerType(64)]])
 
     asm_string: StringAttr = prop_def(StringAttr)
     constraints: StringAttr = prop_def(StringAttr)
--- a/xdsl/dialects/llvm.py
+++ b/xdsl/dialects/llvm.py
@@ -740,6 +740,32 @@ class InlineAsmOp(IRDLOperation):
     has_side_effects: UnitAttr | None = opt_prop_def(UnitAttr)
     is_align_stack: UnitAttr | None = opt_prop_def(UnitAttr)
 
+    def __init__(
+        self,
+        operands_: list[SSAValue | Operation],
+        res_types: list[Attribute],
+        asm_string: str,
+        constraints: str,
+        asm_dialect: int = 0,
+        has_side_effects: bool = False,
+        is_align_stack: bool = False,
+    ):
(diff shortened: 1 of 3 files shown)


superlopuh:
Can you please add a test for the roundtrip with MLIR?

JosseVanDelm:
Added @superlopuh !

xdsl/dialects/llvm.py
```diff
@@ -712,6 +712,61 @@ def __init__(self, input: SSAValue | Operation, ptr_type: Attribute | None = Non
...
+        props: dict[str, Attribute] = {
+            "asm_string": StringAttr(asm_string),
+            "constraints": StringAttr(constraints),
+            "asm_dialect": IntegerAttr.from_int_and_width(asm_dialect, 64),
+        }
+        if has_side_effects:
+            props["has_side_effects"] = UnitAttr()
+        if is_align_stack:
+            props["is_align_stack"] = UnitAttr()
+
```
superlopuh, line 762:
```suggestion
            "asm_dialect": IntegerAttr.from_int_and_width(asm_dialect, 64),
            "has_side_effects": UnitAttr() if has_side_effects else None,
            "is_align_stack": UnitAttr() if is_align_stack else None,
        }
```

superlopuh: can then probably move the dict literal into the super init call

superlopuh (approved):

xdsl/dialects/llvm.py
```diff
@@ -712,6 +712,61 @@ def __init__(self, input: SSAValue | Operation, ptr_type: Attribute | None = Non
...
+    """
+    https://mlir.llvm.org/docs/Dialects/LLVM/#llvminline_asm-llvminlineasmop
+
+    To see what each field means, have a look at:
+    https://llvm.org/docs/LangRef.html#inline-assembler-expressions
+    """
+
+    name = "llvm.inline_asm"
+
+    # operands_, because is already defined?
```
superlopuh, line 726: Can probably remove this comment

xdsl/dialects/llvm.py
JosseVanDelm: Pyright throws an error if I do that, cause props is not allowed to have a None as value in the dictionary.
I've seen some other shorthands here, but just like the poster here I prefer the original code:
https://stackoverflow.com/questions/14263872/only-add-to-a-dict-if-a-condition-is-met

xdsl/dialects/llvm.py
JosseVanDelm: Done!

xdsl/dialects/llvm.py
superlopuh: Wait a second, properties can't be missing, can you please check that in MLIR these are properties and not attributes by printing the generic syntax?

PR #2207, opened 2024-02-21
interactive: remove passes that raise errors from condensed list

superlopuh:
I have a feeling this should have been the case from the get-go. The main reason to condense is to show the available transformations, to make it quick and easy to go towards your target. There are three reasons why a pass might fail:
1. it's mlir-opt and it has weird stdout/stderr behaviour that breaks textual
2. it's a pass that requires parameters to run
3. it actually fails

I would say that none of these are useful by themselves to show in the "quick and easy" list. If the user knows to use a certain pass, they can always "uncondense" and pick the pass that they want, possibly with the parameters they're interested int. The passes that currently require parameters should probably either provide default parameters or provide a range of options to try in the future, but that can be done by another PR.

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -20,16 +20,12 @@ from xdsl.ir import Block, Region
 from xdsl.transforms import (
     canonicalize,
     individual_rewrite,
-    mlir_opt,
     printf_to_llvm,
-    scf_parallel_loop_tiling,
-    stencil_unroll,
     test_lower_linalg_to_snitch,
 )
 from xdsl.transforms.experimental import (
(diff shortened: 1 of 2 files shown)


PapyChacal:
Why?

dshaaban01 (approved):

PapyChacal:
Thanks for the writeup!
So the Chang is that, if a pass is raising on the current input, it won't be in the list of advised passes, is that right?

superlopuh:
Exactly. The previous filter was pessimistic, meaning we only remove things that we know don't change anything. My proposed list is optimistic, meaning we only keep things that we know change something.

PR #2209, opened 2024-02-21
dialects: (ml_program) add ml_program.global and ml_program.global_load_const

kayode-gif:
see https://mlir.llvm.org/docs/Dialects/MLProgramOps/#ml_programglobal-ml_programglobalop.

I'm sure a custom printer and parser is required in order to get this going, however i'm not sure how to start with printer and parsing this one.


superlopuh:
Can you please add `ml_program.global_load_const` also

kayode-gif:
i think the issue is with ```value = parse_attribute()``` perhaps, since that attribute expects a type?

kayode-gif:
> i think the issue is with `value = parse_attribute()` perhaps, since that attribute expects a type?

edit: i wasn't passing the a type attribute needed too

xdsl/dialects/ml_program.py
```diff
@@ -0,0 +1,150 @@
...
+        else:
+            sym_visibility = StringAttr("")
+        if parser.parse_optional_keyword("mutable"):
+            mutability = UnitAttr()
+        sym_name = parser.parse_attribute()
+        if parser.parse_optional_punctuation("("):
+            value = parser.parse_attribute()
+            parser.parse_punctuation(")")
+        parser.parse_punctuation(":")
+        parser.parse_type()
```
superlopuh, line 81:
```suggestion
        type = parser.parse_type()
```

xdsl/dialects/ml_program.py
```diff
@@ -0,0 +1,150 @@
...
+
+        global_op = cls(
+            sym_name=sym_name,
+            type_attr=value.type,
```
superlopuh, line 85:
```suggestion
            type_attr=type,
```

xdsl/dialects/ml_program.py
```diff
@@ -0,0 +1,158 @@
...
+            sym_visibility = StringAttr("public")
+        elif parser.parse_optional_keyword("nested"):
+            sym_visibility = StringAttr("nested")
+        elif parser.parse_optional_keyword("private"):
+            sym_visibility = StringAttr("private")
+        else:
+            sym_visibility = StringAttr("")
+        is_mutable = UnitAttr()
+        sym_name = parser.parse_symbol_name().data
+        if parser.parse_optional_punctuation("("):
```
superlopuh, line 86: I don't think this is optional

kayode-gif: why isnt it, we can have a case like this: ```ml_program.global mutable @foobar : tensor<?xi32>```

xdsl/dialects/ml_program.py
```diff
@@ -0,0 +1,168 @@
...
+    def parse(cls, parser: Parser) -> Self:
+        attrs = parser.parse_optional_attr_dict()
+        if parser.parse_optional_keyword("public"):
+            sym_visibility = StringAttr("public")
+        elif parser.parse_optional_keyword("nested"):
+            sym_visibility = StringAttr("nested")
+        elif parser.parse_optional_keyword("private"):
+            sym_visibility = StringAttr("private")
+        else:
+            sym_visibility = StringAttr("")
```
superlopuh, line 92: This isn't a legal value for visibility. In MLIR, this op has the `Symbol` interface, which we represent with `SymbolOpInterface`. Could you please add that to this op's traits? We'll need it later to find the relevant `global` op when using `global_load_context`

xdsl/dialects/ml_program.py
```diff
@@ -0,0 +1,171 @@
...
+    def parse(cls, parser: Parser) -> Self:
+        attrs = parser.parse_optional_attr_dict()
+        if parser.parse_optional_keyword("public"):
+            sym_visibility = StringAttr("public")
+        elif parser.parse_optional_keyword("nested"):
+            sym_visibility = StringAttr("nested")
+        elif parser.parse_optional_keyword("private"):
+            sym_visibility = StringAttr("private")
+        else:
+            raise VerifyException("Expected 'public', 'private', or 'nested'")
```
superlopuh, line 95: Here we should raise a ParseError instead

kayode-gif: is there a difference with this and ```parser.raise_error```

superlopuh: `raise_error` is preferred, you're right

PR #2210, opened 2024-02-21
core: add Hasher and HashableModule

superlopuh:
These utils are planned to be used in `xdsl-gui`, but may well be useful outside of that project. The Hasher contains logic I've used a lot before, it seems to work well enough. In the future, we might want a more efficient way to compute equality/inequality for modules, but we might as well add a working helper class now and optimise later.

### DIFF
--- a/tests/utils/test_hashable_module.py
+++ b/tests/utils/test_hashable_module.py
@@ -1,0 +1,33 @@
+from collections.abc import Iterable
+
+from xdsl.builder import ImplicitBuilder
+from xdsl.dialects.builtin import ModuleOp, StringAttr
+from xdsl.dialects.test import TestOp
+from xdsl.utils.hashable_module import HashableModule
+
+
+def m(labels: Iterable[str]):
+    module = ModuleOp([])
+    with ImplicitBuilder(module.body):
+        for label in labels:
+            TestOp(attributes={"label": StringAttr(label)})
+    return module
+
+
+def test_hashable_module():
+    assert HashableModule(ModuleOp([])) == HashableModule(ModuleOp([]))
+    assert hash(HashableModule(ModuleOp([]))) == hash(HashableModule(ModuleOp([])))
+
+    a0 = m("a")
+    a1 = m("a")
+
+    assert HashableModule(a0) == HashableModule(a1)
+    assert hash(HashableModule(a0)) == hash(HashableModule(a1))
+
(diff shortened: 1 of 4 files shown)


tests/utils/test_hashable_module.py
```diff
@@ -0,0 +1,33 @@
...
+from xdsl.dialects.test import TestOp
+from xdsl.utils.hashable_module import HashableModule
+
+
+def m(labels: Iterable[str]):
+    module = ModuleOp([])
+    with ImplicitBuilder(module.body):
+        for label in labels:
+            TestOp(attributes={"label": StringAttr(label)})
+    return module
```
AntonLydike, line 14: I think having a more descriptive name here would be helpful

AntonLydike (commented):
I feel that this is quite a convoluted way of implementing a `__hash__`.

What would be the downside of computing a key (tuple of operation names) and calling `hash` on that? I feel that this might result in much simpler code. (It would also introduce less new code that needs testing)

Furthermore, we should take into account that hashes of mutable objects may result in unexpected/broken behavior, I think having a clearer warning on the HashableModule class would be helpful in communicating the problems associated with that.

superlopuh:
I feel like it's pretty standard, it lets users create a hash of complex objects without having all of the things that compose the hash in memory at a time, making it much more efficient than a hash of a tuple of ops. I also don't feel like it's that complicated, I guess that depends on how used to it the reader is. Swift's hash works like this, so I guess I'm used to it. I also didn't want to go just with xor since that drops the order of elements, but I guess few rewrites just reorder operations, so maybe that should be fine for the sake of hashing in the short term.

dshaaban01 (approved):

math-fehr (approved):
Nice!
I wouldn't use just a xor here, as it will have a lot of collisions (to be exact, if you have the same parity of every operation, then it will have the same hash, the order won't matter at all).
But that's something we can always fix later, so I'm fine with merging it now!

AntonLydike:
> I wouldn't use just a xor here, as it will have a lot of collisions (to be exact, if you have the same parity of every operation, then it will have the same hash, the order won't matter at all).

I totally agree, we can replace the xor with `hash((self.hash, other_hash))` for it to fall back to a known good hash instead of rolling our own.

superlopuh:
Good idea!

PR #2211, opened 2024-02-21
transformations: insert function declarations only if called

superlopuh:


dshaaban01 (approved):

xdsl/transforms/experimental/hls_convert_stencil_to_ll_mlir.py
```diff
@@ -1208,24 +1208,14 @@ def match_and_rewrite(self, op: scf.ParallelOp, rewriter: PatternRewriter, /):
 class QualifyInterfacesPass(RewritePattern):
     module: builtin.ModuleOp
     declared_coeff_func: bool = False
+    called_coeff_func: bool = False
```
math-fehr, line 1211: This is read but never set anywhere? Is there some use of it missing?

xdsl/transforms/printf_to_llvm.py
```diff
@@ -145,9 +146,21 @@ class PrintfToLLVM(ModulePass):
     name = "printf-to-llvm"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        has_done_action = False
+
+        def handler(op: Operation, new_results: Sequence[SSAValue | None]):
+            nonlocal has_done_action
+            has_done_action = True
```
math-fehr, line 153: This technically works, but I would instead override the `operation_insertion_handler`, and check if one with the `CallOp("printf")` is being inserted.
Just to make this future-proof and avoiding bug.

I would also rename `has_done_action` with something else, just to make it clear why we have this variable. Something like "has_inserted_printf_call"?

math-fehr (approved):

xdsl/transforms/experimental/hls_convert_stencil_to_ll_mlir.py
superlopuh: Ah yes, good catch

AntonLydike (changes requested):
I would try to avoid adding a handler here and look for simpler solutions to the problem. This seems to be a relatively common problem, handled differently by different rewrites.

AntonLydike:
Also, there are some interactive changes muddled in with this PR, which feels a bit weird, can you give a quick summary of those changes too?

xdsl/transforms/printf_to_llvm.py
superlopuh: The rewrite patterns are ideally stateless. Having state in them makes it much more difficult to reason about. I know that we currently don't have this as a strict restriction, but it would be good not to add extra state to existing patterns if we can avoid it. To me it makes intuitive sense that the Rewriter should track whether it's done an action, as opposed to a pattern.

superlopuh:
The interactive changes are due to the fact that the tests there are not well isolated wrt the rest of the framework. They are kind of integration tests in that they check what will be displayed when the user launches the app. It would be good to refactor them in the future such that they are independent of the passes in xDSL (cc @dshaaban01 ) but I'd say that's not a priority right now.

superlopuh:
I found another way

AntonLydike (approved):
LGTM, thanks for giving it another go!

PR #2215, opened 2024-02-22
interactive: change ListView to 1-level Tree

dshaaban01:
Changing the list view where you can select passes to a 1-level tree . it essentially does the exact same thing as before, but the widget has changed. next PR to add levels and make it a true "tree"

### DIFF
--- a/xdsl/interactive/app.py
+++ b/xdsl/interactive/app.py
@@ -19,14 +19,7 @@ from textual.app import App, ComposeResult
 from textual.containers import Horizontal, ScrollableContainer, Vertical
 from textual.reactive import reactive
 from textual.screen import Screen
-from textual.widgets import (
-    Button,
-    DataTable,
-    Footer,
-    Label,
-    ListView,
-    TextArea,
-)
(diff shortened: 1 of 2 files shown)


xdsl/interactive/app.py
```diff
@@ -216,15 +206,14 @@ def on_mount(self) -> None:
...
-        # initialize ListView to contain the pass options
+        # initialize Tree to contain the pass options
         for n, module_pass in ALL_PASSES:
-            self.passes_list_view.append(
-                PassListItem(Label(n), module_pass=module_pass, pass_spec=None, name=n)
+            self.passes_tree.root.add(
+                label=n,
+                data=PassListItem(
+                    Label(n), module_pass=module_pass, pass_spec=None, name=n
+                ),
```
superlopuh, line 216: I'm a little confused. It seems like the tree API is different to that of the list, where Tree Data can be anything, whereas the List requires a ListItem. Now we're duplicating the label, which seems unnecessary. Can you please exctract the "data" part of the PassListItem, separate from the label, into a NamedTuple, and pass that to the tree node instead?

superlopuh (approved):

PR #2216, opened 2024-02-22
transformations: (riscv_func) handle external function declarations

superlopuh:
I'm not 100% sure what the lowering should be. Godbolt seems to drop external function declarations, should we skip them during assembly printing?


AntonLydike (approved):
Very nice stuff! Really nice to see this mature! :+1:

AntonLydike:
> I'm not 100% sure what the lowering should be. Godbolt seems to drop external function declarations, should we skip them during assembly printing?

I would print them, as far as I'm aware, these are required for the linker. Not sure why goodbolt drops them.

PR #2217, opened 2024-02-22
interactive: make some structural changes that will help tree PR 2

dshaaban01:
To reuse the functionality we already have in the tool where we ask the user to input pass arguments, we make some structural changes so that we can reuse it in future with the tree structure (node expansion). 

self.get_pass_arguments() function computes the pass tuple with the arguments, in a format ready to be added to the pass_pipeline. Actually adding it to the pass_pipeline happens in the update_pass_pipeline() function instead of in the self.get_pass_arguments() function (the way it used to be). 

Since functions to do with screens cannot return the variable (i.e the add_pass_with_arguments_to_pass_pipeline), I create a global `current_argument_pass` variable that holds the variable so that it can be accessed later.


xdsl/interactive/app.py
```diff
@@ -126,6 +126,11 @@ class InputApp(App[None]):
     text areas.
     """
 
+    current_argument_pass: tuple[tuple[type[ModulePass], PipelinePassSpec], ...] = ()
```
superlopuh, line 129: Why do we need this new variable? Can we not just get the last element of the pass pipeline?

dshaaban01: I explained in the description above - 

"Since functions to do with screens cannot return the variable (i.e the add_pass_with_arguments_to_pass_pipeline), I create a global current_argument_pass variable that holds the variable so that it can be accessed later."


- I think that we do not necessarily want to add this to the pass pipeline in the future when doing multi-level-traversal. no?

superlopuh: I don't understand what you mean here

dshaaban01: Okay I have removed it. Is this better?

superlopuh (approved):

PR #2218, opened 2024-02-22
interactive: add update_selected_passes_list_view helper function PR 3

dshaaban01:
Now starts a series of mini-pr's that will add helper functions that will help make the multi-level tree traversal. please trust the vision that we need these helper functions!

update_selected_passes_list_view() is passes a pipeline_variable because in the future we will not want it to contain the last pass in the pipeline (as that will be the root of the tree).


xdsl/interactive/app.py
```diff
@@ -274,6 +283,26 @@ def watch_available_pass_list(
                     ),
                 )
 
+    def update_selected_passes_list_view(
+        self, pass_pipeline: tuple[tuple[type[ModulePass], PipelinePassSpec], ...]
```
superlopuh, line 287: if this is always called with `self.pass_pipeline`, it feels like we can drop the function argument

xdsl/interactive/app.py
```diff
@@ -357,16 +386,7 @@ def watch_pass_pipeline(self) -> None:
...
-            self.selected_passes_list_view.append(
-                PassListItem(
-                    Label(pass_value.name),
-                    module_pass=pass_value,
-                    pass_spec=value_spec,
-                    name=pass_value.name,
-                )
-            )
+        self.update_selected_passes_list_view(self.pass_pipeline)
         self.update_current_module()
```
superlopuh, line 390: I had a look into whether it's possible to have multiple watchers for the same reactive variable, and it seems a little tedious, so calling a few methods seems best.

dshaaban01: So this is okay?

xdsl/interactive/app.py
```diff
@@ -262,6 +271,24 @@ def watch_available_pass_list(
                     data=(value, value_spec),
                 )
 
+    def update_selected_passes_list_view(self) -> None:
+        """
+        Helper function that updates the selected passes ListView to display the passes in pass_pipeline.
+        """
+        self.selected_passes_list_view.clear()
+        if self.pass_pipeline != ():
```
superlopuh, line 279:
```suggestion
        if self.pass_pipeline:
```

superlopuh (approved):

PR #2220, opened 2024-02-22
interactive: add helper function get_root_to_child_pass_list PR 5

dshaaban01:
PR 3

`get_root_to_child_pass_list` computes the path from root to expanded_node. this will be useful to make multi-level tree traversal possible, so that we can i) compute the set of available passes at an expanded node (we would need to get_condensed_pass_list based on the assumption that the root_to_child_pass_list is added to the pass_pipeline) or ii) if node is selected multi-levels in, we can directly add the root_to_child_pass_list to the pass pipeline.


superlopuh (commented):
can you add a test, please?

xdsl/interactive/app.py
```diff
@@ -268,6 +268,38 @@ def watch_available_pass_list(
...
+        ] = ()
+        current = expanded_node
+
+        # traverse the path starting from the child node until we reach the root
+        while current is not None and current.data is not None and not current.is_root:
+            selected_pass_value, selected_pass_spec = current.data
+            pass_list_items = (
+                *pass_list_items,
+                (selected_pass_value, selected_pass_spec),
+            )
```
superlopuh, line 291:
```suggestion
            pass_list_items.append(current.data)
```

xdsl/interactive/app.py
```diff
@@ -268,6 +268,33 @@ def watch_available_pass_list(
             self.passes_tree.clear()
             self.expand_node(self.passes_tree.root, new_pass_list)
 
+    def get_root_to_child_pass_list(
```
superlopuh, line 271: Can you add some tests, please?

dshaaban01: yes on it.

superlopuh (approved):

PR #2229, opened 2024-02-22
interpreter: add args arg to xdsl-run for entry point function arguments

superlopuh:
Great for tests of microkernels


xdsl/tools/xdsl_run.py
```diff
@@ -51,6 +52,12 @@ def register_all_arguments(self, arg_parser: argparse.ArgumentParser):
             type=str,
             help="Name of function to call.",
         )
+        arg_parser.add_argument(
+            "--args",
+            default="",
+            type=str,
+            help="Arguments to pass to entry function.",
+        )
```
PapyChacal, line 60: Why not positional arguments?
Similar to say the Python interpreter

PapyChacal (commented):
Nitpick interface comment
More generally, I'm confused by the reason of the general attribute-to-Python-values infrastructure here; could you motivate it?

AntonLydike:
I think I'm more a fan of what @PapyChacal suggested, i.e. going `xdsl-run <xdsl-run args> -- <command args>` as this is the standard way of doing things. I would even make the `--` optional and just pass al unconsumed args to the program? I think that's the default way of handling this sorta stuff

superlopuh:
I should have followed my own advice, I split out the Interpreter changes from the `xdsl-run` changes, so this is now on top of #2231, which should make things easier to merge. I'll now have a go at addressing the comments in this PR

superlopuh:
I'm struggling to get the positional arguments to work. We already have a positional argument, which is the input file. If that's missing, then stdin is used. Having a variable-length positional input messes with that logic, as arg_parse no longer knows whether any of the positional arguments are the input file path. If anyone feels like playing with this locally, can you please let me know whether you figure out a way to make it work? Otherwise I'd say that a single comma-separated list of inputs is not a bad API either, that way we use a single Parser for the whole expression, and the weirdness in the CLI argument is contained to one string.

PapyChacal:
Oh wow indeed, it's apparently a long running story that argparse does not support this 🙃
I prefer it still but forcing the initial PR to struggle with that was not the point, I'm fine with merging like this.
I would still use a variadic argument rather than a single string, or is there also a caveat?

superlopuh:
Having a variadic number of arguments means that they have to come last. I feel like that could also get confusing, where you go from piping the IR in, then try to add it as the last arg, only to have the program wait forever for stdin.

compor:
> Having a variadic number of arguments means that they have to come last. I feel like that could also get confusing, where you go from piping the IR in, then try to add it as the last arg, only to have the program wait forever for stdin.

It's not ideal, but that's standard *nix behaviour; the best I've seen is some tools (I really don't remember which), print a wee message like "awaiting input from terminal", but you need to be able to detect if input is indeed from an interactive terminal vs a pipe.

TL;DR I'm fine with the current string ATM.

PR #2230, opened 2024-02-22
interpreter: add index_bitwidth to interpreter

superlopuh:
Some of the high-level function semantics depend on the width of the pointer type. This PR adds this information to the interpreter, making it accessible for interpreter impls. Currently none of the impls leverage this, so we're no worse off than before with this PR in, and can get to fixing the existing bugs.

### DIFF
--- a/docs/Toy/toy/__main__.py
+++ b/docs/Toy/toy/__main__.py
@@ -96,7 +96,7 @@ def main(path: Path, emit: str, ir: bool, print_generic: bool):
         printer.print(module_op)
         return
 
-    interpreter = Interpreter(module_op)
+    interpreter = Interpreter(module_op, index_bitwidth=32)
     if emit in ("toy", "toy-opt", "toy-inline", "toy-infer-shapes"):
         interpreter.register_implementations(ToyFunctions())
     if emit in ("affine"):
--- a/docs/Toy/toy/__main__.py
+++ b/docs/Toy/toy/__main__.py
@@ -96,7 +96,7 @@ def main(path: Path, emit: str, ir: bool, print_generic: bool):
         printer.print(module_op)
         return
 
-    interpreter = Interpreter(module_op, index_bitwidth=32)
+    interpreter = Interpreter(module_op)
     if emit in ("toy", "toy-opt", "toy-inline", "toy-infer-shapes"):
         interpreter.register_implementations(ToyFunctions())
     if emit in ("affine"):
--- a/xdsl/interpreter.py
+++ b/xdsl/interpreter.py
@@ -426,6 +426,10 @@ class Interpreter:
             ...
 
     module: ModuleOp
+    index_bitwidth: int = field(kw_only=True)
+    """
+    Number of bits in the binary representation of the index
+    """
     _impls: _InterpreterFunctionImpls = field(default_factory=_InterpreterFunctionImpls)
     _ctx: InterpreterContext = field(
         default_factory=lambda: InterpreterContext(name="root")
--- a/xdsl/interpreter.py
+++ b/xdsl/interpreter.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import platform
 from collections import Counter
 from collections.abc import Callable, Generator, Iterable
 from dataclasses import dataclass, field
 from typing import (
     IO,
     Any,
+    ClassVar,
     NamedTuple,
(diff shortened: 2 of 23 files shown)


Moxinilian (approved):

xdsl/transforms/constant_fold_interp.py
```diff
@@ -80,7 +79,7 @@ class ConstantFoldInterpPass(ModulePass):
     name = "constant-fold-interp"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
-        interpreter = Interpreter(op)
+        interpreter = Interpreter(op, index_bitwidth=32)
```
AntonLydike, line 82: I'm not sure that this is a sound decision? Constant folding index values on 64 bit platforms will yield wrong results if I'm not mistaken?

PapyChacal: Yep, as discussed privately first, I would default to using platform.architecture to figure out a sound host index size

docs/Toy/toy/__main__.py
```diff
@@ -96,7 +96,7 @@ def main(path: Path, emit: str, ir: bool, print_generic: bool):
         printer.print(module_op)
         return
 
-    interpreter = Interpreter(module_op)
```
PapyChacal, line 99: Also I would use the default wherever possible?

PapyChacal (commented):
I know part of the motivation is having proper buffer management in the interpreter so yeah yeah yeah!!
Just some nitpicks, I think we can reduce the impact a lot and get the benefits still

docs/Toy/toy/__main__.py
```diff
@@ -96,7 +96,7 @@ def main(path: Path, emit: str, ir: bool, print_generic: bool):
...
+    interpreter = Interpreter(module_op, index_bitwidth=32)
```
PapyChacal, line 99: Also I would use the default wherever possible?

xdsl/interpreter.py
```diff
@@ -425,7 +427,16 @@ def will_interpret_op(self, op: Operation, args: PythonValues) -> None:
         def did_interpret_op(self, op: Operation, results: PythonValues) -> None:
             ...
 
+    SYSTEM_BITWIDTH: ClassVar[int | None] = {"64bit": 64, "32bit": 32}.get(
+        platform.architecture()[0]
+    )
```
AntonLydike, line 432: I think you meant to add a default value of `None` to the `.get` here, as referenced in the next line.

AntonLydike (approved):
Thank you for this! Very nice work! I just commented on a minor thing, otherwise LGTM!

xdsl/transforms/constant_fold_interp.py
Moxinilian: > I'm not sure that this is a sound decision? Constant folding index values on 64 bit platforms will yield wrong results if I'm not mistaken?

Why would it?

AntonLydike: Hmm, interesting question @Moxinilian, I dug a bit into [the docs](https://mlir.llvm.org/docs/Dialects/IndexOps/) and the provided example there (bottom of the dialect introduction section) helps to construct an example:

Assuming we try to fold `div(0x100000002, 2)` (or any other operation that right shifts)

- if we are 64 bit, but simulate 32, we truncate then divide, giving us `0x01` as a result
- if we are 32 bit, but simulate 64, we get an upper bit set that would have been truncated, giving us `0x80000001` as a result

MLIR always uses 64 bit values for constant folding `index`, and solved this conundrum by not being able to fold right shifting operations on indices. If we are able to pull information through into the folding pass, we will be allowed to actually do such things.

PapyChacal (approved):
:rocket:

xdsl/interpreter.py
AntonLydike: Oooooh, damn, didn't know that!

xdsl/transforms/constant_fold_interp.py
Moxinilian: isn't index lowering also parametrized by the target's index size?

AntonLydike: As far as I can see, the `canonicalize` pass has no index bitwidth argument, and that pass does folding if I'm not mistaken.

PR #2237, opened 2024-02-27
testing: rename test-lower-linalg-to-snitch to test-lower-snitch-stream-to-asm

superlopuh:
The main idea is that we want to keep a part of this stable, since we still have some tests defined as snitch_stream in the RISC-V backend paper experiments. I'd like to propose having two passes, one for linalg to snitch stream, and a second for lowering snitch stream down to assembly.


compor (approved):

compor:
Can you propose this on Zulip or better as an issue in the project? Having a shorthand pass for parts of the lowering makes sense

PR #2248, opened 2024-02-28
core: Spacing fix in EnumAttribute and OpaqueSyntaxAttribute

PapyChacal:
Enables:
- #2249

A mistake slipped in because all initial EnumAttributes where OpaqueSyntaxAttribute. linalg.IteratorType is the first one not to, hence the fix.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -596,7 +596,7 @@ class EnumAttribute(Data[EnumType]):
 
     @final
     def print_parameter(self, printer: Printer) -> None:
-        printer.print(" ", self.data.value)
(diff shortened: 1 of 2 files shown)


superlopuh (approved):

math-fehr (approved):

PapyChacal:
This shouldn't have been merged :smiling_face_with_tear: 
Realized meanwhile why this general spacing is off, with a concrete example in:
- #2251 
through e.g. `#hw<innerSym@sym>`... I guess I'll just make a specific baseclass for opaque syntax enum attributes, as I just want enum attributes to be struggle-free wrappers around Python enums.

PR #2249, opened 2024-02-28
dialects: make linalg.IteratorType an EnumAttribute

PapyChacal:
Stacked on:
- #2248 

Just a little reuse.
Had to remove the `@final` to add brackets.
@math-fehr, do you still believe those brackets belong in individual attributes? Is there an MLIR attribute that does not use them in non-opaque syntax?

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -15,7 +15,6 @@ from typing import (
     Protocol,
     TypeVar,
     cast,
-    final,
     get_args,
     get_origin,
     overload,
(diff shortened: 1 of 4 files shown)


xdsl/ir/core.py
```diff
@@ -594,11 +593,9 @@ def __init_subclass__(cls) -> None:
 
         cls.enum_type = enum_type
 
-    @final
```
superlopuh, line 597: why does the base not use in_angle_brackets?

PapyChacal: Because not all enum attributes use this syntax; in fact, most of them don't:
`omp<schedule static>`: the name part is `omp<schedule` ... `>`, the parameter part is just ` static`

PapyChacal: Also, after #2256, I'm thinking this dominating case of having brackets could be made a baseclass handled in our printer? :thinking:

superlopuh (approved):

PR #2251, opened 2024-02-28
dialects: (hw) Add `InnerSymAttr` attributes to HW dialect.

lucjaulmes:
Another hopefully manageable chunk of HW, which introduces the symbol attributes. These are used in many circt dialects (e.g., seq) to define the names then referenced by the `InnerRefAttr`s and the like.

This is [the bit of the symbol rationale](https://circt.llvm.org/docs/RationaleSymbols/#inner-symbol) that gives details.

This PR also includes a small fix on the tests, simplifying the test file and undoing the over-correction of camel case to snake case on symbol mnemonics (we want those to be compatible with CIRCT).

Thanks a lot to @PapyChacal for all the help in deciphering where CIRCT hides the symbol attr tests.

### DIFF
--- a/tests/dialects/test_hw.py
+++ b/tests/dialects/test_hw.py
@@ -310,3 +310,16 @@ def test_inner_sym_attr():
     assert aggregate_sym_attr.get_sym_name() == StringAttr(
         "sym"
     ), "InnerSymAttr for aggregate types should return name with field ID 0"
+
+    for inner, expected_field_id in zip(aggregate_sym_attr, [0, 1, 2]):
+        assert (
+            inner.field_id.data == expected_field_id
+        ), "InnerSymAttr should allow iterating its properties in order"
+
+    aggregate_without_nested = aggregate_sym_attr.erase(2)
+    assert (
+        aggregate_without_nested.get_sym_if_exists(2) is None
+    ), "InnerSymAttr removal should work"
+    assert (
+        len(aggregate_without_nested) == 2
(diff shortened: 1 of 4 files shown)


lucjaulmes:
- Regarding code formatting, the `...` on the same line is auto-corrected that way by ruff locally. Can revert but not sure why this is happening? I just made sure I have the correct versions (running python 3.11 though), and am rebased on latest main.
- Regarding coverage, will complete soon. Unfortunately local computing of coverage is off too.

PapyChacal:
> * Regarding code formatting, the `...` on the same line is auto-corrected that way by ruff locally. Can revert but not sure why this is happening? I just made sure I have the correct versions (running python 3.11 though), and am rebased on latest main.
> 
>     * Regarding coverage, will complete soon. Unfortunately local computing of coverage is off too.

Did you run `make` in the repository? AFAIU this installs dependencies as expected, i.e. the right versions of toolings.
That said, I recently had a lot of those ... reformatting show up recently, but never had a problem with them. did they trigger CI errors in this case?

lucjaulmes:
Yes CI failed on e9a0829. I updated from requirements.txt in case there were any chances, but could try a full delete and reinstall, see if that changes anything…

lucjaulmes:
Updated on latest main & added missing bits.

PapyChacal (approved):

superlopuh (approved):

PR #2252, opened 2024-02-28
dialects: (snitch_stream) add offsets to StridePattern and use in interpreter

superlopuh:
We'll need this for further testing also in an upcoming PR


tests/interpreters/test_snitch_stream_interpreter.py
```diff
@@ -67,7 +47,7 @@ def test_snitch_stream_interpreter():
     )
 
     with ImplicitBuilder(streaming_region_body) as (a_stream, b_stream, c_stream):
-        count_reg = riscv.LiOp(6).rd
+        count_reg = riscv.LiOp(5).rd
```
superlopuh, line 50: pesky off by one :)

superlopuh: The affine map interpreter implementation didn't catch this, because it just iterated forever, wrapping around to (0, 0, 0)

compor (approved):
Simplifies things, thanks

PR #2254, opened 2024-02-28
core: make ModulePass frozen

superlopuh:
Mainly through using n-tuples instead of lists.

This makes it much easier to reason about pass pipelines in general, but also for the interactive app specifically. If we can assume that all passes are immutable, we don't have to reconstruct them every time just to be safe.

### DIFF
--- a/xdsl/transforms/arith_add_fastmath.py
+++ b/xdsl/transforms/arith_add_fastmath.py
@@ -16,9 +16,9 @@ from xdsl.pattern_rewriter import (
 _FASTMATH_NAMES_TO_ENUM = {str(member.value): member for member in llvm.FastMathFlag}
 
 
-def _get_flag_list(flags: list[str]):
+def _get_flag_list(flags: tuple[str, ...]):
     try:
-        return [_FASTMATH_NAMES_TO_ENUM[flag] for flag in flags]
+        return tuple(_FASTMATH_NAMES_TO_ENUM[flag] for flag in flags)
     except KeyError as e:
         raise ValueError(f"{e} is not a valid fastmath flag.")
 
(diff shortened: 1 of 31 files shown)


compor (commented):
What do you mean by reason about the pass pipeline?

Did we dynamically change options and/or the behaviour of the passes so far?
Did we introspect it in any way?

superlopuh:
We never change options or behaviour, so this is basically an NFC. My proposal is to encode it as part of the type, so that we don't do it in the future. Today, they are difficult to introspect for this reason, and not safe to reuse, leading to a bit of weirdness on the interactive side where we initialise a new pass every time. When trying to exhaustively search the space of possible IRs given a set of rewrite passes, initing a new one every time adds up quickly.

compor:
The introspection aspect is not clear at all to me and how this would help this (maybe slightly), but I do get the interactive use case part. Thanks for elaborating.

compor (approved):

PapyChacal (approved):

dshaaban01 (approved):

xdsl/transforms/arith_add_fastmath.py
```diff
@@ -32,18 +32,20 @@ class AddArithFastMathFlags(RewritePattern):
...
-        | arith.Minimumf
-        | arith.Maximumf,
+        op: (
+            arith.Addf
+            | arith.Subf
+            | arith.Mulf
+            | arith.Divf
+            | arith.Minimumf
+            | arith.Maximumf
+        ),
```
math-fehr, line 42: I think that is an unrelated change? At least, that's not usually the style we are using

math-fehr (approved):
That makes sense to me!
I'm just not a fan of the new syntax for larger Union types!

AntonLydike (approved):
LGTM

xdsl/transforms/arith_add_fastmath.py
superlopuh: There's some weirdness in Black formatting, I didn't make this change manually. I've seen it in some other PRs, I think I have a situation where the formatter in VSCode and in precommit are slightly different, and both slightly ambiguous

superlopuh: #2271

PR #2255, opened 2024-02-29
dialects: adding Pure() to arith dialect

dshaaban01:
adding Pure to arith dialect so that dce works on arith peephole rewrites


superlopuh:
Does `make test` pass locally?

xdsl/dialects/arith.py
```diff
@@ -352,11 +351,15 @@ def infer_overflow_type(input_type: Attribute) -> Attribute:
 class Muli(SignlessIntegerBinaryOp):
```
superlopuh, line 351: aren't all the `SignlessIntegerBinaryOp`s pure?

superlopuh (commented):
Couldn't the abstract superclasses be annotated, instead of the subclasses?

xdsl/transforms/constant_fold_interp.py
```diff
@@ -27,7 +27,7 @@ class ConstantFoldInterpPattern(RewritePattern):
     interpreter: Interpreter
 
     def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
-        if not op.has_trait(Pure):
```
webmiche, line 30: I'd prefer either having Constantlike inherit from Pure or doing this in a different PR.

xdsl/dialects/arith.py
```diff
@@ -767,21 +769,25 @@ def print(self, printer: Printer):
 @irdl_op_definition
 class Addf(FloatingPointLikeBinaryOp):
     name = "arith.addf"
+    traits = frozenset((Pure(),))
```
webmiche, line 772: Can we not just add the trait on the parent op? So on FloatingPointLikeBinaryOp in this case.

webmiche (commented):
Do you think you can easily add tests for this? This does change some interaction with the framework, right?

dshaaban01:
> Does `make test` pass locally?

Yes it does. Not sure why the tests fail.

superlopuh:
Can you reproduce locally?

superlopuh:
dce only acts on Pure ops, so it looks like it's changing the output of that test

superlopuh:
We'll probably want to add the newly-pure ops to that test to make sure we are consistent in eliminating duplicates with the same inputs

xdsl/transforms/constant_fold_interp.py
superlopuh: I agree, could you please split this out into a separate PR?

superlopuh (commented):
Could you please add tests for these?

superlopuh:
Let's close this for now?

superlopuh:
Closing for now, let's reopen when we get back to this, there have been some changes both to arith and the impact of Pure with the effects interfaces, it'll be worth revisiting the PR anyway

PR #2259, opened 2024-02-29
interactive: don't print =none for None value

superlopuh:
It can't roundtrip, `None` is inferred when only the name of the argument is given, with no value.


dshaaban01 (approved):

dshaaban01:
Looks great, thank you!!

PR #2261, opened 2024-02-29
dialects: reformat onnx.Constant to align with MNIST models one

kayode-gif:
wip #2260 - onnx invalid tests need reformatting 

This PR adds a custom printer and parser to cater to the variety of formats `onnx.Constant` can be shown in. I have added `pass` to attributes which aren't used in the `ONNX-MLIR` since i didn't want to assume how they appeared.


tests/filecheck/mlir-conversion/with-mlir/dialects/onnx/ops.mlir
superlopuh, line 1: I don't think we want this test, the point of reference is onnx-mlir, not mlir

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,82 @@ def verify_(self) -> None:
...
+            printer.print_attribute(self.value_int)
+        elif self.value_ints:
+            printer.print_string("value_ints")
+            printer.print_string(" = ")
+            arr: list[int] = [value.value.data for value in self.value_ints]
+            printer.print(arr)
+        elif self.value_string:
+            pass
+        elif self.value_strings:
+            pass
```
superlopuh, line 702: I would recommend adding a helper here, one that will give the only value for this instance, like `self.get_value()`

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,82 @@ def verify_(self) -> None:
...
+            elif parser.parse_optional_keyword("value_strings"):
+                parser.parse_punctuation("=")
+                value_strings = parser.parse_attribute()
+            parser.parse_punctuation("}")
+            parser.parse_punctuation(":")
+            output_type = parser.parse_type()
+        else:
+            value = parser.parse_attribute()
+            if isinstance(value, DenseIntOrFPElementsAttr):
+                output_type = value.type
```
superlopuh, line 739: I'd expect this to work:
``` python
v = parser.parse_attribute()
if isinstance(v, DictionaryAttr):
  parser.parse_punctuation(':')
  t  = parser.parse_attribute()
else:
  # get the type somehow


constant = cls.build(attributes = {k: v}, result_types=(t,))
return constant
```

superlopuh: But it all depends on how onnx-mlir does it

kayode-gif: it doesnt like it

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,85 @@ def verify_(self) -> None:
...
+    def get_value(self) -> tuple[Any | None, str]:
+        if self.value_float:
+            return self.value_float, "value_float"
+        elif self.value_int:
+            return self.value_int, "value_int"
+        elif self.value_ints:
+            arr: list[int] = [value.value.data for value in self.value_ints]
+            return arr, "value_ints"
+        else:
+            # TODO: value_string, value_strings, value_floats
```
superlopuh, line 687: `raise NotImplementedError()`

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,85 @@ def verify_(self) -> None:
                 f"Only one value attribute must be provided, but {used_attrs} were specified"
             )
 
+    def get_value(self) -> tuple[Any | None, str]:
+        if self.value_float:
+            return self.value_float, "value_float"
+        elif self.value_int:
+            return self.value_int, "value_int"
+        elif self.value_ints:
+            arr: list[int] = [value.value.data for value in self.value_ints]
```
superlopuh, line 684: Can you add a comment here about why you do it this way?

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,85 @@ def verify_(self) -> None:
...
+        printer.print_string(" : ")
+        printer.print_attribute(self.output.type)
+
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        value_float = (
+            value_floats
+        ) = (
+            value_int
+        ) = value_ints = value_string = value_strings = value = output_type = None
```
superlopuh, line 716: Can you split these onto their own lines, please?

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,94 @@ def verify_(self) -> None:
...
+            return self.value_int, "value_int"
+        elif self.value_ints:
+            # In the ONNX-MLIR format, integer types are abstracted, so only the integer data is extracted
+            # for printing purposes.
+            arr: list[int] = [value.value.data for value in self.value_ints]
+            return arr, "value_ints"
+        else:
+            # TODO: value_string, value_strings, value_floats
+            raise NotImplementedError()
+
```
superlopuh, line 691: Sorry to change my advice but, thinking about this more, I think we might as well do it a different way. Can you please change this so that it returns an Attribute and a string, not an optional Any? I don't think that it makes sense to add logic that's specific to printing outside of the printing method

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,94 @@ def verify_(self) -> None:
...
+        if value and value_type:
+            if value_type == "value_ints":
+                printer.print_string(value_type)
+                printer.print_string(" = ")
+                printer.print(value)
+            else:
+                printer.print_string(value_type)
+                printer.print_string(" = ")
+                printer.print_attribute(value)
+        printer.print_string("}")
```
superlopuh, line 712: I would recommend changing this to something like this:
```
if self.value is not None:
  printer.print(self.value)
printer.print_attribute_dict(self.attributes, exclude=("value",))
```

xdsl/dialects/onnx.py
```diff
@@ -673,6 +675,94 @@ def verify_(self) -> None:
...
+            elif parser.parse_optional_keyword("value_strings"):
+                parser.parse_punctuation("=")
+                value_strings = parser.parse_attribute()
+            parser.parse_punctuation("}")
+            parser.parse_punctuation(":")
+            output_type = parser.parse_type()
+        else:
+            value = parser.parse_attribute()
+            if isinstance(value, DenseIntOrFPElementsAttr):
+                output_type = value.type
```
superlopuh, line 751: similarly here, we can just parse an attribute. If it's a `dense`, then we continue to parse an optional dict. If it's not `dense`, then it must be an extra attributes dict. I'd expect something like this to work.

xdsl/dialects/onnx.py
kayode-gif: if i change it to this, this is the issue:
`"value_ints" = [1 : i64, 2 : i64, 3 : i64]} `
but onnx-mlir:
`{value_ints = [1, 2, 3]}`
so we would have to alter the printer

superlopuh: Yeah, that's my suggestion

superlopuh: so maybe add a special case for `else if self.value_ints is not None`

kayode-gif: ive just realised that `print_attr_dict` does not have an exclude keyword argument, i think i have to do it manually

superlopuh: `Printer.print_op_attributes(reserver_attr_names=...)`

xdsl/dialects/onnx.py
compor: I agree, it would also avoid the duplication a bit

xdsl/dialects/onnx.py
kayode-gif: > `Printer.print_op_attributes(reserver_attr_names=...)`

just to be clear, you mean you want to change the `Printer` Class not the method that is bounded in the constant class, this is because the attribute name type is printed with strings aswell, so this doesn't work as nicely

kayode-gif: because if thats the case, i don't see a use for my helper as of such

superlopuh (approved):

PR #2271, opened 2024-03-01
misc: update precommit black version

superlopuh:
Turns out we're pretty far behind, so everyone creating new venvs and using black to auto-format in VSCode is introducing slightly incompatible formatting.


math-fehr (approved):
I don't like the ugly type syntax, sad they went in that direction but I understand it

superlopuh:
We seem to have bad luck with types these days :)

superlopuh:
Probably with a trick similar to the pyright one

superlopuh:
#2292

PR #2277, opened 2024-03-01
transformations: add imperfect loop nest support in memref_stream.streaming_region

### LINKED ISSUE #2268: transformations: add imperfect loop nest support in memref_stream.streaming_region
This means handling affine maps with fewer dimensions than the number of bounds

superlopuh:
This is the first step in lowering to imperfect loop nests.

In a perfect loop nest, it makes sense for all the affine maps to have the same number of dimensions: each dimension corresponds to an induction variable in the loop nest. For an imperfect loop nest, it makes sense for the affine map to have fewer dimensions, and for those to correspond to the indices of the loops that it's in, and not the ones that are further in. I propose to model this with a two step transform:
1. add a transform on `memref_stream.generic` to lop off all the reduction indices in the out parameters (upcoming)
2. add lowerings for streaming region (this PR) and `memref_stream.generic` to add the out parameter reads and writes not in the innermost loop (upcoming)

Note that the `streamify` pass is unchanged, it just duplicates the affine maps between the generic and the streaming region, but it must run after the `imperfectize` (naming TBD) pass.

fixes #2268


compor:
I'm not sure I understand how the upcoming 1+2 transforms are going to keep track of what is "lopped off" and the out reads/writes?

compor (approved):

superlopuh:
I'm not 100% sure either, but I have an idea, we basically need to replicate the structure of our targets, where the result is passed around in the iteration arguments

PR #2282, opened 2024-03-03
interactive: don't include the child in the root to child list

superlopuh:
The issue is that the child might need arguments, and we sometimes want to apply the passes up to but not including the pass with arguments, without asking the user what these arguments should be.


dshaaban01 (approved):

tests/interactive/test_app.py
```diff
@@ -467,3 +471,46 @@ async def test_passes():
...
+        for node in root_children:
+            assert node.data is not None
+            pass_val, _ = node.data
+            if pass_val.name == stencil_global_to_local.DistributeStencilPass.name:
+                distribute_stencil_node = node
+
+        assert distribute_stencil_node is not None
+
+        # When running in a test, node selection does not send the expected event
+        # app.passes_tree.select_node(distribute_stencil_node)
```
dshaaban01, line 510: remove commented code

superlopuh: It's kind of apart of the comment, I clarified it

PR #2285, opened 2024-03-03
interpreter: add xtype, a typed memory format string

superlopuh:
RawPtr no longer provides `Any`-typed methods, everything goes through the typed version.


superlopuh:

webmiche (approved):
LGTM, not that I get the exact effects of these changes, but I think the code looks good :wink:

PR #2286, opened 2024-03-03
interpreter: pass interpreter instance to impl_attr

superlopuh:
I missed that sometimes you need to know the bitwidth of the system to accurately construct the object required, this PR makes it possible, and this is leveraged by the next PR.


webmiche (commented):
Could you add tests?

superlopuh:
Sure, for which part?

webmiche:
Nvm, I looked at the codecov and most things seem to be touched by tests already :+1:

superlopuh:

webmiche (commented):
Maybe this is just not visible in the code diff, but I don't quite get where this new argument is needed. AFAICT, you are just changing function signatures around the code, but the additional argument is not really used. Could you point out where it is used? Or add a test that uses it?

webmiche (approved):
Ah, I see the classic I implement it here and use it in a different PR :wink: 
LGTM :+1:

superlopuh:
My aim was to simplify the already complex code changes in the other PR, I'm not sure what other decompisition would have been better than the three PRs I came up with

webmiche:
Yeah sure, it's fine. I think I was lacking context a bit, if I had known the code better, I could have probably seen what you were about to do. Also, I think the [x/n] annotation would have made it easier, then it's clear that these are stacked and the order in which i review is defined :)

PR #2287, opened 2024-03-03
interpreter: replace list data representation in shaped array with bytebuffer

superlopuh:
Use a bytebuffer in the interpreter, instead of a list of individual elements.


superlopuh:

webmiche (approved):
Ah, I understand this now. LGTM

xdsl/interpreters/ptr.py
```diff
@@ -123,7 +124,7 @@ def __getitem__(self, index: int) -> _T:
         raw_at_offset = self.raw + index * self.size
         return struct.unpack(self.format, raw_at_offset.memoryview[: self.size])[0]
 
-    def __setitem__(self, index: int, value: _T):
+    def __setitem__(self: TypedPtr[_T], index: int, value: _T):
```
superlopuh, line 127:
```suggestion
    def __setitem__(self, index: int, value: _T):
```

PR #2289, opened 2024-03-03
ci: use old pytest

superlopuh:
Our pattern is too coarse, the tests all passed locally, but broke with the new minor version.


webmiche:
@superlopuh  do you mind open an issue for this? I think we should fix this (and I'd like to know what "patterns are too coarse" means :wink: )

webmiche:
nvm, let's use the dependabot PR :+1:

PR #2290, opened 2024-03-03
interpreter: (memref_stream) add interpreter implementation for memref_stream

superlopuh:
Also add to the integration test filecheck, to test that a kernel gives the same results at the linalg level, memref_stream level, and snitch_stream level.


AntonLydike (approved):
:chart_with_upwards_trend:

PR #2293, opened 2024-03-04
transformations: (mlir-opt) Add option to specify mlir-opt binary in args

AntonLydike:
This adds the `executable` option to the `mlir-opt` pass that allows one to pass a custom executable (e.g. `mlir-opt-17`, or a path to a binary that is not in `PATH`).

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/mlir_opt_fail.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/mlir_opt_fail.mlir
@@ -1,5 +1,6 @@
 // RUN: xdsl-opt %s -p mlir-opt{arguments='--hello','--mlir-print-op-generic'} --print-op-generic --verify-diagnostics | filecheck %s
 // RUN: xdsl-opt %s -p mlir-opt[this-probably-will-never-be-an-MLIR-pass-name] --print-op-generic --verify-diagnostics | filecheck %s
+// RUN: xdsl-opt %s -p mlir-opt{executable=cat} --print-op-generic --verify-diagnostics | filecheck %s
 
 "builtin.module"() ({
   "func.func"() ({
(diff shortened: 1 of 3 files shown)


tests/filecheck/mlir-conversion/with-mlir/mlir_opt_fail.mlir
```diff
@@ -1,5 +1,6 @@
 // RUN: xdsl-opt %s -p mlir-opt{arguments='--hello','--mlir-print-op-generic'} --print-op-generic --verify-diagnostics | filecheck %s
 // RUN: xdsl-opt %s -p mlir-opt[this-probably-will-never-be-an-MLIR-pass-name] --print-op-generic --verify-diagnostics | filecheck %s
+// RUN: xdsl-opt %s -p mlir-opt{executable=cat} --print-op-generic --verify-diagnostics | filecheck %s
```
superlopuh, line 3: 🤖 🐱

AntonLydike: Sadly `cat` made the test not fail :/

JosseVanDelm (approved):
@AntonLydike didn't you also add a test to check whether it actually works?

tests/filecheck/mlir-conversion/with-mlir/mlir_opt_fail.mlir
superlopuh: Try `dog`

superlopuh (approved):

webmiche (commented):
Hmm, this means we can pretty much run arbitrary command line strings, right? Shouldn't we do at least some input sanity checks? Like, we probably really don't want the "executable" to have whitespaces....

PapyChacal:
I'm not sure xdsl-opt itself should be concerned by injections?

tests/filecheck/mlir-conversion/with-mlir/mlir_opt_fail.mlir
PapyChacal: We *need* `cowsays` filechecks somewhere

PapyChacal (approved):
Yes yes yes
Can you hook it in the pipeline sugar? In the next PR I guess

AntonLydike:
> Hmm, this means we can pretty much run arbitrary command line strings, right? Shouldn't we do at least some input sanity checks? Like, we probably really don't want the "executable" to have whitespaces....

Since you can also provide paths, whitespaces should be okay. It's not interpreted as a shell string, so the executable will really just be either a name or a path to an executable.

webmiche:
Yeah true. And if someone can run xdsl-opt from the commandline, they can already run arbitrary executables from the command line, so it's not like much changed (as long as noone commits a `rm -R ~/` in a filecheck test :wink:)

PR #2294, opened 2024-03-04
dialects: (snitch_stream) make snitch_stream parameters go outwards-in

superlopuh:
Makes things a little less weird for most of the high-level representations, and only treat the ubs and strides as innnermost-out inside the lowering from snitch_stream to snitch.


xdsl/transforms/convert_memref_stream_to_snitch_stream.py
```diff
@@ -106,11 +106,9 @@ def match_and_rewrite(
             for map, shape in zip(op.indexing_maps, shapes, strict=True)
         )
         bounds = tuple(bound.data for bound in op.bounds)
-        # snitch_stream stride patterns are from innermost out
-        # TODO: make consistent
```
superlopuh, line 110: ✅

AntonLydike (approved):
Yes, I think this is a good direction to move in!

PR #2301, opened 2024-03-05
transformations: implement convert-stencil-to-tensor.

PapyChacal:
### Higher-level lowering

This implements a new lowering for the stencil dialect, aiming at `linalg`+`tensor`+`bufferization`.
This new lowering's working name is `convert-stencil-to-tensor`, and is illustrated in [tests/filecheck/transforms/convert-stencil-to-tensor.mlir](https://github.com/xdslproject/xdsl/blob/emilien/stencil-tensor/tests/filecheck/transforms/convert-stencil-to-tensor.mlir)

This avoids doing any bufferization shenanigans in the stencil lowering, modulo some indications using the bufferization dialect to still get what we want out of it, while still decoupling the `load`/`apply`/`store` triplet in the lowering. Thus this can lower a strict superset of the previous lowering: some constraints defined on the `stencil` dialect were only present to keep the internal bufferization work tractable. All those will be lifted once this lowering reaches the same features as the initial one.
Ultimately, I want this to be the only lowering - or at least the only supported one.

### Designed to yield the same code through the right passes

This flow is used in DeVito/PsyClone forks, so to more reasonably move forward, I carefully designed it to yield equivalent IR through the adequate subsequent MLIR passes.

I also implemented an xDSL pass currently named `convert-stenctil-to-tensor-compat` as a simple pipeline containing this new lowering and the mentioned adequate passes. It is illustrated in [tests/filecheck/mlir-conversion/with-mlir/transforms/convert-stencil-to-tensor-compat.mlir](https://github.com/xdslproject/xdsl/blob/emilien/stencil-tensor/tests/filecheck/mlir-conversion/with-mlir/transforms/convert-stencil-to-tensor-compat.mlir).

This composite pass should be usable as a drop-in replacement for the current one.

### Still incomplete

Some cases and ops are not yet lowered by this PR's new lowering, mostly to keep the review bearable. I could also split this one if it's already too much!

Things to do in later PRs:

- Unrolled stencil lowering is not implemented yet. (some design to do)
- Access offset mapping is not yet supported. (trivial)
- `stencil.index` is not lowered yet. (trivial)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ toml<0.11
 pytest-cov
 coverage<8.0.0
 ipykernel
-pytest<8.1
+pytest<8.2
 nbval<0.12
 filecheck<0.0.25
 lit<18.0.0
--- a/tests/dialects/test_bufferization.py
+++ b/tests/dialects/test_bufferization.py
@@ -1,0 +1,31 @@
+from xdsl.dialects.bufferization import AllocTensorOp, ToTensorOp
+from xdsl.dialects.builtin import MemRefType, TensorType, UnitAttr, f64
+from xdsl.dialects.test import TestOp
+
+
+def test_to_tensor():
+    memref_t = MemRefType(f64, [10, 20, 30])
+    tensor_t = TensorType(f64, [10, 20, 30])
+    memref_v = TestOp(result_types=[memref_t]).res[0]
+
+    to_tensor = ToTensorOp(memref_v)
+    assert to_tensor.memref == memref_v
+    assert to_tensor.restrict is None
+    assert to_tensor.writable is None
+    assert to_tensor.tensor.type == tensor_t
+
+    to_tensor = ToTensorOp(memref_v, writable=True, restrict=True)
+    assert to_tensor.memref == memref_v
+    assert to_tensor.restrict == UnitAttr()
+    assert to_tensor.writable == UnitAttr()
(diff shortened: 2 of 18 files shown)


PapyChacal:
> wow! THat's a bit of a PR. Will review asap

Yep!
Please don't review in details; this is intended to be chunked in smaller PRs, and is just here for me to keep track or for having general discussions over the approach 🙂

superlopuh:
Hi all, it's been a really long time since updates on this @georgebisbas @PapyChacal

PapyChacal:
Yes, it triggered some discussions, that I will try to tackle in the coming days. I don't mind closing or converting to draft if you're after a clean PR page?

PapyChacal:
Well, I realized meanwhile why using `tensor` would require more work than I expected, so I'm closing this as this is not pragmatical at all for the current stencil objectives.

For the record, so I can remember it if I have time to spend on this, the main issue is that the stencil dialect allow partial updates, that is, it allow for stencil codes optionally not updating a cell.
mainline MLIR tensor operations do not allow this, as this is close to undefined behaviour: you then basically define value-semantics values with undefined cells.
linalg and scf.forall allow something similar in results, by simply conditionally returning the output initial value. Though this results in conditionally loading and always storing that value.
One could get to the same result by then optimizing away the noop case, by sinking the store in the branches containing the associated loads.

PR #2302, opened 2024-03-05
core: Make SSAValue hashable

AntonLydike:
This PR makes `SSAValue` hashable so that pyright will stop complaining when I use it in the `value_mapper` dict of `Operation.clone`.

The hash of an SSA Value is it's `id`, which should be safe as two SSA Values are only the same if they are literally the same value.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -372,6 +372,12 @@ class SSAValue(ABC):
             )
         self.replace_by(ErasedSSAValue(self.type, self))
 
+    def __hash__(self):
+        """
+        Make SSAValue hashable. Two SSA Values are never the same, therefore
+        the use of `id` is allowed here.
+        """
+        return id(self)
 
 @dataclass
 class OpResult(SSAValue):
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -379,6 +379,7 @@ class SSAValue(ABC):
         """
         return id(self)
 
+
 @dataclass
 class OpResult(SSAValue):
     """A reference to an SSA variable defined by an operation result."""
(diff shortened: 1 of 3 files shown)


AntonLydike:
By default, `__hash__` will return the result of `id(self)`. This is literally only so that pyright stops complaining.

PapyChacal (approved):

xdsl/ir/core.py
```diff
@@ -372,6 +372,13 @@ def erase(self, safe_erase: bool = True) -> None:
             )
         self.replace_by(ErasedSSAValue(self.type, self))
 
+    def __hash__(self):
+        """
+        Make SSAValue hashable. Two SSA Values are never the same, therefore
+        the use of `id` is allowed here.
+        """
+        return id(self)
```
webmiche, line 380: It is a bit weird to have this in an ABC, right? Should we have an assertion here that this is not executed?

webmiche: Hm, yeah, it's kind of redundant there now....

superlopuh:
```
class SSAValue(Hashable, ABC):
```
This seems to work for me locally, minimising the changes, and making fewer constraints on the way to subclass SSAValue. I have some wip branches making the IR elements abstract, leaving the implementation abstract would simplify some future changes I'd like to propose.

math-fehr (approved):
I like this, and I prefer it than the `Hashable` inheritance on `SSAValue` for now. Mostly because there is currently no reasons to change `__hash__` and `__id__`, and having this at the `SSAValue` level makes sense because we deduplicate code, and we make it easier to discover what are the "semantics" of `SSAValue`.

superlopuh:
Ok I guess it doesn't make a difference in practice

PR #2303, opened 2024-03-05
interactive: fix bug - do not allow users to be able to select root node to be added to the pass pipeline

dshaaban01:
Small bug in the code. Selecting root node of tree of passes adds node to pass pipeline - this is not allowed.


webmiche (approved):
Great change, great job, really appreciate you and your work, hope you will reach the sky!

superlopuh (approved):

PR #2305, opened 2024-03-06
dialects: Implement `tensor.insert_slice` and `tensor.extract_slice`

PapyChacal:


xdsl/dialects/bufferization.py
```diff
@@ -0,0 +1,98 @@
...
+        copy: SSAValue | Operation | None = None,
+        size_hint: SSAValue | Operation | None = None,
+    ):
+        super().__init__(
+            operands=(dynamic_sizes, copy, size_hint),
+            result_types=(result_type,),
+        )
+
+    @classmethod
+    def static_type(cls, type: Attribute) -> "AllocTensorOp":
```
superlopuh, line 52: What's the use of this?

PapyChacal: Just being explicit in user code. Felt natural to me while writing it to use it directly, cf #2301.
Do you want it removed?

superlopuh: There are a few things that threw me off here. One is not using Self (from typing_extensions), and using a string instead. One is the name, although I now understand what it means. Maybe `static_shape` could be more explicit? IDK. But also the fact that the implementation is shorter than the method call, I'm not sure that the integral of confusion over time is lowered with this helper...

PapyChacal: Will remove

xdsl/dialects/tensor.py
```diff
@@ -162,10 +173,102 @@ def verify_(self) -> None:
...
+
+        if reduce_rank:
+            result_sizes = list(s for s in sizes if s != 1)
+        else:
+            result_sizes = list(sizes)
+
+        return_type = TensorType[Any](source_t.get_element_type(), result_sizes)
+
+        return ExtractSliceOp.build(
+            operands=[source, [], [], []],
```
superlopuh, line 215: This feels a little weird, why not handle these also? Or at least raise NotImplemented if dynamic offsets are passed in

PapyChacal: This is `from_static_parameters`, it does not take SSAValues arguments for the dynamic offsets,sizes, and strides.

PapyChacal: The more generic constructor should be easier to adapt than `memref.subview`'s, from which I just ported this one, but I'd be glad to worry about it when someone needs it; I don't as I'll use those to lower compile-time sized code.

superlopuh: Oh sorry I misread, I thought it was the init, all good.

PR #2307, opened 2024-03-06
dialects: linalg minor fixes

PapyChacal:
- `doc` and `library_call` are properties
- The parenthesis around output types are mandatory in custom syntax if multiple types are present.

### DIFF
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -35,7 +35,7 @@ from xdsl.irdl import (
     irdl_attr_definition,
     irdl_op_definition,
     operand_def,
-    opt_attr_def,
+    opt_prop_def,
     prop_def,
     region_def,
     var_operand_def,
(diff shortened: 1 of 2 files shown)


xdsl/dialects/linalg.py
```diff
@@ -96,8 +96,8 @@ class Generic(IRDLOperation):
     # Trait attributes
     indexing_maps: ArrayAttr[AffineMapAttr] = prop_def(ArrayAttr[AffineMapAttr])
     iterator_types: ArrayAttr[IteratorTypeAttr] = prop_def(ArrayAttr[IteratorTypeAttr])
-    doc: StringAttr | None = opt_attr_def(StringAttr)
-    library_call: StringAttr | None = opt_attr_def(StringAttr)
+    doc: StringAttr | None = opt_prop_def(StringAttr)
```
superlopuh, line 99: do you think it would be too much of a detour to implement default properties before merging this?

PapyChacal: I really want to - and soon! Though this is actually a sensitive fix: MLIR interop is currently broken without it. Can I have your approval to merge this frustrating `opt_prop_def` so it's functional first? We'll get rid of those eventually!

superlopuh: I'm now wondering whether opt_attr_def will work for you in the short term? If not then let's merge like this.

superlopuh (approved):

PR #2308, opened 2024-03-06
core: add insert_op_at_location to PatternRewriter

superlopuh:
I'm messing about with some helpers that pass around what are effectively insertion points, this functionality makes that code simpler. It also feels like a general entropy minimising step.

### DIFF
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -94,8 +94,7 @@ class PatternRewriter(PatternRewriterListener):
         op = (op,) if isinstance(op, Operation) else op
         if not op:
             return
-        for op_ in op:
-            Rewriter.insert_op_at_location(op_, insertion_point)
+        Rewriter.insert_ops_at_location(op, insertion_point)
 
         for op_ in op:
             self.handle_operation_insertion(op_)
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -8,7 +8,7 @@ from functools import wraps
 from types import UnionType
 from typing import TypeVar, Union, final, get_args, get_origin
 
-from xdsl.builder import Builder, BuilderListener, InsertPoint
+from xdsl.builder import BuilderListener, InsertPoint
 from xdsl.dialects.builtin import ArrayAttr, ModuleOp
 from xdsl.ir import (
     Attribute,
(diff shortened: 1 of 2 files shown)


PapyChacal (approved):

compor (approved):

math-fehr (approved):
Very cool!

xdsl/pattern_rewriter.py
```diff
@@ -92,60 +92,47 @@ class PatternRewriter(PatternRewriterListener):
...
+    def insert_op_at_location(
+        self, op: Operation | Sequence[Operation], insertion_point: InsertPoint
+    ):
+        """Insert operations at a certain location in a block."""
+        self.has_done_action = True
+        op = [op] if isinstance(op, Operation) else op
+        if not op:
+            return
+        for op_ in op:
+            Builder(insertion_point).insert(op_)
```
AntonLydike, line 104: Can we hoist the builder construction out of the loop? I'm not deeply familiar with it's API, but this seems a bit wasteful imo.

superlopuh: Since putting this up I've realised this is not the right approach, we should work with the rewriter directly

superlopuh:
Updated to add equivalent method to Rewriter, seems a bit cleaner now

xdsl/pattern_rewriter.py
```diff
@@ -92,60 +92,47 @@ class PatternRewriter(PatternRewriterListener):
...
+
+        for op_ in op:
+            self.handle_operation_insertion(op_)
```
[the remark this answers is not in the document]
superlopuh: No longer relevant :)

xdsl/pattern_rewriter.py
```diff
@@ -86,60 +86,47 @@ class PatternRewriter(PatternRewriterListener):
...
+    def insert_op_at_location(
+        self, op: Operation | Sequence[Operation], insertion_point: InsertPoint
+    ):
+        """Insert operations at a certain location in a block."""
+        self.has_done_action = True
+        op = (op,) if isinstance(op, Operation) else op
+        if not op:
+            return
+        for op_ in op:
+            Rewriter.insert_op_at_location(op_, insertion_point)
```
math-fehr, line 98: Actually, `Rewriter.insert_op_at_location` should allow a sequence of operations? Or is that not the case?

xdsl/rewriter.py
```diff
@@ -252,19 +252,23 @@ def insert_block_before(block: Block | list[Block], target: Block):
         pos = region.get_block_index(target)
         region.insert_block(block_list, pos)
 
+    @staticmethod
+    def insert_op_at_location(op: Operation, insertion_point: InsertPoint):
```
math-fehr, line 256: Yeah, I would add `Sequence[Operation]` as well to make things simpler

PR #2310, opened 2024-03-06
core: move InsertPoint to Rewriter [NFC]

superlopuh:
I'd like to add some helpers to Rewriter to leverage `InsertPoint`, and right now this would involve a circular reference. Between `Rewriter` and `Builder`, it feels like `InsertPoint` is closer in abstraction to `Rewriter` anyway.


compor (approved):

math-fehr (approved):
I see where it's going and I like it

PR #2311, opened 2024-03-06
core: use InsertPoint for block inlining

superlopuh:
Unifies implementations

### DIFF
--- a/tests/test_rewriter.py
+++ b/tests/test_rewriter.py
@@ -147,7 +147,6 @@ def test_inline_block_at_end():
 "builtin.module"() ({
   %0 = "test.op"() : () -> !test.type<"int">
   "test.op"() ({
-  ^0:
   }) : () -> ()
   %1 = "test.op"() : () -> !test.type<"int">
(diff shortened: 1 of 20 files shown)


tests/test_rewriter.py
```diff
@@ -147,7 +147,6 @@ def test_inline_block_at_end():
 "builtin.module"() ({
   %0 = "test.op"() : () -> !test.type<"int">
   "test.op"() ({
-  ^0:
```
superlopuh, line 150: inlined source blocks are now consistently deleted when inlined into another block

PapyChacal (approved):

compor (approved):

tests/test_rewriter.py
math-fehr: Makes sense to me, that's actually what MLIR does as well!

math-fehr (approved):
Cool

AntonLydike (approved):
Very nice stuff!

PR #2313, opened 2024-03-06
core: declarative format variable fix

PapyChacal:


math-fehr (approved):
Arg, I see the problem :/
One other way is to index the operand definition instead, but I agree that getting it from the name seems to be the way to go!
Thanks!

PR #2316, opened 2024-03-07
install: Consistently use VENV_DIR variable in Makefile

lucjaulmes:
Currently the makefile interchangeably uses `venv` and `${VENV_DIR}` to refer to the virtual environment install directory. This change replaces all instances with the variable.

This allows to have e.g. 2 versions side-by-side:
```sh
$ make  # unchanged, builds into ./venv/
$ git checkout some-other-branch
$ make VENV_DIR=other-venv  # builds into ./other-venv/ without overwriting ./venv/
```

The trailing slash on the target name means you can use a hidden directory (e.g., `make VENV_DIR=.venv`) without the target being confused for a special Makefile target (`.c.o`, `.PHONY`, etc).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,11 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 .ONESHELL:
 
 # these targets don't produce files:
-.PHONY: venv clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
+.PHONY: ${VENV_DIR}/ clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
 .PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report-md
 
 # set up the venv with all dependencies for development
-venv: requirements.txt
+${VENV_DIR}/: requirements.txt
 	python3 -m venv ${VENV_DIR}
 	. ${VENV_DIR}/bin/activate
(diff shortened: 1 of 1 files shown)


math-fehr:
I guess `VENV_DIR` is `venv` by default?

superlopuh:
I wonder if the extra slash is what's making the CI unhappy? Maybe something breaks in the pipeline?

superlopuh:
Ah no it's unrelated, I'll update the branch

AntonLydike (approved):
Thanks for the fix!

lucjaulmes:
> I guess `VENV_DIR` is `venv` by default?

It is indeed:

https://github.com/xdslproject/xdsl/blob/f011cf8d14285946b1544b061d4d94e8411e9ba4/Makefile#L4-L5

On second thought this could also be kept as-is, in which case the PHONY default target venv is a different concept to the default install directory venv, i.e. implicitly the equivalent to:

```mk
venv: ${VENV_DIR}/

# set up the venv with all dependencies for development
${VENV_DIR}/: requirements.txt
	python3 -m venv ${VENV_DIR}
	. ${VENV_DIR}/bin/activate
	python3 -m pip --require-virtualenv install -r requirements.txt
```

math-fehr (approved):

AntonLydike:
Shall we merge this for now?

lucjaulmes:
I think it could be good to have @compor’s take on this, plus he last changed things relating to the venv target.

superlopuh:
I would say let's merge it, and fix things later if anything breaks

compor (approved):

PR #2317, opened 2024-03-07
dialects (hw): Fix Formatting

webmiche:
Quick PR to fix the currently failing CI

### DIFF
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -329,14 +329,12 @@ class InnerSymAttr(
         ...
 
     @overload
-    def __init__(self, syms: str | StringAttr) -> None:
-        ...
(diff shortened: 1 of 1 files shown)


math-fehr (approved):

lucjaulmes:
It’s unclear to me what the correct formatting is for empty functions.

In #2251 I had to insert these line breaks (reverted here) because [same-line ellipses failed the CI](https://github.com/xdslproject/xdsl/actions/runs/8086218106/job/22095477254). Are those now expected to be consistently on the same line?

superlopuh:
We recently updated the version in the precommit, as it was out of date with the one in the CI. Can you run make precommit-install again? Sorry about the confusion, I should probably have made more of an announcement. There's also a vague idea of making it less broken by finding the value to use dynamically from requirements.txt but I can't make any promises about timelines :)

PR #2320, opened 2024-03-07
transformations: (onnx) lower onnx.Gemm to linalg

### LINKED ISSUE #2272: transformations: (onnx) lower onnx.Gemm to linalg

kayode-gif:
resolves #2272

### DIFF
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -34,5 +34,13 @@
 // CHECK-NEXT: %res_constant_2 = ml_program.global_load_const @onnx_constant_2 : tensor<1x5xf32>
 // CHECK-NEXT: ml_program.global private @onnx_constant_1(dense<1> : tensor<1xi64>) : tensor<1xi64>
 // CHECK-NEXT: ml_program.global private @onnx_constant_2(dense<2.000000e+00> : tensor<1x5xf32>) : tensor<1x5xf32>
+
+
+%t5, %t6, %t7 = "test.op"(): () -> (tensor<5x3xf32>, tensor<3x2xf32>, tensor<5x2xf32>)
+%res_gemm = "onnx.Gemm"(%t5, %t6, %t7) {onnx_node_name = "/Gemm"}: (tensor<5x3xf32>, tensor<3x2xf32>, tensor<5x2xf32>) -> tensor<5x2xf32>
+
+
 // CHECK-NEXT:  }
+
+
 
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -30,12 +30,41 @@
 %t5, %t6, %t7 = "test.op"(): () -> (tensor<1x320xf32>, tensor<50x320xf32>, tensor<50xf32>)
 %res_gemm= "onnx.Gemm"(%t5, %t6, %t7) {onnx_node_name = "/Gemm", "alpha" = 1.000000e+00 : f32, "beta" = 1.000000e+00 : f32, "transA" = 0 : si64, "transB" = 1 : si64}: (tensor<1x320xf32>, tensor<50x320xf32>, tensor<50xf32>) -> tensor<1x50xf32>
 
-// CHECK-NEXT:   %t5, %t6, %t7 = "test.op"() : () -> (tensor<1x320xf32>, tensor<50x320xf32>, tensor<50xf32>)
+// CHECK-NEXT:  %t5, %t6, %t7 = "test.op"() : () -> (tensor<1x320xf32>, tensor<50x320xf32>, tensor<50xf32>)
 // CHECK-NEXT:  %3 = tensor.empty() : tensor<320x50xf32>
 // CHECK-NEXT:  %4 = linalg.transpose ins(%t6:tensor<50x320xf32>) outs(%3:tensor<320x50xf32>) permutation = [1, 0]
 // CHECK-NEXT:  %res_gemm = tensor.empty() : tensor<1x50xf32>
 // CHECK-NEXT:  %res_gemm_1 = linalg.mul ins(%t5, %4 : tensor<1x320xf32>, tensor<320x50xf32>) outs(%res_gemm : tensor<1x50xf32>) -> tensor<1x50xf32>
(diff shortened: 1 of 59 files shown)


xdsl/dialects/onnx.py
```diff
@@ -302,11 +302,11 @@ def verify_(self) -> None:
         if tensor_b_type.get_num_dims() != 2:
             raise VerifyException("tensor B should be a 2D tensor")
 
-        if self.trans_a is not None:
-            list(tensor_a_shape).reverse()
+        if self.trans_a is not None and self.trans_a.value.data == 1:
+            tensor_a_shape = tuple(tensor_a_shape)[::-1]
```
superlopuh, line 306:
```suggestion
            tensor_a_shape = tuple(reversed(tensor_a_shape))
```
similarly below

superlopuh:
Looking at the codecov, there are quite a few lines that aren't covered. I think GEMM would be worth really making reliable, and covering in tests completely. Could you please add some tests that would exercise the rest of the possible parameters?

kayode-gif:
> Looking at the codecov, there are quite a few lines that aren't covered. I think GEMM would be worth really making reliable, and covering in tests completely. Could you please add some tests that would exercise the rest of the possible parameters?

that is all done now i believe :)

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -30,13 +30,42 @@
...
+// CHECK-NEXT:  %6 = linalg.transpose ins(%t11:tensor<10x5xf32>) outs(%5:tensor<5x10xf32>) permutation = [1, 0]
+// CHECK-NEXT:  %7 = tensor.empty() : tensor<5x10xf32>
+// CHECK-NEXT:  %8 = arith.constant 5.000000e-01 : f32
+// CHECK-NEXT:  %9 = linalg.mul ins(%8, %6 : f32, tensor<5x10xf32>) outs(%7 : tensor<5x10xf32>) -> tensor<5x10xf32>
+// CHECK-NEXT:  %10 = tensor.empty() : tensor<5x3xf32>
+// CHECK-NEXT:  %11 = arith.constant 5.000000e-01 : f32
+// CHECK-NEXT:  %12 = linalg.mul ins(%11, %t13 : f32, tensor<5x3xf32>) outs(%10 : tensor<5x3xf32>) -> tensor<5x3xf32>
+// CHECK-NEXT:  %res_gemm_2 = tensor.empty() : tensor<5x3xf32>
+// CHECK-NEXT:  %res_gemm_2_1 = linalg.mul ins(%9, %t12 : tensor<5x10xf32>, tensor<10x3xf32>) outs(%res_gemm_2 : tensor<5x3xf32>) -> tensor<5x3xf32>
+// CHECK-NEXT:  %res_gemm_2_2 = linalg.add ins(%res_gemm_2_1, %12 : tensor<5x3xf32>, tensor<5x3xf32>) outs(%res_gemm_2 : tensor<5x3xf32>) -> tensor<5x3xf32>
```
superlopuh: This looks weird to me. My understanding is that the result should be `A @ B * alpha + beta` (modulo transpose). Is that the case? If so, the values for each subexpression (`A @ B`, `A @ B * alpha`, and the result) all have the same shape, and can be stored in the same result buffer, so with only one tensor.empty. Something like this:
```
res0 = tensor.empty()
res1 = linalg.mul(A,B, out=res0)
alpha = arith.constant alpha
res2 = linalg.mul(res1, alpha, out=res1)
beta = arith.constant beta
res3 = linalg.add(res2, beta, out=res2)
```
I'm not 100% sure that this is legal with the tensor semantics, but this is the shape that we'd want with memrefs.

superlopuh: @PapyChacal, Mr Bufferization, do you have advice for this? Is it legal to pass a tensor as both an in and out parameter? is the tensor.empty the way to go here?

superlopuh: I checked and it seems fine to do this

kayode-gif: @superlopuh  does this mean we are good to merge this ?

superlopuh (approved):

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
superlopuh: 👍 thanks for the reminder

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -27,12 +27,53 @@
...
+// CHECK-NEXT:  %t11, %t12, %t13 = "test.op"() : () ->  (tensor<10x5xf32>, tensor<10x3xf32>, tensor<5x3xf32>)
+// CHECK-NEXT:  %5 = tensor.empty() : tensor<5x10xf32>
+// CHECK-NEXT:  %6 = linalg.transpose ins(%t11:tensor<10x5xf32>) outs(%5:tensor<5x10xf32>) permutation = [1, 0]
+// CHECK-NEXT:  %7 = arith.constant 5.000000e-01 : f32
+// CHECK-NEXT:  %8 = linalg.mul ins(%7, %6 : f32, tensor<5x10xf32>) outs(%6 : tensor<5x10xf32>) -> tensor<5x10xf32>
+// CHECK-NEXT:  %9 = arith.constant 5.000000e-01 : f32
+// CHECK-NEXT:  %10 = linalg.mul ins(%9, %t13 : f32, tensor<5x3xf32>) outs(%t13 : tensor<5x3xf32>) -> tensor<5x3xf32>
+// CHECK-NEXT:  %res_gemm_2 = tensor.empty() : tensor<5x3xf32>
+// CHECK-NEXT:  %res_gemm_2_1 = linalg.matmul ins(%8, %t12 : tensor<5x10xf32>, tensor<10x3xf32>) outs(%res_gemm_2 : tensor<5x3xf32>) -> tensor<5x3xf32>
+// CHECK-NEXT:  %res_gemm_2_2 = linalg.add ins(%res_gemm_2_1, %10 : tensor<5x3xf32>, tensor<5x3xf32>) outs(%res_gemm_2 : tensor<5x3xf32>) -> tensor<5x3xf32>
```
superlopuh, line 64: oh sorry one sec here this looks a little off to me, the fact that the out parameter isn't also `%res_gemm_2_1`

kayode-gif: @superlopuh should look better now :)

superlopuh (approved):

Issue #1760, opened 2023-11-07
dialects: (riscv) use register index for identity of RISC-V registers


superlopuh:
I think it would be really interesting to see what happens when the register is represented with a more complex structure that more accurately identifies it than just the string that we use currently. There are a few restrictions:

1. it has to be immutable
2. all register representations with the same bits in the binary encoding should be equal and have the same hash
3. it has to support registers outside of the current binary encodings (we use j0, j1, ... for extra registers)

Maybe something like this could work?

``` python
@dataclass(frozen=True)
class RISCVRegister:
  index: int
  name_hint: str = field(default="", hash=False)

class RISCVRegisterType(Data[RISCVRegister], TypeAttribute, ABC):
  ...

@irdl_attr_definition
class IntRegisterType(RISCVRegisterType):
    name = "riscv.reg"

x0 = IntRegisterType(0)
zero = IntRegisterType(0, "zero")
assert x0 == zero
fp = IntRegisterType(8, "fp")
s0 = IntRegisterType(8, "s0")
j0 = IntRegisterType(-1, "j0")
f0 = FloatRegisterType(0)
ft0 = FloatRegisterType(0, "ft0")
```

superlopuh:
This has been implemented across a few PRs.

PR #2321, opened 2024-03-07
dialects: (riscv) add index to registers

zero9178:
The identity of registers is currently defined by a string. This can be problematic as RISC-V registers have alternative spellings which are then considered as unequal.

This PR is a first step towards fixing that issue by adding an optional `index` parameter to the type. The idea is to have a unique index that identifies the register in the future if it is a predefined register. Non-predefined registers such as the infinite `j[0-9]+` registers set `index` to `None` to signal that the spelling should be used for equality.

A future PR will add the `__eq__` implementation after the representation has proven itself.

Part of https://github.com/xdslproject/xdsl/issues/1760

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -3,16 +3,18 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence, Set
 from io import StringIO
-from typing import IO, Annotated, ClassVar, Generic, Literal, TypeAlias, TypeVar
+from typing import IO, Annotated, Generic, Literal, TypeAlias, TypeVar
 
 from typing_extensions import Self
 
 from xdsl.dialects.builtin import (
     AnyIntegerAttr,
     IndexType,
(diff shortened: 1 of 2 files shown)


xdsl/dialects/riscv.py
```diff
@@ -74,47 +78,63 @@ def __init__(self, flags: None | Sequence[FastMathFlag] | Literal["none", "fast"
...
+        if index is not None:
+            index_attr = IntAttr(index)
+        return index_attr, StringAttr(spelling)
 
     @classmethod
     def unallocated(cls) -> Self:
-        if cls._unallocated is None:
-            cls._unallocated = cls("")
-        return cls._unallocated
+        return cls("")
```
superlopuh, line 107: While we're out it, could you please make this an abstract method, and in the subclasses return either `Registers.UNALLOCATED_INT` or `Registers.UNALLOCATED_FLOAT`?

superlopuh (approved):

compor:
beautiful thanks!

PR #2326, opened 2024-03-10
dialects: (comb) Fix concat op

Moxinilian:
The constructor, parser and printer of the comb concat op are incorrect. This PR fixes it, adds a useful builder for the concat operation and adds a verifier.

### DIFF
--- a/tests/dialects/test_comb.py
+++ b/tests/dialects/test_comb.py
@@ -1,9 +1,11 @@
 import pytest
 
 from xdsl.dialects.builtin import IntegerType, i32
-from xdsl.dialects.comb import ConcatOp, ICmpOp
-from xdsl.dialects.test import TestOp, TestType
-from xdsl.utils.exceptions import VerifyException
+from xdsl.dialects.comb import Comb, ConcatOp, ICmpOp
+from xdsl.dialects.test import Test, TestOp, TestType
+from xdsl.ir import MLContext
+from xdsl.parser import Parser
+from xdsl.utils.exceptions import ParseError, VerifyException
 from xdsl.utils.test_value import TestSSAValue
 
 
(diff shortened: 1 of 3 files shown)


tests/dialects/test_comb.py
```diff
@@ -14,3 +17,38 @@ def test_icmp_incorrect_comparison():
         # 'slet' is an invalid comparison operation
         _icmp_op = ICmpOp(a, b, "slet")
     assert e.value.args[0] == "Unknown comparison mnemonic: slet"
+
+
+def test_comb_concat_builder():
+    a = TestOp(result_types=[IntegerType(5)])
+    b = TestOp(result_types=[IntegerType(3)])
+    c = TestOp(result_types=[IntegerType(1)])
+    foo = TestOp(result_types=[TestType("foo")])
```
math-fehr, line 26: Can you replace these with `TestSSAValue(IntegerType(_))` like in the previous example?
It's just to stay consistent with the rest of the codebase.

math-fehr (approved):
Nice!

tests/dialects/test_comb.py
Moxinilian: I cannot believe I missed this. The infrastructure in this project is just so fancy.

xdsl/dialects/comb.py
```diff
@@ -402,6 +402,19 @@ def print(self, printer: Printer):
...
+    """
+    Gets the sum of the width of the provided integer types. Returns None
+    if one of the provided attributes is not an integer type.
+    """
+    sum_of_width = 0
+    for typ in int_types:
+        if not isinstance(typ, IntegerType):
+            return None
+        sum_of_width += typ.width.data
+    return sum_of_width
```
superlopuh, line 415: For future reference:
``` python
return sum(t.width.data for t in int_types if isinstance(t, IntegerType))
```

PR #2327, opened 2024-03-10
dialects: (comb) Fix extract op

Moxinilian:
The constructor, parser and printer of the comb extract op are incorrect. This PR fixes it and adds a verifier.

### DIFF
--- a/xdsl/dialects/comb.py
+++ b/xdsl/dialects/comb.py
@@ -364,7 +364,12 @@ class ParityOp(IRDLOperation):
 class ExtractOp(IRDLOperation):
     """
     Extract a range of bits into a smaller value, low_bit
-    specifies the lowest bit included.
+    specifies the lowest bit included. Result is the size
+    of the value to extract.
+
+    |-----------------|   input
+           l              low bit
+           <-------->     result
     """
 
     name = "comb.extract"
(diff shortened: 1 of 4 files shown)


webmiche:
comb is a CIRCT dialect, right? If so, which CIRCT commit are we talking about here?

Or is comb in MLIR? In this case, I would prefer to add some interoperation tests to this PR.

Moxinilian:
@webmiche Comb is CIRCT. We are talking about commit `2e23cda6c2cbedb118b92fab755f1e36d80b13f5`.

webmiche:
Hmm, I think this commit hash is important. Is there a way to include it? Maybe add it at the top of the file?

If what was there before is broken, we can only be happy that the almighty Theo decided to bless us mere humans with this correct code. Tysm!

webmiche:
That's fine, it's better than before, so i am happy :+1:

webmiche (approved):

xdsl/dialects/comb.py
```diff
@@ -376,30 +383,55 @@ class ExtractOp(IRDLOperation):
     result: OpResult = result_def(IntegerType)
 
     def __init__(
-        self, operand: Operation | SSAValue, low_bit: IntegerAttr[IntegerType]
+        self,
+        operand: Operation | SSAValue,
+        low_bit: IntegerAttr[IntegerType],
+        result: IntegerType,
```
math-fehr, line 389:
```suggestion
        result_type: IntegerType,
```

xdsl/dialects/comb.py
```diff
@@ -376,30 +383,55 @@ class ExtractOp(IRDLOperation):
...
     ):
         operand = SSAValue.get(operand)
         return super().__init__(
             attributes={"low_bit": low_bit},
             operands=[operand],
-            result_types=[operand.type],
+            result_types=[result],
```
math-fehr, line 395:
```suggestion
            result_types=[result_type],
```

math-fehr (approved):
Nice!

PR #2333, opened 2024-03-12
dialects: (x86) PR1 - Registers

KGrykiel:
This is the first of several pull requests that are intended to upstream the work from the KGrykiel/x86dialect branch. In this one I set up registers that are going to be used by the x86 dialect


AntonLydike (approved):
LGTM! I assume there are more register types coming later on?

webmiche (commented):
Could you add some tests for this? At least instantiate some of the registers in a filecheck test or something.

If you don't want to, feel free to move this to the experimental folder.

webmiche:
(Sidenote: our [naming standard](https://github.com/xdslproject/xdsl/wiki/Tags-for-commit-messages-and-PR-titles) says that the dialect name should be in brackets :wink: )

xdsl/dialects/x86.py
```diff
@@ -0,0 +1,141 @@
...
+        "r10": 10,
+        "r11": 11,
+        "r12": 12,
+        "r13": 13,
+        "r14": 14,
+        "r15": 15,
+    }
+
+
+class Registers(ABC):
```
superlopuh, line 114: I regret this design decision in the riscv dialect file, I find that that file is pretty bloated because we squeeze all the functionality in one place. Could you please split this file into several, with clean dependencies between them? I would like to aim for the following structure:
```
dialects/
  x86/
    __init__.py
    register.py  # X86RegisterType, GeneralRegisterType
    registers.py # RAX, RCX, ...
```
I would probably recommend a similar structure to ir, which bans ir.core, and add the little ruff annotation that bans importing from `x86.register` directly, making users import from `x86` only

xdsl/dialects/x86.py
```diff
@@ -0,0 +1,141 @@
...
+    An x86 register type.
+    """
+
+    _unallocated: ClassVar[Self | None] = None
+
+    @classmethod
+    def unallocated(cls) -> Self:
+        if cls._unallocated is None:
+            cls._unallocated = cls("")
+        return cls._unallocated
```
superlopuh, line 32: We moved away from this pattern in riscv, `UNALLOCATED` should probably be in the `Registers` as one of the options, and the `unallocated` class method abstract

xdsl/dialects/x86.py
```diff
@@ -0,0 +1,141 @@
...
+)
+from xdsl.irdl import (
+    irdl_attr_definition,
+)
+from xdsl.parser import AttrParser
+from xdsl.printer import Printer
+from xdsl.utils.exceptions import VerifyException
+
+
+class X86RegisterType(Data[str], TypeAttribute, ABC):
```
superlopuh, line 21: I'm wondering whether it would make sense to move the shared functionality to a `RegisterType` abstract base class in the `backend` directory, used by all the assembly dialects. I have a feeling that this would make a lot of things simpler, but I don't know how it impacts the other code in your branch. Could you please have a go at this?

KGrykiel:
> Could you add some tests for this? At least instantiate some of the registers in a filecheck test or something.
> 
> If you don't want to, feel free to move this to the experimental folder.

I have some tests that use the register class, they'll be added in later PRs 👍

superlopuh:
Can you please add the tests in this PR so that we can review them at the same time as the code they're testing?

KGrykiel:
the tests make use of the registers, but are for specific instructions so I can't add them until i also add the instructions. Unless you'd like me to write a test just for initialising a register?

webmiche:
Yes, please, I think that would make sense :+1:

xdsl/dialects/x86.py
superlopuh: Just so there's no confusion, could you please do this in this PR?

superlopuh: Yes, maybe as part of this PR, or as a separate PR to be merged before this one. TBH I think it's a small enough change that it can be done in this PR.

xdsl/dialects/x86.py
superlopuh: It wouldn't because I'm only suggesting to move the abstract base class to the backend module, which has mostly abstract methods. The riscv and x86-specific subclasses should remain with their dialects

superlopuh: Ah yes

superlopuh: Maybe we'll need to refactor that then

KGrykiel: sure, sounds good

KGrykiel:
let me know if this is what you had in mind

xdsl/dialects/x86/register.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def instruction_set_name(cls) -> str:
+        return "x86"
+
+    @classmethod
+    def abi_index_by_name(cls) -> dict[str, int]:
+        return X86_INDEX_BY_NAME
+
+
+class Registers(ABC):
```
superlopuh, line 80: I'd recommend getting rid of this and placing the constants at the top level of this file

xdsl/dialects/x86/x86.py
superlopuh, line 1: I'd probably rename this dialect.py

xdsl/dialects/x86/x86.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import annotations
+
+from xdsl.ir import Dialect
+
+from .register import GeneralRegisterType, Registers
+
+# This is where all of the instructions will be placed.
+Registers = Registers  # here to allow the test file to import the Registers class
```
superlopuh, line 8: no need to do this at all

xdsl/dialects/x86/x86.py
superlopuh: Actually, maybe for now the best thing to do would be to move the dialect definition to __init__.py

xdsl/backend/register_type.py
```diff
@@ -0,0 +1,84 @@
...
+    TypeAttribute,
+)
+from xdsl.irdl import ParameterDef
+from xdsl.parser import AttrParser
+from xdsl.printer import Printer
+
+
+class RegisterType(ParametrizedAttribute, TypeAttribute, ABC):
+    """
+    A RISC-V register type.
```
superlopuh, line 25:
```suggestion
    An abstract register type for target ISA-specific dialects.
```

xdsl/dialects/x86/x86.py
KGrykiel: without it the test complains because "module 'xdsl.dialects.x86' has no attribute 'Registers'". Probably an import issue

xdsl/dialects/x86/register.py
KGrykiel: what are you referring to by "this"? Do you mean getting rid of the class structure and placing the registers as loose constants?

xdsl/dialects/x86/x86.py
KGrykiel: the dialect definition will be filled up with instructions in later PRs that I plan to put within x86.py (which I'll rename to dialect.py) so with the future in mind it's perhaps best to leave it as is?

superlopuh: what we can do is split into `register.py`, `ops.py`, and `__init__.py`, that way the ops are in their own file (or files if we want to group them by functionality), and they can still all be imported from `xdsl.dialects.x86`

xdsl/dialects/x86/x86.py
superlopuh: that's because Registers needs to be imported in the `__init__.py`

xdsl/dialects/x86/x86.py
KGrykiel: Like this?:
`from xdsl.ir import Dialect

from .ops import *
from .register import *

X86 = Dialect(
    "x86",
    [],
    [
        GeneralRegisterType,
    ],
)
`
pre-commit doesn't allow me to commit that with star imports with the dialect definition here

KGrykiel:
formatting fails because of a star import in __init__, but not sure how to avoid it.

xdsl/dialects/x86/x86.py
superlopuh: yep take a look at `xdsl/ir/__init__.py`

xdsl/dialects/x86/ops.py
superlopuh, line 1: let's just add this in a later PR?

xdsl/dialects/x86/x86.py
KGrykiel: I tried the #noqa already but it doesn’t seem to affect the linter

xdsl/dialects/x86/__init__.py
```diff
@@ -0,0 +1,11 @@
+from xdsl.ir import Dialect
+
+from .register import *
```
superlopuh, line 3:
```suggestion
from .register import * # noqa: F405
```

xdsl/dialects/x86/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+X86 = Dialect(
+    "x86",
+    [],
+    [
+        GeneralRegisterType,  # noqa: F405
```
superlopuh, line 9:
```suggestion
        GeneralRegisterType,  
```

xdsl/dialects/x86/__init__.py
KGrykiel: @superlopuh This is exactly what I tried in the previous commit, "removed ops.py", but this results in a failed formatting test.

KGrykiel: The error technically fires on line 9, and I can only assume it would fire on each individual addition to the dialect as well

superlopuh: at this stage if the star makes things tedious let's import each class?

KGrykiel: Then I’d need to also import the register constants or otherwise the tests won’t work, but in that case it’s going to complain that the imports are unused in the init file

superlopuh: can do both?

superlopuh: Or have clients import them from the `x86.registers` module, that's sort of ok, since the initial proposal was `x86.Registers` anyway

KGrykiel: I’m fine with whatever but I thought the whole point of doing it this way was to avoid importing directly from the submodules of x86?

superlopuh: I'm not 100% sure how important it is. I think it would be good to have all the operations and attributes available in `xdsl.dialects.x86`, but the register constants I think it's not critical

KGrykiel: alright, will do it that way then

superlopuh:
Looking great! Can you please take a look at the codecov, and cover the remaining lines? We don't always do this, but it feels like for such a foundational type it would be good to cover all the bases. For example, we don't have any tests for parsing and printing, but it would be good to add a filecheck test for these. Something like this:
``` mlir
%a = "test.op"() -> !x86.reg<rax>
```

tests/dialects/test_x86.py
```diff
@@ -0,0 +1,7 @@
+from xdsl.dialects import x86
+
+
+def test_register():
+    rax = x86.register.RAX
+    assert rax.is_allocated
+    assert rax.register_name == "rax"
```
webmiche, line 7: Could you parametrize this to test all the registers? You can see an eample [here](https://github.com/xdslproject/xdsl/blob/main/tests/dialects/test_arith.py#L124)

xdsl/dialects/x86/register.py
```diff
@@ -0,0 +1,95 @@
...
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import Attribute
+from xdsl.irdl import (
+    irdl_attr_definition,
+)
+from xdsl.parser import AttrParser
+from xdsl.utils.exceptions import VerifyException
+
+
+class X86RegisterType(RegisterType):
```
webmiche, line 14: Why do you have both `x86RegisterType` or `GeneralRegisterType` and both have the same description?

xdsl/dialects/x86/register.py
```diff
@@ -0,0 +1,95 @@
...
+RSI = GeneralRegisterType("rsi")
+RDI = GeneralRegisterType("rdi")
+R8 = GeneralRegisterType("r8")
+R9 = GeneralRegisterType("r9")
+R10 = GeneralRegisterType("r10")
+R11 = GeneralRegisterType("r11")
+R12 = GeneralRegisterType("r12")
+R13 = GeneralRegisterType("r13")
+R14 = GeneralRegisterType("r14")
+R15 = GeneralRegisterType("r15")
```
webmiche, line 95: Have you thought about the lower bit size registers? How will you model that?

webmiche (commented):
A couple more comments from my side.

xdsl/dialects/x86/register.py
KGrykiel: x86RegisterType is the parent class for multiple types of register collections. Other than GeneralRegisterType, AVXRegisterType will also inherit from it. A similar system has been implemented in the riscv dialect

xdsl/dialects/x86/register.py
KGrykiel: I haven't made an implementation yet, but I reckon that could be done via making another class that also inherits from x86RegisterType

superlopuh:
I appreciate that they'll be further covered in downstream PRs, but it would still be worth adding the tests for the code being added in the same PR, so we can review the two together. I don't know what the tests are that are still upcoming, so I cannot say now whether they will cover the functionality introduced here in the future.

KGrykiel:
Ok, will add a filecheck now as well then 🫡

xdsl/dialects/x86/register.py
tarinduj: It seems [LLVM does this](https://github.com/llvm/llvm-project/blob/0bcec96f3fe919fc5fd1189d1901fa3a0e501e2c/llvm/lib/Target/X86/X86RegisterInfo.td) by tracking sub-registers as an attribute in the register class. They also track which bits the sub-registers take up in the super-register, which is probably useful for allocation and optimizations.

xdsl/dialects/x86/register.py
webmiche: Could you somehow represent this in the documentation? One might be an abstract registertype, the other one a specifically scalar registertype

tests/filecheck/dialects/x86/x86_ops.mlir
```diff
@@ -0,0 +1,6 @@
+// RUN: XDSL_ROUNDTRIP
+
+"builtin.module"() ({
+    %a = "test.op"() : () -> !x86.reg<rax>
+    // CHECK: %{{.*}} = "test.op"() : () -> !x86.reg<rax>
+}) : () -> ()
```
superlopuh, line 6:
```suggestion
}) : () -> ()

```

superlopuh: Oh actually we can just remove the whole thing

superlopuh, line 6:
```suggestion
%a = "test.op"() : () -> !x86.reg<rax>
// CHECK: %{{.*}} = "test.op"() : () -> !x86.reg<rax>
```
The top-level builtin.module is implicit

superlopuh: Please add a newline at the end of all files, it helps with git diffing

xdsl/dialects/x86/register.py
superlopuh: Ah, that's a good idea
```suggestion
import abc

class X86RegisterType(RegisterType, abc.ABC):
```
Having it marked as explicitly abstract and not possible to instantiate would be useful in itself

tarinduj (approved):

superlopuh (approved):
Looks perfect, thank you!

webmiche (approved):
Nice, LGTM now. I agree that marking the `X86RegisterType` as an `ABC` would make a lot of sense, but if you don't think so, that's fine as well.

PR #2334, opened 2024-03-12
frontend: [onnx] unify type builders

alecerio:


tests/frontend/onnx/test_type.py
```diff
@@ -1,16 +1,54 @@
...
 
-from xdsl.dialects.builtin import TensorType, f32, f64
+from xdsl.dialects.builtin import Float32Type, TensorType, f32, f64
 
 try:
     from onnx import TensorShapeProto, TypeProto  # noqa: E402
 
-    from xdsl.frontend.onnx.shape_type import get_shape, get_tensor_type  # noqa: E402
+    from xdsl.frontend.onnx.type import (
+        get_elem_type,  # noqa: E402
```
superlopuh, line 10: I'm now wondering whether these are necessary. Could you please try removing the `noqa` comments, and seeing what happens?

xdsl/frontend/onnx/type.py
```diff
@@ -0,0 +1,36 @@
...
+
+ELEM_TYPE = {
+    1: f32,
+    11: f64,
+}
+"Dictionary containing information about the type."
+
+
+def get_elem_type(code: int) -> Attribute:
+    "Function that takes an index as input and returns the corresponding Attribute."
```
superlopuh, line 14: Can you please change the single quotes to triple quotes? Also, the `Function` seems a bit redundant, could you please remove it from the docstring, and below?
The term `index` isn't clear here, it seems like `ONNX tensor element type code` is a little more accurate/descriptive, do you know what term they use in their documentation to describe this value?

PR #2335, opened 2024-03-12
dialects: (comb) Fix icmp parser/printer discrepancy and 2-state semantics modelling

Moxinilian:
The printer and parser of the ICmp operation were emitting/expecting an extra comma that is not present in the original dialect.


math-fehr (approved):
,

Moxinilian:
oh well I forgot to ran tests and that yielded a new problem!

Moxinilian:
I applied the same fix to the parity op as well

superlopuh:
Can you please assign yourself, apply the right label, etc?

superlopuh (approved):

PR #2338, opened 2024-03-13
interpreter: (onnx) add elementwise onnx ops and onnx.relu

kayode-gif:
wip #2337

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ typeCheckingMode = "strict"
     "tests/dialects/test_memref.py",
     "xdsl/interpreters/experimental/wgpu.py",
     "xdsl/frontend/onnx/shape_type.py",
+    "xdsl/interpreters/onnx.py",
 ]
 "ignore" = [
     "tests/filecheck/frontend/dialects/builtin.py",
@@ -59,6 +60,7 @@ typeCheckingMode = "strict"
     "tests/filecheck/frontend/dialects/scf.py",
     "tests/filecheck/frontend/programs/invalid.py",
     "xdsl/_version.py",
+
 ]
 
 [tool.isort]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,6 @@ typeCheckingMode = "strict"
     "tests/filecheck/frontend/dialects/scf.py",
     "tests/filecheck/frontend/programs/invalid.py",
     "xdsl/_version.py",
-
 ]
 
 [tool.isort]
--- a/tests/interpreters/test_onnx_interpreter.py
+++ b/tests/interpreters/test_onnx_interpreter.py
@@ -1,7 +1,3 @@
-# try:
-#     import numpy as np
-# except (ImportError, ModuleNotFoundError):
-#     np = None
 import numpy as np
 import pytest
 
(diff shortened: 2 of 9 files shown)


xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,32 @@
...
+    impl,
+    register_impls,
+)
+
+
+@register_impls
+class OnnxFunctions(InterpreterFunctions):
+    @impl(onnx.Add)
+    def run_add(self, interpreter: Interpreter, op: onnx.Add, args: PythonValues):
+        return ReturnedValues(args), ()
```
superlopuh, line 16:
```suggestion
        lhs, rhs = args
        return lhs + rhs
```

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,80 @@
...
+def test_onnx_sub():
+    op = onnx.Sub(
+        TestSSAValue(TensorType(f32, [2, 3])),
+        TestSSAValue(TensorType(f32, [2, 3])),
+        res_type=TensorType(f32, [2, 3]),
+    )
+
+    a = ShapedArray([1, 2, 3, 4, 5, 6], [2, 3])
+    b = ShapedArray([1, 4, 2, 5, 3, 6], [2, 3])
+    c = ShapedArray([0, -2, 1, -1, 2, 0], [2, 3])
```
superlopuh, line 36:
```suggestion
    c = ShapedArray([0, 0, 0, 0, 0, 0], [2, 3])
```

superlopuh: We want to test that we get the expected result in the end

xdsl/interpreters/onnx.py
superlopuh: actually I guess what we really want to do is
1. check that the inputs are all shaped ptrs
2. create numpy arrays with their contents
3. call + on the numpy arrays
4. return result

tests/interpreters/test_onnx_interpreter.py
kayode-gif: but isn't that the expected result, if you subtract a and b? i don't get why [0,0,0,0..]

superlopuh: because you're giving the result as the input, in real life the input will either be initialized to random or 0, and has to be set as part of the calculation

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,107 @@
...
+    pytest.skip("onnx is an optional dependency", allow_module_level=True)
+    # in the case to port numpy, install it anyways
+    try:
+        subprocess.check_call([sys.executable, "-m", "pip", "install", "numpy"])
+        print("NumPy installation successful.")
+        __import__("numpy")
+    except subprocess.CalledProcessError:
+        print("NumPy installation unsuccessful. Please install it manually.")
+
+np = __import__("numpy")
```
superlopuh, line 33: please try doing this in a way that only checks for available packages

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,107 @@
...
+        TestSSAValue(TensorType(f32, [2, 3])),
+        res_type=TensorType(f32, [2, 3]),
+    )
+
+    a = ShapedArray(np.array([1, 2, 3, 4, 5, 6]), [2, 3])
+    b = ShapedArray(np.array([1, 4, 2, 5, 3, 6]), [2, 3])
+    c = ShapedArray(np.array([0, 0, 0, 0, 0, 0]), [2, 3])
+
+    interpreter.run_op(op, (a, b, c))
+    assert np.all(c.data == np.array([0, 0, 0, 0, 0, 0]))
```
superlopuh, line 66:
```suggestion
    assert c == ShapedArray(np.array([0, -2, 1, -1, 2, 0]), [2, 3])
```

superlopuh: We want the input to be 0, and to check that the result is what we expect

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,93 @@
...
+except (ImportError, ModuleNotFoundError):
+    np = None
+
+if np is None:
+    # if numpy is not available, uninstall onnx
+    try:
+        subprocess.check_call([sys.executable, "-m", "pip", "uninstall", "-y", "onnx"])
+        print("ONNX uninstallation successful.")
+    except subprocess.CalledProcessError:
+        print("ONNX uninstallation unsuccessful. Please uninstall it manually.")
```
superlopuh, line 25: we want to avoid all variations of messing with installed packages

xdsl/interpreters/__init__.py
```diff
@@ -27,6 +29,7 @@ def register_implementations(
     ctx: MLContext,
     include_wgpu: bool = True,
```
superlopuh, line 30: I'd rather add a flag here and make the client of this class choose to include onnx, like in the wgpu case

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,93 @@
...
+        def run_sub(
+            self, interpreter: Interpreter, op: onnx.Sub, args: tuple[Any, ...]
+        ) -> tuple[Any, ...]:
+            lhs, rhs = args[0], args[1]
+            assert isinstance(lhs, ShapedArray)
+            assert isinstance(rhs, ShapedArray)
+            lhs = cast(ShapedArray[int], lhs)
+            rhs = cast(ShapedArray[int], rhs)
+            assert lhs.shape == rhs.shape
+            result = [l - r for l, r in zip(np.array(lhs.data), np.array(rhs.data))]
```
superlopuh, line 55:
```suggestion
            result = np.array(lhs.data) - np.array(rhs.data)
```

kayode-gif: this messes up the tests for some reason

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,104 @@
...
+        assert c[0].shape == [2, 3]
+
+    def test_onnx_mul():
+        op = onnx.Mul(
+            TestSSAValue(TensorType(f32, [2, 2])),
+            TestSSAValue(TensorType(f32, [2, 2])),
+            res_type=TensorType(f32, [2, 2]),
+        )
+
+        a = ShapedArray(np.array([1, 4, 7, 1]), [2, 2])  # pyright: ignore
```
superlopuh, line 66: What's the type error? We want to avoid all ignores possible

superlopuh: To have the pyright CI be aware of numpy we need to modify the CI script, just cherry pick the pyright-ci change from this PR: #2334

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,83 @@
...
+    InterpreterFunctions,
+    impl,
+    register_impls,
+)
+from xdsl.interpreters.shaped_array import ShapedArray
+
+try:
+    import numpy as np  # pyright: ignore
+except (ImportError, ModuleNotFoundError) as exc:
+    print(exc)
```
superlopuh, line 15:
```suggestion
import numpy as np
```
Please add this file to the list of files in pyproject.toml that only shows pyright errors if it's open

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+from xdsl.dialects.builtin import ModuleOp, TensorType, f32
+from xdsl.interpreter import Interpreter
+from xdsl.interpreters.shaped_array import ShapedArray
+from xdsl.utils.test_value import TestSSAValue
+
+try:
+    import numpy as np
+except (ImportError, ModuleNotFoundError):
+    np = None
+    pytest.importorskip("numpy", reason="numpy is an optional dependency in xDSL")
```
superlopuh, line 13:
```suggestion
pytest.importorskip("numpy", reason="numpy is an optional dependency in xDSL")
```
does this work?

kayode-gif: nope

superlopuh: what's the error?

kayode-gif: `  import numpy as np
E   ModuleNotFoundError: No module named 'numpy'` 
 this is the error

tests/interpreters/test_wgpu.py
superlopuh, line 1: Ah, I forgot that this is how it works. I'd actually recommend keeping this structure, and using it in the onnx interpreter test also

pyproject.toml
```diff
@@ -59,6 +60,7 @@ typeCheckingMode = "strict"
     "tests/filecheck/frontend/dialects/scf.py",
     "tests/filecheck/frontend/programs/invalid.py",
     "xdsl/_version.py",
+
```
superlopuh, line 63:
```suggestion
```

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,98 @@
...
+from xdsl.interpreters.onnx import OnnxFunctions
+from xdsl.interpreters.shaped_array import ShapedArray
+from xdsl.utils.test_value import TestSSAValue
+
+try:
+    import numpy as np
+except (ImportError, ModuleNotFoundError):
+    np = pytest.importorskip("numpy", reason="numpy is an optional dependency in xDSL")
+
+interpreter = Interpreter(ModuleOp([]))
```
superlopuh, line 15:
```suggestion
pytest.importorskip("numpy", reason="numpy is an optional dependency in xDSL")

import numpy as np  # noqa: I001, E402
```

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,81 @@
...
+    InterpreterFunctions,
+    impl,
+    register_impls,
+)
+from xdsl.interpreters.shaped_array import ShapedArray
+
+try:
+    import numpy as np
+except (ImportError, ModuleNotFoundError) as exc:
+    print(exc)
```
superlopuh, line 15:
```suggestion
import numpy as np
```

superlopuh: This will make some tests fail. Please use the same approach as we did for webgpu to work around them. You can search for "wgpu" in the project to see which parts of the code require special casing

xdsl/interpreters/__init__.py
```diff
@@ -26,6 +26,7 @@ def register_implementations(
     interpreter: Interpreter,
     ctx: MLContext,
```
superlopuh, line 27:
```suggestion
    ctx: MLContext,
    *,
```
Now that we have multiple optional boolean parameters, I'd recommend adding a `*` parameter, to force the boolean flags to be named.

xdsl/interpreters/onnx.py
kayode-gif: okay cool, but there isn't any special workaround that i see that "wgpu" does to handle this, ive done it now slightly differently  tho

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,101 @@
...
+
+where_is_numpy = importlib.util.find_spec("numpy")
+found_numpy = where_is_numpy is not None
+
+if found_numpy:
+    import numpy as np  # noqa: I001, E402
+else:
+    np = None
+
+if np is not None:
```
superlopuh, line 22: This is not necessary

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,82 @@
...
+    impl,
+    register_impls,
+)
+from xdsl.interpreters.shaped_array import ShapedArray
+
+where_is_numpy = importlib.util.find_spec("numpy")
+numpy_found = where_is_numpy is not None
+
+if numpy_found:
+    import numpy as np  # noqa: E402
```
superlopuh, line 17: we now don't need this dance at all, the contract will be that everyone importing this file has to have checked the numpy dependency first. This is the case in this PR, where you test both in the unit test and rely on the flag in the `register_implementation` function

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @impl(onnx.Relu)
+    def run_relu(
+        self, interpreter: Interpreter, op: onnx.Relu, args: tuple[Any, ...]
+    ) -> tuple[Any, ...]:
+        operand = args[0]
+        assert isinstance(operand, ShapedArray)
+        operand = cast(ShapedArray[int], operand)
+        operand_data = np.array(operand.data)
+        result = np.maximum(np.zeros_like(operand.data), operand_data)
```
superlopuh, line 77:
```suggestion
        result = operand_data * (operand_data > 0)
```

xdsl/interpreters/onnx.py
```diff
@@ -0,0 +1,78 @@
+from typing import Any, cast
+
+import numpy as np  # pyright: ignore
```
superlopuh, line 3:
```suggestion
import numpy as np 
```

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,96 @@
...
+def test_onnx_add():
+    op = onnx.Add(
+        TestSSAValue(TensorType(f32, [2, 3])),
+        TestSSAValue(TensorType(f32, [2, 3])),
+        res_type=TensorType(f32, [2, 3]),
+    )
+
+    a = ShapedArray(np.array([1, 2, 3, 4, 5, 6]), [2, 3])
+    b = ShapedArray(np.array([1, 4, 2, 5, 3, 6]), [2, 3])
+    c = ShapedArray(np.array([0, 0, 0, 0, 0, 0]), [2, 3])
```
superlopuh, line 28: nit: 
In my understanding, these don't need to be numpy arrays, changing them to Python lists would make this test a little more readable.
```suggestion
    a = ShapedArray([1, 2, 3, 4, 5, 6], [2, 3])
    b = ShapedArray([1, 4, 2, 5, 3, 6], [2, 3])
    c = ShapedArray([0, 0, 0, 0, 0, 0], [2, 3])
```

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,96 @@
...
+
+    c = interpreter.run_op(op, (a, b, c))
```
superlopuh, line 30: Looks like a bug in the interpreter, we shouldn't allow passing more arguments to this than the number of operands in the op, I'll fix this later, in the meantime we should remove the c from the argument
```suggestion
    c = interpreter.run_op(op, (a, b))
```

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -0,0 +1,96 @@
...
+    assert len(c) == 1
+    assert np.array_equal(c[0].data[0][0], np.array([2, 6, 5, 9, 8, 12]))
```
superlopuh, line 32:
```suggestion
    assert c == ShapedArray([2, 6, 5, 9, 8, 12], [2,3])
```

xdsl/interpreters/onnx.py
superlopuh: https://stackoverflow.com/questions/32109319/how-to-implement-the-relu-function-in-numpy

kayode-gif:
@superlopuh is this dance better?

superlopuh:
beautiful!

superlopuh (approved):
Thank you for all the iteration, I think this is perfect

PR #2344, opened 2024-03-18
backend: (csl) Add a print to csl backend

AntonLydike:
This starts adding the `csl` backend by creating a printing to print to [CSL](https://sdk.cerebras.net/csl/code-examples/) from MLIRs standard dialects (func, builtin, arith, scf...)


webmiche (commented):
Could you add a comment somewhere (briefly) explaining what CSL is and maybe a link to some documentation?

AntonLydike:
> Could you add a comment somewhere (briefly) explaining what CSL is and maybe a link to some documentation?

Done, I added it to the `__init__.py` of the backend folder, hope that clears some things up.

xdsl/backend/csl/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+CSL is the language used to program for the Cerebras Wafer Scale Engine (WSE) a wafer-parallel compute accelerator.
+
+More info here: https://sdk.cerebras.net/
```
webmiche, line 4: :+1:

webmiche (approved):
I have no idea in what form cerebras wants its programs, so that's on you to review. From the xDSL standpoint, this LGTM.

xdsl/backend/csl/print_csl.py
superlopuh, line 1: why have this as a folder instead of `backend/csl.py`? Will there be more files added in the future?

xdsl/backend/csl/print_csl.py
```diff
@@ -0,0 +1,170 @@
...
+        This method does not yet support all the types and will be expanded as needed later.
+        """
+        match type_attr:
+            case Float16Type():
+                return "f16"
+            case Float32Type():
+                return "f32"
+            case (
+                IndexType()
+            ):  # TODO: figure out what the equivalent of intpr_t is for csl!
```
superlopuh, line 80: task plz

xdsl/backend/csl/print_csl.py
```diff
@@ -0,0 +1,170 @@
...
+                return "i64"
+            case IntegerType(
+                width=IntAttr(data=width),
+                signedness=SignednessAttr(data=Signedness.UNSIGNED),
+            ):
+                return f"u{width}"
+            case IntegerType(width=IntAttr(data=width)):
+                return f"i{width}"
+            case unkn:
+                return f"<!unknown type {unkn}>"
```
superlopuh, line 90: nit:
```suggestion
            case _:
                return f"<!unknown type {type_attr}>"
```
`unkn` threw me off a bit

superlopuh: similarly below

xdsl/backend/csl/print_csl.py
```diff
@@ -0,0 +1,170 @@
...
+from xdsl.ir import Attribute, Block, SSAValue
+
+
+@dataclass
+class CslPrintContext:
+    output: IO[str]
+
+    variables: dict[SSAValue, str] = field(default_factory=dict)
+
+    _counter: int = field(default=0)
```
superlopuh, line 28: my first instinct is that it would be neater to have a counter per name, to avoid
```
a0
b1
c2
```
```suggestion
    _counter: int = field(default_factory=lambda: defaultdict(counter))
```

superlopuh: Ah, I see the problem, the potential clashing is annoying. I'm not sure what the right way to deal with that is, everything that doesn't involve a global analysis of all possible names seems broken, so might as well go with your simple solution

xdsl/backend/csl/print_csl.py
AntonLydike: Yes, there will be transformations and possibly dialects, plus some more complex things to figure out data flow etc.

I just thought having this structure here from the start would be the best solution.

xdsl/backend/csl/print_csl.py
AntonLydike: Yeah, almost every aspect of this needs improving. This is a "good enough" thing to get the ball rolling

superlopuh (approved):

xdsl/backend/csl/print_csl.py
```diff
@@ -141,7 +141,7 @@ def print_block(self, body: Block):
                     self.descend().print_block(bdy.block)
                     self.print("}")
                 case scf.For(lb=lower, ub=upper, step=stp, body=bdy):
-                    idx, *iter_args = bdy.block.args
+                    idx, *_ = bdy.block.args
                     self.print(
                         f"for(@range({self.mlir_type_to_csl_type(idx.type)}, {self._get_variable_name_for(lower)}, {self._get_variable_name_for(upper)}, {self._get_variable_name_for(stp)})) |{self._get_variable_name_for(idx)}| {{"
```
AntonLydike, line 146: Hmmm, could you try and make this line fit within the character limit?

xdsl/backend/csl/print_csl.py
```diff
@@ -0,0 +1,188 @@
...
+                        and isa(lower.owner.value, IntegerAttr[IntegerType])
+                        and isa(stp.owner.value, IntegerAttr[IntegerType])
+                        and lower.owner.value.value.data == 0
+                        and stp.owner.value.value.data == 1
+                    ):
+                        # implicit lb==0 and step==1
+                        self.print(
+                            f"for(@range({self.mlir_type_to_csl_type(idx.type)}, {self._get_variable_name_for(upper)})) |{self._get_variable_name_for(idx)}| {{"
+                        )
+                    else:
```
AntonLydike, line 158: Do we need this, or can we drop this as well?

compor (approved):

PR #2348, opened 2024-03-20
interpreter: (ml_program) Add ml_program operations to target MNIST ONNX model 

### LINKED ISSUE #2346: interpreter: (ml_program) Add ml_program operations to target MNIST ONNX model

kayode-gif:
resolves #2346

### DIFF
--- a/tests/interpreters/test_ml_program_interpreter.py
+++ b/tests/interpreters/test_ml_program_interpreter.py
@@ -1,5 +1,23 @@
+from xdsl.dialects import ml_program
+from xdsl.dialects.builtin import ModuleOp, StringAttr, TensorType, i32
+from xdsl.interpreter import Interpreter
+from xdsl.interpreters.ml_program import MLProgramFunctions
+from xdsl.interpreters.shaped_array import ShapedArray
+
+
 def test_ml_program_global():
-    pass
+    interpreter = Interpreter(ModuleOp([]))
+    interpreter.register_implementations(MLProgramFunctions())
+    op = ml_program.Global(
+        StringAttr("global_op"),
+        TensorType(i32, [4]),
+        None,
+        None,
+        StringAttr("private"),
+    )
+    print(op.value)
+    c = interpreter.run_op(op, ())
+    assert c == ShapedArray([0, 1, 2, 3], [4])
 
 
 def test_ml_program_global_load_constant():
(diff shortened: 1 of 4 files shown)


xdsl/interpreters/ml_program.py
```diff
@@ -0,0 +1,41 @@
...
+    def run_global(
+        self, interpreter: Interpreter, op: ml_program.Global, args: tuple[Any, ...]
+    ) -> tuple[Any, ...]:
+        if op.is_mutable is not None:
+            raise NotImplementedError(
+                "mutable global not yet supported in ml_program.global interpreter"
+            )
+        global_value = op.value
+        assert isinstance(global_value, DenseIntOrFPElementsAttr)
+        shape = global_value.get_shape()
```
superlopuh, line 27: I think better raise not implemented if shape is none here, I'm not sure when this would arise but it feels like we want to catch this explicitly

xdsl/interpreters/ml_program.py
```diff
@@ -0,0 +1,41 @@
...
+    def run_global_load_constant(
+        self,
+        interpreter: Interpreter,
+        op: ml_program.GlobalLoadConstant,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        result_type = op.result.type
+        assert isinstance(result_type, TensorType)
+        result_shape = result_type.get_shape()
+        return (ShapedArray([None], list(result_shape)),)
```
superlopuh, line 41: the code above should be here. This function should look up the symbol, get the constant value, and return a result

tests/interpreters/test_ml_program_interpreter.py
```diff
@@ -0,0 +1,55 @@
...
+    interpreter.register_implementations(MLProgramFunctions())
+    op = ml_program.Global(
+        StringAttr("global_op"),
+        TensorType(i32, [4]),
+        None,
+        DenseIntOrFPElementsAttr.tensor_from_list([4], i32, [4]),
+        StringAttr("private"),
+    )
+
+    c = interpreter.run_op(op, ())
```
superlopuh, line 27: we're running into the same issue as before, I think it's time to add an assertion that `run_op` takes and returns as many interpreter values as the op has operands and results, probably in a separate PR

superlopuh: The Global op is a declaration, it does not have a result, so the interpreter shouldn't either

superlopuh (approved):

PR #2351, opened 2024-03-20
dialects: (x86) PR2 - Operations: add

KGrykiel:
This is the second pull request in the series of upstreaming the x86 dialect. In this PR I add the "add" operation and all the necessary infrastructure to support it and it's testing. A lot of the structure is designed to resemble the riscv dialect.

### DIFF
--- a/tests/dialects/test_x86.py
+++ b/tests/dialects/test_x86.py
@@ -1,7 +1,6 @@
 import pytest
 
 from xdsl.dialects import x86
-from xdsl.utils.test_value import TestSSAValue
 
 
 @pytest.mark.parametrize(
@@ -28,11 +27,3 @@ from xdsl.utils.test_value import TestSSAValue
 def test_register(register: x86.register.GeneralRegisterType, name: str):
     assert register.is_allocated
     assert register.register_name == name
-
-
-def test_add_op():
-    r1 = TestSSAValue(x86.register.RAX)
-    r2 = TestSSAValue(x86.register.RDX)
-    add_op = x86.AddOp(r1, r2, result=x86.register.RAX)
-
-    assert add_op.assembly_line() == "    add rax, rdx"
--- a/tests/dialects/test_x86.py
+++ b/tests/dialects/test_x86.py
@@ -1,6 +1,7 @@
 import pytest
 
 from xdsl.dialects import x86
+from xdsl.utils.test_value import TestSSAValue
 
 
 @pytest.mark.parametrize(
@@ -27,3 +28,11 @@ from xdsl.dialects import x86
 def test_register(register: x86.register.GeneralRegisterType, name: str):
     assert register.is_allocated
     assert register.register_name == name
+
+
+def test_add_op():
+    r1 = TestSSAValue(x86.register.RAX)
+    r2 = TestSSAValue(x86.register.RDX)
+    add_op = x86.AddOp(r1, r2, result=x86.register.RAX)
+
+    assert add_op.assembly_line() == "    add rax, rdx"
--- a/tests/filecheck/dialects/x86/x86_ops.mlir
+++ b/tests/filecheck/dialects/x86/x86_ops.mlir
@@ -2,8 +2,7 @@
 %a = "test.op"() : () -> !x86.reg<rax>
 // CHECK: %{{.*}} = "test.op"() : () -> !x86.reg<rax>
 
-%0 = x86.get_register : () -> !x86.reg<>
-%1 = x86.get_register : () -> !x86.reg<>
+%0, %1 = "test.op"() : () -> (!x86.reg<>, !x86.reg<>)
 
 %add = x86.add %0, %1 : (!x86.reg<>, !x86.reg<>) -> !x86.reg<>
 // CHECK: %{{.*}} = x86.add %{{.*}}, %{{.*}} : (!x86.reg<>, !x86.reg<>) -> !x86.reg<>
--- a/tests/filecheck/dialects/x86/x86_ops.mlir
+++ b/tests/filecheck/dialects/x86/x86_ops.mlir
@@ -1,4 +1,9 @@
 // RUN: XDSL_ROUNDTRIP
-
 %a = "test.op"() : () -> !x86.reg<rax>
 // CHECK: %{{.*}} = "test.op"() : () -> !x86.reg<rax>
+
+%0 = x86.get_register : () -> !x86.reg<>
+%1 = x86.get_register : () -> !x86.reg<>
+
+%add = x86.add %0, %1 : (!x86.reg<>, !x86.reg<>) -> !x86.reg<>
+// CHECK: %{{.*}} = x86.add %{{.*}}, %{{.*}} : (!x86.reg<>, !x86.reg<>) -> !x86.reg<>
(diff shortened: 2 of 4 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+
+    def assembly_line_args(self) -> tuple[AssemblyInstructionArg | None, ...]:
+        return self.r1, self.r2
+
+
+@irdl_op_definition
+class AddOp(RROperation[GeneralRegisterType, GeneralRegisterType]):
+    """
+    Adds the registers r1 and r2 and stores the result in r1.
```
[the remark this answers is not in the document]
KGrykiel: My current plan is for there to be seperate classes for immediate and memory reference variants

xdsl/dialects/x86/ops.py
```diff
@@ -0,0 +1,290 @@
...
+def x86_code(module: ModuleOp) -> str:
+    stream = StringIO()
+    print_assembly(module, stream)
+    return stream.getvalue()
+
+
+# endregion
+
+
+class GetAnyRegisterOperation(Generic[R1InvT], IRDLOperation, X86Op):
```
webmiche, line 271: Is this only used for testing? If so, you should be able to use the `test.op`.
Otherwise, I would advise to have a variadic result_def here, so you can create as many registers in one go as you want.

xdsl/dialects/x86/ops.py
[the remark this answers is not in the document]
tarinduj: Were you planning to have a bunch of different AddOps for all possible combinations of operands or just one generic AddOp? If it's just one AddOp, should we add those classes before this patch? We will have a complete add instruction and won't have to modify it later on.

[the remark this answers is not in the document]
KGrykiel: Several different AddOps is probably the simplest way so that was my plan.

xdsl/dialects/x86/ops.py
KGrykiel: Currently it is only for testing purposes however according to the riscv dialect documentation (from where I got this from) it also has the added functionality of "bridging the RISC-V convention that stores the result of function calls in `a0` and `a1` into SSA form". I haven't gotten to that part yet, but x86 has similar convention in places so perhaps it's better to keep it as it might be useful later down the line?

webmiche: My personal feeling would be to go without this operation for now. It can still be added once it's needed. But that's a personal feeling, I don't mind it being here either.

KGrykiel: I like how it keeps a certain level of consistency between the assembly dialects, but if the general consensus is to remove it im fine with that as well 👍

tests/dialects/test_x86.py
```diff
@@ -27,3 +28,11 @@
...
     assert register.is_allocated
     assert register.register_name == name
+
+
+def test_add_op():
+    r1 = TestSSAValue(x86.register.RAX)
+    r2 = TestSSAValue(x86.register.RDX)
+    add_op = x86.AddOp(r1, r2, result=x86.register.RAX)
+
+    assert add_op.assembly_line() == "    add rax, rdx"
```
superlopuh, line 38: This is already tested in the filecheck, we don't need an extra pytest here

xdsl/dialects/x86/__init__.py
```diff
@@ -1,10 +1,11 @@
 from xdsl.ir import Dialect
 
+from .ops import AddOp, GetRegisterOp
 from .register import GeneralRegisterType
 
 X86 = Dialect(
     "x86",
-    [],
+    [AddOp, GetRegisterOp],
```
superlopuh, line 8:
```suggestion
    [AddOp, GetRegisterOp,],
```

tests/filecheck/dialects/x86/x86_ops.mlir
```diff
@@ -1,4 +1,9 @@
 // RUN: XDSL_ROUNDTRIP
-
 %a = "test.op"() : () -> !x86.reg<rax>
 // CHECK: %{{.*}} = "test.op"() : () -> !x86.reg<rax>
+
+%0 = x86.get_register : () -> !x86.reg<>
+%1 = x86.get_register : () -> !x86.reg<>
```
superlopuh, line 6: This seems better done with a test op
```suggestion
%0, %1 = "test.op"() : () -> (!x86.reg<>, !x86.reg<>)
```

xdsl/dialects/x86/ops.py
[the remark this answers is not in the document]
superlopuh: Yeah this is best, we probably want to aim for one-to-one mapping for ops and assembly instructions

superlopuh:
Thank you for opening this PR! I think the code overall looks great, but there's definitely too much of it. I'd recommend splitting this PR and merging the parts one by one, so that each part of the code is easier to review. I'd recommend something like this:
 - add op with no custom syntax and no assembly printing
 - add op assembly printing (use this PR to figure out what should go into "backend" as helpers, such as the comment printing etc)
 - add op custom syntax (use this PR to add a new test for custom assembly for x86 ops)
 - get register op (I don't actually think we want it at all, it was a temporary hack to work around a lack of structured representations of riscv operations, but we might as well build x86 as a structured dialect from the get go, and leverage it for dataflow etc)

tests/dialects/test_x86.py
KGrykiel: I don't think filecheck tests the assembly printing?

superlopuh: ah yes then let's add a filecheck for that, like for riscv :)

xdsl/dialects/x86/ops.py
superlopuh: I agree with Michel, I would propose adding it only when you're sure that there's no alternative

KGrykiel:
I agree it's a lot for a PR. I added the prinitng stuff as well because you mentioned that it's best to have tests for the code in the same PR, but if the pytest in unnecessary then I'll leave printing for later 👍

xdsl/dialects/x86/__init__.py
```diff
@@ -1,10 +1,11 @@
 from xdsl.ir import Dialect
 
+from .ops import AddOp
 from .register import GeneralRegisterType
 
 X86 = Dialect(
     "x86",
-    [],
+    [AddOp],
```
superlopuh, line 8:
```suggestion
    [AddOp,],
```

xdsl/dialects/x86/ops.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+class X86Op(Operation, ABC):
+    """
+    Base class for operations that can be a part of x86 assembly printing.
+    """
+
+    @abstractmethod
+    def assembly_line(self) -> str | None:
+        raise NotImplementedError()
```
superlopuh, line 41: We don't need this functionality yet in the base class, might as well add it later

xdsl/dialects/x86/ops.py
```diff
@@ -0,0 +1,201 @@
...
+    def assembly_instruction_name(self) -> str:
+        """
+        By default, the name of the instruction is the same as the name of the operation.
+        """
+
+        return self.name.split(".", 1)[-1]
+
+    def assembly_line(self) -> str | None:
+        # default assembly code generator
+        raise NotImplementedError()
```
superlopuh, line 151: all of this also can be added later

superlopuh (approved):

tarinduj (approved):

PR #2355, opened 2024-03-21
interpreter: (linalg) Add linalg operations for ONNX model MNIST targeting

### LINKED ISSUE #2354: interpreter: (linalg) Add linalg operations for ONNX model MNIST targeting

kayode-gif:
resolves #2354

### DIFF
--- a/xdsl/interpreters/linalg.py
+++ b/xdsl/interpreters/linalg.py
@@ -68,20 +68,21 @@ class LinalgFunctions(InterpreterFunctions):
     def run_add(
         self, interpreter: Interpreter, op: linalg.AddOp, args: tuple[Any, ...]
     ) -> tuple[Any, ...]:
-        lhs, rhs = args[0], args[1]
+        (
+            lhs,
+            rhs,
+        ) = (
+            args[0],
+            args[1],
+        )
         assert isinstance(lhs, ShapedArray)
         assert isinstance(rhs, ShapedArray)
         lhs = cast(ShapedArray[float], lhs)
         rhs = cast(ShapedArray[float], rhs)
         assert lhs.shape == rhs.shape
-        for i in range(len(lhs.data)):
-            lhs.data[i] += rhs.data[i]
-        if op.results is not None:
-            return (lhs,)
-        return ()
-        # return (
-        #     ShapedArray(list(l + r for l, r in zip(lhs.data, rhs.data)), lhs.shape),
-        # )
+        return (
+            ShapedArray(list(l + r for l, r in zip(lhs.data, rhs.data)), lhs.shape),
+        )
 
     @impl(linalg.FillOp)
     def run_fill(
@@ -93,12 +94,12 @@ class LinalgFunctions(InterpreterFunctions):
         result_type = op.res[0].type
         assert isinstance(result_type, TensorType)
         result_shape = list(result_type.get_shape())
-        operand.data *= math.prod(result_shape)
-        if op.results is not None:
-            return (operand,)
-        # return (
-        #     ShapedArray(list(operand.data * math.prod(result_shape)), result_shape),
-        # )
+        # operand.data *= math.prod(result_shape)
+        # if op.results is not None:
+        #     return (operand,)
+        return (
+            ShapedArray(list(operand.data * math.prod(result_shape)), result_shape),
+        )
 
     @impl(linalg.MulOp)
     def run_mul(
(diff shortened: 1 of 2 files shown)


superlopuh:
I get that this PR is working towards targeting MNIST, but it would be a bit more helpful to add the names of the ops directly instead

xdsl/interpreters/linalg.py
```diff
@@ -61,3 +63,98 @@ def run_yield(
...
+        operand = cast(ShapedArray[float], operand)
+        assert len(operand.shape) == 2
+
+        transposed_data: list[float] = []
+
+        for c in range(operand.shape[1]):
+            for r in range(operand.shape[0]):
+                transposed_data.append(operand.load((r, c)))
+
+        return (ShapedArray(list(transposed_data), operand.shape[::-1]),)
```
superlopuh, line 124: Shaped array has a helper for this

xdsl/interpreters/linalg.py
```diff
@@ -61,3 +63,98 @@ def run_yield(
...
+    ) -> tuple[Any, ...]:
+        lhs, rhs = args[0], args[1]
+        assert isinstance(lhs, ShapedArray)
+        assert isinstance(rhs, ShapedArray)
+        lhs = cast(ShapedArray[float], lhs)
+        rhs = cast(ShapedArray[float], rhs)
+        assert lhs.shape == rhs.shape
+        return (
+            ShapedArray(list(l * r for l, r in zip(lhs.data, rhs.data)), lhs.shape),
+        )
```
superlopuh, line 107: we want to modify the out parameter in place, and return it

xdsl/interpreters/linalg.py
superlopuh: But since we want to modify the out parameter in-place, this won't work

xdsl/interpreters/linalg.py
kayode-gif: to clarify, you mean, do in place addition the the lhs arg and return it in the shaped array,
so return` shapedarray(lhs.data, lhs.shape)`

superlopuh: My understanding of the linalg args is that they pass the destination as the last parameter, in this case the arg, which should be modified. At the `tensor` level, this is sort of implicit, but at the `memref` level the result is not returned, just modified in-place, I think.

superlopuh: I think we need to add the functionality to check the number of operands and arguments to the interpreter. If they're not the same count we have a problem.

superlopuh: If you could open a separate PR for this that would be very helpful

kayode-gif: > I think we need to add the functionality to check the number of operands and arguments to the interpreter. If they're not the same count we have a problem.

this would be modifying `run_op` correct?

kayode-gif: > My understanding of the linalg args is that they pass the destination as the last parameter, in this case the arg, which should be modified. At the `tensor` level, this is sort of implicit, but at the `memref` level the result is not returned, just modified in-place, I think.

oh so modify to the expected result and just `return ()`

superlopuh: I'm not sure how much of this context you're familiar with, so I'll describe my understanding of the whole thing:
In MLIR, there are two high-level shaped pointer abstractions, tensor and memref. Tensor is a value type, meaning that it's conceptually never mutated, and the resulting tensor is returned instead, so you can pretend that it's like a normal SSA value. Memref is a reference type, meaning that operations mutate what the memref contains. There's a process called bufferization that converts between the two. When bufferizing linalg, operations that operate on tensors are transformed into operations on memrefs, and in the process ops that return a value are replaced with ops that do not. So the operation implementation in the interpreter needs to check either the number of results or the types to check what the behaviour should be to be consistent with the behaviour both before and after bufferization.

kayode-gif: i feel like i understand, but ill reclarify: tensors aren't mutable but given memrefs are mutuable the operation value is kept somewhere but not returned explicitly like tensors, so to ensure we are consistent we must check that the no of results from tensor -> memref are the same  before /after we modify the args in place

kayode-gif: if you could include a general pseudocode skeleton of what you mean that be helpful too :)

superlopuh: Basically all these functions need to mutate the out parameters in place, and return the out parameters if the op has results

kayode-gif: okay cool

kayode-gif:
@superlopuh is this fine

superlopuh (commented):
I just realised that it technically isn't. Depending on the operation, linalg may accumulate into the result, instead of updating it. This is a bit difficult to test right now, though. I'd say let's add checks in these functions that the result buffer passed in is all 0.0 for now, and handle the case when it isn't later.

kayode-gif:
> I just realised that it technically isn't. Depending on the operation, linalg may accumulate into the result, instead of updating it. This is a bit difficult to test right now, though. I'd say let's add checks in these functions that the result buffer passed in is all 0.0 for now, and handle the case when it isn't later.

cool, for sure, added them in now

xdsl/interpreters/linalg.py
```diff
@@ -70,6 +70,7 @@ def run_add(
         assert isinstance(lhs, ShapedArray)
         assert isinstance(rhs, ShapedArray)
         assert isinstance(res, ShapedArray)
+        assert all(res.data_ptr[i] == 0 for i in range(len(res.data)))
```
superlopuh: hmm, maybe the best would be to add a helper for this to the class directly, to make the code here a little simpler. Could you please have a go at this? There might be a method on `memoryview` that would give this directly, maybe something like number of set bits == 0

superlopuh: If you feel like you'd rather do that later, please file an issue and we can do it some other time, this seems to work great already

superlopuh (approved):

PR #2359, opened 2024-03-21
interpreter: (onnx) Add rest of onnx operations for ONNX model MNIST targeting 

### LINKED ISSUE #2358: interpreter: (onnx) Add rest of onnx operations for ONNX model MNIST targeting

kayode-gif:
resolves #2358 

Note: Please pay extra attention to onnx.maxpoolsingleout, not so confident on it overall, evnthough its fine

### DIFF
--- a/tests/interpreters/test_onnx_interpreter.py
+++ b/tests/interpreters/test_onnx_interpreter.py
@@ -1,8 +1,20 @@
 import pytest
 
 from xdsl.dialects import onnx
-from xdsl.dialects.builtin import ModuleOp, TensorType, f32
+from xdsl.dialects.builtin import (
+    AnyIntegerAttr,
+    ArrayAttr,
+    DenseIntOrFPElementsAttr,
+    FloatAttr,
+    ModuleOp,
+    NoneType,
+    StringAttr,
+    TensorType,
+    f32,
+    i64,
+)
 from xdsl.interpreter import Interpreter
+from xdsl.interpreters.builtin import BuiltinFunctions
 from xdsl.interpreters.shaped_array import ShapedArray
 from xdsl.utils.test_value import TestSSAValue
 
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/onnx.py
```diff
@@ -66,3 +68,221 @@ def run_relu(self, interpreter: Interpreter, op: onnx.Relu, args: tuple[Any, ...
...
+                if auto_pad == "SAME_UPPER":
+                    pad_top = int(pad_along_height // 2)
+                    pad_bottom = int(pad_along_height - pad_top)
+                    pad_left = int(pad_along_width // 2)
+                    pad_right = int(pad_along_width - pad_left)
+                else:
+                    pad_bottom = int(pad_along_height // 2)
+                    pad_top = int(pad_along_height - pad_bottom)
+                    pad_right = int(pad_along_width // 2)
+                    pad_left = int(pad_along_width - pad_right)
```
superlopuh, line 171: why the ints? I think they're guaranteed to be ints already

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -85,3 +97,542 @@ def test_onnx_relu():
...
+            29,
+            30,
+            31,
+            32,
+            33,
+            34,
+        ],
+        [1, 1, 7, 5],
+    )
+    b = ShapedArray([1, 1, 1, 1, 1, 1, 1, 1, 1], [1, 1, 3, 3])
```
superlopuh, line 536:
```suggestion
    b = ShapedArray([1,] * 9, [1, 1, 3, 3])
```

xdsl/interpreters/onnx.py
kayode-gif: for the function `test_onnx_conv_same_lower_strides` without the ints the pad width is not of integer type

superlopuh: oh then I'd suggest changing the lines for out_height and out_width to add the `int` initialisation, so the downstream calculations are all strictly on integers

superlopuh (approved):

superlopuh:
I'm not 100% sure the implementations are correct, but I look forward to having them tested in the future :)

xdsl/interpreters/onnx.py
kayode-gif: they do have the int stuff

kayode-gif:
> I'm not 100% sure the implementations are correct, but I look forward to having them tested in the future :)

for sure :)

xdsl/interpreters/onnx.py
superlopuh: I don't see it:
```

                out_height = np.ceil(matrix.shape[2] / strides[0])
                out_width = np.ceil(matrix.shape[3] / strides[1])
```

kayode-gif: oh lol, i was looking at the other one, its added now

tests/interpreters/test_onnx_interpreter.py
```diff
@@ -85,3 +97,372 @@ def test_onnx_relu():
...
+def test_onnx_reshape():
+    interpreter = Interpreter(ModuleOp([]))
+    interpreter.register_implementations(OnnxFunctions())
+    op = onnx.Reshape(
+        (TestSSAValue(TensorType(f32, [1, 10]))),
+        (TestSSAValue(TensorType(i64, [2]))),
+        AnyIntegerAttr(0, i64),
+    )
+    a = ShapedArray([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], [1, 10])
+    b = ShapedArray([None], [2])
```
superlopuh, line 134: The more I think about it, the more I'm sure that the inputs have to determine the size of the output, not the output shape. It's likely that the output shape is dynamic, whereas the input is guaranteed to exist. Like in the tensor reshape, we should test that they don't disagree, and use the input data as the new shape

kayode-gif:
@superlopuh just remembered about this pr, seems it has conflicts with that merge you did before

kayode-gif:
@superlopuh turns out max pool can support strides of 2, added a case that seems to work fine

superlopuh (approved):

PR #2362, opened 2024-03-22
interpreter: (interpreter) Add check to see if operand count matches input count in `run_op `

### LINKED ISSUE #2361: interpreter: (interpreter) Add check to see  if operand count matches input count in `run_op`

kayode-gif:
resolves #2361


xdsl/interpreter.py
```diff
@@ -605,6 +605,11 @@ def run_op(self, op: Operation | str, inputs: PythonValues) -> PythonValues:
         if isinstance(op, str):
             op = self.get_op_for_symbol(op)
 
+        if (operands_count := len(op.operands)) != (inputs_count := len(inputs)):
+            raise InterpretationError(
+                f"Number of operands ({operands_count}) doesn't match the number of inputs ({inputs_count})."
+            )
+
```
superlopuh, line 612: please move to _run_op instead, and also check number of results

kayode-gif: > please move to _run_op instead, and also check number of results

this causes a handful of tests to fail, im thinking i should just raise errors rn

superlopuh: Let's fix the tests, that's why I asked you to open this PR :)

kayode-gif: yh, im not to sure how to fix these tests, 
the tests are:
`test_constant in test_arith`
`test_apply and test_functions in test_affine`

superlopuh: The code in interpreter.py still seems broken

superlopuh: the test needs to check that the number of results on the op is the same as the number of results returned by the impl

kayode-gif: okay, it now works i believe, i just have conflicts, you updated the way shapedarray performs

superlopuh:
I fixed the conflicts with my changes

tests/test_interpreter.py
```diff
@@ -134,3 +137,23 @@ class Funcs1(InterpreterFunctions): ...
     assert interpreter.get_data(Funcs0, "d", lambda: {"b": 2}) == {"b": 2}
 
     assert interpreter.get_data(Funcs1, "a", lambda: {"b": 2}) == {"b": 2}
+
+
+def test_run_op_interpreter_args():
```
superlopuh, line 142: Can you please add a test for the number of results, also? Then we're GTG

superlopuh (approved):

tests/test_interpreter.py
```diff
@@ -134,3 +137,23 @@ class Funcs1(InterpreterFunctions): ...
...
+    interpreter = Interpreter(ModuleOp([]))
+    interpreter.register_implementations(TensorFunctions())
+    op = tensor.ReshapeOp(
+        TestSSAValue(TensorType(f32, [4, 1])),
+        TestSSAValue(TensorType(i32, [1])),
+        TensorType(f32, [4]),
+    )
+    a = ShapedArray([1, 2, 3, 4], [4, 1])
```
superlopuh, line 150:
```suggestion
    a = ShapedArray(TypedPtr.new_int32([1, 2, 3, 4]), [4, 1])
```

kayode-gif:
@superlopuh should be gtg :)

xdsl/interpreters/test.py
```diff
@@ -0,0 +1,17 @@
...
+)
+
+
+@register_impls
+class TestFunctions(InterpreterFunctions):
+    @impl(test.TestOp)
+    def run_test(
+        self, interpreter: Interpreter, op: test.TestOp, args: PythonValues
+    ) -> PythonValues:
+        return (range(len(op.res)),)
```
superlopuh, line 17: can you please move this to the test directly? Ideally inside the test function. Test ops do not have any runtime semantics, we should not provide implementations for them.
```suggestion
@register_impls
class TestFunctions(InterpreterFunctions):
    @impl(test.TestOp)
    def run_test(
        self, interpreter: Interpreter, op: test.TestOp, args: PythonValues
    ) -> PythonValues:
        return (1,)
```
I think my initial suggestion over Zulip was not good, the idea is to provide a clearly broken function, which happens to be correct when the number of results of the op matches (in this case, 1 result), and test that the interpreter raises an error if it doesn't match (for example, when given a test op with no results).

tests/test_interpreter.py
```diff
@@ -14,6 +22,10 @@
     register_impls,
 )
 from xdsl.interpreters.builtin import BuiltinFunctions
+from xdsl.interpreters.ptr import TypedPtr
+from xdsl.interpreters.shaped_array import ShapedArray
+from xdsl.interpreters.tensor import TensorFunctions
+from xdsl.interpreters.test import TestFunctions
```
superlopuh, line 28: let's not import these, and only rely on functions defined in this test file

superlopuh: For me the builtin dialect is an unfortunate exception just like in the rest of the framework :)

kayode-gif: okay, that means for op.operands != inputs i need to define another op for that right

kayode-gif: forget this, realised i was doing something dumb

superlopuh (approved):
Can you please add a couple of non-raising cases to the test? With that we're GTG

kayode-gif:
> Can you please add a couple of non-raising cases to the test? With that we're GTG

@superlopuh is it better now>

superlopuh (approved):

PR #2363, opened 2024-03-22
dialects: (linalg) allow for empty results in linalg.fill custom format

### LINKED ISSUE #2228: linalg.fill shouldn't have a fixed result
Hi there,
Haven't really checked the custom format in the operation definition in MLIR, but I've found an example in which `-> type($res)` is optional (I'm referring to https://github.com/xdslproject/xdsl/blob/ae2054d727e261e8873115e182d8bb53bc7ca0f9/xdsl/dialects/linalg.py#L439). See the example here: https://github.com/Xilinx/mlir-air/blob/82d23361cb3843e5d8fc01d746d513f25a224125/test/airhost/51_air_mmult_2x2_channel_broadcast/air.mlir#L30

superlopuh:
Fixes #2228: linalg.fill shouldn't have a fixed result

CC @gabrielrodcanal


Moxinilian (approved):
Could you add a verifier

PR #2364, opened 2024-03-22
interactive : remove bug that happens when argument screen is cleared and entered

dshaaban01:
When pressing a pass that requires argument input, a new screen pops up. when you press "clear" (i.e. clear the screen) and then press "enter" (i.e. dismiss the screen) --> causes error. 

this should fix it.

### DIFF
--- a/xdsl/interactive/add_arguments_screen.py
+++ b/xdsl/interactive/add_arguments_screen.py
@@ -5,7 +5,7 @@ from textual.screen import Screen
 from textual.widgets import Button, TextArea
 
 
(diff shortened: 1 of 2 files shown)


superlopuh:
The 0.00% coverage is a little scary. One thing I'm not sure about is what the behaviour is, it looks from the code that we won't add a pass if the arguments are empy. But we want to provide a way for users to say "just give me the default", which for me is an empty string. It feels like we should instead use `dismiss(None)` for cancel, and "" for defaults

superlopuh (approved):
Thank you!

PR #2366, opened 2024-03-23
dialects: (hw) Add `hw.module` and `hw.output` operations

Moxinilian:
This PR introduces the `hw.module` and its companion `hw.output` from CIRCT, while adding various small enhancements to the general infrastructure to accommodate them. ModuleOps are used to specify self-contained hardware modules, akin to functions in software.

A `hw.module` looks like the following:
```
hw.module @name<param1: i1, param2: i2>(in %a: i8, in %b: i9, out c: i8) {
  %res = arith.constant 0 : i8
  hw.output %res : i8
}
```
While hardware modules have similar semantics to functions, both inputs and outputs in `hw.module` are defined within the parenthesis, as *ports*. This is because some ports may be both input and output in CIRCT (currently not supported). Modules can optionally receive static parameters (in the example, `param1` and `param2`). Those are akin to C++ value templates or Rust const generics, and are useful to factor similar modules together.

`DirectionAttr`, `ModulePort` and `ModuleType` are defined to represent the type of a module and its ports. `ParamDeclAttr` is an attribute to represent the parameter definitions on the module.

The `hw.output` operation is the terminator for `hw.module`, setting the value of the output ports.

cc @superlopuh @math-fehr

### DIFF
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -13,7 +13,6 @@ from typing import overload
 
 from xdsl.dialects.builtin import (
     ArrayAttr,
-    DictionaryAttr,
     FlatSymbolRefAttr,
     IntAttr,
     LocationAttr,
@@ -25,7 +24,6 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.ir import (
     Attribute,
-    Block,
     Data,
     Dialect,
     OpaqueSyntaxAttribute,
@@ -435,9 +433,9 @@ class Direction(Enum):
     Represents the direction of a module port.
     """
 
+    # TODO: support INOUT direction
     INPUT = (0,)
     OUTPUT = (1,)
-    INOUT = (2,)
 
     @staticmethod
     def parse_optional(parser: BaseParser, short: bool = False) -> "Direction | None":
@@ -445,8 +443,6 @@ class Direction(Enum):
             return Direction.INPUT
         elif parser.parse_optional_keyword("output" if not short else "out"):
             return Direction.OUTPUT
-        elif parser.parse_optional_keyword("inout"):
-            return Direction.INOUT
         else:
             return None
 
@@ -461,12 +457,10 @@ class Direction(Enum):
             case Direction.INPUT:
                 printer.print("input" if not short else "in")
             case Direction.OUTPUT:
-                printer.print("output" if not short else "out")
-            case Direction.INOUT:
                 printer.print("inout")
 
     def is_input_like(self) -> bool:
-        return self == Direction.INPUT or self == Direction.INOUT
+        return self == Direction.INPUT
 
 
 @irdl_attr_definition
@@ -581,9 +575,6 @@ _MODULE_OP_ATTRS_HANDLED_BY_CUSTOM_FORMAT: list[str] = [
     "sym_name",
     "module_type",
     "parameters",
-    "per_port_attrs",
-    "comment",
-    "result_locs",
 ]
 
 
@@ -600,25 +591,15 @@ class ModuleOp(IRDLOperation):
     sym_name: SymbolNameAttr = attr_def(SymbolNameAttr)
     module_type: ModuleType = attr_def(ModuleType)
     parameters: ArrayAttr[ParamDeclAttr] | None = opt_attr_def(ArrayAttr[ParamDeclAttr])
-    per_port_attrs: ArrayAttr[DictionaryAttr] | None = opt_attr_def(
-        ArrayAttr[DictionaryAttr]
-    )
-    comment: StringAttr | None = opt_attr_def(StringAttr)
-    result_locs: ArrayAttr[LocationAttr] | None = opt_attr_def(
-        ArrayAttr[LocationAttr], attr_name="resultLocs"
-    )
 
-    body: SingleBlockRegion = region_def()
+    body: SingleBlockRegion = region_def("single_block")
 
     def __init__(
         self,
         sym_name: SymbolNameAttr,
         module_type: ModuleType,
-        body: Region | None = None,
+        body: Region,
         parameters: ArrayAttr[ParamDeclAttr] = ArrayAttr([]),
-        per_port_attrs: ArrayAttr[DictionaryAttr] | None = None,
-        comment: StringAttr | None = None,
-        result_locs: ArrayAttr[LocationAttr] | None = None,
     ):
         attributes: dict[str, Attribute] = {
             "sym_name": sym_name,
@@ -626,32 +607,32 @@ class ModuleOp(IRDLOperation):
             "parameters": parameters,
         }
 
-        if per_port_attrs is not None:
-            attributes["per_port_attrs"] = per_port_attrs
-
-        if comment is not None:
-            attributes["comment"] = comment
-
-        if result_locs is not None:
-            attributes["result_locs"] = result_locs
-
-        if body is None:
-            body = Region(
-                [
-                    Block(
-                        arg_types=(
-                            port.type
-                            for port in module_type.ports.data
-                            if port.dir.data.is_input_like()
-                        )
-                    )
-                ]
-            )
-
         return super().__init__(
             attributes=attributes,
             regions=[body],
         )
+
+    def verify_(self) -> None:
+        if self.parameters is not None:
+            # FIXME: once xDSL supports typed attributes, check that parameter
+            # types are consistent with their default values
+            param_names = [param.port_name.data for param in self.parameters.data]
+            if len(set(param_names)) != len(param_names):
+                raise VerifyException("module has two parameters of same name")
+        block_args = iter(self.body.block.args)
+        for i, port in enumerate(self.module_type.ports.data):
+            if not port.dir.data.is_input_like():
+                continue
+            if (next_block_arg := next(block_args, None)) is None:
+                raise VerifyException("missing block arguments in module block")
+            if port.type != next_block_arg.type:
+                raise VerifyException(
+                    f"input-like port #{i} has inconsistent type with its matching "
+                    f"module block argument (expected {port.type}, block argument "
+                    f"is of type {next_block_arg.type})"
+                )
+        if next(block_args, None) is not None:
+            raise VerifyException("too many block arguments in module block")
 
     @classmethod
     def parse(cls, parser: Parser) -> "ModuleOp":
@@ -689,8 +670,6 @@ class ModuleOp(IRDLOperation):
             if not isinstance(port_type, TypeAttribute):
                 parser.raise_error("port type must be a type attribute")
 
-            # TODO BEFORE PR: handle inout type wrapping
-
             port_attrs = parser.parse_optional_attr_dict()
             port_location = parser.parse_optional_location()
 
@@ -717,10 +696,6 @@ class ModuleOp(IRDLOperation):
         # Extract the ModuleOp attributes and arguments from the parsed arg port list
         region_args: list[Parser.Argument] = []
         module_ports: list[ModulePort] = []
-        per_port_attrs: list[DictionaryAttr] = []
-        has_non_empty_per_port_attr = False
-        result_locs: list[LocationAttr] = []
-        has_known_location = False
         for arg in args:
             if arg.port_ssa:
                 region_args.append(arg.port_ssa)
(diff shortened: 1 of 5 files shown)


xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
             printer.print_string("]")
 
 
+class Direction(Enum):
+    """
+    Represents the direction of a module port.
+    """
+
+    # TODO: support INOUT direction
```
superlopuh, line 434: Can you file an issue please and add it to the comment

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        def parse_port() -> ModulePort:
+            direction = Direction.parse(parser)
+            if (
+                name := parser.parse_optional_identifier()
+                or parser.parse_optional_str_literal()
+            ) is None:
+                parser.raise_error(
+                    "expected port name as identifier or string litteral"
```
superlopuh, line 512:
```suggestion
                    "expected port name as identifier or string literal"
```

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+@irdl_attr_definition
+class ParamDeclAttr(ParametrizedAttribute):
+    name = "hw.param.decl"
+
+    port_name: ParameterDef[StringAttr]
+    type: ParameterDef[TypeAttribute]
+
+    @classmethod
+    def parse_free_standing_parameters(
+        cls, parser: AttrParser, always_parse_name_in_quotes: bool = False
```
superlopuh, line 547: Can you please document the behaviour of this argument?

superlopuh: Probably with an example of what it can and cannot parse

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+            name = parser.parse_optional_identifier()
+        if name is None:
+            parser.raise_error("expected parameter name")
+        parser.parse_punctuation(":")
+        typ = parser.parse_attribute()
+        if not isinstance(typ, TypeAttribute):
+            parser.raise_error("expected type attribute for parameter")
+
+        if parser.parse_optional_punctuation("=") is not None:
+            # TODO: support default values for parameters
```
luisacicolini, line 564: Should this be an issue too? CC: @superlopuh

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+    ) -> Sequence[Attribute]:
+        """
+        Parses the parameter declaration without the encompassing angle brackets.
+        """
+
+        name = parser.parse_optional_str_literal()
+        if name is None and not always_parse_name_in_quotes:
+            name = parser.parse_optional_identifier()
+        if name is None:
+            parser.raise_error("expected parameter name")
```
superlopuh, line 557:
```suggestion
        if name is None:
            if always_parse_name_in_quotes:
                parser.raise_error("expected parameter name as string literal")
            name = parser.expect(parser.parse_optional_identifier, "expected parameter name")
```

xdsl/dialects/hw.py
superlopuh, line 564: can you please file an issue

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+        parser.parse_punctuation(":")
+        typ = parser.parse_attribute()
+        if not isinstance(typ, TypeAttribute):
+            parser.raise_error("expected type attribute for parameter")
+
+        if parser.parse_optional_punctuation("=") is not None:
+            # TODO: support default values for parameters
+            parser.raise_error("default values for parameters are not yet supported")
+
+        return [StringAttr(name), typ]
```
superlopuh, line 567: nit:
```suggestion
        return (StringAttr(name), typ)
```

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+                    f"module block argument (expected {port.type}, block argument "
+                    f"is of type {next_block_arg.type})"
+                )
+        if next(block_args, None) is not None:
+            raise VerifyException("too many block arguments in module block")
+
+    @classmethod
+    def parse(cls, parser: Parser) -> "ModuleOp":
+        @dataclass
+        class ModuleArg:
```
superlopuh, line 662:
```suggestion
        class ModuleArg(NamedTuple):
```

superlopuh: These are a little simpler and have smaller overhead

luisacicolini (approved):
looks good to me!

xdsl/dialects/hw.py
Moxinilian: #2368

xdsl/dialects/hw.py
Moxinilian: #2367

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+
+    def verify_(self) -> None:
+        parent = self.parent_op()
+        assert isinstance(parent, ModuleOp)
+
+        expected_results = [
+            port.type
+            for port in parent.module_type.ports.data
+            if port.dir.data == Direction.OUTPUT
+        ]
```
superlopuh, line 812: nit:
```suggestion
        expected_results = tuple(
            port.type
            for port in parent.module_type.ports.data
            if port.dir.data == Direction.OUTPUT
        )
```

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+                    f"output #{i} is of unexpected type (expected {expected}, got {got.type})"
+                )
+
+    @classmethod
+    def parse(cls, parser: Parser) -> "OutputOp":
+        operands = parser.parse_optional_undelimited_comma_separated_list(
+            parser.parse_optional_unresolved_operand, parser.parse_unresolved_operand
+        )
+        if operands is None:
+            return cls([])
```
superlopuh, line 831: nit:
```suggestion
            return cls(())
```

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,437 @@ def print_parameters(self, printer: Printer):
...
+        printer.print(" ")
+        printer.print_list(self.inputs, printer.print_operand)
+        printer.print(" : ")
+        printer.print_list((x.type for x in self.inputs), printer.print_attribute)
+
+
 HW = Dialect(
     "hw",
-    [],
+    [ModuleOp, OutputOp],
```
superlopuh, line 852:
```suggestion
    [ModuleOp, OutputOp,],
```

xdsl/utils/lexer.py
```diff
@@ -506,6 +506,7 @@ def lex(self) -> Token:
             f"Unexpected character: {current_char}",
         )
 
+    bare_identifier_regex = re.compile(r"[a-zA-Z_][a-zA-Z0-9_$.]*")
     bare_identifier_suffix_regex = re.compile(r"[a-zA-Z0-9_$.]*")
```
superlopuh, line 510: It seems like a good idea to do this in a separate PR, with explicit tests, probably reusing the suffix also
```suggestion
    SUFFIX = r"[a-zA-Z0-9_$.]*"
    bare_identifier_regex = re.compile(r"[a-zA-Z_]" + SUFFIX)
    bare_identifier_suffix_regex = re.compile(SUFFIX)
```

xdsl/printer.py
```diff
@@ -226,12 +227,14 @@ def _print_results(self, op: Operation) -> None:
         self.print_list(op.results, self.print)
         self.print(" = ")
 
-    def print_ssa_value(self, value: SSAValue) -> None:
+    def print_ssa_value(self, value: SSAValue) -> str:
```
superlopuh, line 230: can you please do this in a separate PR also, with associated tests? Probably there are existing tests to update

superlopuh (approved):

xdsl/printer.py
superlopuh: #2377

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,443 @@ def print_parameters(self, printer: Printer):
...
+        return direction
+
+    def print(self, printer: Printer, short: bool = False) -> None:
+        match self:
+            case Direction.INPUT:
+                printer.print("input" if not short else "in")
+            case Direction.OUTPUT:
+                printer.print("output" if not short else "out")
+
+    def is_input_like(self) -> bool:
```
lucjaulmes, line 460: Really feels like `is_input` at the moment. Is this meant to change when INOUT is supported? Ive seen a couple of `isInput() &` !isInOut()` in upstream code but not so much `isInput() ||isInOut()` so I’m curious where this is going.

lucjaulmes (commented):
Thanks, nice contribution! A bit hard to get my head around all the various parts though. Here’s a few thoughts

xdsl/dialects/hw.py
Moxinilian: Yeah this is the case. I used to go further into the inout implementation and this was useful, notably in the parser/printer to decide on whether it should be an SSA value or not.

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,443 @@ def print_parameters(self, printer: Printer):
             printer.print_string("]")
 
 
+class Direction(Enum):
+    """
+    Represents the direction of a module port.
+    """
+
+    # TODO: support INOUT direction (https://github.com/xdslproject/xdsl/issues/2368)
+    INPUT = (0,)
```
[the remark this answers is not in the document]
Moxinilian: I think this is the result of formatting or something, when I wrote it it was indeed int values. I'll try to get rid of this.

xdsl/dialects/hw.py
```diff
@@ -409,12 +426,443 @@ def print_parameters(self, printer: Printer):
...
+
+_MODULE_OP_ATTRS_HANDLED_BY_CUSTOM_FORMAT: list[str] = [
+    "sym_name",
+    "module_type",
+    "parameters",
+]
+
+
+@irdl_op_definition
+class ModuleOp(IRDLOperation):
```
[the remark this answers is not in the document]
lucjaulmes: The trait is just called `InnerRefNamespace` in the circt codebase – the suffix is added due to a name clash (there is also an `InnerRefNamsespace` class). The trait definition is in `include/circt/Dialect/HW/HWOpInterfaces.h` and `include/circt/Dialect/HW/HWOpInterfaces.td`. There’s some brief explanation about it in the “Traits and Classes” section of circt’s `docs/RationaleSymbols.md` document.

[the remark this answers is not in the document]
lucjaulmes: The operation that has that trait is the firrtl `CircuitOp` operation.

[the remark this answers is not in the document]
Moxinilian: Oh of course! Right. I'm just going to add it to the module operation, hopefully this makes sense.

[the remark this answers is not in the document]
Moxinilian: > The operation that has that trait is the firrtl CircuitOp operation.

That does not sound right, HwModules can live in builtin.module too. I think they may also be adding it to MLIR's ModuleOp via an external model, but I cannot find where this is.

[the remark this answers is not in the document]
Moxinilian: Ah no, I think the problem is in the InnerSymbolTable trait implementation in xDSL. It does not check for an InnerRefNamespaceLike but an InnerRefNamespace. I can probably fix that.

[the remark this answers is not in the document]
Moxinilian: Actually I don't understand what is happening in InnerSymbolTableTrait, so if you don't mind I just will not add the trait for now. My understanding is that the implementation of `InnerSymbolTableTrait` is not complete enough to be used anyway.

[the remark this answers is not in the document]
lucjaulmes: I think that’s fine, let’s not make this PR to large. I suppose the propoer way to add it later is to just add a comment and file a follow-up issue. LGTM for the rest.

PR #2375, opened 2024-03-25
dialects: (x86) PR3 - Assembly emission

KGrykiel:
I've re-added the printing functionality with a filecheck to test it. I have temporarily brought back the get_register method to get the PR out, so please let me know what's the better way to do it. test.op no longer works when a particular register needs to be specified (it fails an assertion error during the printing). Riscv did it by starting off with operations that don't need register arguments, but I don't have those yet.

### DIFF
--- a/tests/filecheck/dialects/x86/x86_assembly_emission.mlir
+++ b/tests/filecheck/dialects/x86/x86_assembly_emission.mlir
@@ -1,0 +1,7 @@
+// RUN: xdsl-opt -t x86-asm %s | filecheck %s
+
+%0 = x86.get_register : () -> !x86.reg<rax>
+%1 = x86.get_register : () -> !x86.reg<rdx>
+
+%add = x86.add %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
+// CHECK: add rax, rdx
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
-from abc import ABC
+from abc import ABC, abstractmethod
 from collections.abc import Sequence, Set
-from typing import Generic, TypeAlias, TypeVar
+from io import StringIO
+from typing import IO, Generic, TypeAlias, TypeVar
 
 from typing_extensions import Self
 
 from xdsl.dialects.builtin import (
     AnyIntegerAttr,
+    ModuleOp,
     StringAttr,
 )
 from xdsl.ir import (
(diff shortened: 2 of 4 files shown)


tests/filecheck/dialects/x86/x86_assembly_emission.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: xdsl-opt -t x86-asm %s | filecheck %s
+
+%0 = x86.get_register : () -> !x86.reg<rax>
+%1 = x86.get_register : () -> !x86.reg<rdx>
```
superlopuh, line 4: Please use "test.op" instead
```suggestion
%0, %1 = "test.op"() : () -> (!x86.reg<rax>, !x86.reg<rdx>)
```

superlopuh: With this, the GetRegister op is not needed for this PR

xdsl/dialects/x86/ops.py
```diff
@@ -167,3 +208,83 @@ class AddOp(RROperation[GeneralRegisterType, GeneralRegisterType]):
...
+    name: str,
+    arg_str: str,
+    comment: StringAttr | None = None,
+    is_indented: bool = True,
+) -> str:
+    code = "    " if is_indented else ""
+    code += name
+    if arg_str:
+        code += f" {arg_str}"
+    code = _append_comment(code, comment)
```
superlopuh, line 250: I'd recommend moving the assembly printing helpers to `assembly.py`

webmiche:
> test.op no longer works when a particular register needs to be specified (it fails an assertion error during the printing). 

I am very surprised by that. You "should" be able to generate SSAValues with any type through the test.op. What is your error?

KGrykiel:
> > test.op no longer works when a particular register needs to be specified (it fails an assertion error during the printing).
> 
> I am very surprised by that. You "should" be able to generate SSAValues with any type through the test.op. What is your error?

I'm getting an error at the assertion at line 256:
`        assert isinstance(op, X86Op), f"{op}"`
it says that test_op is not a X86Op. Removing the asserion doesn't help either as the rest of the method stops working due to there not being a "assembly_line" method for a class of type "Operation"

webmiche:
Ah, I see. So you essentially want to print these things using your own printing assembly infrastructure and you do not have the test.op in there.

Other classic suggestions are wrapping this in a block with blockarguments to get the SSAValues from there, I am not sure this will work in this case though.

superlopuh (approved):
My bad for not reading the PR description. I see the issue. I think in that case we might as well add it.

xdsl/dialects/x86/ops.py
superlopuh: no problem, I'll file an issue

KGrykiel:
@webmiche @tarinduj is it fine if I merge and move on to the next PR?

xdsl/dialects/x86/ops.py
superlopuh: #2377

tarinduj (approved):

webmiche:
> @webmiche @tarinduj is it fine if I merge and move on to the next PR?

Sure, I am fine for the moment. I think we should find a more principled way to model registers and global state instead fo spawning them from thin air, but that feels like a more general refactor than what this PR attempts. Feel free to merge!

PR #2378, opened 2024-03-26
transformations: (reconcile-unrealized-casts) Disable warnings in pattern by default

Moxinilian:
The pattern for unrealized casts reconciliation emits a lot of warnings when it fails to match. While this makes sense as a pass application, it clutters output in patterns where reconciliation is expected to fail. This PR makes the emission of warning togglable, disabled by default in the pattern and enabled in the pass.

### DIFF
--- a/xdsl/transforms/reconcile_unrealized_casts.py
+++ b/xdsl/transforms/reconcile_unrealized_casts.py
@@ -1,6 +1,7 @@
 import itertools
 from collections import deque
 from collections.abc import Iterable
+from dataclasses import dataclass
 from warnings import warn
 
 from xdsl.dialects import builtin
@@ -16,7 +17,9 @@ from xdsl.pattern_rewriter import (
 
 
 def _try_remove_cast_chain(
-    op: builtin.UnrealizedConversionCastOp, rewriter: PatternRewriter
+    op: builtin.UnrealizedConversionCastOp,
+    rewriter: PatternRewriter,
+    with_warnings: bool,
 ):
     # casts that either have no uses or have at least
     # one user that isn't an unrealized cast.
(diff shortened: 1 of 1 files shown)


adutilleul (approved):

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -89,27 +94,32 @@ def gen_all_uses_cast(
...
     @op_type_rewrite_pattern
     def match_and_rewrite(
         self, op: builtin.UnrealizedConversionCastOp, rewriter: PatternRewriter
     ):
-        _try_remove_cast_chain(op, rewriter)
+        _try_remove_cast_chain(op, rewriter, self.with_warnings)
 
 
-def reconcile_unrealized_casts(module: ModuleOp):
+def reconcile_unrealized_casts(module: ModuleOp, with_warnings: bool = True):
```
superlopuh, line 114: I'd suggest a couple of small changes here and in the pattern:
```suggestion
def reconcile_unrealized_casts(module: ModuleOp, *, warn_on_error: bool = True):
```
The first is to make the flag a keyword argument. Someone reading code that calls this functions will need to look at the implementation of the function to understand what `reconcile_unrealized_casts(op, True)` means, whereas `reconcile_unrealized_casts(op, warn_on_error=True)` is self-documenting. I'd also say that my suggested renaming is a little clearer about the effect of the flag, although that's a nit/subjective thing, I think there are a bunch of ways to phrase it, and the one in this PR is ok also.

xdsl/transforms/reconcile_unrealized_casts.py
```diff
@@ -89,27 +94,32 @@ def gen_all_uses_cast(
...
 
+@dataclass
 class ReconcileUnrealizedCastsPattern(RewritePattern):
     """
     Removes the chains of `builtin.unrealized_conversion_cast` operations
     that are no longer necessary and that start with the matched
     `builtin.unrealized_conversion_cast`.
     """
 
+    with_warnings: bool = False
```
superlopuh, line 105: Similarly to below, this is the way dataclasses specifies that the properties must be passed as keywords to the dataclass generated init. The alternative is `with_warnings: bool = field(default=True, kw_only=True)`, matter of style
```suggestion
    _: KW_ONLY
    with_warnings: bool = False
```

superlopuh (approved):

PR #2379, opened 2024-03-26
dialects: (x86) PR4 - two-register-operand instructions

KGrykiel:
This one's pretty straightforward, added the instructions sub, imul, and, or. xor and mov for the 2 register operands configuration and the related tests.


superlopuh (approved):
beautiful

tarinduj (approved):

PR #2382, opened 2024-03-26
transformations: (onnx) lower onnx.MaxPoolSingleOut to linalg

kayode-gif:
wip #2381

### DIFF
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -70,7 +70,7 @@
 // CHECK-NEXT:   %t26 = "test.op"() : () -> tensor<1x16x14x14xf32>
 // CHECK-NEXT:  %res_max_pool_single_out = tensor.empty() : tensor<3x3xf32>
 // CHECK-NEXT:  %res_max_pool_single_out_1 = tensor.empty() : tensor<1x16x4x4xf32>
-// CHECK-NEXT:  %res_max_pool_single_out_2 = arith.constant 1.844224e+19 : f64
+// CHECK-NEXT:  %res_max_pool_single_out_2 = arith.constant -inf : f64
 // CHECK-NEXT:  %res_max_pool_single_out_3 = linalg.fill ins(%res_max_pool_single_out_2 : f64) outs(%res_max_pool_single_out_1 : tensor<1x16x4x4xf32>) -> tensor<1x16x4x4xf32>
 // CHECK-NEXT:  %res_max_pool_single_out_4 = linalg.pooling_nchw_max {"dilations" = dense<1> : tensor<2xi64>, "strides" = dense<3> : tensor<2xi64>} ins(%t26, %res_max_pool_single_out : tensor<1x16x14x14xf32>, tensor<3x3xf32>) outs(%res_max_pool_single_out_3 : tensor<1x16x4x4xf32>) -> tensor<1x16x4x4xf32>
 
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -70,7 +70,7 @@
 // CHECK-NEXT:   %t26 = "test.op"() : () -> tensor<1x16x14x14xf32>
 // CHECK-NEXT:  %res_max_pool_single_out = tensor.empty() : tensor<3x3xf32>
 // CHECK-NEXT:  %res_max_pool_single_out_1 = tensor.empty() : tensor<1x16x4x4xf32>
-// CHECK-NEXT:  %res_max_pool_single_out_2 = arith.constant -inf : f64
+// CHECK-NEXT:  %res_max_pool_single_out_2 = arith.constant -1.000000e+308 : f64
 // CHECK-NEXT:  %res_max_pool_single_out_3 = linalg.fill ins(%res_max_pool_single_out_2 : f64) outs(%res_max_pool_single_out_1 : tensor<1x16x4x4xf32>) -> tensor<1x16x4x4xf32>
 // CHECK-NEXT:  %res_max_pool_single_out_4 = linalg.pooling_nchw_max {"dilations" = dense<1> : tensor<2xi64>, "strides" = dense<3> : tensor<2xi64>} ins(%t26, %res_max_pool_single_out : tensor<1x16x14x14xf32>, tensor<3x3xf32>) outs(%res_max_pool_single_out_3 : tensor<1x16x4x4xf32>) -> tensor<1x16x4x4xf32>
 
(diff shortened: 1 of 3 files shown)


superlopuh (commented):
We might be better off with a named op here, like https://mlir.llvm.org/docs/Dialects/Linalg/#linalgpooling_ndhwc_max-linalgpoolingndhwcmaxop

I'm not sure which of the linalg max pooling ops corresponds most closely to the ONNX semantics, can you please take a look into that and document it?

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -64,6 +64,15 @@
 // CHECK-NEXT:  %res_gemm_2_1 = linalg.matmul ins(%8, %t12 : tensor<5x10xf32>, tensor<10x3xf32>) outs(%res_gemm_2 : tensor<5x3xf32>) -> tensor<5x3xf32>
 // CHECK-NEXT:  %res_gemm_2_2 = linalg.add ins(%res_gemm_2_1, %10 : tensor<5x3xf32>, tensor<5x3xf32>) outs(%res_gemm_2_1 : tensor<5x3xf32>) -> tensor<5x3xf32>
 
+%t26 = "test.op"(): () ->  (tensor<1x16x14x14xf32>)
+%res_max_pool_single_out = "onnx.MaxPoolSingleOut"(%t26) {onnx_node_name = "/MaxPoolSingleOut", "auto_pad" = "NOTSET", "ceil_mode" = 0 : si64, "kernel_shape" = [3 : i64, 3 : i64],  "dilations" = [1 : i64, 1 : i64],  "pads" = [0 : i64, 0 : i64, 0 : i64, 0 : i64],  "storage_order" = 0 : si64, strides = [3 : i64, 3 : i64]} : (tensor<1x16x14x14xf32>) -> tensor<1x16x4x4xf32>
+
+// CHECK-NEXT:   %t26 = "test.op"() : () -> tensor<1x16x14x14xf32>
+// CHECK-NEXT:  %res_max_pool_single_out = tensor.empty() : tensor<3x3xf32>
+// CHECK-NEXT:  %res_max_pool_single_out_1 = tensor.empty() : tensor<1x16x4x4xf32>
+// CHECK-NEXT:  %res_max_pool_single_out_2 = arith.constant 0.000000e+00 : f64
```
superlopuh, line 73: this will only store the maximum above 0.0, but we need to include the maximum even if it's negative. This is a little more tricky, we need to initialise the output buffer with -inf. Here's the way that iree-compile does it:
`%cst_0 = arith.constant 0xFF800000 : f32`
This is for a f32, though, I think the f64 version should be `0xFF00000000000000 : f64`

superlopuh: http://bartaz.github.io/ieee754-visualization/

superlopuh: Top 12 bits to 1

xdsl/transforms/convert_onnx_to_linalg.py
```diff
@@ -276,6 +278,58 @@ def match_and_rewrite(self, gemm: onnx.Gemm, rewriter: PatternRewriter, /):
...
+            value.value.data for value in max_pool_single_out.dilations.data
+        ]
+        strides: list[int] = [
+            value.value.data for value in max_pool_single_out.strides.data
+        ]
+        kernel_shape = TensorType(f32, kernel)
+
+        # Lowering of unequal kernel and stride shapes not supported
+        if strides != kernel:
+            raise NotImplementedError()
```
superlopuh, line 300: what's the purpose of this check? My impression is that it should work fine without it, am I missing something?

kayode-gif: yh it should, i removed it

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -64,6 +64,15 @@
 // CHECK-NEXT:  %res_gemm_2_1 = linalg.matmul ins(%8, %t12 : tensor<5x10xf32>, tensor<10x3xf32>) outs(%res_gemm_2 : tensor<5x3xf32>) -> tensor<5x3xf32>
 // CHECK-NEXT:  %res_gemm_2_2 = linalg.add ins(%res_gemm_2_1, %10 : tensor<5x3xf32>, tensor<5x3xf32>) outs(%res_gemm_2_1 : tensor<5x3xf32>) -> tensor<5x3xf32>
 
+%t26 = "test.op"(): () ->  (tensor<1x16x14x14xf32>)
+%res_max_pool_single_out = "onnx.MaxPoolSingleOut"(%t26) {onnx_node_name = "/MaxPoolSingleOut", "auto_pad" = "NOTSET", "ceil_mode" = 0 : si64, "kernel_shape" = [3 : i64, 3 : i64],  "dilations" = [1 : i64, 1 : i64],  "pads" = [0 : i64, 0 : i64, 0 : i64, 0 : i64],  "storage_order" = 0 : si64, strides = [3 : i64, 3 : i64]} : (tensor<1x16x14x14xf32>) -> tensor<1x16x4x4xf32>
+
+// CHECK-NEXT:   %t26 = "test.op"() : () -> tensor<1x16x14x14xf32>
+// CHECK-NEXT:  %res_max_pool_single_out = tensor.empty() : tensor<3x3xf32>
+// CHECK-NEXT:  %res_max_pool_single_out_1 = tensor.empty() : tensor<1x16x4x4xf32>
+// CHECK-NEXT:  %res_max_pool_single_out_2 = arith.constant 1.844224e+19 : f64
```
superlopuh, line 73: as you can see, this is incorrect, unfortunately

superlopuh: I wonder if passing `float('-inf')` would work here

kayode-gif: ` %res_max_pool_single_out_2 = arith.constant -inf : f64`
i think it does

kayode-gif: actually erm

superlopuh: Can you please check that it works with MLIR?

kayode-gif: https://peps.python.org/pep-0754/

so -1e400 is equivalent to float('-inf') so can't it be sufficent to but that or just under that

kayode-gif: > https://peps.python.org/pep-0754/
> 
> so -1e400 is equivalent to float('-inf') so can't it be sufficent to but that or just under that

the sys max is -1e308, so anything higher returns -inf

kayode-gif: @superlopuh is this okay?

superlopuh: I'd say this is still a little broken but at this point it's not a blocker. Could you please file an issue for this, and add it in a comment so readers aren't confused if they come across it in the future? Something like:
"dialects: arith Add a way to represent +-inf arith.constant in xDSL that roundtrips with MLIR for f32 and f64"

kayode-gif: will do, for sure :)

superlopuh (approved):

PR #2389, opened 2024-03-27
transforms: Add function-constant-pinning pass to specialize functions

AntonLydike:
This adds the `function-constant-pinning` pass that takes annotated functions and creates specialized versions of them that pin a value to a constant.

Currently only supports pinning at the top level of a function. Also other limitations, that will be documented on a train ride soon!

See the filecheck for examples on how the pass works!


xdsl/transforms/function_specialization.py
```diff
@@ -0,0 +1,222 @@
...
+        specialization_vals = get_op_specialization(split_op)
+        # drop malformed or empty specializations:
+        if specialization_vals is None or len(specialization_vals) == 0:
+            split_op.attributes.pop(SPEZIALIZE_ON_VALS_ATTR)
+            return
+
+        return_types = tuple(func_op.function_type.outputs)
+
+        function_remainder = split_op.next_op
+        assert function_remainder is not None, "Can't specialize on terminator!"
```
PapyChacal, line 41: Can you avoid making the pass fail and maybe bail out instead?

xdsl/transforms/function_specialization.py
```diff
@@ -0,0 +1,222 @@
...
+    - clone the function
+    - rename it to be uniquely named inside the module
+    - erase all operations up until the specialized operation
+    - replace the specialized operation with a constant instantiation
+    """
+    # clone the function including the body:
+    new_func = func_op.clone()
+    # get the module op
+    module = func_op.parent_op()
+    assert isinstance(module, builtin.ModuleOp), "func must be top-level functions!"
```
PapyChacal, line 124: idem

xdsl/transforms/function_specialization.py
AntonLydike: Done!

xdsl/transforms/function_specialization.py
```diff
@@ -0,0 +1,227 @@
+from collections.abc import Iterable
```
superlopuh, line 1: Can you please add a docstring here, explaining what the pass does, who should use it, and when? I think I have a rough idea but it would be nice to have a little guide. Does MLIR have an equivalent? Or LLVM? Maybe could link to those also?

AntonLydike: Added docstring to ModulePass!

superlopuh (approved):
On one hand, the code looks ok. On the other hand, I would have preferred a slightly decomposed API. It feels like outlining the body should be separate to splitting the code on it being a constant. It feels like these could be separate Python helpers, with their own Python API, instead of a pass which seems quite specialized for a project. My understanding of MLIR and IREE is that this is a pattern that is encouraged. If it unblocks you, I think we should be ok to merge, but I'd be a tiny bit less concerned with the maintainability of this code if it were split into self-contained PRs, one of which provides and tests splitting IR based on a constant value, another one of which provides and tests outlining blocks to functions, and maybe another one that composes them into a pass.

AntonLydike:
> On one hand, the code looks ok. On the other hand, I would have preferred a slightly decomposed API. It feels like outlining the body should be separate to splitting the code on it being a constant. It feels like these could be separate Python helpers, with their own Python API, instead of a pass which seems quite specialized for a project. My understanding of MLIR and IREE is that this is a pattern that is encouraged. If it unblocks you, I think we should be ok to merge, but I'd be a tiny bit less concerned with the maintainability of this code if it were split into self-contained PRs, one of which provides and tests splitting IR based on a constant value, another one of which provides and tests outlining blocks to functions, and maybe another one that composes them into a pass.

I do understand your concerns here, I think this code will evolve over time, and I see a good chance that we yank it out again later in case we don't need it. Maybe we should put this in an experimental folder?

superlopuh:
Oh yeah experimental would make me less hesitant for sure, then YOLO ship it

PR #2396, opened 2024-03-27
transformations: (onnx) lower onnx.Conv to linalg 

kayode-gif:
wip #2395

i used xdsl/tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir to get an idea of how to do it, i feel this is "close"

### DIFF
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -80,7 +80,6 @@
 // CHECK-NEXT:  %t20, %t21, %t22 = "test.op"() : () -> (tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>, none)
 // CHECK-NEXT:  %res_conv_2 = tensor.empty() : tensor<1x1x3x3xf32>
 // CHECK-NEXT:  %res_conv_2_1 = linalg.conv_2d_nchw_fchw {"dilations" = dense<1> : tensor<2xi64>, "strides" = dense<1> : tensor<2xi64>} ins(%t20, %t21 : tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>) outs(%res_conv_2 : tensor<1x1x3x3xf32>) -> tensor<1x1x3x3xf32>
-// CHECK-NEXT:  %res_conv_2_2 = linalg.add ins(%t22 : none) outs(%res_conv_2_1 : tensor<1x1x3x3xf32>) -> tensor<1x1x3x3xf32>
 
 
 
--- a/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
+++ b/tests/filecheck/transforms/convert_onnx_to_linalg.mlir
@@ -80,7 +80,8 @@
 // CHECK-NEXT:  %t20, %t21, %t22 = "test.op"() : () -> (tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>, none)
 // CHECK-NEXT:  %res_conv_2 = tensor.empty() : tensor<1x1x3x3xf32>
 // CHECK-NEXT:  %res_conv_2_1 = linalg.conv_2d_nchw_fchw {"dilations" = dense<1> : tensor<2xi64>, "strides" = dense<1> : tensor<2xi64>} ins(%t20, %t21 : tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>) outs(%res_conv_2 : tensor<1x1x3x3xf32>) -> tensor<1x1x3x3xf32>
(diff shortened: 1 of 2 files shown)


superlopuh (commented):
Why not lower to a named op?
https://mlir.llvm.org/docs/Dialects/Linalg/#linalgpooling_nchw_max-linalgpoolingnchwmaxop

kayode-gif:
> Why not lower to a named op? https://mlir.llvm.org/docs/Dialects/Linalg/#linalgpooling_nchw_max-linalgpoolingnchwmaxop

i guess it was because i dont want to write more interpreter ops, but it may be smoother

kayode-gif:
@superlopuh

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -74,7 +74,13 @@
...
 
+%t20, %t21, %t22 = "test.op"() : () -> (tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>, none)
+%res_conv_2 = "onnx.Conv"(%t20, %t21, %t22) {onnx_node_name = "/Conv", "auto_pad" = "NOTSET", "group" = 1 : i64, "kernel_shape" = [3 : i64, 3 : i64], "dilations" = [1 : i64, 1 : i64], "strides" = [1 : i64, 1 : i64], "pads" = [0 : i64, 0 : i64, 0: i64, 0 : i64]}: (tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>, none) -> tensor<1x1x3x3xf32>
 
+// CHECK-NEXT:  %t20, %t21, %t22 = "test.op"() : () -> (tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>, none)
+// CHECK-NEXT:  %res_conv_2 = tensor.empty() : tensor<1x1x3x3xf32>
+// CHECK-NEXT:  %res_conv_2_1 = linalg.conv_2d_nchw_fchw {"dilations" = dense<1> : tensor<2xi64>, "strides" = dense<1> : tensor<2xi64>} ins(%t20, %t21 : tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>) outs(%res_conv_2 : tensor<1x1x3x3xf32>) -> tensor<1x1x3x3xf32>
+// CHECK-NEXT:  %res_conv_2_2 = linalg.add ins(%t22, %res_conv_2_1 : none, tensor<1x1x3x3xf32>) outs(%res_conv_2 : tensor<1x1x3x3xf32>) -> tensor<1x1x3x3xf32>
```
superlopuh, line 83:
```suggestion
// CHECK-NEXT:  %res_conv_2_2 = linalg.add ins(%t22) outs(%res_conv_2_1 : tensor<1x1x3x3xf32>) -> tensor<1x1x3x3xf32>
```
My understanding of the semantics of the operation is that it adds the input and the output, and stores it to the output. I'm not sure what verification checks MLIR has, but I don't think that it actually handles variadic inputs and outputs. In any case for this case, I'd recommend matching my proposed change.

kayode-gif: ahhh okay, that makes sense, thanks, wasn't too sure on how adding would work, it should be fine now

tests/filecheck/transforms/convert_onnx_to_linalg.mlir
```diff
@@ -74,6 +74,13 @@
 // CHECK-NEXT:  %res_max_pool_single_out_3 = linalg.fill ins(%res_max_pool_single_out_2 : f64) outs(%res_max_pool_single_out_1 : tensor<1x16x4x4xf32>) -> tensor<1x16x4x4xf32>
 // CHECK-NEXT:  %res_max_pool_single_out_4 = linalg.pooling_nchw_max {"dilations" = dense<1> : tensor<2xi64>, "strides" = dense<3> : tensor<2xi64>} ins(%t26, %res_max_pool_single_out : tensor<1x16x14x14xf32>, tensor<3x3xf32>) outs(%res_max_pool_single_out_3 : tensor<1x16x4x4xf32>) -> tensor<1x16x4x4xf32>
 
+%t20, %t21, %t22 = "test.op"() : () -> (tensor<1x1x5x5xf32>, tensor<1x1x3x3xf32>, none)
```
superlopuh, line 77: why none?

superlopuh: Ah, I see! It seems like we should check in the lowering whether the operand is `none` type or not, and not add the linalg.add if it's none. So ideally two checks one with none and one with a real shape

kayode-gif: erm i used a test cases for that, but we should probably check none and add if not now i think

kayode-gif: coool

superlopuh: Code looks great, could you please add a test case for it? Then we should be GTG

kayode-gif: done :)

compor (approved):

PR #2401, opened 2024-03-28
docs: a getting started guide for new contributors

superlopuh:


tobiasgrosser:
This is great. Do you typically have people read the toy tutorials before getting started? Or can they dive right in?

tobiasgrosser (approved):

GETTING_STARTED.md
```diff
@@ -0,0 +1,71 @@
...
+[#1566](https://github.com/xdslproject/xdsl/pull/1566),
+[#1567](https://github.com/xdslproject/xdsl/pull/1567).
+
+The first step is finding a missing optimisation pattern.
+You're welcome to come up with your own, or do one of the following:
+
+- `x * 2^k -> x << k`
+- `x & 0 -> 0`
+- `x | 0 -> x`
+- `x ^ 0 -> x`
```
Moxinilian, line 50: Does this risk to not age well? How do we come up with new work once this is completed?

Moxinilian (approved):
Love this!

GETTING_STARTED.md
compor: I mostly agree with this, but the way I see it is to have something like a "Hello, world!" in xDSL.
There is no novelty in "Hello, world!"  apart from serving as a "kicking the tires" activity.

This might be good for now, but we might need to revise to have a very simple "toy"-like tutorial (e.g., a `foo` dialect with 2 arith-like ops and a simple transform with one of them).

compor (approved):

GETTING_STARTED.md
Moxinilian: Maybe. It's also interesting that the "hello world" example is actually useful, I'm sure it makes things more exciting. But generating an infinite stream of simple, useful work is difficult.

superlopuh: I'd say that it would be worth the effort. The good news is that we're already adding x86, and I'll be surprised if we never add wasm or ARM dialects, so we're unlikely to run out of trivial optimisations to add.

superlopuh:
> This is great. Do you typically have people read the toy tutorials before getting started? Or can they dive right in?

I usually do. We don't have any better slow introduction that really tries to explain the API rather than just give examples. We'll probably want to restructure it at some point. I have a feeling that the first thing I did with xDSL, translating sympy to arith, might be the best place to start for most people, and would avoid the Toy overhead.

PR #2404, opened 2024-03-31
dialects: (seq) Add CompRegOp

Moxinilian:
This op allows modeling registers. This is for now a very basic implementation which only partially matches the CompRegOp upstream. But it should be enough for many practical uses already!

### DIFF
--- a/tests/filecheck/dialects/seq/seq_ops.mlir
+++ b/tests/filecheck/dialects/seq/seq_ops.mlir
@@ -2,6 +2,13 @@
 
 builtin.module {
   %clk = "test.op"() : () -> (!seq.clock)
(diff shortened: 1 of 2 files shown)


Moxinilian:
> I think maybe you could add the optional `$powerOnValue` operand? Otherwise LGTM 🚀

It's a bit silly but adding the power on value triggers a bug in the assembly format thing that is a bit over my head to attempt and fix. Perhaps we can do without for now?

lucjaulmes (approved):

lucjaulmes:
Sure. Maybe file a bug about the double-optionals? We’ll have to revisit compregop anyway.

Moxinilian:
#2413

PR #2407, opened 2024-04-01
core: Add optional attributes to declarative assembly format

lucjaulmes:
I was going to suggest the missing optional `$inner_sym` attr on #2404 but realised optional attributes don’t exist yet in the declarative format. It seemed a rather trivial extension so here’s my shot at it, let me know if something misses/feels wrong.


lucjaulmes:
I’ve added optional property too to fix coverage issues and for completeness.

superlopuh (approved):

PR #2411, opened 2024-04-02
transformations: Add convert-snrt-to-riscv-asm pass to lower snrt ops to RISC-V dialect

AntonLydike:
Add `convert-snrt-to-riscv-asm` transformation that converts (some) snitch runtime ops into their assembly definitions.

This is part one of many, so more patterns coming up soon!


JosseVanDelm (approved):
Looks okay, wondering whether it makes a lot of difference to emit llvm.inlineAsmOps directly here though?

AntonLydike:
> Looks okay, wondering whether it makes a lot of difference to emit llvm.inlineAsmOps directly here though?

My plan is to have a more generic pass that converts riscv dialect to inline assembly ops, that would be most useful for snax-opt I guess

PR #2412, opened 2024-04-02
frontend: (onnx) implementation tests frontend onnx

alecerio:
PR about the tests of the frontend of onnx about the context.

The following tests were implemented:

- test_visit_node_unknown_op_name
- test_visit_node_add
- test_visit_node_sub
- test_visit_graph_add
- test_visit_graph_sub
- test_build_module
- test_visit_value_info

@superlopuh, @compor


superlopuh:
Looks great! It's a little difficult to review, since it looks like it contains changes from #2334. Could you please address the comments on that PR first? That should reduce the diff here.

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,76 @@
...
+from xdsl.dialects.builtin import ModuleOp
+from xdsl.frontend.onnx.type import get_type
+from xdsl.ir import Attribute, SSAValue
+from xdsl.irdl import IRDLOperation
+
+OP_BY_OP_TYPE: dict[str, type[IRDLOperation]] = {
+    "Add": onnx.Add,
+    "Sub": onnx.Sub,
+}
+"Dictionary that associates the name of the operations with the respective operation in ONNX dialect."
```
superlopuh, line 14: I'd recommend skipping `Dictionary` here, and `class` below. The client already knows that this is a dictionary and that `Ctx` is a class. Can you please rephrase the comments to just say what the thing does, and not what kind of thing it is?

Also, could you please use triple quotes for doc comments?

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class Ctx:
+    "Class to represent the Context."
+    type_by_name: dict[str, Attribute]
+    value_by_name: dict[str, SSAValue]
+
+    def __init__(self):
+        "Constructor to initialize the Context class."
```
superlopuh, line 23: Doc comments are visible when hovering over code, I'm not sure if there's a way to see this comment since it's a kind of anonymous function. I'd also say that it's obvious enough what it does to not need documentgation at all.

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,76 @@
...
+from xdsl.irdl import IRDLOperation
+
+OP_BY_OP_TYPE: dict[str, type[IRDLOperation]] = {
+    "Add": onnx.Add,
+    "Sub": onnx.Sub,
+}
+"Dictionary that associates the name of the operations with the respective operation in ONNX dialect."
+
+
+class Ctx:
```
superlopuh, line 17: Can you please give this a better name? Ctx was useful for prorotyping but is not really readable.

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,76 @@
...
+    "Class to represent the Context."
```
superlopuh, line 18: This isn't really clear to me

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,76 @@
...
+    type_by_name: dict[str, Attribute]
+    value_by_name: dict[str, SSAValue]
+
+    def __init__(self):
+        "Constructor to initialize the Context class."
+        self.type_by_name = {}
+        self.value_by_name = {}
+
+
+def visit_node(node: NodeProto, ctx: Ctx) -> None:
```
superlopuh, line 28: Can you please add documentation for how this function and the functions below work?

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,193 @@
...
+    # check keys in context
+    keys = list(ctx.type_by_name.keys())
+    assert keys == ["input_tensor"]
+
+    # check type info
+    type_info = str(ctx.type_by_name["input_tensor"])
+    assert type_info == "tensor<1x3x224x224xf32>"
+
+
+def _create_graph_binary_op(op_name: str, graph_name: str):
```
superlopuh, line 167: Very nice

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # define expected output
+    expected = (
+        "builtin.module {\n"
+        + "  func.func @add_graph(%0 : tensor<0x0xf32>, %1 : tensor<0x0xf32>) -> tensor<0x0xf32> {\n"
+        + "    %2 = onnx.Add(%0, %1) : (tensor<0x0xf32>, tensor<0x0xf32>) -> tensor<0x0xf32>\n"
+        + "    func.return %2 : tensor<0x0xf32>\n"
+        + "  }\n"
+        + "}"
+    )
```
superlopuh, line 128: This might be a little easier to read as a multiline string, like this:
```suggestion
    expected = """\
builtin.module {
  func.func @add_graph(%0 : tensor<0x0xf32>, %1 : tensor<0x0xf32>) -> tensor<0x0xf32> {
    %2 = onnx.Add(%0, %1) : (tensor<0x0xf32>, tensor<0x0xf32>) -> tensor<0x0xf32>
    func.return %2 : tensor<0x0xf32>
  }
}   
"""
```

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,191 @@
...
+        ValueInfoProto,
+        helper,
+    )
+
+    from xdsl.frontend.onnx.context import (
+        OnnxXdslMapping,
+        build_module,  # noqa: E402
+        visit_graph,  # noqa: E402
+        visit_node,  # noqa: E402
+        visit_value_info,  # noqa: E402
```
superlopuh, line 16: can you please remove noqa from these lines?

pyproject.toml
```diff
@@ -59,7 +61,7 @@ typeCheckingMode = "strict"
     "tests/filecheck/frontend/dialects/arith.py",
     "tests/filecheck/frontend/dialects/scf.py",
     "tests/filecheck/frontend/programs/invalid.py",
-    "xdsl/_version.py",
+    "xdsl/_version.py"
```
superlopuh, line 64: this seems like a spurious change, can you please undo it?

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,191 @@
...
+    # visit graph
+    visit_graph(graph, ctx)
+
+    # check value_by_name keys
+    keys = list(ctx.value_by_name.keys())
+    assert keys == ["input1", "input2", "output"]
+
+    # check expected generated ir
+    gen_ir = ctx.value_by_name[keys[2]].owner
+    print(gen_ir)
```
superlopuh, line 84: can you please remove this print?

tests/frontend/onnx/test_context.py
alecerio: Done

tests/frontend/onnx/test_context.py
alecerio: Done

pyproject.toml
superlopuh: thank you

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,189 @@
...
+
+def test_build_module():
+    # create graph composed only of one Add operation
+    graph, _ = _create_graph_binary_op("Add", "add_graph")
+
+    # create module
+    module = build_module(graph)
+
+    # define expected output
+    expected = """builtin.module {
```
superlopuh, line 120: nit:
```suggestion
    expected = """\
builtin.module {
```
It would be nice to have the `builtin.module` be on a new line, do you think you could try this change? I may be remembering the syntax incorrectly, maybe it can be done better by adding the white space we might want at the beginning and end and adding a `.strip()` to remove the white space at the end.

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,189 @@
...
+    ctx = OnnxXdslMapping()
+
+    # create graph composed only of one Sub operation
+    graph, sub_node = _create_graph_binary_op("Sub", "sub_graph")
+
+    # visit graph
+    visit_graph(graph, ctx)
+
+    # visit node (test passes if no exceptions are raised)
+    visit_node(sub_node, ctx)
```
superlopuh, line 65: I might be missing something but it seems like the effect of this is not tested. Am I understanding this correctly?

xdsl/frontend/onnx/context.py
alecerio: Done

tests/frontend/onnx/test_context.py
alecerio: Done

compor (approved):
LGTM module the `visit_node` and co methods

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def visit_graph(g: GraphProto, ctx: OnnxXdslMapping) -> None:
+    """Visit the onnx graph to update the onnx context."""
+    name = g.name
+
+    input_types = tuple(visit_value_info(input, ctx) for input in g.input)
+    output_types = tuple(visit_value_info(output, ctx) for output in g.output)
+
+    fn = func.FuncOp(name, (input_types, output_types))
```
superlopuh, line 62: I forgot about this, but could you please change this function to explicitly build a module instead of implicitly? The way it currently works is a little hard to explain and reason about, as all operations created within this function call are inserted into the module. It would be better for `visit_graph` to return the `func.func` directly, and for `visit_node` to take a builder and insert the node if it's not already in `ctx`

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,80 @@
...
+
+    type_by_name: dict[str, Attribute]
+    value_by_name: dict[str, SSAValue]
+
+    def __init__(self):
+        self.type_by_name = {}
+        self.value_by_name = {}
+
+
+def visit_node(node: NodeProto, ctx: OnnxXdslMapping) -> None:
```
superlopuh, line 28: can you please add a check that if we visit a node twice, it returns the existing xDSL operation in the ctx, and does not overwrite it with a new one?

superlopuh: Similearly for `visit_value_info`

tests/frontend/onnx/test_context.py
alecerio: Done

superlopuh (commented):
Am I right in my understanding that this was superceded by a different PR? Let's close this one if that is the case?

PR #2415, opened 2024-04-03
docs: onboarding papercuts

knickish:
Couple issues with getting set up that I encountered, broken links and such.

PR #2419, opened 2024-04-03
misc: (Makefile) Remove trailing slash from venv target

AntonLydike:
Fixes a stray slash introduced by #2316 that changed the venv target name form `venv` to `venv/`, as noticed in #2415.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,14 +14,17 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 .ONESHELL:
 
 # these targets don't produce files:
-.PHONY: ${VENV_DIR} clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
+.PHONY: ${VENV_DIR}/ venv clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
 .PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report-md
 
 # set up the venv with all dependencies for development
-${VENV_DIR}: requirements.txt
+${VENV_DIR}/: requirements.txt
 	python3 -m venv ${VENV_DIR}
 	. ${VENV_DIR}/bin/activate
 	python3 -m pip --require-virtualenv install -r requirements.txt
+
+# make sure `make venv` always works no matter what $VENV_DIR is
+venv: ${VENV_DIR}/
 
 # remove all caches and the venv
 clean:
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,11 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 .ONESHELL:
 
 # these targets don't produce files:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 .PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report-md
 
 # set up the venv with all dependencies for development
-${VENV_DIR}/: requirements.txt
+${VENV_DIR}: requirements.txt
```
superlopuh, line 21: Seems like we want to update the .PHONY above also?

superlopuh (approved):

compor (approved):

lucjaulmes:
Trailing slash is intentional and rationale is in linked PR #2316: 
> The trailing slash on the target name means you can use a hidden directory (e.g., `make VENV_DIR=.venv`) without the target being confused for a special Makefile target (.c.o, .PHONY, etc).

See for example:

```
> cat Makefile 
.rule-no-trailing:
    @echo rule $@

.rule-with-trailing/:
    @echo rule $@

> make
rule .rule-with-trailing/
```

You can revert if you want to explicitly disallow this use case. Or we can add a comment if it seems weird.

superlopuh:
No, it seems like an important thing to support, a bunch of python tools assume `.venv` as the venv name by default.

superlopuh:
It seems like maybe we want to keep `make venv` (where venv is a phony rule` and for it to create a venv at `VENV_DIR`?

lucjaulmes:
I wasn’t aware `make venv` was a documented use case, I always saw just `make` being used instead – would have updated otherwise.

I think this all stems from the ambiguity from `make venv`, is it “make the virtual environment directory” (whatever the value of `$VENV_DIR`), or is it “make the local target `venv`” (only makes sense when `VENV_DIR=venv`). In other words, the phony target happens to match the default value of `$VENV_DIR`

I see 2 solutions ot keep the `make venv` as a phony rule:
- revert #2316 (maybe adding a comment that `venv` isn’t `$VENV_DIR`).
- prepend a first PHONY rule that is `venv: ${VENV_DIR}/` 

Just dropping the `/` doesn’t seem ideal if `make venv` is the documented way to build the virtual environment.

AntonLydike:
I think I'm in favor of having a new phony target `venv: ${VENV_DIR}/`, as (at least for me), I'd like a canonical way to create the virtual environment no matter where it is actually located.

superlopuh:
OTOH I just use `uv venv venv` these days, and I'm not sure how many of the contributors use the existing make. It might be worth considering updating the documentation instead, to make sure that new contributors don't get tripped up by broken documentation.

AntonLydike:
It's worthwhile to note that `make venv` will also install all dependencies, which is a really nice benefit

AntonLydike:
Updated PR, made `venv` always a target, and added the slash back to `${VENV_DIR}`, hope this makes everyone happy!"

PR #2422, opened 2024-04-03
transformations: (convert-snrt-to-riscv) Add lowering for 2d dma calls

AntonLydike:
This pass adds the 2d and 2d_wideptr dma call translations to RISC-V.


xdsl/transforms/convert_snrt_to_riscv.py
```diff
@@ -198,8 +201,95 @@ def match_and_rewrite(
         )
 
 
+class LowerDMAStart2DBase(RewritePattern, ABC):
+    any_reg = riscv.IntRegisterType.unallocated()
+
+    def gen_insns(
```
superlopuh, line 207:
```suggestion
    def gen_instructions(
```

superlopuh (approved):

xdsl/dialects/snitch_runtime.py
```diff
@@ -334,10 +334,10 @@ class DmaStart2DBaseOperation(SnitchRuntimeBaseOperation, Generic[_T], ABC):
     T = Annotated[Attribute, ConstraintVar("T"), _T]
     dst: Operand = operand_def(_T)
     src: Operand = operand_def(_T)
-    dst_stride: Operand = operand_def(IndexType)
```
cappadokes, line 337: Note to future selves: there may be other places where `IndexType` should be converted to `i32`/`i64`/etc.

xdsl/transforms/convert_snrt_to_riscv.py
```diff
@@ -198,8 +201,95 @@ def match_and_rewrite(
         )
 
 
+class LowerDMAStart2DBase(RewritePattern, ABC):
+    any_reg = riscv.IntRegisterType.unallocated()
```
cappadokes, line 205: Comment about which function is being lowered should be inserted here.

xdsl/transforms/convert_snrt_to_riscv.py
```diff
@@ -198,8 +201,95 @@ def match_and_rewrite(
...
+            riscv_snitch.DMStrideOp(src_stride, dst_stride),
+            riscv_snitch.DMRepOp(repeat),
+            cpy_op := riscv_snitch.DMCopyImmOp(size, 0b10),
+            tx_id := builtin.UnrealizedConversionCastOp.get([cpy_op], [builtin.i32]),
+        ], tx_id.results
+
+    def cast_i32(self, input_val: SSAValue):
+        return builtin.UnrealizedConversionCastOp.get([input_val], [self.any_reg])
+
+    def cast_i64(self, input_val: SSAValue):
```
cappadokes, line 235: Properly casting 64-bit integers to 32-bit RISC-V registers is not implemented right now. Future work!

cappadokes (approved):
This review is more decorative than substantial. Approving by default.

tests/filecheck/backend/riscv/func_and_arith_to_riscv_asm_flow.mlir
```diff
@@ -0,0 +1,31 @@
...
+// CHECK-NEXT:     mv t2, a0
+// CHECK-NEXT:     mv t3, a1
+// CHECK-NEXT:     li t1, 128
+// CHECK-NEXT:     li t0, 512
+// CHECK-NEXT:     li t4, 64
+// CHECK-NEXT:     dmsrc t3, zero
+// CHECK-NEXT:     dmdst t2, zero
+// CHECK-NEXT:     dmstr t1, t1
+// CHECK-NEXT:     dmrep t0
+// CHECK-NEXT:     dmcpyi t0, t0, 2
```
JosseVanDelm, line 29: Do these compile with a standard RISC-V compiler?

JosseVanDelm: Can we print these as `.word` s or `.insn` instead?

JosseVanDelm (commented):
Is there a plan to use acc2 dialect instead of converting the calls to multiple csr-like instructions?

xdsl/dialects/snitch_runtime.py
superlopuh: I would say that we might want to constrain it to the bitwidth of the snitch instructions as early as possible (in the dialects that are snitch-specific)

tests/filecheck/backend/riscv/func_and_arith_to_riscv_asm_flow.mlir
superlopuh: I was under the impression that we want to do that separately, to have a way to print insns for certain instructions as an option?

AntonLydike: Yes, I have some big plans for the risc-v backend right now:
- Better infra to handle target printing (e.g. printing `.insn ...` things for instructions unsupported by our backend)
- A pass `convert-riscv-to-llvm` which wraps riscv instructions inside llvm inline assembly ops. For this we'll need to infer sideffects, which will be fun :D

AntonLydike:
> Is there a plan to use acc2 dialect instead of converting the calls to multiple csr-like instructions?

We could have another lowering that is basically `convert-snrt-to-acc2`, but I'm not sure if the dma engines setup is persistent across calls. Let me ask Federico!

PR #2427, opened 2024-04-04
frontend: (onnx) onnx graph exploration implementation

alecerio:
Changes with this PR:
- fix graph exploration algorithm to generate ir from onnx.
- remove implicit builder to compose the module.


superlopuh (commented):
Am I right in thinking that we don't need these changes any more? Is there anything we want to keep from this?

PR #2429, opened 2024-04-04
dialects: (seq) Add builder and verifier for CompRegOp

Moxinilian:
This PR adds a verifier and a builder to CompRegOp.

### DIFF
--- a/tests/dialects/test_seq.py
+++ b/tests/dialects/test_seq.py
@@ -2,10 +2,13 @@ import pytest
 
 from xdsl.dialects.builtin import (
     IntegerAttr,
+    IntegerType,
+    i1,
     i32,
 )
 from xdsl.dialects.seq import (
     ClockDivider,
+    CompRegOp,
     clock,
 )
 from xdsl.utils.exceptions import VerifyException
@@ -22,3 +25,15 @@ def test_clockdivider_verify():
         match="Operation does not verify: pow2 has to be an 8-bit signless integer",
     ):
         clock_div.verify()
+
+
+def test_compreg_builder():
+    data_val = TestSSAValue(IntegerType(5))
+    bool_val = TestSSAValue(i1)
+    clock_val = TestSSAValue(clock)
+
+    CompRegOp(data_val, clock_val).verify()
+
+    with_reset = CompRegOp(data_val, clock_val, (bool_val, data_val))
+    with_reset.verify()
+    assert with_reset.reset is not None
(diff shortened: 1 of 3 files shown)


tests/dialects/test_seq.py
```diff
@@ -22,3 +25,15 @@ def test_clockdivider_verify():
...
+def test_compreg_builder():
+    data_val = TestSSAValue(IntegerType(5))
+    bool_val = TestSSAValue(i1)
+    clock_val = TestSSAValue(clock)
+
+    CompRegOp(data_val, clock_val).verify()
+
+    with_reset = CompRegOp(data_val, clock_val, (bool_val, data_val))
+    with_reset.verify()
+    assert with_reset.reset is not None
```
AntonLydike, line 39: This test case would benefit from a short explanation what it is actually testing

xdsl/dialects/seq.py
```diff
@@ -106,6 +106,28 @@ class CompRegOp(IRDLOperation):
...
+        self,
+        input: SSAValue,
+        clk: SSAValue,
+        reset: tuple[SSAValue, SSAValue] | None = None,
+    ):
+        super().__init__(
+            operands=[
+                input,
+                clk,
+                reset[0] if reset is not None else None,
```
AntonLydike, line 119: you can use `reset.get(0)` here to avoid this spaghetti code

xdsl/dialects/seq.py
```diff
@@ -106,6 +106,28 @@ class CompRegOp(IRDLOperation):
...
+                reset[1] if reset is not None else None,
+            ],
+            result_types=[input.type],
+        )
+
+    def verify_(self):
+        if (self.reset is not None and self.reset_value is None) or (
+            self.reset_value is not None and self.reset is None
+        ):
+            raise VerifyException("both reset and reset_value must be set when one is")
```
AntonLydike, line 129: Our verifier messages usually start with uppercase letters afaik.

AntonLydike (changes requested):
Thanks for this contribution, I just added a few comments on minor things I'd really like to see improved before merging.

Also, how do magnets work? :thinking:

xdsl/dialects/seq.py
lucjaulmes: It seems to be handling `reset = None` not `reset` being a dict without `0`. You could shorten to `(reset or {0: None})[0]` but frankly that doesn’t seem as readable.

AntonLydike (approved):
LGTM :chart_with_upwards_trend:

PR #2430, opened 2024-04-05
dialects: Add accfg dialect operations

JosseVanDelm:
This PR adds the accfg dialect that @AntonLydike and me designed during a very productive week in Leuven

Accfg dialect contains ops to minimize setup in register-set-up accelerators.
Either through registers connected to a CPU core (like control and status registers) or memory mapped io (like setting registers over an AXI bus).

All operations are documented with docstrings.

Thanks for your reviews!

### DIFF
--- a/tests/filecheck/dialects/accfg/accfg_ops.mlir
+++ b/tests/filecheck/dialects/accfg/accfg_ops.mlir
@@ -1,0 +1,43 @@
+// RUN: XDSL_ROUNDTRIP
+
+"accfg.accelerator"() <{
+    name               = @acc1,
+    fields             = {A=0x3c0, B=0x3c1},
+    launch_addr        = 0x3cf,
+    barrier            = 0x7c3
+}> : () -> ()
+
+func.func @test() {
+    %one, %two = "test.op"() : () -> (i32, i32)
+
+    %state = "accfg.setup"(%one, %two) <{
+        param_names = ["A", "B"],
+        accelerator = "acc1",
+        operandSegmentSizes = array<i32: 2, 0>
+    }> : (i32, i32) -> !accfg.state<"acc1">
+
+    %token = "accfg.launch"(%state) <{accelerator = "acc1"}>: (!accfg.state<"acc1">) -> !accfg.token<"acc1">
+
+    %state2 = "accfg.setup"(%one, %two, %state) <{
+        param_names = ["A", "B"],
+        accelerator = "acc1",
+        operandSegmentSizes = array<i32: 2, 1>
+    }> : (i32, i32, !accfg.state<"acc1">) -> !accfg.state<"acc1">
+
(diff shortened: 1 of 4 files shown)


JosseVanDelm:
It appears I've been a bit sloppy in making this PR! Apologies! python tests and pyright fixes coming up!

AntonLydike (approved):
Very nice work! Thanks!

tests/filecheck/dialects/accfg/accfg_ops.mlir
```diff
@@ -0,0 +1,43 @@
+// RUN: XDSL_ROUNDTRIP
```
superlopuh, line 1: Do we want a generic roundtrip also?

xdsl/dialects/accfg.py
```diff
@@ -0,0 +1,310 @@
...
+
+    traits = frozenset([AcceleratorSymbolOpTrait()])
+
+    name_prop = prop_def(SymbolRefAttr, prop_name="name")
+
+    fields = prop_def(DictionaryAttr)
+
+    launch_addr = prop_def(IntegerAttr)
+
+    barrier = prop_def(IntegerAttr)  # TODO: this will be reworked in a later version
```
superlopuh, line 102: What do you mean by reworked?

xdsl/dialects/accfg.py
```diff
@@ -0,0 +1,310 @@
...
+            operands=[vals, in_state],
+            properties={
+                "param_names": ArrayAttr(param_names_tuple),
+                "accelerator": accelerator,
+            },
+            result_types=[StateType(accelerator)],
+        )
+
+    def iter_params(self) -> Iterable[tuple[str, SSAValue]]:
+        return zip((p.data for p in self.param_names), self.values)
```
superlopuh, line 277: If we expect the values to have the same length always, then it seems like a good idea to add `strict=True` here.

superlopuh (approved):

xdsl/dialects/accfg.py
AntonLydike: The exact translation of `accfg.await` is not final, and as such the information required for this on the `accfg.accelerator` op will probably change again in the future. We're looking into ways of generalizing this aspect currently, but this is a thing that actually works for snitch now.

superlopuh: it might be worth adding this to the comment

tests/filecheck/dialects/accfg/accfg_ops.mlir
JosseVanDelm: Sure thing

xdsl/dialects/accfg.py
JosseVanDelm: Will do!

PR #2431, opened 2024-04-05
misc: (Makefile) Fix coverage reports to depend on coverage generation

webmiche:
Making the coverage reporting targets depend on the coverage generating ones.

The big disadvantage of this is that now, you will always run the test suite when wanting to report the coverage. I am open for suggestions on how to circumvent that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,9 @@ black:
 coverage: coverage-tests coverage-filecheck-tests
 	coverage combine --append
 
+conditional-coverage:
+	if ! [[ -e .coverage ]]; then make coverage; fi
+
 # run coverage over tests
 coverage-tests:
 	COVERAGE_FILE=${TESTS_COVERAGE_FILE} pytest -W error --cov --cov-config=.coveragerc
@@ -87,9 +90,9 @@ coverage-filecheck-tests:
 	lit -v tests/filecheck/ -DCOVERAGE
 
 # generate html coverage report
-coverage-report-html: coverage
+coverage-report-html: conditional-coverage
 	coverage html
 
 # generate markdown coverage report
-coverage-report-md: coverage
+coverage-report-md: conditional-coverage
 	coverage report --format=markdown
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 
 # these targets don't produce files:
 .PHONY: ${VENV_DIR}/ clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
+.PHONY: coverage coverage-pytest-tests coverage-filecheck-tests coverage-report-html coverage-report-md
 
 # set up the venv with all dependencies for development
 ${VENV_DIR}/: requirements.txt
@@ -77,18 +78,27 @@ black:
 .coverage: .coverage_pytest .coverage_filecheck
 	coverage combine --keep .coverage_pytest .coverage_filecheck
 
-# run coverage over tests
 .coverage_pytest:
+	make coverage-pytest-tests
+
+.coverage_filecheck:
+	make coverage-filecheck-tests
+
+coverage: coverage-pytest-tests coverage-filecheck-tests
+	coverage combine --keep .coverage_pytest .coverage_filecheck
+
+# run coverage over tests
+coverage-pytest-tests:
 	COVERAGE_FILE=${TESTS_COVERAGE_FILE} pytest -W error --cov --cov-config=.coveragerc
 	coverage combine --append --data-file=.coverage_pytest ${TESTS_COVERAGE_FILE}
 
 # run coverage over filecheck tests
-.coverage_filecheck:
+coverage-filecheck-tests:
 	lit -v tests/filecheck/ -DCOVERAGE
 	coverage combine --append --data-file=.coverage_filecheck .coverage.*
 
 # generate html coverage report
-htmlcov/index.html: .coverage
+coverage-report-html: .coverage
 	coverage html
 
 # generate markdown coverage report
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 
 # these targets don't produce files:
 .PHONY: ${VENV_DIR}/ clean filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright
-.PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report-md
 
 # set up the venv with all dependencies for development
 ${VENV_DIR}/: requirements.txt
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -87,9 +87,9 @@ coverage-filecheck-tests:
 	lit -v tests/filecheck/ -DCOVERAGE
 
 # generate html coverage report
-coverage-report-html:
+coverage-report-html: coverage
```
JosseVanDelm, line 90: Can't you let it depend on the output file of the coverage command to circumvent your problem?

You can add a `--data-file=DATAFILE` to the `coverage` recipe to explicitely control the output

webmiche: What is the syntax to depend on a file? I have honestly not a lot of clue about make files.
The file is guaranteed to be `.coverage`, AFAIK.

webmiche: I guess I could do this using some shell cmds, but I would prefer to fix it in a makefile-way.

webmiche:
> The big disadvantage of this is that now, you will always run the test suite when wanting to report the coverage. I am open for suggestions on how to circumvent that.

I solved this using some bash magic now. I feel like this is not how makefiles are intended to be used, if someone can tell me the proper syntax, I am happy to adjust this.

JosseVanDelm:
@compor can probably give the ultimate response here, but normally in makefiles the names of your recipes are the actual files you build/output (e.g. your html) , and the dependencies refer to actual files or previous recipes (which at some point refer to actual files) (e.g. your .coverage file).
Make will rebuild recipes only for which the dependencies are younger (according to unix timestamp) than the targets.
So in this case you would add:

```make
htmlcov/index.html: .coverage: 
    coverage html
``` 
Your `coverage-report-html` recipe does not actually generate a file that is called `coverage-report-html` so it should be marked as `.PHONY`
You can still use it as a convenient alias in this way by marking the file generated by the actual recipe as a dependency:
```make
.PHONY: coverage-report-html
 
coverage-report-html: htmlcov/index.html
```

webmiche:
Okay, this is the new iteration now. I removed all the Phony stuff and this now works, AFAICT.
I think I know how to add the Phony stuff back, if you guys still want it. I think it's quite useful to have the target match the name, and, IMO, the file names are descriptive enough. But, if someone really wants their old targets back, I can add them back.

lucjaulmes:
I think it’s a step in the right direction, but not sure how you now re-run coverage? If your `.coverage` files exists, it won’t re-run, as there are never newer dependencies.

I think if you just change the `.coverage` target back to a phony `coverage`, but still depend on `.coverage` for the reports you have the best of both worlds:
- reports are only re-generated for new coverages (when `.coverage` is updated)
- `make coverage` always runs coverage again
- `make coverage-report-md` without `make coverage` before it fails
- reports never trigger a full coverage re-run

webmiche:
I gave it another shot, not sure this is what you meant :sweat_smile:

Makefile
```diff
@@ -77,18 +78,27 @@ black:
 .coverage: .coverage_pytest .coverage_filecheck
 	coverage combine --keep .coverage_pytest .coverage_filecheck
 
-# run coverage over tests
 .coverage_pytest:
+	make coverage-pytest-tests
```
superlopuh, line 82:
```suggestion
.coverage_pytest: coverage-pytest-tests
	
```
Feels weird to call make in make

webmiche: I agree, but this does not behave as I think it should. If I know do `make .coverage_pytest` with the .coverage_pytest file still around, it rebuilds.

superlopuh: oh sorry it should be the other way round, the target should depend on the non-phony target

webmiche: The other way around makes it s.t. `make coverage` does not re-run the pytests for coverage. Because `.coverage_pytest` already exists.

superlopuh: that makes sense. then maybe something like this:
```
.non-phony: SAME_AS
    BEFORE

phony:
    make .non-phony
```

webmiche: Maybe I am confused now, but this does the same thing as above, i.e. does not run pycheck tests for `make coverage`

lucjaulmes: It does seem a bit weird. You would usually `$(MAKE)` to recurse in Makefiles, but that’s mainly to use another Makefile than the one you currently use. Would this work?


```
# run coverage over all tests and combine data files
coverage: coverage-tests coverage-filecheck ${COVERAGE_FILE}

# merge coverage files
${COVERAGE_FILE}: ${TESTS_COVERAGE_FILE} ${FILECHECK_COVERAGE_FILE}
	coverage combine --keep --data-file=$@ ${TESTS_COVERAGE_FILE} ${FILECHECK_COVERAGE_FILE}

# run coverage over tests
coverage-tests:
	COVERAGE_FILE=${TESTS_COVERAGE_FILE} pytest -W error --cov --cov-config=.coveragerc

# run coverage over filecheck tests
coverage-filecheck:
	lit -v tests/filecheck/ -DCOVERAGE
	coverage combine --data-file=${FILECHECK_COVERAGE_FILE} .coverage.*

.PHONY: coverage coverage-tests coverage-filecheck
```

lucjaulmes: So making `coverage-tests` and `coverage-filecheck` re-run their part of tests, making `coverage` runs both + the merge. Anything depending on `${COVERAGE_FILE}` will re-run the merge if needed, but not the actual tests.

webmiche: I tried to implement what you were suggesting. Note that the initial thing failed with `make: *** No rule to make target '.coverage.tests', needed by '.coverage'.  Stop.`. I added some more rules to fix this issue, but the behavior now once again is that calling `make coverage-report-md` runs the coverage target.

webmiche: Feel free to push to this branch yourself, if you know how to make this work :)

lucjaulmes:
I think this works @webmiche, give it a try. I used a `define` but we could also repeat the rule or `$(MAKE) ${COVERAGE_FILE}` from the phony `coverage` rule – whichever seems most readable to people.

Also I had to change the name of `.coverage.tests` to not match `.coverage.*` as it was getting merged into `.coverage.filecheck` and deleted.

webmiche:
I see, thanks for the input! This works if you first run `make coverage` and then `make coverage-report-md`. However, if you are only running `make-coverage-report-md` without running `make coverage` before, it fails with `make: *** No rule to make target '.coveragepart.tests', needed by '.coverage'.  Stop.`.

Let me try to apply your approach to the generation of the two subfiles as well, that should fix the issue...

lucjaulmes:
> This works if you first run `make coverage` and then `make coverage-report-md`. However, if you are only running `make-coverage-report-md` without running `make coverage` before, it fails

Sorry I thought this was the behaviour you intended? If not, the semantics of `make coverage-report-md` are a bit unclear to me.

Basically there’s a risk of repeatedly making a report from a stale coverage. To me, `make coverage-report-md` should either always re-build the coverage, or only work when there is a coverage present. Having (and documenting) `make coverage; make coverage-report-md` as a workflow would make sense.

But I suppose it’s more personal preference than anything else at this point. Just document whatever you end up picking.

webmiche:
Ah, I see. I guess I should have made that more clear then.

I am now honestly not even sure anymore if anything has to change. The behavior you are suggesting makes sense and it's pretty much what already happens, minus the intermediate files.

Still, thanks a lot for you input, I learned a lot about makefiles in this, even if this won't make it into main.

PR #2437, opened 2024-04-08
frontend: [onnx] add test_visit_value_info_multiple_time

alecerio:
Implementation of visit_value_info and related tests.

@superlopuh @compor

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,12 +45,14 @@ enableTypeIgnoreComments = false
 typeCheckingMode = "strict"
 "include" = ["docs", "xdsl", "tests", "bench"]
 "exclude" = [
+    "tests/dialects/test_memref.py",
+    "tests/frontend/onnx/test_context.py",
+    "tests/frontend/onnx/test_type.py",
     "tests/test_frontend_op_resolver.py",
     "tests/test_frontend_python_code_check.py",
-    "tests/frontend/onnx/test_type.py",
-    "tests/dialects/test_memref.py",
-    "xdsl/interpreters/experimental/wgpu.py",
+    "xdsl/frontend/onnx/context.py",
     "xdsl/frontend/onnx/type.py",
+    "xdsl/interpreters/experimental/wgpu.py",
     "xdsl/interpreters/onnx.py",
 ]
 "ignore" = [
--- a/tests/frontend/onnx/test_context.py
+++ b/tests/frontend/onnx/test_context.py
@@ -1,0 +1,82 @@
+import onnx
+import pytest
+
+try:
+    from onnx import TensorProto, ValueInfoProto
+
+    from xdsl.frontend.onnx.context import (
+        OnnxXdslMapping,
+        visit_value_info,
+    )
+except ImportError as exc:
+    print(exc)
+    pytest.skip("onnx is an optional dependency", allow_module_level=True)
+
+
+def test_visit_value_info():
+    # initialize context
+    ctx = OnnxXdslMapping()
+
+    # create ValueInfoProto input tensor
+    input_value_info = ValueInfoProto()
+    input_value_info.name = "input_tensor"
+
+    # define the type of the input tensor
+    input_tensor_type = input_value_info.type.tensor_type
+    input_tensor_type.elem_type = TensorProto.FLOAT
+    input_tensor_type.shape.dim.extend(
+        [
+            onnx.TensorShapeProto.Dimension(dim_value=1),
+            onnx.TensorShapeProto.Dimension(dim_value=3),
+            onnx.TensorShapeProto.Dimension(dim_value=224),
+            onnx.TensorShapeProto.Dimension(dim_value=224),
+        ]
+    )
+
+    # run visit_value_info with empty context
+    visit_value_info(input_value_info, ctx)
+
+    # check keys in context
+    keys = list(ctx.type_by_name.keys())
+    assert keys == ["input_tensor"]
+
+    # check type info
+    type_info = str(ctx.type_by_name["input_tensor"])
+    assert type_info == "tensor<1x3x224x224xf32>"
+
+
+def test_visit_value_info_multiple_time():
+    # initialize context
+    ctx = OnnxXdslMapping()
+
+    # create ValueInfoProto input tensor
+    input_value_info = ValueInfoProto()
+    input_value_info.name = "input_tensor"
+
+    # define the type of the input tensor
+    input_tensor_type = input_value_info.type.tensor_type
+    input_tensor_type.elem_type = TensorProto.FLOAT
+    input_tensor_type.shape.dim.extend(
+        [
+            onnx.TensorShapeProto.Dimension(dim_value=1),
+            onnx.TensorShapeProto.Dimension(dim_value=3),
+            onnx.TensorShapeProto.Dimension(dim_value=224),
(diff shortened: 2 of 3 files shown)


tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,82 @@
...
+    # check keys in context
+    keys = list(ctx.type_by_name.keys())
+    assert keys == ["input_tensor"]
+
+    # check type info
+    type_info = str(ctx.type_by_name["input_tensor"])
+    assert type_info == "tensor<1x3x224x224xf32>"
+
+
+def test_visit_value_info_multiple_time():
```
superlopuh, line 48:
```suggestion
def test_visit_value_info():
```
Let's remove the multiple_time from the name, it's makes it sound like it visits it multiple times.

superlopuh: I mean it doesn't only test that, it checks the behaviour, including what happens when it visits it multiple times.

xdsl/frontend/onnx/context.py
```diff
@@ -0,0 +1,24 @@
...
+
+def visit_value_info(i: ValueInfoProto, ctx: OnnxXdslMapping) -> Attribute:
+    """Given the onnx ValueInforProto, it returns the corresponding Attribute stored in the context."""
+    name = i.name
+    t = get_type(i.type)
+    if name not in list(ctx.type_by_name.keys()):
+        ctx.type_by_name[name] = t
+        return t
+    else:
+        return ctx.type_by_name[name]
```
superlopuh, line 24:
```suggestion
    name = i.name
    if name in ctx.type_by_name:
        return ctx.type_by_name[name]
    t = get_type(i.type)
    ctx.type_by_name[name] = t
    return t
```
this is a pretty inefficient way to check that a key is in a dictionary, as lists have to be scanned linearly to test for whether they contain something, and dictionaries do this in constant time. Here is how I would recommend writing this. It has two advantages: we use the hash of the name to check whether we already have the information in O(1), and we only construct the type if it's not in the dictionary.

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+
+    # run visit_value_info with empty context
+    t1 = visit_value_info(input_value_info, ctx)
+
+    # check keys in context
+    keys = list(ctx.type_by_name.keys())
+    assert keys == ["input_tensor"]
+
+    # run visit_value_info with empty context
```
superlopuh, line 75:
```suggestion
    # run visit_value_info again
```

tests/frontend/onnx/test_context.py
```diff
@@ -0,0 +1,82 @@
...
+        [
+            onnx.TensorShapeProto.Dimension(dim_value=1),
+            onnx.TensorShapeProto.Dimension(dim_value=3),
+            onnx.TensorShapeProto.Dimension(dim_value=224),
+            onnx.TensorShapeProto.Dimension(dim_value=224),
+        ]
+    )
+
+    # run visit_value_info with empty context
+    t1 = visit_value_info(input_value_info, ctx)
```
superlopuh, line 69: Can you please add a check for what is returned here? We expect a tensor with a certain shape and element type

xdsl/frontend/onnx/context.py
superlopuh, line 1: I'm not sure that this is the best name for this file, I'd recommend something like `ir_builder.py`

tests/frontend/onnx/test_context.py
alecerio: Done

xdsl/frontend/onnx/context.py
alecerio: Done

tests/frontend/onnx/test_context.py
alecerio: Done

xdsl/frontend/onnx/context.py
alecerio: Done

tests/frontend/onnx/test_context.py
superlopuh, line 1: can you update the name of this file also, please?

pyproject.toml
```diff
@@ -45,12 +45,14 @@ enableTypeIgnoreComments = false
...
 "exclude" = [
+    "tests/dialects/test_memref.py",
+    "tests/frontend/onnx/test_context.py",
+    "tests/frontend/onnx/test_type.py",
     "tests/test_frontend_op_resolver.py",
     "tests/test_frontend_python_code_check.py",
-    "tests/frontend/onnx/test_type.py",
-    "tests/dialects/test_memref.py",
-    "xdsl/interpreters/experimental/wgpu.py",
+    "xdsl/frontend/onnx/context.py",
```
superlopuh, line 53: this list will also need to be updated

tests/frontend/onnx/test_context.py
superlopuh: Ah, let's merge them

pyproject.toml
alecerio: Done

tests/frontend/onnx/test_context.py
alecerio: Done

tests/frontend/onnx/test_build_ir.py
```diff
@@ -0,0 +1,93 @@
...
+    # check keys in context
+    keys = list(ctx.type_by_name.keys())
+    assert keys == ["input_tensor"]
+
+    # check type info
+    type_info = str(ctx.type_by_name["input_tensor"])
+    assert type_info == "tensor<1x3x224x224xf32>"
+
+
+def test_visit_value_info_multiple_time():
```
superlopuh, line 50: I feel like this test covers all the functionality we care about in this function, we don't really need the one above any more

alecerio: Done

superlopuh (approved):
Very nice

xdsl/frontend/onnx/ir_builder.py
```diff
@@ -0,0 +1,25 @@
+from onnx import ValueInfoProto
+
+from xdsl.frontend.onnx.type import get_type
+from xdsl.ir import Attribute
+
+
+class OnnxXdslMapping:
+    """The representation of the onnx context."""
```
compor, line 8: Yeah, but what does it do? Why is it useful? A better docstring would be appreciated

xdsl/frontend/onnx/ir_builder.py
```diff
@@ -0,0 +1,25 @@
...
+
+    type_by_name: dict[str, Attribute]
+
+    def __init__(self):
+        self.type_by_name = {}
+
+
+def visit_value_info(i: ValueInfoProto, ctx: OnnxXdslMapping) -> Attribute:
+    """Given the onnx ValueInforProto, it returns the corresponding Attribute stored in the context."""
```
compor, line 17:
```suggestion
    """Given the onnx ValueInfoProto, it returns the corresponding Attribute stored in the context."""
```

typo but also a few more words on this? is this a visitor pattern? why? what is the point of taking a `ctx`? etc

compor (approved):
LGTM modulo docstrings

compor:
Can be still addressed post-review in a separate PR if you feel like it.

PR #2439, opened 2024-04-09
dialects: Add accfg deduplication pass

JosseVanDelm:
This adds an accelerator deduplication pass that removes unnecessary state setup by implementing an algorithm that  recursively infers the previous state.

To this extent, it also introduces a HoistSetupCallsIntoConditionals pass which, as its name suggests,
Hoists one setup call happening right after an scf.if block into conditionals, this will never result in more setups, but it might result in less, since there is only one previous state to perform deduplication with, and there are 2 if this setup call is not hoisted in.

@AntonLydike I'm currently opening this PR as draft since I'm having some issues with pyright and I seem unable to address them.
I think our type declarations are somewhat lacking, and we are not really defining what should be the outcome of the `infer_state_of` if there is no previous state?

### DIFF
--- a/xdsl/transforms/accfg/trace_acc_state.py
+++ b/xdsl/transforms/accfg/trace_acc_state.py
@@ -41,7 +41,7 @@ def infer_state_of(state_var: SSAValue) -> State:
         # If the owner is an scf.if that has two possible setups
         # only update the state that is set in both possible setups
         case scf.If() as if_op:
-            return state_union(*infer_states_for_if(if_op, state_var))
+            return state_intersection(*infer_states_for_if(if_op, state_var))
         case Block():
             # TODO: maybe implement something here?
             return {}
@@ -70,5 +70,5 @@ def infer_states_for_if(op: scf.If, state: SSAValue) -> tuple[State, State]:
     return state_tuple
 
 
-def state_union(a: State, b: State) -> State:
-    return {k: a[k] for k in a if a[k] == b[k]}
+def state_intersection(a: State, b: State) -> State:
+    return {k: a[k] for k in a if a[k] == b.get(k)}
--- a/xdsl/transforms/accfg/trace_acc_state.py
+++ b/xdsl/transforms/accfg/trace_acc_state.py
@@ -57,7 +57,7 @@ def infer_states_for_if(op: scf.If, state: SSAValue) -> tuple[State, State]:
     assert state in op.results, "Expected state to be one of the scf.if results!"
     idx = op.results.index(state)
 
-    states = []
+    states: list[State] = []
     for region in op.regions:
         # we know the last op must be the yield
         yield_op = region.block.last_op
@@ -65,8 +65,9 @@ def infer_states_for_if(op: scf.If, state: SSAValue) -> tuple[State, State]:
         # we know the yield op has the same number of operands as the
         # scf.if has results, so [idx] must be defined
         states.append(infer_state_of(yield_op.operands[idx]))
-    assert len(states) == 2
-    return tuple(states)
+    state_tuple = tuple(states)
+    assert len(state_tuple) == 2
+    return state_tuple
 
 
 def state_union(a: State, b: State) -> State:
--- a/xdsl/transforms/accfg/trace_acc_state.py
+++ b/xdsl/transforms/accfg/trace_acc_state.py
@@ -1,0 +1,67 @@
+"""
+These inference helpers are used to figure out which values are
+present at which points in the IR.
+
+Example inference results:
+
+```
+accfg.setup(A = %1, B = %2)  // previous state = {}
+// ...
+accfg.setup(A = %3, C = %1)  // previous state = {A = %1, B = %2}
+```
+
+These inference passes walk the IR backwards.
+"""
+
+from xdsl.dialects import accfg, scf
+from xdsl.ir import Block, SSAValue
+
+State = dict[str, SSAValue]
+
+
+def infer_state_of(state_var: SSAValue) -> State:
+    """
+    Entrance function of the inference pass.
+
+    This walks up the def-use chain to compute all values
+    that are guaranteed to be set in this state.
+    """
+    owner = state_var.owner
+    match owner:
+        case accfg.SetupOp(in_state=None) as setup_op:
+            return {name: val for name, val in setup_op.iter_params()}
+        case accfg.SetupOp(in_state=st) as setup_op if st is not None:
+            in_state = infer_state_of(st)
+            in_state.update(dict(setup_op.iter_params()))
+            return in_state
+        case scf.If() as if_op:
+            return state_union(*infer_states_for_if(if_op, state_var))
+        case Block():
+            # TODO: maybe implement something here?
+            return {}
+        case _:
+            raise ValueError(f"Cannot infer state for op {owner.name}")
+
+
+def infer_states_for_if(op: scf.If, state: SSAValue) -> tuple[State, State]:
+    """
+    Walk both sides of the if/else block and return the computed
+    states for the given state SSA value (`state`)
+    """
+    assert state in op.results, "Expected state to be one of the scf.if results!"
+    idx = op.results.index(state)
+
+    states = []
+    for region in op.regions:
+        # we know the last op must be the yield
+        yield_op = region.block.last_op
+        assert isinstance(yield_op, scf.Yield)
+        # we know the yield op has the same number of operands as the
+        # scf.if has results, so [idx] must be defined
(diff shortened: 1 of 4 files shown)


xdsl/transforms/accfg/trace_acc_state.py
```diff
@@ -0,0 +1,73 @@
...
+
+def infer_states_for_if(op: scf.If, state: SSAValue) -> tuple[State, State]:
+    """
+    Walk both sides of the if/else block and return the computed
+    states for the given state SSA value (`state`)
+    """
+    assert state in op.results, "Expected state to be one of the scf.if results!"
+    idx = op.results.index(state)
+
+    states = []
```
superlopuh, line 60:
```suggestion
    states: list[State] = []
```

xdsl/transforms/accfg_dedup.py
```diff
@@ -0,0 +1,154 @@
...
+    Reduce the number of parameters in setup calls by inferring previously
+    set up values and carefully moving setup calls around.
+    """
+
+    name = "accfg-dedup"
+
+    hoist: bool = True
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        patterns = [
```
superlopuh, line 144:
```suggestion
        patterns: list[RewritePattern] = [
```

xdsl/transforms/accfg/trace_acc_state.py
```diff
@@ -0,0 +1,74 @@
...
+            in_state.update(dict(setup_op.iter_params()))
+            return in_state
+        # If the owner is an scf.if that has two possible setups
+        # only update the state that is set in both possible setups
+        case scf.If() as if_op:
+            return state_union(*infer_states_for_if(if_op, state_var))
+        case Block():
+            # TODO: maybe implement something here?
+            return {}
+        case _:
```
JosseVanDelm, line 48: @AntonLydike, do I get it correctly that this is only an issue if at some point we lower to cf instead of working on scf? Returning an empty dict feels wrong here (can we also just throw an error here ? I don't understand why pyright is not complaining actually, is this not a dict{None: None} that  is returned?

JosseVanDelm:
Thanks for the help with Pyright @superlopuh!
I'm marking this as ready for review, but I'm honestly not too sure what the current limitations are, and I might want to more explicitly put this somewhere/maybe add a negative test for some of the cases that are currently unexplored...

xdsl/transforms/accfg/trace_acc_state.py
AntonLydike: Yes, or if we want to handle state transfers across function calls at some point. But that's far, far away.

xdsl/transforms/accfg/trace_acc_state.py
```diff
@@ -0,0 +1,74 @@
...
+accfg.setup(A = %3, C = %1)  // previous state = {A = %1, B = %2}
+```
+
+These inference passes walk the IR backwards.
+"""
+
+from xdsl.dialects import accfg, scf
+from xdsl.ir import Block, SSAValue
+
+State = dict[str, SSAValue]
```
superlopuh, line 19: I'm not sure what exactly it does but I have the impression that it helps Pyright a little bit if you add an annotation to say explicitly what's happening:
```suggestion
State: TypeAlias = dict[str, SSAValue]
```
Although, in this case, it feels like it might be worth adding some more structure and documentation to say what the object represents etc? Especially since `state_union` seems like it might as well be a method on the object. The lowest overhead way of doing that is this:
``` python
class AccState(NamedTuple):
    known_states: dict[str, SSAValue]
```
Like for `dataclass`, the init is auto generated

xdsl/transforms/accfg/trace_acc_state.py
```diff
@@ -0,0 +1,74 @@
...
+        assert isinstance(yield_op, scf.Yield)
+        # we know the yield op has the same number of operands as the
+        # scf.if has results, so [idx] must be defined
+        states.append(infer_state_of(yield_op.operands[idx]))
+    state_tuple = tuple(states)
+    assert len(state_tuple) == 2
+    return state_tuple
+
+
+def state_union(a: State, b: State) -> State:
```
superlopuh, line 73: seems like more of an intersection to me?

xdsl/transforms/accfg/trace_acc_state.py
```diff
@@ -0,0 +1,74 @@
...
+These inference passes walk the IR backwards.
+"""
+
+from xdsl.dialects import accfg, scf
+from xdsl.ir import Block, SSAValue
+
+State = dict[str, SSAValue]
+
+
+def infer_state_of(state_var: SSAValue) -> State:
```
superlopuh, line 22: Would it not be useful to stop early? If the same values keep getting updated, it seems like we'd want to stop after seeing the latest definitions, not go up the whole chain every time.

superlopuh: It looks like the complexity of getting the known states for n ops is O(n^2) right now, if I understand correctly

xdsl/transforms/accfg/trace_acc_state.py
JosseVanDelm: I agree :D

xdsl/transforms/accfg/trace_acc_state.py
JosseVanDelm: @superlopuh, That's a very good comment. I think the complexity gets even worse (exponentially) the more if statements you have. We can probably stop early if we carry the currently known state with the recursive function and only update the values that are not known. However, right now that means I have to do a symbol table lookup on the accelerator symbol that specifies the amount of parameters that have to be set to get a stopping criterion. Is there a canonical way of accessing the symbol table if the current scope is only the op in that module?

xdsl/transforms/accfg/trace_acc_state.py
JosseVanDelm: @superlopuh I tend to agree, but I can not use the NamedTuple as you suggested (seems only available as factory method) and It doesn't seem to make the current code much more readable.

xdsl/transforms/accfg/trace_acc_state.py
AntonLydike: I would add optimizations when we need to, for now, I'd keep the code as simple as possible.

If we decide to change something, it's infinitely easier to edit un-optimized code.

superlopuh: My point is that I'm not sure that this is as simple as possible :) @JosseVanDelm, what do you mean by "canonical way of accessing the symbol table?"

xdsl/transforms/accfg/trace_acc_state.py
superlopuh: I think you have to import it from `typing` for it to work this way, but I'm not sure, I've used it like this before and it's worked

xdsl/transforms/accfg/trace_acc_state.py
JosseVanDelm: I think I was looking for:
```
SymbolTable.lookup_symbol
```

superlopuh: Ah yeah that was my first thought but I didn't realise that it's what you were looking for :)

JosseVanDelm: If okay I'd also like to postponing this as an optimization. I think we have other more pressing issues to work on first :D

AntonLydike:
@JosseVanDelm what's the status with this? Want me to take over?

xdsl/transforms/accfg/trace_acc_state.py
AntonLydike: Solved!

superlopuh:
I thought I'd try to put in words why I'm hesitant about this PR. It seems to be doing something useful, but there are a few aspects of it that make me uneasy. One is that it is quite a lot of logic that seems quite closely coupled to `scf`, and it's not clear whether it would miscompile things in a more complex confrol flow structure. Another is that it's not clear that this will be easy to adapt as the accfg dialect evolves. Most importantly, it seems like an abstract optimization, in that you didn't write about what impact this has in practice. It would be interesting to see how it fits into the bigger picture of compiling to SNAX, and whether there are any ways to do a similar optimisation at a higher level, without doing a fairly involved traversal at this level.

AntonLydike:
This is on ice for now, we have switched back to doing all of this in snax-mlir instead. Am happy to chat privately and/or present this work in an upcoming lab meeting though!

PR #2441, opened 2024-04-09
frontend: (onnx) visit_node implementation

alecerio:
Implementation of visit_node and related tests.

### DIFF
--- a/tests/frontend/onnx/test_build_ir.py
+++ b/tests/frontend/onnx/test_build_ir.py
@@ -2,7 +2,7 @@ import onnx
 import pytest
 
 from xdsl.dialects.builtin import TensorType, f32
-from xdsl.dialects.onnx import Sub
+from xdsl.dialects.onnx import Add, Sub
 from xdsl.ir import Attribute
 from xdsl.utils.test_value import TestSSAValue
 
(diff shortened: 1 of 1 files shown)


xdsl/frontend/onnx/ir_builder.py
```diff
@@ -1,16 +1,42 @@
...
+        raise ValueError(f"Unknown ONNX op name {node.op_type}")
+
+    op = OP_BY_OP_TYPE[node.op_type]
+
+    operands = tuple(ctx.value_by_name[name] for name in node.input)
+    result_types = tuple(ctx.type_by_name[name] for name in node.output)
+
+    results = op.build(operands=operands, result_types=result_types).results
+
+    for output_name, result in zip(node.output, results, strict=True):
```
superlopuh, line 38: What operations have multiple outputs? It feels like we might be better off raising `NotImplementedError` if the result count is anything but 1, that way we catch the case when it comes up.

tests/frontend/onnx/test_build_ir.py
```diff
@@ -59,3 +62,108 @@ def test_visit_value_info():
...
+
+def _update_context(graph: GraphProto, ctx: OnnxXdslMapping):
+    name = graph.name
+    input_types = tuple(visit_value_info(input, ctx) for input in graph.input)
+    output_types = tuple(visit_value_info(output, ctx) for output in graph.output)
+
+    fn = func.FuncOp(name, (input_types, output_types))
+    with ImplicitBuilder(fn.body) as args:
+        for input, arg in zip(graph.input, args, strict=True):
+            ctx.value_by_name[input.name] = arg
```
superlopuh, line 169: This feels like quite a complex helper to me. Could you please add a bit of code to do this manually? So something like this inlined into the tests above:
``` python
lhs = TestSSAValue()
rhs = TestSSAValue()
ctx = OnnxXdslMapping({
    "input1": lhs,
    "input2": rhs,
}, {})
op = visit_node(node, ctx)
assert op.lhs is lhs
assert op.rhs is rhs
...
```

tests/frontend/onnx/test_build_ir.py
```diff
@@ -59,3 +62,108 @@ def test_visit_value_info():
...
+    # update context
+    _update_context(graph, ctx)
+
+    # expected output before visinting the op node
+    expected_output_pre = """{'input1': <BlockArgument[tensor<0x0xf32>] index: 0, uses: 0>, 'input2': <BlockArgument[tensor<0x0xf32>] index: 1, uses: 0>}"""
+
+    assert str(ctx.value_by_name) == expected_output_pre
+
+    # visit node
+    visit_node(sub_node, ctx)
```
superlopuh, line 100: could you please add a test for the output of this function?

alecerio: @superlopuh , visit_node returns nothing as output, it just update the content of ctx.
Is there something else I could test?

alecerio: I added the Operation as function output and compare the name of the operation in the test.

alecerio: (Done)

tests/frontend/onnx/test_build_ir.py
alecerio: Ok, so instead of calling _update_context in the test, I should do this manually inside the test?

alecerio: Done

tests/frontend/onnx/test_build_ir.py
```diff
@@ -59,3 +62,135 @@ def test_visit_value_info():
...
+    assert op.name == "onnx.Sub"
+
+    # expected output after visinting the op node
+
+    expected_out = """'output': <OpResult[tensor<64xf32>] index: 0, operation: onnx.Sub, uses: 0>"""
+    expected_output_post = (
+        "{" + expected_input1 + ", " + expected_input2 + ", " + expected_out + "}"
+    )
+
+    assert str(ctx.value_by_name) == expected_output_post
```
superlopuh, line 162:
```suggestion
    op = visit_node(sub_node, ctx)

    assert op.is_structurally_equivalent(onnx.Sub(lhs, rhs, out_type))
```
I don't think we need to test the ctx, this test for whether all the operands, results, and attributes are the same should cover everything that we want to be stable, independent of the textual representation of the elements

superlopuh (commented):
Looking great! One tiny change to the test and I think we should be good to go

tests/frontend/onnx/test_build_ir.py
superlopuh: Thank you

superlopuh (approved):
Looking great!

PR #2442, opened 2024-04-09
core: Add UseDefChain Traits

ed741:
Add HasAncestor Trait
 - Constrain the op such that another Op is its transitive parent (before reaching ModuleOp)
 - Has a weak mode that allows also allows None to the be transitive parent such that it is valid to not yet have the op in a larger piece of IR

Add UseDefChain Trait
 - Provides a mechanism to follow use-def chains through control flow primitives
 - Implemented for scf as an example

### DIFF
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
 import abc
+import collections
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, Callable, Generic, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
 
 from xdsl.utils.exceptions import VerifyException
 
@@ -415,7 +416,7 @@ class HasAncestor(OpTrait):
         super().__init__(parameters)
 
     def verify(self, op: Operation) -> None:
-        ancestor_type, weak  = self.parameters
+        ancestor_type, weak = self.parameters
         parent = op.parent_op()
         while not isinstance(parent, ModuleOp) and (weak or parent is not None):
             if parent is None:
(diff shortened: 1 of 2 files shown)


math-fehr:
I'll take a look at it in 2 days, I'm just in the middle of a conference, but it's on my todo list!

superlopuh:
I haven't had an in-depth look yet, but do these traits depend on each other? If not, would you mind splitting this PR into two?

ed741:
I an effort to get type checking to play nicely I've basically rewritten most of it such that different Ops will need to subclass the UseDefChainTrait to provide class methods that do the logic, rather than using lambdas etc.

xdsl/traits.py
```diff
@@ -398,3 +398,92 @@ def verify(self, op: Operation) -> None:
...
+    # @classmethod
+    # def get_operands_leading_to_def(cls, ssa_value: SSAValue) -> set[Use]:
+    #     if isinstance(ssa_value.owner, Operation):
+    #         result = cast(OpResult, ssa_value)
+    #         return cls.get_operands_leading_to_op_result(result.op, result.index)
+    #     elif isinstance(ssa_value.owner, Block):
+    #         arg = cast(BlockArgument, ssa_value)
+    #         return cls.get_operands_leading_to_block_argument(arg.block.parent_op(), arg)
+    #     else:
+    #         raise NotImplementedError(f"UseDefChainTrait does not support ssa values that are: {type(ssa_value)}")
```
superlopuh, line 441: commented code

superlopuh:
Hi @ed741, let's discuss the plan for this? It seems like something like this would be useful, but I wonder if there's a way to make the approach more generic, and avoid implementing the trait for every operation. Happy to chat about this on Zulip or at one of the xDSL meetings.

superlopuh:
As discussed during the social, let's close this, and revisit with a more general approach in the future.

PR #2445, opened 2024-04-10
core: More informative rewrite errors.

PapyChacal:
xDSL currently bails out completely on the first exception raised while rewriting.
This changes this behaviour by capturing the said exception and printing it in the IR's context, just like verification errors.
Thus, when reading the error, the user can see the state of the IR at the point where a specific pattern failed, making it easier to reason about fixes or improvements IMO.

This is mostly intended to help while *developing* patterns, not any statement on the "should they even fail" topic.

### DIFF
--- a/tests/pattern_rewriter/test_pattern_rewriter.py
+++ b/tests/pattern_rewriter/test_pattern_rewriter.py
@@ -1,3 +1,4 @@
+import re
 from collections.abc import Sequence
 
 import pytest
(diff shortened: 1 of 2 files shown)


superlopuh (approved):
🙏 🙏 🙏

webmiche (commented):
Nice, very good idea!

However, I feel that we might run into issues if we are recursively rewriting. AFAIU, this will just recurse out and continue rewriting the rest. In some cases, this might lead to very weird behavior, say the op on which you emitted the exception gets deleted later on. Then people will be extremely confused what happened and why.

IMO, in standard xdsl-opt, I would prefer to add the IR print while keeping the behavior that this backs out of the entire rewriting on the first exception. I would then add a flag that allows the rewriting process to continue even if exceptions are encountered.

PapyChacal:
No worries, I actually thought about that side too, from another perspective.
I'm *slightly* frustrated by the recursive raising, as this just raises from whatever top-level pattern was used, which usually is the greedy rewriter. I would find it way more interesting to simply bubble up the exception from the actual failing pattern.
Then we could directly print the pattern or the exception's exact location in code for super dev-friendly output 🙂
It's a rabbit hole I don't want to go solo in though, and I'd rather have this cute improvement first, but I'm all ears for ideas on getting something like this!

PR #2448, opened 2024-04-11
core: Fix optional operand declarative printing.

PapyChacal:
Hopefully fixes #2413

PR #2450, opened 2024-04-11
dialects: (seq) Finish `seq.compreg` optionals

lucjaulmes:
Now #2448 and #2407 are in we can finally complete what was left out of #2404:
- an optional attribute `inner_sym` 
- the second optional operand `powerOn`


PapyChacal (approved):
LGTM!

superlopuh (approved):

Moxinilian (approved):
Looks good! CIRCT seems to have a lot of places where it cares about SSA names being consistent (with the option to set names separately otherwise), but considering this was fairly complex when I implemented it in HwModule, I think it's better if we only add it when truly necessary. I think all the xDSL features to do it are here but it would probably require some refactoring to do it properly.

PR #2454, opened 2024-04-12
dialects: Implement `stencil.combine`

PapyChacal:
And implement support in verifiers and shape inference.
No support in the existing lowering is implemented yet, as this operation is not meant to be lowered directly but handled by dedicated passes; so will implement those separately.

### DIFF
--- a/xdsl/transforms/experimental/stencil_shape_inference.py
+++ b/xdsl/transforms/experimental/stencil_shape_inference.py
@@ -68,46 +68,103 @@ class CombineOpShapeInference(RewritePattern):
     def match_and_rewrite(self, op: CombineOp, rewriter: PatternRewriter, /):
         # Get each result group
         combined_res = op.results_[0 : len(op.lower)]
-        op.results_[len(op.lower) : len(op.lower) + len(op.lowerext)]
-        op.results_[len(op.lower) + len(op.lowerext) :]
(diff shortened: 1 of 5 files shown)


PapyChacal:
I'm working with the original dialect's test cases, and indeed a whole part of the shape inference is not tested :thinking: 
Extending the test case now, I don't want to be bitten by this ever :upside_down_face:

xdsl/transforms/experimental/stencil_shape_inference.py
```diff
@@ -62,6 +63,110 @@ def infer_core_size(op: LoadOp) -> tuple[IndexAttr, IndexAttr]:
...
+            # Compute the new extreme bounds as usual.
+            lb = IndexAttr.min(c_bounds.lb, lb)
+            # Compute the combine bounds
+            c_bound_c = list(c_bounds.ub)
+            c_bound_c[op.dim.value.data] = op.index.value.data
+            c_bound = IndexAttr.get(*c_bound_c)
+            ub = IndexAttr.max(c_bound, ub)
+            bounds = StencilBoundsAttr(zip(lb, ub))
+            l.type = TempType(bounds, l.type.element_type)
+        # Handle combined upper results
```
georgebisbas, line 98: Add an empty line maube

georgebisbas (approved):

PR #2455, opened 2024-04-12
dialects: (x86) PR5_2 - single operand instructions - one destination

KGrykiel:
next part of PR #2398. added the case where the single operand is a destination + the pop operation


xdsl/dialects/x86/ops.py
```diff
@@ -320,6 +320,39 @@ class PushOp(ROperationSrc[GeneralRegisterType]):
     name = "x86.push"
 
 
+class ROperationDst(Generic[R1InvT], SingleOperandInstruction):
```
superlopuh, line 323: This instruction doesn't have any operands, am I missing something?

KGrykiel: This is by design although worth discussing. The pop function moves data from the top of the stack to a destination register. Whatever is inside the register is irrelevant, so I thought it best to only designate it as a result rather than an operand.

superlopuh: That's exactly the right design, IMO, I don't see why inherit from that class

KGrykiel: oh, right, in that sense. I used the single/double/tripleOperandInstruction classes as a way to differentiate the instructions over how they look in assembly. In the sense how many arguments, whatever that may be, an assembly instruction takes. I suppose pop creates this kind of gray area where the assembly instruction takes an argument (i.e. "pop rax") but it's not *really* an operand. 

In hindsight this might be a little confusing, but I'm not sure how best to handle the software engineering aspect of class hierarchy in this case. Now that I've implemented more instructions since first devising the plan, I couldn't find anything meaningful to put in the single/double/tripleOperandInstruction parent classes so perhaps it's best to get rid of them all together and have everything inherit from x86Instruction directly?

I'm fine with whatever you think is best tbh

superlopuh: I think we might as well get rid of them, we have another bunch of abstract class helpers for the inits already, so we might as well drop the marker classes.

tarinduj: Yeah, I think it makes sense to get rid of them. I am also wondering if it would be useful to encode intrinsic information (like pop is modifying %rsp?) in the class as well?

superlopuh: I'm not sure what you mean by that

tarinduj: Say we want to know what registers were modified/ used by an instruction. With instructions like add you know exactly which ones were affected cause they are the operands. So, you can look up the operands of the class instance. But for an instruction like pop, there is no straightforward way to extract that it modified the %rsp register without reasoning about the instruction itself. I am looking at it in terms of having a uniform API to get this information for all instructions. Not sure if it makes sense.

tarinduj: I was trying to visualize what was happening with pop. Does it make sense to model the parent classes like this? 

add rax, rbx
rax = rax + rbx
R <- R,R 
R-RR-Operation *

pop rax
rax = (rsp)
R <- M
R-M-Operation *

push rax
(rsp) = rax
M <- R
M-R-Operation *

So, rsp basically acts like an intrinsic operand for push and pop. 

* Maybe naming parent classes like [Result]-[Operands]-Operation instead of [Operands]Operation would be more consistent?

superlopuh: Hmm, I didn't know that that's what it did. I would actually model it in SSA as taking a `%rsp` operand, and return a second result with the new `%rsp`, and similarly with all the other ops that have side-effects. That way we can reason about dataflow.

KGrykiel: Ok, I opened a new issue for removing the parent classes (#2466), I'll do it later because now it would mess up merging for the other PR I have open.

Going forward with this, would you rather I categorize the instructions by the number of arguments the assembly instruction takes or by the actual number of operands including the intrinsic ones? In this example the 'push' instruction  for instance could be an ROperation (because in the assembly it would look like 'push rdx') or an RR Operation (because it technically affects 2 registers). What are your thoughts?

superlopuh: I wouldn't think of it as a big taxaonomy, the only reason we added it to riscv was to have less duplication of code for the initializers, I'd just group them by the init signature and not overthink it.

tarinduj: I think our current parent class naming convention (eg: add rax, rbx: RROperatio) only makes sense when one of the operands is also the result. IMO renaming the parent classes to reflect other cases as well would make more sense. Something like [Result]-[Operands]-Operation (eg: add rax, rbx: R-RR-Operation). So, push would technically be M-R-Operation (because it's dereferencing rsp)? 

I think it would be helpful to de-sugar the assembly and keep the information explicitly in the SSA.

KGrykiel: That's also an option, i can change that in a fell swoop with the removal of the parent classes :+1:

tarinduj: If not, we can think of push and pop as a mov and do the same thing we do there. Whichever makes more sense.

KGrykiel:
@tarinduj @superlopuh let me know if this is what you had in mind regarding %rsp. If so I imagine I'll have to give the same treatment to the 'push' instruction in a different PR.

superlopuh (approved):

PR #2465, opened 2024-04-16
docs: Introduce the xDSL book

Moxinilian:
This PR introduces a mdbook that should be used as an entry point for MLIR/xDSL newcomers. This is only an infrastructure setup, actual book content should be discussed later. For now, an example structure has been added for demonstration purposes.

This set up allows embedding Python code within the book. The CI has been updated to verify this Python code will run without error. How to use the infrastructure is detailed in the added book README.

Here is a screenshot of one of the pages, with dark theme enabled:

image


Moxinilian:
I think after rewriting it would be nice to have a chapter about standard dialects, explaining their role and rationale, akin to how CIRCT does it.

docs/book/src/SUMMARY.md
```diff
@@ -0,0 +1,14 @@
...
+
+[Introduction](introduction.md)
+
+- [Starting an xDSL project](starting.md)
+- [An xDSL program in examples](examples.md)
+- [Intermediate Representation (IR)](ir/index.md)
+    - [Dialects](ir/dialects.md)
+    - [Attributes](ir/attributes.md)
+    - [Regions](ir/regions.md)
+    - [Verifiers](ir/verifiers.md)
```
superlopuh, line 11:
```suggestion
    - [Attributes](ir/attributes.md)
    - [Operations](ir/operations.md)
    - [Regions](ir/regions.md)
    - [Verifiers](ir/verifiers.md)
    - [Dialects](ir/dialects.md)
```

superlopuh: Oh also Traits

Moxinilian: Yes. I feel like traits could be introduced later, after rewriting, in a re-visit to introduce more complex concepts like traits/interfaces. But maybe there is not enough material to do this.

superlopuh: Custom syntax also somewhere late

.github/workflows/ci-core.yml
```diff
@@ -49,3 +49,10 @@ jobs:
         export PYTHONPATH=$(pwd)
         lit -v tests/filecheck/
         lit -v docs/Toy/examples/
+    
+    - name: Execute book tests
+      run: |
+        for f in docs/book/**/*.py; do
+          echo "Running book code: '$f'";
+          python "$f";
+        done
```
superlopuh, line 58: can you please add this to make tests? this script should probably be running those tests also, but that's for another PR. Ideally `test book`.

Moxinilian: I am not sure if this script is the best way to do this in the first place. With your Python build system experience, would you do it differently?

superlopuh: Actually why this format and not the way we have it in Toy, as just an extra project with source files and pytest tests?

Moxinilian: Could you detail a bit more? The files that will be added will most likely not have the usual "list of functions" format used with pytest.

superlopuh: This is partially why I suggested moving the Toy tutorial. Even though it's not what we want to have in the book in the long run, it contains realistic snippets of code, and would immediately surface missing functionality.

superlopuh:
I'm happy to reopen the PR, if you'd like me to

PR #2468, opened 2024-04-17
transformations: (convert-riscv-to-llvm)

jorendumoulin:
This PR implements a conversion of riscv dialect to llvm using the inline assembly op.
This is useful such that any generated RISC-V code can also be used with an LLVM target like we use in `snax-mlir`. 
Specifically, this will be used to reuse the lowering of `snrt` to `riscv`, without having to duplicate the entire `snrt-to-riscv` transform.

@AntonLydike


jorendumoulin:
Some further reflection on this approach:

It is functional, but the amount of unrealized conversion casts gets a bit ridiculous (see kuleuven-micas/snax-mlir#116). Also, the emission of riscv arithmetic in llvm inline assembly instead of regular llvm ir is a bit silly (not in the example).

I think the following approach would make more sense:

Create ops for the custom snitch operations (such as `dmdst`, `dmsrc`, `dmstrd`, ...) in the `snrt` dialect as _snrt primitives_, which still operate on basic mlir types (i32, i64) as opposed to riscv registers.

refactor the `convert-snrt-to-riscv` pass to first lower to the _snrt primitives_, and only later convert them to actual `riscv` operations. This makes the alternative path to llvm much easier and as such we can also avoid the unrealized conversion casts entirely.

The lowerings would look like this:

```
              snrt                                                   
              │                                                      
              │ lower-snrt                                           
              │                                                      
              │                                                      
              ▼                                                      
              arith + snrt primitives                                
                              │                                      
              │               │                                      
              │               │                                      
   ┌──────────┘               └───────┐                              
   │  convert-snrt-to-riscv           │  convert-snrt-to-llvm        
   │  convert-arith-to-riscv          │  convert-arith-to-llvm (MLIR)
   ▼                                  ▼                              
riscv                          llvm + llvm inline asm                
```            

I can work on this, but I would like to get your opinion on this first @AntonLydike @JosseVanDelm .

JosseVanDelm:
@jorendumoulin I agree with your suggestion.
Implementing the regular arithmetic directly as riscv operations feels like premature lowering, I think we have support for those arith to riscv lowerings in xdsl anyway. 

CC @cappadokes

AntonLydike:
Oh no, we hit the magic interactive tests again :(

AntonLydike:
@jorendumoulin I did a pass to handle named registers in puts differently. We now just print the register name in the assembly instead of going through the SSA value.

Additionally, with #2797 we can now CSE delete the remaining `reg_register` ops left over by this pass.

I'll fix the problem with ops that return a `riscv.reg<zero>` register tomorrow, and that should hopefully make this PR complete.

superlopuh:
What's the latest on this?

superlopuh:
No longer [WIP]?

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,164 @@
...
+            elif isinstance(arg, SSAValue) and isinstance(arg.type, IntRegisterType):
+                # if the input is allocated to a register, use that register
+                if arg.type.is_allocated and arg.type.index == IntAttr(0):
+                    assembly_args_str.append("x0")
+                # otherwise we need to get the value from the SSA value
+                else:
+                    conversion_op = UnrealizedConversionCastOp.get([arg], [builtin.i32])
+                    ops_to_insert.append(conversion_op)
+                    inputs.append(conversion_op.outputs[0])
+                    constraints.append("r")
```
jorendumoulin, line 78: I think we can use 'rI' here. LLVM is then smart enough to detect if it is possible to use an immediate, else it uses a register
```suggestion
                    constraints.append("rI")
```

AntonLydike:
Yes, this is now ready for your all eyes

jorendumoulin (commented):
Indeed, no longer WIP

Thank you, this looks good to me!

xdsl/tools/command_line_tool.py
```diff
@@ -46,6 +46,11 @@ def get_convert_linalg_to_loops():
 
         return convert_linalg_to_loops.ConvertLinalgToLoopsPass
 
+    def get_convert_riscv_scf_to_llvm():
+        from xdsl.transforms import convert_riscv_to_llvm
+
+        return convert_riscv_to_llvm.ConvertRiscvToLLVMPass
+
```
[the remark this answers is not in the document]
superlopuh: I know it's already not super alphabetical, I'll need to clean this up at some point.

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,159 @@
...
+                    constraints.append("rI")
+                    assembly_args_str.append(f"${len(inputs) + num_results - 1}")
+
+            # constant value used as an immediate
+            elif isinstance(arg, IntegerAttr):
+                assembly_args_str.append(str(arg.value.data))
+
+            # not supported argument
+            else:
+                return
```
superlopuh, line 82: Seems like it might be worth raising a DiagnosticException here or some other erroring out

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,159 @@
...
+                constraints.append("=r")
+                # map final result to a result of the inline asm op
+                result_map.append(num_results)
+                num_results += 1
+
+            # ssa value used as an input operand
+            elif isinstance(arg, SSAValue) and isinstance(arg.type, IntRegisterType):
+                # if the input is allocated to a register, use that register
+                if arg.type.is_allocated and arg.type.index == IntAttr(0):
+                    assembly_args_str.append("x0")
```
superlopuh, line 67: "x0"?

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # construct asm_string
+        iname = op.assembly_instruction_name()
+
+        # check if the operation has a custom insn string (for compatibility reasons)
+        custom_insn = op.get_trait(HasInsnRepresentation)
+        if custom_insn is not None:
+            # generate custom insn inline assembly instruction
+            # thank you pyright for making my code so much better I truly appreciate your presence
+            # put the forbidden ~fruit~ method in a variable with a short name:
```
superlopuh, line 92: Can you please add a comment to the issue for updating pyright instead?

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,159 @@
...
+            n = custom_insn.get_insn
```
superlopuh, line 93:
```suggestion
            get_insn = custom_insn.get_insn
```

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,159 @@
...
+                [register_width_int] * num_results,
+            )
+        )
+        op_results = new_op.results
+
+        # cast output back to original type if necessary
+        if num_results > 0:
+            ops_to_insert.append(
+                output_op := UnrealizedConversionCastOp.get(
+                    new_op.results, [r.type for r in op.results]
```
superlopuh, line 123: won't this break when returning 0?

AntonLydike: You are indeed correct!

superlopuh:
What's the latest here?

jorendumoulin:
This seems very close to being finished. @AntonLydike do you by any chance have local progress on this / some time to finish this? Otherwise I can try and get this through.

superlopuh:
@jorendumoulin, I think if you could update this and get it in, it would be fabulous, I'm sure we'll find more use-cases for this functionality

xdsl/transforms/convert_riscv_to_llvm.py
jorendumoulin: pyright does not complain anymore!

jorendumoulin:
I rebased and resolved all outstanding comments, I just don't really get what is meant with this final comment:

> won't this break when returning 0?

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,157 @@
...
+        result_map: list[int | SSAValue] = []
+
+        # populate assembly_args_str and constraints
+        for arg in op.assembly_line_args():
+            # ssa value used as an output operand
+            if (
+                isinstance(arg, OpResult)
+                and isinstance(arg.type, IntRegisterType)
+                and arg.op is op
+            ):
```
superlopuh, line 47: could you please use match syntax here?

superlopuh: Also, I think `arg.owner is op` is enough to determine whether an assembly arg is an operand or result

tests/filecheck/transforms/convert_riscv_to_llvm.mlir
superlopuh, line 1: could you please name the ssa values in the input, to make it a bit easier to track the outputs?

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,157 @@
...
+        # construct asm_string
+        iname = op.assembly_instruction_name()
+
+        # check if the operation has a custom insn string (for compatibility reasons)
+        custom_insn = op.get_trait(HasInsnRepresentation)
+        if custom_insn is not None:
+            # generate custom insn inline assembly instruction
+            insn_str = custom_insn.get_insn(op)
+            asm_string = insn_str.format(*assembly_args_str)
+
```
superlopuh, line 97:
```suggestion
```

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,157 @@
...
+                assembly_args_str.append(str(arg.value.data))
+
+            # not supported argument
+            else:
+                raise DiagnosticException(
+                    "unsupported argument for conversion to an llvm inline assembly instruction"
+                )
+
+        # construct asm_string
+        iname = op.assembly_instruction_name()
```
superlopuh, line 89:
```suggestion
        instruction_name = op.assembly_instruction_name()
```

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,157 @@
...
+                asm_string,
+                constraints_string,
+                inputs,
+                [register_width_int] * num_results,
+            )
+        )
+        op_results = new_op.results
+
+        # cast output back to original type if necessary
+        if num_results > 0:
```
superlopuh, line 118:
```suggestion
        if num_results:
```

xdsl/transforms/convert_riscv_to_llvm.py
```diff
@@ -0,0 +1,157 @@
...
+
+        else:
+            # generate generic riscv inline assembly instruction
+            asm_string = iname + " " + ", ".join(assembly_args_str)
+
+        # construct constraints_string
+        constraints_string = ",".join(constraints)
+
+        # construct llvm inline asm op
+        register_width_int = builtin.IntegerType(self.xlen)
```
superlopuh, line 106: Why not index type?

superlopuh:
> I rebased and resolved all outstanding comments, I just don't really get what is meant with this final comment:
> 
> > won't this break when returning 0?

I think it may have been outdated, before the get registers for zero... I'm not 100% sure, I can no longer spot an issue.

xdsl/transforms/convert_riscv_to_llvm.py
superlopuh: !!

superlopuh (approved):

jorendumoulin:
Thanks! Apologies for the extreme delay on this! I hope I'll soon find some time to convert our snitch runtime stuff to use this flow :)

superlopuh:
The important thing is that we got there in the end, thank you very much for picking it back up!

PR #2469, opened 2024-04-17
frontend: (onnx) integration matmul op in onnx frontend

alecerio:
Integration of matmul operation in onnx frontend.

@superlopuh @compor

### DIFF
--- a/tests/frontend/onnx/test_build_ir.py
+++ b/tests/frontend/onnx/test_build_ir.py
@@ -2,7 +2,7 @@ import onnx
 import pytest
 
 from xdsl.dialects.builtin import TensorType, f32
-from xdsl.dialects.onnx import Add, Sub
+from xdsl.dialects.onnx import Add, MatMul, Sub
 from xdsl.ir import Attribute
 from xdsl.utils.test_value import TestSSAValue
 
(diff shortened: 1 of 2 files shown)


tests/frontend/onnx/test_build_ir.py
```diff
@@ -187,6 +187,28 @@ def test_visit_graph_sub():
...
 
+def test_visit_graph_matmul():
+    # initialize context
+    ctx = OnnxXdslMapping()
+
+    # create graph composed only of one MatMul operation
+    graph, _ = _create_graph_binary_op("MatMul", "matmul_graph")
+
+    # run visit graph
+    visit_graph(graph, ctx)
```
superlopuh, line 198: can you please run visit_node instead, like in the other tests? Can you please also add a `.verify()` to make sure that the result is correct?

alecerio: Done

superlopuh: Thank you! I'm not sure we need this second test, the one above seems to be sufficient to me, am I missing something?

superlopuh (approved):

PR #2472, opened 2024-04-17
dialects: (llvm) Fix elem_type bug in GEPOp builder

Moxinilian:
There seems to have been a misunderstanding on the meaning of the element type in the llvm.getelementptr implementation in xDSL.


superlopuh (approved):
Any tests for this?

superlopuh:
Oh I see yeah there's probably no further testing needed

PR #2474, opened 2024-04-18
core: Implement MessageConstraint

PapyChacal:
This offers to expose a simple constraint wrapper, attaching a message to some IRDL constraint, enabling dialects to provide informative messages on the constraint's violation.

The constraint system is quite functional but not enabling user-friendly feedback: "This type was expected to be that type" is not as helpful as, i.e., "arith.addf expect both operands to have the same type".

NB: This is conservative, it keeps whatever errors were given before, only *adds* this context.


xdsl/irdl/irdl.py
```diff
@@ -381,6 +381,40 @@ def get_unique_base(self) -> type[Attribute] | None:
...
+            constr = EqAttrConstraint(constr)
+
+        self.constr = constr
+        self.message = message
+
+    def verify(self, attr: Attribute, constraint_vars: dict[str, Attribute]) -> None:
+        try:
+            return self.constr.verify(attr, constraint_vars)
+        except VerifyException as e:
+            raise VerifyException(e.args[0] + f"\n{self.message}", *e.args[1:])
```
superlopuh, line 409: Can you please add a comment explaining the indexing logic here?

superlopuh: Maybe add a check for e.args[0] always being present?

superlopuh (approved):
❤️ 🚀 🙏

xdsl/irdl/irdl.py
PapyChacal: Agreed on this code looking cryptic, OTOH this is simply the generic Exception. Could we maybe make our exception class itself more readable? :thinking:

superlopuh: I'll do it as a bit of a meditative exercise, I think the PR is ok to merge as-is

math-fehr (approved):
Nice! Can you remove the similar class that is in `builtin.py` currently?

PapyChacal:
> Nice! Can you remove the similar class that is in `builtin.py` currently?

Done, @superlopuh, it looks like you were using it, with the behaviour of *replacing* the current error. Do you want to add an option to this one for this?

cf https://github.com/xdslproject/xdsl/pull/2474/commits/49f1122826ceb0f8f66dc465643dabcbc62b284a#diff-1c55fe70b181fc520faa8ef4db978aa53daacef7ca579e1e8963eb1f233979fd

superlopuh:
Ah, I'm glad we caught this and can discuss. I think the best might be to take inspiration from Python's underlying error, and include both strings but put the explanation first. For this PR, it would probably be enough to invert the order of the strings, maybe putting the original message in parens. As you can see in the func message, the "empty array" is confusing/misleading.

PapyChacal:
> Ah, I'm glad we caught this and can discuss. I think the best might be to take inspiration from Python's underlying error, and include both strings but put the explanation first. For this PR, it would probably be enough to invert the order of the strings, maybe putting the original message in parens. As you can see in the func message, the "empty array" is confusing/misleading.

I like the reordering!
I want to avoid parenthesis, that sound like potentially more confusing to me when we start nesting!

tests/filecheck/dialects/func/func_invalid.mlir
```diff
@@ -36,4 +36,5 @@ builtin.module {
 
 "func.call"() { "callee" = @call::@invalid } : () -> ()
 
-// CHECK:  Operation does not verify: Unexpected nested symbols in FlatSymbolRefAttr.
+// CHECK:       Operation does not verify: Unexpected nested symbols in FlatSymbolRefAttr
+// CHECK-NEXT:  expected empty array, but got ["invalid"]
```
superlopuh, line 40:
```suggestion
// CHECK-NEXT:  Underlying verification failure: expected empty array, but got ["invalid"]
```
Maybe something like this to make it explicit?

PapyChacal: Done!

PR #2476, opened 2024-04-18
dialects: (fir) Update to the MLIR hash version of FIR that we use

mesham:
Attributes moved to properties, addition of the new FortranVariableFlagsAttr and new FIR Declare operation added

### DIFF
--- a/xdsl/dialects/experimental/fir.py
+++ b/xdsl/dialects/experimental/fir.py
@@ -4,6 +4,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from enum import Enum
 
+from xdsl.dialects.arith import FastMathFlagsAttr
 from xdsl.dialects.builtin import (
     AnyFloat,
     AnyIntegerAttr,
@@ -35,6 +36,7 @@ from xdsl.irdl import (
     VarOperand,
     VarOpResult,
     VarRegion,
+    attr_def,
     irdl_attr_definition,
     irdl_op_definition,
     operand_def,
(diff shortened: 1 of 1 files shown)


xdsl/dialects/experimental/fir.py
```diff
@@ -22,21 +35,94 @@
     VarOperand,
     VarOpResult,
     VarRegion,
-    attr_def,
     irdl_attr_definition,
     irdl_op_definition,
     operand_def,
-    opt_attr_def,
     opt_operand_def,
+    opt_prop_def,
```
superlopuh, line 42: I would recommend against `opt_prop_def` in general, I don't think MLIR has an equivalent, the equivalent is `MyAttr | NoneAttr` as opposed to Python's `None`

mesham: Is there an example of using prop_def for optional properties in the dialects? I ask because the following fails validation:

`linkName: StringAttr | NoneAttr = prop_def(StringAttr)
linkName: StringAttr  = prop_def(StringAttr | NoneAttr)
linkName: StringAttr | NoneAttr = prop_def(StringAttr | NoneAttr)`

Whereas opt_prop_def obviously does work for this optional property

superlopuh: My bad, I don't think we have the necessary functionality to support this right now, we'd first need default attributes

PapyChacal (approved):

xdsl/dialects/experimental/fir.py
```diff
@@ -2134,6 +2283,7 @@ class ZeroBits(IRDLOperation):
         ZeroBits,
     ],
     [
+        FortranVariableFlagsAttr,
         ReferenceType,
         DeferredAttr,
```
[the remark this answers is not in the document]
mesham: Thanks for that, very good spot - I'd omitted that by accident. Have now added it in :)

PR #2479, opened 2024-04-19
dialects: (onnx) Add onnx.Sigmoid

alecerio:
Implementation of Sigmoid onnx operator.

@superlopuh @compor


xdsl/dialects/onnx.py
```diff
@@ -932,6 +932,57 @@ def verify_(self) -> None:
...
+
+    Returns:
+    - output_tensor (TensorType): The output tensor after applying the sigmoid function element-wise to the input tensor.
+    """
+
+    name = "onnx.Sigmoid"
+
+    T = Annotated[AnyFloat | IntegerType, ConstraintVar("T")]
+    input_tensor = operand_def(TensorType[T])
+    output_tensor = result_def(TensorType[T])
```
superlopuh, line 955:
```suggestion
    T = Annotated[TensorType[AnyFloat], ConstraintVar("T")]
    input_tensor = operand_def(T)
    output_tensor = result_def(T)
```
Is it possible to do sigmoid of integers? It feels safe to only allow floats. The other thing is if the shapes and types are the same, then the types have to be the same, so we can change the declarative constraints, and remove the shape constraint below.

alecerio: Done

xdsl/dialects/onnx.py
```diff
@@ -1066,7 +1066,57 @@ def verify_(self) -> None:
...
+        input_tensor_shape = input_tensor_type.get_shape()
+        output_tensor_shape = output_tensor_type.get_shape()
+
+        # check if input tensor and output tensor have the same shape
+        if input_tensor_shape != output_tensor_shape:
+            raise VerifyException(
+                f"tensor input shape {input_tensor_shape} is not equal to tensor output shape {output_tensor_shape}"
+            )
 
+
```
superlopuh, line 1119:
```suggestion

```

PR #2480, opened 2024-04-21
core: Implement unqualified attributes in declarative syntax.

PapyChacal:
Those were currently using qualified attribute syntax (i.e., wrapping the parameters in the attribute's name).
This is not the intended behaviour: when parsing/printing a known attribute type, it should only print and parse its parameters.
The fully-qualified case is implemented upstream by an explicit wrapper (`qualified($attr)`) and could be added if needed; I see it is mostly used by the LLVM dialect upstream.

### DIFF
--- a/tests/filecheck/dialects/seq/seq_ops.mlir
+++ b/tests/filecheck/dialects/seq/seq_ops.mlir
@@ -1,4 +1,5 @@
 // RUN: XDSL_ROUNDTRIP
+// RUN: XDSL_GENERIC_ROUNDTRIP
 
 builtin.module {
   %clk = "test.op"() : () -> (!seq.clock)
@@ -18,6 +19,7 @@ builtin.module {
   // CHECK: %compreg_all = seq.compreg %data, %clk reset %bool, %data powerOn %on : i14
   %compreg_sym = seq.compreg sym @foo %data, %clk : i14
   // CHECK: %compreg_sym = seq.compreg sym @foo %data, %clk : i14
+  // CHECK-GENERIC:    %compreg_sym = "seq.compreg"(%data, %clk) {"inner_sym" = #hw<innerSym@foo>, "operandSegmentSizes" = array<i32: 1, 1, 0, 0, 0>} : (i14, !seq.clock) -> i14
 
   %const_low = seq.const_clock low
   // CHECK: %const_low = seq.const_clock low
--- a/tests/filecheck/dialects/seq/seq_ops.mlir
+++ b/tests/filecheck/dialects/seq/seq_ops.mlir
@@ -16,8 +16,8 @@ builtin.module {
   // CHECK: %compreg_poweron = seq.compreg %data, %clk powerOn %on : i14
   %compreg_all = seq.compreg %data, %clk reset %bool, %data powerOn %on : i14
   // CHECK: %compreg_all = seq.compreg %data, %clk reset %bool, %data powerOn %on : i14
-  %compreg_sym = seq.compreg sym #hw<innerSym@foo> %data, %clk : i14
-  // CHECK: %compreg_sym = seq.compreg sym #hw<innerSym@foo> %data, %clk : i14
+  %compreg_sym = seq.compreg sym @foo %data, %clk : i14
+  // CHECK: %compreg_sym = seq.compreg sym @foo %data, %clk : i14
 
   %const_low = seq.const_clock low
   // CHECK: %const_low = seq.const_clock low
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -9,9 +9,16 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-from typing import Literal, cast
+from typing import Any, Literal, cast
 
-from xdsl.ir import Attribute, SSAValue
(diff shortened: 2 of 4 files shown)


tests/filecheck/dialects/seq/seq_ops.mlir
```diff
@@ -16,8 +16,8 @@ builtin.module {
   // CHECK: %compreg_poweron = seq.compreg %data, %clk powerOn %on : i14
   %compreg_all = seq.compreg %data, %clk reset %bool, %data powerOn %on : i14
   // CHECK: %compreg_all = seq.compreg %data, %clk reset %bool, %data powerOn %on : i14
-  %compreg_sym = seq.compreg sym #hw<innerSym@foo> %data, %clk : i14
-  // CHECK: %compreg_sym = seq.compreg sym #hw<innerSym@foo> %data, %clk : i14
```
lucjaulmes, line 20: I’m sure we’ve seen this syntax grepping through circt test outputs. Do you mind if we keep both variants in tests?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -708,27 +722,50 @@ class AttributeVariable(FormatDirective):
     The directive will request a space to be printed right after.
     """
 
-    attr_name: str
+    name: str
     """The attribute name as it should be in the attribute or property dictionary."""
     is_property: bool
     """Should this attribute be put in the attribute or property dictionary."""
+    unique_base: type[Attribute] | None
```
lucjaulmes, line 729: Maybe a docstring here?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -738,29 +775,6 @@ class OptionalAttributeVariable(AttributeVariable, OptionalVariable):
...
-        else:
-            attr = op.attributes.get(self.attr_name)
-        if attr:
-            printer.print_attribute(attr)
-            state.should_emit_space = True
-            state.last_was_punctuation = False
-
-    def is_present(self, op: IRDLOperation) -> bool:
-        return getattr(op, self.attr_name) is not None
-
```
lucjaulmes, line 763: Not sure I understand how this is related, just incidentally removing duplicate code?

PapyChacal: Exactly!

xdsl/irdl/declarative_assembly_format.py
PapyChacal: Done!

tests/filecheck/dialects/seq/seq_ops.mlir
PapyChacal: Looking at CIRCT tests, I see this new syntax, i.e.:
https://github.com/llvm/circt/blob/main/test/Dialect/Seq/compreg.mlir

Now, the previous syntax is the *qualified* one, that is, the one you should see for example in generic syntax.

I checked on similar situations against mlir-opt - which is using the same declarative syntax code - and it's only accepting the unqualified one when expecting it, not both - that would be extra tricky to get right!

I don't want to go over you on your dialect though; can you tell me if it solves the question for you?

lucjaulmes: You‘re right – thanks for adding both syntaxes in checks, helpful.

lucjaulmes (approved):

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -708,27 +722,51 @@ class AttributeVariable(FormatDirective):
     The directive will request a space to be printed right after.
     """
 
-    attr_name: str
+    name: str
     """The attribute name as it should be in the attribute or property dictionary."""
     is_property: bool
     """Should this attribute be put in the attribute or property dictionary."""
+    unique_base: type[Attribute] | None
```
math-fehr, line 729: Instead of this being the unique base, even for builtin attributes, I would set this to `None` for builtin attributes.
I think that this design would make more sense, what do you think?
It will also solve later the `qualified` directive, where we will not want the unique base, even though it's known.

I would just change this and rename the docstring to
```
"""
The elided attribute base, if any.
When this value is set, the attribute name is not printed nor parsed.
For instance, `!test.attr<param>` would be parsed and printed as `<param>`.
"""
```

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -369,10 +371,18 @@ def parse_optional_variable(
             match attr_def:
                 case OptAttributeDef() | OptPropertyDef():
                     return OptionalAttributeVariable(
-                        variable_name, attr_or_prop == "property"
+                        variable_name,
+                        attr_or_prop == "property",
+                        attr_def.constr.get_unique_base(),
```
math-fehr, line 376: I would add here the logic about the builtin dialect (and later qualified directive).
probably with a function like `should_qualify_attribute` or something.

math-fehr (approved):
Really cool! I just have a nit comment on where to move some code, but otherwise perfect! Thanks a lot!

xdsl/irdl/declarative_assembly_format_parser.py
PapyChacal: If we're fine for a while without something like `should_qualify_attribute`, I'd be happy to avoid it and just handle the same builtin exception here for now. WDYT?

PR #2481, opened 2024-04-21
core: Remove redundant output from IRDLOperation's verification error.

PapyChacal:
Minor refactor of error reporting, removing superfluous intermediary messages from IRDL.
More importantly, this was capturing the error thrown by constraints, forcing users to scroll back to the underlying exception; this enables to directly print messages as seen in:
- #2482

### DIFF
--- a/docs/irdl.ipynb
+++ b/docs/irdl.ipynb
@@ -971,7 +971,8 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Operation does not verify: arith.addi32 operation does not verify\n",
+      "Operation does not verify: result at position 0 does not verify!\n",
+      "Expected attribute i32 but got i64\n",
       "\n",
       "%0 = \"arith.addi32\"(%1, %1) : (i32, i32) -> i64\n",
       "\n",
@@ -1042,7 +1043,8 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Operation does not verify: binary_op operation does not verify\n",
+      "Operation does not verify: result at position 0 does not verify!\n",
+      "attribute i32 expected from variable 'T', but got i64\n",
       "\n",
       "%0 = \"binary_op\"(%1, %1) : (i32, i32) -> i64\n",
       "\n",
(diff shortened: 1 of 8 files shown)


docs/ret.mlir
superlopuh, line 1: what's the purpose of this file?

math-fehr: We should add this file to our gitignore, this is generated from a notebook I believe.

tests/test_ir.py
```diff
@@ -851,10 +851,9 @@ def test_op_custom_verify_is_done_last():
     a = Constant.from_int_and_width(1, i32)
     # CustomVerify expects a i64, not i32
     b = CustomVerify.get(a.result)
-    with pytest.raises(Exception) as e:
+    with pytest.raises(VerifyException) as e:
```
math-fehr, line 854: Since we are refactoring this a bit:
```suggestion
    with pytest.raises(VerifyException, matches="Custom Verification Check") as e:
```

math-fehr (approved):
Very nice!

docs/irdl.ipynb
```diff
@@ -971,7 +971,8 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Operation does not verify: arith.addi32 operation does not verify\n",
+      "Operation does not verify: result at position 0 does not verify!\n",
```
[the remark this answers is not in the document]
PapyChacal: Cool, wanted to be chill with applying my opinions there, but I share this one :wink: 
Removing now

docs/ret.mlir
PapyChacal: Done

tests/test_ir.py
PapyChacal: The following assert checks that this is ***not*** present in the error messages (Testing that another VerifyException is raised before that one)
I'm not sure how to best refactor this with pytest TBH

docs/irdl.ipynb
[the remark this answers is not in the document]
PapyChacal: Since we're nitting: I'm in favor of a semicolon, since it's followed by a newline and the underlying reason, WDYT?

PR #2482, opened 2024-04-21
core: enable generic coercion of ConstraintVars in AttributeConstraint.

PapyChacal:
This enables nesting those ConstraintVars to enforce, for example, equal nested parameters in operation constructs without enforcing more on the other parameters.

### DIFF
--- a/tests/irdl/test_operation_definition.py
+++ b/tests/irdl/test_operation_definition.py
@@ -284,8 +284,13 @@ def test_nested_constraint_var():
         VerifyException,
         match=re.escape(
             """\
-Operation does not verify: Expected operand and result to have the same shape.
+Operation does not verify: result at position 0 does not verify!
+Expected operand and result to have the same shape.
 Underlying verification failure: attribute [#builtin.int<1>, #builtin.int<2>] expected from variable 'S', but got [#builtin.int<3>, #builtin.int<4>]
+
+%0 = "test.nested_constraint_var_op"(%1) : (memref<1x2xindex>) -> memref<3x4xindex>
+
+
 
 %0 = "test.nested_constraint_var_op"(%1) : (memref<1x2xindex>) -> memref<3x4xindex>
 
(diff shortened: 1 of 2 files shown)


tests/irdl/test_operation_definition.py
```diff
@@ -230,6 +235,96 @@ def test_constraint_var_fail_not_satisfy_constraint():
...
+Operation does not verify: result at position 0 does not verify!
+Expected operand and result to have the same shape.
+Underlying verification failure: attribute [#builtin.int<1>, #builtin.int<2>] expected from variable 'S', but got [#builtin.int<3>, #builtin.int<4>]
+
+%0 = "test.nested_constraint_var_op"(%1) : (memref<1x2xindex>) -> memref<3x4xindex>
+
+
+
+%0 = "test.nested_constraint_var_op"(%1) : (memref<1x2xindex>) -> memref<3x4xindex>
+
```
math-fehr, line 296: I would remove this from the check

tests/irdl/test_operation_definition.py
```diff
@@ -230,6 +235,96 @@ def test_constraint_var_fail_not_satisfy_constraint():
...
+Operation does not verify: result at position 0 does not verify!
+Expected operand and result to have the same shape.
+Underlying verification failure: attribute [#builtin.int<3>, #builtin.int<4>] expected from variable 'S', but got [#builtin.int<1>, #builtin.int<2>]
+
+%0 = "test.nested_constraint_var_op"(%1) : (memref<3x4xindex>) -> memref<1x2xindex>
+
+
+
+%0 = "test.nested_constraint_var_op"(%1) : (memref<3x4xindex>) -> memref<1x2xindex>
+
```
math-fehr, line 318: Same here

xdsl/irdl/irdl.py
```diff
@@ -224,13 +225,9 @@ def attr_constr_coercion(
     Attributes are coerced into EqAttrConstraints,
```
math-fehr, line 225: Should we just delete this function, and replace it with `irdl_to_attr_constraint`?

xdsl/irdl/irdl.py
```diff
@@ -458,7 +454,6 @@ def _irdl_list_to_attr_constraint(
 def irdl_to_attr_constraint(
     irdl: Any,
     *,
-    allow_type_var: bool = False,
```
math-fehr, line 461: Why did you remove this? The reason I added this was because `isa(x, type_variable)` is a dangerous thing, and doesn't really mean anything if you don't have the "context" of the type variable.

math-fehr (commented):
Good for me, but I just don't understand why you removed the "allow_type_var" option?
I would argue that we should keep it for now, unless you encountered a case where this was needed?

xdsl/irdl/irdl.py
PapyChacal: Probably!

PapyChacal:
> Good for me, but I just don't understand why you removed the "allow_type_var" option?
> I would argue that we should keep it for now, unless you encountered a case where this was needed?

As discussed, this isn't clear at all, and I'll have a close look later on!
This just worked and there's probably a subtler way

PapyChacal:
Closing for now as I think I was mostly confusing some layers here;
I can actually do that already, and what I tried to enable is more about dialect-side sugar - which would be welcome but most likely done otherwise!

Issue #2466, opened 2024-04-16
dialects: (x86) remove the Single/Double/TripleOperandInstruction parent classes


KGrykiel:
the Single/Double/TripleOperandInstruction classes in the x86 dialect are unnecessary, so we should just get rid of them

PR #2487, opened 2024-04-22
dialects: (x86) - Reformats and changes in naming convention

KGrykiel:
I've done some changes to the naming conventions following the suggestions of @tarinduj. This includes removing the single/doubleOperand parent classes (issue #2466), writing the class names in the format [Result]-[Operands]-Operation, and changing the individual names of the operations to prepare for the different configurations those instructions can exist in. The prefix signifies the type of the operand passed into the assembly instruction and is removed during assembly emission.

### DIFF
--- a/tests/filecheck/dialects/x86/x86_assembly_emission.mlir
+++ b/tests/filecheck/dialects/x86/x86_assembly_emission.mlir
@@ -4,23 +4,23 @@
 %1 = x86.get_register : () -> !x86.reg<rdx>
 %rsp = x86.get_register : () -> !x86.reg<rsp>
 
-%add = x86.add %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
+%add = x86.rr_add %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
 // CHECK: add rax, rdx
-%sub = x86.sub %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
+%sub = x86.rr_sub %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
 // CHECK: sub rax, rdx
-%imul = x86.imul %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
+%imul = x86.rr_imul %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/x86/x86_assembly_emission.mlir
```diff
@@ -4,23 +4,23 @@
 %1 = x86.get_register : () -> !x86.reg<rdx>
 %rsp = x86.get_register : () -> !x86.reg<rsp>
 
-%add = x86.add %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
+%add = x86.rr_add %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
```
superlopuh, line 7: Why this change? I thought the idea was to just change the abstract base class names? Since the operands and results are clearly denoted already, I'm not sure what this mnemonic adds

KGrykiel: Yeah, it's for the different configurations, all of the operations in the current version have several different configurations. Here's an example snippet from a matrix multiplication program that mixes them up a bit:

```
%20 = x86.rm_add %19, %rdx, 12: (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
x86.mr_mov %rdx, %20, 12 : (!x86.reg<rdx>, !x86.reg<rax>) -> ()
%21 = x86.rm_mov %17, %rsi, 12: (!x86.reg<rcx>, !x86.reg<rsi>) -> !x86.reg<rcx>
%22 = x86.rm_imul %21, %rdi, 12: (!x86.reg<rcx>, !x86.reg<rdi>) -> !x86.reg<rcx>
%23 = x86.rr_add %22, %20: (!x86.reg<rcx>, !x86.reg<rax>) -> !x86.reg<rcx>
x86.mr_mov %rdx, %23, 12 : (!x86.reg<rdx>, !x86.reg<rcx>) -> ()
```

which corresponds to:

```
add rax, [rdx + 12]
mov [rdx + 12], rax
mov rcx, [rsi + 12]
imul rcx, [rdi + 12]
add rcx, rax
mov [rdx + 12], rcx
```

superlopuh: In that case, for the time being, I'd probably recommend making a small change to `x86.rm.add` instead of the underscore, the lexer in my head breaks on `.` more easilty than `_`, hopefully making the IR a bit more readable. It should also make it easier to handle the assembly instruction printing, it'll be the last component in the name split on `.`

PR #2492, opened 2024-04-23
core: Handling true and false literals in DenseBaseAttr

mesham:
This PR firstly will enabling parsing of true and false string literals of type i1 to 1 and 0 respectively for DenseBaseAttr (previously true or false strings resulted in an error, but MLIR can output these as type i1).

This also works around an issue with MLIR, where MLIR will segfault in DenseArrayElementParser when it parses 0 and 1 as type i1, as it expects the false and true strings instead.

### DIFF
--- a/xdsl/parser/base_parser.py
+++ b/xdsl/parser/base_parser.py
@@ -350,7 +350,7 @@ class BaseParser:
             "Expected integer literal" + context_msg,
         )
 
-    def parse_optional_number(self, allow_boolean: bool = False) -> int | float | None:
+    def parse_optional_number(self, *, allow_boolean: bool = False) -> int | float | None:
         """
         Parse a (possibly negative) integer or float literal, if present.
         Can optionally parse 'true' or 'false' into 1 and 0.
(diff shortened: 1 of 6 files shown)


math-fehr:
I just added a commit on top of this to add this feature directly to `parse_number`.
Sorry about that, I didn't realized that the API was that broken for your use case!

math-fehr (approved):

xdsl/parser/base_parser.py
```diff
@@ -350,8 +350,11 @@ def parse_integer(
             "Expected integer literal" + context_msg,
         )
 
-    def parse_optional_number(self) -> int | float | None:
-        """Parse a (possibly negative) integer or float literal, if present."""
+    def parse_optional_number(self, allow_boolean: bool = False) -> int | float | None:
```
superlopuh, line 353: nit:
```suggestion
    def parse_optional_number(self, *, allow_boolean: bool = False) -> int | float | None:
```
To make the `allow_boolean` kw mandatory

superlopuh (approved):

PR #2497, opened 2024-04-25
ci: use a separate workflow for codecov upload

superlopuh:
My understanding is that actions that are triggered by the workflow_run action are run with secrets enabled.

### DIFF
--- a/.github/workflows/ci-codecov.yml
+++ b/.github/workflows/ci-codecov.yml
@@ -1,0 +1,32 @@
+# This workflow will install MLIR, Python dependencies, run tests and lint with a single version of Python
+
+name: CI - Codecov
+
+on:
+  # Trigger the workflow after ci-mlir completes
+  workflow_run:
+    workflows: [CI - Python-based Testing]
+    types:
+      - completed
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        path: xdsl
+
+    - name: Upload coverage to Codecov
+      if: matrix.python-version == '3.10'
+      uses: Wandalen/wretry.action@v1
+      with:
+        action: codecov/codecov-action@v4
+        attempt_delay: 10000
+        attempt_limit: 10
+        with: |
+          fail_ci_if_error: true
+          verbose: true
+          files: coverage.xml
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-codecov.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Upload coverage to Codecov
+      uses: Wandalen/wretry.action@v1
+      with:
+        action: codecov/codecov-action@v4
+        attempt_delay: 10000
+        attempt_limit: 10
+        with: |
+          fail_ci_if_error: true
+          verbose: true
+          files: coverage.xml
```
superlopuh, line 27: I'm not entirely sure how this file gets passed around, I've done a bit of googling and it looks like it should "just work"

georgebisbas (approved):
We can merge and then rebase some PR to see how it behaves

.github/workflows/ci-codecov.yml
PapyChacal: I don't think so, the usual mechanism I know of is artifacts:
https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

.github/workflows/ci-codecov.yml
```diff
@@ -0,0 +1,28 @@
...
+    types: [completed]
+
+jobs:
+  upload-codecov:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        path: xdsl
```
PapyChacal, line 16: I don't think we need to checkout xDSL to upload the report

PapyChacal:
> It looks like the second workflow has to be in main for it to get triggered. Maybe the best thing to do would be to merge this PR (if it looks fine to everyone) and retry with a forked repo PR to see if it works.

I wouldn't block this but I want to give the alternative, as we usually religiously keep main intact: Just merge this on a fork's default branch, and it's good enough for GitHubs' criterion.
I'm guessing it will have issues with secrets not set on the fork and what not; but is good enough for testing the trigger mechanism :slightly_smiling_face:

superlopuh:
IDK, I feel like this is worth trying anyway, I'm hoping this will fix our own dependabot issues also

PR #2498, opened 2024-04-25
dialects: (hw) Revert HWModuleOp's usage of InnerRefNamespaceTrait

Moxinilian:
While upstream MLIR decouples SymbolTables from operations, this is not the case in xDSL. As such, whenever one wants to reuse symbol table functionality, they have to declare an operation as a SymbolTable op. This was done on HWModuleOp to reuse SymbolTable functionality for the InnerSymbolTable, but in contrast to how it is done in upstream CIRCT, this thus requires the InnerSymbolTable to be a regular SymbolTable. This in turn blocks regular symbol referencing outside of the InnerSymbolTable.

This PR reverts this change for now. This should be reimplemented once xDSL supports decoupled symbol tables (or InnerRefNamespaceTrait is updated accordingly).

### DIFF
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -235,7 +235,7 @@ class InnerRefUserOpInterfaceTrait(OpTrait):
 
 @dataclass(frozen=True)
 class InnerRefNamespaceTrait(OpTrait):
-    """Trait for operations defining a new scope for InnerRef’s. Operations with this trait must be a SymbolTable."""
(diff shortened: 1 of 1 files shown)


lucjaulmes (commented):
Can you provide a small example of when `HWModuleOp` being a `SymbolTable` is an issue? That way I can look at tweaking `InnerSymbolTable` later down the line.

lucjaulmes (approved):

Moxinilian:
Sure!

```mlir
hw.module @foo() {
    hw.output
}
hw.module @bar() {
    hw.instance "foo_inst" @foo() -> ()
}
```

PR #2500, opened 2024-04-25
dialects: stencil IRDL update.

PapyChacal:
Just an update to the stencil dialect properly using IRDL to enforce more properties declaratively, and prepare more type inference.


xdsl/dialects/stencil.py
```diff
@@ -683,8 +744,31 @@ class LoadOp(IRDLOperation):
     """
 
     name = "stencil.load"
-    field: Operand = operand_def(FieldType)
-    res: OpResult = result_def(TempType)
+
+    T = Annotated[Attribute, ConstraintVar("T")]
```
georgebisbas, line 748: Should we do some Mixin class to avoid those duplicate things?

PR #2502, opened 2024-04-25
dialects: (x86) PR7 - RI Operations

KGrykiel:
Added support for operations taking one register and an immediate value operand. Very similar to the previous PR.


xdsl/dialects/x86/ops.py
```diff
@@ -550,6 +551,144 @@ class RM_MovOp(RMOperation[GeneralRegisterType, GeneralRegisterType]):
...
+        _print_immediate_value(printer, self.immediate)
+        return {"immediate"}
+
+    def assembly_line(self) -> str | None:
+        instruction_name = self.assembly_instruction_name()
+        destination = _assembly_arg_str(self.r1)
+        immediate = _assembly_arg_str(self.immediate)
+        return _assembly_line(
+            instruction_name, f"{destination}, {immediate}", self.comment
+        )
```
superlopuh, line 612: do you need this? don't the `assembly_line_args` already specify the same behaviour?

KGrykiel: yeah, you're right, I removed it

superlopuh (approved):

PR #2507, opened 2024-04-26
core: Factor out visibility keyword parsing

Moxinilian:
This PR factors out the parsing of visibility keyword into core (like in MLIR) as visibility keywords are a property of the symbol infrastructure. While the symbol lookup in xDSL does not take it into account yet, it is useful to have it be part of the symbol infrastructure itself as those keywords are used in many places (for example, in func.func or hw.module).

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -853,3 +853,24 @@ def test_parse_location():
     ctx = MLContext()
     attr = Parser(ctx, "loc(unknown)").parse_optional_location()
     assert attr == LocationAttr()
+
+
+@pytest.mark.parametrize(
+    "keyword,expected",
+    [
+        ("public", StringAttr("public")),
+        ("nested", StringAttr("nested")),
+        ("private", StringAttr("private")),
+        ("privateeee", None),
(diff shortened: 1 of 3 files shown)


tests/test_parser.py
```diff
@@ -853,3 +853,24 @@ def test_parse_location():
...
+        ("unknown", None),
+    ],
+)
+def test_parse_visibility(keyword: str, expected: StringAttr | None):
+    assert Parser(MLContext(), keyword).parse_optional_visibility_keyword() == expected
+
+    parser = Parser(MLContext(), keyword)
+    if expected is None:
+        with pytest.raises(ParseError):
+            parser.parse_visibility_keyword()
```
PapyChacal, line 874: Can you add a `match=` to the `pytest.raises()`?
It might sound overkill but I find it neat to catch error-reporting regressions, which otherwise are just noticed months later by confused users and are tricky to trace back.

PapyChacal (approved):
Didn't even know that, wonderful, thanks!
Just a testing comment but good to go!

PR #2508, opened 2024-04-26
dialects: (hw) Add visibility keyword to hw.module

Moxinilian:
Depends on #2507


Moxinilian:
Ah I thought GitHub would be smarter about showing changes. Some of the changes are from #2507.

lucjaulmes (approved):

PR #2510, opened 2024-04-26
dialects: (x86) PR8 - MR Operations

KGrykiel:
Added support for operations that have one memory access and one register operand. The result is saved to the memory location.

### DIFF
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -713,7 +713,16 @@ class M_MR_Operation(Generic[R1InvT, R2InvT], IRDLOperation, X86Instruction, ABC
         )
 
     def assembly_line_args(self) -> tuple[AssemblyInstructionArg | None, ...]:
-        return self.r1, self.r2
+        destination = _assembly_arg_str(self.r1)
+        if self.offset is not None:
+            offset = _assembly_arg_str(self.offset)
+            if self.offset.value.data > 0:
+                destination = f"[{destination}+{offset}]"
+            else:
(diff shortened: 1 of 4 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -681,6 +681,144 @@ class RI_MovOp(R_RI_Operation[GeneralRegisterType]):
...
+                    self.comment,
+                )
+            else:
+                return _assembly_line(
+                    instruction_name, f"[{destination}{offset}], {source}", self.comment
+                )
+        else:
+            return _assembly_line(
+                instruction_name, f"[{destination}], {source}", self.comment
+            )
```
superlopuh, line 753: Could you give a go at doing this via the assembly line args instead? The method you implemented above seems to have different behaviour, and will never be called.

KGrykiel: hm, not sure how I could represent the memory access format [register + offset] with just assembly_line_args. Also not sure what you mean by saying this method will never be called?

superlopuh: `assembly_line_args` is called in the superclass, to construct the assembly line. If `assembly_line` is overridden, then `assembly_line_args` is not called during assembly emission.

You can construct the memory access format into a string in the `assembly_line_args` method, and return that as one of the args, seems like it would be marginally neater.

KGrykiel:
@superlopuh is this what you had in mind?

superlopuh (approved):

PR #2512, opened 2024-04-29
dialects: (hw) Fix printing of module visibility

Moxinilian:
While the visibility test was working, the practical roundtripping of private modules did not work properly. The reason for this is that there was a typo in the definition of the visibility attribute, which was cancelled out by the fact that roundtriping goes through generic format!

This seems to be a more fundamental issue of the testing infrastructure, so I am not sure how to add more tests that would cover this. @math-fehr @superlopuh

### DIFF
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -42,7 +42,6 @@ from xdsl.irdl import (
     irdl_attr_definition,
     irdl_op_definition,
     opt_attr_def,
-    opt_prop_def,
     region_def,
     traits_def,
     var_operand_def,
@@ -619,6 +618,7 @@ _MODULE_OP_ATTRS_HANDLED_BY_CUSTOM_FORMAT: list[str] = [
     "sym_name",
     "module_type",
     "parameters",
+    "sym_visibility",
 ]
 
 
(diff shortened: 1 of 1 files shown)


PapyChacal:
Nice catch!
The only way to test generic syntax is to test it explicitly IMO.
I see those filechecks are using // RUN: `XDSL_ROUNDTRIP` and `// CHECK:`.
There are complementary helpers; you can just add a `// RUN: XDSL_GENERIC_ROUNDTRIP` and use `// CHECK-GENERIC:` wherever in the same file to test the generic syntax output specifically :slightly_smiling_face:

Moxinilian:
No i think you misunderstood the problem.
`XDSL_ROUNDTRIP` does `xdsl-opt %s --print-op-generic --split-input-file | xdsl-opt --split-input-file`. While `xdsl-opt %s | xdsl-opt` would have spotted the problem, this specific problem is such that doing `xdsl-opt --print-op-generic | xdsl-opt` hides it.

superlopuh (approved):

superlopuh:
I'm confused, which tests are broken, and what do they miss?

superlopuh:
Ah, indeed, I'd expect it to not print the generic syntax. When I change the macro locally one test fails, but I assume it's for a different reason. Maybe we can change the macro?

PR #2514, opened 2024-04-29
core: Implement HasAncestor.

PapyChacal:
Implement HasAncestor, the transitive cousin of HasParent.

### DIFF
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+from collections.abc import Iterator
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, TypeVar
 
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -77,10 +77,8 @@ class HasAncestor(OpTrait):
 
     parameters: tuple[type[Operation], ...]
 
-    def __init__(self, *parameters: type[Operation]):
-        if not parameters:
-            raise ValueError("parameters must not be empty")
-        super().__init__(parameters)
+    def __init__(self, head_param: type[Operation], *tail_params: type[Operation]):
+        super().__init__((head_param, *tail_params))
 
     def verify(self, op: Operation) -> None:
         if self.get_ancestor(op) is None:
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -68,6 +68,40 @@ class HasParent(OpTrait):
         raise VerifyException(f"'{op.name}' expects parent op to be one of {names}")
 
 
+@dataclass(frozen=True)
+class HasAncestor(OpTrait):
+    """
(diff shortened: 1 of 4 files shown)


xdsl/traits.py
```diff
@@ -68,6 +68,40 @@ def verify(self, op: Operation) -> None:
...
+    Constraint the operation to have a specific operation as ancestor, i.e. transitive
+    parent.
+    """
+
+    parameters: tuple[type[Operation], ...]
+
+    def __init__(self, *parameters: type[Operation]):
+        if not parameters:
+            raise ValueError("parameters must not be empty")
+        super().__init__(parameters)
```
superlopuh, line 83: One way to enforce this syntactically is to have a first param, and remaining params
```suggestion
    def __init__(self, head_param: type[Operation], *tail_parameters: type[Operation]):
        super().__init__((head_param, *parameters))
```
Seems like the tuple construction cost is minimal compared to the runtime error annoyance, but that's kind of a matter of taste :)

xdsl/traits.py
```diff
@@ -68,6 +68,40 @@ def verify(self, op: Operation) -> None:
...
+                f"'{op.name}' expects ancestor op to be one of {names}"
+            )
+
+    def get_ancestor(self, op: Operation) -> Operation | None:
+        parent = op.parent_op()
+        while not isinstance(parent, self.parameters):
+            if parent is None:
+                return None
+            parent = parent.parent_op()
+        return parent
```
superlopuh, line 102: Should this be a helper on operation? It has a very similar vibe to `walk` and `walk_backwards`, I think a `walk_ancestors` would make sense. To minimise the scope creep, I'd like to propose splitting the iteration from the predicate first, putting both helpers here, and then eventually migrate them.
```suggestion
    def walk_ancestors(self, op: Operation) -> Iterator[Operation]:
        """Iterates over the ancestors of an operation, including the input"""
        curr = op
        yield curr
        while (curr := curr.parent_op()) is not None:
            yield curr

    def get_ancestor(self, op: Operation) -> Operation | None:
        ancestors = self.walk_ancestors(op)
        matching_ancestors = (a for a in ancestors if isinstance(a, self.parameters))
        return next(matching_ancestors, None)
```

xdsl/traits.py
PapyChacal: Right, will do 🙂

xdsl/traits.py
PapyChacal: Why not!

PapyChacal: Will do but if anyone else disagrees and we start bikeshedding on core, I'm reversing to add functionality first :+1:

PapyChacal: Oh wait; didn't see your idea of splitting but still in traits. Perfect on the concern ahah

superlopuh (approved):

PR #2517, opened 2024-04-30
dialects: Implement helpers on stencil attributes and simplify shape inference.

PapyChacal:


xdsl/transforms/experimental/dmp/stencil_global_to_local.py
```diff
@@ -39,13 +39,13 @@ class ChangeStoreOpSizes(RewritePattern):
...
-            integer_attr.data == 0 for integer_attr in op.lb.array.data
+            integer_attr.data == 0 for integer_attr in op.bounds.lb.array.data
         ), "lb must be 0"
         shape: tuple[int, ...] = tuple(
-            integer_attr.data for integer_attr in op.ub.array.data
+            integer_attr.data for integer_attr in op.bounds.ub.array.data
         )
         new_shape = self.strategy.calc_resize(shape)
-        op.ub = stencil.IndexAttr.get(*new_shape)
+        op.bounds.ub = stencil.IndexAttr.get(*new_shape)
```
superlopuh, line 48: I'm surprised this works, aren't Attributes supposed to be immutable? I thought we had made them frozen

superlopuh (approved):

xdsl/transforms/experimental/dmp/stencil_global_to_local.py
PapyChacal: Well, you use the frozen dataclass constructor in irdl_attr_definition, but AFAIU this is just a typechecker functionality, and AFAIU pyright has no clue that irdl_attr_definition returns such a thing currently. Not sure if it's doable after a few tentative!
Anyway, made it into a more frozen style now.

PR #2526, opened 2024-05-02
transformations: stencil conversion fixes and improvements

PapyChacal:
Some improvements on the conversion:
- Stop relying on explicit loop indices for accesses and returns; `stencil.index` is modelling just that, use it instead as a proxy.
- `stencil.apply` now simply creates its own loop indices and replaces `stencil.index`s appropriately.
- split `stencil.store` in its own pattern rather than a loop, didn;t serve any purpose at this point - resulting in the minor recording of subviews.
- Accessorily fix a corner-case bug when not cleaning up the unused cloned op in storage materialization.

As can be seen on the tests, it's a nonfunctional change on whatever was tested, but the current passes did not work on a proper open earth compiler kernel. It now does :tada:

### DIFF
--- a/xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
+++ b/xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
@@ -1,7 +1,6 @@
-from collections.abc import Iterable
 from dataclasses import dataclass
 from itertools import product
-from typing import TypeVar, cast
+from typing import TypeVar
 from warnings import warn
 
 from xdsl.dialects import arith, builtin, memref, scf
@@ -9,7 +8,6 @@ from xdsl.dialects.builtin import (
     MemRefType,
     UnrealizedConversionCastOp,
 )
-from xdsl.dialects.func import FuncOp
 from xdsl.dialects.stencil import (
     AccessOp,
     ApplyOp,
@@ -97,7 +95,7 @@ def collectBlockArguments(number: int, block: Block):
 
 
 def update_return_target(
-    return_targets: dict[ReturnOp, list[SSAValue | None]],
+    return_targets: dict[ApplyOp, list[SSAValue | None]],
     old_target: SSAValue,
     new_target: SSAValue | None,
 ):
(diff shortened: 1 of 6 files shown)


xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
```diff
@@ -349,15 +325,15 @@ def match_and_rewrite(self, op: ApplyOp, rewriter: PatternRewriter, /):
...
+            unroll = list(u for u in unroll)
 
-        dim = res_type.get_num_dims()
-        body = prepare_apply_body(op, rewriter, dim)
-        body.block.add_op(scf.Yield())
+        rank = op.get_rank()
+        body = prepare_apply_body(op)
         if unroll is None:
-            unroll = [1] * dim
+            unroll = [1] * rank
```
georgebisbas, line 336: why rank?

xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
```diff
@@ -340,7 +316,7 @@ def match_and_rewrite(self, op: BufferOp, rewriter: PatternRewriter, /):
 
 @dataclass
 class ApplyOpToParallel(RewritePattern):
-    return_targets: dict[ReturnOp, list[SSAValue | None]]
+    return_targets: dict[ApplyOp, list[SSAValue | None]]
```
georgebisbas, line 319: why this change?

PapyChacal: It just made things simpler in one of the pattern :)
They are sort of interchangeable, as one is the only terminator of the other, they have to exist in unique pairs!

xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
PapyChacal: It is the number of dimensions, which I'm used to calling rank, and I believe is the common term in MLIR too!

georgebisbas (approved):
Needs rebase and GTG

PR #2531, opened 2024-05-03
dialects: (x86) PR13 - M Operations

KGrykiel:
Implemented some operations taking only one memory access as operand. I decided to make 'push' into its own function as opposed to the way I did it with the register version as I don't think there's any other method that would inherit from that. I will probably reformat the register version as well in a PR later.


KGrykiel:
I am not sure how to tackle the M version of the 'pop' operation. Since it only has a destination that is a memory access, there is no registers involved other than RSP. Where would the information of what the memory access should be come from in the IR? I can only think of something like;
`x86.m.pop %rsp : (!x86.reg<rsp>) -> (!x86.reg<rsp>)`
but that's clearly lacking info.

PR #2532, opened 2024-05-03
transformations: add malloc to riscv function call lowering

superlopuh:
This seems to be similar to what `--finalize-memref-to-llvm` does.

### DIFF
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -32,12 +32,32 @@ from xdsl.traits import SymbolTable
 from xdsl.utils.exceptions import DiagnosticException
 
 
+def bitwidth_of_type(type_attribute: Attribute) -> int:
+    """
+    Returns the width of an element type in bits, or raises ValueError for unknown inputs.
+    """
+    match type_attribute:
+        case IntegerType():
+            bitwidth = type_attribute.width.data
+            return bitwidth
+        case Float32Type():
+            return 32
+        case Float64Type():
+            return 64
+        case _:
+            raise NotImplementedError(
+                f"Unsupported memref element type for riscv lowering: {type_attribute}"
+            )
+
+
 class ConvertMemrefAllocOp(RewritePattern):
 
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: memref.Alloc, rewriter: PatternRewriter) -> None:
         assert isinstance(op_memref_type := op.memref.type, memref.MemRefType)
-        size = prod(op_memref_type.get_shape()) * 8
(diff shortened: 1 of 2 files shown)


xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -26,13 +28,29 @@
...
 from xdsl.utils.exceptions import DiagnosticException
 
 
 class ConvertMemrefAllocOp(RewritePattern):
+
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: memref.Alloc, rewriter: PatternRewriter) -> None:
-        raise DiagnosticException("Lowering memref.alloc not implemented yet")
+        assert isinstance(op_memref_type := op.memref.type, memref.MemRefType)
+        size = prod(op_memref_type.get_shape()) * 8
```
AntonLydike, line 40: I think you can't assume that the types size is 8 bytes long, in the example code above you have an i32, which is only 4 bytes long, while an i128 might be more than 8 bytes.

I think supporting this in general will require us to add a way to get the size of a type. I think @Moxinilian worked a bit on this.

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -264,10 +282,12 @@ class ConvertMemrefToRiscvPass(ModulePass):
     name = "convert-memref-to-riscv"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
+        contains_malloc = PatternRewriteWalker(ConvertMemrefAllocOp()).rewrite_module(
+            op
+        )
```
AntonLydike, line 287: Nit: I think I would prefer to track "inserted malloc op" as an attribute on the rewrite for slightly cleaner code. Especially since we will need to repeat this for free and realloc.

AntonLydike (commented):
Thanks for adding this! I think this would work as a first prototype, but I would change a few things to at least assert that you are on a "happy code path" instead of the current behavior.

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: We have this! I'll add it now

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: Agree to disagree :)

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: Turns out we didn't have this, or I couldn't find it

AntonLydike (approved):
:rocket:

PR #2536, opened 2024-05-04
interpreter: use function type when parsing xdsl-run arguments

superlopuh:
There was an aspect of the interpreter that I didn't account for when opening the original PR that allowed passing function arguments, which is that different levels of abstractions use different runtime representations of the same thing. Notably, buffers are represented by `ShapedArray` for values of type `MemrefType`, and with `RawPtr` for RISC-V register values. This PR adds an extra parameter to the `impl_attr`-annotated impls, and changes the interpreter logic to use those to find the implementation, instead of the type of the parameter passed via the command line.

Note stacked PR


xdsl/interpreters/builtin.py
```diff
@@ -53,23 +54,41 @@ def run_cast(
...
+    def float32_attr_value(
+        self, interpreter: Interpreter, attr: Attribute, attr_type: Float32Type
+    ) -> float:
+        interpreter.interpreter_assert(isa(attr, AnyFloatAttr))
+        attr = cast(AnyFloatAttr, attr)
         return attr.value.data
 
-    @impl_attr(IntegerAttr)
+    @impl_attr(IntegerType)
     def integer_attr_value(
```
PapyChacal, line 74: Should this be `integer_type_value`/`integer_type_attr_value` then?

PapyChacal (approved):
I'm not a fan 

Edit: Oopsie, started a message and figured it was not helping, ended up posting the least helpful bit :sweat_smile:

xdsl/interpreters/builtin.py
superlopuh: At the end of the day, it's kind of equivalent, since it can only handle integer attrs

PR #2537, opened 2024-05-04
dialects: (x86) - mem access helper

KGrykiel:
As per suggestion, I made a helper function for the memory accesses


superlopuh (approved):
Thank you!

PR #2539, opened 2024-05-05
dialects: (x86) PR15 - S Operations

KGrykiel:
Added support for S operations and the unconditional jump. I'll do conditional jumps in a seperate PR as they would involve creating a new register type.


superlopuh:
Ok now we're getting to the juicy part of representing assembly control flow in xDSL.

I'm still of the opinion that we should always follow the MLIR invariant that if a region is in SSA, then when a block is entered all the instructions are executed. This jump, for me, should be a terminator, as it is in the `riscv_cf` dialect. I still think we should remove the non-terminator jumps from the riscv dialect and replace them with the `riscv_cf` ops instead. The criticism there is that we might not be able to represent dynamic jumping in our IR in that case. The question for me is whether x86 allows dynamic jumps, and whether we want to be able to represent them. A jump to a label can be statically reasoned about, so there we could definitely model it as a terminator op with a successor. Are you planning to add jumps with dynamic offsets?

KGrykiel:
I haven't really thought as far as dynamic offsets. My goal for now is to implement a complete basic version that would handle most simple cases and then potentially develop it further based on people's needs. I suppose it's best to keep to static jumps for now?

superlopuh:
In that case, I would recommend adding the equivalents to the `riscv_cf` instructions to x86, with successors and block arguments

webmiche:
> I'm still of the opinion that we should always follow the MLIR invariant that if a region is in SSA, then when a block is entered all the instructions are executed. This jump, for me, should be a terminator, as it is in the `riscv_cf` dialect. I still think we should remove the non-terminator jumps from the riscv dialect and replace them with the `riscv_cf` ops instead. The criticism there is that we might not be able to represent dynamic jumping in our IR in that case. The question for me is whether x86 allows dynamic jumps, and whether we want to be able to represent them. A jump to a label can be statically reasoned about, so there we could definitely model it as a terminator op with a successor. Are you planning to add jumps with dynamic offsets?

Big :+1: from me. Note that the MLIR community has been discussing this lately (I remember a talk at eurollvm2023 by Jeff and the mojo people about this). So, let's see what they come up with to do the things they want.

KGrykiel:
@superlopuh @webmiche  Let me know if this is more or less what you had in mind. I didn't do assembly emission tests yet because those need to supposedly be in a block to work and I don't have that functionality yet. I kind of made a hack to make the other filecheck work in the first place.

tests/filecheck/dialects/x86/x86_ops.mlir
```diff
@@ -117,5 +117,13 @@ x86.directive ".align" "2"
...
+func.func @funcyasm() {
+    x86.s.jmp ^labelblock(%arg : !x86.reg<>)
+    // CHECK: x86.s.jmp ^{{.+}}(%arg : !x86.reg<>)
+    ^labelblock(%arg : !x86.reg<>):
+    // CHECK-NEXT: ^{{.+}}(%arg : !x86.reg<>):
+    x86.label "then"
+    // CHECK-NEXT: x86.label "then"
+    x86.s.jmp ^labelblock(%arg : !x86.reg<>)
+    // CHECK-NEXT: x86.s.jmp ^{{.+}}(%arg : !x86.reg<>)
+}
```
superlopuh, line 129:
```suggestion
}

```

superlopuh (approved):

PR #2540, opened 2024-05-05
transformations: (riscv_scf) add a pass to fuse perfectly nested loops

superlopuh:


AntonLydike (approved):
Does this pass consider data dependencies between the loops? I think it'd be better to fuse at a higher level (before we get memrefs even) such that we can infer this more easily.

AntonLydike:
Disregard my uninformed comment

xdsl/transforms/riscv_scf_loop_fusion.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if (inner_lb := get_constant_value(inner_loop.lb)) is None:
+            return
+        if inner_lb.value.data != 0:
+            return
+
+        if (inner_ub := get_constant_value(inner_loop.ub)) is None:
+            return
+        if (outer_step := get_constant_value(op.step)) is None:
+            return
```
AntonLydike, line 39: Why only fuse on constant bounds/step? Can't you emit instructions that calculate the new bounds?

tests/filecheck/dialects/riscv_scf/loop_fusion.mlir
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// CHECK-NEXT:    %{{.*}} = riscv.li 0 : () -> !riscv.reg<>
+// CHECK-NEXT:    %{{.*}} = riscv.li 1 : () -> !riscv.reg<>
+// CHECK-NEXT:    %{{.*}} = riscv.li 8 : () -> !riscv.reg<>
+// CHECK-NEXT:    %{{.*}} = riscv.li 64 : () -> !riscv.reg<>
+// CHECK-NEXT:    riscv_scf.for %{{.*}} : !riscv.reg<> = %{{.*}} to %{{.*}} step %{{.*}} {
+// CHECK-NEXT:      %{{.*}} = riscv.li 8 : () -> !riscv.reg<>
+// CHECK-NEXT:      "test.op"(%{{.*}}) : (!riscv.reg<>) -> ()
+// CHECK-NEXT:    }
```
AntonLydike, line 26: I think here it is critical that the filecheck actually tests the ssa variable names.

AntonLydike (approved):
Thanks for this really neat rewrite! This is very interesting! I would really like to see some more comments and documentation on the conditions and calculations (e.g. it has to be a nested for loop where outer_step = len(inner_range), inner_lb == 0, etc...)

On a side note, I am not sure if I would call this loop fusion. For me it implies that two loop bodies are merged into one. Isn't this closer to a canonicalization? It does not change any actual iteration domains, just simplifies IR.

AntonLydike (approved):
Thanks!

PR #2542, opened 2024-05-06
transformations: Implement and test `memref-to-gpu`.

PapyChacal:
This new pass converts all `memref.(de)alloc`s to `gpu.(de)alloc`s.

The motivation is that my stencil conversion is now completely target-agnostic, except for allocations, for which there currently exists no way to express transparently and select a platform later, for example.

It is very rough at this point; i.e., could extend to more targets, or to more settings, but is good enough for what I want to do with it at the moment :slightly_smiling_face:


superlopuh (approved):

georgebisbas (commented):
Maybe memref-to-gpu  or sth similar?

PR #2543, opened 2024-05-06
transformations: handle scalar arguments to generic ops when lowering

superlopuh:
We need this for fill


PapyChacal (approved):
IIRC this is valid in MLIR, so not a question about your port but more OOC on the initial rationale, if you know or have an opinion;

Why even struggle with scalar arguments in the first place instead of just using it from outside the `linal.generic` body? :thinking:

superlopuh:
Not sure :)

PR #2545, opened 2024-05-07
core: Fix xDSL printing multiple values with the same name.

PapyChacal:
Found this adorable little fundamental crack while working on the OEC kernels.

xDSL is currently still able to print distinct values using the same name; which is a big no no no.

## Headlines:

Basically our brave printer, willing to honor user-friendly name hints given to values, deduplicates value names when a given name_hint clashes with a previously printed name, by adding a `_1`, `_2`, or ... to the name.

Func fact is, if the printer already printed `%v` twice, i.e., actually printed them as `%v` and `%v_1`, it only sees `%v_1` printed as a deduplication of `v`. So if the next value is hinted as `v_1`, then no problem seen, and the printer will print the three ***distinct*** values as `%v`, `%v_1`, and `%v_1` again :warning: 

Here is @math-fehr 's solution: just strip all suffixes in the form of underscore+number from name_hints, always. The rationale being, they would be printed with those again, in order, anyway.
More importantly, this makes xDSL unable (to the best of our knowledge again :smiling_face_with_tear:) to print invalid IR :tada:

### DIFF
--- a/tests/filecheck/backend/riscv/riscv_register_allocation.mlir
+++ b/tests/filecheck/backend/riscv/riscv_register_allocation.mlir
@@ -42,8 +42,8 @@ riscv_func.func @main() {
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:        %{{\d+}} = riscv.mv %9 : (!riscv.reg<t0>) -> !riscv.reg<t0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:        riscv_scf.yield %{{\d+}} : !riscv.reg<t0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      }
-//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_0 = riscv.li 0 : () -> !riscv.reg<zero>
-//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_1 = riscv.li 0 : () -> !riscv.reg<a0>
+//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_1 = riscv.li 0 : () -> !riscv.reg<zero>
+//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_2 = riscv.li 0 : () -> !riscv.reg<a0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      riscv_func.return
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:    }
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:  }
@@ -64,8 +64,8 @@ riscv_func.func @main() {
 //   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:        %{{\d+}} = riscv.mv %9 : (!riscv.reg<j0>) -> !riscv.reg<j0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:        riscv_scf.yield %{{\d+}} : !riscv.reg<j0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:      }
-//   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:      %zero_0 = riscv.li 0 : () -> !riscv.reg<zero>
-//   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:      %zero_1 = riscv.li 0 : () -> !riscv.reg<a0>
+//   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:      %zero_1 = riscv.li 0 : () -> !riscv.reg<zero>
+//   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:      %zero_2 = riscv.li 0 : () -> !riscv.reg<a0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:      riscv_func.return
 //   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:    }
 //   CHECK-LIVENESS-BLOCK-NAIVE-J-NEXT:  }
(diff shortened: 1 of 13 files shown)


tests/filecheck/backend/riscv/riscv_register_allocation.mlir
```diff
@@ -42,8 +42,8 @@ riscv_func.func @main() {
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:        %{{\d+}} = riscv.mv %9 : (!riscv.reg<t0>) -> !riscv.reg<t0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:        riscv_scf.yield %{{\d+}} : !riscv.reg<t0>
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      }
-//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_0 = riscv.li 0 : () -> !riscv.reg<zero>
-//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_1 = riscv.li 0 : () -> !riscv.reg<a0>
+//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %zero_1 = riscv.li 0 : () -> !riscv.reg<zero>
```
georgebisbas, line 45: but why starting from 1 now?

tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -3,15 +3,15 @@
...
 
-"test.op"(%addf_0, %addf_vector_0) : (f32, vector<4xf32>) -> ()
+"test.op"(%addf, %addf_vector) : (f32, vector<4xf32>) -> ()
 
 // CHECK:        %lhsf32, %rhsf32 = "test.op"() : () -> (f32, f32)
 // CHECK-NEXT:   %lhsvec, %rhsvec = "test.op"() : () -> (vector<4xf32>, vector<4xf32>)
-// CHECK-NEXT:   %addf_0 = arith.addf %lhsf32, %rhsf32 : f32
-// CHECK-NEXT:   %addf_vector_0 = arith.addf %lhsvec, %rhsvec : vector<4xf32>
-// CHECK-NEXT:   "test.op"(%addf_0, %addf_vector_0) : (f32, vector<4xf32>) -> ()
+// CHECK-NEXT:   %addf = arith.addf %lhsf32, %rhsf32 : f32
```
georgebisbas, line 15: hm

PapyChacal: What do you mean on this one?

tests/filecheck/backend/riscv/riscv_register_allocation.mlir
PapyChacal: Mmmh
We didn't change that logic; so it raises the question "how did it start at zero before" on my side :upside_down_face:

PapyChacal: Oooh no got it
The input has `%zero_0`, hence the name_hint used to be `zero_0` and printed back as such.
This is such a case where any other value name `zero` added in that IR would trigger the described bug.

PapyChacal: FWIW, I just tried offsetting the deduplication to start at 0 instead, it just makes even more filechecks fail, meaning more meaningless changes like this one in the same PR :upside_down_face:

PapyChacal: Tried starting from 0 instead, it just have more meaningless impacts like this on tests overall!

tests/filecheck/dialects/arith/canonicalize.mlir
georgebisbas: That now,we do not start from zero if one occurrence is there.


I would expect that:

if one occurrence exists: addf
if multiple: addf_0, addf_1, addf_2

PapyChacal: As in, never printing the first one and directly starting at 0?
That would pretty costly, we would need to check ahead!
I'd also like to steer discussion: If we want to discuss better styles of doing it, can we maybe open an issue and discuss there? To keep this PR about the correct *behaviour*; I see it as relatively critical, as this currently is able to just parse valid IR and print it directly as invalid IR.

PapyChacal: But yeah, the choice of this fix is to just strip any meaning from `_[0-9]+` suffixes in value names. That's a super efficient one, both to ensure it's safe and at low cost.
Wasn't my first idea either, but I like it for those reasons.
Note that, you can still do fancy value naming, with anything other than numbers.

PS: Oh no. just remembered how much we like our numeric suffixes in DeVito :smiling_face_with_tear:. I would rather have a workaround there (We can add a suffix after the number  :shrug: ) and have xDSL core be correct before all

superlopuh (commented):
I was waiting for this to happen :) what's MLIR's behaviour?

superlopuh:
Have you tried the "add _i in a for loop until we have an unseen variable" approach?

PapyChacal:
> I was waiting for this to happen :) what's MLIR's behaviour?

AFAIU, there is no MLIR behaviour; MLIR just forgets all values names and sometimes, on some ops, give the results cute name like "cst_0" for arith.constant 0; but nothing about carrying names, does that make sense to you?

My point being, we could try and make it look slightly more MLIRish, but a) that's purely aesthetic and b) We have a fundamentally different handling of things here, so we only could try to mimic results in very specific cases.

PapyChacal:
> Have you tried the "add _i in a for loop until we have an unseen variable" approach?

I tried to try but it's already slightly more complex (or costly, usual tradeoff) than this.
The current approach is to store counters of how many times a given name was deduplicating, to keep going from that index. So we could mimic something like this by just removing one `_[0-9]+` suffix and check against deduplication of that "root"..
Except that now, you have to think about arbitrarily many `_[0-9]+` suffixes in names and all. It gets crazy fast; Hence I'm in favor of Mathieu's solution of just always stripping such prefixes from name_hints in the first place; then we know that adding one can *never* clash with an existing hint that would happen to carry such a suffix and call it a day.

superlopuh (approved):

PR #2546, opened 2024-05-07
misc: TestLowerSnitchStreamToAsm inherits from PipelinePass [NFC]

superlopuh:
I want to add a callback to print between passes in a notebook, this seems to be the cleanest way to do that.


AntonLydike (approved):
:guitar:

PR #2550, opened 2024-05-09
misc: Create a devcontainer for container setup

superlopuh:
wgpu breaks the container, we probably want to have a dual docker + devcontainer setup going, I'll keep a list of tasks here to keep track of work:

 - [ ] migrate the dev requirements to pyproj
 - [ ] don't install wgpu by default for dev reqs


superlopuh:
I just tried this, and it took 10 minutes to install everything and fail. This is probably due to the container on which I've based this one. I think I might as well abandon this PR and start again whenever I get inspired to do so, as I don't have the motivation to debug this right now.

PR #2552, opened 2024-05-09
dialects: (x86) PR16 - Conditional Jumps

KGrykiel:
I added conditional jumps and some infrastructure around it such as the RFLAGS register as well as the 'cmp' operation. For completeness I implemented almost all of the different conditional branch instructions even though some of them are functionally identical. i didn't test all of them individually because they're all the same in essence, but let me know if I should anyway.


superlopuh (commented):
I think we might as well add the tests, so we can detect regressions if there are any in the future. Hopefully it won't be too annoying if you write a little Python script to generate the lit test

superlopuh (approved):
Thank you!

PR #2554, opened 2024-05-09
installation: Add nix flake

alexarice:
Makes the following two changes to make building easier on nixos:
- Adds direnv files to gitignore
- Adds a basic flake that allows `make tests` to be run

@AntonLydike

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,5 @@
 {
-  description = "XDSL devshell";
+  description = "xDSL devshell";
 
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,25 @@
+{
+  description = "XDSL devshell";
+
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { nixpkgs, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
+        let
+          pkgs = import nixpkgs {
+            inherit system;
+          };
+        in
+          {
+            devShells.default = with pkgs; mkShell {
+              LD_LIBRARY_PATH = "${stdenv.cc.cc.lib}/lib";
+              buildInputs = [
+                nodejs_22
+              ];
+            };
+          }
+    );
+}
--- a/.gitignore
+++ b/.gitignore
@@ -158,3 +158,7 @@ docs/ret.xdsl
 
 # asv
 .asv
+
+# direnv
+.direnv
+.envrc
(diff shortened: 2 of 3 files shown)


superlopuh (approved):
I know very little about nix, but if it works on your machine then I think we're safe to merge

flake.nix
```diff
@@ -0,0 +1,25 @@
+{
+  description = "XDSL devshell";
```
superlopuh, line 2: nit:
```suggestion
  description = "xDSL devshell";
```

tobiasgrosser:
> I know very little about nix, but if it works on your machine then I think we're safe to merge

Yes, nice to see better nix support.

alexarice:
One thing I'm unsure about is whether committing the lock file is good practice. Also might want to include `ruff` in the shell

superlopuh:
I did a bit of googling, and it looks like it's mostly a matter of personal preference, although I found some (fixed) issues where nix used to break if the lockfile was gitignored, so at least some people thought ignoring it was a good idea. I'm happy to go either way, it doesn't seem to be a very big file.

AntonLydike (approved):
:partying_face:

AntonLydike:
> One thing I'm unsure about is whether committing the lock file is good practice. Also might want to include `ruff` in the shell

I always thought committing the lock file was good practice for a project, similarly to lock files used by other package managers (notably absent from `requirements.txt` based development...)

georgebisbas (commented):
Would it be an overkill to add a NixOS github actions test?https://github.com/marketplace/actions/install-nix

AntonLydike:
> Would it be an overkill to add a NixOS github actions test?https://github.com/marketplace/actions/install-nix

Maybe we can do this in the future if we find the flake to break regularly?

superlopuh:
My understanding is that the requirements is supposed to be a kind of lockfile, it's just not used that way consistently. The pyproj is supposed to be more flexible in terms of dependency versions, and the requirements.txt is supposed to have a fixed, known-to-work set of exact dep versions.

Anyway, I think this is good to merge, feel free to press the button at your convenience, @alexarice

georgebisbas (approved):

AntonLydike:
> My understanding is that the requirements is supposed to be a kind of lockfile, it's just not used that way consistently. The pyproj is supposed to be more flexible in terms of dependency versions, and the requirements.txt is supposed to have a fixed, known-to-work set of exact dep versions.
> 
> Anyway, I think this is good to merge, feel free to press the button at your convenience, @alexarice

Notably, even `pip freeze` will not freeze versions to exact hashes, making this "lock" not as strong as e.g. poetrys lockfiles. Anyway, we are getting super off-track

PR #2555, opened 2024-05-09
dialects: Add qssa dialect

alexarice:
Adds the qssa dialect for quantum gates. Currently contains only a hadamard gate

Currently rebased on top of my other pr #2554


alexarice:
Closing this until we have a better idea on how we want to implement this

PR #2428, opened 2024-04-04
transformations: (convert-snrt-to-riscv) Add lowerings for even more info ops

AntonLydike:
Missing ops:
- `snrt.global_compute_core_idx`
- `snrt.global_compute_core_num`
- `snrt.global_dm_core_num`
- `snrt.cluster_compute_core_idx`
- `snrt.cluster_dm_core_idx`

These seem to be relatively unused in practice and are skipped for now.

PR #2559, opened 2024-05-10
transforms: (convert-snrt-to-riscv) Add smaller number of ops

AntonLydike:
(hopefully) smaller version of #2428 stripped of all the more complex lowerings introduced in that PR

### DIFF
--- a/xdsl/transforms/convert_snrt_to_riscv.py
+++ b/xdsl/transforms/convert_snrt_to_riscv.py
@@ -1,7 +1,8 @@
 from abc import ABC
 from collections.abc import Sequence
+from dataclasses import dataclass
 
-from xdsl.dialects import builtin, riscv, riscv_snitch, snitch_runtime
+from xdsl.dialects import arith, builtin, riscv, riscv_snitch, snitch_runtime
 from xdsl.dialects.builtin import IntegerAttr
 from xdsl.ir import MLContext, Operation, SSAValue
 from xdsl.passes import ModulePass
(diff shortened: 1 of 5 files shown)


xdsl/transforms/convert_snrt_to_riscv.py
```diff
@@ -14,6 +15,19 @@
...
+class SnrtConstants:
+    """
+    Constants used when compiling the snitch runtime, depend on the exact snitch
+    architecture target.
+    """
+
+    SNRT_CLUSTER_NUM: int = 4
+    SNRT_CLUSTER_CORE_NUM: int = 9
+    SNRT_BASE_HARTID: int = 0
+    SNRT_CLUSTER_DM_CORE_NUM: int = 1
```
superlopuh, line 28: this feels weird, shouldn't these be lower_snake_case?

xdsl/transforms/convert_snrt_to_riscv.py
```diff
@@ -395,10 +411,141 @@ def match_and_rewrite(
...
+
     name = "convert-snrt-to-riscv"
 
+    cluster_num: int = 4
+    cluster_core_num: int = 9
+    base_hartid: int = 0
+    cluster_dm_core_num: int = 1
+
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        constants = SnrtConstants(self.cluster_num)
```
superlopuh, line 547:
```suggestion
        constants = SnrtConstants(self.cluster_num, self.cluster_core_num, self.base_hartid, self.cluster_dm_core_num)
```

xdsl/transforms/convert_snrt_to_riscv.py
superlopuh: Also, I'd recommend removing the default values here, since they're duplicated in the pass. Another alternative would be to move these constants to the module level, and reusing them as default values on the pass. One concern is there might be a new constant added in the future, and it would be nice for the type system to tell us that the pass isn't passing this new value along, instead of silently just using a different value to what's expected.

AntonLydike: Hmm, I have an idea :bulb:

xdsl/transforms/convert_snrt_to_riscv.py
AntonLydike: Fixed this by making the pass inherit from the `SnrtConstants` dataclass, it now integrates very nicely.

superlopuh (approved):

PR #2560, opened 2024-05-10
dialects: (x86) PR18 - m_pop + added rsp dataflow to push

KGrykiel:
Added the memory variant of the pop function and the dataflow of the rsp register for the push operations. Also reformatted a bit to remove unncecessary parent functions as im pretty sure pop and push are unique in their workings. 

Do let me know if you know of a better way to do m.pop because the current way is a bit hacked


xdsl/dialects/x86/ops.py
```diff
@@ -1195,6 +1191,65 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
...
+    """
+    Copies the value at the top of the stack into [r1] and increases %rsp.
+    https://www.felixcloutier.com/x86/pop
+    """
+
+    name = "x86.m.pop"
+
+    rsp_input = operand_def(GeneralRegisterType("rsp"))
+    destination = operand_def(
+        R1InvT
```
superlopuh, line 1205: This seems off, shouldn't it be a Generic parameter?

xdsl/dialects/x86/ops.py
```diff
@@ -1195,6 +1191,65 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
...
+    )  # not an operand, but couldn't think of a better way to make it part of the IR
```
superlopuh, line 1206: Can you please describe what's happening and what you're aming for a bit more?

xdsl/dialects/x86/ops.py
KGrykiel: Not sure what you mean? I define operands the same way in all the other instructions inheriting directly from IRDLOperation. It wouldn't let me pass Generic arguments iirc.

xdsl/dialects/x86/ops.py
KGrykiel: So the problem is that the value within that memory location doesn't matter so it's not really an operand. We did it in a similar way in the register version of pop. In this case however the destination is a memory access, not a register so no SSA value is created either. As such, I didn't see any way to include the information of what the memory access actually is in the IR because it doesn't fit neither operand_def nor result_def.

xdsl/dialects/x86/ops.py
superlopuh: Hmm, not sure how I missed this. This is a TypeVar, it's designed to be passed to `Generic` at the type level. When used like this, it's not any different to just passing the `bound` of the TypeVar, and possibly technically invalid Python, in my understanding. I'd say we should either make the base classes generic in their register types, or switch the operand constraints to pass the bound instead.

xdsl/dialects/x86/ops.py
```diff
@@ -300,59 +300,61 @@ class RR_MovOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
     name = "x86.rr.mov"
 
 
-class M_R_Operation(Generic[R1InvT], IRDLOperation, X86Instruction, ABC):
+@irdl_op_definition
+class R_PushOp(IRDLOperation, X86Instruction, ABC):
```
superlopuh, line 304: This seems to not be an ABC any more, if it defines an __init__

xdsl/dialects/x86/ops.py
superlopuh: The register holds a pointer, not a value, right? My understanding right now is that an operand is clearly the way to model this, although maybe I'm missing something. Could you maybe explain the relevant values in registers and memory before and after this instruction is executed? The link in the docstring is not very clear.

xdsl/dialects/x86/ops.py
```diff
@@ -300,59 +300,61 @@ class RR_MovOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
...
+            result_types=[rsp_output],
         )
 
     def assembly_line_args(self) -> tuple[AssemblyInstructionArg | None, ...]:
         return (self.source,)
 
 
 @irdl_op_definition
-class R_PushOp(M_R_Operation[GeneralRegisterType]):
+class R_PopOp(IRDLOperation, X86Instruction, ABC):
```
superlopuh, line 340:
```suggestion
class R_PopOp(IRDLOperation, X86Instruction):
```

superlopuh, line 340:
```suggestion
class R_PopOp(IRDLOperation, X86Instruction):
```

xdsl/dialects/x86/ops.py
```diff
@@ -370,16 +372,6 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg | None, ...]:
...
-class R_PopOp(R_M_Operation[GeneralRegisterType]):
-    """
-    Copies the value at the top of the stack into r1 and increases %rsp.
-    https://www.felixcloutier.com/x86/pop
-    """
-
-    name = "x86.r.pop"
-
-
 class R_R_Operation(Generic[R1InvT], IRDLOperation, X86Instruction, ABC):
```
superlopuh, line 375:
```suggestion
class R_R_Operation(Generic[R1InvT], IRDLOperation, X86Instruction):
```

xdsl/dialects/x86/ops.py
superlopuh: The issue is that ABC is supposed to override `__new__` and not let users init it directly, but so does our IRDL infrastructure, so we never catch this... All the ops in this file that have an `@irdl_op_definition` annotation should not be ABC

KGrykiel: oh right ok, I'll see if any stragglers slipped through in next PR as well.

PR #2563, opened 2024-05-12
core: fix line numbers in split input mode.

PapyChacal:
I think the effect it is self-explanatory in [the one changed test!](https://github.com/xdslproject/xdsl/pull/2563/files#diff-024119b9d92381358b7618228e4198476beb0ddce5f25b66d204eba6374ee83b).

### DIFF
--- a/xdsl/tools/command_line_tool.py
+++ b/xdsl/tools/command_line_tool.py
@@ -9,6 +9,7 @@ from xdsl.ir import Dialect, MLContext
 from xdsl.parser import Parser
 from xdsl.passes import ModulePass
 from xdsl.utils.exceptions import ParseError
+from xdsl.utils.lexer import Span
 
 
 def get_all_dialects() -> dict[str, Callable[[], Dialect]]:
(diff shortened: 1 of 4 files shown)


xdsl/utils/lexer.py
```diff
@@ -112,6 +117,7 @@ def print_with_context(self, msg: str | None = None) -> str:
         are highlighted by up-carets beneath them (`^`). The message msg is printed
         along these.
         """
+        print("line_offset:", self.line_offset)
```
superlopuh, line 120: Do we want to capture the output here?

xdsl/tools/command_line_tool.py
```diff
@@ -722,9 +725,9 @@ def parse_chunk(self, chunk: IO[str], file_extension: str) -> ModuleOp | None:
...
             return self.available_frontends[file_extension](chunk)
         except ParseError as e:
+            s = e.span
+            e.span = Span(s.start, s.end, s.input, start_offset)
             if "parsing_diagnostics" in self.args and self.args.parsing_diagnostics:
                 print(e.with_context())
             else:
                 raise Exception("Failed to parse:\n" + e.with_context()) from e
-        finally:
-            chunk.close()
```
superlopuh, line 730: Do we no longer want this? Why would this not be needed?

PapyChacal: Good catch, forgot this change, it can stay!
Note that we never needed it, it's just cleaner file/memory management

xdsl/utils/lexer.py
PapyChacal: How did this not trigger anything :exploding_head: Thanks!

superlopuh (approved):

PR #2564, opened 2024-05-12
testing: remove dense test kernel from bottom-up tests

superlopuh:
We can add it back later, but right now it's slowing my bottom-up test migration a bit.


AntonLydike (approved):
Big red number = good patch :chart_with_upwards_trend:

PR #2568, opened 2024-05-13
backend: (riscv) move results of lowered riscv_scf.for loops

superlopuh:
During register allocation, the registers of the inputs, results, and iteration arguments of the for loop must all be the same, so we need to defensively add moves all over the place, which will hopefully be eliminated after register allocation. We are currently missing moves after the for loop, this PR fixes this.


AntonLydike (approved):
Sure thing! :partying_face:

Issue #2377, opened 2024-03-25
dialects: (x86) move assembly helpers to assembly.py


superlopuh:
We'd like to keep `ops.py` dedicated to the ops in the dialect

KGrykiel:
Hi, just about ready to look into that, is the idea to make the assembly.py file for x86 specifically or also to extract the common functionality from both x86 and riscv and put it into xdsl/backend?

superlopuh:
I think x86 is the priority, but factoring out common functionality to the two of them could also be nice, and make it slightly easier to add ARM in the future

PR #2570, opened 2024-05-13
dialects: (x86) PR19 - move assembly helpers to separate class

KGrykiel:
I moved most of the printing and parsing methods to a separate file as per issue #2377. I also moved the LabelAttr to attributes.py because it was needed in both .assembly and .ops + it seemed to make sense.

I left the print_assembly and x86_code in .ops as otherwise it would create a circular import.

### DIFF
--- a/xdsl/dialects/x86/attributes.py
+++ b/xdsl/dialects/x86/attributes.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from xdsl.ir import (
-    Data,
+    Data
 )
 from xdsl.irdl import (
     irdl_attr_definition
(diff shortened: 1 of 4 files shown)


xdsl/dialects/x86/attributes.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+from xdsl.ir import (
+    Data,
+)
+from xdsl.irdl import (
+    irdl_attr_definition,
```
superlopuh, line 7:
```suggestion
    irdl_attr_definition
```

PR #2571, opened 2024-05-13
interpreter: (riscv) add fmv.d implementation

superlopuh:


Moxinilian:
Not an expert, but my understanding of this instruction is that it will cast from arbitrary int to float. Is this properly handled here? (I don't know the interpreter infra)

superlopuh:
This just moves a float to a float, the int casting op is [fcvt.w.s](https://msyksphinz-self.github.io/riscv-isadoc/html/rvfd.html#fcvt.w.s).

xdsl/interpreters/riscv.py
```diff
@@ -518,6 +518,17 @@ def run_fld(
...
 
+    @impl(riscv.FMvDOp)
+    def run_fmv_d(
+        self,
+        interpreter: Interpreter,
+        op: riscv.FMvDOp,
+        args: tuple[Any, ...],
+    ):
+        args = RiscvFunctions.get_reg_values(interpreter, op.operands, args)
+        results = args
```
AntonLydike, line 529: I think you could get away without this line?

AntonLydike (approved):
Tiny nit comment, otherwise neat stuff! :racehorse:

xdsl/interpreters/riscv.py
superlopuh: Defo, I'm just trying to make it explicit what's happening here, I feel like it makes it a bit clearer that it's on purpose. Maybe it could be resolved better with a comment...

PR #2573, opened 2024-05-13
dialects: (x86) PR20 - remaining x86Lite instructions + cleanup (season finale)

KGrykiel:
Added the remaining missing instructions and alternative configurations of existing ones. Moreover removed ABC from where it was not supposed to be and made some visual changes to the tests to keep it more consistent.


superlopuh (approved):
We have liftoff

PR #2574, opened 2024-05-13
backend: (snitch) insert fmv ops when necessary when lowering streams

superlopuh:


AntonLydike (approved):
Nice stuff! :sparkles:

PR #2577, opened 2024-05-13
dialects: (hw) Add HWModuleLike trait

Moxinilian:
This only adds the HWModuleLike methods that are currently relevant. Eventually it may be nice to add them all, but as they are pretty niche and would require quite a few changes on HWModuleOp...

### DIFF
--- a/tests/dialects/test_hw.py
+++ b/tests/dialects/test_hw.py
@@ -375,7 +375,7 @@ def test_hwmoduleop_hwmodulelike():
     )
 
     hw_module_like = hw_module.get_trait(HWModuleLike)
-    assert hw_module_like
+    assert hw_module_like is not None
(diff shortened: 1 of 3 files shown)


tests/dialects/test_hw.py
```diff
@@ -359,3 +365,25 @@ def test_instance_builder():
...
+
+def test_hwmoduleop_hwmodulelike():
+    module_type = ModuleType((ArrayAttr(()),))
+
+    hw_module = HWModuleOp(
+        StringAttr("foo"), module_type, Region((Block((OutputOp(),)),))
+    )
+
+    hw_module_like = hw_module.get_trait(HWModuleLike)
+    assert hw_module_like
```
superlopuh, line 378: nit: 
```suggestion
    assert hw_module_like is not None
```

superlopuh (approved):

PR #2578, opened 2024-05-13
testing: raise riscv_scf to scf in bottom-up tests

superlopuh:


AntonLydike (approved):
I don't see any raising happening here, but you do you! :rocket:

PR #2579, opened 2024-05-13
testing: (snitch) bottom-up tests have no riscv

superlopuh:
We can express all our kernels in an architecture-independent manner


AntonLydike (approved):
hells yeah! Let's gooo!

PR #2580, opened 2024-05-13
rewriting: Add convert-ml-program-to-memref

superlopuh:
I'm not 100% sure that this is the intended way that ml_program should be lowered but it seems to work for our ONNX kernels.


xdsl/dialects/memref.py
```diff
@@ -359,10 +359,10 @@ class GetGlobal(IRDLOperation):
...
 
     @staticmethod
-    def get(name: str, return_type: Attribute) -> GetGlobal:
-        return GetGlobal.build(
-            result_types=[return_type], properties={"name": SymbolRefAttr(name)}
-        )
+    def get(name: str | SymbolRefAttr, return_type: Attribute) -> GetGlobal:
+        if isinstance(name, str):
+            name = SymbolRefAttr(name)
+        return GetGlobal.build(result_types=[return_type], properties={"name": name})
```
PapyChacal, line 365: Let's make it `__init__` if you're already making a change?

xdsl/transforms/convert_ml_program_to_memref.py
```diff
@@ -0,0 +1,71 @@
...
+    ) -> None:
+        assert isinstance(op_type := op.result.type, TensorType)
+        op_type = cast(TensorType[Any], op_type)
+        new_type = memref.MemRefType(op_type.element_type, op_type.shape)
+        rewriter.replace_matched_op(
+            (
+                mem := memref.GetGlobal.get(op.global_attr, new_type),
+                bufferization.ToTensorOp(mem.memref),
+            )
+        )
```
PapyChacal, line 58: This is dark magic :mage: 
But necessary dark magic when you have to deal with linking and tensor I guess.
Can you simply comment on that here?

PapyChacal: NB: As said, this is necessary in general as you want it to work with external tensor symbols.
What about special casing it for when the symbol actually is in the same module?

PapyChacal (approved):

PR #2585, opened 2024-05-14
backend: (riscv) flatten loops that don't use induction variables

superlopuh:
Turns out we have those often when lowering from memref_stream


xdsl/transforms/riscv_scf_loop_flatten.py
```diff
@@ -0,0 +1,145 @@
...
+#      for j in range(k):
+#        f(A[i+j])
+#
+#    factor = (iu - il) // is
+#    for o in range(ol, ou * factor, os):
+#      # o is not used
+#
+
+
+class FuseNestedLoopsPattern(RewritePattern):
```
AntonLydike, line 41: Nit: Still called fuse

AntonLydike (approved):
Nit: Maybe a neater approach would be to check if *every* use is an expression like `i+K*j`, which would catch both the empty case, as well as any un-cse'd version of the IR.

Anyway, this is a neat fix already! LGTM

superlopuh:
Woops, I haven't actually implemented the i+K*j, only i+j, I'll fix the doc. The issue is both are possible, and have pretty different matching patterns, so we can't get away with just i+k*j or i+j anyway, so the case where neither gets used can be seen as a special case for either of these possibilities.

PR #2586, opened 2024-05-14
dialects: (x86) - AVX registers + some instructions (bonus PR!)

KGrykiel:
I added AVX registers and some example instructions to go with them. The reason for this specific selection is because I believe @tarinduj needed them for something. Let me know if that wasn't the case.


superlopuh (approved):
Looking great! CC @tarinduj

tests/dialects/test_x86.py
```diff
@@ -33,3 +33,45 @@ def test_rflags_register():
...
+        (x86.register.ZMM25, "zmm25"),
+        (x86.register.ZMM26, "zmm26"),
+        (x86.register.ZMM27, "zmm27"),
+        (x86.register.ZMM28, "zmm28"),
+        (x86.register.ZMM29, "zmm29"),
+        (x86.register.ZMM30, "zmm30"),
+        (x86.register.ZMM31, "zmm31"),
+    ],
+)
+def test_avx_register(register: x86.register.AVXRegisterType, name: str):
```
AntonLydike, line 75: I'm not sure we need all these additional tests? I feel that they bloat test time which is already getting quite large imo

KGrykiel: @webmiche suggested doing a parametrised test like that for general purposes registers (https://github.com/xdslproject/xdsl/pull/2333#discussion_r1526190469) so I wanted to keep it consistent.

PR #2588, opened 2024-05-15
testing: make loop bounds match parallel iterators in bottom-up test

superlopuh:
This is what converting memref_stream to loops generates for the parallel iterators, I'll do reductions in a later PR.


AntonLydike (approved):
Coolio! Nice stuff! Just wondering, we should have canonicalization patterns that help a bit with these loops (especially the single-iteration kind), so we could use them here, right?

PR #2593, opened 2024-05-15
backend: (csl) Add printing for CSL dialect ops

AntonLydike:
This PR adds printing for the CSL dialect operations (stacked on top of #2591)

### DIFF
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from typing import IO
 
-from xdsl.dialects import arith, func, scf
+from xdsl.dialects import arith, csl, scf
 from xdsl.dialects.builtin import (
     Float16Type,
     Float32Type,
@@ -29,14 +29,14 @@ class CslPrintContext:
 
     _prefix: str = field(default="")
 
-    def print(self, text: str, prefix: str = ""):
+    def print(self, text: str, prefix: str = "", end: str = "\n"):
         """
         Print `text` line by line, prefixed by self._prefix and prefix.
         """
         for l in text.split("\n"):
-            print(self._prefix + prefix + l, file=self.output)
+            print(self._prefix + prefix + l, file=self.output, end=end)
 
-    def _get_variable_name_for(self, val: SSAValue) -> str:
+    def _get_variable_name_for(self, val: SSAValue, hint: str | None = None) -> str:
         """
         Get an assigned variable name for a given SSA Value
         """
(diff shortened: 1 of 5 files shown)


xdsl/backend/csl/print_csl.py
```diff
@@ -133,17 +141,51 @@ def print_block(self, body: Block):
...
-                case func.FuncOp(sym_name=name, body=bdy, function_type=ftyp) if len(
+                case csl.ImportModuleConstOp(module=module, params=params, result=res):
+                    name = self._get_variable_name_for(res)
+
+                    params_str = ""
+                    if params is not None:
+                        params_str = f", {self._get_variable_name_for(params)}"
+
+                    self.print(
+                        f'const {name} : comptime_struct = @import_module("{module.data}"{params_str});'
```
dk949, line 152: Giving the result of `@import_module` an explicit type of `comptime_struct` crashes the cslc compiler. The type can be left out or explicitly specified as `imported_module`. AFAICT this is the only difference between`comptime_struct` and `imported_module`, so no other changes would be necessary (except in the filecheck).

xdsl/backend/csl/print_csl.py
```diff
@@ -29,23 +29,27 @@ class CslPrintContext:
...
+    def print(self, text: str, prefix: str = "", end: str = "\n"):
         """
         Print `text` line by line, prefixed by self._prefix and prefix.
         """
         for l in text.split("\n"):
-            print(self._prefix + prefix + l, file=self.output)
+            print(self._prefix + prefix + l, file=self.output, end=end)
 
-    def _get_variable_name_for(self, val: SSAValue) -> str:
+    def _get_variable_name_for(self, val: SSAValue, hint: str | None = None) -> str:
```
superlopuh, line 39: I'm not sure how much we want to rely on name hints for naming variables in the output, especially now that we mess with them at parsing time, like stripping `_0`, `_1`, ... suffixes. The two alternatives that come to mind is either a custom type for variables or an attribute on ops that own the variables. The second seems vaguely preferable. I don't want to derail your csl mergeathon too much, but if you agree that not relying on name hints is a good change in the future, it would be great to file a PR and mention it in a comment.

superlopuh (approved):

xdsl/backend/csl/print_csl.py
AntonLydike: Good to know! I added a new type `ImportedModuleType` and an alias `StructLike = ImportedModuleType | ComptimeStructType`, which we can extend with any other struct like objects in the future, I feel this gives us the best balance between accuracy and flexibility without lying in the IR types (compared to when we just print a different return type from an import)

xdsl/backend/csl/print_csl.py
AntonLydike: Yeah, not sure how we want to continue to go on with name hints from here. I'd see how it evolves over time, maybe we don't actually need readable CSL (the official examples don't seem to bother too much with that either).

PR #2596, opened 2024-05-15
dialects: Add optional result to stencil.store to forward the stored values with halo

PapyChacal:
This enables further stencil computations to act on the stored values, but also on out-of-computed-bounds ones, read from the stored-to buffer.
Otherwise, those would have to be somehow loaded from the field. But one cannot load and store from the same field!


PapyChacal:
Not needed anymore thanks to stencil bufferization

PR #2592, opened 2024-05-15
dialect,backend: Experimental CSL dialect and printer

dk949:
An experimental implementation of the CSL dialect modeling the CSL language used by the Cerebras WSE based systems.

Also includes a backend generating the CSL language and a tool to split output of the backend into the files consumable by cslc.

**Files will need to be moved to the experimental directory before merging**

Currently implemented features:

## Types

* `csl.ptr`
* `csl.comptime_struct`
* `csl.color`
* `csl.string`
* `csl.type`

## Attributes

* `csl.module_kind`
* `csl.task_kind`
* `csl.ptr_kind`
* `csl.ptr_const`

## Ops

* `csl.module`
* `csl.layout`
* `csl.param`
* `csl.import_module`
* `csl.member_access`
* `csl.member_call`
* `csl.addressof`
* `csl.func`
* `csl.task`
* `csl.call`
* `csl.return`
* `csl.export`
* `csl.set_rectangle`
* `csl.set_tile_code`
* `csl.rpc`
* `csl.get_color`
* `csl.const_struct`
* `csl.const_str`
* `csl.const_type`

## Non-CSL ops supported by the printer

* `arith.Addf`
* `arith.Addi` 
* `arith.Constant`
* `arith.ExtFOp` 
* `arith.ExtSIOp` 
* `arith.ExtUIOp`
* `arith.FPToSIOp` 
* `arith.IndexCastOp` 
* `arith.Mulf`
* `arith.Muli` 
* `arith.SIToFPOp` 
* `arith.TruncFOp` 
* `arith.TruncIOp` 
* `memref.GetGlobal`
* `memref.Global`
* `memref.Load`
* `memref.Store`
* `scf.Yield`

Previously discussed with @AntonLydike and @n-io

PR #2603, opened 2024-05-16
dialects: (csl) Added call op and tasks

dk949:
* The three kinds of tasks
  * Tasks carry IDs they will be bound to (by the printer, coming in future PR)
* `csl.call`, for tasks and functions
* Removed `AttrSizedOperandSegments` from `csl.member_call` following a discussion with @AntonLydike

### DIFF
--- a/xdsl/dialects/csl.py
+++ b/xdsl/dialects/csl.py
@@ -9,7 +9,6 @@ This is meant to be used in conjunction with the `-t csl` printing option to gen
 
 from __future__ import annotations
 
-from abc import abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import TypeAlias
@@ -89,7 +88,7 @@ class TaskKind(StrEnum):
     CONTROL = "control"
 
 
-class _FuncBase:
+class _FuncBase(IRDLOperation):
     """
     Base class for the shared functionalty of FuncOp and TaskOp
     """
@@ -99,11 +98,6 @@ class _FuncBase:
     function_type: FunctionType = prop_def(FunctionType)
     arg_attrs = opt_prop_def(ArrayAttr[DictionaryAttr])
     res_attrs = opt_prop_def(ArrayAttr[DictionaryAttr])
-
-    @abstractmethod
-    def _get_name(self) -> str: ...
-    @abstractmethod
-    def _get_attribs_and_props(self) -> dict[str, Attribute]: ...
 
     def _props_region(
         self,
@@ -118,9 +112,7 @@ class _FuncBase:
             inputs, output = function_type
             function_type = FunctionType.from_lists(inputs, [output] if output else [])
         if len(function_type.outputs) > 1:
-            raise ValueError(
-                f"Can't have a {self._get_name()} return more than one value!"
-            )
+            raise ValueError(f"Can't have a {self.name} return more than one value!")
         if not isinstance(region, Region):
             region = Region(Block(arg_types=function_type.inputs))
         properties: dict[str, Attribute | None] = {
@@ -150,7 +142,7 @@ class _FuncBase:
             self.sym_name,
             self.function_type,
             self.body,
-            self._get_attribs_and_props(),
+            self.attributes | self.properties,
             arg_attrs=self.arg_attrs,
             reserved_attr_names=(
                 "sym_name",
@@ -350,7 +342,7 @@ class MemberCallOp(IRDLOperation):
 
 
 @irdl_op_definition
-class FuncOp(IRDLOperation, _FuncBase):
+class FuncOp(_FuncBase):
     """
     Almost the same as func.func, but only has one result, and is not isolated from above.
 
@@ -407,15 +399,9 @@ class FuncOp(IRDLOperation, _FuncBase):
     def print(self, printer: Printer):
         _FuncBase._print(self, printer)
 
-    def _get_name(self) -> str:
-        return self.name
-
-    def _get_attribs_and_props(self) -> dict[str, Attribute]:
-        return self.attributes | self.properties
-
 
 @irdl_op_definition
-class TaskOp(IRDLOperation, _FuncBase):
+class TaskOp(_FuncBase):
     """
     Represents a task in CSL. All three types of task are represented by this Op.
 
@@ -528,12 +514,6 @@ class TaskOp(IRDLOperation, _FuncBase):
 
     def print(self, printer: Printer):
         _FuncBase._print(self, printer)
-
-    def _get_name(self) -> str:
-        return self.name
-
-    def _get_attribs_and_props(self) -> dict[str, Attribute]:
-        return self.attributes | self.properties
 
 
 @irdl_op_definition
--- a/xdsl/dialects/csl.py
+++ b/xdsl/dialects/csl.py
@@ -21,6 +21,8 @@ from xdsl.dialects.builtin import (
     ModuleOp,
     StringAttr,
     SymbolRefAttr,
+    IntegerAttr,
+    IntegerType,
 )
 from xdsl.dialects.utils import parse_func_op_like, print_func_op_like
 from xdsl.ir import (
@@ -62,6 +64,7 @@ from xdsl.traits import (
 )
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.str_enum import StrEnum
+from xdsl.utils.hints import isa
 
 
 class PtrKind(StrEnum):
@@ -77,6 +80,12 @@ class PtrConst(StrEnum):
 class ModuleKind(StrEnum):
     LAYOUT = "layout"
     PROGRAM = "program"
+
+
+class TaskKind(StrEnum):
+    LOCAL = "local"
+    DATA = "data"
+    CONTROL = "control"
 
 
 class FuncBase:
@@ -226,6 +235,18 @@ class ModuleKindAttr(EnumAttribute[ModuleKind], SpacedOpaqueSyntaxAttribute):
     """Attribute representing the kind of CSL module, either layout or program"""
 
     name = "csl.module_kind"
+
+
+@irdl_attr_definition
+class TaskKindAttr(EnumAttribute[TaskKind], SpacedOpaqueSyntaxAttribute):
+    name = "csl.task_kind"
+
+    def get_color_bits(self):
+        match self.data:
+            case TaskKind.LOCAL | TaskKind.DATA:
+                return 5
+            case TaskKind.CONTROL:
+                return 6
 
 
 @irdl_attr_definition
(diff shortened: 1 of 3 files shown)


xdsl/dialects/csl.py
```diff
@@ -344,20 +404,121 @@ def parse(cls, parser: Parser) -> FuncOp:
...
+            "id": id,
+        }
+        super().__init__(properties=properties, regions=[region])
+
+    def verify_(self) -> None:
+        FuncBase._verify(self)
+        if len(self.function_type.outputs.data) != 0:
+            raise VerifyException(f"{self.name} cannot have return values")
+
+        # TODO(dk949): Need to check at some point that we're not reusing the same color multiple times
```
AntonLydike, line 462: I think we could do this as part of a pass instead? Maybe have some pass that allocates ids to tasks?

xdsl/dialects/csl.py
```diff
@@ -344,20 +404,121 @@ def parse(cls, parser: Parser) -> FuncOp:
...
+    def __init__(
+        self,
+        name: str,
+        function_type: FunctionType | tuple[Sequence[Attribute], Attribute | None],
+        region: Region | type[Region.DEFAULT] = Region.DEFAULT,
+        *,
+        task_kind: TaskKindAttr | TaskKind,
+        arg_attrs: ArrayAttr[DictionaryAttr] | None = None,
+        res_attrs: ArrayAttr[DictionaryAttr] | None = None,
+        id: IntegerAttr[IntegerType] | int,
```
AntonLydike, line 438: I think task kind and task ID should be optional, right? Can tasks not have IDs?

xdsl/dialects/csl.py
```diff
@@ -372,14 +533,14 @@ class ReturnOp(IRDLOperation):
...
-    traits = frozenset([HasParent(FuncOp), IsTerminator()])
+    traits = frozenset([HasParent(FuncOp, TaskOp), IsTerminator()])
 
     def __init__(self, return_val: SSAValue | Operation | None = None):
         super().__init__(operands=[return_val])
 
     def verify_(self) -> None:
         func_op = self.parent_op()
-        assert isinstance(func_op, FuncOp)
+        assert isinstance(func_op, FuncOp) or isinstance(func_op, TaskOp)
```
AntonLydike, line 543: This is handled by the `HasParent` traits verifier

xdsl/dialects/csl.py
```diff
@@ -412,6 +573,25 @@ def print(self, printer: Printer):
...
+
+    name = "csl.call"
+
+    callee = prop_def(SymbolRefAttr)
+    args = var_operand_def(Attribute)
+    result = opt_result_def(Attribute)
+
+    # TODO(dk949): verify that callee corresponds to a real symbol
+
+    # TODO(dk949): verify that function type of callee matches args and result
```
AntonLydike, line 590: Not sure these are required in MLIR...

xdsl/dialects/csl.py
```diff
@@ -412,6 +573,25 @@ def print(self, printer: Printer):
...
+
+    # TODO(dk949): verify that if Call is used outside of a csl.func or csl.task it has a result
```
AntonLydike, line 592: Is this a restriction given by CSL?

xdsl/dialects/csl.py
dk949: This seems more sensible than doing it in the `verify`. Should I remove the TODO?

xdsl/dialects/csl.py
```diff
@@ -80,6 +82,82 @@ class ModuleKind(StrEnum):
     PROGRAM = "program"
 
 
+class TaskKind(StrEnum):
+    LOCAL = "local"
+    DATA = "data"
+    CONTROL = "control"
+
+
+class FuncBase:
```
AntonLydike, line 91: We can make this module private, right? It's not supposed to be used by anyone except `TaskOp` and `FuncOp`.

AntonLydike (commented):
Really nice stuff! I just think there are a few tiny things here we can improve

xdsl/dialects/csl.py
AntonLydike: In general, we can have a pass called `csl-legalize` that looks for broken programs, and tries to fix them (maybe inserting foreign function definitions, assigning task IDs or something like that) and if it can't fix an issue it bails out. Basically as a sanity check that we will print valid CSL code.

xdsl/dialects/csl.py
dk949: Yes, a call outside of a function or task is a `comptime` call, so can't have sideeffects. I could've articulated that better. 

It could be done as a pass instead.

xdsl/dialects/csl.py
dk949: Kind enforces restrictions on what args the task can have, I think it should be mandatory.

The id I was using to immediately bind tasks upon declaration, but I guess this can be done at runtime too.

I can make it optional and only bind the task if id is present (in the printer).

This would necessitate some sort of `csl.bind_task` op, which we don't currently have, but it'd be pretty simple to add.

xdsl/dialects/csl.py
dk949: This was just to appease the type checker. I can use a `cast` instead

AntonLydike: Oh! I missed that you had more in the verifier! Sorry

xdsl/dialects/csl.py
AntonLydike: Hmm, I feel this can be done in the verifier? Not sure either way.

AntonLydike: We can leave it a todo for now

xdsl/dialects/csl.py
AntonLydike: Hmm, whatever you prefer. Just wanted to note this. Is there overlap in what args are allowed?

AntonLydike (approved):
Nice!

xdsl/dialects/csl.py
dk949: Local: no args
Data: 1 - 4 args of no more than 64 bits total
Control: 0 - 4 args of no more than 64 bits total

The way the meaning of the args also changes between data and control, control tasks get passed their IDs, data tasks just get the data.

n-io:
Have you considered introducing subclasses - however small - for the different types of tasks?

It seems to be in the spirit of xdsl (see for instance the `Addf` and `Subf` ops in the `arith` dialect). There also seems to be a stronger distinction in CSDK 1.1 which removed `@bind_task` in favour of `@bind_control_task`, `@bind_data_task`, and `@bind_local_task`, on top of the existing `@get_control_task_id`, `@get_data_task_id` and `@get_local_task_id` as well as the types `control_task_id`, `data_task_id`, `local_task_id`.

dk949:
@n-io I am not sure the difference between the task kinds warrants different classes.

In terms of binding the tasks, I'm already using the `@bind_<kind>_task` syntax in #2592 printer (see the `_bind_task` method [here](https://github.com/xdslproject/xdsl/pull/2592/files#diff-04dab167c57fd9b1b69fa2b32ae20f4b571e8fe718d04fed8298d4920be7e782R211))

That said, the implementation would boil down to a `match...case` whether using inheritance or enums, so it'd probably be equally clean.

AntonLydike:
I think this is an interesting question, and I had a quick chat with Markus on this. Basically, the key takeaway here is that if very few optimizations care about the enum (e.g. if canonicalization patterns, or optimizations matching on tasks don't care what kind of task it is), then we don't need to introduce extra operations for each kind of task there is. But if the enum significantly changes the semantics of the operation, we probably want to create multiple different operations.

I think (but please correct me if I'm wrong), that for us the semantics of the operation don't really change depending on the value of the enum, and therefore we don't need separate operations for this.

n-io:
> That said, the implementation would boil down to a `match...case` whether using inheritance or enums, so it'd probably be equally clean.

I think the op rewriter matches on classes rather than enum members. I'm not suggesting to not have a shared base class, I think that's good to have (and can still be used for matching, as @AntonLydike suggested).

Given that the function signatures and return types are more or less fixed, offering simple constructors (or static builders) might save us writing quite a bit of verification code. We might also have different verification methods, where the data task might verify that it has one 32bit or two 16bit args (iirc), a check the others don't need to do.

dk949:
> We might also have different verification methods, where the data task might verify that it has one 32bit or two 16bit args (iirc), a check the others don't need to do.

The verification code would be pretty simple [(it's mostly already there)](https://github.com/xdslproject/xdsl/blob/c70fe53fdac35c878e0c36666f8b0704f5fd2ee2/xdsl/dialects/csl.py#L469-L482). It doesn't verify that `sum(bit_size_of(arg) for args) <= 64`, mostly because I don't know if we have a `bit_size_of` function in xDSL.

I think @AntonLydike is right and the tasks are similar enough to each other that we can get away with just an enum and not introduce more code than we have to.

xdsl/dialects/csl.py
```diff
@@ -80,6 +82,82 @@ class ModuleKind(StrEnum):
...
+                "input types"
+            )
+
+    def _print(self, printer: Printer):
+        print_func_op_like(
+            printer,
+            self.sym_name,
+            self.function_type,
+            self.body,
+            getattr(self, "attributes", {}) | getattr(self, "properties", {}),
```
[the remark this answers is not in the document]
dk949: That would be a better solution, can do the same with `name` in the comment as well.

AntonLydike:
> Given that the function signatures and return types are more or less fixed, offering simple constructors (or static builders) might save us writing quite a bit of verification code. We might also have different verification methods, where the data task might verify that it has one 32bit or two 16bit args (iirc), a check the others don't need to do.

We can still have specific constructors for all the different kinds of tasks, just as static methods on the operation. Adding operations on such a fine-grained level would imo just make the dialect bigger.

xdsl/dialects/csl.py
```diff
@@ -344,20 +405,135 @@ def parse(cls, parser: Parser) -> FuncOp:
...
+            task_kind=extra_attrs.data["kind"],
+            id=id,
+        )
+        return task
+
+    def print(self, printer: Printer):
+        _FuncBase._print(self, printer)
+
+    def _get_name(self):
+        return self.name
```
AntonLydike, line 533: Also, you are missing a type annotation here

xdsl/dialects/csl.py
```diff
@@ -80,6 +83,84 @@ class ModuleKind(StrEnum):
...
+    body: Region = region_def()
+    sym_name: StringAttr = prop_def(StringAttr)
+    function_type: FunctionType = prop_def(FunctionType)
+    arg_attrs = opt_prop_def(ArrayAttr[DictionaryAttr])
+    res_attrs = opt_prop_def(ArrayAttr[DictionaryAttr])
+
+    @abstractmethod
+    def _get_name(self) -> str: ...
+    @abstractmethod
+    def _get_attribs_and_props(self) -> dict[str, Attribute]: ...
```
AntonLydike, line 106: Another minor nit comment:

I don't quite understand why these are needed here, as they are both implemented as just getters for the `name` attribute, which you can just grab the same way using `self.name`

Also, I would make `_FuncBase` actually inherit from `ABC` *and* `IRDLOperation` so that it is a proper abstract class and can access all the IRDL fields (such as name).

dk949: Now that I am looking at it, I am not entirely sure why _FuncBase doesn't just inherit from `IRDLOperation`. I think I ran into some issues early on when I didn't quite get how the `@irdl_op_definition` worked.

xdsl/dialects/csl.py
```diff
@@ -80,6 +82,77 @@ class ModuleKind(StrEnum):
     PROGRAM = "program"
 
 
+class TaskKind(StrEnum):
+    LOCAL = "local"
+    DATA = "data"
+    CONTROL = "control"
+
+
+class _FuncBase(IRDLOperation):
```
AntonLydike, line 91: This should probably also inherit from `ABC` for completeness sake, just so that it's clear that this is not to be instantiated

superlopuh (approved):

PR #2610, opened 2024-05-20
backend: (csl) Add printing some scf and memref features

dk949:
Add csl backend support for the following operations:
* `scf.For`
* `scf.yield`
* `memref.global`
* `memref.get_global`

Also added the following types:
* `memref`
* `index`


AntonLydike (approved):
Good stuff!

superlopuh (approved):

PR #2614, opened 2024-05-21
core: Make printer scoped.

### LINKED ISSUE #2609: Fix SSA value name re-use
The following IR has SSA value name re-use in MLIR but not in xDSL:

```
// test.mlir
  func.func @a() {
    %0 = "test.op"() : () -> i32
    return
  }
  func.func @b() {
    %0 = "test.op"() : () -> i32
    return
  }
```

See:

```
$ mlir test.mlir
module {
  func.func @a() {
    %0 = "test.op"() : () -> i32
    return
  }
  func.func @b() {
    %0 = "test.op"() : () -> i32
    return
  }
}
```

```
$ xdsl-opt test.mlir
builtin.module {
  func.func @a() {
    %0 = "test.op"() : () -> i32
    func.return
  }
  func.func @b() {
    %1 = "test.op"() : () -> i32
    func.return
  }
}
```

The second result of `test.op` is named `%1` even though `%0` would be a valid name.

This makes composing filechecks together more difficult, as previously inserted regions (e.g. functions) will mess with the names in functions introduced later.

PapyChacal:
Fixes #2609

### DIFF
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -88,8 +88,8 @@ class Printer:
     """
     _ssa_names: dict[str, int] = field(default_factory=dict, init=False)
     _block_names: dict[Block, int] = field(default_factory=dict, init=False)
-    _next_valid_name_id: int = field(default=0, init=False)
-    _next_valid_block_id: int = field(default=0, init=False)
+    _next_valid_name_id: list[int] = field(default_factory=lambda: [0], init=False)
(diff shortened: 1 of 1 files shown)


AntonLydike:
Thank you so much for this :clap: :clap: :clap:

AntonLydike:
I'm gonna take a shot at the failing tests that I know about

math-fehr (approved):
Really cool!

PR #2615, opened 2024-05-21
transformations: Implement stencil inlining.

PapyChacal:
Apologies for the monster PR; I might be able to split in two :thinking:


tests/filecheck/transforms/stencil-inlining.mlir
```diff
@@ -0,0 +1,355 @@
...
+// CHECK-NEXT:      }
+// CHECK-NEXT:      %4 = stencil.buffer %1 : !stencil.temp<[0,64]x[0,64]x[0,60]xf64>
+// CHECK-NEXT:      %5 = stencil.apply(%arg2_1 = %4 : !stencil.temp<[0,64]x[0,64]x[0,60]xf64>) -> (!stencil.temp<[0,64]x[0,64]x[0,60]xf64>) {
+// CHECK-NEXT:        %6 = stencil.access %arg2_1[0, 0, 0] : !stencil.temp<[0,64]x[0,64]x[0,60]xf64>
+// CHECK-NEXT:        %7 = stencil.store_result %6 : !stencil.result<f64>
+// CHECK-NEXT:        stencil.return %7 : !stencil.result<f64>
+// CHECK-NEXT:      }
+// CHECK-NEXT:      stencil.store %5 to %arg1 ([0, 0, 0] : [64, 64, 60]) : !stencil.temp<[0,64]x[0,64]x[0,60]xf64> to !stencil.field<[-3,67]x[-3,67]x[-3,67]xf64>
+// CHECK-NEXT:      func.return
+// CHECK-NEXT:    }
```
superlopuh, line 355:
```suggestion
// CHECK-NEXT:    }

```

tobiasgrosser:
How nice. I am curious about the performance numbers.

xdsl/tools/command_line_tool.py
```diff
@@ -572,6 +572,11 @@ def get_replace_incompatible_fpga():
 
         return replace_incompatible_fpga.ReplaceIncompatibleFPGA
 
+    def get_stencil_inlining():
+        from xdsl.transforms import stencil_inlining
```
georgebisbas, line 576: Is there a cyclic import? Why not on top?

georgebisbas (approved):

xdsl/tools/command_line_tool.py
superlopuh: We have a lazy import interface that avoids the 1+s import of all the files in xDSL by default when invoking `xdsl-opt`

PapyChacal:
> How nice. I am curious about the performance numbers.

Me too!

There is some polishing to do, it does not seem to work exactly as expected on big kernels.
Because of my KISS implementation, this pass as-is also combinatorial explode on big kernels.
So I'll first integrate with #2623, to cut off those explosions and be able to pinpoint where exactly I missed something still!

xdsl/tools/command_line_tool.py
PapyChacal: (For completeness)
We avoid those 1+s imports because all decorators are reapplied on import, and this just makes xDSL startup time explode

georgebisbas (commented):
Is this now ready to review?

tobiasgrosser:
Nice. The performance looks good. Why do we get a relative error here? Should inlining not perform exactly the same compuation?

georgebisbas:
Are these seconds?

PapyChacal:
> Nice. The performance looks good. Why do we get a relative error here? Should inlining not perform exactly the same compuation?

The relative error is xDSL inlined vs OEC inlined.

I'm not sure what exactly causes the slight differences yet, but could look into it; it might be quite the rabbit hole, given that different versions of CUDA, MLIR and clang are at play.

Regarding inlining doing the same computations; I'm not confident about OEC - or me missing something - on this side yet. In my tests, OEC plains out crashes on some examples if I don't use inlining. On some other examples that run without issue, OEC's inlining appear to change the results relatively significantly; I could look into that too, most likely in priority to xDSL vs old MLIR.

That's why I reported the relative error of both frameworks with inlining enabled for now. It is the case without surprise from OEC and at least demonstrating that things are consistent there.

I don't mind waiting until all those grey areas are clarified if anybody prefers

PapyChacal:
> Are these seconds?

milliseconds! Just the first thing I got to work on that side, I can now fine-tune if anyone wants to see different measures now.

Those are 512 iteratons over 64x64x64 domains with a halo size of 4 in all directions (i.e. 72x72x72 buffers, computation over the central 64x64x64)

georgebisbas (approved):
I am happy with the reported relative error, so I approve

superlopuh:
What's the plan for this?

superlopuh:
Since there's no activity, I'll close this, feel free to reopen and merge!

superlopuh:
@n-io, would it be possible to split this into smaller reviewable chunks and to make the tests less integrationy and more unit-y? As it stands, if someone introduces a bug in the pass, I have no idea where to begin debugging it. It would be great to have clear inputs and outputs for each of the rewrite patterns.

n-io:
> @n-io, would it be possible to split this into smaller reviewable chunks and to make the tests less integrationy and more unit-y? As it stands, if someone introduces a bug in the pass, I have no idea where to begin debugging it.

I am faced with a very similar challenge regarding the suggestion you are making.

superlopuh:
Do you mean you don't know where to begin making the tests more targeted to the individual rewrites?

n-io:
@superlopuh What I'm not clear about is that from my pov the filecheck tests appear to already be in the form you are asking for, unit test-like rather than integration-like. For instance, the last test covers the pass operating on the `stencil.buffer` op, which hadn't been included in any previous tests. There is one for dyn_access, and some to cover small unit test-like specific scenarios.

xdsl/transforms/stencil_inlining.py
```diff
@@ -0,0 +1,368 @@
...
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: ApplyOp, rewriter: PatternRewriter, /):
+        for operand in (consumer := op).operands:
+            if isinstance(producer := operand.owner, ApplyOp):
+                if has_single_consumer(producer, consumer) and is_inlining_possible(
+                    producer, consumer
+                ):
+                    return self.inline_producer(producer, consumer, rewriter)
+        pass
```
superlopuh, line 349:
```suggestion
```

xdsl/transforms/stencil_unroll.py
```diff
@@ -42,10 +42,9 @@ def offseted_block_clone(apply: ApplyOp, unroll_offset: Sequence[int]):
                     offset_mapping = list(range(0, len(op.offset)))
                 else:
                     offset_mapping = op.offset_mapping
-
-                new_offset = [o for o in op.offset]
-                for i in offset_mapping:
-                    new_offset[i] += unroll_offset[i]
+                new_offset = [
+                    o + unroll_offset[m] for o, m in zip(op.offset, offset_mapping)
```
superlopuh, line 46:
```suggestion
                    o + unroll_offset[m] for o, m in zip(op.offset, offset_mapping, strict=True)
```

xdsl/transforms/stencil_inlining.py
```diff
@@ -0,0 +1,368 @@
...
+    """
+    # Don't inline any producer with conditional writes.
+    r = not any(
+        store_result.arg is None
+        for store_result in producer.walk()
+        if isinstance(store_result, StoreResultOp)
+    ) and not any(
+        # Don't inline any dynamic accesses.
+        isinstance(use.operation, DynAccessOp)
+        for consumer_operand in consumer.operands
```
superlopuh, line 126: nit:
```suggestion
        for consumer_operand, arg in zip(consumer.operands, consumer.region.block.args, strict=True)
```

xdsl/transforms/stencil_inlining.py
```diff
@@ -0,0 +1,368 @@
...
+    # Perform producer consumer inlining instead
+    if has_single_consumer(producer, consumer):
+        return False
+    for operand in consumer.operands:
+        if isinstance(operand.owner, Operation):
+            if (operand.owner is not producer) and is_before_in_block(
+                producer, operand.owner
+            ):
+                return False
+    return True
```
superlopuh, line 111:
```suggestion
    return not any(
        isinstance(operand.owner, Operation) and (operand.owner is not producer) and is_before_in_block(producer, operand.owner)
        for operand in consumer.operands
    )
```

superlopuh (approved):
I still think the tests could be more targeted but if you promise to help fix any bugs that might arise in this part of the codebase in the future I'm good to merge :)

xdsl/transforms/stencil_inlining.py
n-io: I think the code is doing a lookup rather than a zip at this point.

xdsl/transforms/stencil_inlining.py
superlopuh: Not sure why you resolved this but it's not a big deal

PR #2616, opened 2024-05-21
dialects: (csl) Added operations to manage symbols

dk949:
This PR adds:

* `csl.addressof`
    * Take address of a value (required for `csl.export`)
* `csl.export`
    * Exports symbol by ssa value and name (as a string) or by symbol name
* `csl.rpc`
    * equivalend to `@rpc` in the language
    * _NOTE:_ I thought `@rpc` was going away in SDK 1.1, but now I can't find a reference to it in the documentation.
* `csl.param`
    * Module import parameters (with optional default values)

### DIFF
--- a/tests/filecheck/dialects/csl/ops.mlir
+++ b/tests/filecheck/dialects/csl/ops.mlir
@@ -84,6 +84,9 @@ csl.func @initialize() {
   type = !csl.ptr<i16, #csl<ptr_kind single>, #csl<ptr_const var>>
 }> : (!csl.ptr<i16, #csl<ptr_kind single>, #csl<ptr_const var>>) -> ()
 
+%rpc_col = "test.op"() : () -> !csl.color
+"csl.rpc"(%rpc_col) : (!csl.color) -> ()
+
 }) {sym_name = "program"} :  () -> ()
 
 "csl.module"() <{kind = #csl<module_kind layout>}> ({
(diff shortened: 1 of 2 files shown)


xdsl/dialects/csl.py
```diff
@@ -643,6 +656,142 @@ class SetTileCodeOp(IRDLOperation):
...
+                )
+        elif res_elem_ty == val_ty:
+            if res_ty.kind.data != PtrKind.SINGLE:
+                raise VerifyException(
+                    f"The kind of array pointer to array has to be {PtrKind.SINGLE.value}"
+                )
+        else:
+            raise VerifyException(
+                "Contained type of the result pointer must match the contained type of the operand memref or the memref itself"
+            )
```
AntonLydike, line 733: Can you add a docstring to this method? It seems to be quite complex

tests/filecheck/dialects/csl/ops.mlir
```diff
@@ -63,11 +63,35 @@ csl.func @initialize() {
...
+
+
+
   csl.return
 }
+
+%other_global_ptr = "test.op"() : () -> !csl.ptr<i16, #csl<ptr_kind single>, #csl<ptr_const var>>
+
+"csl.export"() <{var_name = @initialize, type = () -> ()}> : () -> ()
+"csl.export"() <{var_name = @global_ptr, type = !csl.ptr<f32, #csl<ptr_kind many>, #csl<ptr_const const>>}> : () -> ()
```
AntonLydike, line 81: Where is `@global_ptr`? I think we should only allow function exporting using symbols, or do we have an op to "bind" a value to a symbol?

AntonLydike (commented):
This is already pretty neat, I just have a few questions here where I'm not 100% sure if we are modelling it the right way.

AntonLydike (approved):
Nice! Thank you!

PR #2619, opened 2024-05-21
backend: (riscv) Allocate ub and step before loop, and lb after

superlopuh:
First of two PRs fixing our register allocator, this one allows for reuse of the iteration argument, which is currently never freed.


tests/filecheck/backend/riscv/riscv_register_allocation.mlir
```diff
@@ -29,18 +29,18 @@ riscv_func.func @main() {
 //   CHECK-LIVENESS-BLOCK-NAIVE:       builtin.module {
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:    riscv_func.func @external() -> ()
 //   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:    riscv_func.func @main() {
-//   CHECK-LIVENESS-BLOCK-NAIVE-NEXT:      %{{.+}} = riscv.li 0 : () -> !riscv.reg<zero>
```
superlopuh, line 32: I think the anonymous ssa value here actually makes it harder to read the test

AntonLydike (approved):

zero9178 (approved):

PR #2624, opened 2024-05-22
core: enabling pattern rewrite for subscripted ops

n-io:
This PR enables the pattern rewriter to support matching things like:

```
@op_type_rewrite_pattern
def match_and_rewrite(
    self, op: BinaryOperation[Attribute], rewriter: PatternRewriter, /
):
   ....
```

Currently, matching is supported for un-subscripted types (like `BinaryOperation`) but pyright will note that there are incomplete type information. Adding subscripts (as shown above) will pass pyright checks and we'd like to support it by extending `isa` and by using `isa` rather than `isinstance` in the pattern rewriter.

An additional test function shows how `arith.Addf` is matched into the above type. It is an interesting case, because accessing the underlying type is not immediately straightforward, as several things need resolving:
* calling `get_args` will not work directly as it doesn't expose any args, instead we have to go to the superclass
* the superclass is a type alias which needs resolving
* the args are `Annotated` which evaluates to the type of its first param

It might be interesting to match `BinaryOperation[Float32Type]`, but note that it appears the operation does not actually appear to carry that type information.

### DIFF
--- a/tests/dialects/test_arith.py
+++ b/tests/dialects/test_arith.py
@@ -49,6 +49,8 @@ from xdsl.dialects.arith import (
 from xdsl.dialects.builtin import (
     AnyTensorType,
     AnyVectorType,
+    Float16Type,
+    Float32Type,
     FloatAttr,
     IndexType,
     IntegerType,
@@ -368,3 +370,17 @@ def test_extui_incorrect_bitwidth():
     # bitwidth of b has to be larger than the one of a
     with pytest.raises(VerifyException):
         _extui_op = ExtUIOp(a, i32).verify()
+
+
+def test_op():
+    one = Constant(FloatAttr(1.0, Float16Type()))
+    two = Constant(FloatAttr(2.0, Float32Type()))
+    addf = Addf(one, two)
+    assert isa(addf, BinaryOperation[Attribute])
+    # it'd be great if the following would pass
+    assert not isa(addf, BinaryOperation[Float16Type | Float32Type])
+    assert isa(addf, BinaryOperation)
+    assert isa(addf, FloatingPointLikeBinaryOp)
+    assert not isa(
+        addf, BinaryOperation[int]  # pyright: ignore [reportGeneralTypeIssues]
+    )
--- a/tests/dialects/test_arith.py
+++ b/tests/dialects/test_arith.py
@@ -378,7 +378,7 @@ def test_op():
     addf = Addf(one, two)
     assert isa(addf, BinaryOperation[Attribute])
     # it'd be great if the following would pass
-    assert not isa(addf, BinaryOperation[Float16Type | Float32Type])
+    # assert not isa(addf, BinaryOperation[Float16Type | Float32Type])
     assert isa(addf, BinaryOperation)
     assert isa(addf, FloatingPointLikeBinaryOp)
     assert not isa(
--- a/tests/test_is_satisfying_hint.py
+++ b/tests/test_is_satisfying_hint.py
@@ -1,24 +1,25 @@
-from typing import Any, Generic, Literal, TypeAlias, TypeVar
+from typing import Annotated, Any, Generic, Literal, TypeAlias, TypeVar
 
-from xdsl.dialects.arith import (
-    Addf,
-    BinaryOperation,
-    Constant,
-    FloatingPointLikeBinaryOp,
-)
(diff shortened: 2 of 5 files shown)


tests/test_is_satisfying_hint.py
```diff
@@ -1,8 +1,16 @@
 from typing import Any, Generic, Literal, TypeAlias, TypeVar
 
+from xdsl.dialects.arith import (
```
superlopuh, line 3: Can you please replace the arith use with a custom operation defined for this unit test? We try to avoid using non-builtin dialects for unit tests

n-io: That sounds like a clean way of testing. I replaced usages of the arith package with custom ops, and moved the code testing arith ops to the arith test file.

tests/dialects/test_arith.py
```diff
@@ -368,3 +370,17 @@ def test_extui_incorrect_bitwidth():
...
+    one = Constant(FloatAttr(1.0, Float16Type()))
+    two = Constant(FloatAttr(2.0, Float32Type()))
+    addf = Addf(one, two)
+    assert isa(addf, BinaryOperation[Attribute])
+    # it'd be great if the following would pass
+    # assert not isa(addf, BinaryOperation[Float16Type | Float32Type])
+    assert isa(addf, BinaryOperation)
+    assert isa(addf, FloatingPointLikeBinaryOp)
+    assert not isa(
+        addf, BinaryOperation[int]  # pyright: ignore [reportGeneralTypeIssues]
```
superlopuh, line 385: why the pyright ignore?

tests/dialects/test_arith.py
```diff
@@ -368,3 +370,17 @@ def test_extui_incorrect_bitwidth():
...
         _extui_op = ExtUIOp(a, i32).verify()
+
+
+def test_op():
+    one = Constant(FloatAttr(1.0, Float16Type()))
+    two = Constant(FloatAttr(2.0, Float32Type()))
+    addf = Addf(one, two)
+    assert isa(addf, BinaryOperation[Attribute])
+    # it'd be great if the following would pass
+    # assert not isa(addf, BinaryOperation[Float16Type | Float32Type])
```
superlopuh, line 381: what's the failure here?

xdsl/pattern_rewriter.py
```diff
@@ -378,6 +378,8 @@ def op_type_rewrite_pattern(
     expected_types = (expected_type,)
     if get_origin(expected_type) in [Union, UnionType]:
         expected_types = get_args(expected_type)
+    elif (t := get_origin(expected_type)) is not None:
```
superlopuh, line 381: might as well get origin only once, I recall this being an expensive function

xdsl/utils/hints.py
superlopuh, line 1: it seems like this `isa` change deserves its own PR

math-fehr (changes requested):
I feel that your change on `isa` is incorrect, or at least acts weirdly compared to what `isa` is supposed to do. `isa` should be the equivalent to `isinstance`, however in your case I think you have a bug:

My understanding is that you are doing this:
You have a generic class A, and a child class B:
```python
class A(Generic[T]):
  pass

class B(A[int]):
  pass
```
Now, you want to check if an object you have is an `A[int]`:
```python
b = B()
assert isa(b, A[int])
```
You do this by checking if the `int` you passed in your `isa` is a subclass of the `int` that is in the definition of `B`, which it is, so this is passed.

However, what if we have the following example:
```python
class Parent: pass
class Child1: pass
class Child2: pass

class A(Generic[T]):
  a: T
class B(A[Parent]):
  pass

b = B(Child1)
assert isa(b, A[Child2])
```

My understanding, is that your check will check if `Child2` is a subclass of `Parent`. However, here, what we want is that `Child2` is a subclass of `Child1`, which it is not.

Fundamentally, you are doing your check on the definition of the class, but not on the actual values that are stored by the object, if that makes sense.
Sadly, AFAIK, it is impossible to write such generic checks in Python, which is why I had to write it in a specific way for attributes, and if we want to do that for operations, we should do the same as well. However, I'm not sure how easy that is, as operations are a bit more complex than attributes.

superlopuh:
What's the plan for this? Let's close this if we're not going forward with it for the time being?

PR #2630, opened 2024-05-23
core: Expect Data's carride type to be Hashable.

PapyChacal:
Rationale reminder:
We already made `Data` a frozen dataclass, which force-provides a specific `hash` implementation assuming the data element is washable. Though we do not enforce this anywhere and it ungracefully crashes on non-compliant cases.

Also:
- Accordingly freeze AffineExpr constructs.
- Use `tuple` instead of `list` in a test-intended Data attribute.
- Tidy up the `@irdl_attr_definition` Data case:
  - Do not throw away base classes
  - Then no need to refreeze it? This silently overwrites a lot of stuff the user might want to provide on their specific Data attribute. It just needs to be Hashable, the read-only `data` part is already enforced by the `Data` base class, now carried through.
  - It typechecks :exploding_head:

### DIFF
--- a/xdsl/irdl/irdl.py
+++ b/xdsl/irdl/irdl.py
@@ -2468,12 +2468,10 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
         data_class = cast(type[Data[Any]], cls)
         data_metaclass = cast(ABCMeta, data_class.__class__)
         irdled = runtime_final(
-            dataclass(frozen=True)(
-                data_metaclass(
-                    cls.__name__,
-                    (cls, *cls.__bases__),
-                    dict(cls.__dict__),
-                )
+            data_metaclass(
+                cls.__name__,
+                (cls, *cls.__bases__),
+                dict(cls.__dict__),
             )
         )
         return cast(TypeAttributeInvT, irdled)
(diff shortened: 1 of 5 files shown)


superlopuh (approved):

xdsl/irdl/irdl.py
```diff
@@ -2465,15 +2465,16 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
...
-                    (cls,),  # pyright: ignore[reportUnknownArgumentType]
-                    dict(cls.__dict__),
-                )
+        data_class = cast(type[Data[Any]], cls)
+        data_metaclass = cast(ABCMeta, data_class.__class__)
+        irdled = runtime_final(
+            data_metaclass(
+                cls.__name__,
+                (cls, *cls.__bases__),
+                dict(cls.__dict__),
```
math-fehr, line 2474: Can you add a comment somewhere that explains what this does, for instance that it will make the class frozen because of the dataclass?

math-fehr (approved):
Looks good to me, thanks a lot!

xdsl/irdl/irdl.py
PapyChacal: Thank you for discussing this;
Explaining it, I realized that this was now only rebuilding the class itself as-is... There is no need for a dynamic class at all :yum: 
Added a comment!

PR #2631, opened 2024-05-23
transformations: Implement generic CSE.

PapyChacal:
Implement generic CSE mimicking MLIR's.
We do not have the notion of Region kind, so this implementation is pessimisticly handling everything as a Graph region. That means it is slightly less powerful than MLIR's, but as safe.

The included file check is exactly the one in MLIR, with the same expected output, unless specified otherwise on top of differing examples.

The only unsupported cases are regarding side effects; given that we only have `Pure` in xDSL (i.e., no operation of this type has any side effect ever), this does not implement more subtle cases.
Those are CSEing consecutive read-only side-effects, if not interrupted by write side-effects, and probably more importantly: recursive side effect. (A very scary way to say "I have side effect if and only if my regions contain side effects :shrug: ). That would allow  CSEing side-effect-less conditionals and what not.


xdsl/dialects/affine.py
```diff
@@ -48,6 +48,8 @@ class ApplyOp(IRDLOperation):
     map = prop_def(AffineMapAttr)
     result = result_def(IndexType)
 
+    traits = frozenset([Pure()])
+
```
PapyChacal, line 52: Chose not to split in a 2-lines PR as this had no effect elsewhere

xdsl/transforms/dead_code_elimination.py
```diff
@@ -5,24 +5,29 @@
...
 
+def is_trivially_dead(op: Operation):
+    # Check that operation is side-effect-free
+    if not op.has_trait(Pure):
+        return False
+
+    # Check whether any of the results are used
+    results = op.results
+    for result in results:
+        if len(result.uses):
```
superlopuh, line 16:
```suggestion
        if result.uses:
```

xdsl/transforms/dead_code_elimination.py
```diff
@@ -5,24 +5,29 @@
...
+            # At least one of the results is used
+            return False
+
+    return True
```
superlopuh, line 20:
```suggestion
    for result in results:
        if len(result.uses):
            # At least one of the results is used
            return False

    return not any(result.uses for result in op.results)
```

xdsl/transforms/common_subexpression_elimination.py
```diff
@@ -0,0 +1,185 @@
...
+            + sum(hash(i) for i in self.op.attributes.items())
+            + sum(hash(i) for i in self.op.properties.items())
+            + sum(hash(i.type) for i in self.op.results)
+            + sum(hash(i) for i in self.op.operands)
+        )
+
+    def __eq__(self, other: object):
+        if not isinstance(other, OperationInfo):
+            return False
+        if hash(self) != hash(other):
```
PapyChacal, line 39: Those need proper comparison too! 🤔

xdsl/transforms/common_subexpression_elimination.py
```diff
@@ -0,0 +1,185 @@
...
+            name = self.op.op_name.data
+        else:
+            name = self.op.name
+        return (
+            hash(name)
+            + sum(hash(i) for i in self.op.attributes.items())
+            + sum(hash(i) for i in self.op.properties.items())
+            + sum(hash(i.type) for i in self.op.results)
+            + sum(hash(i) for i in self.op.operands)
+        )
```
AntonLydike, line 34: May I recommend using `hash((a, b, c))` instead of `hash(a+b+c)` to combine hashes? This makes the concatenation non-symmetrical.

AntonLydike (approved):
Really nice work!

xdsl/transforms/common_subexpression_elimination.py
PapyChacal: Sure, good to know, thanks! That should help with performance if it ever comes down to that.
NB I kept the sum on the dicts because we want it to be symmetrical there.

xdsl/dialects/test.py
```diff
@@ -130,4 +173,4 @@ def print_parameter(self, printer: Printer) -> None:
             printer.print_string_literal(self.data)
 
 
-Test = Dialect("test", [TestOp, TestTermOp], [TestType])
+Test = Dialect("test", [TestOp, TestPureOp, TestTermOp], [TestType])
```
superlopuh, line 176: nit:
```suggestion
Test = Dialect("test", [TestOp, TestPureOp, TestTermOp,], [TestType,])
```

xdsl/ir/core.py
```diff
@@ -761,6 +761,17 @@ def __iter__(self) -> Iterator[SSAValue]:
...
+        if not isinstance(other, OpOperands):
+            return False
+        return (
+            self._op._operands  # pyright: ignore[reportPrivateUsage]
+            == other._op._operands  # pyright: ignore[reportPrivateUsage]
+        )
+
+    def __hash__(self):
+        return hash(self._op._operands)  # pyright: ignore[reportPrivateUsage]
+
```
superlopuh, line 774: this would be a fantastic thing to split out into its own PR with tests

PapyChacal: https://github.com/xdslproject/xdsl/pull/2639

xdsl/transforms/common_subexpression_elimination.py
```diff
@@ -0,0 +1,197 @@
...
+        # Doesmn't mean we can't just simplify what we can with our simpler model :)
+        if not op.has_trait(Pure):
+            return
+
+        # This operation rings a bell!
+        if existing := self.known_ops.get(op):
+
+            # Just replace results
+            def wasVisited(use: Use):
+                return self.known_ops.get(use.operation) is None
```
superlopuh, line 136: Seems like making `use.operation in self.known_ops` work would be preferable, hopefully not too complicated

superlopuh (approved):

xdsl/transforms/common_subexpression_elimination.py
PapyChacal: Not too complicated, as it already works :smile: Just didn't use it

PR #2632, opened 2024-05-23
dialects: arith.constant is Pure (no side effects)

PapyChacal:
Make arith.constant Pure, and fix constant folding that breaks when facing this.


math-fehr (approved):
Nice!

superlopuh (approved):

Issue #2604, opened 2024-05-18
API: `builtin.DenseArrayBase.create_dense_int_or_index` fails during verification when an `IndexType` is passed as `data_type`


watermelonwolverine:
The method name and the type hints for the `data_type` suggest that I can pass a `IndexType` as `data_type`. Doing so causes an exception during verification.

```python
indices = builtin.DenseArrayBase.create_dense_int_or_index(
    builtin.IndexType(),
    [0],
)
```

throws:

```
openpme_xdsl/transforms/convert_particles_to_ll_mlir.py:266: in match_and_rewrite
    indices = builtin.DenseArrayBase.create_dense_int_or_index(
.venv/lib/python3.10/site-packages/xdsl/dialects/builtin.py:1077: in create_dense_int_or_index
    return DenseArrayBase([data_type, ArrayAttr(attr_list)])
.venv/lib/python3.10/site-packages/xdsl/ir/core.py:627: in __init__
    super().__init__()
<string>:3: in __init__
    ???
.venv/lib/python3.10/site-packages/xdsl/ir/core.py:451: in __post_init__
    self._verify()
.venv/lib/python3.10/site-packages/xdsl/ir/core.py:666: in _verify
    super()._verify()
.venv/lib/python3.10/site-packages/xdsl/ir/core.py:456: in _verify
    self.verify()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = DenseArrayBase(parameters=(IndexType(parameters=()), ArrayAttr(data=(IntAttr(data=0),))), elt_type=IndexType(parameters=()), data=ArrayAttr(data=(IntAttr(data=0),)))

    def verify(self):
        if isinstance(self.elt_type, IntegerType):
            for d in self.data.data:
                if isinstance(d, FloatData):
                    raise VerifyException(
                        "dense array of integer element type "
                        "should only contain integers"
                    )
        else:
            for d in self.data.data:
                if isinstance(d, IntAttr):
>                   raise VerifyException(
                        "dense array of float element type "
                        "should only contain floats"
                    )
E                   xdsl.utils.exceptions.VerifyException: dense array of float element type should only contain floats
```

superlopuh:
[...]

PR #2633, opened 2024-05-23
api: Fixed builtin.DenseArrayBase not supporting IndexType

watermelonwolverine:
api: Fixed builtin.DenseArrayBase not supporting IndexType as element type
Fixes issue #2604


math-fehr (approved):
Really nice, thanks!

superlopuh (approved):

PR #2640, opened 2024-05-24
core: Implement a simplified MemoryEffect interface.

PapyChacal:
Implement a simplified MemoryEffect interface, just saying if the op has any side effect or not.
(The upstream version says if it's a write, read, alloc or free too; but that can be used already, and is part of the API of the upstream one too :slightly_smiling_face:)

Implement NoMemoryEffect and RecursiveMemoryEffect in those terms, also make Pure properly implemented in those terms, and use in CSE.

### DIFF
--- a/xdsl/traits.py
+++ b/xdsl/traits.py
@@ -34,10 +34,6 @@ class OpTrait:
 OpTraitInvT = TypeVar("OpTraitInvT", bound=OpTrait)
 
 
-class Pure(OpTrait):
-    """A trait that signals that an operation has no side effects."""
-
-
 class ConstantLike(OpTrait):
     """
     Operation known to be constant-like.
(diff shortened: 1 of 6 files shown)


xdsl/traits.py
```diff
@@ -446,3 +442,59 @@ def verify(self, op: Operation) -> None:
...
+
+def is_side_effect_free(op: Operation):
+    """
+    Boilerplate helper to check if a generic operation is side effect free for sure.
+    """
+    # If it doesn't say, safely assume it has side effects.
+    if not (trait := op.get_trait(MemoryEffect)):
+        return False
+    return not trait.has_effects(op)
+
```
superlopuh, line 471: Maybe better on MemoryEffect?

superlopuh (approved):

xdsl/traits.py
PapyChacal: Well no, the main point of the helper is to say it might have side effect if the op does not implement the MemoryEffect interface (e.g., unregistered or just experimental new dialect!)

superlopuh: I mean as a static function in the MemoryEffect class, as a namespacing thing

georgebisbas (approved):

PR #2641, opened 2024-05-24
dialects: Update arith, math, scf and stencil side-effect traits.

PapyChacal:
NB: `scf.reduce` is missing IsTerminator/HasParent traits too, but that needs updating some tests so sounded out of scope of this PR.

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from collections.abc import Mapping, Sequence
-from dataclasses import dataclass
 from typing import Annotated, Generic, Literal, TypeVar, cast, overload
 
 from xdsl.dialects.builtin import (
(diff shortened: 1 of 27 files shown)


xdsl/dialects/arith.py
```diff
@@ -225,8 +229,6 @@ def __hash__(self) -> int:
 class BinaryOperationWithFastMath(Generic[_T], BinaryOperation[_T]):
     fastmath = opt_prop_def(FastMathFlagsAttr)
 
-    traits = frozenset((Pure(),))
```
superlopuh, line 228: Why this change?

PapyChacal: Well they are not all Pure, and if we start actually using those traits, this is a timebomb!

Pure is NoMemoryEffect+AlwaysSpeculatable;
AlwaysSpeculatable implies that an operation never has undefined behaviour,
which is not the case of e.g. divisions, where dividing by 0 is UB

georgebisbas (approved):

xdsl/dialects/arith.py
superlopuh: Ah nice, thats very clear

PR #2646, opened 2024-05-27
[RFC] misc: implement PatternUnfoldListener

PapyChacal:
Currently confused by some behaviour of a pattern-based pass I'm working on, I wrote this to be able to see the steps done in order.

I think it would make sense to expose this as a CLI option of `xdsl-opt` directly, but getting there requires extra plumbing AFAIU.
Meanwhile, what do people think? This can be used in development by adding it directly to the pass you're working on.

### DIFF
--- a/xdsl/utils/diagnostic.py
+++ b/xdsl/utils/diagnostic.py
@@ -16,14 +16,9 @@ class Diagnostic:
         """Add a message to an operation."""
         self.op_messages.setdefault(op, []).append(message)
 
-    def raise_exception(
-        self,
-        message: str,
-        ir: IRNode,
-        exception_type: type[Exception] = DiagnosticException,
-        underlying_error: Exception | None = None,
-    ) -> NoReturn:
-        """Raise an exception, that will also print all messages in the IR."""
+    def get_output(self, ir: IRNode) -> str:
+        """Get all messages in the IR."""
+
         from xdsl.printer import Printer
 
         f = StringIO()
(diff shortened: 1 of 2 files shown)


xdsl/utils/diagnostic.py
```diff
@@ -16,14 +16,9 @@ def add_message(self, op: Operation, message: str) -> None:
...
-        self,
-        message: str,
-        ir: IRNode,
-        exception_type: type[Exception] = DiagnosticException,
-        underlying_error: Exception | None = None,
-    ) -> NoReturn:
-        """Raise an exception, that will also print all messages in the IR."""
+    def get_output(self, ir: IRNode) -> str:
+        """Get all messages in the IR."""
+
```
PapyChacal, line 21: This is lready a separate commit if someone wants to split it.

It is a basic factoring of the string building method, unmodified, to use it without necessarily raising an exception containing it.

superlopuh: What's the difference with `f"{ir}"` in the case of Operation? 

``` python
    def __str__(self) -> str:

        res = StringIO()
        printer = Printer(stream=res)
        printer.print_op(self)
        return res.getvalue()
```

Seems like it's only used for Operation below, anyway?

PapyChacal: Main difference is the diagnostic parameter to the printer:
```
p = Printer(stream=f, diagnostic=self, print_generic_format=True)
```
Printing the IR with diagnostic messages inline.

PapyChacal: (Also the generic syntax - because it's not generally safe to print custom syntax of unverifying IR, this can mess up error reporting quite a lot!)

superlopuh: Ah sounds good

PapyChacal:
Closing as I solved my problem and I don't see specific interest for this feature from others; happy to reopen if this changes!

PR #2647, opened 2024-05-27
backend: (csl) printing for functions, tasks and layout related things

dk949:
Printing for:
* `csl.call`
* `csl.task`
* `csl.func`
    * Merged with `csl.task`
* `csl.export`
* `csl.layout`

Also introduces a divider between the program and layout, for ease of filechecking and splitting

### DIFF
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -38,6 +38,9 @@ class CslPrintContext:
     _counter: int = field(default=0)
 
     _prefix: str = field(default="")
+    _symbols_to_export: dict[str, tuple[TypeAttribute, bool | None]] = field(
+        default_factory=dict
+    )
 
     def print(self, text: str, prefix: str = "", end: str = "\n"):
         """
@@ -114,6 +117,15 @@ class CslPrintContext:
             return f"{self._get_variable_name_for(val)}"
         else:
             return f"{intro} {self._get_variable_name_for(val)} : {self.mlir_type_to_csl_type(val.type)}"
+
+    def _export_sym_constness(self, ty: FunctionType | csl.PtrType):
+        if isinstance(ty, FunctionType):
+            return None
+        match ty.constness.data:
+            case csl.PtrConst.VAR:
+                return True
+            case csl.PtrConst.CONST:
+                return False
 
     def _get_variable_name_for(self, val: SSAValue, hint: str | None = None) -> str:
         """
@@ -299,6 +311,18 @@ class CslPrintContext:
                 case memref.GetGlobal(name_=name, memref=res):
                     # We print the array definition when the global is defined
                     self.variables[res] = name.string_value()
+                case csl.SymbolExportOp(value=val, type=ty) as exp:
+                    name = exp.get_name()
+                    q_name = f'"{name}"'
+                    self._symbols_to_export[name] = (ty, self._export_sym_constness(ty))
+                    ty = self.attribute_value_to_str(ty)
+                    if val is not None:
+                        export_val = self._get_variable_name_for(val)
+                    else:
+                        # Use symbol ref name if operand not provided
+                        export_val = name
+                    with self._in_block("comptime"):
+                        self.print(f"@export_symbol({export_val}, {q_name});")
                 case anyop:
                     self.print(f"unknown op {anyop}", prefix="//")
 
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -103,6 +103,10 @@ class CslPrintContext:
     def _print_bind_task(
         self, name: str, kind: csl.TaskKind, id: csl.ColorIdAttr | None
     ):
+        """
+        Generate a call to `@bind_<kind>_task` if task ID was specified as a
+        property of the task. Otherwise we assume binding will be done at runtime.
+        """
         if id is None:
             return
         with self._in_block("comptime"):
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -85,9 +85,9 @@ class CslPrintContext:
             if len(ftyp.outputs) == 0
             else self.mlir_type_to_csl_type(ftyp.outputs.data[0])
         )
-        self.print(f"\n{kind} {name.data}({args}) {ret} {{")
-        self.descend().print_block(bdy.block)
-        self.print("}")
+        signature = f"\n{kind} {name.data}({args}) {ret}"
+        with self.descend(signature) as inner:
+            inner.print_block(bdy.block)
 
     def _wrap_task_id(self, kind: csl.TaskKind, id: int) -> str:
         """
@@ -109,8 +109,8 @@ class CslPrintContext:
         """
         if id is None:
             return
-        with self._in_block("comptime"):
-            self.print(
+        with self.descend("comptime") as inner:
+            inner.print(
                 f"@bind_{kind.value}_task({name}, @get_{kind.value}_task_id({self._wrap_task_id(kind, id.value.data)}));"
             )
 
@@ -333,11 +333,9 @@ class CslPrintContext:
                     lower_name, upper_name, stp_name, idx_name = map(
                         self._get_variable_name_for, (lower, upper, stp, idx)
                     )
-                    self.print(
-                        f"\nfor(@range({idx_type}, {lower_name}, {upper_name}, {stp_name})) |{idx_name}| {{"
-                    )
-                    self.descend().print_block(bdy.block)
-                    self.print("}")
+                    loop_definition = f"\nfor(@range({idx_type}, {lower_name}, {upper_name}, {stp_name})) |{idx_name}|"
+                    with self.descend(loop_definition) as inner:
+                        inner.print_block(bdy.block)
                 case scf.Yield():
                     pass
                 case memref.Global(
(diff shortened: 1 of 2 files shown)


xdsl/backend/csl/print_csl.py
```diff
@@ -42,6 +50,45 @@ def print(self, text: str, prefix: str = "", end: str = "\n"):
...
 
+    @contextmanager
+    def _in_block(self, block_name: str):
+        self.print(f"{block_name} {{")
+        old_prefix = self._prefix
+        self._prefix += self._INDENT
+        yield
+        self._prefix = old_prefix
+        self.print("}")
+        pass
```
AntonLydike, line 61: Does this replace the previous `descend` method?

xdsl/backend/csl/print_csl.py
```diff
@@ -42,6 +50,45 @@ def print(self, text: str, prefix: str = "", end: str = "\n"):
...
+            for arg in bdy.block.args
+        )
+        ret = (
+            "void"
+            if len(ftyp.outputs) == 0
+            else self.mlir_type_to_csl_type(ftyp.outputs.data[0])
+        )
+        self.print(f"\n{introducer} {name.data}({args}) {ret} {{")
+        self.descend().print_block(bdy.block)
+        self.print("}")
```
AntonLydike, line 77: I would really like to have all functions that print prefixed with `print_`

xdsl/backend/csl/print_csl.py
```diff
@@ -42,6 +50,45 @@ def print(self, text: str, prefix: str = "", end: str = "\n"):
...
+            return f"@get_color({id})"
+        return str(id)
+
+    def _bind_task(self, name: str, kind: csl.TaskKind, id: csl.ColorIdAttr | None):
+        if id is None:
+            return
+        with self._in_block("comptime"):
+            self.print(
+                f"@bind_{kind.value}_task({name}, @get_{kind.value}_task_id({self._wrapp_task_id(kind, id.value.data)}));"
+            )
```
AntonLydike, line 90: Same here, something like `print_task_bind`

xdsl/backend/csl/print_csl.py
```diff
@@ -72,6 +119,15 @@ def _var_use(
...
 
+    def _export_sym_constness(self, ty: FunctionType | csl.PtrType):
+        if isinstance(ty, FunctionType):
+            return None
+        match ty.constness.data:
+            case csl.PtrConst.VAR:
+                return True
+            case csl.PtrConst.CONST:
+                return False
+
```
AntonLydike, line 130: I don't quite understand what this function does. Type annotations and docstrings would go a long way in making this more reviewable!

xdsl/backend/csl/print_csl.py
dk949: Not really, this just wraps a set of prints in `<block_name> { ... }`. I could combine it with descend, but it's more for cases where we don't have an MLIR `Block` and need to wrap something in braces.

xdsl/backend/csl/print_csl.py
dk949: When exporting symbols we have to specify if they are host-writable, except for functions, where it's a compile error to do so. 

Will add a docstring and return type

xdsl/backend/csl/print_csl.py
```diff
@@ -266,15 +346,36 @@ def descend(self) -> CslPrintContext:
             output=self.output,
             variables=self.variables.copy(),
             _counter=self._counter,
-            _prefix=self._prefix + "  ",
+            _prefix=self._prefix + self._INDENT,
         )
 
 
+def _get_layout_program(ops: BlockOps) -> tuple[csl.CslModuleOp, csl.CslModuleOp]:
```
AntonLydike, line 353: I think it'd be more appropriate to pass the module here instead of the iterable of it's ops?

AntonLydike: Also, docstrings!

dk949:
@AntonLydike for the `print_` prefix, should these still be private (i.e. `_print_`)?

AntonLydike (changes requested):
I think this is overall a pretty good first printer PR, although I would really really like to see more docstrings here!

AntonLydike:
> @AntonLydike for the `print_` prefix, should these still be private (i.e. `_print_`)?

I think they can easily be private as well!

xdsl/backend/csl/print_csl.py
AntonLydike: I see it only being used for comptime regions, if that is the case, maybe don't need a form this generic? Just some food for thought.

dk949: It's also used for `layout`, though I can rewrite that as `descend`, since layout does have an MLIR `Block`, then make `_in_block` `_in_comptime` instead, since there are no other CSL block types.

dk949: Technically, descend can be used to replace all uses of `_in_block`, but this way felt cleaner, since it's always the same pattern:
```
<block-name> {
    <statements>
}
```

dk949: What would your thoughts be on:

```python
with self._in_block("comptime") as in_comptime:
    in_comptime.print(f"@export_symbol({export_val}, {q_name});")
```

Then we can reuse `descend` in `_in_block` and avoid repetition there?

AntonLydike: Maybe we can instead replace descend with this instead then? I also feel that the context manager "feels" nicer, I would just argue that we should focus on one method for now.

xdsl/backend/csl/print_csl.py
```diff
@@ -52,6 +52,12 @@ def print(self, text: str, prefix: str = "", end: str = "\n"):
 
     @contextmanager
     def _in_block(self, block_name: str):
+        """
+        To be used in a `with` statement.
+        Surrounds everything in the `with` with curly braces and .
+
+        This is used instead of descend when
```
AntonLydike, line 59: premature end of docstring, I think

xdsl/backend/csl/print_csl.py
```diff
@@ -60,9 +66,16 @@ def _in_block(self, block_name: str):
         self.print("}")
         pass
 
-    def _task_or_fn(
-        self, introducer: str, name: StringAttr, bdy: Region, ftyp: FunctionType
+    def _print_task_or_fn(
+        self,
+        kind: Literal["fn"] | Literal["task"],
```
AntonLydike, line 71: Can be simplified to `Literal["fn", "task"]`

AntonLydike (commented):
Much better already!

xdsl/backend/csl/print_csl.py
```diff
@@ -42,7 +50,75 @@ def print(self, text: str, prefix: str = "", end: str = "\n"):
...
+        ret = (
+            "void"
+            if len(ftyp.outputs) == 0
+            else self.mlir_type_to_csl_type(ftyp.outputs.data[0])
+        )
+        self.print(f"\n{kind} {name.data}({args}) {ret} {{")
+        self.descend().print_block(bdy.block)
+        self.print("}")
+
+    def _wrapp_task_id(self, kind: csl.TaskKind, id: int) -> str:
```
AntonLydike, line 92: nit: typo - should be `wrap_task_id`

xdsl/backend/csl/print_csl.py
```diff
@@ -266,15 +385,43 @@ def descend(self) -> CslPrintContext:
...
     Takes a module op and prints it to the given output stream.
     """
     ctx = CslPrintContext(output)
-    for mod in prog.body.block.ops:
-        assert isinstance(mod, csl.CslModuleOp)
-        ctx.print_block(mod.body.block)
+    layout, program = _get_layout_program(prog)
+    ctx.print_block(program.body.block)
+    ctx.print(ctx.DIVIDER)
+    ctx.print_block(layout.body.block)
```
AntonLydike, line 427: I like the previous approach more of iterating over the modules, I feel this provides more flexibility for e.g. multiple programs or multiple layouts. This feels overly restrictive

dk949: We have to print program first as that's when we collect symbols to be exported.

We can allow for multiple programs, but there still needs to be some logic to make sure they all get processed before the one layout.

xdsl/backend/csl/print_csl.py
```diff
@@ -361,32 +359,53 @@ def print_block(self, body: Block):
...
                             # If specified, get mutability as true/false from python bool
                             mut = str(val[1]).lower() if val[1] is not None else ""
-                            self.print(f'@export_name("{name}", {ty}, {mut});')
+                            inner.print(f'@export_name("{name}", {ty}, {mut});')
                 case anyop:
                     self.print(f"unknown op {anyop}", prefix="//")
 
-    def descend(self) -> CslPrintContext:
+    @contextmanager
+    def descend(self, block_start: str = ""):
```
AntonLydike, line 376: nit: you can add a type annotation here so that we know the type of the context manager:

```py
from typing import ContextManager
# ...
@contextmanager
def descend(self, block_start: str = "") ->  ContextManager[CslPrintContext]:
    ....
```

AntonLydike (approved):
I think this is in a good-to-go state now!

xdsl/backend/csl/print_csl.py
superlopuh: I'm curious whether this will work, I would have expected it to rather be `-> Iterator[CslPrintContext]`

AntonLydike: iirc pyright (at least when I tested it) were happy with both, but I find `ContextManager` to be the most correct imo

dk949: When I try `ContextManager[CslPrintContext]`, pyright produces all sorts of errors. ruff also complains that `typing.ContextManager` is deprecated in favor of `AbstractContextManager`, `AbstractContextManager[CslPrintContext]` causes all the same errors, since it's not compatible with `Generator`.

`Iterator[CslPrintContext]` works, as does `Generator[CslPrintContext, None, None]`, but I think both of these make the function signature more confusing. Personally, I would leave the return type auto-deduced, and rely on the docstring

dk949:
> I think this is in a good-to-go state now!

I forgot to remove the old `_in_block` method, but I recon it's good now.

xdsl/backend/csl/print_csl.py
superlopuh: Yeah, the type annotation is for the function passed into the decorator, so `ContextManager` is incorrect

PR #2648, opened 2024-05-28
dialects: (csl) Data structure descriptors

n-io:
A basic implementation of dsd types and the `@get_dsd`. Custom array index expressions are not yet supported, and the current version should be understood (and printed) as accessing `A[i,j,k]` according to the number of dimensions and generate induction vars accordingly.

### DIFF
--- a/tests/filecheck/dialects/csl/ops.mlir
+++ b/tests/filecheck/dialects/csl/ops.mlir
@@ -70,10 +70,10 @@ csl.func @initialize() {
     %many_arr_ptr = "csl.addressof"(%arr) : (memref<10xf32>) -> !csl.ptr<f32, #csl<ptr_kind many>, #csl<ptr_const const>>
     %single_arr_ptr = "csl.addressof"(%arr) : (memref<10xf32>) -> !csl.ptr<memref<10xf32>, #csl<ptr_kind single>, #csl<ptr_const const>>
 
-    %dsd_1d = "csl.get_dsd"(%arr, %scalar) : (memref<10xf32>, i32) -> !csl<dsd mem1d_dsd>
-    %dsd_2d = "csl.get_dsd"(%arr, %scalar, %scalar) : (memref<10xf32>, i32, i32) -> !csl<dsd mem4d_dsd>
-    %dsd_3d = "csl.get_dsd"(%arr, %scalar, %scalar, %scalar) : (memref<10xf32>, i32, i32, i32) -> !csl<dsd mem4d_dsd>
-    %dsd_4d = "csl.get_dsd"(%arr, %scalar, %scalar, %scalar, %scalar) : (memref<10xf32>, i32, i32, i32, i32) -> !csl<dsd mem4d_dsd>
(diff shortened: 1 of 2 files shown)


tests/filecheck/dialects/csl/ops.mlir
```diff
@@ -70,6 +70,10 @@ csl.func @initialize() {
     %many_arr_ptr = "csl.addressof"(%arr) : (memref<10xf32>) -> !csl.ptr<f32, #csl<ptr_kind many>, #csl<ptr_const const>>
     %single_arr_ptr = "csl.addressof"(%arr) : (memref<10xf32>) -> !csl.ptr<memref<10xf32>, #csl<ptr_kind single>, #csl<ptr_const const>>
 
+    %dsd_1d = "csl.get_dsd"(%arr, %scalar) : (memref<10xf32>, i32) -> !csl<dsd mem1d_dsd>
+    %dsd_2d = "csl.get_dsd"(%arr, %scalar, %scalar) : (memref<10xf32>, i32, i32) -> !csl<dsd mem4d_dsd>
```
AntonLydike, line 74: Two questions here: 
1. Can we not just read the sizes from the memref type?
2. Can we have two sizes passed to the get_dsd if the underlying array is 1D?

AntonLydike (approved):
If this is all that is needed to get DSDs supported in xDSL then that's amazing! I'm afraid that I don't know much about the internals of DSDs though, so for that @dk949 is probably the better conversation partner!

From my side this looks good!

tests/filecheck/dialects/csl/ops.mlir
n-io: 1. What if you don't want to iterate over the entire buffer?
2. It will fail to verify.

n-io:
> > > Would `fabin_dsd` and `fabout_dsd` be part of the same type and op, or should they have their own?
> > 
> > 
> > Since they need different fields specified, making it a separate op is actually a viable option.
> 
> In that case it would make sense to rename this op (and class) something like `csl.get_mem_dsd`.
> 
> I think it makes sense for both ops to still return a `csl.dsd`, since all the maths builtins accept either a mem dsd or a fab dsd.
> 
> EDIT: by both ops I mean this and the hypothetical future `csl.get_fab_dsd`.

I've added `fabin_dsd` and `fabout_dsd` into the same `get_dsd` op. It ended up with a lot of the fields being optional and needs verification that all (+only) the right fields are populated.

xdsl/dialects/csl.py
```diff
@@ -656,6 +673,41 @@ class SetTileCodeOp(IRDLOperation):
...
+    """
+
+    name = "csl.get_dsd"
+
+    array_var = operand_def(MemRefType)
+    sizes = var_operand_def(IntegerType)
+
+    result = result_def(DsdType)
+    # ind_vars # todo: generate
+    # expr     # todo: accept as arg
```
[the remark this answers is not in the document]
n-io: Added strides and offsets.

n-io:
`GetDsdOp` is now an abstract base class and there are subclasses for `csl.get_mem_dsd` and `csl.get_fab_dsd`. Overall, this looks a lot cleaner.

PR #2657, opened 2024-05-29
transformations: scf-for-loop-folding handle iter args

superlopuh:


PapyChacal (commented):
This seems to assume iteration arguments are just passed around as-is and without any permutation, which might be okay for the use case, but is not checked by the pattern currently!

superlopuh:
Fantastic spot, thank you!

PapyChacal (approved):
Nice, LGTM!

PR #2660, opened 2024-05-29
core: unify (Pattern)Rewriter insertion and inlining APIs around InsertPoint

### LINKED ISSUE #859: `PatternRewriter` and `Rewriter` have different orders (and signatures) for `insert_op_after`
PatternRewriter:
```python
def insert_op_before(self, op: Operation | list[Operation], target_op: Operation):
```

Rewriter:
```python
def insert_op_before(op: Operation, new_op: Operation):
```

This means that `PatternRewriter.insert_op_before(new_op, target)` has reversed semantics to `Rewriter.insert_op_before(new_op, target)` (which will actually raise an exception, luckily enough).

Ideally we would get these APIs to agree to some extent...

superlopuh:
This reduces the surface area of the Rewriter and PatternRewriter. Note that it doesn't modify the `matched_op` helpers, partially so that the diff doesn't baloon, and partially because I think they are actually quite helpful as they are. At least I'd love to talk about the matched_op helpers separately from this PR, if possible.

Resolves #859

### DIFF
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -251,18 +251,6 @@ class PatternRewriter(PatternRewriterListener):
         self._replace_all_uses_with(arg, None, safe_erase=safe_erase)
         arg.block.erase_arg(arg, safe_erase)
 
-    def inline_block_at_location(
-        self,
-        block: Block,
-        insertion_point: InsertPoint,
-        arg_values: Sequence[SSAValue] = (),
-    ):
-        """
-        Move the block operations to the specified insertion point.
-        """
-        self.has_done_action = True
-        Rewriter.inline_block_at_location(block, insertion_point, arg_values=arg_values)
-
     def inline_block_at_end(
         self, block: Block, target_block: Block, arg_values: Sequence[SSAValue] = ()
     ):
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -257,7 +257,7 @@ class PatternRewriter(PatternRewriterListener):
         self._replace_all_uses_with(arg, None, safe_erase=safe_erase)
         arg.block.erase_arg(arg, safe_erase)
 
-    def inline_block_at_location(
+    def inline_block(
         self,
         block: Block,
         insertion_point: InsertPoint,
@@ -267,7 +267,7 @@ class PatternRewriter(PatternRewriterListener):
         Move the block operations to the specified insertion point.
         """
         self.has_done_action = True
-        Rewriter.inline_block_at_location(block, insertion_point, arg_values=arg_values)
+        Rewriter.inline_block(block, insertion_point, arg_values=arg_values)
 
     def inline_block_at_end(
         self, block: Block, target_block: Block, arg_values: Sequence[SSAValue] = ()
@@ -276,7 +276,7 @@ class PatternRewriter(PatternRewriterListener):
         Move the block operations to the end of another block.
         This block should not be a parent of the block to move to.
         """
-        self.inline_block_at_location(
+        self.inline_block(
             block, InsertPoint.at_end(target_block), arg_values=arg_values
         )
 
@@ -287,7 +287,7 @@ class PatternRewriter(PatternRewriterListener):
         Move the block operations to the start of another block.
         This block should not be a parent of the block to move to.
         """
-        self.inline_block_at_location(
+        self.inline_block(
             block, InsertPoint.at_start(target_block), arg_values=arg_values
         )
 
@@ -307,9 +307,7 @@ class PatternRewriter(PatternRewriterListener):
         Move the block operations before the given operation.
         The block should not be a parent of the operation.
         """
-        self.inline_block_at_location(
-            block, InsertPoint.before(op), arg_values=arg_values
-        )
+        self.inline_block(block, InsertPoint.before(op), arg_values=arg_values)
 
     def inline_block_after_matched_op(
         self, block: Block, arg_values: Sequence[SSAValue] = ()
@@ -327,9 +325,7 @@ class PatternRewriter(PatternRewriterListener):
         Move the block operations after the given operation.
         The block should not be a parent of the operation.
         """
-        self.inline_block_at_location(
-            block, InsertPoint.after(op), arg_values=arg_values
-        )
+        self.inline_block(block, InsertPoint.after(op), arg_values=arg_values)
 
     def move_region_contents_to_new_regions(self, region: Region) -> Region:
         """Move the region blocks to a new region."""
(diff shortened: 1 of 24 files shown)


xdsl/pattern_rewriter.py
```diff
@@ -288,7 +304,9 @@ def inline_block_before(
         The block should not be a parent of the operation.
         """
         self.has_done_action = True
-        Rewriter.inline_block_before(block, op, arg_values=arg_values)
+        Rewriter.inline_block_at_location(
+            block, InsertPoint.before(op), arg_values=arg_values
+        )
```
math-fehr, line 309: I'm not sure what to think here to be honest.
I feel that it was better and more concise previously?

I understand you want to have the API simpler, but in that example I'm not really sure I see why it is beneficial? Having concise methods is really useful for users, as writing everytime `InsertPoint` is a bit annoying in these cases?

I'm wondering what the others are thinking? If they think it's fine I'm fine having this, but I would make that decision consciously at least

superlopuh: As a client of APIs, I definitely prefer fewer methods and data structures to think about. Another consideration for me is that `Rewriter` is a lower-level API than `PatternRewriter`, which still has the extended API.

superlopuh: Just to make my position clear, I think that the API with a single method and a single datastructure is more ergonomic in general than a collection of helpers.

superlopuh: Happy to wait for @AntonLydike and @PapyChacal's opinions, might as well ping @webmiche, @compor, and @zero9178, in case they also have strong opinions.

math-fehr: Yeah, we might also just talk about it on Monday for our meeting, as this is probably a general position we should take.
For instance, this would mean removing `rewriter.replace_matched_op` as well, as this is also an API we do not *need*.

superlopuh: I'm not sure about that, I think we can also have a clear distinction between a pattern rewriter and a plain rewriter. Also, the matched op seems like a distinct thing, it's no longer about arbitrary locations in code, but about the thing you're locally rewriting.

xdsl/rewriter.py
```diff
@@ -262,16 +215,6 @@ def insert_ops_at_location(ops: Sequence[Operation], insertion_point: InsertPoin
         else:
             insertion_point.block.add_ops(ops)
 
-    @staticmethod
-    def insert_op_after(op: Operation, new_op: Operation):
-        """Inserts a new operation after another operation."""
-        Rewriter.insert_ops_at_location((new_op,), InsertPoint.after(op))
-
-    @staticmethod
-    def insert_op_before(op: Operation, new_op: Operation):
```
superlopuh, line 271: Please note that this also removes a weird feature of our APIs, where `Rewriter` and `PatternRewriter` flip the order of the `target` op and `new_op` in `inser_op_before`, the `InsertPoint` removes the possibility of ambiguity about which is which.
#859

xdsl/pattern_rewriter.py
PapyChacal: I do like this change; I understand the big discussion about a general direction, I do think it might be overstating it a little, but if I had to choose, I'd rather have the style proposed here by @superlopuh honestly.

Might be a bit overfitted, but when I type with autocompletion, it's cognitively heavy (more than necessary IMO) to read all the variants and pick exactly the right one. With this change, it goes more like "Oh yes, I just want to inline a block", then see I'm asked a location, and pick the option that makes sense for the location *specifically*.

those more specific helpers are just one liners, and can be used in individual rewrites if they save some LOC in specific transformations, IMO :slightly_smiling_face:

PapyChacal (approved):

math-fehr:
Just writing a note here, I think that if we remove these methods, we have to remove the same methods from the `PatternRewriter`, otherwise we will be quite inconsistent. I think we should have a proper discussion for that as well, because consistency is really important here. And to me, that would include the `inline_after_matched_op` for instance.

Also, how do we plan to let new users find how to use this method?
Before, they would just see the method `insert_op_before` and that would be relatively clear. Now, they have to additionally understand how to create an `InsertPoint` with the `InsertPoint.before(op)` method, which I think just adds friction for newcomers.
While I understand that you want a simple API, a simple API sometimes makes the learning curve worse, if that makes sense.

superlopuh:
I agree, and that's why I would actually recommend to leave the `matched_op` API as-is. The InsertPoint I really don't think is that difficult for newcomers to understand, it's a fairly self-explanatory thing, in my opinion.

math-fehr:
But then, why would we keep `rewriter.replace_matched_op(new_op)` when it can be written with `rewriter.replace_op(new_op, matched_op)`?
or `rewriter.insert_after_matched_op(new_op)` compared to `rewriter.insert(new_op, InsertPoint.after(matched_op))` ?

To me, that is the exact thing we want to remove from the PatternRewriter no? Otherwise, the API looks kind of inconsistent?
I think we can remove these as safely as the ones you currently want to remove, as the `matched_op` is anyway passed as the `match_and_rewrite` method, and can be accessed with `rewriter.matched_op` if we want to.

superlopuh:
I see your point. In that case, I'll give that a go so we can look at the diff, I'll make this into a draft in the meantime.

PapyChacal (approved):

xdsl/rewriter.py
AntonLydike: I Just stumbled on this last week again. Thank you so much :heart:

AntonLydike (approved):
Not sure I'm a giant fan of having the function be `insert_ops` (plural), but that's so super duper bikeshedding that I will just say lgtm

xdsl/rewriter.py
```diff
@@ -255,22 +257,24 @@ def insert_block_before(block: Block | list[Block], target: Block):
         region.insert_block(block_list, pos)
 
     @staticmethod
-    def insert_ops_at_location(ops: Sequence[Operation], insertion_point: InsertPoint):
+    def insert_ops(ops: Sequence[Operation], insertion_point: InsertPoint):
         """Insert operations at a certain location in a block."""
         if insertion_point.insert_before is not None:
             insertion_point.block.insert_ops_before(ops, insertion_point.insert_before)
         else:
             insertion_point.block.add_ops(ops)
```
AntonLydike, line 265: Nevermind, I will comment on the naming. Can we have this be `insert_op(op: Operation | Sequence[Operation], ...) to mirror the other API?

superlopuh: Messaged in Zulip to ask whether to do this before or after this one is in, or as part of this PR

PR #2665, opened 2024-05-30
dialects: (memref_stream) use memref_stream iterator type, not linalg

superlopuh:
We don't support window iterators. I also want to try adding our own "interleave" iterator type to model our current interleaving of operations.


PapyChacal (approved):
LGTM code-wise;
not sure about the interleaving iterator, that would justify creating a new attribute type. Otherwise, I think I'd rather restrict to the ones you want through verification!

superlopuh:
I still have to play around with it to see what the best direction to go is

PR #2668, opened 2024-05-30
dialects: (riscv) mark assembly region as isolated from above

superlopuh:
Mostly impacts SSA counters in filechecks, and one assembly printing test.


PapyChacal (approved):
Yay!

PR #2669, opened 2024-05-30
dialects: (riscv_func) mark riscv_func.func as isolated from above

superlopuh:
Mostly filecheck ssa name changes again, had to move a few values around, found one "bug" in an assembly printing test.


PapyChacal (approved):
Yay!

PR #2671, opened 2024-05-30
tests: Add lit support for spaces in folders and files

math-fehr:
This PR adds support in lit for spaces in folders and files.
It is an alternative implementation of #2628, where we change the configuration file rather than expecting
tests to wrap `%s` with quotes.

Note that this is only implemented for `%s` and `%S`, as these are the only substitutions that we are using.


tests/filecheck/test spaces/file with space.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: xdsl-opt %s | filecheck %s
+// RUN: xdsl-opt %S/file\ with\ space.mlir | filecheck %s
+
+// Check that files (and directories) with spaces work with filecheck
+
+builtin.module {}
+
+// CHECK: builtin.module
```
superlopuh, line 8:
```suggestion
// CHECK: builtin.module

```

tests/filecheck/lit.cfg
```diff
@@ -19,3 +18,15 @@ if "COVERAGE" in lit_config.params:
...
+def my_sh_test_execute(self, test, litConfig):
+    sourcepath = test.getSourcePath()
+    sourcedir = os.path.dirname(sourcepath)
+    self.extra_substitutions.append(('%s', sourcepath.replace(" ", "\\ ")))
+    self.extra_substitutions.append(('%S', sourcedir.replace(" ", "\\ ")))
+    return shtest_execute(self, test, litConfig)
+
+lit.formats.ShTest.execute = my_sh_test_execute
+
+config.test_format = lit.formats.ShTest(preamble_commands=[f"cd {xdsl_src}"])
```
superlopuh, line 32:
```suggestion
config.test_format = lit.formats.ShTest(preamble_commands=[f"cd {xdsl_src}"])

```

superlopuh (approved):

math-fehr:
Can anyone of you manage to reproduce this bug?
I don't seem to reproduce it locally?

superlopuh:
I ran it locally, and it didn't fail in the same place, but the new test fails:
``
xdsl-opt /Users/sasha/Developer/xdslproject/xdsl/tests/filecheck/test spaces/file with space.mlir | filecheck /Users/sasha/Developer/xdslproject/xdsl/tests/filecheck/test spaces/file with space.mlir
# executed command: xdsl-opt /Users/sasha/Developer/xdslproject/xdsl/tests/filecheck/test spaces/file with space.mlir
# .---command stderr------------
# | usage: xdsl-opt [-h] [-f {mlir}] [--disable-verify]
# |                 [--allow-unregistered-dialect] [--no-implicit-module]
# |                 [-t {mlir,riscv-asm,x86-asm,riscemu,csl}] [-o OUTPUT_FILE]
# |                 [-p PASSES] [--print-between-passes] [--verify-diagnostics]
# |                 [--parsing-diagnostics] [--split-input-file]
# |                 [--print-op-generic] [--print-no-properties]
# |                 [--print-debuginfo] [-v]
# |                 [input_file]
# | xdsl-opt: error: unrecognized arguments: spaces/file with space.mlir
# `-----------------------------
# error: command failed with exit status: 2
# executed command: filecheck /Users/sasha/Developer/xdslproject/xdsl/tests/filecheck/test spaces/file with space.mlir
# .---command stdout------------
# | Could not open check file '/Users/sasha/Developer/xdslproject/xdsl/tests/filecheck/test': No such file or directory
# `-----------------------------
# error: command failed with exit status: 2
```

superlopuh:
Here is a relevant LLVM PR: https://reviews.llvm.org/D29185 In it, they recommend not adding any files with spaces to the repo, which makes sense to me. They also mention `%^s`, which doesn't work for me locally. I'd say let's stick to the same policy of no files with spaces, and maybe think of another way of solving cloning xdsl into a repo with spaces.

superlopuh:
https://discourse.llvm.org/t/spaces-in-path-with-lit-tests/79283

math-fehr:
Yeah, let's stop trying this here and instead fix this in lit. For now we'll just have the policy of not having spaces in directories, which I believe is sensible.
Sad to see the world is still not ready for spaces in files/directory.

PR #2675, opened 2024-05-31
core: implement RangeConstraint and use it without functional change

PapyChacal:
Had this in mind for a while now. This PR offers an implementation of RangeConstraint: a constraint not over single attributes, but over *sequences* of attributes.

Make IRDL operands and results use RangeConstraint instead of AttrConstraint internally. For a non-functional change, implement two basic RangeConstraints:

- `RangeOf`: constrain an attribute sequence to any number of attributes respecting the passed AttrConstraint.
- `SingleOf` : constrain an attribute sequence to a single attribute respecting the passed AttrConstraint.

I.e., `RangeOf` simply reflects the current behaviour of variadic/optional operands/results, and `SingleOf` redundantly mimic the current behaviour of non-variadic operands/results. (IRDL already checks lengths! The redundancy is only here for clarity, and paving the way to different choices if we want. Not sure we'll keep it!)

So all the currently used AttrConstraint in dialects are implicitly wrapped in either of those RangeConstraints, for an invisible first introduction of the concept :slightly_smiling_face: 

Why bother? We can now declaratively reason about much wider properties, e.g, `scf.for` results types having to match element wise the iter_args :slightly_smiling_face: Only needs the introduction of the corresponding RangeVariables.


xdsl/irdl/irdl.py
```diff
@@ -1805,51 +1926,61 @@ def get_operand_result_or_region(
...
-    construct: VarIRConstruct,
+    construct: Literal[VarIRConstruct.OPERAND, VarIRConstruct.RESULT],
     constraint_vars: dict[str, Attribute],
 ) -> None:
     """Verify the argument list of an operation."""
     arg_sizes = get_variadic_sizes(op, op_def, construct)
     arg_idx = 0
     var_idx = 0
-    args = get_op_constructs(op, construct)
+    args = cast(
```
superlopuh, line 1949: why the casts?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -207,14 +203,22 @@ def resolve_operand_types(self, state: ParsingState, op_def: OpDef) -> None:
...
+                    range_length, state.constraint_variables
+                )
+                if isinstance(operand_def, OptionalDef):
+                    operand_type = (
+                        list[Attribute | None]()
+                        if len(operand_type) == 0
+                        else operand_type[0]
+                    )
+                elif isinstance(operand_def, VariadicDef):
+                    operand_type = cast(list[Attribute | None], operand_type)
```
superlopuh, line 218: also surprised you need the cast here

superlopuh: and below

xdsl/irdl/irdl.py
```diff
@@ -429,6 +429,115 @@ def infer(self, constraint_vars: dict[str, Attribute]) -> Attribute:
...
 
+class RangeConstraint(ABC):
+
+    @abstractmethod
+    def verify(
+        self, attrs: Sequence[Attribute | None], constraint_vars: dict[str, Attribute]
+    ) -> None:
+        """
+        Check if the range satisfies the constraint,
+        or raise an exception otherwise.
```
math-fehr, line 440: Weird line break?

xdsl/irdl/irdl.py
```diff
@@ -429,6 +429,115 @@ def infer(self, constraint_vars: dict[str, Attribute]) -> Attribute:
         return self.constr.infer(constraint_vars)
 
 
+class RangeConstraint(ABC):
+
+    @abstractmethod
+    def verify(
+        self, attrs: Sequence[Attribute | None], constraint_vars: dict[str, Attribute]
```
math-fehr, line 436: I guess the `constraint_vars` is going to be changed later on when you create the new ranges variables?

xdsl/irdl/irdl.py
```diff
@@ -429,6 +429,115 @@ def infer(self, constraint_vars: dict[str, Attribute]) -> Attribute:
...
+    ) -> None:
+        """
+        Check if the range satisfies the constraint,
+        or raise an exception otherwise.
+        The range can contain Nones, which represent an attribute not to be checked.
```
math-fehr, line 441: I forgot if you told me before, but why is there possibly attributes that shouldn't be checked?

xdsl/irdl/irdl.py
```diff
@@ -429,6 +429,115 @@ def infer(self, constraint_vars: dict[str, Attribute]) -> Attribute:
...
+        Raises an exception if the range cannot be inferred. If `can_infer`
+        returns `True` with the given constraint variables, this method should
+        not raise an exception.
+        """
+        raise ValueError("Cannot infer range from constraint")
+
+
+@dataclass
+class RangeOf(RangeConstraint):
+
```
math-fehr, line 475:
```suggestion
"""Constraint each element in a range to satisfy a given constraint."""
```

xdsl/irdl/irdl.py
```diff
@@ -429,6 +429,115 @@ def infer(self, constraint_vars: dict[str, Attribute]) -> Attribute:
...
+    def infer(
+        self, length: int, constraint_vars: dict[str, Attribute]
+    ) -> list[Attribute]:
+        return [self.constr.infer(constraint_vars)] * length
+
+
+@dataclass
+class SingleOf(RangeConstraint):
+
```
math-fehr, line 500:
```suggestion
"""Constraint a range to only contain a single element, which should satisfy a given constraint."""
```

xdsl/irdl/irdl.py
```diff
@@ -429,6 +429,115 @@ def infer(self, constraint_vars: dict[str, Attribute]) -> Attribute:
...
+    if isinstance(attr, RangeConstraint):
+        return attr
+    return RangeOf(attr_constr_coercion(attr))
+
+
+def single_range_constr_coercion(
+    attr: Attribute | type[Attribute] | AttrConstraint,
+) -> RangeConstraint:
+    if isinstance(attr, SingleOf):
+        return attr
```
math-fehr, line 536: This should not happen, as `SingleOf` is a `RangeConstraint`.

xdsl/irdl/irdl.py
```diff
@@ -1513,10 +1634,10 @@ def verify(self, op: Operation):
         irdl_op_verify_arg_list(op, self, VarIRConstruct.RESULT, constraint_vars)
 
         # Verify regions.
-        irdl_op_verify_arg_list(op, self, VarIRConstruct.REGION, constraint_vars)
+        irdl_op_verify_regions(op, self)
 
-        # Verify successors.
-        irdl_op_verify_arg_list(op, self, VarIRConstruct.SUCCESSOR, constraint_vars)
+        # verify successors
```
math-fehr, line 1639: Why is this now in lowercase without a comma, where everything else is?

math-fehr (approved):
Really cool!
My only question is why some attributes do not need to be verified in ranges?

xdsl/irdl/irdl.py
PapyChacal: Indeed, probably by a more complete object encapsulating whatever state we want to add? At first just both variable types

xdsl/irdl/irdl.py
PapyChacal: I'm honestly not sure this is needed; will check if it's an issue to revert this for now!

Idea was to use for inference of not yet fully known sequences

xdsl/irdl/irdl.py
PapyChacal: Oh right, we might not want to allow RangeConstraints here; let's keep that kind of controversy for hopefully never

xdsl/irdl/irdl.py
PapyChacal: Because strong opinions

Nah, good catch 😋

PapyChacal:
> Really cool!
> My only question is why some attributes do not need to be verified in ranges?

Guessing this refers to operation attributes and properties? I'm just not sure where it would be useful and thought it would be less controversial to keep them as-is. Would you prefer to switch to ranges everywhere for consistency?

Or were you referring to something else?

math-fehr:
> > Really cool!
> > My only question is why some attributes do not need to be verified in ranges?
> 
> Guessing this refers to operation attributes and properties? I'm just not sure where it would be useful and thought it would be less controversial to keep them as-is. Would you prefer to switch to ranges everywhere for consistency?
> 
> Or were you referring to something else?

Sorry, I wrote garbage!
What I meant was why in the `verify` method you have `list[Attribute | None]`. But you answered that on the comment ;)

xdsl/irdl/irdl.py
superlopuh: *Constrain

xdsl/irdl/irdl.py
PapyChacal: Done, forgetting about None for now, it was only needed for Pyright at this point.
I suspect with more complex RangeConstraint, it might becone useful, but let's think about it then

xdsl/irdl/irdl.py
PapyChacal: get_op_constructs return `Sequence[SSAValue] | list[OpResult] | list[Region] | list[Successor]`, depending on the enum argument.
I could use overload for literal arguments, but not union of them, so used a cast.

xdsl/irdl/declarative_assembly_format.py
PapyChacal: `Argument of type "list[Attribute | None] | Attribute | list[Attribute]" cannot be assigned to parameter "value" of type "Attribute | list[Attribute | None] | None"` ?

superlopuh (approved):
LGTM!

PR #2676, opened 2024-05-31
backend: (csl) Maths and memory operatoin printing

dk949:
* Types:
    * `PtrType`
    * `FunctionType`
* Arith operations:
    * `IndexCastOp`
    * `SIToFPOp`
    * `FPToSIOp`
    * `ExtFOp`
    * `TruncFOp`
    * `TruncIOp`
    * `ExtSIOp`
    * `ExtUIOp`
    * `Muli` 
    * `Mulf`
    * `Addi` 
    * `Addf`
* Memref operations:
    * `AddressOfOp`
    * `Store`
    * `Load`


tests/filecheck/backend/csl/print_csl.mlir
```diff
@@ -39,6 +39,21 @@
     csl.return
   }
 
+  csl.func @casts() {
+    %constI32 = arith.constant 0 : i32
+    %constU16 = arith.constant 0 : ui16
+    %constF32 = arith.constant 0 : f32
+    %castIndex = "arith.index_cast"(%constF32) : (f32) -> index
```
AntonLydike, line 46: `index_cast` can only cast between index and integer data types I'm afraid

AntonLydike: Not sure why this isn't checked by xDSL

xdsl/backend/csl/print_csl.py
```diff
@@ -333,6 +382,24 @@ def print_block(self, body: Block):
...
+                    arr_name = self._get_variable_name_for(arr)
+                    idx_args = ", ".join(map(self._get_variable_name_for, idxs))
+                    # Use the array access syntax instead of copying the value out
+                    self.variables[res] = f"({arr_name}[{idx_args}])"
+                case csl.AddressOfOp(value=val, res=res):
+                    val_name = self._get_variable_name_for(val)
+                    ty = res.type
+                    assert isinstance(
+                        ty, csl.PtrType
+                    ), f"Result of {csl.AddressOfOp.name} has to be a pointer"
```
AntonLydike, line 400: This is checked by the `verify` (although I guess this is needed to make pyright happy?). Not sure we need the assert here, I think we could replace it by a `cast`, as it is never going to be wrong?

I guess this is a trade-off between more readable printing code vs. better error messages when printing invalid csl code...

I guess imo we don't need this assertion here.

AntonLydike (approved):
LGTM overall, just a few minor nits

xdsl/backend/csl/print_csl.py
dk949: Yes, this assert is just for the sake of the type checker, it does get caught by `verify`. A `cast` should be safe here.

tests/filecheck/backend/csl/print_csl.mlir
dk949: I didn't realise that. Fixed

AntonLydike: PR open #2682 btw :)

PR #2682, opened 2024-06-03
dialects: (arith) Add verifier to 'arith.index_cast' op and fix returns in inits

AntonLydike:
Add a verifier to `arith.index_cast` op mirroring MLIR.

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -967,6 +967,8 @@ class IndexCastOp(IRDLOperation):
 
     traits = frozenset([Pure()])
 
+    assembly_format = "$input attr-dict `:` type($input) `to` type($result)"
+
     def __init__(self, input_arg: SSAValue | Operation, target_type: Attribute):
         super().__init__(operands=[input_arg], result_types=[target_type])
 
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -961,14 +961,18 @@ class Minnumf(FloatingPointLikeBinaryOp):
 class IndexCastOp(IRDLOperation):
     name = "arith.index_cast"
 
-    input: Operand = operand_def()
+    input: Operand = operand_def(IntegerType | IndexType)
 
-    result: OpResult = result_def()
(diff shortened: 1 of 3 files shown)


xdsl/dialects/arith.py
```diff
@@ -961,15 +961,25 @@ class Minnumf(FloatingPointLikeBinaryOp):
...
     traits = frozenset([Pure()])
 
     def __init__(self, input_arg: SSAValue | Operation, target_type: Attribute):
         return super().__init__(operands=[input_arg], result_types=[target_type])
 
+    def verify_(self) -> None:
+        if (
+            isinstance(self.input.type, IndexType)
+            + isinstance(self.result.type, IndexType)
+            != 1
```
superlopuh, line 977: Meganit: I'd say that != each other is a bit kore idiomatic

superlopuh (approved):

xdsl/dialects/arith.py
AntonLydike: Hmm, how would we do that such that it works when `i32` and `i16` are used?

AntonLydike: I guess it would then be `type(self.input.type) != type(self.result.type)`... Hmmm. Yeah, I can see that!

superlopuh: Oh I meant the isinstances, since you're checking that exactly one of the isinstance checks returns True. I thought one of the two had to be `index`, does it handle `i32` to `i16` casting, also?

AntonLydike: Oh sorry, the above comment was misleading. The only allowed mode is exactly one of input or output being index.

I meant that checking for `input.type != output.type` would not work, as one would need to check the type of the ssa value type^^

But it should be possible (and shorter)

AntonLydike:
```
xdsl/dialects/arith.py:974:12: E721 Do not compare types, use `isinstance()`
```


But `if isinstance(self.input.type, type(self.result.type)):` is just less readable imo

PR #2683, opened 2024-06-03
Dialects: (csl) Simplify DSD types in IR

n-io:
Simplify DSD types in IR by not carrying around element type information. Since this is only needed for the printing of `@increment_dsd_offset()`, we can provide a helper method instead to retrieve this information.


AntonLydike (approved):
Not sure I'm 100% on board here, doesn't this allow us to type-confuse an integer dsd for a float one and create "bad" code?

dk949 (approved):

dk949:
I thin it's important to keep in mind that the type here is only used to denote size, using i32 should be the same as using f32 (no matter the underlying type).

PR #2685, opened 2024-06-03
dialects: Implement MemoryEffect on RISCV op, just reuse main CSE.

PapyChacal:
Remove the `riscv-cse` pass, instead implement AllocatedMemoryEffect on some RISCV ops and let generic CSE do the job.

This makes CSE more powerful, with dead-code elimination and common subexpression elimination across regions.


tests/filecheck/transforms/riscv_cse.mlir
```diff
@@ -10,16 +10,22 @@
...
     %d8 = riscv.li 8 : () -> !riscv.reg<>
     %e8 = riscv.li 8 : () -> !riscv.reg<>
+
+    "test.op"(%d8, %e8) : (!riscv.reg<>, !riscv.reg<>) -> ()
 }
 
 %f8 = riscv.li 8 : () -> !riscv.reg<>
 
+"test.op"(%a8, %b8, %c8, %a7, %a7, %b7, %f8) : (!riscv.reg<>, !riscv.reg<>, !riscv.reg<>, !riscv.reg<>, !riscv.reg<>, !riscv.reg<>, !riscv.reg<>) -> ()
+
```
PapyChacal, line 20: Added two `test.op`s to use value and prevent CSE to rightfully CSE away the whole module

superlopuh (approved):
Fantastico

AntonLydike (commented):
Why only add the trait to two of the operations? Shouldn't this work for all RISC-V operations? And don't we also want to model memory effects to all load and stores?

PapyChacal:
> Why only add the trait to two of the operations? Shouldn't this work for all RISC-V operations?

Left as an exercise for the reader I guess; I see this is enough to do what's tested, but I don't know every corner of this dialect stack. Feel free to add the trait to any missing operation you see!

> And don't we also want to model memory effects to all load and stores?

What do you mean by that? If you mean modelling read/write effects in more detail, yes, this is left as further work. I can help with if it would be useful somewhere specifically?

superlopuh:
I would say that this is good to merge as-is, in an incomplete state, we can fill in the rest of the ops with the effect later.

PapyChacal:
> I would say that this is good to merge as-is, in an incomplete state, we can fill in the rest of the ops with the effect later.

Alright, will update and merge soonish :slightly_smiling_face:

tests/filecheck/transforms/riscv_cse.mlir
```diff
@@ -75,7 +80,6 @@ riscv_scf.for %13 : !riscv.reg = %11 to %8 step %12 {
 // CHECK-NEXT:        %{{.*}} = riscv_scf.for %{{.*}} : !riscv.reg = %{{.*}} to %{{.*}} step %{{.*}} iter_args(%{{.*}} = %{{.*}}) -> (!riscv.freg) {
 // CHECK-NEXT:          %{{.*}} = riscv_snitch.read from %{{.*}} : !riscv.freg
 // CHECK-NEXT:          %{{.*}} = riscv_snitch.read from %{{.*}} : !riscv.freg
-// CHECK-NEXT:          %{{.*}} = riscv.fmul.d %{{.*}}, %{{.*}} : (!riscv.freg, !riscv.freg) -> !riscv.freg
```
superlopuh, line 78: oh actually this is weird

PapyChacal:
> Just so no-one merges accidentally, this is currently not working.

We called and this was about a misread of the test-case: An unused operation was correctly erased, the example simply intended to have it *used* instead.

superlopuh (approved):

PR #2686, opened 2024-06-03
Dialects: (csl) Adding builtin ops for DSDs

n-io:
DSD builtin operations typically come with a variety of overloads. This PR proposes a base class handling ops and verification. The subclasses correspond to CSL builtin ops. Their implementation only needs to specify op name and the valid function signatures. The goal is to keep everything really simple.

### DIFF
--- a/xdsl/dialects/csl.py
+++ b/xdsl/dialects/csl.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 from abc import ABC
 from collections.abc import Sequence
 from dataclasses import dataclass
-from typing import Annotated, TypeAlias
+from typing import Annotated, TypeAlias, TypeVar
 
 from xdsl.dialects.builtin import (
     AnyFloatAttr,
@@ -46,6 +46,7 @@ from xdsl.ir import (
     TypeAttribute,
 )
 from xdsl.irdl import (
+    AttrConstraint,
     IRDLOperation,
     ParameterDef,
     ParametrizedAttribute,
@@ -285,6 +286,14 @@ DsdElementType: TypeAlias = (
     | Annotated[IntegerType, IntegerType(16, Signedness.UNSIGNED)]
     | Annotated[IntegerType, IntegerType(32, Signedness.SIGNED)]
     | Annotated[IntegerType, IntegerType(32, Signedness.UNSIGNED)]
+)
+
+
+f16_pointer = PtrType(
+    [Float16Type(), PtrKindAttr(PtrKind.SINGLE), PtrConstAttr(PtrConst.VAR)]
+)
+f32_pointer = PtrType(
+    [Float32Type(), PtrKindAttr(PtrKind.SINGLE), PtrConstAttr(PtrConst.VAR)]
 )
 
 
@@ -883,6 +892,39 @@ class SetDsdStrideOp(IRDLOperation):
             raise VerifyException(f"{self.name} can only operate on mem1d_dsd type")
 
 
+class _BuiltinDsdOpBase(IRDLOperation, ABC):
+    ops = var_operand_def()
+    signature: list[tuple[AttrConstraint | Attribute | type[Attribute] | TypeVar, ...]]
+
+    def verify_(self) -> None:
+        def typcheck(
+            op_typ: Attribute,
+            sig_typ: AttrConstraint | Attribute | type[Attribute] | TypeVar,
+        ) -> bool:
+            if isinstance(sig_typ, type):
+                return isinstance(op_typ, sig_typ)
+            else:
+                return op_typ == sig_typ
+
+        for sig in self.signature:
+            if len(self.ops) == len(sig):
+                if all(typcheck(op.type, sig_t) for (op, sig_t) in zip(self.ops, sig)):
+                    return
+        raise VerifyException("Cannot find matching type signature")
+
+
+@irdl_op_definition
+class FaddsOp(_BuiltinDsdOpBase):
+    name = "csl.fadds"
+
+    signature = [
+        (DsdType, DsdType, DsdType),
+        (DsdType, Float16Type, DsdType),
+        (DsdType, DsdType, Float16Type),
+        (f16_pointer, Float16Type, DsdType),
+    ]
+
+
 @irdl_op_definition
 class SymbolExportOp(IRDLOperation):
     """
@@ -1064,6 +1106,7 @@ CSL = Dialect(
         IncrementDsdOffsetOp,
         SetDsdLengthOp,
         SetDsdStrideOp,
+        FaddsOp,
         AddressOfOp,
         SymbolExportOp,
         RpcOp,
--- a/tests/filecheck/dialects/csl/ops.mlir
+++ b/tests/filecheck/dialects/csl/ops.mlir
@@ -87,6 +87,11 @@ csl.func @initialize() {
     %fabin_dsd = "csl.get_fab_dsd"(%scalar) : (i32) -> !csl<dsd fabin_dsd>
     %fabout_dsd = "csl.get_fab_dsd"(%scalar) : (i32) -> !csl<dsd fabout_dsd>
 
+    %f16_ptr, %f16_val, %f32_ptr = "test.op"() : () -> (!csl.ptr<f16, #csl<ptr_kind single>, #csl<ptr_const var>>, f16, !csl.ptr<f32, #csl<ptr_kind single>, #csl<ptr_const var>>)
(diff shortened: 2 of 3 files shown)


xdsl/dialects/csl.py
```diff
@@ -883,6 +892,50 @@ def verify_(self) -> None:
...
+]
+
+
+class _BuiltinDsdOpBase(IRDLOperation, ABC):
+    ops = var_operand_def()
+
+    @staticmethod
+    @abstractmethod
+    def get_signatures() -> FunctionSignatures:
+        raise NotImplementedError()
```
dk949, line 906: I think you should be able to just do `def get_signatures() -> FunctionSignatures: ...`

xdsl/dialects/csl.py
```diff
@@ -883,6 +892,50 @@ def verify_(self) -> None:
...
+
+    def verify_(self) -> None:
+        def typcheck(
+            op_typ: Attribute,
+            sig_typ: AttrConstraint | Attribute | type[Attribute] | TypeVar,
```
dk949, line 911: I am not sure we need `AttrConstraint` here, since we don't really use them anywhere else in the dialect (not directly any way). If you do think we should keep it, maybe consider using `attr_constr_coercion` from `irdl`, to convert everything to an  `AttrConstraint` then using the `verify` method.

xdsl/dialects/csl.py
```diff
@@ -883,6 +892,50 @@ def verify_(self) -> None:
             raise VerifyException(f"{self.name} can only operate on mem1d_dsd type")
 
 
+FunctionSignatures = list[
+    tuple[AttrConstraint | Attribute | type[Attribute] | TypeVar, ...]
```
dk949, line 896: Ditto for use of `AttrConstraint`.

dk949 (commented):
I am still not sure about implementing a class for every version of each
builtin. I agree that this is the simpler solution to initially implement, I
feel like it will be harder to maintain and a bit more difficult to transform. 

I think it should be possible to roll any number of builtins into the same
class. The two ideas I had yesterday were to:

1. Provide the suffixes as part of the signature and a public function for the
   printer to call, performing a match and returning the associated suffix. This
   method would be quite easy to implement, but it adds printer specific
   functionality to the dialect, which is undesirable.
2. Derive the suffix from the signature(s). There appears to be some sort of a
   system for naming these builtins, however, some functions are
   indistinguishable by signature alone, as they use untyped DSDs.
3. (This is the 3rd of the two ideas). Disregard concerns in 1st paragraph and
   implement each builtin as its own class 🤷

xdsl/dialects/csl.py
n-io: It's the type signature copied over from `operand_def`

n-io:
> I am still not sure about implementing a class for every version of each builtin. I agree that this is the simpler solution to initially implement, I feel like it will be harder to maintain and a bit more difficult to transform.

I agree, but regrettably haven't really gotten to that part yet. Happy to include it in this PR if you like.

> I think it should be possible to roll any number of builtins into the same class. The two ideas I had yesterday were to:
> 
> 1. Provide the suffixes as part of the signature and a public function for the
>    printer to call, performing a match and returning the associated suffix. This
>    method would be quite easy to implement, but it adds printer specific
>    functionality to the dialect, which is undesirable.

This may not really work with the current design, since the signature is based off of the var args, and we can't pass the precision as an operand. However, what we could do is pass the precision as an operand. Then in the `get_signatures` method we could check the precision property value and return the correct set of signatures. Do we have to provide a method for the name as well or do we figure it out in the printer?

> 2. Derive the suffix from the signature(s). There appears to be some sort of a
>    system for naming these builtins, however, some functions are
>    indistinguishable by signature alone, as they use untyped DSDs.

This would be really really good - however, I do see practical problems. For instance, if we get `(DsdType, DsdType, DsdType)` as the function signature, this would match both `faddh`, `faddhs`, and `fadds`, since we don't carry around element type information on the DSDs.

> 3. (This is the 3rd of the two ideas). Disregard concerns in 1st paragraph and
>    implement each builtin as its own class 🤷

Yes, I suppose it's between this and option 1.

AntonLydike:
So I guess to summarise previous discussion:

- having one generic `csl.dsd_builtin(%a, %b, %c) <{name = "faddh", }> : (...) -> ...` would be painful and complex to verify?
- having one class per "name" and precision might be excessive 
- having one class per op "name" with precision as a property seems to be the sweet spot?

I would favour having one op per function name emitted in the resulting CSL, except if there's a good reason not to.

xdsl/dialects/csl.py
```diff
@@ -883,6 +925,527 @@ def verify_(self) -> None:
...
+    name = "csl.fscale"
+
+    def get_signatures(self) -> FunctionSignatures:
+        match self.precision.data:
+            case BuiltinOpPrecisionKind.half:
+                return [
+                    (f16_pointer, Float16Type, i16_value),
+                ]
+            case BuiltinOpPrecisionKind.mixed:
+                return []
```
AntonLydike, line 1293: This is a way of saying "I support no mixed-precision modes", right?

AntonLydike (commented):
image

:face_holding_back_tears: Can we split this into two or three PRs? This is quite a lot to review all at once

xdsl/dialects/csl.py
n-io: yes

n-io: It will cause it to fail the verification performed by the superclass

xdsl/dialects/csl.py
```diff
@@ -883,6 +925,527 @@ def verify_(self) -> None:
...
+@irdl_op_definition
+class Fh2xp16Op(_BuiltinDsdOp):
+    """
+    Implements @fh2xp16
+    """
+
+    name = "csl.fh2xp16"
+
+    def get_signatures(self) -> FunctionSignatures:
+        return [(DsdType, DsdType), (DsdType, Float16Type), (i16_pointer, Float16Type)]
```
AntonLydike, line 1098: Can you add a comma at the end of this array, so that `black` splits it across the lines like it does with others?

dk949:
@AntonLydike

> I would favour having one op per function name emitted in the resulting CSL, except if there's a good reason not to.

This is one reason...

I wasn't proposing one single op to rule them all. But maybe all additions together, all `mac`s together etc..

dk949:
@n-io i really appreciate you adding all the possible DSD operations, but we will realistically never use 90% of these...

I don't think I've ever used a clz in regular code, let alone CSL 😅

n-io:
> @n-io i really appreciate you adding all the possible DSD operations, but we will realistically never use 90% of these...
> 
> I don't think I've ever used a clz in regular code, let alone CSL 😅

Easy to do, but dito for control tasks 😂

n-io:
If we want one op per function name emitted, that's done quickly and I'm happy to do it. It may actually not be a bad approach going forward, I suppose. We had a similar discussion regarding tasks, where in the end we decided to combine all types of tasks into one parameterised class. I'm happy either way, to be honest.

One op to rule them all might actually make op rewriting rather complicated, also making it generally a lot harder to easily access common functionalities that xdsl provides us with.

AntonLydike (approved):
Yeah, lgtm

xdsl/dialects/csl.py
```diff
@@ -883,6 +908,437 @@ def verify_(self) -> None:
...
+            if isinstance(sig_typ, type):
+                return isinstance(op_typ, sig_typ)
+            else:
+                return op_typ == sig_typ
+
+        for sig in self.SIGNATURES:
+            if len(self.ops) == len(sig):
+                if all(typcheck(op.type, sig_t) for (op, sig_t) in zip(self.ops, sig)):
+                    return
+        raise VerifyException("Cannot find matching type signature")
```
AntonLydike, line 935: Can you add a unit test for this (just in python, not filecheck). I think this would increase confidence here

dk949 (commented):
It's 👍 from me as well, except the `AttrConstraint` thing in the type checking function

xdsl/dialects/csl.py
n-io: Added a python test as suggested.

n-io:
> It's 👍 from me as well, except the `AttrConstraint` thing in the type checking function

It was the full union definition used by `operand_def` which I thought might be what we wanted. But since we're not actually exposing anything here, you're completely right that we don't technically need it. Removed as suggested.

PR #2687, opened 2024-06-03
core: (irdl) Allow ClassVar with all uppercase name

AntonLydike:
This change makes IRDL accept non-irdl fields on `@irdl_op_definition` decorated classes, if they are all uppercase and annotated as `ClassVar`.

### DIFF
--- a/xdsl/irdl/irdl.py
+++ b/xdsl/irdl/irdl.py
@@ -1487,7 +1487,13 @@ class OpDef:
                 if (
                     field_name in annotations
                     and field_name.isupper()
-                    and get_origin(annotations[field_name]) is ClassVar
+                    and (
+                        get_origin(annotations[field_name]) is ClassVar
+                        or (
+                            isinstance(annotations[field_name], str)
+                            and annotations[field_name].startswith("ClassVar")
+                        )
+                    )
                 ):
                     continue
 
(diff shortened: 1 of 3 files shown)


xdsl/irdl/irdl.py
```diff
@@ -1483,6 +1483,12 @@ def wrong_field_exception(field_name: str) -> PyRDLOpDefinitionError:
                 if field_name in field_names:
                     # already registered value for field name
                     continue
+                if (
+                    field_name in annotations
+                    and field_name.isupper()
+                    and annotations[field_name].startswith("ClassVar[")
```
AntonLydike, line 1489: This does not feel the cleanest, but I'm not sure what the proper way to do this would be...

math-fehr (approved):
That makes sense to me!

n-io:
There is something really subtle going on, in that the test will fail if we add `from __future__ import annotations` at the beginning of the file.

I've created a separate test file to showcase this scenario, and it currently breaks (as expected).

The reason is somewhat subtle:
`cls.__dict__` without the import: `{... '__annotations__': {'VAR': ClassVar[str]} ... }`
`cls.__dict__` with the import: `{... '__annotations__': {'VAR': 'ClassVar[str]'} ... }`

It appears that the import will cause `ClassVar[str]` to be a string rather than a type in the annotations.

math-fehr:
Yeah, because of this feature, some python types will become `str`, and can only be converted to actual types with an annoying function that's quite costly.
Personally, I would be fine checking that the string starts with `ClassVar`. I feel this is quite sensible, will not hurt performance, and is a clear improvement over what we currently have.

PR #2690, opened 2024-06-05
core: Allowing overload of ClassVar from superclass

n-io:
This is to enhance PR [#2687](https://github.com/xdslproject/xdsl/pull/2687), which allowed uppercase `ClassVar` fields on ops.

This PR adds support for the case that an uppercase `ClassVar` is defined (without a value) in a superclass and assigned a value in a subclass. This is confirmed by performing the same check as in the other PR.

### DIFF
--- a/xdsl/irdl/irdl.py
+++ b/xdsl/irdl/irdl.py
@@ -1472,6 +1472,14 @@ class OpDef:
             annotations = parent_cls.__annotations__
 
             for field_name in annotations:
+                if field_name.isupper() and (
+                    get_origin(annotations[field_name]) is ClassVar
+                    or (
+                        isinstance(annotations[field_name], str)
+                        and annotations[field_name].startswith("ClassVar")
+                    )
+                ):
+                    continue
                 if field_name not in clsdict:
                     raise wrong_field_exception(field_name)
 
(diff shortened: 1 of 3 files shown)


xdsl/irdl/irdl.py
```diff
@@ -1472,6 +1472,14 @@ def wrong_field_exception(field_name: str) -> PyRDLOpDefinitionError:
...
 
             for field_name in annotations:
+                if field_name.isupper() and (
+                    get_origin(annotations[field_name]) is ClassVar
+                    or (
+                        isinstance(annotations[field_name], str)
+                        and annotations[field_name].startswith("ClassVar")
+                    )
+                ):
+                    continue
```
AntonLydike, line 1482: Seeing as we have this in two places now, could we factor this out to make it easier to see that the two checks are identical? And maybe also make this code denser again?

AntonLydike (approved):
Just a minor nit on factoring out the check. Otherwise LGTM, thanks for the core contribution!

xdsl/irdl/irdl.py
n-io: That sounds very reasonable, have factored out the check as suggested.

math-fehr (approved):
Very nice!

PR #2697, opened 2024-06-07
backend: (csl) conditional statements

n-io:
Adding support for `scf.if` conditionals to CSL printer. Also fixing print for boolean type.

### DIFF
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -331,14 +331,19 @@ class CslPrintContext:
                 case csl.ReturnOp(ret_val=val) if val is not None:
                     self.print(f"return {self._get_variable_name_for(val)};")
                 case scf.If(
-                    cond=cond, true_region=true_region, false_region=false_region
+                    cond=cond,
+                    output=outputs,
+                    true_region=true_region,
+                    false_region=false_region,
                 ):
+                    for o in outputs:
+                        self.print(f"{self._var_use(o, 'var')};")
                     with self.descend(
                         f"if ({self._get_variable_name_for(cond)})"
                     ) as inner:
                         inner.print_block(true_region.block)
                     if false_region:
-                        if not (
+                        if len(outputs) > 0 or not (
                             len(false_region.block.ops) == 1
                             and isinstance(false_region.block.first_op, scf.Yield)
                         ):
@@ -368,7 +373,13 @@ class CslPrintContext:
                     with self.descend(loop_definition) as inner:
                         inner.print_block(bdy.block)
                 case scf.Yield():
-                    pass
+                    if (
+                        isinstance(op.parent, Block)
+                        and isinstance(op.parent.parent, Region)
+                        and isinstance(ifop := op.parent.parent.parent, scf.If)
+                    ):
+                        for r, y in zip(ifop.results, op.operands):
+                            self.print(f"{self._var_use(r)} = {self._var_use(y)};")
                 case (
                     arith.IndexCastOp(input=inp, result=res)
                     | arith.SIToFPOp(input=inp, result=res)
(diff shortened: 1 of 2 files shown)


xdsl/backend/csl/print_csl.py
```diff
@@ -352,7 +373,13 @@ def print_block(self, body: Block):
                     with self.descend(loop_definition) as inner:
                         inner.print_block(bdy.block)
                 case scf.Yield():
-                    pass
+                    if (
+                        isinstance(op.parent, Block)
+                        and isinstance(op.parent.parent, Region)
+                        and isinstance(ifop := op.parent.parent.parent, scf.If)
+                    ):
```
AntonLydike, line 380: Can be replaced with `isinstance(op.parent_op(), scf.If)`

AntonLydike (approved):
Just a minor nitpick, but otherwise this looks good to me! :star:

xdsl/backend/csl/print_csl.py
```diff
@@ -328,6 +330,25 @@ def print_block(self, body: Block):
...
+                    for o in outputs:
+                        self.print(f"{self._var_use(o, 'var')};")
+                    with self.descend(
+                        f"if ({self._get_variable_name_for(cond)})"
+                    ) as inner:
+                        inner.print_block(true_region.block)
+                    if false_region:
+                        if len(outputs) > 0 or not (
+                            len(false_region.block.ops) == 1
+                            and isinstance(false_region.block.first_op, scf.Yield)
```
dk949, line 348: Do we need this conditional? worst case scenario it just prints an empty `else {}` block.

dk949 (approved):

xdsl/backend/csl/print_csl.py
n-io: @dk949 does it slightly differently in `scf.For` loops - basically, instead of passing a value to the `scf.Yield` op, it is assigned to the outside var directly. I'm matching this for now. Happy to hear your thoughts on this?

xdsl/backend/csl/print_csl.py
n-io: @dk949 does printing of `scf.Yield` slightly differently for `scf.For` loops - basically, instead of passing a value to the `scf.Yield` op, it is assigned to the outside var directly. I'm matching this for now. Happy to hear your thoughts on this?

xdsl/backend/csl/print_csl.py
n-io: We must avoid at all costs printing an empty else. Yikes. 😉

xdsl/backend/csl/print_csl.py
AntonLydike: I don't have any opinion on it, it should result in the same code anyway, right?

xdsl/backend/csl/print_csl.py
```diff
@@ -338,6 +338,13 @@ def print_block(self, body: Block):
                 ):
                     for o in outputs:
                         self.print(f"{self._var_use(o, 'var')};")
+                    # Search for all yield operations and match yield argument names to for argument names
+                    for blk in [true_region, false_region]:
+                        for op in blk.block.ops:
+                            if not isinstance(op, scf.Yield):
+                                continue
+                            for yield_arg, o in zip(op.arguments, outputs):
+                                self.variables[yield_arg] = self.variables[o]
```
AntonLydike, line 347: You should be able to grab the yield op by using `true_region.block.last_op` instead, this way we don't need to traverse the list at all.

AntonLydike (commented):
I'm generally in favour of simpler implementations rather than more complex ones, so I would prefer the approach with less code.

xdsl/backend/csl/print_csl.py
n-io: I like that very much, implementing it for both `scf.If` and `scf.For`

PR #2699, opened 2024-06-09
core: allow to define a RangeConstraint on a region's entry arguments.

PapyChacal:


xdsl/irdl/irdl.py
```diff
@@ -989,11 +989,13 @@ class OptResultDef(VarResultDef, OptionalDef):
 
 
 @dataclass(init=True)
-class RegionDef(Region):
```
math-fehr, line 992: Nice catch ;)

xdsl/irdl/irdl.py
```diff
@@ -1581,6 +1604,20 @@ def get_constraint(
...
+                        | AttrConstraint
+                        | Attribute
+                        | type[Attribute]
+                        | TypeVar
+                    ),
+                ) -> RangeConstraint:
+                    if isinstance(pyrdl_constr, RangeConstraint):
+                        return pyrdl_constr
+                    return RangeOf(get_constraint(pyrdl_constr))
+
```
math-fehr, line 1620: I'm not sure I would define a new function if there is only a single use? Unless you expect it to have more uses?

math-fehr (approved):
Really nice!

xdsl/irdl/irdl.py
PapyChacal: :dart:

xdsl/irdl/irdl.py
PapyChacal: Yep, next PR makes it three uses :wink: 
Happy to move it to the next PR if it comes to that

PapyChacal: Assuming this is okay as it is an approval's comment, so merging as is; yell at me if I was wrong ^^

tests/irdl/test_operation_definition.py
```diff
@@ -627,3 +631,41 @@ def test_multiple_inheritance_op():
...
+    op = EntryArgsOp.create(regions=[Region(Block(arg_types=[i32, i32]))])
+    op.verify()
+
+    op = EntryArgsOp.create(regions=[Region(Block(arg_types=[i64]))])
+    with pytest.raises(
+        DiagnosticException,
+        match="""\
+Operation does not verify: region #0 entry arguments do not verify:
+Expected attribute i32 but got i64""",
+    ):
```
georgebisbas, line 661: SHouldnt this be less hard coded?

tests/irdl/test_operation_definition.py
```diff
@@ -627,3 +631,41 @@ def test_multiple_inheritance_op():
...
+        op.verify()
+
+    op = EntryArgsOp.create(regions=[Region(Block(arg_types=[i64, i32]))])
+    with pytest.raises(
+        DiagnosticException,
+        match="""\
+Operation does not verify: region #0 entry arguments do not verify:
+Expected attribute i32 but got i64""",
```
georgebisbas, line 669: same

tests/irdl/test_operation_definition.py
PapyChacal: What do you mean?

PapyChacal: I like testing explicitely the expressed error, because an easy regression in that area is, if you just test that verification fails, for some reasons verification might fail for an unrelated reason. Then, nothing is checking the exact tested verification feature anymore!

PR #2700, opened 2024-06-09
core: Implement and use ConstraintContext

PapyChacal:
Just define ConstraintContext to encapsulate the constraint variables used in verification and inference.
Also, define a default value for this argument in verification, to verify with an empty context.

### DIFF
--- a/xdsl/irdl/irdl.py
+++ b/xdsl/irdl/irdl.py
@@ -105,7 +105,7 @@ class AttrConstraint(ABC):
     def verify(
         self,
         attr: Attribute,
-        constraint_context: ConstraintContext | None = None,
+        constraint_context: ConstraintContext,
     ) -> None:
         """
         Check if the attribute satisfies the constraint,
@@ -128,7 +128,7 @@ class AttrConstraint(ABC):
         # By default, we cannot infer anything.
         return False
 
-    def infer(self, constraint_context: ConstraintContext | None = None) -> Attribute:
+    def infer(self, constraint_context: ConstraintContext) -> Attribute:
         """
         Infer the attribute given the constraint variables that are already set.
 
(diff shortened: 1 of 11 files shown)


xdsl/irdl/irdl.py
```diff
@@ -81,12 +81,32 @@ class IRDLAnnotations(Enum):
...
 
 
+@dataclass
+class ConstraintContext:
+    """
+    Context for constraint variables.
+    """
+
+    variables: dict[str, Attribute] = field(default_factory=dict)
+    """The constraint variables."""
```
math-fehr, line 91:
```suggestion
    """The assignment of constraint variables."""
```

xdsl/irdl/irdl.py
```diff
@@ -81,12 +81,32 @@ class IRDLAnnotations(Enum):
 #
 
 
+@dataclass
+class ConstraintContext:
+    """
+    Context for constraint variables.
```
math-fehr, line 87:
```suggestion
    Contains the assignment of constraint variables.
```
Otherwise the comment was not really helping much I think ;)

math-fehr (approved):
Nice!
One thing I'm kind of disagreeing on though, is allowing to call `verify` without the `ConstraintContext`.
I would say it is the role of the caller to define a `ConstraintContext`, not the constraint. What do you think?

PapyChacal:
> Nice! One thing I'm kind of disagreeing on though, is allowing to call `verify` without the `ConstraintContext`. I would say it is the role of the caller to define a `ConstraintContext`, not the constraint. What do you think?

Sure, I can just revert that, here are my cents:
- The caller can define one if they want to, but doesn't have to worry about it everytime
- Basically all the erased `, {}`s would become `, ConstraintContext()`s, including an extra import. I just think it brings much noise to undesired places, e.g. the docs introducing constraints as a whole.

I want to insist on **I can just revert** :yum: I see this as bikeshedding, so please first just confirm you want it reverted, and let's continue the discussion separately from the feature :slightly_smiling_face: This is one small piece amongst quite a few small pieces, that will yield very nice expressivity to IRDL, which is what I'm after rather than the API

PapyChacal:
As discussed in a quick call, I'll revert the default argument thing!

PR #2702, opened 2024-06-10
dialects: Define memref.atomic rmw

PapyChacal:
Also add minor helpers to builtin: `I1`,`I32`,`I64`, to use in type expressions.


georgebisbas (approved):
Thanks!

PR #2705, opened 2024-06-11
core: [NFC] Split irdl.py

PapyChacal:


AntonLydike (approved):

georgebisbas (approved):

superlopuh (approved):
Finally :)

xdsl/irdl/common.py
```diff
@@ -0,0 +1,18 @@
+from enum import Enum
```
math-fehr, line 1: Should we rename this to `error.py` for now?

math-fehr (approved):
Doing the lord's work, thanks a lot 🙏

xdsl/irdl/common.py
PapyChacal: Why not!

superlopuh:
I moved around the branches to merge the split before the RangeConstraint PR, to unblock other changes that would have touched irdl.py

PR #2706, opened 2024-06-11
dialects: Add NoMemoryEffects to memref and snitch_runtime dialect ops

AntonLydike:
This PR adds `NoMemoryEffects` trait to ops from the memref and snitch_runtime dialect that should have them.

### DIFF
--- a/xdsl/dialects/memref.py
+++ b/xdsl/dialects/memref.py
@@ -58,6 +58,7 @@ from xdsl.traits import (
     HasParent,
     IsTerminator,
     SymbolOpInterface,
+    NoMemoryEffect,
 )
 from xdsl.utils.bitwise_casts import is_power_of_two
 from xdsl.utils.deprecation import deprecated_constructor
@@ -379,6 +380,10 @@ class GetGlobal(IRDLOperation):
     memref: OpResult = result_def(MemRefType[Attribute])
     name_: SymbolRefAttr = prop_def(SymbolRefAttr, prop_name="name")
 
+    traits = frozenset([NoMemoryEffect()])
+
+    assembly_format = "$name `:` type($memref) attr-dict"
+
     def __init__(self, name: str | SymbolRefAttr, return_type: Attribute):
         if isinstance(name, str):
             name = SymbolRefAttr(name)
@@ -388,8 +393,6 @@ class GetGlobal(IRDLOperation):
     @staticmethod
     def get(name: str | SymbolRefAttr, return_type: Attribute) -> GetGlobal:
         return GetGlobal(name, return_type)
-
-    assembly_format = "$name `:` type($memref) attr-dict"
 
     # TODO how to verify the types, as the global might be defined in another
     # compilation unit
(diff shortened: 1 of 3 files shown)


xdsl/dialects/memref.py
```diff
@@ -492,6 +493,8 @@ class AlterShapeOp(IRDLOperation):
         "$src $reassociation attr-dict `:` type($src) `into` type($result)"
     )
 
+    traits = frozenset([NoMemoryEffect()])
+
```
PapyChacal, line 497: Not sure about this one as I see no reference to it upstream, but it sounds sensible given the name :shrug:

xdsl/dialects/snitch_runtime.py
```diff
@@ -45,6 +46,8 @@ class SnitchRuntimeGetInfo(SnitchRuntimeBaseOperation, ABC):
 
     result: OpResult = result_def(i32)
 
+    traits = frozenset([NoMemoryEffect()])
+
```
PapyChacal, line 50: (Same comment for all snitch ops)
They seem strange to me. Given that those operation have no operands, `NoMemoryEffect` implies that the result will always be the same, i.e., you could replace those op by a given `arith.constant`. Is that the case?

PapyChacal (approved):
I have doubts regarding the snitch ones, but giving approval if given my comments it's still considered right, as I don't know the dialect's contract or semantics.

xdsl/dialects/snitch_runtime.py
PapyChacal: Ah, nevermind, I'm overinterpreting the thing; it can be context-dependent, like `gpu.thread_id` for example. So LGTM as long as it sounds right to for this dialect!

xdsl/dialects/memref.py
AntonLydike: These are marked `NoMemoryEffect` also in upstream MLIR: see both [memref.collapse_shape](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefcollapse_shape-memrefcollapseshapeop) and [memref.expand_shape](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefexpand_shape-memrefexpandshapeop)

KGrykiel (approved):

AntonLydike:
@PapyChacal any idea what's wrong in the memref canonicalization test? It seems fishy to me.

PapyChacal:
> @PapyChacal any idea what's wrong in the memref canonicalization test? It seems fishy to me.

What sounds fishy to you? You rightfully marked `memref.subview` as having no side effect, so canonicalization will safely erase any unused subview 🙂

AntonLydike:
> > @PapyChacal any idea what's wrong in the memref canonicalization test? It seems fishy to me.
> 
> What sounds fishy to you? You rightfully marked `memref.subview` as having no side effect, so canonicalization will safely erase any unused subview 🙂

I think your eyes deceive you:

I added a use to the memref, and so now we basically have:

````
func A(x: memref) {
 x = cast(cast(x))
 opaque_op(x)
}
```

And `xdsl-opt -p canonicalize` folds the two casts into one. But the filecheck expects there to be two casts!

AntonLydike:
Ahh, I figured it out!

The memref canonicalization pass did not remove the old cast, it basically transformed the IR to:

```
func a(x: memref) {
x_ = old_cast(x)
new = new_cast(x)
 opaque_op(new)
}
```
and `cse` then righfully garbage collected the first cast!

PR #2709, opened 2024-06-12
core: update parser rules to decide on memref layout or memory space attribute

jorendumoulin:
This PR changes the way the parser decides on whether an attribute of a memref is a layout or a memory space.
My main motivation for this is that I want to use `StringAttr` attributes as memory spaces.
This decision is difficult in xdsl because it lacks the `MemRefLayoutAttrInterface` of mlir.

The current decision flow is as follows:

- If there are two attributes, first is layout and second is memory_space
- Else, if the attribute is an integer, it is a memory space
- Else, if the attribute is a stridedlayout or affinemap layout, it is a layout
- Else, throw an error

My problems with this approach are:

1. We can only use integers as memory spaces. This is causing some headaches over at `snax-mlir`. See [here](https://discourse.llvm.org/t/rfc-memref-memory-shape-as-attribute/2229), for a discussion on this in MLIR. They extended the memory space to be allowed as _any_ attribute.
2. The parsing behaviour is different when using either one or both attributes: for example, my custom layout attribute is completely valid according to xdsl, as long as I also specify a memory space for the memref. Otherwise, it is checked to be either a stridedlayout or affinemap, which it isn't, and then considered to be an invalid attribute.

Therefore, I suggest to extend the memory space to integers and strings, as these are the most widely used attribute types for this, and to remove the subsequent check on memory layout, as this obstructs the use of custom layout attributes

### DIFF
--- a/xdsl/parser/attribute_parser.py
+++ b/xdsl/parser/attribute_parser.py
@@ -517,7 +517,7 @@ class AttrParser(BaseParser):
         memory_or_layout = self.parse_attribute()
 
         # If there is both a memory space and a layout, we know that the
-        # layout is the second one
+        # layout is the first one
         if self.parse_optional_punctuation(",") is not None:
             memory_space = self.parse_attribute()
             return MemRefType(type, shape, memory_or_layout, memory_space)
--- a/xdsl/parser/attribute_parser.py
+++ b/xdsl/parser/attribute_parser.py
@@ -522,24 +522,17 @@ class AttrParser(BaseParser):
             memory_space = self.parse_attribute()
             return MemRefType(type, shape, memory_or_layout, memory_space)
 
-        # Otherwise, there is a single argument, so we check based on the
-        # attribute type. If we don't know, we return an error.
-        # MLIR base itself on the `MemRefLayoutAttrInterface`, which we do not
-        # support.
+        # Otherwise, there is a single argument, so we check based the attribute type.
+        # MLIR bases itself on the `MemRefLayoutAttrInterface`, which we do not support.
+        # Therefore, integers and strings are considered to be memory spaces,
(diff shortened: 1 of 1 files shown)


superlopuh:
What would it take to port the interface? Also, do we have enough tests for the current and proposed behaviour? It would be nice to have a diff in the tests to reflect the change of what we support.

PapyChacal:
Having a quick look at `MemRefLayoutAttrInterface` upstream, its main interface is:
```cpp
/// Get the MemRef layout as an AffineMap, the method must not return NULL
  ::mlir::AffineMap getAffineMap() const;
```
So I guess the intent is only to manipulate the existing StridedLayout as an AffineMap transparently. To put it differently, any such layout should be reducible to an AffineMap, so could be one.

@jorendumoulin, does what you're trying to achieve need something more generic as a layout? If so, we only have to change this interface a bit and allow more generic things, but I would love to use an Interface based approach too (Would be as simple as a base class in xDSL, and I'm happy to help with the technicalities :slightly_smiling_face: )

jorendumoulin:
Alright, seems like integrating the interface is the way to go then :)
It does indeed seem simple enough to integrate. 

@PapyChacal, the custom layout attribute we are using can generalize to an affine map, so this is definitely doable.

I could use some advice on how to properly integrate this into xDSL:

- Are there any existing attribute interfaces?
- Where to best put this?
- Is there something using affinemap layouts that you know of? (I dont't directly find any tests or examples)

superlopuh:
How does the interface relate to parsing? I thought it was, but now it seems like it isn't, maybe this should be two separate PRs?

PapyChacal:
We had a alook with @jorendumoulin and implemented:
- #2718

PR #2710, opened 2024-06-12
backend: (csl) Rework printing logic for printing individual modules

AntonLydike:
This PR addresses a few minor details in csl printing QOL:

- Remove the requirement of there being exactly two csl modules (one layout and one program)
- Change the divider to the standard file-split divider
- Print the file name at the start of each file block

### DIFF
--- a/tests/filecheck/backend/csl/print_csl.mlir
+++ b/tests/filecheck/backend/csl/print_csl.mlir
@@ -315,7 +315,7 @@ csl.func @builtins() {
   csl.return
 }
 
-}) {sym_name = "program"} : () -> ()
+}) {sym_name = "program.csl"} : () -> ()
 
 
 "csl.module"() <{kind=#csl<module_kind layout>}> ({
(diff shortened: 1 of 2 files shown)


tests/filecheck/backend/csl/print_csl.mlir
```diff
@@ -626,12 +628,12 @@ csl.func @builtins() {
 // CHECK-NEXT:   @set_rectangle(x_dim, y_dim);
 // CHECK-NEXT:   const x_coord0 : i32 = 0;
 // CHECK-NEXT:   const y_coord : i32 = 0;
-// CHECK-NEXT:   @set_tile_code(x_coord0, y_coord, "file.csl", )
+// CHECK-NEXT:   @set_tile_code(x_coord0, y_coord, "file.csl", );
```
AntonLydike, line 631: Is this trailing comma in the call wanted / correct?

n-io: Correct yes, wanted not really.

n-io (approved):

dk949 (approved):
This does assume modules come in the right order (program -> layout). I'm happy to make this assumption if we're planning to move the exporting stuff in layout into a pass as discussed.

AntonLydike:
> This does assume modules come in the right order (program -> layout). I'm happy to make this assumption if we're planning to move the exporting stuff in layout into a pass as discussed.

Will fix this, one sec!

AntonLydike:
@dk949 layout ops will now always be printed last. Should be an okay compromise for now?

dk949:
> @dk949 layout ops will now always be printed last. Should be an okay compromise for now?

:+1:  This seems to do the trick (I've never seen `yield from` before :sweat_smile:)

PR #2711, opened 2024-06-12
dialects: (csl) Small incremental fixes to dialect structure (stacked PR)

AntonLydike:
- Make `@get_color` take an operand instead of property.
- Make `ParamOp` take a default value as operand instead of property.


n-io:
Can you please also add type `direction` with values `EAST`, `WEST`, `NORTH`, `SOUTH`? See the comms library for reference.

xdsl/backend/csl/print_csl.py
```diff
@@ -451,7 +451,7 @@ def print_block(self, body: Block):
                     if init is None:
                         init = ""
                     else:
-                        init = f" = { self.attribute_value_to_str(init)}"
+                        init = f" = { self._var_use(init)}"
```
dk949, line 454: `_var_use` potentially initialises the variable, I think `_get_variable_name_for` would be better here.

(strictly speaking, I don't think there is any way this can generate `const foo = const bar`, since `init` is an operand, but it could in theory?)

AntonLydike:
> Can you please also add type `direction` with values `EAST`, `WEST`, `NORTH`, `SOUTH`? See the comms library for reference.

Will be another PR

xdsl/backend/csl/print_csl.py
AntonLydike: I actually have a case where that happens, will fix!

dk949 (approved):

n-io (commented):
Looks good!

PR #2712, opened 2024-06-12
dialects: (csl) set_tile_code fix traits

dk949:
`set_tile_code` does not have to be a direct child of `layout`, e.g. it can be used in a loop. But it does have to occur in a `layout`.

_NOTE_: the parameter to `set_tile_code` is specified as `ComptimeStructType`, not `StructLike`. This is semantically correct (though undocumented), i.e. you can't do this:

```zig
const my_params = @import_module("my_params.csl")
...
@set_tile_code(..., my_params);
```


AntonLydike (approved):
:pray: Literally just ran into this!

PR #2713, opened 2024-06-12
transformations: (csl) added transformation for builtin constructs to csl

dk949:
This transformation converts `func.func` and `func.return` to their `csl` counterparts, wraps everything in a program module and adds a layout module.

### DIFF
--- a/xdsl/transforms/builtin_to_csl.py
+++ b/xdsl/transforms/builtin_to_csl.py
@@ -1,0 +1,168 @@
+from dataclasses import dataclass, field
+
+from xdsl.dialects import builtin, csl, func
+from xdsl.ir import Block, MLContext, Operation, Region
+from xdsl.parser import Parser
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    GreedyRewritePatternApplier,
+    PatternRewriter,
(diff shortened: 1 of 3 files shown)


dk949:
This depends on #2712 to make `set_tile_code` verify

xdsl/transforms/builtin_to_csl.py
```diff
@@ -0,0 +1,168 @@
...
+    }}
+
+  }}
+}}) {{sym_name = "layout"}} :  () -> ()
+""",
+        ).parse_op()
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: csl.CslModuleOp, rewriter: PatternRewriter, /):
+        rewriter.insert_op(self._build_layout_module(), InsertPoint.after(op))
```
AntonLydike, line 132: Not sure this is the right approach, let's chat once we have the baseline testfiles ready! :smile:

dk949: I am not sure either, TBH. Happy to discuss alternatives! This will need fixing anyway after all the other stuff is merged.

dk949:
I don't think this is needed anymore, so I'll close it. It can be reopened if that's not the case

PR #2718, opened 2024-06-13
core: Implement MemrefLayoutAttr base class and use in parser.

PapyChacal:
This is a barebone mimic of the upstream MemrefLayoutAttrInterface, as it does not implement any of the methods; only mark attributes as acceptable by the parser. This solve the parser problem though, and allow the same extensibility on accepted layouts.

The uniplemented methods were not implemented on the existing layout attributes anyway, so I'm in favor of just merging, moving on, and think about implementing them in an interface style later.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1175,7 +1175,7 @@ class OpaqueAttr(ParametrizedAttribute):
         return OpaqueAttr([StringAttr(name), StringAttr(value), type])
 
 
-class MemrefLayoutAttr(Attribute):
+class MemrefLayoutAttr(Attribute, ABC):
     """
     Interface for any attribute acceptable as a memref layout.
     """
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1177,7 +1177,7 @@ class OpaqueAttr(ParametrizedAttribute):
 
 class MemrefLayoutAttr(Attribute):
     """
-    Some explanation here
+    Interface for any attribute acceptable as a memref layout.
     """
 
     name = "memref_layout_att"
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1180,7 +1180,7 @@ class MemrefLayoutAttr(Attribute, ABC):
     Interface for any attribute acceptable as a memref layout.
     """
 
-    name = "memref_layout_att"
+    name = "abstract.memref_layout_att"
 
     pass
 
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1175,8 +1175,18 @@ class OpaqueAttr(ParametrizedAttribute):
         return OpaqueAttr([StringAttr(name), StringAttr(value), type])
 
 
+class MemrefLayoutAttr(Attribute):
+    """
+    Some explanation here
+    """
+
+    name = "memref_layout_att"
+
+    pass
+
+
 @irdl_attr_definition
-class StridedLayoutAttr(ParametrizedAttribute):
+class StridedLayoutAttr(MemrefLayoutAttr, ParametrizedAttribute):
     """
     An attribute representing a strided layout of a shaped type.
     See https://mlir.llvm.org/docs/Dialects/Builtin/#stridedlayoutattr
@@ -1218,7 +1228,7 @@ class StridedLayoutAttr(ParametrizedAttribute):
 
 
 @irdl_attr_definition
-class AffineMapAttr(Data[AffineMap]):
+class AffineMapAttr(MemrefLayoutAttr, Data[AffineMap]):
     """An Attribute containing an AffineMap object."""
 
     name = "affine_map"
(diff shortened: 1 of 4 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1175,8 +1175,18 @@ def from_strings(name: str, value: str, type: Attribute = NoneAttr()) -> OpaqueA
         return OpaqueAttr([StringAttr(name), StringAttr(value), type])
 
 
+class MemrefLayoutAttr(Attribute):
```
superlopuh, line 1178:
```suggestion
class MemrefLayoutAttr(Attribute, abc.ABC):
```

xdsl/parser/attribute_parser.py
```diff
@@ -520,6 +521,8 @@ def _parse_memref_attrs(
         # layout is the second one
         if self.parse_optional_punctuation(",") is not None:
             memory_space = self.parse_attribute()
+            if not isinstance(memory_or_layout, MemrefLayoutAttr):
+                self.raise_error("Expected a MemRef layout attribute!")
```
superlopuh, line 525:
```suggestion
                self.raise_error("Expected a MemRef layout attribute")
```

xdsl/dialects/memref.py
```diff
@@ -167,7 +168,7 @@ def get(
         alignment: int | AnyIntegerAttr | None = None,
         shape: Iterable[int | IntAttr] | None = None,
         dynamic_sizes: Sequence[SSAValue | Operation] | None = None,
-        layout: Attribute = NoneAttr(),
+        layout: MemrefLayoutAttr | NoneAttr = NoneAttr(),
```
superlopuh, line 171: excellent

xdsl/dialects/builtin.py
PapyChacal: Just doing it, but why? It is not declaring any abstract method, what's the point?

superlopuh: Seems like a more protective default, if it's not designed to be initialized, then why not make that throw an error at runtime?

xdsl/dialects/builtin.py
```diff
@@ -1175,8 +1175,18 @@ def from_strings(name: str, value: str, type: Attribute = NoneAttr()) -> OpaqueA
         return OpaqueAttr([StringAttr(name), StringAttr(value), type])
 
 
+class MemrefLayoutAttr(Attribute, ABC):
+    """
+    Interface for any attribute acceptable as a memref layout.
+    """
+
+    name = "memref_layout_att"
```
superlopuh, line 1183: Why provide a name?

xdsl/dialects/builtin.py
PapyChacal: Oh, forgot about the instanciation guard. Good reminder, thanks!

xdsl/dialects/builtin.py
PapyChacal: Oh no, this didn't just slip through :face_with_peeking_eye: 

This is a can of worm, and just putting a name here avoids to open it, can we separate the can of worm from this very practical PR?

But, to answer:
When using such a type in an IRDL constant, some verification errors try to access its name for the error reporting, erorring the error :upside_down_face:  (Also, it's defined at the `Attribute` baseclass level that they should all have a name, so Pyright says "okay" to that.) So there's a whole little discussion to have around that, but practically and currently, we need any base attribute used in Constraints to have a name. And per type consistency, they really just all should have a `name` - Pyright assumes so.

superlopuh: Very interesting. Surely this is worth at least an issue? I think it would be great to have at least a bit of a discussion of some potential future.

PapyChacal: Definitely worth an issue, will file one in a moment if not done

And yeah, worth the discussion (or a type system 😂), I just really want this change independently in, as this is holding back a concrete project of our Belgium collaborators 🙂

superlopuh: Thinking about it a little more, I'd recommend adding abstract in the name, filing an issue, and leaving the rest for later

jorendumoulin (approved):
I have tested this on the snax side, and everything works beautifully. Thanks a lot for your help on this!

Somewhat unrelated: the changes mix the use of `isinstance` and `isa`. When should one use which?

superlopuh:
`isinstance` is a built-in Python thing, and I would say is the recommended thing to use when it's sufficient. `isa` is an xDSL function, that can handle more things, notably Python generics and PyRDL constraints.

xdsl/dialects/builtin.py
PapyChacal: I'm okay with that!

math-fehr: Oh god 🙈

xdsl/parser/attribute_parser.py
```diff
@@ -520,27 +521,18 @@ def _parse_memref_attrs(
...
+            if not isinstance(memory_or_layout, MemrefLayoutAttr):
+                self.raise_error("Expected a MemRef layout attribute")
             return MemRefType(type, shape, memory_or_layout, memory_space)
 
-        # Otherwise, there is a single argument, so we check based on the
-        # attribute type. If we don't know, we return an error.
-        # MLIR base itself on the `MemRefLayoutAttrInterface`, which we do not
-        # support.
+        # If the argument is a MemrefLayoutAttr, use it as layout
+        if isa(memory_or_layout, MemrefLayoutAttr):
```
math-fehr, line 529: This can be an `isinstance` right?

math-fehr (approved):
Thanks for implementing this!
I do remember not implementing it properly, and I do really HATE what MLIR is doing here, but that's life.
It is one of the rare case in MLIR where parsing depends on an interface, which means here that if you want to use a memory layout that is unregistered, then the parsing will fail even though we may want it to work (Just saying this so you know in case you encounter this).

PapyChacal:
> Thanks for implementing this! I do remember not implementing it properly, and I do really HATE what MLIR is doing here, but that's life. It is one of the rare case in MLIR where parsing depends on an interface, which means here that if you want to use a memory layout that is unregistered, then the parsing will fail even though we may want it to work (Just saying this so you know in case you encounter this).

Good point; but how do you know if your unregistered attribute is a memory space or layout then?

PR #2720, opened 2024-06-13
transform: stencil to csl - generate layout module

n-io:
Adding a transform from stencil to CSL.

* Perform initial pass over the `stencil.apply` ops to figure out stencil pattern and compute grid size (done)
* Generate CSL layout module (done, except for exported pointers)
* Generate basic CSL program module (params/imports to setup module structure, translating stencil computation will be done separately)


xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,304 @@
...
+    def add_var_to_layout(self, name: str, op: Operation):
+        self.layout_module.body.block.add_op(op)
+        self.layout_vars[name] = op
+
+
+def setup_program_module(ctx: TranslationContext) -> None:
+    ctx.add_params_to_program(
+        ParamOp.get("stencil_comms_params", ComptimeStructType()),
+        ParamOp.get("memcpy_params", ComptimeStructType()),
+        ParamOp.get("z_dim", IntegerType(16, Signedness.SIGNED)),
```
AntonLydike, line 96: I'd recommend not using signedness unless you *really* need an unsigned integer!

AntonLydike: dito for all other integer variables

n-io: I'd caution against that - most of what is happening here is setting up module (library) imports, calling module functions, and using that to set params for the `pe.csl`. It might just be harder to reverse-engineer it in the printer.

dk949: From the point of view of the printer using signed or signless is exactly the same (they both get printed as signed).

n-io: we might need to look into that, some functions have specific requirements and may only accept one of the two

n-io:
> Please add comments and docstrings to this PR, it's really hard to follow what you are doing / want to do because of the distinct lack of comments.

Added comments and doc strings as suggested

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,336 @@
...
+    ctx.add_var_to_program(
+        "stencil_comms_params_combined",
+        stencil_comms_params_combined := ConcatStructOp.get(
+            stencil_comms_params_ext, ctx.program_params["stencil_comms_params"]
+        ),
+    )
+
+    ctx.add_import_module_to_program(
+        ImportModuleConstOp.get("stencil_comms.csl", stencil_comms_params_combined),
+    )
```
AntonLydike, line 163: I would really like to see comments what these do and why they are added

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+
+def generate_program_module(ctx: TranslationContext) -> None:
+    """
+    Generates the basic structure of a program module needed for stencil computations.
+    This function should *not* translate the actual stencil computation, but set up everything
+    needed for these, including imports, params, and top-level constants.
+    Params are generated to match `@set_tile_code` invocation of the
+    layout module. Also imports memcpy, utils, and stencil_comms modules. Vars/consts
+    are set up as required to perform these imports.
+    """
```
AntonLydike, line 122: I assume that this method makes some assumptions about the given context, (e.g. the number of dimensions of the computation). Can we get these explicitly mentioned in the docstring and maybe with an assert?

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
+from attr import dataclass
+
+from xdsl.dialects import func
+from xdsl.dialects.arith import Cmpi, Constant, MinUI, Muli, OrI
```
AntonLydike, line 4: Instead of importing individual operations, we usually import the module, and then access operations as `arith.Muli`. This makes the resulting code much more readable, as you can immediately tell to which dialect an operation belongs.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+        s_params := ParamOp.get(Named.stencil_comms_params, ComptimeStructType()),
+        m_params := ParamOp.get(Named.memcpy_params, ComptimeStructType()),
+        ParamOp.get(Named.is_border_region_pe, IntegerType(1)),
+    )
+
+    # size of the z dimension
+    ctx.add_var_to_program(
+        Named.z_dim,
+        z_dim := Constant(
+            IntegerAttr(ctx.grid_dim[2], IntegerType(16, Signedness.UNSIGNED))
```
AntonLydike, line 135: Is there a reason why this is explicitly marked unsigned? See https://mlir.llvm.org/docs/Rationale/Rationale/#integer-signedness-semantics for more info on signedness.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+    """
+
+    pattern = ("pattern",)
+    stencil_comms_params = ("stencil_comms_params",)
+    memcpy_params = ("memcpy_params",)
+    z_dim = ("z_dim",)
+    is_border_region_pe = ("is_border_region_pe",)
+
+
+class TranslationContext:
```
AntonLydike, line 51: I feel like this would benefit from a `@dataclass` annotation

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+    program_module: CslModuleOp
+    layout_module: CslModuleOp
+    program_sym_name: StringAttr
+
+    program_params: dict[str, ParamOp] = {}
+    layout_params: dict[str, ParamOp] = {}
+    program_imports: dict[str, ImportModuleConstOp] = {}
+    layout_imports: dict[str, ImportModuleConstOp] = {}
+    program_vars: dict[str, Operation] = {}
+    layout_vars: dict[str, Operation] = {}
```
AntonLydike, line 71: Having mutable defaults here is a footgun of python, see e.g. https://docs.astral.sh/ruff/rules/mutable-dataclass-default/

You probably want to have these be initialized empty on every object instantiation.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+    )
+
+    # when sending / receiving buffers with stencil_comms, use padded_z_dim rather than z_dim as buffer size
+    ctx.add_var_to_program("padded_z_dim", Muli(num_chunks, chunk_size))
+
+    # stencil pattern, i.e. neighbours in any direction + 1
+    ctx.add_var_to_program(
+        Named.pattern,
+        pattern_op := Constant(
+            IntegerAttr(ctx.pattern, IntegerType(16, Signedness.UNSIGNED))
```
AntonLydike, line 176: Same signedness argument here

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+            SetTileCodeOp.get(
+                ctx.program_module.sym_name,
+                x_id,
+                y_id,
+                set_tile_params_ext,
+            ),
+            Yield(),
+        ]
+    )
+    outer_loop_body.add_op(Yield())
```
AntonLydike, line 341: I would highly suggest you look into `ImplicitBuilder`, which is our attempt at making IR construction readable :)

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+            attributes={"sym_name": StringAttr("pe.csl")},
+        )
+        t_ctx.layout_module = CslModuleOp(
+            regions=[Region(Block())],
+            properties={"kind": ModuleKindAttr(ModuleKind.LAYOUT)},
+            attributes={"sym_name": StringAttr("layout.csl")},
+        )
+        # iterate over stencil.apply ops to find the maximum stencil width ('arm length')
+        assert len(op.body.block.ops) == 1
+        assert isinstance(stencil_func := op.body.block.first_op, func.FuncOp)
```
AntonLydike, line 363: I don't think asserting something this strong about program structure is a great idea, as it will just crash in most cases.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+                arg.type.element_type, TensorType[Attribute]
+            ):
+                shape = arg.type.get_shape()
+                t_ctx.grid_dim = tuple(
+                    max(a, b)
+                    for a, b in zip(
+                        t_ctx.grid_dim,
+                        (shape[0], shape[1], arg.type.element_type.get_shape()[0]),
+                    )
+                )
```
AntonLydike, line 386: Let's think about how we actually want this conversion to happen. I don't think this is a very scalable approach.

xdsl/dialects/csl.py
```diff
@@ -376,6 +376,14 @@ class ImportModuleConstOp(IRDLOperation):
 
     result = result_def(ImportedModuleType)
 
+    @staticmethod
+    def get(name: str, *params: SSAValue | Operation) -> ImportModuleConstOp:
+        return ImportModuleConstOp(
+            operands=[[]] if len(params) == 0 else params,
+            result_types=[ImportedModuleType()],
+            properties={"module": StringAttr(name)},
+        )
```
AntonLydike, line 385: I would prefer to see this as a custom `__init__` instead of a `.get`. We are trying to move away from these.

xdsl/dialects/csl.py
```diff
@@ -398,6 +406,19 @@ def verify_(self) -> None:
...
+        operands: list[Operation] = []
+        fields: list[StringAttr] = []
+        for fname, op in args:
+            fields.append(StringAttr(fname))
+            operands.append(op)
+        return ConstStructOp(
+            operands=[operands],
+            result_types=[ComptimeStructType()],
+            properties={"ssa_fields": ArrayAttr(fields)},
+        )
```
AntonLydike, line 420: dito

xdsl/dialects/csl.py
```diff
@@ -406,6 +427,10 @@ class GetColorOp(IRDLOperation):
     id = operand_def(IntegerType)
     res = result_def(ColorType)
 
+    @staticmethod
+    def get(op: Operation) -> GetColorOp:
+        return GetColorOp(operands=[op], result_types=[ColorType()])
+
```
AntonLydike, line 433: dito

xdsl/dialects/csl.py
```diff
@@ -438,6 +463,21 @@ class MemberCallOp(IRDLOperation):
...
+        struct: Operation,
+        *params: SSAValue | Operation,
+    ) -> MemberCallOp:
+        return MemberCallOp(
+            operands=[struct, params],
+            result_types=[result_type],
+            properties={
+                "field": StringAttr(fname),
+            },
+        )
```
AntonLydike, line 479: dito

xdsl/dialects/csl.py
```diff
@@ -1486,6 +1538,14 @@ class ParamOp(IRDLOperation):
 
     res = result_def(T)
 
+    @staticmethod
+    def get(name: str, result_type: T) -> ParamOp:
```
AntonLydike, line 1542: `T` is not a valid type here, I think? @PapyChacal

xdsl/dialects/csl.py
```diff
@@ -1511,6 +1571,20 @@ def verify_(self) -> None:
...
+        assert isinstance(typ, IntegerType)
+        return SignednessCastOp(operands=[op], result_types=[IntegerType(typ.width)])
+
+    @staticmethod
+    def get_u(op: SSAValue | Operation) -> SignednessCastOp:
+        typ = op.results[0].type if isinstance(op, Operation) else op.type
+        assert isinstance(typ, IntegerType)
+        return SignednessCastOp(
+            operands=[op], result_types=[IntegerType(typ.width, Signedness.UNSIGNED)]
+        )
```
AntonLydike, line 1586: I would prefer to see these two as a single `def __init__(self, inp: SSAValue | Operation, result_type: IntegerType)`

AntonLydike (commented):
I think this is a great start, but I am not sure if the way we integrate this is the right path right now. I would prefer to have something closer to how GPU dialect does offloading.

I think that would result in much more usable passes and would make our life significantly easier.

I think we need a mid-level intermediate dialect for that though.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
```diff
@@ -0,0 +1,399 @@
...
+    """
+    Enum type to store names used across the program, for instance
+    names of parameters passed from the layout to the program module.
+    """
+
+    pattern = ("pattern",)
+    stencil_comms_params = ("stencil_comms_params",)
+    memcpy_params = ("memcpy_params",)
+    z_dim = ("z_dim",)
+    is_border_region_pe = ("is_border_region_pe",)
```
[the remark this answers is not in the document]
n-io: Cause `make precommit` turned them into tuples 🤷‍♂️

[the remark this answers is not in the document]
AntonLydike: Ah, because you had a trailing comma! That makes it implicitly a tuple. You don't need commas in enum definitions!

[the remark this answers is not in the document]
n-io: probably because I ended each line with a comma

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: good point

xdsl/dialects/csl.py
n-io: actually, we don't need the result type if the behaviour is simply to flip between unsigned and signless (since we don't want explicit 'signed')

n-io: otherwise done as suggested

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: (see discussion above)

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: thanks, will take a look. It seems mostly used for testing though?

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: I'm afraid we will need some variant of this, as the communications library is bound to a fixed stencil size (which I suppose needs to be the maximum we encounter throughout the program).

The grid dimensions are primarily (though not exclusively) needed for the 2-dimensional loop performed in layout.csl to `set_tile_code` for each PE. The idea here is to derive the size of the compute grid from the size of the buffers passed in to the stencil function - the maximum size, to be safe.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: There now is an assert in the main apply and in the type

n-io:
> I think this is a great start, but I am not sure if the way we integrate this is the right path right now. I would prefer to have something closer to how GPU dialect does offloading.

This PR doesn't actually do very much except initialising a layout module and a stub program module that takes care of some administrative tasks such as importing things and initialising the stencil comms library. I'm hoping this should be non-committal as to how we go about further translating.

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
AntonLydike: What do you mean by main dialect? But usually this is done for all dialects.

n-io: I mean csl, sorry

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: But then I can't do a standard dict access to the fields..

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: This is actually quite neat, using the ImplicitBuilder now

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: Added explicit mention in the above docstring

xdsl/backend/csl/lowering/convert_stencil_to_csl.py
n-io: done as suggested

n-io:
We expect that we will need to generate csl modules very soon and are benching this PR until then.

superlopuh:
Any updates on this? Let's close it? @n-io

PR #2721, opened 2024-06-13
dialects: (memref_stream) Add some more verification for generic op

superlopuh:
We currently don't check as much as we could, this PR adds some more verification.


xdsl/dialects/memref_stream.py
```diff
@@ -530,6 +530,43 @@ def verify_(self) -> None:
...
+        # Whether or not the operation represents an imperfect loop nest, verify that the
+        # bounds of the outer + inner nests match the domain of the input affine maps
+        input_count = len(self.inputs)
+        input_maps = self.indexing_maps.data[:input_count]
+
+        for i, m in enumerate(input_maps):
+            if m.data.num_symbols:
+                raise NotImplementedError(
+                    f"Symbols currently not implemented in {self.name} indexing maps"
+                )
```
PapyChacal, line 547: I find this strange in a verifier; what is not implemented? The lowering?

xdsl/dialects/memref_stream.py
```diff
@@ -530,6 +530,43 @@ def verify_(self) -> None:
...
+        if min_dims != max_dims:
+            raise VerifyException(
+                "The number of dims in output indexing maps must all be the same"
+            )
+
+        if min_dims not in (len(iterator_types), num_parallel):
+            raise VerifyException(
+                "The number of dims in output indexing maps must be "
+                f"{len(iterator_types)} or {num_parallel}"
+            )
```
PapyChacal, line 568: This one bugs me too. I guess this is about special handling of reductions? Can you explain what it means to have `len(iterator_types)` dimensions in the output indexing map in the presence of reduction iterators?

superlopuh:
I added a comment, and moved the not implemented error, does that make it clearer?

PapyChacal (approved):

PR #2722, opened 2024-06-13
dependencies: move dev dependencies to pyproject.toml

superlopuh:
This PR unifies the dependency tracking information into `pyproject.toml`. This lets us do some useful things, like installing all the dev requirements without the "extras", for CI, and for GitHub dev containers (#2550). (Or for local developement, when spinning up a quick venv into which you don't feel like installing wgpu and onnx and others.)


PapyChacal (approved):
Thanks!!

georgebisbas (approved):
I think that this info should also be somehow communicated to the installation instructions as well.
https://github.com/xdslproject/xdsl/blob/main/README.md

PR #2723, opened 2024-06-13
dependencies: split "extras" dependencies into named subprojects

superlopuh:
Note that this does not change the flow for local install, which still goes via `requirements.txt`, it just means that `pip install xdsl[extras]` must now be changed to specify exactly which subprojects you want as the dependency. This gives us the flexibility of adding potentially quite large dependencies without forcing users who use just one of them to install them all.


PapyChacal (approved):
Thanks a lot!

Obvious annoying question: is there a way to add some kind of "etra-all" if that makes sense? Just concatenating all subprojects for whenever we don't want to think about it and get it all?

Defnitely in favor of splitting first anyway :)

georgebisbas (approved):

superlopuh:
@PapyChacal, I have found a pep that seems to propose this, I don't know of a way to do this right now. I would just say that people who want to use xDSL are unlikely to use any of the subprojects, and that people contributing to xDSL should continue to use `pip install -r requirements.txt`.

PR #2734, opened 2024-06-16
misc: Implement Dialect stub generation

PapyChacal:
Implement Python stub file generation for an irdl dialect.


xdsl/ir/core.py
```diff
@@ -388,6 +389,11 @@ def __eq__(self, other: object) -> bool:
         return self is other
 
 
+Operand: TypeAlias = SSAValue
+VarOperand: TypeAlias = list[Operand]
```
superlopuh, line 393: oh I forgot this was a list, that's weird

xdsl/utils/dialect_stub.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+
+    def _import(self, object: Any, alias: str = ""):
+        module = object.__module__
+        if module == "builtins":
+            return
+        if module.startswith("xdsl.ir."):
+            module = "xdsl.ir"
+        if module.startswith("xdsl.irdl."):
+            module = "xdsl.irdl"
```
superlopuh, line 66: this seems a little hacky, don't we want to make this work anywhere, including in dependencies of xDSL?

PapyChacal: What do you mean by make this work anywhere?

This simply removes the banned nested modules to make Pyright happy about the output stubs, it, don't import from `xdsl.ir.core` but just from `xdsl.ir`. Then, be it one of those modules or any other one, it will be imported, including external dependencies. Note that, this is intended to generate stubs for parsed irdl dialect, so I'm not sure when irdl will reach a stage able to map to external Python dependencies?

The first line is about not importing builtin types like `int` or `str` - not sure which one triggered it now :thinking:

xdsl/ir/core.py
PapyChacal: Indeed, let's think about it independently though if that's fine to you? Apparently, it ain't broken

xdsl/ir/core.py
```diff
@@ -387,6 +388,11 @@ def __eq__(self, other: object) -> bool:
...
```
math-fehr, line 392: It feels weird to me that this is now moved to `core.py`, where these are `irdl` concepts?
Did you have to do this because of cycle of reference? If that's the case, then that's fine.

math-fehr (commented):
Really cool! Whenever you add the documentation it should be good to me!
I'm not a big fan of moving things like `OptOperand` in `irdl.core`, as I feel they are out of place there.
What was the problem you had when they were in `irdl.py`? Just trying to understand, if this is too much of a problem I'm fine moving them in `core.py`, even if it's ugly.

xdsl/utils/dialect_stub.py
superlopuh: Ah, I see. I think I need to check out the branch and play with it a little bit for more insightful comments than "this looks weird" :)

PapyChacal: I guess I can offer comments too 😉

xdsl/ir/core.py
PapyChacal: It's more sneaky than this;

Type hints are for typecheckers only, right? So, as a Python object, Operand actually *is* SSAValue.
Which means `Operand.__module__` actually *is* `SSAValue.__module__` AKA `xdsl.ir.core`.

So, when generating their import statements in stub, it would need special-casing to do it right, or just resort back to type them `SSAValue` on `IRDLOperation`s, which sounds too bad :thinking: 

Seeing it from that point of view, having `Operand` defined in `xdsl.irdl.core` to end up having `__module__` as `xdsl.ir.core` is what felt weird to me.

Now that I mention it though, it solves half the problem, 'cause I'll have to special case `VarOperand` anyway, which as `list` would have `builtins` as `__module__` :smiling_face_with_tear: So, I'll think about it and maybe just make it less fancy on the type stub side..

PapyChacal: Reverted the moving around, and solved inside dialect_stub.py

superlopuh (approved):

PapyChacal:
> Really cool! Whenever you add the documentation it should be good to me! I'm not a big fan of moving things like `OptOperand` in `irdl.core`, as I feel they are out of place there. What was the problem you had when they were in `irdl.py`? Just trying to understand, if this is too much of a problem I'm fine moving them in `core.py`, even if it's ugly.

Reverted core changes and added doc!

math-fehr (approved):
👍, thanks a lot!
I would just add a `_generate` prefix to your methods (right now they look like they return a field), but otherwise that looks good to me!

PR #2735, opened 2024-06-16
misc: Implement Python module importing from `.irdl` files

PapyChacal:
Stacked on:
- #2749 

This achieves just importing `.irdl` files (MLIR IR files supposed to contain an irdl dialect) transparently as Python modules, automatically translating the irdl dialect to a PyRDL one and initializing a Python module accordingly, and also generating Python type stubs to be able to programmatically use the said Dialect with type-checking :rocket:

There is *no* way to make pyright benefit from this transparency though, because *it will never run Python code* :face_exhaling: So, I also provide `xdsl-stubgen`, which simply go over `.irdl` files in the `xdsl.dialect` module's directory (Read "including in the install destination if that's what's run"), and importing them, just manually triggering all stubs generation. Add this as a prelude to pyright in CI and Makefile, for the rule of least surprise?

Demo included with `xdsl/dialects/cmath.py` replaced by `xdsl/dialects/cmath.irdl` but registered as any dialect!
NB: This renamed the dialect object to Cmath instead of CMath, because, well, automatic capitalization :shrug: Sounds minimal to me - @math-fehr, maybe we should just use capitalized names in irdl operations?


PapyChacal:
Note: it worked when running tests sequentially; but failed in CI because Pyright is not triggering the actual cmath import, itself triggering the type stubs generation.

Does anyone have an idea on where to best hook that in?

xdsl/ir/context.py
```diff
@@ -0,0 +1,533 @@
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from xdsl.ir import Attribute, Dialect, Operation
+
+
+def get_all_dialects() -> "dict[str, Callable[[], Dialect]]":
```
superlopuh, line 9: relatively minor, but if we're splitting things, I'd rather move this to its own file, and it feels a bit more sense to put it in the dialects folder, but that may be a conversation for a different PR

PapyChacal: Why not, I do think this PR could be split, it's a bit drowning the actual added functionality.
I'll split it after the bases are merged for sanity though :slightly_smiling_face:

.gitignore
```diff
@@ -156,6 +156,9 @@ docs/opt-out.mlir
 docs/out.mlir
 docs/ret.xdsl
 
+# Type stubs
```
math-fehr, line 159:
```suggestion
# Type stubs
# These are generated type stubs, we do not otherwise use type stubs in the project
```

xdsl/dialects/cmath.py
```diff
@@ -1,74 +0,0 @@
-from __future__ import annotations
```
math-fehr, line 1: Nice

xdsl/ir/context.py
math-fehr, line 9: Why do you need the "" here? Can you `from __future__ import annotations`? I forgot if this is supposed to work or not.

math-fehr: Yeah, I would agree moving it to its own file, for me it's just because I really don't think it goes well with `MLContext`, which is supposedly a dialect-agnostic thing

xdsl/__init__.py
```diff
@@ -1,3 +1,79 @@
...
 
 __version__ = _version.get_versions()["version"]
+
+import importlib.abc
+import importlib.util
+import os
+import sys
+
+
+class CustomFileLoader(importlib.abc.Loader):
```
math-fehr, line 18: Can you add a bit of documentation here and in the next class?

math-fehr: And can we rename the class? :)

xdsl/__init__.py
```diff
@@ -1,3 +1,79 @@
...
+                    fullname, potential_path, loader=loader
+                )
+        return None
+
+
+def parse_custom_file(content: str):
+    pass
+
+
+sys.meta_path.insert(0, CustomFileFinder())
```
math-fehr, line 79: Really nice

math-fehr (commented):
Really cool to see this working ;)

xdsl/__init__.py
PapyChacal: What a direct attack against state-of-the-art ChatGPT :open_mouth: 
Forgot to rename the boilerplate I kept ahah, will rename more IRDLy

xdsl/ir/context.py
PapyChacal: I'm not sure why annotations? quoted type hints are for not adding runtime dependencies because of type hints. There is a cyclic dependency if I don't use quotes here; Should be solved by moving this function to its own file!

.gitignore
PapyChacal: Good point! But as discussed, we will try to commit the stubs, until we find something more satisfactory for Pyright

PapyChacal: (more like a self-note)

xdsl/ir/context.py
PapyChacal: moved `get_all_dialects` to `xdsl.dialects`, and `MLContext` to root `xdsl`, WDYT?

PapyChacal: Indeed, solved

PapyChacal:
> would it make sense to put the move from the context object out of the `ir` module in a separate PR? I feel it introduces a lot of noise in this one.

Yes, this is planned as soon as the base is merged

.gitignore
PapyChacal: I'm actually more for the current state I'm offering; aka explicitly generating stub before pyright when we want to be sure. I really don't like committing autogenerated stuff if I can avoid it.
So, adding this clearer comment :yum:

xdsl/ir/context.py
PapyChacal: All moved out to:
- #2749

Makefile
```diff
@@ -69,6 +69,7 @@ precommit:
 
 # run pyright on all files in the current git commit
 pyright:
+	xdsl-stubgen
```
math-fehr, line 72:
```suggestion
        # We make sure to generate the python typing stubs before running pyright
	xdsl-stubgen
```

math-fehr (approved):
Really nice, thanks a lot for splitting the PRs!

xdsl/__init__.py
```diff
@@ -1,3 +1,96 @@
...
+        # Open the irdl file
+        with open(self.path) as file:
+            # Parse it
+            ctx = MLContext()
+            for dialect_name, dialect_factory in get_all_dialects().items():
+                ctx.register_dialect(dialect_name, dialect_factory)
+            irdl_module = Parser(ctx, file.read(), self.path).parse_module()
+
+            # Make it a PyRDL Dialect
+            dialect_name = os.path.basename(self.path)[:-5]
```
superlopuh, line 50: why -5?

PapyChacal: It's "remove the .irdl extension"; would you rather have it be a library call or smth?

superlopuh: I considered suggesting using Path but maybe it's overkill here. It may be worth adding `Drop ".irdl"` as a comment or using Path to remove the extension, your call.

superlopuh (approved):

xdsl/__init__.py
PapyChacal: Done the call thing!

PR #2742, opened 2024-06-17
dialects: (csl) adding constructors for CSL ops

n-io:
Adding constructors for many CSL ops.

This was originally part of #2720

### DIFF
--- a/xdsl/dialects/csl.py
+++ b/xdsl/dialects/csl.py
@@ -376,9 +376,9 @@ class ImportModuleConstOp(IRDLOperation):
 
     result = result_def(ImportedModuleType)
 
-    def __init__(self, name: str, *params: SSAValue | Operation):
+    def __init__(self, name: str, params: SSAValue | Operation | None = None):
         super().__init__(
-            operands=[[]] if len(params) == 0 else params,
+            operands=[params],
             result_types=[ImportedModuleType()],
             properties={"module": StringAttr(name)},
         )
@@ -745,7 +745,7 @@ class SetTileCodeOp(IRDLOperation):
         fname: str | StringAttr,
         x_coord: SSAValue | Operation,
         y_coord: SSAValue | Operation,
-        params: SSAValue | Operation,
+        params: SSAValue | Operation | None = None,
     ):
         name = StringAttr(fname) if isinstance(fname, str) else fname
         super().__init__(operands=[x_coord, y_coord, params], properties={"file": name})
--- a/xdsl/dialects/csl.py
+++ b/xdsl/dialects/csl.py
@@ -376,6 +376,13 @@ class ImportModuleConstOp(IRDLOperation):
 
     result = result_def(ImportedModuleType)
 
+    def __init__(self, name: str, *params: SSAValue | Operation):
+        super().__init__(
+            operands=[[]] if len(params) == 0 else params,
+            result_types=[ImportedModuleType()],
+            properties={"module": StringAttr(name)},
+        )
+
 
 @irdl_op_definition
 class ConstStructOp(IRDLOperation):
@@ -398,6 +405,18 @@ class ConstStructOp(IRDLOperation):
             "Number of ssa_fields has to match the number of arguments"
         )
 
+    def __init__(self, *args: tuple[str, Operation]):
+        operands: list[Operation] = []
+        fields: list[StringAttr] = []
+        for fname, op in args:
+            fields.append(StringAttr(fname))
+            operands.append(op)
+        super().__init__(
+            operands=[operands],
+            result_types=[ComptimeStructType()],
+            properties={"ssa_fields": ArrayAttr(fields)},
+        )
+
 
 @irdl_op_definition
 class GetColorOp(IRDLOperation):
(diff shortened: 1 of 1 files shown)


xdsl/dialects/csl.py
```diff
@@ -376,6 +376,13 @@ class ImportModuleConstOp(IRDLOperation):
 
     result = result_def(ImportedModuleType)
 
+    def __init__(self, name: str, *params: SSAValue | Operation):
+        super().__init__(
+            operands=[[]] if len(params) == 0 else params,
```
AntonLydike, line 381: Two things:

1. Params should be a single `SSAValue | Operation | None = None` operand, as `import_module` only takes a single argument 
2. I'm 99% sure that you can just write ` operands=[param]` (in both cases where param is a vararg, and if it's optional)

xdsl/dialects/csl.py
```diff
@@ -438,6 +460,21 @@ class MemberCallOp(IRDLOperation):
 
     result = opt_result_def(Attribute)
 
+    def __init__(
+        self,
+        fname: str,
+        result_type: Attribute,
+        struct: Operation,
+        *params: SSAValue | Operation,
```
AntonLydike, line 468: I think I'd like this to not be a vararg, and instead be a `Sequence[SSAValue | Operation]`.

Having a vararg after a bunch of other arguments would make the code less readable at the call site imo.

xdsl/dialects/csl.py
```diff
@@ -703,6 +740,16 @@ class SetTileCodeOp(IRDLOperation):
     y_coord = operand_def(IntegerType)
     params = opt_operand_def(ComptimeStructType)
 
+    def __init__(
+        self,
+        fname: str | StringAttr,
+        x_coord: SSAValue | Operation,
+        y_coord: SSAValue | Operation,
+        params: SSAValue | Operation,
```
AntonLydike, line 748: This should be `SSAValue | Operation | None` as it's optional

xdsl/dialects/csl.py
```diff
@@ -1486,6 +1533,13 @@ class ParamOp(IRDLOperation):
 
     res = result_def(T)
 
+    def __init__(self, name: str, result_type: T):
```
AntonLydike, line 1536: I'm not sure if this is allowed in python... @PapyChacal do you have any input on this? I think you know more on this than I do...

xdsl/dialects/csl.py
```diff
@@ -1511,6 +1565,18 @@ def verify_(self) -> None:
...
+        typ = op.results[0].type if isinstance(op, Operation) else op.type
+        assert isinstance(typ, IntegerType)
+        res_signedness = (
+            Signedness.SIGNLESS
+            if typ.signedness.data == Signedness.UNSIGNED
+            else Signedness.UNSIGNED
+        )
+        super().__init__(
+            operands=[op], result_types=[IntegerType(typ.width, res_signedness)]
+        )
```
AntonLydike, line 1578: I think I'd prefer a much simpler interface that is in-line with other cast operations:

```suggestion
    def __init__(self, op: SSAValue | Operation, result_type: IntegerType):
        super().__init__(
            operands=[op], result_types=[result_type]
        )
```

This way the caller can decide what he wants instead of the init smartly guessing.

AntonLydike (approved):

AntonLydike:
A very nice set of changes, I just have a bunch of nitpicks strewn about :D

xdsl/dialects/csl.py
n-io: If you prefer a simple interface, I suggest it's simplest to do what the op promises to do, and just flip the signedness.

xdsl/dialects/csl.py
n-io: I thought that's what make TypeVars so neat
https://mypy.readthedocs.io/en/stable/generics.html

xdsl/dialects/csl.py
n-io: changing as suggested

xdsl/dialects/csl.py
AntonLydike: I don't think it's necessarily wrong, but it's not expressing what you may think it is.

xdsl/dialects/csl.py
n-io: As long as it's limited to function name+type, it should hopefully be fine enough to avoid me doing cross-PR changes (:

xdsl/dialects/csl.py
AntonLydike: But you may want to cast from `i32` to `si32`, which this constructor does not allow.

xdsl/dialects/csl.py
n-io: done as suggested

xdsl/dialects/csl.py
n-io: I understood we didn't want to have `si32`, but I added an optional arg to explicitly state the result type, and if it's `None` we derive it.

xdsl/dialects/csl.py
AntonLydike: I'm strongly in favour of switching to `Sequence` here, as it's otherwise out of line with the other call operations...

xdsl/dialects/csl.py
n-io: I misread, it's not a TypeVar. I think it's not an IRDL constraint either, though, it should just be a union type with some additional meta data.

PR #2743, opened 2024-06-17
interpreter: (memref_stream) add interpreter support for imperfect nesting

superlopuh:
This makes it a little easier to test for miscompilations in our flow from linalg to snitch


PapyChacal (approved):
Code looks good to me, I'm gonna trust you on the semantics!

PR #2744, opened 2024-06-17
dialects: (builtin) Fix printing and parsing of `UnitAttr`

superlopuh:
It's actually `unit` not the empty string.

### DIFF
--- a/tests/filecheck/parser-printer/builtin_attrs.mlir
+++ b/tests/filecheck/parser-printer/builtin_attrs.mlir
@@ -8,12 +8,9 @@
 
   // CHECK: (index)
 
-  "func.func"() ({
-    ^bb0(%arg0: unit):
-    "func.return"() : () -> ()
-  }) {function_type = (unit) -> (), sym_name = "unit_type_func"} : () -> ()
+  "func.func"() ({}) {function_type = () -> (), sym_name = "unit_attr_func", unitarray = [unit]} : () -> ()
 
-  // CHECK: (unit)
+  // CHECK: "unitarray" = [unit]
 
   "func.func"() ({
     ^bb0(%arg0: i32, %arg1: i64, %arg2: i1):
--- a/tests/filecheck/parser-printer/builtin_attrs.mlir
+++ b/tests/filecheck/parser-printer/builtin_attrs.mlir
@@ -8,6 +8,12 @@
 
   // CHECK: (index)
 
+  "func.func"() ({
+    ^bb0(%arg0: unit):
+    "func.return"() : () -> ()
+  }) {function_type = (unit) -> (), sym_name = "unit_type_func"} : () -> ()
+
+  // CHECK: (unit)
 
   "func.func"() ({
     ^bb0(%arg0: i32, %arg1: i64, %arg2: i1):
--- a/xdsl/parser/attribute_parser.py
+++ b/xdsl/parser/attribute_parser.py
@@ -595,7 +595,6 @@ class AttrParser(BaseParser):
             return BytesAttr(bytes_lit)
 
         attrs = (
-            self.parse_optional_unit_attr,
             self.parse_optional_builtin_int_or_float_attr,
             self._parse_optional_array_attr,
             self._parse_optional_symref_attr,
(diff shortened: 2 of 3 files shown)


xdsl/parser/attribute_parser.py
```diff
@@ -1250,6 +1250,7 @@ def _parse_optional_integer_or_float_type(self) -> Attribute | None:
         Parse as integer or float type, if present.
           integer-or-float-type ::= index-type | integer-type | float-type
           index-type            ::= `index`
+          unit-attribute        ::= `unit`
```
superlopuh, line 1253: I first tried to add this as a normal attribute not type, but then the test failed, saying "type expected". I have a supsicion it's actually a type in MLIR also, but my quick search for the definition didn't yield much so I thought I'd open a PR anyway.

math-fehr: It is not a type, it's an attribute though: https://mlir.llvm.org/docs/Dialects/Builtin/#unitattr

superlopuh: Should I change the test instead? When I put the code for parsing into `parse_attribute` it fails, as the block argument expects a type.

xdsl/parser/attribute_parser.py
```diff
@@ -1262,6 +1263,11 @@ def _parse_optional_integer_or_float_type(self) -> Attribute | None:
             self._consume_token()
             return IndexType()
 
+        # Unit type
+        if name == "unit":
+            self._consume_token()
+            return UnitAttr()
+
```
math-fehr, line 1270: This is not the right place to add this though, this is neither an integer nor attribute type?

math-fehr (commented):
Thanks for looking at this!
`unit` is an attribute, not a type in MLIR, and we should stay the same.
So here I would not add it in the `parse_optional_integer_or_float_type`

xdsl/parser/attribute_parser.py
math-fehr: Which test? And yes, we should change it I think

tests/filecheck/parser-printer/builtin_attrs.mlir
```diff
@@ -8,6 +8,9 @@
 
   // CHECK: (index)
 
+  "func.func"() ({}) {function_type = () -> (), sym_name = "unit_attr_func", unitarray = [unit]} : () -> ()
```
AntonLydike, line 11: Can I suggest using the `test` dialect here? I don't think we want to depend on `func` here...

xdsl/parser/attribute_parser.py
```diff
@@ -651,6 +652,22 @@ def _parse_strided_layout_attr(self, name: Span) -> Attribute:
...
         return StridedLayoutAttr(strides, None if offset == "?" else offset)
 
+    def parse_optional_unit_attr(self) -> Attribute | None:
+        """
+        Parse a value of `unit` type.
+        unit-attribute ::= `unit`
+        """
+        if self._current_token.kind != Token.Kind.BARE_IDENT:
+            return None
+        name = self._current_token.span.text
```
AntonLydike, line 662: supernit: Not sure this extra variable assignment is needed here?

AntonLydike (approved):
LGTM, just some small misc. comments

georgebisbas (approved):

tests/filecheck/parser-printer/builtin_attrs.mlir
superlopuh: I agree in general, although here it seemed like a good fit for consistency, happy to change it if you or anyone else insist

xdsl/parser/attribute_parser.py
superlopuh: definitely not, although it doesn't really hurt either

xdsl/parser/attribute_parser.py
```diff
@@ -651,6 +652,22 @@ def _parse_strided_layout_attr(self, name: Span) -> Attribute:
...
+
+        # Unit type
```
PapyChacal, line 664:
```suggestion
        # Unit attribute
```

PapyChacal (approved):
Seeing the low footprint, I'm guessing this preserves the behaviour of unit attribute being printed as "just a name" in attribute dictionaries?

I had no clue they were handled otherwise in other places 😃

math-fehr (approved):
Looks good to me now (I would just fix the comment that Emilien pointed)

PR #2745, opened 2024-06-18
dialects: (csl) Improve some constructor inits

AntonLydike:
Fix some of the things that accidentally went through in #2742.

Also moves the `__init__` to the top of the method definition and adds some docstrings.


dk949 (approved):
Looks very sensible 👍

n-io (approved):
thanks, looks good

PR #2746, opened 2024-06-18
dialects: Add qssa dialect

alexarice:
Add a simple quantum ssa dialect and a simple roundtrip filecheck test. Thought it would be better to split the work I was doing into multiple pull requests to make it easier to review so this doesn't have any rewrites yet

### DIFF
--- a/tests/filecheck/dialects/qssa/ops.mlir
+++ b/tests/filecheck/dialects/qssa/ops.mlir
@@ -21,7 +21,7 @@
 
 // CHECK-NEXT: %0 = qssa.measure %q6
 
-// CHECK-GENERIC: %q0, %q1 = "qssa.alloc"() <{"num_qubits" = 2 : i64}> : () -> (!qssa.qubit, !qssa.qubit)
+// CHECK-GENERIC: %q0, %q1 = "qssa.alloc"() : () -> (!qssa.qubit, !qssa.qubit)
 // CHECK-GENERIC-NEXT: %q2 = "qssa.h"(%q0) : (!qssa.qubit) -> !qssa.qubit
 // CHECK-GENERIC-NEXT: %q3, %q4 = "qssa.cz"(%q1, %q2) : (!qssa.qubit, !qssa.qubit) -> (!qssa.qubit, !qssa.qubit)
 // CHECK-GENERIC-NEXT: %q5, %q6 = "qssa.cnot"(%q3, %q4) : (!qssa.qubit, !qssa.qubit) -> (!qssa.qubit, !qssa.qubit)
--- a/tests/filecheck/dialects/qssa/ops.mlir
+++ b/tests/filecheck/dialects/qssa/ops.mlir
@@ -11,3 +11,4 @@
 
 // CHECK-GENERIC: %q0 = "test.op"() : () -> !qssa.qubits<1>
 // CHECK-GENERIC-NEXT: %0 = "qssa.h"(%q0) : (!qssa.qubits<1>) -> !qssa.qubits<1>
+
--- a/tests/filecheck/dialects/qssa/ops.mlir
+++ b/tests/filecheck/dialects/qssa/ops.mlir
@@ -5,24 +5,24 @@
 
 // CHECK: %q0, %q1 = qssa.alloc<2>
 
-%0 = qssa.h %q0
+%q2 = qssa.h %q0
 
-// CHECK-NEXT: %0 = qssa.h %q0
+// CHECK-NEXT: %q2 = qssa.h %q0
 
-%q2:2 = qssa.cz %q1 %0
+%q3, %q4 = qssa.cz %q1, %q2
 
-// CHECK-NEXT: %q2, %q2_1 = qssa.cz %q1 %0
+// CHECK-NEXT: %q3, %q4 = qssa.cz %q1, %q2
 
-%q3, %q4 = qssa.cnot %q2#0 %q2#1
+%q5, %q6 = qssa.cnot %q3, %q4
 
-// CHECK-NEXT: %q3, %q4 = qssa.cnot %q2 %q2_1
+// CHECK-NEXT: %q5, %q6 = qssa.cnot %q3, %q4
 
-%1 = qssa.measure %q3
+%0 = qssa.measure %q6
 
-// CHECK-NEXT: %1 = qssa.measure %q3
+// CHECK-NEXT: %0 = qssa.measure %q6
 
-// CHECK-GENERIC: %q0, %q1 = "qssa.alloc"() <{"qubits" = 2 : i64}> : () -> (!qssa.qubit, !qssa.qubit)
-// CHECK-GENERIC-NEXT: %0 = "qssa.h"(%q0) : (!qssa.qubit) -> !qssa.qubit
-// CHECK-GENERIC-NEXT: %q2, %q2_1 = "qssa.cz"(%q1, %0) : (!qssa.qubit, !qssa.qubit) -> (!qssa.qubit, !qssa.qubit)
-// CHECK-GENERIC-NEXT: %q3, %q4 = "qssa.cnot"(%q2, %q2_1) : (!qssa.qubit, !qssa.qubit) -> (!qssa.qubit, !qssa.qubit)
-// CHECK-GENERIC-NEXT: %1 = "qssa.measure"(%q3) : (!qssa.qubit) -> i1
+// CHECK-GENERIC: %q0, %q1 = "qssa.alloc"() <{"num_qubits" = 2 : i64}> : () -> (!qssa.qubit, !qssa.qubit)
+// CHECK-GENERIC-NEXT: %q2 = "qssa.h"(%q0) : (!qssa.qubit) -> !qssa.qubit
+// CHECK-GENERIC-NEXT: %q3, %q4 = "qssa.cz"(%q1, %q2) : (!qssa.qubit, !qssa.qubit) -> (!qssa.qubit, !qssa.qubit)
+// CHECK-GENERIC-NEXT: %q5, %q6 = "qssa.cnot"(%q3, %q4) : (!qssa.qubit, !qssa.qubit) -> (!qssa.qubit, !qssa.qubit)
+// CHECK-GENERIC-NEXT: %0 = "qssa.measure"(%q6) : (!qssa.qubit) -> i1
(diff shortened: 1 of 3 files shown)


alexarice:
Am I good to merge or should I ask for more reviews?

AntonLydike:
Maybe @n-io can give a thumbs up as well? Otherwise I don't know who'd care about a `qssa` dialect.

General rule of thumb is that you want to be somewhat sure that anyone interested or invested had the chance to say okay or no. If nobody said anything, that's not enough. Sasha goes along the lines of one check is enough, I usually try to get two or more.

n-io (commented):
Looks good, some minor comments (somehow didn't click submit review, apologies)

xdsl/dialects/qssa.py
```diff
@@ -0,0 +1,142 @@
...
+
+class QubitBase(IRDLOperation, ABC):
+    pass
+
+
+@irdl_op_definition
+class QubitAllocOp(QubitBase):
+    name = "qssa.alloc"
+
+    qubits = prop_def(IntegerAttr)
```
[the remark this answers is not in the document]
AntonLydike: Actually, looking at this again, I don't think we need this attribute at all. The number of qubits is simply the number of results. So the attribute can be dropped entirely

tests/filecheck/dialects/qssa/ops.mlir
```diff
@@ -0,0 +1,28 @@
...
+
+%0 = qssa.h %q0
+
+// CHECK-NEXT: %0 = qssa.h %q0
+
+%q2:2 = qssa.cz %q1 %0
+
+// CHECK-NEXT: %q2, %q2_1 = qssa.cz %q1 %0
+
+%q3, %q4 = qssa.cnot %q2#0 %q2#1
```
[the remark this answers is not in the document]
AntonLydike: Just to make sure, you don't mean the notation of `%q2#0 %q2#1`? Because that's just a really weird alternative way for MLIR to represent multiple return values

tests/filecheck/dialects/qssa/ops.mlir
```diff
@@ -0,0 +1,28 @@
...
+
+%q0, %q1 = qssa.alloc<2>
+
+// CHECK: %q0, %q1 = qssa.alloc<2>
+
+%0 = qssa.h %q0
+
+// CHECK-NEXT: %0 = qssa.h %q0
+
+%q2:2 = qssa.cz %q1 %0
```
AntonLydike, line 12: I think it's customary for custom syntax to have commas between operands, so I'd suggest adding them if you don't have a reason to omit them.

tests/filecheck/dialects/qssa/ops.mlir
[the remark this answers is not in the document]
AntonLydike: `%q2:2 = ...` is the same as saying `%q2_1, %q2_2 = ...`, just that the two ssa vals are accessed with `%q2#0` and `%q2#1` instead.

xdsl/dialects/qssa.py
```diff
@@ -0,0 +1,144 @@
...
+
+    res: VarOpResult = var_result_def(qubit)
```
n-io, line 41: From the semantics, this reads like it allocates a series of single qubits, rather than an array of qubits as the op is described in the paper. You could consider parameterising `QubitAttr` similar to `IntegerAttr`, or maybe @AntonLydike could advise?

tests/filecheck/dialects/qssa/ops.mlir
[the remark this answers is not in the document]
n-io: Yeah that's actually what I was confused about, it looked like an array.

xdsl/dialects/qssa.py
AntonLydike: What I was saying is that these two ops are both completely valid, just one has redundant information:


```
@irdl_op_definition
class QubitAllocOp(QubitBase):
    name = "qssa.alloc"

    qubits = prop_def(IntegerAttr)

    res: VarOpResult = var_result_def(qubit)
```

and 
```
@irdl_op_definition
class QubitAllocOp(QubitBase):
    name = "qssa.alloc"

    res: VarOpResult = var_result_def(qubit)
```

Simply because we know that `qbits == len(res)`. Therefore we don't need to carry this additional information at all. The resulting IR would look like this:


```
%qbit1, %qbit2 = qssa.alloc
%qbit3 = qssa.alloc
```

So you just specify how many qbits you allocate by the number of results.

n-io: For reference: 3.3.1 Allocation. `qssa.alloc` allocates an array of qubits of fixed or variable length. `%q = qssa.alloc : qubit<10>`

AntonLydike: I think @alexarice does not talk about qbit arrays, but models each qbit as a single SSA value. But that's for him to answer :smile:

n-io: I seem to remember that semantically, an array of qubits may be understood to mean something different as compared to n single qubits, but I may be mistaken here.

xdsl/dialects/qssa.py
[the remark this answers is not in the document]
n-io: I agree. I'd generally encourage to print some type information in the mlir, though.

xdsl/dialects/qssa.py
alexarice: This is not directly based on the qssa dialect from the paper and is currently representing each individual qubit as a separate ssa value. I'd be open to renaming this dialect if this will cause confusion in the future

alexarice: The rest of the discussion didn't load for me on github, but I believe dropping the parameter is not ideal as the parser does not have access to the number of results (at least from my testing)

xdsl/dialects/qssa.py
```diff
@@ -0,0 +1,144 @@
...
+        return QubitAllocOp.create(
+            operands=(),
+            result_types=tuple(qubit for _ in range(0, num_qubits)),
+            properties={"num_qubits": IntegerAttr(num_qubits, 64)},
+            attributes=attr_dict or {},
+        )
+
+    def print(self, printer: Printer):
+        with printer.in_angle_brackets():
+            printer.print(self.num_qubits.value.data)
```
n-io, line 64:
```suggestion
            printer.print(len(self.res))
```
One approach would be to still print the number (since you're not printing a type) but not to store it on the op as a property.

xdsl/dialects/qssa.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @classmethod
+    def parse(cls, parser: Parser) -> "QubitAllocOp":
+        with parser.in_angle_brackets():
+            num_qubits = parser.parse_integer()
+        attr_dict = parser.parse_optional_attr_dict()
+        return QubitAllocOp.create(
+            operands=(),
+            result_types=tuple(qubit for _ in range(0, num_qubits)),
+            properties={"num_qubits": IntegerAttr(num_qubits, 64)},
```
n-io, line 58:
```suggestion
```
2/4
still parsing, just not storing

xdsl/dialects/qssa.py
```diff
@@ -0,0 +1,144 @@
...
+
+    num_qubits = prop_def(IntegerAttr)
+
+    res: VarOpResult = var_result_def(qubit)
+
+    def __init__(self, num_qubits: int):
+        super().__init__(
+            operands=(),
+            result_types=([qubit for _ in range(0, num_qubits)],),
+            properties={"num_qubits": IntegerAttr(num_qubits, 64)},
```
n-io, line 47:
```suggestion
```
3/4

xdsl/dialects/qssa.py
```diff
@@ -0,0 +1,144 @@
...
+
+class QubitBase(IRDLOperation, ABC):
+    pass
+
+
+@irdl_op_definition
+class QubitAllocOp(QubitBase):
+    name = "qssa.alloc"
+
+    num_qubits = prop_def(IntegerAttr)
```
n-io, line 39:
```suggestion
```
4/4

tests/filecheck/dialects/qssa/ops.mlir
```diff
@@ -0,0 +1,28 @@
...
+
+%q5, %q6 = qssa.cnot %q3, %q4
+
+// CHECK-NEXT: %q5, %q6 = qssa.cnot %q3, %q4
+
+%0 = qssa.measure %q6
+
+// CHECK-NEXT: %0 = qssa.measure %q6
+
+// CHECK-GENERIC: %q0, %q1 = "qssa.alloc"() <{"num_qubits" = 2 : i64}> : () -> (!qssa.qubit, !qssa.qubit)
```
n-io, line 24:
```suggestion
// CHECK-GENERIC: %q0, %q1 = "qssa.alloc"() : () -> (!qssa.qubit, !qssa.qubit)
```
5/4 (don't mind my counting)

n-io (changes requested):
Some suggested changes on avoiding storing the number of qubits explicitly.

alexarice:
> Suggestions on how to avoid storing the number of qubits explicitly.

Is there a reason this is desirable? It feels strange to me to have a disconnect between the printed data and the data stored on the operation

AntonLydike:
I would recommend to not print the count at all. Just rely on people being able to count the number of results?

AntonLydike:
> > Suggestions on how to avoid storing the number of qubits explicitly.
> 
> Is there a reason this is desirable? It feels strange to me to have a disconnect between the printed data and the data stored on the operation

I generally agree, I think printing derived (redundant) data is not the best idea... I'd just not print the count at all

alexarice:
> I would recommend to not print the count at all. Just rely on people being able to count the number of results?

you then can't parse the operation though

AntonLydike:
> > I would recommend to not print the count at all. Just rely on people being able to count the number of results?
> 
> you then can't parse the operation though

I think you should be able to. I'll come over

AntonLydike:
@alexarice and I looked into the parser internals and spoke with Markus on how MLIR handles it, and sadly we have to print redundant information here (the number of results), as MLIR allows one to write `qssa.alloc<4>` without specifying the results, if they are not used. This is imo not great, and makes me wrong in saying "I think you should be able to. I'll come over"

n-io (commented):
Looks good :)

tests/filecheck/dialects/qssa/ops.mlir
n-io: does this actually work now, @AntonLydike ?

n-io: hang on, it's been removed

n-io (approved):

PR #2748, opened 2024-06-18
dialects: Add memref.extract_strided_metadata op

JosseVanDelm:
I'm adding the memref.extract_strided_metadataop because I need an op to fetch the dynamic offset created by a subview, and otherwise I can not do  that. 
https://mlir.llvm.org/docs/Dialects/MemRef/#memrefextract_strided_metadata-memrefextractstridedmetadataop

### DIFF
--- a/xdsl/dialects/memref.py
+++ b/xdsl/dialects/memref.py
@@ -513,6 +513,10 @@ class ExpandShapeOp(AlterShapeOp):
 
 @irdl_op_definition
 class ExtractStridedMetaDataOp(IRDLOperation):
+    """
+    https://mlir.llvm.org/docs/Dialects/MemRef/#memrefextract_strided_metadata-memrefextractstridedmetadataop
+    """
+
     name = "memref.extract_strided_metadata"
 
     source: Operand = operand_def(MemRefType)
@@ -528,6 +532,10 @@ class ExtractStridedMetaDataOp(IRDLOperation):
     def from_source_memref(
         source: SSAValue | Operation, source_type: MemRefType[Attribute]
     ):
+        """
+        Create an ExtractStridedMetaDataOp that extracts the metadata from the
+        operation (source) that produces a memref.
+        """
         source_shape = source_type.get_shape()
         # Return a rank zero memref with the memref type
         base_buffer_type = MemRefType(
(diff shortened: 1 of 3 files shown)


xdsl/dialects/memref.py
```diff
@@ -509,6 +511,40 @@ class ExpandShapeOp(AlterShapeOp):
...
+    offset: OpResult = result_def(IndexType)
+    sizes: VarOpResult = var_result_def(IndexType)
+    strides: VarOpResult = var_result_def(IndexType)
+
+    irdl_options = [AttrSizedResultSegments()]
+
+    @staticmethod
+    def from_source_memref(
+        source: SSAValue | Operation, source_type: MemRefType[Attribute]
+    ):
```
AntonLydike, line 530: Can you make this an `__init__`? That's what we are trying to move towards in xDSL

xdsl/dialects/memref.py
```diff
@@ -509,6 +511,40 @@ class ExpandShapeOp(AlterShapeOp):
...
+            source_type.memory_space,
+        )
+        offset_type = IndexType()
+        strides_type = [IndexType() for _ in source_shape]
+        sizes_type = [IndexType() for _ in source_shape]
+        return_type = [base_buffer_type, offset_type, strides_type, sizes_type]
+        return ExtractStridedMetaDataOp.build(
+            operands=[source], result_types=[*return_type]
+        )
+
```
AntonLydike, line 546: In the init you can replace this with `super().__init__(operands=[source], result_types=[*return_type])`

AntonLydike: Also `[*return_type]` is the same as `return_type`

AntonLydike (approved):
Nice work, just some minor stylistic comments! Structure looks solid!

xdsl/dialects/memref.py
```diff
@@ -509,6 +511,40 @@ class ExpandShapeOp(AlterShapeOp):
...
+        # Return a rank zero memref with the memref type
+        base_buffer_type = MemRefType(
+            source_type.element_type,
+            [],
+            NoneAttr(),
+            source_type.memory_space,
+        )
+        offset_type = IndexType()
+        strides_type = [IndexType() for _ in source_shape]
+        sizes_type = [IndexType() for _ in source_shape]
```
AntonLydike, line 541: You could write this as `[IndexType()] * len(source_shape)`, but I don't know if you want to^^

xdsl/dialects/memref.py
JosseVanDelm: Is this an __init__ though? creating a bare op of this kind should not really include a source type, should it?

PapyChacal: I remember being torn by the very same thought on subview itself;
I propose you just do what's still done there, and @AntonLydike could file an issue or Zulip thread to discuss it independently?

AntonLydike: Hmmm, I don't know if I can follow. How else would you create this op if not from a source memref?

AntonLydike: Maybe you could clear this up by adding a docstring to the method saying exactly what it's for and what each operand does?

PapyChacal (approved):
LGTM, thanks!

xdsl/dialects/memref.py
n-io: May I ask why `source` and `source_type` are passed in separately? I'd expect `source` to come with a type

PR #2749, opened 2024-06-18
core: Move MLContext out of ir, move get_all_dialects to dialects.

PapyChacal:
Just some refactoring to decouple a few things.


AntonLydike (commented):
Nice to see this refactored!

AntonLydike (approved):
oops

georgebisbas (approved):

xdsl/dialects/__init__.py
```diff
@@ -0,0 +1,310 @@
+from collections.abc import Callable
```
math-fehr, line 1: Oh, I would not have thought about adding this to the `__init__.py` directly, but that's a really good idea!

xdsl/context.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    def register_all_dialects(self):
+        """
+        Register all dialects that can be used.
+
+        Add other/additional dialects by overloading this function.
+        """
+        for dialect_name, dialect_factory in get_all_dialects().items():
+            self.register_dialect(dialect_name, dialect_factory)
```
math-fehr, line 35: I am not entirely sure that anyone should inherit `MLContext`, so I would drop this comment.
Which is also why I'm not sure we should add this method here? Maybe we should just have `register_dialects` to make this simpler and just have `context.register_dialects(get_all_dialects())`? Or keep what we had before.

At least, I would prefer keep this a conscious decision for the user to add all dialects or not, and for that asking them to call `get_all_dialects` make sense to me if that makes sense? I don't know, what do you think?

xdsl/tools/command_line_tool.py
```diff
@@ -726,15 +419,6 @@ def get_input_stream(self) -> tuple[IO[str], str]:
...
 
-    def register_all_dialects(self):
-        """
-        Register all dialects that can be used.
-
-        Add other/additional dialects by overloading this function.
-        """
-        for dialect_name, dialect_factory in get_all_dialects().items():
-            self.ctx.register_dialect(dialect_name, dialect_factory)
-
```
math-fehr, line 737: I would still keep this here actually, and remove it from `MLContext`, what do you think?
This class is supposed to be overloaded for each tool, so adding the dialects here would make sense to me. While `MLContext` is not.

math-fehr (commented):
Really nice!
I would just have a question relative to `MLContext.register_all_dialects`.
To me, this is a bit weird, as we do not really want users to overload `MLContext` I think? Or at least this is not how I was currently seeing it, and should change my view.

But to me, having the `register_all_dialects` in the command line tool made sense as it was a tool that was meant to be inherited for each use case. However, for the `MLContext`, I don't see why we would overload this for each tool besides to add these dialects (which we can add in another way?).

But maybe I'm just missing a use case you have on this?

Otherwise, thanks a lot for moving the `get_all_dialects`!!!!

xdsl/tools/command_line_tool.py
PapyChacal: Ah, I missed the comment stating that people should override it.

Problem is, to parse and interpret irdl dialects, I need the dependent (i.e., at least irdl?) dialexts to be available.
Our lazy loading mechanism is perfect for this, as it will only load actually needed dialects.

Now, without moving register_all_dialects, I would have to create an xDSLOptMain for this purpose, which can be wanted completely out of a tool's context.

I guess a compromise would be to keep a `register_all_dialects` method on xDSLOptMain, of which the default implementation is to just call MLContext's one? WDYT?

xdsl/context.py
superlopuh: I think this gets especially weird for users of xDSL that want to add their own dialects, it seems a little cleaner to let them do their own combinations of different dialects rather than strongly encouraging inheritance in this way.

xdsl/tools/command_line_tool.py
superlopuh: I'm not 100% sure what the answer is to this but I would propose changing the mechanism in a different PR

superlopuh (approved):

PR #2751, opened 2024-06-18
installation: Add ruff to flake.nix

alexarice:
The python installation of ruff doesn't work on nixos, so it'd be good to add it to the flake


superlopuh (approved):
I assume you've tested this, LGTM if that's the case

PR #2752, opened 2024-06-18
backend: (riscv) Add prologue epilogue insertion

zero9178:
The RISC-V call ABI requires a few registers, if clobbered, to be restored by the callee prior to the function returning. This has so far not been implemented meaning that any C compiler calling an xDSL generated function may cause undefined behaviour.

This PR adds a pass performing the required register saving and restoring by pushing and popping them from the stack. The registers are saved in the prologue at the beginning of the function and an epilogue restoring them is emitted before every return operation.

The pass is required to run after register allocation to see the clobbering of callee-preserved registers. It itself does not require register allocation to be run afterward.


tests/interactive/test_get_all_available_passes.py
```diff
@@ -30,21 +30,15 @@ def test_get_all_available_passes():
...
-
     res = get_available_pass_list(
         input_text,
         pass_pipeline,
         condense_mode=True,
         rewrite_by_names_dict=individual_rewrite.REWRITE_BY_NAMES,
     )
 
-    assert res == expected_res
+    assert AvailablePass(
```
zero9178, line 40: Note: This test is very fragile and me just adding a new pass changed the result. 
I am not sure what the original intent of the test is however

superlopuh: I need to change this test it's getting a little annoying. I'd say just update the test with the new expected result for now, I'm not sure we want to change the structure of the test.

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -0,0 +1,92 @@
...
+            IntRegisterType | FloatRegisterType, None
+        ]()
+        for op in func.walk():
+            # Special cases ops that do not actually write to the register.
+            if isinstance(op, riscv.GetRegisterOp | riscv.GetFloatRegisterOp):
+                continue
+
+            for res in op.results:
+                if res.type in Registers.S or res.type in Registers.FS:
+                    used_callee_preserved_registers[res.type] = None
```
superlopuh, line 41:
```suggestion
        used_callee_preserved_registers = {
            res.type
            for op in func.walk()
            if not isinstance(op, riscv.GetRegisterOp | riscv.GetFloatRegisterOp)
            for res in op.results
            if res.type in Registers.S or res.type in Registers.FS
        }
```
Might as well use a set here instead of a dict to None, no?

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Note: This assumes RV32D aka 4 byte integer registers, 8 byte FP registers.
+        def get_register_size(r: RISCVRegisterType):
+            if isinstance(r, IntRegisterType):
+                return 4
+            return 8
```
superlopuh, line 47: Could you please pass these as xlen and flen? Ideally we'd want these as parameters on the pass, and to have a way to share the logic between passes.

superlopuh: By that I mean that in the future, we would find a way to pass xlen and flen only once ƒor all the passes, but for now it would be good to at least state them explicitly

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -0,0 +1,92 @@
...
+                reg_op = builder.insert(riscv.GetRegisterOp(reg))
+                op = riscv.SwOp(rs1=sp_register, rs2=reg_op, immediate=offset)
+            else:
+                reg_op = builder.insert(riscv.GetFloatRegisterOp(reg))
+                op = riscv.FSdOp(rs1=sp_register, rs2=reg_op, immediate=offset)
+
+            builder.insert(op)
+            offset += get_register_size(reg)
+
+        # Now build the epilogue right before every return operation.
```
superlopuh, line 66: Is there not a requirement of a single exit block?

superlopuh (commented):
Do MLIR passes guarantee idempotency? For some of the things I'd like to do, like automatic pipeline construction, idempotency is a very nice feature to have. I wonder if we could structure the stack handling in our backend somehow to guarantee that if we run a pass twice we get the same result. One obvious way might be to add an attribute to denote a function that's been processed already, that would be skipped if processed again, but maybe in a more integrated way with the register allocator? Happy to brainstorm.

xdsl/backend/riscv/prologue_epilogue_insertion.py
zero9178: This would make the compiler output not-deterministic as sets have unspecified iteratior order

xdsl/backend/riscv/prologue_epilogue_insertion.py
zero9178: There is not, see the tests + any of the kernels they generate. They contain multiple blocks and each block may have their own return operation. (You often even want this for hot-cold basic block splitting).
Given how trivial it is support this I don't see a reason not to support it.

xdsl/backend/riscv/prologue_epilogue_insertion.py
AntonLydike: Then you'd need to make sure to sort it before you use it otherwise you get funky stuff :ghost:

zero9178:
> Do MLIR passes guarantee idempotency? 

They do not no.

> One obvious way might be to add an attribute to denote a function that's been processed already, that would be skipped if processed again, but maybe in a more integrated way with the register allocator? Happy to brainstorm.

We can do that if you want. An attribute on the function makes sense to me if you want to track a phase in that sense. Do you want to see this already as part of this PR?

I am not sure how this pass requires any integration with the register allocator in any way. It only needs to see its output basically. This will change if we ever start spilling memory to stack as we'll probably want a meta-op representing a not-yet materialized stack allocation that this pass would materialize, but that is the only thing I can think of. That is future talk either way.

xdsl/backend/riscv/prologue_epilogue_insertion.py
zero9178: Sorted did not work immediately as there is no comparison operator between `IntRegisterType` and `FloatRegisterType`. I applied your changes but kept it as a dict for now. I can add that comparison operator (or can this be auto generated by some annotator?) if you feel strongly about this

xdsl/backend/riscv/prologue_epilogue_insertion.py
zero9178: I made these a pass option now as that was actually really nice and easy

tests/filecheck/backend/riscv/prologue_epilogue_insertion.mlir
```diff
@@ -0,0 +1,52 @@
...
+  // CHECK-NEXT: lw %{{.*}}, 8
+  // CHECK-SAME: (!riscv.reg<sp>) -> !riscv.freg<s5>
+  // CHECK-NEXT: addi %{{.*}}, 12
+  // CHECK-SAME: (!riscv.reg<sp>) -> !riscv.freg<sp>
+  riscv_func.return
+}
+
+// SMALL_FLEN: func @main
+// SMALL_FLEN: addi %{{.*}}, -8
+// SMALL_FLEN-SAME: (!riscv.reg<sp>) -> !riscv.reg<sp>
```
superlopuh, line 52:
```suggestion
// SMALL_FLEN-SAME: (!riscv.reg<sp>) -> !riscv.reg<sp>

```

tests/filecheck/backend/riscv/prologue_epilogue_insertion.mlir
```diff
@@ -0,0 +1,52 @@
+// RUN: xdsl-opt --split-input-file -p "riscv-prologue-epilogue-insertion" %s | filecheck %s
+// RUN: xdsl-opt --split-input-file -p "riscv-prologue-epilogue-insertion{flen=4}" %s | filecheck %s --check-prefix=SMALL_FLEN
```
superlopuh, line 2: nit:
```suggestion
// RUN: xdsl-opt --split-input-file -p "riscv-prologue-epilogue-insertion{flen=4}" %s | filecheck %s --check-prefix=CHECK-SMALL-FLEN
```

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -0,0 +1,92 @@
+from dataclasses import dataclass, field
+
+from xdsl.builder import Builder
+from xdsl.dialects import builtin, riscv, riscv_func
+from xdsl.dialects.riscv import (
+    IntRegisterType,
+    Registers,
+    RISCVRegisterType,
+)
+from xdsl.ir import MLContext
```
superlopuh, line 10:
```suggestion
from xdsl.context import MLContext
```

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -0,0 +1,92 @@
...
+    def _process_function(self, func: riscv_func.FuncOp) -> None:
+        # Find all callee-preserved registers that are clobbered. We define clobbered
+        # as it being the result of some operation and therefore written to.
+        used_callee_preserved_registers = {
+            res.type: None
+            for op in func.walk()
+            if not isinstance(op, riscv.GetRegisterOp | riscv.GetFloatRegisterOp)
+            for res in op.results
+            if res.type in Registers.S or res.type in Registers.FS
+        }
```
superlopuh, line 41: Hopefully this will give the same behaviour:
```suggestion
        used_callee_preserved_registers = orderedset(
            res.type: None
            for op in func.walk()
            if not isinstance(op, riscv.GetRegisterOp | riscv.GetFloatRegisterOp)
            for res in op.results
            if res.type in Registers.S or res.type in Registers.FS
        )
```

xdsl/backend/riscv/prologue_epilogue_insertion.py
superlopuh: Yep sounds good, thanks for the clarification

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -0,0 +1,94 @@
...
+        # as it being the result of some operation and therefore written to.
+        used_callee_preserved_registers = OrderedSet(
+            res.type
+            for op in func.walk()
+            if not isinstance(op, riscv.GetRegisterOp | riscv.GetFloatRegisterOp)
+            for res in op.results
+            if res.type in Registers.S or res.type in Registers.FS
+        )
+
+        # Note: This assumes RV32D aka 4 byte integer registers, 8 byte FP registers.
```
superlopuh, line 45:
```suggestion
```

superlopuh (approved):
LGTM

PR #2753, opened 2024-06-19
core: Move register_all_dialects to MLContext

PapyChacal:
Just make it the default implementation of CLI's hook.


xdsl/context.py
```diff
@@ -60,6 +62,13 @@ def registered_dialect_names(self) -> Iterable[str]:
         """
         return self._registered_dialects.keys()
 
+    def register_all_dialects(self):
+        """
+        Register all listed dialects.
+        """
+        for dialect_name, dialect_factory in get_all_dialects().items():
+            self.register_dialect(dialect_name, dialect_factory)
+
```
superlopuh, line 71: I wonder if there's another place we could put this logic. I see your point about the command-line tool not being the best place for it if you want to use it for non command-line purposes. I'm also not sure that we want clients of xDSL to subclass MLContext to add their own dialect hooks. I would actually propose splitting this out somewhere else, potentially, or just asking people to copy/paste this two-line snippet for their own use-case until we find a better way to structure it.

PapyChacal: Did you see that register_all_dialects is still defined on CommandLineTool and simply reusing that logic here?
AKA users can just subclass CommandLineTool and use it as before? I'm really confused here by how it says a different story to CommandLineTool users

superlopuh: Sure, command-line tool is fine, I'm just saying for other use-cases where a client might want to add dialects for things to work, now they'd have to do it in two places, and I think that subclassing MLContext to add it here also is a little weird.

superlopuh: I have some suggestions about this, but it might be better to do a short call before I outline them here? I think it would be good to have a bit more of an insight into your short-term plans on top of this. The main thing to consider is what tools like `xdsl-gui` should do to integrate new dialects in dependencies.

PR #2756, opened 2024-06-20
transforms: stencil-tensorize-z-dim accept stencil.field as function arg

n-io:
A minor fix (plus test case) that allows the transformation to work on `stencil.field` passed in as function arguments.


PapyChacal (approved):
Yay :grinning:

n-io:
> Yay 😀

As you suggested, this is actually a lot neater (thanks!), but best to allow both since we already support the other.

PR #2758, opened 2024-06-20
transformations: do not insert affine.apply ops when streaming

superlopuh:
We encode the affine map constraints in the streaming region setup, so we don't need the results of the affine.apply operation, unlike in the memref.load case. We could dce them away, but it feels better to not insert them in the first place.


PapyChacal (approved):
I'm a big fan of "just DCE it later" myself when it just works, but LGTM!

NB: You can also just manually reuse DCE *from your pass*, i.e., don't complexity your pattern and still get your pass outputting simpler IR. Just mentioning the idea, again, this LGTM :slightly_smiling_face:

superlopuh:
I completely agree, hence this PR :)

PR #2760, opened 2024-06-20
transformations: fix yielding of values in memref_stream.generic lowering

superlopuh:


PapyChacal (approved):
I'm not sure what this fixes exactly, but what I see looks good!

superlopuh:
The values yielded in the loop body were the results of the loop itself 💥

PR #2763, opened 2024-06-21
dialects: (memref_stream) Add an inits field to memref_stream.generic [1/3]

superlopuh:
This is PR 1/3 that introduces constant initialisation to our kernel representations. This PR adds the new attribute, and adds NotImplemented to the streamification and lowering to loops, which are handled in the next two pull requests.

Note stacked PR.

### DIFF
--- a/tests/filecheck/dialects/memref_stream/ops.mlir
+++ b/tests/filecheck/dialects/memref_stream/ops.mlir
@@ -50,18 +50,19 @@ memref_stream.generic {
         affine_map<(d0, d1) -> (d1)>,
         affine_map<(d0, d1) -> (d0, d1)>
     ],
-    iterator_types = ["parallel", "parallel"]
+    iterator_types = ["parallel", "parallel"],
+    inits = [unit]
 } ins(%A, %B : memref<2xf32>, memref<3xf32>) outs(%C : memref<3x2xf64>) attrs = {hello = "world"} {
 ^bb0(%arg3: f32, %arg4: f32):
     memref_stream.yield %arg3 : f32
 }
 
-// CHECK-NEXT:          memref_stream.generic {bounds = [#builtin.int<3>, #builtin.int<2>], indexing_maps = [affine_map<(d0, d1) -> (d0)>, affine_map<(d0, d1) -> (d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = ["parallel", "parallel"]} ins(%A, %B : memref<2xf32>, memref<3xf32>) outs(%C : memref<3x2xf64>) attrs =  {"hello" = "world"} {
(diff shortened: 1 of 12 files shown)


xdsl/dialects/memref_stream.py
```diff
@@ -310,6 +319,12 @@ class GenericOp(IRDLOperation):
     pattern defines the order in which the elements of the input buffers will be written
     to.
     """
+    inits = prop_def(ArrayAttr[AnyFloatAttr | AnyIntegerAttr | UnitAttr])
+    """
+    Initial values for outputs. If `NoneAttr`, then the value is read from the output
+    buffer. Otherwise, the value is created at runtime with an `arith.constant` operation
+    during lowering. The inits may be set only for the imperfectly nested form.
```
PapyChacal, line 326: The comment mentions NoneAttr but the prop uses UnitAttr!

I skimmed your unit PR and I'm not sure I get it; why not NoneAttr? The name sounds more like the use here?

PapyChacal (commented):
Just a question about unit, otherwise this PR is a bit too complex for me and my knowledge of the streaming situation to review! Just tell me if there are specific aspects you'd like my review

xdsl/dialects/memref_stream.py
superlopuh: Our printing and parsing of it is somewhat broken, NoneAttr doesn't roundtrip in the IR, as `#none` is interpreted as an alias, and `none` is NoneType, which seems like the wrong thing to use. This is one bit I was particularly hoping for feedback/ideas of how to do better. I could add custom parsing and printing for this, and spell the missing values as `_` or add them somewhere else in the custom syntax, also. Recommendations welcome.

tests/filecheck/dialects/memref_stream/ops.mlir
```diff
@@ -83,10 +83,80 @@ memref_stream.generic {
...
+        affine_map<(d0, d1, d2) -> (d2, d1)>,
+        affine_map<(d0, d1) -> (d0, d1)>
+    ],
+    iterator_types = ["parallel", "parallel", "reduction"]
+} ins(%E, %F : memref<4x2xf64>, memref<2x3xf64>) outs(%G : memref<4x3xf64>) inits(%H : f64) {
+^0(%e : f64, %f : f64, %acc_old : f64):
+    %prod = arith.mulf %e, %f : f64
+    %acc_new = arith.addf %acc_old, %prod : f64
+    linalg.yield %acc_new : f64
+}
```
AntonLydike, line 108: nit: These filechecks don't actually test the inits added in this PR.

AntonLydike (approved):
LGTM overall, just a small nitpick

tests/filecheck/dialects/memref_stream/ops.mlir
superlopuh: This filecheck is just to check that we roundtrip correctly with printing and parsing of custom syntax, so I added a couple of ops with inits with them, there's no other functionality intended to test here

PR #2747, opened 2024-06-18
(dialects and transforms): csl stencil lowering

n-io:
This PR proposes a `csl_stencil` dialect that manages stencil accesses across the stencil pattern, including communication between PEs, with the goal of preparing the stencil dialect for lowering to CSL.

PR #2766, opened 2024-06-21
(dialect+transform): stencil to new csl_stencil dialect and transform

n-io:
Introduces an intermediate dialect with ops:
* `csl_stencil.prefetch` to indicate prefetched buffer transfers
* `csl_stencil.access` performs a `stencil.access` to a prefetched buffer

The `stencil-to-csl-stencil` transform:
* lowers `dmp.swap` to `csl_stencil.prefetch`
* adds prefetched buffers to signature of `stencil.apply`
* lowers `stencil.access` to `csl_stencil.access` iff they are accesses to prefetched buffers

For a more detailed description see the document in #2747. This PR implements Step 1 outlined there.

### DIFF
--- a/tests/filecheck/dialects/csl/stencil-to-csl-stencil.mlir
+++ b/tests/filecheck/dialects/csl/stencil-to-csl-stencil.mlir
@@ -1,0 +1,63 @@
+// RUN: xdsl-opt %s -p "stencil-to-csl-stencil" | filecheck %s
+
+builtin.module {
+  func.func @gauss_seidel(%a : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>, %b : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>) {
+    %0 = stencil.load %a : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>> -> !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+    "dmp.swap"(%0) {"topo" = #dmp.topo<1022x510>, "swaps" = [#dmp.exchange<at [1, 0, 0] size [1, 1, 510] source offset [-1, 0, 0] to [1, 0, 0]>, #dmp.exchange<at [-1, 0, 0] size [1, 1, 510] source offset [1, 0, 0] to [-1, 0, 0]>, #dmp.exchange<at [0, 1, 0] size [1, 1, 510] source offset [0, -1, 0] to [0, 1, 0]>, #dmp.exchange<at [0, -1, 0] size [1, 1, 510] source offset [0, 1, 0] to [0, -1, 0]>]} : (!stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>) -> ()
+    %1 = stencil.apply(%2 = %0 : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>) -> (!stencil.temp<[0,1]x[0,1]xtensor<510xf32>>) {
+      %3 = arith.constant 1.666600e-01 : f32
+      %4 = stencil.access %2[1, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %5 = "tensor.extract_slice"(%4) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+      %6 = stencil.access %2[-1, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %7 = "tensor.extract_slice"(%6) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+      %8 = stencil.access %2[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %9 = "tensor.extract_slice"(%8) <{"static_offsets" = array<i64: 1>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+      %10 = stencil.access %2[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %11 = "tensor.extract_slice"(%10) <{"static_offsets" = array<i64: -1>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+      %12 = stencil.access %2[0, 1] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %13 = "tensor.extract_slice"(%12) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+      %14 = stencil.access %2[0, -1] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %15 = "tensor.extract_slice"(%14) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+      %16 = arith.addf %15, %13 : tensor<510xf32>
+      %17 = arith.addf %16, %11 : tensor<510xf32>
+      %18 = arith.addf %17, %9 : tensor<510xf32>
+      %19 = arith.addf %18, %7 : tensor<510xf32>
+      %20 = arith.addf %19, %5 : tensor<510xf32>
+      %21 = tensor.empty() : tensor<510xf32>
+      %22 = linalg.fill ins(%3 : f32) outs(%21 : tensor<510xf32>) -> tensor<510xf32>
+      %23 = arith.mulf %20, %22 : tensor<510xf32>
+      stencil.return %23 : tensor<510xf32>
+    }
+    stencil.store %1 to %b ([0, 0] : [1, 1]) : !stencil.temp<[0,1]x[0,1]xtensor<510xf32>> to !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>
+    func.return
+  }
+}
+
+// CHECK-NEXT: builtin.module {
+// CHECK-NEXT:   func.func @gauss_seidel(%a : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>, %b : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>) {
+// CHECK-NEXT:     %0 = stencil.load %a : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>> -> !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+// CHECK-NEXT:     %1 = "csl_stencil.prefetch"(%0) <{"size" = 510 : i64, "topo" = #dmp.topo<1022x510>, "swaps" = [#csl_stencil.exchange<to [1, 0]>, #csl_stencil.exchange<to [-1, 0]>, #csl_stencil.exchange<to [0, 1]>, #csl_stencil.exchange<to [0, -1]>]}> : (!stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>) -> memref<4xtensor<510xf32>>
+// CHECK-NEXT:     %2 = stencil.apply(%3 = %0 : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>, %4 = %1 : memref<4xtensor<510xf32>>) -> (!stencil.temp<[0,1]x[0,1]xtensor<510xf32>>) {
+// CHECK-NEXT:       %5 = arith.constant 1.666600e-01 : f32
+// CHECK-NEXT:       %6 = csl_stencil.access %4[1, 0] : memref<4xtensor<510xf32>>
+// CHECK-NEXT:       %7 = csl_stencil.access %4[-1, 0] : memref<4xtensor<510xf32>>
+// CHECK-NEXT:       %8 = stencil.access %3[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+// CHECK-NEXT:       %9 = "tensor.extract_slice"(%8) <{"static_offsets" = array<i64: 1>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+// CHECK-NEXT:       %10 = stencil.access %3[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+// CHECK-NEXT:       %11 = "tensor.extract_slice"(%10) <{"static_offsets" = array<i64: -1>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
+// CHECK-NEXT:       %12 = csl_stencil.access %4[0, 1] : memref<4xtensor<510xf32>>
+// CHECK-NEXT:       %13 = csl_stencil.access %4[0, -1] : memref<4xtensor<510xf32>>
+// CHECK-NEXT:       %14 = arith.addf %13, %12 : tensor<510xf32>
+// CHECK-NEXT:       %15 = arith.addf %14, %11 : tensor<510xf32>
+// CHECK-NEXT:       %16 = arith.addf %15, %9 : tensor<510xf32>
+// CHECK-NEXT:       %17 = arith.addf %16, %7 : tensor<510xf32>
+// CHECK-NEXT:       %18 = arith.addf %17, %6 : tensor<510xf32>
+// CHECK-NEXT:       %19 = tensor.empty() : tensor<510xf32>
+// CHECK-NEXT:       %20 = linalg.fill ins(%5 : f32) outs(%19 : tensor<510xf32>) -> tensor<510xf32>
+// CHECK-NEXT:       %21 = arith.mulf %18, %20 : tensor<510xf32>
+// CHECK-NEXT:       stencil.return %21 : tensor<510xf32>
+// CHECK-NEXT:     }
+// CHECK-NEXT:     stencil.store %2 to %b ([0, 0] : [1, 1]) : !stencil.temp<[0,1]x[0,1]xtensor<510xf32>> to !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>
+// CHECK-NEXT:     func.return
+// CHECK-NEXT:   }
+// CHECK-NEXT: }
--- a/xdsl/dialects/csl_/csl_stencil.py
+++ b/xdsl/dialects/csl_/csl_stencil.py
@@ -1,0 +1,270 @@
+from collections.abc import Sequence
+from itertools import pairwise
+from typing import Annotated, cast
+
+from xdsl.dialects import builtin, memref, stencil
+from xdsl.dialects.builtin import IntegerAttr, IntegerType, TensorType
+from xdsl.dialects.experimental import dmp
+from xdsl.ir import Attribute, Dialect, Operation, ParametrizedAttribute, SSAValue
+from xdsl.irdl import (
+    ConstraintVar,
+    IRDLOperation,
+    Operand,
+    ParameterDef,
+    attr_def,
+    irdl_attr_definition,
+    irdl_op_definition,
+    operand_def,
+    opt_attr_def,
+    opt_prop_def,
+    prop_def,
+    result_def,
+)
+from xdsl.parser import AttrParser, Parser
+from xdsl.printer import Printer
+from xdsl.traits import HasAncestor, Pure
+from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
+
+
+@irdl_attr_definition
+class ExchangeDeclarationAttr(ParametrizedAttribute):
+    """
+    A simplified version of dmp.exchange, from which it should be lowered
+    """
+
+    name = "csl_stencil.exchange"
+
+    neighbor_: ParameterDef[builtin.DenseArrayBase]
+
+    def __init__(
+        self,
+        neighbor: Sequence[int],
+    ):
+        data_type = builtin.i64
+        super().__init__(
+            [
+                builtin.DenseArrayBase.from_list(data_type, neighbor),
+            ]
+        )
+
+    @classmethod
+    def from_dmp_exch_decl_attr(cls, src: dmp.ExchangeDeclarationAttr):
+        return cls(src.neighbor)
+
+    @property
+    def neighbor(self) -> tuple[int, ...]:
+        data = self.neighbor_.as_tuple()
+        assert isa(data, tuple[int, ...])
+        return data
+
+    def print_parameters(self, printer: Printer) -> None:
+        printer.print_string(f"<to {list(self.neighbor)}>")
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
+        parser.parse_characters("<")
+        parser.parse_characters("to")
+        to = parser.parse_comma_separated_list(
+            parser.Delimiter.SQUARE, parser.parse_integer
+        )
+        parser.parse_characters(">")
+
+        return [builtin.DenseArrayBase.from_list(builtin.i64, to)]
+
+
+@irdl_op_definition
+class PrefetchOp(IRDLOperation):
+    """
+    An op to indicate buffer prefetches.
+    """
+
+    name = "csl_stencil.prefetch"
+
+    input_stencil: Operand = operand_def(
+        stencil.TempType[Attribute] | memref.MemRefType[Attribute]
+    )
+
+    swaps: builtin.ArrayAttr[ExchangeDeclarationAttr] | None = opt_prop_def(
+        builtin.ArrayAttr[ExchangeDeclarationAttr]
+    )
+
+    size = prop_def(IntegerAttr[IntegerType])
+
+    topo: dmp.RankTopoAttr | None = opt_prop_def(dmp.RankTopoAttr)
+
+    result = result_def(memref.MemRefType)
+
+    def __init__(
+        self,
+        input_stencil: SSAValue | Operation,
+        size: IntegerAttr[IntegerType],
+        topo: dmp.RankTopoAttr | None = None,
+        swaps: Sequence[ExchangeDeclarationAttr] = [],
+        result_type: memref.MemRefType[Attribute] | None = None,
+    ):
+        # if result_type is None:
+        #
+        #     result_type = builtin.TensorType()
+        super().__init__(
+            operands=[input_stencil],
+            properties={"size": size, "topo": topo, "swaps": builtin.ArrayAttr(swaps)},
+            result_types=[result_type],
+        )
+
+
+@irdl_op_definition
+class AccessOp(IRDLOperation):
+    """
+    A CSL stencil access that operates on data prefetched by `csl_stencil.prefetch`
+    """
+
+    T = Annotated[Attribute, ConstraintVar("T")]
+
+    name = "csl_stencil.access"
+    op: Operand = operand_def(memref.MemRefType)
+    offset: stencil.IndexAttr = attr_def(stencil.IndexAttr)
+    offset_mapping = opt_attr_def(stencil.IndexAttr)
+    result = result_def(TensorType)
+
+    traits = frozenset([HasAncestor(stencil.ApplyOp), Pure()])
+
+    def __init__(
+        self,
+        op: Operand,
+        offset: stencil.IndexAttr,
+        result_type: TensorType[Attribute],
+        offset_mapping: stencil.IndexAttr | None = None,
+    ):
+        super().__init__(
+            operands=[op],
+            attributes={"offset": offset, "offset_mapping": offset_mapping},
+            result_types=[result_type],
+        )
+
+    def print(self, printer: Printer):
+        printer.print(" ")
+        printer.print_operand(self.op)
+        printer.print_op_attributes(
+            self.attributes,
+            reserved_attr_names={"offset", "offset_mapping"},
+            print_keyword=True,
+        )
+
+        mapping = self.offset_mapping
+        if mapping is None:
+            mapping = range(len(self.offset))
+        offset = list(self.offset)
+
+        printer.print("[")
+        index = 0
+        for i in range(len(self.offset)):
+            if i in mapping:
+                printer.print(offset[index])
+                index += 1
+            else:
+                printer.print("_")
+            if i != len(self.offset) - 1:
+                printer.print(", ")
+        printer.print("]")
+
+        printer.print_string(" : ")
+        printer.print_attribute(self.op.type)
+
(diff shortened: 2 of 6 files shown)


tests/filecheck/dialects/csl/stencil-to-csl-stencil.mlir
```diff
@@ -0,0 +1,63 @@
...
+// CHECK-NEXT:       %18 = arith.addf %17, %6 : tensor<510xf32>
+// CHECK-NEXT:       %19 = tensor.empty() : tensor<510xf32>
+// CHECK-NEXT:       %20 = linalg.fill ins(%5 : f32) outs(%19 : tensor<510xf32>) -> tensor<510xf32>
+// CHECK-NEXT:       %21 = arith.mulf %18, %20 : tensor<510xf32>
+// CHECK-NEXT:       stencil.return %21 : tensor<510xf32>
+// CHECK-NEXT:     }
+// CHECK-NEXT:     stencil.store %2 to %b ([0, 0] : [1, 1]) : !stencil.temp<[0,1]x[0,1]xtensor<510xf32>> to !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>
+// CHECK-NEXT:     func.return
+// CHECK-NEXT:   }
+// CHECK-NEXT: }
```
superlopuh, line 63:
```suggestion
// CHECK-NEXT: }

```

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,270 @@
...
+    A simplified version of dmp.exchange, from which it should be lowered
+    """
+
+    name = "csl_stencil.exchange"
+
+    neighbor_: ParameterDef[builtin.DenseArrayBase]
+
+    def __init__(
+        self,
+        neighbor: Sequence[int],
```
superlopuh, line 42: I would strongly recommend allowing users of inits to pass the actual objects that will be stored directly. As it stands it's impossible to init this object with the exact data one would want in it, which is not great for ergonomics.

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,270 @@
...
+
+@irdl_attr_definition
+class ExchangeDeclarationAttr(ParametrizedAttribute):
+    """
+    A simplified version of dmp.exchange, from which it should be lowered
+    """
+
+    name = "csl_stencil.exchange"
+
+    neighbor_: ParameterDef[builtin.DenseArrayBase]
```
superlopuh, line 38: Is there a way to avoid the tralining underscore? It just looks a little weird and isn't super easy to distinguish from its tuple counterpart. I'd probably recommend removing the underscore here, and renaming `neighbor` to `neighbor_tuple`, but an alternative would be to make this something like `neighbor_attr`, or something along those lines.

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,270 @@
...
+        self,
+        input_stencil: SSAValue | Operation,
+        size: IntegerAttr[IntegerType],
+        topo: dmp.RankTopoAttr | None = None,
+        swaps: Sequence[ExchangeDeclarationAttr] = [],
+        result_type: memref.MemRefType[Attribute] | None = None,
+    ):
+        # if result_type is None:
+        #
+        #     result_type = builtin.TensorType()
```
superlopuh, line 108: commented code

n-io: thanks, removing

xdsl/dialects/csl_/csl_stencil.py
n-io: renamed to remove trailing underscore

xdsl/dialects/csl_/csl_stencil.py
n-io: accepting both types now

n-io:
I'd love suggestions on file structure/layout.

As for the dialect, riscv takes the approach of having several `riscv_*.py` files in `xdsl/dialects`, while `x86` has its own subdir. A subdir named `csl` would clash with the current `csl.py`, so I wanted to get feedback on how to structure that and where to place the intermediate dialect file (to be clear, the current `csl_` dir name is not supposed to be committed, only for the purposes of reviewing). Alternatively, it can go into experimental.

As for the transform, it is currently placed in `transform/experimental`. Alternatively, it can also be moved to `transform` or to `backend/csl/lowering`. I'd tend towards the latter, while wondering if it has any impact on where to place the dialect.

PapyChacal (approved):
Minor comment but LGTM generally!

On the file layout situation, I would be in favor of a `csl` directory containing a nested dialect file. Nothing enforces it to be directly in `xdsl/dialects`.
If you're concerned about the module path, i.e `import xdsl.dialects.csl.csl`, you could, in `xdsl/dialects/csl/__init__.py` import the dialect to expose it from there, enabling `from xdsl.dialects.csl import CSL` event if it is nested.
Note: I would personally not like `xdsl/dialects/csl/__init__.py` to contain the dialect definition itself though.
All of this being in answer to your mention of the topic; I wouldn't be against merging for any file layout point myself!

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+from xdsl.traits import HasAncestor, Pure
+from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
+
+
+@irdl_attr_definition
+class ExchangeDeclarationAttr(ParametrizedAttribute):
+    """
+    A simplified version of dmp.exchange, from which it should be lowered
+    """
```
AntonLydike, line 34: This is just a direction type, right? Could we represent this directly with `csl.direction` instead?

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+        assert isa(data, tuple[int, ...])
+        return data
+
+    def print_parameters(self, printer: Printer) -> None:
+        printer.print_string(f"<to {list(self.neighbor)}>")
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
+        parser.parse_characters("<")
+        parser.parse_characters("to")
```
AntonLydike, line 71: I would simplify the custom syntax to `csl_stencil.direction<-1,0>` (or even `csl.direction<NORTH>`)

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+
+    input_stencil: Operand = operand_def(
+        stencil.TempType[Attribute] | memref.MemRefType[Attribute]
+    )
+
+    swaps: builtin.ArrayAttr[ExchangeDeclarationAttr] | None = opt_prop_def(
+        builtin.ArrayAttr[ExchangeDeclarationAttr]
+    )
+
+    size = prop_def(IntegerAttr[IntegerType])
```
AntonLydike, line 96: This implies that we are always exchanging a 1d slice, right?

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+        )
+
+
+@irdl_op_definition
+class AccessOp(IRDLOperation):
+    """
+    A CSL stencil access that operates on data prefetched by `csl_stencil.prefetch`
+    """
+
+    T = Annotated[Attribute, ConstraintVar("T")]
```
AntonLydike, line 123: I don't see `T` used anywhere, did I miss something?

xdsl/dialects/csl_/__init__.py
superlopuh, line 1: why `csl_`?

xdsl/transforms/experimental/stencil_to_csl_stencil.py
superlopuh, line 1: I would recommend moving this out of experimental, and adding a test for this, even if it's fairly minimal

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+
+@dataclass(frozen=True)
+class ApplyOpAddPrefetch(RewritePattern):
+
+    prefetch_op: csl_stencil.PrefetchOp
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.ApplyOp, rewriter: PatternRewriter, /):
+
+        pass
```
superlopuh, line 30: what is the purpose of this rewrite pattern?

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+                    ]
+                ),
+                walk_reverse=False,
+                apply_recursively=False,
+            )
+
+            nested_rewriter.rewrite_op(new_apply_op)
+
+        # erase dmp.swap
+        # rewriter.erase_matched_op(False)
```
superlopuh, line 146: commented code

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+
+    name = "csl_stencil.access"
+    op: Operand = operand_def(memref.MemRefType)
+    offset: stencil.IndexAttr = attr_def(stencil.IndexAttr)
+    offset_mapping = opt_attr_def(stencil.IndexAttr)
```
AntonLydike, line 128: Is there a reason why we are using attributes instead of properties here?

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+        parser.parse_characters(">")
+
+        return [builtin.DenseArrayBase.from_list(builtin.i64, to)]
+
+
+@irdl_op_definition
+class PrefetchOp(IRDLOperation):
+    """
+    An op to indicate buffer prefetches.
+    """
```
AntonLydike, line 84: I would clarify that this is an op that talks about communication. I.e. that this represents a symmetrical exchange with neighbouring nodes.

In that sense I would strongly recommend naming this op similar to `dmp.swap`, as it is conceptuially quite close.

xdsl/transforms/experimental/stencil_to_csl_stencil.py
AntonLydike: Empty rewrite?

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: doing as suggested

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # index of the stencil access to be replaced by the prefetched buffer in the last arg
+    arg_index: int
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.AccessOp, rewriter: PatternRewriter, /):
+        assert len(op.offset) == 2
+        if tuple(op.offset) == (0, 0):
+            return
+        prefetched_arg = op.get_apply().region.block.args[-1]
```
AntonLydike: How do we know that the prefetched arg is always the last one?

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: Good spot, one that I didn't end up needing. Removing it.

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+            offset_mapping=op.offset_mapping,
+            result_type=t_type,
+        )
+
+        # if the matched op is immediately followed by tensor.ExtractSliceOps to remove ghost cells, replace both
+        if len(op.res.uses) == 1 and isinstance(
+            use := list(op.res.uses)[0].operation, tensor.ExtractSliceOp
+        ):
+            rewriter.replace_op(use, csl_access_op)
+            rewriter.erase_op(op)
```
AntonLydike: Can I get some comments here assuring and explaining tome that this is correct? I have a hard time understanding how removing the slice op here does not invalidate the whole semantics of the code...

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+                TensorType(t_type.get_element_type(), (uniform_size,)),
+                (len(op.swaps),),
+            ),
+        )
+
+        # replace dmp.swap (`replace_matched_op` does not work because `prefetch` produces a result)
+        # rewriter.insert_op(prefetch_op, InsertPoint.before(op))
+
+        # a dirty hack to get around a check that prevents me from replacing a no-results op with an n-results op
+        rewriter.replace_matched_op(prefetch_op, new_results=[])
```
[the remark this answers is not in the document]
n-io: removed commented code and changed 'dirty hack' to 'little hack'

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+        uniform_size = op.swaps.data[0].size[2]
+        assert all(swap.size[2] == uniform_size for swap in op.swaps)
+
+        assert isa(
+            op.input_stencil.type,
+            memref.MemRefType[Attribute] | stencil.TempType[Attribute],
+        )
+        assert isa(
+            t_type := op.input_stencil.type.get_element_type(), TensorType[Attribute]
+        )
```
AntonLydike: I really don't know how I feel about all these asserts in the rewrites... @PapyChacal @superlopuh do you know of a better way to encode specific requirements of the input IR on the pass?

I would at least like to see these asserts produce a meaningful warning about what is allowed and what isn't...

xdsl/transforms/experimental/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,163 @@
...
+    prefetch_op: csl_stencil.PrefetchOp
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.ApplyOp, rewriter: PatternRewriter, /):
+
+        pass
+
+
+@dataclass(frozen=True)
+class AccessOpFromPrefetch(RewritePattern):
```
AntonLydike: We usually give our rewrite patterns the "Pattern" suffix, so that it's easier to tell what's a pass and what's a pattern when importing/referencing them...

xdsl/transforms/experimental/stencil_to_csl_stencil.py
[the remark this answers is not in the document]
AntonLydike: Yep, I wouldn't call this a dirty hack

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # As promised by HasAncestor(ApplyOp)
+        trait = cast(HasAncestor, AccessOp.get_trait(HasAncestor, (stencil.ApplyOp,)))
+        apply = trait.get_ancestor(self)
+        assert isinstance(apply, stencil.ApplyOp)
+
+        # TODO This should be handled by infra, having a way to verify things on ApplyOp
+        # **before** its children.
+        # cf https://github.com/xdslproject/xdsl/issues/1112
+        apply.verify_()
```
AntonLydike, line 222: I'm not sure if this will not result in the verification errors being "attached" to the wrong operation right now, i.e. in wrong IR the error message of the invalid apply will be printed on this access instead of the apply...

xdsl/transforms/experimental/stencil_to_csl_stencil.py
AntonLydike: Commented code

xdsl/dialects/csl_/csl_stencil.py
```diff
@@ -0,0 +1,271 @@
...
+    size = prop_def(IntegerAttr[IntegerType])
+
+    topo: dmp.RankTopoAttr | None = opt_prop_def(dmp.RankTopoAttr)
+
+    result = result_def(memref.MemRefType)
+
+    def __init__(
+        self,
+        input_stencil: SSAValue | Operation,
+        size: IntegerAttr[IntegerType],
```
AntonLydike, line 105: Isn't there an implicit constraint here, that the resulting memref is always a `memref<${len(self.swaps}xtensor<${self.size}x??>>`?

I feel we don't need to pass both size and result type, as this is redundant information?

Some documentation on each attribute would definitely help people to understand this op much better.

AntonLydike (commented):
I think I have conceptual questions on both the `csl_stencil` dialect and the rewrite, maybe it makes sense to split the two off?

Also, this PR is severely lacking comments and docstrings. A lot of things are difficult to review for me, as I have to reconstruct what the method/op/rewrite/pass does from the code. This makes the review error prone as well, as I am never sure if you intended something or if it's a bug...

xdsl/dialects/csl_/csl_stencil.py
n-io: It's actually direction and distance.

xdsl/dialects/csl_/csl_stencil.py
n-io: Yes, correct

n-io:
> Minor comment but LGTM generally!
> 
> On the file layout situation, I would be in favor of a `csl` directory containing a nested dialect file. Nothing enforces it to be directly in `xdsl/dialects`. If you're concerned about the module path, i.e `from xdsl.dialects.csl.csl import CSL`, you could, in `xdsl/dialects/csl/__init__.py` import the dialect to expose it from there, enabling `from xdsl.dialects.csl import CSL` event if it is nested. Note: I would personally not _like_ `xdsl/dialects/csl/__init__.py` to contain the dialect definition itself though. All of this being in answer to your mention of the topic; I wouldn't be against merging for any file layout point myself!

@PapyChacal I believe that doing something like
```
from .csl import *

__all__ = ["csl"]
```
will unfortunately result in an [F405](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/) error. Not entirely sure how to go about it.

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: Removed.

xdsl/dialects/csl_/csl_stencil.py
PapyChacal: This is copy-pasta from the stencil dialect.

This verifier needs the apply's verification to be done before, so this is the current solution.
MLIR has two different verification hooks for this kind of situations, we only have one :shrug:

PapyChacal:
> > Minor comment but LGTM generally!
> > On the file layout situation, I would be in favor of a `csl` directory containing a nested dialect file. Nothing enforces it to be directly in `xdsl/dialects`. If you're concerned about the module path, i.e `from xdsl.dialects.csl.csl import CSL`, you could, in `xdsl/dialects/csl/__init__.py` import the dialect to expose it from there, enabling `from xdsl.dialects.csl import CSL` event if it is nested. Note: I would personally not _like_ `xdsl/dialects/csl/__init__.py` to contain the dialect definition itself though. All of this being in answer to your mention of the topic; I wouldn't be against merging for any file layout point myself!
> 
> @PapyChacal I believe having an `__init__.py` with something like
> 
> ```
> from .csl import *
> 
> __all__ = ["csl"]
> ```
> 
> will unfortunately result in an [F405](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/) error. Not entirely sure how to go about it.

We do this in a few places already, you could just add `# noqa: TID251` on that line

Or list explicitly what you want to expose

AntonLydike:
Hmm, I think it might instead be the relative part not the star part of the import that's the problem here, no?


On June 24, 2024 12:28:38 PM GMT+01:00, Emilien Bauer ***@***.***> wrote:
>> > Minor comment but LGTM generally!
>> > On the file layout situation, I would be in favor of a `csl` directory containing a nested dialect file. Nothing enforces it to be directly in `xdsl/dialects`. If you're concerned about the module path, i.e `from xdsl.dialects.csl.csl import CSL`, you could, in `xdsl/dialects/csl/__init__.py` import the dialect to expose it from there, enabling `from xdsl.dialects.csl import CSL` event if it is nested. Note: I would personally not _like_ `xdsl/dialects/csl/__init__.py` to contain the dialect definition itself though. All of this being in answer to your mention of the topic; I wouldn't be against merging for any file layout point myself!
>> 
>> @PapyChacal I believe having an `__init__.py` with something like
>> 
>> ```
>> from .csl import *
>> 
>> __all__ = ["csl"]
>> ```
>> 
>> will unfortunately result in an [F405](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/) error. Not entirely sure how to go about it.
>
>We do this in a few places already, you could just add `# noqa: TID251` on that line
>
>Or list explicitly what you want to expose
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/xdslproject/xdsl/pull/2766#issuecomment-2186350510
>You are receiving this because your review was requested.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: This is run by the `nested_rewriter` inside the `SwapToPrefetch` rewrite pattern. I've added comments to clarify.

xdsl/dialects/csl_/csl_stencil.py
n-io: Just because that's what `stencil.access` does. Happy to change it though

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: done as suggested

xdsl/dialects/csl_/csl_stencil.py
n-io: It's technically a point in the stencil, meaning direction + distance. I was thinking of translating it later (when lowering to calling the API functions), in order to keep `csl_stencil.access` in line with `stencil.access`, since in the current design plan they can co-exist.

xdsl/dialects/csl_/csl_stencil.py
n-io: Great spot, thanks!

xdsl/dialects/csl_/csl_stencil.py
n-io: That's a really good point actually. It might be easy for the symmetric nature of the communication to get lost in the word prefetch. At the same time, I wonder if `dmp.swap` semantics allow for more of a targeted point-to-point communication (i.e., true swapping) than CS hardware can actually support. I'll change the comment as suggested, and am not fully sure about the name right now.

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: It's complicated. `stencil-tensorize-z-dimension` adds extract slice ops to handle ghost cells and pass type validation. Since ghost cells are not prefetched, these ops can be removed again. If left in, they have no effect. Adding some additional checks on shape etc. to make sure it's safe to remove.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -59,7 +59,7 @@ def match_and_rewrite(self, op: stencil.AccessOp, rewriter: PatternRewriter, /):
 
 
 @dataclass(frozen=True)
-class SwapToPrefetch(RewritePattern):
+class SwapToPrefetchPattern(RewritePattern):
```
AntonLydike, line 62: nit: These usually start with a verb, e.g. `ConvertSwapToPrefetchPattern`

xdsl/dialects/csl_/csl_stencil.py
n-io: Excellent, removing it immediately! Thanks

xdsl/dialects/csl_/csl_stencil.py
n-io: changed as suggested.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: changed as suggested

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+        uniform_size = op.swaps.data[0].size[2]
+        assert all(swap.size[2] == uniform_size for swap in op.swaps)
+
+        assert isa(
+            op.input_stencil.type,
+            memref.MemRefType[Attribute] | stencil.TempType[Attribute],
+        )
+        assert isa(
+            t_type := op.input_stencil.type.get_element_type(), TensorType[Attribute]
+        )
```
AntonLydike, line 100: Still no messages on these asserts

xdsl/transforms/stencil_to_csl_stencil.py
AntonLydike: No? Am I missing something?

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    Rebuilds stencil.access by csl_stencil.access which operates on prefetched accesses.
+
+    stencil.access operates on stencil.temp types found at arg_index
+    csl_stencil.access operates on memref< num_neighbors x tensor< buf_size x data_type >> found at last arg index
+
+    Note: This is intended to be called in a nested pattern rewriter, such that the above precondition is met.
+    """
+
+    arg_index: int
```
AntonLydike, line 33: `arg_index` is not actually used in this pattern

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+            apply_op.region.block.insert_arg(
+                prefetch_op.result.type, len(apply_op.args)
+            )
+
+            # rebuild stencil.apply op
+            r_types = [r.type for r in apply_op.results]
+            assert isa(r_types, Sequence[stencil.TempType[Attribute]])
+            new_apply_op = stencil.ApplyOp.get(
+                list(apply_op.args) + [prefetch_op.result],
+                apply_op.region.clone(),
```
AntonLydike, line 143: I think detatching the region would be cheaper here.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+            # add the prefetched buffer as the last arg to stencil.access
+            apply_op.region.block.insert_arg(
+                prefetch_op.result.type, len(apply_op.args)
+            )
+
+            # rebuild stencil.apply op
+            r_types = [r.type for r in apply_op.results]
+            assert isa(r_types, Sequence[stencil.TempType[Attribute]])
+            new_apply_op = stencil.ApplyOp.get(
+                list(apply_op.args) + [prefetch_op.result],
```
AntonLydike, line 142: `[*apply_op.args, prefetch_op.result]` is more concise imo.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+            nested_rewriter.rewrite_op(new_apply_op)
+
+
+@dataclass(frozen=True)
+class StencilToCslStencilPass(ModulePass):
+    name = "stencil-to-csl-stencil"
+
+    def apply(self, ctx: MLContext, op: ModuleOp) -> None:
+        module_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier([SwapToPrefetchPattern()]),
```
AntonLydike, line 169: No need for the `GreedyRewritePatternApplier([SwapToPrefetchPattern()])` I think? You should be able to just `PatternRewriteWalker(SwapToPrefetchPattern()).rewrite_module()`

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+            walk_reverse=False,
```
AntonLydike, line 170: nit: this is the default

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+
+            # replace stencil.access (operating on stencil.temp at arg_index)
+            # with csl_stencil.access (operating on memref at last arg index)
+            nested_rewriter = PatternRewriteWalker(
+                GreedyRewritePatternApplier(
+                    [
+                        AccessOpFromPrefetchPattern(arg_idx),
+                    ]
+                ),
+                walk_reverse=False,
```
AntonLydike, line 156: nit: this is the default value

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,173 @@
...
+            and tuple(d.data for d in use.static_offsets.data) == (0,)
+            and tuple(d.data for d in use.static_strides.data) == (1,)
+            and len(use.offsets) == 0
+            and len(use.sizes) == 0
+            and len(use.strides) == 0
+        ):
+            rewriter.replace_op(use, csl_access_op)
+            rewriter.erase_op(op)
+        else:
+            rewriter.replace_matched_op(csl_access_op)
```
AntonLydike, line 67: This feels like it should be a canonicalization pattern on the `tensor.ExtractSliceOp` instead of part of this pattern, or am I missing something?

AntonLydike (commented):
Added more code-level comments to the rewrite itself

xdsl/dialects/csl_/csl_stencil.py
n-io: Done

xdsl/transforms/experimental/stencil_to_csl_stencil.py
n-io: Ok, I'll leave in the code until then if that's ok.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Good point, yes.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -0,0 +1,165 @@
...
+                GreedyRewritePatternApplier(
+                    [
+                        AccessOpFromPrefetchPattern(arg_idx),
+                    ]
+                ),
+                walk_reverse=False,
+                apply_recursively=False,
+            )
+
+            nested_rewriter.rewrite_op(new_apply_op)
```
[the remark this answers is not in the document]
n-io: I need to invoke it with that argument.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Changing as suggested

xdsl/transforms/stencil_to_csl_stencil.py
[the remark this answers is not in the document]
AntonLydike: Maybe this will become a simpler rewrite once we get the `csl_stencil.apply` ops?

It's usually a sign that the abstractions you are thinking with don't quite map to MLIRs cleanly. We can leave this in for now, but I think it's a sign that we are doing something not quite right yet.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Fixing this.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Now :)

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Added assert messages for non-type check asserts

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -0,0 +1,265 @@
...
+
+    input_stencil: Operand = operand_def(
+        stencil.TempType[Attribute] | memref.MemRefType[Attribute]
+    )
+
+    swaps: builtin.ArrayAttr[ExchangeDeclarationAttr] | None = opt_prop_def(
+        builtin.ArrayAttr[ExchangeDeclarationAttr]
+    )
+
+    topo: dmp.RankTopoAttr | None = opt_prop_def(dmp.RankTopoAttr)
```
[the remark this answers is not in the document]
n-io: Removing as suggested

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Nice

xdsl/transforms/stencil_to_csl_stencil.py
n-io: I strongly prefer to finish the dialect and transform first, as the design doc suggests there will be some follow-up rewriting, for which I'd prefer not to have the extract slice ops if at all possible.

PR #2767, opened 2024-06-21
core: make VarOperand, VarOpResult, and VarRegion tuple instead of list

superlopuh:
Comes with some minimal impacts on uses, but seems like a win in terms of communicating immutability.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -563,7 +563,7 @@ class Operation(IRNode):
     _operands: tuple[SSAValue, ...] = field(default=())
     """The operation operands."""
 
-    results: list[OpResult] = field(default_factory=list)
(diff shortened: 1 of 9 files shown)


xdsl/ir/core.py
```diff
@@ -563,7 +563,7 @@ class Operation(IRNode):
     _operands: tuple[SSAValue, ...] = field(default=())
     """The operation operands."""
 
-    results: list[OpResult] = field(default_factory=list)
+    results: tuple[OpResult, ...] = field(default=())
     """The results created by the operation."""
 
     successors: list[Block] = field(default_factory=list)
```
superlopuh, line 569: seems like this should also be a tuple, but I wasn't sure whether this PR was the place to do that

math-fehr (approved):
Nice!

PR #2768, opened 2024-06-24
dialects: Add qref dialect

alexarice:
Introduces the qref dialect as a companion to the qssa dialect. Also adds transforms to convert back and forth between these two dialects


alexarice:
Under advice from @AntonLydike I am going to break this into two prs

PR #2770, opened 2024-06-24
transformations: add qref/qssa conversions

alexarice:
Built on top of #2768 

Adds two passes which convert back and forth between the qssa and qref dialects. Also adds filecheck tests that test each pass and roundtrip of the passes

I think the `convert-qref-to-qssa` is particularly hacky, with the saving the old operands of the operation into a tuple, running replace on the operands and then inserting back the old ones. Would be keen to hear better ways to achieve the same behaviour

### DIFF
--- a/xdsl/dialects/qref.py
+++ b/xdsl/dialects/qref.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
-from abc import ABC
+from abc import ABC, abstractmethod
 
+from xdsl.dialects import qssa
 from xdsl.dialects.builtin import IntegerType
 from xdsl.ir import Dialect, ParametrizedAttribute, SSAValue, TypeAttribute
 from xdsl.irdl import (
@@ -30,12 +31,35 @@ qubit = QubitAttr()
 
 
 class QRefBase(IRDLOperation, ABC):
-    pass
+    @abstractmethod
+    def ssa_op(self) -> qssa.QssaBase:
+        """
+        Get corresponding qssa operation
+        """
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def is_gate(self) -> bool:
+        """
+        Is this operation a gate
+        """
+        raise NotImplementedError()
 
 
 @irdl_op_definition
 class QRefAllocOp(QRefBase):
     name = "qref.alloc"
+
+    def ssa_op(self) -> qssa.QubitAllocOp:
+        return qssa.QubitAllocOp.create(
+            result_types=[qssa.qubit] * self.num_qubits,
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return False
 
     res: VarOpResult = var_result_def(qubit)
 
@@ -70,6 +94,17 @@ class QRefAllocOp(QRefBase):
 class HGateOp(QRefBase):
     name = "qref.h"
 
+    def ssa_op(self) -> qssa.HGateOp:
+        return qssa.HGateOp.create(
+            operands=self.operands,
+            result_types=(qssa.qubit,),
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return True
+
     input = operand_def(qubit)
 
     assembly_format = "$input attr-dict"
@@ -84,6 +119,17 @@ class HGateOp(QRefBase):
 @irdl_op_definition
 class CNotGateOp(QRefBase):
     name = "qref.cnot"
+
+    def ssa_op(self) -> qssa.CNotGateOp:
+        return qssa.CNotGateOp.create(
+            operands=self.operands,
+            result_types=(qssa.qubit, qssa.qubit),
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return True
 
     in1 = operand_def(qubit)
 
@@ -102,6 +148,17 @@ class CNotGateOp(QRefBase):
 class CZGateOp(QRefBase):
     name = "qref.cz"
 
+    def ssa_op(self) -> qssa.CZGateOp:
+        return qssa.CZGateOp.create(
+            operands=self.operands,
+            result_types=(qssa.qubit, qssa.qubit),
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return True
+
     in1 = operand_def(qubit)
 
     in2 = operand_def(qubit)
@@ -118,6 +175,17 @@ class CZGateOp(QRefBase):
 @irdl_op_definition
 class MeasureOp(QRefBase):
     name = "qref.measure"
+
+    def ssa_op(self) -> qssa.MeasureOp:
+        return qssa.MeasureOp.create(
+            operands=self.operands,
+            result_types=(IntegerType(1),),
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return False
 
     input = operand_def(qubit)
 
--- a/xdsl/dialects/qref.py
+++ b/xdsl/dialects/qref.py
@@ -31,10 +31,17 @@ qubit = QubitAttr()
 
 
 class QRefBase(IRDLOperation, ABC):
+    """
+    Base class for qref operations, with methods to help convert to the qssa dialect.
+
+    Invariant:
+    self.is_gate == self.ssa_op().is_gate
+    """
+
     @abstractmethod
     def ssa_op(self) -> qssa.QssaBase:
         """
-        Get corresponding qssa operation
+        Build corresponding qssa operation
         """
         raise NotImplementedError()
 
@@ -42,7 +49,10 @@ class QRefBase(IRDLOperation, ABC):
     @abstractmethod
     def is_gate(self) -> bool:
         """
-        Is this operation a gate
+        Is this operation a gate?
+        Qref gates represent standard quantum logic gates
+        They should have no results
+        The results of the generated gate must be rewired when converting to the qssa dialect
         """
         raise NotImplementedError()
 
--- a/xdsl/dialects/qref.py
+++ b/xdsl/dialects/qref.py
@@ -61,16 +61,6 @@ class QRefBase(IRDLOperation, ABC):
 class QRefAllocOp(QRefBase):
     name = "qref.alloc"
 
-    def ssa_op(self) -> qssa.QubitAllocOp:
-        return qssa.QubitAllocOp.create(
-            result_types=[qssa.qubit] * self.num_qubits,
-            attributes=self.attributes,
-        )
-
-    @property
-    def is_gate(self) -> bool:
-        return False
-
     res: VarOpResult = var_result_def(qubit)
 
     def __init__(self, num_qubits: int):
@@ -99,10 +89,30 @@ class QRefAllocOp(QRefBase):
 
         printer.print_op_attributes(self.attributes)
 
+    def ssa_op(self) -> qssa.QubitAllocOp:
+        return qssa.QubitAllocOp.create(
+            result_types=[qssa.qubit] * self.num_qubits,
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return False
+
 
 @irdl_op_definition
 class HGateOp(QRefBase):
     name = "qref.h"
+
+    input = operand_def(qubit)
+
+    assembly_format = "$input attr-dict"
+
+    def __init__(self, input: SSAValue):
+        super().__init__(
+            operands=(input,),
+            result_types=(),
+        )
 
     def ssa_op(self) -> qssa.HGateOp:
         return qssa.HGateOp.create(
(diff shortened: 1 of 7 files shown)


xdsl/dialects/qref.py
```diff
@@ -30,13 +31,36 @@ class QubitAttr(ParametrizedAttribute, TypeAttribute):
...
+        Get corresponding qssa operation
+        """
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def is_gate(self) -> bool:
+        """
+        Is this operation a gate
+        """
```
AntonLydike, line 46: Can you add a docstring to explain why you need this helper?

xdsl/transforms/convert_qref_to_qssa.py
```diff
@@ -0,0 +1,46 @@
...
+)
+
+
+class QRefToQssaRewritePattern(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: QRefBase, rewriter: PatternRewriter):
+        # Create replacement operation
+        new_op: QssaBase = op.ssa_op()
+        # For gates there are no results to replace
+        new_results = () if new_op.is_gate else new_op.results
```
AntonLydike, line 20: Couldn't this check be replaced with a check of "has the new_op results that are a qubit"?

xdsl/transforms/convert_qref_to_qssa.py
```diff
@@ -0,0 +1,46 @@
...
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class QRefToQssaRewritePattern(RewritePattern):
```
AntonLydike, line 14: nit: Name suggestion `ConvertQrefToQssaPattern`

xdsl/transforms/convert_qssa_to_qref.py
```diff
@@ -0,0 +1,26 @@
...
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class QssaToQRefRewritePattern(RewritePattern):
```
AntonLydike, line 13: nit: `ConvertQssaToQrefPattern`

AntonLydike (commented):
Would love to see docstrings on the patterns and/or passes that explain what they do!

xdsl/transforms/convert_qref_to_qssa.py
alexarice: `is_gate` should be equivalent to "The qubit-typed operands match one-to-one with the qubit-typed results".

alexarice: This property could be removed and replaced by an actual check of this. Essentially the `gates` are the things we actually need to modify in the transform as they switch between having results and having no results depending on the dialect. Alloc and measure on the other hand are basically unchanged between dialects (apart from their types)

xdsl/dialects/qref.py
superlopuh: I agree, also a more explicit description of what a gate is in this context.

xdsl/dialects/qref.py
```diff
@@ -30,13 +31,36 @@ class QubitAttr(ParametrizedAttribute, TypeAttribute):
...
+        return qssa.QubitAllocOp.create(
+            result_types=[qssa.qubit] * self.num_qubits,
+            attributes=self.attributes,
+        )
+
+    @property
+    def is_gate(self) -> bool:
+        return False
+
     res: VarOpResult = var_result_def(qubit)
```
n-io, line 64: Style comment: Can you move `res` before the methods please

xdsl/dialects/qref.py
```diff
@@ -70,6 +94,17 @@ def print(self, printer: Printer):
 class HGateOp(QRefBase):
     name = "qref.h"
 
+    def ssa_op(self) -> qssa.HGateOp:
```
n-io, line 97: Best to keep dialect conversion functions in the converter rather than on the dialect.

xdsl/dialects/qref.py
```diff
@@ -85,6 +120,17 @@ def __init__(self, input: SSAValue):
 class CNotGateOp(QRefBase):
     name = "qref.cnot"
 
+    def ssa_op(self) -> qssa.CNotGateOp:
```
n-io, line 123: Likewise

xdsl/dialects/qref.py
```diff
@@ -102,6 +148,17 @@ def __init__(self, in1: SSAValue, in2: SSAValue):
 class CZGateOp(QRefBase):
     name = "qref.cz"
 
+    def ssa_op(self) -> qssa.CZGateOp:
```
n-io, line 151: Likewise

xdsl/dialects/qref.py
```diff
@@ -119,6 +176,17 @@ def __init__(self, in1: SSAValue, in2: SSAValue):
 class MeasureOp(QRefBase):
     name = "qref.measure"
 
+    def ssa_op(self) -> qssa.MeasureOp:
```
n-io, line 179: Likewise

xdsl/dialects/qssa.py
```diff
@@ -29,14 +30,39 @@ class QubitAttr(ParametrizedAttribute, TypeAttribute):
...
+
     pass
 
 
 @irdl_op_definition
-class QubitAllocOp(QubitBase):
+class QubitAllocOp(QssaBase):
     name = "qssa.alloc"
 
+    def ref_op(self) -> qref.QRefAllocOp:
```
n-io, line 56: Likewise

xdsl/dialects/qssa.py
```diff
@@ -67,9 +93,20 @@ def print(self, printer: Printer):
 
 
 @irdl_op_definition
-class HGateOp(QubitBase):
+class HGateOp(QssaBase):
     name = "qssa.h"
 
+    def ref_op(self) -> qref.HGateOp:
```
n-io, line 99: Likewise

xdsl/dialects/qssa.py
```diff
@@ -84,9 +121,20 @@ def __init__(self, input: SSAValue):
 
 
 @irdl_op_definition
-class CNotGateOp(QubitBase):
+class CNotGateOp(QssaBase):
     name = "qssa.cnot"
 
+    def ref_op(self) -> qref.CNotGateOp:
```
n-io, line 127: Likewise

xdsl/dialects/qssa.py
```diff
@@ -105,9 +153,20 @@ def __init__(self, in1: SSAValue, in2: SSAValue):
 
 
 @irdl_op_definition
-class CZGateOp(QubitBase):
+class CZGateOp(QssaBase):
     name = "qssa.cz"
 
+    def ref_op(self) -> qref.CZGateOp:
```
n-io, line 159: Likewise

xdsl/dialects/qssa.py
```diff
@@ -126,9 +185,20 @@ def __init__(self, in1: SSAValue, in2: SSAValue):
 
 
 @irdl_op_definition
-class MeasureOp(QubitBase):
+class MeasureOp(QssaBase):
     name = "qssa.measure"
 
+    def ref_op(self) -> qref.MeasureOp:
```
n-io, line 191: Likewise

n-io (commented):
Looks good, I'd suggest to separate out dialect and transform code

xdsl/dialects/qref.py
n-io: @AntonLydike I maybe incorrectly believed that this would cause some style check to fail, since it is done this way across all xdsl ops I've seen.

xdsl/dialects/qref.py
alexarice: I am unsure if this is desirable, as I don't see a way around having a big `if isinstance` chain in the converter without doing it this way. Another advantage of this method is that if a new operation is added, then the implementation of these two functions is forced

n-io: You could match on the specific op, create a new one, and move any shared code to a helper function., e.g. something along the lines of:

```
class ConvertHGateToQssaPattern(RewritePattern):
    @op_type_rewrite_pattern
    def match_and_rewrite(self, op: qref.HGateOp, rewriter: PatternRewriter):
        replace_with_ssa(
            op,
            qssa.HGateOp.create(
                operands=self.operands,
                result_types=(qssa.qubit,),
                attributes=self.attributes,
            ),
            rewriter,
        )
```

alexarice: This still runs into the problem of not forcing the conversion to be written for new ops, though I guess a test could be written that ensures this

xdsl/dialects/qssa.py
```diff
@@ -29,12 +30,35 @@ class QubitAttr(ParametrizedAttribute, TypeAttribute):
...
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def is_gate(self) -> bool:
+        """
+        Is this operation a gate?
+        Qssa gates represent standard quantum logic gates
+        They should have an equal number of qubit-typed operands and results
+        Their results must be rewired when converting to the qref dialect
```
superlopuh, line 55: nit:
```suggestion
        Is this operation a gate?
        Qssa gates represent standard quantum logic gates.
        They should have an equal number of qubit-typed operands and results.
        Their results must be rewired when converting to the qref dialect.
```
And in other doc comments

xdsl/dialects/qref.py
superlopuh: I kind of like the current setup. The only thing I might suggest is to have a `QuantumGate` trait instead of `is_gate()` but even that seems like a nit, and like less good a way to guarantee that each op will have to implement these methods when it's added. It also feels like the two dialects are coupled in a pretty fundamental way, which seems like an ok design decision, making the abstract method style of conversion quite natural. I'm not sure in what situation in the future we would regret merging this PR as-is.

xdsl/transforms/convert_qref_to_qssa.py
```diff
@@ -0,0 +1,57 @@
...
+
+        rewriter.replace_matched_op(new_op, new_results)
+
+        if not new_op.is_gate:
+            return
+
+        # For gates we replace any other occurences of the original operands
+        # with the results of the new operation
+
+        old_operands = tuple(new_op.operands)
```
alexarice, line 36: @math-fehr I requested a review to see if you knew a better way to do this section, but feel free to ignore if you're too busy

superlopuh (approved):

xdsl/dialects/qref.py
```diff
@@ -30,7 +31,30 @@ class QubitAttr(ParametrizedAttribute, TypeAttribute):
...
+
+    @property
+    @abstractmethod
+    def is_gate(self) -> bool:
+        """
+        Is this operation a gate?
+        Qref gates represent standard quantum logic gates.
+        They should have no results.
+        The results of the generated gate must be rewired when converting to the qssa dialect.
+        """
```
AntonLydike, line 56: :+1: Nice and concise docstring!

xdsl/dialects/qssa.py
```diff
@@ -29,12 +30,35 @@ class QubitAttr(ParametrizedAttribute, TypeAttribute):
 qubit = QubitAttr()
 
 
-class QubitBase(IRDLOperation, ABC):
-    pass
+class QssaBase(IRDLOperation, ABC):
+    """
```
AntonLydike, line 34: I like this renaming, big fan :muscle:

xdsl/transforms/convert_qref_to_qssa.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Converts uses of the qref dialect to the qssa dialect in a module.
+    Inverse to the "convert-qssa-to-qref" pass.
+    """
+
+    name = "convert-qref-to-qssa"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            ConvertQRefToQssaPattern(), apply_recursively=False
```
AntonLydike, line 56: Supernit: I'm 99% sure you don't need to specify `apply_recursively=False` here as it should not impact the result at all.

Every non-standard option specified unnecessarily introduces reviewer overhead.

AntonLydike (approved):
Good stuff!

xdsl/transforms/convert_qref_to_qssa.py
superlopuh: I disagree, it has a performance impact, and should be added whenever the rewrites don't insert operations that could be recursively matched.

xdsl/dialects/qref.py
n-io: > This still runs into the problem of not forcing the conversion to be written for new ops, though I guess a test could be written that ensures this

This could also be a feature. Forcing a conversion to be made also implies an unwritten guarantee that there will always be a valid 1:1 conversion, and that neither of these dialects will ever hold concepts that are not directly convertible, now or in the future. I just wanted to draw attention to the fact that this tight coupling might be a rather committal decision. I don't fully know the plans for these dialects, so if you're happy then please go ahead and merge.

xdsl/transforms/convert_qref_to_qssa.py
alexarice: You are probably right, think I was worried about things being converted in the "wrong" order which could maybe cause verification problems (I'm not actually sure when the verifier is run)

n-io (approved):
If we're happy with a tight coupling of these dialects (see discussion above) I'm happy to approve.

xdsl/transforms/convert_qref_to_qssa.py
alexarice: The tests still succeed without it, so I guess it's just a performance thing

PR #2776, opened 2024-06-25
transformations: Add `memref-stream-(infer|generalize)-fill` passes

superlopuh:
These let us detect fill operations, fold them if possible, and convert them back to a form that we can lower if they could not be folded. Also, updates the bottom_up tests to go all the way from linalg.generic to our snitch assembly. Note that this is not the case yet for the matrix op, as that one is interleaved and peeled, those are the two steps remaining.

Note stacked PR.


tests/filecheck/dialects/memref/canonicalize.mlir
superlopuh, line 1: part of another PR

xdsl/dialects/memref.py
superlopuh, line 1: part of another PR

xdsl/transforms/canonicalization_patterns/memref.py
superlopuh, line 1: part of another PR

tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
```diff
@@ -500,20 +519,30 @@ func.func public @pooling_nchw_sum_d1_s2_3x3(
...
+        indexing_maps = [
+            affine_map<(d0, d1, d2, d3) -> ()>,
+            affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+        ],
+        iterator_types = ["parallel", "parallel", "parallel", "parallel"]
+    } ins(%zero_float : f64) outs(%Y : memref<1x1x7x7xf64>) {
+    ^bb0(%in: f64, %out: f64):
+        linalg.yield %in : f64
+    }
+    %alloc = memref.alloc() {alignment = 64 : i64} : memref<3x3xf32>
```
superlopuh, line 532: note that these allocs get canonicalized away, as they're used only for their shapes

zero9178 (approved):

PR #2777, opened 2024-06-25
(minor): allow csl_stencil.access to operate on own data

n-io:
Follow-up from #2766

The original plan was to lower `stencil.access` to `csl_stencil.access` if and only if the access is to neighbour data (i.e. prefetched), while accesses to own data were supposed to remain `stencil.access` without being lowered.

This runs into the very practical problem, that `stencil.access` can only exist within a `stencil.apply` op - which we also want to lower, but this cannot be done with `stencil.access` ops.

### DIFF
--- a/xdsl/dialects/csl/csl_stencil.py
+++ b/xdsl/dialects/csl/csl_stencil.py
@@ -124,11 +124,17 @@ class PrefetchOp(IRDLOperation):
 @irdl_op_definition
 class AccessOp(IRDLOperation):
     """
-    A CSL stencil access that operates on data prefetched by `csl_stencil.prefetch`
+    A CSL stencil access that operates on own data or data prefetched from neighbors via `csl_stencil.prefetch`
+
+    The source of data determines the type `op` is required to have:
+
+      ${type(op) == memref.MemRefType}  -  for accesses to data prefetched from neighbors
+      ${type(op) == stencil.Temp}       -  for accesses to own data
+
     """
 
     name = "csl_stencil.access"
-    op = operand_def(memref.MemRefType)
+    op = operand_def(memref.MemRefType | stencil.TempType)
     offset = prop_def(stencil.IndexAttr)
     offset_mapping = opt_prop_def(stencil.IndexAttr)
     result = result_def(TensorType)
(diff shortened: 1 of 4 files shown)


xdsl/dialects/csl/csl_stencil.py
```diff
@@ -124,11 +124,17 @@ def __init__(
 @irdl_op_definition
 class AccessOp(IRDLOperation):
     """
-    A CSL stencil access that operates on data prefetched by `csl_stencil.prefetch`
+    A CSL stencil access that operates on own data or data prefetched from neighbors via `csl_stencil.prefetch`
+
+    The source of data determines the type `op` is required to have:
+
+      ${type(op) == memref.MemRefType}  -  for accesses to data prefetched from neighbors
+      ${type(op) == stencil.Temp}       -  for accesses to own data
```
AntonLydike, line 132: I'm a bit concerned about the reckless mixing of value and reference semantics here. Not sure what to think of it...

n-io: That's a valid concern, and I suppose is owed to accesses to neighbour data operating on prefetched data, having effectively gone through one more round of lowering as compared to accesses to own data.

The plan was to not touch accesses to own data and leave them as `stencil.access` (without lowering to `csl_stencil.access`). What this means in the code is that there already is a mixing of value and reference semantics, just on different ops.

The question then partly becomes, how to do this cleanly. I see three ways:
* allow `stencil.access` to have parents other than `stencil.apply`
* allow mixing of value semantics (own data) and reference semantics (neighbour data), as proposed here
* separate the ops into something along the lines of `csl_stencil.access_own_data` and `csl_stencil.access_neighbor_data` - this would work but not be my personal preference because of potential confusion around the `get_accesses` function and I struggle to see what we might gain from this

n-io: @PapyChacal what would you suggest?

AntonLydike: Well, if in doubt, press merge. Having slept over this for a night, I think it's a symptom of us not having figured out quite how/when bufferisation kicks in. Maybe this isn't a problem we need to solve, maybe it is.

Let's see where this PR takes us instead of sweating too much with the semantics right now.

PR #2778, opened 2024-06-25
backend: (riscv) riscv-prologue-epilogue-insertion

bpervan:
[Riscv-prologue-epilogue-insertion pass](https://github.com/xdslproject/xdsl/blob/main/xdsl/backend/riscv/prologue_epilogue_insertion.py) ensures compatibility with RISC-V ABI in terms of preserving the content of callee-saved registers. This brief PR adds the immediate return from the pass in case no callee-saved registers are used by the function, thus preventing the issuing of unnecessary `addi sp, sp, 0`.

`test_get_all_available_passes` is changed to adjust the output to the expected one (temporary workaround).


bpervan:
@zero9178 @superlopuh

zero9178:
Could you also add a test-case to the existing tests in https://github.com/xdslproject/xdsl/blob/ad76f511b94bdfa4d51979d0e41fd52908a61521/tests/filecheck/backend/riscv/prologue_epilogue_insertion.mlir? 

A simple function that does not use callee-saved registers + a `CHECK-NOT:` for the absence of a `addi sp, sp` should be good enough.

xdsl/backend/riscv/prologue_epilogue_insertion.py
```diff
@@ -42,6 +42,9 @@ def _process_function(self, func: riscv_func.FuncOp) -> None:
             if res.type in Registers.S or res.type in Registers.FS
         )
 
+        if len(used_callee_preserved_registers) == 0:
```
superlopuh, line 45:
```suggestion
        if not used_callee_preserved_registers:
```
This is the idiomatic way to check if a sequence is empty in Python, it avoids a potentially costly length computation.

bpervan: Sure, I'm not yet completely familiar with those 👍

superlopuh (approved):

bpervan:
> Could you also add a test-case to the existing tests in https://github.com/xdslproject/xdsl/blob/ad76f511b94bdfa4d51979d0e41fd52908a61521/tests/filecheck/backend/riscv/prologue_epilogue_insertion.mlir?
> 
> A simple function that does not use callee-saved registers + a `CHECK-NOT:` for the absence of a `addi sp, sp` should be good enough.

Updated, take a look please

tests/filecheck/backend/riscv/prologue_epilogue_insertion.mlir
```diff
@@ -50,3 +50,43 @@ riscv_func.func @main() {
...
+  %3 = riscv.fmv.s %1 : (!riscv.freg<fa1>) -> !riscv.freg<ft1>
+  %4 = riscv.fadd.s %2, %3 : (!riscv.freg<ft0>, !riscv.freg<ft1>) -> !riscv.freg<ft0>
+  %5 = riscv.fmv.s %4 : (!riscv.freg<ft0>) -> !riscv.freg<fa0>
+
+  // CHECK-NOT: %{{.*}} = riscv.addi %{{.*}}, 0 : (!riscv.reg<sp>) -> !riscv.reg<sp>
+
+  // CHECK: riscv_func.return %{{.*}}
+  // CHECK-SAME: !riscv.freg<fa0>
+  riscv_func.return %5 : !riscv.freg<fa0>
+}
```
zero9178, line 92:
```suggestion
}

```

zero9178 (approved):

PR #2781, opened 2024-06-26
(dialects): csl_stencil apply op

n-io:
This PR implements the `csl_stencil.apply` op as outlined in Step 2 of #2747 

This operation combines a `csl_stencil.prefetch` (symmetric buffer communication across a given stencil shape) with a `stencil.apply`. Please see the doc string of the op for a detailed description.

### DIFF
--- a/tests/dialects/test_csl_stencil.py
+++ b/tests/dialects/test_csl_stencil.py
@@ -41,7 +41,6 @@ def test_access_patterns():
         ],
         result_types=[tens_t],
     )
-    # apply = ApplyOp.get((temp, temp), apply_op_region.detach_block(0), [typ])
 
     r0_t0_acc, r0_t1_acc, r1_t0_acc, r1_t1_acc = tuple(apply.get_accesses())
 
--- a/tests/dialects/test_csl_stencil.py
+++ b/tests/dialects/test_csl_stencil.py
@@ -1,0 +1,57 @@
+from xdsl.builder import Builder
+from xdsl.dialects.builtin import IntegerAttr, IntegerType, MemRefType, TensorType, f32
+from xdsl.dialects.csl.csl_stencil import AccessOp, ApplyOp
+from xdsl.dialects.stencil import IndexAttr, TempType
+from xdsl.ir import Region, SSAValue
+from xdsl.utils.test_value import TestSSAValue
+
+
+def test_access_patterns():
+    temp_t = TempType(5, f32)
+    temp = TestSSAValue(temp_t)
+    mref = TestSSAValue(mref_t := MemRefType(tens_t := TensorType(f32, (5,)), (4,)))
+
+    @Builder.implicit_region((mref_t, temp_t))
+    def region0(args: tuple[SSAValue, ...]):
+        t0, t1 = args
+        for x in (-1, 1):
+            AccessOp(t0, IndexAttr.get(x, 0), tens_t)
+        for y in (-1, 1):
+            AccessOp(t0, IndexAttr.get(0, y), tens_t)
+
+        AccessOp(t1, IndexAttr.get(1, 1), tens_t)
+        AccessOp(t1, IndexAttr.get(-1, -1), tens_t)
+
+    @Builder.implicit_region((temp_t, temp_t))
+    def region1(args: tuple[SSAValue, ...]):
+        t0, t1 = args
+        AccessOp(t0, IndexAttr.get(0, 0), tens_t)
+        AccessOp(t1, IndexAttr.get(0, 0), tens_t)
+
+    apply = ApplyOp(
+        operands=[temp, mref, []],
+        properties={
+            "swaps": None,
+            "topo": None,
+            "num_chunks": IntegerAttr(1, IntegerType(64)),
+        },
+        regions=[
+            Region(region0.detach_block(0)),
+            Region(region1.detach_block(0)),
+        ],
+        result_types=[tens_t],
+    )
+    # apply = ApplyOp.get((temp, temp), apply_op_region.detach_block(0), [typ])
+
+    r0_t0_acc, r0_t1_acc, r1_t0_acc, r1_t1_acc = tuple(apply.get_accesses())
+
+    assert r0_t0_acc.visual_pattern() == " X \nXOX\n X "
+    assert r0_t1_acc.visual_pattern() == "X  \n O \n  X"
+
+    assert not r0_t0_acc.is_diagonal
+    assert r0_t1_acc.is_diagonal
+
+    assert len(tuple(r0_t1_acc.get_diagonals())) == 2
+
+    assert r1_t0_acc.visual_pattern() == "X"
+    assert r1_t1_acc.visual_pattern() == "X"
--- a/xdsl/dialects/csl/csl_stencil.py
+++ b/xdsl/dialects/csl/csl_stencil.py
@@ -1,11 +1,21 @@
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 from itertools import pairwise
 from typing import cast
 
 from xdsl.dialects import builtin, memref, stencil
-from xdsl.dialects.builtin import TensorType
+from xdsl.dialects.builtin import IndexType, IntegerAttr, TensorType
 from xdsl.dialects.experimental import dmp
-from xdsl.ir import Attribute, Dialect, Operation, ParametrizedAttribute, SSAValue
+from xdsl.dialects.utils import AbstractYieldOperation
+from xdsl.ir import (
+    Attribute,
+    Block,
+    BlockArgument,
+    Dialect,
+    Operation,
+    ParametrizedAttribute,
+    Region,
+    SSAValue,
+)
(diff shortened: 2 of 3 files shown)


n-io:
Discussion:

It is not entirely clear how best to deal with the block args of the two regions. To recap briefly, both the op and its two regions have some required args plus optional args, e.g.:
```
op(op_req_args, op_opt_args...)
block1(block1_req_args, block1_opt_args...)
block2(block2_req_args, block2_opt_args...)
```

The approach taken here (also see the doc string for a more detailed discussion), is that `op_opt_args = block1_opt_args + block2_opt_args`.
Alternatively, we could simply say that `op_opt_args == block1_opt_args == block2_opt_args` to simply use additional args on both blocks. Easier to handle, some redundancy.

More than happy to invite comments on this. Alternatively, once work on the transform commences and it may become clear that whatever solution proposed now is practically cumbersome, this may of course be changed again to allow manageable effort in the transform.

Bear in mind, that the API calls this is supposed to be lowered to have fixed function signatures, so this is predominantly to deal with `IsolatedFromAbove`, which I would prefer to avoid resolving at this stage.

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
...
+    args = var_operand_def(Attribute)
+
+    chunk_reduce = region_def()
+    post_process = region_def()
+
+    swaps = prop_def(builtin.ArrayAttr[ExchangeDeclarationAttr])
+
+    topo = opt_prop_def(dmp.RankTopoAttr)
+
+    num_chunks = prop_def(IntegerAttr)
```
AntonLydike, line 198: There seem to be some undocumented attributes here still, can you add some info on what they do? (Maybe as a docstring directly attached to the attributes?

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
...
+    iter_arg = operand_def(TensorType[Attribute])
+
+    args = var_operand_def(Attribute)
+
+    chunk_reduce = region_def()
+    post_process = region_def()
+
+    swaps = prop_def(builtin.ArrayAttr[ExchangeDeclarationAttr])
+
+    topo = opt_prop_def(dmp.RankTopoAttr)
```
AntonLydike, line 196: Can you elaborate on why this is optional? what happens when this is not given?

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
...
+        assert len(result_types) > 0
+
+        if isinstance(body, Block):
+            body = Region(body)
+
+        return ApplyOp.build(
+            operands=[list(args)],
+            regions=[body],
+            result_types=[result_types],
+        )
```
AntonLydike, line 280: Can you make this an `__init__`?

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
...
+        parser.parse_punctuation("(")
+        chunk_reduce = parser.parse_region()
+        parser.parse_punctuation(",")
+        post_process = parser.parse_region()
+        parser.parse_punctuation(")")
+        return cls(
+            operands=[operands[0], operands[1], operands[2:]],
+            result_types=[result_types],
+            regions=[chunk_reduce, post_process],
+            properties={"topo": topo, "swaps": swaps, "num_chunks": num_chunks},
```
AntonLydike, line 261: nit: I think you can just pass `props` here instead of unpacking the dict and then re-packing it.

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
...
+                if not isinstance(use.operation, AccessOp):
+                    continue
+                access: AccessOp = use.operation
+                # grab the offsets as a tuple[int, ...]
+                offsets = tuple(access.offset)
+                # account for offset_mappings:
+                if access.offset_mapping is not None:
+                    offsets = tuple(offsets[i] for i in access.offset_mapping)
+                accesses.append(offsets)
+            yield stencil.AccessPattern(tuple(accesses))
```
AntonLydike, line 372: Can you add a test that this is still working as expected?

AntonLydike (commented):
Great progress! Feels interesting to see how the stencil dialect is adapted here.

As usual, I would just like some more documentation :)

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -139,7 +390,7 @@ class AccessOp(IRDLOperation):
     offset_mapping = opt_prop_def(stencil.IndexAttr)
     result = result_def(TensorType)
 
-    traits = frozenset([HasAncestor(stencil.ApplyOp), Pure()])
+    traits = frozenset([HasAncestor(stencil.ApplyOp, ApplyOp), Pure()])  # todo
```
dk949, line 393: What should be done for this todo?

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
         )
 
 
+@irdl_op_definition
+class ApplyOp(IRDLOperation):
```
dk949, line 145: I'm not 100% sure I understand why this is a single op with 2 regions, rather than something like this:

```mlir
%reduced = csl_stencil.chunk_reduce(%args) {
	...
}
%res = csl_stencil.post_process(%reduced) {
	...
}
```


I'm not proposing to change this, I'm just confused..

xdsl/dialects/csl/csl_stencil.py
n-io: Nothing, whoops :)

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -121,6 +141,237 @@ def __init__(
...
+    Note, that %pref can be dropped (as communication is done by the op rather than before the op),
+    and that a new %iter_arg is required, an empty tensor which is filled by `chunk_reduce` and
+    consumed by `post_process`
+    """
+
+    name = "csl_stencil.apply"
+
+    communicated_stencil = operand_def(
+        stencil.TempType[Attribute] | memref.MemRefType[Attribute]
+    )
```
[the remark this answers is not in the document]
n-io: > the stencil was being communicated

That's what I meant by 'communicated'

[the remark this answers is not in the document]
n-io: passive voice as opposed to past tense

xdsl/dialects/csl/csl_stencil.py
n-io: Adding info on what they do

xdsl/dialects/csl/csl_stencil.py
n-io: It's optional because it's optional in `dmp.swap`. Not sure why, but changing it to non-optional.

xdsl/dialects/csl/csl_stencil.py
n-io: I probably want to take this out for now.

xdsl/dialects/csl/csl_stencil.py
n-io: I was hoping it would be helpful to preserve the semantics of the `stencil.apply`, since it essentially splits its body into two parts. There is also some implied interaction between the two regions, where the result yielded by the one is an input arg to the other. Finally, the whole thing would likely get lowered to one API call (with two callbacks), so I was wondering if it'd just be easier. I honestly hadn't actually considered splitting up the op entirely.

xdsl/dialects/csl/csl_stencil.py
n-io: Ok

AntonLydike (approved):

tests/dialects/test_csl_stencil.py
```diff
@@ -0,0 +1,56 @@
...
+            AccessOp(t0, IndexAttr.get(0, y), tens_t)
+
+        AccessOp(t1, IndexAttr.get(1, 1), tens_t)
+        AccessOp(t1, IndexAttr.get(-1, -1), tens_t)
+
+    @Builder.implicit_region((temp_t, temp_t))
+    def region1(args: tuple[SSAValue, ...]):
+        t0, t1 = args
+        AccessOp(t0, IndexAttr.get(0, 0), tens_t)
+        AccessOp(t1, IndexAttr.get(0, 0), tens_t)
```
AntonLydike, line 29: What happens with the accesses if this doesn't just access (0,0)?

n-io: Since the `get_accesses` function loops over
```
for arg in self.chunk_reduce.block.args + self.post_process.block.args:
```
there should in theory be no difference between whether a particular access is in the first or second region, so this is merely testing that this works, without re-doing the more in-depth testing for the first region.

PR #2784, opened 2024-06-26
dialects: (riscv) Add HasInsntrait and implement trait for some operations

AntonLydike:
In order to make progress on #2468 I factored out the insn representation. This will hopefully also make it usable for the RISC-V backend efforts.

### DIFF
--- a/tests/dialects/test_riscv_snitch.py
+++ b/tests/dialects/test_riscv_snitch.py
@@ -32,4 +32,10 @@ ground_truth = {
 def test_insn_repr(op: RISCVInstruction):
     trait = op.get_trait(HasInsnRepresentation)
     assert trait is not None
-    assert trait.get_insn(op) == ground_truth[op.name[13:]]
+    # Limitation of Pyright, see https://github.com/microsoft/pyright/issues/7105
+    # We are currently stuck on an older version of Pyright, the update is
+    # tracked in https://github.com/xdslproject/xdsl/issues/2791
+    assert (
+        trait.get_insn(op)  # pyright: ignore[reportGeneralTypeIssues]
+        == ground_truth[op.name[13:]]
+    )
--- a/tests/dialects/test_riscv_snitch.py
+++ b/tests/dialects/test_riscv_snitch.py
@@ -1,0 +1,35 @@
+import pytest
+
+from xdsl.dialects import riscv_snitch
+from xdsl.dialects.riscv import RISCVInstruction
+from xdsl.traits import HasInsnRepresentation
+
+ground_truth = {
+    "dmsrc": ".insn r 0x2b, 0, 0, x0, {0}, {1}",
+    "dmdst": ".insn r 0x2b, 0, 1, x0, {0}, {1}",
+    "dmcpyi": ".insn r 0x2b, 0, 2, {0}, {1}, {2}",
+    "dmcpy": ".insn r 0x2b, 0, 3, {0}, {1}, {2}",
+    "dmstati": ".insn r 0x2b, 0, 4, {0}, {1}, {2}",
+    "dmstat": ".insn r 0x2b, 0, 5, {0}, {1}, {2}",
+    "dmstr": ".insn r 0x2b, 0, 6, x0, {0}, {1}",
+    "dmrep": ".insn r 0x2b, 0, 7, x0, {0}, x0",
+}
+
+
+@pytest.mark.parametrize(
+    "op",
+    (
+        riscv_snitch.DMSourceOp,
+        riscv_snitch.DMDestinationOp,
+        riscv_snitch.DMCopyImmOp,
+        riscv_snitch.DMCopyOp,
+        riscv_snitch.DMStatImmOp,
+        riscv_snitch.DMStatOp,
+        riscv_snitch.DMStrideOp,
+        riscv_snitch.DMRepOp,
+    ),
+)
+def test_insn_repr(op: RISCVInstruction):
+    trait = op.get_trait(HasInsnRepresentation)
+    assert trait is not None
+    assert trait.get_insn(op) == ground_truth[op.name[13:]]
--- a/xdsl/backend/riscv/traits.py
+++ b/xdsl/backend/riscv/traits.py
@@ -1,0 +1,21 @@
+from xdsl.ir import Operation
+from xdsl.traits import HasInsnsRepresentation
+
+
+class StaticInsnsRepresentation(HasInsnsRepresentation):
+    """
+    Returns the first parameter as an insn template string.
+
+    See https://sourceware.org/binutils/docs/as/RISC_002dV_002dDirectives.html for more information
+    """
+
+    def get_insn(self, op: Operation) -> str:
+        """
+        Return the insns representation of the operation for printing.
+        """
+        insn_str = self.parameters
+        if not isinstance(insn_str, str):
+            raise ValueError(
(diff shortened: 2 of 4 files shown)


jorendumoulin:
If these functions are still implemented as risc-v ops, does that not undo the progress of pr #2784 ? As these ops operate on riscv-register operands in comparison to just i32 or index variables, we will have to do an `arith-to-riscv` pass anyway or insert a lot of `UnrealizedConversionCastOps` instead. Why not implement them as operations in the `snrt` dialect (where the operands are not risc-v registers) where this trait could be used?

AntonLydike:
> If these functions are still implemented as risc-v ops, does that not undo the progress of pr #2784 ?

I think you linked to the wrong PR here.

> As these ops operate on riscv-register operands in comparison to just i32 or index variables, we will have to do an `arith-to-riscv` pass anyway or insert a lot of `UnrealizedConversionCastOps` instead. 

I don't know if I can quite follow. The riscv backend of @superlopuh is already able to convert `arith` to `riscv` dialect, but for the snax work we want to go through the `llvm` dialect, to which arith can also already be lowered.

>Why not implement them as operations in the `snrt` dialect (where the operands are not risc-v registers) where this trait could be used?

Because snrt dialect ops map to multiple individual instructions, making it impractical to carry this information imo.

With this plan we gain both:

- A more flexible conversion pattern that converts `riscv-to-llvm`
- The ability to convert riscv operations to `.insn`s further down the line in the riscv backend, so that that project can move away from a custom riscv toolchain as well.

xdsl/dialects/riscv_snitch.py
```diff
@@ -462,6 +463,8 @@ class DMSourceOp(IRDLOperation, RISCVInstruction):
     ptrlo = operand_def(riscv.IntRegisterType)
     ptrhi = operand_def(riscv.IntRegisterType)
 
+    traits = frozenset([StaticInsnsRepresentation(".insn r 0x2b, 0, 0, x0, {0}, {1}")])
```
superlopuh, line 466: What would be the best way to test this?

xdsl/backend/riscv/traits.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+
+    def get_insn(self, op: Operation) -> str:
+        """
+        Return the insns representation of the operation for printing.
+        """
+        insn_str = self.parameters
+        if not isinstance(insn_str, str):
+            raise ValueError(
+                "Paramter of StaticInsnsRepresentation must be the insn string"
```
jorendumoulin, line 19:
```suggestion
                "Parameter of StaticInsnsRepresentation must be the insn string"
```

xdsl/backend/riscv/traits.py
```diff
@@ -0,0 +1,21 @@
+from xdsl.ir import Operation
+from xdsl.traits import HasInsnsRepresentation
+
+
+class StaticInsnsRepresentation(HasInsnsRepresentation):
```
jorendumoulin, line 5:
```suggestion
class StaticInsnRepresentation(HasInsnRepresentation):
```
not really important but having it without the s should make the usage more consistent

xdsl/traits.py
```diff
@@ -498,3 +498,21 @@ def has_effects(cls, op: Operation) -> bool:
 
 class Pure(NoMemoryEffect):
     """A trait that signals that an operation has no side effects."""
+
+
+class HasInsnsRepresentation(OpTrait, abc.ABC):
+    """
+    A trait providing information on how to encode an operation using a .insns assember directive.
```
jorendumoulin, line 505:
```suggestion
    A trait providing information on how to encode an operation using a .insn assember directive.
```

xdsl/traits.py
```diff
@@ -498,3 +498,21 @@ def has_effects(cls, op: Operation) -> bool:
...
+    The returned string contains python string.format placeholders where formatted operands are inserted during
+    printing.
+
+    See https://sourceware.org/binutils/docs/as/RISC_002dV_002dDirectives.html for more information.
+    """
+
+    @abc.abstractmethod
+    def get_insn(self, op: Operation) -> str:
+        """
+        Return the insns representation of the operation for printing.
```
jorendumoulin, line 516:
```suggestion
        Return the insn representation of the operation for printing.
```

jorendumoulin (approved):
Okay, here again I wast just a bit worried about the unrealized conversian cast resolution

xdsl/dialects/riscv_snitch.py
jorendumoulin: I have tested this in verilator simulations, and once this is in we can convert the tests in ci to use this system over at `snax-mlir`. 
Without simulations, I'm not sure.

superlopuh: Here a pytest test that checks that these operations have the exact string you would expect would be good, just to make sure that something in the formatting infrastructure doesn't break, for example

tests/dialects/test_riscv_snitch.py
```diff
@@ -0,0 +1,35 @@
...
+        riscv_snitch.DMStatImmOp,
+        riscv_snitch.DMStatOp,
+        riscv_snitch.DMStrideOp,
+        riscv_snitch.DMRepOp,
+    ),
+)
+def test_insn_repr(op: RISCVInstruction):
+    trait = op.get_trait(HasInsnRepresentation)
+    assert trait is not None
+    assert trait.get_insn(op) == ground_truth[op.name[13:]]
```
AntonLydike, line 35: Pyright complains that: `Method "get_insn" cannot be called because it is abstract and unimplemented (reportGeneralTypeIssues)`

But `get_insn` is an abstract method on `HasInsnRepresentation`, which itself is ABC, so I can never get it "not implemented" as I will always get a subclass of this trait (that by definition has to implement the method). How the heck am I supposed to fix this? @superlopuh @math-fehr

xdsl/traits.py
```diff
@@ -498,3 +498,21 @@ def has_effects(cls, op: Operation) -> bool:
...
+    """
+    A trait providing information on how to encode an operation using a .insn assember directive.
+
+    The returned string contains python string.format placeholders where formatted operands are inserted during
+    printing.
+
+    See https://sourceware.org/binutils/docs/as/RISC_002dV_002dDirectives.html for more information.
+    """
+
+    @abc.abstractmethod
```
n-io, line 513: What happens when you remove this? Since you provide an implementation, does it actually need the decorator?

superlopuh: Pyright helpfully warns users if a method with this decorator has not been implemented

tests/dialects/test_riscv_snitch.py
superlopuh: I think it's because the `get_insn` type parameter is invariant, meaning it says that it returns only instances of this type exactly, not subclasses. I think we want the covariant version instead.

xdsl/traits.py
n-io: With an `@abstractmethod` you only need to say `...` instead of a function body. Since the method throws a `NotImplementedError`, my suspicion is one, that we probably don't need both, and two, that it might potentially make the other error go away if we lose the decorator.

tests/dialects/test_riscv_snitch.py
superlopuh: Turns out I was wrong, it was indeed a Pyright bug

PR #2786, opened 2024-06-26
dialects: (riscv, x86) do not print empty <> for unallocated registers

superlopuh:
I think this removes quite a bit of noise when reading the IR.

Note that the code changes are the three bottom files at the bottom of the files view, the rest is filecheck changes.


AntonLydike (approved):
Damn, this does look much cleaner.

PR #2787, opened 2024-06-26
documentation: allow marimo notebooks as documentation

superlopuh:
I like these a lot for local development, and have struggled for a while with getting them to fail properly when the xDSL code updates. This PR introduces both the a sample notebook, but also a new test script that just runs all the notebooks in the marimo folder in headless mode, failing on error. This should give us plenty of coverage and confidence in the notebooks working properly.


pyproject.toml
```diff
@@ -26,6 +26,7 @@ dev = [
     "nbval<0.12",
     "filecheck<0.0.25",
     "lit<19.0.0",
+    "marimo==0.6.10",
```
superlopuh, line 29: note that this is a new dev dependency

docs/marimo/IR_Chapter_0_Attributes.py
```diff
@@ -0,0 +1,33 @@
...
+    For example, `{IntAttr.__name__}` is an attribute representing an integer.
+    """)
+    return IntAttr,
+
+
+@app.cell
+def __():
+    raise Exception("a")
+    return
+
```
superlopuh, line 30: test code to make sure the CI fails as expected

superlopuh:
OK it failed in the way I expected!

docs/marimo/IR_Chapter_0_Attributes.py
superlopuh: It did fail as expected

math-fehr (commented):
I think it's fine to have some marimo files in some projects (like the riscv project for instance), and having way to test them in the CI.

But otherwise, I feel like having both jupyter notebooks and marimo in our documentation just make things more complex for users to see the documentation no?
I'm just saying this because it seems you are writing a book using marimo, and I'm not sure this is the right approach. If this is just a test file, then I would rename it accordingly though, but would not keep it in `docs`

superlopuh:
I would propose migrating all the jupyter notebooks. I actually started with Toy, but since that imports the `toy` module by being in the same folder, that one might be a little difficult. I'd say that all the other notebooks I would rather migrate to the marimo folder.

I'm not sure exactly what you're proposing for this PR, though, would you like me to replace the placeholder notebook with one with more interesting content?

superlopuh:
My attempt at swapping the notebook for a more interesting one was not successful. I have a small collection of notebooks locally that all require some modifications to the code that I've not merged yet, so I don't actually have a more compelling notebook to add immediately. I'd still like to merge the infrastructure to pave the way for these notebooks being added in the future, after I merge all the dependent changes.

tobiasgrosser (approved):
I find this is a perfect patch. LGTM.

Makefile
```diff
@@ -51,8 +51,15 @@ pytest-toy:
 
 tests-toy: filecheck-toy pytest-toy
 
+tests-marimo:
```
AntonLydike, line 54: Should this be marked PHONY?

superlopuh: defo

PR #2789, opened 2024-06-27
misc: Small refactor to `PatternRewriter.replace_op`

alexarice:
Slight code cleanup for the replace_op function, removing duplicate check of the type of `new_ops`, and making a couple of things more pythony

### DIFF
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -193,16 +193,15 @@ class PatternRewriter(PatternRewriterListener):
         Otherwise, replace its uses with ErasedSSAValue.
         """
         self.has_done_action = True
+
         if isinstance(new_ops, Operation):
-            new_ops = [new_ops]
+            new_ops = (new_ops,)
 
         # First, insert the new operations before the matched operation
         self.insert_op(new_ops, InsertPoint.before(op))
 
-        if isinstance(new_ops, Operation):
-            new_ops = [new_ops]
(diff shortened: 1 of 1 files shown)


math-fehr (approved):

xdsl/pattern_rewriter.py
```diff
@@ -193,16 +193,15 @@ def replace_op(
...
+            new_ops = (new_ops,)
 
         # First, insert the new operations before the matched operation
         self.insert_op(new_ops, InsertPoint.before(op))
 
-        if isinstance(new_ops, Operation):
-            new_ops = [new_ops]
         if new_results is None:
-            new_results = [] if len(new_ops) == 0 else new_ops[-1].results
+            new_results = [] if not new_ops else new_ops[-1].results
```
superlopuh, line 204: supermeganit:
```suggestion
            new_results = new_ops[-1].results if new_ops else []
```

superlopuh (approved):

xdsl/pattern_rewriter.py
georgebisbas: good nit, agree

georgebisbas (approved):

alexarice:
One small thing, the `self.has_done_action = True` line is unnecessary as it is also done in `self.insert_op` but I guessed that this is preferable to leave in anyway just so it's clear

georgebisbas (approved):

PR #2790, opened 2024-06-27
core: Implement block dominance

PapyChacal:
Thought I wanted to use the concept, realized I do not right now, but figured I might as well open the PR.

Free bikeshedding:

I like rethinking vocabulary that I find a bit outdatedly heavy on the tone, like how we finally call the default branch `main` rather then `master`. Anyone has ideas on a name sounding right but lighter than "dominance"?

### DIFF
--- a/xdsl/irdl/dominance.py
+++ b/xdsl/irdl/dominance.py
@@ -2,29 +2,47 @@ from xdsl.ir import Block, Region
 
 
 class DominanceInfo:
+    """
+    Computes and exposes the dominance relation amongst blocks of a region.
+
+    https://en.wikipedia.org/w/index.php?title=Dominator_(graph_theory)&oldid=1189814332
+    """
+
     _dominfo: dict[Block, set[Block]]
 
     def __init__(self, region: Region):
+        """
+        Compute (improper) dominance.
+
+        https://en.wikipedia.org/w/index.php?title=Dominator_(graph_theory)&oldid=1189814332
+        """
+
         self._dominfo = {}
+
+        # No block, no work
         if not (region.blocks):
             return
 
+        # Build the preceding relationship
         pred: dict[Block, set[Block]] = {}
-
         for b in region.blocks:
             pred[b] = set()
-
         for b in region.blocks:
             if b.last_op is not None:
                 for s in b.last_op.successors:
                     pred[s].add(b)
 
+        # Get entry and other blocks
         entry, *blocks = region.blocks
 
+        # The entry block is only dominated by itself
         self._dominfo[entry] = {entry}
 
+        # Instantiate other blocks dominators to all blocks
         for b in blocks:
             self._dominfo[b] = set(region.blocks)
+
+        # Iteratively filter out dominators until it converges
         changed = True
         while changed:
             changed = False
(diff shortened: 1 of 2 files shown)


xdsl/irdl/dominance.py
```diff
@@ -0,0 +1,91 @@
...
+        return False
+    if a.parent is None:
+        raise ValueError("Block `a` has no parent region")
+    if a.parent is not b.parent:
+        raise ValueError("Blocks `a` and `b` are not in the same region")
+
+    return DominanceInfo(a.parent).properly_dominates(a, b)
+
+
+def properly_dominates(a: Block, b: Block) -> bool:
```
math-fehr, line 90: Why not renaming the private function to this? Not sure if I'm missing something.
Otherwise, can you add the documentation here as well?

xdsl/irdl/dominance.py
```diff
@@ -0,0 +1,91 @@
...
+    def dominates(self, a: Block, b: Block) -> bool:
+        """
+        Return if `a` dominates `b`.
+        """
+        return a in self._dominfo[b]
+
+
+def _properly_dominates_block(a: Block, b: Block) -> bool:
+    """
+    Returns true if block `a` properly dominates block `b`, assuming they are in the
```
math-fehr, line 77: Can you explain here what `properly` dominates here?

math-fehr (approved):
Nice, thanks a lot ;)

xdsl/irdl/dominance.py
```diff
@@ -0,0 +1,91 @@
...
+                oldie = self._dominfo[b].copy()
+                self._dominfo[b] = {b} | (
+                    set[Block].intersection(*(self._dominfo[p] for p in pred[b]))
+                    if pred[b]
+                    else set()
+                )
+                if oldie != self._dominfo[b]:
+                    changed = True
+
+    def properly_dominates(self, a: Block, b: Block) -> bool:
```
superlopuh, line 59:
```suggestion
    def strictly_dominates(self, a: Block, b: Block) -> bool:
```
I think strict is the technical term here, for the transitive closure of a relation that doesn't include the source

xdsl/irdl/dominance.py
superlopuh: `strictly` would make it clearer

superlopuh (approved):

xdsl/irdl/dominance.py
PapyChacal: Yeah but MLIR uses properly; do you prefer strictly?

superlopuh: It's just a bit more natural for me, it's like strict relations (https://en.wikipedia.org/wiki/Partially_ordered_set), it immediately implies non-reflexivity. Maybe there's a term of art for dominance that would overrule this if all the dominance experts will get confused by an unfamiliar term :)

xdsl/irdl/dominance.py
```diff
@@ -0,0 +1,91 @@
...
+    _dominfo: dict[Block, set[Block]]
+
+    def __init__(self, region: Region):
+        """
+        Compute (improper) dominance.
+
+        https://en.wikipedia.org/w/index.php?title=Dominator_(graph_theory)&oldid=1189814332
+        """
+
+        self._dominfo = {}
```
georgebisbas, line 20: self._dominance?

georgebisbas (approved):

xdsl/irdl/dominance.py
PapyChacal: The public one ise supposed to be overridden to also support (Operation, Operation) and (Value, Operation) arguments :slightly_smiling_face: This is half the work because I realized I don't need it right now, but this function's purpose is to become the overriden one, to dispatch to implementations/boilerplate accordingly.

xdsl/irdl/dominance.py
```diff
@@ -0,0 +1,91 @@
...
+        # Instantiate other blocks dominators to all blocks
+        for b in blocks:
+            self._dominfo[b] = set(region.blocks)
+
+        # Iteratively filter out dominators until it converges
+        changed = True
+        while changed:
+            changed = False
+            for b in blocks:
+                oldie = self._dominfo[b].copy()
```
georgebisbas, line 50: `old` ?

xdsl/irdl/dominance.py
PapyChacal: Sure, will use strict, this is just picking who gets confused though :wink: 
I like choosing the mathematical consensus too, though :+1:

PR #2793, opened 2024-06-27
core: Implement EffectKind

PapyChacal:
Implement EffectKind, sepcifying READ, WRITE, ALLOC or FREE side-effects.

Use it to enable our MemoryEffect to return this more precise information, and leverage it in DCE (an unused read-only effect is safe to erase) and CSE (A duplicate read-only is safe to replace with an earlier one if we can ensure there was no write in-between)

MLIR further enables to express side-effects tied to specific resources and values, but this a little step forward.


xdsl/traits.py
```diff
@@ -444,30 +445,111 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
         raise NotImplementedError()
 
 
+class EffectKind(Enum):
```
superlopuh, line 448: These seem memory-specific
```suggestion
class MemoryEffectKind(Enum):
```

xdsl/transforms/common_subexpression_elimination.py
```diff
@@ -92,6 +99,23 @@ def pop(self, k: Operation):
...
 
 
+def has_other_side_effecting_op_in_between(
+    from_op: Operation, to_op: Operation
+) -> bool:
+    """
+    Returns if there *may* be a 'write' effecting operation between `from_op` and
+    `to_op`.
+    """
+    assert from_op.parent_block() == to_op.parent_block()
```
superlopuh, line 109: nit:
```suggestion
    assert from_op.parent is to_op.parent
```

zero9178 (commented):
PR and code LGTM and matches MLIR with the more limited feature set (any reason this is still a draft acutally? I'd stamp otherwise).

One thing worth considering at your discretion:
One annoyance I've always had with this interface in MLIR is that if an Op implemented the trait/interface, it was impossible for the op to dynamically (based on attributes e.g.) be able to say "I am side effect free" or "I affect the universe" (the same as an op not implementing the interface basically). This is because both in this PR and in MLIR, returning an empty set means "no side effect". 
If it instead returned a `set | None` (or similar) then we could encode `None` to mean the same as a missing trait implementation (effects the universe). 

A common ues case for this would be a call op e.g. The call op may be annotated with a `pure` attribute and in that case you want it to not have any side effects. If it doesn't, you'll want it to affect the universe

Thought I'd mention this now so that if you do want it, that you don't have to go through the refacotring churn twice.

PapyChacal:
Interesting point, thanks!! I'll give it some thought and would love to discuss it, as I'm lass experienced on that side of things. My first thought is to ask if that option isn't still limited. It still means arbitrary but finite effects of some kind, or absolutely everything, right?

What about reading the universe while writing to a defined part? Genuinely wondering as I'm not sure it would either represent something practical or enable more powerful anything :thinking:

zero9178:
> It still means arbitrary but finite effects of some kind, or absolutely everything, right?

It generally means absolutely everything yes. The most conservative answer.

> What about reading the universe while writing to a defined part? Genuinely wondering as I'm not sure it would either represent something practical or enable more powerful anything 🤔

Good question. This is currently unimplemented in MLIR as well as there is no special value for the universe right now. I've seen this come up from time to time in discussions but haven't seen anyone actually try to implement it. MLIR does have the concept of `Resources` on side effects but these represent distinct aka non-aliasing regions. Ideas to implement this would be making a universe resource and/or allowing resources to specify the aliasing relationship between them.

PapyChacal:
> > What about reading the universe while writing to a defined part? Genuinely wondering as I'm not sure it would either represent something practical or enable more powerful anything 🤔
> 
> Good question. This is currently unimplemented in MLIR as well as there is no special value for the universe right now. I've seen this come up from time to time in discussions but haven't seen anyone actually try to implement it. MLIR does have the concept of `Resources` on side effects but these represent distinct aka non-aliasing regions. Ideas to implement this would be making a universe resource and/or allowing resources to specify the aliasing relationship between them.

Right, I guess the best end-game answer is building an aliasing framework for resources indeed, having a Universe defined as aliasing everything.
But yeah, interesting questions in themselves, but in practice, I'm trying to get basics running in xDSL here rather than exploring that side of things; so I think I'll just go with the easy None escape-hatch here and stay out of the rabbit hole myself :rabbit: 

Thanks a lot for the discussion though, didn't cross my mind yet!

xdsl/dialects/test.py
```diff
@@ -153,6 +153,92 @@ def __init__(
         )
 
 
+@irdl_op_definition
```
math-fehr, line 156: Big fan of adding new operations to the test dialect

math-fehr (approved):
Really cool! I will not lie, I did not try to understand the entire algorithm, I just trust you there ;)

xdsl/dialects/test.py
PapyChacal: That - and the algorithms - is all copy-pasta-porta from MLIR implementations :yum:

xdsl/traits.py
```diff
@@ -444,30 +445,111 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
+        """
         raise NotImplementedError()
 
+    @final
+    @classmethod
+    def has_effects(cls, op: Operation) -> bool:
+        """
+        Returns if the operation has any side effects.
+        """
+        return len(cls.get_effects(op)) > 0
```
superlopuh, line 499:
```suggestion
        return cls.get_effects(op)
```

xdsl/traits.py
```diff
@@ -444,30 +445,111 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
 
-def is_side_effect_free(op: Operation):
+    @final
+    @classmethod
+    def only_has_effect(cls, op: Operation, effect: MemoryEffectKind) -> bool:
+        """
+        Returns if the operation has the given side effects and no others.
+        """
+        return cls.get_effects(op) == {effect}
```
superlopuh, line 507: nit: is this helper necessary? It would probably be clearer to write out the implementation in the client code. This might also apply to `has_effects`

xdsl/traits.py
```diff
@@ -444,30 +445,111 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
+            return False
+        # If it said it has no effect, and does not have recursive effects, done
+        if not op.get_trait(RecursiveMemoryEffect):
+            return True
+    # If it implements neither trait, play safe and assume it has effects
+    elif not op.get_trait(RecursiveMemoryEffect):
         return False
-    return not trait.has_effects(op)
+
+    # If it only has recursive effects, well, recurse
```
superlopuh, line 525: nit:
```suggestion
    # Recurse if it only has recursive effects
```

xdsl/traits.py
```diff
@@ -444,30 +445,111 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
-    if not (trait := op.get_trait(MemoryEffect)):
+    if trait := op.get_trait(MemoryEffect):
+        # If the op implements the trait and says it has effects, done
+        if trait.has_effects(op):
+            return False
+        # If it said it has no effect, and does not have recursive effects, done
+        if not op.get_trait(RecursiveMemoryEffect):
+            return True
+    # If it implements neither trait, play safe and assume it has effects
+    elif not op.get_trait(RecursiveMemoryEffect):
```
superlopuh, line 522: nit: I would swap around this elif and the fallthrough, or add a final else, to make it a tiny bit clearer that the deafault behaviour is `return False`, and the only way to return True is with one of the two effect traits

xdsl/transforms/common_subexpression_elimination.py
```diff
@@ -92,6 +99,23 @@ def pop(self, k: Operation):
...
+    `to_op`.
+    """
+    assert from_op.parent is to_op.parent
+    next_op: Operation | None = from_op
+    while next_op and (next_op is not to_op):
+        effects = get_side_effects_recursively(next_op)
+        if effects is None or (MemoryEffectKind.WRITE in effects):
+            return True
+        next_op = next_op.next_op
+    return False
```
superlopuh, line 116: nit:
```suggestion
    next_op = from_op
    while next_op is not to_op:
        effects = get_side_effects_recursively(next_op)
        if effects is None or (MemoryEffectKind.WRITE in effects):
            return True
        next_op = next_op.next_op
        assert next_op is not None, "Incorrect order of ops in side-effect search"
    return False
```
since from_op is before to_op, there should never be a None, right?

xdsl/transforms/dead_code_elimination.py
```diff
@@ -3,19 +3,58 @@
...
-        and all(not result.uses for result in op.results)
+        all(not result.uses for result in op.results)
+        and (not op.get_trait(IsTerminator))
+        and (not op.get_trait(SymbolOpInterface))
+        and would_be_trivially_dead(op)
     )
 
 
+def would_be_trivially_dead(rootOp: Operation) -> bool:
+    effecting_ops = {rootOp}
```
superlopuh, line 26: Please add docstring, and maybe change the name. Is it lifted from MLIR? I'm not sure what the "would be" means

xdsl/dialects/riscv.py
```diff
@@ -2809,11 +2810,19 @@ class RegisterAllocatedMemoryEffect(MemoryEffect):
     """
 
     @classmethod
-    def has_effects(cls, op: Operation) -> bool:
-        return any(
+    def get_effects(cls, op: Operation) -> set[MemoryEffectKind]:
```
AntonLydike, line 2813: I am wondering if it makes sense to have MemoryEffectKind be a [Flag](https://docs.python.org/3/library/enum.html#enum.Flag) instead?

xdsl/traits.py
```diff
@@ -444,30 +445,111 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
+    """
+    Indicates that the operation allocates from some resource. An 'allocate' effect
+    implies only allocation of the resource, and not any visible mutation or dereference.
+    """
+    FREE = auto()
+    """
+    Indicates that the operation frees some resource that has been allocated. A 'free'
+    effect implies only de-allocation of the resource, and not any visible allocation,
+    mutation or dereference.
+    """
```
AntonLydike, line 477: nit: spacing

AntonLydike (commented):
I am very happy to see this make its way into MLIR.

Right now we have a `Read` and a `Write` effect trait. I am a bit concerned that this does not compose well, as right now we ignore all but the first `EffectTrait` on an operation.

If we have an operation that has both the read and write trait, I would expect it to be marked both reading and writing. We should be able to fix that though by just getting all traits of the type instead of the first one.

xdsl/traits.py
PapyChacal: No it's not necessary, but it's supposed to get more complex soon. I'd rather keep it than refactor out all copy pasted code soon - if you insist, I can remove it

xdsl/transforms/dead_code_elimination.py
PapyChacal: It is lifted from MLIR.

xdsl/transforms/common_subexpression_elimination.py
PapyChacal: Right, lifted from MLIR

I think point is, this is an internal helper, not mean to be robust against user input but just avoid the extra hot loop test inside.

I don't mind adding it

xdsl/traits.py
PapyChacal: Lifted from MLIR, I like to keep it close to what they do, because I'm only porting subsets and it makes it easier to filter out cases and add them when I add the necessary boilerplate.

Do you mind if we keep it as is until it is at least on par? Again, I'm just porting stuff, not exploring ideas on *how* best to do it or what not.

xdsl/traits.py
PapyChacal: See above discussion with Markus; will probably change the return of get_effects to None | set so postponing this one

xdsl/traits.py
PapyChacal: What ?

xdsl/dialects/riscv.py
PapyChacal: No I don't think it does, this kind of effect is one aspect of an effect, a next stage equips them with different resources or effected values to make up an effect and I don't think having a Flag composed in there would make anything simpler.

Otherwise, similar comment to other comments from Sasha; I'm just partially porting stuff from MLIR here; I'd rather have it as similar as possible at least until it's on par; I don't see the value of nitting on those details at the cost of having the port be error-prone on the behaviour first.

xdsl/transforms/dead_code_elimination.py
superlopuh: Then a docstring would_be really helpful :)

xdsl/transforms/common_subexpression_elimination.py
superlopuh: your call, I think it's ok as-is

xdsl/traits.py
superlopuh: this is fine!

xdsl/traits.py
superlopuh: no, it's ok

xdsl/traits.py
superlopuh: kk

xdsl/transforms/common_subexpression_elimination.py
PapyChacal: Will add more comments

More generally, I'll also try to just add links to the MLIR implementation and mention inline the intent where I'm not even trying to be Pythonic-first.

xdsl/dialects/riscv.py
PapyChacal: Same addition as on sasha's discussion, will mention these points inline in places

PapyChacal:
> I am very happy to see this make its way into MLIR.

Do you mean in xDSL? This is relatively old behaviour in MLIR, that I'm just porting as literally as possible. I'll try to include explicit pointers to the MLIR side in the code, as in many places I purposefully decide to keep it very similar to keep the progressive porting easy, and not a full-on exploration-style track - I just want this existing behaviour available to xDSL dialects.

> Right now we have a `Read` and a `Write` effect trait. I am a bit concerned that this does not compose well, as right now we ignore all but the first `EffectTrait` on an operation.

> If we have an operation that has both the read and write trait, I would expect it to be marked both reading and writing. We should be able to fix that though by just getting all traits of the type instead of the first one.

Note that those traits you mentioned are commented by "A trait that signals that an operation always and *only* have write(resp. read) side effects.", emphasize *only*: they say it. The intent is indeed to implement just one, and handle composition there.
We could provide another, parametrized, helper that says "This operation always and only have *those* effects" ?

I think it's sugar though, I'd like to keep that separated from the actual base feature if possible? You can already define such a subclass yourself. Point here is to define the base class API and handling in core features. Let's work on user-friendliness as a second step?

I hear this style is not very xDSL-y and could consider composing different MemoryEffect traits on operations. I'm worried it would complexify things in remote places, but if it's of limited footprint, we could just handle the union of different MemoryEffect implementations on each operation.

AntonLydike:
> Do you mean in xDSL?
Yep, slipped up^^


> > Right now we have a `Read` and a `Write` effect trait. I am a bit concerned that this does not compose well, as right now we ignore all but the first `EffectTrait` on an operation.
> 
> > If we have an operation that has both the read and write trait, I would expect it to be marked both reading and writing. We should be able to fix that though by just getting all traits of the type instead of the first one.
> 
> Note that those traits you mentioned are commented by "A trait that signals that an operation always and _only_ have write(resp. read) side effects.", emphasize _only_: they say it. The intent is indeed to implement just one, and handle composition there. We could provide another, parametrized, helper that says "This operation always and only have _those_ effects" ?

Hmm, I think it still is quite unintuitive that it's perfectly fine to have multiple traits attached to the operation but that it just silently fails... In theory we should be able to just union the effects sets returned by all effects traits. right?

> I think it's sugar though, I'd like to keep that separated from the actual base feature if possible? You can already define such a subclass yourself. Point here is to define the base class API and handling in core features. Let's work on user-friendliness as a second step?

I don't know if I necessarily agree here. You already use the classes, and the way you use them results in imo un-intuitive behaviour when the op has multiple traits that signal effects. Doesn't need to be the ones you specifically said need to be unique. IMO the trait interface gives you the expectation of composability, but the implementation ignores it.

PapyChacal:
I hear your point; to clarify mine, by sugar, I meant my own proposal of helper base classes, not the composability, which indeed is a bit more core.

I'll add composability; unless I realize an awful consequence or what not

On the sugar : I think the vast majority of implementations will be "have effect(s) a if b(op)", so a cute wrapper parametrized by an effect and predicate would help a lot I think; especially with composability included!

AntonLydike:
> I hear your point; to clarify mine, by sugar, I meant my own proposal of helper base classes, not the composability, which indeed is a bit more core.

Ah yes, this makes much more sense! 


> I'll add composability; unless I realize an awful consequence or what not

Let's hope it's all fine! :crossed_fingers: 

> On the sugar : I think the vast majority of implementations will be "have effect(s) a if b(op)", so a cute wrapper parametrized by an effect and predicate would help a lot I think; especially with composability included!

Agree 100%. I guess I could see two traits for different kinds of efffects (e.g. `HasReadEffectIf...` and `HasWriteEffectIf...`) that may be mixed and matched by ops. At least this feels like a somewhat natural thing to do.

Looking forward to us being at that stage!

superlopuh (approved):

xdsl/transforms/dead_code_elimination.py
PapyChacal: Done both!

xdsl/traits.py
PapyChacal: Accordingly went with `return bool(cls.get_effects(op))`

`return cls.get_effects(op)` does not implicitely convert

xdsl/traits.py
```diff
@@ -444,30 +445,108 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
     Boilerplate helper to check if a generic operation is side effect free for sure.
     """
-    # If it doesn't say, safely assume it has side effects.
-    if not (trait := op.get_trait(MemoryEffect)):
-        return False
-    return not trait.has_effects(op)
+    return get_side_effects_recursively(op) == set()
+
+
+def get_side_effects_recursively(rootOp: Operation) -> set[MemoryEffectKind] | None:
```
AntonLydike, line 522: I would name this "get_effects".

The recursive checks should be implemented on the `RecursiveMemoryEffect` interface, not here imo.

PapyChacal: Interestingly, implementing RecursiveMemoryEffect as a MemoryEffect is not possible if following the MLIR implementation closely, as RecursiveMemoryEffect needs to be able to answer unconclusively, as it could contain unknown operations.
But adding support for this as recommend by @zero9178 makes it possible, and I prefer too; done

xdsl/traits.py
```diff
@@ -444,30 +445,98 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
+
+def is_side_effect_free(op: Operation) -> bool:
     """
     Boilerplate helper to check if a generic operation is side effect free for sure.
     """
-    # If it doesn't say, safely assume it has side effects.
-    if not (trait := op.get_trait(MemoryEffect)):
-        return False
-    return not trait.has_effects(op)
+    return get_effects(op) == set()
```
superlopuh, line 519: nit:
```suggestion
    effects = get_effects(op)
    return effects is not None and bool(effects)
```
Avoids the empty set construction

superlopuh (approved):

xdsl/traits.py
```diff
@@ -476,28 +545,54 @@ class NoMemoryEffect(MemoryEffect):
...
     @classmethod
-    def has_effects(cls, op: Operation) -> bool:
-        return False
+    def get_effects(cls, op: Operation) -> set[MemoryEffectKind]:
+        return set()
+
+
+class MemoryReadEffect(MemoryEffect):
+    """
+    A trait that signals that an operation always and only have read side effects.
```
AntonLydike, line 554: nit: we can drop the "and only" here.

```suggestion
    A trait that signals that an operation always have read side effects.
```

xdsl/traits.py
```diff
@@ -476,28 +545,54 @@ class NoMemoryEffect(MemoryEffect):
...
+    """
+
+    @classmethod
+    def get_effects(cls, op: Operation) -> set[MemoryEffectKind]:
+        return {MemoryEffectKind.READ}
+
+
+class MemoryWriteEffect(MemoryEffect):
+    """
+    A trait that signals that an operation always and only have write side effects.
```
AntonLydike, line 564:
```suggestion
    A trait that signals that an operation always have write side effects.
```
same

xdsl/transforms/dead_code_elimination.py
```diff
@@ -3,19 +3,37 @@
...
+        and result_only_effects(op)
     )
 
 
+def result_only_effects(rootOp: Operation) -> bool:
+    """
+    Returns if we can ensure the operation would have no observable effect beyond its
+    returned values.
+
+    cf MLIR's WouldOpBeTriviallyDead
```
AntonLydike, line 31: nit: maybe link docs for quick reference, but not necessary

AntonLydike (approved):
Nice :rocket:

xdsl/traits.py
PapyChacal: Is Python really not optimizing this?
Your proposal is off but I can get it to a similar form if you really want to.

I'm wondering if this is not the kind of things we should just keep readable and just hope the most basic optimization gets rid of though?

xdsl/traits.py
superlopuh: "has" instead of "have", probably also

xdsl/traits.py
```diff
@@ -444,30 +445,98 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
-    if not (trait := op.get_trait(MemoryEffect)):
-        return False
-    return not trait.has_effects(op)
+    return get_effects(op) == set()
+
+
+def get_effects(op: Operation) -> set[MemoryEffectKind] | None:
+    """
+    Helper to get known side effects of an operation, including recursive effects.
+    None means that the operation has unknown effects, for safety.
```
webmiche, line 525: I think None here is a bit weird. would it make sense to add an `unknown` to the memory effect kinds? IMO, this would be much clearer and you could also express that something is guaranteed to have a write effect and maybe others and that something is guaranteed to have a write effect and no others. Currently, this distinction is not possible, AFAIU.

webmiche: but I guess we are copying MLIR here so....

webmiche (commented):
Cool concept, really like it :+1:

xdsl/traits.py
```diff
@@ -444,30 +445,98 @@ def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
...
+    def has_effects(cls, op: Operation) -> bool:
+        """
+        Returns if the operation has any side effects.
+        """
+        effects = cls.get_effects(op)
+        return (effects is not None) and len(effects) > 0
+
 
-def is_side_effect_free(op: Operation):
+def only_has_effect(op: Operation, effect: MemoryEffectKind) -> bool:
```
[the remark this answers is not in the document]
PapyChacal: This is direct copy from MLIR but yeah it bugs me too; I'll give it some thought, as I think what I personally would use is a variant anyway 🤔

xdsl/traits.py
PapyChacal: Actually not on this one ! This is an extension from MLIR already, which only allows to return known effects or not implement the interface at all.

This is a minimal extension code wise, and already allows unifying RecursiveEffect, which otherwise has to be a separate trait (as it could, well, have recursively unknown effects)

If you're fine with it, I would like to keep this probably non-final extension as-is; I am just after porting MLIR functionality here initially, not improve the design fundamentally, if that makes sense

I want to be clear, I think there's a lot of interesting discussions to have on those nuances and not dismissing that. I'd just like to push it out of this initial xDSL PR.

xdsl/traits.py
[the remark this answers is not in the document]
webmiche: I would be fine with having both and one just dispatches to the other

xdsl/traits.py
webmiche: Okay, sure, I am fine with leaving it as-is :+1:

xdsl/traits.py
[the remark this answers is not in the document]
PapyChacal: But to me they mine different things though? If you tell me "It has this exact effect", I do not deduce "It has no other effects", which is what's tested here. Do you? I don't mind adding it if you hear they mean the same thing, I just want to make sure we agree on the named thing

xdsl/transforms/dead_code_elimination.py
AntonLydike: Sure, also feel free to ignore if there is not docs page (it's MLIR after all)

xdsl/traits.py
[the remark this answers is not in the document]
webmiche: yes, I was thinking of the `exact`-version as "has exactly this effect and no other". 

I think it's fine to just leave as-is, I mainly dislike the current name, but it's from MLIR, so our are hands are tied a bit.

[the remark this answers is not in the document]
PapyChacal: Will add a wrapper with the name you like then, it's no big deal 🙂

PR #2794, opened 2024-06-27
core: add RegionBlocks to abstract away the block storage [1/2]

superlopuh:
PR one out of two to migrate blocks of regions to a doubly-linked-list data structure, instead of a list. This is more in line with how MLIR stores things, and may be beneficial in some scenarios, when there are many blocks.

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -232,7 +232,8 @@ class FuncOp(IRDLOperation):
         if len(self.body.blocks) != 1:
             raise VerifyException("Expected FuncOp to contain one block")
 
-        block = self.body.block
+        block = self.body.blocks.first
+        assert block is not None
 
         if not block.ops:
             raise VerifyException("Expected FuncOp to not be empty")
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -232,8 +232,7 @@ class FuncOp(IRDLOperation):
         if len(self.body.blocks) != 1:
             raise VerifyException("Expected FuncOp to contain one block")
 
-        block = self.body.blocks.first
-        assert block is not None
+        block = self.body.blocks[0]
 
         if not block.ops:
             raise VerifyException("Expected FuncOp to not be empty")
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -232,7 +232,7 @@ class FuncOp(IRDLOperation):
         if len(self.body.blocks) != 1:
             raise VerifyException("Expected FuncOp to contain one block")
 
-        block = self.body.blocks[0]
+        block = self.body.block
 
         if not block.ops:
             raise VerifyException("Expected FuncOp to not be empty")
--- a/tests/dialects/test_func.py
+++ b/tests/dialects/test_func.py
@@ -71,8 +71,8 @@ def test_func_II():
     # Use this region to create a func0
     func1 = FuncOp.from_region("func1", [], [], region0)
 
-    ops0 = list(list(func1.regions[0].blocks)[0].ops)
-    ops1 = list(list(func1.regions[0].blocks)[1].ops)
+    ops0 = list(func1.regions[0].blocks[0].ops)
+    ops1 = list(func1.regions[0].blocks[1].ops)
 
     assert len(ops0) == 3
     assert len(ops1) == 3
--- a/tests/dialects/test_func.py
+++ b/tests/dialects/test_func.py
@@ -71,8 +71,8 @@ def test_func_II():
     # Use this region to create a func0
     func1 = FuncOp.from_region("func1", [], [], region0)
 
-    ops0 = list(func1.regions[0].blocks[0].ops)
-    ops1 = list(func1.regions[0].blocks[1].ops)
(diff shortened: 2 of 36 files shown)


xdsl/dialects/riscv_cf.py
```diff
@@ -130,8 +130,10 @@ def verify_(self) -> None:
         if parent_region is None:
             return
 
-        this_index = parent_region.blocks.index(parent_block)
-        else_index = parent_region.blocks.index(self.else_block)
+        # TODO: use next_block to check this instead
```
superlopuh, line 133: https://github.com/xdslproject/xdsl/pull/2795/files#diff-6488d07a29ced65b2e14906fd38c48f7fed25d6997cd210eff83b0fa9b98ec4c

xdsl/printer.py
```diff
@@ -344,8 +343,16 @@ def print_region(
             print_block_args=print_entry_block_args,
             print_block_terminator=print_block_terminators,
         )
-        for block in region.blocks[1:]:
-            self.print_block(block, print_block_terminator=print_block_terminators)
+        # TODO: Replace the explicit iteration below with `next_block` accesses
```
superlopuh, line 346: https://github.com/xdslproject/xdsl/pull/2795/files#diff-31fd89d915ff630760b1c736cd9e78c508627efd7678225c39bbc389b5ea0272

tests/dialects/test_func.py
```diff
@@ -71,8 +71,8 @@ def test_func_II():
     # Use this region to create a func0
     func1 = FuncOp.from_region("func1", [], [], region0)
 
-    ops0 = list(func1.regions[0].blocks[0].ops)
-    ops1 = list(func1.regions[0].blocks[1].ops)
+    ops0 = list(list(func1.regions[0].blocks)[0].ops)
+    ops1 = list(list(func1.regions[0].blocks)[1].ops)
```
AntonLydike, line 75: This seems to be a strict downgrade, can we phrase this differently?

docs/Toy/toy/dialects/toy.py
```diff
@@ -232,7 +232,7 @@ def verify_(self):
         if len(self.body.blocks) != 1:
             raise VerifyException("Expected FuncOp to contain one block")
 
-        block = self.body.blocks[0]
+        block = self.body.block
```
AntonLydike, line 235: Technically this is semantically different, as it will now fail when there are more than one block, while it previously ignored subsequent blocks.

A more correct replacement would be `.first_block`.

AntonLydike (commented):
I don't really see why we would need this change, do we see any performance problems with this?

This seems to complicate block accesses a lot (maybe we are just missing the right helpers here? (`first_block() -> Block` without the `None` maybe?).

I'd like to see some benchmarks showing we are leaving important performance behind here before merging such a change...

docs/Toy/toy/dialects/toy.py
superlopuh: Oh I thought I had gotten rid of all of these, I'll fix this.

tests/dialects/test_func.py
superlopuh: I thought that since this was just a test, I wouldn't make things too weird, but I'm happy to rewrite if you think it's too messy

superlopuh:
Yeah, maybe we could just add an accessor, I'd prefer `assert_first_block` or something like it. Happy to take suggestions.

I ran perf measurements, and it has no measureable impact on performance. The main motivation for me is from an API standpoint. On one hand, this is closer to the structure of MLIR, which seems like a win in itself. On the other, it hides the fact that the blocks of regions is a list, which for me is already a win from an encapsulation standpoint.

superlopuh:
The perf measurements I did were running `time make tests` a couple of times, inserting a block at the start of a region, and inserting a block at the end of the region, all had more noise than performance difference.

superlopuh:
After a bit more thinking about it, I think the non-None helper would promote bad code. When I was making this change, I was stopping to look for clues about whether it was correct, and I'm really not sure that it was most of the time. Hopefully it will lead users to consider whether the first block is guaranteed to exist more carefully.

math-fehr (approved):
I would say that this is only an API issue for now, and not a performance one.
Having `first_block` and `last_block` being optional makes sense to me, the same way we have `first_op` and `last_op` being optional.
It would be nice however to have something like `first_block_unwrap` or something like this that returns the first block and asserts that it exists, just to make the API a bit simpler.
I'm not sure of a better naming though.

superlopuh:
Thinking about it a bit more, I don't think there's a need to be too patronising here and let the users use indices if they want. There's a small chance that the region has a million blocks, and if the user wants to index into the last one using counting I guess that's their problem. I'll push the functionality that reduces the downstream impact, and can revert it if you like it less @math-fehr

xdsl/irdl/operations.py
```diff
@@ -1348,8 +1348,8 @@ def irdl_op_verify_regions(
                 f"Region '{name}' at position {i} expected a single block, but got "
                 f"{len(region.blocks)} blocks"
             )
-        if len(region.blocks) > 0:
-            entry_args_types = tuple(a.type for a in region.blocks[0].args)
+        if (first_block := region.blocks.first) is not None:
```
georgebisbas, line 1351: I think the previous one was better.

PR #2795, opened 2024-06-27
core: store blocks as doubly-linked list on Region [2/2]

superlopuh:
This migrates the blocks to be a doubly linked list.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dev = [
     "lit<19.0.0",
     "marimo==0.6.10",
     "pre-commit==3.7.1",
-    "ruff==0.4.10",
+    "ruff==0.5.0",
     "asv<0.7",
     "isort==5.13.2",
     "nbconvert>=7.7.2,<8.0.0",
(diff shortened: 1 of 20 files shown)


xdsl/rewriter.py
```diff
@@ -283,46 +283,34 @@ def insert_op_before(op: Operation, new_op: Operation):
     def move_region_contents_to_new_regions(region: Region) -> Region:
         """Move the region blocks to a new region."""
         new_region = Region()
-        for block in region.blocks:
```
math-fehr, line 286: Nice

math-fehr (approved):
Makes sense to me, though I think we *really* should abstract away the double linked list.
I'm pretty sure we can abstract a `DoubleLinkedList[T]` class that will contain almost all of the logic, and then with a bit of inheritance we can get back what you have (in both blocks and operations).
Otherwise, I'm fine merging it like it is

superlopuh:
It would be nice to have such an abstraction, but I would say that this is something we want to hide from users, at least the mutability aspect. Let's chat about this at some point.

xdsl/ir/core.py
```diff
@@ -1560,37 +1615,48 @@ def __getitem__(self, idx: int) -> Block: ...
...
+                        return b
+                raise IndexError
+        else:
+            # This is possible but would require a bit of work to handle complex slices
+            raise NotImplementedError("Indexing of RegionBlocks not yet implemented")
 
     def __len__(self):
-        return len(self._blocks)
+        i = 0
+        for _ in self:
```
georgebisbas, line 1635: why this way?

superlopuh: as opposed to enumerating or `while (next := next.next) is not None`?

Issue #2122, opened 2024-02-08
interpreter: (riscv) external function call in riscv_func does not handle register dictionary in RiscvFunctions


superlopuh:
When calling an external function call, there is currently an error of a0 not being set.

compor:
what's the minimal reproduction example? is there a filecheck?
I'd like to tackle how we explicitly use registers for func arguments, so this might be related eventually.

superlopuh:
This is super low priority, it's just the interpreter, I'm not actually sure where the problem is, but it's easy to reproduce in the branch that references this issue

compor:
Yeah, no worries, I'm just doing some issue janitorial work.

PR #2796, opened 2024-06-27
docs: add onnx demo notebook

superlopuh:
Note that this is missing the interpreter part, since the bug is still there with the ABI calling convention not being respected when calling from the interpreter object directly.

### DIFF
--- a/.github/workflows/ci-mlir.yml
+++ b/.github/workflows/ci-mlir.yml
@@ -115,6 +115,8 @@ jobs:
     - name: Test MLIR dependent marimo notebooks
       run: |
         cd xdsl
+        # Add mlir-opt to the path
+        export PATH=$PATH:${GITHUB_WORKSPACE}/llvm-project/build/bin/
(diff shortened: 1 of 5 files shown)


tobiasgrosser:
This looks good to me. At some point we had the interpreter give us performance numbers. Did you drop them deliberately

superlopuh:
Yes, it's because we currently have a bug in the interpreter. I wanted to merge the first part first, to not block on this issue to add a first real notebook:
#2122: interpreter: (riscv) external function call in riscv_func does not handle register dictionary in RiscvFunctions

superlopuh:
The way it works in my local branch is by commenting some code that is needed for other functionality, and I'd like to resolve it properly, by setting the registers specified by the ABI

tobiasgrosser:
Cool. Happy to merge it without the interpreter.

superlopuh:
I'll interpret that as an "approve" :)

PR #2801, opened 2024-06-28
dialects: (wasm) Add wat encoding

superlopuh:


xdsl/xdsl_opt_main.py
```diff
@@ -209,6 +209,19 @@ def _output_riscv_asm(prog: ModuleOp, output: IO[str]):
...
             x86.ops.print_assembly(prog, output)
 
+        def _output_wat(prog: ModuleOp, output: IO[str]):
+            from xdsl.dialects.wasm import WasmModule
+            from xdsl.dialects.wasm.wat import WatPrinter
+
+            wasm_modules = tuple(op for op in prog.walk() if isinstance(op, WasmModule))
+            if len(wasm_modules) != 1:
+                print(f"invalid number of wasm modules ({len(wasm_modules)})")
+                return
```
superlopuh, line 219: I'm not sure if we want this or just print in a loop

PapyChacal: What would be a reason *not* to print in a loop?

superlopuh: I can't think of any, I'll change it

Moxinilian (approved):
Super neat but your simple change makes my next PR even bigger 💀

PR #2803, opened 2024-06-28
(transform): convert to csl_stencil.apply

n-io:
This PR implements the conversion to csl_stencil.apply op as outlined in Step 2 of https://github.com/xdslproject/xdsl/pull/2747

The `csl_stencil.apply` op combines a csl_stencil.prefetch (symmetric buffer communication across a given stencil shape) with a stencil.apply.

The transformation consists of several steps:
* When rewriting a `stencil.apply`, select the `csl_stencil.prefetch` with the biggest memory overhead (if several are present) to be fused with the apply op into a `csl_stencil.apply`
* Find a suitable split of ops to be divided across the two code blocks
  * Re-order arithmetic e.g. `(a+b)+c -> (c+a)+b` to access, consume, and reduce data of neighbours (for input stencil only)
  * Move this into first code block, move all other ops into second code block
  * Fallback strategy: Move everything into first code block
* Add `tensor.InsertSliceOp` to insert computed chunk into returned z-value tensor
* Set up code block args
* Move ops into new regions according to determined split
* Translate arg usage to match new block args
* Set up yield ops
* Run tensor update shape on chunk region

### DIFF
--- a/xdsl/dialects/csl/csl_stencil.py
+++ b/xdsl/dialects/csl/csl_stencil.py
@@ -116,24 +116,24 @@ class PrefetchOp(IRDLOperation):
         stencil.TempType[Attribute] | memref.MemRefType[Attribute]
     )
 
-    swaps = opt_prop_def(builtin.ArrayAttr[ExchangeDeclarationAttr])
+    swaps = prop_def(builtin.ArrayAttr[ExchangeDeclarationAttr])
 
-    topo = opt_prop_def(dmp.RankTopoAttr)
+    topo = prop_def(dmp.RankTopoAttr)
 
     result = result_def(memref.MemRefType)
 
     def __init__(
         self,
         input_stencil: SSAValue | Operation,
-        topo: dmp.RankTopoAttr | None = None,
-        swaps: Sequence[ExchangeDeclarationAttr] | None = None,
+        topo: dmp.RankTopoAttr,
+        swaps: Sequence[ExchangeDeclarationAttr],
         result_type: memref.MemRefType[Attribute] | None = None,
     ):
         super().__init__(
             operands=[input_stencil],
             properties={
                 "topo": topo,
-                "swaps": builtin.ArrayAttr(swaps if swaps else []),
+                "swaps": builtin.ArrayAttr(swaps),
             },
             result_types=[result_type],
         )
--- a/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
+++ b/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
@@ -311,6 +311,10 @@ class AccessOpUpdateShape(RewritePattern):
 
 
 class CslStencilAccessOpUpdateShape(RewritePattern):
+    """
+    Updates the result type of a tensorized `csl_stencil.access` op
+    """
+
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: csl_stencil.AccessOp, rewriter: PatternRewriter, /):
         if typ := get_required_result_type(op):
@@ -387,13 +391,17 @@ class FillOpUpdateShape(RewritePattern):
 
 @dataclass(frozen=True)
 class BackpropagateStencilShapes(ModulePass):
+    """
+    Greedily back-propagates the result types of tensorized ops.
+    Use after creating/modifying tensorization.
+    """
+
     name = "backpropagate-stencil-shapes"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         backpropagate_stencil_shapes = PatternRewriteWalker(
             GreedyRewritePatternApplier(
                 [
-                    # AccessOpUpdateShape(),
                     CslStencilAccessOpUpdateShape(),
                     ExtractSliceOpUpdateShape(),
                     EmptyOpUpdateShape(),
--- a/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
+++ b/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
@@ -4,6 +4,7 @@ from typing import TypeGuard, cast
 from attr import dataclass
 
 from xdsl.context import MLContext
+from xdsl.dialects import builtin
 from xdsl.dialects.arith import (
     Addf,
     BinaryOperation,
@@ -14,6 +15,7 @@ from xdsl.dialects.arith import (
 )
 from xdsl.dialects.builtin import (
     AnyFloat,
+    ArrayAttr,
     ContainerType,
     IntAttr,
     MemRefType,
@@ -21,6 +23,7 @@ from xdsl.dialects.builtin import (
     ShapedType,
     TensorType,
 )
+from xdsl.dialects.csl import csl_stencil
 from xdsl.dialects.func import FuncOp
 from xdsl.dialects.linalg import FillOp
 from xdsl.dialects.stencil import (
@@ -35,7 +38,7 @@ from xdsl.dialects.stencil import (
     StoreOp,
     TempType,
 )
-from xdsl.dialects.tensor import EmptyOp, ExtractSliceOp
+from xdsl.dialects.tensor import EmptyOp, ExtractSliceOp, InsertSliceOp
 from xdsl.ir import (
     Attribute,
     Operation,
@@ -69,6 +72,15 @@ def get_required_result_type(op: Operation) -> TensorType[Attribute] | None:
                             return r_type
                 # abort when encountering an un-tensorized ReturnOp successor
                 return None
+            if (
+                isinstance(use.operation, InsertSliceOp)
+                and is_tensor(use.operation.result.type)
+                and isa(use.operation.static_sizes.data, ArrayAttr[IntAttr])
+            ):
+                return TensorType(
+                    use.operation.result.type.get_element_type(),
+                    use.operation.static_sizes.data,
+                )
             for ret in use.operation.results:
                 if isa(r_type := ret.type, TensorType[Attribute]):
                     return r_type
@@ -298,6 +310,24 @@ class AccessOpUpdateShape(RewritePattern):
                 )
 
 
+class CslStencilAccessOpUpdateShape(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: csl_stencil.AccessOp, rewriter: PatternRewriter, /):
+        if typ := get_required_result_type(op):
+            if needs_update_shape(op.result.type, typ) and (
+                isa(op.op.type, TempType[TensorType[Attribute]])
+                or isa(op.op.type, MemRefType[TensorType[Attribute]])
+            ):
+                rewriter.replace_matched_op(
+                    csl_stencil.AccessOp(
+                        op.op,
+                        op.offset,
+                        op.op.type.get_element_type(),
+                        op.offset_mapping,
+                    )
+                )
+
+
 class ExtractSliceOpUpdateShape(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: ExtractSliceOp, rewriter: PatternRewriter, /):
@@ -356,6 +386,28 @@ class FillOpUpdateShape(RewritePattern):
 
 
 @dataclass(frozen=True)
+class BackpropagateStencilShapes(ModulePass):
+    name = "backpropagate-stencil-shapes"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        backpropagate_stencil_shapes = PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    # AccessOpUpdateShape(),
+                    CslStencilAccessOpUpdateShape(),
+                    ExtractSliceOpUpdateShape(),
+                    EmptyOpUpdateShape(),
+                    FillOpUpdateShape(),
+                    ArithOpUpdateShape(),
+                ]
+            ),
+            walk_reverse=True,
+            apply_recursively=False,
+        )
+        backpropagate_stencil_shapes.rewrite_module(op)
+
+
+@dataclass(frozen=True)
 class StencilTensorizeZDimension(ModulePass):
     name = "stencil-tensorize-z-dimension"
 
@@ -386,17 +438,4 @@ class StencilTensorizeZDimension(ModulePass):
             apply_recursively=False,
         )
         stencil_pass.rewrite_module(op)
-        backpropagate_stencil_shapes = PatternRewriteWalker(
-            GreedyRewritePatternApplier(
-                [
-                    # AccessOpUpdateShape(),
-                    ExtractSliceOpUpdateShape(),
-                    EmptyOpUpdateShape(),
-                    FillOpUpdateShape(),
-                    ArithOpUpdateShape(),
-                ]
-            ),
-            walk_reverse=True,
-            apply_recursively=False,
-        )
-        backpropagate_stencil_shapes.rewrite_module(op)
+        BackpropagateStencilShapes().apply(ctx=ctx, op=op)
--- a/xdsl/transforms/stencil_to_csl_stencil.py
+++ b/xdsl/transforms/stencil_to_csl_stencil.py
@@ -289,9 +289,9 @@ def get_op_split(
     This function does not attempt to arithmetically re-structure the computation to obtain a good split. To do this,
     `RestructureSymmetricReductionPattern()` may be executed first.
     """
-    a: Sequence[Operation] = []
-    b: Sequence[Operation] = []
-    rem: Sequence[Operation] = []
+    a: list[Operation] = []
+    b: list[Operation] = []
+    rem: list[Operation] = []
     for op in ops:
         if isinstance(op, csl_stencil.AccessOp):
             (b, a)[op.op == buf].append(op)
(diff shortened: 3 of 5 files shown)


xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+    has_changes = True
+    while (
+        len(
+            a_exports := set(
+                op
+                for op in a
+                for r in op.results
+                for u in r.uses
+                if u.operation not in a
+            )
```
AntonLydike, line 330: I am not sure what this is computing exactly. May I suggest you put this in an aptly named helper function with a docstring? I feel this would make the code much more readable.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+    - `a` contains neighbour accesses plus the minumum set of instructions to reduce the accessed data to 1 thing
+    - `b` contains everything else
+
+    If no valid split can be found, return `(ops, [])`.
+
+    This function does not attempt to arithmetically re-structure the computation to obtain a good split. To do this,
+    `RestructureSymmetricReductionPattern()` may be executed first.
+    """
+    a: Sequence[Operation] = []
+    b: Sequence[Operation] = []
```
AntonLydike, line 313: I found these one-letter names to become confusing later down the line when used in list comprehensions. Would you mind giving these more descriptive names (e.g. `dependent_ops` and `independent_ops`?)

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+    Returns a split of ops into an `(a,b)` tuple, such that:
+
+    - `a` contains neighbour accesses plus the minumum set of instructions to reduce the accessed data to 1 thing
+    - `b` contains everything else
+
+    If no valid split can be found, return `(ops, [])`.
+
+    This function does not attempt to arithmetically re-structure the computation to obtain a good split. To do this,
+    `RestructureSymmetricReductionPattern()` may be executed first.
+    """
```
AntonLydike, line 311: This is a good start, although I would like to see you mention what each input argument represents (currently `prefetch` is unexplained)

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+        movable = [
+            op
+            for op in a_deps
+            if all(x.op in a for x in op.operands if isinstance(x, OpResult))
+        ]
+        if len(movable) > 0:
+            has_changes = True
+            for op in movable:
+                a.append(op)
+                rem.remove(op)
```
AntonLydike, line 348: Hmmm, I'm having a hard time looking through this. Could you add some comments here what each check is fore, and how we modify our global variables

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+
+@dataclass(frozen=True)
+class ConvertApplyOpPattern(RewritePattern):
+    """
+    Fuses a `csl_stencil.prefetch` and a `stencil.apply` to build a `csl_stencil.apply`.
+
+    If there are several candidate prefetch ops, the one with the largest result buffer size is selected.
+    The selection is greedy, and could in the future be expanded into a more global selection optimising for minimal
+    prefetch overhead across multiple apply ops.
+    """
```
AntonLydike, line 366: Nice! Really helpful already! 

Ideally you'd mention what parameters your pattern takes (e.g. `num_chunks`) so that people know how to call it!

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -116,24 +116,24 @@ class PrefetchOp(IRDLOperation):
...
 
     result = result_def(memref.MemRefType)
 
     def __init__(
         self,
         input_stencil: SSAValue | Operation,
-        topo: dmp.RankTopoAttr | None = None,
-        swaps: Sequence[ExchangeDeclarationAttr] | None = None,
+        topo: dmp.RankTopoAttr,
+        swaps: Sequence[ExchangeDeclarationAttr],
```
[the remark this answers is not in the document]
n-io: The params can't be optional anymore, so can't be initialised to `None`.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+                    IntegerType(64),
+                    (static_size,),
+                ),
+                "static_strides": DenseArrayBase.from_list(IntegerType(64), (1,)),
+                "operandSegmentSizes": DenseArrayBase.from_list(
+                    IntegerType(32), (1, 1, 1, 0, 0)
+                ),
+            },
+            result_types=[result_type],
+        )
```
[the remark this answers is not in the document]
n-io: Yes, good call. Bear in mind that this will be a static method, in order to prevent re-factoring calls in the rest of the code base that uses the existing inherited `init`, as this op is already established.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,146 @@
...
+from xdsl.rewriter import InsertPoint
+from xdsl.transforms.experimental.stencil_tensorize_z_dimension import (
+    BackpropagateStencilShapes,
+)
 from xdsl.utils.hints import isa
 
 
+def get_stencil_accessed_symbols(op: Operand) -> set[Operand]:
+    """
+    Returns the symbols of all stencil accessess by op and all its dependencies.
```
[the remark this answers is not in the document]
n-io: Correct, as `stencil.access`/`csl_stencil.access` are accessing `Operand`s.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Ok

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+            properties={
+                "static_offsets": DenseArrayBase.from_list(IntegerType(64), (0,)),
+                "static_sizes": DenseArrayBase.from_list(
+                    IntegerType(64),
+                    (static_size,),
+                ),
+                "static_strides": DenseArrayBase.from_list(IntegerType(64), (1,)),
+                "operandSegmentSizes": DenseArrayBase.from_list(
+                    IntegerType(32), (1, 1, 1, 0, 0)
+                ),
```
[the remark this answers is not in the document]
n-io: Thanks

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Ok

xdsl/transforms/stencil_to_csl_stencil.py
n-io: This is the aptly named helper function with a docstring :)

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Simplified the code and added comments.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+    If there are several candidate prefetch ops, the one with the largest result buffer size is selected.
+    The selection is greedy, and could in the future be expanded into a more global selection optimising for minimal
+    prefetch overhead across multiple apply ops.
+    """
+
+    num_chunks: int
+
+    def _translate_operands(
+        self, ops: Sequence[Operand], table: dict[Operand, Operand]
+    ) -> Sequence[Operand]:
```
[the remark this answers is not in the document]
n-io: Inlined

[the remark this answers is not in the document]
n-io: the logic had slightly simplified after a rewrite, it didn't acty need to be its own function anymore

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,146 @@
...
+    Uses helper function `get_stencil_accessed_symbols` to check which bufs are stencil-accessed in each of these args,
+    and to distinguish the following three cases:
+
+     (1) all accesses in an arg tree are to `buf`  - arg should be moved forward in the computation
+     (2) no accesses are to `buf`                  - arg should be moved backward in the computation
+     (3) there's a mix                             - unknown, take any or no action
+
+    If two args are identified that should be moved forward, or two args are identified that should be moved backwards,
+    the computation is restructured accordingly.
+    """
```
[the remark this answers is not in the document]
n-io: > I wonder if a better way to think about it would be a sort-of graph partitioning algorithm, where you want to re-order your graph such that in the end you have as few edges as possible between the two partitions.

It should only be a single edge, to be precise.

> I was thinking that it should be possible to do this with a local rule like:
> 
> match `(a+b)+c` if `source(a) != source(b)` then rewrite to `(same_source_as(a)+a)+not_same_source_as(a))`.

I was hoping that too, but to untangle the structure you'd want to move good things forward *and* bad things backward.

> Also, the terminology of forward/backward seems to be a bit odd, as we are looking at some sort of DAG.

One that is flattened to a list.

> Finally, since all arith ops are pure anyway, the only order that really matters here is use-def order, so I don't really see a point in moving these ops around to be above/below a certain point...

Without the insert point maths, it kept creating scenarios where values are used before they are defined. It seemed best to insert each newly created instruction at a point where both its operands are present to avoid this side-effect of tree restructuring.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Splitting a list into two parts a and b is probably one of the few cases where I'd use single-char variable names, but I have renamed things in the list comprehensions hoping to make them more readable. Come back on this please if you don't agree.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +295,314 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+                ),
+                "static_strides": DenseArrayBase.from_list(IntegerType(64), (1,)),
+                "operandSegmentSizes": DenseArrayBase.from_list(
+                    IntegerType(32), (1, 1, 1, 0, 0)
+                ),
+            },
+            result_types=[result_type],
+        )
+
+    @op_type_rewrite_pattern
```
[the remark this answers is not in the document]
n-io: I recon this may come with small print to test things separately. It's really just translating one op anyways 🥲

xdsl/transforms/stencil_to_csl_stencil.py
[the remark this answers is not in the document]
n-io: @PapyChacal suggested to orient this along the lines of what constant folding does.

xdsl/transforms/stencil_to_csl_stencil.py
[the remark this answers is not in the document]
AntonLydike:
```suggestion
    Returns all operands of all `csl_stencil.access` ops that are (transitive) inputs to the given operation `op`.
```

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,145 @@
...
     op_type_rewrite_pattern,
 )
+from xdsl.rewriter import InsertPoint
+from xdsl.transforms.experimental.stencil_tensorize_z_dimension import (
+    BackpropagateStencilShapes,
+)
 from xdsl.utils.hints import isa
 
 
+def get_stencil_accessed_symbols(op: Operand) -> set[Operand]:
```
AntonLydike, line 33: Maybe something closer to

```suggestion
def find_transitive_stencil_access_operands(op: Operand) -> set[Operand]:
```

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -171,10 +294,282 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
...
+                if use.operation not in a
+            )
+        )
+        > 1
+        and has_changes
+    ):
+        has_changes = False
+
+        # ops that directly depend on `a` but are not themselves in `a`
+        fontier = set(
```
AntonLydike, line 339: Consider more descriptive names:

```suggestion
        depending_on_a = set(
```

xdsl/transforms/stencil_to_csl_stencil.py
AntonLydike: The thing is that `a` and `b` have semantic meaning attached, so when there's a check like `x in a` it's less clear than `x in independent_ops`.

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,145 @@
...
+            Builds `(one+two)+three` where `'+' == type(op)`
+            """
+            first_compute = type(op)(one, two)
+            second_compute = type(op)(first_compute, three)
+
+            # insert ops at the earliest point after both of its dependencies
+            rewriter.insert_op(
+                first_compute,
+                InsertPoint.after(
+                    max(
```
dk949, line 98: Could this be simplified or extracted into a function with a docstring? I've had a really hard time following the logic here :sweat_smile:

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,145 @@
...
+        self, op: arith.Addf | arith.Mulf, rewriter: PatternRewriter, /
+    ):
+        # this rewrite requires exactly 1 use which is the same type of operation
+        if len(op.result.uses) != 1 or not isinstance(
+            use := list(op.result.uses)[0].operation, type(op)
+        ):
+            return
+        c_op = use.operands[0] if use.operands[1] == op.result else use.operands[1]
+
+        def rewrite(one: Operand, two: Operand, three: Operand):
```
dk949, line 87: As far as I can tell, `one`, `two` and `three` have to have the same parent for this to work?

Could we assert this to make sure the logic below always works (or breaks loudly)?

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,145 @@
...
+            second_compute = type(op)(first_compute, three)
+
+            # insert ops at the earliest point after both of its dependencies
+            rewriter.insert_op(
+                first_compute,
+                InsertPoint.after(
+                    max(
+                        (o.op.parent.get_operation_index(o.op), o.op)
+                        for o in [one, two]
+                        if isinstance(o, OpResult) and isinstance(o.op.parent, Block)
```
dk949, line 101: What happens when both `one` and `two` are block args? can this happen?

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,145 @@
...
+)
 from xdsl.utils.hints import isa
 
 
+def get_stencil_accessed_symbols(op: Operand) -> set[Operand]:
+    """
+    Returns the symbols of all stencil accessess by op and all its dependencies.
+    """
+    res: set[Operand] = set()
+    frontier: set[Operand] = set((op,))
```
dk949, line 38: nit: I personally find `{op}` easier to read than `set((op, ))`

xdsl/transforms/stencil_to_csl_stencil.py
```diff
@@ -1,23 +1,145 @@
...
+        elif self.move_back(a) and self.move_back(b):
+            return
+        elif self.move_fwd(a) and self.move_fwd(c):
+            rewrite(a_op, c_op, b_op)
+        elif self.move_fwd(b) and self.move_fwd(c):
+            rewrite(b_op, c_op, a_op)
+        elif self.move_back(a) and not self.move_back(c):
+            rewrite(c_op, b_op, a_op)
+        elif self.move_back(b) and not self.move_back(c):
+            rewrite(c_op, a_op, b_op)
```
dk949, line 134: I _think_ this encodes some of the answers to the questions I asked in comments above. regarding ops being block args? It's a bit challenging for me to follow, if I'm honest.

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -355,6 +389,32 @@ def match_and_rewrite(self, op: FillOp, rewriter: PatternRewriter, /):
...
 
 
+@dataclass(frozen=True)
+class BackpropagateStencilShapes(ModulePass):
+    """
+    Greedily back-propagates the result types of tensorized ops.
+    Use after creating/modifying tensorization.
+    """
+
+    name = "backpropagate-stencil-shapes"
```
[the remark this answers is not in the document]
n-io: Yes, it's a helper more than a proper thing.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Changed to `get_stencil_access_operands`

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Changing as suggested

xdsl/transforms/stencil_to_csl_stencil.py
n-io: I found a way to simplify it

xdsl/transforms/stencil_to_csl_stencil.py
n-io: This should no longer be an issue after the rewrite.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Handling this differently now

xdsl/transforms/stencil_to_csl_stencil.py
n-io: Given that :128 and :134, as well as :130 and :132 result in equivalent rewrites, I can probably shorten it at the cost of trickier-to-read conditions.

xdsl/transforms/stencil_to_csl_stencil.py
n-io: `frontier` sounds like good compiler terminology no?

xdsl/transforms/stencil_to_csl_stencil.py
n-io: The semantic meaning that comes to mind is to refer to elements of a tuple 😅

AntonLydike (approved):

PR #2809, opened 2024-07-01
core: Make print_string aware of indentation

Moxinilian:
This PR makes the `print_string` function in the printer aware of indentation, and will print newlines in strings accordingly.

For cases where this is not desired (such as printing a newline character in the indentation-aware newline logic), `print_string_raw` has been added to print the string ignoring indentation.

Additionally, a helper `indented` function for with statements has been added to control indentation within a scope. Feedback on this would be much appreciated, as a possible concern is that it may not be super intuitive (it indents newlines, not lines).

### DIFF
--- a/tests/test_printer.py
+++ b/tests/test_printer.py
@@ -885,3 +885,37 @@ def test_get_printed_name():
     printed = StringIO()
     picked_name = Printer(printed).print_ssa_value(val)
     assert f"%{picked_name}" == printed.getvalue()
+
+
+def test_indented():
+    output = StringIO()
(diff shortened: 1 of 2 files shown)


superlopuh:
Amazing, I was just implementing it locally!

superlopuh (approved):
great timing, I also need it for pretty syntax in the kernel ops

superlopuh:
I would say that the previous behaviour was also not intuitive, and the new functions add some more structure, so we're better off with the new functionality.

math-fehr (approved):
Very cool

PR #2814, opened 2024-07-02
(transform): csl_stencil canonicalization pass

n-io:
Adds a canonicalisation pass for `csl_stencil.apply`. The op takes an empty tensor as`iter_arg`, which it does not manage itself. The conversion pass in #2803 initialises an `iter_arg` for each instance of `apply`. This canonicalisation pass identifies where this can be re-used, effectively removing redundant allocations.

### DIFF
--- a/xdsl/dialects/csl/csl_stencil.py
+++ b/xdsl/dialects/csl/csl_stencil.py
@@ -29,9 +29,11 @@ from xdsl.irdl import (
     var_result_def,
 )
 from xdsl.parser import AttrParser, Parser
+from xdsl.pattern_rewriter import RewritePattern
 from xdsl.printer import Printer
 from xdsl.traits import (
     HasAncestor,
+    HasCanonicalisationPatternsTrait,
     HasParent,
     IsolatedFromAbove,
     IsTerminator,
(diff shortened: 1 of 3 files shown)


xdsl/transforms/canonicalization_patterns/csl_stencil.py
```diff
@@ -0,0 +1,33 @@
...
+class RedundantIterArgInitialisation(RewritePattern):
+    """
+    Removes redundant allocations of empty tensors with no uses other than passed
+    as `iter_arg` to `csl_stencil.apply`. Prefer re-use where possible.
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: csl_stencil.ApplyOp, rewriter: PatternRewriter
+    ) -> None:
```
dk949, line 20: Would match all `csl_stencil.ApplyOp`s? I am guessing this doesn't really matter, since only the first one will get rewritten?

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -139,6 +141,16 @@ def __init__(
         )
 
 
+class ApplyOpHasCanonicalizationPatternsTrait(HasCanonicalisationPatternsTrait):
```
dk949, line 144: Am I correct that this gets automatically applied because `ApplyOpHasCanonicalizationPatternsTrait` trait?

dk949 (approved):

xdsl/dialects/csl/csl_stencil.py
n-io: Yes, the `canonicalize` pass picks up on this trait, so it can be scheduled anywhere in the pipeline. The GPU pipeline actually has a few `canonicalize` at different points.

xdsl/transforms/canonicalization_patterns/csl_stencil.py
n-io: It actually just recycles the arg being used.

PR #2815, opened 2024-07-02
dialects: (memref_stream) print memref_stream.generic with newlines

superlopuh:
Makes it much easier to read the filechecks IMO.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
     "nbval<0.12",
     "filecheck<0.0.25",
     "lit<19.0.0",
-    "marimo==0.6.23",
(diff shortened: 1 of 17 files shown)


PapyChacal (approved):
I'm not a fan, it reads like testing more than what we care about to me - i.e., any spacing change now needs functional test updating

This is an opinion though and the PR looks fine to me in itself, so this is indeed an approval 😉

superlopuh:
The whitespace-strictness is only for the dialect roundtrip test, not for the rest of them. Or do you mean that if we add more newlines more tests will have to change? They would probably change anyway, so it doesn't seem like a big concern.

Moxinilian (approved):

PR #2819, opened 2024-07-02
ir: Add forward slash ('/') to lexer

alexarice:
Adds a forward slash token to the lexer. Mlir seems to accept this as a token so there seems little reason not to accept in in xdsl too (unless someone has a good reason not to). 

It's possible there are other missing tokens so if someone knows the appropriate place to look in mlir to see all the punctuation then I'd be happy to add more.

### DIFF
--- a/tests/test_lexer.py
+++ b/tests/test_lexer.py
@@ -53,13 +53,14 @@ def assert_token_fail(input: str):
         ("|", Token.Kind.VERTICAL_BAR),
         ("{-#", Token.Kind.FILE_METADATA_BEGIN),
         ("#-}", Token.Kind.FILE_METADATA_END),
+        ("/", Token.Kind.FORWARD_SLASH),
     ],
 )
 def test_punctuation(text: str, kind: Token.Kind):
     assert_single_token(text, kind)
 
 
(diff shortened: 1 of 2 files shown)


math-fehr:
I'll take a look tonight to see how MLIR does this exactly!

math-fehr:
Hmm, looking at it, it seems that MLIR doesn't actually accept it in its lexer: https://github.com/llvm/llvm-project/blob/main/mlir/lib/AsmParser/TokenKinds.def

I have to say, I am really not sure why we saw it accepted? But I think it might be because it was an unregistered attribute, so it probably behaved differently (though it shouldn't have).

So I would say that it's better to not add this, to stay the same as MLIR, if that makes sense?

alexarice:
I had a look myself and it seems that mlir decides to bypass its own lexer when parsing a dialect attribute body https://github.com/llvm/llvm-project/blob/3b73cb3bf138d2ad7f354c9d26037d769cfe9d53/mlir/lib/AsmParser/DialectSymbolParser.cpp#L101

I guess this should be closed and i should just use something like `div` instead

alexarice:
I made a discussion on zulip but I think it's best to close this for now, the issue seems more complicated than I first thought

PR #2820, opened 2024-07-02
misc: Fix the way --split-input-file is handled

AntonLydike:
Make `--split-input-file` only split on `// -----` if it's the first thing on a line (modulo whitespace)


PapyChacal (approved):
Yes, this annoyed me so many times 🤩

Why the modulo whitespace btw? Unless clear reason (or over complex implementation), I would also not split if preceding whitespace

math-fehr:
How do we handle this for non-`.mlir` files? For instance the frontend python files we still have?
I think it's common to write something like `# // -----` for splitting these, why can we not `CHECK: -----` for checking splits?

AntonLydike:
> Yes, this annoyed me so many times 🤩
> 
> Why the modulo whitespace btw? Unless clear reason (or over complex implementation), I would also not split if preceding whitespace

I found a bunch of tests that had indented splits, and I found that it's quite irritating to debug why the file wasn't split at a certain point if the split is indented by a single whitespace^^

Also, MLIR also allows indented splits, so yeah.

AntonLydike:
> How do we handle this for non-`.mlir` files? For instance the frontend python files we still have? I think it's common to write something like `# // -----` for splitting these, why can we not `CHECK: -----` for checking splits?

Seems weird that the tests pass... But good point!

Looking at MLIR, it just splits on any line containing `// -----` anywhere, and deletes the whole line. Maybe we copy that?

superlopuh (approved):
fantastic

math-fehr:
> > How do we handle this for non-`.mlir` files? For instance the frontend python files we still have? I think it's common to write something like `# // -----` for splitting these, why can we not `CHECK: -----` for checking splits?
> 
> Seems weird that the tests pass... But good point!
> 
> Looking at MLIR, it just splits on any line containing `// -----` anywhere, and deletes the whole line. Maybe we copy that?

To me we should copy that yes, but what's currently implemented then? I thought it was already that

AntonLydike:
> > > How do we handle this for non-`.mlir` files? For instance the frontend python files we still have? I think it's common to write something like `# // -----` for splitting these, why can we not `CHECK: -----` for checking splits?
> > 
> > 
> > Seems weird that the tests pass... But good point!
> > Looking at MLIR, it just splits on any line containing `// -----` anywhere, and deletes the whole line. Maybe we copy that?
> 
> To me we should copy that yes, but what's currently implemented then? I thought it was already that

We currently split on `// -----` anywhere, but keep the remaining bits of the lines

I fixed it to eat the whole line.

AntonLydike:
Wait, I was wrong. Current behaviour mimics MLIRs, so we are good!

Issue #2807, opened 2024-07-01
Assembly format does not print/parse dialect attributes correctly


alexarice:
I believe that dialect attributes are not being parsed correctly by the assembly format. Take the following minimal example, which I have arbitrarily defined as part of the qssa dialect:
```
@irdl_op_definition
class MyTest(IRDLOperation):
    name = "qssa.mytest"

    test = prop_def(LLVMArrayType)

    assembly_format = "$test attr-dict"
```
The following can be parsed:
```
"qssa.mytest"()<{test = !llvm.array<2xi32>}>: () -> ()
```
but it is printed as follows:
```
qssa.mytest <2 x i32>
```
My understanding is that it should be printed as:
```
qssa.mytest !llvm.array<2 x i32>
```
I believe this is a bug in the assembly format parser, though this could just be a problem with my understanding/ the way I'm using the assembly format.

superlopuh:
@PapyChacal, is this expected?

PapyChacal:
It is expected in our implementation; I'm genuinely not sure if this is expected in MLIR's implementation on the top of my head.

The former example:
```
"qssa.mytest"()<{test = !llvm.array<2xi32>}>: () -> ()
```
Is in generic syntax, so expects fully qualified attributes as above.

The latter:
```
qssa.mytest <2 x i32>
```
Is in *custom* syntax inferred by the declarative syntax framework.
The logic is, because you constraint the attribute in a way that we can *infer* it is an `!llvm.array`, it just elides the name in custom syntax.

So:
- As long as it roundtrips (parses and prints correctly between both syntax, i.e, always keep representing the same thing and not erroring), yes, it is the expected behaviour
- If you really want the fully qualified version to be used in declarative assembly syntax, MLIR's way is to have `qualified($test)` instead of just `$test` in your syntax, to enforce it; but this is not implemented yet in xDSL; but shouldn't be a big deal

For the record, I agree that the resulting syntax here is weird to read ahah - but I think it might be *functionally* correct

If you want `qualified` implemented in xDSL's declarative syntax, let's open a dedicated issue?
Also, could you then ping me again here or there? I'm not following all issues :sweat_smile:

alexarice:
It makes sense that the attribute name can always be inferred here, it just seemed strange that this was the output to me. I have no need for the qualified flag so I'll close the issue for now

PR #2821, opened 2024-07-02
dialects: add parametrised rz gate

alexarice:
This pr does the following:
- Adds 'quantum' dialect which contains an angle attribute which is shared by both the 'qref' and 'qssa' dialects.
- angles are stored as rational multiples of 'pi' kept to the range [0,2pi)
- Adds rz gate which is parametrised by an angle (at the moment this is given via an attribute.
- Standardises the gate set to the one used in 'http://arxiv.org/abs/1710.07345'. I can revert the removal of the cz gate if this seems like a poor idea.

~Built on top of #2819 to allow '/' in the angle syntax~
Now using ':' instead to avoid the lexing change

### DIFF
--- a/tests/filecheck/transforms/convert_qssa_to_qref.mlir
+++ b/tests/filecheck/transforms/convert_qssa_to_qref.mlir
@@ -3,14 +3,14 @@
 
 %q0, %q1 = qssa.alloc<2>
 %q2 = qssa.h %q0
-%q3 = qssa.rz <pi/2> %q1
+%q3 = qssa.rz <pi:2> %q1
 %q4, %q5 = qssa.cnot %q2, %q3
 %0 = qssa.measure %q4
 
 // CHECK:       builtin.module {
 // CHECK-NEXT:    %q0, %q1 = qref.alloc<2>
 // CHECK-NEXT:    qref.h %q0
-// CHECK-NEXT:    qref.rz <pi/2> %q1
+// CHECK-NEXT:    qref.rz <pi:2> %q1
 // CHECK-NEXT:    qref.cnot %q0, %q1
 // CHECK-NEXT:    %0 = qref.measure %q0
 // CHECK-NEXT:  }
@@ -18,7 +18,7 @@
 // CHECK-ROUNDTRIP:       builtin.module {
 // CHECK-ROUNDTRIP-NEXT:    %q0, %q1 = qssa.alloc<2>
 // CHECK-ROUNDTRIP-NEXT:    %q0_1 = qssa.h %q0
-// CHECK-ROUNDTRIP-NEXT:    %q1_1 = qssa.rz <pi/2> %q1
+// CHECK-ROUNDTRIP-NEXT:    %q1_1 = qssa.rz <pi:2> %q1
 // CHECK-ROUNDTRIP-NEXT:    %q0_2, %q1_2 = qssa.cnot %q0_1, %q1_1
 // CHECK-ROUNDTRIP-NEXT:    %0 = qssa.measure %q0_2
 // CHECK-ROUNDTRIP-NEXT:  }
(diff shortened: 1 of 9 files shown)


superlopuh (approved):
beautiful

xdsl/dialects/quantum.py
```diff
@@ -0,0 +1,57 @@
...
+    Attribute that wraps around a fraction, implicitly multiplying by pi and keeping the result in the range [0,2pi)
+    """
+
+    name = "quantum.angle"
+
+    @classmethod
+    def parse_parameter(cls, parser: AttrParser) -> Fraction:
+        with parser.in_angle_brackets():
+            i = parser.parse_optional_integer()
+            numerator = 1 if i is None else i
```
AntonLydike, line 23: I don't think an integer numerator is not covered by filecheck?

tests/filecheck/transforms/convert_qssa_to_qref.mlir
```diff
@@ -3,23 +3,23 @@
...
+%q3 = qssa.rz <pi/2> %q1
+%q4, %q5 = qssa.cnot %q2, %q3
+%0 = qssa.measure %q4
 
 // CHECK:       builtin.module {
 // CHECK-NEXT:    %q0, %q1 = qref.alloc<2>
 // CHECK-NEXT:    qref.h %q0
-// CHECK-NEXT:    qref.cz %q1, %q0
-// CHECK-NEXT:    qref.cnot %q1, %q0
+// CHECK-NEXT:    qref.rz <pi/2> %q1
```
AntonLydike, line 13: @PapyChacal do you know why the attribute name is not printed here?

AntonLydike (commented):
Looks good, just missing some tests covering anything that's not `<pi/2>`^^

tests/filecheck/transforms/convert_qssa_to_qref.mlir
alexarice: see https://github.com/xdslproject/xdsl/issues/2807

xdsl/dialects/quantum.py
```diff
@@ -0,0 +1,57 @@
...
+                printer.print(0)
+                return
+
+            if self.data.numerator != 1:
+                printer.print(self.data.numerator)
+
+            printer.print("pi")
+            if self.data.denominator != 1:
+                printer.print("/")
+                printer.print(self.data.denominator)
```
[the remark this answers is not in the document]
alexarice: I've simplified this slightly

n-io:
> * I can revert the removal of the cz gate if this seems like a poor idea.

This likely seems a higher-level, strategic consideration than could be done justice here. In my understanding there are some equivalences/semantic-preserving transformations between gates. I suppose you removed CZ in order to keep the dialect minimal? If you are wondering whether CZ should be in or not (without intending to take a position here myself), one possible consideration could be to offer a dialect with a wider variety of gates and lowering it to one with fewer gates.

n-io (approved):

xdsl/dialects/quantum.py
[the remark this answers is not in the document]
superlopuh: Oh yes my bad I misread the original

alexarice:
Added another commit that fixed some of the arithmetic operations on angles. I also added an `AngleAttr.from_fraction` method but It would be good to know if there is another better way of doing this

alexarice:
Realised I would like to be able to also parse `!quantum.angle<-pi>`, working on this now

AntonLydike:
@alexarice what's the status on `-pi`?

alexarice:
I had it working but it looks like I didn't push it, I'll get it pushed over lunch

alexarice:
I'm very confused, it seems I managed to push the commit https://github.com/xdslproject/xdsl/commit/8e178a1e345a5590af228582109b45e0040103db , but it doesn't seem to be appearing in this pull request, even though if you click on the branch for the pull request the commit is there. Anyone know what is going on?

alexarice:
@AntonLydike I've rebased and force pushed and now the commit shows up, so this pr should hopefully be ready

PR #2823, opened 2024-07-02
transformations: add memref-stream-interleave

superlopuh:


nazavode (approved):
Beautiful.

xdsl/transforms/memref_stream_interleave.py
```diff
@@ -0,0 +1,154 @@
...
+                        )
+                        + (
+                            AffineExpr.dimension(interleave_bound_index)
+                            * self.pipeline_depth
+                            + AffineExpr.dimension(m.data.num_dims),
+                        )
+                        + tuple(
+                            AffineExpr.dimension(i)
+                            for i in range(
+                                interleave_bound_index + 1, m.data.num_dims + 2
```
georgebisbas, line 102: Where do these numbers stem from?

superlopuh: The idea is that there is an extra dimension added somewhere in the middle of the affine map domain. This is the slice of the domain after the updated value, so it's going from `inserted_index + 1 to old_len + 1` inclusive.

superlopuh:
Lowering works OK, let's merge this and keep going.

PR #2827, opened 2024-07-03
tests: Do not run marimo mlir tests without MLIR installed

### LINKED ISSUE #2826: Tests on main fail without mlir installed
Without mlir-opt installed, `make tests` on master fails. Could these tests come up as "Unsupported" instead of failing like the mlir integration tests?

The tests fail at `Running docs/marimo/mlir/onnx_demo.py`

webmiche:
title says it all.

closes #2826

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,12 @@ tests-marimo:
 	@echo "All marimo tests passed successfully."
 
 tests-marimo-mlir:
+	@if command -v mlir-opt &> /dev/null; then
+		@echo "MLIR is installed, running tests."
+	else
+		@echo "MLIR is not installed, skipping tests."
+		exit 0
+	fi
 	@for file in docs/marimo/mlir/*.py; do \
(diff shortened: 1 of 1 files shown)


webmiche:
I am a bit surprised that this made it through our Python-based testing, this kind of error should be caught there, I would expect.

alexarice (approved):

webmiche:
I guess we removed the notebooks from the Python-based testing to its own thing, and this thing only runs the non-mlir marimo stuff.

Not sure how to work around this, though I think we should test that this kind of mistake does not happen somewhere in the CI. What do you think, @superlopuh?

superlopuh:
I forgot about this use-case, I think we might benefit from a CI dedicated to running `make tests` with only `xdsl[dev]` installed, as opposed to the full `requirements.txt`.

superlopuh (approved):

PR #2829, opened 2024-07-03
misc: Don't add default arguments to PassPipelineSpec

AntonLydike:
@superlopuh


superlopuh:
Oh, now that I think about it, maybe this could be an option on the helper? I think this is used to populate the pass argument text box in xdsl-gui. I'm also surprised that none of the unit tests needed to change, I guess this means that we would benefit from some new ones.

superlopuh:
If you're not feeling like doing this I'm happy to take over.

PR #2830, opened 2024-07-03
(core): convenience functions for InsertPoint

n-io:
Adding two convenience functions on InsertPoint:

```
@staticmethod
def after_all(ops: Sequence[Operation]) -> InsertPoint:
    """Gets the insertion point after all operations, which must all belong to the same block."""
```

```
@staticmethod
def after_operands(ops: Sequence[Operand]) -> InsertPoint:
    """Gets the insertion point after all operands (which must be of the same block) are present."""
```

### DIFF
--- a/tests/test_rewriter.py
+++ b/tests/test_rewriter.py
@@ -365,6 +365,76 @@ def test_insert_op_after():
     rewrite_and_compare(prog, expected, transformation)
 
 
+def test_insert_op_after_all():
+    """Test the insertion of an operation after another operation."""
+    prog = """\
+"builtin.module"() ({
+  %0 = "arith.constant"() <{"value" = 40 : i32}> : () -> i32
+  %1 = "arith.constant"() <{"value" = 41 : i32}> : () -> i32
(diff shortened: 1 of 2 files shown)


superlopuh (commented):
Thank you for opening this PR! I have some concerns about the API. I'm not sure in which scenarios such a specific helper would be useful. My preference would be to keep the InsertPoint API as minimal as possible, but to add helper functions to get the first or last op in a block, and just constructing the InsertPoint once the operation is available. Similarly with a block.

One more thing to note is that `op.parent.get_operation_index(op)` is O(len(op.parent.ops)), meaning that the `after_all` complexity is number of ops in the block * number of ops in the sequence, and might benefit from a slightly a different implementation.

My suggestion would be to first add the helpers in the bit of code that would call them, and then place them in a shared place if there's a second bit of code that needs to call them.

math-fehr:
I would agree with Sasha here that we probably want to not complexify that much the API.

The utility that might be useful is getting the last operation in a block given a list of operation, but even that I would agree we don't have enough use cases yet.

PR #2831, opened 2024-07-03
transformations: Handle interleaving in convert-memref-stream-to-loops

superlopuh:
WIP PR while I write more tests and get the rest of the ops to compile properly. Currently stuck on how to tile operations when there are too many dimensions for streaming on Snitch, and whether it makes sense to have interleaved iterators vs a single interleave factor property on the operation.


tests/filecheck/transforms/convert_memref_stream_to_loops.mlir
```diff
@@ -374,9 +370,61 @@ func.func @interleaved_no_init(%A : memref<3x5xf64>, %B : memref<5x8xf64>, %C :
...
+// CHECK-NEXT:          %{{.*}} = affine.apply affine_map<(d0, d1) -> (((d0 * 4) + d1))> (%{{.*}}, %{{.*}})
+// CHECK-NEXT:          memref.store %{{.*}}, %C[%{{.*}}, %{{.*}}] : memref<3x8xf64>
+// CHECK-NEXT:          %{{.*}} = affine.apply affine_map<(d0, d1) -> (((d0 * 4) + d1))> (%{{.*}}, %{{.*}})
+// CHECK-NEXT:          memref.store %{{.*}}, %C[%{{.*}}, %{{.*}}] : memref<3x8xf64>
+// CHECK-NEXT:          %{{.*}} = affine.apply affine_map<(d0, d1) -> (((d0 * 4) + d1))> (%{{.*}}, %{{.*}})
+// CHECK-NEXT:          memref.store %{{.*}}, %C[%{{.*}}, %{{.*}}] : memref<3x8xf64>
+// CHECK-NEXT:        }
+// CHECK-NEXT:      }
+// CHECK-NEXT:      func.return %C : memref<3x8xf64>
+// CHECK-NEXT:    }
```
zero9178, line 427: I know this is contrary to previous practice, but would it be okay to use SSA names instead of the `%{{.*}}` here? I think some subtelties like the iteration bounds and which loop corresponds to what are lost otherwise.

xdsl/transforms/convert_memref_stream_to_loops.py
```diff
@@ -42,62 +41,109 @@ def _insert_load(
...
+                    rewriter,
+                    insertion_point,
+                    affine_map,
+                    tuple(ind_vars) + extra_dim(source_index),
+                )
+                store_op = memref.Store.get(value, destination, indices)
+            else:
+                store_op = memref_stream.WriteOp(value, destination)
+            rewriter.insert_op(store_op, insertion_point)
+            return store_op
```
zero9178, line 146: Could you put some high level comments here as to what these does, when they are called and specifically the special behaviour regarding the implementation of the interleaving (that in that case it needs to be mapped back to the original memref/stream from the "unrolled" block argument)

Ditto for the load

PR #2835, opened 2024-07-04
testing: fix mlir-opt check for MacOS make

superlopuh:
For some reason, running the version in `main` complains about a syntax error, the updated version works, and will hopefully be more portable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,11 +59,11 @@ tests-marimo:
 	@echo "All marimo tests passed successfully."
 
 tests-marimo-mlir:
-	@if command -v mlir-opt &> /dev/null; then
-		@echo "MLIR is installed, running tests."
-	else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -59,12 +59,11 @@ tests-marimo:
 	@echo "All marimo tests passed successfully."
 
 tests-marimo-mlir:
-	@if command -v mlir-opt &> /dev/null; then
```
AntonLydike, line 62: is `&>` not supported on mac `bash`? O.o

AntonLydike (approved):
LGTM

Makefile
superlopuh: that bit was fine, it's the linux that didn't like this

superlopuh: for mac it was the nested `@` and missing `\`

PR #2839, opened 2024-07-04
api: add printing and parsing of dynamic lists as in MLIR

superlopuh:
This is used in a few places, notably in memref.subview custom syntax, which is the next PR

### DIFF
--- a/tests/test_dialect_utils.py
+++ b/tests/test_dialect_utils.py
@@ -56,6 +56,19 @@ def test_print_dynamic_index_list():
     print_dynamic_index_list(printer, [], [])
     assert stream.getvalue() == "[]"
 
+    # Test case 6: Mix of integers and SSA values with custom dynamic index
+    dynamic_index = -42
+    stream = StringIO()
+    printer = Printer(stream)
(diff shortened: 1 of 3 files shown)


nazavode (approved):

webmiche:
So, the weird thing about this is that MLIR is not fully consistent itself with how this works, AFAIK. In particular, the default "take the operand"-value is different among different dialects (GEP from the LLVM dialect uses INT_MIN, whereas memref uses -1 apparently).

Providing this functionality within the parser feels like we are preferring one version over the other. I guess, one could overwrite the `DYNAMIC_INDEX` variable, but that feels very forced. What do you think about adding this value as a keyword argument? The default can still be -1, I guess.

superlopuh:
That's exactly my plan, I'll add it as a parameter, and fix my implementation, turns out that memref actually uses `-9223372036854775808`, and I ported the wrong function.

superlopuh:
The PR on top of this one roundrips perfectly with MLIR, so I'm now pretty confident the helper functions have at least that functionality working.

PR #2843, opened 2024-07-05
transformations: Fix subview hack in stencil conversion

PapyChacal:
Directly allocate buffers with the expected layout instead.

Reminder: This is a shallow layout, simply used to offset indices used in accesses, not modifying any physical layout. This avoids the need to propagate more information to `stencil.access`, which can simply access the lowered memref using its own coordinate system.


superlopuh (approved):

AntonLydike:
I faintly remember this being relevant in the DMP to MPI lowering or something. Do you know how this interacts with that?

PR #2845, opened 2024-07-05
transformations: Handle more division sizes in memref-stream-interleave

superlopuh:


xdsl/transforms/memref_stream_interleave.py
```diff
@@ -47,42 +46,52 @@ def match_and_rewrite(
             # No parallel dimension
             return
 
-        if interleave_bound % self.pipeline_depth:
-            raise DiagnosticException(
-                "Ranges of interleave factors not implemented yet"
-            )
+        interleave_factor = 1
+        # Search factors until the next number divisible by pipeline_depth
+        for potential_factor in range(self.pipeline_depth, self.pipeline_depth * 2):
```
zero9178, line 51: The upper range here seems somewhat arbitrary and could theoretically go up to `interleave_bound` if I understand correctly. 
Could you add a comment here explaining the tradeoff of choosing a higher vs lower `interleave_factor`? (higher `interleave_factor` leading to better FPSS usage but higher register pressure as well).

nazavode (approved):

PR #2847, opened 2024-07-05
transformations: Add memref.subview lowering to convert-memref-to-riscv

superlopuh:


xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -296,6 +367,100 @@ def match_and_rewrite(self, op: memref.GetGlobal, rewriter: PatternRewriter):
...
+
+        identity_strides = tuple(
+            accumulate(reversed(result_shape), lambda acc, new: acc * new, initial=1)
+        )[::-1]
+
+        if identity_strides[-len(result_strides) :] != result_strides:
+            # The subview may be rank reducing, we want the suffix of the original strides
+            raise DiagnosticException(
+                "Cannot lower strides that do not result in a contiguous subview"
+            )
```
zero9178, line 410: Why is this a limitation? 
And if this is due to our stack not supporting it, shouldn't the issue be caught with appropiate errors in the access operations rather than here?
I'd have expected the `subview` to perform an offset on the bare pointer regardless of the result type.

superlopuh: I didn't think too much about it, but you're right, that should be checked in the store/load lowering, I'll add it

- [ ] check layout on store and load lowering

zero9178: I'd also do this in a separate PR given its unrelated to the subview lowering

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -296,6 +367,100 @@ def match_and_rewrite(self, op: memref.GetGlobal, rewriter: PatternRewriter):
...
+            if static_offset == memref.Subview.DYNAMIC_INDEX:
+                pairs.append((offsets[dynamic_offset_index], stride))
+                dynamic_offset_index += 1
+            else:
+                if static_offset:
+                    # No need to insert arithmetic ops that will be multiplied by zero
+                    rewriter.insert_op_before_matched_op(
+                        offset_op := riscv.LiOp(static_offset)
+                    )
+                    pairs.append((offset_op.rd, stride))
```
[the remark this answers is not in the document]
superlopuh: Fixed, modulo still requiring the shape to be statically known and strides contiguous

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: Thinking about it a bit more, the strides are dangerous in that our riscv pointer representation assumes that everything is contiguous. I think it's still a good idea to check for it when lowering subview operations, just to increase the chances of catching the errors early.

zero9178: Do you know what operations support strides and which don't?

I am very much passing non-contigious memref to xdsl (they are basically required if you want to any kind of tiling besides workgroup tiling) and currently investigating as to why the xDSL workflow returns different results than LLVM in Qudditch at HEAD. Do note that the IR I am passing does not contain any subviews at all, just linalg operations, so these checks would not help in that case (which is why I'd rather see them at the memref consumer that cannot support them, rather than the memref producer that can)

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -141,7 +145,74 @@ def memref_shape_ops(
...
+    res_ptr = src_ptr
+
+    for index, factor in pairs:
+        if factor == 1:
+            ops.append(
+                head_op := riscv.AddOp(
+                    res_ptr, index, rd=riscv.IntRegisterType.unallocated()
+                )
+            )
+            res_ptr = head_op.rd
```
zero9178, line 190:
```suggestion
```

zero9178 (approved):

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -141,7 +145,74 @@ def memref_shape_ops(
...
+        [
+            bytes_per_element_op := riscv.LiOp(bytes_per_element),
+            offset_bytes := riscv.MulOp(
+                res_ptr,
+                bytes_per_element_op.rd,
+                rd=riscv.IntRegisterType.unallocated(),
+                comment="multiply by element size",
+            ),
+            ptr := riscv.AddOp(
+                src_ptr, offset_bytes, rd=riscv.IntRegisterType.unallocated()
```
zero9178, line 215: The code here seems problematic. Specifically, in the `for` before we keep on adding onto `res_ptr` and then we are multiplying the `res_ptr` with `bytes_per_element`. 

That means we are not multiplying the indices but rather the pointer value. `offset_bytes` is also misleadin given its actually the pointer/memref

zero9178 (changes requested):
GitHub doesn't let me request changes without a commet apparently...

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: OK then let's work out how best to deal with non-contiguous memrefs, it would be good to have a better idea of the kinds of patterns we need to support.

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: Yep, I think I got a little lost in the sauce here, the new version should be correct.

zero9178 (approved):

PR #2849, opened 2024-07-05
transformations: Handle affine.apply in lower-affine

superlopuh:
We need this in the generic to loops lowering when not all the arguments are streamed


tests/filecheck/transforms/lower_affine.mlir
```diff
@@ -1,46 +1,61 @@
...
+// CHECK-NEXT:      scf.yield %v3 : f32
+  "affine.yield"(%v3) : (f32) -> ()
+
+// CHECK-NEXT:    }
+}) : (f32) -> f32
+
+// CHECK-NEXT:    %apply_dim, %apply_sym = "test.op"() : () -> (index, index)
+// CHECK-NEXT:    %apply_res = arith.constant 42 : index
+// CHECK-NEXT:    %apply_res_1 = arith.muli %apply_dim, %apply_res : index
+// CHECK-NEXT:    %apply_res_2 = arith.addi %apply_sym, %apply_res_1 : index
```
zero9178, line 53: This seems to be a bug. `apply_dim` and `apply_sym` should be switched since `s0` binds to `apply_sym` and `d0` to `apply_dim`.

superlopuh: nice catch! I'll take a look

zero9178 (approved):
LGTM!

(Side note. Unless we get variables in filecheck, I am starting to think it is much much better to hardode SSA values... The `%{{.*}}` everywhere makes it hard to review tests and would not have cought errors like the one in the first commit)

superlopuh:
I agree completely, I think the key here might be name hints, I'll look into it a bit more in subsequent PRs

PR #2852, opened 2024-07-06
dialects: (linalg) Fix matmul custom syntax and custom init

qaco:


superlopuh (commented):
Thank you for opening this PR! Let's take some time to address some things that would be good to have, such as filecheck tests for the updated assembly format (we usually have two for dialects that are mirrored from MLIR, one for xdsl parsing and printing, and one for mlir compatibility), and also for the init fix (these are usually in pytest).

The text of the PR would also be good to delete, once all the points have been addressed.

superlopuh:
Oh and I would recommend setting up the precommit hooks with `make precommit-install`

qaco:
> Thank you for opening this PR! Let's take some time to address some things that would be good to have, such as filecheck tests for the updated assembly format (we usually have two for dialects that are mirrored from MLIR, one for xdsl parsing and printing, and one for mlir compatibility), and also for the init fix (these are usually in pytest).
> 
> The text of the PR would also be good to delete, once all the points have been addressed.

Thank you for your advices. I have added the suggested tests.

tests/filecheck/parser-printer/matmul_on_memrefs_parsing.mlir
superlopuh, line 1: could you please move this to `tests/filecheck/dialects/linalg/linalg_ops.mlir`?

tests/dialects/test_linalg.py
```diff
@@ -36,6 +36,20 @@ def body(args: tuple[Any, ...]):
...
+    def funcBody(args: tuple[Any, ...]):
+        A = memref.Alloc.get(f32, shape=[100, 50])
+        B = memref.Alloc.get(f32, shape=[50, 100])
+        C = memref.Alloc.get(f32, shape=[100, 100])
+
+        linalg.MatmulOp(inputs=(A.results[0], B.results[0]), outputs=(C.results[0],))
+
+        func.Return()
+
+    func.FuncOp("foo", ([], []), funcBody)
```
superlopuh, line 50:
```suggestion
    A = memref.Alloc.get(f32, shape=[100, 50])
    B = memref.Alloc.get(f32, shape=[50, 100])
    C = memref.Alloc.get(f32, shape=[100, 100])

    matmul_op = linalg.MatmulOp(inputs=(A.results[0], B.results[0]), outputs=(C.results[0],))

    assert tuple(result.type for result in matmul_op.results) == ()
```
We still need to test that the result types are what we expect

tests/filecheck/parser-printer/matmul_on_memrefs_parsing.mlir
qaco: Ok, I just did it !

superlopuh (approved):
LGTM!

PR #2853, opened 2024-07-06
backend: reserve registers not preserved across function calls in RISC-V

superlopuh:
A little correctness improvement, we should only be allocating to s registers if there's a function call, of which there should be 0 for the foreseeable future, but it's nice not to silently miscompile.

### DIFF
--- a/xdsl/backend/riscv/register_allocation.py
+++ b/xdsl/backend/riscv/register_allocation.py
@@ -4,7 +4,7 @@ from itertools import chain
 from ordered_set import OrderedSet
 
 from xdsl.backend.riscv.register_queue import RegisterQueue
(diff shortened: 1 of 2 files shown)


xdsl/backend/riscv/register_allocation.py
```diff
@@ -26,6 +26,12 @@ def gather_allocated(func: riscv_func.FuncOp) -> set[RISCVRegisterType]:
         if not isinstance(op, RISCVAsmOperation):
             continue
 
+        if isinstance(op, riscv_func.CallOp):
```
zero9178, line 29: Probably worth putting a command here as to why this is done for whoever is going to implement call ABI to find and remove this.

zero9178 (approved):
LGTM with small comment

PR #2859, opened 2024-07-07
dialects: (snitch_stream) add custom syntax to streaming region

superlopuh:
Makes some things easier to read.

### DIFF
--- a/xdsl/dialects/snitch_stream.py
+++ b/xdsl/dialects/snitch_stream.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 from collections.abc import Iterator, Sequence
 from itertools import product
+from typing import cast
 
 from xdsl.dialects import riscv
 from xdsl.dialects.builtin import (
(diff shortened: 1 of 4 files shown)


xdsl/dialects/snitch_stream.py
```diff
@@ -248,6 +249,110 @@ def __init__(
...
+
+        if self.outputs:
+            printer.print_string(" outs(")
+            printer.print_list(self.outputs, printer.print_ssa_value)
+            printer.print_string(" : ")
+            printer.print_list((o.type for o in self.outputs), printer.print_attribute)
+            printer.print_string(")")
+
+        if self.attributes:
+            printer.print(" attrs = ")
```
zero9178, line 283: Would it make sense to change this to just `" attributes "` like `attr-with-keyword` in MLIR?

zero9178 (approved):

xdsl/dialects/snitch_stream.py
superlopuh: oh yeah

PR #2863, opened 2024-07-08
docs: (toy) remove all custom instructions from Toy

superlopuh:
I added a bunch of hacks to Toy to get it to work, but we now have all the functionality we need to remove the custom behaviour.


AntonLydike (approved):
Big Red :100: :chart_with_downwards_trend:

PR #2866, opened 2024-07-08
transformations: handle strided layout attr in memref_stream lowering

superlopuh:
We currently treat all memrefs as raw pointers in our memref_stream.streaming_region lowering, but that's not correct, we want to take the strided layout into account when possible. This PR adds some helpers to compute the appropriate stride map, and uses them to construct the correct snitch stride pattern.

### DIFF
--- a/tests/dialects/test_memref.py
+++ b/tests/dialects/test_memref.py
@@ -3,6 +3,7 @@ import pytest
 from xdsl.builder import Builder
 from xdsl.dialects import arith, builtin, func, memref, scf
 from xdsl.dialects.builtin import (
+    AffineMapAttr,
     ArrayAttr,
     FloatAttr,
     IndexType,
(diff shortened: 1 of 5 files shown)


tests/dialects/test_memref.py
```diff
@@ -486,3 +488,20 @@ def test_memref_copy():
         match="Expected source and destination to have the same element type.",
     ):
         copy.verify()
+
+
+def test_get_strides():
+    strided = StridedLayoutAttr((12, 4, 1), 5)
```
zero9178, line 494:
```suggestion
    strided = StridedLayoutAttr((24, 4, 1), 5)
```
This should use a non-identity layout so that the test at line 507 actually checks whether the strides come from `strided`

xdsl/transforms/convert_memref_stream_to_snitch_stream.py
```diff
@@ -166,7 +165,7 @@ def match_and_rewrite(
             rewriter.modify_block_argument_type(arg, stream_type)
 
 
-def offset_map_from_shape(shape: Sequence[int], factor: int) -> AffineMap:
+def strides_map_from_memref_type(memref_type: MemRefType[AttributeCovT]) -> AffineMap:
     """
     Given a list of lengths for each dimension of a memref, and the number of bytes per
     element, returns the map from indices to an offset in bytes in memory. The resulting
```
zero9178, line 171:
```suggestion
    Given a memref returns the map from indices to an offset in bytes in memory. The resulting
```

zero9178 (approved):

PR #2869, opened 2024-07-08
transformations: Add memref-stream-tile-outer-loops

superlopuh:
We sometimes have too high a rank when lowering to snitch_stream, this avoids that at the memref_stream level, before streamification.

### DIFF
--- a/xdsl/transforms/memref_stream_tile_outer_loops.py
+++ b/xdsl/transforms/memref_stream_tile_outer_loops.py
@@ -28,20 +28,30 @@ from xdsl.utils.exceptions import DiagnosticException
 
 
 def insert_subview(
-    val: SSAValue,
-    m: AffineMapAttr,
+    memref_val: SSAValue,
+    affine_map: AffineMap,
     apply_operands: Sequence[SSAValue],
     upper_bounds: Sequence[int],
     location: InsertPoint,
 ):
-
-    name_hint_prefix = val.name_hint + "_" if val.name_hint is not None else ""
+    """
+    A helper method to insert a subview from the input `memref_val` to one with new upper
+    bounds, given that it will be accessed with the specified affine map.
+    `apply_operands` are the operands to use to determine the new offset,
+    and `upper_bounds` the new shape.
+    Any new operations should be inserted at `location`.
+    """
+    name_hint_prefix = (
+        memref_val.name_hint + "_" if memref_val.name_hint is not None else ""
+    )
     apply_ops = tuple(
         affine.ApplyOp(
             apply_operands,
-            AffineMapAttr(AffineMap(m.data.num_dims, m.data.num_symbols, (res,))),
+            AffineMapAttr(
+                AffineMap(affine_map.num_dims, affine_map.num_symbols, (res,))
+            ),
         )
-        for res in m.data.results
+        for res in affine_map.results
     )
     offset_vals = tuple(apply_op.result for apply_op in apply_ops)
     for offset_val in offset_vals:
@@ -54,7 +64,7 @@ def insert_subview(
         apply_op.map.data.eval(max_values, ())[0] + 1 for apply_op in apply_ops
     )
 
-    source_type = val.type
+    source_type = memref_val.type
     if not isinstance(source_type, memref.MemRefType):
         raise DiagnosticException("Cannot create subview from non-memref type")
     source_type = cast(MemRefType[Attribute], source_type)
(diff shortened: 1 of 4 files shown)


xdsl/transforms/memref_stream_tile_outer_loops.py
```diff
@@ -0,0 +1,216 @@
...
+    memref_val: SSAValue,
+    affine_map: AffineMap,
+    apply_operands: Sequence[SSAValue],
+    upper_bounds: Sequence[int],
+    location: InsertPoint,
+):
+    """
+    A helper method to insert a subview from the input `memref_val` to one with new upper
+    bounds, given that it will be accessed with the specified affine map.
+    `apply_operands` are the operands to use to determine the new offset,
```
zero9178, line 40:
```suggestion
    `offset` are the indices to use to determine the new offset,
```
The `apply_operands` name is an implementation detail. If it represents the offset then lets name it as such

xdsl/transforms/memref_stream_tile_outer_loops.py
```diff
@@ -0,0 +1,216 @@
...
+    offset_vals = tuple(apply_op.result for apply_op in apply_ops)
+    for offset_val in offset_vals:
+        offset_val.name_hint = name_hint_prefix + "offset"
+    Rewriter.insert_op(apply_ops, location)
+
+    # New subview shape
+    max_values = tuple(ub - 1 for ub in upper_bounds)
+    sizes = tuple(
+        apply_op.map.data.eval(max_values, ())[0] + 1 for apply_op in apply_ops
+    )
```
zero9178, line 65: Can you explain the story and logic here? I.e. why specifically this value is chosen (it being one larger than the highest accessed based on the access pattern given by the affine map)

xdsl/transforms/memref_stream_tile_outer_loops.py
```diff
@@ -0,0 +1,216 @@
...
+
+    source_type = memref_val.type
+    if not isinstance(source_type, memref.MemRefType):
+        raise DiagnosticException("Cannot create subview from non-memref type")
```
zero9178, line 69: (would be cool to have a typed SSA value in xDSL to move such checks to `pyright`)

xdsl/transforms/memref_stream_tile_outer_loops.py
```diff
@@ -0,0 +1,216 @@
...
+    source_type = cast(MemRefType[Attribute], source_type)
+    layout_attr = source_type.layout
+    strides = tuple(source_type.get_strides())
+    if isinstance(layout_attr, NoneAttr):
+        layout_attr = StridedLayoutAttr(strides, None)
+        dest_type = MemRefType(source_type.element_type, sizes, layout_attr)
+    else:
+        dest_type = source_type
```
zero9178, line 77: I don't think is correct in general, but works for this specific pass since you always keep the strides, and make all offsets dynamic. 

Could you either:
* Leave a comment here that the `else` branch depends on this assumption? Maybe add an assertion too
or
* Fix it so that if your source type already has offsets, but they happen to be static offsets, that it turns it to a dynamic offset as well.

xdsl/transforms/memref_stream_tile_outer_loops.py
```diff
@@ -0,0 +1,216 @@
...
+    subview_op.result.name_hint = name_hint_prefix + "subview"
+    Rewriter.insert_op(subview_op, location)
+    return subview_op.result
+
+
+def materialize_loop(
+    rewriter: PatternRewriter, generic_op: memref_stream.GenericOp, index: int
+) -> Sequence[Operation]:
+    """
+    Replaces a given generic op with a for loop containing an op with the ub at the
```
zero9178, line 95:
```suggestion
    Replaces a given generic op with a for loop containing an op with the upper bound at the
```

xdsl/transforms/memref_stream_tile_outer_loops.py
```diff
@@ -0,0 +1,216 @@
...
+    specified index set to 1.
+    """
+    if (
+        generic_op.iterator_types.data[index].data
+        != memref_stream.IteratorType.PARALLEL
+    ):
+        raise DiagnosticException("Cannot ")
```
zero9178, line 102: This comment littarly "Cannot " right now :slightly_smiling_face:

xdsl/transforms/memref_stream_tile_outer_loops.py
superlopuh: I agree :)

xdsl/transforms/memref_stream_tile_outer_loops.py
superlopuh: Oh yeah forgot about this

zero9178 (approved):

PR #2871, opened 2024-07-09
core: (rewriting) Fix recursive type conversion in block arguments for nested regions

### LINKED ISSUE #2857: Type Conversion of function arguments
I am unsure if this is intended behavior or I am doing something incorrectly, but when I try using a TypeConversionPattern on the arguments of a FuncOp, the argument types do not change even with the recursive flag set to True.

An example:  
```python
from xdsl.dialects.func import Func
    Builtin,
    IndexType,
    IntegerType,
)
from xdsl.pattern_rewriter import (
    PatternRewriteWalker,
    TypeConversionPattern,
    attr_type_rewrite_pattern,
)


prog = """\
"builtin.module"() ({
  "func.func"() <{function_type = (memref<2x4xui16>, memref<2x3xui8>) -> (), sym_name = "main", sym_visibility = "private"}> ({
  ^bb0(%arg0: memref<2x4xui16>, %arg1: memref<2x3xui8>):
    "func.return"() : () -> ()
  }) : () -> ()
}) : () -> ()
"""

ctx = MLContext()
ctx.load_dialect(Builtin)
ctx.load_dialect(Func)
parser = Parser(ctx, prog)
module = parser.parse_module()


class Rewrite(TypeConversionPattern):
    @attr_type_rewrite_pattern
    def convert_type(self, typ: IntegerType) -> IndexType:
        return IndexType()


PatternRewriteWalker(Rewrite(recursive=True), apply_recursively=True).rewrite_module(
    module
)
```

The output of the above code snippet is:
```mlir
builtin.module {
  func.func private @main(%arg0 : memref<2x4xui16>, %arg1 : memref<2x3xui8>) {
    func.return
  }
}
```
while I expected:
```mlir
builtin.module {
  func.func private @main(%arg0 : memref<2x4xindex>, %arg1 : memref<2x3xindex>) {
    func.return
  }
}
```

webmiche:
closes #2857 

Solves the issue mentioned there and adds the reported testcase as a regression test.

Before, this was actually quite broken as not applying the recursive rewrite, but the normal rewrite, to the region would actually mean that the produced func_op did not even verify (the `function_type` attribute got rewritten properly, whereas the entry_block args would still have the old type). 

The reason why the old type was being printed is that the `print_func_op_like`-utility function defaults to printing the entry_block arguments in some cases.

### DIFF
--- a/tests/pattern_rewriter/test_pattern_rewriter.py
+++ b/tests/pattern_rewriter/test_pattern_rewriter.py
@@ -1610,6 +1610,41 @@ def test_type_conversion():
     )
 
 
+def test_recursive_type_conversion_in_regions():
+
+    prog = """\
+"builtin.module"() ({
+  "func.func"() <{"function_type" = (memref<2x4xui16>) -> (), "sym_name" = "main", "sym_visibility" = "private"}> ({
+  ^bb0(%arg0 : memref<2x4xui16>):
+    "func.return"() : () -> ()
+  }) : () -> ()
+}) : () -> ()
+"""
+    expected_prog = """\
+"builtin.module"() ({
+  "func.func"() <{"function_type" = (memref<2x4xindex>) -> (), "sym_name" = "main", "sym_visibility" = "private"}> ({
+  ^0(%arg0 : memref<2x4xindex>):
+    "func.return"() : () -> ()
+  }) : () -> ()
+}) : () -> ()
+"""
+
+    class IndexConversion(TypeConversionPattern):
+
+        @attr_type_rewrite_pattern
+        def convert_type(self, typ: IntegerType) -> IndexType:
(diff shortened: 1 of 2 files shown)


webmiche:
@francescodaghero feel free to also comment/leave a review :)

superlopuh (approved):

webmiche:
@superlopuh do you think we can merge this without @PapyChacal s approval? It seems like fairly obvious fix to me, but I am not very deep in the intricacies of the TypeConversion stuff.

webmiche:
Hmm, yeah fair enough. I am honestly not sure what we want this API to look like exactly, but your change sounds reasonable.

webmiche:
I will wait for the CI to pass and then merge.

PR #2872, opened 2024-07-09
dialects: (riscv_snitch) Add f32 mul, add, pack from Snitch packed SIMD extension

nazavode:
This PR adds a small set of operations to `riscv_snitch` directly mapping on top of the Snitch custom packed SIMD ISA. Further ops are going to need operand constraints to be correctly register allocated.

### DIFF
--- a/tests/filecheck/dialects/riscv_snitch/ops.mlir
+++ b/tests/filecheck/dialects/riscv_snitch/ops.mlir
@@ -87,11 +87,11 @@ riscv_func.func @xdma() {
 }
 
 riscv_func.func @simd() {
-  // CHECK-LABEL: @simd
   %v = riscv.get_float_register : !riscv.freg
+  // CHECK: %v = riscv.get_float_register : !riscv.freg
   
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/riscv_snitch/ops.mlir
```diff
@@ -86,6 +86,22 @@ riscv_func.func @xdma() {
   riscv_func.return
 }
 
+riscv_func.func @simd() {
+  // CHECK-LABEL: @simd
```
AntonLydike, line 90: Nit: `CHECK-LABEL` is not (yet) supported by our filecheck, we are working on switching to a new one that does support everything though!

AntonLydike (approved):
Nice work! Just one nitpick about the use of `CHECK-LABEL`

tests/filecheck/dialects/riscv_snitch/ops.mlir
nazavode: Ouch, thanks for spotting!

superlopuh (commented):
could you please add assembly printing tests?

superlopuh (approved):

PR #2878, opened 2024-07-11
(transform): csl-stencil to csl-wrapper pass

n-io:
Pass to lower `csl_stencil` to `csl_wrapper`.

* Creates one `csl_wrapper.module` for each module-level function
* Extracts stencil-related params to initialise `csl_wrapper.modue` and both of its `layout_module` and `program_module` regions
* Initialises `layout_module` with everything required for running stencil programs and yields it to `program_module`, specifically: memcpy params, routes config, and `isBorderRegionPE` bool

Not in this PR:
* ~~Handling host-device transferable symbols and setting up exports~~
* ~~Setting up and exporting program entry point~~

~~As such, the translated func is currently simply placed inside the `program_module`.~~

Update:
* Host-device transferable symbols are now added to the `program_module` block args
* The function op is resolved
* The function `sym_name` is stored as a property on the module
* ~~The function body is moved into the program module body~~
* We set up a main entry point `csl.func` named after the translated `func.func`, transfer the ops into the new csl main func, and add the function to the program module body

### DIFF
--- a/tests/filecheck/transforms/csl-stencil-to-csl-wrapper.mlir
+++ b/tests/filecheck/transforms/csl-stencil-to-csl-wrapper.mlir
@@ -57,39 +57,36 @@ builtin.module {
 // CHECK-NEXT:     %22 = arith.ori %21, %19 : i1
 // CHECK-NEXT:     "csl_wrapper.yield"(%11, %10, %22) <{"fields" = ["memcpy_params", "stencil_comms_params", "isBorderRegionPE"]}> : (!csl.comptime_struct, !csl.comptime_struct, i1) -> ()
 // CHECK-NEXT:   }, {
-// CHECK-NEXT:   ^1(%23 : i16, %24 : i16, %25 : i16, %26 : i16, %memcpy_params : !csl.comptime_struct, %stencil_comms_params : !csl.comptime_struct, %isBorderRegionPE : i1):
-// CHECK-NEXT:     func.func @gauss_seidel(%a : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>, %b : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>) {
-// CHECK-NEXT:       %27 = stencil.load %a : !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>> -> !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
-// CHECK-NEXT:       %28 = "csl_stencil.prefetch"(%27) <{"topo" = #dmp.topo<1022x510>, "swaps" = [#csl_stencil.exchange<to [1, 0]>, #csl_stencil.exchange<to [-1, 0]>, #csl_stencil.exchange<to [0, 1]>, #csl_stencil.exchange<to [0, -1]>]}> : (!stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>) -> memref<4xtensor<510xf32>>
-// CHECK-NEXT:       %29 = tensor.empty() : tensor<510xf32>
-// CHECK-NEXT:       %30 = csl_stencil.apply(%27 : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>, %29 : tensor<510xf32>) <{"swaps" = [#csl_stencil.exchange<to [1, 0]>, #csl_stencil.exchange<to [-1, 0]>, #csl_stencil.exchange<to [0, 1]>, #csl_stencil.exchange<to [0, -1]>], "topo" = #dmp.topo<1022x510>, "num_chunks" = 2 : i64}> -> (!stencil.temp<[0,1]x[0,1]xtensor<510xf32>>) ({
-// CHECK-NEXT:       ^2(%31 : memref<4xtensor<255xf32>>, %32 : index, %33 : tensor<510xf32>):
-// CHECK-NEXT:         %34 = csl_stencil.access %31[1, 0] : memref<4xtensor<255xf32>>
-// CHECK-NEXT:         %35 = csl_stencil.access %31[-1, 0] : memref<4xtensor<255xf32>>
-// CHECK-NEXT:         %36 = csl_stencil.access %31[0, 1] : memref<4xtensor<255xf32>>
-// CHECK-NEXT:         %37 = csl_stencil.access %31[0, -1] : memref<4xtensor<255xf32>>
-// CHECK-NEXT:         %38 = arith.addf %35, %34 : tensor<255xf32>
-// CHECK-NEXT:         %39 = arith.addf %37, %36 : tensor<255xf32>
-// CHECK-NEXT:         %40 = arith.addf %39, %38 : tensor<255xf32>
-// CHECK-NEXT:         %41 = "tensor.insert_slice"(%40, %33, %32) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 255>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 1, 1, 0, 0>}> : (tensor<255xf32>, tensor<510xf32>, index) -> tensor<510xf32>
-// CHECK-NEXT:         csl_stencil.yield %41 : tensor<510xf32>
-// CHECK-NEXT:       }, {
-// CHECK-NEXT:       ^3(%42 : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>, %43 : tensor<510xf32>):
-// CHECK-NEXT:         %44 = csl_stencil.access %42[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
-// CHECK-NEXT:         %45 = csl_stencil.access %42[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
-// CHECK-NEXT:         %46 = arith.constant 1.666600e-01 : f32
-// CHECK-NEXT:         %47 = "tensor.extract_slice"(%44) <{"static_offsets" = array<i64: 1>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
-// CHECK-NEXT:         %48 = "tensor.extract_slice"(%45) <{"static_offsets" = array<i64: -1>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
-// CHECK-NEXT:         %49 = arith.addf %43, %48 : tensor<510xf32>
-// CHECK-NEXT:         %50 = arith.addf %49, %47 : tensor<510xf32>
-// CHECK-NEXT:         %51 = tensor.empty() : tensor<510xf32>
-// CHECK-NEXT:         %52 = linalg.fill ins(%46 : f32) outs(%51 : tensor<510xf32>) -> tensor<510xf32>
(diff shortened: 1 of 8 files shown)


tests/filecheck/transforms/csl-stencil-to-csl-wrapper.mlir
```diff
@@ -0,0 +1,95 @@
...
+      %10 = csl_stencil.access %4[0, -1] : memref<4xtensor<255xf32>>
+      %11 = arith.addf %8, %7 : tensor<255xf32>
+      %12 = arith.addf %10, %9 : tensor<255xf32>
+      %13 = arith.addf %12, %11 : tensor<255xf32>
+      %14 = "tensor.insert_slice"(%13, %6, %5) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 255>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 1, 1, 0, 0>}> : (tensor<255xf32>, tensor<510xf32>, index) -> tensor<510xf32>
+      csl_stencil.yield %14 : tensor<510xf32>
+    }, {
+    ^1(%15 : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>, %16 : tensor<510xf32>):
+      %17 = csl_stencil.access %15[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
+      %18 = csl_stencil.access %15[0, 0] : !stencil.temp<[-1,2]x[-1,2]xtensor<512xf32>>
```
AntonLydike, line 22: Is there any particular reason why we need two access ops here?

xdsl/dialects/csl/csl_wrapper.py
```diff
@@ -303,6 +311,18 @@ def layout_yield_op(self) -> YieldOp:
...
         return cast(YieldOp, self.layout_module.block.last_op)
 
+    @property
+    def exported_symbols(self) -> Sequence[BlockArgument]:
+        """
+        Get the exported symbols.
+        """
+        return self.program_module.block.args[
+            2 + len(self.params) + len(self.layout_yield_op.fields) :
+        ]
```
AntonLydike, line 321: Can we get a test for this? Index magic always makes me anxious :sweat_smile:

xdsl/dialects/csl/csl_wrapper.py
```diff
@@ -303,6 +311,18 @@ def layout_yield_op(self) -> YieldOp:
...
+
+    def set_program_name(self, name: StringAttr):
+        self.program_name = name
```
AntonLydike, line 324: Why would you use this helper instead of assigning to the property directly? Am I missing something here?

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+        height: int = 1
+        z_dim_no_ghost_cells: int = 1
+        z_dim: int = 1
+        for apply_op in apply_ops:
+            # loop over accesses to get `neighbour` (from which we build `pattern`)
+            for ap in apply_op.get_accesses():
+                if ap.is_diagonal:
+                    raise ValueError("Diagonal accesses are currently not supported")
+                # if ap.dims != 2:
+                #     raise ValueError("Stencil accesses must be 2-dimensional at this stage")
```
AntonLydike, line 50: We have some commented code here, maybe replace it with a comment on why we don't need to check that the access is 2d?

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+                if len(ap.offsets) > 0:
+                    neighbours = max(
+                        neighbours, max(abs(d) for offset in ap.offsets for d in offset)
+                    )
```
AntonLydike, line 54: This is computing the max radius, right? I would love to see (in increasing order of happiness for me):

- a comment here explaining what you are calculating and renaming `neighbors` to `max_radius`
- moving this to a helper method for increased readability
- moving this logic to `AccessPattern.radius() -> int` and adding tests

Imagine how readable the code would be as:

```py
max_radius = max(max_radius, ap.radius())
```
no comments needed! It's literally self-documenting!

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+    def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
+        # find csl_stencil.apply ops, abort if there are none
+        apply_ops = self.get_csl_stencil_apply_ops(op)
+        if len(apply_ops) == 0:
+            return
+        neighbours: int = 1
+        width: int = 1
+        height: int = 1
+        z_dim_no_ghost_cells: int = 1
+        z_dim: int = 1
```
AntonLydike, line 43: Would love to see a quick comment on these explaining what they are tracking (especially `neighbors`, as I suspect this is in actuality `max_radius`?).

xdsl/dialects/csl/csl_wrapper.py
```diff
@@ -142,12 +143,18 @@ class ModuleOp(IRDLOperation):
 
     The layout module has two additional block args `x` and `y` as part of the `@set_tile_code` loop nest.
     Operations using these args need to be lowered to the correct place in the loop nest.
+
+    The program module has the following args (in order):
+      * general params:        `width` and `height` followed by everything specified in `params`
+      * params from layout:    everything defined by `layout_yield_op.fields`
+      * input-output symbols:  any arg from the function lowered into this op, which are exported symbols supporting host-device transfers
```
AntonLydike, line 150: Are there some properties of these in-out symbols that we need to verify?

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+                block_arg.type, len(module_op.program_module.block.args)
+            )
+
+        # replace func.return
+        func_return = op.body.block.last_op
+        assert isinstance(func_return, func.Return)
+        assert (
+            len(func_return.arguments) == 0
+        ), "Non-empty returns currently not supported"
+        rewriter.replace_op(func_return, csl.ReturnOp())
```
AntonLydike, line 106: How would we handle return values returned by the function?

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+
+        self.initialise_layout_module(module_op)
+        module_op.update_program_block_args_from_layout()
+        module_op.set_program_name(op.sym_name)
+
+        # add func args to program_module block args
+        for block_arg in op.body.block.args:
+            module_op.program_module.block.insert_arg(
+                block_arg.type, len(module_op.program_module.block.args)
+            )
```
AntonLydike, line 98: I wonder if it'd make sense to move this onto `csl_wrapper.module` just like the others above? Maybe something along the lines `ModuleOp.add_program_in_outs(self, types: Sequence[Attribtue]) -> Sequence[BlockArgument]`?

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+    def initialise_layout_module(self, module_op: csl_wrapper.ModuleOp):
+        """Initialises the layout_module (wrapper block) by setting up (esp. stencil-related) program params"""
+
+        # extract layout module params as the function has linear complexity
+        param_width = module_op.get_layout_param("width")
+        param_height = module_op.get_layout_param("height")
+        param_x = module_op.get_layout_param("x")
+        param_y = module_op.get_layout_param("y")
+        param_pattern = module_op.get_layout_param("pattern")
+
```
AntonLydike, line 140: I think I'd personally prefer the style of:

```py
width, height, x, y, pattern = module_op.get_layout_params()
```

But alas, this is what we are doing now. Feel free to ignore my nitpicks

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -0,0 +1,232 @@
...
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        module_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    ConvertStencilFuncToModuleWrappedPattern(),
+                ]
+            ),
+            walk_reverse=False,
+            apply_recursively=False,
```
AntonLydike, line 230: Is there a reason why you specify these two parameters here?

AntonLydike (commented):
Nice stuff! Would always love more comments, but the code is already quite a bit more readable than before!

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: By the above assert 🤓

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: Just so that it's clear what it does.

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: That would prescribe an implementation-specific order and elevate pattern to always-required, hard-coding something I'm hoping not to hard-code.

xdsl/transforms/csl_stencil_to_csl_wrapper.py
AntonLydike: I think it's safe to drop at least `walk_reverse`, as it's the default value anyway...

tests/filecheck/transforms/csl-stencil-to-csl-wrapper.mlir
n-io: We could canonicalise it at an earlier stage, it's a relic from the tensorisation pass.

xdsl/dialects/csl/csl_wrapper.py
n-io: changing as suggested

xdsl/dialects/csl/csl_wrapper.py
AntonLydike: Can they be anything, or do we have some type restrictions? I'm asking because I'm not sure myself. I would guess that they need to be `memref`?

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: Combined this with the `update_program_args` method to avoid having two methods that need to be called in the correct order.

xdsl/dialects/csl/csl_wrapper.py
n-io: Added as suggested

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: ok

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: It breaks but I think it shouldn't, need to investigate.

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: I'm actually not sure how we could export un-named return values for host-device transfers.

xdsl/dialects/csl/csl_wrapper.py
n-io: @dk949 ?

n-io: Actually, I noticed we're using it for stencil.fields right now. My suggestion is to not restrict it for the time being.

AntonLydike (approved):

PR #2887, opened 2024-07-15
backend: (riscv) add tied operands/results to register allocation

nazavode:
This PR adds the concept of *tied operands* found in LLVM to be able to model 2-address instructions, a.k.a. when an operand register is both read and written in-place by an instruction. Both the operand and the result must be represented in their own SSA values, but the regalloc must allocate both on the same register to be able to generate correct code.

**Keeping as draft for feedback**, things that I would like to change:

* I don't like the APIs, it could be nice to have our own `TiedConstraint` instead of a generic `ConstraintVar("T")`
* ~~The code that collects all operands/results definitions that share the same `ConstraintVar` is a mess, to be cleaned up.~~

### DIFF
--- a/tests/backend/riscv/test_register_allocation_constraints.py
+++ b/tests/backend/riscv/test_register_allocation_constraints.py
@@ -42,7 +42,7 @@ class TestOp4(Generic[RD1InvT, RD2InvT, RS1InvT, RS2InvT], RISCVInstruction, ABC
 SameIntRegisterType: TypeAlias = Annotated[IntRegisterType, ConstraintVar("T")]
 
 
-def test_constrained_single():
+def test_tie_constrained_single():
     @irdl_op_definition
     class ConstrainedOp(
         TestOp4[SameIntRegisterType, IntRegisterType, IntRegisterType, IntRegisterType]
@@ -82,7 +82,7 @@ def test_constrained_single():
     assert constr.operand_is_constrained_to(3) is None
 
 
-def test_constrained_all_unallocated():
+def test_tie_constrained_all_unallocated():
     @irdl_op_definition
     class ConstrainedOp(
         TestOp4[
@@ -127,7 +127,7 @@ def test_constrained_all_unallocated():
     assert constr.operand_is_constrained_to(1) == Registers.A1
 
 
-def test_constrained_all_allocated():
+def test_tie_constrained():
     @irdl_op_definition
     class ConstrainedOp(
         TestOp4[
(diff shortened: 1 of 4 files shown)


tests/backend/riscv/test_register_allocation_constraints.py
```diff
@@ -0,0 +1,177 @@
...
+
+class TestOp4(Generic[RD1InvT, RD2InvT, RS1InvT, RS2InvT], RISCVInstruction, ABC):
+
+    rd1: OpResult = result_def(RD1InvT)
+    rd2: OpResult = result_def(RD2InvT)
+    rs1: Operand = operand_def(RS1InvT)
+    rs2: Operand = operand_def(RS2InvT)
+
+    def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
+        return self.rd1, self.rd2, self.rs1, self.rs2
```
superlopuh, line 39: I was thinking you were going to do something a bit more like this:
```suggestion
class TestOp4(Generic[RdRsInvT, RD2InvT, RS2InvT], RISCVInstruction, ABC):

    SameIntRegisterType: TypeAlias = Annotated[RdRsInvT, ConstraintVar("RdRs")]

    rd1: OpResult = result_def(SameIntRegisterType)
    rd2: OpResult = result_def(RD2InvT)
    rs1: Operand = operand_def(SameIntRegisterType)
    rs2: Operand = operand_def(RS2InvT)

    def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
        return self.rd1, self.rd2, self.rs1, self.rs2
```
So embedding in IRDL the fact that it's the same variable, as opposed to constraining the generic parameter itself.

superlopuh (commented):
It might be interesting to split this in two, to first add the constraint logic and use it in our current loop allocation code (to constrain that the iter args, block args, and results are in the same registers) and then to use it for operations with same register constraints like in this PR

tests/backend/riscv/test_register_allocation_constraints.py
nazavode: Much nicer! In this case tho I'm instantiating the generic class with different tie constraints for testing reasons, your suggestion would be 100% error proof when defining real world operations.

nazavode:
> It might be interesting to split this in two, to first add the constraint logic and use it in our current loop allocation code (to constrain that the iter args, block args, and results are in the same registers) and then to use it for operations with same register constraints like in this PR

This PR adds just the bare logic and a a handful of fictional ops just for testing purposes. I think trying to apply it to loops allocation first would be a strong stress test before having the whole thing merged. Trying it out on a stacked branch!

superlopuh:
What's the plan for this? Did we merge the equivalent logic or does this branch go further than what we came up with in Zurich?

nazavode:
> What's the plan for this? Did we merge the equivalent logic or does this branch go further than what we came up with in Zurich?

I think that all we need has been merged already, happy to close this!

superlopuh:
Fantastico

PR #2889, opened 2024-07-16
api: add generic parameter to attribute constraint

superlopuh:
PR 1/2 to make the Pylance errors in operation definitions go away.

The main idea is for the constraint type to carry the upper bound of the Attribute types that will pass verification. It is not guaranteed to be a strict upper bound.

Pyright errors: 489.

### DIFF
--- a/xdsl/utils/isa.py
+++ b/xdsl/utils/isa.py
@@ -9,6 +9,10 @@ from xdsl.utils.hints import isa
 def isattr(
     arg: Any, hint: type[AttributeInvT] | GenericAttrConstraint[AttributeInvT]
 ) -> TypeGuard[AttributeInvT]:
+    """
+    A helper method to check whether a given attribute has a given type or conforms to a
+    given constraint.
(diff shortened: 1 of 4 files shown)


xdsl/utils/isa.py
```diff
@@ -0,0 +1,21 @@
...
+
+from xdsl.ir import AttributeInvT
+from xdsl.irdl import GenericAttrConstraint
+from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
+
+
+def isattr(
+    arg: Any, hint: type[AttributeInvT] | GenericAttrConstraint[AttributeInvT]
+) -> TypeGuard[AttributeInvT]:
```
AntonLydike, line 11: can I has docstring?

AntonLydike (approved):

math-fehr (approved):
Nice, I like where this is going.
Thanks for doing the unicorn's work :pray:

PR #2890, opened 2024-07-16
api: use constraints to carry type information in pyrdl

superlopuh:
PR 2/2 to make the Pylance errors in operation definitions go away.

The proposal is to introduce a `constr` mentod to construct a constraint given a PyRDL type, add `|` and `&` operators on constraints to combine them, and to require a constraint instead of a type for operand, result, attribute, and property definitions.

Pylance errors: 317 (-172)


xdsl/irdl/attributes.py
```diff
@@ -431,3 +432,9 @@ def irdl_to_attr_constraint(
         )
 
     raise ValueError(f"Unexpected irdl constraint: {irdl}")
+
+
+def constr(
+    irdl: GenericAttrConstraint[AttributeInvT] | AttributeInvT | type[AttributeInvT],
+) -> GenericAttrConstraint[AttributeInvT]:
+    return cast(GenericAttrConstraint[AttributeInvT], irdl_to_attr_constraint(irdl))
```
AntonLydike, line 440: can I has docstring to explain why it's necessary?

AntonLydike (approved):
Can we get some documentation on what `constr` is, how we should use it, and why we need it? 

Otherwise GTG

xdsl/irdl/constraints.py
```diff
@@ -72,12 +72,17 @@ def get_unique_base(self) -> type[Attribute] | None:
         """Get the unique base type that can satisfy the constraint, if any."""
         return None
 
+    def __or__(
+        self, value: GenericAttrConstraint[AttributeInvT], /
+    ) -> AnyOf[AttributeCovT | AttributeInvT]:
```
math-fehr, line 77:
```suggestion
    def __or__(
        self, value: GenericAttrConstraint[AttributeCovT2], /
    ) -> AnyOf[AttributeCovT | AttributeCovT2]:
```
Both arguments should be covariant here I believe

xdsl/irdl/test.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Generic, TypeGuard, TypeVar
+
```
math-fehr, line 5: What is this testing?

xdsl/irdl/attributes.py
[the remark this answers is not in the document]
superlopuh: Actually I'll look into how bad things would get if I were to give it this signature now, maybe it won't be that big a diff.

xdsl/irdl/constraints.py
superlopuh: I'll give this a go

docs/Toy/toy/dialects/toy.py
```diff
@@ -46,10 +47,12 @@
 )
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import isa
+from xdsl.utils.isattr import isattr
 
 TensorTypeF64: TypeAlias = TensorType[Float64Type]
 UnrankedTensorTypeF64: TypeAlias = UnrankedTensorType[Float64Type]
 AnyTensorTypeF64: TypeAlias = TensorTypeF64 | UnrankedTensorTypeF64
+AnyTensorTypeF64Constr = constr(TensorTypeF64) | constr(UnrankedTensorTypeF64)
```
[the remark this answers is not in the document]
superlopuh: I think it is allowed right now, it's the union that isn't, this is what this line provides.

xdsl/irdl/test.py
superlopuh: Oops

xdsl/irdl/attributes.py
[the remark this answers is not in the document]
superlopuh: OK it's a necessary temporary thing, at least until the things that irdl_to_attr_constraint are a little more strongly typed than just being `Any`, which is the current case in a few places it's called.

superlopuh: Added docstrings

PR #2891, opened 2024-07-16
(transform): csl_wrapper program module init

n-io:
This PR initialises the program_module of a CSL wrapper. Computes additional known values to remove any dependency on `utils.csl`.

Minor fixes:
* Add utility function `get_param_value` to csl_wrapper.ModuleOp
* small fix to `get_apply` on csl_stencil.AccessOp to work for both `csl.Apply` or `csl_stencil.Apply`


xdsl/dialects/csl/csl_wrapper.py
```diff
@@ -323,6 +323,20 @@ def get_program_param(self, name: str) -> BlockArgument:
         # not found = value error
         raise ValueError(f"{name} does not refer to a block arg of this program_module")
 
+    def get_param_value(self, name: str) -> IntegerAttr[IntegerType]:
```
dk949, line 326: This looks really useful to avoid having to use params as SSA values, but I don't fully understand how this works when params have no default values (or where the default value has been replaced by a compiler flag).

n-io: In that case, you'd get a ValueError saying `"Parameter name is unknown or has no value"`. The alternative would have been to return `IntegerAttr[IntegerType] | None`, but it leads to ugly type-checking all over the place. So this function should only be called for a param that you know has a value.

n-io: Instead of leading to "type-checking all over the place", it practically probably leads to errors being raised in those placed, given that I only call this function when I need the param to have a value and otherwise I want to throw an error either here or at the caller site.

dk949: I am somewhat concerned for the case where the param has a default value that has been overridden. E.g. pattern is defaulted to 2, then set to 5 with a compiler flag, but all offset calculations would be done as if the value is 2.

This could also result in fairly fragile transformations. Removing the value of a param will not result in a verification error in the wrapper but will cause the transformation to fail. I'm not sure how much we care about this though.

Perhaps there should be two kinds of params, ones where value can be changed by compiler flags and ones where it can't. Though that is outside of the scope of this PR.

dk949 (approved):
I'm generally ok merging this, but I would like to hear @AntonLydike 's opining on `get_param_value`

xdsl/dialects/csl/csl_wrapper.py
n-io: Yes, I don't think these param values would be safe to change after being set.

PR #2904, opened 2024-07-19
core: Add FixedbitWidthType Interface

jorendumoulin:
These functions to determine the bit/byte width of an element are useful in many places. I think I have them implemented in a couple of different places myself, so I think this abstraction is quite useful in this regard. The functions were now defined in a pass, I simply moved them over.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -353,7 +353,7 @@ class SignednessAttr(Data[Signedness]):
                 raise ValueError(f"Invalid signedness {data}")
 
 
-class BitWidthType(TypeAttribute, ABC):
+class FixedBitwidthType(TypeAttribute, ABC):
     """
     A type attribute with a defined bitwidth
     """
@@ -385,7 +385,7 @@ class BitWidthType(TypeAttribute, ABC):
 
 
 @irdl_attr_definition
-class IntegerType(ParametrizedAttribute, BitWidthType):
+class IntegerType(ParametrizedAttribute, FixedBitwidthType):
     name = "integer_type"
     width: ParameterDef[IntAttr]
     signedness: ParameterDef[SignednessAttr]
@@ -534,7 +534,7 @@ class _FloatType(ABC):
 
 
 @irdl_attr_definition
-class BFloat16Type(ParametrizedAttribute, BitWidthType, _FloatType):
+class BFloat16Type(ParametrizedAttribute, FixedBitwidthType, _FloatType):
     name = "bf16"
 
     @property
@@ -543,7 +543,7 @@ class BFloat16Type(ParametrizedAttribute, BitWidthType, _FloatType):
 
 
 @irdl_attr_definition
-class Float16Type(ParametrizedAttribute, BitWidthType, _FloatType):
+class Float16Type(ParametrizedAttribute, FixedBitwidthType, _FloatType):
     name = "f16"
 
     @property
@@ -552,7 +552,7 @@ class Float16Type(ParametrizedAttribute, BitWidthType, _FloatType):
 
 
 @irdl_attr_definition
-class Float32Type(ParametrizedAttribute, BitWidthType, _FloatType):
+class Float32Type(ParametrizedAttribute, FixedBitwidthType, _FloatType):
     name = "f32"
 
     @property
@@ -561,7 +561,7 @@ class Float32Type(ParametrizedAttribute, BitWidthType, _FloatType):
 
 
 @irdl_attr_definition
-class Float64Type(ParametrizedAttribute, BitWidthType, _FloatType):
+class Float64Type(ParametrizedAttribute, FixedBitwidthType, _FloatType):
     name = "f64"
 
     @property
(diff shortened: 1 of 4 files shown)


xdsl/dialects/builtin.py
```diff
@@ -353,8 +353,39 @@ def print_parameter(self, printer: Printer) -> None:
                 raise ValueError(f"Invalid signedness {data}")
 
 
+class BitWidthType(TypeAttribute, ABC):
```
superlopuh, line 356:
```suggestion
class FixedBitWidthType(TypeAttribute, ABC):
```
?

xdsl/dialects/builtin.py
```diff
@@ -353,8 +353,39 @@ def print_parameter(self, printer: Printer) -> None:
...
+    @property
+    @abstractmethod
+    def get_bitwidth(self) -> int:
+        """
+        Returns the width of an element type in bits.
+        """
+        raise NotImplementedError()
+
+    @property
+    def get_element_width(self) -> int:
```
superlopuh, line 372:
```suggestion
    def get_exact_size_in_bytes(self) -> int | None:
```
Some ugly but explicit naming would be nice if we're moving this to shared infrastructure. Also, the client should decide how to handle bitwidths that are not evenly divisible by 8.

The cleanest option for me would be something like `MemoryLayout` in Swift: https://developer.apple.com/documentation/swift/memorylayout, which describes the alignment, size, and stride of a structure in memory, precisely for these kinds of calculations.

superlopuh: Another option would be to go directly for this:
```
    @property
    def size(self) -> int:
        """Size of the type in bytes, defined as `ceil(bitwidth / 8)`"""
        return  bitwidth // 8 + bool(bitwidth % 8)
```

xdsl/dialects/builtin.py
```diff
@@ -353,8 +353,39 @@ def print_parameter(self, printer: Printer) -> None:
...
+
+    name = "abstract.bitwidth_type"
+
+    @property
+    @abstractmethod
+    def get_bitwidth(self) -> int:
+        """
+        Returns the width of an element type in bits.
+        """
+        raise NotImplementedError()
```
superlopuh, line 369:
```suggestion
    @property
    @abstractmethod
    def bitwidth(self) -> int:
        """
        Returns the width of an element type in bits.
        """
        raise NotImplementedError()
```
If it's a property it might as well not have the get prefix

xdsl/dialects/builtin.py
superlopuh: Oh but we actually want `stride` for the calculation below!

superlopuh: My understanding of `stride` is that it's the smallest multiple of `alignment` that is greater than or equal to `size`, and the alignment is target-dependent, so the type itself can only provide the size, and the rest of the logic should be in the lowering.

jorendumoulin: I thought `alignment` would be the alignment when allocating, as in `llvm.alloca`, typically much larger than the `stride`. Or how do you see this?

jorendumoulin: I do like the `stride` attribute, as this is essentially what everybody is using `get_exact_size_in_bytes` for.

jorendumoulin: How about:
```
bitwidth: int
    Contiguous memory footprint in bits.
size: int | None
    Contiguous memory footprint in bytes. None if not applicable
stride: int | None
    Number of bytes from one element to the next when stored contiguously in memory. None if not applicable
```
If my understanding is correct, `alignment` is indeed target-dependent and not really a property of the type. Bitwidth must be defined, but size and stride can return None in the case of an `i4` type stored contiguously for example.

superlopuh: I would say we almost always want the size to be the number of bites that fits the number of bits, so it might as well not be optional. If the client cares about the divisibility they can actually do that check themselves.

And I would drop the stride, I'm not sure how correct it is without the alignment.

jorendumoulin: Okay, I kept it as mandatory, changed the name to `size` as this seems a clear enough destinction from bitwidth, and used in other places as well. I included a default implementation to the `ceil(bitwidth / 8)` function

xdsl/dialects/builtin.py
```diff
@@ -353,8 +353,31 @@ def print_parameter(self, printer: Printer) -> None:
...
+        Contiguous memory footprint in bits
+        """
+        raise NotImplementedError()
+
+    @property
+    def size(self) -> int:
+        """
+        Contiguous memory footprint in bytes, defaults to `ceil(bitwidth / 8)`
+        """
+        return ceil(self.bitwidth / 8)
```
superlopuh, line 376: This returns a float. I would do the weird int modulo thing tbh, since the bitwidth is guaranteed to be positive, and division is expensive
```suggestion
        bitwidth = self.bitwidth
        return bitwidth >> 3 + bool(bitwidth % 8)
```

superlopuh (approved):

superlopuh (approved):

PR #2916, opened 2024-07-21
transformations: don't check element type of memref_stream operands

superlopuh:
I'm no longer sure that this is a beneficial check, or what to replace it with, since the element types are checked when lowering stream read and write ops.


superlopuh:
There's some more work to do

zero9178 (approved):

nazavode (approved):

PR #2921, opened 2024-07-22
interpreter: support callable operation impls

### LINKED ISSUE #2122: interpreter: (riscv) external function call in riscv_func does not handle register dictionary in RiscvFunctions

superlopuh:
This gives the implementor of a dialect the flexibility for custom hooks of callable objects during interpretation, such as our tracking of mappings of registers to SSA values. We currently crash when calling a function with arguments as the entry point for interpretation, and this PR fixes this.

Fixes #2122


xdsl/interpreters/riscv_func.py
```diff
@@ -31,3 +32,19 @@ def run_call(
...
         results = RiscvFunctions.set_reg_values(interpreter, op.results, results)
         return results
+
+    @impl_callable(riscv_func.FuncOp)
+    def run_func(
+        self, interpreter: Interpreter, op: riscv_func.FuncOp, args: tuple[Any, ...]
+    ) -> tuple[Any, ...]:
+        if (first_block := op.body.blocks.first) is None or not first_block.ops:
+            results = interpreter.call_external(op.sym_name.data, op, args)
+        else:
```
zero9178, line 42:
```suggestion
        if (first_block := op.body.blocks.first) is None or not first_block.ops:
            return interpreter.call_external(op.sym_name.data, op, args)
        
```

zero9178 (approved):

PR #2922, opened 2024-07-22
core: Implement getAffine for MemrefLayout attributes

jorendumoulin:
This PR defines `get_affine_map` in the `MemrefLayoutAttr`, and implements this for the available layout attributes.
This is also defined for the memref type. `memreftype.get_affine_map` still returns the element offset, while `memreftype.get_affine_map_in_bytes` returns the byte offset (accounting for element type width).

Additionally, it replaces all code that (partially) implemented this functionality in their respective transforms (mostly `convert-memref-stream-to-snitch-stream`), by using the new interface in the builtin dialect.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1248,10 +1248,10 @@ class StridedLayoutAttr(MemrefLayoutAttr, ParametrizedAttribute):
         super().__init__([strides, offset])
 
     def get_strides(self) -> Sequence[int | None]:
-        return [
+        return tuple(
             None if isinstance(stride, NoneAttr) else stride.data
             for stride in self.strides
-        ]
+        )
 
     def get_offset(self) -> int | None:
         if isinstance(self.offset, NoneAttr):
@@ -1663,10 +1663,7 @@ class MemRefType(
     def get_affine_map(self) -> AffineMap:
         """
         Return the affine mapping from the iteration space of this
-        memref's layout to the byte offset in linear memory.
-
-        Unlike the get_affine_map of `MemrefLayoutAttr`, this
-        function accounts for element width.
+        memref's layout to the element offset in linear memory.
         """
         if isinstance(self.layout, NoneAttr):
             # empty shape not supported
@@ -1679,6 +1676,18 @@ class MemRefType(
             map = StridedLayoutAttr(strides).get_affine_map()
         else:
             map = self.layout.get_affine_map()
+
+        return map
+
+    def get_affine_map_bytes(self) -> AffineMap:
+        """
+        Return the affine mapping from the iteration space of this
+        memref's layout to the byte offset in linear memory.
+
+        Unlike the get_affine_map, this function accounts for element width.
+        """
+
+        map = self.get_affine_map()
 
         # account for element width
         assert isinstance(self.element_type, FixedBitwidthType)
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -69,7 +69,7 @@ from xdsl.traits import (
     OptionalSymbolOpInterface,
     SymbolTable,
 )
-from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.exceptions import DiagnosticException, VerifyException
 
 if TYPE_CHECKING:
     from xdsl.parser import AttrParser, Parser
@@ -1668,6 +1668,12 @@ class MemRefType(
         function accounts for element width.
         """
         if isinstance(self.layout, NoneAttr):
+            # empty shape not supported
+            if self.get_shape() == ():
+                raise DiagnosticException(
+                    f"Unsupported empty shape in memref of type {self}"
+                )
+
             strides = self.strides_for_shape(self.get_shape())
             map = StridedLayoutAttr(strides).get_affine_map()
         else:
@@ -1676,8 +1682,10 @@ class MemRefType(
         # account for element width
         assert isinstance(self.element_type, FixedBitwidthType)
 
-        return AffineMap(1, 0, (AffineConstantExpr(self.element_type.size),)).compose(
-            map
+        return AffineMap(
+            map.num_dims,
+            map.num_symbols,
+            tuple(result * self.element_type.size for result in map.results),
         )
 
     def get_strides(self) -> Sequence[int | None] | None:
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -32,7 +32,13 @@ from xdsl.ir import (
     TypeAttribute,
     TypedAttribute,
 )
-from xdsl.ir.affine import AffineMap, AffineSet
+from xdsl.ir.affine import (
+    AffineConstantExpr,
+    AffineDimExpr,
+    AffineSymExpr,
+    AffineMap,
+    AffineSet,
+)
 from xdsl.irdl import (
     AllOf,
     AnyAttr,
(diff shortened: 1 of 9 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1222,19 +1247,51 @@ def __init__(
...
+    def get_affine_map(self) -> AffineMap:
+        """
+        Return the affine mapping from the iteration space of this
+        layout to the element offset in linear memory. The resulting
+        affine map thus has only one result.
+
+        For dynamic strides, this results in an affinemap with a number
+        of symbols, ordered in the following manner:
+            (1) Symbols for every dynamic stride of the layout
+            (2) Symbol for the dynamic offset of the layout
```
superlopuh, line 1271: Is this MLIR's behaviour as well?

jorendumoulin: I'll check this

jorendumoulin: Oh, it seems to be the other way around: https://github.com/llvm/llvm-project/blob/56ad7cc0126f9899fd884391cfa10b6359206c01/mlir/lib/IR/BuiltinAttributes.cpp#L1790

I'll change it so it is the same

xdsl/ir/affine/affine_expr.py
```diff
@@ -180,6 +180,13 @@ def _simplify_mul(self, other: AffineExpr) -> AffineExpr | None:
         if isinstance(self, AffineBinaryOpExpr) and self.kind == AffineBinaryOpKind.Mul:
             if fold := self.rhs._try_fold_constant(other, AffineBinaryOpKind.Mul):
                 return self.lhs * fold
+        # Fold (expr + expr) * constant.
+        if (
+            isinstance(self, AffineBinaryOpExpr)
+            and self.kind == AffineBinaryOpKind.Add
+            and isinstance(other, AffineConstantExpr)
+        ):
+            return self.lhs * other + self.rhs * other
```
jorendumoulin, line 189: This little addition makes the generated affine maps much more readable:
changes
`((((d0 * 2) + d1) * 2 + d2) * 2 + d3) * 2)`
into
`(d0 * 16) + (d1 * 8) + (d2 * 4) + (d3 * 2)`

I don't know if this is legal or if this needs to be aligned with MLIR in some way?

superlopuh: Do you know if MLIR does this? I personally am not a fan of the automatic optimisation, but this one seems safe enough

jorendumoulin: > I personally am not a fan of the automatic optimisation

Isn't this ... what a compiler does?

There are some automizations in `simplify_mul` here: https://github.com/llvm/llvm-project/blob/e67f2cc3fc38cec2041cfb197ac4688ed3d16e7e/mlir/lib/IR/AffineExpr.cpp#L769, but it doesn't seem to implement this exact one. But maybe it should?

I just did this because I know running `simplify` in matlab (no judging) does exactly this, and is really nice.

xdsl/dialects/builtin.py
```diff
@@ -1222,19 +1247,52 @@ def __init__(
...
-        for stride in self.strides:
-            if isinstance(stride, NoneAttr):
-                yield None
-            else:
-                yield stride.data
+    def get_strides(self) -> Sequence[int | None]:
+        return [
+            None if isinstance(stride, NoneAttr) else stride.data
+            for stride in self.strides
+        ]
```
superlopuh, line 1254: nit:
```suggestion
        return tuple(
            None if isinstance(stride, NoneAttr) else stride.data
            for stride in self.strides
        )
```

xdsl/ir/affine/affine_expr.py
superlopuh: My stance is it would be nice for this to be explicit, and I'd be happier if there were a simplify method like in matlab :) I'm happy to add this one since the others are already there

xdsl/transforms/convert_memref_stream_to_snitch_stream.py
```diff
@@ -232,16 +172,29 @@ def strides_for_affine_map(
     """
     if affine_map.num_symbols:
         raise ValueError("Cannot create strides for affine map with symbols")
-    offset_map = strides_map_from_memref_type(memref_type)
+    offset_map = memref_type.get_affine_map()
+    if offset_map.num_symbols:
+        # only allowed for dynamic offsets, since we can ignore this here
+        # in that case, there would be one symbol, and all shapes are static
+        if not (
+            all(x != -1 for x in memref_type.get_shape())
```
superlopuh, line 180: Could you please use a constant here? Would make the intent more explicit

xdsl/transforms/convert_memref_stream_to_snitch_stream.py
```diff
@@ -232,16 +172,29 @@ def strides_for_affine_map(
     """
     if affine_map.num_symbols:
         raise ValueError("Cannot create strides for affine map with symbols")
-    offset_map = strides_map_from_memref_type(memref_type)
+    offset_map = memref_type.get_affine_map()
+    if offset_map.num_symbols:
+        # only allowed for dynamic offsets, since we can ignore this here
+        # in that case, there would be one symbol, and all shapes are static
```
superlopuh, line 178: I'm not sure exactly what this means

xdsl/ir/affine/affine_expr.py
zero9178: I think its also reasonable (if easy to do) that an affine map expressing something should only have one single canonical representation

jorendumoulin:
> > One difference between layout.get_affine_map and memreftype.get_affine_map is that the layout map returns the element offset, and the memref map returns the byte offset (accounting for element type width).
> 
> Why is this difference? I think at the memref level its way more common to work on the element type level and this may create confusion given the same name. I think it'd make more sense to name it differently, make `get_affine_map` work on the element type, and then compose the returned affine map with the coefficient for the size

Do you mean something like this?:
```
class MemrefType:
  get_affine_map():
      # return element offset
  get_affine_map_bytes():
       # return byte offset
```

xdsl/dialects/builtin.py
jorendumoulin: ok! but why is this better?

xdsl/transforms/convert_memref_stream_to_snitch_stream.py
jorendumoulin: I tried changing the check for clarity. We don't want any symbols for dynamic strides, as this is not supported yet. However, a dynamic offset is supported, as the offset can be ignored in this function. I just changed the check to make sure all shapes are statically known. If there are any symbols left in the affine map, they are related to the offset, and can just be set to 0.

zero9178 (approved):

xdsl/dialects/builtin.py
```diff
@@ -1602,7 +1660,45 @@ def print_parameters(self, printer: Printer) -> None:
...
+                )
+
+            strides = self.strides_for_shape(self.get_shape())
+            map = StridedLayoutAttr(strides).get_affine_map()
+        else:
+            map = self.layout.get_affine_map()
+
+        return map
+
+    def get_affine_map_bytes(self) -> AffineMap:
```
superlopuh, line 1682:
```suggestion
    def get_affine_map_in_bytes(self) -> AffineMap:
```
?

superlopuh (approved):

PR #2924, opened 2024-07-22
backend: (riscv) add allocate_same method to register allocator

superlopuh:
This is useful as a general factoring out of our loop allocation logic, but is also necessary for future operations that have constraints for results and operands that must be the same register.

### DIFF
--- a/xdsl/backend/riscv/register_allocation.py
+++ b/xdsl/backend/riscv/register_allocation.py
@@ -1,4 +1,5 @@
 import abc
+from collections.abc import Sequence
 from itertools import chain
 from typing import cast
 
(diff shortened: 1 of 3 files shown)


xdsl/backend/riscv/register_allocation.py
```diff
@@ -116,6 +120,54 @@ def allocate(self, reg: SSAValue) -> bool:
 
         return False
 
+    def allocate_same(self, vals: Sequence[SSAValue]) -> bool:
```
zero9178, line 123: Could you leave a doc comment here explaining more in detail what it does?

My initial thought of the name was that it would make sure that it'd allocate the same register for all `vals` but it seems to be smarter than that and if one consistent register exists that has been allocated then it seemingly just uses that. 

I think this is worth documenting as this can cause incorrectness issues if misused.

superlopuh: Yep good idea

nazavode (approved):

PR #2927, opened 2024-07-22
dialects: (snitch_stream) add repetition to stride pattern

superlopuh:
Without the repeat setting when the last offset is 0, Snitch just fetches every time without caching the value, repeat avoids this.

CC @jorendumoulin @JosseVanDelm seems like more and more accelerator config! One of the things that's somewhat annoying is that we now have to defensively reset the repetition count, it would be great to avoid this eventually.


JosseVanDelm:
@superlopuh this is exactly what `accfg` is useful for indeed!!! @AntonLydike also recently added a `accfg.reset` op that should be useful to reset required at the end of MLIRs scope.

nazavode (approved):
Beautiful.

PR #2928, opened 2024-07-23
dialects: (memref_stream) add doc and library call support

jorendumoulin:
We are using the `library_call` attribute to designate to which accelerator we want to dispatch a certain linalg generic to. Right now, this information gets lost when converting from linalg.generic to memref_stream.generic, which is somewhat annoying. This PR adds support for these generic attributes in the memref stream dialect.

### DIFF
--- a/xdsl/dialects/memref_stream.py
+++ b/xdsl/dialects/memref_stream.py
@@ -43,6 +43,7 @@ from xdsl.irdl import (
     irdl_attr_definition,
     irdl_op_definition,
     operand_def,
+    opt_prop_def,
     prop_def,
     region_def,
     var_operand_def,
@@ -397,6 +398,9 @@ class GenericOp(IRDLOperation):
     Indices into the `outputs` that correspond to the initial values in `inits`.
     """
 
+    doc: StringAttr | None = opt_prop_def(StringAttr)
+    library_call: StringAttr | None = opt_prop_def(StringAttr)
+
     body: Region = region_def("single_block")
 
     traits = frozenset((GenericOpHasCanonicalizationPatternsTrait(),))
@@ -413,6 +417,8 @@ class GenericOp(IRDLOperation):
         iterator_types: ArrayAttr[Attribute],
         bounds: ArrayAttr[IntegerAttr[IndexType]],
         init_indices: ArrayAttr[IntAttr],
+        doc: StringAttr | None = None,
+        library_call: StringAttr | None = None,
     ) -> None:
         for m in indexing_maps:
             if m.data.num_symbols:
@@ -426,6 +432,8 @@ class GenericOp(IRDLOperation):
                 "init_indices": init_indices,
                 "indexing_maps": indexing_maps,
                 "iterator_types": iterator_types,
+                "doc": doc,
+                "library_call": library_call,
             },
             regions=[body],
         )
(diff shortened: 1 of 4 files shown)


xdsl/dialects/memref_stream.py
```diff
@@ -397,6 +398,9 @@ class GenericOp(IRDLOperation):
     Indices into the `outputs` that correspond to the initial values in `inits`.
     """
 
+    doc: StringAttr | None = opt_prop_def(StringAttr)
+    library_call: StringAttr | None = opt_prop_def(StringAttr)
```
superlopuh, line 402: I would lean towards this:
```suggestion
    doc = prop_def(StringAttr)
    library_call = prop_def(StringAttr)
```
How is an empty doc and empty library call functionally different from None?

xdsl/dialects/memref_stream.py
```diff
@@ -513,6 +521,12 @@ def print(self, printer: Printer):
                 lambda iterator_type: printer.print_string_literal(iterator_type.data),
             )
             printer.print_string("]")
+            if self.doc:
```
superlopuh, line 524: We generally prefer `is None` to differentiate between empty cases and `None`, here my recommendation would also be to rather check for this:
```suggestion
            if self.doc.data:
```
And for the default value for the argument to be `StringAttr("")`

xdsl/dialects/memref_stream.py
jorendumoulin: Then it's no longer optional and all current tests fail with `Operation does not verify: property doc expected`. I think we want to keep them optional as in mlir? I just copied this from the `linalg.generic` implementation

xdsl/dialects/memref_stream.py
jorendumoulin: Having a default value of `StringAttr("")` still prints the attribute:
```
  memref_stream.generic {
    bounds = [4, 2, 3],
    indexing_maps = [
      affine_map<(d0, d1, d2) -> (d0, d2)>,
      affine_map<(d0, d1, d2) -> (d2, d1)>,
      affine_map<(d0, d1) -> (d0, d1)>,
      affine_map<(d0, d1) -> (d0, d1)>
    ],
    iterator_types = ["parallel", "parallel", "reduction"],
    library_call = ""
  } ins(%E, %F : memref<4x2xf64>, memref<2x3xf64>) outs(%G, %I : memref<4x3xf64>, memref<4x3xf64>) inits(%H : f64, None) {
  ^4(%e_1 : f64, %f_1 : f64, %acc_old_1 : f64, %acc_old_2 : f64):
    %prod_1 = arith.mulf %e_1, %f_1 : f64
    %acc_new_1 = arith.addf %acc_old_1, %prod_1 : f64
    %acc_new_2 = arith.addf %acc_old_2, %prod_1 : f64
    linalg.yield %acc_new_1, %acc_new_2 : f64, f64
  }
  ```
As it is an optional attribute, I don't think this is desirable

superlopuh: But that's something you added in this PR, you could just not print it if it's the empty string, similarly make the parsing step optional with the same default value

jorendumoulin: The same holds for `--print-op-generic`, where it can't be changed.

superlopuh: Ah yes, I wonder what MLIR does there

superlopuh: OK it does the same thing, then your approach is best

xdsl/dialects/memref_stream.py
superlopuh: You're right, this is good

superlopuh (approved):

PR #2929, opened 2024-07-23
transformations: (memref_stream) add memref_stream.generic legalization pass

nazavode:
This PR adds a legalization pass for `memref_stream.generic` operations. The process involves making the payload body legal wrt types that can be streamed by the target.

### DIFF
--- a/tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
+++ b/tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt -p convert-linalg-to-memref-stream,memref-stream-legalize,test-optimise-memref-stream,test-lower-memref-stream-to-snitch-stream,test-lower-snitch-stream-to-asm -t riscv-asm %s | filecheck %s
+// RUN: xdsl-opt -p convert-linalg-to-memref-stream,test-optimise-memref-stream,test-lower-memref-stream-to-snitch-stream,test-lower-snitch-stream-to-asm -t riscv-asm %s | filecheck %s
 
 
 func.func public @ssum(
--- a/tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
+++ b/tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt -p convert-linalg-to-memref-stream,test-optimise-memref-stream,test-lower-memref-stream-to-snitch-stream,test-lower-snitch-stream-to-asm -t riscv-asm %s | filecheck %s
+// RUN: xdsl-opt -p convert-linalg-to-memref-stream,memref-stream-legalize,test-optimise-memref-stream,test-lower-memref-stream-to-snitch-stream,test-lower-snitch-stream-to-asm -t riscv-asm %s | filecheck %s
 
 
 func.func public @ssum(
@@ -15,9 +15,9 @@ func.func public @ssum(
     ],
     iterator_types = ["parallel", "parallel"]
   } ins(%X, %Y : memref<8x16xf32>, memref<8x16xf32>) outs(%Z : memref<8x16xf32>) {
-  ^1(%in : vector<2xf32>, %in_1 : vector<2xf32>, %out : vector<2xf32>):
-    %3 = arith.addf %in, %in_1 : vector<2xf32>
-    memref_stream.yield %3 : vector<2xf32>
+  ^1(%in : f32, %in_1 : f32, %out : f32):
+    %3 = arith.addf %in, %in_1 : f32
+    memref_stream.yield %3 : f32
   }
   func.return
 }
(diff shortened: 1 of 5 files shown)


xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,150 @@
...
+@dataclass(frozen=True)
+class MemrefStreamGenericLegalize(RewritePattern):
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: memref_stream.GenericOp, rewriter: PatternRewriter
+    ) -> None:
+        # FIXME add reduction support
+        if op.iterator_types.data[-1].data != IteratorType.PARALLEL:
+            return
```
superlopuh, line 77: I would say we should actually raise a DiagnosticError and crash early right now, since we're planning to add it soon

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,150 @@
...
+        # Collect block arguments that need to be legalized
+        legalizations: dict[int, Attribute] = {
+            i: _legalize_attr(arg.type)
+            for i, arg in enumerate(op.body.block.args)
+            if not _is_legal_attr(arg.type)
+        }
+        if len(legalizations) == 0:
```
superlopuh, line 84:
```suggestion
        if not legalizations):
```

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,163 @@
...
+                return True
+            case Float16Type(), 4:
+                return True
+            case _:
+                return False
+    return False
+
+
+def _is_legal_attr(attr: Attribute) -> bool:
+    return _is_legal_vector(attr) or isa(attr, StreamingLegalType)
```
superlopuh, line 49: It seems like these helpers should be on the memref_stream.generic type itself, or maybe in the dialect definition file?

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,163 @@
...
+class MemrefStreamGenericLegalize(RewritePattern):
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: memref_stream.GenericOp, rewriter: PatternRewriter
+    ) -> None:
+        if op.iterator_types.data[-1].data != IteratorType.PARALLEL:
+            raise DiagnosticException(
+                "iterators other than 'parallel' are not supported yet"
+            )
```
superlopuh, line 97: Actually could you please move this below the "should legalize" check?

xdsl/transforms/memref_stream_legalize.py
superlopuh: Actually no they're snitch constraints not memref_stream constraints

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,163 @@
...
+        # Collect block arguments that need to be legalized
+        legalizations: dict[int, Attribute] = {
```
superlopuh, line 99: It feels a little off to me to cast below. Maybe we're missing another helper that returns an optional `StreamingVectorType` or None of there's no legalization to do?
```suggestion
        legalizations: dict[int, StreamingVectorType] = {
```

tests/filecheck/transforms/memref_stream_legalize.mlir
superlopuh, line 1: Could you please add an example with a reduction on f64 to make sure we don't error out when no legalization is needed?

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,174 @@
...
+        return VectorType(f16, (4,))
+    else:
+        raise DiagnosticException(f"Cannot legalize {attr} for streaming")
+
+
+def _legalize_block(block: Block, rewriter: PatternRewriter) -> None:
+    # Start from all uses of all block arguments
+    to_be_legalized: set[Operation] = {
+        use.operation for arg in block.args for use in arg.uses
+    }
```
superlopuh, line 76: shouldn't these be the uses of the legalized arguments, as opposed to all of them?

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,174 @@
...
+        if not illegal_results:
+            continue
+        new_op = op.create(
+            operands=op.operands,
+            result_types=[_legalize_attr(res.type) for res in op.results],
+            attributes=op.attributes,
+        )
+        rewriter.replace_op(op, new_op)
+        for idx in illegal_results:
+            to_be_legalized.update(use.operation for use in new_op.results[idx].uses)
```
superlopuh, line 95:
```suggestion
        to_be_legalized.update(
            use.operation 
            for idx in illegal_results
            for use in new_op.results[idx].uses
        )
```

xdsl/transforms/memref_stream_legalize.py
```diff
@@ -0,0 +1,174 @@
...
+        illegal_results: list[int] = [
+            result.index for result in op.results if not _is_legal_attr(result.type)
+        ]
+        if not illegal_results:
+            continue
+        new_op = op.create(
+            operands=op.operands,
+            result_types=[_legalize_attr(res.type) for res in op.results],
+            attributes=op.attributes,
+        )
```
superlopuh, line 92: It feels like we should check for a set of operations we know this is safe to do, like `FloatingPointLikeBinaryOp` in arith.

nazavode: It looks both `isa` and `isinstance` break with the nice `FloatingPointLikeBinaryOp`, resorted to enumeration of operations known to be safe to stream-legalize.

tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
superlopuh, line 1: does it work with linalg.generic? Seems like it should, it would be cool to go all the way

superlopuh (approved):

tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
nazavode: It works like a charm.

tests/filecheck/projects/riscv-backend-paper/bottom_up.mlir
```diff
@@ -1,23 +1,22 @@
-// RUN: xdsl-opt -p convert-linalg-to-memref-stream,test-optimise-memref-stream,test-lower-memref-stream-to-snitch-stream,test-lower-snitch-stream-to-asm -t riscv-asm %s | filecheck %s
+// RUN: xdsl-opt -p convert-linalg-to-memref-stream,memref-stream-legalize,test-optimise-memref-stream,test-lower-memref-stream-to-snitch-stream,test-lower-snitch-stream-to-asm -t riscv-asm %s | filecheck %s
```
superlopuh, line 1: Others might disagree, but I view this as part of the scheduling, and not really a separate thing, so I would recommend adding it to `test-optimise-memref-stream` instead of a standalone test in the pipeline

nazavode: Agreed!

superlopuh (approved):

PR #2939, opened 2024-07-25
misc: use constraints in more operation definitions

superlopuh:
Pylance errors: 216 -> 151

### DIFF
--- a/xdsl/dialects/csl/csl_wrapper.py
+++ b/xdsl/dialects/csl/csl_wrapper.py
@@ -152,8 +152,8 @@ class ModuleOp(IRDLOperation):
 
     name = "csl_wrapper.module"
 
-    width = prop_def(IntegerAttr)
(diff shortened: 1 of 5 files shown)


math-fehr (approved):
O u I

xdsl/dialects/csl/csl_wrapper.py
```diff
@@ -152,8 +152,8 @@ class ModuleOp(IRDLOperation):
 
     name = "csl_wrapper.module"
 
-    width = prop_def(IntegerAttr)
-    height = prop_def(IntegerAttr)
+    width = prop_def(AnyIntegerAttr)
+    height = prop_def(AnyIntegerAttr)
```
n-io, line 156: Why do we want to allow index types here?

n-io: Technically speaking, it probably already was allowed. Will fix.

PR #2940, opened 2024-07-25
api: remove parameters field from Trait

superlopuh:
These make the API more complicated than it needs to be, since all traits are frozen dataclasses we can let the dataclass annotation do the heavy lifting for comparison and hashability.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -939,8 +939,8 @@ class Operation(IRNode):
     @classmethod
     def has_trait(
         cls,
-        trait: type[OpTrait],
-        parameters: Any = None,
+        trait: type[OpTrait] | OpTrait,
+        *,
         value_if_unregistered: bool = True,
     ) -> bool:
         """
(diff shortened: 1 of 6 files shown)


xdsl/ir/core.py
```diff
@@ -953,18 +953,21 @@ def has_trait(
...
             return value_if_unregistered
 
-        return cls.get_trait(trait, parameters) is not None
+        return cls.get_trait(trait) is not None
 
     @classmethod
-    def get_trait(
-        cls, trait: type[OpTraitInvT], parameters: Any = None
-    ) -> OpTraitInvT | None:
+    def get_trait(cls, trait: type[OpTraitInvT] | OpTraitInvT) -> OpTraitInvT | None:
```
superlopuh, line 959: This is a breaking change, but the fix is just a couple of backspaces to delete the comma and instantiate the class instead of passing the parameters separately

webmiche (approved):
LGTM in general.

I am not sure we want to remove parameters from traits. It feels to me like parameters are something that traits should have. Or at least, that the concept of a parametrized trait deserves its own class.
So, to me, this is fine in a technical sense, but it feels to me like what I understand as a trait and what the class implements is not quite the same anymore.

That's not a blocking comment though, I am fine with merging this.

n-io (approved):
Merged main to resolve a conflict in `csl_stencil`, check if you're happy, lgmt.

superlopuh:
Thank you! To me the thing that makes this still have the property of "traits with parameters" is that frozen dataclasses have fields that you can inspect and reason about, so you still have this notion of a named introspectable structure, but without more infrastructure on our end.

xdsl/traits.py
```diff
@@ -47,22 +43,20 @@ class ConstantLike(OpTrait):
...
     """Constraint the operation to have a specific parent operation."""
 
-    parameters: tuple[type[Operation], ...]
+    op_types: tuple[type[Operation], ...]
 
-    def __init__(self, *parameters: type[Operation]):
-        if not parameters:
-            raise ValueError("parameters must not be empty")
-        super().__init__(parameters)
+    def __init__(self, head_param: type[Operation], *tail_params: type[Operation]):
```
math-fehr, line 48: Why not super() init?

xdsl/traits.py
```diff
@@ -73,18 +67,18 @@ class HasAncestor(OpTrait):
     parent.
     """
 
-    parameters: tuple[type[Operation], ...]
+    op_types: tuple[type[Operation], ...]
 
     def __init__(self, head_param: type[Operation], *tail_params: type[Operation]):
-        super().__init__((head_param, *tail_params))
+        object.__setattr__(self, "op_types", (head_param, *tail_params))
```
math-fehr, line 73: Same here

xdsl/traits.py
superlopuh: There is no super init here that would set the op types property, as far as I understand

PR #2942, opened 2024-07-25
core: (affine) override python floordiv symbol

jorendumoulin:
we can override the floordiv symbol, which makes for much nicer experience when using `AffineMap.from_callable`, among others.


superlopuh (approved):
Neat!

PR #2946, opened 2024-07-26
misc: move IRDLDialectLoader to own file and remove circular dependency

superlopuh:
It seems unfortunate to me to import the builtin dialect when importing from xdsl.ir. Since this is a dialects thing, it feels like the best place to put the import override hook is in the `xdsl.dialects` `__init__.py`, and localize the dynamism a little bit.

My aim for this was for the only functional change to be to make the import happen later, as far as I can tell it does exactly the same thing, please let me know if I've missed something.


webmiche (approved):
Sounds reasonable to me. I like not importing more than is needed.

PR #2947, opened 2024-07-26
(dialects): improving stencil attr syntax mlir compatibility

n-io:
MLIR interoperability needs the (generic) format changed from `#stencil.index[0]` to `#stencil.index<[0]>`, and likewise for `#stencil.bounds` where it also doesn't like the colon.

A practical implication of this is that the syntax of `stencil.store` has changed, for instance, from
`stencil.store %5 to %3 ([0, 0, 0] : [64, 64, 64])` to
`stencil.store %5 to %3 (<[0, 0, 0], [64, 64, 64]>)`

This PR does not modify the syntax of any ops, and only modified the printing and parsing of the index and bounds attributes. In a future PR, the syntax of the store op (maybe also others) could be reviewed.


PapyChacal (approved):
Thanks!!

PR #2950, opened 2024-07-27
misc: make version lazy

superlopuh:
Shaves off about 40ms when launching xdsl-opt, or 1/5 of the time.

`make filecheck` goes from 18.26s to 13.10s on my computer with this change.

### DIFF
--- a/xdsl/__init__.py
+++ b/xdsl/__init__.py
@@ -1,3 +1,7 @@
+import importlib.abc
+import importlib.util
+import os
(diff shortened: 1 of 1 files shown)


zero9178 (approved):

webmiche (approved):
LGTM! I honestly don't get what's happening 100%, but I am surprised that this would not be an optimization that is done somewhere internally. Still, good catch!

PR #2951, opened 2024-07-28
api: deprecate `print_string_raw` in favor of indent=0

superlopuh:
This is an effort to minimise the API surface a little bit, and to make it a little more efficient. When there are no newlines, it's wasteful to create a list of them, so the first step when printing a string is now to check whether they're there. Raw printing is equivalent to printing with an indentation of 0, modulo next line callbacks. I don't think it's likely to end up in a situation where the difference is visible, so where the user prints a raw string and expects the next line callback to not be executed, and instead be saved for the next line that is printed with the non-raw printing, please let me know if I'm missing something. In all other cases the behaviour is the same as before.


xdsl/printer.py
```diff
@@ -261,12 +270,18 @@ def _print_new_line(
         self, indent: int | None = None, print_message: bool = True
     ) -> None:
         indent = self._indent if indent is None else indent
-        self.print_string_raw("\n")
```
superlopuh, line 264: This change avoids the slightly weird circular calling of the functions

Moxinilian (approved):
Yeah that's better.

AntonLydike (approved):

xdsl/printer.py
```diff
@@ -142,44 +144,51 @@ def print(self, *argv: Any) -> None:
...
         This function takes into account indentation level when
         printing new lines.
-
-        Indentation level for newlines within the printed string
-        can be overriden by specifying the `indent` argument.
+        If the indentation level is specified as 0, the string is printed as-is, if `None`
+        then the `Printer` instance's indentation level is used.
         """
 
+        num_newlines = text.count("\n")
```
math-fehr, line 165: Is this more or less the same speed as a precomputed regex?

xdsl/printer.py
```diff
@@ -261,12 +270,15 @@ def _print_new_line(
         self, indent: int | None = None, print_message: bool = True
     ) -> None:
         indent = self._indent if indent is None else indent
-        self.print_string_raw("\n")
+        print(file=self.stream)
```
math-fehr, line 273: Can you add a comment to make it obvious?

xdsl/printer.py
```diff
@@ -261,12 +270,15 @@ def _print_new_line(
...
+        self._current_line += 1
         if print_message:
             for callback in self._next_line_callback:
                 callback()
             self._next_line_callback = []
-        self.print_string_raw(" " * indent * indentNumSpaces)
+        num_spaces = indent * indentNumSpaces
+        print(" " * num_spaces, end="", file=self.stream)
```
math-fehr, line 280: Same here

math-fehr (approved):
🦄

xdsl/printer.py
superlopuh: I'm not sure, but I'll probably fiddle with this only if I see it coming up in the traces

xdsl/printer.py
superlopuh: To make what obvious?

superlopuh: I added a couple of comments, not sure if they're what you expected

xdsl/printer.py
AntonLydike: It doesn't really make a difference:

```py
# a.py
import random
import time
random.seed(9999)
# text with a few instances of `#b` in there
alphabet = list("abcdefghijklmopqrstuvwxyzABCDEFGHIJKLMOPQRSTUVWXYZ0123456789 ") + ['#b']
text = "".join(
    random.choice(alphabet) for _ in range(500)
)
```

**`str.count`**:

```bash
$ python3 -m timeit -s "from a import text, re" "text.count('#b')"
10000 loops, best of 5: 34 usec per loop
```

**`re.findall` (with compilation)**

```bash
$ python3 -m timeit -s "from a import text, re" "len(re.findall('#b', text))"
10000 loops, best of 5: 33.8 usec per loop
```

**`re.findall` (precompiled)**
```bash
$ python3 -m timeit -s "from a import text, re; expr = re.compile('#b')" "len(expr.findall(text))"
10000 loops, best of 5: 33.7 usec per loop
```

These are all well within error of each other.

Smaller strings seem to even favour `str.count`, probably because the regex machinery is actaully somewhat heavy...

PR #2961, opened 2024-07-30
backend: (riscv) fix register allocation stats

superlopuh:
We were overcounting the registers, since there was no uniquing going on.


xdsl/backend/riscv/register_allocation.py
```diff
@@ -70,12 +70,16 @@ def count_reg_types(regs: Iterable[Attribute]) -> tuple[int, int]:
     """
     Returns a tuple containing the count of IntRegister and FloatRegister in the iterable.
     """
-    num_complex = sum(
```
superlopuh, line 73: too clever by half :)

zero9178 (approved):

PR #2963, opened 2024-07-30
transformations: lift arith to linalg named ops

n-io:
A small pass to lift arith ops to linalg *named* ops.

Currently supported: `addf`, `mulf`, `subf`


tests/filecheck/transforms/lift-arith-to-linalg.mlir
superlopuh, line 1: Could you please remove the stencil bit from this test? Ideally, it would be great if it contained only the arith ops that are handled by the transform, and some `"test.op"` to create the inputs

xdsl/transforms/lift_arith_to_linalg.py
```diff
@@ -0,0 +1,99 @@
...
+            curr_op = curr_op.next_op
+
+        # insert a new tensor.empty if none can be found
+        rewriter.insert_op(
+            new_op := tensor.EmptyOp((), tensor_shape),
+            InsertPoint.at_start(p_region.first_block),
+        )
+        return new_op.tensor
+
+    assert False, "Op must be inside a region with a block"
```
superlopuh, line 50: Why not insert a tensor.alloc just before the linalg operation? It seems like this would most closely match the behaviour of the arith op, which implicitly allocs a tensor as part of its operation.

n-io: The answer has to do with bufferization, as bufferization will create a new `memref.alloc` for each `tensor.empty`. However, having looked into this again, it's probably much easier to simply pass the lhs into outs. This is valid because the semantics of for instance `linalg.add` is `out[0] = ins[0] + ins[1]` and *not* `out[0] += ins[0] + ins[1]`, as per the docs and the hidden region code.

n-io: Specifically, as per the docs, in tensor mode to the `outs` is only referred to for its shape.

tests/filecheck/transforms/lift-arith-to-linalg.mlir
n-io: Done as suggested.

superlopuh (approved):
Beautiful, thank you

PR #2964, opened 2024-07-30
misc: use specific print method in printer.py

superlopuh:
`make filecheck` times down from 13.51s to 12.74s, or 6% faster.


math-fehr (approved):
#interpreted-language

PR #2966, opened 2024-07-31
testing: (linalg-snitch) change order of incrementally added passes

superlopuh:


tests/transforms/test_test_lower_linalg_to_snitch.py
jorendumoulin, line 1: this is neat. Is this meant to replicate `-O3` behaviour of gcc?

jorendumoulin (approved):

tests/transforms/test_test_lower_linalg_to_snitch.py
superlopuh: Not really, it's just something for the paper, to show how the performance gradually improves with certain optimisations. I still have to write the bit of the evaluation about this, hopefully it'll be clear what's going on, my previous attempt in the pres was just confusing because it seemed like one pass was doing all the heavy lifting, but actually they're all necessary.

PR #2967, opened 2024-07-31
misc: don't import riscv and x86 in xdsl-opt

superlopuh:
Filecheck tests go from 12.68s to 11.41s.


zero9178 (approved):

webmiche (approved):
Software engineers hate this Python trick

PR #2968, opened 2024-07-31
dialects: (tensor) add tensor.dim and tensor.cast op

jorendumoulin:
This PR adds the `tensor.dim` and `tensor.cast` operation

### DIFF
--- a/xdsl/dialects/tensor.py
+++ b/xdsl/dialects/tensor.py
@@ -33,6 +33,41 @@ from xdsl.utils.exceptions import VerifyException
 
 
 @irdl_op_definition
+class DimOp(IRDLOperation):
+    """
+    The tensor.dim operation takes a tensor and a dimension operand of type index.
+    It returns the size of the requested dimension of the given tensor.
+    If the dimension index is out of bounds, the behavior is undefined
+    """
+
+    name = "tensor.dim"
+
+    source = operand_def(TensorType[Attribute])
+    index = operand_def(IndexType)
+    result = result_def(IndexType)
+
+    def __init__(self, source: SSAValue | Operation, index: SSAValue | Operation):
+        super().__init__(operands=(source, index), result_types=(IndexType(),))
+
+    def print(self, printer: Printer):
+        printer.print_string(" ")
+        printer.print_ssa_value(self.source)
+        printer.print_string(", ")
+        printer.print_ssa_value(self.index)
+        printer.print_string(" : ")
+        printer.print_attribute(self.result.type)
+
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        source = parser.parse_operand()
+        parser.parse_punctuation(",")
+        index = parser.parse_operand()
+        parser.parse_punctuation(":")
+        parser.parse_type()
+        return cls(source, index)
+
+
+@irdl_op_definition
 class EmptyOp(IRDLOperation):
     name = "tensor.empty"
 
@@ -306,6 +341,7 @@ class InsertSliceOp(IRDLOperation):
 Tensor = Dialect(
     "tensor",
     [
+        DimOp,
         EmptyOp,
         ExtractSliceOp,
         InsertSliceOp,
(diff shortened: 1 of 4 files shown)


superlopuh (commented):
Could you please add MLIR interop tests?

xdsl/dialects/tensor.py
```diff
@@ -32,6 +32,41 @@
...
+        printer.print_attribute(self.source.type)
+
+    @classmethod
+    def parse(cls, parser: Parser) -> Self:
+        source = parser.parse_operand()
+        parser.parse_punctuation(",")
+        index = parser.parse_operand()
+        parser.parse_punctuation(":")
+        parser.parse_type()
+        return cls(source, index)
```
superlopuh, line 67: This seems like it could work with a custom assembly format. Could you please have a go? It might simplify things a little bit. Importantly, the attribute dictionary is missing.
```
operation ::= `tensor.dim` attr-dict $source `,` $index `:` type($source)
```
Hopefully something pretty close to the mlir version should work, if not this exactly

jorendumoulin: I only discovered this possibility yesterday but I really like it and definitely want to give it a go
It's a bit tricky here because the result and index type do not appear in the custom format (index, index).

There should be some sort of exception when handling indextypes in the parser:

In `xdsl/irdl/constraints.py`:
```python
class BaseAttr():

    attr: type[AttributeCovT]
    """The expected attribute base type"""

    # overload can_infer
    def can_infer(self, constraint_names: set[str]) -> bool:
        # check here if it is indextype
        if is_index_type:
            return True
        return super().can_infer(constraint_names)
```

However, importing from builtin dialect in irdl classes seems wrong, and additionaly also results in circular imports.

Is the following a viable strategy?

- create traits for attributes
- create `ShouldNotInferType` trait 
- if the constraining type has this attribute, there is no need for `can_infer`, and we can just return true

jorendumoulin: However, if this is overkill, the following dirty fix does the trick:
```python

    def can_infer(self, constraint_names: set[str]) -> bool:
        if self.attr.name == 'index':
            return True
        return super().can_infer(constraint_names)
```

superlopuh: Ah I see. I think MLIR is smarter here about inferring types in the assembly format, and we could also do better. A part of the problem is that our parser is missing a method of parsing an attribute given its type, which MLIR has. Let's drop my suggestion of the assembly format for now, although maybe it's worth filing an issue. I would mostly recommend adding some support for the attribute dictionary, and adding the MLIR interop tests.

superlopuh: This is all with the idea of not derailing this PR and your work too much, but if you'd like to hack on the parsing infrastructure and the assembly format, that would be very welcome!

superlopuh: I would expect that some traits/abstract superclasses would be the thing to use here. We're currently blocked on `memref.global` for a similar reason, the type of the memref is in a different place to the content, and we can't handle that currently.

jorendumoulin:
I've fixed the missing attribute and missing mlir interop tests! I kept with custom parsing and printing, but I'll file an issue about this problem.
If it's alright also sneaked in the `tensor.cast` op here, which works nicely with the assembly format.

xdsl/dialects/tensor.py
```diff
@@ -32,6 +32,62 @@
 from xdsl.utils.exceptions import VerifyException
 
 
+@irdl_op_definition
+class CastOp(IRDLOperation):
+    """
+    Convert a tensor from one type to an equivalent type without changing any data elements.
+    The source and destination types must both be tensor types with the same element type.
+    If both are ranked, then the rank should be the same and static dimensions should match.
+    The operation is invalid if converting to a mismatching constant dimension.
```
jorendumoulin, line 41: I'm quickly going to add a verfier for this

superlopuh (approved):

superlopuh (approved):

PR #2969, opened 2024-07-31
dialects: (arith, core) Constant op to accept tensor and memref

n-io:
Upstream mlir accepts the following arith.constant:
```
%t_const = arith.constant dense<1.234500e-01> : tensor<16xf32>
%m_const = arith.constant dense<1.678900e-01> : memref<64xf32>
```

As for tensors, xdsl currently accepts it in generic form and print it in both generic and non-generic form, but will fail to parse the non-generic form.

As for memrefs, xdsl will not accept it and requires an extensive change to the attribute parser to add support for memrefs in `RankedVectorOrTensorOf`. Some code had to be moved after the definition of `MemRefType` to support this change.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1587,7 +1587,7 @@ class UnrankedMemrefType(
 
 AnyUnrankedMemrefType: TypeAlias = UnrankedMemrefType[Attribute]
 
-RankedVectorOrTensorOf: TypeAlias = (
+RankedStructure: TypeAlias = (
     VectorType[AttributeCovT] | TensorType[AttributeCovT] | MemRefType[AttributeCovT]
 )
 
@@ -1598,9 +1598,9 @@ class DenseIntOrFPElementsAttr(
 ):
     name = "dense"
     type: ParameterDef[
-        RankedVectorOrTensorOf[IntegerType]
-        | RankedVectorOrTensorOf[IndexType]
-        | RankedVectorOrTensorOf[AnyFloat]
+        RankedStructure[IntegerType]
+        | RankedStructure[IndexType]
+        | RankedStructure[AnyFloat]
     ]
     data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
 
@@ -1631,7 +1631,7 @@ class DenseIntOrFPElementsAttr(
 
     @staticmethod
     def create_dense_index(
-        type: RankedVectorOrTensorOf[IndexType],
+        type: RankedStructure[IndexType],
         data: Sequence[int] | Sequence[IntegerAttr[IndexType]],
     ) -> DenseIntOrFPElementsAttr:
         if len(data) and isinstance(data[0], int):
@@ -1645,7 +1645,7 @@ class DenseIntOrFPElementsAttr(
 
     @staticmethod
     def create_dense_int(
-        type: RankedVectorOrTensorOf[IntegerType],
+        type: RankedStructure[IntegerType],
         data: Sequence[int] | Sequence[IntegerAttr[IntegerType]],
     ) -> DenseIntOrFPElementsAttr:
         if len(data) and isinstance(data[0], int):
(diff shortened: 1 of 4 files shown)


xdsl/dialects/builtin.py
```diff
@@ -896,147 +892,6 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
         constraint.verify(attr, constraint_context)
 
 
-@irdl_attr_definition
```
superlopuh, line 899: Why move this down? What's the change here? Git isn't being very helpful

n-io: As briefly mentioned in the PR description, though likely not explained in sufficient detail, I had to move `RankedVectorOrTensorOf` after `MemRefType`, as I changed it to now also accept `MemRefType`. As a follow-up, `DenseIntOrFPElementsAttr` also had to be moved down since it uses `RankedVectorOrTensorOf`.

As far as I remember, the only change required to `DenseIntOrFPElementsAttr` was to change the signature of the `type` argument in `from_list` from `RankedVectorOrTensorOf[AnyFloat | IntegerType | IndexType]` to:
```
RankedVectorOrTensorOf[AnyFloat | IntegerType | IndexType]
            | RankedVectorOrTensorOf[AnyFloat]
            | RankedVectorOrTensorOf[IntegerType]
            | RankedVectorOrTensorOf[IndexType]
```

Not fully clear why, but it does make pyright happy.

BTW, accepting new name suggestions for `RankedVectorOrTensorOf`.

superlopuh: Makes a lot of sense, thank you

xdsl/dialects/builtin.py
```diff
@@ -1732,6 +1587,167 @@ def from_type(
 
 AnyUnrankedMemrefType: TypeAlias = UnrankedMemrefType[Attribute]
 
+RankedVectorOrTensorOf: TypeAlias = (
+    VectorType[AttributeCovT] | TensorType[AttributeCovT] | MemRefType[AttributeCovT]
+)
```
superlopuh, line 1592: Maybe this?
```suggestion
RankedStructureOf: TypeAlias = (
    VectorType[AttributeCovT] | TensorType[AttributeCovT] | MemRefType[AttributeCovT]
)
```

superlopuh (commented):
Code looks good to me, I would probably wait for someone more familiar with MLIR than I am to comment on how similar this is to how things happen over there. My understanding is that some of the names follow the MLIR API.

Mabe @math-fehr  or @zero9178 can advise?

n-io:
> Code looks good to me, I would probably wait for someone more familiar with MLIR than I am to comment on how similar this is to how things happen over there. My understanding is that some of the names follow the MLIR API.
> 
> Mabe @math-fehr or @zero9178 can advise?

I'd appreciate that, not super familiar with that part of the API myself, but it appears `RankedVectorOrTensorOf` was an xdsl concept, where MLIR *if I'm not mistaken* has no similar concept and would more generically use `Attribute`

AntonLydike:
I don't think `memref` makes all that much sense to be accepted here? Seeing as arith does not work on `memref` anyway? (big debate on this happening rn)...

PapyChacal:
> I don't think `memref` makes all that much sense to be accepted here? Seeing as arith does not work on `memref` anyway? (big debate on this happening rn)...

I don't think it makes much sense either; but this is about matching MLIR's implementation, not changing it, IMO?

n-io:
> It appears that `arith.constant` can instantiate memrefs (although I don't know how it would even lower that?), but other arith ops can't operate on memrefs...

It's lowered during bufferization, e.g.:
```
%0 = arith.constant dense<1.234500e-01> : tensor<8xf32>
```
```
// RUN: mlir-opt tests/filecheck/transforms/test.mlir -allow-unregistered-dialect --eliminate-empty-tensors --one-shot-bufferize="allow-unknown-ops"
module {
  memref.global "private" constant @__constant_8xf32 : memref<8xf32> = dense<1.234500e-01> {alignment = 64 : i64}
  %0 = memref.get_global @__constant_8xf32 : memref<8xf32>
}
```

AntonLydike (approved):

PR #2696, opened 2024-06-07
core: Allow to parse string enums

math-fehr:
Add support in the parser to parse an `StrEnum`.
It parses a bare identifier, and check if it is part of the `StrEnum`.

PR #2970, opened 2024-07-31
transform: (stencil-tensorize-z-dimension) Tensorize arith.constant directly

n-io:
Add support for directly tensorising `arith.constant`, e.g.
```
%0 = arith.constant dense<1.666600e-01> : tensor<510xf32>
```

The more generic way of tensorising any scalar operand (independent of its source) would be `linalg.fill`ing the value a newly created empty tensor. This still works as a generic way. However, in the special case that the scalar operand is a constant, this is much shorter.

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -110,9 +110,7 @@ class Constant(IRDLOperation):
 
     @overload
     def __init__(
-        self,
-        value: AnyIntegerAttr | FloatAttr[AnyFloat] | Attribute,
-        value_type: None = None,
+        self, value: AnyIntegerAttr | FloatAttr[AnyFloat], value_type: None = None
     ) -> None: ...
 
     @overload
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -110,7 +110,9 @@ class Constant(IRDLOperation):
 
     @overload
     def __init__(
-        self, value: AnyIntegerAttr | FloatAttr[AnyFloat], value_type: None = None
+        self,
+        value: AnyIntegerAttr | FloatAttr[AnyFloat] | Attribute,
+        value_type: None = None,
     ) -> None: ...
 
     @overload
--- a/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
+++ b/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
@@ -157,7 +157,7 @@ def arithBinaryOpTensorize(
                 rewriter.replace_op(scalar_op.op, tens_const)
             return tens_const.result
         emptyop = EmptyOp((), dest_typ)
-        fillop = FillOp((scalar_op,), (emptyop,), (dest_typ,))
+        fillop = FillOp((scalar_op,), (emptyop.tensor,), (dest_typ,))
         rewriter.insert_op(emptyop, InsertPoint.before(op))
         rewriter.insert_op(fillop, InsertPoint.before(op))
         return fillop.res[0]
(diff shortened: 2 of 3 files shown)


xdsl/dialects/arith.py
```diff
@@ -110,7 +110,9 @@ class Constant(IRDLOperation):
 
     @overload
     def __init__(
-        self, value: AnyIntegerAttr | FloatAttr[AnyFloat], value_type: None = None
+        self,
+        value: AnyIntegerAttr | FloatAttr[AnyFloat] | Attribute,
```
superlopuh, line 114: This doesn't seem like a meaningful overload any more, AnyIntegerAttr and FloatAttr[AnyFloat] are all Attribute, so the `|` are superfluous. Maybe what we want here instead is `TypedAttribute`?

superlopuh: Also, this really feels like it should be in a separate PR, it is not related to stencils directly, I'm glad I clicked through to see what's in this PR despite the title not saying anything about the standard dialects.

n-io: I'm getting confused with PRs. I thought we had merged changes to arith and I'd forgotten to make one change, but it's actually still open in #2696, where the `| Attribute` added here is `| DenseIntOrFPElementsAttr`. Rolling back the change here.

n-io:
This PR depends on #2969, re-run test when this is merged.

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -140,6 +144,24 @@ def arithBinaryOpTensorize(
...
+            if len(scalar_op.uses) > 1:
+                rewriter.insert_op(tens_const, InsertPoint.before(scalar_op.op))
+            else:
+                rewriter.replace_op(scalar_op.op, tens_const)
+            return tens_const.result
+        emptyop = EmptyOp((), dest_typ)
+        fillop = FillOp((scalar_op,), (emptyop.tensor,), (dest_typ,))
+        rewriter.insert_op(emptyop, InsertPoint.before(op))
+        rewriter.insert_op(fillop, InsertPoint.before(op))
+        return fillop.res[0]
```
superlopuh, line 163: one danger with nesting function definitions in Python is that it doesn't act as a namespace, and actually redefines the function every time. It would probably be better to move this function to the top level, and make it private. It would also benefit from a doc comment, as well as the function above it.

n-io: Thanks, fixed as suggested.

superlopuh (approved):
Code changes look good, but I'm not sure about the pass in general, and the test file. It seems like there's quite a lot of inspection and rewriting of operands, when the usual approach is to match directly on the uses. This matters a bit for composability of rewrites, and rewriting some of this tensorization as transformations on the uses would make it easier to read and to test. For example, the linalg.fill where the operand is a constant could be replaced with a new arith.constant of a tensor, leaving the original constant in place (and letting it be dce-ed away in a later stage in the pipeline). Similarly for the arith operations.

n-io:
> Code changes look good, but I'm not sure about the pass in general, and the test file. It seems like there's quite a lot of inspection and rewriting of operands, when the usual approach is to match directly on the uses. This matters a bit for composability of rewrites, and rewriting some of this tensorization as transformations on the uses would make it easier to read and to test. For example, the linalg.fill where the operand is a constant could be replaced with a new arith.constant of a tensor, leaving the original constant in place (and letting it be dce-ed away in a later stage in the pipeline). Similarly for the arith operations.

Thanks, done as discussed :)

PapyChacal (approved):
Nice!

Issue #2959, opened 2024-07-30
Linalg generic-print named ops with hidden regions to match upstream linalg


n-io:
PR #2957 describes a discrepancy between the linalg dialect re-implemented in xdsl and the upstream linalg dialect.

Specifically, linalg *named* ops are printed correctly in the default pretty printing mode, but the generic printing mode should also print a hidden region.

The above PR #2957 makes the xdsl-opt *generic* print for AddOp, SubOp, and MulOp, which previously was:
```
xdsl-opt --print-op-generic tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/ops.mlir

[...]
%sum = "linalg.add"(%2, %2, %3) <{"operandSegmentSizes" = array<i32: 2, 1>}> : (tensor<2x3xf32>, tensor<2x3xf32>, tensor<2x3xf32>) -> tensor<2x3xf32>
```
... match the mlir-opt *generic* print:
```
mlir-opt --mlir-print-op-generic tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/ops.mlir

[...]
%2 = "linalg.add"(%1#0, %1#0, %1#1) <{operandSegmentSizes = array<i32: 2, 1>}> ({
^bb0(%arg0: f32, %arg1: f32, %arg2: f32):
  %21 = "arith.addf"(%arg0, %arg1) <{fastmath = #arith.fastmath<none>}> : (f32, f32) -> f32
  "linalg.yield"(%21) : (f32) -> ()
}) : (tensor<2x3xf32>, tensor<2x3xf32>, tensor<2x3xf32>) -> tensor<2x3xf32>
```

Help is wanted in checking if the generic print differs for the following ops between xdsl-opt and mlir-opt, and to fix this in the code:

 * [x] FillOp
 * [x] TransposeOp
 * [x] MatmulOp
 * [x] PoolingNchwMaxOp
 * [x] Conv2DNchwFchwOp
 * [x] BroadcastOp
 * [x] Is there a non-manual way to verify MLIR compatibility (e.g. in tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/ops.mlir) ?

The `NamedOpBase` class provides standard printing/parsing methods for the common `ins`/`outs`+variadic syntax employed by linalg named ops, and may be re-used for any of these ops.

jorendumoulin:
#2971 addresses matmul and fill op!

jorendumoulin:
[...]

PR #2971, opened 2024-07-31
dialects: (linalg) add correct generic printing for fill, and matmul

jorendumoulin:
I happened to need the fill operator and decided to do the matmul as well. This continues the work of #2959 
I also made sure that the code works on more than just floating points values.
I moved some duplicate code to determine the type arguments of the hidden region into a common method of the `NamedOp` base class.
I also used the implicit builder to build the regions as I think this is much more readable.

Doing the matmul was a bit more tricky:

- for some reason this op prints the attributes in front of the `ins` and `outs`
- generic from includes the `linalg.memoized_indexing_maps` attribute, which are the indexing maps that would be generated when converting to a linalg

### DIFF
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -3,10 +3,11 @@ from __future__ import annotations
 from abc import ABC
 from collections.abc import Mapping, Sequence
 from enum import auto
-from typing import cast
+from typing import ClassVar, cast
 
 from typing_extensions import Self
 
+from xdsl.builder import Builder
 from xdsl.dialects import arith
 from xdsl.dialects.builtin import (
     AffineMapAttr,
@@ -29,14 +30,14 @@ from xdsl.dialects.utils import (
 )
 from xdsl.ir import (
     Attribute,
-    Block,
+    BlockArgument,
     Dialect,
     EnumAttribute,
-    Operation,
     Region,
     SSAValue,
 )
 from xdsl.ir.affine import AffineMap
+from xdsl.ir.core import Operation
 from xdsl.irdl import (
     AttrSizedOperandSegments,
     IRDLOperation,
@@ -391,9 +392,9 @@ class YieldOp(AbstractYieldOperation[Attribute]):
     traits = frozenset([IsTerminator()])
 
 
-class NamedOpBase(IRDLOperation):
+class NamedOpBase(IRDLOperation, ABC):
     """
-    Base class for named ops with hidden region.
+    Abstract base class for named ops with hidden region.
     """
 
     inputs = var_operand_def()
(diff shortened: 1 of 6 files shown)


xdsl/dialects/linalg.py
```diff
@@ -534,6 +530,30 @@ def print(self, printer: Printer):
...
+
+        result: Sequence[AnyFloat | IntegerType] = []
+
+        for op_type in (op.type for op in operands):
+            if isa(op_type, MemRefType[Attribute]):
+                element_type = op_type.get_element_type()
+            elif isa(op_type, TensorType[Attribute]):
+                element_type = op_type.get_element_type()
+            elif isinstance(op_type, Operation):
+                element_type = op_type.results[0].type
```
n-io, line 549: Since `Operation` does not inherit from `Attribute` this case should be invalid imo.

xdsl/dialects/linalg.py
```diff
@@ -475,26 +465,42 @@ def parse(cls, parser: Parser):
...
+            else:
+                attrs = {}
 
         if parser.parse_optional_punctuation("->"):
             res_types = parser.parse_optional_comma_separated_list(
                 parser.Delimiter.PAREN, parser.parse_attribute
             )
             if res_types is None:
-                res_types = [[parser.parse_attribute()]]
+                res_types = [parser.parse_attribute()]
```
n-io, line 482: Somehow I thought the double list is needed by virtue of results being a variadic.

xdsl/dialects/linalg.py
```diff
@@ -534,6 +530,30 @@ def print(self, printer: Printer):
...
+    def body_arg_types(
+        operands: Sequence[SSAValue],
+    ) -> Sequence[AnyFloat | IntegerType]:
+        """
+        Return the element types of the arguments of the body of this operation
+        """
+
+        result: Sequence[AnyFloat | IntegerType] = []
+
+        for op_type in (op.type for op in operands):
```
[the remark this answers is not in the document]
jorendumoulin: Yes, this is probably somewhat cleaner

xdsl/dialects/linalg.py
jorendumoulin: Indeed! This was added because of a failing test that still managed to pass an operation as an attribute, but I now just fixed the test properly instead.

xdsl/dialects/linalg.py
jorendumoulin: Aha yes, they were put in triple brackets because of an error somewhere else, but i fixed this and put the double brackets back!

n-io (approved):
Thanks, looks good!

superlopuh (approved):

xdsl/dialects/linalg.py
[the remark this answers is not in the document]
n-io: Very minor, but the `op_type = op.type` line was more to show equivalence rather than meant to be a concrete suggestion. Happy either way though!

PR #2973, opened 2024-08-01
dialects: (transform) Add some transform dialect operations

JonasCrols:
Adds 7 operations of the transform dialect (SequenceOp, YieldOp, TileOp, TileToForallOp, SelectOp, NamedSequenceOp, CastOp), as well as 4 new types (OperationType, ParamType, TypeParamType,TransformHandleType). All with tests and filecheck


JonasCrols:
@jorendumoulin @JosseVanDelm please sort out reviewers

tests/dialects/test_transform.py
```diff
@@ -0,0 +1,102 @@
...
+    affineMapType = transform.AffineMapType()
+    assert affineMapType.name == "transform.affine_map"
+
+
+test_transform_op_type()
+test_transform_param_type()
+test_transform_any_op_type()
+test_transform_any_param_type()
+test_transform_any_value_type()
+test_transform_affine_map_type()
```
jorendumoulin, line 46: you don't have to call these

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,263 @@
...
+        scalable_sizes: DenseArrayBase | None = None,
+    ):
+        super().__init__(
+            operands=(target, dynamic_sizes),
+            properties={
+                "static_sizes": static_sizes,
+                "interchange": interchange,
+                "scalable_sizes": scalable_sizes,
+            },
+            # TODO: Figure out how to handle the result types with var_result_def()
```
jorendumoulin, line 161: delete comment

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,263 @@
...
+
+@irdl_attr_definition
+class TypeParamType(TransformHandleType):
+    name = "transform.type"
+
+
+@irdl_attr_definition
+class FailurePropagationModeAttr(ParametrizedAttribute):
+    name = "failures"
+
```
jorendumoulin, line 93:
```suggestion
from xdsl.utils.str_enum import StrEnum
from xdsl.ir import EnumAttribute

class FailurePropagationModeType(StrEnum):
    PROPAGATE = "propagate"
    SUPPRESS = "suppress"

@irdl_attr_definition
class FailurePropagationModeAttr(EnumAttribute[FailurePropagationModeType]):
    name = "transform.failures"

```

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,263 @@
...
+
+@irdl_op_definition
+class SequenceOp(IRDLOperation):
+    name = "transform.sequence"
+
+    # TODO: Find out how to also use the enum FailurePropagationModeAttr as well as AnyIntegerAttr
+
+    body = region_def("single_block")
+    failure_propagation_mode = prop_def(AnyIntegerAttr)
```
jorendumoulin, line 109: why does it need to be anyintegerattr? can it not just be failureproagationmodeattr?

JonasCrols: both IntegerAttr and FailurePropagationModeAttr are supported. for most use cases, it looks like the IntegerAttr is used so I just used this until FailurePropagationModeAttr was also completed

tests/dialects/test_transform.py
```diff
@@ -0,0 +1,90 @@
...
+        ]
+    )
+
+    target = block.args[0]
+    static_sizes = DenseArrayBase.create_dense_int_or_index(IndexType(), [8, 8])
+    tile_op = transform.TileOp(
+        target=target,
+        dynamic_sizes=[],
+        static_sizes=static_sizes,
+    )
```
JosseVanDelm, line 82: same thing here, can you also let the init accept a python list of integers maybe?

xdsl/dialects/transform.py
JosseVanDelm, line 1: Can you add some simple docstrings?, I'll add one suggestion as an example

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,263 @@
...
 @irdl_attr_definition
-class AffineMapType(ParametrizedAttribute, TypeAttribute):
+class AffineMapType(TransformHandleType):
     name = "transform.affine_map"
 
 
 @irdl_attr_definition
-class AnyOpType(ParametrizedAttribute, TypeAttribute):
+class AnyOpType(TransformHandleType):
     name = "transform.any_op"
```
JosseVanDelm, line 54:
```suggestion
class AnyOpType(TransformHandleType):
    """
    https://mlir.llvm.org/docs/Dialects/Transform/#anyoptype
    """
    name = "transform.any_op"
```

JosseVanDelm (commented):
Thank you @JonasCrols! This is a very cool PR!

I made a few nit comments on:
* Adding docstrings to certain ops.
* Support other inits (e.g. also support a list of integers instead of just a DenseArrayAttr).
  I think some more refactoring could make that easily possible, although I don't think this is    essential for this PR.
  
My biggest issue is that currently it is not very clear which version of the transform dialect we would be supporting. 
I think this is an important discussion to have first, especially since it seems quite unstable (as you have noticed). I'm curious what you and others think about this.

I think it would be nice/important(?) if we could also add transform interop tests to this PR.

tests/filecheck/mlir-conversion/with-mlir/dialects/transform/transform_types.mlir
```diff
@@ -1,13 +1,49 @@
-// RUN: xdsl-opt %s | xdsl-opt | mlir-opt --allow-unregistered-dialect | filecheck %s
```
superlopuh, line 1: WHy change this to print generic?

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,325 @@
...
+        )
+
+
+@irdl_op_definition
+class TileOp(IRDLOperation):
+    """
+    https://mlir.llvm.org/docs/Dialects/Transform/#transformstructuredtile_using_for-transformtileusingforop
+    """
+
+    name = "transform.structured.tile"  # "transform.structured.tile_using_for" as of mlir 18.0
```
superlopuh, line 182: I would say let's stick to the name as it is in the commit version that xDSL is coupled with, it's a little confusing to have both of these at the same time. When we update the MLIR commit, we will need to change this name to whichever is more appropriate.

tests/filecheck/dialects/transform/transform_types.mlir
```diff
@@ -1,13 +1,49 @@
 // RUN: XDSL_ROUNDTRIP
```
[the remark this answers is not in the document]
superlopuh: It's in the readme, I would say that we want to match the syntax 1:1 with the version of MLIR in that commit, and not the latest MLIR, and if there's a reason to update the MLIR version then we can do that also.

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,325 @@
...
+
+@irdl_op_definition
+class SequenceOp(IRDLOperation):
+    """
+    https://mlir.llvm.org/docs/Dialects/Transform/#transformsequence-transformsequenceop
+    """
+
+    name = "transform.sequence"
+
+    # TODO: Find out how to also use the enum FailurePropagationModeAttr as well as AnyIntegerAttr
```
superlopuh, line 149: What do you mean by this?

superlopuh (commented):
Really fantastic to see this, thank you!

superlopuh:
Please reach out on the Zulip if you need any help with the tests

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,325 @@
...
+    result_def,
+    var_operand_def,
+    var_result_def,
 )
+from xdsl.traits import IsolatedFromAbove, IsTerminator
+from xdsl.utils.str_enum import StrEnum
+
+
+class TransformHandleType(ParametrizedAttribute, TypeAttribute, ABC):
+    name: ClassVar[str] = field(init=False, repr=False)
```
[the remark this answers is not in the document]
JonasCrols: If I don't specify it. it raises an AttributeError where the class does not have the name Attribute, even though it's superclass has it. It is weird, but adding that line fixes it.

xdsl/dialects/transform.py
JonasCrols: In milt, both integers can be used, as well as an attribute-type specifically made for this. I needed to figure out how I can support both at the same time. but this is now fixed

superlopuh: Then it would be good to remove the todo

xdsl/dialects/transform.py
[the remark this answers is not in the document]
superlopuh: On which line is the error raised?

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,376 @@
...
+    https://mlir.llvm.org/docs/Dialects/Transform/#transformsequence-transformsequenceop
+    """
+
+    name = "transform.sequence"
+
+    T = Annotated[AnyIntegerOrFailurePropagationModeAttr, ConstraintVar("T")]
+
+    body = region_def("single_block")
+    failure_propagation_mode: Property = prop_def(
+        T  # pyright: ignore[reportArgumentType]
```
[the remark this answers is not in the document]
superlopuh: I'm currently trying to refactor some things for the newer Pyright version to be happy, but for the time being please use `make pyright` as the source of truth

xdsl/dialects/transform.py
```diff
@@ -1,41 +1,376 @@
...
+        YieldOp,
+        TileOp,
+        TileToForallOp,
+        SelectOp,
+        NamedSequenceOp,
+        CastOp,
+    ],
+    [
+        # Types
+        TransformHandleType,
```
superlopuh, line 365: this is an abstract helper class, and should not be included here

xdsl/dialects/transform.py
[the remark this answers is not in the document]
superlopuh: I have a feeling that now that it's not passed as one of the Attributes in the dialect init it might be ok without the name

superlopuh (approved):
Looks great, thank you!

What's your plan for this? Are you interested in implementing an interpreter for the dialect in xDSL, or are you planning to use the MLIR transforms directly?

JonasCrols:
> Looks great, thank you!
> 
> What's your plan for this? Are you interested in implementing an interpreter for the dialect in xDSL, or are you planning to use the MLIR transforms directly?

We'll be using this to automatically tile all matmul operations. As of now, our plan is to use the mlir `--transform-interpreter` pass directly as it works fine for our goals. we're not currently planning on adding the interpreter to xDSL

jorendumoulin (approved):

PR #2982, opened 2024-08-05
dialects: Enable bufferized stencil.apply

PapyChacal:
I.e., make stencil.apply able to express computations on reference-semantic fields through destination operands, instead of only value-semantic through results.

### DIFF
--- a/xdsl/dialects/stencil.py
+++ b/xdsl/dialects/stencil.py
@@ -481,6 +481,9 @@ class ApplyOp(IRDLOperation):
         printer.print(") ")
         printer.print_op_attributes(self.attributes, print_keyword=True)
         printer.print_region(self.region, print_entry_block_args=False)
+        if self.bounds is not None:
+            printer.print(" to ")
+            self.bounds.print_parameters(printer)
 
     @classmethod
     def parse(cls: type[ApplyOp], parser: Parser):
--- a/xdsl/dialects/stencil.py
+++ b/xdsl/dialects/stencil.py
@@ -439,8 +439,11 @@ class ApplyOp(IRDLOperation):
     B = Annotated[Attribute, ConstraintVar("B")]
 
     args: VarOperand = var_operand_def(Attribute)
+    dest = var_operand_def(FieldType)
     region: Region = region_def()
     res: VarOpResult = var_result_def(TempType)
+
+    bounds = opt_prop_def(StencilBoundsAttr)
 
     traits = frozenset(
         [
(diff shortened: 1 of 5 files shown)


xdsl/dialects/stencil.py
```diff
@@ -1315,7 +1379,14 @@ def verify_(self) -> None:
...
         apply = cast(ApplyOp, self.parent_op())
-        res_types = [cast(TempType[Attribute], r.type).element_type for r in apply.res]
+        if len(apply.res) > 0:
+            res_types = [
+                cast(TempType[Attribute], r.type).element_type for r in apply.res
+            ]
+        else:
+            res_types = [
+                cast(FieldType[Attribute], o.type).element_type for o in apply.dest
+            ]
```
n-io, line 1389: I think `apply.dest` always has to be present and match, while `apply.res` only has to be there and match in unbufferized mode, no?

xdsl/dialects/stencil.py
```diff
@@ -458,16 +463,27 @@ def print_assign_argument(args: tuple[BlockArgument, SSAValue, Attribute]):
...
         printer.print_list(
             zip(self.region.block.args, self.args, (a.type for a in self.args)),
             print_assign_argument,
         )
         printer.print(") -> (")
-        printer.print_list((r.type for r in self.res), printer.print_attribute)
+        if self.dest:
+            printer.print_list(self.dest, print_destination_operand)
+        else:
+            printer.print_list((r.type for r in self.res), printer.print_attribute)
```
n-io, line 480: Not so keen on this syntax, if I'm understanding correctly that it hides that the bufferized version does not actually have a SSA return value.

Also, can you not have dest+res in the unbufferized mode?

xdsl/dialects/stencil.py
PapyChacal: To simplify a lot other parts, I went with apply having to have either res or dest, never both.

xdsl/dialects/stencil.py
PapyChacal: I went with apply only able to be fully value-semantic (having results in res) or reference-semantics (having destinations in dest), never a mix of both

PapyChacal: I could clarify that in a docstring I guess

n-io: My point was something else, that dest is printed after the `->` (which unhelpfully isn't shown in the 4-line snippet above but comes just before)

n-io: To keep it simple, you could have `-> ( ... )` for results and `dest( .... )` or `outs(....)` for the destination-style passing.

n-io: You could also consider re-using `print_assign_argument`

PapyChacal: Done with outs, WDYT?

n-io: lgtm, thanks for this

PR #2983, opened 2024-08-05
transformations: First implementation of stencil-bufferize

PapyChacal:
A first implementation of stencil bufferization, enough to demonstrate on the main tests.

It is not enough for all goals in scope, but figured it's big enough to merge a first step.

PR #2985, opened 2024-08-05
dialects: Add stencil features meant for bufferization

PapyChacal:
- Add stencil.alloc to allocate a new field
- Enable a field result on stencil.buffer, meaning "The underlying buffer of the passed values"

Split from:
- #2983

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/stencil.py
+++ b/xdsl/transforms/canonicalization_patterns/stencil.py
@@ -11,7 +11,15 @@ from xdsl.rewriter import InsertPoint
 from xdsl.transforms.common_subexpression_elimination import cse
 
 
-class RedundantOperands(RewritePattern):
+class AllocUnused(RewritePattern):
+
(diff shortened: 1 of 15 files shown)


xdsl/transforms/canonicalization_patterns/stencil.py
```diff
@@ -11,7 +11,15 @@
 from xdsl.transforms.common_subexpression_elimination import cse
 
 
-class RedundantOperands(RewritePattern):
+class AllocUnused(RewritePattern):
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.AllocOp, rewriter: PatternRewriter) -> None:
+        if not op.field.uses:
+            rewriter.erase_op(op)
```
PapyChacal, line 19: This would be best done through usual DCE with an implementation of side-effect values; but I felt like moving to the application for now.

n-io (approved):

PR #2988, opened 2024-08-07
dialects: (riscv_snitch) add vfmax.s

superlopuh:
Useful for relu


tests/filecheck/dialects/riscv_snitch/assembly_emission.mlir
superlopuh, line 1: Small refactor, all the result names are now independent of the order so we can move them around, and each of the register operands and results is different so the test keeps track of the order these are printed in.

nazavode (approved):

PR #2995, opened 2024-08-08
(core) Support UnitAttr in assembly format

n-io:
This PR enables support for UnitAttr in the assembly format parser.

The upstream MLIR assembly format syntax to represent a UnitAttr may be for instance ([source](https://mlir.llvm.org/docs/Dialects/BufferizationOps/#bufferizationto_memref-bufferizationtomemrefop)):
```
`bufferization.to_memref` $tensor (`read_only` $read_only^)? attr-dict `:` type($memref)
```

Specifically, if the keyword `read_only` is present, then the UnitAttr (of the same name) will be set. In xdsl, it would simply be set by assigning `= UnitAttr()`. The approach here is similar, if the optional keyword is present, then the value should be set.

This change consists of two parts:
* The first is to not change the `unique_base` variable in the assembly format parser in the way it's usually done for builtins. This causes `AttributeVariable.parse` to not execute the `parser.parse_attribute()` in the `unique_base is None` case (expecting to parse an actual value), but instead to use the custom parsing method of UnitAttr in the `issubclass(.. , ParametrizedAttribute)` case
* The second is to implement a `UnitAttr.parse_parameters` method returning an empty list, overriding the parse method of the superclass which attempts to parse an actual value.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -463,7 +463,7 @@ def test_optional_property(program: str, generic_program: str):
             '"test.optional_unit_attr_prop"() : () -> ()',
         ),
         (
-            "test.optional_unit_attr_prop unit_attr",
+            "test.optional_unit_attr_prop unit_attr ",  # todo this prints an extra whitespace
             '"test.optional_unit_attr_prop"() <{"unit_attr"}> : () -> ()',
         ),
     ],
@@ -494,7 +494,7 @@ def test_optional_unit_attr_property(program: str, generic_program: str):
             '"test.optional_unit_attr"() : () -> ()',
         ),
         (
-            "test.optional_unit_attr unit_attr",
+            "test.optional_unit_attr unit_attr ",  # todo this prints an extra whitespace
             '"test.optional_unit_attr"() <{"unit_attr"}> : () -> ()',
         ),
     ],
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -8,7 +8,7 @@ from typing import Annotated, Generic, TypeVar
 import pytest
 
 from xdsl.context import MLContext
-from xdsl.dialects.builtin import I32, IntegerAttr, ModuleOp
+from xdsl.dialects.builtin import I32, IntegerAttr, ModuleOp, UnitAttr
 from xdsl.dialects.test import Test, TestType
 from xdsl.ir import (
     Attribute,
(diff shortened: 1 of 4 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -11,6 +11,7 @@
 from dataclasses import dataclass, field
 from typing import Any, Literal, cast
 
+from xdsl.dialects.builtin import UnitAttr
```
superlopuh, line 14: this seems unfortunate, I would like to avoid importing any dialects here, it seems like it breaks the layering conceptually

n-io: Can we live with the fact that it would print an extra whitespace?

superlopuh: Where is the whitespace printed?

xdsl/dialects/builtin.py
```diff
@@ -431,6 +431,11 @@ def bitwidth(self) -> int:
 class UnitAttr(ParametrizedAttribute):
     name = "unit"
 
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        """Parse the attribute parameters."""
+        return []
+
```
superlopuh, line 438: why is this necessary?

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -455,6 +455,68 @@ class OptionalPropertyOp(IRDLOperation):
...
+)
+def test_optional_unit_attr_attribute(program: str, generic_program: str):
+    """Test the parsing of optional UnitAttr operands"""
+
+    @irdl_op_definition
+    class OptionalUnitAttrOp(IRDLOperation):
+        name = "test.optional_unit_attr"
+        unit_attr = opt_prop_def(UnitAttr)
+
+        assembly_format = "(`unit_attr` $unit_attr^)? attr-dict"
```
superlopuh, line 510: Could you please make the strings different to disambiguate? Maybe something like this?

```suggestion
        unit_attr_prop = opt_prop_def(UnitAttr)

        assembly_format = "(`unit_attr_str` $unit_attr_prop^)? attr-dict"
```

xdsl/irdl/declarative_assembly_format.py
PapyChacal: > this seems unfortunate, I would like to avoid importing any dialects here, it seems like it breaks the layering conceptually

I don't think we could avoid it *and* give special treatment to those here, which is necessary for compatibility

superlopuh: But I don't understand the special treatment. Could you maybe post what happens today without these changes?

superlopuh: Ah I see it was added in the latest commit. Which is the line of code that prints the extra space? Shouldn't it be avoided? Do we want this to work differently for UnitAttr to any other parametrized attribute with no parameters?

PapyChacal: I think the support for UnitAttr should be structured differently too; I think a custom directive would make more sense, rather than adapting the current one to special case UnitAttr.

PapyChacal: More specifically, I think that rather than hacking UnitAttr, the generic name-eliding feature and all to behave syntactically like we want, it would be way clearer to have a custom directive (maybe an OptionaVariable subclass) encoding the corner case of UnitAttr.
UnitAttr in declarative syntax is a corner-case by itself, and I think it would be clearer this way.
https://mlir.llvm.org/docs/DefiningDialects/Operations/#unit-attributes

xdsl/dialects/builtin.py
PapyChacal: See my comment on the other thread on this
https://github.com/xdslproject/xdsl/pull/2995#discussion_r1709570381

xdsl/irdl/declarative_assembly_format.py
PapyChacal: I'm happy to have a short call to clarify if it's unclear - I know that part of the code base can be a bit overwhelming :upside_down_face:

tests/irdl/test_declarative_assembly_format.py
n-io: Before doing this, I'd note that this would diverge from how the other tests do it, as for the ones immediately above and below

superlopuh: OK fair enough, please disregard

superlopuh:
OK, thank you both for the explanations and discussion. I see why the special handling is needed, indeed there seems to not be a much better way to do this than with special handling in the printer and parser. I suggest reverting to the state before my suggestions and continuing from there?

xdsl/irdl/declarative_assembly_format.py
superlopuh: No time for calls today, let's just keep iterating here

PapyChacal: Change of direction done!

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -801,6 +802,21 @@ class OptionalAttributeVariable(AttributeVariable, OptionalVariable):
     """
 
 
+class OptionalUnitAttrVariable(OptionalAttributeVariable):
+    """
+    An optional UnitAttr variable
```
superlopuh, line 807: Could you please expand a bit here why this is needed? Something about UnitAttr not being printed in the IR, so special handling is needed, maybe with a link to the MLIR documentation you posted in the other comment.

superlopuh (approved):
Thank you!

PapyChacal (approved):

xdsl/irdl/declarative_assembly_format.py
n-io: Doing as suggested

PR #3001, opened 2024-08-09
transformations: stencil bufferization polish

PapyChacal:
Step back on the idea of allocating all apply results to then fold away and instead just fold stores into apply. It's cheaper, simpler, and so far works just as well!

Also rely on the new side-effect value information to improve side-effect analysis.

Both changes allow to bufferize a full open earth compiler example now!

Minor test changes, the bufferization used to blindly allocate buffers for unused results, now it scraps them. I guess it's debatable if that should be its role? Note that it does so by reusing the canonicalization pattern, no logic is actually duplicated.


xdsl/dialects/stencil.py
```diff
@@ -561,17 +561,20 @@ def parse_operand():
     def get(
         args: Sequence[SSAValue] | Sequence[Operation],
         body: Block | Region,
-        result_types: Sequence[TempType[Attribute]],
+        result_types: Sequence[TempType[Attribute]] | None = None,
+        bounds: StencilBoundsAttr | None = None,
     ):
-        assert len(result_types) > 0
-
+        assert result_types or bounds
```
superlopuh, line 567: What happens if result_types is an empty sequence? Surely you want to have no results?
```suggestion
        assert result_types is not None or bounds is not None
```

xdsl/dialects/stencil.py
```diff
@@ -646,6 +649,14 @@ def get_accesses(self) -> Iterable[AccessPattern]:
                 accesses.append(offsets)
             yield AccessPattern(tuple(accesses))
 
+    def get_bounds(self):
+        if self.bounds is not None:
+            return self.bounds
+        else:
+            assert len(self.res) > 0
```
superlopuh, line 656:
```suggestion
            assert self.res
```

xdsl/dialects/stencil.py
PapyChacal: Yeah that's semantic!
Basically the op needs a non-empty result list or bounds, so the `(result_types = [], bounds = None)` case is invalid too :slightly_smiling_face:

superlopuh: Would this work? `result_types: Sequence[TempType[Attribute]] = ()`

PapyChacal: Sure, but I'd still want to check it for non-emptiness? I'm confused

superlopuh: This avoids the `None`/ empty ambiguity

PapyChacal: Done!

n-io (approved):

superlopuh (approved):

PR #3002, opened 2024-08-09
dialects: (irdl) Update for mlir compatibility

alexarice:
Added the `irdl.c_pred` operation as well as adding variadicity attributes to operands and results. Also updated the pyrdl-to-irdl script so that it prints symbols correctly.


xdsl/dialects/irdl/pyrdl_to_irdl.py
```diff
@@ -51,7 +51,7 @@ def op_def_to_irdl(op: type[IRDLOperation]) -> OperationOp:
     if result_values:
         builder.insert(ResultsOp(result_values))
 
-    return OperationOp(op_def.name, Region([block]))
+    return OperationOp(op_def.name.split(".")[-1], Region([block]))
```
superlopuh, line 54: this doesn't quite feel right, are there any tests for this? My understanding is the dialect name is everything up to the first dot, and the op name is everything after it

superlopuh: We probably want to centralise this logic somewhere

xdsl/dialects/irdl/irdl.py
```diff
@@ -35,12 +38,47 @@
...
+
+
+@irdl_attr_definition
+class VariadicityArrayAttr(ParametrizedAttribute, SpacedOpaqueSyntaxAttribute):
+    name = "irdl.variadicity_array"
+
+    value: ParameterDef[ArrayAttr[VariadicityAttr]]
+
+    def __init__(self, variadicities: Iterable[VariadicityEnum]) -> None:
+        array_attr = ArrayAttr(map(lambda x: VariadicityAttr(x), variadicities))
```
superlopuh, line 66: Please convert the maps to tuples, they're noticeably faster and I would argue easier to read
```suggestion
        array_attr = ArrayAttr(tuple(VariadicityAttr(x) for x in variadicities))
```

xdsl/dialects/irdl/irdl.py
```diff
@@ -223,21 +296,34 @@ class OperandsOp(IRDLOperation):
...
-        super().__init__(operands=[args])
+    def __init__(self, args: Sequence[tuple[VariadicityEnum, SSAValue] | SSAValue]):
+        args_list = [
+            (VariadicityEnum.Single, x) if isinstance(x, SSAValue) else x for x in args
+        ]
+        operands = [x[1] for x in args_list]
+        attributes = {
+            "variadicity": VariadicityArrayAttr(map(lambda x: x[0], args_list))
+        }
+        super().__init__(operands=[operands], attributes=attributes)
```
superlopuh, line 311:
```suggestion
        args_list = [
            (VariadicityEnum.Single, x) if isinstance(x, SSAValue) else x for x in args
        ]
        operands = tuple(operand for _, operand in args_list)
        attributes = {
            "variadicity": VariadicityArrayAttr(tuple(v for v, _ in args_list))
        }
        super().__init__(operands=[operands], attributes=attributes)
```

xdsl/dialects/irdl/irdl.py
```diff
@@ -115,6 +153,24 @@ def qualified_name(self):
...
+
+    output: OpResult = result_def(AttributeType())
+
+    assembly_format = "$pred attr-dict"
+
+    def __init__(self, pred: str | StringAttr):
+        if isinstance(pred, str):
+            pred = StringAttr(pred)
+        super().__init__(attributes={"pred": pred}, result_types=[AttributeType()])
+
```
PapyChacal, line 172: De we really reaaally want this one? I like my denial of its necessity :smiling_face_with_tear:

xdsl/dialects/irdl/pyrdl_to_irdl.py
PapyChacal: Yep, should be the other way around! Split on the first dot and keep whatever's after.

More concretely, some ops have nested-like names, say `scf.forall.in_parallel` : dialect still is just `scf`, op name is `forall.in_parallel`, dot including.

And yes, we probably want to centraluze it ahah

PapyChacal (approved):
Yay :heart:

xdsl/dialects/irdl/pyrdl_to_irdl.py
alexarice: Any idea where the centralisation should live?

xdsl/dialects/irdl/irdl.py
alexarice: Probably necessary if you ever want to read something made by the tblgen-to-irdl script

xdsl/dialects/irdl/pyrdl_to_irdl.py
```diff
@@ -69,7 +69,7 @@ def attr_def_to_irdl(
         param_values.append(constraint_to_irdl(builder, param[1]))
     builder.insert(ParametersOp(param_values))
 
-    return AttributeOp(attr_def.name, Region([block]))
+    return AttributeOp(attr_def.name.split(".", 1)[1], Region([block]))
```
superlopuh, line 72: Can you please add a helper function with a test somewhere? Maybe on `Dialect` itself, as a static method? Something like `Dialect.split_name(name: str) -> tuple[str, str]`

xdsl/dialects/irdl/pyrdl_to_irdl.py
alexarice: Now rebased onto #3003

xdsl/dialects/irdl/irdl.py
```diff
@@ -35,12 +38,47 @@
...
 
 ################################################################################
 # Dialect, Operation, and Attribute definitions                                #
 ################################################################################
 
 
+class VariadicityEnum(StrEnum):
+    Single = "single"
+    Optional = "optional"
+    Variadic = "variadic"
```
superlopuh, line 51:
```suggestion
    SINGLE = "single"
    OPTIONAL = "optional"
    VARIADIC = "variadic"
```
It's ugly but standard :)

xdsl/dialects/irdl/irdl.py
```diff
@@ -35,12 +38,47 @@
...
+
+
+@irdl_attr_definition
+class VariadicityAttr(EnumAttribute[VariadicityEnum], SpacedOpaqueSyntaxAttribute):
+    name = "irdl.variadicity"
```
superlopuh, line 56:
```suggestion
class VariadicityAttr(EnumAttribute[VariadicityEnum], SpacedOpaqueSyntaxAttribute):
    name = "irdl.variadicity"
    
    SINGLE = VariadicityAttr(VariadicityEnum.SINGLE)
```
etc

superlopuh (approved):

PR #3003, opened 2024-08-09
misc: Add `Dialect.split_name` method

alexarice:
Adds a `Dialect.split_name` method as discussed in #3002

### DIFF
--- a/tests/test_dialect_utils.py
+++ b/tests/test_dialect_utils.py
@@ -165,3 +165,11 @@ def test_split_name(name: str, expected_1: str, expected_2: str):
     result_1, result_2 = Dialect.split_name(name)
     assert result_1 == expected_1
     assert result_2 == expected_2
+
+
+def test_split_name_failure():
+    with pytest.raises(ValueError) as e:
+        Dialect.split_name("test")
+
+    print(e)
+    assert e.value.args[0] == ("Invalid operation or attribute name test.")
--- a/tests/test_dialect_utils.py
+++ b/tests/test_dialect_utils.py
@@ -11,7 +11,7 @@ from xdsl.dialects.utils import (
     parse_dynamic_index_without_type,
     print_dynamic_index_list,
 )
-from xdsl.ir import SSAValue
+from xdsl.ir import Dialect, SSAValue
 from xdsl.parser import Parser, UnresolvedOperand
 from xdsl.printer import Printer
 from xdsl.utils.test_value import TestSSAValue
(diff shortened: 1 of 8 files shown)


xdsl/ir/core.py
```diff
@@ -61,6 +61,14 @@ def attributes(self) -> Iterator[type[Attribute]]:
     def name(self) -> str:
         return self._name
 
+    @staticmethod
+    def split_name(name: str) -> tuple[str, str]:
+        x = name.split(".", 1)
+        if len(x) == 2:
+            return (x[0], x[1])
+        else:
+            return (name, "")
```
superlopuh, line 70: I don't think we want the special logic, we might as well assert that there are two components, it's user error to provide a string that is neither an operation nor an attribute name.
```suggestion
        d, rest = name.split(".", 1)
        return (d, rest)
```

alexarice: That makes sense, is it worth throwing a custom exception or just let the IndexError get thrown?

superlopuh: I guess it would be a little nicer to have a custom exception, I'll reword my suggestion

xdsl/ir/core.py
```diff
@@ -61,6 +61,12 @@ def attributes(self) -> Iterator[type[Attribute]]:
     def name(self) -> str:
         return self._name
 
+    @staticmethod
+    def split_name(name: str) -> tuple[str, str]:
+        names = name.split(".", 1)
+        assert len(names) == 2, f"Invalid operation or attribute name {name}"
+        return tuple(names)
```
superlopuh, line 68: Oh no

alexarice: yeah that didn't go well

superlopuh:
```suggestion
        first, second = names
        return (first, second)
```

alexarice: any reason for this?

superlopuh: Pyright isn't smart enough to see that when a list of length 2 is passed to the tuple constructor from an iterable, the resulting tuple will also have length 2 (`tuple[str, str]`) as opposed to dynamic (`tuple[str, ...]`).

alexarice: ah right, it fixes the pyright error

tests/test_dialect_utils.py
superlopuh, line 1: Since we've added an error, it would be good to add a pytest that tracks the error message. The more I think of it, the more it seems that raising `ValueError` instead of asserting would be a little cleaner. Sorry about the churn. There are a bunch of examples in the tests of how to catch errors and test for contents.

tests/test_dialect_utils.py
```diff
@@ -152,3 +152,24 @@ def test_parse_dynamic_index_list_with_custom_delimiter():
...
+    result_1, result_2 = Dialect.split_name(name)
+    assert result_1 == expected_1
+    assert result_2 == expected_2
+
+
+def test_split_name_failure():
+    with pytest.raises(ValueError) as e:
+        Dialect.split_name("test")
+
+    print(e)
```
superlopuh, line 174:
```suggestion
    print(e)
```

tests/test_parser_error.py
```diff
@@ -15,15 +15,15 @@
 
 
 @irdl_op_definition
-class UnkownOp(IRDLOperation):
-    name = "unknown"
+class Unknown(IRDLOperation):
```
superlopuh, line 18:
```suggestion
class UnknownOp(IRDLOperation):
```
This change seems unnecessary

superlopuh (approved):

PR #3004, opened 2024-08-09
transformations: (csl_stencil) Add bufferization pass

n-io:
Change `stencil.field` to memrefs, change all tensors to memrefs, and perform a 'partial bufferization' by inserting `bufferization.to_tensor` and `bufferization.to_memref` ops where needed to that other dialects can still operate on tensors.

Todo:
* [x] fix verification
* [x] add docstrings
* [x] review setting of `restrict`, `writable`, and `read_only` of the `bufferization` ops
* [x] add extract_slice needed by linalg bufferize

Edit: I think I re-used a branch name of a merged branch, ignore some of the commits


superlopuh:
Can you please assign yourself and add the appropriate labels?

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+def tensor_to_memref_type(t: TensorType[Attribute]) -> memref.MemRefType[Attribute]:
+    """Type conversion from tensor to memref."""
+    return memref.MemRefType(t.get_element_type(), t.get_shape())
+
+
+def to_memref_op(op: SSAValue) -> bufferization.ToMemrefOp:
+    """Creates a `bufferization.to_memref` operation."""
+    assert isa(op.type, TensorType[Attribute])
```
PapyChacal, line 38: nit: Shaped & Container[Attribute]?

PapyChacal: nvm, it's expected to be a memref by the op, not only for the type

n-io: usually ok, but the op takes a `TensorType[Attribute]` as input anyway, so might be fine?

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -0,0 +1,313 @@
...
+            ],
+            [
+                (
+                    tensor_to_memref_type(t.get_element_type())
+                    if isa(t, stencil.FieldType[TensorType[Attribute]])
+                    else t
+                )
+                for t in op.function_type.outputs
+            ],
+        )
```
PapyChacal, line 264: I did that in a few places but it would be advisable to replace with a new op having this new result type;
For things like proper worklist management and all; doing it this way, you might get surprises if you expect it to trigger other patterns!
I don't mind myself about merging it either way; just a note

PapyChacal (approved):
I'm not getting all semantics used so just reviewed the code itself!

xdsl/transforms/csl_stencil_bufferize.py
n-io: Thanks good point, done as suggested.

PR #3005, opened 2024-08-10
misc: don't set pass spec arguments from instance when they are the default value

superlopuh:
This is useful in a few places, it makes the `xdsl-gui` pass pipeline a little cleaner, as well as the marimo notebook showing the different stages of the pipeline in the slides.

### DIFF
--- a/tests/test_pass_to_spec.py
+++ b/tests/test_pass_to_spec.py
@@ -46,7 +46,7 @@ class SimplePass(ModulePass):
 
 
 @pytest.mark.parametrize(
-    "test_pass, test_spec_with_default, test_spec_no_default",
+    "test_pass, test_spec",
     (
         (
             CustomPass(3, (1, 2), None, ("clown", "season")),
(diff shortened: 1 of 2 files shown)


xdsl/passes.py
```diff
@@ -116,32 +116,39 @@ def required_fields(cls: type[ModulePassT]) -> set[str]:
             field.name for field in dataclasses.fields(cls) if not _is_optional(field)
         }
 
-    def pipeline_pass_spec(self) -> PipelinePassSpec:
+    def pipeline_pass_spec(self, *, include_default: bool = False) -> PipelinePassSpec:
         """
         This function takes a ModulePass and returns a PipelinePassSpec.
+
+        If `include_default` is `True`, then add arguments that are optional are not
+        included in the spec.
```
webmiche, line 124: This sentence is kinda hard for me to parse. I think I understand it, but how about this?

```suggestion
        If `include_default` is `True`, then optional arguments are not included in the spec.
```

tests/test_pass_to_spec.py
webmiche, line 1: Why not split the tests to one testing with default and one without? This makes it much simpler and more straight-forward to understand what failed, if anything failed.

webmiche (approved):

dshaaban01 (approved):

tests/test_pass_to_spec.py
superlopuh: Sure I'll do that

webmiche (approved):
LGTM!

PR #3007, opened 2024-08-10
CI: use .lint when required by ruff

superlopuh:
As a consequence of updating ruff, it now raises warnings. This updates the config in accordance with ruff's suggestion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,8 +95,8 @@ typeCheckingMode = "strict"
 profile = "black"
 
 [tool.ruff]
-select = ["E", "F", "W", "I", "UP", "PT", "TID251"]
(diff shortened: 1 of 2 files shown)


tobiasgrosser (commented):
I do not understand why this worked before and now this is required? Maybe add 1-2 sentences into the commit message?

superlopuh:
It's a consequence of updating ruff, it now raises warnings. You will see the warnings if you run `ruff check` in the repo. We didn't notice them because it's the precommit version that runs normally, and that was super out of date.

tobiasgrosser (approved):
Maybe copy this info into the commit message. Then its good to go from my side.

PR #3012, opened 2024-08-10
CI: include onnx interpreter for pyright

superlopuh:
ONNX used to not have type annotations but now does.


tobiasgrosser (approved):
Nice.

PR #3017, opened 2024-08-12
transformations: (lift-arith-to-linalg) Add support for partially bufferized programs

n-io:
This PR improves the selection of suitable `outs` arguments for the created linalg ops (see [here](https://github.com/llvm/llvm-project/issues/101709)).

Partially bufferized programs have `bufferization.to_tensor` ops. These can be flagged as `writable`, and the pass will look for one with a matching size. This may in fact be after an additional `tensor.extract_slice`.

This mechanism should only be triggered for the translation of the first op in a chained calculation, as subsequent ops will re-use the `outs` of their operands.

### DIFF
--- a/tests/filecheck/transforms/lift-arith-to-linalg.mlir
+++ b/tests/filecheck/transforms/lift-arith-to-linalg.mlir
@@ -1,15 +1,64 @@
 // RUN: xdsl-opt %s -p lift-arith-to-linalg | filecheck %s
 
 builtin.module {
+// CHECK: builtin.module {
+
+  func.func @a() {
     %t0, %t1, %t2, %t3 = "test.op"() : () -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, tensor<8xf32>)
     %0 = arith.addf %t0, %t1 : tensor<8xf32>
     %1 = arith.subf %0, %t2 : tensor<8xf32>
     %2 = arith.mulf %1, %t3 : tensor<8xf32>
-}
+    func.return
+  }
 
-// CHECK-NEXT: builtin.module {
+// CHECK-NEXT: func.func @a() {
 // CHECK-NEXT:   %t0, %t1, %t2, %t3 = "test.op"() : () -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, tensor<8xf32>)
 // CHECK-NEXT:   %0 = linalg.add ins(%t0, %t1 : tensor<8xf32>, tensor<8xf32>) outs(%t0 : tensor<8xf32>) -> tensor<8xf32>
 // CHECK-NEXT:   %1 = linalg.sub ins(%0, %t2 : tensor<8xf32>, tensor<8xf32>) outs(%0 : tensor<8xf32>) -> tensor<8xf32>
 // CHECK-NEXT:   %2 = linalg.mul ins(%1, %t3 : tensor<8xf32>, tensor<8xf32>) outs(%1 : tensor<8xf32>) -> tensor<8xf32>
+// CHECK-NEXT:   func.return
+// CHECK-NEXT: }
+
+
+  func.func @choose_correct_tensor() {
+    %t0, %t1, %t2, %m0 = "test.op"() : () -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, memref<8xf32>)
+    %not_this = bufferization.to_tensor %m0 restrict : memref<8xf32>
+    %this = bufferization.to_tensor %m0 restrict writable : memref<8xf32>
+    %also_not_this = bufferization.to_tensor %m0 restrict : memref<8xf32>
+    %3 = arith.addf %t0, %t1 : tensor<8xf32>
+    %4 = arith.mulf %3, %t2 : tensor<8xf32>
+    func.return
+  }
+
+// CHECK-NEXT: func.func @choose_correct_tensor() {
+// CHECK-NEXT:   %t0, %t1, %t2, %m0 = "test.op"() : () -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, memref<8xf32>)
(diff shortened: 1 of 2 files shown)


tests/filecheck/transforms/lift-arith-to-linalg.mlir
```diff
@@ -1,15 +1,64 @@
...
 // CHECK-NEXT:   %2 = linalg.mul ins(%1, %t3 : tensor<8xf32>, tensor<8xf32>) outs(%1 : tensor<8xf32>) -> tensor<8xf32>
+// CHECK-NEXT:   func.return
+// CHECK-NEXT: }
+
+
+  func.func @choose_correct_tensor() {
+    %t0, %t1, %t2, %m0 = "test.op"() : () -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, memref<8xf32>)
+    %not_this = bufferization.to_tensor %m0 restrict : memref<8xf32>
+    %this = bufferization.to_tensor %m0 restrict writable : memref<8xf32>
+    %also_not_this = bufferization.to_tensor %m0 restrict : memref<8xf32>
```
superlopuh, line 27: These seem not to be used?

n-io: Well, `%this` is used, but what isn't used is `%not_this` and `%also_not_this` 🥹

n-io: The outer two shouldn't be found by the discovery function because they're not exposing `writable` tensors.

superlopuh: I see, I didn't spot it. It would probably be good to add the explanation in a comment somewhere, but that's more of a nit.

superlopuh: As a meta point, I do feel that this might be another case of doing two things in one pass. Why not always create a tensor.empty, and then infer whether it should be reused in another pass? Also, isn't bufferization already supposed to do this sort of analysis?

n-io: Thanks, added a comment as suggested.

I also thought that bufferization is supposed to do this analysis, but to keep the pass simple bufferization treats every `outs` as an init tensor, and with that as something having more semantic meaning than "shape-only" meaning of non-init-tensors. There is an `isInitTensor` function that I can't see being called in bufferization if I'm not mistaken, and there may be a simplifying assumption that treating everything as an init tensor means bufferization does not need to do such analysis, thus keeping it simpler. As any fix to upstream would take a little while as well as an mlir version update in xdsl, I'd recommend addressing it on our side.

Finally, the question is where to address it. Personally, I can easily fix it elsewhere, but I do like the idea of setting the args on translated ops in the best way we can.

n-io:
whoops sorry

tests/filecheck/transforms/lift-arith-to-linalg.mlir
```diff
@@ -1,15 +1,66 @@
...
+
+// CHECK-NEXT: func.func @choose_correct_tensor() {
+// CHECK-NEXT:   %t0, %t1, %t2, %m0 = "test.op"() : () -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, memref<8xf32>)
+// CHECK-NEXT:   %0 = bufferization.to_tensor %m0 restrict : memref<8xf32>
+// CHECK-NEXT:   %1 = bufferization.to_tensor %m0 restrict writable : memref<8xf32>
+// CHECK-NEXT:   %2 = bufferization.to_tensor %m0 restrict : memref<8xf32>
+// CHECK-NEXT:   %3 = linalg.add ins(%t0, %t1 : tensor<8xf32>, tensor<8xf32>) outs(%1 : tensor<8xf32>) -> tensor<8xf32>
+// CHECK-NEXT:   %4 = linalg.mul ins(%3, %t2 : tensor<8xf32>, tensor<8xf32>) outs(%3 : tensor<8xf32>) -> tensor<8xf32>
+// CHECK-NEXT:   func.return
+// CHECK-NEXT: }
```
superlopuh, line 42: This example worries me, where do you check that `%2` is not used after `%1` has been written to?

superlopuh: I'm sorry to be annoying, but I can't shake the feeling that this is just the wrong place to add the bufferization logic. Conceptually, arith operations create a new tensor, they're not in dps, so it makes sense for them to alloc a new tensor. A pass that helps bufferization and tries to look for a memref to reuse should be separated and have these kinds of corner cases checked separately from the arith and linalg dialects

n-io: It looks for a tensor to re-use rather than a memref.

n-io: > This example worries me, where do you check that `%2` is not used after `%1` has been written to?

This is actually why I included it in the filecheck.

superlopuh: I see that in this example it's not used, but I'm missing the bit of the code that examines the number of uses of the tensor in the pass. I agree that eliminating empty tensors is important, all I'm saying is that this custom logic would be better placed in a different pass that would eliminate them whether they were produced by lifting arith or any other method.

n-io: It sounds like what you are proposing might be a canonicalisation pass, but this would be a canonicalisation pass that would have to be done on linalg and would differ from upstream and may also have wider impact than we want. I do not think of allocating empty tensors here as a viable option tbh.

superlopuh: I don't see why this would be a canonicalization pass, it could just be its own pass with a single rewrite that implements exactly the logic added here, but on any empty tensor in the code. That way it would not be coupled to either arith or linalg, and could be tested separately.

PR #3020, opened 2024-08-12
dialects: (stablehlo) Add initial contribution for stablehlo dialect

efferifick:
This PR adds the `abs` operation for the stablehlo dialect. More to come.

### DIFF
--- a/tests/filecheck/dialects/stablehlo/test.mlir
+++ b/tests/filecheck/dialects/stablehlo/test.mlir
@@ -1,9 +1,8 @@
 // RUN: xdsl-opt %s --parsing-diagnostics --verify-diagnostics --split-input-file | filecheck %s
 
 builtin.module {
-  func.func @different_shapes(%0: tensor<i32>) {
-    // CHECK: "stablehlo.abs"(%{{.*}}) : (tensor<i32>) -> tensor<i32>
-    %result = "stablehlo.abs"(%0) : (tensor<i32>) -> tensor<i32>
-    func.return
-  }
+  %t0 = "test.op"() : () -> tensor<i32>
+
+  // CHECK: "stablehlo.abs"(%{{.*}}) : (tensor<i32>) -> tensor<i32>
+  %r0 = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
 }
--- a/tests/filecheck/dialects/stablehlo/test.mlir
+++ b/tests/filecheck/dialects/stablehlo/test.mlir
@@ -1,0 +1,9 @@
+// RUN: xdsl-opt %s --parsing-diagnostics --verify-diagnostics --split-input-file | filecheck %s
+
+builtin.module {
+  func.func @different_shapes(%0: tensor<i32>) {
+    // CHECK: "stablehlo.abs"(%{{.*}}) : (tensor<i32>) -> tensor<i32>
+    %result = "stablehlo.abs"(%0) : (tensor<i32>) -> tensor<i32>
+    func.return
+  }
+}
--- a/xdsl/dialects/stablehlo.py
+++ b/xdsl/dialects/stablehlo.py
@@ -35,7 +35,7 @@ class AbsOp(IRDLOperation):
     name = "stablehlo.abs"
 
     # TODO: Remove this constraint for complex types.
-    T = Annotated[TensorType, ConstraintVar("T")]
+    T = Annotated[AnyTensorType, ConstraintVar("T")]
 
     operand = operand_def(T)
     result = result_def(T)
--- a/xdsl/dialects/stablehlo.py
+++ b/xdsl/dialects/stablehlo.py
@@ -8,7 +8,7 @@ ML frameworks that produce StableHLO programs are compatible with ML compilers t
 
 from typing import Annotated
 
-from xdsl.dialects.builtin import TensorType
+from xdsl.dialects.builtin import AnyTensorType
 from xdsl.ir import Attribute, Dialect, SSAValue
 from xdsl.irdl import (
     ConstraintVar,
--- a/xdsl/dialects/stablehlo.py
+++ b/xdsl/dialects/stablehlo.py
@@ -1,0 +1,56 @@
+"""
+https://github.com/openxla/stablehlo/blob/main/docs/spec.md
+
+StableHLO is an operation set for high-level operations (HLO) in machine learning (ML) models.
+StableHLO works as a portability layer between different ML frameworks and ML compilers:
+ML frameworks that produce StableHLO programs are compatible with ML compilers that consume StableHLO programs.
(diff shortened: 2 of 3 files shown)


xdsl/dialects/stablehlo.py
```diff
@@ -0,0 +1,41 @@
+from xdsl.dialects.builtin import TensorType
+from xdsl.ir import Attribute, Dialect, SSAValue
+from xdsl.irdl import IRDLOperation, irdl_op_definition, operand_def, result_def
+from xdsl.utils.exceptions import VerifyException
+
+
+@irdl_op_definition
+class AbsOp(IRDLOperation):
```
superlopuh, line 8: Could you please add a docstring to this class, and to the top of this file? Ideally with links to the documentation for the dialect/op, and the overall purpose of the dialect.

xdsl/dialects/stablehlo.py
```diff
@@ -0,0 +1,41 @@
...
+    name = "stablehlo.abs"
+    operand = operand_def(TensorType)
+    result = result_def(TensorType)
```
superlopuh, line 11: xdsl uses IRDL Constraints to do things like verifying that some types are the same. There are a few examples of this around, in this case it would look something like this:
```suggestion
    T = Annotated[TensorType, ConstraintVar("T")]

    operand = operand_def(T)
    result = result_def(T)
```
The mismatch in either the shape or the element will be checked automatically. One thing I'm not sure about is whether the operation really requires the element type and the shape to be the same, or whether there is any broadcasting or dynamic shapes possible? A link to the doc would help disambiguate what level of abstraction this op is in.

xdsl/dialects/stablehlo.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, operand: SSAValue, result: Attribute):
+        super().__init__(operands=(operand,), result_types=(result,))
```
superlopuh, line 14: We usually have the result_type be optional when it can be inferred from the operand:
```suggestion
    def __init__(self, operand: SSAValue, result_type: Attribute | None):
        if result_type is None:
            result_type = operand.type
        super().__init__(operands=(operand,), result_types=(result_type,))
```

xdsl/dialects/stablehlo.py
efferifick: The [specification for this op](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#abs) states the following:

> Constraints
>
>    (C1) shape(result) = shape(operand).
>    (C2) baseline_element_type(result) is defined as:
>        complex_element_type(element_type(operand)) if is_complex(operand).
>        baseline_element_type(operand) otherwise.

I'll add a link to the documentation

tests/dialects/test_stablehlo.py
superlopuh, line 1: We rely on `lit` tests whenever possible, in this case I'd recommend migrating these tests to something similar to these:

https://github.com/xdslproject/xdsl/blob/main/tests/filecheck/dialects/memref/memref_ops.mlir
https://github.com/xdslproject/xdsl/blob/main/tests/filecheck/dialects/memref/invalid_ops.mlir

One filecheck tests that the operations can be printed and parsed correctly (and that they are hooked into `xdsl-opt`, which they are currently not in this PR), and that the verification errors raised are what we expect. Could you please do something similar in this PR?

xdsl/dialects/stablehlo.py
superlopuh: OK I think we're good with my suggestion for now, I'm not sure exactly what they mean with the complex stuff, but it sounds like we can handle it in the future

efferifick: It is just for complex<{f32,f64, ....}>

xdsl/dialects/stablehlo.py
efferifick: Just a note, I would prefer to keep the names as close as possible to the stablehlo specification. In the definition of `absOp`, `result` is used as opposed to `result_type`. So, if possible, I would rather keep the name.

xdsl/dialects/stablehlo.py
superlopuh: Yeah but the actual pairs of types aren't super clear to me. Let's cross that bridge when we get to it, though, I don't think it's very likely to come up soon.

efferifick: Ah sorry, I have been just doing a lot of this without thinking, but `abs` of a complex type just computes the norm.

efferifick: Maybe I'll skip the constraint for now, since for complex values it is different.

tests/dialects/test_stablehlo.py
efferifick: Do you want me to register the stablehlo dialect to `xdsl-opt`? It seems a bit premature, but happy to do so.

tests/filecheck/dialects/stablehlo/test.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: xdsl-opt %s --parsing-diagnostics --verify-diagnostics --split-input-file | filecheck %s
+
+builtin.module {
+  func.func @different_shapes(%0: tensor<i32>) {
+    // CHECK: "stablehlo.abs"(%{{.*}}) : (tensor<i32>) -> tensor<i32>
+    %result = "stablehlo.abs"(%0) : (tensor<i32>) -> tensor<i32>
+    func.return
+  }
```
superlopuh, line 8: Just as a little nit that will make life easier later, we try to make our tests as specific as possible, and for that we use the `test` dialect, which lets us just conjure up ssa values:
```suggestion
  %t0 = "test.op"() : () -> tensor<i32>

  // CHECK: "stablehlo.abs"(%{{.*}}) : (tensor<i32>) -> tensor<i32>
  %r0 = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
```

xdsl/dialects/stablehlo.py
```diff
@@ -0,0 +1,56 @@
...
+    * For signed integers: integer modulus.
+    * For floats: abs from IEEE-754.
+    * For complex numbers: complex modulus.
+    * For quantized types: dequantize_op_quantize(abs, operand, type(result)).
+    """
+
+    name = "stablehlo.abs"
+
+    # TODO: Remove this constraint for complex types.
+    T = Annotated[TensorType, ConstraintVar("T")]
```
superlopuh, line 38: Ah right, now it's complaining about the missing element type, I'd recommend this:
```suggestion
    T = Annotated[AnyTensorType, ConstraintVar("T")]
```

superlopuh (approved):
Beautiful, thank you!

superlopuh (approved):

PR #3023, opened 2024-08-13
CI: use ruff to format xdsl

superlopuh:
Runs in `0.04s` on my computer.

The Ruff format is designed to be a drop-in replacement for black. As you see, most of the diff is adding a new line at the start of every class, which seems like a good change. I'll comment on the config files with changes for easier reviewing.

Note stacked PR.


.pre-commit-config.yaml
superlopuh, line 1: Config change 1/3

pyproject.toml
superlopuh, line 1: Config change 2/3

xdsl/utils/comparisons.py
```diff
@@ -1,35 +1,35 @@
...
+
+The main idea is to not have the signedness be a property of the type of the value,
+and rather be a property of the operation. That means that a signless value can be
+interpreted as either a signed or unsigned value at runtime, depending on the
+operation that acts on it.
+
+During interpretation, this gets a little tricky, as the same bit pattern can be
+interpreted as two runtime values, meaning that comparing signless values is a little
+involved. For example, a signless value of 5 is equal to a signless value of -3, since
+their bit representations are the same.
```
PapyChacal, line 32: :heart:

PapyChacal (approved):
Nice, just a comment on the underlying one about lines length:
- #3022

PR #3024, opened 2024-08-13
core: Implement simple helpers for construct type access.

PapyChacal:
Just provide `result_types` and `operands_types` on `Operation` and `args_types` on `Block` to get rid of a lot of (IMO) annoying comprehensions.

IRDLOperation hopefully following!


xdsl/dialects/gpu.py
```diff
@@ -245,7 +245,7 @@ def verify_(self) -> None:
                 f"{self.operand.type}. They must be the same type for gpu.all_reduce"
             )
 
-        non_empty_body = any(b.ops for b in self.body.blocks)
+        non_empty_body = len(self.body.blocks) > 0
```
superlopuh, line 248:
```suggestion
        non_empty_body = bool(self.body.blocks)
```

xdsl/ir/core.py
```diff
@@ -614,6 +614,14 @@ class Operation(IRNode):
     def parent_node(self) -> IRNode | None:
         return self.parent
 
+    @property
+    def results_types(self) -> Sequence[Attribute]:
+        return tuple(r.type for r in self.results)
+
+    @property
+    def operands_types(self) -> Sequence[Attribute]:
+        return tuple(operand.type for operand in self.operands)
```
superlopuh, line 623: is this following a convention? I would expect `result_types` and `operand_types`, even though the original names are plurals. This follows `result_types` as the argument name to `build`, and `arg_types` in the Block init.

PapyChacal: Can do!

PapyChacal: Done!

PR #3025, opened 2024-08-13
core: helpers for IRDL construct type access.

PapyChacal:
Implement simple type access helpers, but for IRDL


xdsl/irdl/operations.py
```diff
@@ -341,7 +344,10 @@ def __init__(
         self.constr = range_constr_coercion(attr)
 
 
-VarOpResult: TypeAlias = tuple[OpResult, ...]
+class VarOpResult(tuple[OpResult, ...]):
+    @property
+    def types(self):
+        return tuple(r.type for r in self)
```
n-io, line 350: this is great if this doesn't need an `or [[]]`

PapyChacal: This wasn't actually needed where you used it :slightly_smiling_face: Happy to discuss it in chat if you're curious

PR #3026, opened 2024-08-13
[RFC] rewriting: Implement operation update methods

PapyChacal:
With @n-io (and potentially others?) we write a lot of patterns ultimately boiling down to "Just update this field on this op". Now, it's best practice to actually replace the op right, for patterns recursion and worklist management, if only that.

So it yields a lot of "Let me create this op of this type, from that op of this same type", re-expanding all fields to populate a constructor, and replace the old with the new. Also some error-prone (or just bad performance prone) stuff like region management, like region cloning or inlining, where they can just be carried over.

So I think it would make sense to have some API on the rewrite side to take care of this common pattern!

### DIFF
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import inspect
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Iterable, Mapping, Sequence
+from collections.abc import Callable, Iterable, Sequence
+from contextlib import contextmanager
 from dataclasses import dataclass, field
 from functools import wraps
 from types import UnionType
@@ -89,54 +90,17 @@ class PatternRewriter(PatternRewriterListener):
     has_done_action: bool = field(default=False, init=False)
     """Has the rewriter done any action during the current match."""
 
-    def update_op(
-        self,
-        op: Operation,
-        new_results: Sequence[SSAValue | None] | None = None,
-        safe_erase: bool = True,
-        *,
-        operands: Sequence[SSAValue] | None = None,
-        result_types: Sequence[Attribute] | None = None,
-        properties: Mapping[str, Attribute] | None = None,
-        attributes: Mapping[str, Attribute] | None = None,
-        successors: Sequence[Block] | None = None,
-        regions: Sequence[Region] | None = None,
-    ):
-        self.has_done_action = Rewriter.update_op(
-            op,
-            new_results,
-            safe_erase,
-            operands=operands,
-            result_types=result_types,
-            properties=properties,
-            attributes=attributes,
-            successors=successors,
-            regions=regions,
-        )
-
-    def update_matched_op(
-        self,
-        new_results: Sequence[SSAValue | None] | None = None,
-        safe_erase: bool = True,
-        *,
-        operands: Sequence[SSAValue] | None = None,
-        result_types: Sequence[Attribute] | None = None,
-        properties: Mapping[str, Attribute] | None = None,
-        attributes: Mapping[str, Attribute] | None = None,
-        successors: Sequence[Block] | None = None,
-        regions: Sequence[Region] | None = None,
-    ):
-        self.update_op(
-            self.current_operation,
-            new_results,
-            safe_erase,
-            operands=operands,
-            result_types=result_types,
-            properties=properties,
-            attributes=attributes,
-            successors=successors,
-            regions=regions,
-        )
+    @contextmanager
+    def modify_op_in_place(self, op: Operation):
+        """
+        Modify an operation in place.
+        This is useful when the operation is not replaced, but modified.
+        """
+        try:
+            yield
+        finally:
+            self.has_done_action = True
+            self.handle_operation_modification(op)
 
     def insert_op(
         self, op: Operation | Sequence[Operation], insertion_point: InsertPoint
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import inspect
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Iterable, Sequence
+from collections.abc import Callable, Iterable, Mapping, Sequence
 from dataclasses import dataclass, field
 from functools import wraps
 from types import UnionType
@@ -88,6 +88,55 @@ class PatternRewriter(PatternRewriterListener):
 
     has_done_action: bool = field(default=False, init=False)
     """Has the rewriter done any action during the current match."""
+
+    def update_op(
+        self,
+        op: Operation,
+        new_results: Sequence[SSAValue | None] | None = None,
+        safe_erase: bool = True,
+        *,
+        operands: Sequence[SSAValue] | None = None,
+        result_types: Sequence[Attribute] | None = None,
+        properties: Mapping[str, Attribute] | None = None,
+        attributes: Mapping[str, Attribute] | None = None,
+        successors: Sequence[Block] | None = None,
+        regions: Sequence[Region] | None = None,
+    ):
+        self.has_done_action = Rewriter.update_op(
+            op,
+            new_results,
+            safe_erase,
+            operands=operands,
+            result_types=result_types,
+            properties=properties,
+            attributes=attributes,
+            successors=successors,
+            regions=regions,
+        )
+
+    def update_matched_op(
+        self,
+        new_results: Sequence[SSAValue | None] | None = None,
+        safe_erase: bool = True,
+        *,
+        operands: Sequence[SSAValue] | None = None,
+        result_types: Sequence[Attribute] | None = None,
+        properties: Mapping[str, Attribute] | None = None,
+        attributes: Mapping[str, Attribute] | None = None,
+        successors: Sequence[Block] | None = None,
+        regions: Sequence[Region] | None = None,
+    ):
+        self.update_op(
+            self.current_operation,
+            new_results,
+            safe_erase,
+            operands=operands,
+            result_types=result_types,
+            properties=properties,
+            attributes=attributes,
+            successors=successors,
+            regions=regions,
+        )
 
     def insert_op(
         self, op: Operation | Sequence[Operation], insertion_point: InsertPoint
--- a/xdsl/rewriter.py
+++ b/xdsl/rewriter.py
@@ -132,6 +132,13 @@ class Rewriter:
         successors: Sequence[Block] | None = None,
         regions: Sequence[Region] | None = None,
     ):
+        """
+        Recreate an operation with any passed field updated. All other fields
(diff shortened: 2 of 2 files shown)


xdsl/pattern_rewriter.py
```diff
@@ -89,6 +89,55 @@ class PatternRewriter(PatternRewriterListener):
     has_done_action: bool = field(default=False, init=False)
     """Has the rewriter done any action during the current match."""
 
+    def update_op(
+        self,
+        op: Operation,
+        new_results: Sequence[SSAValue | None] | None = None,
```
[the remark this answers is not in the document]
math-fehr: Yeah agreed, I don't think you can make this work like it.

xdsl/pattern_rewriter.py
```diff
@@ -89,6 +89,55 @@ class PatternRewriter(PatternRewriterListener):
...
+        safe_erase: bool = True,
+        *,
+        operands: Sequence[SSAValue] | None = None,
+        result_types: Sequence[Attribute] | None = None,
+        properties: Mapping[str, Attribute] | None = None,
+        attributes: Mapping[str, Attribute] | None = None,
+        successors: Sequence[Block] | None = None,
+        regions: Sequence[Region] | None = None,
+    ):
+        self.has_done_action = Rewriter.update_op(
```
math-fehr, line 105: You probably also need to add the operation to the worklist if you rewrite recursively

xdsl/rewriter.py
```diff
@@ -119,6 +119,52 @@ def replace_op(
...
+        new_op = op.create(
+            operands=op.operands if operands is None else operands,
+            result_types=op.result_types if result_types is None else result_types,
+            # No need to copy the properties, attributes, regions: we replace the
+            # original op so it won't use them anymore!
+            properties=op.properties if properties is None else properties,
+            attributes=op.attributes if attributes is None else attributes,
+            successors=op.successors if successors is None else successors,
+            regions=(
+                tuple(op.detach_region(0) for _ in op.regions)
```
math-fehr, line 160:
```suggestion
                tuple(region.detach() for region in list(op.regions))
```
Makes it more clear to me?

math-fehr:
I'm not sure how I feel about it.
I feel that `update` is a bit of an anti-pattern, because we are losing all the nice utilities that IRDL can offer.
For instance, writing `my_op.my_attr = ...` feels much better than `my_op.update(my_op.attr_dict & {"my_attr" = new_attr})`, and does type checking as well.
Also, I'm not sure how this will work when you change the number of results?

Just to add some background, the way this works in MLIR is this:
```python
rewriter.modifyOpInPlace(my_op, lambda op: ...)
```
or this:
```python
rewriter.startOpModification(my_op)
# Do the modification
rewriter.stopOpModification(my_op)
```

I would personally prefer a bit the MLIR version, but I'm wondering if there is a pattern that often arise in your case that made you design this current version?

math-fehr:
Btw, since we are in python:
```python
with rewriter.modifyOpInPlace(op):
   # Do modifications
```
Which is way more elegant than the versions I've written from MLIR

PapyChacal:
> Btw, since we are in python:
> 
> ```python
> with rewriter.modifyOpInPlace(op):
>    # Do modifications
> ```
> 
> Which is way more elegant than the versions I've written from MLIR

My first thought reading your comment :grin: 

I like it, and if it's good enough, maybe it's the best call :thinking: 

> I would personally prefer a bit the MLIR version, but I'm wondering if there is a pattern that often arise in your case that made you design this current version?

The first thing coming to mind is the checking if we should actually checking the op or not, to prevent infinite recursion.
I kinda like having infrastructure check itself if it actually did some work or not.
I'm guessing this could be managed by the very same contextmanager though!

math-fehr:
I think this is up to the user to do these checks.
If they enable the `recursive` option in the pattern rewriter, then they need to only do the modifications (same with rewrites) if that change something.
Currently, they would replace an op with the same op, and that will recurse as well

superlopuh:
Yeah I think I also prefer Mathieu's suggestion

PapyChacal:
Interesting, I think it would be pretty cheap to do most recursion checks for the user! If it was transparent, would you still prefer not to for performance - or other - concerns ?

math-fehr:
I would be fine if we did the checks, though I don't think it's *that* cheap?
We would need to check for equality between dictionary attributes

math-fehr:
And operands, properties, results. But I'm not sure how we could check results equality though?

PapyChacal:
So if it doesn't do that, what would
```python3
with rewriter.modifyOpInPlace(op):
   # Do modifications
```
actually mean? Isn't it just `rewriter.has_done_action = True` at this point?

math-fehr:
Yep, same thing when you replace an operation with literally the same operation.
If you don't want to do the modification, you could do:
```python
if condition:
  with rewriter.modifyOpInPlace(op):
    # Do modifications
```

PapyChacal:
Understood, but if reduced to that, I don't really get the point of using a `ContextManager`? :thinking:

math-fehr:
Oh, it's just because I think that a with statement looks much better than `rewriter.startOpModification()` and `rewriter.stopOpModification()`. Because then you don't forget the other.
The Context manager would just call the two functions, nothing else

n-io:
The mechanism I think would be most helpful would be more of a modify-or-rebuild-in-place, along the lines of:

* modify operands, properties, or attributes without rebuilding the op, and trigger rewrite handlers
* modify result_types, rebuild op in place, and trigger rewrite handlers

A `with` block definitely sounds pythonically pretty, but also has implications that are worth considering. For instance, it'd be fun to do:

```
new_result_types: Sequence[Attribute] = ...
with rewriter.modify_op_in_place(op) as mod:
    mod.result_types = new_result_types
```

This particular case needs to trigger a rebuild, although it is not clear whether this should best be done immediately or at the end of the block. (In my attempts, the above example uses a wrapper which forwards all `__setattr__` except for `result_types`, but I ran into the issue that intersection types don't currently exist).

TL;DR
If we don't want `update_op` as proposed by @PapyChacal, then I'd advocate for having two things:
* a `modify_op_in_place` yielding a `with` block for setting properties that do not require a rebuild, and always triggers handlers
* a `rebuild_op_in_place(op, operands, result_types, props, attr, regions)`, and triggers handlers

xdsl/rewriter.py
```diff
@@ -119,6 +119,52 @@ def replace_op(
...
+        new_results: Sequence[SSAValue | None] | None = None,
+        safe_erase: bool = True,
+        *,
+        operands: Sequence[SSAValue] | None = None,
+        result_types: Sequence[Attribute] | None = None,
+        properties: Mapping[str, Attribute] | None = None,
+        attributes: Mapping[str, Attribute] | None = None,
+        successors: Sequence[Block] | None = None,
+        regions: Sequence[Region] | None = None,
+    ):
```
n-io, line 134:
```suggestion
    @staticmethod
    def rebuild_op_in_place(
        op: Operation,
        operands: Sequence[SSAValue] | None = None,
        result_types: Sequence[Attribute] | None = None,
        properties: Mapping[str, Attribute] | None = None,
        attributes: Mapping[str, Attribute] | None = None,
        successors: Sequence[Block] | None = None,
        regions: Sequence[Region] | None = None,
    ):
```

suggested method name and signature

PapyChacal:
> TL;DR I'd advocate for having two things:
> 
>     * a `modify_op_in_place` yielding a `with` block for setting properties that do not require a rebuild, and always triggers handlers
> 
>     * a `rebuild_op_in_place(op, operands, result_types, props, attr, regions)`, and triggers handlers

* I believe this is what was agreed upon here, this is what I'll steer this specific PR to now

* I believe this is what @math-fehr and @superlopuh didn't like here, and what I initially thought about. Let's discuss that in another PR/issue then?

superlopuh:
The main reason I don't like this is that it goes against all the work we've been doing to try to leverage Python's type system to make it easier to construct operations and attributes. This mechanism completely ignores the IRDL stuff, updating the raw backing things in the operation, which could potentially be invalid, and you wouldn't know until runtime. For example, the updated operands might need a different segment size attribute, as the old one's attributes might just leave it as incorrect. At another level, this also bypasses the pattern rewriter's worklist, meaning that we drop the tracking of the operations to re-process after they are rewritten. I have a gut feeling that more things will break when this sort of in-place mutability is introduced.

n-io:
Prob without having a full picture of the IRDL side-of-things, if we were aiming for more usable and safe(r) setter methods in order to avoid expensive rebuilds, something like this might do the trick:

```
# in PatternRewriter
@contextmanager
def modify_op_in_place(op: Operation) -> Generator[None, None, None]
try:
    yield
finally:
    self.has_done_action = True
```

What I'd personally like, however, is not actually the above and more just syntactic sugar for:
```
rewriter.replace_matched_op(
    func.FuncOp.build(
        operands=op.operands,
        result_types=op.result_types,
        regions=[op.detach_region(op.body)],
        properties={**op.properties, "function_type": function_type},
        attributes=op.attributes.copy(),
    )
)

# could be simplified to:
rewriter.rebuild_matched_op_in_place(
    properties={**op.properties, "function_type": function_type},
)
```
Avoids having to specify the type (esp. when dealing with type hierarchies), and makes rebuilding with new result types shorter. The question is if sugar is good for the system..

PapyChacal:
> The main reason I don't like this is that it goes against all the work we've been doing to try to leverage Python's type system to make it easier to construct operations and attributes. This mechanism completely ignores the IRDL stuff, updating the raw backing things in the operation, which could potentially be invalid, and you wouldn't know until runtime. For example, the updated operands might need a different segment size attribute, as the old one's attributes might just leave it as incorrect.

Agreed, this PR offered this at the Operation level, to discuss the rewriting side of things first, before hooking to IRDL for `IRDLOperation`, as I did for type accessors. IRDL level stuff is just always more technically involved so I like to start by the low-level aspect, that's all. So if only that concern remain, we could discuss it in a two-staged manner.

> At another level, this also bypasses the pattern rewriter's worklist, meaning that we drop the tracking of the operations to re-process after they are rewritten. I have a gut feeling that more things will break when this sort of in-place mutability is introduced.

That I'm completely confused by; The build-and-replace approach is precisely thought to minimize mutability usage? (as-in, just replace the op instead of mutating it)
The tracking of the operation missing in the PR is a typo and was the core point; this API was for the rewritepattern to handle those and the user only having to provide the may-be-changed fields of the operation.

Either way, I heard in this discussion that you were both more interested in actually mutating operations in-place with infrastructure support than I thought, which is good enough for my purposes, so I'll steer this PR to that with the contextmanager now; I don't have enough time to open and consider varied ideas anyway, and will leave that to other participants :slightly_smiling_face: I believe those are important points though :wink:

superlopuh:
Let's close this for the time being?

superlopuh:
Closing for now, let's reopen when there's demand for this again

PR #3027, opened 2024-08-13
transformations: Loop Invariant Code Motion

RRavikiran66:
This pull request initiates the development of loop invariant code motion (LICM) optimizations. I have implemented helper functions like `canBeHoisted` , `move_out_of_the_region`, 'isMemoryeffectfree'. I am not sure about speculatable so for now it contains same body as 'isMemoryeffectfree'  I think it is working for scf.for, and I have added few tests cases to verify as well.


superlopuh:
Oh actually the documentation lies about the side effects check, that's done outside of the helper in CPP, we might as well follow their logic for now.

RRavikiran66:
> Oh actually the documentation lies about the side effects check, that's done outside of the helper in CPP, we might as well follow their logic for now.

I have included your changes in the new commit but I forgot rename the canBeHoisted I will do it in the next commit. I have implemented helper functions like  `move_out_of_the_region`, 'isMemoryeffectfree'. I am not sure about speculatable so for now it contains same body as 'isMemoryeffectfree'  I think it is working for scf.for, and I have added few tests cases to verify as well. I think code can be optimized, I will take that up in the next commit as well.

superlopuh:
Can you please fix all the linting errors? Running `make install-precommit` and `make precommit` should help, it's quite hard to review otherwise. Also, if it's ready for review, could you please remove the WIP label and hit the `Ready for review` button in GitHub?

superlopuh (commented):
This is great! I think the PR is ready to review, fantastic to see it working. Could you please mark it as such in the GitHub interface?

RRavikiran66:
I will create three PR for rest of the Helper Functions

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -0,0 +1,141 @@
...
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.traits import (
+    IsTerminator,
+    Pure,
+    RecursiveMemoryEffect,
+    # is_side_effect_free,
+    # only_has_effect,
```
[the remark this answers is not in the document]
superlopuh: Please let the person who commented resolve the conversation themselves in the PR. Why did you resolve my comment? We generally try to not have any commented code in the xDSL code base, so I would recommend deleting the above two lines.

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -0,0 +1,99 @@
...
+    # the loop body.
+    for child in op.walk():
+        for operand in child.operands:
+            for own in operand.owner.walk():
+                if not isinstance(own, scf.For):
+                    if op.is_ancestor(operand.owner):
+                        continue
+                    if region_target.is_ancestor(own):
+                        return False
+    return True
```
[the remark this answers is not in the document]
superlopuh: It may seem like more PRs will mean that merging it takes more time but it will almost certainly take less time overall to merge the changes.

xdsl/transforms/loop_invariant_code_motion.py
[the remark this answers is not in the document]
RRavikiran66: I resolved them, Cause I thought running precommit will remove unnecessary imports which I realized is not true for commented code, I didn't look for it either. I will be cautious of this next time

xdsl/transforms/loop_invariant_code_motion.py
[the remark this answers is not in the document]
RRavikiran66: I don't understand few things:
1)If I create a helper function, How can I test the helper function without calling in a class?
2)The test setup should also be a part of this PR right?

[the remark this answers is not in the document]
superlopuh: I would recommend adding a pytest test for the function specifically, it's saved me a few times when writing tricky code, as it's easier to pinpoint the error than with the filecheck. In this case, I would recommend constructing a loop nest manually, using a builder, and checking that all the branches are covered, in a file like `tests/transforms/test_loop_invariant_code_motion.py`.

superlopuh:
I had a chat with Ravikiran, and he's currently working towards a paper with a deadline on the 22 of November, will wait until then to update this.

superlopuh:
I spoke with @RRavikiran66, and he won't have time to get to this in the near future. I think it'll be quite helpful to have this for a few projects, so I'll take over this PR and get this merged. Thank you for getting this started @RRavikiran66!

superlopuh:
@math-fehr @alexarice thoughts? I mainly ported over MLIR's implementation, but made it scf.for-specific since we don't have LoopLikeInterface yet.

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,82 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
-            assert operand_owner is not None
             if op.is_ancestor(operand_owner):
                 continue
             if target_region.is_ancestor(operand_owner):
                 return False
     return True
+
+
+def _move_loop_invariant_code(
+    regions: Sequence[Region],
```
math-fehr, line 52: Is there any case where we cannot handle this region by region? I would just take a single region as argument here no?

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,82 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
+    should_move_out_of_region: Callable[[Operation, Region], bool],
+    move_out_of_region: Callable[[Operation, Region], None],
+):
```
math-fehr, line 55: I would also return `None` (I'm not sure why we return the number of operation moved), or mark explicitely that the return value is `int` (and document that)

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,82 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
                 continue
             if target_region.is_ancestor(operand_owner):
                 return False
     return True
+
+
+def _move_loop_invariant_code(
+    regions: Sequence[Region],
+    should_move_out_of_region: Callable[[Operation, Region], bool],
+    move_out_of_region: Callable[[Operation, Region], None],
```
math-fehr, line 54: Is there a reason we are passing callables here?
I would instead just use the functions we define later, no need to generalize everything that early on before people have any use for it?

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,82 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
+    """
+    Moves operations without side effects out of loops, provided they do not depend on
+    values defined the loops.
+    """
+
+    name = "licm"
+
+    def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            LoopInvariantCodeMotion(), walk_reverse=True, walk_regions_first=True
```
math-fehr, line 121: Maybe you should explicitely mark it as not recursive? Otherwise it will run multiple times, even though the second time nothing should change AFAIK

math-fehr (commented):
I added some nits comments, but otherwise it looks good!
Can you also add documentation on the different functions that you are adding?

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,82 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
+
+def _move_loop_invariant_code(
+    regions: Sequence[Region],
+    should_move_out_of_region: Callable[[Operation, Region], bool],
+    move_out_of_region: Callable[[Operation, Region], None],
+):
+    num_moved = 0
+    for region in regions:
+        # add top-level operations in the loop body to the worklist
+        worklist = [op for block in region.blocks for op in block.ops]
```
alexarice, line 59: Can we reuse `Worklist` from `pattern_rewriter` which I think does this more efficiently?

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: I followed MLIR API here as close as possible, but happy to inline the function definitions.

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: I think this is there for loop ops with multiple regions, maybe like a while loop?

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: Yeah I'll change to None

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: Will do

xdsl/transforms/loop_invariant_code_motion.py
math-fehr: What I mean is that in that case, we should probably write `for region in op.regions: _move_loop_invariant_code(region)`, which kind of makes the code simpler here I think

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: Done

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: I don't think that Worklist would be more efficient here, but I replaced this with a deque, which is probably better than a list.

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: Done

math-fehr (approved):

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,66 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
+    """
+
+    name = "licm"
+
+    def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            LoopInvariantCodeMotion(),
+            apply_recursively=False,
+            walk_regions_first=True,
+            walk_reverse=True,
```
alexarice, line 105: sorry if this already got answered but why do we `walk_reverse`?

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -32,9 +41,66 @@ def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
...
             if op.is_ancestor(operand_owner):
                 continue
             if target_region.is_ancestor(operand_owner):
                 return False
     return True
+
+
+def _move_loop_invariant_code(region: Region, builder: Builder):
+    # add top-level operations in the loop body to the worklist
+    worklist = deque(op for block in region.blocks for op in block.ops)
```
alexarice, line 53: I just thought the `Worklist` class might be better as it avoids duplicates as far as I know, but this probably doesn't matter much.

alexarice (approved):

xdsl/transforms/loop_invariant_code_motion.py
superlopuh: It feels like it might be the right thing to do but I couldn't come up with an example where it changes anything, so I'll leave it out.

PR #3030, opened 2024-08-14
dialects: (stablehlo) add elementwise add, sub, mul

superlopuh:
Also reorganizes the test.

CC @efferifick

### DIFF
--- a/tests/filecheck/dialects/stablehlo/test.mlir
+++ b/tests/filecheck/dialects/stablehlo/test.mlir
@@ -1,8 +1,0 @@
-// RUN: xdsl-opt %s --parsing-diagnostics --verify-diagnostics --split-input-file | filecheck %s
-
-builtin.module {
-  %t0 = "test.op"() : () -> tensor<i32>
-
-  // CHECK: "stablehlo.abs"(%{{.*}}) : (tensor<i32>) -> tensor<i32>
-  %r0 = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
-}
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/stablehlo/test.mlir
```diff
@@ -1,8 +0,0 @@
-// RUN: xdsl-opt %s --parsing-diagnostics --verify-diagnostics --split-input-file | filecheck %s
```
alexarice, line 1: Would be good it this could have been `git mv`ed but not necessary

xdsl/dialects/stablehlo.py
```diff
@@ -18,18 +19,41 @@
     result_def,
 )
 
+# region Abstract Base Classes
+
+
+class ElementwiseBinaryOperation(IRDLOperation, abc.ABC):
+    # TODO: Remove this constraint for complex types.
+    T = Annotated[AnyTensorType, ConstraintVar("T")]
+
```
alexarice, line 28: Should these have some verification that the inputs are "numeric"

alexarice (commented):
Do these operations not have assemblyFormat/custom parsing?

xdsl/dialects/stablehlo.py
superlopuh: Eventually we'll make these constrained to the values that the original stablehlo dialect supports, I don't think it's critical right now.

superlopuh:
> Do these operations not have assemblyFormat/custom parsing?

nope

alexarice (approved):

PR #3031, opened 2024-08-14
dialects: (stablehlo) add transpose

superlopuh:
Note stacked PR

### DIFF
--- a/tests/filecheck/dialects/stablehlo/ops.mlir
+++ b/tests/filecheck/dialects/stablehlo/ops.mlir
@@ -1,17 +1,28 @@
 // RUN: XDSL_ROUNDTRIP
 
-builtin.module {
-  %t0 = "test.op"() : () -> tensor<i32>
+%t0 = "test.op"() : () -> tensor<i32>
 
-  // CHECK: %abs = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
-  %abs = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
+// CHECK: %abs = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
+%abs = "stablehlo.abs"(%t0) : (tensor<i32>) -> tensor<i32>
 
-  // CHECK: %add = "stablehlo.add"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
-  %add = "stablehlo.add"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+// CHECK: %add = "stablehlo.add"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+%add = "stablehlo.add"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
 
-  // CHECK: %multiply = "stablehlo.multiply"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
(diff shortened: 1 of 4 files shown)


tests/filecheck/dialects/stablehlo/ops.mlir
```diff
@@ -1,17 +1,28 @@
...
+// CHECK: %multiply = "stablehlo.multiply"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+%multiply = "stablehlo.multiply"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
 
-  // CHECK: %subtract = "stablehlo.subtract"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
-  %subtract = "stablehlo.subtract"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
-}
+// CHECK: %subtract = "stablehlo.subtract"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+%subtract = "stablehlo.subtract"(%t0, %t0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+
+%transpose_operand = "test.op"() : () -> tensor<2x3x2xi32>
```
alexarice, line 17: This seems to be missing a `CHECK`?

xdsl/dialects/stablehlo.py
```diff
@@ -126,13 +128,69 @@ class SubtractOp(ElementwiseBinaryOperation):
...
+
+    def verify_(self) -> None:
+        # Operand and result types are checked before the custom `verify_`
+        o_type = cast(TensorType[Attribute], self.operand.type)
+        r_type = cast(TensorType[Attribute], self.result.type)
+
+        o_shape = o_type.get_shape()
+        r_shape = r_type.get_shape()
+
+        # TODO: C1 - quantization
```
alexarice, line 169: What do `C1` `C2` and `C3` mean here?

xdsl/dialects/stablehlo.py
```diff
@@ -126,13 +128,69 @@ class SubtractOp(ElementwiseBinaryOperation):
...
+    name = "stablehlo.transpose"
+
+    ElementType = Annotated[Attribute, ConstraintVar("ElementType")]
+
+    operand = operand_def(TensorType[ElementType])
+    result = result_def(TensorType[ElementType])
+    permutation = attr_def(DenseArrayBase)
+
+    def __init__(
+        self, operand: SSAValue, permutation: DenseArrayBase, result_type: Attribute
```
alexarice, line 150: Would it be worth also supporting inputting a permutation as a sequence of ints?

xdsl/dialects/stablehlo.py
superlopuh: They're the constraint names in the original stablehlo documentation

xdsl/dialects/stablehlo.py
superlopuh: Others may disagree, but I think that it's not a big overhead for the caller to just do this bit themselves (`DenseArrayBase.create_dense_int_or_index(i64, (2, 1, 0))`)

xdsl/dialects/stablehlo.py
alexarice: Would it be worth explaining this/linking in the comment?

superlopuh: Actually I just removed them, they're not even consistently numbered throughout the document

alexarice (approved):
Assuming the quantization constraints can't be done because we don't have this "is_per_axis_quantized" predicate? In any case LGTM

superlopuh:
More that I don't want to do them right now :)

PR #3032, opened 2024-08-14
misc: remove mentions of black formatter

superlopuh:
A bit of cleanup.

1. exclude ruff transition from git blame
2. remove `make black`
3. clean up comments
4. Makefile now also has 88 column width, because why not
5. Add jupyter notebooks and type stubs to ruff formatted file list

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,10 @@ TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 .ONESHELL:
 
 # these targets don't produce files:
-.PHONY: ${VENV_DIR}/ venv clean clean-caches filecheck pytest pytest-nb tests-toy tests rerun-notebooks precommit-install precommit black pyright tests-marimo tests-marimo-mlir
-.PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report-md
+.PHONY: ${VENV_DIR}/ venv clean clean-caches filecheck pytest pytest-nb tests-toy tests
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -127,8 +127,7 @@ To automate the formatting, we use pre-commit hooks from the
 make precommit-install
 # to run the hooks:
 make precommit
-# alternatively, running black on all staged files:
-make black  # or simply black $(git diff --staged --name-only)
+# Alternatively, run `ruff format`
```
alexarice, line 130: Should this have a command below it?

Makefile
```diff
@@ -105,12 +113,6 @@ pyright:
 	xdsl-stubgen
 	pyright $(shell git diff --staged --name-only  -- '*.py')
 
-# run black on all files currently staged
-black:
```
alexarice, line 109: Is it worth having a "make ruff" replacement for this or is it just unnecessary?

alexarice (approved):
Can't find any usages of the word black in the repository so looks good

Makefile
superlopuh: I initially changed this to `make format` but then I thought that was silly since it's the same number of characters and doesn't seem particularly useful.

README.md
superlopuh: I thought the inline command was enough, do you think it would be clearer below?

alexarice:
```suggestion
# Alternatively, run ruff directly:
ruff format
```

PR #3034, opened 2024-08-14
interpreter: use interpreter default bitwidth in run main

superlopuh:


xdsl/tools/xdsl_run.py
```diff
@@ -62,6 +62,7 @@ def register_all_arguments(self, arg_parser: argparse.ArgumentParser):
         arg_parser.add_argument(
             "--index-bitwidth",
             choices=(32, 64),
+            default=Interpreter.DEFAULT_BITWIDTH,
```
alexarice, line 65: This is defined as the system bitwidth anyway

alexarice: Not clear to me what happened before if this wasn't specified

superlopuh:
I'll look into this and try to determine the difference in behaviour

xdsl/tools/xdsl_run.py
superlopuh: I added a comment, the argument value is None by default, so it was assigning None explicitly in the init, which was propagated

alexarice (approved):

PR #3037, opened 2024-08-15
dependencies: add JAX as an optional dependency of xDSL

superlopuh:
There will be a number of projects using xDSL + JAX, might as well make this easy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dev = [
     "pyright==1.1.345",
 ]
 gui = ["textual==0.76.0", "pyclip==0.7"]
+jax = ["jax==0.4.31"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -34,6 +34,7 @@ dev = [
     "pyright==1.1.345",
 ]
 gui = ["textual==0.76.0", "pyclip==0.7"]
+jax = ["jax==0.4.31"]
```
superlopuh, line 37: We will be relying on unstable things in the API, hence the minor version pin.

webmiche (approved):

pyproject.toml
webmiche: That does not at all sound like an upcoming nightmare :smile:

PR #3039, opened 2024-08-15
dependencies: update ruff-pre-commit to v0.6.0

superlopuh:
bleep bloop


alexarice (commented):
I expect this comes as no surprise but this is a breaking change on nixos

alexarice (approved):
Manually updated ruff and this works fine, though would appreciate if the pr could be held for a few days until the ruff update reaches repositories (new version seems barely older than this pr?)

superlopuh:
I'll merge it next week

PapyChacal (approved):
bleep bleep bloop bleep

PR #3041, opened 2024-08-15
[RFC] core: Better support for splat in DenseIntOrFPElementsAttr

n-io:
When DenseIntOrFPElementsAttr consists of a single data element, it is currently replicated across the entire shape, which is inefficient for large shapes, and handled as a special case in upstream MLIR (the [docs](https://mlir.llvm.org/docs/Dialects/Builtin/#denseintorfpelementsattr) refer to this as a splat value).

This PR removes the replication from the parser and makes it lazy (when `data` is accessed). This is done by making the `data` field protected ( -> `_data`) and hiding it behind a property of the same name.

Ideally, we'd fix the places in the code that rely on this redundancy and remove the `@property data` again.

### DIFF
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -142,11 +142,10 @@ class CslPrintContext:
             case UnitAttr():
                 return ""
             case DenseIntOrFPElementsAttr():
-                data = init.data.data
                 assert (
-                    len(data) == 1
-                ), f"Memref global initialiser has to have 1 value, got {len(data)}"
-                return f" = @constants({type}, {self.attribute_value_to_str(data[0])})"
+                    data := init.get_single_element()
+                ), "Memref global initialiser has to have a single value"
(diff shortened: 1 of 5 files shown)


n-io:
The test breaking seems to highlight that different parts of the code deal with splat values in different ways.

superlopuh:
What's the motivation for this? Is it too slow?

n-io:
The primary motivation is to have a unified way of representing splat elements. See some of the changed code snippets.

superlopuh:
That doesn't seem like a motivation in my understanding, more of an implementation... Why make this change?

n-io:
I'm highlighting the problem that
```
data.count(data[0]) == len(data)
```
and
```
len(data) == 1
```
are used to check for to the same thing.

superlopuh:
I'm sorry, I don't understand, what wasn't working before that works with the proposed changes?

superlopuh:
@n-io, what's the latest on this? I'm still not sure I understand the motivation behind this PR, is it still relevant?

PR #3044, opened 2024-08-16
core: Improved treatment of whitespace in assembly format

alexarice:
This makes the following changes:
- Makes assembly format recognise the empty whitespace directive ``` `` ```
- Fixes a bug in the assembly format parsing so that this is actually parsed correctly
- Allows whitespace at the start of an optional group
- Modifies existing stencil tests now that they print correctly (i.e. without a space in `stencil.store`
- Adds tests for new functionality

These changes bring the assembly format closer to the assembly format of mlir

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -822,13 +822,13 @@ class OptionalUnitAttrVariable(OptionalAttributeVariable):
 class WhitespaceDirective(FormatDirective):
     """
     A whitespace directive, with the following format:
-      whitespace-directive ::= `\n` | ` `
+      whitespace-directive ::= `\n` | ` ` | ``
     This directive is only applied during printing, and has no effect during
     parsing.
     The directive will not request any space to be printed after.
     """
 
-    whitespace: Literal[" ", "\n"]
+    whitespace: Literal[" ", "\n", ""]
     """The whitespace that should be printed."""
 
     def parse(self, parser: Parser, state: ParsingState) -> None:
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -836,7 +836,7 @@ class WhitespaceDirective(FormatDirective):
 
     def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> None:
         printer.print(self.whitespace)
-        state.last_was_punctuation = len(self.whitespace) == 0
+        state.last_was_punctuation = self.whitespace == ""
         state.should_emit_space = False
 
 
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -908,6 +908,7 @@ class KeywordDirective(OptionallyParsableDirective):
 @dataclass(frozen=True)
 class OptionalGroupDirective(FormatDirective):
     anchor: AnchorableDirective
+    then_whitespace: tuple[WhitespaceDirective, ...]
     then_first: OptionallyParsableDirective
     then_elements: tuple[FormatDirective, ...]
 
@@ -944,5 +945,9 @@ class OptionalGroupDirective(FormatDirective):
 
     def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> None:
         if self.anchor.is_present(op):
-            for element in (self.then_first, *self.then_elements):
+            for element in (
+                *self.then_whitespace,
+                self.then_first,
+                *self.then_elements,
+            ):
                 element.print(printer, state, op)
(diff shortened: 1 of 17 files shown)


xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -487,12 +487,21 @@ def parse_optional_group(self) -> FormatDirective:
                 anchor = then_elements[-1]
         self.parse_punctuation("?")
 
-        if not then_elements:
+        # Pull whitespace element of front, as they are not parsed
```
alexarice, line 490: Would be keen to know if there's a more "pythonic" way of doing this.

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -487,12 +487,21 @@ def parse_optional_group(self) -> FormatDirective:
...
+        whitespace_counter = 0
+        while whitespace_counter < len(then_elements) and isinstance(
+            then_elements[whitespace_counter], WhitespaceDirective
+        ):
+            whitespace_counter += 1
```
PapyChacal, line 495: Here's a maybe Pythonic way?
```suggestion
        whitespace_counter = next(
            (
                i
                for i, e in enumerate(then_elements)
                if not isinstance(e, WhitespaceDirective)
            ),
            len(then_elements),
        )
```
I don't find it necessarily clearer myself, but answering as requested!

PapyChacal (approved):
Thanks!! stencil dialect be pretty :ok_hand: 

And I thought it was partially supported ahah, but it was actually accidentally accepted as a degenerate keyword, is that right?

alexarice:
The content parsed for any whitespace directive was whatever whitespace followed the directive. It just so happens that in nearly every case this is a single space for both cases.

This meant ``` `` ...`` actually parsed as the single space whitespace directive, the opposite of what you wanted to convey

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: I think I personally find it less clear what this does, but happy to go either way

alexarice:
Also I don't know much about the stencil dialect, but is the result of this pr correct for it?

xdsl/irdl/declarative_assembly_format_parser.py
PapyChacal: > I think I personally find it less clear what this does, but happy to go either way

Yeah I don't mind myself; as long as it works and is understandable - I'd say both styles are both for the everyday xDSL dev!

alexarice: if all the elements are whitespace then it should throw an error (as the next part of the code does)

PapyChacal: Yep this one kept it, you get the length of the list out so you would trigger the exception down

Otherwise, just remove the len(then_elements) from there and next() itself would throw an error, but not a nice hand-written one..

You could replace len(then_elements) in there by None and just
```python
if whitespace_counter is None:
    # fancy_error
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -822,21 +822,21 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
+    whitespace: Literal[" ", "\n", ""]
     """The whitespace that should be printed."""
 
     def parse(self, parser: Parser, state: ParsingState) -> None:
         pass
 
     def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> None:
         printer.print(self.whitespace)
-        state.last_was_punctuation = False
+        state.last_was_punctuation = len(self.whitespace) == 0
```
superlopuh, line 839: Length of strings is O(n) :)
```suggestion
        state.last_was_punctuation = not bool(self.whitespace)
```

alexarice: This string always has length 0 or 1

alexarice: maybe we could compromise on the simpler ```self.whitespace == ""```?

xdsl/irdl/declarative_assembly_format_parser.py
superlopuh: I would recommend something like this:
```suggestion
        first_non_whitespace_index = 0
        for first_non_whitespace_index, el in enumerate(then_elements):
            if isinstance(el, WhitespaceDirective):
                break
```

superlopuh (approved):

xdsl/irdl/declarative_assembly_format.py
superlopuh: Sounds good

alexarice:
I feel that this is pretty clean now (an amalgamation of some of the suggestions):
```
# Pull whitespace element of front, as they are not parsed
first_non_whitespace_index = None
for i, x in enumerate(then_elements):
    if not isinstance(x, WhitespaceDirective):
        first_non_whitespace_index = i
        break

if first_non_whitespace_index is None:
    self.raise_error("An optional group must have a non-whitespace directive")
```

alexarice:
If no-one objects to the above way of getting the index I think this is ready to merge?

PR #3045, opened 2024-08-16
rewriting: Make `modify_block_argument_type` `modify_value_type`

PapyChacal:
This requires absolutely no change in itself, I'm not sure why it was typed BlockArgument specifically?

Also, consider modifying a block argument's type as a modification of the parent operation, if any. Is anyone finding it problematic? It didn't trigger any problem on tested stuff, and it's natural for my use-case; stacked PR to come.
No big deal if it's an alarm, I'll hoist it in my user code.

Note/Usecase: Was intended for:
- #3046 

But I meanwhile just including the logic in my own files instead of waiting on core.


superlopuh (commented):
I would say that the code looks good, but would love to hear @math-fehr's opinion on the change before merging

math-fehr (approved):

superlopuh (approved):

PR #3046, opened 2024-08-16
transformations: stencil shape inference polish

PapyChacal:
Just update stencil shape inference to cooperative (i.e, bailing not crashing everything) and order-independent (proper use of the worklist instead of application-order assumptions) patterns.


n-io (approved):

PapyChacal:
Just ported the feature from:
- #3045 
In the stencil code for less friction.

PR #3047, opened 2024-08-16
transformations: Implement `shape-inference` pass

PapyChacal:
Implement a trait-based generic `shape-inference` pass, following the style of `canonicalize`, allowing for shape inference patterns to interoperate amongst dialect and all; Just hook existing stencil shape inference patterns in there for now.

Stacked on:
- #3046


superlopuh (commented):
Very nice! When I did this in the Toy tutorial, I followed the MLIR version's approach of just modifying the type directly, instead of using a rewrite pattern: https://mlir.llvm.org/docs/Tutorials/Toy/Ch-4/
I'm now not sure which approach is prefereable. It feels like the one used in Toy is more straightforward, why not use that directly?

PapyChacal:
> Very nice! When I did this in the Toy tutorial, I followed the MLIR version's approach of just modifying the type directly, instead of using a rewrite pattern: https://mlir.llvm.org/docs/Tutorials/Toy/Ch-4/ I'm now not sure which approach is prefereable. It feels like the one used in Toy is more straightforward, why not use that directly?

I belive it's less generic, right? I need some level of worklist management for my pass to work independently of the application order. stencil shape inference also works backwards, inferring operand sizes from results. So to interoperate both direction, I think we need that?

I would argue for hooks to hook a simpler case like the Toy one in the bigger case, if you really want to keep Toy simple yet integrate in this one. I also think it might be okay to just have a simple one on Toy for demo purposes, and it doesn't have to imply constraints on the core one?

superlopuh (approved):

n-io (approved):
Looks great, thanks

PR #3049, opened 2024-08-16
dialect: (irdl) Add region and attributes operations

alexarice:
Adds `RegionsOp`, `RegionOp`, `RegionType`, and `AttributesOp` to irdl

### DIFF
--- a/tests/filecheck/dialects/irdl/testd.irdl.mlir
+++ b/tests/filecheck/dialects/irdl/testd.irdl.mlir
@@ -168,7 +168,10 @@ builtin.module {
     // CHECK:      irdl.operation @attr_op {
     // CHECK-NEXT:   %{{.*}} = irdl.any
     // CHECK-NEXT:   %{{.*}} = irdl.is i64
-    // CHECK-NEXT:   irdl.attributes { "attr1" = %{{.*}}, "attr2" = %{{.*}} }
+    // CHECK-NEXT:   irdl.attributes {
+    // CHECK-NEXT:     "attr1" = %{{.*}},
+    // CHECK-NEXT:     "attr2" = %{{.*}}
+    // CHECK-NEXT:   }
     // CHECK-NEXT: }
     irdl.operation @attr_op {
       %0 = irdl.any
--- a/tests/filecheck/dialects/irdl/testd.irdl.mlir
+++ b/tests/filecheck/dialects/irdl/testd.irdl.mlir
@@ -146,14 +146,14 @@ builtin.module {
     }
 
     // CHECK:      irdl.operation @op_with_regions {
-    // CHECK-NEXT:   %{{.*}} = irdl.region
-    // CHECK-NEXT:   %{{.*}} = irdl.region()
-    // CHECK-NEXT:   %{{.*}} = irdl.is i32
-    // CHECK-NEXT:   %{{.*}} = irdl.is i64
-    // CHECK-NEXT:   %{{.*}} = irdl.region(%{{.*}}, %{{.*}})
-    // CHECK-NEXT:   %{{.*}} = irdl.region with size 3
-    // CHECK-NEXT:   irdl.regions(%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}})
-    // CHECK-NEXT: }
+    // CHECK-NEXT:    %r0 = irdl.region
+    // CHECK-NEXT:    %r1 = irdl.region()
+    // CHECK-NEXT:    %v0 = irdl.is i32
+    // CHECK-NEXT:    %v1 = irdl.is i64
+    // CHECK-NEXT:    %r2 = irdl.region(%v0, %v1)
+    // CHECK-NEXT:    %r3 = irdl.region with size 3
+    // CHECK-NEXT:    irdl.regions(%r0, %r1, %r2, %r3)
+    // CHECK-NEXT:  }
     irdl.operation @op_with_regions {
       %r0 = irdl.region
       %r1 = irdl.region()
--- a/tests/filecheck/dialects/irdl/testd.irdl.mlir
+++ b/tests/filecheck/dialects/irdl/testd.irdl.mlir
@@ -166,11 +166,11 @@ builtin.module {
     }
 
     // CHECK:      irdl.operation @attr_op {
-    // CHECK-NEXT:   %{{.*}} = irdl.any
-    // CHECK-NEXT:   %{{.*}} = irdl.is i64
+    // CHECK-NEXT:   %[[#first:]] = irdl.any
+    // CHECK-NEXT:   %[[#second:]] = irdl.is i64
     // CHECK-NEXT:   irdl.attributes {
-    // CHECK-NEXT:     "attr1" = %{{.*}},
-    // CHECK-NEXT:     "attr2" = %{{.*}}
+    // CHECK-NEXT:     "attr1" = %[[#first]],
+    // CHECK-NEXT:     "attr2" = %[[#second]]
     // CHECK-NEXT:   }
     // CHECK-NEXT: }
     irdl.operation @attr_op {
--- a/tests/filecheck/dialects/irdl/testd.irdl.mlir
+++ b/tests/filecheck/dialects/irdl/testd.irdl.mlir
@@ -164,5 +164,19 @@ builtin.module {
 
       irdl.regions(%r0, %r1, %r2, %r3)
     }
+
+    // CHECK:      irdl.operation @attr_op {
+    // CHECK-NEXT:   %{{.*}} = irdl.any
+    // CHECK-NEXT:   %{{.*}} = irdl.is i64
+    // CHECK-NEXT:   irdl.attributes { "attr1" = %{{.*}}, "attr2" = %{{.*}} }
+    // CHECK-NEXT: }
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/irdl/testd.irdl.mlir
```diff
@@ -144,5 +144,39 @@ builtin.module {
...
     }
+
+    // CHECK:      irdl.operation @op_with_regions {
+    // CHECK-NEXT:   %{{.*}} = irdl.region
+    // CHECK-NEXT:   %{{.*}} = irdl.region()
+    // CHECK-NEXT:   %{{.*}} = irdl.is i32
+    // CHECK-NEXT:   %{{.*}} = irdl.is i64
+    // CHECK-NEXT:   %{{.*}} = irdl.region(%{{.*}}, %{{.*}})
+    // CHECK-NEXT:   %{{.*}} = irdl.region with size 3
+    // CHECK-NEXT:   irdl.regions(%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}})
```
superlopuh, line 155: why anonymize?

tests/filecheck/dialects/irdl/testd.irdl.mlir
```diff
@@ -144,5 +144,39 @@ builtin.module {
...
+    // CHECK-NEXT:   %{{.*}} = irdl.any
+    // CHECK-NEXT:   %{{.*}} = irdl.is i64
+    // CHECK-NEXT:   irdl.attributes { "attr1" = %{{.*}}, "attr2" = %{{.*}} }
+    // CHECK-NEXT: }
+    irdl.operation @attr_op {
+      %0 = irdl.any
+      %1 = irdl.is i64
+      irdl.attributes {
+        "attr1" = %0,
+        "attr2" = %1
```
superlopuh, line 178: how does this work?

superlopuh: Oh I see it's fully custom syntax

tests/filecheck/dialects/irdl/testd.irdl.mlir
alexarice: thought it was the done thing

tests/filecheck/dialects/irdl/testd.irdl.mlir
alexarice: yes, I also wasn't sure if it should be printed in this way or if the single line print is acceptable?

tests/filecheck/dialects/irdl/testd.irdl.mlir
superlopuh: Yes and no, we don't want spurious changes in the filecheck tests if we add something in the middle, but if you print the IR you'll see that the names of the values are preserved, so it will be stable, and they're actually really useful in this case to track how the constraints are propagated, I'd recommend deanonymizing. I added an option to `filecheckize` to do this: `--xdsl-anonymize`, this will only anonymize `%[\d]+`, not ssa values with name hints.

tests/filecheck/dialects/irdl/testd.irdl.mlir
superlopuh: I think this looks nicer

xdsl/dialects/irdl/irdl.py
```diff
@@ -379,6 +394,64 @@ def print(self, printer: Printer) -> None:
...
+        )
+        return AttributesOp(attributes)
+
+    def print(self, printer: Printer) -> None:
+        dictionary = dict(zip(self.attribute_value_names, self.attribute_values))
+        printer.print(" ")
+        printer.print_dictionary(
+            dictionary,
+            lambda x: printer.print_attribute(x),
+            lambda x: printer.print_operand(x),
```
superlopuh, line 437: does this work?
```suggestion
            printer.print_attribute,
            printer.print_operand,
```

tests/filecheck/dialects/irdl/testd.irdl.mlir
alexarice: Ok I'll try to figure out how new lines/indentation work

xdsl/dialects/irdl/irdl.py
alexarice: Just checked, seems to work

xdsl/dialects/irdl/irdl.py
```diff
@@ -379,6 +394,64 @@ def print(self, printer: Printer) -> None:
...
+    attribute_values = var_operand_def(AttributeType())
+
+    attribute_value_names = attr_def(ArrayAttr[StringAttr])
+
+    def __init__(self, attributes: dict[str, SSAValue]):
+        operands = tuple(attributes.values())
+        names = ArrayAttr(StringAttr(x) for x in attributes.keys())
+        super().__init__(
+            operands=(operands,), attributes={"attribute_value_names": names}
+        )
```
superlopuh, line 420: I would recommend adding a dumb init with the fields of the op, and a `.get` helper with the smarts, similarly below, there's currently no way to create these ops with exactly the things you want

xdsl/printer.py
```diff
@@ -259,12 +259,17 @@ def print_dictionary(
...
+            self.print("{}")
+        else:
+            self.print("{ ")
+            for i, (key, value) in enumerate(elems.items()):
+                if i:
+                    self.print_string(delimiter)
+                print_key(key)
+                self.print_string(" = ")
+                print_value(value)
+            self.print(" }")
```
superlopuh, line 272: Just noticed this, I don't think it belongs in a PR tagged as `dialects`, since it's actually a core change. Why is it necessary?

alexarice: It's not "necessary", I could just put this logic in the printer for AttributesOp, but I thought it would make sense to use the print_dictionary function and I (maybe incorrectly) thought that since it was used nowhere else that it would be safe to modify/improve

xdsl/dialects/irdl/irdl.py
alexarice: Is there example of such a `get` helper somewhere?

xdsl/printer.py
alexarice: Would be happy to split this change off into a separate pr if wanted

tests/filecheck/dialects/irdl/testd.irdl.mlir
superlopuh: `cat my_ir.mlir | filecheckize`

xdsl/dialects/irdl/irdl.py
superlopuh: Not sure, I just mean a class/static method that returns the Operation

tests/filecheck/dialects/irdl/testd.irdl.mlir
alexarice: and where would one get filecheckize from?

xdsl/printer.py
superlopuh: Hmm, that's weird that it's not used... I'm also surprised none of the tests fail... My first instinct is to remove the helper if it's not used. In either case, I would probably recommend do this in a custom way in your custom syntax.

tests/filecheck/dialects/irdl/testd.irdl.mlir
superlopuh: `pip`, as usual :)

alexarice: Ah I was trying to find it as a subproject of xdsl

xdsl/printer.py
alexarice: I have reverted the core change and put improved printing in `AttributesOp`

alexarice:
I simplified the printing a bit, removing the empty case as this should never occur in practice (in fact mlir does not even parse it). I also added the verifier that mlir has to check that there are the same number of names and values, is it worth adding a test for this?

alexarice:
Just realised that mlir does parse the empty case, but does it by omitting the curly braces entirely

tests/filecheck/dialects/irdl/testd.irdl.mlir
```diff
@@ -144,5 +144,49 @@ builtin.module {
...
+
+      irdl.regions(%r0, %r1, %r2, %r3)
+    }
+
+    // CHECK:      irdl.operation @attr_op {
+    // CHECK-NEXT:   %{{.*}} = irdl.any
+    // CHECK-NEXT:   %{{.*}} = irdl.is i64
+    // CHECK-NEXT:   irdl.attributes {
+    // CHECK-NEXT:     "attr1" = %{{.*}},
+    // CHECK-NEXT:     "attr2" = %{{.*}}
```
superlopuh, line 173: can you deanon this, please?

superlopuh (approved):

tests/filecheck/dialects/irdl/testd.irdl.mlir
alexarice: Is this deanonification sufficient?

superlopuh: it's perfect, thank you!

PR #3051, opened 2024-08-16
transformations: (csl-stencil-to-csl-wrapper) Translate func args to exportable memref.global

n-io:
Top-level func args are no longer set up as block args in the newly created program module, but are instead translated to (exportable) `memref.global` ops.


PapyChacal (approved):

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -136,6 +136,54 @@ def get_csl_stencil_apply_ops(
...
+                arg_ops.append(
+                    memref.Global.get(
+                        sym_name=StringAttr(arg_name),
+                        sym_type=csl_stencil_bufferize.tensor_to_memref_type(
+                            arg.type.get_element_type()
+                        ),
+                        initial_value=UnitAttr(),
+                        sym_visibility=StringAttr("public"),
+                        # constant: UnitAttr | None = None,
+                        # alignment: int | IntegerAttr[IntegerType] | None = None,
```
superlopuh, line 163: commented code

n-io: @dk949 Do we want to specify any of them?

dk949: I don't believe we have any way to specify alignment in the backend. As for constness, I _think_ these will always be non-const. So you can leave both as default.

tests/filecheck/transforms/csl-stencil-to-csl-wrapper.mlir
```diff
@@ -56,38 +56,44 @@ builtin.module {
...
+// CHECK-NEXT:   ^1(%26 : i16, %27 : i16, %28 : i16, %29 : i16, %30 : i16, %31 : i16, %32 : i16, %memcpy_params : !csl.comptime_struct, %stencil_comms_params : !csl.comptime_struct, %isBorderRegionPE : i1):
 // CHECK-NEXT:     %33 = "csl_wrapper.import"(%memcpy_params) <{"module" = "<memcpy/memcpy>", "fields" = [""]}> : (!csl.comptime_struct) -> !csl.imported_module
 // CHECK-NEXT:     %34 = "csl_wrapper.import"(%29, %31, %stencil_comms_params) <{"module" = "stencil_comms.csl", "fields" = ["pattern", "chunkSize", ""]}> : (i16, i16, !csl.comptime_struct) -> !csl.imported_module
+// CHECK-NEXT:     "memref.global"() <{"sym_name" = "a", "type" = memref<512xf32>, "initial_value", "sym_visibility" = "public"}> : () -> ()
+// CHECK-NEXT:     %35 = memref.get_global @a : memref<512xf32>
+// CHECK-NEXT:     %a = builtin.unrealized_conversion_cast %35 : memref<512xf32> to !stencil.field<[-1,1023]x[-1,511]xtensor<512xf32>>
```
dk949, line 64: I don't quite follow how this `unrealized_conversion_cast`  gets passed to the `stencil.load` below. Is that done with the `arg_op_mapping`?

dk949 (approved):

tests/filecheck/transforms/csl-stencil-to-csl-wrapper.mlir
n-io: Any `stencil.load` should be resolved by `--stencil-bufferize` and not be there at this stage, and everything should be based on `stencil.field<AxBxtensor<Cxf32>>`. This in turn is translated to a memref in `--csl-stencil-bufferize`, which takes the element_type (tensor) and translates it to a memref. Maybe it should translate to `memref<1x1xCxf32>` instead of `memref<Cxf32>`, but this is all linked to DMP which has changes in the pipeline that allow it to operate on `stencil.field`, so it's best to wait for this to be merged.

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: Thanks, removing commented code.

PR #3052, opened 2024-08-17
CI: update pre-commit-hooks

superlopuh:
I've subscribed to both the precommit hook repos so I'll just keep these updated manually.


PapyChacal (approved):
tick

PR #3050, opened 2024-08-16
dialects: dmp: make strategies attributes and carry them in IR

PapyChacal:
As of now, the decomposition strategies are dataclasses and the information to construct them is expected to match in multiple bits of the pipeline (actually fused in `distribute-stencil`, which happens to do shape inference too, but I'm trying to decouple things here)

So instead, define them as attributes, and make the initial distribution pass pop that attribute on the swaps. The swap shape inference then do not need extra information to carry on.

Teaser: I want this part to be just collaborating shape inference as in:
- #3047

PR #3054, opened 2024-08-17
transformations: Decouple `distribute-stencil` from `shape-inference`

PapyChacal:
Make dmp shape inference part of the general `shape-inference`, and remove `shape-inference` as a part of `distribute-stencil`.

Stacked on:
- #3050


n-io (approved):
Looks good, thanks! Please can you leave a comment whether dmp needs to run before/after shape inference, or does either work equally well?

PapyChacal:
> Looks good, thanks! Please can you leave a comment whether dmp needs to run before/after shape inference, or does either work equally well?

Shape inference would need to run after dmp. DMP is just structurally placing "when" communications might need to happen. Shape inference allows to infer what needs to be communicated at those points.
Ideally, shape inference could happen before too, and just happen again on the swaps if needed. This is not tested at this point though, and I would expect it to fail from experience.

This drop-in replacement order is indicated by the update of `distribute-stencil.mlir`

PR #3057, opened 2024-08-17
backend: (JAX) add export to JAX executable

superlopuh:
PR 1/2, allowing for the creation of executable objects, with an `execute` method that just calls the JAX underlying executable.


xdsl/backend/jax_executable.py
```diff
@@ -0,0 +1,87 @@
...
+    Usage:
+
+    ``` python
+    my_module_str =\"""
+    func.func @main(%arg0: tensor<5xf32>) -> tensor<5xf32> {
+        ...
+    }
+    \"""
+    my_module = parser.parse_module(my_module_str)
+    @JaxExecutable.compile(module)
```
zero9178, line 50:
```suggestion
    @JaxExecutable.compile(my_module)
```

zero9178, line 50: I don't see a unit test using the decorator syntax nor a `__call__` that would delegate to `execute`. Is this intentional?

zero9178: Just saw the stacked PR which implements this. The documentation seems to just be ahead of its time :slightly_smiling_face:

zero9178 (approved):

xdsl/backend/jax_executable.py
superlopuh: 🔮

superlopuh: I'll wait for both to be approved to merge this one

PR #3059, opened 2024-08-18
installation: Update nix flake

alexarice:
Switches nixpkgs version to unstable to get latest version of `ruff` (0.6.1).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,15 +20,18 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


superlopuh (approved):

superlopuh:
I assume you've tested this! Is there a way to get a nix CI?

alexarice:
If you have some ideas on what you would want a nix ci to do, then I could give it a go. What is currently there only deals with getting ruff and pyright working correctly

PR #3062, opened 2024-08-19
dependencies: fix numpy version for subprojects that depend on it

superlopuh:
Numpy changed something in their typing annotations and now our main is broken. We should probably move to something that locks our dependencies for our CI, in the meantime, this PR specifies a numpy version and fixes the type errors.


tobiasgrosser (approved):
Thank you.

PR #3065, opened 2024-08-19
core: (irdl) Add regions to declarative assembly format

kimxworrall:
Added support for regions to the declarative assembly format.

The declarative assembly format now supports:
- Referring to region variables
- Referring to optional or variadic region variables
- Printing custom assembly for region variables

The format requires that:
- The attr-dict-with-keyword directive is present and not attr-dict. This is to maintain consistency with mlir.
- The regions appear after the attributes.

Also added tests.


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -18,7 +20,9 @@
     ParametrizedAttribute,
     SSAValue,
     TypedAttribute,
+    Region,
 )
+from xdsl.ir.core import Block, Operation
```
alexarice, line 25: Please install the precommit, this shouldn't pass the linter

alexarice (commented):
Could you comment on why regions must come after attributes? I can't see anything about this in the mlir documentation (which doesn't mean it's not true)

xdsl/irdl/declarative_assembly_format.py
kimxworrall: That's the part I'm not completely sure about from the documentation, I know that the generic format does this, and that the attribute dict has to be attr-dict-with-keyword, but am unsure if its allowed to have the attributes in the middle of regions (but labelled).

alexarice: I tried some stuff locally and the format "$body attr-dict-with-keyword" seems to work absolutely fine

.github/workflows/update_xdsl_pyodide_build.py
superlopuh, line 1: why this change?

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1087,6 +1090,194 @@ class OptionalResultOp(IRDLOperation):
...
+    """Test the parsing of multiple optional regions"""
+
+    """Test that multiple optional regions requires the ABCMeta PyRDL option."""
+    with pytest.raises(
+        PyRDLOpDefinitionError,
+        match="Operation test.optional_regions defines more than two variadic regions",
+    ):
+
+        @irdl_op_definition
+        class OptionalOperandsOp(IRDLOperation):
```
superlopuh, line 1273:
```suggestion
        class OptionalOperandsOp(IRDLOperation):  # pyright: ignore[reportUnusedClass]
```

alexarice: Surely the warning shouldn't be ignored and is an indication that something is wrong?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -159,11 +163,25 @@ def parse(
...
+        regions = cast(
+            Sequence[
+                Region
+                | Sequence[Operation]
+                | Sequence[Block]
+                | Sequence[Region | Sequence[Operation] | Sequence[Block]]
+                | None
+            ],
+            state.regions,
+        )
```
superlopuh, line 177: I'm surprised this is needed, I'll clone and play with this locally quickly

xdsl/irdl/operations.py
```diff
@@ -1735,7 +1735,7 @@ def fun(self: Any, idx: int = arg_idx, previous_vars: int = previous_variadics):
         not any(isinstance(o, arg_size_option) for o in op_def.options)
     ):
         arg_size_option_name = type(arg_size_option).__name__
-        raise Exception(
+        raise PyRDLOpDefinitionError(
```
superlopuh, line 1738: 👍

xdsl/tools/xdsl-opt
superlopuh, line 1: why this change?

tests/irdl/test_declarative_assembly_format.py
alexarice: Ignore me, didn't see this was just a test

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1087,6 +1090,194 @@ class OptionalResultOp(IRDLOperation):
...
+        @irdl_op_definition
+        class NoRegionTypeOp(IRDLOperation):  # pyright: ignore[reportUnusedClass]
+            name = "test.region_op_missing_keyword"
+            region = region_def()
+
+            assembly_format = "attr-dict $region"
+
+
+def test_region_before_attr():
+    """Test that regions should appear after the 'attr-dict-with-keyword' directive."""
```
alexarice, line 1126: My mlir testing suggests that this should be valid assembly format

xdsl/irdl/declarative_assembly_format.py
kimxworrall: Hmm okay, maybe its because I'm running an older version, will fix

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -723,6 +741,80 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
+        state.should_emit_space = True
+
+
+@dataclass(frozen=True)
+class VariadicRegionVariable(
+    VariadicVariable, VariableDirective, OptionallyParsableDirective
+):
+    """
+    A variadic region variable, with the following format:
+      region-directive ::= ( percent-ident ( `,` percent-id )* )?
```
alexarice, line 770: Is this the correct format? SHould it not still be `region-directive ::= dollar-ident`?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -723,6 +741,80 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
             state.should_emit_space = True
 
 
+@dataclass(frozen=True)
+class RegionVariable(VariableDirective):
```
alexarice, line 745: Surely this can be `OptionallyParsable` too

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -723,6 +741,80 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
+        if region:
+            printer.print_list(region, printer.print_region, delimiter=" ")
+            state.last_was_punctuation = False
+            state.should_emit_space = True
+
+
+class OptionalRegionVariable(OptionalVariable, OptionallyParsableDirective):
+    """
+    An optional region variable, with the following format:
+      region-directive ::= ( percent-ident )?
```
alexarice, line 797: This format seems incorrect again (see above)?

alexarice (commented):
Haven't read through everything but here's some comments

xdsl/irdl/declarative_assembly_format.py
alexarice: From my testing it really only seems to complain if you have an attr-dict immediately followed by a region

xdsl/tools/xdsl-opt
kimxworrall: Ah I think this is just the cache auto-updated and it overwrote the file without changing anything.

xdsl/irdl/declarative_assembly_format.py
alexarice: https://github.com/llvm/llvm-project/blob/ea28668f8c980c9987b0e2f305b0a8eff9797401/mlir/tools/mlir-tblgen/OpFormatGen.cpp#L3010 Here is the check

kimxworrall: As in does not accept "attr-dict $body" but does accept "attr-dict-with-keyword $body"?

alexarice: Yes, it looks like the only thing that is denied is attr-dict followed directly by a region variable

alexarice: I also tested ```attr-dict `,` $body``` and that worked as expected too.

xdsl/irdl/declarative_assembly_format.py
superlopuh: https://github.com/xdslproject/xdsl/commit/a7ba1d417b16f0386e4a019704ea750dcdb43715

superlopuh: I don't think the `None` in the inner `list[Region | None]` is needed, might as well remove?

kimxworrall: Thank you, that seems sensible

superlopuh: Feel free to merge those changes into this PR, then I think this is GTG

xdsl/tools/xdsl-opt
superlopuh, line 1: can revert?

xdsl/irdl/declarative_assembly_format.py
kimxworrall: Thanks, have done so now

xdsl/irdl/declarative_assembly_format.py
kimxworrall: Have made these updates and added tests, should work now

alexarice (commented):
There's still two files that you've not touched being pulled in somehow. I've left a few minor comments but looking good, will give it a test tomorrow

alexarice:
I also have no idea why the mlir integration tests are failing

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -48,19 +49,20 @@ class ParsingState:
...
     properties: dict[str, Attribute]
     constraint_context: ConstraintContext
 
     def __init__(self, op_def: OpDef):
-        if op_def.regions or op_def.successors:
+        if op_def.successors:
             raise NotImplementedError(
-                "Operation definitions with regions "
-                "or successors are not yet supported"
+                "Operation definitions with " "successors are not yet supported"
```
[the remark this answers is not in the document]
superlopuh:
```suggestion
                "Operation definitions with successors are not yet supported"
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -159,11 +161,13 @@ def parse(
             properties = state.properties
         else:
             properties = op_def.split_properties(state.attributes)
+
```
superlopuh, line 164:
```suggestion
```

superlopuh: I feel like it's easier just to post suggestions then it's two clicks in the GH UI to apply them

alexarice: I just forgot this was a feature

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -723,6 +727,85 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
 
+@dataclass(frozen=True)
+class RegionVariable(VariableDirective, OptionallyParsableDirective):
+    """
+    A region variable, with the following format:
+      region-directive ::= dollar-ident
+    The directive will request a space to be printed after.
+    """
+
+    def parse(self, parser: Parser, state: ParsingState) -> None:
```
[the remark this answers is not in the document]
PapyChacal: Oh wait: Only the OptionalRegionVariabl needs to implement it, not this one, indeed :+1: Good catch!

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -723,6 +727,85 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
+        state.should_emit_space = True
+
+
+@dataclass(frozen=True)
+class VariadicRegionVariable(
+    VariadicVariable, VariableDirective, OptionallyParsableDirective
+):
+    """
+    A variadic region variable, with the following format:
+      region-directive ::= ( dollar-ident ( `,` dollar-id )* )?
```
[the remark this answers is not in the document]
PapyChacal: Agreed, it should just be dollar-ident.

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -723,6 +726,85 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
+        if region:
+            printer.print_list(region, printer.print_region, delimiter=" ")
+            state.last_was_punctuation = False
+            state.should_emit_space = True
+
+
+class OptionalRegionVariable(OptionalVariable, OptionallyParsableDirective):
+    """
+    An optional region variable, with the following format:
+      region-directive ::= ( dollar-ident )?
```
PapyChacal, line 787: Same here, should just be dollar-ident

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1087,6 +1090,213 @@ class OptionalResultOp(IRDLOperation):
...
+    check_roundtrip(program, ctx)
+    check_equivalence(program, generic_program, ctx)
+
+
+@irdl_op_definition
+class MiscOp(IRDLOperation):
+    name = "test.typed_attr"
+    attr = attr_def(IntegerAttr[I32])
+
+    assembly_format = "$attr attr-dict"
```
PapyChacal, line 1163: Minor: could you maybe populate the below tests with test.op (TestOp) instead of creating a new one

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1087,6 +1090,213 @@ class OptionalResultOp(IRDLOperation):
...
+        (
+            "attr-dict-with-keyword $region",
+            'test.variadic_region {\n  "test.op"() : () -> ()\n} {\n  "test.op"() : () -> ()\n}',
+            '"test.variadic_region"() ({ "test.op"() : () -> ()}, { "test.op"() : () -> ()}) : () -> ()',
+        ),
+        (
+            "attr-dict-with-keyword $region",
+            'test.variadic_region {\n  "test.op"() : () -> ()\n} {\n  "test.op"() : () -> ()\n}',
+            '"test.variadic_region"() ({ "test.op"() : () -> ()}, {"test.op"() : () -> ()}) : () -> ()',
+        ),
```
PapyChacal, line 1228: Those two tests are the same!
I'm fine just keeping one, or making the last one a 3-regions example for the show

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1087,6 +1090,213 @@ class OptionalResultOp(IRDLOperation):
...
+    ctx.load_dialect(Test)
+
+    check_roundtrip(program, ctx)
+    check_equivalence(program, generic_program, ctx)
+
+
+def test_multiple_optional_regions():
+    """Test the parsing of multiple optional regions"""
+
+    """Test that multiple optional regions requires the ABCMeta PyRDL option."""
```
PapyChacal, line 1285: This just requires the right IRDLOption, and I believe this is what this test tests, and it shouldn't be here.
The declarative format framework should check that two variadic regions do not directly follow each other (ambiguous). It might already through baseclasses, but here it catches the irdl_op_definition even before that.

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1087,6 +1090,213 @@ class OptionalResultOp(IRDLOperation):
...
+
+        @irdl_op_definition
+        class OptionalOperandsOp(IRDLOperation):  # pyright: ignore[reportUnusedClass]
+            name = "test.optional_regions"
+            region1 = opt_region_def()
+            region2 = opt_region_def()
+
+            assembly_format = "attr-dict-with-keyword $region1 $region2"
+
+
```
PapyChacal, line 1299: You implemented it in such a nice way that it should work with optional groups, maybe we wont to test those?

Things like `` ($opt_region^ `keyword`)? `` and `` (`keyword` $opt_region^)? ``

Don't take this as a requirement, would only be even nicer if you have time for it too, but I'd be glad to merge it without that too!

PapyChacal (approved):
Amazing to see that in!

The implementation and existing tests seem perfect to me, only have some minor polish comments :) Making this an approval as at this point I consider any further polishing a bonus, the feature seems ready to benefit xDSL in itself!

Thanks a lot!

xdsl/irdl/declarative_assembly_format.py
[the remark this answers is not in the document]
alexarice: Sorry I wasn't clear, I meant that I think `parse` gets a default implementation from the definition of `parse_optional`. I think you could concoct a weird example where you could still use a region in an optional group even if it's non optional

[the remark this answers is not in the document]
kimxworrall: You're right, it does get a default implementation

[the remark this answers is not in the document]
PapyChacal: > Sorry I wasn't clear, I meant that I think `parse` gets a default implementation from the definition of `parse_optional`. I think you could concoct a weird example where you could still use a region in an optional group even if it's non optional

You're right; just for full context: xDSL doesn't implement that case yet (optional group's else-elements) so a lot of other similar declarative variables do not implement it either.

So for the sake of this PR, both removing it or keeping it would be fine :slightly_smiling_face:

If keeping the optional, using the default `parse` implementation if it is working would be nice indeed!

tests/irdl/test_declarative_assembly_format.py
PapyChacal: My first comment is a bit cryptic;
`OptionalOperandsOp` should contain
`irdl_options = [AttrSizedRegionSegments()]`
to make `irdl_op_definition` happy with the op.

Allowing to test if the declarative syntax is indeed emitting its own error in that case!

alexarice (approved):
I've had a little test locally and this seems reasonable. Thanks for implementing this. Happy for this to be merged once the remaining comments are resolved

xdsl/irdl/declarative_assembly_format.py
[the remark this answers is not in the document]
kimxworrall: Keeping just parse_optional and allowing the default parse implementaion to be used

superlopuh (approved):

tests/irdl/test_declarative_assembly_format.py
kimxworrall: Fixed now

tests/irdl/test_declarative_assembly_format.py
kimxworrall: Have added these two tests

PR #3067, opened 2024-08-19
core: Implement IRDL SameSize options

### LINKED ISSUE #3064: core: We need more variadic inference for compatibility
## Symptom: xDSL's `memref.extract_strided_metadata` is incompatible with MLIR's
(I wasn't sure which one should be title or subtitle)

Introduced by:
- #2748 
(The op was not defined before so at least handlable as unregistered - not possible to interoperate it in custom syntax now.)

This operation has two variadic results, which in xDSL enforces to have 
```python
    irdl_options = [AttrSizedResultSegments()]
```
And thus expect the related attribute.

Except MLIR prints it without it:
```mlir
%1:6 = "memref.extract_strided_metadata"(%0) : (memref<512x384xi32>) -> (memref<i32>, index, index, index, index, index)
```

What trickery is this, you might ask :mage: 

Well they provide an internal trait `SameVariadicResultSize`, bypassing the need of an explicit encoding of each variadic result's size here, because as cleary stated by the name, one can just infer it easily.

Except xDSL cannot at the moment. :cry:

PapyChacal:
Should help fix:
- #3064 

(Hey GitHub, I said should help fix #3064?)

### DIFF
--- a/xdsl/irdl/operations.py
+++ b/xdsl/irdl/operations.py
@@ -1391,7 +1391,7 @@ def get_variadic_sizes(
     # If the operation has to related SameSize option, equally distribute the
     # variadic arguments between the variadic definitions.
     option = next((o for o in op_def.options if isinstance(o, same_size_option)), None)
-    if option:
+    if option is not None:
         non_variadic_defs = len(defs) - len(variadic_defs)
         variadic_args = len(args) - non_variadic_defs
         if variadic_args % len(variadic_defs):
--- a/xdsl/irdl/operations.py
+++ b/xdsl/irdl/operations.py
@@ -249,6 +249,30 @@ class AttrSizedSuccessorSegments(AttrSizedSegments):
     """Name of the attribute containing the variadic successor sizes."""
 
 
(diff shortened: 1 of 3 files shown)


math-fehr:
LGTM, I would just add a test with 2 variadics and 1 non variadic!

PapyChacal:
> LGTM, I would just add a test with 2 variadics and 1 non variadic!

I directly added a non-variadic between the two variadics instead to avoid bloating!

I also added a test for the verification exception.

xdsl/irdl/operations.py
```diff
@@ -1331,6 +1388,19 @@ def get_variadic_sizes(
             )
         return [len(args) - len(defs) + 1]
 
+    # If the operation has to related SameSize option, equally distribute the
+    # variadic arguments between the variadic definitions.
+    option = next((o for o in op_def.options if isinstance(o, same_size_option)), None)
+    if option:
```
superlopuh, line 1394:
```suggestion
    if option is not None:
```

superlopuh (approved):

PR #3069, opened 2024-08-20
dialects: Update `tensor` side-effect traits

n-io:
(@PapyChacal  edit)
Just implement the right side effect interfaces on the `tensor` dialect, and demonstrate usage in stencil bufferization.
- Checked against MLIR.
- Removed `Pure` from tensor.empty as the upstream is marked ConditionallySpeculatable. Again, we don't have those traits yet.

### DIFF
--- a/xdsl/transforms/stencil_bufferize.py
+++ b/xdsl/transforms/stencil_bufferize.py
@@ -213,6 +213,9 @@ class ApplyStoreFoldPattern(RewritePattern):
     def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
         # Check that the destination is not used between the apply and store.
         dest = store.field
+        start = dest.owner
+        if isinstance(start, Block):
+            start = cast(Operation, start.first_op)
         effecting = [
             o
             for o in walk_from_to(apply, store)
--- a/xdsl/transforms/stencil_bufferize.py
+++ b/xdsl/transforms/stencil_bufferize.py
@@ -213,11 +213,9 @@ class ApplyStoreFoldPattern(RewritePattern):
     def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
         # Check that the destination is not used between the apply and store.
         dest = store.field
-        if apply.next_op is None:
-            return True
         effecting = [
             o
-            for o in walk_from_to(apply.next_op, store)
+            for o in walk_from_to(apply, store)
             if might_effect(o, {MemoryEffectKind.READ, MemoryEffectKind.WRITE}, dest)
         ]
         return not effecting
--- a/xdsl/transforms/stencil_bufferize.py
+++ b/xdsl/transforms/stencil_bufferize.py
@@ -213,9 +213,6 @@ class ApplyStoreFoldPattern(RewritePattern):
     def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
         # Check that the destination is not used between the apply and store.
         dest = store.field
-        start = dest.owner
-        if isinstance(start, Block):
-            start = cast(Operation, start.first_op)
         effecting = [
             o
             for o in walk_from_to(apply, store)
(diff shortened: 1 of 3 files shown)


PapyChacal (changes requested):
No, we definitely want to check the apply

What if it is already reading from the destination buffer you're folding as a destination? It might violate dependencies.
If I get the time, I intend to make it a bit more nuanced for the apply itself, to allow at least in-place stencils (where reading and writing to the same one is okay because it is an element wise mapping, for example)
But we can't just ignore it in general.

I'm happy to discuss what concrete issue you faced that lead to trying this change though?

n-io:
> No, we definitely want to check the apply
> What if it is already reading from the destination buffer you're folding as a destination? It might violate dependencies.

If the folded store happens by implication at the end of the apply, then I do not see the difference if it happens at the end of the apply instruction or immediately in the next instruction after. The only difference should be if there is something else going on *in-between* the end of the apply op and the store op. Is this not also what the comment in the method suggests?

```
    def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
        # Check that the destination is not used between the apply and store.
```


> I'm happy to discuss what concrete issue you faced that lead to trying this change though?

Please see the included filecheck.

PapyChacal:
> > No, we definitely want to check the apply
> > What if it is already reading from the destination buffer you're folding as a destination? It might violate dependencies.
> 
> If the folded store happens by implication at the end of the apply, then I do not see the difference if it happens at the end of the apply instruction or immediately in the next instruction after. The only difference should be if there is something else going on _in-between_ the end of the apply op and the store op. Is this not also what the comment in the method suggests?
> 
> ```
>     def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
>         # Check that the destination is not used between the apply and store.
> ```
> 
> > I'm happy to discuss what concrete issue you faced that lead to trying this change though?
> 
> Please see the included filecheck.

The apply represents a computation

> > No, we definitely want to check the apply
> > What if it is already reading from the destination buffer you're folding as a destination? It might violate dependencies.
> 
> If the folded store happens by implication at the end of the apply, then I do not see the difference if it happens at the end of the apply instruction or immediately in the next instruction after. The only difference should be if there is something else going on _in-between_ the end of the apply op and the store op. Is this not also what the comment in the method suggests?
> 
> ```
>     def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
>         # Check that the destination is not used between the apply and store.
> ```
> 
> > I'm happy to discuss what concrete issue you faced that lead to trying this change though?
> 
> Please see the included filecheck.

The apply is applying a stencil pattern to a whole field.
If every point depends on the previous value of its neighbours, you can't both update neighbours before the point and the point before neighbours?
It technically is a corner case in your example because you are expressing a one-point-stencil, so there is only one point to update, so what you said apply.
Again, the stencil dialect as a whole is not meant to represent this case. We cannot just ignore the apply in the side effects analysis.

A corner-case could be added to allow ignoring it for a 1-point stencil, but again, I'm not sure it really makes sense.

Also, the included filecheck should indeed bufferize the way you have written it, even without any corner case. I can look into getting it to work. But this is not a fix

PapyChacal (approved):

PapyChacal:
Done

n-io:
> The apply is applying a stencil pattern to a whole field. If every point depends on the previous value of its neighbours, you can't both update neighbours before the point and the point before neighbours?

But doesn't the implied store happen only in the stencil.return? If so, the store is the last instruction in the apply body, and I don't see a difference to it being the first instruction immediately after the stencil body, such that I think we only need to analyse instructions that happen between the apply body's stencil.return and stencil.store. At least, I can't think of a counter example. Do we have any tests that break if we don't analyse the body?

xdsl/transforms/stencil_bufferize.py
```diff
@@ -213,6 +213,9 @@ class ApplyStoreFoldPattern(RewritePattern):
     def is_dest_safe(apply: ApplyOp, store: StoreOp) -> bool:
         # Check that the destination is not used between the apply and store.
         dest = store.field
+        start = dest.owner
+        if isinstance(start, Block):
+            start = cast(Operation, start.first_op)
```
n-io, line 218: `start` is unused

PapyChacal: Sure, just trying to make this PR a dialect PR now. Redeleted unused start. Merge at your convenience

PR #3071, opened 2024-08-20
dialects: stencil: Allow loading and storing from a same field.

PapyChacal:
Now that `stencil-bufferize` can deal with those cases, lift this restriction at the stencil dialect level.
Keep the check in the specific old conversion patterns still assuming this property for least surprise. Point user to use `stencil-bufferize`, which if successful makes those operations disappear.

### DIFF
--- a/tests/dialects/test_stencil.py
+++ b/tests/dialects/test_stencil.py
@@ -529,27 +529,6 @@ def test_stencil_store():
     assert store.bounds is bounds
 
 
-def test_stencil_store_load_overlap():
-    temp_type = TempType([(0, 5), (0, 5)], f32)
-    temp_type_ssa_val = TestSSAValue(temp_type)
-
-    field_type = FieldType([(0, 2), (0, 2)], f32)
-    field_type_ssa_val = TestSSAValue(field_type)
-
-    lb = IndexAttr.get(1, 1)
-    ub = IndexAttr.get(64, 64)
-    bounds = StencilBoundsAttr.new((lb, ub))
-
-    load = LoadOp.get(field_type_ssa_val, lb, ub)
-    store = StoreOp.get(temp_type_ssa_val, field_type_ssa_val, bounds)
-
-    with pytest.raises(VerifyException, match="Cannot Load and Store the same field!"):
-        load.verify()
-
-    with pytest.raises(VerifyException, match="Cannot Load and Store the same field!"):
-        store.verify()
-
-
 def test_stencil_index():
     dim = IntAttr(10)
     offset = IndexAttr.get(1)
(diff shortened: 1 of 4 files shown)


tests/dialects/test_stencil.py
```diff
@@ -529,27 +529,6 @@ def test_stencil_store():
...
-    load = LoadOp.get(field_type_ssa_val, lb, ub)
-    store = StoreOp.get(temp_type_ssa_val, field_type_ssa_val, bounds)
-
-    with pytest.raises(VerifyException, match="Cannot Load and Store the same field!"):
-        load.verify()
-
-    with pytest.raises(VerifyException, match="Cannot Load and Store the same field!"):
-        store.verify()
-
-
```
PapyChacal, line 552: Was only testing the specific verification error.
I'm planning to add filechecks to test the new lowering error properly.

tests/filecheck/transforms/convert-stencil-to-ll-mlir.mlir
```diff
@@ -844,3 +844,54 @@ func.func @buffered_combine(%115 : !stencil.field<?x?xf64>) {
...
+      %xyz_1 = arith.addi %xy_1, %z_1 : index
+      stencil.return %xyz_1 : index
+    }
+    stencil.store %95 to %93 (<[0, 0, 0], [64, 64, 64]>) : !stencil.temp<[0,64]x[0,64]x[0,64]xindex> to !stencil.field<[0,64]x[0,64]x[0,64]xindex>
+    func.return
+  }
+
+// CHECK:           "stencil.store"(%2, %0) {"bounds" = #stencil.bounds<[0, 0, 0], [64, 64, 64]>} : (!stencil.temp<[0,64]x[0,64]x[0,64]xindex>, !stencil.field<[0,64]x[0,64]x[0,64]xindex>) -> ()
+// CHECK-NEXT:      ^^^^^^^^^^^^^^^-------------------------------------------------------------------------------------------------------------------------
+// CHECK-NEXT:      | Error while applying pattern: Cannot lower directly if storing to the same field multiple times! Try running `stencil-bufferize` before.
```
georgebisbas, line 874: so, we still keep this, as expected

PapyChacal: Kinda, as a transition; This is now only an error to try and use the old lowering on such a case!
So you can without hacks express it, and in most cases lower it by first running `stencil-bufferize`

PapyChacal: (Where before, it was considered an error to just write it!)

georgebisbas (approved):

PR #3073, opened 2024-08-20
transformations: Enable bufferized `dmp.swap` lowering.

PapyChacal:
This makes `dmp-to-mpi` expected to run right before stencil lowering, as opposed to right after!
Mostly because of my brutal type conversion for stencil.
But I don't think I'll have time for a finer type conversion infrastructure soon, and this makes things work in Devito!


superlopuh (approved):

georgebisbas:
I missed this, thanks!

PR #3074, opened 2024-08-20
CI: Python 3.12

PapyChacal:
Just figured I've been working on Python 3.12 for a while now and everything's fine so why not; 3.13 is planned for October!


superlopuh (approved):
Yeah seems like we should support the full range

PR #3078, opened 2024-08-21
transformations: (licm) add `can_be_hoisted` helper function

RRavikiran66:
Helper function `can_be_hoisted` for Loop invariant code motion. `can_be_hoisted` return True if the operation is invariant to be the Scf.For loop. I think the pytests can be improved a lot, I am not sure how to write pytests properly. So, I created below tests by seeing existing tests. If you can be suggest to good testcases, I will happily include those :)


tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,138 @@
...
+    IntegerAttr,
+    ModuleOp,
+    f32,
+    i32,
+)
+from xdsl.dialects.scf import For, Yield
+from xdsl.ir import Block, BlockArgument, Region
+from xdsl.transforms.loop_invariant_code_motion import can_be_hoisted
+
+index = IndexType()
```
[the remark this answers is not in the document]
superlopuh: My bad, this doesn't exist, probably should

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,138 @@
...
+    # regoin1 = Region([body0, outer_loop_body.block])
+    # # Create the function
+    func_type = FunctionType.from_lists([], [])
+    function = func.FuncOp("invariant_loop_dialect", func_type, region0)
+
+    # Wrap all in a ModuleOp
+    mod = ModuleOp([function])
+    # Assertions to check loop invariant and hoisting
+    for op in mod.body.walk():
+        for regoin in mod.regions:
```
superlopuh, line 102:
```suggestion
        for region in mod.regions:
```
here and in some other places

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,138 @@
...
+    ub = Constant.from_int_and_width(42, i32)
+    step = Constant.from_int_and_width(3, i32)
+
+    op1 = test.TestOp(result_types=[i32])
+    op2 = test.TestOp(result_types=[i32])
+
+    for_op = For(lb, ub, step, [], Block([op1, op2], arg_types=[i32]))
+
+    assert can_be_hoisted(op1, for_op.body)
+    assert can_be_hoisted(op2, for_op.body)
```
superlopuh, line 31: This is great, just that there's no need for the arith ops.
```suggestion
def test_is_loop_dependent_no_dep():
    lb = test.TestOp(result_types=[i32])
    ub = test.TestOp(result_types=[i32])
    step = test.TestOp(result_types=[i32])

    op1 = test.TestOp(result_types=[i32])
    op2 = test.TestOp(result_types=[i32])

    for_op = For(lb, ub, step, [], Block([op1, op2], arg_types=[i32]))

    assert can_be_hoisted(op1, for_op.body)
    assert can_be_hoisted(op2, for_op.body)
```

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def test_for_with_loop_invariant_verify():
```
superlopuh, line 34: Would it be possible to unify the tests below? I think we could get away with a single test, that exercises all the codepaths in the helper function. It would ideally not call any arith operations, and just have `TestOp`s.

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,138 @@
...
+    for_op = For(lower, upper, step, [carried], body)
+
+    for op in for_op.body.walk():
+        for regoin in for_op.regions:
+            # MLIR iterates the for loop's body and hoists operations that
+            # can_be_hoisted. However, the DivUI operation cannot be hoisted immediately
+            # because its operands are still within the loop's body. It can only be
+            # hoisted once the instructions generating those operands have been hoisted.
+            if not (isinstance(op, Yield) or isinstance(op, DivUI)):
+                assert can_be_hoisted(op, regoin)
```
superlopuh, line 61: Functions encapsulate variables, but we want to be able to test them directly, so the other way of building blocks is probably preferable here:
```suggestion
    block = Block(arg_types=(index, index,))
    region = Region(block)
    with ImplicitBuilder(block) as i_arg, carried_arg:
        a = Constant(IntegerAttr.from_int_and_width(1, 32), i32)
        b = Constant(IntegerAttr.from_int_and_width(2, 32), i32)
        # Operations on these constants
        c = Addi(a, b)
        d = Muli(a, b)
        e = DivUI(c, d)
        f = Yield(e)

    for_op = For(lower, upper, step, [carried], region)

    assert can_be_hoisted(region, a)
    assert can_be_hoisted(region, b)
    assert can_be_hoisted(region, c)
    assert can_be_hoisted(region, d)
    assert not can_be_hoisted(region, e)
    assert not can_be_hoisted(region, f)
```

RRavikiran66: I have address your comments

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,92 @@
...
+        op3 = test.TestOp([op1, op2, arg0], result_types=[i32])
+        _f = Yield(op3)
+    for_op = For(lower, upper, step, [carried], region)
+    block0 = Block([a, b, for_op])
+    _region_outer = Region(block0)
+
+    for region in for_op.regions:
+        for op in region.block.walk():
+            if not isinstance(op, scf.Yield):
+                assert can_be_hoisted(op3, region)
```
superlopuh, line 48: There's just a single region, and we have all the ops in scope, so we can just test each of the ops directly
```suggestion
    assert can_be_hoisted(op1, region)
    assert can_be_hoisted(op2, region)
    assert can_be_hoisted(op3, region)
    assert not can_be_hoisted(f, region)
```
This code currently does not test that calling the op with yield returns False

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_invariant_loop_dialect():
```
superlopuh, line 51: Can you please confert this function to the version above? Then we'll be GTG

RRavikiran66: Modified the test cases with your suggestions

superlopuh: I'm sorry, I wasn't very clear, the functions below should be `with` blocks as above, and for all the asserts to be explicit for each operation in the test.
Also, it seems like a test is missing for the case where an inner op has a region with values defined inside it, but can still be moved out.

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,94 @@
...
+        op2 = test.TestOp(result_types=[i32])
+        op3 = test.TestOp([op1, op2, arg0], result_types=[i32])
+        f = Yield(op3)
+    for_op = For(lower, upper, step, [carried], region)
+    block0 = Block([a, b, for_op])
+    _region_outer = Region(block0)
+
+    assert can_be_hoisted(op1, region)
+    assert can_be_hoisted(op2, region)
+    assert can_be_hoisted(op3, region)
```
PapyChacal, line 46: I'm not sure what exactly is wrong, but to my understanding this assertion should fail; `op3` is using `op1` and `op2` which are inside the loop at this point, so `op3` should *not* be considered hoistable.

The point is to check if it is immediately hoistable, and hoist them one by one; i.e., `op3` would be hoistable if `op1` and `op2` were hoisted *before* checking it.

PapyChacal: Just noticed my unclear selection of lines; I'm referring to `can_be_hoisted(op3, region)` specifically; `op1` and `op2` seem right to me

RRavikiran66: You are correct, I have modified the logic and added explicit testcase that Sasha was mentioning.

tests/transforms/test_transform_can_be_hoisted.py
```diff
@@ -0,0 +1,99 @@
...
+
+    region = Region(block1)
+    with ImplicitBuilder(block1) as (_arg1, _):
+        scf.For(ci0, ci10, ci1, [], region_inner)
+
+    for_op = For(co0, co1, coi, [], region)
+    body0 = Block([cf7, cf8, for_op])
+    region_outer = Region(body0)
+    func_type = FunctionType.from_lists([], [])
+    function = func.FuncOp("invariant_loop_dialect", func_type, region_outer)
```
superlopuh, line 91: This is now looking great, but the structure of the module is a little hard to read. The `ImplicitBuilder` regions can be nested, which will make it a little easier to figure out what's going on. Here is what I would suggest to do:

```suggestion
    outer_region = Region([Block()])
    with ImplicitBuilder(outer_region):
        ci32, cf32 = test.TestOp(result_types=[i32, f32]).results
        
        outer_for = For(ci32, ci32, ci32, [], Region(Block(arg_types=[index, index])
        with ImplicitBuilder(outer_for.body) as (_outer_i, _):
            inner_for = For(ci32, ci32, ci32, [], Region(Block(arg_types=[index, index])
            with ImplicitBuilder(inner_for.body) as (_inner_i, _):
                no_args = test.TestOp(result_types=[i32])
                args_in_inner = test.TestOp([no_args], result_types=[i32])
                no_results = test.TestOp([args_in_inner], result_types=[])
```

I removed the arith operation to avoid depending on that dialect, which is not relevant to this test. With this refactor, it is clear that the nested region functionality is not tested. It would be great to add another two testop in the loop, one of which has a region that is self-contained, the other of which depends on something outside of the for loop. I'm also not sure that the nested for loops get us anything.

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -0,0 +1,43 @@
...
+from xdsl.ir import Operation, Region
+from xdsl.traits import IsTerminator
+
+
+def can_be_hoisted(op: Operation, target_region: Region) -> bool | None:
+    """
+    Checks whether the given op can be hoisted by checking that
+    - the op and none of its contained operations depend on values inside of the
+     loop (by means of calling definedOutside).
+    - the op has no side-effects.
```
superlopuh, line 26: This seems to not be tested in this function, let's remove it?

tests/transforms/test_transform_can_be_hoisted.py
superlopuh: In general, could you please check that each of the conditions in the `can_be_hoisted` function are tested? I'm not 100% sure that this is the last missing example.

tests/transforms/test_transform_can_be_hoisted.py
PapyChacal: This now seems sensible to me, thanks!

PapyChacal (approved):
There are some possible PR improvements as mentionned by @superlopuh already, but the content now seems good to me!

xdsl/transforms/loop_invariant_code_motion.py
RRavikiran66: Done

tests/transforms/test_transform_can_be_hoisted.py
RRavikiran66: I have made the changes

tests/transforms/test_transform_can_be_hoisted.py
superlopuh, line 1: I would recommend renaming this to `test_loop_invariant_code_motion`, to make it a bit easier to search locally

RRavikiran66: Will do

xdsl/transforms/loop_invariant_code_motion.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Walk the nested operations and check that all used values are either
+    # defined outside of the loop or in a nested region, but not at the level of
+    # the loop body.
+    for child in op.walk():
+        for operand in child.operands:
+            operand_owner = operand.owner
+            assert operand_owner is not None
+            if op.is_ancestor(operand_owner):
+                continue
```
superlopuh, line 39: I just ran the test locally, and this line is never hit.

superlopuh: I pushed my suggested changes to `sasha/licm/can-be-hoisted`. I added breakpoints to every line in this function, and added test cases that would hit every one of them. I also removed scf and func dialects from the test, since they're not used by this helper. I would recommend reviewing those changes, and merging them to your branch if you're happy with them.

superlopuh: Then we can move onto the next step

RRavikiran66: I reused your testcases, Thank you Sasha. Can I move to add IsMemoryEffectFree helper function?

superlopuh (approved):
Fantastic! Thank you very much for iterating on this, it was more complex than I thought

PR #3084, opened 2024-08-22
transformations: (csl-stencil-bufferize) Fold `csl_stencil.access` that have no effect

n-io:
The `csl-stencil-bufferize` pass lowers `stencil.field` and `tensor` types to `memref`.

The type conversion pass creates the scenario that some `csl_stencil.access` ops are equal input and output types, for instance, `(memref<512xf32>) -> memref<512xf32>`. This only happens for ops accessing own data. In this case, the access op has no effect and can safely be folded away.


dk949 (approved):
LGTM, but I'm not 100% sure this is the right place to do this transformation. This rewrite ends up _either_ transforming the access _or_ converting it to a `to_tensor`

n-io:
It only folds accesses that have no effect.

PR #3089, opened 2024-08-23
dialects: (stablehlo) add stablehlo.precision

superlopuh:
One of the attributes necessary to express dot products

CC @efferifick

### DIFF
--- a/xdsl/dialects/stablehlo.py
+++ b/xdsl/dialects/stablehlo.py
@@ -10,11 +10,19 @@ import abc
 from typing import Annotated, TypeAlias, cast
 
 from xdsl.dialects.builtin import AnyTensorType, DenseArrayBase, IntegerType, TensorType
-from xdsl.ir import Attribute, Dialect, SSAValue
+from xdsl.ir import (
+    Attribute,
+    Dialect,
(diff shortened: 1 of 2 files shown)


xdsl/dialects/stablehlo.py
```diff
@@ -43,6 +51,32 @@ def __init__(
...
+
+# region Attributes
+
+
+class Precision(StrEnum):
+    """
+    XLA precision for an operand. Has backend specific meaning.
+    """
+
+    DEFAULT = "DEFAULT"
```
alexarice, line 64: I believe you could use `auto()` here

alexarice (approved):

xdsl/dialects/stablehlo.py
superlopuh: For some reason it made the string lowercase and my test failed

superlopuh: `# |                                    Expected `default`, `HIGH`, or `HIGHEST`.`

alexarice: Yep, documentation seems to say that `auto` lowercases

PR #3093, opened 2024-08-24
dialects: (stablehlo) add stablehlo.token

efferifick:
Adds stablehlo.token type.


efferifick:
I am not sure if `stablehlo.token` should be parametrized attribute. All `TypeAttribute`s must be either `ParametrizedAttribute` or `Data` so I just picked one. Let me know if I should remove this restriction from `TypeAttribute` or if `stablehlo.token` is fine as `ParametrizedAttribute`.

superlopuh (approved):

superlopuh:
This is perfect, you did the right thing!

PR #3094, opened 2024-08-24
dialects: (vector) add pure trait to vector.broadcast and vector.fma

### LINKED ISSUE #3019: `vector.broadcast` and `vector.fma` are missing the Pure trait
`vector.broadcast` and `vector.fma` are missing the Pure trait

knickish:
Add `Pure` trait to `vector.broadcast` and `vector.fma`, fixes #3019

### DIFF
--- a/tests/filecheck/dialects/vector/vector_pure_ops.mlir
+++ b/tests/filecheck/dialects/vector/vector_pure_ops.mlir
@@ -1,0 +1,12 @@
+// RUN: xdsl-opt %s --verify-diagnostics -p cse | filecheck %s
+
+%m0, %i0 = "test.op"() : () -> (memref<4x4xindex>, index)
+%load = "vector.load"(%m0, %i0, %i0) : (memref<4x4xindex>, index, index) -> vector<2xindex>
+"vector.store"(%load, %m0, %i0, %i0) : (vector<2xindex>, memref<4x4xindex>, index, index) -> ()
+%broadcast = "vector.broadcast"(%i0) : (index) -> vector<1xindex>
+%fma = "vector.fma"(%load, %load, %load) : (vector<2xindex>, vector<2xindex>, vector<2xindex>) -> vector<2xindex>
+
+/// Check that unused results from vector.broadcast and vector.fma are eliminated
+// CHECK:       %m0, %i0 = "test.op"() : () -> (memref<4x4xindex>, index) 
+// CHECK-NEXT:  %load = "vector.load"(%m0, %i0, %i0) : (memref<4x4xindex>, index, index) -> vector<2xindex>
+// CHECK-NEXT:  "vector.store"(%load, %m0, %i0, %i0) : (vector<2xindex>, memref<4x4xindex>, index, index) -> ()
--- a/xdsl/dialects/vector.py
+++ b/xdsl/dialects/vector.py
@@ -22,6 +22,7 @@ from xdsl.irdl import (
     result_def,
     var_operand_def,
 )
+from xdsl.traits import Pure
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import assert_isa, isa
 
@@ -89,6 +90,7 @@ class Broadcast(IRDLOperation):
     name = "vector.broadcast"
     source: Operand = operand_def(AnyAttr())
     vector: OpResult = result_def(VectorType)
+    traits = frozenset((Pure(),))
 
     def verify_(self):
         assert isa(self.vector.type, VectorType[Attribute])
@@ -113,6 +115,7 @@ class FMA(IRDLOperation):
     rhs: Operand = operand_def(VectorType)
     acc: Operand = operand_def(VectorType)
     res: OpResult = result_def(VectorType)
+    traits = frozenset((Pure(),))
 
     def verify_(self):
         assert isa(self.lhs.type, VectorType[Attribute])


superlopuh:
Thank you for this! Could you please add a minimal filecheck test that tests that these get dce'd away? Should be a minimal test with `"test.op"`s that produce the required inputs for the ops, and no uses, and that checks that the operations are removed after calling the `dce` pass

webmiche:
Honestly, I am surprised that the broadcast op would be pure. If MLIR models it this way, it's fine to have it here aswell, I guess.

superlopuh:
https://mlir.llvm.org/docs/Dialects/Vector/#vectorbroadcast-vectorbroadcastop
You're right

knickish:
Is that indicated by the `Effects: MemoryEffects::Effect{}` line? Looks like nearly all of them have that

superlopuh:
Yep! We now have a similar trait in xDSL, if you could add the `NoMemoryEffect` trait instead of Pure that would be ideal. And then also a filecheck test that check that they still get removed by `dce`

PapyChacal:
@webmiche @superlopuh As I discovered similarly on #3103 (resolved discussion); MLIR tend to not report some AlwaysSpeculatable operations properly in their documentation.

See:
- https://github.com/llvm/llvm-project/blob/98e674c9f16d677d95c67bc130e267fae331e43c/mlir/include/mlir/Dialect/Vector/IR/VectorOps.td#L348-L349
- https://github.com/llvm/llvm-project/blob/98e674c9f16d677d95c67bc130e267fae331e43c/mlir/include/mlir/Dialect/Vector/IR/VectorOps.td#L597-L598

Those ops are indeed Pure and @knickish was right :slightly_smiling_face:

PapyChacal (approved):
(Plus, I personally don't think we should check traits' behaviour on dialect operations everytime, those are tested on their own through test operations, just my 2 cents on that front)

superlopuh:
We don't need to go into a full debate here but to me they're two separate things, the trait is just one way to achieve the desired behaviour, and are an implementation detail, and I think it's worth having a test for each dialect to make sure that the operations have the expected behaviour, in this case that the operations can be eliminated if their results are not used.

knickish:
Ok, so leave as `Pure` and add the filecheck tests, correct?

superlopuh:
Yes, please :)

knickish:
I gave the test a shot, not sure if that's exactly what it's supposed to be though

superlopuh (approved):

tests/filecheck/dialects/vector/vector_pure_ops.mlir
```diff
@@ -0,0 +1,21 @@
...
+
+builtin.module {
+  func.func private @vector_cse_test(%0 : memref<4x4xindex>, %1 : vector<1xi1>, %2 : index) {
+    %3 = "vector.load"(%0, %2, %2) : (memref<4x4xindex>, index, index) -> vector<2xindex>
+    "vector.store"(%3, %0, %2, %2) : (vector<2xindex>, memref<4x4xindex>, index, index) -> ()
+    %4 = "vector.broadcast"(%2) : (index) -> vector<1xindex>
+    %5 = "vector.fma"(%3, %3, %3) : (vector<2xindex>, vector<2xindex>, vector<2xindex>) -> vector<2xindex>
+    func.return
+  }
+}
```
superlopuh, line 11: nit:
```suggestion
%m0, %i0 = "test.op"() : () -> (memref<4x4xindex>, index)
%load = "vector.load"(%m0, %i0, %i0) : (memref<4x4xindex>, index, index) -> vector<2xindex>
"vector.store"(%load, %m0, %i0, %i0) : (vector<2xindex>, memref<4x4xindex>, index, index) -> ()
%broadcast = "vector.broadcast"(%i0) : (index) -> vector<1xindex>
%fma = "vector.fma"(%load, %load, %load) : (vector<2xindex>, vector<2xindex>, vector<2xindex>) -> vector<2xindex>
```
1. we prefer to use test.op to create fake values for tests, and generally minimise the dependencies on other dialects to avoid spurious test breaks
2. adding names to values makes the naming more stable, if you add an op in the middle it doesn't mess up the check lines below
3. the `builtin.module` is implied if not present, makes the test a little easier to read.

superlopuh: I think it's ok to merge as-is, bit making these changes would make the test a little more future-proof, even if it tests the same functionality

knickish: The naming is definitely clearer. Do you want me to rebase this change in to the last commit?

knickish: Ah I see, need to update the checks anyway. Will do that and then squash the test changes together

tests/filecheck/dialects/vector/vector_pure_ops.mlir
```diff
@@ -0,0 +1,17 @@
...
+%m0, %i0 = "test.op"() : () -> (memref<4x4xindex>, index)
+%load = "vector.load"(%m0, %i0, %i0) : (memref<4x4xindex>, index, index) -> vector<2xindex>
+"vector.store"(%load, %m0, %i0, %i0) : (vector<2xindex>, memref<4x4xindex>, index, index) -> ()
+%broadcast = "vector.broadcast"(%i0) : (index) -> vector<1xindex>
+%fma = "vector.fma"(%load, %load, %load) : (vector<2xindex>, vector<2xindex>, vector<2xindex>) -> vector<2xindex>
+
+/// Check that unused results from vector.broadcast and vector.fma are eliminated
+
+// CHECK:      builtin.module {
+// CHECK-NEXT:   func.func private @vector_cse_test(%0 : memref<4x4xindex>, %1 : vector<1xi1>, %2 : index) {
```
superlopuh, line 12: ok but now the func is gone so one more change :)

tests/filecheck/dialects/vector/vector_pure_ops.mlir
knickish: updated the checks and squashed all the test commits into one

superlopuh:
Thank you!

PR #3098, opened 2024-08-26
ci: Enable pip caching

n-io:


superlopuh:
I'm going to re-run the CI action to see what the impact is. Currently the 3.10 one took 23 minutes, presumably to cache things, and 3.11 took 30s.

superlopuh (approved):
OK it's not any faster but at least it's not redownloading the world on every PR. LGTM! Thank you for this

webmiche (approved):

n-io:
My high-score was 95m yesterday.

Let me try to add explicit caching for setup.py ([link](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)) and see if it has any effect.

n-io:
I think this looks good to me, I've tried re-running non-cached jobs on a different PR, but the network seems to be working fine today so it's harder to test.

superlopuh (changes requested):
Oh actually no, sorry, we specify all our dependencies in the pyproject.toml, not requirements or in setup.py

superlopuh:
> Does that mean that it may not use some updated dependency if not strictly upgraded by the dependabot? Havr you checked if this PR causes a diff in the used versions ?

My understanding is that if `check-latest` is not set to `true`, then it will use the cached version, yes. But that seems ok to me? For the dependencies where this matters we merge the dependabot things pretty quickly, for the ones it doesn't I'm personally happy with the CI passing with at least one version.

n-io:
> Oh actually no, sorry, we specify all our dependencies in the pyproject.toml, not requirements or in setup.py

Yes that's very much correct, but the top-level command executed is `pip install -r requirements.txt`, which has an `.e[...]` clause that I understand is handled by setup.py, at which point I'm not fully clear how it gets down to the specified dependencies in pyproject.toml.

As for whether it works and how exactly it works, I've compared some of the build logs to non-cached build logs:
old ("Install the package locally")
```
Collecting immutabledict<4.2.1 (from xdsl==0+untagged.3.g28dc517)
  Downloading immutabledict-4.2.0-py3-none-any.whl.metadata (3.4 kB)
```

new (this PR):
```
Collecting immutabledict<4.2.1 (from xdsl==0+untagged.3.gaecc4c1)
  Using cached immutabledict-4.2.0-py3-none-any.whl.metadata (3.4 kB)
```

From what I understand and from what the logs seem to say, this should purely be a download cache.

superlopuh:
Isn't the idea of that flag to specify which files to watch for changes to update the cache? It just seems weird to me that pyproject.toml isn't in that list, but maybe I'm misunderstanding something.

n-io:
This is actually not fully clear to me either. In the one place where it is mentioned in the docs (see [caching poetry dependencies](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)) it seems to suggest it knows what's going on and doesn't need to specify the pyproject.toml explicitly. In comparison to the first build of this PR, the logs now report that it's hitting the cache as expected. It looks like it works just specifying the top-level file.

superlopuh (approved):

superlopuh:
I saw the first version of this PR also hit the cache. If it works with both versions then I'm all for it. TBH I'm thinking of moving all our things to `uv` in the nearish future, so this intermediate step seems like a improvement until then, but I wouldn't spend too much time on it.

PR #3099, opened 2024-08-26
frontend: (stablehlo) add ir generation from jaxpr

shreya-um:
This will let us create an xDSL module from a add function jitted with JAX.

### DIFF
--- a/xdsl/frontend/jaxpr/__init__.py
+++ b/xdsl/frontend/jaxpr/__init__.py
@@ -1,15 +1,36 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from typing import NoReturn
 
 from jax._src.core import ClosedJaxpr
 
-from xdsl.builder import Builder
+from xdsl.builder import Builder, ImplicitBuilder
 from xdsl.dialects.builtin import FunctionType, ModuleOp, TensorType, f32
 from xdsl.dialects.func import FuncOp, Return
-from xdsl.ir import Block, Region
+from xdsl.ir import Block, Region, SSAValue
+
+from .primitive_builder import PrimitiveBuilder
 
 
 class IRGenError(Exception):
     pass
+
+
+@dataclass
+class ScopedSymbolTable:
+    "A mapping from variable names to SSAValues, append-only"
+
+    table: dict[str, SSAValue] = field(default_factory=dict)
+
+    def __contains__(self, __o: object) -> bool:
+        return __o in self.table
+
+    def __getitem__(self, __key: str) -> SSAValue:
+        return self.table[__key]
+
+    def __setitem__(self, __key: str, __value: SSAValue) -> None:
+        if __key in self:
+            raise AssertionError(f"Cannot add value for key {__key} in scope {self}")
+        self.table[__key] = __value
 
 
 @dataclass(init=False)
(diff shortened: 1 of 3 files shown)


xdsl/frontend/jaxpr/primitive_builder.py
```diff
@@ -0,0 +1,33 @@
...
+
+from jax._src.core import JaxprEqn
+
+from xdsl.dialects.stablehlo import AddOp
+from xdsl.ir import Operation, SSAValue
+
+
+class PrimitiveBuilder:
+    BUILDER_BY_PRIMITIVE: ClassVar[
+        dict[str, Callable[[JaxprEqn, tuple[SSAValue, ...]], Operation]]
```
shreya-um, line 12: as “add” is not primitive in Jax, I changed `dict[jax_primitive, …. ]` to `dict[str, …. ]`

xdsl/frontend/jaxpr/__init__.py
```diff
@@ -1,17 +1,38 @@
...
+    def __contains__(self, __o: object) -> bool:
+        return __o in self.table
+
+    def __getitem__(self, __key: str) -> SSAValue:
+        return self.table[__key]
+
+    def __setitem__(self, __key: str, __value: SSAValue) -> None:
+        if __key in self:
+            raise AssertionError(f"Cannot add value for key {__key} in scope {self}")
+        self.table[__key] = __value
```
superlopuh, line 33: if we're reusing this helper from Toy, could you please first move this into `utils` in a separate PR, add some tests, and then use that version here? It feels like something we'll be reusing in the future, and MLIR has a version for clients to use.

xdsl/frontend/jaxpr/__init__.py
```diff
@@ -26,58 +47,105 @@ class IRGen:
...
+        """
+        Declare a variable in the current scope, return success if the variable
+        wasn't declared yet."""
+        assert self.symbol_table is not None
+        if var in self.symbol_table:
+            return False
+        self.symbol_table[var] = value
+        return True
+
+    def ir_gen_variable_expr(self, name: str) -> SSAValue:  # pyright: ignore[reportReturnType]
```
superlopuh, line 81:
```suggestion
    def ir_gen_variable_expr(self, name: str) -> SSAValue: 
```
Why this ignore?

xdsl/frontend/jaxpr/__init__.py
```diff
@@ -26,58 +47,105 @@ class IRGen:
...
+        """
+        This is a reference to a variable in an expression. The variable is
+        expected to have been declared and so should have a value in the symbol
+        table, otherwise emit an error and return nullptr."""
+        assert self.symbol_table is not None
+        try:
+            variable = self.symbol_table[name]
+            return variable
+        except Exception as e:
```
superlopuh, line 90: This is bad practice in general, we should catch the exact Exception we expect

xdsl/frontend/jaxpr/__init__.py
```diff
@@ -26,58 +47,105 @@ class IRGen:
...
+            self.error(f"error: unknown variable `{name}`", e)
```
superlopuh, line 91: this is also not great, we're just silently ignoring errors, let's propagate them with a new error instead, or just let the compiler crash with the original error

xdsl/frontend/jaxpr/primitive_builder.py
superlopuh: It would be great to try and find exactly what Jax uses for this. Strings are quite dangerous, because `add` is so generic, what if someone comes and calls their expression `add` too? Ideally we'd want to go from some notion of type or function identity. I'm happy to help with this

xdsl/frontend/jaxpr/primitive_builder.py
```diff
@@ -0,0 +1,33 @@
...
+        if primitive.name not in cls.BUILDER_BY_PRIMITIVE:
+            raise ValueError(f"No builder for primitive: {primitive}")
+
+        builder = cls.BUILDER_BY_PRIMITIVE[primitive.name]
+        res = builder(eqn, args)
+        return res
+
+
+class JaxPrimitiveBuilder(PrimitiveBuilder):
+    pass
```
superlopuh, line 33: why this separation?

xdsl/frontend/jaxpr/__init__.py
```diff
@@ -26,58 +47,105 @@ class IRGen:
...
-        outputVars = jaxpr.jaxpr.outvars[0]
+        inputVars = jaxpr.jaxpr.invars
+        outputVars = jaxpr.jaxpr.outvars
+
+        for i in range(1, len(inputVars)):
+            assert inputVars[i].aval.shape == inputVars[i - 1].aval.shape
 
         input_types = [
-            TensorType(f32, [inputVars.aval.size])
+            TensorType(f32, list(inputVars[0].aval.shape))
```
superlopuh, line 118: Does this work?
```suggestion
            TensorType(f32, inputVars[0].aval.shape)
```

xdsl/frontend/jaxpr/primitive_builder.py
shreya-um: in Jax add_p is a primitive, add is just a wrapper function around it. Using lax.add_p I can still use dict[jax_primitive, …. ]. :)

xdsl/frontend/jaxpr/primitive_builder.py
shreya-um: Its not required, Ill update

xdsl/frontend/jaxpr/primitive_builder.py
```diff
@@ -0,0 +1,30 @@
...
+
+class PrimitiveBuilder:
+    BUILDER_BY_PRIMITIVE: ClassVar[
+        dict[Primitive, Callable[[JaxprEqn, tuple[SSAValue, ...]], Operation]]
+    ] = {}
+
+    def __init__(self) -> None:
+        super().__init_subclass__()
+        # register jaxpr primitives
+        self.BUILDER_BY_PRIMITIVE[lax.add_p] = lambda _, args: AddOp(args[0], args[1])  # pyright: ignore[reportUnknownMemberType]
```
superlopuh, line 19: Why the pyright ignore?

xdsl/frontend/jaxpr/primitive_builder.py
```diff
@@ -0,0 +1,30 @@
...
+        print(f"Registered in {self.__class__}: {self.BUILDER_BY_PRIMITIVE}")
```
superlopuh, line 20:
```suggestion
```

xdsl/frontend/jaxpr/__init__.py
```diff
@@ -26,58 +47,105 @@ class IRGen:
...
-            TensorType(f32, [inputVars.aval.size])
+            TensorType(f32, inputVars[0].aval.shape)
             for _ in range(len(jaxpr.jaxpr.invars))
         ]
+
         output_types = [
-            TensorType(f32, [outputVars.aval.size])
+            TensorType(f32, outputVars[0].aval.shape)
             for _ in range(len(jaxpr.jaxpr.outvars))
         ]
```
superlopuh, line 125: Shouldn't these be the shapes for each invar, not always the first one?

shreya-um: As of now Im asserting that all invars are with same shape (assert inputVars[i].aval.shape == inputVars[i - 1].aval.shape). and Im using 1st invars to get shape.

superlopuh: If they're equivalent, then why not use each shape since it seems simpler and more future-proof?

shreya-um: okk. Ill update

xdsl/frontend/jaxpr/__init__.py
superlopuh: I've started doing this locally, as I'd like it to be in a specific place, and potentially reused in the interpreter. Will open a PR today.

superlopuh: #3355

superlopuh:
@shreya-um, let's reopen this when you have time to iterate, always happy to do this, just ping me :)

PR #3101, opened 2024-08-27
transformations: (csl-stencil-to-csl-wrapper) Generate symbol export and no memref.globals

n-io:


xdsl/dialects/csl/csl.py
```diff
@@ -1445,6 +1446,29 @@ class SymbolExportOp(IRDLOperation):
 
     type = prop_def(base(PtrType) | base(FunctionType))
 
+    def __init__(self, sym_name: str | StringAttr, type_or_op: SSAValue | FunctionType):
+        """
+        Accepts one of:
+          1. var_name + PtrType + operand
+          2. var_name + FunctionType
```
dk949, line 1453: This was a bit confusing, as the function accepts 2 arguments in each case, but the second argument can be either an ssa value or a function type

xdsl/dialects/csl/csl.py
```diff
@@ -1445,6 +1446,29 @@ class SymbolExportOp(IRDLOperation):
...
+        """
+        var_name: StringAttr | SymbolRefAttr = (
```
dk949, line 1455: I don't really understand how this can be a `SymbolRefAttr`

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -144,43 +144,64 @@ def _translate_function_args(
...
+                                    csl.PtrKindAttr(csl.PtrKind.MANY),
+                                    csl.PtrConstAttr(csl.PtrConst.VAR),
+                                ]
+                            )
+                        ],
                     )
                 )
-                arg_ops.append(get_global_op := memref.GetGlobal(arg_name, arg_t))
-                arg_ops.append(
+                export_ops.append(csl.SymbolExportOp(arg_name, SSAValue.get(address)))
```
dk949, line 173: Using the name hint can easily break. `SymbolExportOp` will export the symbol using the name given to it, but this is not necessarily the name the symbol will be printed with by the printer

xdsl/dialects/csl/csl.py
n-io: A few lines below there is a conversion to SymbolRefAttr. There is some complicated verification logic happening, that I sought to simplify in the constructor.

xdsl/dialects/csl/csl.py
n-io: Sorry, this should be what the class accepts. The init simplifies this.

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: Is there any alternative?

dk949: Some suggestions:
1. We could give `SymbolExportOp` the SSA value instead, so that in the printer it definitely uses the same name that the value will be printed with.
2. Another option is to give a symbol name to `csl.zeros` and `csl.constants`, but I don't really like that either.
3. A `csl.exported_array` (or something along those lines) which contains all functionality for defining and exporting arrays (we'd also then need some attribute on functions to make those exported).

All these options make it basically impossible to know the name of the symbol for interfacing from the host, but I think that's something to be resolved at some other time.

Option 3 is I think too much of a departure from the semantics of the CSL language (also a lot of work)

Option 2 would make creating new arrays needlessly difficult since not all arrays need to be exported (though I suppose we can make the symbol optional). This also complicates printing and does not escape from the problem of having to figure out what the symbol should be in the first place

Option 1 is the easiest to implement, but likely makes least sense from the point of view of the user accessing the exported symbols from the host.

I personally vote Option 1. I think it will cause the least number of changes and will be the easiest to implement. More suggestions are welcome :slightly_smiling_face: 

(I think this may have been one of the reasons I used a `memref.global` initially, but its hard to remember now :sweat_smile:)

n-io: Option 1 sounds good - we actually do give it the symbol, iff it's a pointer rather than a function.  The one advantage of 'sym_name' properties is that they can be persisted when invoking mlir-opt, whereas name hints are all re-named.

xdsl/dialects/csl/csl.py
n-io: Updated

dk949 (commented):
~See [#r1733265639](https://github.com/xdslproject/xdsl/pull/3101/files#r1733265639) ~ resolved

dk949 (approved):
All looks good

PR #3102, opened 2024-08-27
core: Implement speculability trait and helper.

PapyChacal:
It is way simpler than side effects, as an op is either speculatable or is not!

Similarly to what was decided on side effects, I only changed that MLIR has an extra return value signifying "recursively speculatable", where here it is handled by an implementation of the trait, rather than in user code.
This prevents some amount of ninja-coding in said user code, but also a lot of ninja headaches.


superlopuh (commented):
This is great! Seems like it's missing tests ;)

superlopuh (approved):

PR #3103, opened 2024-08-27
transformations: Implement `control-flow-hoist`

PapyChacal:
Stacked on:
- #3102 

A couple of splittable but minor changes:
- Update on some dialects for required traits
- Make AffineEx hashable (as all attributes are expected to be! I just randomly crashed some CSE I was playing around with otherwise)

Just a simple and naive pass hoisting everything from region branches.

Thinking a bit, I realized this is all I missed for some slightly tricky cases to work the way I want in MLIR; see filecheck for more!

In a nutshell, MLIR provides some nice bits already, like `loop-invariant-code-motion` (Get stuff out of loop) and `control-flow-sink` (Nest stuff in branches) which are locally okay and smart, but contradict sometimes leading to frustrating things like, to reduce it as much as possible:

```mlir
func.func @nested_loop_invariant(%n : index) {
    %0 = arith.constant 0 : index
    %1 = arith.constant 1 : index
    %100 = arith.constant 100 : index
    scf.for %i = %0 to %100 step %1 {
        %cond = "test.op"() : () -> (i1)
        %thing = "scf.if"(%cond)  ({
            %n100 = arith.muli %n, %100 : index 
            scf.yield %n100 :index
        }, {
            scf.yield %n : index
        }) : (i1) -> index
        "test.op"(%thing) : (index) -> ()
        scf.yield
    }
    return
}
```

From the `scf.if`'s perspective, one should really keep the `arith.muli` nested. What a waste if you compute it and don't take this branch!
Thus from the loop's perspective, it's not as easy as hoisting an operation from its body and is not done.

So this pass does a typically-bad thing to help get out of this kind of local-minima frustration. Just naively hoist everything out of the conditionals to expose the loop-invariant bits, and sink the rest back if desired!

It should also help if someone wants to go from branches to precomputation + select, as I remember was a cool shader trick some amount of years ago at least :yum:

### DIFF
--- a/xdsl/dialects/affine.py
+++ b/xdsl/dialects/affine.py
@@ -38,7 +38,12 @@ from xdsl.irdl import (
 )
 from xdsl.parser import Parser
 from xdsl.printer import Printer
-from xdsl.traits import IsTerminator, Pure
+from xdsl.traits import (
+    IsTerminator,
+    Pure,
+    RecursivelySpeculatable,
+    RecursiveMemoryEffect,
+)
 from xdsl.utils.exceptions import VerifyException
 
 
@@ -205,6 +210,8 @@ class If(IRDLOperation):
     then_region = region_def("single_block")
     else_region = region_def()
 
+    traits = frozenset([RecursiveMemoryEffect(), RecursivelySpeculatable()])
+
 
 @irdl_op_definition
 class ParallelOp(IRDLOperation):
@@ -350,7 +357,7 @@ class Yield(IRDLOperation):
     name = "affine.yield"
     arguments: VarOperand = var_operand_def(AnyAttr())
 
-    traits = frozenset([IsTerminator()])
+    traits = frozenset([IsTerminator(), Pure()])
 
     @staticmethod
     def get(*operands: SSAValue | Operation) -> Yield:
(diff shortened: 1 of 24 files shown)


xdsl/dialects/affine.py
```diff
@@ -350,7 +357,7 @@ class Yield(IRDLOperation):
     name = "affine.yield"
     arguments: VarOperand = var_operand_def(AnyAttr())
 
-    traits = frozenset([IsTerminator()])
+    traits = frozenset([IsTerminator(), Pure()])
```
superlopuh, line 360: I see on the MLIR page that it's ConditionallySpeculatable and NoMemoryEffect, is that the same as Pure, effectively?

superlopuh (approved):

xdsl/dialects/affine.py
PapyChacal: Ah, no, Pure is AlwaysSpeculatable and NoMemoryEffect
Thanks, will have a quick look at the speculatability implementation there 🤔

PapyChacal: Eeeerm
It is not marked as Pure in the docs: https://mlir.llvm.org/docs/Dialects/Affine/#affineyield-affineaffineyieldop
But it is Pure as per the code: https://github.com/llvm/llvm-project/blob/9cf052df90418697c05810f69c588064f7b3ce71/mlir/include/mlir/Dialect/Affine/IR/AffineOps.td#L901
:melting_face: 

I'd just consider it pure myself. does that sound right to you?

PR #3107, opened 2024-08-28
transformations: from `csl_wrapper` to `csl` modules

dk949:
Transforming `csl_wrapper` to the two `csl.module`s


xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,144 @@
...
+
+    name = "csl-wrapper-to-csl"
+    prog_name: str
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        module_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    RemoveUnusedYield(),
+                    ExtractLayoutProgramModule(prog_name=self.prog_name),
```
n-io: You could read this from `csl_wrapper.module.program_name`

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,144 @@
...
+        if op.parent_region() == parent.program_module:
+            rewriter.erase_matched_op()
+
+
+@dataclass(frozen=True)
+class ExtractLayoutProgramModule(RewritePattern):
+    prog_name: str
+
+    @staticmethod
+    def replace_arg(
```
n-io: this seems unused

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,144 @@
...
+
+        assert isa(yield_op := op.layout_module.block.last_op, csl_wrapper.YieldOp)
+        yield_op.detach()
+
+        with ImplicitBuilder(module_block):
+            const_0 = arith.Constant.from_int_and_width(0, builtin.IndexType())
+            const_1 = arith.Constant.from_int_and_width(1, builtin.IndexType())
+
+            const_width = arith.Constant(op.width)
+            param_width = csl.ParamOp("width", op.width.type, const_width)
```
n-io, line 83: Having separate `const_width` and `param_width` values means that the values can diverge. For instance, csl.param values have a default but can be passed a different value from command line, which would not affect the value in const_width. Best to remove const_width and only use param_width, same for *_height

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,205 @@
...
+            const_height = arith.Constant(op.height)
+            param_height = csl.ParamOp("height", op.height.type, const_height)
+
+            new_args = _collect_params(op)
+
+            layout = csl.LayoutOp(Region())
+            with ImplicitBuilder(layout.body.block):
+                scf.For(
+                    lb=const_0,
+                    ub=const_width,
```
n-io, line 94:
```suggestion
                    ub=param_width,
```

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,205 @@
...
+                    step=const_1,
+                    iter_args=[x],
+                    body=outer_loop_block,
+                )
+
+                with ImplicitBuilder(outer_loop_block):
+                    scf.For(
+                        lb=const_0,
+                        ub=const_height,
```
n-io, line 103:
```suggestion
                        ub=param_height,
```

xdsl/transforms/csl_wrapper_to_csl.py
n-io: Ok actually the confusion might have been the use of const_width in the loop.

xdsl/transforms/csl_wrapper_to_csl.py
dk949: :+1: 

I did keep the `prog_name` in as a back up, since `op.program_name` can be `None`, but I'm very much not marries to it

xdsl/transforms/csl_wrapper_to_csl.py
dk949: Removed

xdsl/transforms/csl_wrapper_to_csl.py
dk949: Fixed

xdsl/transforms/csl_wrapper_to_csl.py
n-io: Given we have it already, my preference would be not to have it as a command line arg unless there's genuine need for it.

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,242 @@
...
+        inner_loop_block = Block()
+        inner_loop_block.insert_arg(builtin.IntegerType(16), 0)
+        y = inner_loop_block.args[0]
+
+        new_args = list[SSAValue]()
+
+        assert isa(yield_op := op.layout_module.block.last_op, csl_wrapper.YieldOp)
+        rewriter.erase_op(yield_op)
+
+        with ImplicitBuilder(module_block):
```
n-io, line 99:
```suggestion
        with ImplicitBuilder(module_block := Block()):
```

Some of these can be inlined, also inner/outer loop block. The block arg types can be passed directly as params.

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -0,0 +1,242 @@
...
+        layout_mod = csl.CslModuleOp(
+            regions=[Region(module_block)],
+            properties={"kind": csl.ModuleKindAttr(csl.ModuleKind.LAYOUT)},
+            attributes={"sym_name": builtin.StringAttr(f"{prog_name}_layout")},
+        )
+
+        _add_to_toplevel(rewriter, op, layout_mod)
+
+
+# program: width, height, *params, *layout-yields
```
n-io, line 161: Commented code

n-io (approved):

PR #3113, opened 2024-08-29
[WIP] Experiments in e-graphs in IR

superlopuh:
CC @tobiasgrosser @jianyicheng


xdsl/interpreters/experimental/eqsat_pdl.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+@dataclass
+class EqsatPDLMatcher(PDLMatcher):
+    def match_operand(
+        self, ssa_val: SSAValue, pdl_op: pdl.OperandOp, xdsl_val: SSAValue
+    ):
+        owner = xdsl_val.owner
+        assert isinstance(owner, eqsat.EClassOp)
+        assert (
```
jianyicheng, line 26: I guess this will no longer be true after the first rewrite, i.e. an eclass may have more than one operand?

jianyicheng:
Not sure if I am looking at the right PR but it seems there is no test on eqsatpdl... Is this expected? @superlopuh Happy to write one by myself if there isn't one.

superlopuh:
This is the everything PR, it would definitely be useful to add tests, but especially in the PRs that take bits of code from here that we think are ready and merge to main

superlopuh:
Oh the one caveat I forgot to mention about pushing to this branch is that the tests should always pass so that we can all build on it, can you please fix the tests here?

xdsl/dialects/pdl_interp.py
```diff
@@ -0,0 +1,508 @@
...
+    name = "pdl_interp.get_attribute"
+    constraint_name = prop_def(StringAttr, prop_name="name")
+    input_op = operand_def(OperationType)
+    value = result_def(Attribute)
+
+    def __init__(self, name: str | StringAttr, input_op: SSAValue) -> None:
+        if isinstance(name, str):
+            name = StringAttr(name)
+        super().__init__(
+            operands=[input_op], properties={"name": name}, result_types=[Attribute]
```
superlopuh, line 228:
```suggestion
            operands=[input_op], properties={"name": name}, result_types=[AttributeType()]
```

xdsl/dialects/pdl_interp.py
```diff
@@ -0,0 +1,508 @@
...
+@irdl_op_definition
+class GetValueTypeOp(IRDLOperation):
+    """
+    https://mlir.llvm.org/docs/Dialects/PDLInterpOps/#pdl_interpget_value_type-pdl_interpgetvaluetypeop
+    """
+
+    name = "pdl_interp.get_value_type"
+    value = operand_def(
+        ValueType | ArrayAttr[ValueType]
+    )  # TODO: base(ValueType) | base(ArrayAttr[ValueType]) ???
```
superlopuh, line 339:
```suggestion
    value = operand_def(TypeType | RangeType[ValueType])  
```

xdsl/dialects/pdl_interp.py
```diff
@@ -0,0 +1,508 @@
...
+
+    result = result_def(TypeType | ArrayAttr[TypeType])
+
+    def __init__(self, value: SSAValue) -> None:
+        super().__init__(operands=[value], result_types=[TypeType()])
```
superlopuh, line 344:
```suggestion
        super().__init__(operands=[value], result_types=[RangeType(TypeType()) if isinstance(value.type, RangeType) else TypeType()])
```

jumerckx:
> Oh the one caveat I forgot to mention about pushing to this branch is that the tests should always pass so that we can all build on it, can you please fix the tests here?

Sorry, I'd missed this comment (and code suggestions, thanks!!). Will fix CI.

alexarice:
@superlopuh is this PR still relevant?

superlopuh:
Yep

superlopuh:
We're using it as an active development branch

superlopuh:
I need to merge the last parts that are missing into xDSL-proper so that we do development in xDSL directly

jumerckx:
> I need to merge the last parts that are missing into xDSL-proper so that we do development in xDSL directly

Do you want to move away from a shared development branch then?
I quite like this to experiment with things like the list dialect that shouldn't immediately be added to xDSL.

superlopuh:
I feel like list should live in our experiments repo, and the core mechanism should live in xDSL

superlopuh:
After discussions with Jules, it seems like we might want to use this code in the future, but aren't sure, I'll close this for now.

PR #3114, opened 2024-08-29
dialects: (stim) add qubit attribute and qubit coordinate attribute

kimxworrall:
Add a base attribute to the stim dialect - StimAttr which is StimPrintable so that all implementing attributes must implement print_stim.

Add two initial attributes - QubitAttr to indicate qubits, and QubitCoordAttr which provides a pair of a coordinate of a physical qubit to a QubitAttr


xdsl/dialects/stim/ops.py
```diff
@@ -1,24 +1,119 @@
...
 
-from .stim_printer_parser import StimPrintable, StimPrinter
 
+class StimAttr(StimPrintable):
+    "Base Stim attribute"
 
-class StimOp(IRDLOperation, ABC):
-    def print_stim(self, printer: StimPrinter) -> None:
-        raise (PyRDLOpDefinitionError("print_stim not implemented!"))
+    name = "stim.attr"
```
superlopuh, line 21: What is the purpose of this class?

xdsl/dialects/stim/ops.py
```diff
@@ -1,24 +1,119 @@
...
+        if not isinstance(qubit, IntAttr):
+            qubit = IntAttr(qubit)
+        super().__init__(parameters=[qubit])
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[IntAttr]:
+        parser.parse_punctuation("<")
+        qubit = parser.parse_integer(allow_negative=False, allow_boolean=False)
+        parser.parse_punctuation(">")
+        return [IntAttr(qubit)]
```
superlopuh, line 44: nit:
```suggestion
        with parser.in_angle_brackets():
            qubit = parser.parse_integer(allow_negative=False, allow_boolean=False)
            return [IntAttr(qubit)]
```

xdsl/dialects/stim/ops.py
```diff
@@ -1,24 +1,119 @@
...
+
+    def print_parameters(self, printer: Printer) -> None:
+        with printer.in_angle_brackets():
+            printer.print(self.qubit.data)
+
+    def print_stim(self, printer: StimPrinter):
+        printer.print_string(self.qubit.data.__str__())
```
superlopuh, line 51:
```suggestion
        printer.print_string(f"{self.qubit.data}")
```

xdsl/dialects/stim/stim_printer_parser.py
```diff
@@ -19,6 +23,32 @@ def in_braces(self):
         finally:
             self.print_string("}")
 
+    @contextmanager
+    def in_parens(self):
+        self.print_string("(")
+        try:
+            yield
+        finally:
+            self.print_string(")")
```
superlopuh, line 32: This was never required, I'm not sure why we have this pattern
```suggestion
        yield
        self.print_string(")")
```

superlopuh (commented):
This PR adds some code and infrastructure for stim printing, but never tests it. We should either add the printing code in a future PR or add the tests here.

tests/dialects/stim/test_stim.py
superlopuh, line 1: could you please move the testing for attribute syntax to a filecheck file, something like `tests/filecheck/dialects/stablehlo/attrs.mlir`?

xdsl/dialects/stim/ops.py
kimxworrall: Everything in this file is stim printable - in mlir the convention seems to be making a base class for the attributes and operations - is it okay to not do that here?

superlopuh: We generally try to do it only when necessary. In this case, the fact that it has a name, and is a base class without inheriting from ABC are both a little confusing. I would just remove it for the time being.

superlopuh (approved):

superlopuh:
Let's merge this?

PR #3118, opened 2024-08-30
transformations: (stencil-tensorize-z-dim) Support loops

n-io:
Add support for enclosing unknown ups such as scf.for

### DIFF
--- a/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
+++ b/xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
@@ -297,15 +297,20 @@ class DmpSwapOpTensorize(RewritePattern):
 class StoreOpTensorize(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: StoreOp, rewriter: PatternRewriter, /):
-        rewriter.replace_matched_op(
-            StoreOp.get(
-                op.temp,
-                op.field,
-                StencilBoundsAttr(
-                    zip(list(op.bounds.lb.array)[:-1], list(op.bounds.ub)[:-1])
-                ),
+        if (
+            is_tensorized(op.field.type)
+            and isinstance(op.field.type, ShapedType)
+            and len(op.bounds.lb) != len(op.field.type.get_shape())
+        ):
+            rewriter.replace_matched_op(
+                StoreOp.get(
+                    op.temp,
+                    op.field,
+                    StencilBoundsAttr(
+                        zip(list(op.bounds.lb.array)[:-1], list(op.bounds.ub)[:-1])
+                    ),
+                )
             )
-        )
 
 
 class AccessOpUpdateShape(RewritePattern):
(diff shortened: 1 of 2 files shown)


dk949:
Could you please add a description as to what this PR aims to achieve?

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -262,29 +267,18 @@ def is_scalar(typ: Attribute) -> TypeGuard[AnyFloat]:
...
-class ExternalLoadOpTensorize(RewritePattern):
-    @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: ExternalLoadOp, rewriter: PatternRewriter, /):
-        if not is_tensorized(op.result.type):
-            assert isa(op.result.type, FieldType[Attribute])
-            rewriter.replace_matched_op(
-                ExternalLoadOp.get(op.field, stencil_field_to_tensor(op.result.type))
-            )
-
-
```
dk949, line 274: Is this no longer needed

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -303,20 +297,15 @@ def match_and_rewrite(self, op: dmp.SwapOp, rewriter: PatternRewriter, /):
 class StoreOpTensorize(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: StoreOp, rewriter: PatternRewriter, /):
-        if (
-            is_tensorized(op.field.type)
```
dk949, line 307: I don't really understand why we no longer need the `is_tensorized` checks here and in `LoadOpTensorize`, is it because of the type conversion pass?

dk949 (commented):
I don't fully understand what this change does. Is it to relax the restrictions on what can be tensorized?

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
n-io: It's handled by the type conversion pass.

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
n-io: I think this was more of a "is it actually needed and will it still run when I simply remove the checks?". Can put it back in probably, not really important.

n-io:
> I don't fully understand what this change does. Is it to relax the restrictions on what can be tensorized?

The type conversion pass helps dealing with unknown ops. Notably, including scf.for. Running the type conversion pass first has the effect of the old checks "input is tensorised and output is not tensorised" no longer working. So the ExternalLoadOp conversion was not needed at all, and some other rewrite passes needed their checks (esp for output) revisited.

dk949:
> > I don't fully understand what this change does. Is it to relax the restrictions on what can be tensorized?
> 
> The type conversion pass helps dealing with unknown ops. Notably, including scf.for. Running the type conversion pass first has the effect of the old checks "input is tensorised and output is not tensorised" no longer working. So the ExternalLoadOp conversion was not needed at all, and some other rewrite passes needed their checks (esp for output) revisited.

Makes sense, thank you

dk949 (approved):

PR #3119, opened 2024-08-30
CI: update ruff precommit

superlopuh:
Brings the precommit version in line with the dependency version. I'm still not sure what the official way to do this is, it feels weird to keep the two in sync manually. IIUC the dependency is used in the VSCode extension, so is also useful...


webmiche (approved):
can we do this together with the dependabot update? I know we used to have something weird that kinda parsed the dependency file to figure out its own version on the CI at some point (at least, AFAIR)

superlopuh:
Never for the precommit, that was for a CI workflow, so I've just manually subscribed in GitHub to all releases to our precommit dependencies, and open PRs for when new ones come out. It's not really a big hassle, just slightly annoying.

PR #3120, opened 2024-08-30
misc: remove type annotations from operand definitions [NFC]

superlopuh:
I regret writing the original sed script to add the type annotations, as they're not necessary.


remove_annotated.sed
superlopuh, line 1: no longer necessary

superlopuh:
I double-checked all the files and they're fine

superlopuh:
@math-fehr, as you can imagine, this PR accumulates conflicts very quickly, do you think we could time your review somehow? I'm happy to update it whenever you're ready to take a look at it, please ping me.

math-fehr:
Sorry for the delay, let's merge this now

math-fehr (approved):

PR #3121, opened 2024-08-31
dialects: (riscv) canonize x ^ x = 0

mamanain:
Hello! Canonizing this pattern.

Should it be just `li 0` instead of it's canonized form in the rewrite pattern? In this case it kind of seems like canonization of `li 0` is split to two places.


superlopuh (approved):
Perfect, thank you!

PR #3123, opened 2024-08-31
core: Assembly format shouldn't require `$*SegmentSizes`

alexarice:
Removes `*SegmentSizes` from the required properties seen if the appropriate option is set of the operation


superlopuh:
I'm not sure what change this has in practice, could you please add a test of something that works with your change but didn't before?

alexarice:
Added a test, the problem arises when `AttrSizedOperandSegments(as_property=True)` is used with assembly format

superlopuh (approved):

PR #3125, opened 2024-08-31
tests: Increase timeout on lit tests

alexarice:
The test `/tests/filecheck/dialects/stencil/oec-kernels/fvtp2d_qi.mlir` keeps timing out on my system. Could we either increase the timeout (as in this pr) or fix the test.


superlopuh (approved):
You're the third person to hit this, and the first person to actually open a PR to fix it 🥇

PR #3126, opened 2024-08-31
dialects: (scf) Add parse/print to `scf.if`

alexarice:
Adds custom parsing to some scf ops and fixes the relevant tests.

I noticed during this that `scf.reduce` is now quite different to upstream mlir, but didn't attempt to change this.


alexarice:
The implementation of `reduce` seems to disagree with the fixed mlir commit too, so will likely have to revert that for now

superlopuh:
jsut scf.if would already be great as a standalone PR, will also fix #2544

superlopuh (approved):

PR #3133, opened 2024-09-02
dialects: (scf) Add IsTerminator to scf.reduce

AntonLydike:
Add `IsTerminator` to `scf.reduce` op as per upstream

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -580,7 +580,7 @@ class ReduceOp(IRDLOperation):
 
     body: Region = region_def("single_block")
 
-    traits = frozenset([RecursiveMemoryEffect()])
+    traits = frozenset([RecursiveMemoryEffect(), IsTerminator()])
 
     def __init__(
         self,


alexarice:
I looked at this last week, upstream is significantly different, and in the pinned version of mlir `scf.reduce` is not a terminator I believe

JosseVanDelm:
It indeed became a terminator after the current MLIR version that is pinned:
https://github.com/llvm/llvm-project/pull/75314

I have no clue whether this is a big issue though?
Are there a lot of people who require stuff to be after an scf.reduce po?

superlopuh:
This seems like a slippery slope, couldn't we just update MLIR version instead? I would just suggest doing that after the CGO deadline

alexarice:
If my understanding is correct, then in the pinned mlir version you can have multiple `scf.reduce` operations in one `scf.parallel`, which would not be possible if it were a terminator. The up to date mlir version has been changed so that you can attach multiple regions/arguments. I don't really have any stake in this but it seems like a bad idea to make some strange hybrid `scf.reduce` op?

AntonLydike:
Yes, sorry, meant to close this earlier!

PR #3134, opened 2024-09-02
transformations: Add pass to lower csl-stencil dialect

n-io:
This pass lowers csl_stencil to csl and api calls.

~~Todo: Depends on a direction type and function pointers.~~

The pass works as follows:

* `csl_stencil.access` are lowered to api call (emitting dsd) + UnrealizedConversionCastOp (converting dsd to memref).
* The UnrealizedConversionCastOps are erased in the `memref-to-dsd` pass
* `csl_stencil.apply` is lowered to an api call. Its two regions are placed into csl.funcs that are passed as callbacks to the api call.
* `csl_stencil.yield` ops are lowered to `csl.return` as they terminate what are now callback functions with no return values.


xdsl/transforms/lower_csl_stencil.py
```diff
@@ -0,0 +1,219 @@
...
+    op: csl_stencil.AccessOp,
+) -> tuple[csl.DirectionOp, arith.Constant]:
+    """
+    Given an access op, return the distance and direction, assuming as access
+    to a neighbour (not self) in a star-shape pattern
+    """
+
+    offset = tuple(op.offset)
+    assert len(offset) == 2, "Expecting 2-dimensional access"
+    assert (offset[0] == 0) ^ (
```
dk949, line 36: I tend to find `!=` to be more readable than `^` when it comes to booleans

xdsl/transforms/lower_csl_stencil.py
```diff
@@ -0,0 +1,219 @@
...
+class LowerApplyOp(RewritePattern):
+    """
+    Lowers csl_stencil.apply to an API call. Places the two regions in csl.funcs and
+    passes them as callbacks.
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: csl_stencil.ApplyOp, rewriter: PatternRewriter, /):
+        if not (module_wrapper_op := _get_module_wrapper(op)):
+            return
```
dk949, line 107: Can this happen and not be an error?

xdsl/transforms/lower_csl_stencil.py
```diff
@@ -0,0 +1,219 @@
...
+            parent_func = op.parent_op()
+        assert (
+            parent_func
+        ), "Expected csl_stencil.apply to be inside a func.func or csl.func"
+
+        # set up csl funcs
+        reduce_fn = csl.FuncOp("chunk_reduce_cb", FunctionType.from_lists([i16], []))
+        post_fn = csl.FuncOp(
+            "post_process_cb", FunctionType.from_lists([], []), Region(Block())
+        )
```
dk949, line 124: This will cause issues if we ever have multiple `apply`s, but I don't think we support that anyway :shrug:

dk949 (commented):
Just a few minor things. Looks good otherwise :+1:

xdsl/transforms/lower_csl_stencil.py
n-io: Done as suggested.

xdsl/transforms/lower_csl_stencil.py
n-io: It may be valid in mlir, just this pass can't do anything with it.

xdsl/transforms/lower_csl_stencil.py
n-io: fixed

dk949 (approved):
Looks good!

PR #3135, opened 2024-09-02
dialects: (csl) added `csl.addressof_fn`

dk949:
Functions like `csl.addressof`, but takes a prop `fn_name` instead of an SSAValue. Result type is a single const pointer to a function (this is a limitation imposed by the CSL language.


n-io (approved):
Looks great, thanks!

PR #3136, opened 2024-09-02
dialects: (csl) Add `direction` type

n-io:
Adding `direction` type to csl

### DIFF
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -552,7 +552,7 @@ class CslPrintContext:
                     use = self._var_use(res, ty.constness.data.value)
                     self.print(f"{use} = &{name.string_value()};")
                 case csl.DirectionOp(dir=d, res=res):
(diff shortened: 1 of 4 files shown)


xdsl/backend/csl/print_csl.py
```diff
@@ -549,6 +551,9 @@ def print_block(self, body: Block):
                     ty = cast(csl.PtrType, res.type)
                     use = self._var_use(res, ty.constness.data.value)
                     self.print(f"{use} = &{name.string_value()};")
+                case csl.DirectionOp(dir=d, res=res):
+                    use = self._var_use(res, "const")
```
dk949, line 555: nit: `"const"` is the default introducer for `_var_use`, so you could drop it if you want to
```suggestion
                    use = self._var_use(res)
```

dk949 (approved):
Minus the nit (which you can feel free to ignore) it LGTM

dk949 (approved):

AntonLydike:
:heart:

PR #3137, opened 2024-09-02
transformations: Lower `csl_wrapper.import` to `csl.import`

dk949:
Hoists the import operation to the top of the `csl.module` with all of its dependencies. Constructs the struct needed for importing, then imports the module.


n-io (commented):
Would it make sense for this to be part of the csl-wrapper-to-csl pass?

dk949:
> Would it make sense for this to be part of the csl-wrapper-to-csl pass?

I thought about it, but figured that that pass is already quite long.

It would simplify the pipeline, so I'm not opposed to combining the passes

dk949:
> [!NOTE]
> Still need to translate empty field names to `concat_structs`

dk949:
Empty field names handled :+1:

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -227,6 +227,76 @@ def lower_program_module(
...
+                structs_to_concat.append(op)
+        out_structs: list[Operation] = [csl.ConstStructOp(*fields)]
+        for s in structs_to_concat:
+            out_structs.append(csl.ConcatStructOp(out_structs[-1], s))
+        return out_structs
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: csl_wrapper.ImportOp, rewriter: PatternRewriter, /):
+        csl_mod = self._get_csl_mod(op)
+        ops = self._collect_ops(op, [])
```
n-io, line 292: This seems to be running a series of `detach()` ops on the use-def chain. If you're concerned about leaving unused ops in the code, it might be easier to let the `dce`/`cse` pass handle the cleanup, and we might not need this method.

xdsl/transforms/csl_wrapper_to_csl.py
```diff
@@ -227,6 +227,76 @@ def lower_program_module(
...
+        structs = self._make_import_struct(op)
+        import_ = csl.ImportModuleConstOp(op.module, structs[-1])
```
n-io, line 294:
```suggestion
        import_ = csl.ImportModuleConstOp(op.module, structs[-1] if len(structs) > 0 else None)
```

xdsl/transforms/csl_wrapper_to_csl.py
dk949: This method moves the import statement to the module scope along with all of its dependencies. It does `detach() -> insert_op()` to preserve the links between all the operations as they are being moved . If there is an existing function that does this I'd love to use that instead.

xdsl/transforms/csl_wrapper_to_csl.py
dk949: Technically `structs` will never be empty with the current implementation, but I think this is still a good idea in case this changes in the future

xdsl/transforms/csl_wrapper_to_csl.py
n-io: Agreed. I think placing all ops inside the loop nest is a good strategy, but any existing pass that does hoisting/loop-invariant-code-motion would help here. My thinking is that we do not currently analyse whether any of the instructions depend on the loop vars, but this may not be allowed in the first place so it's prob ok. Theoretically, the hoisting could always go in a small separate pass, but not strictly necessary.

n-io (approved):

PR #3138, opened 2024-09-03
dialects: (scf) make iteration variable type explicit

alexarice:
In mlir, if the iteration variable of `scf.for` does not have an explicit type, then it is assumed to be `index`. This change matches this behaviour in xDSL.


dk949:
> I see that you touched some csl stuff. Does this affect anything there? (@dk949, @n-io)
> 
> Otherwise, LGTM!

We use `i16`s for loop indices which used to be inferred from `ub`, `lb` and `step`, but I'm guessing no longer is.

So, no, I don't think this affects anything.

superlopuh (approved):

PR #3139, opened 2024-09-03
backend: (csl) Add comparison operators to CSL

dk949:
Adds integer and floating point comparisons as well as `and` and `or`.

Signed and unsigned integers are compared the same way.

Ordered and unordered floats are compared the same way.

`true` and `false` float comparisons are printed as `true` and `false` literals.

`ord` and `uno` comparisons result in an error.

`arith.andi` and `arith.ori` produce `and` and `or` operators when operating on `i1` (bools) and `&` and `|` otherwise

### DIFF
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -53,6 +53,8 @@ class CslPrintContext:
     Maps operation name => operand for binary operands
     """
 
+    _cmp_ops: dict[str, dict[str, str | None]] = field(default_factory=dict)
+
     def register_binops(self):
         self._binops.update(
             {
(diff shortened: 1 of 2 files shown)


n-io (approved):

xdsl/backend/csl/print_csl.py
```diff
@@ -68,8 +71,63 @@ def register_binops(self):
...
+                    "ole": "<=",
+                    "one": "!=",
+                    "ord": None,
+                    "ueq": "==",
+                    "ugt": ">",
+                    "uge": ">=",
+                    "ult": "<",
+                    "ule": "<=",
+                    "une": "!=",
+                    "uno": None,
```
n-io, line 107: What happens in this case?

n-io (approved):

xdsl/backend/csl/print_csl.py
dk949: There is a check on line 122 for when the op maps to `None`.

Specifically for `ord` and `uno` we throw an error (line 127), because I think for now we don't need to handle this, and if we do, we will find out quickly when the error is thrown.

PR #3143, opened 2024-09-04
transformations: (csl) add `csl.rpc` op

dk949:
When constructing the program module in `csl-wrapper-to-csl`:

* find the `memcpy/memcpy` module
* get the `LAUNCH` color from it 
* use it in the `csl.rpc` op.


dk949:
> LGTM, though it might also go into the pass setting up a csl_wrapper

It could, but this felt like the better place for it, since it's where we fix all the things relating to CSL, like layout and imports.

PR #3144, opened 2024-09-04
transformations: add call to `unblock_cmd_stream`

dk949:
This is needed at the end of the post-processing function to signal
program end.


n-io (commented):
> This is needed at the end of the post-processing function to signal
program end.

This isn't quite the right fix, although for the right fix we're missing a pass. It should go at the end of the main function, and a good place for this is csl-stencil-to-csl-wrapper.

dk949:
@n-io with the outer loop transformation you are working on, can I close this?

n-io:
> @n-io with the outer loop transformation you are working on, can I close this?

I believe the cleanest way is to add it in the csl-stencil-to-csl-wrapper pass, simply at the end of the main func. I've opened #3198 for this

dk949:
> > @n-io with the outer loop transformation you are working on, can I close this?
> 
> I believe the cleanest way is to add it in the csl-stencil-to-csl-wrapper pass, simply at the end of the main func. I've opened #3198 for this

Cool, I'll close this one then

PR #3147, opened 2024-09-05
dialects: llvm.InLineAsmOp has_side_effects, is_align_stack are properties

JosseVanDelm:
I apparently made a small yet significant error when I implemented this operation the first time, which sometimes gives weird errors in mixed pipelines.

`has_side_effects`, `is_align_stack` are optional properties defined as UnitAttributes.
I apparently gave a very difficult to find error because I tested them as attributes, and not as properties. and because `has_side_effects` is not exactly the same as `has_side_effect`

I'm not sure the implementation I'm using here for defining optional properties is correct. It seems there are multiple ways of doing this in xdsl?


AntonLydike (approved):
:heart:

PR #3149, opened 2024-09-06
fix: Add the missing call to `@set_rectangle`

dk949:
When generating the layout module in `csl_wrapper`, `@set_rectangle` needs to be called with the values of `width` and `height` (via `csl.set_rectangle`).


n-io (approved):
Thanks for this, LGTM

PR #3152, opened 2024-09-09
backend: (csl) Don't use CSL keywords as var names

dk949:
Added a list of CSL language keywords which will be checked along with taken variable names when choosing the name of a new variable.


AntonLydike (approved):
Good stuff! This was nagging me on the back of my mind for a while!

PR #3156, opened 2024-09-10
dialects: (cf) add switch

alexarice:
Adds `cf.switch` and simple test

### DIFF
--- a/xdsl/dialects/cf.py
+++ b/xdsl/dialects/cf.py
@@ -24,14 +24,15 @@ from xdsl.irdl import (
     attr_def,
     irdl_op_definition,
     operand_def,
-    opt_attr_def,
+    opt_prop_def,
+    prop_def,
     successor_def,
     var_operand_def,
     var_successor_def,
 )
 from xdsl.parser import Parser
 from xdsl.printer import Printer
-from xdsl.traits import IsTerminator
+from xdsl.traits import IsTerminator, Pure
 from xdsl.utils.exceptions import VerifyException
 
 
@@ -115,7 +116,7 @@ class Switch(IRDLOperation):
 
     name = "cf.switch"
 
-    case_values = opt_attr_def(DenseIntOrFPElementsAttr)
+    case_values = opt_prop_def(DenseIntOrFPElementsAttr)
 
     flag = operand_def(AnySignlessIntegerOrIndexType)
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/cf.py
```diff
@@ -91,12 +110,218 @@ def __init__(
...
+
+    name = "cf.switch"
+
+    case_values = opt_attr_def(DenseIntOrFPElementsAttr)
+
+    flag = operand_def(AnySignlessIntegerOrIndexType)
+
+    default_operands = var_operand_def(AnyAttr())
+
+    case_operands = var_operand_def(AnyAttr())
```
superlopuh, line 125:
```suggestion
    default_operands = var_operand_def()

    case_operands = var_operand_def()
```

xdsl/dialects/cf.py
```diff
@@ -91,12 +110,218 @@ def __init__(
...
+        case_operands: Sequence[Sequence[Operation | SSAValue]] = [],
+        attr_dict: dict[str, Attribute] | None = None,
+    ):
+        case_operand_segments: list[int] = []
+        c_operands: list[Operation | SSAValue] = []
+        for case_operand in case_operands:
+            case_operand_segments.append(len(case_operand))
+            c_operands += list(case_operand)
+        operands = [flag, default_operands, c_operands]
+        attributes: dict[str, Attribute] = attr_dict or {}
```
superlopuh, line 154: Instead of always creating a new dict if the one passed in is empty, only create one if the one passed in is None
```suggestion
        attributes = attr_dict if attr_dict is not None else
```

xdsl/dialects/cf.py
```diff
@@ -91,12 +110,218 @@ def __init__(
...
+        self,
+        flag: Operation | SSAValue,
+        default_block: Successor,
+        default_operands: Sequence[Operation | SSAValue],
+        case_values: DenseIntOrFPElementsAttr | None = None,
+        case_blocks: Sequence[Successor] = [],
+        case_operands: Sequence[Sequence[Operation | SSAValue]] = [],
+        attr_dict: dict[str, Attribute] | None = None,
+    ):
+        case_operand_segments: list[int] = []
```
superlopuh, line 148:
```suggestion
        case_operand_segments = tuple(len(o) for o in case_operands)
```

xdsl/dialects/cf.py
```diff
@@ -91,12 +110,218 @@ def __init__(
...
+        c_operands: list[Operation | SSAValue] = []
```
superlopuh, line 149: Might as well iterate twice, IMO makes the code a little more readable
```suggestion
        c_operands = tuple(o for os in case_operand for o in os)
```

superlopuh (approved):

xdsl/dialects/cf.py
alexarice: I don't understand this, python's `or` should be lazy

superlopuh:
```
>>> x = {}
>>> y = x or {}
>>> x is y
False
```

alexarice: I see, thanks

alexarice:
Apparently mlir uses properties when it says it's using attributes and so I've switched this over and am rerequesting review

superlopuh:
Yeah I've found this really confusing, I think it's because no-one's responsible for finishing the properties transition, and because attributes can be put in properties...

superlopuh (approved):

PR #3157, opened 2024-09-10
dialects: (scf) add index_switch

alexarice:
Adds `scf.index_switch` and test

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -7,8 +7,10 @@ from typing_extensions import Self
 
 from xdsl.dialects.builtin import (
     AnySignlessIntegerOrIndexType,
+    DenseArrayBase,
     IndexType,
     IntegerType,
+    i64,
 )
 from xdsl.dialects.utils import (
     AbstractYieldOperation,
@@ -21,11 +23,13 @@ from xdsl.irdl import (
     AttrSizedOperandSegments,
     ConstraintVar,
     IRDLOperation,
+    attr_def,
     irdl_op_definition,
     operand_def,
     region_def,
     traits_def,
     var_operand_def,
+    var_region_def,
     var_result_def,
 )
 from xdsl.parser import Parser, UnresolvedOperand
@@ -156,7 +160,11 @@ class Yield(AbstractYieldOperation[Attribute]):
 
     traits = traits_def(
         lambda: frozenset(
-            [IsTerminator(), HasParent(For, If, ParallelOp, While), Pure()]
+            [
+                IsTerminator(),
+                HasParent(For, If, ParallelOp, While, IndexSwitchOp),
+                Pure(),
+            ]
         )
     )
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/scf.py
```diff
@@ -689,6 +697,107 @@ def parse(cls, parser: Parser) -> Self:
...
 
 
+@irdl_op_definition
+class IndexSwitchOp(IRDLOperation):
+    name = "scf.index_switch"
+
+    arg = operand_def(IndexType)
+    cases = attr_def(DenseArrayBase)
+
+    output = var_result_def(AnyAttr())
```
superlopuh, line 707:
```suggestion
    output = var_result_def()
```

xdsl/dialects/scf.py
```diff
@@ -689,6 +697,107 @@ def parse(cls, parser: Parser) -> Self:
...
+    def __init__(
+        self,
+        arg: Operation | SSAValue,
+        cases: DenseArrayBase,
+        default_region: Region,
+        case_regions: Sequence[Region],
+        result_types: Sequence[Attribute],
+        attr_dict: dict[str, Attribute] | None = None,
+    ):
+        attributes: dict[str, Attribute] = attr_dict or {}
```
superlopuh, line 721:
```suggestion
        attributes = attr_dict or if attr_dict is not None else {}
```

xdsl/dialects/scf.py
```diff
@@ -689,6 +697,107 @@ def parse(cls, parser: Parser) -> Self:
...
+        def verify_region(region: Region, name: str):
+            yield_op = region.block.last_op
+            if not yield_op or not isinstance(yield_op, Yield):
+                raise VerifyException(f"expected region {name} to end with `scf.yield`")
+
+            if yield_op.operand_types != self.result_types:
+                raise VerifyException(
+                    f"region {name} returns values of types {yield_op.operand_types}"
+                    f"but expected {self.result_types}"
+                )
```
superlopuh, line 753: could you please add tests for these?

xdsl/dialects/scf.py
```diff
@@ -689,6 +697,107 @@ def parse(cls, parser: Parser) -> Self:
...
+
+        verify_region(self.default_region, "default")
+        for name, region in zip(cases, self.case_regions):
+            verify_region(region, str(name))
```
superlopuh, line 757: is this necessary?

xdsl/dialects/scf.py
alexarice: Also removed all the other AnyAttr's in this file

xdsl/dialects/scf.py
alexarice: Similar to other pr, I believe python's `or` is lazy and so the original should be ok

xdsl/dialects/scf.py
alexarice: The verification is a direct translation of the verification for the mlir operation. It's as necessary as any verification I guess

alexarice: I expect I could use `SingleBlockImplicitTerminator` and leave it at that?

alexarice:
Added the appropriate traits as they should have been there anyway

xdsl/dialects/scf.py
superlopuh: that's not my point, it's that if it's empty it's still Falsey, so it'll create a new one. Also, I think it's bad style to do Falsey things in Python when None is involved, partially for this reason.

xdsl/dialects/scf.py
superlopuh: My impression is that the regions of the operation are already verified by the time the custom verifier is called, is that not the case? Can you check that they're not verified twice?

alexarice: `verify_region` is a custom function defined above

superlopuh: Ah, I see, I thought we had `Operation.verify` call into helper methods on the operation, and that this was also doing this manually. In that case all good, although maybe it might be better as a private method?

superlopuh (approved):

xdsl/dialects/scf.py
superlopuh: I should really be a bit more attentive in PRs, I'm not sure how I missed it. It's private indeed, but that just means that you define it every time that you verify this operation. Ideally, it should be a static method with an underscore, so that it's defined only once, when the file is imported.

alexarice: No problem, I'll make a quick pr

PR #3178, opened 2024-09-17
dialects: (scf) small refactor for index_switch

alexarice:
As discussed in https://github.com/xdslproject/xdsl/pull/3157#discussion_r1762987533


superlopuh (approved):
🚀

PR #3179, opened 2024-09-17
transformations: (scf-to-cf) add lowering for index_switch

alexarice:
Again this is mainly copied from the implementation in mlir

### DIFF
--- a/xdsl/transforms/convert_scf_to_cf.py
+++ b/xdsl/transforms/convert_scf_to_cf.py
@@ -1,9 +1,12 @@
+from typing import cast
+
 from xdsl.context import MLContext
 from xdsl.dialects import builtin
-from xdsl.dialects.arith import Addi, Cmpi
-from xdsl.dialects.cf import Branch, ConditionalBranch
-from xdsl.dialects.scf import For, If
-from xdsl.ir import Block
+from xdsl.dialects.arith import Addi, Cmpi, IndexCastOp
+from xdsl.dialects.builtin import DenseIntOrFPElementsAttr, i32
+from xdsl.dialects.cf import Branch, ConditionalBranch, Switch
+from xdsl.dialects.scf import For, If, IndexSwitchOp, Yield
+from xdsl.ir import Block, Region
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
     GreedyRewritePatternApplier,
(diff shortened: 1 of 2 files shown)


xdsl/transforms/convert_scf_to_cf.py
```diff
@@ -149,6 +152,67 @@ def match_and_rewrite(self, for_op: For, rewriter: PatternRewriter):
...
+            assert block is not None
+
+            # Convert yield op to a branch to the continue block
+            yield_op = block.last_op
+            assert isinstance(yield_op, Yield)
+            rewriter.replace_op(yield_op, Branch(continue_block, *yield_op.operands))
+
+            # Inline the region
+            rewriter.inline_region_before(region, continue_block)
+            return block
```
superlopuh, line 184: I would recommend outlining this helper so that it's not redefined for every switch operation in the IR

xdsl/transforms/convert_scf_to_cf.py
```diff
@@ -149,6 +152,67 @@ def match_and_rewrite(self, for_op: For, rewriter: PatternRewriter):
...
+        rewriter.insert_op(case_value, InsertPoint.at_end(condition_block))
+
+        # Create the switch
+        case_operands = tuple(() for _ in case_successors)
+        rewriter.insert_op(
+            Switch(
+                case_value,
+                default_block,
+                (),
+                DenseIntOrFPElementsAttr.vector_from_list(case_values, i32),
```
superlopuh, line 206: can this not be the `op.cases` above? Does the type change?

xdsl/transforms/convert_scf_to_cf.py
```diff
@@ -160,5 +224,5 @@ class ConvertScfToCf(ModulePass):
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         PatternRewriteWalker(
-            GreedyRewritePatternApplier([IfLowering(), ForLowering()])
+            GreedyRewritePatternApplier([SwitchLowering(), IfLowering(), ForLowering()])
```
superlopuh, line 227:
```suggestion
            GreedyRewritePatternApplier([SwitchLowering(), IfLowering(), ForLowering(),])
```

xdsl/transforms/convert_scf_to_cf.py
alexarice: Yes, `op.cases` is a [`DenseArrayBase`](https://mlir.llvm.org/docs/Dialects/Builtin/#densearrayattr) but this is a [`DenseIntOrFPElementsAttr`](https://mlir.llvm.org/docs/Dialects/Builtin/#denseintorfpelementsattr) and these appear to not be the same

PR #3183, opened 2024-09-17
dialect: (csl) Adds variables to the dialect

dk949:
Variables work very much like `memref.global`, but they are scalars.

Ops:
* `csl.variable`: declare variable with optional default value
* `csl.get_var`: get SSA value (constant) of the variable
* `csl.update_var` set variable from ssa value

Types:
* `csl.var<T>`: type of a variable

### DIFF
--- a/tests/filecheck/backend/csl/print_csl.mlir
+++ b/tests/filecheck/backend/csl/print_csl.mlir
@@ -180,6 +180,17 @@
     csl.return
   }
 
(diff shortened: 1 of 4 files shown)


dk949:
This is a little rough around the edges (no docstrings and whatnot). I'll clean it up if we're happy with the general concept

n-io (approved):
Thanks, looks great

PR #3186, opened 2024-09-17
Added initial eqsat front end pass

jianyicheng:
This PR addresses https://github.com/xdslproject/xdsl/issues/3170:
* Added initial front end pass `eqsat-create-eclasses` for the minimal example
* Updated the test case to match the pass

### DIFF
--- a/xdsl/tools/command_line_tool.py
+++ b/xdsl/tools/command_line_tool.py
@@ -405,7 +405,7 @@ def get_all_passes() -> dict[str, Callable[[], type[ModulePass]]]:
 
         return stencil_bufferize.StencilBufferize
 
-    def get_create_eclasses():
+    def get_eqsat_create_eclasses():
         from xdsl.transforms import eqsat_create_eclasses
 
         return eqsat_create_eclasses.EqsatCreateEclasses
@@ -488,7 +488,7 @@ def get_all_passes() -> dict[str, Callable[[], type[ModulePass]]]:
         "stencil-unroll": get_stencil_unroll,
         "stencil-bufferize": get_stencil_bufferize,
         "test-lower-linalg-to-snitch": get_test_lower_linalg_to_snitch,
-        "eqsat-create-eclasses": get_create_eclasses,
+        "eqsat-create-eclasses": get_eqsat_create_eclasses,
     }
 
 
--- a/xdsl/tools/command_line_tool.py
+++ b/xdsl/tools/command_line_tool.py
@@ -405,6 +405,11 @@ def get_all_passes() -> dict[str, Callable[[], type[ModulePass]]]:
 
         return stencil_bufferize.StencilBufferize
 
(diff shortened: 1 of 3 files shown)


xdsl/tools/command_line_tool.py
```diff
@@ -405,6 +405,11 @@ def get_stencil_bufferize():
 
         return stencil_bufferize.StencilBufferize
 
+    def get_create_eclasses():
```
superlopuh, line 408:
```suggestion
    def get_eqsat_create_eclasses():
```

xdsl/transforms/eqsat_create_eclasses.py
```diff
@@ -1,32 +1,91 @@
...
 
 def insert_eclass_ops(block: Block):
+    # Insert eqsat.eclass for each operation
     for op in block.ops:
         results = op.results
         if len(results) != 1:
-            raise DiagnosticException("Ops with non-single results not handled")
+            continue
+            # TODO: ignore operations with no result for now
+            # raise DiagnosticException("Ops with non-single results not handled")
```
superlopuh, line 22: for our WIP branch, this is fine, but for main, we generally avoid TODO comments without associated tasks to track them, in a comment in the code. In this case, I'd just raise the diagnostic exception, and add a filecheck test to track that the diagnostic error is raised as expected (with a `--verify-diagnostics` flag passed to `xdsl-opt`). With that setup we tend not to file issues, and just fix the test when we come across the error.

xdsl/transforms/eqsat_create_eclasses.py
```diff
@@ -1,32 +1,91 @@
...
 
         eclass_op = eqsat.EClassOp(results[0])
-        Rewriter.insert_op(op, InsertPoint.after(eclass_op))
-        eclass_op.results[0].replace_by(eclass_op.result)
+        insertion_point = InsertPoint.after(op)
+        Rewriter.insert_op(eclass_op, insertion_point)
+        results[0].replace_by(eclass_op.results[0])
+        # Redirect eclassop operand back to the original value
+        # TODO: do we need a `replace_by_except` function, e.g.
+        # eclass_op.result.replace_by_except(eclass_op.results[0], [eclass_op.result])
```
superlopuh, line 30: I'm not sure what MLIR has for this, but this works ok, no?

jianyicheng: I am thinking of this: https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html#a6f5567fd62feeaf0cbba4877732c512a

xdsl/transforms/eqsat_create_eclasses.py
jianyicheng: Fixed here: https://github.com/xdslproject/xdsl/pull/3189

PR #3188, opened 2024-09-18
dialects: (tosa) turn attributes into properties

jorendumoulin:
The attributes are actually properties in tosa.
I got fooled by the `ParsePropInAttrDict` ☹️


superlopuh (commented):
This makes me think that it's probably worth adding a roundtrip test with MLIR to make sure we're consistent

jorendumoulin:
That's the funny thing, there are mlir interop tests, however, as most interop tests, I only used it with custom printing. The check thus didn't notice a difference between attributes or properties that are custom printed as attributes.

Does it make sense to change
```
// RUN: xdsl-opt %s | xdsl-opt | mlir-opt | filecheck %s
```
Into
```
// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | mlir-opt --mlir-print-op-generic | xdsl-opt | mlir-opt | filecheck %s
```
?

superlopuh:
Oh interesting. Yeah maybe we should have both indeed, I doubt it will have a significant overhead in time

superlopuh (approved):

PR #3189, opened 2024-09-18
transformations: (eqsat) add pass to convert non-eclass functions to eclass

jianyicheng:
This PR addresses https://github.com/xdslproject/xdsl/issues/3170:

- [x] Added initial front end pass `eqsat-create-eclasses` for the minimal example
- [x] Added an initial test case for the pass


tests/filecheck/transforms/eqsat-create-eclasses.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: xdsl-opt -p eqsat-create-eclasses --verify-diagnostics %s | filecheck %s
+
+func.func @test(%x : index) -> (index) {
+    %c2 = arith.constant 2 : index
+    %res = arith.muli %x, %c2 : index
+    func.return %res : index
+}
+
+// CHECK: Error while applying pattern: Ops with non-single results not handled
```
superlopuh, line 9: Oh interesting, I forgot about function returns, they seem useful enough to have from the get-go, could you please change the check to rather be `len(results) > 1`? I would then recommend using the `test` dialect in filecheck, to make sure we're testing exactly what we mean to test:
```suggestion
func.func @test(%x : index) -> (index) {
    %r0, %r1 = "test.op"() : () -> (index, index)
}

// CHECK: Error while applying pattern: Only single-result operations can be converted to eclasses, got 2 results
```
(I also changed the message do include the length of the results, should hopefully help debug this in the future)

xdsl/transforms/eqsat_create_eclasses.py
```diff
@@ -0,0 +1,79 @@
...
+from xdsl.rewriter import InsertPoint, Rewriter
+from xdsl.utils.exceptions import DiagnosticException
+
+
+def insert_eclass_ops(block: Block):
+    # Insert eqsat.eclass for each operation
+    for op in block.ops:
+        results = op.results
+        if len(results) != 1:
+            raise DiagnosticException("Ops with non-single results not handled")
```
superlopuh, line 21:
```suggestion
        results = op.results
        if not results:
            # No need to convert ops with no results
            continue
        if len(results) != 1:
            raise DiagnosticException("Ops with non-single results not handled")
```

superlopuh: This seems to handle the no results case

superlopuh (commented):
Could you please add a test with a successful transform also?

xdsl/transforms/eqsat_create_eclasses.py
jianyicheng: It is safe to ignore return ops, but I feel that there is a need to handle other ops with no result. I meant to flag that so we know this is something needing more work. Ignoring it might cause problem when dealing with other cases such as `memref.store`, but I am not sure. Shall we skip return op only and keep `DiagnosticException` for other ops with `len(results) != 1`? i.e.:

```python
if len(results) != 1 and not isinstance(op, func.ReturnOp):
    raise DiagnosticException("Ops with non-single results not handled")
```
Happy to commit your version as well if we plan to address it later.

superlopuh: Sounds good to me! MLIR has traits that might be helpful for this, like `ReturnLike`, which we might want to migrate to in the future, but I have a feeling that this will be enough for now

xdsl/transforms/eqsat_create_eclasses.py
```diff
@@ -0,0 +1,79 @@
...
+        eclass_op.operands[0] = results[0]
+
+    # Insert eqsat.eclass for each arg
+    for arg in block.args:
+        eclass_op = eqsat.EClassOp(arg)
+        insertion_point = InsertPoint.at_start(block)
+        Rewriter.insert_op(eclass_op, insertion_point)
+        arg.replace_by(eclass_op.results[0])
+        # Redirect eclassop operand back to the original value
+        eclass_op.operands[0] = arg
```
superlopuh, line 37: I opened a PR for the helper you suggested, should work here #3196

xdsl/transforms/eqsat_create_eclasses.py
jianyicheng: Cool. I will work on refactoring this this week!

xdsl/transforms/eqsat_create_eclasses.py
jianyicheng: Lovely, will use that instead

superlopuh (approved):
Perfect, thank you!

PR #3192, opened 2024-09-18
transformations: (csl-stencil) Add pass to handle async ops and enclosing cf

n-io:
Adds a pass to handle the async control flow of `csl_stencil.apply` and any enclosing loops by translating control flow into a csl.func call graph.


xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+
+        post_func = csl.FuncOp(f"for_post{self.counter}", no_params)
+        rewriter.inline_block(post_block, InsertPoint.at_start(post_func.body.block))
+
+        cond_func = csl.TaskOp(
+            f"for_cond{self.counter}",
+            no_params,
+            task_kind=csl.TaskKind.LOCAL,
+            id=cond_task_id,
+        )
```
AntonLydike, line 122: Can you add some documentation why `for_post` can be a function, but `for_cond` needs to be a task?

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+        assert isa(op.iter_args[0].type, MemRefType[csl.ZerosOp.T])
+        element_type = op.iter_args[0].type.get_element_type()
+        assert all(
+            isa(a.type, MemRefType[csl.ZerosOp.T])
+            and element_type == a.type.get_element_type()
+            for a in op.iter_args
+        )
+
+        no_params = FunctionType.from_lists([], [])
+        cond_task_id = self.counter + 1
```
AntonLydike, line 107: Are you sure that this is a robust enough solution? I remember us thinking about having task-id allocation and referring to them by symbols in the meantime. What happened to that plan? (jsut curious, not necessarily criticism)

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+            stepped = arith.Addi(iv_load, step)
+            csl.StoreVarOp(iv, stepped)
+            load_vars = [csl.LoadVarOp(v) for v in iter_vars]
+            for iter_var, yielded_var in zip(iter_vars, terminator.arguments):
+                idx = (
+                    op.body.block.args.index(yielded_var) - 1
+                )  # subtract 1 as the first block arg is the loop var
+                csl.StoreVarOp(iter_var, load_vars[idx])
+            csl.ActivateOp(cond_task_id, csl.TaskKind.LOCAL)
+            # csl.CallOp(SymbolRefAttr(cond_func.sym_name))
```
AntonLydike, line 169: Again, a tiny bit of information why you choose to use task instead of a func would be appreciated.

Also, commented out code should probably be deleted.

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+            step = arith.Constant.from_int_and_width(op.step.op.value.value, 16)
+            iv_load = csl.LoadVarOp(iv)
+            stepped = arith.Addi(iv_load, step)
+            csl.StoreVarOp(iv, stepped)
+            load_vars = [csl.LoadVarOp(v) for v in iter_vars]
+            for iter_var, yielded_var in zip(iter_vars, terminator.arguments):
+                idx = (
+                    op.body.block.args.index(yielded_var) - 1
+                )  # subtract 1 as the first block arg is the loop var
+                csl.StoreVarOp(iter_var, load_vars[idx])
```
AntonLydike, line 167: Why are we inserting a load and then an immediate store for all loop-carried vars?

tests/filecheck/transforms/csl-stencil-handle-async-flow.mlir
```diff
@@ -0,0 +1,174 @@
...
+// CHECK-NEXT:     }
+// CHECK-NEXT:     csl.func @for_inc0() {
+// CHECK-NEXT:       %47 = arith.constant 1 : i16
+// CHECK-NEXT:       %48 = "csl.load_var"(%38) : (!csl.var<i16>) -> i16
+// CHECK-NEXT:       %49 = arith.addi %48, %47 : i16
+// CHECK-NEXT:       "csl.store_var"(%38, %49) : (!csl.var<i16>, i16) -> ()
+// CHECK-NEXT:       %50 = "csl.load_var"(%39) : (!csl.var<memref<512xf32>>) -> memref<512xf32>
+// CHECK-NEXT:       %51 = "csl.load_var"(%40) : (!csl.var<memref<512xf32>>) -> memref<512xf32>
+// CHECK-NEXT:       "csl.store_var"(%39, %51) : (!csl.var<memref<512xf32>>, memref<512xf32>) -> ()
+// CHECK-NEXT:       "csl.store_var"(%40, %50) : (!csl.var<memref<512xf32>>, memref<512xf32>) -> ()
```
AntonLydike, line 164: This sequence of load then stores appears to be unnecessary?

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+        assert isa(op.step.op.value, IntegerAttr[IndexType])
+
+        # limitation: all iter_args must be memrefs (stencil buffers) and have the same data type
+        assert isa(op.iter_args[0].type, MemRefType[csl.ZerosOp.T])
+        element_type = op.iter_args[0].type.get_element_type()
+        assert all(
+            isa(a.type, MemRefType[csl.ZerosOp.T])
+            and element_type == a.type.get_element_type()
+            for a in op.iter_args
+        )
```
AntonLydike, line 104: Nit: I'm generally not a fan of asserts in rewrite patterns, but I can see the appeal here. When the pass just aborts you may end up with silent failures, which are a pain to debug... I guess this shows that xDSL is lacking some sort of warning system for not-applied passes...

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+        assert isinstance(op.ub, OpResult)
+        assert isinstance(op.step, OpResult)
+        assert isinstance(op.lb.op, arith.Constant)
+        assert isinstance(op.ub.op, arith.Constant)
+        assert isinstance(op.step.op, arith.Constant)
+        assert isa(op.lb.op.value, IntegerAttr[IndexType])
+        assert isa(op.ub.op.value, IntegerAttr[IndexType])
+        assert isa(op.step.op.value, IntegerAttr[IndexType])
+
+        # limitation: all iter_args must be memrefs (stencil buffers) and have the same data type
```
AntonLydike, line 97: This limitation is not mentioned in the docstring above. Do you have any explanation for why this is?

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,233 @@
...
+        assert isa(op.iter_args[0].type, MemRefType[csl.ZerosOp.T])
+        element_type = op.iter_args[0].type.get_element_type()
+        assert all(
+            isa(a.type, MemRefType[csl.ZerosOp.T])
```
AntonLydike, line 101: Also, I'm really not sure what `csl.ZerosOp.T` is doing here. Why add this? That `T` is only well-defined inside the context of the `csl.ZerosOp`, right? Why add it here?

AntonLydike (commented):
Nice progress! I have some comments here and there

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: Technically, they can all be functions. One should be a task, however, to avoid a potential call stack in memory for long-running loops. Will add a comment.

tests/filecheck/transforms/csl-stencil-handle-async-flow.mlir
n-io: It's doing a flip, as the scf.yield is swapping the arguments. In this example, we load them and store them in the order prescribed, ie. reversed.

n-io: Added a comment.

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: This is to handle scf.yield yielding things out of order. `idx` should help translate.

Suppose our iter_args/iter_vars/block_args[1:] are `a, b` and we yield `b, a`. Then in the first iteration of the loop we have `iter_var == a` and `yielded_var == b` meaning we want to store `b` to `a`. We look up the original index of `b`, take what we pre-loaded (`load_vars`), and store it to `a` (`iter_var`).

Added a comment.

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: I think in MLIR there are separate match and apply functions, which separates out code that checks if a pattern should be applied. We also need quite a few of these asserts to make pyright happy, and the new `make precommit` forces us to split `assert isinstance(op.lb, OpResult) and isinstance(op.lb.op, arith.Constant)` into separate asserts.

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: This is primarily a "todo: not implemented".

n-io: Good spot though, added a comment.

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: A relic from a previous implementation. Changed to `Attribute`.

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: Yes and no, with our current comms library we have ids 1-3 free for local task ids. We could prob put an assert not to exceed that, but I think it'll simply not compile at a later stage. The question is how much knowledge this pass should have of things downstream? That said, it is the first thing in the pipeline to handle tasks and task ids, so it has a bit of freedom.

xdsl/dialects/csl/csl.py
```diff
@@ -400,7 +400,7 @@ def get_element_type(self):
         return self.res.type.get_element_type()
 
     @staticmethod
-    def from_type(child_type: TypeAttribute) -> VariableOp:
```
dk949, line 403: Should this not be a `TypeAttribute`?

xdsl/transforms/csl_stencil_handle_async_flow.py
```diff
@@ -0,0 +1,238 @@
...
+        if (
+            op.next_op
+            and isinstance(call_op := op.next_op, csl.CallOp)
+            and op.next_op.next_op
+            and isinstance(op.next_op.next_op, csl.ReturnOp)
+        ):
+            terminator = op.post_process.block.last_op
+            assert isinstance(terminator, csl_stencil.YieldOp)
+            rewriter.insert_op(call_op.clone(), InsertPoint.before(terminator))
+            rewriter.erase_op(call_op)
```
dk949, line 49: Does this still work if `call_op` had a result?

I'm also somewhat confused as to what this rewrite does in general

dk949 (commented):
Just a couple of things from me

xdsl/dialects/csl/csl.py
n-io: Then pyright complains in this line: `[csl.VariableOp.from_type(arg_t) for arg_t in op.iter_args.types]`

xdsl/transforms/csl_stencil_handle_async_flow.py
n-io: There is an implicit assumption that the `call_op` was placed there by the other rewrite pass. It's not finished yet though.

xdsl/dialects/csl/csl.py
dk949: This seems like the wrong place to fix this then..?

Although, @AntonLydike, is there a reason some types do not inherit `TypeAttribute`? In which case this should be an `Attribute`.

AntonLydike: Well, good question. All types *should* inherit from `TypeAttribute`, as that will have them be printed as `!dialect.name` instead of `#dialect.name` (which is attribute specific and can't be used as a type according to MLIR).

dk949 (approved):
Looks good AFAICT

PR #3196, opened 2024-09-19
api: add replace_by_if helper to control ssa value replacement

### LINKED ISSUE #3190: api: supporting replaceUsesWithIf
Now we only have `replace_by` in the core, and in some cases we only want to replace part of the uses with a new value. This is implemented as `replaceUsesWithIf` in MLIR: https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html#a6f5567fd62feeaf0cbba4877732c512a

Given the current implementation of `replace_by`:
https://github.com/xdslproject/xdsl/blob/cd2d5857e2c14deda0ca276aacc44a022e9a1e6b/xdsl/ir/core.py#L157-L164

I wonder if it would be useful to add the following function: 
```python
    def replace_by_except(self, value: SSAValue, exception: [SSAValue]) -> None:
        """Replace the value by another value in its uses except the given exception."""
        for use in self.uses.copy():
            if use not in exception:
                use.operation.operands[use.index] = value
        assert len(self.uses) <= len(exception), "unexpected error in xdsl"
```

superlopuh:
Resolves #3190

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -2,7 +2,15 @@ from __future__ import annotations
 
 import re
 from abc import ABC, abstractmethod
-from collections.abc import Hashable, Iterable, Iterator, Mapping, Reversible, Sequence
+from collections.abc import (
+    Callable,
+    Hashable,
+    Iterable,
+    Iterator,
+    Mapping,
+    Reversible,
+    Sequence,
+)
 from dataclasses import dataclass, field
 from io import StringIO
 from itertools import chain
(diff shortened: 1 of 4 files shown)


jianyicheng (approved):

xdsl/ir/core.py
```diff
@@ -161,7 +169,19 @@ def replace_by(self, value: SSAValue) -> None:
...
+    def replace_by_if(self, value: SSAValue, test: Callable[[Use], bool]):
+        """
+        Replace the value by another value in all its uses that pass the given test
+        function.
+        """
+        for use in self.uses.copy():
+            if test(use):
+                use.operation.operands[use.index] = value
+        # carry over name if possible
+        if value.name_hint is None:
```
math-fehr, line 183: Can we have an option for changing the name?
Compared to `replace_by`, we don't change all uses, so we might run this function multiple times per value, making this name change a bit weird?

superlopuh: I doubt we'll ever hit the case you describe, it seems to be primarily useful for "replace everything except" as opposed to some targeted replacing. I'll merge this PR, and happily add an option if you are sure that it would be a good idea

AntonLydike:
:heart: Really cool stuff! Thanks @superlopuh!

PR #3199, opened 2024-09-20
backend: (csl) added arith.select to csl backend

dk949:
`arith.select` can be translated to a CSL ternary operator
`if (cond) lhs else rhs`.


dk949:
I thought we might need it. Turns out we didn't really, but I figured I'd push it anyway since it's already written.

n-io (approved):

PR #3202, opened 2024-09-20
backend: (csl) allow `get_dir` to be inlined

dk949:
Using `_print_or_promote_to_inline_expr` instead of `print`


n-io:
This may need a small test case.

n-io (approved):

PR #3203, opened 2024-09-20
transformations: (lower-csl-stencil) Store results to apply.dest

n-io:
Lowering csl_stencil.apply's yield op now stores the results to the dest buffer.


dk949 (approved):
Thank you! Looks good

PR #3207, opened 2024-09-23
bug: Print csl namehint fix

n-io:
Print via `_use_var` instead of `name_hint`, as that can be `None`.


dk949 (approved):
Looks good AFAICT

PR #3209, opened 2024-09-23
fix: (csl) adjusted width and height of the PE

dk949:
They need to be 2 greater than the dmp topology

@n-io Could you please let me know if this looks right?


n-io (commented):
I think it's better not to hard-code this, though if we want a simple fix, we can adjust width and height by adding `2*(pattern-1)`, which I believe is `2*max_distance`. Best done outside the loop.

dk949:
> I think it's better not to hard-code this, though if we want a simple fix, we can adjust width and height by adding `2*(pattern-1)`, which I believe is `2*max_distance`. Best done outside the loop.

:+1: 

Sounds good.

n-io (approved):

PR #3210, opened 2024-09-23
tool: tblgen-to-py script

alexarice:
This is the start of a "tblgen-to-py" script for reading the json output of `llvm-tblgen --dump-json` and creating an xdsl dialect. I would like some help/advice on how to best integrate this into the existing tooling. I think something similar to the "irdl-to-pyrdl" script would be good, where the irdl file is used to create a python file.

I've included some json files for reference, but as they are huge I expect we would want to cache the result of this script in practice.

Quite a few constraints are missing at the moment but this can be iterated on.


alexarice:
Would it likely be better to go with "script that prints out some python code" than what I've done here? Programatically generating the classes seems to introduce some difficulties and makes caching harder

tobiasgrosser:
> Would it likely be better to go with "script that prints out some python code" than what I've done here? Programatically generating the classes seems to introduce some difficulties and makes caching harder

I think that would make a lot of sense.

alexarice:
Just pushed version 2 which just dialect creates a python file. Think this is close to being ready, just need to work out how to test it

alexarice:
Struggling to come up with a way to test this, as I'd ideally like a filecheck test similar to the one for `irdl-to-pyrdl`. The optional I can think of are:
- Include a 17000 line json in the repository (surely not ok)
- Manually cull `test.json` to the parts that are actually relevant
- Dynamically generate a `test.json` (needs access to `llvm-tblgen` and the mlir include files). This has the advantage of testing stuff a bit more end-to-end but I can't think of a good way to ensure access to the include files

Would love to hear any good ideas for this.

tobiasgrosser:
> Manually cull test.json to the parts that are actually relevant

Either this or a manually crafted json file that tests the key features?

AntonLydike (commented):
Couldn't get through all the code yet, but can you add some more comments, and maybe a readme for this tool? Oh, also some minimal tests would be super neat to see what it actually does!

AntonLydike:
@alexarice I think hand-crafting a json example would a good way to test this

index.json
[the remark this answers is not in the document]
alexarice: > I've included some json files for reference, but as they are huge I expect we would want to cache the result of this script in practice.

I intend to get rid of these when this becomes not a draft. It was just so people could see the sort of input the script gets

alexarice:
I've refactored the script a fair bit, added auto-formatting, added a way to cull json files to only the necessary bits, and added a simple test.

xdsl/tools/tblgen_to_py.py
```diff
@@ -0,0 +1,566 @@
...
+    @property
+    def dialect(self) -> str:
+        return self["opDialect"]["def"]
+
+    @property
+    def opName(self) -> str:
+        return self["opName"]
+
+    @property
+    def assemblyFormat(self) -> str | None:
```
math-fehr, line 47:
```suggestion
    def assembly_format(self) -> str | None:
```
And same things on other camelCase functions to make them snake_case (to follow Python style).

xdsl/tools/tblgen_to_py.py
```diff
@@ -0,0 +1,566 @@
...
+
+    def _get_attr(self, name: str) -> TblgenAttr:
+        self.used_records.add(name)
+        return TblgenAttr(self.js[name])
+
+    def _get_record(self, name: str) -> TblgenRecord:
+        self.used_records.add(name)
+        return TblgenRecord(self.js[name])
+
+    def load_dialect(self, dialect_name: str):
```
math-fehr, line 119: I would rename these functions (and the class name), since now they generate a python xDSL dialect rather than loading one.

xdsl/tools/tblgen_to_py.py
```diff
@@ -0,0 +1,566 @@
...
+                )
+                """)  # TODO can't represent APInt properly
+
+            case "StrAttr":
+                return "BaseAttr(StringAttr)"
+            case "SymbolNameAttr":
+                return "BaseAttr(SymbolNameAttr)"
+            case "UnitAttr":
+                return "EqAttrConstraint(UnitAttr())"
+            case _:
```
math-fehr, line 299: Nit: Since anyway all other cases are doing a `return`, you can de indent this one

math-fehr (approved):
Sorry for the delay! I was in the middle of travelling and got a bit sick.
I think this is going in a nice direction.
Is there a reason why you go directly from TableGen, instead of doing the TableGen -> IRDL -> Python route? My guess is that IRDL cannot express what you currently want?
Just so you know, my long-term objective is to have everything passing by IRDL, but I understand that it is not feasible currently, so this tool makes sense to me in the meantime!
I would just add a few comments on some functions, and drop a few comments here and there.

superlopuh (commented):
This is quite a dump of code, no? Is there no way to make this smaller/split the changes up?

xdsl/tools/tblgen_to_py.py
alexarice: I think I mainly kept it like this because pyright wasn't happy if the case was omitted entirely, but happy to dedent

alexarice:
> This is quite a dump of code, no? Is there no way to make this smaller/split the changes up?

I could remove all the argument constraints for a first pass, which would simplify it a little

> Is there a reason why you go directly from TableGen, instead of doing the TableGen -> IRDL -> Python route? My guess is that IRDL cannot express what you currently want?

I agree with the end goal. At least short term my problems with it are:
1. People aren't writing dialects in irdl, and probably won't be soon.
2. There are a lot of constraints that can't be represented nicely in irdl at the moment (e.g. things like the shaped types)
3. To add a new feature to IRDL, you need to update IRDL in mlir, update the tblgen-to-irdl script, and update the xdsl irdl importer. In contrast this only relies on the json tablegen dump which shouldn't need modification.
4. It is still unclear to me how to nicely interface with builtin types in irdl. For example, in mlir the builtin `Complex` type is not parametrized. In this tool we can ignore the mlir representation and just use the xdsl one which *is* parametrized.

xdsl/tools/tblgen_to_py.py
alexarice: I went with `import` rather than `load` but am not really sure if this is better

math-fehr: `generate` could also work maybe?

xdsl/tools/tblgen_to_py.py
math-fehr: If pyright wasn't happy let's not make it angry then ;)

math-fehr:
> > This is quite a dump of code, no? Is there no way to make this smaller/split the changes up?
> 
> I could remove all the argument constraints for a first pass, which would simplify it a little
> 
> > Is there a reason why you go directly from TableGen, instead of doing the TableGen -> IRDL -> Python route? My guess is that IRDL cannot express what you currently want?
> 
> I agree with the end goal. At least short term my problems with it are:
> 
>     1. People aren't writing dialects in irdl, and probably won't be soon.
> 
>     2. There are a lot of constraints that can't be represented nicely in irdl at the moment (e.g. things like the shaped types)
> 
>     3. To add a new feature to IRDL, you need to update IRDL in mlir, update the tblgen-to-irdl script, and update the xdsl irdl importer. In contrast this only relies on the json tablegen dump which shouldn't need modification.
> 
>     4. It is still unclear to me how to nicely interface with builtin types in irdl. For example, in mlir the builtin `Complex` type is not parametrized. In this tool we can ignore the mlir representation and just use the xdsl one which _is_ parametrized.

Yeah I agree, let's do this right now, we'll see later if IRDL turns out to be usable enough to make it pass through it ;)

xdsl/tools/tblgen_to_py.py
```diff
@@ -0,0 +1,583 @@
...
+    arg_parser.add_argument(
+        "-i", "--input_file", required=False, type=str, help="path to input file"
+    )
+    arg_parser.add_argument(
+        "-c",
+        "--cull",
+        action="store_true",
+        help="Output a culled json with only necessary fields",
+    )
+    args = arg_parser.parse_args()
```
superlopuh, line 484: Could you please separate the args parsing from the business logic?

tests/tblgen_to_py/test.py
```diff
@@ -0,0 +1,157 @@
...
+    name = "confined"
+
+    tensor = operand_def(AnyAttr())
+
+    vector = operand_def(AnyAttr())
+
+
+@irdl_op_definition
+class Test_Integers(IRDLOperation):
+    """"""
```
superlopuh, line 92: might as well not print the docstring if it's missing?

tests/tblgen_to_py/test.py
```diff
@@ -0,0 +1,157 @@
...
+    any_integer = operand_def(BaseAttr(IntegerType))
+
+
+@irdl_op_definition
+class Test_OrOp(IRDLOperation):
+    """"""
+
+    name = "or"
+
+    in_ = operand_def(
```
superlopuh, line 112: it feels like we should still have the name be "in" in the `operand_def`

superlopuh (commented):
The actual dialect definition is missing, do we want to include it? Also, the dialect name is missing from all the ops.

It would be great to have more documentation about this, like how the json was generated

xdsl/tools/tblgen_to_py.py
```diff
@@ -0,0 +1,642 @@
...
+    culled: dict[str, Any] = {key: cull_field(js[key]) for key in loader.used_records}
+    culled["!instanceof"] = {
+        key: js["!instanceof"][key] for key in ("TypeDef", "AttrDef", "Op", "Dialect")
+    }
+
+    if output_file is not None:
+        with open(output_file, "w") as out_file:
+            print(json.dumps(culled), file=out_file)
+    else:
+        print(json.dumps(culled))
```
superlopuh, line 537: it feels like the caller should do this dance, and the out parameter here should be `IO[str] | None`

tests/tblgen_to_py/test.py
```diff
@@ -0,0 +1,170 @@
...
+# ruff: noqa: F403, F405
+
+from xdsl.dialects.builtin import *
+from xdsl.ir import *
+from xdsl.irdl import *
+
+
+@irdl_attr_definition
+class Test_SingletonAType(ParametrizedAttribute, TypeAttribute):
+    """"""
```
superlopuh, line 16: Might as well also drop these empty docstrings

alexarice:
Ready for review again

superlopuh (approved):

PR #3213, opened 2024-09-24
transformations: (lower-csl-stencil) Check isBorderRegionPE

n-io:
The `csl_stencil.apply.done_exchange` region is wrapped in an check to `isBorderRegionPE`, such that the compute and (most importantly) the store do not happen for PEs of the border region. While technically possible to wrap only the store, it may be beneficial for code optimisation to wrap all rather than wrap minimally.

### DIFF
--- a/xdsl/transforms/lower_csl_stencil.py
+++ b/xdsl/transforms/lower_csl_stencil.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 
 from xdsl.context import MLContext
-from xdsl.dialects import arith, func, memref
+from xdsl.dialects import arith, func, memref, scf
 from xdsl.dialects.builtin import (
     FunctionType,
     IndexType,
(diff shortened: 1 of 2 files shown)


xdsl/transforms/lower_csl_stencil.py
```diff
@@ -238,13 +270,17 @@ class LowerCslStencil(ModulePass):
...
+      compute and (most importantly) the store do not happen for PEs of the border region. While technically possible
+      to wrap only the store, it may be beneficial for code optimisation to wrap all rather than wrap minimally.
     """
 
     name = "lower-csl-stencil"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
+        PatternRewriteWalker(LowerYieldOp()).rewrite_module(op)
         PatternRewriteWalker(
-            LowerYieldOp(),
```
AntonLydike, line 247: Can these two rewrites not be applied in one pass? :thinking:

AntonLydike (approved):
Cool stuff! Very nice to see progress being made here! :partying_face:

PR #3218, opened 2024-09-26
dialects: (mod_arith) create mod_arith dialect and add initial operation

emmau678:
Create a mod_arith dialect with a basic initial operation, based on the Mod Arith dialect in HEIR.
(link to the original dialect: https://github.com/google/heir/tree/main/lib/Dialect/ModArith)

### DIFF
--- a/tests/filecheck/dialects/mod_arith/mod_arith.mlir
+++ b/tests/filecheck/dialects/mod_arith/mod_arith.mlir
@@ -1,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+"builtin.module"() ({
+  %lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
+
+  %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
+
+  // CHECK: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
+
+}) : () -> ()
--- a/tests/filecheck/dialects/mod_arith/mod_arith.mlir
+++ b/tests/filecheck/dialects/mod_arith/mod_arith.mlir
@@ -4,5 +4,5 @@
 %lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
 
 // CHECK: %add_res = mod_arith.add %lhsi1, %rhsi1 {"modulus" = 17 : i64} : i1
-// CHECK-GENERIC: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) {"modulus" = 17 : i64} : (i1, i1) -> i1
+// CHECK-GENERIC: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) <{"modulus" = 17 : i64}> : (i1, i1) -> i1
 %add_res = "mod_arith.add"(%lhsi1, %rhsi1) {modulus=17} : (i1, i1) -> i1
--- a/tests/filecheck/dialects/mod_arith/mod_arith.mlir
+++ b/tests/filecheck/dialects/mod_arith/mod_arith.mlir
@@ -2,5 +2,5 @@
 
 %lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
 
-// CHECK: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
-%add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
+// CHECK: %add_res = mod_arith.add %lhsi1, %rhsi1 {"modulus" = 17 : i64} : i1
+%add_res = "mod_arith.add"(%lhsi1, %rhsi1) {modulus=17} : (i1, i1) -> i1
--- a/tests/filecheck/dialects/mod_arith/mod_arith.mlir
+++ b/tests/filecheck/dialects/mod_arith/mod_arith.mlir
@@ -1,6 +1,8 @@
 // RUN: XDSL_ROUNDTRIP
+// RUN: XDSL_GENERIC_ROUNDTRIP
 
 %lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
 
 // CHECK: %add_res = mod_arith.add %lhsi1, %rhsi1 {"modulus" = 17 : i64} : i1
+// CHECK-GENERIC: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) {"modulus" = 17 : i64} : (i1, i1) -> i1
 %add_res = "mod_arith.add"(%lhsi1, %rhsi1) {modulus=17} : (i1, i1) -> i1
--- a/tests/filecheck/dialects/mod_arith/mod_arith.mlir
+++ b/tests/filecheck/dialects/mod_arith/mod_arith.mlir
@@ -1,10 +1,6 @@
 // RUN: XDSL_ROUNDTRIP
 
-"builtin.module"() ({
-  %lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
+%lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
 
-  %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
-
-  // CHECK: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
-
-}) : () -> ()
+// CHECK: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
+%add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
--- a/xdsl/dialects/mod_arith.py
+++ b/xdsl/dialects/mod_arith.py
@@ -1,14 +1,14 @@
 from abc import ABC
 
 from xdsl.dialects.arith import signlessIntegerLike
-from xdsl.ir import Attribute, Dialect, SSAValue
+from xdsl.ir import Attribute, Dialect, Operation, SSAValue
 from xdsl.irdl import IRDLOperation, irdl_op_definition, operand_def, result_def
 from xdsl.traits import Pure
 
 
 class ModArithOp(IRDLOperation, ABC):
-    # assemblyFormat = "operands attr-dict `:` functional-type(operands, results)"
     pass
+
 
 class BinaryOp(ModArithOp):
     lhs = operand_def(signlessIntegerLike)
@@ -17,7 +17,12 @@ class BinaryOp(ModArithOp):
     # assemblyFormat ="$lhs $rhs attr-dict `:` type($output)"  # operands directive not in asm format
     traits = frozenset((Pure(),))
 
-    def __init__(self, lhs: SSAValue, rhs: SSAValue, result_type: Attribute | None):
+    def __init__(
+        self,
+        lhs: SSAValue | Operation,
+        rhs: SSAValue | Operation,
+        result_type: Attribute | None,
+    ):
         if result_type is None:
             result_type = SSAValue.get(lhs).type
 
@@ -34,6 +39,6 @@ class AddOp(BinaryOp):
 
 
 ModArith = Dialect(
-    "modarith",
+    "mod_arith",
     [AddOp],
 )
--- a/xdsl/dialects/mod_arith.py
+++ b/xdsl/dialects/mod_arith.py
@@ -25,15 +25,12 @@ class ModArithOp(IRDLOperation, ABC):
     pass
 
 
-_T = TypeVar("_T", bound=Attribute)
-
-
-class BinaryOp(ModArithOp, ABC, Generic[_T]):
+class BinaryOp(ModArithOp, ABC):
     """
     Simple binary operation
     """
 
-    T = Annotated[Attribute, ConstraintVar("T"), _T]
+    T = Annotated[Attribute, ConstraintVar("T"), signlessIntegerLike]
     modulus = prop_def(AnyIntegerAttr)
     lhs = operand_def(T)
     rhs = operand_def(T)
--- a/xdsl/dialects/mod_arith.py
+++ b/xdsl/dialects/mod_arith.py
@@ -10,7 +10,6 @@ class ModArithOp(IRDLOperation, ABC):
     # assemblyFormat = "operands attr-dict `:` functional-type(operands, results)"
     pass
 
-
 class BinaryOp(ModArithOp):
     lhs = operand_def(signlessIntegerLike)
     rhs = operand_def(signlessIntegerLike)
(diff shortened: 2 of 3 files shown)


xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,40 @@
...
+from xdsl.irdl import IRDLOperation, irdl_op_definition, operand_def, result_def
+from xdsl.traits import Pure
+
+
+class ModArithOp(IRDLOperation, ABC):
+    # assemblyFormat = "operands attr-dict `:` functional-type(operands, results)"
+    pass
+
+
+class BinaryOp(ModArithOp):
```
tobiasgrosser, line 14:
```suggestion

class BinaryOp(ModArithOp):
```

tobiasgrosser: Not sure we want double-newlines, but I guess others know the coding style of xDSL better.

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,40 @@
...
+        return super().verify_()
+
+
+@irdl_op_definition
+class AddOp(BinaryOp):
+    name = "mod_arith.add"
+
+
+ModArith = Dialect(
+    "modarith",
```
alexarice, line 38:
```suggestion
    "mod_arith",
```

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,40 @@
+from abc import ABC
+
+from xdsl.dialects.arith import signlessIntegerLike
+from xdsl.ir import Attribute, Dialect, SSAValue
+from xdsl.irdl import IRDLOperation, irdl_op_definition, operand_def, result_def
+from xdsl.traits import Pure
+
+
+class ModArithOp(IRDLOperation, ABC):
+    # assemblyFormat = "operands attr-dict `:` functional-type(operands, results)"
```
alexarice, line 10: Please remove old comments

xdsl/dialects/mod_arith.py
alexarice, line 14: Should this also inherit from `ABC`? (I'm not 100% sure how this works in python so I could be wrong

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class BinaryOp(ModArithOp):
+    lhs = operand_def(signlessIntegerLike)
+    rhs = operand_def(signlessIntegerLike)
+    output = result_def(signlessIntegerLike)
+    # assemblyFormat ="$lhs $rhs attr-dict `:` type($output)"  # operands directive not in asm format
+    traits = frozenset((Pure(),))
+
+    def __init__(self, lhs: SSAValue, rhs: SSAValue, result_type: Attribute | None):
```
alexarice, line 21: In a lot of other dialects the type `SSAValue | Operation` is used so that you can pass in single output operations

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,40 @@
...
+        if result_type is None:
+            result_type = SSAValue.get(lhs).type
```
alexarice, line 23: Does `lhs.type` not work?

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,40 @@
...
+class ModArithOp(IRDLOperation, ABC):
+    # assemblyFormat = "operands attr-dict `:` functional-type(operands, results)"
+    pass
+
+
+class BinaryOp(ModArithOp):
+    lhs = operand_def(signlessIntegerLike)
+    rhs = operand_def(signlessIntegerLike)
+    output = result_def(signlessIntegerLike)
+    # assemblyFormat ="$lhs $rhs attr-dict `:` type($output)"  # operands directive not in asm format
```
alexarice, line 18: What does this comment mean? Does the assembly format not work?

alexarice (commented):
Left a few comments but looks like a good start

xdsl/dialects/mod_arith.py
kimxworrall: We copied this from a function in arith - I was presuming it forces the generation of a new SSAValue

xdsl/dialects/mod_arith.py
emmau678: I'm not fully sure, do I need to include it again when ModArithOp inherits from ABC?

AntonLydike: Yes, in general, every class that is "incomplete" (i.e. should not be instantiated itself but only serves to encapsulate common abstractions) should ideally be marked `ABC` to make it a type error when someone tries to instantiate them.

xdsl/dialects/mod_arith.py
AntonLydike: Hmm, I think the problem is that the magic fairy dust that is the assembly format printer/parser cannot infer all types with this format. If you want `$lhs`, `$rhs` and `$output` to always have the same type you can use a type variable to achieve this. Happy to show you how that's done!

xdsl/dialects/mod_arith.py
AntonLydike: `SSAValue.get(...)` in needed to convert from `SSAValue | Operation` to `SSAValue`, so this is the correct approach!

AntonLydike (commented):
Welcome @emmau678! How exciting to see the project grow, this is a very strong start already!

superlopuh:
Can you please delete the "Thank you for submitting a PR to xDSL!" bit from the PR description? The PR descriptions get added as commit messages when merged, so it's good to craft these to be helpful when looking at this commit in git history in the future.

tests/filecheck/dialects/mod_arith/mod_arith.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+"builtin.module"() ({
+  %lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)
+
+  %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
+
+  // CHECK: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
+
+}) : () -> ()
```
superlopuh, line 10: I'd suggest a few changes:
```suggestion
// RUN: XDSL_ROUNDTRIP

%lhsi1, %rhsi1 = "test.op"() : () -> (i1, i1)

// CHECK: %add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
%add_res = "mod_arith.add"(%lhsi1, %rhsi1) : (i1, i1) -> i1
```
The builtin.module is implied, so we can skip that entirely.
The LLVM approach is usually to add check lines before the thing, as opposed to after. We're not consistent with this, but I've transitioned to this format now, and would recommend it also.

tobiasgrosser:
> Can you please delete the "Thank you for submitting a PR to xDSL!" bit from the PR description? The PR descriptions get added as commit messages when merged, so it's good to craft these to be helpful when looking at this commit in git history in the future.

Actually, I believe that everything below the line won't be included. But it probably does not hurt to clean this up.

xdsl/dialects/mod_arith.py
superlopuh: This is entirely handled by Ruff :)

xdsl/dialects/mod_arith.py
emmau678: that would be great, thanks @AntonLydike

alexarice: Bit of an aside but doesn't `$operands` expand to something more like 
```
`(` $lhs `,` $rhs `)`
```
?

AntonLydike (approved):
:rocket:

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,71 @@
...
+class BinaryModuloOp(BinaryOp[_T], ABC, Generic[_T]):
+    """
+    Adds the modulus attribute
+    """
+
+    modulus = attr_def(AnyIntegerAttr)
+
+
+@irdl_op_definition
+class AddOp(BinaryModuloOp[Annotated[Attribute, signlessIntegerLike]]):
```
alexarice, line 64: Is there a reason for putting the bound here instead of on the superclass as upstream does?

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,71 @@
...
+        rhs: SSAValue | Operation,
+        result_type: Attribute | None,
+    ):
+        if result_type is None:
+            result_type = SSAValue.get(lhs).type
+
+        super().__init__(operands=[lhs, rhs], result_types=[result_type])
+
+
+class BinaryModuloOp(BinaryOp[_T], ABC, Generic[_T]):
```
alexarice, line 55: Why make another class here where upstream just has `ModArith_Op` > `ModArith_BinaryOp` > `ModArith_AddOp`

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Adds the modulus attribute
+    """
+
+    modulus = attr_def(AnyIntegerAttr)
```
alexarice, line 60: Tablegen "attributes" are actually properties (not sure if you were trying to be compatible.

alexarice (commented):
Some small comments, probably won't block merging

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@irdl_op_definition
+class AddOp(BinaryModuloOp[Annotated[Attribute, signlessIntegerLike]]):
+    name = "mod_arith.add"
+
+
+ModArith = Dialect(
+    "mod_arith",
+    [AddOp],
```
superlopuh, line 70: Presumably, there will be more ops
```suggestion
    [AddOp,],
```

xdsl/dialects/mod_arith.py
AntonLydike: Yeah, upstream has modulus as an attribute not a property

xdsl/dialects/mod_arith.py
AntonLydike: Hmm, not 100% sure what you mean, but there are `ModArith` ops that don't have a modulus.

I was thinking we could eventually model `HasModulus` as a mixin instead of a subclass. I guess this becomes relevant once we get the unary and ternary operators in?

xdsl/dialects/mod_arith.py
AntonLydike: No particular reason, I guess this becomes visible as duplicated code once we add more ops to the dialect.

xdsl/dialects/mod_arith.py
alexarice: In the file I'm looking at, all `ModArith_BinaryOp` have a modulus (maybe I'm looking at an out of date file

xdsl/dialects/mod_arith.py
alexarice: In [here](https://github.com/google/heir/blob/main/lib/Dialect/ModArith/IR/ModArithOps.td) I am 90% sure it is a property

AntonLydike: Seeing the assembly format only mentions `attr-dict`:

> `let assemblyFormat = "operands attr-dict `:` type($output)";`

[ModArithOps.td:75](https://github.com/google/heir/blob/4ca0a98a682a0f2f2569bca50007a3b2d5e50ca6/lib/Dialect/ModArith/IR/ModArithOps.td#L75C1-L75C63)

And the op is printed as `%reduce = mod_arith.reduce %c4 { modulus = 17 } : i10` in [tests/mod_arith/syntax.mlir:37](https://github.com/google/heir/blob/4ca0a98a682a0f2f2569bca50007a3b2d5e50ca6/tests/mod_arith/syntax.mlir#L37)

I'm 99.9% sure it's an attribute.

alexarice: I tried to install the tool for an hour and failed, but the real test will be printing generic syntax

alexarice: It took me all evening and installing a virtual machine but can confirm that it is a property:
image

emmau678: I'm not fully clear on what I need to do here - should I change attr_def to prop_def?

docs/marimo/linalg_snitch.py
superlopuh, line 1: please revert this change

docs/marimo/mlir/onnx_demo.py
superlopuh, line 1: same here

pyproject.toml
superlopuh, line 1: this seems unrelated

superlopuh:
Something's off, did a merge go wrong?

alexarice:
> Something's off, did a merge go wrong?

image
Kind of looks like you did this

emmau678:
@superlopuh @alexarice yes, I ran git pull when I should not have, I'm working on it :)

emmau678:
> @superlopuh @alexarice yes, I ran git pull when I should not have, I'm working on it :)

I think I have fixed it now

alexarice:
The merge looks fixed. Just the one comment still up then should be good to merge from my end

alexarice (approved):

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,74 @@
...
+
+_T = TypeVar("_T", bound=Attribute)
+
+
+class BinaryOp(ModArithOp, ABC, Generic[_T]):
+    """
+    Simple binary operation
+    """
+
+    T = Annotated[Attribute, ConstraintVar("T"), _T]
```
superlopuh, line 36: Is there any future where the type isn't `signlessIntegerLike`? I'm doing some related work at the moment, and it would make it easier if we could specialize as part of this PR, something like this:
```suggestion
class BinaryOp(ModArithOp, ABC):
    """
    Simple binary operation
    """

    T = Annotated[Attribute, ConstraintVar("T"), signlessIntegerLike]
```

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,74 @@
...
+
+        super().__init__(
+            operands=[lhs, rhs],
+            result_types=[result_type],
+            properties={"modulus": modulus},
+        )
+
+
+@irdl_op_definition
+class AddOp(BinaryOp[Annotated[Attribute, signlessIntegerLike]]):
```
superlopuh, line 65:
```suggestion
class AddOp(BinaryOp):
```
remaining change if you acept the one above

xdsl/dialects/mod_arith.py
```diff
@@ -0,0 +1,74 @@
...
+    irdl_op_definition,
+    operand_def,
+    prop_def,
+    result_def,
+)
+from xdsl.traits import Pure
+
+
+class ModArithOp(IRDLOperation, ABC):
+    pass
```
superlopuh, line 25: I'm not sure that this is necessary, I'd advise against abstracting things before there's a real use for it.

superlopuh (approved):

PR #3224, opened 2024-09-27
[WIP]: Jax tensor donation

mamanain:
This is still work in progress, just wanted to get some feedback for this version as well.

The main area of improvement: donation logic should be more in line with the jax [documentation](https://jax.readthedocs.io/en/latest/faq.html#buffer-donation).
There are currently two problems here:
1) Donated buffers can still be used in the function but here it can be used as a buffer straight away. This can lead to data being overwritten and then used for some computations which will lead to errors. So we need to change the logic so that the buffer becomes available only after its last usage.
2) Right now buffer can be used only once. Should we keep it in the dictionary and reuse in other places where it fits?


superlopuh (commented):
Thank you for this, it's a great first step! A few bigger comments before the nitty-gritty:
 - I'm a little confused by the name of the pass, it's not really converting JAX to linalg, just making sure that the operands marked as donating are actually reused as destinations, so another name feels more appropriate. (I would recommend making the prefix of the pass something that's not convert, as that's usually used for converting one dialect to another, and in this case it's more of an optimisation.) Maybe `jax-use-donated-arguments`?
 - The test is quite specific, and uses operations from dialects that aren't strictly involved. I would recommend using the "test" dialect as much as possible to generate values that are necessary for the test.
 - It's quite a bit more difficult to review PRs with lots of Pyright errors, could you please fix the errors locally and ping me again to take a look?
 - It would be good to be very clear about the limitations of the proposed approach, and to raise helpful messages if used in an unexpected context, could you please add `DiagnosticException`s if the assumptions are not met and tests for those?

xdsl/tools/command_line_tool.py
```diff
@@ -86,6 +86,11 @@ def get_convert_stencil_to_ll_mlir():
 
         return convert_stencil_to_ll_mlir.ConvertStencilToLLMLIRPass
 
+    def get_jax_use_donated_arguments():
+        from xdsl.transforms.experimental import jax_use_donated_arguments
```
superlopuh, line 90: I would recommend against experimental, it's a place for untested code, that we will probably remove at some point, let's just put it in the main transforms folder

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,61 @@
...
+class SubstituteDonatedTensors(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: FuncOp, rewriter: PatternRewriter, /):
+        if op.arg_attrs is None:
+            return
+
+        donated_inputs: list[BlockArgument] = []
+        for inp, attr in zip(op.args, op.arg_attrs):
+            if type(inp.type) is TensorType and "tf.aliasing_output" in attr.data:
+                donated_inputs.append(inp)
```
superlopuh, line 30: Does this work?
```suggestion
        donated_inputs = [
            inp
            for inp, attr in zip(op.args, op.arg_attrs)
            if isinstance(inp.type, TensorType) and "tf.aliasing_output" in attr.data
        ]
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        the_one_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier([SubstituteDonatedTensors()]),
+            apply_recursively=False,
+            walk_reverse=True,
+            walk_regions_first=True,
+        )
+        the_one_pass.rewrite_module(op)
+        MLIROptPass(arguments=("--linalg-fuse-elementwise-ops",)).apply(ctx, op)
```
superlopuh, line 61: This should be done separately by the user, and not included in this pass

tests/filecheck/mlir-conversion/with-mlir/jax-use-donated-arguments.mlir
superlopuh, line 1: there's no need to put the test in the with-mlir folder, it's better to separate it into a pure xDSL thing, and let the user call into mlir-opt if they want to

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -14,60 +14,53 @@
...
     @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: FuncOp, rewriter: PatternRewriter, /):
-        if op.arg_attrs is None:
+    def match_and_rewrite(self, op: Return, rewriter: PatternRewriter, /):
+        func_op = op.parent_op()
+        if func_op is None:
+            raise VerifyException("Return operation should be tied to a FuncOp")
+
+        arg_attrs = getattr(func_op, "arg_attrs")
+        args = getattr(func_op, "args")
```
superlopuh, line 33: we try to avoid getattr when possible, could you please check that the parent is a func, and then access properties directly?

xdsl/dialects/tensor.py
```diff
@@ -176,6 +180,25 @@ def parse(cls, parser: Parser) -> Self:
...
+    name = "tensor.collapse_shape"
+
+    src = operand_def(TensorType[Attribute])
+    result = result_def(TensorType[Attribute])
+    reassociation = prop_def(ReassociationAttr)
+    assembly_format = (
+        "$src $reassociation attr-dict `:` type($src) `into` type($result)"
+    )
+
+    traits = frozenset([NoMemoryEffect()])
```
superlopuh, line 199: I don't think we need this as part of this PR

superlopuh: Would be great to add in a separate PR, though

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,65 @@
...
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.utils.exceptions import VerifyException
+
+
+def make_materialize_op(source: SSAValue, dest: SSAValue) -> MaterializeInDestination:
+    return MaterializeInDestination(operands=[source, dest], result_types=[source.type])
```
superlopuh, line 21: I don't think we need this

tests/filecheck/transforms/jax-use-donated-arguments.mlir
```diff
@@ -0,0 +1,55 @@
+// RUN: xdsl-opt %s -p jax-use-donated-arguments --split-input-file --verify-diagnostics | filecheck %s
+
+builtin.module {
+func.func public @main(%arg0: tensor<2x3xf32>, %arg1: tensor<3x4xf32>, %arg2: tensor<2x4xf32> {tf.aliasing_output = 0 : i32}) -> (tensor<2x4xf32>) {
+    %0 = tensor.empty() : tensor<2x4xf32>
+    %cst = arith.constant 0.000000e+00 : f32
+    %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<2x4xf32>) -> tensor<2x4xf32>
```
superlopuh, line 7: We try to make the tests as targeted as possible. In this case, the fact that the results are produced by linalg is incidental, and any other operation should work just as well.
```suggestion
    %res = "test.op"() : () -> tensor<2x4xf32>
```
Similarly below

xdsl/dialects/builtin.py
```diff
@@ -796,6 +796,19 @@ def get_shape(self) -> tuple[int, ...]:
...
+        other_shape = list(other_tensor.shape)
+        if len(current_shape) != len(other_shape):
+            return False
+
+        return (
+            len(list(filter(lambda x: x[0] != x[1], zip(current_shape, other_shape))))
+            == 0
+            and self.element_type == other_tensor.element_type
+            and self.encoding == other_tensor.encoding
+        )
```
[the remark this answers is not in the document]
mamanain: for some reason I thought that it didn't work, but it does... I guess it didn't work for some other reason at that time

tests/filecheck/transforms/jax-use-donated-arguments.mlir
superlopuh, line 1: could  you please add a test with two arguments of the same type being donated and returned, to check that we don't incorrectly add aliasing?

xdsl/dialects/bufferization.py
```diff
@@ -168,12 +168,33 @@ class ToMemrefOp(IRDLOperation):
     assembly_format = "$tensor (`read_only` $read_only^)?  `:` attr-dict type($memref)"
 
 
+# now only works for (tensor, tensor) arguments. need to add memref support as well.
+@irdl_op_definition
+class MaterializeInDestination(IRDLOperation):
```
superlopuh, line 173: Can you please add this in a separate PR?

tests/filecheck/transforms/jax-use-donated-arguments.mlir
```diff
@@ -0,0 +1,57 @@
...
+// CHECK: builtin.module {
+// CHECK-NEXT: builtin.module {
+// CHECK-NEXT:   func.func public @main(%arg0 : tensor<2x3xf32>, %arg1 : tensor<3x4xf32>, %arg2 : tensor<2x4xf32> {"tf.aliasing_output" = 0 : i32}) -> tensor<2x4xf32> {
+// CHECK-NEXT:      %res = "test.op"() : () -> tensor<2x4xf32>
+// CHECK-NEXT:      %0 = bufferization.materialize_in_destination %res in %arg2 : (tensor<2x4xf32>, tensor<2x4xf32>) -> tensor<2x4xf32>
+// CHECK-NEXT:      func.return %0 : tensor<2x4xf32>
+// CHECK-NEXT:   }
+// CHECK-NEXT: }
+
+builtin.module {
```
superlopuh, line 19:
```suggestion
```

And in other tests

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,66 @@
...
+        func_op = op.parent_op()
+        if func_op is None or type(func_op) is not FuncOp:
+            raise VerifyException("Return operation should be tied to a FuncOp")
+
+        if func_op.arg_attrs is None:
+            return
+
+        donated_inputs = [
+            inp
+            for inp, attr in zip(func_op.args, func_op.arg_attrs)
```
superlopuh, line 33:
```suggestion
            for inp, attr in zip(func_op.args, func_op.arg_attrs, strict=True)
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,66 @@
...
+)
+from xdsl.utils.exceptions import VerifyException
+
+
+@dataclass
+class SubstituteDonatedTensors(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: Return, rewriter: PatternRewriter, /):
+        func_op = op.parent_op()
+        if func_op is None or type(func_op) is not FuncOp:
```
superlopuh, line 25:
```suggestion
        assert isinstance(func_op, FuncOp)
```
The input can be assumed to be verified. There's no way to handle a VerifyException here anyway, so we might as well assert.

superlopuh:
As per our discussion, I'm not sure that it's a good idea to merge this until we find a way to get to linalg from JAX again... Let's close the PR for now, keep the changes in the branch, and reopen when we figure out a new way to get to the kernels we want?

mamanain:
Yeah, we can close it for now. It turned out to be a known [issue](https://github.com/openxla/stablehlo/issues/2597#issuecomment-2429571058) in stablehlo. I think, I'll be able to send a pr to them to fix it this week.

superlopuh:
Closing for now

PR #3229, opened 2024-09-30
transformations: New stencil-shape-minimize pass

n-io:


xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -0,0 +1,73 @@
...
+        self.seen.add(typ)
+
+
+@dataclass
+class ShapeMinimisation(TypeConversionPattern):
+    shape: stencil.StencilBoundsAttr | None = None
+
+    @attr_type_rewrite_pattern
+    def convert_type(self, typ: stencil.FieldType[Attribute], /) -> Attribute | None:
+        if typ.bounds != self.shape and self.shape:
```
dk949, line 33:
```suggestion
    shape: stencil.StencilBoundsAttr

    @attr_type_rewrite_pattern
    def convert_type(self, typ: stencil.FieldType[Attribute], /) -> Attribute | None:
        if typ.bounds != self.shape:
```


Since we always pass the shape on construction anyway

dk949 (approved):
Looks good

xdsl/transforms/stencil_shape_minimize.py
n-io: That unfortunately doesn't seem to work with `dataclass`

dk949: Ohh, `TypeConversionPattern` has default values, I didn't realise that.

PR #3231, opened 2024-10-01
transformations: adds a single pass from `stencil` to `csl`

dk949:
This pass combines (using a `PipelinePass`) all previously written passes and makes it easier to keep track of the order in which the passes need to be ran.

### DIFF
--- a/xdsl/tools/command_line_tool.py
+++ b/xdsl/tools/command_line_tool.py
@@ -400,11 +400,6 @@ def get_all_passes() -> dict[str, Callable[[], type[ModulePass]]]:
 
         return convert_stencil_to_csl_stencil.ConvertStencilToCslStencilPass
 
-    def get_convert_stencil_to_csl():
-        from xdsl.transforms import convert_stencil_to_csl
-
-        return convert_stencil_to_csl.ConvertStencilToCslPass
-
     def get_stencil_unroll():
         from xdsl.transforms import stencil_unroll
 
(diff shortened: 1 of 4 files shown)


xdsl/tools/command_line_tool.py
```diff
@@ -442,6 +447,7 @@ def get_stencil_bufferize():
         "convert-scf-to-riscv-scf": get_convert_scf_to_riscv_scf,
         "convert-snitch-stream-to-snitch": get_convert_snitch_stream_to_snitch,
         "convert-stencil-to-csl-stencil": get_convert_stencil_to_csl_stencil,
+        "convert-stencil-to-csl": get_convert_stencil_to_csl,
```
superlopuh, line 450: I would recommend calling it `test-convert-stencil-to-csl`, to minimise the confusion of alternative ways of getting to places. Ideally the csl work would actually have its own driver, and not use xdsl-opt, something like the `toy` command line tool that is toy-specific.

dk949: Renamed as suggested. 

Though, this is less of a driver, and more just a way to keep track of the order of passes developed so far.

AntonLydike: > Though, this is less of a driver, and more just a way to keep track of the order of passes developed so far.

Well... Isn't that what the "driver" is though?

This would also allow us to add some simple logic to print the layout and program to multiple files and such. I guess we should discuss how to build actual compiler tools with xDSL (where you get an interface like `stencilcc --march=cs2 my_input_file.mlir --out_dir=out` that will write multiple files into `./out` for the csl stencil work)

superlopuh:
What's the latest on this? It seems useful to me. I would just advise outlining the pass pipeline to a tuple available at the top of the file, I've found it useful for the equivalent riscv functionality.

superlopuh:
Closing due to no activity for months, please reopen if you would still like to merge it

PR #3235, opened 2024-10-02
core: Block predecessors

alexarice:
Adds a list of uses to blocks, and adds a function to get the predecessors of a block.

I'm slightly unsure about the code re-use in this PR. It feels slightly wrong to use the `Use` class for both operand usage and successor usage, though the data is the same. I'm also unsure if I should have attempted to unify `OpOperands` and `OpSuccessors` under a single interface, as they are very similar.

### DIFF
--- a/tests/test_use.py
+++ b/tests/test_use.py
@@ -1,8 +1,6 @@
 from xdsl.context import MLContext
 from xdsl.dialects.builtin import Builtin, ModuleOp
-from xdsl.dialects.cf import Cf
-from xdsl.dialects.func import Func, FuncOp
-from xdsl.dialects.test import Test
+from xdsl.dialects.test import Test, TestOp
 from xdsl.ir import Use
 from xdsl.parser import Parser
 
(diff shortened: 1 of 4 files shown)


xdsl/ir/core.py
```diff
@@ -1756,6 +1779,50 @@ def __next__(self):
         return next_block
 
 
+@dataclass
+class OpSuccessors(Sequence[Block]):
```
math-fehr, line 1783: Really cool!

tests/test_use.py
```diff
@@ -28,3 +30,37 @@ def test_main():
...
+
+    func.verify()
+    assert isinstance(func, FuncOp)
+
+    block1, block2, block3, block4 = func.body.blocks
+
+    assert block1.predecessors() == ()
+    assert block2.predecessors() == (block1,)
+    assert block3.predecessors() == (block1,)
+    assert set(block4.predecessors()) == set((block2, block3))
```
superlopuh, line 66: meganit:
```suggestion
    assert set(block4.predecessors()) == {block2, block3}
```

tests/test_use.py
```diff
@@ -28,3 +30,37 @@ def test_main():
...
+}
+"""
+
+
+def test_predecessor():
+    ctx = MLContext()
+    ctx.load_dialect(Builtin)
+    ctx.load_dialect(Test)
+    ctx.load_dialect(Func)
+    ctx.load_dialect(Cf)
```
superlopuh, line 53: nit:
it feels like it would be good to do it all with just the test dialect, without `func` and `cf`.

superlopuh (approved):

Issue #2876, opened 2024-07-10
Wrong values on large dense arrays after xdsl-opt


gabrielrodcanal:
Hi there,
I have found a strange behaviour when running an MLIR code through xdsl-opt (no passes). The code is a NN that has very large dense arrays, and so they are formatted as a string in hex. After running the code through xdsl-opt however, it replaces it by a single float, even though the tensor is multidimensional. A minimal example:
```
"builtin.module"() ({
[...]

PR #3239, opened 2024-10-03
core: Fix parsing for hex denseAttrs

vinayakdsci:
Fixes https://github.com/xdslproject/xdsl/issues/2876.

The original parsing implementation for hex denseAttrs uses format strings in `struct.unpack_from()`, but the format strings actually do not include the count of the chunks contained in the bytes list.

This patch fixes this behavior by embedding the number of chunks in the respective format string.


vinayakdsci:
cc @superlopuh @gabrielrodcanal @JosseVanDelm.

superlopuh:
Could you please add a unit test that fails before your patch, and works after?

vinayakdsci:
@superlopuh Added.

superlopuh (approved):
Very nice, thank you!

JosseVanDelm:
Thanks @vinayakdsci !

PR #3241, opened 2024-10-03
dialects: (varith) Add varith (variadic arithmetic) dialect

AntonLydike:
Add a variadic arithmetic dialect.

The goal is to provide a place for nice-to-use arithmetic operations. Also makes re-writing arithmetic easier in some cases, e.g. when you want to split a summation over one set of values into two summations by selecting certain values from the set.

Coming next is two rewrites to canonicalize arith to varith, and then to de-canonicalize back.

### DIFF
--- a/xdsl/dialects/varith.py
+++ b/xdsl/dialects/varith.py
@@ -1,0 +1,76 @@
+from typing import Annotated
+
+from xdsl.dialects.builtin import (
+    BFloat16Type,
+    ContainerOf,
+    Float16Type,
+    Float32Type,
+    Float64Type,
+    Float80Type,
+    Float128Type,
+    IndexType,
+    IntegerType,
+)
+from xdsl.ir import Attribute, Dialect, Operation, SSAValue
+from xdsl.irdl import (
+    AnyOf,
+    ConstraintVar,
+    IRDLOperation,
+    irdl_op_definition,
+    result_def,
+    var_operand_def,
+)
+from xdsl.traits import Pure
+
+integerOrFloatLike: ContainerOf = ContainerOf(
+    AnyOf(
+        [
+            IntegerType,
+            IndexType,
+            BFloat16Type,
+            Float16Type,
+            Float32Type,
+            Float64Type,
+            Float80Type,
+            Float128Type,
+        ]
+    )
+)
+
+
+class VarithOp(IRDLOperation):
+    """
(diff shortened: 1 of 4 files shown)


n-io (approved):

superlopuh:
Interesting! What's the advantage of this? Smaller IR? I'm a little worried that it might be more difficult to reason about the individual arithmetic operations that would make up a variadic one

n-io:
> Interesting! What's the advantage of this? Smaller IR? I'm a little worried that it might be more difficult to reason about the individual arithmetic operations that would make up a variadic one

Imo it looks quite useful for restructuring larger computations.

As for specifying what arithmetic operation made up the variadic one, linalg.reduce solves this by having a region:

```
  %reduce = linalg.reduce
      ins(%input:tensor<16x32x64xf32>)
      outs(%init:tensor<16x64xf32>)
      dimensions = [1]
      (%in: f32, %out: f32) {
        %0 = arith.addf %out, %in: f32
        linalg.yield %0: f32
      }
```

Though having separate ops seems quite straightforward actually.

AntonLydike:
> Interesting! What's the advantage of this? Smaller IR? I'm a little worried that it might be more difficult to reason about the individual arithmetic operations that would make up a variadic one

In our usecase, we have a giant sum that we need to split into two. This is much harder if you have a taller def-use tree vs a shorter one:

```
arith:
%a  %b
 \ /  
  +--- %c
  |
  +--- %d
  |
  +--- %c
  .
  .
  .
  %res


varith:

%a %b %c &d ... %z
 |  |  |  |     |
 ----------------
        |
        %res
``` 

This allows us to basically "filter" the `varith.add` operands to either be "dirty" or "clean", and move them around accordingly, instead of having a weird tree-like graph.

superlopuh:
Interesting! I'm curious to see where this goes

xdsl/dialects/varith.py
```diff
@@ -0,0 +1,76 @@
...
+@irdl_op_definition
+class VarithMulOp(VarithOp):
+    name = "varith.mul"
+
+
+Varith = Dialect(
+    "varith",
+    [
+        VarithAddOp,
+        VarithMulOp,
```
n-io, line 74: How about `VarithSumOp` and `VarithProductOp`?

n-io (approved):

PR #3245, opened 2024-10-04
rewriting: Move dce to rewriter

alexarice:
In mlir, dead code elimination seems to be part of the rewriter itself. I am planning to write a full (region-level) dead code elimination pass but as a start I have moved the "trivial" dead code elimination to the rewriter. I think this change will make more sense with the full pass but I thought this was already enough for a PR.

This also makes canonicalization a little bit more aggressive as dead code elim and canonicalization patterns can be interleaved. This has changed one of the stencil tests so it would be good to get a confirmation that this is OK.

Looks like a lot of changes but it's mostly moving code around which git hasn't managed to keep track of.

### DIFF
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -23,6 +23,7 @@ from xdsl.ir import (
     SSAValue,
 )
 from xdsl.rewriter import InsertPoint, Rewriter
+from xdsl.transforms.dead_code import is_trivially_dead
 from xdsl.utils.hints import isa
 
 
@@ -381,6 +382,15 @@ class RewritePattern(ABC):
         ...
 
 
+class NoRewrite(RewritePattern):
+    """
+    Rewrite pattern which performs no rewrite.
+    """
+
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter):
+        pass
+
+
 _RewritePatternT = TypeVar("_RewritePatternT", bound=RewritePattern)
 _OperationT = TypeVar("_OperationT", bound=Operation)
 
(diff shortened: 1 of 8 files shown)


xdsl/pattern_rewriter.py
```diff
@@ -650,7 +660,7 @@ class PatternRewriteWalker:
     The walker will also walk recursively on the created operations.
     """
 
-    pattern: RewritePattern
+    pattern: RewritePattern = field(default_factory=NoRewrite)
```
superlopuh, line 663: Why this change?

xdsl/transforms/dead_code.py
superlopuh, line 1: Why a new file?

alexarice: Import loops

xdsl/pattern_rewriter.py
alexarice: It made the toy example a bit neater, but it could be dropped

superlopuh: My gut says drop it FWIW

xdsl/transforms/dead_code.py
superlopuh: If it's just the PatternRewriteWalker I would add the import inside the dce function and call it a day, it feels a little messy to split this file

alexarice:
I think a better approach to doing this will be to keep dce separate from the rewrite machinery, but add an option on the pattern rewriter to perform a module pass at the end of each rewrite phase. This allows the same functionality as the mlir version but is a bit cleaner in my opinion and much more modular.

I'm going to close this PR in favor of doing the above.

Issue #3154, opened 2024-09-10
Function _parse_builtin_densearray_attr fails when parsing a dense array with element_type IndexType 


watermelonwolverine:
The function `_parse_builtin_densearray_attr` does not accept dense arrays with element_type `IndexType`. On the other hand, the function `DenseArrayBase.create_dense_int_or_index` does accept `IndexType` as element_type.

superlopuh:
Tagged as "good first issue", I expect resolving this is looking up what MLIR does for this, and making sure that both the functions do the expected thing.

PR #3247, opened 2024-10-06
(bug): Teach densearray parse function to accept IndexType

vinayakdsci:
Resolves https://github.com/xdslproject/xdsl/issues/3154.

This patch eases the type constraints in `_parse_builtin_densearray_attr()` to include `IndexType`.
Also adds a test for the change.

### DIFF
--- a/tests/filecheck/dialects/builtin/attrs.mlir
+++ b/tests/filecheck/dialects/builtin/attrs.mlir
@@ -19,24 +19,26 @@
     %x7 = "arith.constant"() {"value" = 0 : i64, "test" = array<i32: 2, 3, 4>} : () -> i64
     // CHECK:  "test" = array<f32: 2.1, 3.2, 4.3>
     %x8 = "arith.constant"() {"value" = 0 : i64, "test" = array<f32: 2.1, 3.2, 4.3>} : () -> i64
+    // CHECK:  "test" = array<index: 2, 3, 4>
+    %x9 = "arith.constant"() {"value" = 0 : i64, "test" = array<index: 2, 3, 4>} : () -> i64
     // CHECK:  "test" = #builtin.signedness<signless>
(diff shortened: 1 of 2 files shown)


vinayakdsci:
cc @superlopuh @watermelonwolverine.

superlopuh (changes requested):
```
❯ cat hello.mlir
%x9 = "arith.constant"() {"value" = 0 : i64, "test" = array<index: 2, 3, 4>} : () -> i64
❯ mlir-opt hello.mlir
Assertion failed: (isIntOrFloat() && "only integers and floats have a bitwidth"), function getIntOrFloatBitWidth, file Types.cpp, line 124.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
...
```

MLIR segfaults with this, I'm not sure we should support it if MLIR doesn't. This makes sense to me, as dense is for fixed-bitwidth dense data, and index is platform-dependent.

vinayakdsci:
@superlopuh hi, sorry for the late reply. I understand your point, and agree with it. But this would also mean that the type constraints on `DenseArrayBase.create_dense_int_or_index` should then be tightened to not accept an Index type, if I understand the situation correctly?

superlopuh:
It seems like it, that would be a great change to make if you have the bandwidth to implement it

vinayakdsci:
> It seems like it, that would be a great change to make if you have the bandwidth to implement it

Sure can!  Should I also rename the function to remove the `index` part?

vinayakdsci:
> Sure can! Should I also rename the function to remove the `index` part?

OK, I have renamed the function to remove int (in a new PR). This is so because otherwise the function's name would be misleading, as it mentions IndexType, but does not accept it.

vinayakdsci:
This PR has errors in its approach, which are fixed by https://github.com/xdslproject/xdsl/pull/3258. Therefore, closing this PR to follow up on the latter.

PR #3253, opened 2024-10-07
transforms: (dce) region-level dce

alexarice:
Introduces a region level algorithm for dead-code elimination, which is capable of deleting unused operations which occur in cycles and deleting blocks. It is modeled on the [mlir implementation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Transforms/Utils/RegionUtils.cpp). While the mlir implementation separates this into two separate passes (removing dead blocks and then removing dead operations) they have a todo comment for merging them, and so I have implemented it here as a single pass.

I have not implemented removal of block argument removal, for the sake of keeping this PR smaller and my own sanity, as it seems to add a fair bit of complexity.

### DIFF
--- a/xdsl/ir/post_order.py
+++ b/xdsl/ir/post_order.py
@@ -1,0 +1,29 @@
+from collections import deque
+from collections.abc import Iterator
+from dataclasses import dataclass
+
+from xdsl.ir import Block, Operation
+from xdsl.traits import IsTerminator
+
+
+@dataclass(init=False)
+class PostOrderIterator(Iterator[Block]):
+    stack: deque[Block]
+    seen: set[Block]
+
+    def __init__(self, block: Block) -> None:
+        self.stack = deque((block,))
+        self.seen = {block}
+
+    def __iter__(self) -> Iterator[Block]:
+        return self
+
+    def __next__(self) -> Block:
+        if not self.stack:
+            raise StopIteration
+        block = self.stack.popleft()
+        term = block.last_op
+        if isinstance(term, Operation) and term.has_trait(IsTerminator()):
+            self.stack.extend(x for x in term.successors if x not in self.seen)
+            self.seen.update(term.successors)
+        return block
--- a/xdsl/ir/post_order.py
+++ b/xdsl/ir/post_order.py
@@ -1,4 +1,3 @@
-from collections import deque
 from collections.abc import Iterator
 from dataclasses import dataclass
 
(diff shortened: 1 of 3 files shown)


xdsl/ir/post_order.py
```diff
@@ -0,0 +1,29 @@
...
+from dataclasses import dataclass
+
+from xdsl.ir import Block, Operation
+from xdsl.traits import IsTerminator
+
+
+@dataclass(init=False)
+class PostOrderIterator(Iterator[Block]):
+    stack: deque[Block]
+    seen: set[Block]
```
superlopuh, line 12: Could you add some dostrings, please?

Also, maybe `stack` should be `queue`?

xdsl/transforms/dead_code_elimination.py
```diff
@@ -11,18 +14,21 @@
 )
 
 
-def is_trivially_dead(op: Operation):
-    """
-    Returns if the operation has no observable effect.
-    """
+def would_be_trivially_dead(op: Operation):
     return (
```
superlopuh, line 18: Can you add a doc string, please?

xdsl/transforms/dead_code_elimination.py
```diff
@@ -73,8 +79,80 @@ def dce(op: ModuleOp):
...
+    return live_set.changed
+
+
 class DeadCodeElimination(ModulePass):
     name = "dce"
 
     def apply(self, ctx: MLContext, op: ModuleOp) -> None:
-        dce(op)
+        for region in op.regions:
+            region_dce(region)
```
superlopuh, line 158: This only does regions of the root operation, and not recursively, right?

superlopuh: It would be good to add a test with a module op, that has a test op with a region with dead code

xdsl/ir/post_order.py
alexarice: Looking back at this I think I implemented it entirely incorrectly. It should iterate through the children of each node before the node itself. I'll fix it and add docstrings

xdsl/transforms/dead_code_elimination.py
alexarice: This should act recursively, I'll add a test

xdsl/transforms/dead_code_elimination.py
```diff
@@ -73,8 +82,80 @@ def dce(op: ModuleOp):
...
+
+
+def region_dce(region: Region) -> bool:
+    live_set = LiveSet()
+
+    while live_set.changed:
+        live_set.changed = False
+        live_set.propagate_region_liveness(region)
+
+    live_set.changed = False
```
superlopuh, line 151:
```suggestion
```
Am I missing something?

superlopuh (approved):

xdsl/transforms/dead_code_elimination.py
alexarice: no

PR #3254, opened 2024-10-07
dialects: (stencil) add StencilType constraints

superlopuh:
This adds some constraint constants and a new `.constr` helper to construct parametrized constraints for stencil types. Pyright errors 143 -> 133

### DIFF
--- a/xdsl/transforms/csl_stencil_to_csl_wrapper.py
+++ b/xdsl/transforms/csl_stencil_to_csl_wrapper.py
@@ -84,7 +84,7 @@ class ConvertStencilFuncToModuleWrappedPattern(RewritePattern):
                     apply_op.done_exchange.block.args[1].type.get_shape()[-1],
                 )
 
-            # retrieve z_dim from post_process arg[0]
+            # retrieve z_dim from done_exchange arg[0]
             if isattr(
                 field_t := apply_op.done_exchange.block.args[0].type,
                 stencil.StencilTypeConstr,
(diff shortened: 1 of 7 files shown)


xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -84,15 +84,16 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
                     apply_op.done_exchange.block.args[1].type.get_shape()[-1],
                 )
 
-            # retrieve z_dim from done_exchange arg[0]
-            if isa(
+            # retrieve z_dim from post_process arg[0]
```
superlopuh, line 87:
```suggestion
            # retrieve z_dim from done_exchange arg[0]
```

xdsl/transforms/csl_stencil_to_csl_wrapper.py
```diff
@@ -85,14 +85,15 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
...
-            if isa(
+            if isattr(
                 field_t := apply_op.done_exchange.block.args[0].type,
-                stencil.StencilType[
-                    TensorType[Attribute] | memref.MemRefType[Attribute]
-                ],
+                stencil.StencilTypeConstr,
+            ) and isattr(
+                el_type := field_t.element_type,
+                AnyTensorTypeConstr | AnyMemRefTypeConstr,
```
superlopuh, line 93: this is the only kind of functional change, splitting the check into two, it should still be equivalent

math-fehr (approved):
Nice!
I think we should try to follow this on the other dialects later on

xdsl/transforms/csl_stencil_to_csl_wrapper.py
n-io: LGTM, thanks for doing this

n-io (approved):

PR #3255, opened 2024-10-07
api: (irdl) ParamAttrConstraint base_attr -> base_constr

superlopuh:
In Python, it's not legal to call `isinstance` with a first parameter that is a specialized generic (like `MemRefType[Attribute]`), which is problematic for the ParamAttrConstraint that constrains a generic attribute. Using a BaseAttr works around this constraint.

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -351,21 +351,24 @@ class ParamAttrConstraint(
     and also constrain its parameters with additional constraints.
     """
 
-    base_attr: type[ParametrizedAttributeCovT]
-    """The base attribute type."""
+    base_constr: BaseAttr[ParametrizedAttributeCovT]
+    """The base attribute type constraint."""
 
     param_constrs: tuple[GenericAttrConstraint[Attribute], ...]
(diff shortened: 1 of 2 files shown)


xdsl/irdl/constraints.py
```diff
@@ -351,31 +351,34 @@ class ParamAttrConstraint(
...
+        object.__setattr__(self, "base_constr", base_attr)
         object.__setattr__(self, "param_constrs", constrs)
 
     def verify(
         self,
         attr: Attribute,
         constraint_context: ConstraintContext,
     ) -> None:
-        if not isinstance(attr, self.base_attr):
+        if not isinstance(attr, self.base_constr.attr):
```
superlopuh, line 379: still isinstance instead of `.verify` because that function by itself does not communicate to the type system that it raises an error if the types aren't right.

alexarice:
I've used these constraints in `tblgen-to-py` so will likely need to modify something there

superlopuh:
After some discussions with Mathieu, closing this PR in favour of better helpers

PR #3258, opened 2024-10-08
bug: Tighten type constraints on DenseArrayBase.create_dense_int_or_index

### LINKED ISSUE #3154: Function _parse_builtin_densearray_attr fails when parsing a dense array with element_type IndexType

vinayakdsci:
Fixes #3154.

MLIR supports the following data types for `DenseArrayAttr`: `bool`, `i8`, `i16`, `i32`, `i64`, `f32`, and `f64`.

However, we also supported `index` in DenseArrayAttrs through the function `DenseArrayBase.create_dense_int_or_index()`, which caused an incompatibility with MLIR, causing `mlir-opt` to crash on IR produced by xdsl in this context.

This patch renames the function `DenseArrayBase.create_dense_int_or_index()` to `DenseArrayBase.create_dense_int()` so that the function's name is not misleading, and tightens type constraints on the function's inputs to only be of IntegerType.

This removes the incompatibility with MLIR produced from earlier behavior.


vinayakdsci:
cc @superlopuh @watermelonwolverine.

superlopuh:
Could you please add back the `create_int_or_index` function with a deprecated warning, advising users to use the new one instead? We can delete it later, but clients that depend on xDSL have one less breaking change

xdsl/dialects/builtin.py
```diff
@@ -941,6 +941,18 @@ def verify(self):
...
+    @staticmethod
+    def create_dense_int_or_index(
+        data_type: IntegerType | IndexType, data: Sequence[int] | Sequence[IntAttr]
+    ) -> DenseArrayBase:
+        if len(data) and isinstance(data[0], int):
+            attr_list = [IntAttr(d) for d in cast(Sequence[int], data)]
+        else:
+            attr_list = cast(Sequence[IntAttr], data)
+
+        return DenseArrayBase([data_type, ArrayAttr(attr_list)])
```
superlopuh, line 954:
```suggestion
        assert not isinstance(data_type, IndexType), "Index type is not supported"
        return DenseArrayBase.create_int(data_type, data)
```

vinayakdsci:
@math-fehr @superlopuh does this PR look good to go?

math-fehr (approved):
Yes, sorry I missed the previous pings!

vinayakdsci:
@math-fehr tests have passed, can we merge the PR?

Thanks a lot!

math-fehr:
> @math-fehr tests have passed, can we merge the PR?
> 
> 
> 
> Thanks a lot!

Sorry I got distracted when updating it, merging it now, thanks a lot!

PR #2704, opened 2024-06-10
core: Implement RangeVarConstraint

PapyChacal:
This is similar to VarConstraint, but constraining *ranges* to match instead.

PR #3261, opened 2024-10-08
core: Implement RangeVarConstraint

alexarice:
Continuation of #2704 

Co-authored-by: Emilien Bauer <papychacal@gmail.com>

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -20,7 +20,7 @@ class ConstraintContext:
     variables: dict[str, Attribute] = field(default_factory=dict)
     """The assignment of constraint variables."""
 
-    ranges: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
+    range_variables: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
     """The assignment of constraint range variables."""
 
     def copy(self):
@@ -517,15 +517,15 @@ class RangeVarConstraint(GenericRangeConstraint[AttributeCovT]):
         constraint_context: ConstraintContext | None = None,
     ) -> None:
         constraint_context = constraint_context or ConstraintContext()
-        if self.name in constraint_context.ranges:
-            if tuple(attrs) != constraint_context.ranges[self.name]:
+        if self.name in constraint_context.range_variables:
+            if tuple(attrs) != constraint_context.range_variables[self.name]:
                 raise VerifyException(
-                    f"attributes {tuple(str(x) for x in constraint_context.ranges[self.name])} expected from range variable "
+                    f"attributes {tuple(str(x) for x in constraint_context.range_variables[self.name])} expected from range variable "
                     f"'{self.name}', but got {tuple(str(x) for x in attrs)}"
                 )
         else:
             self.constraint.verify(attrs, constraint_context)
-            constraint_context.ranges[self.name] = tuple(attrs)
+            constraint_context.range_variables[self.name] = tuple(attrs)
 
     def get_resolved_variables(self) -> set[str]:
         return {self.name, *self.constraint.get_resolved_variables()}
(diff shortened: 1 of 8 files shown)


alexarice:
Would be nice to get the commit attribution correct as this is still mainly @PapyChacal 's work, I just made pyright happy and put some match statements on the tests

superlopuh:
I edited the description, which should propagate authorship when the PR is merged

xdsl/irdl/constraints.py
```diff
@@ -20,6 +20,9 @@ class ConstraintContext:
     variables: dict[str, Attribute] = field(default_factory=dict)
     """The assignment of constraint variables."""
 
+    ranges: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
+    """The constraint range variables."""
```
superlopuh, line 24: is this also an assignment?

superlopuh: Would be good to have the docstrings be consistent if yes

superlopuh (approved):
Would be great to have @math-fehr and @PapyChacal's ticks before merging

xdsl/irdl/constraints.py
```diff
@@ -20,6 +20,9 @@ class ConstraintContext:
     variables: dict[str, Attribute] = field(default_factory=dict)
     """The assignment of constraint variables."""
 
+    ranges: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
+    """The assignment of constraint range variables."""
+
     def copy(self):
```
alexarice, line 26: I've realised that the copy and update method now look incorrect. Am I right in thinking they should also copy/update the ranges? These functions are only used in the `AnyOf` constraint (maybe this constraint can't be used with ranges and so this is fine?

xdsl/irdl/constraints.py
alexarice: Have changed the docstring

xdsl/irdl/constraints.py
```diff
@@ -20,6 +20,9 @@ class ConstraintContext:
     variables: dict[str, Attribute] = field(default_factory=dict)
     """The assignment of constraint variables."""
 
+    ranges: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
+    """The assignment of constraint range variables."""
```
math-fehr, line 24: Should we somehow check somewhere that the variable and range variables have different names?
Otherwise I think the `can_infer` and `infer` funcitons might be incorrect if someone have twice the same name for range variable and variable.

If it's too annoying to check, jsut make an issue for it, as I don't think it's actually gonna happen in practice.

xdsl/irdl/constraints.py
```diff
@@ -20,6 +20,9 @@ class ConstraintContext:
     variables: dict[str, Attribute] = field(default_factory=dict)
     """The assignment of constraint variables."""
 
+    ranges: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
```
math-fehr, line 23:
```suggestion
    range_variables: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
```
?

math-fehr (approved):
Nice!

xdsl/irdl/constraints.py
alexarice: I've put this in as the last commit, can revert it if it is deemed too messy

alexarice:
@superlopuh @math-fehr any thoughts on whether to keep the last commit?

superlopuh:
IDK, I'd rather add helpers to add and get variables on the original objects, and keep all the logic to one class, from an architecture point of view there's some dubious coupling going on there, it would nice for the unicity logic to be encapsulated.

alexarice:
Sure, I'll do that

alexarice:
I removed the last commit, implemented it as @superlopuh suggested. In the process of this I rebased past the pyright changes and now had to fix some other stuff. I think this is because I generalised `AnyOf` at some point in this commit mistakenly. I can try to revert these changes if wanted.

math-fehr (approved):
To me that looks good!

PR #3263, opened 2024-10-08
misc: add pyright ignores for places where it's not going to be happy

superlopuh:
Pyright used to be happy to accept type expressions as types, but is happy no longer. This adds the required pyright ignores to work around this restriction.

Pylance errors: 94 -> 81

### DIFF
--- a/tests/test_is_satisfying_hint.py
+++ b/tests/test_is_satisfying_hint.py
@@ -372,18 +372,18 @@ def test_parametrized_attribute():
 
 
 def test_literal():
-    assert isa("string", Literal["string"])
-    assert isa("string", Literal["string", "another string"])
-    assert isa("another string", Literal["string", "another string"])
-    assert not isa("not this string", Literal["string", "another string"])
+    assert isa("string", Literal["string"])  # pyright: ignore[reportArgumentType]
+    assert isa("string", Literal["string", "another string"])  # pyright: ignore[reportArgumentType]
+    assert isa("another string", Literal["string", "another string"])  # pyright: ignore[reportArgumentType]
+    assert not isa("not this string", Literal["string", "another string"])  # pyright: ignore[reportArgumentType]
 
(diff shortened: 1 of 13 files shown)


tests/test_is_satisfying_hint.py
```diff
@@ -372,18 +372,18 @@ def test_parametrized_attribute():
...
-    assert not isa("not this string", Literal["string", "another string"])
-
-    assert isa(1, Literal[1])
-    assert isa(1, Literal[1, 2])
-    assert isa(2, Literal[1, 2])
-    assert not isa(3, Literal[1, 2])
-
-    assert not isa(1, Literal["1"])
-    assert not isa("1", Literal[1])
+    assert isa("string", Literal["string"])  # pyright: ignore[reportArgumentType]
```
math-fehr, line 375: Is it the error that `Literal` is a `TypeExpr` and not a `type`?

xdsl/tools/tblgen_to_py.py
```diff
@@ -474,7 +474,7 @@ def generate_op(self, tblgen_op: TblgenOp, dialect_name: str):
                     fields[name] = "var_region_def()"
                 case (True, True):
                     fields[name] = 'var_region_def("single_block")'
-                case _:
+                case _:  # pyright: ignore[reportUnnecessaryComparison]
```
math-fehr, line 477: Wait, should we just remove the default case then? We had this just because pyright wasn't happy, but now it is

math-fehr (approved):
Nice! I just had one comment in one place

xdsl/tools/tblgen_to_py.py
superlopuh: Yep I'll remove the default case with the new pyright, the old one still wants it. I guess I could remove it now and ignore the pyright error on the match case directly, I'll do that quickly.

PR #3264, opened 2024-10-08
misc: use VarConstraint instead of Annotated[ConstraintVar] in operation definitions

superlopuh:
Using Annotated in the way that we do is not well defined by the Python type system, this fixes the current misuse to get the number of Pylance errors to 0.

The current proposal is to keep both implementations working, and to deprecate the `Annotated` API at some point in the future.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import textwrap
 from collections.abc import Callable
 from io import StringIO
-from typing import Annotated, Generic, TypeVar
+from typing import Annotated, ClassVar, Generic, TypeVar
 
 import pytest
 
@@ -30,6 +30,7 @@ from xdsl.irdl import (
     ParamAttrConstraint,
     ParameterDef,
     ParsePropInAttrDict,
+    VarConstraint,
     VarOperand,
     VarOpResult,
     attr_def,
@@ -1578,7 +1579,7 @@ def test_basic_inference(format: str):
 
     @irdl_op_definition
     class TwoOperandsOneResultWithVarOp(IRDLOperation):
-        T = Annotated[Attribute, ConstraintVar("T")]
+        T: ClassVar[VarConstraint[Attribute]] = VarConstraint("T", AnyAttr())
 
         name = "test.two_operands_one_result_with_var"
         res = result_def(T)
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -22,9 +22,12 @@ from xdsl.irdl import (
     AttrSizedOperandSegments,
     AttrSizedRegionSegments,
     AttrSizedResultSegments,
+    BaseAttr,
     ConstraintVar,
     EqAttrConstraint,
+    GenericAttrConstraint,
     IRDLOperation,
+    ParamAttrConstraint,
     ParameterDef,
     ParsePropInAttrDict,
     VarOperand,
@@ -1559,7 +1562,7 @@ def test_optional_successor(program: str, generic_program: str):
 # Inference                                                                   #
 ################################################################################
 
-_T = TypeVar("_T", bound=Attribute)
+_T = TypeVar("_T", bound=Attribute, covariant=True)
 
 
 @pytest.mark.parametrize(
@@ -1661,6 +1664,18 @@ def test_nested_inference():
         p: ParameterDef[_T]
         q: ParameterDef[Attribute]
 
+        @classmethod
+        def constr(
+            cls,
+            *,
+            n: GenericAttrConstraint[Attribute] | None = None,
+            p: GenericAttrConstraint[_T] | None = None,
+            q: GenericAttrConstraint[Attribute] | None = None,
+        ) -> BaseAttr[ParamOne[Attribute]] | ParamAttrConstraint[ParamOne[_T]]:
+            if n is None and p is None and q is None:
+                return BaseAttr(cls)
+            return ParamAttrConstraint(cls, (n, p, q))
+
     @irdl_op_definition
     class TwoOperandsNestedVarOp(IRDLOperation):
         T = Annotated[Attribute, ConstraintVar("T")]
@@ -1694,6 +1709,16 @@ def test_non_verifying_inference():
     class ParamOne(ParametrizedAttribute, TypeAttribute, Generic[_T]):
         name = "test.param_one"
         p: ParameterDef[_T]
+
+        @classmethod
+        def constr(
+            cls,
+            *,
+            p: GenericAttrConstraint[_T] | None = None,
+        ) -> BaseAttr[ParamOne[Attribute]] | ParamAttrConstraint[ParamOne[_T]]:
+            if p is None:
+                return BaseAttr(cls)
+            return ParamAttrConstraint(cls, (p,))
 
     @irdl_op_definition
     class OneOperandOneResultNestedOp(IRDLOperation):
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import textwrap
 from collections.abc import Callable
 from io import StringIO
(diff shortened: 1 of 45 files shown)


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1661,13 +1664,25 @@ class ParamOne(ParametrizedAttribute, TypeAttribute, Generic[_T]):
         p: ParameterDef[_T]
         q: ParameterDef[Attribute]
 
+        @classmethod
```
alexarice, line 1667: This seems overkill for this test?

tests/irdl/test_operation_definition.py
```diff
@@ -227,6 +230,78 @@ def test_constraint_var_fail_not_satisfy_constraint():
         op.verify()
 
 
+@irdl_op_definition
+class GenericConstraintVarOp(IRDLOperation):
+    name = "test.constraint_var_op"
+
+    T: ClassVar[VarConstraint[IntegerType | IndexType]] = VarConstraint(
+        "T", base(IntegerType) | base(IndexType)
+    )
```
alexarice, line 239: In general do we want these constraints to be a visible part of the operation interface? Would it be better to use `_T` instead?

tests/irdl/test_operation_definition.py
```diff
@@ -227,6 +230,78 @@ def test_constraint_var_fail_not_satisfy_constraint():
...
+def test_generic_constraint_var_fail_non_equal():
+    """Check that all uses of a constraint variable are of the same attribute."""
+    i32_operand = TestSSAValue(i32)
+    index_operand = TestSSAValue(IndexType())
+
+    # Fail because of operand
+    op = GenericConstraintVarOp.create(
+        operands=[index_operand], result_types=[i32], attributes={"attribute": i32}
+    )
+    with pytest.raises(DiagnosticException):
```
alexarice, line 271: It would be nice for these to match on the error

tests/tblgen_to_py/test.py
```diff
@@ -50,13 +50,7 @@ class Test_AttributesOp(IRDLOperation):
...
 
     int_attr = prop_def(
-        ParamAttrConstraint(
-            IntegerAttr,
-            (
-                AnyAttr(),
-                EqAttrConstraint(IntegerType(16)),
-            ),
-        )
+        IntegerAttr[IntegerType].constr(type=EqAttrConstraint(IntegerType(16)))
```
alexarice, line 53: This file was automatically generated and probably shouldn't be manually changed

xdsl/dialects/builtin.py
```diff
@@ -644,6 +665,7 @@ def __init__(
 
 
 AnyFloatAttr: TypeAlias = FloatAttr[AnyFloat]
+AnyFloatAttrConstr: BaseAttr[AnyFloatAttr] = BaseAttr(FloatAttr)
```
alexarice, line 668: Do we need to put the type here?

xdsl/dialects/ltl.py
```diff
@@ -51,11 +51,9 @@ class AndOp(IRDLOperation):
 
     name = "ltl.and"
 
-    T = Annotated[
-        Attribute,
-        AnyOf([Sequence, Property, IntegerType(1, signedness=Signedness.SIGNLESS)]),
-        ConstraintVar("T"),
-    ]
+    T: ClassVar[VarConstraint[Attribute]] = VarConstraint(
+        "T", AnyOf([Sequence, Property, IntegerType(1, signedness=Signedness.SIGNLESS)])
```
alexarice, line 55: `Signedness.SINGLESS` is the default I believe

xdsl/irdl/constraints.py
```diff
@@ -221,6 +221,10 @@ def verify(
         pass
 
 
+# TODO: merge after pyright is updated, and update all constrs
+ANY_ATTR = AnyAttr()
```
alexarice, line 225: Why is this necessary?

xdsl/tools/tblgen_to_py.py
```diff
@@ -319,16 +319,12 @@ def _resolve_prop_constraint(self, rec: TblgenRecord | str) -> str:
                 return "BaseAttr(BoolAttr)"
             case "IndexAttr":
                 return textwrap.dedent("""
-                ParamAttrConstraint(
-                    IntegerAttr, (AnyAttr(), EqAttrConstraint(IndexType()))
-                )
+                IntegerAttr[IndexType].constr(type=IndexTypeConstr)
```
alexarice, line 322: Nice, ignore the earlier comment

alexarice (commented):
Very nice. Not a comprehensive review but I left some comments. Is the `constr` field a part of the `ParametrizedAttr` superclass?

xdsl/irdl/constraints.py
superlopuh: It's not really, it's more of a perf thing, just feels wrong to me to create lots of the same object, but I'm not planning on including it in any of the upcoming PRs, at least not before updating Pyright.

tests/irdl/test_operation_definition.py
superlopuh: What `_T`?

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1678,11 +1678,11 @@ def constr(
 
     @irdl_op_definition
     class TwoOperandsNestedVarOp(IRDLOperation):
-        T = Annotated[Attribute, ConstraintVar("T")]
+        T: ClassVar[VarConstraint[Attribute]] = VarConstraint("T", AnyAttr())
 
         name = "test.two_operands_one_result_with_var"
         res = result_def(T)
-        lhs = operand_def(ParamOne[T])
+        lhs = operand_def(ParamOne[Attribute].constr(p=T))
```
math-fehr, line 1685: I do have to say, I really dislike the verbosity of this, it will make it harder for newcomers to get started I think.

math-fehr (approved):
🚢

tests/irdl/test_declarative_assembly_format.py
superlopuh: Yeah it definitely got less pretty. One potential alternative is `ParamOneConstr(p=T)`, would you like me to try that instead? It's one more thing to import, but potentially worth the effort.

superlopuh: I just tested it locally and that API works and makes Pyright happy, it really doesn't want you to omit the type parameter even if it could be inferred in this case. I'm still inclined towards this more verbose API for the time being, with the hope that Pyright gets smart enough at some point to drop the unnecessary type parameter. We could also file it as a bug.

superlopuh: https://github.com/microsoft/pyright/issues/9190

alexarice: What about making `constr` a staticmethod instead? I just tried it locally and my pyright at least seems happy?

alexarice: This allows you to do `ParamOne.constr(p=T)` instead of `ParamOne[Attribute].constr(p=T)`

tests/irdl/test_operation_definition.py
superlopuh: Oh yeah I guess they could be, I tried to modify a minimal amount with these changes.

tests/irdl/test_declarative_assembly_format.py
superlopuh: it doesn't work for me

superlopuh: How do you feel about iterating on this after this is merged and Pyright is updated?

alexarice: Fine with me

alexarice: https://peps.python.org/pep-0696/ also potentially relevant for making this api nicer

superlopuh: Yep, I look forward to being able to use the new, shiny APIs. I'm not sure when it would be wise to drop support for older Pythons, according to this calendar 3.10 is supported for another couple of years: https://endoflife.date/python

PR #3271, opened 2024-10-09
transformations: (lower-csl-stencil) Optimise full-stencil access

n-io:
For full-stencil accesses followed by an *immediate* reduction of a uniform reduction function, use the catch-all `getRecvBufDsd` API call instead of the neighbour-by-neighbour `getRecvBufDsdByNeighbor` function.

Iff this optimisation is applicable, the accumulator dsd might look like this:
```zig
  const accumulatorDsd = @get_dsd(mem4d_dsd, .{
    .tensor_access = |i,j,k|{directionCount, pattern, chunkSize} -> accumulator[k]
  });
```

This is currently implemented by setting the strides of `i` and `j` to 0, which might require a printer update (tbc).

### DIFF
--- a/xdsl/transforms/lower_csl_stencil.py
+++ b/xdsl/transforms/lower_csl_stencil.py
@@ -260,7 +260,23 @@ class FullStencilAccessImmediateReductionOptimization(RewritePattern):
     """
     If an apply op accesses all points in the stencil shape *and* immediately performs a reduction,
     lower to an API call that iterates over all receive buffers at once. This requires setting up a
-    4d dsd that disregards all but one dimensions.
+    4d dsd that disregards all but one dimension.
+
+    The optimisation checks if it can be applied, and if so, sets up a new mem4d_dsd accumulator, lowers all
+    relevant `csl_stencil.access` calls to a single mem4d_dsd API call, and replaces all relevant reduction ops
+    with a single reduction op over the two mem4d_dsds.
+
+    Note, if the optimisation is not applied, `csl_stencil.access` calls are left untouched to be handled by
+    the `LowerAccessOp` pass instead and translated to individual mem1d_dsd API calls.
+
+    The optimisation is applied on the `csl_stencil.apply.receive_chunk` region iff:
+     * each point in the stencil shaped is accessed
+     * each `csl_stencil.access` has exactly one use
+     * each access is immediately processed by the same (type of) reduction op
+     * each reduction op uses the same accumulator to store a result
+     * each reduction op uses no inputs except from the above access ops
+     * todo: the data of the accumulator is not itself an input of the reduction
+     * todo: no other ops modify the accumulator in-between reduction ops
     """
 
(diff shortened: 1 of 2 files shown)


xdsl/transforms/lower_csl_stencil.py
```diff
@@ -252,6 +255,123 @@ def match_and_rewrite(self, op: csl_stencil.ApplyOp, rewriter: PatternRewriter,
                 block_arg.replace_by(SSAValue.get(new_arg))
 
 
+@dataclass(frozen=True)
+class FullStencilAccessImmediateReductionOptimization(RewritePattern):
+    """
+    If an apply op accesses all points in the stencil shape *and* immediately performs a reduction,
+    lower to an API call that iterates over all receive buffers at once. This requires setting up a
+    4d dsd that disregards all but one dimensions.
```
dk949, line 263:
```suggestion
    4d dsd that disregards all but one dimension.
```

xdsl/transforms/lower_csl_stencil.py
```diff
@@ -252,6 +255,123 @@ def match_and_rewrite(self, op: csl_stencil.ApplyOp, rewriter: PatternRewriter,
...
+            csl.FaddsOp,
+            csl.FmulsOp,
+        ]:
+            return
+        red_ops = cast(set[csl.BuiltinDsdOp], reduction_ops)
+
+        # safety check 1: each access has one use
+        acc_ops_uses = set(len(a.result.uses) for a in access_ops)
+        if len(acc_ops_uses) > 1 or acc_ops_uses.pop() != 1:
+            return
```
dk949, line 300:
```suggestion
        if any(len(a.result.uses) > 1 for a in access_ops):
            return
```

I think this is equivalent(?). I find it more readable (and it avoids constructing a `set`)

dk949 (changes requested):
How does this pass interact with the "Option 3" method that used to be emitted? 

If the optimization is performed, does Option 3 still apply?

xdsl/transforms/lower_csl_stencil.py
n-io: Done as suggested.

xdsl/transforms/lower_csl_stencil.py
n-io: Done as suggested.

n-io:
> How does this pass interact with the "Option 3" method that used to be emitted?
> 
> If the optimization is performed, does Option 3 still apply?

For `csl_stencil.apply` ops on which the optimisation is applied, this is clarified in the docstring.
For `csl_stencil.apply` ops on which the optimisation is not applied, there is now a filecheck showing the lowering to individual per-neighbour `@get_dsd` ops, which is the known and trusted default behaviour. Commenting out the optimisation in the `PatternRewriteWalker` will also trigger the default behaviour, in case this is helpful for testing or running.

dk949 (approved):

PR #3273, opened 2024-10-09
misc: add more .constr helpers and constraints

superlopuh:
Pylance errors: 84 -> 74.

Part of #3264

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -22,9 +22,12 @@ from xdsl.irdl import (
     AttrSizedOperandSegments,
     AttrSizedRegionSegments,
     AttrSizedResultSegments,
+    BaseAttr,
     ConstraintVar,
     EqAttrConstraint,
+    GenericAttrConstraint,
     IRDLOperation,
+    ParamAttrConstraint,
     ParameterDef,
     ParsePropInAttrDict,
     VarOperand,
@@ -1559,7 +1562,7 @@ def test_optional_successor(program: str, generic_program: str):
 # Inference                                                                   #
 ################################################################################
 
-_T = TypeVar("_T", bound=Attribute)
+_T = TypeVar("_T", bound=Attribute, covariant=True)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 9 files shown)


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1559,7 +1562,7 @@ class OptSuccessorOp(IRDLOperation):
 # Inference                                                                   #
 ################################################################################
 
-_T = TypeVar("_T", bound=Attribute)
+_T = TypeVar("_T", bound=Attribute, covariant=True)
```
superlopuh, line 1565: Necessary for constr definition below, that way `BaseAttr[ParamOne[_T]]` is a subclass of `BaseAttr[ParamOne[Attribute]]`.

xdsl/dialects/builtin.py
```diff
@@ -1436,7 +1457,7 @@ def print(self, printer: Printer) -> None:
 f128 = Float128Type()
 
 
-_MemRefTypeElement = TypeVar("_MemRefTypeElement", bound=Attribute)
+_MemRefTypeElement = TypeVar("_MemRefTypeElement", bound=Attribute, covariant=True)
```
superlopuh, line 1460: this change was made to appease Pyright below, but I'm not really sure that it's kosher... My understanding is that, even though a memref is mutable, and hence a memref type parameter of the memref itself should be covariant, a memref constraint is immutable, so a memref constraint that only accepts cats should be ok to give whenever the client expects a constraint that only accepts animals. Does this seem right? Am I missing something?

math-fehr (approved):
🚢

xdsl/dialects/builtin.py
alexarice: If you can store stuff in the memref then this is technically unsound I believe

alexarice: Actually, I can't even really figure out why this type is generic at all?

superlopuh: This is the type not the memref itself. I am now pretty convinced that it's sound, just because all attributes and attribute constraints are immutable, so it makes sense to have their type parameters be covariant. It is not really related to the type of the objects in the IR, as far as the Python type system is concerned.

alexarice: Looking through the uses of `MemRefType`, I'm pretty sure this generic argument serves no function whatsoever

PR #3272, opened 2024-10-09
dialects (arith) add fastmath option for float comparisons

knickish:
Add an optional fastmath attribute to `arith.cmpf` and include the attribute while lowering to riscv

Hopefully closes #2725.

PR #3275, opened 2024-10-10
dialects (arith): add fastmath optional attribute to cmpf

knickish:
`arith`-only part of #3272, with a new dialect test added

### DIFF
--- a/tests/filecheck/dialects/arith/arith_ops.mlir
+++ b/tests/filecheck/dialects/arith/arith_ops.mlir
@@ -173,7 +173,7 @@
 
   %cmpf_fm = "arith.cmpf"(%lhsf32, %rhsf32) {"predicate" = 2 : i64, "fastmath" = #arith.fastmath<fast>} : (f32, f32) -> i1
 
-  // CHECK-NEXT: %cmpf_fm = arith.cmpf ogt, %lhsf32, %rhsf32 fastmath<fast> : f32
+  // CHECK-NEXT: %cmpf_fm = arith.cmpf ogt, %lhsf32, %rhsf32 {"fastmath" = #arith.fastmath<fast>} : f32
 
   %selecti = "arith.select"(%lhsi1, %lhsi32, %rhsi32) : (i1, i32, i32) -> i32
   %selectf = "arith.select"(%lhsi1, %lhsf32, %rhsf32) : (i1, f32, f32) -> f32
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -829,8 +829,8 @@ class Cmpf(ComparisonOperation):
         printer.print(", ")
         printer.print_operand(self.rhs)
         if self.fastmath != FastMathFlagsAttr("none"):
-            printer.print_string(" fastmath")
-            self.fastmath.print_parameter(printer)
+            printer.print(" ")
+            printer.print_attr_dict({"fastmath": self.fastmath})
         printer.print(" : ")
         printer.print_attribute(self.lhs.type)
 
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_cmp.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_cmp.mlir
@@ -15,7 +15,7 @@
   %11 = "arith.select"(%10, %4, %5) : (i1, index, index) -> index
 }) : ()->()
 
-// CHECK:        "arith.cmpf"(%0, %1) <{"fastmath" = #arith.fastmath<none>, "predicate" = 2 : i64}> : (f64, f64) -> i1
+// CHECK:        "arith.cmpf"(%0, %1) <{"predicate" = 2 : i64, "fastmath" = #arith.fastmath<none>}> : (f64, f64) -> i1
 // CHECK:        "arith.select"(%6, %0, %1) : (i1, f64, f64) -> f64
 // CHECK:        "arith.cmpi"(%2, %3) <{"predicate" = 1 : i64}> : (i32, i32) -> i1
 // CHECK:        "arith.select"(%8, %2, %3) : (i1, i32, i32) -> i32


tests/filecheck/dialects/arith/arith_ops.mlir
```diff
@@ -167,6 +167,10 @@
 
   // CHECK-NEXT: %cmpf = arith.cmpf ogt, %lhsf32, %rhsf32 : f32
 
+  %cmpf_fm = "arith.cmpf"(%lhsf32, %rhsf32) {"predicate" = 2 : i64, "fastmath" = #arith.fastmath<fast>} : (f32, f32) -> i1
+
+  // CHECK-NEXT: %cmpf_fm = arith.cmpf ogt, %lhsf32, %rhsf32 fastmath<fast> : f32
```
alexarice, line 172: This doesn't agree with the output of mlir (or at least the version of mlir we are compatible with).
```
 %58 = arith.cmpf ogt, %4#0, %4#1 {fastmath = #arith.fastmath<fast>} : f32
```
I see that it does agree with how this is parsed in mlir master though. Not sure what the correct thing to do here is.

xdsl/dialects/arith.py
```diff
@@ -765,13 +768,16 @@ def parse(cls, parser: Parser):
         operand1 = parser.parse_unresolved_operand()
         parser.parse_punctuation(",")
         operand2 = parser.parse_unresolved_operand()
+        fastmath = None
+        if parser.parse_optional_keyword("fastmath") is not None:
+            fastmath = FastMathFlagsAttr(FastMathFlagsAttr.parse_parameter(parser))
```
alexarice, line 773:
```suggestion
        fastmath = None if parser.parse_optional_keyword("fastmath") is None else FastMathFlagsAttr(FastMathFlagsAttr.parse_parameter(parser))
```

alexarice (commented):
@superlopuh @math-fehr how do we want to manage mlir version mismatches (see comments)

xdsl/dialects/arith.py
superlopuh: Hmm, I'm not sure I prefer your suggestion TBH, I'd maybe suggest moving the `fastmath = None` into an `else` clause but otherwise I think it's OK to split into two lines.

tests/filecheck/dialects/arith/arith_ops.mlir
superlopuh: yeah `fastmath` should be in the properties dict in the generic format

alexarice: That too

alexarice:
I believe this property should not be optional, as `DefaultValuedAttr` in tablegen is not optional

knickish:
Ok, thanks for the feedback. Planning to:
- Make `FastMathFlagsAttr` non-optional
- Print as member of property/attribute dict rather than `fastmath<fast>`

Why is `FastMathFlagsAttr` expected to be printed in the property dict rather than in the `fastmath<fast>` format? I see that the `addf` printer does not output it in the properties dict, should that also be updated?

```mlir
  %faddf = arith.addf %lhsf32, %rhsf32 fastmath<fast> : f32
  %faddf_vector = arith.addf %lhsvec, %rhsvec fastmath<fast> : vector<4xf32>

  // CHECK-NEXT: %faddf = arith.addf %lhsf32, %rhsf32 fastmath<fast> : f32
  // CHECK-NEXT: %faddf_vector = arith.addf %lhsvec, %rhsvec fastmath<fast> : vector<4xf32>
  ```

**Comment by alexarice** [comment]
Likely no-one checked the correct mlir version when `arith.addf` was made? It might be worth holding off on changing the syntax as I'm pretty sure we want to update to a newer version of mlir soon(?).

**Comment by superlopuh** [comment]
I'm not sure when we'll get around to updating MLIR but we should be consistent with our current version, and if we're not that's a bug. I have a feeling that some MLIR interop tests are missing here to make sure that our syntax is consistent.

We have a file to check that the generic format is consistent (`tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_fp_ops.mlir`) but we would probably benefit from the custom format also being checked.

**Comment by knickish** [comment]
Assigned a default of `FastMathFlagsAttr("none")` when no value is passed at init (or found while parsing). Also updated the print method to print the flag as a property dict member. Is the printing done correctly? I wasn't able to find a method that felt very clean for that part

**Comment by knickish** [comment]
> I believe this property should not be optional, as `DefaultValuedAttr` in tablegen is not optional

Using `prop_def` instead of `opt_prop_def` made a lot of tests fail parsing. Is an `opt_prop_def` sufficient for this as long as it is always set?

File: xdsl/dialects/arith.py [hunk 3]
```diff
@@ -718,13 +718,15 @@ class Cmpf(ComparisonOperation):
     predicate = prop_def(AnyIntegerAttr)
     lhs = operand_def(floatingPointLike)
     rhs = operand_def(floatingPointLike)
+    fastmath = opt_prop_def(FastMathFlagsAttr)
```
alexarice, line 721: What goes wrong with `prop_def(FastMathFlagsAttr, default = FastMathFlagsAttr("none"))`

xdsl/dialects/arith.py
```diff
@@ -780,6 +786,9 @@ def print(self, printer: Printer):
         printer.print_operand(self.lhs)
         printer.print(", ")
         printer.print_operand(self.rhs)
+        if self.fastmath is not None and self.fastmath != FastMathFlagsAttr("none"):
```
alexarice, line 789: I'm unsure that this printing is correct if there are other attributes on the function (is the attribute dictionary being printed anywhere?

xdsl/dialects/arith.py
```diff
@@ -765,13 +768,16 @@ def parse(cls, parser: Parser):
         operand1 = parser.parse_unresolved_operand()
         parser.parse_punctuation(",")
         operand2 = parser.parse_unresolved_operand()
+        fastmath = FastMathFlagsAttr("none")
+        if parser.parse_optional_keyword("fastmath") is not None:
+            fastmath = FastMathFlagsAttr(FastMathFlagsAttr.parse_parameter(parser))
```
alexarice, line 773: This is parsing the old(/new mlir) version right?

xdsl/dialects/arith.py
alexarice: Apparently `default` isn't really implemented so maybe ignore this suggestion

xdsl/dialects/arith.py
knickish: I actually have no idea how to answer this. [This page](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithcmpf-arithcmpfop) seems to show it in the `'fastmath<fast>` format separate from the property dict, but it looks like the attribute dict is also used in some other places in `xdsl`. How can I tell which version should be supported, and see the documentation for that version?

xdsl/dialects/arith.py
knickish: I don't think so, as nothing `fastmath`-related is printed when it's set as an attribute without this change

xdsl/dialects/arith.py
alexarice: That page is the documentation for mlir master. I believe the version of mlir we have current compatibility with is 98e674c9f16d677d95c67bc130e267fae331e43c , though the link to this seems to have gone missing from the readme at some point

superlopuh (commented):
I just checked, and our version of MLIR doesn't have fastmath on comparison operations. This makes me think that we should hold off on merging this until we update MLIR, and that we should prioritise this update, as the last one was in last December.

superlopuh:
The properties are in review, and hopefully we'll update MLIR soon also, I have a feeling that it won't take very long. Thank you for your patience :)

alexarice:
Should have said earlier, but I think the changes from this PR have all been pulled into https://github.com/xdslproject/xdsl/pull/3310. Let me know if there's anything missing

knickish:
Still need to do the MLIR update part, but amended the commit to use a `prop_def` with Alex's new `default_value` change

knickish:
Oh you commented between drafting my comment and sending it. I think it should all be good for now, thanks @alexarice

knickish:
looks like this got done in the mlir bump, will go ahead and close

PR #3276, opened 2024-10-10
dialects: (riscv) add fastmath flag to RdRsRs Float Float Int operations

knickish:
`riscv`-only part of https://github.com/xdslproject/xdsl/pull/3272, with a new dialect test added. Thanks for the suggestion of separate tests, I didn't realize those existed and caught an error in the LiOp implementation while adding the test cases.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -43,7 +43,6 @@ from xdsl.irdl import (
     irdl_op_definition,
     operand_def,
     opt_attr_def,
-    opt_prop_def,
     region_def,
     result_def,
     var_operand_def,
@@ -3028,7 +3027,7 @@ class RdRsRsFloatFloatIntegerOperationWithFastMath(RISCVInstruction, ABC):
     rd = result_def(IntRegisterType)
     rs1 = operand_def(FloatRegisterType)
     rs2 = operand_def(FloatRegisterType)
-    fastmath = opt_prop_def(FastMathFlagsAttr)
+    fastmath = attr_def(FastMathFlagsAttr)
 
     def __init__(
         self,
@@ -3048,11 +3047,9 @@ class RdRsRsFloatFloatIntegerOperationWithFastMath(RISCVInstruction, ABC):
 
         super().__init__(
             operands=[rs1, rs2],
-            properties={
-                "fastmath": fastmath,
-            },
             attributes={
                 "comment": comment,
+                "fastmath": fastmath,
             },
             result_types=[rd],
         )
@@ -3066,8 +3063,7 @@ class RdRsRsFloatFloatIntegerOperationWithFastMath(RISCVInstruction, ABC):
         fast = FastMathFlagsAttr("none")
         if parser.parse_optional_keyword("fastmath") is not None:
             fast = FastMathFlagsAttr(FastMathFlagsAttr.parse_parameter(parser))
-            if fast != FastMathFlagsAttr("none"):
-                attributes["fastmath"] = fast
+        attributes["fastmath"] = fast
         return attributes
 
     def custom_print_attributes(self, printer: Printer) -> Set[str]:
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -43,6 +43,7 @@ from xdsl.irdl import (
     irdl_op_definition,
     operand_def,
     opt_attr_def,
+    opt_prop_def,
     region_def,
     result_def,
     var_operand_def,
@@ -2532,7 +2533,7 @@ class LiOp(RISCVInstruction, ABC):
     def custom_print_attributes(self, printer: Printer) -> Set[str]:
         printer.print(" ")
         print_immediate_value(printer, self.immediate)
-        return {"immediate"}
+        return {"immediate", "fastmath"}
 
     @classmethod
     def parse_op_type(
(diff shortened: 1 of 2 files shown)


xdsl/dialects/riscv.py
```diff
@@ -2491,6 +2491,7 @@ class LiOp(RISCVInstruction, ABC):
 
     rd = result_def(IntRegisterType)
     immediate = attr_def(base(Imm32Attr) | base(LabelAttr))
+    fastmath = opt_attr_def(FastMathFlagsAttr)
```
alexarice, line 2494: Is there a specific reason this is an attribute here where it was a property on the other PR?

xdsl/dialects/riscv.py
```diff
@@ -2527,12 +2530,19 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
     def custom_parse_attributes(cls, parser: Parser) -> dict[str, Attribute]:
         attributes = dict[str, Attribute]()
         attributes["immediate"] = parse_immediate_value(parser, i32)
+        fast = FastMathFlagsAttr("none")
```
alexarice, line 2533: Is this flag meant to be optional? It seems you can't parse it with the flag unset?

xdsl/dialects/riscv.py
```diff
@@ -3016,6 +3026,65 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
...
+    A base class for RISC-V operations that have two source floating-point
+    registers with an integer destination register, and can be annotated with fastmath flags.
+
+    This is called R-Type in the RISC-V specification.
+    """
+
+    rd = result_def(IntRegisterType)
+    rs1 = operand_def(FloatRegisterType)
+    rs2 = operand_def(FloatRegisterType)
+    fastmath = opt_attr_def(FastMathFlagsAttr)
```
alexarice, line 3040: Similar comments here about attr/prop and this being optional

xdsl/dialects/riscv.py
```diff
@@ -3016,6 +3026,65 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
         return self.rd, self.rs1, self.rs2
 
 
+class RdRsRsFloatFloatIntegerOperationWithFastMath(RISCVInstruction, ABC):
```
alexarice, line 3029: Can this not inherit from `RdRsRSFloatFloatIntegerOperation`? Looks like duplicated code

xdsl/dialects/riscv.py
superlopuh: I would recommend changing to a property here

xdsl/dialects/riscv.py
superlopuh: Nope, because the inits get weird, I have a feeling that we should stick with this for now

xdsl/dialects/riscv.py
superlopuh, line 2494: What is the semantics change of li being optionally fastmath?

knickish:
Ok thanks, will do this stuff this evening hopefully

Need to:
- [ ] Make `fastmath` a property rather than attribute on both op types
- [ ] Document semantic changes to `LiOp` now that it has the optional `fastmath` property
- [ ] Use `None` as default instead of `FastMathFlagsAttr("none")`

xdsl/dialects/riscv.py
knickish: I spent a while trying to find any information on this, and I'm not seeing that there is any actual effect on the `li` instr at all when compiled with fastmath. `Chapter 15. Load Immediate` of the [riscv-asm-manual](https://github.com/riscv-non-isa/riscv-asm-manual.git) doesn't mention fastmath at all, and [godbolt](https://godbolt.org/z/Yf4jP4z8d) doesn't seem to show any difference when compiled with/without `-ffast-math`. 

Is there anywhere else that I should look for information on this? If not, should I go ahead and remove the `fastmath` attribute from `LiOp`?

xdsl/dialects/riscv.py
knickish: As on the other PR, it's not meant to be optional, `op_prop_def` just doesn't actually use the `default` parameter at all, so parsing fails when it's set as a non-`opt` `prop_def`

xdsl/dialects/riscv.py
superlopuh: I also don't know how it would affect anything. I don't think arith.constant can have fastmath, can it? It seems safe to remove from this PR, and add it later if we find a way to leverage it.

xdsl/dialects/riscv.py
superlopuh: I would recommend making it non-optional, and to do the default value handling manually. What's the error when you change the declaration to `prop_def`?

superlopuh: I guess this is more a question for the equivalent line below, if we're removing the LiOp changes

knickish:
```
# |     raise VerifyException(f"property {prop_name} expected")
# | xdsl.utils.exceptions.VerifyException: property fastmath expected
```
That's the error I get, even with setting a default via `fastmath = prop_def(FastMathFlagsAttr, default=FastMathFlagsAttr("none")) (which also doesn't typecheck)

xdsl/dialects/riscv.py
knickish: I've removed it now, and rebased everything into one commit since it was getting messy

xdsl/dialects/riscv.py
superlopuh: yep, the default field here doesn't work. I think @alexarice is now looking into this, but for the time being we can just set it directly in the init and in the custom parser. So all that would be necessary in this PR is to change `opt_prop_def` to `prop_def`

knickish: I've just pushed a commit that does exactly that, but it's failing the filecheck test with the same error as above. Is there something else I need to do ? I am setting the `fastmath` property in both the parse and init methods

xdsl/dialects/riscv.py
```diff
@@ -3016,6 +3017,68 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
...
+            operands=[rs1, rs2],
+            properties={
+                "fastmath": fastmath,
+            },
+            attributes={
+                "comment": comment,
+            },
+            result_types=[rd],
+        )
+        self.fastmath = fastmath
```
superlopuh, line 3059: There's no need, the `super().__init__` sets this property for us
```suggestion
```

xdsl/dialects/riscv.py
```diff
@@ -3016,6 +3017,68 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
...
+
+    @classmethod
+    def custom_parse_attributes(cls, parser: Parser) -> dict[str, Attribute]:
+        attributes = dict[str, Attribute]()
+        fast = FastMathFlagsAttr("none")
+        if parser.parse_optional_keyword("fastmath") is not None:
+            fast = FastMathFlagsAttr(FastMathFlagsAttr.parse_parameter(parser))
+            if fast != FastMathFlagsAttr("none"):
+                attributes["fastmath"] = fast
+        cls.fastmath = fast
```
superlopuh, line 3072: This sets fastmath on the class directly, not on the instance. What happens without this line? I would expect that there is some magic to convert attributes to properties, but if not we'll need to add this to our infrastructure. I'll pull this branch locally and see what's up.

superlopuh: OK so the current parsing infrastructure only handles attributes. I would recommend changing fastmath to a non-optional `attr_def`, and not dealing with the whole property infrastructure side of things

superlopuh: Eventually we'll want to migrate the riscv dialect to use properties instead of attributes, but that's not critical

xdsl/dialects/riscv.py
knickish: removed this

xdsl/dialects/riscv.py
knickish: it is now using `attr_def`

superlopuh (approved):
Thank you!

PR #3283, opened 2024-10-10
dialects: (cf) cond_br folding

alexarice:
First two `cond_br` canonicalization patterns:
- Constant folding (`cf.cond_br %true ^0 ^1` == `cf.br ^0`)
- Passthrough (conditional branch to a branch with just a single `br` op gets forwarded)

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/cf.py
+++ b/xdsl/transforms/canonicalization_patterns/cf.py
@@ -9,6 +9,7 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 from xdsl.rewriter import InsertPoint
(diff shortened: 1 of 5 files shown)


xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -9,6 +9,7 @@
     op_type_rewrite_pattern,
 )
 from xdsl.rewriter import InsertPoint
+from xdsl.transforms.canonicalization_patterns.scf import const_evaluate_operand
```
alexarice, line 12: Should this be moved somewhere else, and if so where?

superlopuh (approved):

PR #3284, opened 2024-10-10
transformations: (cf) cf.cond_br identical successors

alexarice:
Adds simplification to `cf.cond_br` when both branches point to the same successor.

The mlir version only performs the reduction when either all the operands are the same, or the common destination branch only has one predecessor. I have no idea why this is and didn't include this condition.

Also updates some of the other tests which now reduce further

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/cf.py
+++ b/xdsl/transforms/canonicalization_patterns/cf.py
@@ -172,3 +172,40 @@ class SimplifyPassThroughCondBranch(RewritePattern):
                 op.cond, new_then, new_then_args, new_else, new_else_args
             )
         )
+
+
+class SimplifyCondBranchIdenticalSuccessors(RewritePattern):
+    """
+    cf.cond_br %cond, ^bb1(A, ..., N), ^bb1(A, ..., N)
+     -> br ^bb1(A, ..., N)
+
+    cf.cond_br %cond, ^bb1(A), ^bb1(B)
+     -> %select = arith.select %cond, A, B
+        br ^bb1(%select)
+    """
+
+    @staticmethod
+    def _merge_operand(
+        op1: SSAValue,
+        op2: SSAValue,
+        rewriter: PatternRewriter,
+        cond_br: cf.ConditionalBranch,
+    ) -> SSAValue:
+        if op1 == op2:
+            return op1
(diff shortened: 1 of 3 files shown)


alexarice:
@zero9178 maybe you know why the precondition exists in mlir?

zero9178:
> @zero9178 maybe you know why the precondition exists in mlir?

I'd have to do more code archealogy given that the code is rather old but if I had to guess as to why, I think its just opinionated about the canonical form and arguably the pattern you're using is worse if you measure the "better canonical form" as the one that has fewer operations.

Consider the following example:
```mlir
cf.cond_br %bool, ^final_dest(%args...), ^inner_block

^inner_block:
   cf.cond_br ^final_dest(%other_args1...), ^final_dest(%other_args2)

^final_dest(%bb_args...):
   ...
```
as you have it implemented right now, `^inner_block` will have a `arith.select` per block argument followed by a `cf.br` instead of the `cf.cond_br`. Meaning that you now traded `n` more operations for removing exactly one edgein the CFG  (one of the two that goes from `inner_block` to `^final_dest), while keeping the shape of the CFG identical otherwise. 

The condition that `^inner_block` must be the only predecessor of `^final_dest` ensures that the `cf.br` that gets created will be able to be folded and the CFG therefore simplified by merging blocks together.

alexarice:
Interesting, I think I'm relatively happy with it working either way so will see what other people think

xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -172,3 +172,40 @@ def match_and_rewrite(self, op: cf.ConditionalBranch, rewriter: PatternRewriter)
...
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: cf.ConditionalBranch, rewriter: PatternRewriter):
+        # Check that the true and false destinations are the same
+        if op.then_block != op.else_block:
+            return
+
+        merged_operands = tuple(
+            self._merge_operand(op1, op2, rewriter, op)
+            for (op1, op2) in zip(op.then_arguments, op.else_arguments)
```
superlopuh, line 208: for good measure:
```suggestion
            for (op1, op2) in zip(op.then_arguments, op.else_arguments, strict=True)
```

superlopuh (approved):
neat

PR #3286, opened 2024-10-10
misc: remove unnecessary pyright ignore comments [NFC]

superlopuh:
I added `reportUnnecessaryTypeIgnoreComment = true` to pyproject.toml to find these

### DIFF
--- a/tests/backend/test_jax_executable.py
+++ b/tests/backend/test_jax_executable.py
@@ -170,7 +170,7 @@ def test_return_annotation_tuple_type():
         ),
     ):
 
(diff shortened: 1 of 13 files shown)


math-fehr:
Oh that's nice! Should we keep this option in the `.toml`?

math-fehr (approved):

alexarice:
I assume keeping the option on makes it difficult to support multiple python versions?

superlopuh:
Yeah I removed it ultimately because I have a feeling that it will make the kind of migration we just did more difficult in the future.

PR #3287, opened 2024-10-10
api: constr helpers for generic attributes now mostly static methods

superlopuh:


tests/tblgen_to_py/test_tblgen.py
```diff
@@ -11,5 +11,4 @@ def test_run_tblgen_to_py():
     with open("tests/tblgen_to_py/test.py") as f:
         expected = f.read()
 
-        assert len(out_str.strip()) == len(expected.strip())
```
alexarice, line 14: Seems unrelated but guess it got annoying?

xdsl/dialects/builtin.py
```diff
@@ -1598,29 +1597,28 @@ def get_strides(self) -> Sequence[int | None] | None:
...
+        and memory_space is None
+    ):
+        return BaseAttr[MemRefType[_MemRefTypeElement]](MemRefType)
+    return ParamAttrConstraint[MemRefType[_MemRefTypeElement]](
+        MemRefType, (shape, element_type, layout, memory_space)
+    )
 
 
 AnyMemRefType: TypeAlias = MemRefType[Attribute]
-AnyMemRefTypeConstr = BaseAttr[MemRefType[Attribute]](MemRefType)
```
alexarice, line 1623: I would personally prefer to remove extra aliases. We already have quite a few ways to build these constraints.

xdsl/dialects/stencil.py
```diff
@@ -359,19 +359,20 @@ def __init__(
...
+
+def StencilTypeConstr(
+    *,
+    bounds: GenericAttrConstraint[Attribute] | None = None,
+    element_type: GenericAttrConstraint[_FieldTypeElement] = AnyAttr(),
+) -> (
+    BaseAttr[StencilType[Attribute]]
+    | ParamAttrConstraint[StencilType[_FieldTypeElement]]
+):
+    if bounds is None and element_type == AnyAttr():
```
alexarice, line 371: Not specific to here, but what's the benefit to using `BaseAttr` over a `ParamAttrConstraint`?

xdsl/dialects/stencil.py
```diff
@@ -406,9 +419,21 @@ class TempType(
...
+    element_type: GenericAttrConstraint[_FieldTypeElement] = AnyAttr(),
+) -> BaseAttr[TempType[Attribute]] | ParamAttrConstraint[TempType[_FieldTypeElement]]:
+    if bounds is None and element_type == AnyAttr():
+        return BaseAttr[TempType[Attribute]](TempType)
+    return ParamAttrConstraint[TempType[_FieldTypeElement]](
+        TempType, (bounds, element_type)
+    )
+
+
+AnyStencilTypeConstr = StencilTypeConstr()
```
alexarice, line 434: Same comment as above

alexarice (approved):
I think I'd personally prefer these as `@staticmethod`, as it couples them more to their class and there is possibly an option of providing a default implementation in non generic cases, but this is a good improvement on what is there currently.

tests/tblgen_to_py/test_tblgen.py
superlopuh: Oh yeah it was actually making the test harder to debug, pytest shows a nice diff for string comparisons, and not for `len`s. I meant to put it in a different PR, but if you're ok to merge it then we could just do it in this one.

xdsl/dialects/stencil.py
superlopuh: Perf

xdsl/dialects/builtin.py
superlopuh: I just feel unhappy about constructing these objects many times, but that may well be premature optimisation...

alexarice: This is surely an immeasurable difference

tests/tblgen_to_py/test_tblgen.py
superlopuh: I reverted it here, and will open a separate PR for this.

superlopuh:
@math-fehr @alexarice I tried a different approach, which is not to reuse the type parameter on the class definition, seems to work quite well. The only exception is the stencil case, where they actually use type hierarchies, so that bit of the codebase can be a bit more verbose. WDYT?

tests/tblgen_to_py/test.py
```diff
@@ -49,9 +49,7 @@ class Test_AnyOp(IRDLOperation):
 class Test_AttributesOp(IRDLOperation):
     name = "test.attributes"
 
-    int_attr = prop_def(
-        IntegerAttr[IntegerType].constr(type=EqAttrConstraint(IntegerType(16)))
-    )
+    int_attr = prop_def(IntegerAttr.constr(type=EqAttrConstraint(IntegerType(16))))
```
alexarice, line 52: If you wanted to simplify syntax further, you could allow this to take an attribute to be coerced into a Constraint

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1710,23 +1710,22 @@ class ParamOne(ParametrizedAttribute, TypeAttribute, Generic[_T]):
         name = "test.param_one"
         p: ParameterDef[_T]
 
-        @classmethod
+        @staticmethod
```
alexarice, line 1713: I think this looks good, much closer to what I imagined

alexarice: Is there any hope to have all `ParametrisedAttribute`s define this? perhaps with a method defined in `@irdl_attr_definition`? (Perhaps not as this PR)

alexarice (commented):
This looks nice to me, I can't see the stencil stuff you were talking about though

tests/tblgen_to_py/test.py
superlopuh: I'd rather not go down that route, but it can be done in a different PR

PR #3291, opened 2024-10-11
transformations: (cf) switch canonicalization

alexarice:
Adds all the switch canonicalization patterns present in mlir. I'm happy to anonymise some of the block names in the filecheck tests if this is deemed worth doing.

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/cf.py
+++ b/xdsl/transforms/canonicalization_patterns/cf.py
@@ -1,8 +1,14 @@
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
+from typing import cast
 
 from xdsl.dialects import arith, cf
-from xdsl.dialects.builtin import BoolAttr, IntegerAttr
-from xdsl.ir import Block, BlockArgument, SSAValue
+from xdsl.dialects.builtin import (
+    AnyIntegerAttr,
+    BoolAttr,
+    DenseIntOrFPElementsAttr,
+    IntegerAttr,
+)
+from xdsl.ir import Block, BlockArgument, Operation, SSAValue
 from xdsl.pattern_rewriter import (
     PatternRewriter,
     RewritePattern,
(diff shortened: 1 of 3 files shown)


xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -172,3 +177,276 @@ def match_and_rewrite(self, op: cf.ConditionalBranch, rewriter: PatternRewriter)
...
+    """
+    switch %flag : i32, [
+      default:  ^bb1
+    ]
+     -> br ^bb1
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: cf.Switch, rewriter: PatternRewriter):
+        if len(op.case_blocks) == 0:
```
superlopuh, line 192:
```suggestion
        if not op.case_blocks:
```

xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -172,3 +177,276 @@ def match_and_rewrite(self, op: cf.ConditionalBranch, rewriter: PatternRewriter)
...
+        return
+    requires_change = False
+
+    new_case_values: list[int] = []
+    new_case_blocks: list[Block] = []
+    new_case_operands: list[Sequence[Operation | SSAValue]] = []
+
+    for switch_case, block, operands in zip(
+        case_values.data.data, op.case_blocks, op.case_operand
+    ):
```
superlopuh, line 214:
```suggestion
    for switch_case, block, operands in zip(
        case_values.data.data, op.case_blocks, op.case_operand, strict=True
    ):
```

xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -172,3 +177,276 @@ def match_and_rewrite(self, op: cf.ConditionalBranch, rewriter: PatternRewriter)
...
+    Helper for folding a switch with a constant value.
+    switch %c_42 : i32, [
+      default: ^bb1 ,
+      42: ^bb2,
+      43: ^bb3
+    ]
+    -> br ^bb2
+    """
+    case_values = switch.case_values
+    if case_values is not None:
```
superlopuh, line 275: does it make sense for this to be optional? Shouldn't it be `()` by default?

xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -172,3 +177,276 @@ def match_and_rewrite(self, op: cf.ConditionalBranch, rewriter: PatternRewriter)
...
+        for i, c in enumerate(case_values.data.data):
+            if flag == c.value.data:
+                rewriter.replace_matched_op(
+                    cf.Branch(switch.case_blocks[i], *switch.case_operand[i])
+                )
+                return
+
+    rewriter.replace_matched_op(
+        cf.Branch(switch.default_block, *switch.default_operands)
+    )
```
superlopuh, line 285: I would recommend something like this:
```suggestion
    case_values = () if switch.case_values is None else switch.case_values.data.data

    new_block, new_operands = next(
        ((block, operand) 
        for (c, block, operand) in zip(case_values, switch.case_blocks, switch.case_operands, strict=True)
        if flag == c.value.data),
        (switch.default_block, switch.default_operands)

    rewriter.replace_matched_op(
        cf.Branch(new_block, *new_operands)
    )
```

xdsl/transforms/canonicalization_patterns/cf.py
alexarice: It is an optional property in mlir. I don't personally think it makes any sense but it is what it is

alexarice:
@superlopuh happy for this to be merged?

superlopuh (approved):
Thank you!

PR #3294, opened 2024-10-12
installation: Switch to UV for dependency management

EdmundGoodman:
## Summary

Adopt `uv` for dependency management, replacing the existing `pip` approach.

## Motivation

`uv` is developed by [Astral](https://astral.sh/), who also wrote `ruff` which is already in use in xDSL CI pipelines. It offers the benefits of much faster dependency resolution and installation, along with leveraging the more modern `pyproject.toml` configuration format defined in PEPs [518](https://peps.python.org/pep-0518/) and [618](https://peps.python.org/pep-0621/).

Requested by @superlopuh

## Changes

This replacement requires the following changes:

- [x] Modification to the `venv` target of the Makefile to create the virtual environment
- [x] Update CI flows to install with `uv` as opposed to `pip`, including testing
  - [x] `ci-core.yaml`
  - [x] `ci-mlir.yaml`
  - [x] `ci-notebooks.yaml`
  - [x] `ci-pyright-fails.yaml`
  - [x] `code-formatting.yaml`
  - [ ] `jupyterlite.yaml` (cannot easily be spawned due to workflow triggers)
  - [ ] `pythonpublish.yaml` (cannot easily be spawned due to required secrets)
  - [x] `release-notes.yaml` (requires no changes)
- [x] Update documentation to reflect these changes
- [x] Test across operating systems
  - [x] Windows
  - [x] MacOS
  - [x] NixOS
- [x] Resolve changes required to Dependabot

## Post-merge checklist

Once merged, we should check the following things:

- [ ] The release flow worked correctly (including versioneer)
- [ ] Users are able to switch to `uv` and instructions are clear
- [ ] The `update-bot` workflow creates helpful and correct PRs in a timely manner


superlopuh (commented):
Could you please also migrate the makefile? Maybe with a phony target that checks whether uv is installed, and directs developers to installation instructions. The goal would be for make tests to be runnable outside of an activated venv

math-fehr (commented):
Thanks a lot for the PR!
A few quick questions/requests I have:
* Can we move the removed requirements.txt in the pyproject.toml? That way we can still track them with dependabot, and also this is the only change that's missing to still support virtualenv. I'd personally would like to still continue supporting venv if possible, as for now it is still way more popular than uv.
* Do you think we can automate the update of uv.lock with dependabot? Currently, we get daily automated updates of dependencies with dependabot, and it would quite annoying to have to update the `uv.lock` manually every time.
* We will probably have to test pythonpublish after the PR gets merged. Just ping me and I'll try to run it to see if it works or not, but looking at it this should be fine I think.

alexarice (commented):
We've been using `uv` for our quantum stack recently and it seems to work nicely. I have a couple of comments:
- When to use `uv run`. It seems in the test ci you are activating the venv anyway so I don't see the reason to us `uv run` everywhere.
- If you did want to use `uv run` everywhere, you would need to make the filecheck tests use it (i.e. every `xdsl-opt` should be `uv run xdsl-opt`). You can get most of these by modifying `lit.cfg`. I'm not sure if this is actually worth it?
- I've always had problems with the `wgpu` package and had to manually remove it. Will there be an easy way to do that with the new set up?
- `uv` should be added to `flake.nix` (I can do this afterwards if necessary)

EdmundGoodman:
> Could you please also migrate the makefile? Maybe with a phony target that checks whether uv is installed, and directs developers to installation instructions.

I have added a `uv-installed` phony to do this, which is then a dependency for creating the virtual environment:

```Makefile
.PHONY: uv-installed
uv-installed:
	@command -v uv &> /dev/null ||\
		(echo "UV doesn't seem to be installed, try the following instructions:" &&\
		echo "https://docs.astral.sh/uv/getting-started/installation/" && false)
```

> The goal would be for make tests to be runnable outside of an activated venv

Per Alex's review, I have switched all the Make targets to have the `uv run` prefix, as opposed to activating the venv when creating it with the Make target (which would die each shell session?). This could be something to change if people have strong opinions about it though?

Whilst I was doing all this, I have moved all the Phony definitions to immediately before the target definition (as suggested here <https://clarkgrubb.com/makefile-style-guide#phony-targets>), as in the past I've found it makes it a lot easier to maintain which targets should be phonies than a big list at the top of the file, and in the current state we were already missing a couple. Again, this can be reverted if people have strong opinions about it

EdmundGoodman:
@math-fehr 

> Thanks a lot for the PR! A few quick questions/requests I have:
> 
>     * Can we move the removed requirements.txt in the pyproject.toml? That way we can still track them with dependabot, and also this is the only change that's missing to still support virtualenv. I'd personally would like to still continue supporting venv if possible, as for now it is still way more popular than uv.

As far as I could tell, these were already in the `pyproject.toml` file as well, then the `requirements.txt` file just specified they should be installed as extras? Because of this, I expect something similar to the following should still work to install via `pip`. Perhaps this could be a backup phony target if there is still a use case for it?

```bash
python3 -m pip --require-virtualenv install -e .[gui,jax,riscv,wgpu,onnx,dev]
```

> 
>     * Do you think we can automate the update of uv.lock with dependabot? Currently, we get daily automated updates of dependencies with dependabot, and it would quite annoying to have to update the `uv.lock` manually every time.

It looks like this isn't supported yet, but there is ongoing work on it <https://github.com/dependabot/dependabot-core/issues/10039>?

However, in the meantime Renovate (<https://docs.astral.sh/uv/guides/integration/dependency-bots/>, <https://github.com/renovatebot/renovate>) is supported, and there might be some hacks around it <https://browniebroke.com/blog/keep-uv.lock-file-up-to-date-with-dependabot-updates/>

> 
>     * We will probably have to test pythonpublish after the PR gets merged. Just ping me and I'll try to run it to see if it works or not, but looking at it this should be fine I think.

Cool - I'll ping you when all the rest of the review stuff is done. Perhaps something to consider adding in future is <https://github.com/devpi/devpi>, which might be able to help testing the release flow without such high consequences of broken releases?

math-fehr:
> As far as I could tell, these were already in the pyproject.toml file as well, then the requirements.txt file just specified they should be installed as extras?

My bad, I misunderstood how this worked! Everything good then!

math-fehr:
> and there might be some hacks around it https://browniebroke.com/blog/keep-uv.lock-file-up-to-date-with-dependabot-updates/

I would love having this! I feel that to me, dependabot is quite of a blocker, and I would rather have no lock file than not having dependabot. So having this additional workflow would be perfect so we have everything!

> Something that might not help in this case, but is perhaps worth considering adding in future is https://github.com/devpi/devpi, which might be able to help testing the release flow without such high consequences of broken releases?

Yeah I can try with the test PyPi as well, it's just that I don't really think that your PR will break things ;)
If I manage to get some time, I can test it, but I just have very limited time this and next month 🙃

superlopuh:
For me, the big advantage of dependabot is the notification of the new dependency, it feels like not a lot of effort to run a command locally when a dependency gets updated. We can probably combine this with a lower precision of version control dependency, so we should get way fewer updates.

.github/workflows/jupyterlite.yml
```diff
@@ -1,8 +1,7 @@
 name: Deploy JupyterLite Page
 
 on:
-  # Trigger the workflow on push or pull request,
-  # but only for the master branch
+  # Trigger the workflow every day at 4:15am
```
superlopuh, line 4: 👍

Makefile
```diff
@@ -1,76 +1,85 @@
...
 TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
 
 # make tasks run all commands in a single shell
 .ONESHELL:
 
-# these targets don't produce files:
-.PHONY: ${VENV_DIR}/ venv clean clean-caches filecheck pytest pytest-nb tests-toy tests
-.PHONY: rerun-notebooks precommit-install precommit pyright tests-marimo
-.PHONY: tests-marimo-mlir coverage coverage-tests coverage-filecheck-tests
-.PHONY: coverage-report-html coverage-report-md
```
superlopuh, line 20: nice

requirements.txt
```diff
@@ -1 +0,0 @@
--e .[gui,jax,riscv,wgpu,onnx,dev]
```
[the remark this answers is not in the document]
superlopuh: No, these are in the pyproject.toml, and are installed via `--all-extras`

EdmundGoodman:
@alexarice 

> We've been using `uv` for our quantum stack recently and it seems to work nicely. I have a couple of comments:

Great to hear there is some precedent for using `uv`!


> When to use `uv run`. It seems in the test ci you are activating the venv anyway so I don't see the reason to us `uv run` everywhere.

Unlike `poetry`/`pipenv`/... there isn't a command like `poetry shell` to activate the virtual environment -- it needs to be done manually (for example `source ./venv/bin/activate`, which is OS, shell, and venv location specific). As such, I thought it's probably best to just use `uv run` where possible - with the only exception being the `pyright` action where it can't be used, and we control the environment.

> If you did want to use `uv run` everywhere, you would need to make the filecheck tests use it (i.e. every `xdsl-opt` should be `uv run xdsl-opt`). You can get most of these by modifying `lit.cfg`. I'm not sure if this is actually worth it?

Good catch on `lit.cfg` - I hadn't noticed that! I guess this is a drawback of the `uv run` approach, and there might be a few other weird edge cases to squash?

> I've always had problems with the `wgpu` package and had to manually remove it. Will there be an easy way to do that with the new set up?

This should be fine by directly specifying which extensions in `uv sync`. If wgpu is particularly bad I could add another Make target to specifically exclude it?

>  `uv` should be added to `flake.nix` (I can do this afterwards if necessary)

Awesome, thanks! I don't have much experience with Nix so would appreciate any help, but I guess as you say this can happen a bit later/whenever is convenient to you

alexarice:
For the filecheck tests, there are a bunch of tests that run a custom command at the top of the file. Will probably want to change each `xdsl-opt` to `uv run xdsl-opt` in these though it should be a simple find and replace.

What's the issue with `uv run pyright`?

EdmundGoodman:
I think the final remaining blocker on this is something to replace Dependabot. To keep the desired functionality, we could either:

- Switch to using an action which automatically versions bumps dependencies when pull requesting <https://browniebroke.com/blog/keep-uv.lock-file-up-to-date-with-dependabot-updates/>. This is simple, but could result in including unexpected upstream changes from packages if `pyproject.toml` is not specific enough, and possibly extraneous merges of lock files
- Switch to using [Renovate](https://github.com/renovatebot/renovate) instead of Dependabot. This has feature parity, but may require self-hosting some infrastructure which is very undesirable
- Do nothing, staying with `pip` and Dependabot for installation/version management

tests/filecheck/lit.cfg
```diff
@@ -15,7 +15,7 @@ irdl_to_pyrdl = "xdsl/tools/irdl_to_pyrdl.py"
 config.substitutions.append(('XDSL_ROUNDTRIP', "xdsl-opt %s --print-op-generic --split-input-file | xdsl-opt --split-input-file | filecheck %s"))
```
alexarice, line 15: If I'm not mistaken, this and the line below also need a `uv run`

EdmundGoodman:
> For the filecheck tests, there are a bunch of tests that run a custom command at the top of the file. Will probably want to change each `xdsl-opt` to `uv run xdsl-opt` in these though it should be a simple find and replace.

Do you mean this kind of thing:

```
// RUN: xdsl-opt --split-input-file -t riscemu %s | filecheck %s
```

to something like this?

```
// RUN: uv run xdsl-opt --split-input-file -t riscemu %s | filecheck %s
```

> What's the issue with `uv run pyright`?

Pyright provides [its own action](https://github.com/jakebailey/pyright-action) which looks something like this

```yaml
      - name: Pyright
        uses: jakebailey/pyright-action@v2
        with:
          version: PATH
```

But this needs access to the python dependencies to type check against them, but because the action doesn't let you change the command to have the `uv run` suffix, so the environment needs to be sourced as described here: <https://github.com/jakebailey/pyright-action#providing-a-pyright-version-sourced-from-preexisting-dependencies>.

alexarice:
> 
> Do you mean this kind of thing:
> 
> ```
> // RUN: xdsl-opt --split-input-file -t riscemu %s | filecheck %s
> ```
> 
> to something like this?
> 
> ```
> // RUN: uv run xdsl-opt --split-input-file -t riscemu %s | filecheck %s
> ```
> 

Exactly

alexarice:
It might be possible to do all the filecheck tests with substitutions in `lit.cfg` though I'm not an expert on this

superlopuh:
Wait, why do we need the uv run in the filecheck? I'd rather avoid the code substitution, and just run the test using a uv command if possible

tests/filecheck/lit.cfg
superlopuh: I don't see the need for this

superlopuh:
The only thing to do should be to have a uv equivalent of running `lit -v my_file.mlir` outside of the repo, something like `uv run lit -v my_file.mlir` should work

EdmundGoodman:
Yes - you're right, it all seems to work correctly without adding `uv run` inside the `lit.cfg` if it is invoked from the Makefile. I'll revert [0218bba](https://github.com/xdslproject/xdsl/pull/3294/commits/0218bba7ab9ab862a01ba7c86a14c9f3f2b2e1a0) to get back to this.

alexarice:
Just to check `uv run lit ...` works outside of venv and without any substitutions? If so I apologise for the misinformed suggestion

EdmundGoodman:
As far as I can tell, it seems to give the same results for me as the `main` branch. If you want to try it out yourself:

```bash
git clone -b uv-packaging https://github.com/EdmundGoodman/xdsl
cd xdsl
make venv
make filecheck
```

This doesn't throw any Python errors, and gives this result (the same as the main branch), which I think is because it doesn't like spaces in my directory names...?

```
uv run lit -vv tests/filecheck --order=smart --timeout=20
-- Testing: 362 tests, 8 workers --
FAIL: xDSL :: transforms/convert-scf-to-openmp.mlir (1 of 362)
******************** TEST 'xDSL :: transforms/convert-scf-to-openmp.mlir' FAILED
********************
Exit Code: 127

Command Output (stdout):
--
# preamble command line
cd /Users/edjg/Desktop/60-69 Masters/61 Thesis/xdsl
# executed command: cd /Users/edjg/Desktop/60-69 Masters/61 Thesis/xdsl
# .---command stderr------------
# | 'cd' supports only one argument
# `-----------------------------
# error: command failed with exit status: 127

--
[snip...]
  xDSL :: version.mlir
  xDSL :: with-riscemu/riscv_emulation.mlir
  xDSL :: with-riscemu/rvscf_lowering_emu.mlir
  xDSL :: xdsl_opt/split_input.mlir


Testing Time: 20.23s

Total Discovered Tests: 362
  Unsupported:  54 (14.92%)
  Failed     : 308 (85.08%)
```

EdmundGoodman:
Sorry for taking up so much of you folks' time with this review and thanks for your patience - if this is low priority on the project roadmap feel free to put it on the back burner for a bit

superlopuh:
No need to worry about this, you can trust us to review and comment when we have time :)

I followed your steps and got this:

```
~/Developer/temp ❯ git clone -b uv-packaging https://github.com/EdmundGoodman/xdsl
Cloning into 'xdsl'...
remote: Enumerating objects: 29121, done.
remote: Counting objects: 100% (2330/2330), done.
remote: Compressing objects: 100% (1532/1532), done.
remote: Total 29121 (delta 1611), reused 1294 (delta 760), pack-reused 26791 (from 1)
Receiving objects: 100% (29121/29121), 9.37 MiB | 7.86 MiB/s, done.
Resolving deltas: 100% (21974/21974), done.
~/Developer/temp ❯ cd xdsl                                             17:08:18
~/Dev/t/xdsl uv-packaging ❯ make filecheck                             17:08:23
uv run lit -vv tests/filecheck --order=smart --timeout=20
error: unrecognized subcommand 'run'

Usage: uv [OPTIONS] <COMMAND>

For more information, try '--help'.
make: *** [filecheck] Error 2
~/Dev/t/xdsl uv-packaging ❯ make venv                                  17:08:27
uv sync --all-extras
error: unrecognized subcommand 'sync'

  tip: a similar subcommand exists: 'uv pip sync'

Usage: uv [OPTIONS] <COMMAND>

For more information, try '--help'.
make: *** [.venv] Error 2
```

EdmundGoodman:
Does `uv run` work on its own outside of the Makefile? In that case something is unhappy there (tabs vs spaces is a usual suspect)

Otherwise, could you have an older or strangely symlinked/installed version of uv? `uv run` should work as a subcommand <https://docs.astral.sh/uv/getting-started/features/#scripts>.

To debug, do you get anything different (or some sensible variant if you're not on MacOS) from the following:
```
$ uv --version
uv 0.4.15 (Homebrew 2024-09-21)
$ which uv
/opt/homebrew/bin/uv
```

alexarice:
Also +1 on the not worrying about taking up reviewer time, thanks for putting the work into making this happen.

superlopuh:
TBH I think we should move the wgpu code to a different repo, the cost of having the dependency is not worth the benefit of the project right now. I just installed and ran everything on a github workspace, it would be nice for the makefile to guide the user a bit more towards the ideal state.

This is what I get when I run `make tests` and `make uv-installed` with no `uv`, something doesn't seem quite right:
```
@superlopuh ➜ /workspaces/xdsl (uv-packaging) $ make tests
uv run pytest tests -W error -vv
make: uv: Command not found
make: *** [Makefile:49: pytest] Error 127
@superlopuh ➜ /workspaces/xdsl (uv-packaging) $ make uv-installed
```

The easiest way to test it is probably just to press `.` when looking at this PR, spinning up a codespace, and trying to follow the readme instructions.

superlopuh:
To ease short term pain, I would probably recommend make venv to only install `"[gui,dev]"` and not any of the other optional dependencies, instead of the `all-extras`.

EdmundGoodman:
> TBH I think we should move the wgpu code to a different repo, the cost of having the dependency is not worth the benefit of the project right now. I just installed and ran everything on a github workspace, it would be nice for the makefile to guide the user a bit more towards the ideal state.
>
> To ease short term pain, I would probably recommend make venv to only install "[gui,dev]" and not any of the other optional dependencies, instead of the all-extras.

For now I'll change the Makefile to only do these two as extras, then wgpu can be removed in a later PR?

> This is what I get when I run `make tests` and `make uv-installed` with no `uv`, something doesn't seem quite right:

This still looks like `uv` isn't installed properly, could you run the following?

```
$ uv --version
uv 0.4.21 (Homebrew 2024-10-14)
$ which uv
/opt/homebrew/bin/uv
```

superlopuh:
I mean that's the point, once this is merged, everyone contributing to the project will suddenly not be able to run make tests, and it'll fail with an unknown message. I would love for the makefile targets to print a nice message to guide everyone through the process.

superlopuh:
When uv is installed everything is ok

EdmundGoodman:
Ok - I definitely see now what you mean with wanting to have installation instructions for `uv` when it isn't installed.

I guess the question then is why is the `uv-installed` make target not failing as it should on your machine when `uv` isn't installed? If we fix that the instructions should print and people should be good to go

It might be the semantics of the `command` builtin differs between shells/OSes? When I change the target to check something like `command -v hello` (which isn't installed for me) instead of `command -v uv` it gives the following as I'd expect?

```
$ command -v hello &> /dev/null ||\
    (echo "UV doesn't seem to be installed, try the following instructions:" &&\
    echo "https://docs.astral.sh/uv/getting-started/installation/" && false)
UV doesn't seem to be installed, try the following instructions:
https://docs.astral.sh/uv/getting-started/installation/
make: *** [uv-installed] Error 1
```

Because I'm struggling to replicate it on my machine, would it be possible to get some information about your system (OS/shell) so I can try to understand it better/debug it?

EdmundGoodman:
Ok - I think I see the problem. All the make targets using uv need to have the `uv-installed` dependency, not just creating the venv file.

I assumed that people would be coming in without having cloned it already, make their venv (and install uv then), then do the other targets. In your case, you already have a venv so went straight to the other targets and it failed.

I will update the PR tonight when I'm back from a commitment to hopefully fix this.

EdmundGoodman:
Now the Makefile stuff is hopefully fixed, I've also hacked together something which might be able to be a stopgap for Dependabot until it officially supports `uv`.

I've made a demo repo here: <https://github.com/EdmundGoodman/update-bot>, but the idea is a cron job which runs every week and creates a PR to bump the `uv` lockfile. This slightly differs from Dependabot in it is scheduled and does all version bumps not just security ones, but might give enough functionality for now?

If it seems reasonable, the change will just be adding that workflow YAML file to this PR

superlopuh:
This is beautiful. If I understand you correctly, the idea would be to loosen the dependabot requirements, and update the dev versions in the lockfile once a week? My understanding is that we are already too fine-grained in our dependency constraints.

EdmundGoodman:
> This is beautiful.

😃 ! I have added the new workflow to the PR now

> If I understand you correctly, the idea would be to loosen the dependabot requirements, and update the dev versions in the lockfile once a week?

I think Dependabot doesn't support `uv` at all currently, so we would turn it off (now in the PR), then just bump any dependencies with new versions weekly by merging the automatically generated PR if it passes all the tests.

> My understanding is that we are already too fine-grained in our dependency constraints.

I guess some things like `filecheck==1.0.1` would need to be modified to be `filecheck>=1.0.1` as otherwise `uv` alone wouldn't bump them even if they are known to require a security patch, yes

.github/workflows/update-bot.yml
```diff
@@ -0,0 +1,38 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update uv lockfile
+          title: Update uv lockfile
+          body-path: uv_output.md
+          branch: update-uv
+          base: main
+          labels: install
+          delete-branch: true
+          add-paths: uv.lock
+          assignees: math-fehr, georgebisbas
```
superlopuh, line 38:
```suggestion
          assignees: math-fehr, georgebisbas, superlopuh
```

.github/dependabot.yml.bak
superlopuh, line 1: what is the impact of this change?

superlopuh: Should we just delete it?

EdmundGoodman: Dependabot will no longer run - as it would do nothing anyway or throw an error as the package ecosystem is no longer `pip` as it expects.

An approximation to the behaviour of Dependabot is then supported by the new custom update bot GitHub Action workflow.

When Dependabot (hopefully) eventually supports uv, it can be reverted and the package ecosystem changed to `uv`

If not disabling Dependabot is a strong constraint (which is fairly reasonable), then moving to `uv` is blocked till it offers support for it

EdmundGoodman: Oops, didn't see the follow up message - this is would make it marginally easier to (remember to?) revert it when support is added, but realistically `git revert` would make more sense. I'll just delete it

superlopuh: Yeah I'm pretty sure I'll see on either the twitter or GH when they enable Dependabot for UV, so I don't think it's a big risk.

superlopuh (commented):
What's left on the TODO list to open it up for review?

EdmundGoodman:
> What's left on the TODO list to open it up for review?

I think mostly just the stuff in the top comment:

- Testing against non-MacOS operating systems
- Checking the release flow/hard-to-run pipelines won't explode because of the changes
- (additionally) Checking the update-bot workflow crontab schedules correctly in this instance

It may also be worth ensuring there is buy-in from maintainers and key users before what is a fairly impactful if not large change.

alexarice:
Hi, I added a commit to update the nix flake, hope that is ok. `make tests` works for me locally

alexarice (approved):

superlopuh:
This LGTM, I guess Mathieu will test the release situation once this is merged

EdmundGoodman:
> Hi, I added a commit to update the nix flake, hope that is ok. `make tests` works for me locally

Amazing, thanks!

EdmundGoodman:
> This LGTM, I guess Mathieu will test the release situation once this is merged

Sounds reasonable - if anything explodes for users when it gets merged feel free to ping me here/on Zulip and I can try to them debug things (but hopefully everything will go smoothly 😄)

superlopuh (approved):
@math-fehr

.github/dependabot.yml
```diff
@@ -1,19 +0,0 @@
...
-    schedule:
-      interval: "daily"
-    # Add assignees
-    assignees:
-      - "math-fehr"
-      - "georgebisbas"
-    commit-message:
-      prefix: "pip prod"
-      prefix-development: "pip dev"
-      include: "scope"
```
math-fehr, line 19: Could we actually keep dependabot, though maybe making the schedule to "weekly" instead?

I feel that `update-bot` is not as strong as dependabot, as it will not update the actual version, it will just update the minor ones (and miss security updates).
Having both at the same time is probably what we want here.

README.md
```diff
@@ -89,13 +89,16 @@ To contribute to the development of xDSL follow the subsequent steps.
 
 ### Developer Installation
```
math-fehr, line 90: Can we keep a subsection that just gives the commands for the standard pip installation?
Even though we want to make `uv` the "default", I feel it's important to keep compatibility with `pip`.
We just need to write the `python -m venv venv`, and `source venv/bin/activate`, the `pip install -e .`, and explain that we don't need the `uv run` prefixes. in the rest of the documentation.
That way, we always have a fallback, especially for people already using pip/venv.

math-fehr (commented):
Almost good for me, I just want to be sure we are not regressing too much with dependabot, and keep the documentaiton for doing a `pip` install.

I'll check the release once it's merged, hopefully it won't break!

.github/dependabot.yml
EdmundGoodman: > tl;dr, I've just now caught up with you in realising it might be possible to use Dependabot in this case despite the docs saying its not supported. I'll try to make changes to go back to just using Dependabot and see if something breaks


Long chain of thought to justify to myself this when I come back to it tomorrow:

I've just had a further stare at this and am wondering if Dependabot might just work for `uv` despite not being explicitly supported?

On modern python projects the requirements are often loosely defined in the `pyproject.toml` file (e.g. [`package = ^1.2.1`](https://github.com/Textualize/textual/blob/main/pyproject.toml#L59)), then the specific versions set in a lock file such as `Pipenv.lock`/`poetry.lock`/`uv.lock` - then Dependabot acts [on this lock file](https://github.com/xdslproject/xdsl/pull/3366). Since `uv` isn't a supported lock file type, Dependabot can't be used like this.

However, I've just now realised that since this project is pinning dependency versions in `pyproject.toml` (as `pip` doesn't have a lockfile), Dependabot acts on the `pyproject.toml` [directly](https://github.com/xdslproject/xdsl/pull/3366) -- which I hadn't seen before so didn't think to check. As such if we either: don't commit the `uv` lock file; or use Dependabot in conjunction with update-bot to bump the lock file, it should all work correctly

In future, it might be worth considering using a lock file instead of pinning dependencies in `pyproject.toml` -- but for now I think it saves us from this problem so we can just leave it.

README.md
```diff
@@ -89,13 +89,41 @@ To contribute to the development of xDSL follow the subsequent steps.
...
+For some systems and workflows, changing to a new dependency management system
+may be inconvenient, impractical, or impossible. If this is the case for you,
+xDSL can still be installed using `pip`.
+
+To create the required virtual environment (the equivalent of `make venv`):
+
+```bash
+python -m venv venv
+source venv/bin/activate
+pip install -e .
```
superlopuh, line 115:
```suggestion
pip install -e ".[dev]"
```

math-fehr (approved):
It's good for me to merge now, sorry for the delay!

Makefile
```diff
@@ -1,78 +1,86 @@
 MAKEFLAGS += --no-builtin-rules
 MAKEFLAGS += --no-builtin-variables
 
-# allow overriding the name of the venv directory
-VENV_DIR ?= venv
```
compor, line 5: why is this elided? don't the make rules work with a `VENV_DIR` variable?

.github/workflows/ci-core.yml
```diff
@@ -22,21 +22,17 @@ jobs:
...
-    - name: Upgrade pip
-      run: |
-        pip install --upgrade pip
+    - name: Set up Python ${{ matrix.python-version }}
+      run: uv python install ${{ matrix.python-version }}
 
-    - name: Install the package and dev dependencies locally
-      run: pip install -e ".[dev]"
+    - name: Install the project
+      run: uv sync --all-extras
```
superlopuh, line 35: this one should just install the dev requirements

EdmundGoodman:
Have just resolved the merge conflicts, revealing two final issues:

- `flake.lock` will need to be regenerated to include both `nodejs` and `uv`
- Failing pipeline runs for wgpu tests (https://github.com/xdslproject/xdsl/actions/runs/12200467352/job/34036682236?pr=3294) to be resolved

alexarice:
>     * `flake.lock` will need to be regenerated to include both `nodejs` and `uv`

If you have `nix` installed, you can run `nix flake update`, otherwise I'll can do it and commit to this branch.

Makefile
alexarice: I believe `uv` sets the venv directory with `UV_PROJECT_ENVIRONMENT`
https://docs.astral.sh/uv/configuration/environment/#uv_project_environment

EdmundGoodman: Unfortunately `uv` doesn't support custom virtual environment names currently (https://github.com/astral-sh/uv/issues/1422), so this variable cannot be used for anything and so should be elided to avoid confusion

.github/workflows/ci-core.yml
EdmundGoodman: Fixed by https://github.com/xdslproject/xdsl/pull/3294/commits/cc1743df8a6d9c538f557fc66fcd281edaf9e2f2 switching to the `--only-dev` flag

Makefile
compor: I think this was resolved in what @alexarice via https://github.com/astral-sh/uv/pull/6834, so it would be nice to hook it up to what we currently have been using

EdmundGoodman: Good spot! I've updated the Makefile to use this to respect it as previously

compor (approved):

superlopuh:
Almost there!

The three things I see that are still not working are:
1. the Python-based Testing CI seems not to install pytest in the preparation step, should be easy enough to debug locally
2. the ONNX type errors are similarly because onnx is not installed in that CI action, this is weird to me because it's the same command as the MLIR-based Testing one... There are some differences in the steps taken before the install command, one of which changes the location of the ci, maybe that affects whether uv finds the lockfile, or something like that? Just an idea
3. The test failure to do with a mismatching string is a little weird, because we already test 3.10, 3.11, and 3.12, so it looks like the uv-distributed Pythons have a slightly different error message, it should be easy enough to add it to one of the options in the regex in the test

If you give me commit access I'm happy to do point 3. quicky

superlopuh:
Thanks again @EdmundGoodman !

PR #3301, opened 2024-10-14
dialects: (bufferization) add materialize_in_destination

mamanain:
Currently we need only the command with tensor arguments but we should extend it to memrefs as well at some point.

### DIFF
--- a/xdsl/dialects/bufferization.py
+++ b/xdsl/dialects/bufferization.py
@@ -175,17 +175,17 @@ class MaterializeInDestination(IRDLOperation):
 
     source = operand_def(
         TensorMemrefInferenceConstraint(
-            "T", AnyOf([AnyTensorTypeConstr, AnyUnrankedTensorTypeConstr])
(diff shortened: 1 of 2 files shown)


superlopuh (commented):
could you please add a roundtrip test case to the filechecks of the dialect and mlir interoperation?

xdsl/dialects/bufferization.py
```diff
@@ -168,12 +168,39 @@ class ToMemrefOp(IRDLOperation):
...
 
 
+# now only works for (tensor, tensor) arguments. need to add memref support as well.
+@irdl_op_definition
+class MaterializeInDestination(IRDLOperation):
+    name = "bufferization.materialize_in_destination"
+
+    source = operand_def(
+        TensorMemrefInferenceConstraint(
+            "T", AnyOf([AnyTensorTypeConstr, AnyUnrankedTensorTypeConstr])
```
superlopuh, line 178:
```suggestion
            "T", AnyTensorTypeConstr | AnyUnrankedTensorTypeConstr
```

superlopuh (approved):

PR #3302, opened 2024-10-14
Tensor dialect: CollapseShapeOp

mamanain:
Operation is used in convolution implementation produced by jax.


superlopuh (commented):
code looks great, could you please add roundtrip tests for the dialect and mlir interoperation?

superlopuh (approved):
beautiful

PR #3303, opened 2024-10-15
core: always print unregistered ops

alexarice:
`scf.for` prints it's blocks without terminators if there are no `iter_args`. Unfortunately the check for whether an operation is a terminator returns true by default for unregistered operations (presumably so that they don't break verification), and so they don't get printed.

This change should make it so that these ops always get printed.

### DIFF
--- a/tests/filecheck/dialects/scf/unregistered.mlir
+++ b/tests/filecheck/dialects/scf/unregistered.mlir
@@ -1,0 +1,12 @@
+// RUN: xdsl-opt %s --print-op-generic --allow-unregistered-dialect | xdsl-opt --split-input-file | filecheck %s
+
+  func.func @for_unregistered() {
+    %lb = arith.constant 0 : index
+    %ub = arith.constant 42 : index
+    %s = arith.constant 3 : index
+    scf.for %iv = %lb to %ub step %s {
+      "unregistered_op"() : () -> ()
+      scf.yield
+    }
+    func.return
+  }
--- a/tests/filecheck/dialects/scf/unregistered.mlir
+++ b/tests/filecheck/dialects/scf/unregistered.mlir
@@ -1,12 +1,22 @@
 // RUN: xdsl-opt %s --print-op-generic --allow-unregistered-dialect | xdsl-opt --split-input-file | filecheck %s
 
-  func.func @for_unregistered() {
-    %lb = arith.constant 0 : index
-    %ub = arith.constant 42 : index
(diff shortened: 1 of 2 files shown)


alexarice:
@AntonLydike feel free to add your example as a test case

alexarice:
If not I can make one

superlopuh (commented):
Code change looks good and a simple `"unregistered_op"() : () -> ()` should work as a test, right?

alexarice:
I added a small test, but in the process found that the implicit terminator stuff in `scf.for` doesn't roundtrip. It seems to me that the cause is the `ensure_terminator` function, which from my understanding of the docstring should add a terminator if it is not present, but it instead just throws a verify error

alexarice:
@compor I believe you last touched this code, do you know what the intended behaviour is supposed to be?

tests/filecheck/dialects/scf/unregistered.mlir
```diff
@@ -0,0 +1,22 @@
...
+// CHECK-NEXT:   func.return
+// CHECK-NEXT: }
+
+func.func @for_unregistered() {
+  %lb = arith.constant 0 : index
+  %ub = arith.constant 42 : index
+  %s = arith.constant 3 : index
+  scf.for %iv = %lb to %ub step %s {
+    "unregistered_op"() : () -> ()
+    scf.yield
```
superlopuh, line 19: Am I missing something?
```suggestion
    "unregistered_op"() : () -> ()
```

alexarice: You're missing that I believe this doesn't verify as the implicit terminator isn't added, due to an unrelated problem this pr was trying to solve. (`scf.for` expects to be terminated by an `scf.yield`)

superlopuh: I'm confused, does this test not pass on main?

superlopuh: OK I just tested it out myself and became more confused :) I think a part of the problem is that the code in the printer is a little backwards, and would benefit from a comment. It might also benefit from a bit of a refactor, it feels a little wasteful to check for terminator trait for every operation that we print, when they can only be the last operation in the block. That's probably out of scope for this PR though.

superlopuh (approved):

math-fehr (approved):
Thanks for the fix!

PR #3304, opened 2024-10-15
transformations: (lower-csl-stencil) Add iter args to first region

n-io:
Translate block args for both regions alike in `csl_stencil.apply` lowering.


dk949 (approved):
Looks good, thank you!

PR #3305, opened 2024-10-15
core: Add default valued attributes

alexarice:
Adds the equivalents for the tablegen `DefaultValuedAttribute` and `DefaultValuedOptionalAttribute`. 

This is a blocker for updating mlir.

-----------

@zero9178, would you be able to check if the test cases agree with what you think mlir should be doing, as you seemed to have some idea about what was going on here when I last asked about this.

Lastly is there a nice python way to specify that an `opt_prop_def` should return an `Attribute` instead of `Attribute | None` when it is specified with a default value?


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -2041,3 +2041,79 @@ class MixedOp(IRDLOperation):
...
+
+    parsed = Parser(ctx, "test.default").parse_operation()
+
+    assert isinstance(parsed, DefaultOp)
+
+    assert parsed.def_prop.value.data == 0
+
+    assert parsed.def_opt_prop is not None
+
+    assert parsed.def_opt_prop.value.data == 1
```
zero9178, line 2119: Can we also add a test here where we parse in the operation as a generic but don't specify in the dictionary an optioanl attribute/property? 

In MLIR the auto-generated getter (say `getDefOptProp`) should then still return `True`, while `getAttr("def_opt_propt")` would return null/missing

zero9178 (commented):
Looks great to me and the implementation seems to capture the nuances of the differences between optional and non-optional default values attributes/properties as far as I can tell. Just one test-case is missing for full branch coverage I believe.

superlopuh (approved):

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -2041,3 +2041,133 @@ class MixedOp(IRDLOperation):
...
+    assert parsed.attr.value.data == 0
+
+    assert parsed.attributes.get("opt_attr") is None
+
+    assert parsed.opt_attr is not None
+
+    assert parsed.opt_attr.value.data == 1
+
+
+def test_generic_accessors():
```
alexarice, line 2149: I feel this test particularly but also the one above it should probably not live here, but am not sure exactly where they should go

zero9178 (approved):

tests/irdl/test_declarative_assembly_format.py
alexarice: Moving it to `test_operation_definition.py`

alexarice:
I hit a bug with renamed properties when continuing to update mlir. Should hopefully be fixed in latest commit

math-fehr (approved):
Very cool!
I find ODS semantics quite weird, I didn't know you could have optional attributes with default values, I think it completely breaks the roundtripping. 
Thanks for looking at it and implementing it!

PR #3307, opened 2024-10-16
transformations: Split varith into neighbour and own data across csl_stencil regions

n-io:
Background: The `csl_stencil.apply` op does communicate-and-compute on a given stencil buffer. It holds two regions, one for processing chunks of neighbour data (of this one buffer only), and one region for processing everything else after the exchange is done. The `convert-stencil-to-csl-stencil` pass splits the computation of the `stencil.apply` op across these two regions.

The split was done in two steps, first re-ordering arith ops in the `RestructureSymmetricReductionPattern`, and then calling the `get_ops_split` function on the re-shuffled arith ops. Intuitively, the re-order pass would identify chained reductions (`arith.addf`, `arith.mulf`) and restructure them such that all neighbour data which should end up in the first region is consumed first, and the chained arith ops become easily splittable.

This PR replaces this logic by converting arith to varith, splitting the varith op into neighbour/other data in the `SplitVarithOpPattern` rewrite, and then proceeding with `get_ops_split` and everything else as before. At the end, varith is converted back to arith.

Minor improvements:
* Constants are now always duplicated and appear on both regions, which `dce` can clean up

Note:
* ~~This PR currently has a fix from #3330 merged, will revert changes to `convert-arith-to-varith.mlir` and `varith_transformations.py` as they are not intended to be merged from this PR~~

### DIFF
--- a/xdsl/transforms/convert_stencil_to_csl_stencil.py
+++ b/xdsl/transforms/convert_stencil_to_csl_stencil.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from math import prod
 
 from xdsl.context import MLContext
-from xdsl.dialects import arith, stencil, tensor
+from xdsl.dialects import arith, stencil, tensor, varith
 from xdsl.dialects.builtin import (
     AnyMemRefTypeConstr,
     AnyTensorType,
@@ -15,7 +15,15 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.dialects.csl import csl_stencil
 from xdsl.dialects.experimental import dmp
-from xdsl.ir import Attribute, Block, BlockArgument, Operation, OpResult, Region
+from xdsl.ir import (
+    Attribute,
+    Block,
+    BlockArgument,
+    Operation,
+    OpResult,
+    Region,
+    SSAValue,
+)
 from xdsl.irdl import Operand, base
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
@@ -53,7 +61,40 @@ def get_stencil_access_operands(op: Operand) -> set[Operand]:
     return res
 
 
-@dataclass(frozen=True)
+def _get_prefetch_buf(op: stencil.ApplyOp) -> int | None:
+    # calculate memory cost of all prefetch operands
+    def get_prefetch_overhead(o: OpResult):
+        assert isa(o.type, TensorType[Attribute])
+        buf_count = o.type.get_shape()[0]
+        buf_size = prod(o.type.get_shape()[1:])
+        return buf_count * buf_size
+
+    candidate_prefetches = [
+        (get_prefetch_overhead(o), o)
+        for o in op.operands
+        if isinstance(o, OpResult) and isinstance(o.op, csl_stencil.PrefetchOp)
+    ]
+    if len(candidate_prefetches) == 0:
+        return
+
+    # select the prefetch with the biggest communication overhead to be fused with matched stencil.apply
+    prefetch = max(candidate_prefetches)[1]
+    return op.operands.index(prefetch)
+
+
+def _get_apply_op(op: Operation) -> stencil.ApplyOp | None:
+    """
+    Return the enclosing csl_wrapper.module
+    """
+    parent_op = op.parent_op()
+    while parent_op:
+        if isinstance(parent_op, stencil.ApplyOp):
+            return parent_op
+        parent_op = parent_op.parent_op()
+    return None
+
+
+@dataclass(frozen=False)
 class RestructureSymmetricReductionPattern(RewritePattern):
     """
     Consume data first where that data comes from stencil accesses to `buf`.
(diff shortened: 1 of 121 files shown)


xdsl/transforms/varith_transformations.py
```diff
@@ -209,6 +197,7 @@ def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
                     MergeVarithOpsPattern(),
                 ]
             ),
+            walk_reverse=True,
         ).rewrite_op(op)
```
tobiasgrosser, line 203: This change seems to be already in https://github.com/xdslproject/xdsl/pull/3330/files. Maybe rebase after merging #3330.

n-io: Good spot, I had merged the fix from #3330 into this branch to finish development. I have merged main and the changes have disappeared from this PR's diff.

xdsl/transforms/convert_stencil_to_csl_stencil.py
```diff
@@ -55,69 +67,37 @@ def get_stencil_access_operands(op: Operand) -> set[Operand]:
...
-class RestructureSymmetricReductionPattern(RewritePattern):
-    """
-    Consume data first where that data comes from stencil accesses to `buf`.
+def _get_prefetch_buf(op: stencil.ApplyOp) -> int | None:
+    # calculate memory cost of all prefetch operands
+    def get_prefetch_overhead(o: OpResult):
+        assert isa(o.type, TensorType[Attribute])
+        buf_count = o.type.get_shape()[0]
+        buf_size = prod(o.type.get_shape()[1:])
+        return buf_count * buf_size
```
dk949, line 76:
```suggestion
        return prod(o.type.get_shape())
```

This appears to just do a product of all dimensions?

xdsl/transforms/convert_stencil_to_csl_stencil.py
```diff
@@ -331,14 +313,61 @@ def get_op_split(
                         a.append(use.operation)
                         rem.remove(use.operation)
 
-    if len(a_exports) == 1:
-        return a, b + rem
+    cnst_exports = [cnst for cnst in a_exports if isinstance(cnst, arith.Constant)]
+    if len(a_exports) == 1 + len(cnst_exports):
```
dk949, line 317: A comment here would be great

xdsl/transforms/convert_stencil_to_csl_stencil.py
```diff
@@ -331,14 +313,61 @@ def get_op_split(
...
+        recv_chunk_ops, done_exch_ops = list[Operation](), list[Operation]()
+        for op in ops:
+            if op in a:
+                recv_chunk_ops.append(op)
+                if op in cnst_exports:
+                    done_exch_ops.append(cln := op.clone())
+                    op.result.replace_by_if(
```
dk949, line 324: I'm not sure, but I think this might have to call `rewriter.handle_operation_modification`? Though I have no idea how this works with `replace_by_if`

dk949, line 324: nit: this function is called `get_...`, but it may modify input ops

xdsl/transforms/convert_stencil_to_csl_stencil.py
```diff
@@ -55,69 +67,37 @@ def get_stencil_access_operands(op: Operand) -> set[Operand]:
     return res
 
 
-@dataclass(frozen=True)
-class RestructureSymmetricReductionPattern(RewritePattern):
-    """
-    Consume data first where that data comes from stencil accesses to `buf`.
+def _get_prefetch_buf(op: stencil.ApplyOp) -> int | None:
```
dk949, line 70: nit: can this be `_get_prefetch_buf_index`? Since it doesn't return the buffer, but just its index

n-io: done as suggested.

xdsl/transforms/convert_stencil_to_csl_stencil.py
n-io: Renaming to `split_ops`

xdsl/transforms/convert_stencil_to_csl_stencil.py
n-io: Added comments.

xdsl/transforms/convert_stencil_to_csl_stencil.py
n-io: True, slightly favour the readability here. Changed anyway.

xdsl/transforms/convert_stencil_to_csl_stencil.py
n-io: Theoretically possible, practically the op refactoring of the pass is quite substantial, dividing the ops over two regions. This particular `replace_by_if` replaces an arith.const operand with an exact copy of the arith.const operand inserted in the correct region, which may not have to be flagged as a meaningful change.

n-io: Added some more comments in that region of the code.

dk949 (approved):
Looks good to go :+1:

PR #3309, opened 2024-10-16
transforms: Add convert-varith-to-arith pass

emmau678:
Adding a convert-varith-to-arith pass, which reverses the convert-arith-to-varith pass, rewriting varith operations as a sequence of arith operations.

---

### DIFF
--- a/xdsl/transforms/varith_transformations.py
+++ b/xdsl/transforms/varith_transformations.py
@@ -99,6 +99,11 @@ class VarithToArithPattern(RewritePattern):
 
         # Break the varith op down into a sequence of arith ops
         first_arg = op.operands[0]
+
+        if len(op.operands) == 1:
+            rewriter.replace_matched_op([], new_results=[first_arg])
+            return
+
         for i in range(1, len(op.operands)):
             newop = target_arith_type(first_arg, op.operands[i])
             arith_ops.append(newop)
--- a/xdsl/transforms/varith_transformations.py
+++ b/xdsl/transforms/varith_transformations.py
@@ -28,12 +28,7 @@ ARITH_TO_VARITH_TYPE_MAP: dict[
 VARITH_TO_ARITH_TYPE_MAP: dict[
     type[varith.VarithOp],
     type[arith.SignlessIntegerBinaryOperation | arith.FloatingPointLikeBinaryOperation],
-] = {
-    varith.VarithAddOp: arith.Addi,
-    varith.VarithAddOp: arith.Addf,
-    varith.VarithMulOp: arith.Muli,
-    varith.VarithMulOp: arith.Mulf,
-}
+] = {varith.VarithAddOp: arith.Addi}
 
 
 class ArithToVarithPattern(RewritePattern):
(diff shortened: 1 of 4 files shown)


xdsl/transforms/varith_transformations.py
```diff
@@ -64,6 +75,36 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
...
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
+        # check that the op is of a type that we can convert to arith
+        if type(op) not in VARITH_TO_ARITH_TYPE_MAP:
+            return
+
+        # this must be true, as all keys of VARITH_TO_ARITH_TYPE_MAP are varith ops
+        op = cast(
+            varith.VarithOp,
+            op,
+        )
```
n-io, line 92:
```suggestion
    def match_and_rewrite(self, op: varith.VarithOp, rewriter: PatternRewriter, /):
```
One really nice feature of the API, it can match on exactly what you want it to match on. This can be either the base class, or more specific types like `varith.add`

xdsl/transforms/varith_transformations.py
```diff
@@ -64,6 +75,36 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
...
+
+        dest_type = VARITH_TO_ARITH_TYPE_MAP[type(op)]
```
n-io, line 94: The map currently has duplicate keys, but you can distinguish between int/float by looking at `op.res.type` (try a debugger)

n-io (commented):
Looks really good, thanks for that. There's an edge case where the varith op only has one operand, in which case you could just output the operand without generating an arith op, e.g. `replace_matched_op([], new_results=[first_arg])`

xdsl/transforms/varith_transformations.py
```diff
@@ -171,3 +221,22 @@ def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
...
+
+    name = "convert-varith-to-arith"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    VarithToArithPattern(),
+                ]
+            ),
```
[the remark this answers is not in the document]
n-io: nit: The `apply_recursively` flag is probably optional here. Some passes need it to be set in a specific way in order to work as intended.

xdsl/transforms/varith_transformations.py
```diff
@@ -64,6 +67,53 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
...
+        target_arith_type = ARITH_TYPES[(type_name, kind)]
+
+        arith_ops: list[Operation] = []
+
+        # Break the varith op down into a sequence of arith ops
+        first_arg = op.operands[0]
+
+        if len(op.operands) == 1:
+            rewriter.replace_matched_op([], new_results=[first_arg])
+            return
```
n-io, line 105: 👍

xdsl/transforms/varith_transformations.py
```diff
@@ -64,6 +67,53 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
...
+
+        for i in range(1, len(op.operands)):
+            newop = target_arith_type(first_arg, op.operands[i])
+            arith_ops.append(newop)
+            first_arg = newop.result
+
+        rewriter.replace_matched_op(arith_ops)
+
+        return
+
```
n-io, line 115:
```suggestion
```
nit: not really needed

xdsl/transforms/varith_transformations.py
```diff
@@ -64,6 +67,53 @@ def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
...
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
+        # check that the op is of a type that we can convert to arith
+        if type(op) not in VARITH_TYPES:
+            return
+
+        # this must be true, as all keys of VARITH_TO_ARITH_TYPE_MAP are varith ops
+        op = cast(
+            varith.VarithOp,
+            op,
+        )
```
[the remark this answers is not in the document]
n-io: xdsl actually provides this logic for you with the annotation, allowing you to match on the specified type (only) directly

n-io (commented):
Looks really good! Just minor things and API suggestions left

n-io (approved):
LGTM, thanks for this

tobiasgrosser:
Nice. Amazing to see this merged so quickly.

PR #3310, opened 2024-10-16
Update mlir to d401987fe349a87c53fe25829215b080b70c0c1a

### LINKED ISSUE #3296: Update mlir version
Going to keep a running log of what I'm doing. See https://github.com/xdslproject/xdsl/tree/alexarice/mlir-update

- Updated mlir hash in 3 locations:
  - https://github.com/xdslproject/xdsl/blob/12dc46b5854a24621230ca2a07359a0562228db1/README.md?plain=1#L47
  - https://github.com/xdslproject/xdsl/blob/12dc46b5854a24621230ca2a07359a0562228db1/docs/mlir_interoperation.md?plain=1#L28
  - https://github.com/xdslproject/xdsl/blob/12dc46b5854a24621230ca2a07359a0562228db1/.github/workflows/ci-mlir.yml#L23

### LINKED ISSUE #3297: Update `scf.parallel`
In mlir master, `scf.parallel` is now terminated by a single `scf.reduce`, which can now store multiple reductions.

### LINKED ISSUE #3298: arith.overflow
Seems to be a new(?) attribute `arith.overflow` in mlir master

alexarice:
Updates mlir.

Still work in progress. This is obviously a huge PR but I've tried to split each commit into a reasonable logical block. 

As a note for the future: don't leave it this long.

Fixes #3298 
Fixes #3297 
Fixes #3296


alexarice:
Main changes so far are:
- [x] scf.parallel is now terminated by a single `scf.reduce` which may have multiple regions
- [x] `arith.{addi,subi,muli,shli}` now have a new `arith.overflow` attr, which defaults to `#arith.overflow<"none">`
- [x] `arith.cmpf` has a fastmath property (see https://github.com/xdslproject/xdsl/pull/3275)
- [x] Some `gpu.launch` has more operands and `gpu.module` takes it's symbol as a property
- [x] `omp.parallel` has an extra operand
- [ ] `omp.wsloop` seems to have been changed quite a bit
- [ ] `llvmir` is broken a bit
- [ ] `memref` has changed

alexarice:
Just missing llvm fixes now and then should be ready for review

math-fehr (approved):
Thanks a lot for all the work!
Yes, we should make sure to update it more frequently 🙃

tests/interpreters/test_pdl_interpreter.py
```diff
@@ -70,6 +71,9 @@ def test_rewrite_swap_inputs_pdl():
         apply_recursively=False,
     ).rewrite_module(input_module)
 
+    printer = Printer(print_generic_format=True)
+    print(input_module.print(printer))
+    print(output_module.print(printer))
```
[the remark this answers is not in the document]
alexarice: Good spot

xdsl/traits.py
```diff
@@ -50,6 +50,9 @@ def __init__(self, head_param: type[Operation], *tail_params: type[Operation]):
 
     def verify(self, op: Operation) -> None:
         parent = op.parent_op()
+        # Don't check parent when op is detached
+        if parent is None:
+            return
```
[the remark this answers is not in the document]
alexarice: Nothing has changed with `mlir-opt` for this, the reasons for this were the following:
- At least for me the original meaning was not the obvious meaning of the trait (i.e. I didn't realise it prohibited the operation from being detached.)
- It makes the tests for `scf.reduce` cleaner as they don't have to be put in an `scf.parallel` (see https://github.com/xdslproject/xdsl/blob/b396db84899c11a36c8cdbf8af2054983c8a32c9/tests/dialects/test_scf.py#L280 for an example)

That said I would understand if this change isn't wanted and can revert it and change the tests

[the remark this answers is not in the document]
superlopuh: what does MLIR do here? it feels like this would be better-off done in a separate PR

tests/filecheck/transforms/scf-parallel-loop-tiling-partial.mlir
```diff
@@ -54,9 +54,9 @@ func.func @tile_partial() {
 // CHECK-FIRST-NEXT:     ^{{.*}}(%{{.*}} : index, %{{.*}} : index, %{{.*}} : index):
 // CHECK-FIRST-NEXT:       %{{.*}} = arith.addi %{{.*}}, %{{.*}} : index
 // CHECK-FIRST-NEXT:       %{{.*}} = arith.addi %{{.*}}, %{{.*}} : index
-// CHECK-FIRST-NEXT:       scf.yield
+// CHECK-FIRST-NEXT:       "scf.reduce"() : () -> ()
```
superlopuh, line 57: this is definitely a nit, but it's a bit of a shame to add so many ops in generic syntax when the custom syntax is not very involved. Do you know if the custom syntax has changed from our version? I could quickly open a PR with the custom syntax with our current MLIR, to avoid this micro regression

xdsl/traits.py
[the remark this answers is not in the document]
math-fehr: I would agree that a separate PR would make more sense, and I would agree that this change is beneficial.
I don't think MLIR does anything that is relevant here, as it is not common in MLIR to create new operations without a parent. Usually there is only a single `ModuleOp` with no parents, and all new operations are directly added in it.

tests/pattern_rewriter/test_pattern_rewriter.py
superlopuh, line 1: this makes me a little bit sad :) we should probably migrate this to the test dialect at some point

tests/test_builtin_traits.py
```diff
@@ -68,26 +68,6 @@ class HasMultipleParentOp(IRDLOperation):
...
-
-    has_multiple_parent_op = HasMultipleParentOp()
-    message = (
-        "'test.has_multiple_parent' expects parent op to "
-        "be one of 'test.parent', 'test.parent2'"
-    )
-    with pytest.raises(VerifyException, match=message):
-        has_multiple_parent_op.verify()
-
-
```
superlopuh, line 90: if we're changing the behaviour then I would expect an update of the test, to one that succeeds if the operation has no parents. (I still think a different PR would be a better place for this change)

xdsl/dialects/scf.py
```diff
@@ -559,79 +555,78 @@ def verify_(self) -> None:
...
+            ]
+        )
+    )
 
     def __init__(
         self,
-        argument: SSAValue | Operation,
-        block: Block,
+        args: Sequence[SSAValue | Operation] = (),
+        blocks: Sequence[Block] = (),
```
superlopuh, line 588: it feels to me like this should take a sequence of regions, to a void a situation where you have a list of regions, and need to create a tuple of first blocks, only to then create new regions with those blocks.

xdsl/dialects/memref.py
```diff
@@ -504,12 +504,8 @@ def from_memref(memref: Operation | SSAValue):
 
 
 class AlterShapeOp(IRDLOperation):
```
superlopuh, line 506:
```suggestion
class AlterShapeOperation(IRDLOperation, abc.ABC):
```

tests/pattern_rewriter/test_pattern_rewriter.py
superlopuh: #3322

xdsl/traits.py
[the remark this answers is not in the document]
superlopuh: Yeah makes sense

tests/filecheck/transforms/scf-parallel-loop-tiling-partial.mlir
alexarice: `scf.reduce` has a non-trivial custom syntax I believe, and it'll be completely different to the mlir used for xdsl main

xdsl/traits.py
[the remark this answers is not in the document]
alexarice: I'll move this now, and change the test

alexarice:
Made #3323 for the HasParent change and rebased on top

xdsl/dialects/gpu.py
```diff
@@ -360,7 +360,7 @@ class ModuleOp(IRDLOperation):
     name = "gpu.module"
 
     body = region_def("single_block")
-    sym_name = attr_def(StringAttr)
+    sym_name = prop_def(StringAttr)
```
kimxworrall, line 363: Are these fixed everywhere now?

kimxworrall (approved):

xdsl/dialects/gpu.py
alexarice: no

tests/pattern_rewriter/test_pattern_rewriter.py
superlopuh: no, I don't think it should be blocked by this

tests/filecheck/transforms/scf-parallel-loop-tiling-partial.mlir
superlopuh: Ah, ok, it looks fairly straightforward in the new docs, would it make sense to copy that as part of this PR? I think if it works without too much effort then it would probably be beneficial

superlopuh (approved):
Thank you very much! The reduce thing is more of a nit, but I would appreciate it if you could try it to see if it's easy

tests/filecheck/transforms/scf-parallel-loop-tiling-partial.mlir
alexarice: You're right, it doesn't look as bad as I thought (`scf.parallel` on the othre hand is bad)

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -432,7 +433,7 @@ def parse_optional_variable(
             match region_def:
                 case OptRegionDef():
                     return OptionalRegionVariable(variable_name, idx)
-                case VarRegionDef():
+                case VarRegionDef() | VarSingleBlockRegionDef():
```
alexarice, line 436: @superlopuh needed to fix this bug but was easy other than that

alexarice: Could alternatively make `VarSingleBlockRegionDef` inherit from `VarRegionDef`

superlopuh: LGTM, would be good for @math-fehr to also approve this fix. I think the `|` is neater than inheritance, but I'm not sure if I'm missing some semantic nuance

math-fehr:
Sorry, I didn't realized you wanted my approval again!
Feel free to merge it as soon as you can, so there won't be any more conflict (I see one in arith right now)

PR #3312, opened 2024-10-16
transformations: Support devito timers in the csl pipeline

n-io:


dk949 (approved):
Looks good to me, I'm guessing we'll never be in a situation where we have two timers (if we are, the exported names will clash)

n-io:
> Looks good to me, I'm guessing we'll never be in a situation where we have two timers (if we are, the exported names will clash)

There is a scenario where this can happen, despite (because we're?) taking the names from the name hints.

PR #3315, opened 2024-10-17
transformations: add eqsat-add-costs

### LINKED ISSUE #3171: transformations: (eqsat) add pass to add costs to saturated eqsat code

cowardsa:
Fixes #3171


tests/filecheck/transforms/eqsat-simplify-expressions.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: xdsl-opt -p eqsat-add-costs %s | filecheck %s
```
alexarice, line 1: Should this file be called `eqsat-add-costs.mlir`?

xdsl/transforms/eqsat_add_costs.py
```diff
@@ -0,0 +1,66 @@
...
+        }
+       ```
+    Output example:
+        ```mlir
+        func.func @test(%a : index, %b : index) -> (index) {
+           %c_ab = arith.addi %a, %b : index
+           func.return %c_ab : index
+        }
+        ```
+    """
```
alexarice, line 54: This doesn't seem to agree with what the pass does?

xdsl/transforms/eqsat_add_costs.py
```diff
@@ -0,0 +1,66 @@
...
+    name = "eqsat-add-costs"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    AddCostEclass(),
+                ]
+            ),  # list of rewrite patterns
+            apply_recursively=True,  # do we apply rewrites in a while loop
```
alexarice, line 65: This is the default

xdsl/transforms/eqsat_add_costs.py
```diff
@@ -0,0 +1,66 @@
...
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: eqsat.EClassOp, rewriter: PatternRewriter):
+        for operand in op.operands:
+            if not isinstance(operand, OpResult):
+                # only add costs to operators
+                continue
+
+            # Add cost attribute to operators
+            operation = operand.op
+            operation.attributes["cost"] = IntAttr(1)
```
superlopuh, line 29: this is a pretty common name, and it might clash with another attribute in the future. Let's rename this to `"eqsat_cost"`? Also, could you please extract this to a top-level constant called `EQSAT_COST_LABEL` or similar so that other passes get automatically updated if this is changed in the future?

xdsl/transforms/eqsat_add_costs.py
```diff
@@ -0,0 +1,66 @@
...
+
+    name = "eqsat-add-costs"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            GreedyRewritePatternApplier(
+                [
+                    AddCostEclass(),
+                ]
+            ),  # list of rewrite patterns
```
superlopuh, line 64: I'm not sure the comments are very informatice, I would recommend deleting them. Also, I would expect this pass to work the same without the greedy rewrite applier and with `apply_recursively` set to `False`

tests/filecheck/transforms/eqsat-simplify-expressions.mlir
cowardsa: Agreed - will update

xdsl/transforms/eqsat_add_costs.py
cowardsa: Have moved this upwards in the pass hierarchy but not sure where it should best be defined? In the dialect definition file?

xdsl/transforms/eqsat_add_costs.py
cowardsa: Don't think my pass is actually working it only assigns costs to some operators - e.g. in the example the multiplier does not get assigned a cost? Any ideas why it's not matching on all the e-class operators?

alexarice: I think your pass works but the printer for some of the arith operations is faulty and doesn't print attributes

alexarice: See here:
https://github.com/xdslproject/xdsl/blob/7c86cfba1c8617f9da62b5a752222310008ec86d/xdsl/dialects/arith.py#L204-L210

cowardsa: Nice thanks for the clarification!

xdsl/transforms/eqsat_add_costs.py
```diff
@@ -0,0 +1,67 @@
...
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: eqsat.EClassOp, rewriter: PatternRewriter):
+        for operand in op.operands:
+            if not isinstance(operand, OpResult):
+                # only add costs to operators
+                continue
+
+            # Add cost attribute to operators
+            operation = operand.op
+            operation.attributes[self.EQSAT_COST_LABEL] = IntAttr(1)
```
superlopuh, line 30: I have a feeling that it would be good to start with a slightly different approach, I propose the following:
1. walk the IR to find all the eclass ops
2. get their parent blocks and store in a `set`
3. on each block call a helper called `add_eclass_costs(block: Block` that walks the ops and looks for a cost function for each op class
4. do not add costs for unknown functions
5. add a dummy cost function for the ops in this example, we can get away with a hardcoded dictionary of cost functions for the foreseeable future, can make it extensible after the end-to-end flow works

What do you think about this? One of the things that worries me about the example in this PR is that the first eclass op is the op of one of the operands, but I don't think we want to add costs to those here.

jianyicheng: This might be too complicated for the toy example, and could be a separate pass?
I am thinking of the following:
1. A pass to add costs as attributes (cost = 0 for every non-eclass op, basically what this PR does)
2. A pass to update cost values (this is what you proposed, I think)

jianyicheng: Maybe I misunderstood the code... Yeah, that would be helpful. Thanks!

superlopuh:
As per our discussion, let's merge these changes to main

tests/filecheck/transforms/eqsat-add-costs.mlir
```diff
@@ -0,0 +1,70 @@
...
+
+// CHECK:         func.func @test(%a : index, %b : index) -> index {
+// CHECK-NEXT:      %a_eq = eqsat.eclass %a {"eqsat_cost" = #builtin.int<0>} : index
+// CHECK-NEXT:      %one = arith.constant {"eqsat_cost" = #builtin.int<1>} 1 : index
+// CHECK-NEXT:      %one_eq = eqsat.eclass %one {"eqsat_cost" = #builtin.int<1>} : index
+// CHECK-NEXT:      %amul = arith.muli %a_eq, %one_eq {"eqsat_cost" = #builtin.int<2>} : index
+// CHECK-NEXT:      %out = eqsat.eclass %amul, %a_eq {"eqsat_cost" = #builtin.int<0>} : index
+// CHECK-NEXT:      func.return %out : index
+// CHECK-NEXT:    }
+func.func @test(%a : index, %b : index) -> (index) {
```
alexarice, line 11: Is there a more informative name than test?

xdsl/transforms/__init__.py
```diff
@@ -558,5 +563,6 @@ def get_varith_fuse_repeated_operands():
         "test-add-timers-to-top-level-funcs": get_test_add_timers_to_top_level_funcs,
         "test-lower-linalg-to-snitch": get_test_lower_linalg_to_snitch,
         "eqsat-create-eclasses": get_eqsat_create_eclasses,
+        "eqsat-add-costs": get_eqsat_add_costs,
```
alexarice, line 566: I thought this list was meant to roughly be in alphabetical order but it seems not looking at the rest of it

xdsl/transforms/eqsat_add_costs.py
alexarice: The comment still doesn't agree with what the parse does

xdsl/transforms/__init__.py
superlopuh: Yeah it got messed up recently, I want to do a separate PR sorting this list, I'll open it in parallel to this and adjust the one that gets approved second.

superlopuh: #3426

xdsl/transforms/eqsat_add_costs.py
```diff
@@ -0,0 +1,70 @@
...
+        op.attributes[EQSAT_COST_LABEL] = IntAttr(cost)
+
+
+class EqsatAddCostsPass(ModulePass):
+    """
+    Add costs to all operations in blocks that contain eqsat.eclass ops.
+    The cost of an eclass operation is the minimum of all the costs of the operations of
+    the operands, if these are all non-`None`, and `None` otherwise.
+    The cost for all other operations is currently set to the costs of all the
+    operations of the operands, if these are all non-`None` + 1, and `None` otherwise.
```
alexarice, line 56:
```suggestion
    operations of the operands + 1, if these are all non-`None`, and `None` otherwise.
```

alexarice: Think this is slightly easier to read

alexarice (approved):

PR #3318, opened 2024-10-17
core: introduce TypedAttributeConstraint

alexarice:
Experiment in using Constraint variables to grab a type from an attribute, and an example of using this for `arith.constant`.

Unsure if this a good idea or cooking too hard.

Edit: I also couldn't really figure out how `TypedAttribute` worked, and was not sure if the functionality could be reused (I have use cases in mind where the associated type would not be a parameter)

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -8,6 +8,7 @@ from xdsl.dialects.builtin import (
     AnyFloat,
     AnyFloatConstr,
     AnyIntegerAttr,
+    AnyIntegerAttrConstr,
     ContainerOf,
     DenseIntOrFPElementsAttr,
     Float16Type,
@@ -25,8 +26,11 @@ from xdsl.dialects.builtin import (
 from xdsl.dialects.utils import FastMathAttrBase, FastMathFlag
 from xdsl.ir import Attribute, BitEnumAttribute, Dialect, Operation, SSAValue
 from xdsl.irdl import (
+    AnyAttr,
     AnyOf,
+    BaseAttr,
     IRDLOperation,
+    TypedAttributeConstraint,
     VarConstraint,
     base,
     irdl_attr_definition,
@@ -48,7 +52,6 @@ from xdsl.traits import (
     Pure,
 )
 from xdsl.utils.exceptions import VerifyException
-from xdsl.utils.isattr import isattr
 from xdsl.utils.str_enum import StrEnum
 
 boolLike = ContainerOf(IntegerType(1))
@@ -124,10 +127,20 @@ class IntegerOverflowAttr(BitEnumAttribute[IntegerOverflowFlag]):
 @irdl_op_definition
 class Constant(IRDLOperation):
     name = "arith.constant"
-    result = result_def(Attribute)
-    value = prop_def(Attribute)
+    _T: ClassVar = VarConstraint("T", AnyAttr())
+    result = result_def(_T)
+    value = prop_def(
+        TypedAttributeConstraint(
+            AnyIntegerAttrConstr
+            | BaseAttr[FloatAttr[AnyFloat]](FloatAttr)
+            | BaseAttr(DenseIntOrFPElementsAttr),
+            _T,
+        )
+    )
 
     traits = traits_def(ConstantLike(), Pure())
+
+    assembly_format = "attr-dict $value"
 
     @overload
     def __init__(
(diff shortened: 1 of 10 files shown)


alexarice:
I also believe both failing tests are actually bugs exposed by this change rather than bugs with this change

superlopuh (commented):
On one hand, you're cooking pretty hard indeed, on the other this is great. Probably worth integrating with `TypedAttribute[Attribute]`

math-fehr:
Do you think you could have an inheritance like `WithType` -> `TypedAttribute`?
`WithType` just means your attribute has a canonical type. On the other hand, `TypedAttribute` means that you have a type, that the type is stored in one of the parameters, and that the type is parsed after the `:`. So `TypedAttribute` should be a strict subset of `WithType`.

My goal with `TypedAttribute` was originally to handle all cases that are (or can be) parsed with `:`.
You said you had in mind attributes that do not have types in their parameters, what example do you have in mind?

alexarice:
@math-fehr with the `TypedAttribute` class, is it guaranteed that there is a `type` attribute? Pyright doesn't seem to know this but I found a place where I think it was being runtime enforced

alexarice:
This is the code I was referring to:
https://github.com/xdslproject/xdsl/blob/61bf63244207133d9394260d5a7a4bcb568dcc4e/xdsl/irdl/attributes.py#L163-L181
perhaps this is just checking that the type parameter can be found in the correct place.

I'll rebase this in the meantime

alexarice:
I looked again and `IntegerAttr` seems to be the only `TypedAttribute` in xdsl

alexarice:
Would you expect `DenseIntOrFPElementsAttr` to be a `TypedAttribute`? I'm beginning to think that the functionality of "having an associated type" and "printing and parsing with/without a type" are slightly separate

alexarice:
I want to add that I've been experimenting with traits like this a bit more, and it's all kind of evil as it all depends on what order you "verify" constraints in. Ideally, I feel that the "verify" methods should not solve any constraints and all solving of constraint variables should be done by a separate system which does not depend on the order of constraints, though this would need a bigger refactor.

If anyone has had any ideas on this in general, it would be good to know.

xdsl/dialects/arith.py
```diff
@@ -123,11 +127,21 @@ def __init__(self, flags: None | Sequence[IntegerOverflowFlag] | Literal["none"]
 @irdl_op_definition
 class Constant(IRDLOperation):
     name = "arith.constant"
-    result = result_def(Attribute)
-    value = prop_def(Attribute)
+    _T: ClassVar[AttrConstraint] = VarConstraint("T", AnyAttr())
```
superlopuh, line 130:
```suggestion
    _T: ClassVar = VarConstraint("T", AnyAttr())
```

math-fehr:
> Would you expect `DenseIntOrFPElementsAttr` to be a `TypedAttribute`? I'm beginning to think that the functionality of "having an associated type" and "printing and parsing with/without a type" are slightly separate

Sorry I thought I answered this but I forgot to.
If I recall, these are merged in MLIR. In MLIR, all attribute have a `type` field (most of them are just `None` though).
So if you have an associated type, it will be printed/parsed with the colon. That's my recollection at least.
We are free to do however we want, though I would likely say that these should be merged.

math-fehr:
> I want to add that I've been experimenting with traits like this a bit more, and it's all kind of evil as it all depends on what order you "verify" constraints in. Ideally, I feel that the "verify" methods should not solve any constraints and all solving of constraint variables should be done by a separate system which does not depend on the order of constraints, though this would need a bigger refactor.
> 
> If anyone has had any ideas on this in general, it would be good to know.

I think we need to provide an actual order on the traits yes.
This is something that is also broken in MLIR btw, and I am not completely sure on how to fix this in general

superlopuh:
Temporary solutions have a tendency to become permanent. A week is not a long time, and I would feel better if we could all be happy with this PR before merging it. If you see a better to implement it in the relatively near future, then let's wait a little bit longer.

superlopuh:
Oh I forgot about this PR. I'm not sure we reached a concrete conclusion, it would be good to chat about it now that PLDI rush is over.

superlopuh (approved):

alexarice:
I've rebased on top of the other changes. Finally think this is ready.

superlopuh (approved):

PR #3319, opened 2024-10-17
transformations: (csl) use constants instead of params

dk949:
`csl.param`s allow for flexibility we do not utilise whilst providing a possibility of accidentally overriding a value which was not intended to be overridden. In the `ModuleOp.get_param_value` function we even use the value of the param as if it will always be there.

After this PR: `csl_wrapper.ParamAttribute` can no longer omit its value (now referred to as "value" not "default")

In the layout module:
* All `param`s are now `arith.constant`s

In the program module:
* All params from the `csl_wrapper.ParamAttribute` are now `arith.constant`s
* Params from the layout module `yield` arguments are still `csl.param`s.

### DIFF
--- a/tests/filecheck/dialects/csl/csl-wrapper-ops.mlir
+++ b/tests/filecheck/dialects/csl/csl-wrapper-ops.mlir
@@ -3,7 +3,7 @@
 
 builtin.module {
     "csl_wrapper.module"() <{"width"=10 : i16, "height"=10: i16, "params" = [
-        #csl_wrapper.param<"z_dim" default=4: i16>, #csl_wrapper.param<"pattern" : i16>
+        #csl_wrapper.param<"z_dim" value=4: i16>, #csl_wrapper.param<"pattern" value=2: i16>
     ]}> ({
         ^0(%x: i16, %y: i16, %width: i16, %height: i16, %z_dim: i16, %pattern: i16):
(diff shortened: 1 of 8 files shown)


n-io (commented):
Yes, unfortunately our programs are currently not general enough to support freely modifying the `csl.param`s. I still retain some hope that we (or others in other contexts?) may eventually get there, and this discrepancy may not be the fault of the param translation. Translating `ParamAttribute` to `arith.constant` seems the right fix for us, but semantically - and long-term - translating `ParamAttribute` to `csl.ParamOp` appears logical.

Introducing a `ConstAttribute` alongside the param attribute might be cleaner but also cumbersome. We could potentially introduce a flag to lower-csl-wrapper e.g. a boolean `lower-params-as-const` that enables lowering params with a default/value to consts when set?

dk949:
I have a suggestion then: at the moment, we rename `ParamAttribute` to `ConstAttribute` and if/when we ever need to pass `param`s to the layout module, we can reintroduce the `ParamAttribute`.

Note that it's still possible to pass `param`s to the progr6midule by yielding from the layout region.

superlopuh:
What's the latest on this? Can we close the PR until a solution is found?

dk949:
> What's the latest on this? Can we close the PR until a solution is found?

Yes, I think this can be sorted out at some other time

PR #3324, opened 2024-10-18
transformations: (lower-csl-wrapper) Add params_as_consts flag

n-io:
CSL module wrapper params are by default lowered to CSL params, which can have a value specified in code that can be overridden manually by providing command line inputs. If this is not desired, module wrapper params can be lowered to constants instead by setting the `params_as_consts` flag. For this to happen, the module wrapper params must be numerical and have a specified default value.

Use this flag for programs if the generated program does not support arbitrary param values.


tobiasgrosser:
The PR does not indicate why this flag is needed and also lacks a testcase. Either add a test case or explain in the commit message why no test case can be added.

tests/filecheck/transforms/lower-csl-wrapper.mlir
```diff
@@ -153,29 +156,124 @@ builtin.module {
...
+// CONST-NEXT:         "csl.fmuls"(%arg3_1, %arg3_1, %7) : (memref<510xf32>, memref<510xf32>, f32) -> ()
+// CONST-NEXT:         csl_stencil.yield %arg3_1 : memref<510xf32>
+// CONST-NEXT:       }) to <[0, 0], [1, 1]>
+// CONST-NEXT:       csl.return
+// CONST-NEXT:     }
+// CONST-NEXT:     %0 = "csl.member_access"(%memcpyMod_2) <{"field" = "LAUNCH"}> : (!csl.imported_module) -> !csl.color
+// CONST-NEXT:     "csl.rpc"(%0) : (!csl.color) -> ()
+// CONST-NEXT:   }) {"sym_name" = "gauss_seidel_func_program"} : () -> ()
+// CONST-NEXT: }
+// CONST-EMPTY:
```
tobiasgrosser, line 279: This test case is huge. Do we need such a large test case to test this feature? Or can there be a smaller minimal test case?

emmau678 (approved):
looks good to me!

tobiasgrosser (approved):
Besides my one comment, this looks great.

tests/filecheck/transforms/lower-csl-wrapper.mlir
n-io: I added a shorter test, with and without the flag. @dk949

tobiasgrosser (approved):

PR #3326, opened 2024-10-18
misc: remove type annotations from VarConstraint ClassVars

superlopuh:
It turns out that Pyright can infer them, makes relevant code much cleaner IMO

### DIFF
--- a/xdsl/irdl/operations.py
+++ b/xdsl/irdl/operations.py
@@ -966,6 +966,23 @@ def traits_def(
 _OPERATION_DICT_KEYS = {key for cls in Operation.mro()[:-1] for key in cls.__dict__}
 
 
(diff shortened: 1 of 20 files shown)


xdsl/irdl/operations.py
```diff
@@ -966,6 +966,23 @@ def traits_def(
...
+    where the constness is defined by convention with an UPPER_CASE name and enforced by
+    pyright.
+    The type annotation can be one of
+     * `ClassVar[MyType]`,
+     * `ClassVar`, or
+     * `"ClassVar[MyType]"`.
+    """
+    return field_name.isupper() and (
+        get_origin(annotation) is ClassVar
+        or annotation is ClassVar
```
superlopuh, line 981: new

PR #3339, opened 2024-10-24
fix: (csl) Print float zero as `0.0`, not `0`

dk949:
When printed as `0`, type is assumed to be `comptime_int` which cannot be implicitly cast to float.


emmau678 (approved):
looks good to me

PR #3342, opened 2024-10-25
dialects: (varith) add select

alexarice:
Adds a `varith.select function`

I'm not 100% the syntax is ideal.

I have also not included any sort of default case, but don't think this should be too troublesome? (Maybe it could be an optional argument?)

### DIFF
--- a/tests/filecheck/dialects/varith/varith_ops.mlir
+++ b/tests/filecheck/dialects/varith/varith_ops.mlir
@@ -29,3 +29,7 @@
 %x6 = "varith.mul"(%ta, %tb, %tc, %td) : (tensor<10xf32>, tensor<10xf32>, tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
 // CHECK:  %x6 = varith.mul %ta, %tb, %tc, %td : tensor<10xf32>
 // CHECK-GENERIC: %x6 = "varith.mul"(%ta, %tb, %tc, %td) : (tensor<10xf32>, tensor<10xf32>, tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
+
+%x7 = "varith.select"(%ia, %fa, %fb, %fc, %fd) : (i32, f32, f32, f32, f32) -> f32
+// CHECK: %x7 = varith.select(%ia : i32) %fa, %fb, %fc, %fd
+// CHECK-GENERIC: %x7 = "varith.select"(%ia, %fa, %fb, %fc, %fd) : (i32, f32, f32, f32, f32) -> f32
--- a/xdsl/dialects/varith.py
+++ b/xdsl/dialects/varith.py
@@ -13,10 +13,12 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.ir import Dialect, Operation, SSAValue
 from xdsl.irdl import (
+    AnyAttr,
     AnyOf,
     IRDLOperation,
     VarConstraint,
     irdl_op_definition,
+    operand_def,
     result_def,
     var_operand_def,
 )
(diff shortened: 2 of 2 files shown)


tests/filecheck/dialects/varith/varith_ops.mlir
```diff
@@ -29,3 +29,7 @@
 %x6 = "varith.mul"(%ta, %tb, %tc, %td) : (tensor<10xf32>, tensor<10xf32>, tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
 // CHECK:  %x6 = varith.mul %ta, %tb, %tc, %td : tensor<10xf32>
 // CHECK-GENERIC: %x6 = "varith.mul"(%ta, %tb, %tc, %td) : (tensor<10xf32>, tensor<10xf32>, tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
+
+%x7 = "varith.select"(%ia, %fa, %fb, %fc, %fd) : (i32, f32, f32, f32, f32) -> f32
+// CHECK: %x7 = varith.select(%ia : i32) %fa, %fb, %fc, %fd
```
AntonLydike, line 34: This is missing the return type annotation, no?

xdsl/dialects/varith.py
```diff
@@ -67,10 +69,38 @@ class VarithMulOp(VarithOp):
     name = "varith.mul"
 
 
+@irdl_op_definition
+class VarithSelectOp(IRDLOperation):
+    """
+    Variadic selection operation
+    """
```
AntonLydike, line 76: Can you add a tiny bit more info here what the semantics are? Especially if the index is out of range?

alexarice: I've added to the docstring but am not certain that I am happy with the undefined behaviour

n-io (approved):
Looks good, I recommend supporting it the default conversion passes in `varith_transformations.py`

AntonLydike:
How about `varith.select(%a, %b, %c, %d)[%i: i32] default(%d)-> f32`

I would argue that default should be mandatory as well, as otherwise this is a UB bomb and I wouldn't like that in this dialect.

alexarice:
The other option is to make it look more like `cf.switch` and `scf.index_switch` and do something like
```
varith.select[%i : i32] 42 -> %a, 43 -> %b, default %c
```

AntonLydike:
> The other option is to make it look more like `cf.switch` and `scf.index_switch` and do something like
> 
> ```
> varith.select[%i : i32] 42 -> %a, 43 -> %b, default %c
> ```

I like that! Maybe we can name it `varith.switch` then?

alexarice:
Completely reworked it based on discussion

tests/filecheck/dialects/varith/varith_ops.mlir
```diff
@@ -29,3 +29,13 @@
 %x6 = "varith.mul"(%ta, %tb, %tc, %td) : (tensor<10xf32>, tensor<10xf32>, tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
 // CHECK:  %x6 = varith.mul %ta, %tb, %tc, %td : tensor<10xf32>
 // CHECK-GENERIC: %x6 = "varith.mul"(%ta, %tb, %tc, %td) : (tensor<10xf32>, tensor<10xf32>, tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
+
+%x7 = "varith.switch"(%ia, %fa, %fb, %fc, %fd) <{"case_values" = dense<[0, 1, 2]> : vector<3xi32>}> : (i32, f32, f32, f32, f32) -> f32
```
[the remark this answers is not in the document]
alexarice: I was conforming to the other tests, want me to change it?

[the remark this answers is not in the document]
superlopuh: Either way is fine with me

AntonLydike (approved):
Really cool stuff, thanks for this!

alexarice:
> Looks good, I recommend supporting it the default conversion passes in `varith_transformations.py`

Should be relatively simple to lower this operation to `arith.select`s and comparisons, but I think it'll be best to leave this for a different PR. @superlopuh are you happy for this to be merged?

superlopuh (approved):
LGTM!

PR #3343, opened 2024-10-25
transformations: (csl-stencil-bufferize) Inject accumulator in all csl-stencil linalg ops

n-io:
This is in preparation for `one-shot-bufferize` and significantly reduces the number of intermediate buffer allocations we need to keep.

### DIFF
--- a/xdsl/transforms/csl_stencil_bufferize.py
+++ b/xdsl/transforms/csl_stencil_bufferize.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from xdsl.context import MLContext
 from xdsl.dialects import arith, bufferization, func, linalg, memref, stencil, tensor
 from xdsl.dialects.builtin import (
+    AnyTensorType,
     AnyTensorTypeConstr,
     DenseArrayBase,
     DenseIntOrFPElementsAttr,
@@ -139,9 +140,6 @@ class ApplyOpBufferize(RewritePattern):
             else:
                 done_exchange_arg_mapping.append(arg)
 
-        assert isa(typ := op.receive_chunk.block.args[0].type, TensorType[Attribute])
-        chunk_type = TensorType(typ.get_element_type(), typ.get_shape()[1:])
-
         # inline blocks from old into new regions
         rewriter.inline_block(
             op.receive_chunk.block,
@@ -153,10 +151,6 @@ class ApplyOpBufferize(RewritePattern):
             op.done_exchange.block,
             InsertPoint.at_end(buf_apply_op.done_exchange.block),
             done_exchange_arg_mapping,
-        )
-
-        self._inject_iter_arg_into_linalg_outs(
-            buf_apply_op, rewriter, chunk_type, chunk_region_arg_mapping[2]
         )
 
         # insert new op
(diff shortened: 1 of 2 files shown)


xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -371,6 +290,104 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+                    break
+
+            # create `extract_slice` op if none exists
+            if not extract_slice:
+                extract_slice = tensor.ExtractSliceOp(
+                    operands=[acc_bufferization, [p_region.block.args[1]], [], []],
+                    result_types=[target_t],
+                    properties={
+                        "static_offsets": DenseArrayBase.from_list(
+                            i64, (memref.Subview.DYNAMIC_INDEX,)
```
dk949, line 362: Would the offset not be `(0,)`? Or am I misunderstanding what this does?

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -371,6 +290,104 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+            for use in acc_bufferization.tensor.uses:
+                if (
+                    isinstance(use.operation, tensor.ExtractSliceOp)
+                    and use.operation.result.type == target_t
+                ):
+                    extract_slice = use.operation
+                    break
+
+            # create `extract_slice` op if none exists
+            if not extract_slice:
```
dk949, line 356: I don't really get this part. Why do we make an `extract_slice` rather than not doing anything?

dk949 (commented):
I just don't get one part, other than that, all looks good

xdsl/transforms/csl_stencil_bufferize.py
n-io: This is in the 'chunk' region, where the offset is from the `offset` block arg. When applying a dynamic offset, it wants the above constant in all indices for which a value is supplied dynamically.

xdsl/transforms/csl_stencil_bufferize.py
n-io: This part actually existed, but the logic got refactored. When communicating in chunks, the `extract_slice` op is needed for bufferization to be able to use the accumulator as an accumulator. Even though there is an insert_slice op later, which logically is all we need, bufferization does not draw the implication that it could be using the slice as an accumulator before inserting there.

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -371,6 +290,104 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+    def match_and_rewrite(
+        self, op: linalg.NamedOpBase | linalg.Generic, rewriter: PatternRewriter, /
+    ):
+        # check if there is an output to inject the accumulator into
+        if len(op.outputs) != 1 or not isa(
+            target_t := op.outputs[0].type, AnyTensorType
+        ):
+            return
+
+        # find parent `csl_stencil.apply` and which of the regions `op` is in
```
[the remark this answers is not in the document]
n-io: I usually like this (see `_get_module_wrapper` in `csl_stencil_to_csl_wrapper.py`), but its fetching two things at once, but easy enough to do.

dk949 (approved):

PR #3344, opened 2024-10-25
transformations: (lift-arith-to-linalg) Add generic FMA

n-io:
Adds the option to lift arith to generic fused multiply-add ops, with various flags to control behaviour.


superlopuh (commented):
This feels like it's doing two things at the same time, and breaking the composability of rewrites. Could the ops be fused by a separate pass?

n-io:
The behaviour can be switched on and off by a flag, but doing this in a separate pass is (mildly) more complicated, as the regions contain `arith.addf` and `arith.mulf` which the pass as-is would attempt to translate.

AntonLydike:
I agree with Sasha here, lifting arith to linalg is one thing, but fusing operations together is a different thing entirely

n-io:
Closing in favour of doing this in a separate pass.

PR #3346, opened 2024-10-25
CI: test 3.13 on the CI

superlopuh:
`uv venv` picks the newest Python available, and 3.13 seems to be failing for me locally. This is a problem with an easy workaround (pick an older Python), but I'd rather we just worked out of the box. I'll look into how to actually fix it in the next few days, help would be appreciated.

### DIFF
--- a/.github/workflows/ci-mlir.yml
+++ b/.github/workflows/ci-mlir.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-core.yml
superlopuh, line 1: For now only in core, because ONNX doesn't support 3.13

superlopuh: Yep it was to unblock this

PR #3347, opened 2024-10-26
transformations: New linalg-fuse-multiply-add pass

n-io:
Introduces a new pass to fuse `linalg.mul` and `linalg.add` into `linalg.generic` with the functionality of an FMA op.

### DIFF
--- a/xdsl/transforms/linalg_transformations.py
+++ b/xdsl/transforms/linalg_transformations.py
@@ -1,0 +1,128 @@
+from dataclasses import dataclass
+
+from xdsl.builder import Builder
+from xdsl.context import MLContext
+from xdsl.dialects import arith, linalg
+from xdsl.dialects.builtin import AffineMapAttr, DenseIntOrFPElementsAttr, ModuleOp
+from xdsl.ir import BlockArgument, OpResult, SSAValue
+from xdsl.ir.affine import AffineMap
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+def build_generic_fma(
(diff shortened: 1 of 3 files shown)


n-io:
This is a replacement for #3344 @superlopuh @AntonLydike

xdsl/transforms/linalg_transformations.py
```diff
@@ -0,0 +1,128 @@
...
+            add_operand = (
+                add.inputs[0] if mul.res[0] == add.inputs[1] else add.inputs[1]
+            )
+
+            # # prefer dest that is not used as an input
+            # dest = (
+            #     mul.outputs[0]
+            #     if mul.outputs[0] not in [*mul.inputs, *add.inputs]
+            #     else add.outputs[0]
+            # )
```
superlopuh, line 80: commented code

xdsl/transforms/linalg_transformations.py
```diff
@@ -0,0 +1,128 @@
...
+        inputs,
+        outputs,
+        body,
+        4 * [AffineMapAttr(AffineMap.from_callable(lambda i,: (i,)))],
+        [linalg.IteratorTypeAttr.parallel()],
+        [out.type],
+    )
+
+
+@dataclass
```
superlopuh, line 42: nit:
```suggestion
@dataclass(frozen=True)
```

superlopuh (approved):
The code looks good, although my understanding is that linalg is not designed to be used in this way, and instead fuses after tiling. I wonder if there's a similar direction for you to look into, maybe for later.

xdsl/transforms/linalg_transformations.py
n-io: Done as suggested.

xdsl/transforms/linalg_transformations.py
n-io: Whoops, removed.

n-io:
It made sense in my mind to do this sort of op fusion before bufferization, though maybe it doesn't make a difference?

PR #3350, opened 2024-10-28
[WIP]: Introducing ptr dialect and its riscv conversion

mamanain:
Porting the `ptr` dialect proposed [here](https://discourse.llvm.org/t/rfc-ptr-dialect-modularizing-ptr-ops-in-the-llvm-dialect/75142/21). Transfering the riscv compilation to it for `load` and `store` operations


xdsl/dialects/ptr.py
```diff
@@ -0,0 +1,70 @@
...
+    name = "ptr.ptradd"
+
+    addr = operand_def(PtrType)
+    offset = operand_def(AnyOf([IntegerType, IndexType]))
+    result = result_def(PtrType)
+
+    assembly_format = "$addr `,` $offset attr-dict `:` `(` type($addr) `,` type($offset) `)` `->` type($result)"
+
+
+# haven't managed to pass a type here yet. so did it with a hack.
```
superlopuh, line 30: The way that should work but is a bit inelegant is `base(IntegerType) | base(IndexType)`, there might be `IntegerTypeConstr` already defined, if not `BaseAttr(IntegerType)` should also work

mamanain: I got it to work but only via passing the type through attributes. Is it possible to pass it as an operand? Xdsl complains that it expects an operand there if write `"op"(i32)`

xdsl/dialects/ptr.py
```diff
@@ -34,7 +35,7 @@ class PtrAddOp(IRDLOperation):
 class TypeOffsetOp(IRDLOperation):
     name = "ptr.type_offset"
 
-    elem_type = prop_def(base(IntegerType))
+    elem_type = prop_def(base(Attribute))
```
superlopuh, line 38:
```suggestion
    elem_type = prop_def()
```

mamanain:
Notebooks fail because I deleted Store and Load conversions from the MemrefToRiscv pass. Should we keep the old operation there or make the pass through pointers the default?

superlopuh:
Should we close this PR now that the others are merged?

mamanain:
> Should we close this PR now that the others are merged?

Technically, not all of it. There is still the loop folding part and the integration into the main compilation pipeline. 
I would keep it just as a beacon in the final steps.
But if it's better to close it we can, I have all of this stashed.

superlopuh:
If you have the motivation to keep working on this it's welcome! In that case, could you please resolve the merge conflicts to make the remaining steps clearer?

mamanain:
Yeah I surely want to carry this to the end. Sorry for the delay, got distracted. 

I think I’ll jump back to this next week.

tests/filecheck/backend/riscv/memref_to_riscv.mlir
superlopuh, line 1: let's revert these changes, we still want to make sure that the old flow works

tests/filecheck/backend/riscv/memref_to_riscv_opt.mlir
superlopuh, line 1: I would revert these changes as well

tests/filecheck/transforms/ptr_loop_folding.mlir
superlopuh, line 1: My first reaction to this file is that the file name doesn't correspond to a transform, which is confusing. I'm not sure that we really need it, I think we can avoid it for now.

superlopuh (commented):
Are there any changes to the matmul operation in the linalg-snitch notebook?

superlopuh:
Let's iterate on this when you have more time

PR #3353, opened 2024-10-29
fix: (csl) Use `u32` as the counter in async loops

dk949:
Previously, `i16` was used, this significantly limited the maximum loop iterations.


n-io (commented):
Looks good, but do we need that many iterations? What is the max number of iterations we intend to run, and could this be modelled with a `u16`?

dk949:
> Looks good, but do we need that many iterations? What is the max number of iterations we intend to run, and could this be modelled with a `u16`?

The runtime I get when running for ~30000 iterations is extremely low, on the order of 10ms, having the ability to bump it higher would let us do longer runs.

Is there any benefit to `u16` vs `u32` that I'm missing?

n-io:
Would you expect there to be any performance implications?

dk949:
Not from reading the docs, no.

PR #3354, opened 2024-10-29
transformations: application of PDL patterns from source

qaco:
This PR unlocks the application of PDL patterns present in the source code (either in the same file as the code to be transformed, or in another file). It also corrects a few bugs in the interoperation of PDL operations.

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -17,7 +17,16 @@ from xdsl.dialects.builtin import (
     UnrankedTensorType,
     f64,
 )
-from xdsl.ir import Attribute, Block, Dialect, Operation, OpResult, Region, SSAValue
+from xdsl.ir import (
+    Attribute,
+    Block,
+    Dialect,
+    Operation,
+    OpResult,
+    OpTraits,
+    Region,
+    SSAValue,
+)
 from xdsl.irdl import (
     IRDLOperation,
     attr_def,
@@ -77,7 +86,7 @@ class ConstantOp(IRDLOperation):
     value = attr_def(DenseIntOrFPElementsAttr)
     res = result_def(TensorTypeF64)
 
-    traits = frozenset((Pure(),))
+    traits = OpTraits.get(Pure())
 
     def __init__(self, value: DenseIntOrFPElementsAttr):
         super().__init__(result_types=[value.type], attributes={"value": value})
@@ -138,7 +147,7 @@ class AddOp(IRDLOperation):
     rhs = operand_def(AnyTensorTypeF64Constr)
     res = result_def(AnyTensorTypeF64Constr)
 
-    traits = frozenset((Pure(), InferAddOpShapeTrait()))
+    traits = OpTraits.get(Pure(), InferAddOpShapeTrait())
 
     def __init__(self, lhs: SSAValue, rhs: SSAValue):
         if isa(lhs.type, TensorTypeF64):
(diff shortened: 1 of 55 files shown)


xdsl/interpreters/experimental/pdl.py
superlopuh, line 1: Could you please add these in a separate PR?

xdsl/transforms/apply_pdl.py
```diff
@@ -0,0 +1,67 @@
...
+from xdsl.parser import Parser
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import PatternRewriter
+
+
+@dataclass(frozen=True)
+class ApplyPDLPass(ModulePass):
+    name = "apply-pdl"
+
+    pdl_file: str | None = None
```
superlopuh, line 21: Could you please add a test that specifies the file explicitly, as well?

xdsl/utils/parse_pipeline.py
```diff
@@ -26,6 +26,7 @@ class Kind(Enum):
         STRING_LIT = object()
         MLIR_PIPELINE = object()
         COMMA = ","
+        DOT = "."
```
superlopuh, line 29: is this still necessary with the quotes fix?

xdsl/transforms/apply_pdl.py
qaco: Ok I put the extra file in the filecheck directory as well ?

superlopuh: Yes, please!

xdsl/utils/parse_pipeline.py
qaco: Ok done

tests/filecheck/interpreters/pdl_definition.mlir
```diff
@@ -0,0 +1,16 @@
+pdl.pattern : benefit(1) {
```
math-fehr, line 1: Can you add a comment here to say that this is used in the other test?
Just to not forget that this file is actually used.

tests/filecheck/interpreters/pdl_definition.mlir
```diff
@@ -0,0 +1,16 @@
...
+    %root_addi = pdl.operation "arith.addi" (%x,%y: !pdl.value, !pdl.value)  -> (%in_type: !pdl.type)
+    %a = pdl.result 0 of %root_addi
+    %root_subi = pdl.operation "arith.subi" (%a,%y: !pdl.value, !pdl.value)  -> (%in_type: !pdl.type)
+    pdl.rewrite %root_subi {
+      %zero_attr = pdl.attribute = 0
+      %zero_op = pdl.operation "arith.constant" {"value" = %zero_attr} -> (%in_type: !pdl.type)
+      %zero = pdl.result 0 of %zero_op
+      %new_addi = pdl.operation "arith.addi" (%x,%zero: !pdl.value, !pdl.value)  -> (%in_type: !pdl.type)
+      pdl.replace %root_subi with %new_addi
+      pdl.erase %root_addi
```
math-fehr, line 14: Big nit, but if you want to replace `x + y - y` with `x` (instead of `x + 0`), you can just write `pdl.replace %root_subi with (%x)`.
That would make this example more simple.

xdsl/transforms/apply_pdl.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        payload_module = op
+        # Target the file containing the PDL specification
+        if self.pdl_file:
+            assert os.path.exists(self.pdl_file)
```
math-fehr, line 28: Can you raise an exception here with an error message instead?

math-fehr (approved):
Nice! I left some small comments but otherwise that looks good to me!

superlopuh:
@math-fehr FYI there's a second PR open with a simpler test for the pass itself #3359

PR #3355, opened 2024-10-29
misc: add a ScopedDict util, similar to LLVM's ScopedHashTable

superlopuh:


xdsl/utils/scoped_dict.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import annotations
+
+from typing import Generic, TypeVar
+
+_Key = TypeVar("_Key")
```
math-fehr, line 5: Should this be hashable or something? I'm not sure why pyright doesn't complain here

xdsl/utils/scoped_dict.py
```diff
@@ -0,0 +1,62 @@
...
+    def __getitem__(self, key: _Key) -> _Value:
+        """
+        Fetch key from environment. Attempts to first fetch from current scope,
+        then from parent scopes. Raises KeyError error if not found.
+        """
+        local = self._local_scope.get(key)
+        if local is not None:
+            return local
+        if self.parent is None:
+            raise KeyError()
```
math-fehr, line 45: Should this raise a better error message?

xdsl/utils/scoped_dict.py
```diff
@@ -0,0 +1,62 @@
...
+        return self.parent[key]
+
+    def __setitem__(self, key: _Key, value: _Value):
+        """
+        Assign key to current scope. Raises InterpretationError if key already
+        assigned to.
+        """
+        if key in self._local_scope:
+            raise ValueError(
```
math-fehr, line 54: Should this be a `KeyError`?

math-fehr (approved):
👍

xdsl/utils/scoped_dict.py
superlopuh: I considered it, but I have a feeling that this is not the intended purpose of KeyError, which to me signals that something does not exist for a given key, whereas this error is raised when something does already exist. In either case, I'm not sure what the user can do to recover.

xdsl/utils/scoped_dict.py
superlopuh: Hashable isn't a real type, AFAIK, and the `dict` generic type has the same `TypeVar` definition

PR #3356, opened 2024-10-30
dialects: (llvm) Registering llvm.icmp

lfrenot:


tests/filecheck/dialects/llvm/arithmetic.mlir
```diff
@@ -70,3 +70,6 @@
 
 %cst32 = llvm.mlir.constant(25 : i32) : i32
 // CHECK: %cst32 = llvm.mlir.constant(25 : i32) : i32
+
+%icmp_eq = llvm.icmp "eq" %arg0, %arg1 : i32
+// CHECK: %icmp_eq = llvm.icmp "eq" %arg0, %arg1 : i32
```
tobiasgrosser, line 75: I guess you want examples for all predicate flags.

tobiasgrosser (approved):
Otherwise, this looks great to me. Maybe @superlopuh and @luisacicolini have comments?

lfrenot:
I'm getting this error
```
# |     raise ParseError(at_position, msg)
# | xdsl.utils.exceptions.ParseError: stdin:26:72
# |   %icmp_eq = "llvm.icmp"(%arg0, %arg1) <{"predicate" = #llvm.predicateeq}> : (i32, i32) -> i32
# |                                                                         ^
# |                                                                         'llvm.predicateeq' is not registered
```
Since the predicate is supposed to be an i64 (according to `mlir-opt --mlir-print-op-generic`), I guess I have to change the way I'm representing it right now

xdsl/dialects/llvm.py
```diff
@@ -610,6 +610,106 @@ def verify(self, verify_nested_ops: bool = True):
...
+        elif i == 5:
+            return ICmpPredicateFlag.SGT
+        elif i == 6:
+            return ICmpPredicateFlag.ULT
+        elif i == 7:
+            return ICmpPredicateFlag.ULE
+        elif i == 8:
+            return ICmpPredicateFlag.UGE
+        elif i == 9:
+            return ICmpPredicateFlag.UGT
```
superlopuh, line 646: There has to be a better way of doing this, I'd expect some helper to exist in the Python enum implementation, otherwise we should cache this mapping.

superlopuh:
This PR adds an operation and an attribute, and only registers the new op in the dialect.

lfrenot:
```
# | ValueError: 0 is not a valid ICmpPredicateFlag
```
I don't get why I'm getting this error with the current def of `ICmpPredicateFlag`

lfrenot:
@superlopuh
Does it look good for you? 
Do you know how to properly do the enum thing?

xdsl/dialects/llvm.py
```diff
@@ -610,6 +610,106 @@ def verify(self, verify_nested_ops: bool = True):
...
+
+@irdl_op_definition
+class ICmpOp(IRDLOperation, ABC):
+    name = "llvm.icmp"
+    T: ClassVar = VarConstraint("T", BaseAttr(IntegerType))
+
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+    res = result_def(T)
+    predicate = prop_def(Attribute)
```
superlopuh, line 658: why this change? it was better before as a strenum, no?

superlopuh:
I would recommend reverting your last change, and looking into how to index the enum cases

superlopuh:
```
from enum import StrEnum


class MyEnum(StrEnum):
    a = "a"
    b = "b"


MY_ENUM_CASES = list(MyEnum)

print(MY_ENUM_CASES[1])

```

This seems to work

superlopuh:
and is not O(number of cases)

xdsl/dialects/llvm.py
```diff
@@ -610,6 +610,83 @@ def verify(self, verify_nested_ops: bool = True):
...
+
+
+@irdl_attr_definition
+class ICmpPredicateAttr(EnumAttribute[ICmpPredicateFlag]):
+    name = "llvm.predicate"
+    TUPLE = tuple(ICmpPredicateFlag)
+
+    @classmethod
+    def from_int(cls, i: int):
+        return ICmpPredicateAttr(cls.TUPLE[i])
```
superlopuh, line 633:
```suggestion
    ALL_PREDICATES = tuple(ICmpPredicateFlag)

    @classmethod
    def from_int(cls, i: int):
        return ICmpPredicateAttr(cls.ALL_PREDICATES[i])
```

lfrenot:
@superlopuh do you have any idea on how to fix the current error?

superlopuh:
Ah no, it looks like I was too quick to approve. I would recommend marking the PR as draft if the tests don't pass, and reaching out on Zulip to iterate

superlopuh:
I've actually never used this part of the API before, I'm happy to take a look in a bit if you're stuck, but it would be worth messaging on Zulip in case someone with more experience with your issue knows the answer directly. In this case pinging @alexarice would probably be a good idea, I think he's dealt with these attributes somewhat recently.

xdsl/dialects/llvm.py
```diff
@@ -610,6 +610,83 @@ def verify(self, verify_nested_ops: bool = True):
...
+    SGT = "sgt"
+    SGE = "sge"
+    ULT = "ult"
+    ULE = "ule"
+    UGT = "ugt"
+    UGE = "uge"
+
+
+@irdl_attr_definition
+class ICmpPredicateAttr(EnumAttribute[ICmpPredicateFlag]):
```
alexarice, line 627: I believe you want `OpaqueSyntaxAttribute` or `SpacedOpaqueSyntaxAttribute` here to get it to print properly

xdsl/dialects/llvm.py
```diff
@@ -610,6 +610,83 @@ def verify(self, verify_nested_ops: bool = True):
...
+    name = "llvm.predicate"
+    ALL_PREDICATES = tuple(ICmpPredicateFlag)
+
+    @classmethod
+    def from_int(cls, i: int):
+        return ICmpPredicateAttr(cls.ALL_PREDICATES[i])
+
+
+@irdl_op_definition
+class ICmpOp(IRDLOperation, ABC):
```
alexarice, line 637: Why is this marked as an abstract class?

xdsl/dialects/llvm.py
```diff
@@ -610,6 +610,83 @@ def verify(self, verify_nested_ops: bool = True):
...
+        parser.parse_characters(":")
+        type = parser.parse_type()
+        operands = parser.resolve_operands([lhs, rhs], [type, type], parser.pos)
+        return cls(operands[0], operands[1], predicate, attributes)
+
+    def print_predicate(self, printer: Printer):
+        self.predicate.print_parameter(printer)
+
+    def print(self, printer: Printer):
+        printer.print(' "')
```
alexarice, line 682: I think in general, print_string is preferred to print but this is a super nit

lfrenot:
@alexarice Thank you for the suggestions, that worked

@superlopuh are we good to merge then?

alexarice (commented):
I just tested the mlir version and it does this:
```
bin/mlir-opt ../../test.mlir --mlir-print-op-generic
"builtin.module"() ({
  %0:2 = "test.op"() : () -> (i32, i32)
  %1 = "llvm.icmp"(%0#0, %0#1) <{predicate = 0 : i64}> : (i32, i32) -> i1
}) : () -> ()
```
I feel the generic output of your function is not the same. Unsure if this is a blocker but I just wanted to point it out.

superlopuh (approved):
LGTM! The issue Alex pointed out is worth looking into, but I don't want to block on that necessarily. It would be great if you could take a quick look at the generic syntax before merging, and comment here.

tobiasgrosser:
Let's merge this for now. @lfrenot, can you still check @alexarice's comment.

tobiasgrosser:
> I just tested the mlir version and it does this:
> 
> ```
> bin/mlir-opt ../../test.mlir --mlir-print-op-generic
> "builtin.module"() ({
>   %0:2 = "test.op"() : () -> (i32, i32)
>   %1 = "llvm.icmp"(%0#0, %0#1) <{predicate = 0 : i64}> : (i32, i32) -> i1
> }) : () -> ()
> ```
> 
> I feel the generic output of your function is not the same. Unsure if this is a blocker but I just wanted to point it out.

@lfrenot, we should indeed be consistent with MLIR here. I just tried to use your icmp implementation to parse our test cases, but it failed as MLIR emits an integer attribute and you added your own attribute type. It would be great if this could be fixed.

PR #3358, opened 2024-10-30
dialects: (pdl) Filecheck test documenting how to match an MLIR operation without return type

qaco:
Filecheck test documenting how to match an MLIR operation without return type


superlopuh:
This looks good but I'm not sure what the purpose is. Is it documentation, or to make sure that we don't break this syntax in the future?

qaco:
> This looks good but I'm not sure what the purpose is. Is it documentation, or to make sure that we don't break this syntax in the future?

Both. Since the explicit syntax with ```-> ()``` is also intuitive but not yet implemented, this test shows - and guarantees - that there is a correct way to program this behavior. This saves the person who needs it from having to hack the parser for nothing.

superlopuh (approved):
Sounds good, thank you!

PR #3359, opened 2024-10-30
transformations: application of PDL patterns from source

qaco:
transformations: application of PDL patterns from source

### DIFF
--- a/tests/filecheck/interpreters/extra_file.mlir
+++ b/tests/filecheck/interpreters/extra_file.mlir
@@ -1,0 +1,9 @@
+pdl.pattern : benefit(1) {
+    %zero_attr = pdl.attribute = 0
+    %root = pdl.operation "test.op" {"attr" = %zero_attr}
+    pdl.rewrite %root {
+      %one_attr = pdl.attribute = 1
+      %new_op = pdl.operation "test.op" {"attr" = %one_attr}
+      pdl.replace %root with %new_op
+    }
+}
--- a/tests/filecheck/interpreters/pdl_definition.mlir
+++ b/tests/filecheck/interpreters/pdl_definition.mlir
@@ -1,0 +1,9 @@
+pdl.pattern : benefit(1) {
+    %zero_attr = pdl.attribute = 0
+    %root = pdl.operation "test.op" {"attr" = %zero_attr}
+    pdl.rewrite %root {
+      %one_attr = pdl.attribute = 1
+      %new_op = pdl.operation "test.op" {"attr" = %one_attr}
+      pdl.replace %root with %new_op
+    }
+}
--- a/tests/filecheck/interpreters/test_pdl_different_files.mlir
+++ b/tests/filecheck/interpreters/test_pdl_different_files.mlir
@@ -1,0 +1,7 @@
+// RUN: xdsl-opt %s -p 'apply-pdl{pdl_file="%p/pdl_definition.mlir"}' | filecheck %s
+
+"test.op"() {attr = 0} : () -> ()
+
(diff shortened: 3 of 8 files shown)


superlopuh (commented):
Why not reuse the logic in `PDLRewritePattern`?

qaco:
> Why not reuse the logic in `PDLRewritePattern`?

In some cases, such as in the test I provide, you need to match several operations with the same ```PDLMatcher``` before rewrite. For the moment, however, ```PDLRewritePattern``` only matches a single operation.

qaco:
> Oh, interesting. I'm not sure that I understand the exact limitation, but I feel that this might be worth putting somewhere closer to the rewrite pattern anyway? The thing I was suggesting initially was to just creating `PDLRewritePatterns` in a loop, and stopping when nothing matched in the loop at all. This should already work, and if it doesn't then wemight want to fix it in the interpreter implementation file, as opposed to this helper that should ideally be relatively minimal.

It may happen that the matching of an op2 operation depends on the matching of a previous op1 operation, for example when the result of op1 must be an argument of op2.

I understand the point. I thought that this implementation of ```PDLRewritePattern``` was required for some specialized usage. So, this change can be done in two ways:

 requires to feed the ```PDLRewritePattern``` with pairs (patternop,containerop) instead of (rewriteop,innerop). Are you ok with this change ?

superlopuh:
Hmm, but the matching already handles this? It matches starting from the root op, and can go up an arbitrary number of operations.

qaco:
> Hmm, but the matching already handles this? It matches starting from the root op, and can go up an arbitrary number of operations.

I must be missing the point. I have no change with the following:
```
for pdl_op in pattern.walk():
    if isinstance(pdl_op, pdl.RewriteOp):
        stream = StringIO()
        pattern_rewriter = PatternRewriter(payload_module)
        pdl_rewrite_pattern = PDLRewritePattern(pdl_op,ctx,stream)
        pdl_rewrite_pattern.match_and_rewrite(payload_module,pattern_rewriter)
```

I guess I misunderstand how ```PDLRewritePattern``` works. I'll dig tomorrow.

qaco:
> > Hmm, but the matching already handles this? It matches starting from the root op, and can go up an arbitrary number of operations.
> 
> I must be missing the point. I have no change with the following:
> 
> ```
> for pdl_op in pattern.walk():
>     if isinstance(pdl_op, pdl.RewriteOp):
>         stream = StringIO()
>         pattern_rewriter = PatternRewriter(payload_module)
>         pdl_rewrite_pattern = PDLRewritePattern(pdl_op,ctx,stream)
>         pdl_rewrite_pattern.match_and_rewrite(payload_module,pattern_rewriter)
> ```
> 
> I guess I misunderstand how `PDLRewritePattern` works. I'll dig tomorrow.



> Hmm, but the matching already handles this? It matches starting from the root op, and can go up an arbitrary number of operations.

It is corrected now. Thank you for your careful advices.

superlopuh (commented):
Nice! Although I think it might be possible to get the same result via a greedy pattern applier, and without modifying the existing pdl pattern. The way I see this working is by first getting all the pdl pattern ops, creating an xDSL pattern for each op, and then putting them in a greedy pattern, like a normal pass would. Could you please try this? Hopefully should also work

qaco:
> Nice! Although I think it might be possible to get the same result via a greedy pattern applier, and without modifying the existing pdl pattern. The way I see this working is by first getting all the pdl pattern ops, creating an xDSL pattern for each op, and then putting them in a greedy pattern, like a normal pass would. Could you please try this? Hopefully should also work

It's done and it works !

PR #3360, opened 2024-10-30
transformations: ```pdl.type```, ```pdl.erase``` and multi-operations ```pdl.replace``` are now supported

qaco:
transformations: ```pdl.type```, ```pdl.erase``` and multi-operations ```pdl.replace``` are now supported


superlopuh (commented):
Could you please test this with unit tests instead? I tried to write the other interpreter unit tests in a way that would be as targeted as possible, please let me know if you'd like some help with this. Currently this PR also includes the pass, which is orthogonal

qaco:
> Could you please test this with unit tests instead? I tried to write the other interpreter unit tests in a way that would be as targeted as possible, please let me know if you'd like some help with this. Currently this PR also includes the pass, which is orthogonal

Ok !

PR #3361, opened 2024-10-30
misc: fix tests that secretly rely on optional dependencies

superlopuh:
Also changes CI-Core to test only the core dev dependencies, letting CI-MLIR check the full coverage


tests/interactive/test_app.py
```diff
@@ -27,6 +27,15 @@
...
 from xdsl.utils.parse_pipeline import PipelinePassSpec, parse_pipeline
 
+# Mock pyclip_copy for testing
+_copied_text: str | None = None
+
+
+def pyclip_copy(text: str) -> None:
+    """Mock version of pyclip_copy that stores copied text for inspection during tests"""
+    global _copied_text
+    _copied_text = text
```
math-fehr, line 37: Why was that necessary?

math-fehr (approved):
👍

tests/interactive/test_app.py
superlopuh: Because pyclip is also an optional dependency, and it would be nice to actually test what the button copies, even if we don't right now. I could add this quickly

superlopuh: I actually couldn't get this to work, and added a different API instead.

PR #3362, opened 2024-10-30
transformations: Canonicalizations for arith on float constants

n-io:
This adds the first of two canonicalisation patterns `FloatingPointLikeBinaryOperation`s on `arith.constants`.

* Ops with two constant operands are folded by canonicalisation, regardless of math flags
* ~~Chained ops of the form `(const1 * var) * const2` are rewritten as `(const1 * const2) * var` iff the `fastmath<reassoc>` or `fastmath<fast>` flags are set, and if the ops are multiplications or additions. Note, this is not currently implemented in upstream mlir.~~

See [here](https://godbolt.org/z/1KKf6vbfv) to check how it works in upstream mlir.

Note, these canonicalisations currently do not support container types.

### DIFF
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -15,3 +15,22 @@
 // CHECK-NEXT:   %addf = arith.addf %lhsf32, %rhsf32 : f32
 // CHECK-NEXT:   %addf_vector = arith.addf %lhsvec, %rhsvec : vector<4xf32>
 // CHECK-NEXT:   "test.op"(%addf, %addf_vector) : (f32, vector<4xf32>) -> ()
+
+func.func @test_const_const() {
+    %a = arith.constant 2.9979 : f32
+    %b = arith.constant 3.1415 : f32
+    %1 = arith.addf %a, %b  : f32
+    %2 = arith.subf %a, %b  : f32
+    %3 = arith.mulf %a, %b  : f32
+    %4 = arith.divf %a, %b  : f32
+    "test.op"(%1, %2, %3, %4) : (f32, f32, f32, f32) -> ()
+
+    return
+
+    // CHECK-LABEL: @test_const_const
+    // CHECK-NEXT:   %0 = arith.constant 6.139400e+00 : f32
+    // CHECK-NEXT:   %1 = arith.constant -1.436000e-01 : f32
+    // CHECK-NEXT:   %2 = arith.constant 9.417903e+00 : f32
+    // CHECK-NEXT:   %3 = arith.constant 9.542894e-01 : f32
+    // CHECK-NEXT:   "test.op"(%0, %1, %2, %3) : (f32, f32, f32, f32) -> ()
+}
--- a/xdsl/transforms/canonicalization_patterns/arith.py
+++ b/xdsl/transforms/canonicalization_patterns/arith.py
@@ -38,11 +38,11 @@ class FoldConstConstOp(RewritePattern):
                 case arith.Mulf:
                     val = l.value.data * r.value.data
                 case arith.Divf:
-                    if rhs.value.data == 0.0:
+                    if r.value.data == 0.0:
                         # this mirrors what mlir does
                         val = float("inf")
                     else:
-                        val = lhs.value.data / rhs.value.data
+                        val = l.value.data / r.value.data
                 case _:
                     return
             rewriter.replace_matched_op(arith.Constant(builtin.FloatAttr(val, l.type)))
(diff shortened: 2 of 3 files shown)


superlopuh (approved):
This looks good to me! I'm happy to see that the code doesn't match the description, and that it only has the first of two patterns.

n-io:
Ok, happy to have them in separate PRs!

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -16,3 +17,28 @@ def match_and_rewrite(self, op: arith.Addi, rewriter: PatternRewriter) -> None:
...
+        ):
+            match type(op):
+                case arith.Addf:
+                    val = l.value.data + r.value.data
+                case arith.Subf:
+                    val = l.value.data - r.value.data
+                case arith.Mulf:
+                    val = l.value.data * r.value.data
+                case arith.Divf:
+                    val = l.value.data / r.value.data
```
math-fehr, line 41: What happens if `val` is `NaN`?
We should probably check for it and only do the replacement if `val` is not `NaN`, as `arith.constant` doesn't support `NaN` afaik?

math-fehr (approved):
Besides one issue with `NaN`, looks good to me!

xdsl/transforms/canonicalization_patterns/arith.py
n-io: Excellent point. Theoretically, python should handle it. For instance, `3 / float('nan')` will evaluate to `nan`.

However, I can't seem to get nan to round-trip to verify this. `str(builtin.FloatAttr(float('nan'), 32))` returns `'nan : f32'`, but `arith.constant nan : f32` will throw a parser error saying:
```
    %a = arith.constant nan : f32
                        ^^^
                        attribute expected
```
Am I doing this right?

tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -15,3 +15,22 @@
 // CHECK-NEXT:   %addf = arith.addf %lhsf32, %rhsf32 : f32
 // CHECK-NEXT:   %addf_vector = arith.addf %lhsvec, %rhsvec : vector<4xf32>
 // CHECK-NEXT:   "test.op"(%addf, %addf_vector) : (f32, vector<4xf32>) -> ()
+
```
alexarice, line 18: Nit: I believe we prefer to put filecheck tests above the test now? If so this file is small enough to fix imo

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -16,3 +17,28 @@ def match_and_rewrite(self, op: arith.Addi, rewriter: PatternRewriter) -> None:
...
+
+class FoldConstConstOp(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: arith.FloatingPointLikeBinaryOperation, rewriter: PatternRewriter, /
+    ):
+        if (
+            isinstance(op.lhs.owner, arith.Constant)
+            and isinstance(op.rhs.owner, arith.Constant)
+            and isa(l := op.lhs.owner.value, builtin.AnyFloatAttr)
```
alexarice, line 30: Might be easier to use `isinstance(l, FloatAttr)` here but doesn't make a lot of difference.

xdsl/transforms/canonicalization_patterns/arith.py
alexarice, line 41: This will throw an exception if `r.value.data == 0` which is probably not desirable

xdsl/transforms/canonicalization_patterns/arith.py
n-io: Unfortunately, this will cause pyright to say that the type is partially unknown when using `l`, though I agree that isinstance is more lightweight and preferable.

xdsl/transforms/canonicalization_patterns/arith.py
n-io: I just checked how upstream does it, and it appears they output `inf`

alexarice: That sounds reasonable to me. Python crashes by default so will have to manually check/catch

superlopuh: can mlir parse that? it's possible that xDSL has a bug here

superlopuh: Do you have the bandwidth to fix xDSL printing of nans in that case?

n-io: It's not straightforward, xdsl likely parses hexadecimal float literals as hexadecimal int literals, from which it builds a float rather than using it as a bit-string. See the [floatattr](https://mlir.llvm.org/docs/Dialects/Builtin/#floatattr) docs

PR #3363, opened 2024-10-30
API: use OpTraits to define traits on operations

superlopuh:
Using `@property` + `@classmethod` is deprecated since Python 3.11, and removed in 3.13. The only place wehere we rely on this working is in the traits mechanism, that does a bit of spooky aliasing of the traits class variable to the OpDef traits property. This makes the following changes:
 - changes the API of operation definitions to use OpTraits directly (in line with how the other ClassVars are set on operations)
 - unifies the lazy/eager traits definitions (always just OpTraits)
 - aliases the ClassVar and traits property so that if one is mutated the other is also
 - adds a test for adding a trait to an operation


alexarice (commented):
Haven't gone through the whole thing but is there a reason you don't create a `trait_def` function to make it fit the look of other fields? (Perhaps we do not want to make it look like other fields.

superlopuh:
So I did try with that first, but it got a little weird. I quickly gave up on making it not weird because it's a ClassVar, and not a field, so it's probably better off without the `_def` pattern, to make it distinct.

xdsl/ir/core.py
```diff
@@ -708,6 +708,33 @@ def __hash__(self):
...
+    def __init__(self, traits: set[OpTrait] | Callable[[], set[OpTrait]]) -> None:
+        self._traits = traits
+
+    def get_traits(self) -> set[OpTrait]:
+        """Returns a copy of this instance's traits."""
+        if not isinstance(self._traits, set):
+            self._traits = self._traits()
+        return set(self._traits)
+
+    def add_trait(self, trait: OpTrait):
```
alexarice, line 723: Is this dead code?

alexarice (approved):

xdsl/ir/core.py
superlopuh: Tested here, but may well be dead otherwise...
```
def test_modify_traits():
    op = TestModifyTraitsOp()

    op.verify()

    TestModifyTraitsOp.traits.add_trait(AlwaysFailsTrait())

    with pytest.raises(VerifyException, match="Nope"):
        op.verify()
```

alexarice: I would vote to get rid of it, as I see no reason for operation definitions to be mutable, but will leave it up to you.

tests/test_traits.py
```diff
@@ -546,3 +547,24 @@ class SupeculatabilityTestOp(IRDLOperation):
...
+    def verify(self, op: Operation) -> None:
+        raise VerifyException("Nope")
+
+
+def test_modify_traits():
+    op = TestModifyTraitsOp()
+
+    op.verify()
+
+    TestModifyTraitsOp.traits.add_trait(AlwaysFailsTrait())
```
math-fehr, line 567: Nice

xdsl/ir/core.py
math-fehr: I would actually like to keep it, and I have some use cases for it.
If you are a downstream user, and you want to add an interface to an upstream dialect (let say, adding a semantic interface to an existing operation), then you need to actually modify the trait interfaces.

xdsl/ir/core.py
```diff
@@ -708,6 +708,33 @@ def __hash__(self):
         return hash(self._op._operands)  # pyright: ignore[reportPrivateUsage]
 
 
+class OpTraits(Iterable[OpTrait]):
```
math-fehr, line 711: Can you add one bit of documentation on the class?

xdsl/ir/core.py
```diff
@@ -708,6 +708,33 @@ def __hash__(self):
...
+    _traits: set[OpTrait] | Callable[[], set[OpTrait]]
```
math-fehr, line 712: Can we make this a dataclass? It would still not allow you to access the member right?

xdsl/ir/core.py
```diff
@@ -708,6 +708,33 @@ def __hash__(self):
...
+
+    def __init__(self, traits: set[OpTrait] | Callable[[], set[OpTrait]]) -> None:
+        self._traits = traits
+
+    def get_traits(self) -> set[OpTrait]:
```
math-fehr, line 717: Wouldn't having `@property` and `def traits` being a bit more pythonic here?

xdsl/irdl/operations.py
```diff
@@ -2,7 +2,7 @@
```
math-fehr, line 2: Nice cleanup

math-fehr (approved):
Nice!

xdsl/ir/core.py
superlopuh: the more I look at traces, the more I think that dataclasses are not really pulling their weight most of the time, in terms of being quite slow.

xdsl/ir/core.py
superlopuh: Yep, will do this now.

superlopuh: I didn't want to before but I've changed my mind. I think it might make sense to make it a frozenset also, since mutation will be so infrequent.

xdsl/ir/core.py
math-fehr: But that's only on core data structures, right?
Traits is not something that is remotely on the critical path?

math-fehr: At least I wouldn't want to ditch dataclasses in general, though if we can see that they are bad performance wise, then yeah

xdsl/ir/core.py
superlopuh: Done and done

xdsl/ir/core.py
superlopuh: Hope it's roughly the documentation you were hoping for

PR #3364, opened 2024-10-30
transformations: (canonicalize) Arith const reassociation

n-io:
This adds the second of two canonicalisation patterns for `FloatingPointLikeBinaryOperation`s on `arith.constants`.

* Ops with two constant operands are folded by canonicalisation, regardless of math flags (see #3362 )
* Chained ops of the form `(const1 * var) * const2` are rewritten as `(const1 * const2) * var` iff the `fastmath<reassoc>` or `fastmath<fast>` flags are set, and if the ops are multiplications or additions. Note, this is not currently implemented in upstream mlir.

See [here](https://godbolt.org/z/1KKf6vbfv) to check how it works in upstream mlir.

Note, these canonicalisations currently do not support container types.

### DIFF
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -58,7 +58,7 @@ func.func @test_const_var_const() {
     // CHECK-NEXT:   %b = arith.constant 3.141500e+00 : f32
     // CHECK-NEXT:   %2 = arith.mulf %0, %a : f32
     // CHECK-NEXT:   %3 = arith.mulf %2, %b : f32
-    // CHECK-NEXT:   %4 = arith.constant 2.129352e+01 : f32
+    // CHECK-NEXT:   %4 = arith.constant 2.129352e+01 fastmath<fast> : f32
     // CHECK-NEXT:   %5 = arith.mulf %4, %0 : f32
     // CHECK-NEXT:   "test.op"(%3, %5) : (f32, f32) -> ()
 }
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -34,3 +34,31 @@ func.func @test_const_const() {
     // CHECK-NEXT:   %3 = arith.constant 9.542894e-01 : f32
     // CHECK-NEXT:   "test.op"(%0, %1, %2, %3) : (f32, f32, f32, f32) -> ()
 }
+
+func.func @test_const_var_const() {
+    %0, %1 = "test.op"() : () -> (f32, f32)
+    %a = arith.constant 2.9979 : f32
+    %b = arith.constant 3.1415 : f32
+    %c = arith.constant 4.1415 : f32
+    %d = arith.constant 5.1415 : f32
+
+    %2 = arith.mulf %0, %a : f32
+    %3 = arith.mulf %2, %b : f32
+
+    %4 = arith.mulf %0, %c fastmath<reassoc> : f32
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -34,3 +34,31 @@ func.func @test_const_const() {
...
+    return
+
+    // CHECK-LABEL: @test_const_var_const
+    // CHECK-NEXT:   %0, %1 = "test.op"() : () -> (f32, f32)
+    // CHECK-NEXT:   %a = arith.constant 2.997900e+00 : f32
+    // CHECK-NEXT:   %b = arith.constant 3.141500e+00 : f32
+    // CHECK-NEXT:   %2 = arith.mulf %0, %a : f32
+    // CHECK-NEXT:   %3 = arith.mulf %2, %b : f32
+    // CHECK-NEXT:   %4 = arith.constant 2.129352e+01 : f32
+    // CHECK-NEXT:   %5 = arith.mulf %4, %0 fastmath<fast> : f32
```
math-fehr, line 62: Is the resulting flag the `or` of both operations?
I am out of my depth here, is it what MLIR does?

math-fehr (approved):
Nice! I just had a qustion about the flags to add in the new created operation, but otherwise everything is good! Thanks!

tests/filecheck/dialects/arith/canonicalize.mlir
n-io: Yes, since we're essentially folding two ops into one, I wasn't entirely sure how to set the flags on the new op. The union set of both ops seemed like a cleaner option as compared to picking either one at random. I am not entirely sure as to how mlir would resolve such a case.

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -29,16 +52,49 @@ def match_and_rewrite(
...
+            or len(op.result.uses) != 1
+            or not isinstance(u := list(op.result.uses)[0].operation, type(op))
+            or not isinstance(
+                const2 := u.lhs.owner if u.rhs == op.result else u.rhs.owner,
+                arith.Constant,
+            )
+            or op.fastmath is None
+            or u.fastmath is None
+            or arith.FastMathFlag.REASSOC not in op.fastmath.flags
+            or arith.FastMathFlag.REASSOC not in u.fastmath.flags
```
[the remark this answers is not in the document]
n-io: It's just that my use-cases are covered by ops that happen to be commutative. In my understanding, `(var - cst) - cst` could just as easily be reassociated and simplified, although non-commutative ops have more patterns that would need to be considered.

The [docs](https://llvm.org/docs/LangRef.html#rewrite-based-flags) are a bit light on the detail, though mention that this can have a significant effect on fp results. I understand llvm has a reassociate pass, for which I can't find an equivalent in mlir.

tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -34,3 +34,31 @@ func.func @test_const_const() {
...
+    %3 = arith.mulf %2, %b : f32
+
+    %4 = arith.mulf %0, %c fastmath<reassoc> : f32
+    %5 = arith.mulf %4, %d fastmath<fast> : f32
+
+    "test.op"(%3, %5) : (f32, f32) -> ()
+
+    return
+
+    // CHECK-LABEL: @test_const_var_const
```
alexarice, line 55: I would also prefer to see the checks appear above the tests here, but feel free to ignore

alexarice (approved):

PR #3365, opened 2024-10-30
analysis: (liveness) Port of the liveness analysis in MLIR.

gabrielrodcanal:
Port of the liveness analysis in MLIR: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Analysis/Liveness.cpp

It passes the tests provided in https://github.com/llvm/llvm-project/blob/98e674c9f16d677d95c67bc130e267fae331e43c/mlir/test/Analysis/test-liveness.mlir
Note that this corresponds to commit 98e674c9f16d677d95c67bc130e267fae331e43c of LLVM, whereas xDSL now is up to date with d401987fe349a87c53fe25829215b080b70c0c1a. I'm happy to update this port to the new version if necessary (there is only one extra test that was added with the newer commit).

There are a few functions that had to be added in core xDSL to support the liveness analysis. Pre order walking was one of them that might need testing. Please, let me know if I should open a separate PR with these changes and corresponding tests before accepting this PR.

Finally, I placed the analysis under transforms for now, but I don't think it's the right place for it. Suggestions?

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1436,7 +1436,8 @@ class Block(IRNode, IRWithUses):
         )
 
     def successors(self) -> tuple[Block, ...]:
-        terminator = list(self.ops)[-1]
+        terminator = self.last_op
+        assert isinstance(terminator, Operation)
         return tuple(successor for successor in terminator.successors)
 
     def parent_op(self) -> Operation | None:
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -953,12 +953,12 @@ class Operation(IRNode):
 
     def get_parent_of_type(self, parent_type: type[Operation]) -> Operation | None:
         current_op = self
-        
+
         while parent := current_op.parent_op():
             if isinstance(parent, parent_type):
                 return parent
             current_op = parent
-            
+
         return None
 
     def walk(
@@ -984,7 +984,6 @@ class Operation(IRNode):
                 for op in block.ops:
                     yield from op.walk_blocks_preorder()
 
-
     def get_attr_or_prop(self, name: str) -> Attribute | None:
         """
         Get a named attribute or property.
@@ -995,12 +994,14 @@ class Operation(IRNode):
         if name in self.attributes:
             return self.attributes[name]
         return None
-    
-    def is_before_in_block(self, other_op : Operation) -> bool:
+
+    def is_before_in_block(self, other_op: Operation) -> bool:
         parent_block = self.parent_block()
         assert isinstance(parent_block, Block)
 
-        if parent_block.get_operation_index(self) < parent_block.get_operation_index(other_op):
+        if parent_block.get_operation_index(self) < parent_block.get_operation_index(
+            other_op
+        ):
             return True
         else:
             return False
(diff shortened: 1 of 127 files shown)


xdsl/ir/core.py
```diff
@@ -1234,6 +1261,7 @@ def emit_error(
 
         diagnostic = Diagnostic()
         diagnostic.add_message(self, message)
+        print("OPERATION ERROR: ", self)
```
superlopuh, line 1265: rogue print

xdsl/ir/core.py
```diff
@@ -967,6 +977,14 @@ def walk(
         if region_first:
             yield self
 
+    def walk_blocks_preorder(self) -> Iterator[Block]:
+        for region in self.regions:
+            for block in region.blocks:
+                yield block
+                for op in block.ops:
+                    yield from op.walk_blocks_preorder()
```
superlopuh, line 985: this would be great in its own PR with a separate unit test

xdsl/ir/core.py
```diff
@@ -1406,6 +1434,12 @@ def predecessors(self) -> tuple[Block, ...]:
         return tuple(
             p for use in self.uses if (p := use.operation.parent_block()) is not None
         )
+    
+    def successors(self) -> tuple[Block, ...]:
+        terminator = list(self.ops)[-1]
+        return tuple(
+            successor for successor in terminator.successors
+        )
```
superlopuh: does this deserve a helper? If so, I would strongly recommend against creating a list only to get its last element, `self.last_op` will do the same avoiding the O(n) list creation

superlopuh: Also better in own PR with small unit test

superlopuh:
This is great! I'd love to have this in. It would be great to split this PR into many smaller PRs, hopefully independent of each other to start with, to merge the helpers in this one before getting to the liveness analysis. Where is this in MLIR? Is it also a pass?

gabrielrodcanal:
> This is great! I'd love to have this in. It would be great to split this PR into many smaller PRs, hopefully independent of each other to start with, to merge the helpers in this one before getting to the liveness analysis. Where is this in MLIR? Is it also a pass?

You can find the liveness analysis in MLIR in https://github.com/llvm/llvm-project/blob/main/mlir/lib/Analysis/Liveness.cpp. So it's a pass, just not a transformation pass. That's why I wasn't sure where to place it in xDSL! The analysis is run with the option `test-print-liveness` as you can see in https://github.com/llvm/llvm-project/blob/98e674c9f16d677d95c67bc130e267fae331e43c/mlir/test/Analysis/test-liveness.mlir

I agree it'd be best to have several PRs with the helpers. I'll try to do that in the following days

gabrielrodcanal:
Added PR for the preorder walker #3367

superlopuh:
@gabrielrodcanal, how's it going? Will you have time to get back to this at some point? Do you have any guidance on time you'll be able to get back to it?

gabrielrodcanal:
@superlopuh I will work on this today and tomorrow so we can finally merge it this week.

superlopuh:
Amazing

superlopuh:
Closing this in favour of the new PR

PR #3367, opened 2024-10-31
core: preorder walk of blocks in nested operations

gabrielrodcanal:
Preorder walk of blocks in nested operations. This corresponds to the behaviour of WalkOrder::Preorder in MLIR. The test in the PR was developed in parallel with an analysis pass in MLIR to ensure equivalent behaviour in xDSL. For reproducibility, the relevant method in MLIR is:
```
void PreorderBlocks::print(raw_ostream &os) const {
  operation->walk<WalkOrder::PreOrder>([&](Block *block) {
    os << "--- Block: \n";
    block->print(os);
    os << "\n";
  });
}
```


tests/test_preorder_walk.py
```diff
@@ -0,0 +1,81 @@
...
+  }) : () -> ()
+"""
+
+
+def test_preorder_walk():
+    ctx = MLContext()
+    ctx.load_dialect(Builtin)
+    ctx.load_dialect(Arith)
+    ctx.load_dialect(Func)
+    ctx.load_dialect(Scf)
```
superlopuh, line 54: could you please change this test to work with only the `test` dialect, to avoid spurious dependencies?

xdsl/ir/core.py
```diff
@@ -967,6 +967,13 @@ def walk(
         if region_first:
             yield self
 
+    def walk_blocks_preorder(self) -> Iterator[Block]:
+        for region in self.regions:
+            for block in region.blocks:
```
alexarice, line 972: This is not what I would have expected a preorder traversal of blocks to do. Does llvm simply iterate over them in the order they appear?

superlopuh: Now that I think about it, it would probably be worth adding a `walk` method on `Block` that mirrors the API of the Operation walk, and to change this to forward to that one. I also prefer our `walk_regions_first` to `Preorder` as it's more declarative, maybe for blocks it could be `walk_child_blocks_first`? pre-post order doesn't make sense in the absence of left and right children

alexarice: I more meant that I would have expected a "traversal" to explore blocks via a depth first search of successors from each block, rather than just return each block in order

superlopuh: Ah no that's not what I would expect TBH. @gabrielrodcanal, if you don't feel like implementing reversed iteration and post-order then I think it would still be worth adding `walk` on Block, and `walk_blocks` on Operation, and we can add the optional parameters later without API breaking.

tests/test_preorder_walk.py
```diff
@@ -0,0 +1,119 @@
...
+}) : () -> ()
+"""
+
+
+def test_preorder_walk():
+    ctx = MLContext()
+    ctx.load_dialect(Test)
+
+    parser = Parser(ctx, test_prog)
+    op = parser.parse_op()
```
superlopuh, line 95: sorry to be annoying, but do you think you could change this from parsing the text to constructing it in Python? Something like [this](https://github.com/xdslproject/xdsl/blob/main/tests/transforms/test_loop_invariant_code_motion.py) would make the test a little easier to read and understand, IMO. then we could just `assert tuple(op.walk_blocks()) == (...)`

xdsl/ir/core.py
```diff
@@ -1677,6 +1684,13 @@ def walk(
         for op in reversed(self.ops) if reverse else self.ops:
             yield from op.walk(reverse=reverse, region_first=region_first)
 
+    def walk_child_blocks_first(self) -> Iterator[Block]:
+        for op in self.ops:
+            for region in op.regions:
+                for block in region.blocks:
+                    yield block
+                    yield from block.walk_child_blocks_first()
```
superlopuh, line 1692: It feels natural for it to have this name, WDYT? It would be worth testing also, probably in the same pytest
```suggestion
    def walk(self) -> Iterator[Block]:
        for op in self.ops:
            for region in op.regions:
                for block in region.blocks:
                    yield block
                    yield from block.walk()
```

superlopuh:
Any updates on this? Will you have some time to iterate on it this week?

xdsl/ir/core.py
```diff
@@ -967,6 +967,13 @@ def walk(
         if region_first:
```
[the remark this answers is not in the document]
superlopuh: I think it makes sense to keep this logic in ir.py

compor:
> Taking over this PR, hope that's ok @gabrielrodcanal

I'm interested in having a look if you don't mind; I'll relinquish if it's too much for my time ATM

math-fehr (approved):
👍

PR #3368, opened 2024-10-31
dialects: (arith) select canonicalization patterns

alexarice:
Adds some simple canonicalization patterns for `arith.select`

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/arith.py
+++ b/xdsl/transforms/canonicalization_patterns/arith.py
@@ -147,3 +147,14 @@ class SelectTrueFalsePattern(RewritePattern):
 
         if lhs_value.value.data == 0 and rhs_value.value.data == 1:
             rewriter.replace_matched_op(arith.XOrI(op.cond, rhs))
+
+
+class SelectSamePattern(RewritePattern):
+    """
+    arith.select %x %y %y = %y
+    """
+
+    @op_type_rewrite_pattern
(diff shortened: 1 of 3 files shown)


superlopuh (approved):

alexarice:
I'll merge after #3362  and #3364  to avoid conflicts

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -16,3 +16,48 @@ def match_and_rewrite(self, op: arith.Addi, rewriter: PatternRewriter) -> None:
...
+        if const_cond.value.data == 1:
+            rewriter.replace_matched_op((), (op.lhs,))
+        if const_cond.value.data == 0:
+            rewriter.replace_matched_op((), (op.rhs,))
+
+
+class SelectTrueFalsePattern(RewritePattern):
+    """
+    arith.select %x %true %false = %x
+    arith.select %x %false %true = %x xor 1
```
n-io, line 43: For an `i1` result type you may even consider folding the `%true %true` and `%false %false` cases if you think they may realistically come up.

n-io (approved):

xdsl/transforms/canonicalization_patterns/arith.py
alexarice: More generally, `arith.select %x %y %y` should be %y right? I realise I didn't include this as I didn't immediately find it in mlir but it would probably be nice to add

alexarice:
Added select x y y = y and rebased

PR #3371, opened 2024-10-31
misc: move `get_all_passes` to xdsl.transforms

alexarice:
Moves `get_all_passes` from xdsl.tools.command_line_tool (why was it here?) to `xdsl.transforms`.

This aligns with where `get_all_dialects` is kept.


alexarice:
Is there a way to help git understand that this is mainly a code move rather than rewrite?

superlopuh (approved):
Much better, thank you!

PR #3372, opened 2024-10-31
misc: refactor xdsl-gui

alexarice:
Creates a new class `CommandLineToolWithPasses` which contains the shared functionality of `xdsl-opt` and `xdsl-gui` and inherits from `CommandLineTool`.

Makes `xdsl-gui` inherit from this new class, allowing the available dialects and passes to be overriden in the same way as for xdsl-opt.

As part of this change the ctx becomes static instead of being remade of each input change. I am unsure if this will cause problems so would be good if anyone is aware of any dodgy cases.

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -31,7 +31,7 @@ from xdsl.utils.parse_pipeline import PipelinePassSpec, parse_pipeline
 @pytest.mark.asyncio
 async def test_inputs():
     """Test different inputs produce desired result."""
-    async with InputApp().run_test() as pilot:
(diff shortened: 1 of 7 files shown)


alexarice:
Steps to improve:
- Make this PR a refactor to reduce dependency on `get_all_passes` and `get_all_dialects`
- Move all registering of dialects/targets/frontends/passes into one object which is shared by all tools and passed in
- Remove all dependency on `self.args`
- Remove `self.args`
- Classmethod approach to arguments

alexarice:
Going to close this as it's significantly different to what we eventually went with.

PR #3373, opened 2024-10-31
transformations: (csl-stencil-bufferize) Reselect linalg outs

n-io:
This PR sets destination buffers for linalg ops, if the destination buffer is one of its inputs.

Known caveats:
* Setting the outs for all ops to one buffer known to be writable causes read-after-write conflicts, which the bufferization does not analyse for
* The default behaviour (of mlir-opt's `linalg-fuse-elementwise-ops` pass) is to set `outs` to `input[0]`, which in some cases causes a significant overhead of `memref.alloc`s

Iff the `outs` of a linalg op is its own `input[n]`, this PR re-selects `n` in one of two ways:
* prefer an input with the `writable` flag and no later use, or if none is found
* prefer a linalg op input with no later use (i.e. chaining)

### DIFF
--- a/tests/filecheck/transforms/csl_stencil_bufferize.mlir
+++ b/tests/filecheck/transforms/csl_stencil_bufferize.mlir
@@ -20,9 +20,9 @@ builtin.module {
       %15 = arith.constant dense<1.666600e-01> : tensor<510xf32>
       %16 = "tensor.extract_slice"(%14) <{"static_offsets" = array<i64: 2>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
       %17 = "tensor.extract_slice"(%14) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
-      %18 = linalg.add ins(%13, %17 : tensor<510xf32>, tensor<510xf32>) outs(%13 : tensor<510xf32>) -> tensor<510xf32>
-      %19 = linalg.add ins(%18, %16 : tensor<510xf32>, tensor<510xf32>) outs(%18 : tensor<510xf32>) -> tensor<510xf32>
-      %20 = linalg.mul ins(%19, %15 : tensor<510xf32>, tensor<510xf32>) outs(%19 : tensor<510xf32>) -> tensor<510xf32>
+      %18 = linalg.add ins(%13, %17 : tensor<510xf32>, tensor<510xf32>) outs(%17 : tensor<510xf32>) -> tensor<510xf32>
+      %19 = linalg.add ins(%18, %16 : tensor<510xf32>, tensor<510xf32>) outs(%16 : tensor<510xf32>) -> tensor<510xf32>
+      %20 = linalg.mul ins(%19, %15 : tensor<510xf32>, tensor<510xf32>) outs(%15 : tensor<510xf32>) -> tensor<510xf32>
       csl_stencil.yield %20 : tensor<510xf32>
     }) to <[0, 0], [1, 1]>
     func.return
--- a/xdsl/transforms/csl_stencil_bufferize.py
+++ b/xdsl/transforms/csl_stencil_bufferize.py
@@ -395,7 +395,7 @@ class ReselectLinalgOutsFromInputs(RewritePattern):
         if (
             op.outputs[0] not in op.inputs
             or self.is_writable(op.outputs[0])
-            and len(op.outputs) == 1
+            or len(op.outputs) != 1
         ):
             return
 
--- a/xdsl/transforms/csl_stencil_bufferize.py
+++ b/xdsl/transforms/csl_stencil_bufferize.py
@@ -13,7 +13,15 @@ from xdsl.dialects.builtin import (
     i64,
 )
 from xdsl.dialects.csl import csl_stencil
-from xdsl.ir import Attribute, Block, BlockArgument, Operation, Region, SSAValue
+from xdsl.ir import (
+    Attribute,
+    Block,
+    BlockArgument,
+    Operation,
+    OpResult,
+    Region,
+    SSAValue,
+)
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
     GreedyRewritePatternApplier,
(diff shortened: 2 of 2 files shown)


tests/filecheck/transforms/csl_stencil_bufferize.mlir
```diff
@@ -20,9 +20,9 @@ builtin.module {
...
       %16 = "tensor.extract_slice"(%14) <{"static_offsets" = array<i64: 2>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
       %17 = "tensor.extract_slice"(%14) <{"static_offsets" = array<i64: 0>, "static_sizes" = array<i64: 510>, "static_strides" = array<i64: 1>, "operandSegmentSizes" = array<i32: 1, 0, 0, 0>}> : (tensor<512xf32>) -> tensor<510xf32>
-      %18 = linalg.add ins(%13, %17 : tensor<510xf32>, tensor<510xf32>) outs(%13 : tensor<510xf32>) -> tensor<510xf32>
-      %19 = linalg.add ins(%18, %16 : tensor<510xf32>, tensor<510xf32>) outs(%18 : tensor<510xf32>) -> tensor<510xf32>
-      %20 = linalg.mul ins(%19, %15 : tensor<510xf32>, tensor<510xf32>) outs(%19 : tensor<510xf32>) -> tensor<510xf32>
+      %18 = linalg.add ins(%13, %17 : tensor<510xf32>, tensor<510xf32>) outs(%17 : tensor<510xf32>) -> tensor<510xf32>
+      %19 = linalg.add ins(%18, %16 : tensor<510xf32>, tensor<510xf32>) outs(%16 : tensor<510xf32>) -> tensor<510xf32>
+      %20 = linalg.mul ins(%19, %15 : tensor<510xf32>, tensor<510xf32>) outs(%15 : tensor<510xf32>) -> tensor<510xf32>
```
n-io, line 25: This is messing up the previously well-chosen outs, to show that the pass can select from writable inputs (re-selecting `%13`) and chaining of ops (re-selecting `%18` and `%19`).

n-io: Note, there are no changes to the check-next to show this working.

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -371,13 +379,77 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+            and len(op.outputs) == 1
+        ):
+            return
+
+        # the new `outs` to re-select
+        out: SSAValue | None = None
+
+        for arg in op.inputs:
+            # reselect outs that has no later use to avoid read-after-write conflicts
+            if len(arg.uses) == 1:
```
dk949, line 407: Does this check for later uses or all uses including ones before this one?

E.g.

```mlir
%0 = ...
%1 = "test.op"() : () -> memref<10xf32>
%2 = ...
"test.op"(%1) : (memref<10xf32>) -> ()
linalg.add ins(%1, %2 : memref<10xf32>, memref<10xf32>) outs(%0 : memref<10xf32>)

// %1 not used after here, but len(uses) > 1?
```

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -371,13 +379,77 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+                # check for a `writable` input with no later uses and break immediately
+                if self.is_writable(arg):
+                    out = arg
+                    break
+
+                # check for a linalg op input with no later uses and keep looking
+                if isinstance(arg, OpResult) and isinstance(
+                    arg.op, linalg.NamedOpBase | linalg.Generic
+                ):
+                    out = arg
```
dk949, line 417: If input is not writable, why do we still make it the output? Since we have asserted that at this point the output must be one of the inputs and not writable. Would this not needlessly rebuild the op?

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -371,13 +379,77 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+
+        # replace the op with `out` as `output[0]`
+        if out:
+            rewriter.replace_matched_op(
+                type(op).build(
+                    operands=[op.inputs, [out]],
+                    result_types=op.result_types,
+                    regions=[op.detach_region(r) for r in op.regions],
+                    properties=op.properties,
+                    attributes=op.attributes,
```
dk949, line 427: I don't suppose there's some way to `clone()` then only change the `outputs`? This makes it a bit harder to figure out what actually changes in the rebuild

tests/filecheck/transforms/csl_stencil_bufferize.mlir
dk949, line 25: Would it be possible to have some tests where results/destination ends up being not writable? 

And if these are such tests, could you please add a comment?

dk949 (commented):
A few minor things. All good otherwise :+1:

xdsl/transforms/csl_stencil_bufferize.py
n-io: Difficult with varargs unfortunately, and not the cleanest.

tests/filecheck/transforms/csl_stencil_bufferize.mlir
n-io: lines 24 and 25 don't have writable results

xdsl/transforms/csl_stencil_bufferize.py
n-io: It checks for no other uses at all, as the check for no later uses isn't quite straightforward to implement. Both should logically work, though in theory there might be some additional edge cases that might be caught checking for later uses.

xdsl/transforms/csl_stencil_bufferize.py
dk949: That makes sense. Although I would've expected `mlir` to figure that out when it does the bufferisation

dk949 (approved):

xdsl/transforms/csl_stencil_bufferize.py
n-io: It doesn't really do much analysis, and generally has two options it considers: re-use the buffer given as outs, or create a copy. Chaining ops, such as providing the result of a prev op as outs (not: the outs of a prev op) helps with the first option. But if outs is a tensor from an arith.const, it creates a buffer copy. Try it on https://godbolt.org/z/YjrTG54oP

PR #3375, opened 2024-11-01
dialects: (llvm) fixed generic printing of llvm.icmp

lfrenot:


lfrenot:
I'm working on adding the rest of the test cases

lfrenot:
just realised the printing should be without quotes around "predicate"
I'll fix that

lfrenot:
> just realised the printing should be without quotes around "predicate" I'll fix that

Oh
I now realize it's also the case on the overflow flags
I'm not sure how to fix it then

lfrenot:
> > just realised the printing should be without quotes around "predicate" I'll fix that
> 
> Oh I now realize it's also the case on the overflow flags I'm not sure how to fix it then

@alexarice @superlopuh do you have an idea?

tobiasgrosser:
> > just realised the printing should be without quotes around "predicate" I'll fix that
> 
> Oh I now realize it's also the case on the overflow flags I'm not sure how to fix it then

icmp does not have overflow flags, does it? In case this is not related to icmp, I suggest to first fix icmp and then worry about overflow flags in the next PR.

lfrenot:
> > > just realised the printing should be without quotes around "predicate" I'll fix that
> > 
> > 
> > Oh I now realize it's also the case on the overflow flags I'm not sure how to fix it then
> 
> icmp does not have overflow flags, does it? In case this is not related to icmp, I suggest to first fix icmp and then worry about overflow flags in the next PR.

Yes of course, I zqs just pointing out that they have the same issue and that it will have to be fixed eventually

tobiasgrosser:
Right. Is this PR ready to go from your side or are you waiting for some specific feedback?

I suggest to make this clear in a comment at the bottom of this PR such that others can take action to advance your PR.

lfrenot:
I don't think the PR is ready
Since the generic output of `xdsl-opt` is still different from `mlir-opt`'s and I don't know if that difference matters

tobiasgrosser:
Ah, I missed that the `"` around the predicate are still different. Can you check if `mlir-opt` accepts the output that you are currently generating?

xdsl/dialects/llvm.py
```diff
@@ -633,6 +634,13 @@ class ICmpPredicateAttr(SpacedOpaqueSyntaxAttribute, EnumAttribute[ICmpPredicate
     def from_int(cls, i: int):
         return ICmpPredicateAttr(cls.ALL_PREDICATES[i])
 
+    @classmethod
+    def int_from(cls, flag: ICmpPredicateFlag) -> int:
+        return cls.ALL_PREDICATES.index(flag)
```
math-fehr, line 639: Can you call this function `to_int` and add it as a method of `ICmpPredicateFlag` directly?

tests/filecheck/dialects/llvm/generic-syntax.mlir
```diff
@@ -0,0 +1,61 @@
+// RUN: xdsl-opt %s --print-op-generic | filecheck %s
```
math-fehr, line 1:
```suggestion
// RUN: xdsl-opt %s | xdsl-opt --print-op-generic | filecheck %s
```
To check that it roundtrips as well

tests/filecheck/dialects/llvm/generic-syntax.mlir
```diff
@@ -0,0 +1,61 @@
...
+
+"builtin.module"() ({
+
+    %arg0, %arg1 = "test.op"() : () -> (i32, i32)
+    
+    // CHECK: "builtin.module"() ({
+    // CHECK-NEXT:  %arg0, %arg1 = "test.op"() : () -> (i32, i32)
```
math-fehr, line 8:
```suggestion
```
No need for these two

math-fehr (commented):
Looks good to me, can you just add a similar test in
`tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/`?
To check in the CI that it works with MLIR.

Also, can you rename the file to `icmp.mlir`?

tests/filecheck/dialects/llvm/generic-syntax.mlir
alexarice: This is the command `XDSL_GENERIC_ROUNDTRIP` right?

tests/filecheck/dialects/llvm/generic-syntax.mlir
```diff
@@ -0,0 +1,61 @@
+// RUN: xdsl-opt %s --print-op-generic | filecheck %s
+
+"builtin.module"() ({
```
alexarice, line 3: The module can also be left implicit

xdsl/dialects/llvm.py
```diff
@@ -641,22 +649,22 @@ class ICmpOp(IRDLOperation):
...
+    predicate = prop_def(IntegerAttr[i64])
 
     traits = frozenset([NoMemoryEffect()])
 
     def __init__(
         self,
         lhs: SSAValue,
         rhs: SSAValue,
-        predicate: ICmpPredicateAttr,
+        predicate: IntegerAttr[IntegerType],
```
alexarice, line 661: Should this not also be `IntegerAttr[i64]`?

tests/filecheck/dialects/llvm/generic-syntax.mlir
lfrenot: For the "This is the command XDSL_GENERIC_ROUNDTRIP right?" comment:

Yes it is, but if I read `config.substitutions.append(("XDSL_GENERIC_ROUNDTRIP", "xdsl-opt %s --print-op-generic --split-input-file | filecheck %s --check-prefix=CHECK-GENERIC"))` correctly,
that command doesn't do the roundtrip Mathieu suggested

xdsl/dialects/llvm.py
lfrenot: no, I get the error `Variable not allowed in type expression` if I do that
Doing `IntegerAttr[IntegerType(64)]` gives the error `Call expression not allowed in type expression`

xdsl/dialects/llvm.py
lfrenot: I would rather not, since I'm using `ALL_PREDICATES`

tests/filecheck/dialects/llvm/generic-syntax.mlir
alexarice: Fair enough, go with Mathieu's suggestion instead then

xdsl/dialects/llvm.py
superlopuh: Yep that's because the Python type system cannot do the things we do at runtime for verification.

xdsl/dialects/llvm.py
superlopuh: What is the point of the attribute if it's never used in the IR? Could the helpers not be on the enum directly?

superlopuh: Also, this is also O(n), I would recommend doing something like this:
```suggestion
    INDEX_BY_PREDICATE = {f: i for (i, f) in enumerate(ALL_PREDICATES)}

    @classmethod
    def int_from(cls, flag: ICmpPredicateFlag) -> int:
        return INDEX_BY_PREDICATE[flag]
```

lfrenot:
@tobiasgrosser @superlopuh are we good to merge?

tobiasgrosser:
From my side, this is good to go as soon as all comments are addressed. If the comments of the others are also addressed and they give a 👍 we are good to go.

superlopuh:
I just chatted with Tobias, and he asked me to accelerate this a bit, I'll push my changes as a hotfix, exceptionally!

superlopuh:
I just noticed this, but this branch does not have an author prefix, which makes it harder to manage all the branches on the remote. Could you please prefix your branches with some identifier, like `leon/llvm-icmp-generic`?

lfrenot:
Oh, okay, I'll do that from now
Can/Should I rename this one?

tobiasgrosser:
let's leave it for this one. @superlopuh can you push merge when you are ready and ping me.

Issue #3370, opened 2024-10-31
Add support for hexadecimal float literals including nan/inf


n-io:
When supplied with a hexadecimal float literal, xdsl instead parses it as a hexadecimal int literal, which is subsequently converted to a float. The hexadecimal literal should instead by interpreted as a bit-string for the resulting float, to ensure compatibility with upstream. Special values such as `nan`/`-inf`/`inf` should always be printed as hexadecimal float literals.

While this theoretically affects all float values, it is particularly important for representing special values such as `nan`, `-inf`, and `inf`, which in mlir-opt are printed as hexadecimal float literals (see the [FloatAttr docs](https://mlir.llvm.org/docs/Dialects/Builtin/#floatattr) or try it on [godbolt](https://godbolt.org/z/5Ea4qPzTG)). Note, these are bit values rather than set constants, and differ depending on the bit width of the data type.

xdsl currently prints special floats using python's built-in mechanism, which is different from upstream mlir and subsequently (correctly!) fails to parse:
```
>>> str(arith.Constant(builtin.FloatAttr(float('inf'),32)))
'%0 = arith.constant inf : f32'
```

This consists of two parts:
* [x] Print special float values as hexadecimal literals w.r.t. the bitwidth #3377
* [x] Parse hexadecimal literals as bit strings for the respective target bitwidth, converting neither via int nor via python's builtin float type. #3378

n-io:
I'm wondering if going via [ctypes](https://stackoverflow.com/a/1592200) would help to support different bit widths, python's own hexadecimal float format is quite different from the hexadecimal bit strings that mlir produces.

n-io:
Some musings:
```
# parsing
import numpy as np
np.uint16(0x7C00).view('float16')  == np.inf
np.isneginf(np.uint16(0xFC00).view('float16'))

#printing
hex(np.float16(float('-inf')).view('uint16'))
```

PR #3377, opened 2024-11-01
core: Print FloatAttr special values

n-io:
This prints `FloatAttr` special values `nan`, `inf`, `-inf` as a hexadecimal binary representation to align with mlir-opt as described in #3370

Also see the [FloatAttr docs](https://mlir.llvm.org/docs/Dialects/Builtin/#floatattr) or give it a try on [godbolt](https://godbolt.org/z/89TfsvaKT).


xdsl/printer.py
```diff
@@ -7,6 +7,7 @@
 from itertools import chain
 from typing import Any, TypeVar, cast
 
+import numpy as np
```
alexarice, line 10: Were we already depending on numpy? This seems like a pretty sizable import if not

xdsl/printer.py
```diff
@@ -534,10 +535,25 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                elif isinstance(attr_type, Float32Type):
+                    self.print_string(
+                        f"{hex(np.float32(value.data).view('uint32'))} : "
+                    )
+                elif isinstance(attr_type, Float64Type):
+                    self.print_string(
+                        f"{hex(np.float64(value.data).view('uint64'))} : "
+                    )
+                else:
+                    # todo
```
alexarice, line 553: what is the todo?

tests/test_printer.py
```diff
@@ -761,6 +763,28 @@ def test_densearray_attr():
     assert_print_op(parsed, prog, None)
 
 
+def test_float_attr_specials():
```
alexarice, line 766: Do these also parse correctly?

xdsl/printer.py
n-io: To add support for other types such as f80, f128

tests/test_printer.py
n-io: Absolutely not 🫠 Working on it though

n-io: We currently can't parse special floats at all

xdsl/printer.py
n-io: Updated this in the code.

xdsl/printer.py
n-io: I think only JAX uses it so far.

superlopuh (changes requested):
We do not depend on numpy, and JAX is an optional dependency, so we can't use it for core features.

xdsl/printer.py
```diff
@@ -534,10 +535,25 @@ def print_attribute(self, attribute: Attribute) -> None:
 
         if isinstance(attribute, FloatAttr):
             value = attribute.value
-            attr_type = cast(
-                FloatAttr[Float16Type | Float32Type | Float64Type], attribute
-            ).type
-            self.print_string(f"{value.data:.6e} : ")
+            attr_type = cast(AnyFloatAttr, attribute).type
+            if np.isnan(value.data) or np.isinf(value.data):
```
superlopuh, line 539:
```suggestion
            if math.isnan(value.data) or math.isinf(value.data):
```

xdsl/printer.py
```diff
@@ -534,10 +535,25 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                if isinstance(attr_type, Float16Type):
+                    self.print_string(
+                        f"{hex(np.float16(value.data).view('uint16'))} : "
+                    )
+                elif isinstance(attr_type, Float32Type):
+                    self.print_string(
+                        f"{hex(np.float32(value.data).view('uint32'))} : "
```
superlopuh, line 546:
```suggestion
                        f"{hex(convert_f32_to_u32(inf))} : "
```

superlopuh: Would need a few more functions

xdsl/printer.py
n-io: done as suggested

xdsl/printer.py
n-io: I'm not sure what `convert_f32_to_u32` is or how it would be implemented?

xdsl/printer.py
```diff
@@ -534,10 +536,25 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                    self.print_string(
+                        f"{hex(struct.unpack('<I', struct.pack('<f', value.data))[0])} : "
+                    )
+                elif isinstance(attr_type, Float64Type):
+                    self.print_string(
+                        f"{hex(struct.unpack('<Q', struct.pack('<d', value.data))[0])} : "
+                    )
+                else:
+                    # todo add support for further bitwidths
+                    self.print_string(f"{value.data:.6e} : ")
```
n-io, line 555: I'm wondering if this should throw an error

superlopuh: Yeah seems like the right thing to do! Also might want to refactor this to a match statement, and throw an error in the default branch

n-io: Raising it as an error.

xdsl/printer.py
superlopuh: It's in xDSL already :) Could you please let the original commenter resolve the conversations, that way they don't accidentally get missed in subsequent reviews

xdsl/printer.py
```diff
@@ -534,10 +536,27 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                        f"{hex(struct.unpack('<I', struct.pack('<f', value.data))[0])} : "
+                    )
+                elif isinstance(attr_type, Float64Type):
+                    self.print_string(
+                        f"{hex(struct.unpack('<Q', struct.pack('<d', value.data))[0])} : "
+                    )
+                else:
+                    # todo add support for further bitwidths
+                    raise ValueError(
+                        f"Cannot print '{str(value.data)}' value for float type {str(attr_type)}"
```
superlopuh, line 556:
```suggestion
                        f"Cannot print '{value.data}' value for float type {str(attr_type)}"
```

n-io: Done as suggested.

xdsl/printer.py
n-io: The point was to remove np?

n-io: ctypes unfortunately does not support f16.

superlopuh: Ah OK, thanks for looking into it

superlopuh (approved):
Thank you!

xdsl/printer.py
```diff
@@ -534,10 +536,27 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                        f"{hex(struct.unpack('<H', struct.pack('<e', value.data))[0])} : "
+                    )
+                elif isinstance(attr_type, Float32Type):
+                    self.print_string(
+                        f"{hex(struct.unpack('<I', struct.pack('<f', value.data))[0])} : "
+                    )
+                elif isinstance(attr_type, Float64Type):
+                    self.print_string(
+                        f"{hex(struct.unpack('<Q', struct.pack('<d', value.data))[0])} : "
+                    )
```
superlopuh, line 552: Now that I think about it, it feels like it would be better to use the helpers in `bitwise_casts` after all, using the struct or ctypes approach. We will need the logic in a few places, such as when storing float data in assembly, and other optimisations, so it would be worth having it in an easily reusable place.

xdsl/printer.py
```diff
@@ -534,10 +536,27 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                else:
+                    # todo add support for further bitwidths
+                    raise ValueError(
```
superlopuh, line 555:
```suggestion
                    raise NotImplementedError(
```

n-io:
I have added further `bitwise_cast` functions and outlined the print float logic, as it is also used for printing `dense` elements.

xdsl/printer.py
n-io: Done as suggested.

PR #3378, opened 2024-11-01
core: Parse FloatAttr from hexadecimal bit representation

n-io:
`FloatAttr`s can have their values specified by a hexadecimal literal. This should be interpreted as the bit representation of the float. However, it is currently interpreted as an int number and then converted to a float, leading to incorrect behaviour.

This PR is the second part of #3370 after #3377

### DIFF
--- a/xdsl/parser/attribute_parser.py
+++ b/xdsl/parser/attribute_parser.py
@@ -1144,6 +1144,10 @@ class AttrParser(BaseParser):
         if bool is not None:
             return bool
 
+        hexadecimal_token: Token | None = None
+        if self._current_token.text[:2] in ["0x", "0X"]:
+            hexadecimal_token = self._current_token
+
         # Parse the value
         if (value := self.parse_optional_number()) is None:
             return None
@@ -1158,6 +1162,8 @@ class AttrParser(BaseParser):
         type = self._parse_attribute_type()
 
         if isinstance(type, AnyFloat):
+            if hexadecimal_token:
+                return FloatAttr(hexadecimal_token.get_float_value(), type)
             return FloatAttr(float(value), type)
 
         if isinstance(type, IntegerType | IndexType):
--- a/xdsl/utils/lexer.py
+++ b/xdsl/utils/lexer.py
@@ -7,6 +7,8 @@ from io import StringIO
 from string import hexdigits
 from typing import Literal, TypeAlias, TypeGuard, cast, overload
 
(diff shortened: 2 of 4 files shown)


xdsl/parser/attribute_parser.py
```diff
@@ -1144,6 +1144,10 @@ def parse_optional_builtin_int_or_float_attr(
         if bool is not None:
             return bool
 
+        hexadecimal_token: Token | None = None
+        if self._current_token.text[:2] in ["0x", "0X"]:
+            hexadecimal_token = self._current_token
+
```
n-io, line 1150: I'm not sure this is the best way to do this, but it prevents losing the information that the token is a hexadecimal literal instead of a generic integer, until the attribute type is parsed.

xdsl/utils/lexer.py
```diff
@@ -7,6 +7,8 @@
 from string import hexdigits
 from typing import Literal, TypeAlias, TypeGuard, cast, overload
 
+import numpy as np
```
superlopuh, line 10: same here, numpy is not a core dependency, we need to use other helpers for functionality below

superlopuh (changes requested):

xdsl/utils/lexer.py
```diff
@@ -347,6 +348,18 @@ def get_float_value(self):
...
+                # the number corresponds to `len('0x') + (float_bitwidth / 4)`
+                case 6:
+                    return struct.unpack("<e", struct.pack("<H", int(self.text, 16)))[0]
+                case 10:
+                    return struct.unpack("<f", struct.pack("<I", int(self.text, 16)))[0]
+                case 18:
+                    return struct.unpack("<d", struct.pack("<Q", int(self.text, 16)))[0]
+                case _:
+                    # todo support further bitwidths
+                    pass
```
superlopuh, line 362: shouldn't the bitwidth be determined by the floating point type? I would expect this to be in a helper function that takes an int and a bitwidth, possibly in the bitwise_casts file, with dedicated tests, and for the logic to be done inline in `parse_optional_number`

n-io: Good spot, let me fix that

n-io: Hadn't really considered implied leading zeros, but mlir-opt supports it

superlopuh (approved):
LGTM!

PR #3379, opened 2024-11-01
interactive: refactor to one use of get_all_passes/get_all_dialects

alexarice:
Changes the interactive app object to store a pass list and context, which are set for now in the initialization function. This will allow further refactors which allow these to be overridden more easily by downstream projects.

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -46,13 +46,13 @@ async def test_inputs():
         await pilot.pause()
         assert (
             app.output_text_area.text
-            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     unregistered operation dkjfd!\n"
         )
 
         assert isinstance(app.current_module, ParseError)
         assert (
             str(app.current_module)
-            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     unregistered operation dkjfd!\n"
         )
 
         # Test corect input
@@ -274,7 +274,9 @@ async def test_buttons():
         await pilot.pause()
         # assert after "Condense Button" is clicked that the state and condensed_pass list change accordingly
         assert app.condense_mode is True
-        assert app.available_pass_list == get_condensed_pass_list(expected_module)
+        assert app.available_pass_list == get_condensed_pass_list(
+            expected_module, app.all_passes
+        )
 
         # press "Uncondense" button
         await pilot.click("#uncondense_button")
@@ -321,7 +323,9 @@ async def test_rewrites():
         # assert after "Condense Button" is clicked that the state and get_condensed_pass list change accordingly
         assert app.condense_mode is True
         assert isinstance(app.current_module, ModuleOp)
-        condensed_list = get_condensed_pass_list(app.current_module) + (addi_pass,)
+        condensed_list = get_condensed_pass_list(app.current_module, app.all_passes) + (
+            addi_pass,
+        )
         assert app.available_pass_list == condensed_list
 
         # Select a rewrite
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -41,18 +41,18 @@ async def test_inputs():
         assert app.output_text_area.text == "No input"
         assert app.current_module is None
 
-        # Test inccorect input
+        # Test incorrect input
         app.input_text_area.insert("dkjfd")
         await pilot.pause()
         assert (
             app.output_text_area.text
-            == "<unknown>:1:5\ndkjfd\n     ^\n     unregistered operation dkjfd!\n"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
         )
 
         assert isinstance(app.current_module, ParseError)
         assert (
             str(app.current_module)
-            == "<unknown>:1:5\ndkjfd\n     ^\n     unregistered operation dkjfd!\n"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
         )
 
         # Test corect input
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -9,7 +9,7 @@ from xdsl.backend.riscv.lowering import (
     convert_func_to_riscv_func,
 )
 from xdsl.builder import ImplicitBuilder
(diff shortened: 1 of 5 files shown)


xdsl/interactive/get_all_available_passes.py
```diff
@@ -16,6 +15,8 @@
 
 
 def get_available_pass_list(
+    ctx: MLContext,
```
superlopuh, line 18: why pass the ctx here?

alexarice: It's needed for the parser

xdsl/interactive/app.py
```diff
@@ -91,6 +91,11 @@ class InputApp(App[None]):
         }
         """
 
+    ctx: MLContext
```
superlopuh, line 94: I think it's quite important to not add mutable state to the app, could you please refactor this to avoid passing the context? Probably something like this:
```suggestion
    all_dialects: tuple[tuple[str, Callable[[], Dialect]]]
```
And just pass that in the init. Similarly, `all_passes` below should be passed in the init. This will also massively improve our tests, as we can initialize the test app with a smaller set of dialects and passes for it to condense.

alexarice: This seems like a good idea. I'm leaning towards making both "all_passes" and "all_dialects" into dictionaries. I don't really see why they should be tuples of tuples?

tests/interactive/test_app.py
```diff
@@ -42,7 +42,7 @@ async def test_inputs():
         assert app.output_text_area.text == "No input"
         assert app.current_module is None
 
-        # Test inccorect input
```
alexarice, line 45: Just occurred to me this might have been a joke that I wasn't funny enough to understand

alexarice:
@superlopuh, I've had a second pass at this given the feedback, let me know what you think.

xdsl/interactive/app.py
superlopuh: because tuples are frozen, mostly. a frozendict would also work, but requires an extra dependency

xdsl/interactive/passes.py
```diff
@@ -1,15 +1,13 @@
...
 
 from xdsl.context import MLContext
 from xdsl.dialects import builtin, get_all_dialects
+from xdsl.ir import Dialect
 from xdsl.passes import ModulePass, PipelinePass
-from xdsl.transforms import get_all_passes
 from xdsl.transforms.mlir_opt import MLIROptPass
 from xdsl.utils.parse_pipeline import PipelinePassSpec
 
-ALL_PASSES = tuple(sorted((p_name, p()) for (p_name, p) in get_all_passes().items()))
```
superlopuh, line 10: the immutability is why this was a tuple also

alexarice: Ah, I assumed it was for the ordered Ness (python dicts are now ordered). Happy to make both tuples or make them private or both

alexarice:
Reverted to using tuples for both dialects and passes

xdsl/interactive/app.py
```diff
@@ -144,7 +150,14 @@ def __init__(
         file_path: str | None = None,
         input_text: str | None = None,
         pass_pipeline: tuple[tuple[type[ModulePass], PipelinePassSpec], ...] = (),
+        all_dialects: dict[str, Callable[[], Dialect]] = get_all_dialects(),
+        all_passes: dict[str, Callable[[], type[ModulePass]]] = get_all_passes(),
     ):
+        self.all_dialects = tuple((d_name, d) for d_name, d in all_dialects.items())
+        self.all_passes = tuple(
+            (p_name, p()) for (p_name, p) in sorted(all_passes.items())
+        )
```
superlopuh, line 159: This is code smell to me, if someone already has the dialects and passes as a tuple, they would need to first build a dict to pass to the init, it's a shame to make it impossible to do the dumb/fast thing.

alexarice: I kind of get this, though to me the dictionary is the "correct" interface, in that it stops you assigning the same name to two different dialects/passes, whereas storing it as a tuple is really just an implementation detail.

xdsl/interactive/app.py
```diff
@@ -131,8 +138,7 @@ class InputApp(App[None]):
     """DataTable displaying the operation names and counts of the input text area."""
     diff_operation_count_datatable: DataTable[str | int]
     """
-    DataTable displaying the diff of operation names and counts of the input and output
-    text areas.
+    DataTable displaying the diff of operation names and counts of the input and     text areas.
```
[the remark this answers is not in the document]
alexarice: I have no idea how this happened

xdsl/interactive/app.py
alexarice: Note that `all_passes` has to have every pass actually called at some point as well. Unsure what the nicest interface is, realistically this is called once per project

superlopuh: I agree in general, and this is where frozendict is nice, as it guarantees both unicity and frozenness, but in this case I think we'll be ok with the potential for errors in this way. In either case, this is a separate discussion from whether the init allows for passing the object to be stored directly or not

superlopuh: Not dictionary, since it's mutable, so tuple for now, please. In theory, we could add a dependency to frozendict, but I'm not sure it's worth it for this PR.

alexarice: I meant just for the init function

superlopuh: Ah, I don't really see the point, if it's just `tuple(all_dialects.items())` on the client side, and avoids `isinstance` checks in the init.

superlopuh: I had a secret plan of making that change after you merge your PR :)

superlopuh: Because it seems like too big a change to do both at the same time, but there's not a particularly good reason to make it eager at the app level.

alexarice: What's the solution then? 
- dictionaries with default and convert in init
- choice of dictionary/tuple and need isinstance checks
- only tuple but don't give defaults

superlopuh: only tuple, don't give defaults

alexarice: Done, tests aren't pretty atm but I'm guessing you're hoping fine tune the passes they need in a future pr

alexarice: Realised I could have put the default for dialects, as this shouldn't be too expensive, let me know if you want it in

superlopuh (approved):
Thank you!

xdsl/interactive/app.py
superlopuh: I think the user needs to know they're doing something potentially very expensive, and to expose how to customize it for downstream users such as yourself

alexarice:
Rebased past conflict, will merge if ci is green

PR #3381, opened 2024-11-01
core: Adding option to print full precision fp

n-io:
`mlir-opt` attempts to print floats as 6dp scientific notation, but round-trips to ensure there is no precision loss and the printed number will bitwise reproduce exactly. If this fails, it will choose a different printing method to ensure bit-reproducible printing for a given precision of fp types.

~~This PR introduces a flag to print with full precision. This should eventually become the default.~~
~~This PR prints with full precision by default, but provides a `--print-reduced-precision-fp` flag to enable the legacy behaviour.~~
This PR takes the approach of mlir-opt: attempt to print to scientific 6dp notation iff the resulting string is losslessly reproducible, or else print to full precision using repr. This has the advantage of not affecting a vast number of filechecks, mirroring upstream, and not having a flag for either full or reduced precision.


alexarice:
I think I've missed a bit of discussion, what's wrong with having this as the default?

n-io:
> I think I've missed a bit of discussion, what's wrong with having this as the default?

I completely agree, but having this as a default requires changing a vast number of tests. Also, I suspect that full-precision is not actually very good for cross-system reproducible file checks, as this is currently breaking on the server but passing on my system.

My preference would be to have a reduced-precision flag and use this on file checks, with full-precision mode being the default.

n-io:
I've changed the default behaviour to print with full precision.

tests/filecheck/mlir-conversion/with-mlir/print-fp-precision.mlir
```diff
@@ -0,0 +1,17 @@
...
+builtin.module {
+
+// CHECK:            %0 = arith.constant 3.141593e+00 : f64
+// FULL-PRECISION:   %0 = arith.constant 3.141592653589793 : f64
+// MLIR-AND-BACK:    %0 = arith.constant 3.141592653589793 : f64
+
+%0 = arith.constant 3.141592653589793 : f64
+
+"test.op"(%0) : (f64) -> ()
+}
```
superlopuh, line 17:
```suggestion
// CHECK:            %0 = arith.constant 3.141593e+00 : f64
// FULL-PRECISION:   %0 = arith.constant 3.141592653589793 : f64
// MLIR-AND-BACK:    %0 = arith.constant 3.141592653589793 : f64
%0 = arith.constant 3.141592653589793 : f64
```

tests/filecheck/parser-printer/print-fp-precision.mlir
```diff
@@ -0,0 +1,17 @@
...
+builtin.module {
+
+// REDUCED-PRECISION:  %0 = arith.constant 3.141593e+00 : f64
+// FULL-PRECISION:     %0 = arith.constant 3.141592653589793 : f64
+// ROUNDTRIP:          %0 = arith.constant 3.141592653589793 : f64
+
+%0 = arith.constant 3.141592653589793 : f64
+
+"test.op"(%0) : (f64) -> ()
+}
```
superlopuh, line 17:
```suggestion
// REDUCED-PRECISION:  %0 = arith.constant 3.141593e+00 : f64
// FULL-PRECISION:     %0 = arith.constant 3.141592653589793 : f64
// ROUNDTRIP:          %0 = arith.constant 3.141592653589793 : f64
%0 = arith.constant 3.141592653589793 : f64
```

tests/filecheck/mlir-conversion/with-mlir/print-fp-precision.mlir
superlopuh, line 1: what's the purpose of this check? If it's to check roundtrip with MLIR, shouldn't the first two checks also check via MLIR? Otherwise we can just delete it?

xdsl/xdsl_opt_main.py
```diff
@@ -164,6 +164,13 @@ def register_all_arguments(self, arg_parser: argparse.ArgumentParser):
             help="Print operations with debug info annotation, such as location.",
         )
 
+        arg_parser.add_argument(
+            "--print-reduced-precision-fp",
+            default=False,
+            action="store_true",
+            help="Prints floating point literals with reduced precision.",
+        )
+
```
superlopuh, line 173: I'm really not sure that this is a positive addition to the framework. I don't see why one would want to print reduced precision if the output isn't nicer. I'm happy to help with this PR, as it will will take some time, but I would feel much better if we didn't merge this flag.

tests/filecheck/mlir-conversion/with-mlir/print-fp-precision.mlir
n-io: removed the filecheck.

tests/filecheck/parser-printer/print-fp-precision.mlir
n-io: removed the filecheck.

tests/filecheck/mlir-conversion/with-mlir/print-fp-precision.mlir
n-io: removed the flag and the filecheck.

n-io:
Here's a minimally intrusive suggestion, which takes the same approach of mlir-opt: attempt to print to scientific 6dp notation iff the resulting string is losslessly reproducible, or else print to full precision using `repr`. This has the advantage of not affecting a vast number of filechecks, mirroring upstream, and not having a flag for either full or reduced precision.

superlopuh (approved):
Thank you!

alexarice (approved):
Looks good.

I feel I would still personally prefer that "simple" floats print in non scientific format, but I guess this is just too large of a filecheck diff

n-io:
> Looks good.
> 
> I feel I would still personally prefer that "simple" floats print in non scientific format, but I guess this is just too large of a filecheck diff

It can always be done in the future. I'm not sure about the performance of parsing the str back to a float to ensure there is no precision loss, and printing it straight through `repr` seems perfectly fine in my opinion. Though it seemed a quick way to ensure not losing precision, which was causing issues to us.

PR #3382, opened 2024-11-02
API: add back traits_def, and also lazy_traits_def

superlopuh:
Fixing things


math-fehr (approved):
Thanks a lot for fixing this!

PR #3383, opened 2024-11-03
transformations: Convert `memref` to `ptr`dialect

mamanain:
Introducing lowering of some memref operations to a ptr dialect. This will be used to translate riscv compilation to ptr dialect in the future (this is the second pr in the series).

### DIFF
--- a/xdsl/transforms/convert_memref_to_ptr.py
+++ b/xdsl/transforms/convert_memref_to_ptr.py
@@ -17,10 +17,11 @@ from xdsl.pattern_rewriter import (
 from xdsl.utils.exceptions import DiagnosticException
 
 
-# I think we also need to pass the adress width.
 def offset_calculations(
     memref_type: memref.MemRefType[Any], indices: Iterable[SSAValue]
 ) -> tuple[list[Operation], SSAValue]:
+    """Get operations calculating an offset which needs to be added to memref's base pointer to access an element referenced by indices."""
+
     assert isinstance(memref_type.element_type, builtin.FixedBitwidthType)
 
     match memref_type.layout:
@@ -99,6 +100,8 @@ def get_target_ptr(
     memref_type: memref.MemRefType[Any],
     indices: Iterable[SSAValue],
 ) -> tuple[list[Operation], SSAValue]:
+    """Get operations returning a pointer to an element of a memref referenced by indices."""
+
     ops: list[Operation] = [
         memref_ptr := ptr.ToPtrOp(
             operands=[target_memref], result_types=[ptr.PtrType()]
--- a/xdsl/transforms/convert_memref_to_ptr.py
+++ b/xdsl/transforms/convert_memref_to_ptr.py
@@ -1,0 +1,164 @@
+from collections.abc import Iterable
+from dataclasses import dataclass
+from typing import cast
+
+from xdsl.context import MLContext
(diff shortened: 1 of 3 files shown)


xdsl/transforms/convert_memref_to_ptr.py
```diff
@@ -0,0 +1,164 @@
...
+    memref_type: memref.MemRefType[Any],
+    indices: Iterable[SSAValue],
+) -> tuple[list[Operation], SSAValue]:
+    ops: list[Operation] = [
+        memref_ptr := ptr.ToPtrOp(
+            operands=[target_memref], result_types=[ptr.PtrType()]
+        )
+    ]
+
+    if len(list(indices)) == 0:
```
superlopuh, line 108:
```suggestion
    if not indices:
```

superlopuh: It looks like the indices passed below are actually a Collection, so this should be possible, and would avoid an extra list construction

xdsl/transforms/convert_memref_to_ptr.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@dataclass(frozen=True)
+class ConvertMemrefToPtr(ModulePass):
+    name = "convert-memref-to-ptr"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        the_one_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier([ConvertStoreOp(), ConvertLoadOp()]),
+            apply_recursively=False,
```
superlopuh, line 160: Is this necessary?

superlopuh: Similarly for the other options below

xdsl/transforms/convert_memref_to_ptr.py
```diff
@@ -0,0 +1,161 @@
...
+    GreedyRewritePatternApplier,
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.utils.exceptions import DiagnosticException
+
+
+# I think we also need to pass the adress width.
```
superlopuh, line 20: Isn't it computed below?

superlopuh (approved):
Could you please add some doc strings to the helpers? LGTM otherwise

xdsl/transforms/convert_memref_to_ptr.py
mamanain: yeah, forgot to delete this comment from a previous version

xdsl/transforms/convert_memref_to_ptr.py
```diff
@@ -0,0 +1,164 @@
...
+
+    assert isinstance(memref_type.element_type, builtin.FixedBitwidthType)
+
+    match memref_type.layout:
+        case builtin.NoneAttr():
+            strides = builtin.ShapedType.strides_for_shape(memref_type.get_shape())
+        case builtin.StridedLayoutAttr():
+            strides = memref_type.layout.get_strides()
+        case _:
+            raise DiagnosticException(f"Unsupported layout type {memref_type.layout}")
```
superlopuh, line 33: Oh I forgot about this, could you please add a small PR with a test that catches this diagnostic exception?

PR #3384, opened 2024-11-03
misc: move pdl, sgpu, and wgsl_printer out of experimental

superlopuh:
These are actually all fairly well-tested, so I think the experimental bit is confusing. Some of the code would probably benefit from being in a different place, but we can worry about that another time.


tests/interpreters/test_wgpu.py
```diff
@@ -13,7 +13,7 @@
 
 pytest.importorskip("wgpu", reason="wgpu is an optional dependency")
 
-from xdsl.interpreters.experimental.wgpu import WGPUFunctions  # noqa: E402
+from xdsl.interpreters.wgpu import WGPUFunctions  # noqa: E402
```
alexarice, line 16: What's this lint for? Realise it's unchanged but could there be an explanation in the comment

alexarice (approved):
Seems reasonable to me, though I have no context for why they were separated to begin with

alexarice:
Also fixed titles, was slightly confused on how you were manipulating pdfs with xdsl

PR #3389, opened 2024-11-04
transformations: `ptr` dialect to `riscv`, run snitch compilation through `ptr` dialect

mamanain:
Third part of integrating ptr dialect into snitch compilation pipeline. Here we:
* Introduce ptr -> riscv conversion
* Check that loop folding works on pointer arithmetic
* Integrate ptr conversions to microkernels' compilation pipeline


mamanain:
Splitted this pr to two.

PR #3390, opened 2024-11-04
dialects: (llvm) added the 'exact' keyword to the corresponding operations

lfrenot:


tobiasgrosser:
@lfrenot, can you rebase to ensure this PR has a clean history?

tobiasgrosser:
You still seem to have unrelated changes in the commits, no?

xdsl/dialects/llvm.py
```diff
@@ -474,6 +475,64 @@ def print(self, printer: Printer) -> None:
...
 
+class ArithmeticBinOpExact(IRDLOperation, ABC):
+    """Class for arithmetic binary operations that use an exact flag."""
+
+    T: ClassVar = VarConstraint("T", BaseAttr(IntegerType))
+
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+    res = result_def(T)
+    isExact = opt_prop_def(BoolAttr)
```
superlopuh, line 486: This is the bridge we use to make sure that our Python code conforms to snake_case and that the interop with MLIR works, I'd recommend using it in this case:
```suggestion
    is_exact = opt_prop_def(BoolAttr, prop_name="isExact")
```

superlopuh (approved):
LGTM modulo prop_name. @tobiasgrosser, which are the extra files you mean?

xdsl/dialects/llvm.py
```diff
@@ -474,6 +475,64 @@ def print(self, printer: Printer) -> None:
...
 
+class ArithmeticBinOpExact(IRDLOperation, ABC):
+    """Class for arithmetic binary operations that use an exact flag."""
+
+    T: ClassVar = VarConstraint("T", BaseAttr(IntegerType))
+
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+    res = result_def(T)
+    is_exact = opt_prop_def(BoolAttr, prop_name="isExact")
```
superlopuh, line 486: Oh, one more thing, is that I don't think it makes sense for this property to be optional, isn't it the same as default 0?
```suggestion
    is_exact = prop_def(BoolAttr, prop_name="isExact")
```

superlopuh: also, `default_value` might be useful here

tobiasgrosser:
> LGTM modulo prop_name. @tobiasgrosser, which are the extra files you mean?

No extra files, but https://github.com/xdslproject/xdsl/pull/3390/commits/6460b2d5327bf97e43daf66b0410180fdca16498 is included in the list of commits here. We should avoid this in the future and start with clean PR histories. Given that this PR is almost closed, we are fine, I guess.

superlopuh:
When we merge in xDSL, the history is deleted, as the merges are rebase+squash, so rebasing is not a pattern I've seen very much.

superlopuh (approved):
One final thought is that it would probably be worth adding the interop tests with MLIR

lfrenot:
These wouldn't work right now, since exact flags are not in MLIR yet
I've started looking into it

tobiasgrosser:
> When we merge in xDSL, the history is deleted, as the merges are rebase+squash, so rebasing is not a pattern I've seen very much.

Sure, but PR start typically with a clean history and not with some random other PRs being part of it.

lfrenot:
@tobiasgrosser are we good to merge?

tobiasgrosser:
Yes, good to merge if @superlopuh is happy.

tobiasgrosser (approved):

PR #3391, opened 2024-11-04
transformations: (pdl) Apply PDL patterns from source

qaco:


tests/filecheck/interpreters/test_pdl_interpreter_extra_file.mlir
superlopuh, line 1: This seems redundant, if I'm not missing anything

xdsl/transforms/apply_pdl.py
```diff
@@ -0,0 +1,43 @@
...
+            assert os.path.exists(self.pdl_file)
+            with open(self.pdl_file) as f:
+                pdl_module_str = f.read()
+                parser = Parser(ctx, pdl_module_str)
+                pdl_module = parser.parse_module()
+        else:
+            pdl_module = payload_module
+        stream = StringIO()
+        rewrite_patterns = [
+            cast(RewritePattern, PDLRewritePattern(op, ctx, stream))
```
superlopuh, line 38: is this cast necessary?

superlopuh, line 38:
```suggestion
            cast(RewritePattern, PDLRewritePattern(op, ctx, None))
```

qaco: Yeah because ```list``` is invariant so ```list[PDLRewritePattern]``` cannot be seen as ```list[RewritePattern]``` (```pyright``` error)

superlopuh: Ah, but if you put it inline, it will work, right?

qaco: > Ah, but if you put it inline, it will work, right?

Right ^^ It is done !

superlopuh (approved):

PR #3392, opened 2024-11-04
transformations: check invalid cases in `memref` -> `ptr` conversion

mamanain:
Add missing test cases.


superlopuh (approved):
Thank you!

PR #3393, opened 2024-11-04
transformations: `ptr` -> `riscv` conversion

mamanain:
Part 3 of introducing `ptr` dialect to snitch compilation. Pass to lower `ptr` to `riscv`.

### DIFF
--- a/tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
+++ b/tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
@@ -22,5 +22,4 @@ ptr_xdsl.store %v, %p : i32, !ptr_xdsl.ptr
 %m2 = "test.op"() : () -> (memref<3x2xf128>)
 %p2 = ptr_xdsl.to_ptr %m2 : memref<3x2xf128> -> !ptr_xdsl.ptr
 %v1 = ptr_xdsl.load %p2 : !ptr_xdsl.ptr -> f128
-
 // CHECK: Unexpected floating point type f128
--- a/tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
+++ b/tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
@@ -1,0 +1,19 @@
+// RUN: xdsl-opt %s -p convert-ptr-to-riscv --split-input-file | filecheck %s
+
+%m, %idx, %v = "test.op"() : () -> (memref<3x2xi32>, index, i32)
+
+%p = ptr_xdsl.to_ptr %m : memref<3x2xi32> -> !ptr_xdsl.ptr
+// CHECK: %p = builtin.unrealized_conversion_cast %m : memref<3x2xi32> to index
+
+%r0 = ptr_xdsl.ptradd %p, %idx : (!ptr_xdsl.ptr, index) -> !ptr_xdsl.ptr
+// CHECK-NEXT:  %r0 = arith.addi %p, %idx : index
+
+ptr_xdsl.store %v, %p : i32, !ptr_xdsl.ptr
+// CHECK-NEXT:  %p_1 = builtin.unrealized_conversion_cast %p : index to !riscv.reg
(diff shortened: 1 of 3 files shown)


tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: xdsl-opt %s -p convert-ptr-to-riscv --split-input-file | filecheck %s
+
+%m, %idx, %v = "test.op"() : () -> (memref<3x2xi32>, index, i32)
+
+%p = ptr_xdsl.to_ptr %m : memref<3x2xi32> -> !ptr_xdsl.ptr
+// CHECK: %p = builtin.unrealized_conversion_cast %m : memref<3x2xi32> to index
```
superlopuh, line 6: this should lower to a cast to riscv.reg

tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
```diff
@@ -0,0 +1,19 @@
...
+
+%r0 = ptr_xdsl.ptradd %p, %idx : (!ptr_xdsl.ptr, index) -> !ptr_xdsl.ptr
+// CHECK-NEXT:  %r0 = arith.addi %p, %idx : index
```
superlopuh, line 9: This should lower to casts and riscv add assembly op

xdsl/transforms/convert_ptr_to_riscv.py
```diff
@@ -0,0 +1,133 @@
...
+                        )
+                    case Float64Type():
+                        new_op = riscv.FSdOp(
+                            addr,
+                            value,
+                            0,
+                            comment="store double value to pointer",
+                        )
+                    case _:
+                        assert False, f"Unexpected floating point type {float_type}"
```
superlopuh, line 72:
```suggestion
                        raise DiagnosticException(f"Unexpected floating point type {float_type}")
```
This should raise a diagnostic exception, and we should add a filecheck test to check that we print a helpful string.

xdsl/transforms/convert_ptr_to_riscv.py
```diff
@@ -0,0 +1,133 @@
...
+                lw_op = riscv.LwOp(addr, 0, comment="load word from pointer")
+            case riscv.FloatRegisterType:
+                float_type = cast(AnyFloat, op.res.type)
+                match float_type:
+                    case Float32Type():
+                        lw_op = riscv.FLwOp(addr, 0, comment="load float from pointer")
+                    case Float64Type():
+                        lw_op = riscv.FLdOp(addr, 0, comment="load double from pointer")
+                    case _:
+                        assert False, f"Unexpected floating point type {float_type}"
```
superlopuh, line 100: similarly here, would be good to add a test here, since we know that there are other existing floating point types

xdsl/transforms/convert_ptr_to_riscv.py
```diff
@@ -0,0 +1,133 @@
...
+
+            case _:
+                assert False, f"Unexpected register type {result_register_type}"
+
+        rewriter.replace_matched_op(
+            [lw := lw_op, UnrealizedConversionCastOp.get(lw.results, (op.res.type,))]
+        )
```
superlopuh, line 107:
```suggestion
        rewriter.replace_matched_op(
            (lw := lw_op, UnrealizedConversionCastOp.get(lw.results, (op.res.type,)),)
        )
```

tests/filecheck/backend/riscv/convert_ptr_to_riscv.mlir
superlopuh: I would say no, that it's ok to repeat in this way

superlopuh: the way not to repeat in this case would be to split the pass into two, one that lowers to arith, the other to riscv, but I don't think this is worth it in this case

superlopuh (approved):

PR #3394, opened 2024-11-04
transformations: convert type_offsets in `ptr` to `arith.constant`

mamanain:
It seems like having a separate test for this is a little wasteful, so I decided to test it in loop folding.


tests/filecheck/mlir-conversion/with-mlir/ptr_loop_folding.mlir
```diff
@@ -0,0 +1,90 @@
+// RUN: xdsl-opt -p convert-memref-to-ptr,convert-ptr-type-offsets --print-op-generic --split-input-file %s | mlir-opt --allow-unregistered-dialect --mlir-print-op-generic -pass-pipeline 'builtin.module(func.func(scf-for-loop-canonicalization,scf-for-loop-range-folding,scf-for-loop-canonicalization))' | xdsl-opt -p scf-for-loop-flatten --print-op-generic | mlir-opt --allow-unregistered-dialect --mlir-print-op-generic -pass-pipeline 'builtin.module(func.func(scf-for-loop-canonicalization,scf-for-loop-range-folding,scf-for-loop-canonicalization))'
```
superlopuh, line 1: could you please try using the mlir opt pass in xDSL? should simplify the pipeline a little bit

superlopuh:
It would be great to add an individual filecheck test for the pass also

tests/filecheck/mlir-conversion/with-mlir/ptr_loop_folding.mlir
```diff
@@ -0,0 +1,79 @@
+// RUN: xdsl-opt -p convert-memref-to-ptr,convert-ptr-type-offsets,mlir-opt[scf-for-loop-canonicalization,scf-for-loop-range-folding,scf-for-loop-canonicalization],scf-for-loop-flatten,mlir-opt[scf-for-loop-canonicalization,scf-for-loop-range-folding,scf-for-loop-canonicalization] --split-input-file %s | filecheck %s
```
superlopuh, line 1: How much are we missing in xDSL from the mlir-opt pipeline here? If it's not a lot, I'd much rather have this logic in xDSL.

mamanain: I believe both `scf-for-loop-canonicalization` and `scf-for-loop-range-folding` are missing. Do you think it's worth porting them?

superlopuh: scf-for-loop-range-folding is already in, and was less buggy than the MLIR one until recently, where I upstreamed a bug fix to MLIR after noticing the difference with xDSL :)

mamanain: How about we bench it first? I'm just curious to see if it makes any improvements. Plus, if there is no speed increase, maybe the port won't be worth it.

mamanain: Oh, right, there is. But I think it's only for the riscv loops. Do we want a general scf version?

superlopuh: I can guarantee speed improvements, and register pressure improvements in the final assembly, if that's what you mean

superlopuh: yep, I think it would be worth having an scf version for this

mamanain: ok, I'll port them to xdsl

superlopuh: Do we now have everything we need?

superlopuh (approved):
beautiful, thank you

xdsl/transforms/convert_ptr_type_offsets.py
```diff
@@ -0,0 +1,37 @@
...
+                "Type offset is currently only supported for fixed size types"
+            )
+        elem_type = cast(FixedBitwidthType, op.elem_type)
+        rewriter.replace_matched_op(
+            arith.ConstantOp.from_int_and_width(elem_type.size, IndexType())
+        )
+
+
+class ConvertPtrTypeOffsetsPass(ModulePass):
+    name = "convert-ptr-type-offsets"
```
compor, line 32: A simple docstring wouldn't hurt here

compor (approved):

PR #3395, opened 2024-11-04
transformations: Persist func arg names as arg_attr

n-io:
Persisting function arg names to arg attributes as suggested [here](https://discourse.llvm.org/t/mlir-capi-how-to-set-function-argument-names-arg0/3892/2).

Note, upstream mlir will say `op arguments may only have dialect attributes`, requiring arg attribute names to be prefixed with a dialect.

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -257,6 +257,7 @@ async def test_buttons():
         with ImplicitBuilder(expected_module.body):
             function = func.FuncOp("hello", ((index,), (index,)))
             with ImplicitBuilder(function.body) as (n,):
+                n.name_hint = "n"
                 two = arith.Constant(IntegerAttr(2, index)).result
                 res = arith.Muli(n, two)
                 func.Return(res)
--- a/xdsl/transforms/function_transformations.py
+++ b/xdsl/transforms/function_transformations.py
@@ -13,6 +13,9 @@ from xdsl.pattern_rewriter import (
 class ArgNamesToArgAttrsPass(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
+        if not any(arg.name_hint for arg in op.args):
+            return
+
         arg_attrs = (
             op.arg_attrs.data
             if op.arg_attrs is not None
--- a/xdsl/transforms/function_transformations.py
+++ b/xdsl/transforms/function_transformations.py
@@ -1,0 +1,37 @@
+from xdsl.context import MLContext
+from xdsl.dialects import builtin, func
+from xdsl.dialects.builtin import ArrayAttr, DictionaryAttr, StringAttr
+from xdsl.passes import ModulePass
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class ArgNamesToArgAttrsPass(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
+        arg_attrs = list(op.arg_attrs or (len(op.args) * [DictionaryAttr({})]))
+        new_arg_attrs: list[DictionaryAttr] = []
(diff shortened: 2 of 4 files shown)


xdsl/transforms/function_transformations.py
```diff
@@ -0,0 +1,37 @@
...
+            if arg.name_hint:
+                d_attr = DictionaryAttr(
+                    {"llvm.name": StringAttr(arg.name_hint), **arg_attr.data}
+                )
+            else:
+                d_attr = arg_attr
+            new_arg_attrs.append(d_attr)
+
+        op.arg_attrs = ArrayAttr(new_arg_attrs)
+        rewriter.has_done_action = True
```
superlopuh, line 28: - [ ] It feels like the first thing to do would be to check whether any of the args aren't anonymous, and to bail out early if that's the case. Could you add a test case for this, also?
- [ ] could you also check whether the old attrs are the same as the new attrs, and avoid has_done_action if they are the same?
- [ ] nit: it would be nice to refactor this to a tuple comprehension instead of appending to a list, just for style reasons

n-io: I had it as a tuple-comp, and after refactoring it am glad that is encouraged 😃

superlopuh: I'm a fan :)

xdsl/transforms/function_transformations.py
```diff
@@ -0,0 +1,47 @@
...
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+class ArgNamesToArgAttrsPass(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
+        arg_attrs = list(op.arg_attrs or (len(op.args) * [DictionaryAttr({})]))
```
superlopuh, line 16:
```suggestion
        arg_attrs = op.arg_attrs.data if op.arg_attrs is not None else or (DictionaryAttr({}),) * len(op.args)
```

xdsl/transforms/function_transformations.py
```diff
@@ -0,0 +1,47 @@
...
+
+        new_arg_attrs = ArrayAttr(
+            DictionaryAttr(
+                {"llvm.name": StringAttr(arg.name_hint), **arg_attr.data}
+                if arg.name_hint
+                else arg_attr.data
+            )
+            for arg, arg_attr in zip(op.args, arg_attrs)
```
superlopuh, line 24:
```suggestion
            for arg, arg_attr in zip(op.args, arg_attrs, strict=True)
```

xdsl/transforms/function_transformations.py
superlopuh, line 16: Would it not be worth doing something like this?
```suggestion
    def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
        if not any(arg.name_hint for arg in op.args):
            return
        arg_attrs = list(op.arg_attrs or (len(op.args) * [DictionaryAttr({})]))
```
This way there's no intermediate thing created if there's nothing to do

n-io: `else or`

n-io: At the cost of an extra pass over the args? I actually didn't think that's what we wanted.

n-io: done as suggested (minus the 'else or')

tests/interactive/test_app.py
```diff
@@ -257,6 +257,7 @@ def callback(x: str):
         with ImplicitBuilder(expected_module.body):
             function = func.FuncOp("hello", ((index,), (index,)))
             with ImplicitBuilder(function.body) as (n,):
+                n.name_hint = "n"
```
n-io, line 260: This will otherwise show up as a difference in `get_available_passes` below.

xdsl/transforms/function_transformations.py
n-io: Done as suggested.

tests/interactive/test_app.py
superlopuh: Looks like this fixes a kind of bug in the test, which is that the name hints are not the same in the text and this IR

superlopuh (approved):

PR #3396, opened 2024-11-04
bug: Fix cse for FloatData

n-io:
FloatData recently received an `__eq__` method to handle `nan` inequalities. Because of this change, it also needs to provide a `__hash__` method.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -636,6 +636,9 @@ class FloatData(Data[float]):
             math.isnan(self.data) and math.isnan(other.data) or self.data == other.data
         )
 
+    def __hash__(self):
+        return super().__hash__()
+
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/builtin.py
```diff
@@ -636,6 +636,9 @@ def __eq__(self, other: Any):
             math.isnan(self.data) and math.isnan(other.data) or self.data == other.data
         )
 
+    def __hash__(self):
+        return super().__hash__()
```
superlopuh, line 640: Does this work? Would cut out the middle man
```suggestion
        return hash(self.data)
```

superlopuh (approved):

xdsl/dialects/builtin.py
n-io: It does indeed

PR #3399, opened 2024-11-04
transformations: Add further csl name hints

n-io:


xdsl/transforms/lower_csl_wrapper.py
```diff
@@ -311,9 +320,11 @@ def match_and_rewrite(self, op: csl_wrapper.ImportOp, rewriter: PatternRewriter,
         import_ = csl.ImportModuleConstOp(
             op.module, structs[-1] if len(structs) > 0 else None
         )
+        import_.result.name_hint = self._get_name_hint(op)
 
         rewriter.insert_op(ops, InsertPoint.at_start(csl_mod.body.block))
-        rewriter.replace_matched_op([*structs, import_])
+        rewriter.insert_op(structs, InsertPoint.before(op))
+        rewriter.replace_matched_op(import_)
```
n-io, line 327: This is to prevent the structs from having the same name as the buffer, and the buffer turning into `buf_8` for instance, which may be confusing.

xdsl/transforms/lower_csl_wrapper.py
```diff
@@ -303,6 +306,12 @@ def _make_import_struct(self, import_op: csl_wrapper.ImportOp):
             out_structs.append(csl.ConcatStructOp(out_structs[-1], s))
         return out_structs
 
+    def _get_name_hint(self, import_op: csl_wrapper.ImportOp) -> str | None:
```
dk949, line 309: Could this be `Path(import_op.module.data.strip("<>" )).stem`?

dk949 (approved):
I've got a comment, but I'm not 100% it's correct

PR #3400, opened 2024-11-05
transformations: Add lowering of affine if to standard

gabrielrodcanal:
Port of the lowering of affine if in to standard in MLIR: https://github.com/llvm/llvm-project/blob/ea859005b5ddb14548b9dc5b94d54d78754f5918/mlir/lib/Conversion/AffineToStandard/AffineToStandard.cpp#L272.

I will be adding tests soon.


superlopuh:
Let's reopen this when you have the time to get back to this

PR #3402, opened 2024-11-05
transformations: (pdl) Better support for pdl.type and pdl.replace

qaco:


superlopuh (commented):
Could you please split this into three mini PRs, and add unit tests for them? The filecheck tests I find are generally less useful for debugging, as they don't pinpoint the exact point of failure. In this case, the error might be that the matcher doesn't match, and then we would still need to search in the code for where things went wrong. In contrast, unit tests hopefully cover the core parts of functionality for each function, and will help locate the bug immediately.

xdsl/interpreters/pdl.py
```diff
@@ -109,9 +109,11 @@ def match_type(self, ssa_val: SSAValue, pdl_op: pdl.TypeOp, xdsl_attr: Attribute
...
             return self.matching_context[ssa_val] == xdsl_attr
 
-        self.matching_context[ssa_val] = xdsl_attr
-
-        return True
+        if pdl_op.constantType is None or pdl_op.constantType == xdsl_attr:
+            self.matching_context[ssa_val] = xdsl_attr
+            return True
+        else:
+            return False
```
superlopuh, line 116: This is being added in #3405

PR #3403, opened 2024-11-05
transformations: Materialise stores before bufferization

n-io:
This technically performs some (all?) of the work of `csl-stencil-materialize-stores`, but before bufferization. Not sure if instead of this injection I should rewrite the pass?

### DIFF
--- a/xdsl/dialects/csl/csl_stencil.py
+++ b/xdsl/dialects/csl/csl_stencil.py
@@ -393,9 +393,9 @@ class ApplyOp(IRDLOperation):
                     f"Unexpected block argument type of done_exchange, got {arg.type} != {expected_type} at index {arg.index}"
                 )
 
-        if (len(self.res) == 0) == (len(self.dest) == 0):
+        if (len(self.res) != 0) and (len(self.dest) != 0):
             raise VerifyException(
-                "Expected stencil.apply to have either results or dest specified"
+                "Expected stencil.apply to have results or dest specified, but not both"
             )
 
     def get_rank(self) -> int:
--- a/xdsl/dialects/csl/csl_stencil.py
+++ b/xdsl/dialects/csl/csl_stencil.py
@@ -393,9 +393,9 @@ class ApplyOp(IRDLOperation):
                     f"Unexpected block argument type of done_exchange, got {arg.type} != {expected_type} at index {arg.index}"
                 )
 
-        if (len(self.res) != 0) and (len(self.dest) != 0):
+        if (len(self.res) == 0) == (len(self.dest) == 0):
             raise VerifyException(
-                "Expected stencil.apply to have results or dest specified, but not both"
+                "Expected stencil.apply to have either results or dest specified"
             )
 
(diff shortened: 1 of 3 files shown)


xdsl/dialects/csl/csl_stencil.py
```diff
@@ -393,9 +393,9 @@ def verify_(self) -> None:
                     f"Unexpected block argument type of done_exchange, got {arg.type} != {expected_type} at index {arg.index}"
                 )
 
-        if (len(self.res) == 0) == (len(self.dest) == 0):
+        if (len(self.res) != 0) and (len(self.dest) != 0):
```
superlopuh, line 396:
```suggestion
        if self.res and self.dest:
```

n-io: reverting changes to this file

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -379,6 +381,92 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+                linalg_op,
+                type(linalg_op).build(
+                    operands=[linalg_op.inputs, [extract_slice_op.result]],
+                    result_types=linalg_op.result_types,
+                    regions=[linalg_op.detach_region(r) for r in linalg_op.regions],
+                    properties=linalg_op.properties,
+                    attributes=linalg_op.attributes,
+                ),
+            )
+        if len(additional_args) > 0:
```
superlopuh, line 449:
```suggestion
        if additional_args:
```

superlopuh: Using `len` to check for whether something is empty is recommended against, as it can potentially be O(n)

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -379,6 +381,92 @@ def match_and_rewrite(self, op: arith.Constant, rewriter: PatternRewriter, /):
...
+            arg = op.done_exchange.block.insert_arg(
+                arg.type, len(op.done_exchange.block.args)
+            )
+            arg_to_tensor = to_tensor_op(arg, writable=True)
+            symmetric_offsets = tuple(
+                (s - d) // 2
+                for s, d in zip(
+                    arg_t.get_shape(), yld_arg.type.get_shape(), strict=True
+                )
+            )
```
dk949, line 423: Could you please put a note here? I'm not 100% sure I understand why we do `(s - d) // 2`

dk949 (commented):
Looks good :+1:

xdsl/transforms/csl_stencil_bufferize.py
n-io: This calculates the offset of the core data, assuming a symmetric (i.e. even) number of ghost cells in each direction. Done some renaming and added a comment as suggested.

dk949 (approved):
Thank you

PR #3405, opened 2024-11-06
transformations: (pdl) Different type matching between polymorphic and fixed case

qaco:


superlopuh (commented):
could you please add a pytest unit test?

qaco:
> could you please add a pytest unit test?

Ok done !

tests/interpreters/test_pdl_interpreter.py
```diff
@@ -482,6 +482,65 @@ def even_length_string(attr: Attribute) -> bool:
...
+        with ImplicitBuilder(pdl.PatternOp(2, None).body):
+            pdl_type = pdl.TypeOp(IntegerType(32)).result
+            pdl_op = pdl.OperationOp(StringAttr("test.op"), type_values=[pdl_type]).op
+            with ImplicitBuilder(pdl.RewriteOp(pdl_op).body):
+                pdl.EraseOp(pdl_op)
+
+    pdl_rewrite_op = next(
+        op for op in pdl_module.walk() if isinstance(op, pdl.RewriteOp)
+    )
+
```
superlopuh, line 536: I apologise, I should have been more precise! I didn't mean for you to do all this work to reimplement the inputs etc. In my head, it was a much smaller function, and I didn't realise that there was no precedent already in this file for it. Here is what I was hoping we would already have in this file:
``` python

def test_match_type():
    matcher = PDLMatcher()

    pdl_op = pdl.TypeOp()
    ssa_value = pdl_op.result
    xdsl_value = StringAttr("a")

    # New value
    assert matcher.match_type(ssa_value, pdl_op, xdsl_value)
    assert matcher.matching_context == {ssa_value: xdsl_value}

    # Same value
    assert matcher.match_type(ssa_value, pdl_op, xdsl_value)
    assert matcher.matching_context == {ssa_value: xdsl_value}

    # Other value
    assert not matcher.match_type(ssa_value, pdl_op, StringAttr("b"))
    assert matcher.matching_context == {ssa_value: xdsl_value}
```
I was thinking that it would be just 5 minutes work to add the new test cases for the functionality you added. As it stands these tests are duplicating the functionality tested in the filecheck test, which is probably the better place for them now that you've added the filecheck functionality.

I don't want to block you any more, so if you would like to merge as-is, I think that would also be ok, but if you have the time to replace these two tests with a more targeted test like the one above that just checks the one function you modified I think it would help debug in the future.

superlopuh (approved):

tests/interpreters/test_pdl_interpreter.py
qaco: Oh, ok, I didn't get what you requested. That's fine, I'll do it !

qaco: Ok it's done I guess

superlopuh (approved):
Thank you!

qaco:
> Thank you!

Dos it mean that I do the merge ?

superlopuh:
Yep, all you need is one green tick, and no leftover comments from the review

PR #3407, opened 2024-11-07
transformations: New test-add-timers-to-top-level-funcs pass

n-io:
Adds `timer_start()` and `timer_end()` calls at the beginning and end of top-level functions, intended for testing and benchmarking purposes.

### DIFF
--- a/tests/interactive/test_get_all_available_passes.py
+++ b/tests/interactive/test_get_all_available_passes.py
@@ -62,6 +62,4 @@ def test_get_all_available_passes():
         rewrite_by_names_dict=individual_rewrite.REWRITE_BY_NAMES,
     )
 
-    for e in expected_res:
-        assert e in res
-    assert len(res) < len(all_passes)
+    assert res == expected_res
--- a/tests/interactive/test_get_all_available_passes.py
+++ b/tests/interactive/test_get_all_available_passes.py
@@ -62,4 +62,6 @@ def test_get_all_available_passes():
         rewrite_by_names_dict=individual_rewrite.REWRITE_BY_NAMES,
     )
 
-    assert res == expected_res
+    for e in expected_res:
+        assert e in res
+    assert len(res) < len(all_passes)
--- a/xdsl/transforms/function_transformations.py
+++ b/xdsl/transforms/function_transformations.py
@@ -12,6 +12,7 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 from xdsl.rewriter import InsertPoint
+from xdsl.traits import SymbolTable
 
 
 class ArgNamesToArgAttrsPass(RewritePattern):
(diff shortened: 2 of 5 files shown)


tests/interactive/test_get_all_available_passes.py
```diff
@@ -62,4 +62,6 @@ def test_get_all_available_passes():
         rewrite_by_names_dict=individual_rewrite.REWRITE_BY_NAMES,
     )
 
-    assert res == expected_res
+    for e in expected_res:
+        assert e in res
+    assert len(res) < len(all_passes)
```
n-io, line 67: Comparing a dynamic list (available passes) to a fixed list (expected available passes) results in test failure when adding passes. Possible solutions:
1. Make both available and expected available passes fixed lists, e.g. by filtering `all_passes` to only contain passes known at the time of creating the test.
2. Make both available and expected available passes dynamic lists, e.g. by testing that each available pass actually has an effect.
3. Soften the check (shown above).

n-io: @superlopuh

superlopuh: I disagree with your change, and agree with your sentiment, if you have the bandwidth to fix this test then it would be welcome. In the meantime, could you please update the expected pass list instead?

superlopuh (changes requested):

tests/interactive/test_get_all_available_passes.py
n-io: I disagree with the method of testing this and cannot offer the change requested.

superlopuh: I'll just fix the test today, it's been made much easier recently since Alex fixed the underlying infrastructure, thanks for providing a good reason to do it now rather than later.

n-io: Thanks, I'm rolling back changes to this file in this PR.

xdsl/transforms/function_transformations.py
```diff
@@ -36,6 +40,70 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
...
+            [builtin.Float64Type()], [builtin.Float64Type()]
+        )
+        start_func = func.FuncOp(TIMER_START, start_func_t, Region([]))
+        end_func = func.FuncOp(TIMER_END, end_func_t, Region([]))
+
+        PatternRewriteWalker(
+            AddBenchTimersPattern(start_func_t, end_func_t), apply_recursively=False
+        ).rewrite_module(op)
+
+        op.body.block.add_ops((start_func, end_func))
```
superlopuh, line 104: I would recommend using SymbolTable APIs to look for functions and insert them:

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py

n-io: Sweet, I like this

xdsl/transforms/function_transformations.py
```diff
@@ -36,6 +40,70 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
...
+    end_func_t: func.FunctionType
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
+        if (
+            not (top_level := op.parent_op())
+            or not isinstance(top_level, builtin.ModuleOp)
+            or top_level.parent
+        ):
+            return
```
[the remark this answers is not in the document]
n-io: We currently don't have a use case for this.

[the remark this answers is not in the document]
n-io: We're looking at annotating top-level functions that are the equivalent of a main function. Should prob add a check that it's not being called. Don't think the SymbolTable API supports this though by any chance?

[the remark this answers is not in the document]
superlopuh: Would the name of the function to annotate be worth adding as a pass parameter? If it works with your flow it might be the easiest way to make it safe by default.

superlopuh (approved):

xdsl/transforms/function_transformations.py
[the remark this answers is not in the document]
n-io: Good suggestion, let me come back to that after the deadline if alright

PR #3411, opened 2024-11-08
transforms: add restrict flag to StencilShapeMinimize pass

emmau678:
Add a restrict flag to the StencilShapeMinimize pass which restricts the grid size of a stencil computation.

### DIFF
--- a/xdsl/transforms/shape_inference.py
+++ b/xdsl/transforms/shape_inference.py
@@ -33,7 +33,7 @@ class ShapeInferencePass(ModulePass):
     name = "shape-inference"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
-        PatternRewriteWalker(ShapeInferenceRewritePattern()).rewrite_module(op)
+        shape_inference_pass_helper(op)
 
 
 def shape_inference_pass_helper(op: builtin.ModuleOp):
--- a/xdsl/transforms/shape_inference.py
+++ b/xdsl/transforms/shape_inference.py
@@ -33,10 +33,10 @@ class ShapeInferencePass(ModulePass):
     name = "shape-inference"
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
-        shape_inference_pass_helper(op)
+        infer_shapes(op)
 
 
-def shape_inference_pass_helper(op: builtin.ModuleOp):
+def infer_shapes(op: builtin.ModuleOp):
     """
     A helper function for ShapeInferencePass which allows it to be called from
     within other passes while exposing the least restrictive API.
--- a/xdsl/transforms/shape_inference.py
+++ b/xdsl/transforms/shape_inference.py
@@ -36,7 +36,7 @@ class ShapeInferencePass(ModulePass):
         PatternRewriteWalker(ShapeInferenceRewritePattern()).rewrite_module(op)
 
 
-def ShapeInferencePassHelper(op: builtin.ModuleOp):
+def shape_inference_pass_helper(op: builtin.ModuleOp):
     """
     A helper function for ShapeInferencePass which allows it to be called from
     within other passes while exposing the least restrictive API.
--- a/xdsl/transforms/shape_inference.py
+++ b/xdsl/transforms/shape_inference.py
@@ -34,3 +34,12 @@ class ShapeInferencePass(ModulePass):
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         PatternRewriteWalker(ShapeInferenceRewritePattern()).rewrite_module(op)
+
+
+def ShapeInferencePassHelper(op: builtin.ModuleOp):
+    """
+    A helper function for ShapeInferencePass which allows it to be called from
+    within other passes while exposing the least restrictive API.
+    """
+
+    PatternRewriteWalker(ShapeInferenceRewritePattern()).rewrite_module(op)
--- a/xdsl/transforms/stencil_shape_minimize.py
+++ b/xdsl/transforms/stencil_shape_minimize.py
@@ -79,11 +79,14 @@ class StencilShapeMinimize(ModulePass):
     restrict: tuple[int, ...] | None = None
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
-        PatternRewriteWalker(InvalidateTemps(), apply_recursively=False).rewrite_module(
-            op
-        )
-        restrict_store_op = RestrictStoreOp(restrict=self.restrict)
-        PatternRewriteWalker(restrict_store_op).rewrite_module(op)
+        if restrict:
+            PatternRewriteWalker(
+                [
+                    InvalidateTemps(),
+                    RestrictStoreOp(restrict=self.restrict),
+                ],
+                apply_recursively=False,
+            ).rewrite_module(op)
         analysis = ShapeAnalysis(seen=set())
         PatternRewriteWalker(analysis).rewrite_module(op)
         bounds = set(
--- a/xdsl/transforms/stencil_shape_minimize.py
+++ b/xdsl/transforms/stencil_shape_minimize.py
@@ -13,7 +13,7 @@ from xdsl.pattern_rewriter import (
     attr_type_rewrite_pattern,
     op_type_rewrite_pattern,
 )
-from xdsl.transforms.shape_inference import ShapeInferencePassHelper
+from xdsl.transforms.shape_inference import shape_inference_pass_helper
 
 
 @dataclass
@@ -92,7 +92,7 @@ class StencilShapeMinimize(ModulePass):
                     ]
                 )
             ).rewrite_module(op)
-            ShapeInferencePassHelper(op)
+            shape_inference_pass_helper(op)
         analysis = ShapeAnalysis(seen=set())
         PatternRewriteWalker(analysis).rewrite_module(op)
         bounds = set(
--- a/xdsl/transforms/stencil_shape_minimize.py
+++ b/xdsl/transforms/stencil_shape_minimize.py
@@ -92,7 +92,7 @@ class StencilShapeMinimize(ModulePass):
                     ]
                 )
             ).rewrite_module(op)
-        ShapeInferencePass().apply(ctx, op)
+            ShapeInferencePass().apply(ctx, op)
         analysis = ShapeAnalysis(seen=set())
         PatternRewriteWalker(analysis).rewrite_module(op)
         bounds = set(
--- a/xdsl/transforms/stencil_shape_minimize.py
+++ b/xdsl/transforms/stencil_shape_minimize.py
@@ -13,6 +13,7 @@ from xdsl.pattern_rewriter import (
     attr_type_rewrite_pattern,
     op_type_rewrite_pattern,
 )
+from xdsl.transforms.shape_inference import ShapeInferencePass
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


superlopuh:
Could you please complete the checklist in the description?

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -50,7 +76,14 @@ class StencilShapeMinimize(ModulePass):
 
     name = "stencil-shape-minimize"
 
+    restrict: tuple[int, ...] | None
```
n-io, line 79:
```suggestion
    restrict: tuple[int, ...] | None = None
```
This would set a default value to not restrict the size.

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -50,7 +76,14 @@ class StencilShapeMinimize(ModulePass):
...
+
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(InvalidateTemps(), apply_recursively=False).rewrite_module(
+            op
+        )
+        restrict_store_op = RestrictStoreOp(restrict=self.restrict)
+        PatternRewriteWalker(restrict_store_op).rewrite_module(op)
```
n-io, line 86:
```suggestion
        if restrict:
            PatternRewriteWalker(
                [
                    InvalidateTemps(),
                    RestrictStoreOp(restrict=self.restrict),
                ],
                apply_recursively=False,
            ).rewrite_module(op)
```
This might be able to work in one pass (not sure though if this needs recursive application). There might also be a `shape-inference` pass inside the conditional, since this invalidates temps.

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -42,6 +50,24 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
             op.update_function_type()
 
 
+@dataclass
+class RestrictStoreOp(RewritePattern):
+    restrict: tuple[int, ...] | None = None
```
n-io, line 55: This class might potentially always have a `restrict`, since it can't do much otherwise.

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -42,6 +50,24 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
...
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.StoreOp, rewriter: PatternRewriter, /):
+        if self.restrict:
+            new_bounds: list[tuple[int, int]] = []
+            for lower_bound, upper_bound, bound_lim in zip(
+                op.bounds.lb, op.bounds.ub, self.restrict
+            ):
+                new_bounds.append(
+                    (min(lower_bound, bound_lim), min(upper_bound, bound_lim))
```
n-io, line 65: This looks really good. It could probably be a simple and beautiful tuple-comprehension.

xdsl/transforms/stencil_shape_minimize.py
emmau678: that makes sense, I'll remove the default None value. thanks!

xdsl/transforms/stencil_shape_minimize.py
emmau678: this code gives a few errors in the IDE:
1. "restrict" is not defined" -> I'll change it to self.restrict
2. 
```
Argument of type "list[InvalidateTemps | RestrictStoreOp]" cannot be assigned to parameter "pattern" of type "RewritePattern" in function "__init__"
  "list[InvalidateTemps | RestrictStoreOp]" is not assignable to "RewritePattern"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```
```
Argument of type "list[InvalidateTemps | RestrictStoreOp]" cannot be assigned to parameter "pattern" of type "RewritePattern" in function "__init__"
  "list[InvalidateTemps | RestrictStoreOp]" is not assignable to "RewritePattern"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```
-> looking at other examples, if I add in an enclosed GreedyRewritePatternApplier and remove the ```apply_recursively=False```, it seems to get rid of this error but I don't fully understand it - I'll add it into a new commit and remove if necessary

xdsl/transforms/stencil_shape_minimize.py
emmau678: thanks! I'll attempt to do this

xdsl/transforms/stencil_shape_minimize.py
n-io: Yes, apologies!

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -42,6 +50,23 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
             op.update_function_type()
 
 
+@dataclass
+class RestrictStoreOp(RewritePattern):
+    restrict: tuple[int, ...]
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.StoreOp, rewriter: PatternRewriter, /):
+        if self.restrict:
```
n-io, line 59:
```suggestion
```
This is no longer needed

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -42,6 +50,23 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
...
+            new_bounds = [
+                (min(lower_bound, bound_lim), min(upper_bound, bound_lim))
+                for lower_bound, upper_bound, bound_lim in zip(
+                    op.bounds.lb, op.bounds.ub, self.restrict
+                )
+            ]
+            new_bounds_attr = stencil.StencilBoundsAttr(new_bounds)
+            self.bounds = new_bounds_attr
```
n-io, line 67: Assigning to `op.bounds` would actually mean the changes pass the rewriter. It may be safer to rebuild the op via `rewriter.replace_matched_op(stencil.StoreOp.get(...))`.

To prevent this from infinitely rewriting the op, it also needs either `apply_recursively=False` in the PatternRewriteWalker, or a check in this class whether this should be applied, for instance, by checking if `new_bounds_attr != op.bounds` to know if this needs updating, and bail out early otherwise.

n-io (commented):
Great, while there are ways to combine this into one filecheck mlir file (if you grep for `--check-prefix`), you could also make a new one. Happy to help with either way.

xdsl/transforms/stencil_shape_minimize.py
emmau678: ok, thanks for that. I'll work on this now

emmau678: I think that because I now have GreedyRewritePatternApplier, I can't have apply_recursively=False.. so I will go with the other option for now but please let me know if I'm missing something simpler

n-io:
> > Great, while there are ways to combine this into one filecheck mlir file (if you grep for `--check-prefix`), you could also make a new one. Happy to help with either way.
> 
> any help on how to get started on this would be greatly appreciated!

I just pushed a filecheck to your branch, but it will likely break. The last thing to do is to run shape inference (`ShapeInferencePass().apply(ctx, op)`) inside the `if self.restrict` conditional to tie everything together. I'll let you do the honours :)

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -91,6 +92,7 @@ def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
                     ]
                 )
             ).rewrite_module(op)
+        ShapeInferencePass().apply(ctx, op)
```
superlopuh, line 95: I would recommend against this

superlopuh: @n-io, what's the motivation for doing this inside the pass and not as part of the pass pipeline?

n-io: The most immediate purpose is to undo the `InvalidateTemps()` rewrite pattern. The pass in general is for testing purposes, to allow restricting the stencil grid size to something small enough that for instance could run in a simulator. The idea is simple: invalidate temps, modify size on store ops, un-invalidate temps. This needs to happen before `stencil-shape-minimize` is run in order for that to work. The only alternative would be to separate this out from shape-minimize altogether, though running `stencil-shape-minimize{restrict=30,30,30}` seems to make some structural sense in my head instead of having it as a fully independent pass.

superlopuh: Ah OK, that makes sense. The main thing that triggered me was calling a pass inside of another pass, for the same reason, as the contract is that a pass leaves the IR in a valid state, if the IR is in a valid state to begin with. In that sense it would make sense to split out the body of the shape inference pass into a helper method, and to call into that from here.

n-io: I've seen this done in a few places, though was never sure why helper methods are preferred, including by `post_walk_func`?

superlopuh: My understanding is that it's preferable to expose the least restrictive API when possible. For example, there's no need to pass the context when performing shape inference. It's probably also to make things more flexible in general, like you may want to only perform shape inference on a subset of your module.

emmau678: I'm not sure I follow - do I need to make some updates here?

superlopuh: Yes, please! Could pull out the body of the shape inference pass into a python function that takes any operation, and no MLContext, and call that helper function from here instead of the pass?

xdsl/transforms/shape_inference.py
```diff
@@ -34,3 +34,12 @@ class ShapeInferencePass(ModulePass):
 
     def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
         PatternRewriteWalker(ShapeInferenceRewritePattern()).rewrite_module(op)
+
+
+def ShapeInferencePassHelper(op: builtin.ModuleOp):
```
superlopuh, line 39: Our coding style is to use snake_case for function names, in this case giving something like this:
```suggestion
def shape_inference_pass_helper(op: builtin.ModuleOp):
```
Could you please move this above the pass definition, and also call it from the pass itself?

superlopuh (approved):
It would probably be good to get @n-io's tick before merging, but LGTM!

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -42,6 +51,27 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter, /):
...
+    restrict: tuple[int, ...]
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: stencil.StoreOp, rewriter: PatternRewriter, /):
+        new_bounds = [
+            (min(lower_bound, bound_lim), min(upper_bound, bound_lim))
+            for lower_bound, upper_bound, bound_lim in zip(
+                op.bounds.lb, op.bounds.ub, self.restrict
+            )
+        ]
```
n-io, line 65: nit: An alternative solution that didn't come to mind earlier would be to build (or pass in) a `restrict_bounds` StencilBoundsAttr and use one of its overload, i.e. `op.bounds & restrict_bounds` to build the new bounds. Feel free to ignore, but wanted to mention.

n-io (approved):
LGTM, feel free to merge!

xdsl/transforms/stencil_shape_minimize.py
emmau678: noted, thanks for this

PR #3412, opened 2024-11-08
dialects: move FastMathFlag and FastMathAttrBase to xdsl.dialects.utils

superlopuh:
Removes the dependency of arith and riscv dialects on llvm, as well as some passes that operate on arith and riscv but not llvm.

### DIFF
--- a/xdsl/dialects/utils.py
+++ b/xdsl/dialects/utils.py
@@ -394,7 +394,7 @@ def parse_dynamic_index_list_without_types(
     return values, indices
 
 
-# Fast Math Flags
+# region Fast Math Flags
 
 
 class FastMathFlag(StrEnum):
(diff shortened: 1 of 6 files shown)


xdsl/dialects/utils.py
```diff
@@ -388,3 +392,30 @@ def parse_dynamic_index_list_without_types(
             values.append(value_or_index)
 
     return values, indices
+
+
+# Fast Math Flags
```
alexarice, line 397: Please remove

alexarice (approved):

xdsl/dialects/utils.py
superlopuh: How do you feel about a region for them instead of a stray comment?

PR #3416, opened 2024-11-09
misc: raise error if variadic results types aren't referenced in assembly format

superlopuh:
The number of results is not passed in when parsing operations. In the generic format, the type of the operation always specifies the types of the results, and `resultSegmentSizes` specifies the ranges of of the results if multiple are variadic. In order to support variadic results, the types an length of all variadic results must be present in the custom syntax.

### DIFF
--- a/xdsl/dialects/experimental/air.py
+++ b/xdsl/dialects/experimental/air.py
@@ -154,7 +154,7 @@ class ChannelPutOp(IRDLOperation):
     src_sizes = var_operand_def(IndexType())
     src_strides = var_operand_def(IndexType())
 
-    async_token = opt_result_def(AsyncTokenAttr())
+    async_token = result_def(AsyncTokenAttr())
 
     irdl_options = [AttrSizedOperandSegments()]
 
@@ -242,7 +242,7 @@ class DmaMemcpyNdOp(IRDLOperation):
     src_sizes = var_operand_def(IndexType())
     src_strides = var_operand_def(IndexType())
 
-    async_token = opt_result_def(AsyncTokenAttr())
+    async_token = result_def(AsyncTokenAttr())
 
     irdl_options = [AttrSizedOperandSegments(as_property=True), ParsePropInAttrDict()]
 
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -238,22 +238,19 @@ class FormatProgram:
             zip(state.result_types, op_def.results, strict=True)
         ):
             if result_type is None:
-                result_type = state.result_types[i]
-                range_length = len(result_type) if isinstance(result_type, list) else 1
-                result_type = result_def.constr.infer(
+                # The number of results is not passed in when parsing operations.
+                # In the generic format, the type of the operation always specifies the
+                # types of the results, and `resultSegmentSizes` specifies the ranges of
+                # of the results if multiple are variadic.
+                # In order to support variadic results, the types an length of all
+                # variadic results must be present in the custom syntax.
+                assert not isinstance(result_def, OptionalDef | VariadicDef)
+                range_length = 1
+                inferred_result_types = result_def.constr.infer(
                     range_length, state.constraint_context
                 )
-                if isinstance(result_def, OptionalDef):
-                    result_type = (
-                        list[Attribute | None]()
-                        if len(result_type) == 0
-                        else result_type[0]
-                    )
-                elif isinstance(result_def, VariadicDef):
-                    result_type = cast(list[Attribute | None], result_type)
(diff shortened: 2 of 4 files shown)


xdsl/dialects/experimental/air.py
superlopuh, line 1: @gabrielrodcanal 
This seemed like the most natural change, since the init always specifies the result type anyway.

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -238,22 +238,19 @@ def resolve_result_types(self, state: ParsingState, op_def: OpDef) -> None:
             zip(state.result_types, op_def.results, strict=True)
         ):
             if result_type is None:
-                result_type = state.result_types[i]
-                range_length = len(result_type) if isinstance(result_type, list) else 1
```
superlopuh, line 242: The `result_type` is already known to be `None`, so couldn't possibly be a `list`

alexarice: This doesn't seem true? `result_type` was set in the line above

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -291,11 +291,13 @@ def verify_results(self, seen_variables: set[str]):
             self.seen_result_types, self.op_def.results, strict=True
         ):
             if not result_type:
-                if not result_def.constr.can_infer(seen_variables):
+                can_infer_type = result_def.constr.can_infer(seen_variables)
+                can_infer_length = not isinstance(result_def, VariadicDef | OptionalDef)
```
alexarice, line 295: This doesn't look correct to me. The length could be inferred by a range variable constraint.

alexarice (changes requested):
I think this change is an oververification and blocks some currently valid definitions

xdsl/irdl/declarative_assembly_format_parser.py
superlopuh: Excellent, how would one go about doing that? apparently we have no examples in the code.

xdsl/irdl/declarative_assembly_format.py
superlopuh: where does `result_type` come from?

alexarice: Ah I see

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: Didn't test this so might have syntax errors but I believe the following should work
```
@irdl_op_definition
class RangeVarOp(IRDLOperation):
    name = "test.range_var"
    T: ClassVar = RangeVarConstraint("T", EqAttrConstraint(IntegerType(1)))
    ins = var_operand_def(T)
    outs = var_result_def(T)

    assembly_format = "$ins attr-dict"
```

superlopuh: I added a test case for it, with a not implemented error. Would you like me to file an issue to track the implementation of length inference?

alexarice: I believe the length inference should already work in this case, am I incorrect?

superlopuh: Exactly, it should be done on the lines you commented on, but the current code "implements" this by taking the length of `None` if `None` is a `list`, which is never the case. It should be possible to infer the information for the case you've provided, but I'm not aware of an implementation of a working mechanism in the code we have.

alexarice: Ok, I've tried to implement this and it seems you are right in that we don't currently have the means to make a working test.

xdsl/dialects/experimental/air.py
```diff
@@ -154,7 +154,7 @@ class ChannelPutOp(IRDLOperation):
     src_sizes = var_operand_def(IndexType())
     src_strides = var_operand_def(IndexType())
 
-    async_token = opt_result_def(AsyncTokenAttr())
+    async_token = result_def(AsyncTokenAttr())
```
alexarice, line 157: Should these be in this PR?

alexarice (approved):

xdsl/dialects/experimental/air.py
superlopuh: These tests fail with my changes. This is because we silently always set variadic (meaning also optional) results to have length 1. Our filecheck passes whether or not the results are present in the CHECK lines, because we match on substrings, so we silently ignore whether there are 0 or 1 results. Without the change on this line, the tests crash with the `NotImplementedError` I added in this PR.

PR #3417, opened 2024-11-09
transformations: (scf) loop range folding

mamanain:
Porting mlir pass of the same name. The logic is almost the same. We need this to keep snitch compilation fully inside xdsl.

### DIFF
--- a/tests/filecheck/transforms/scf_for_loop_range_folding.mlir
+++ b/tests/filecheck/transforms/scf_for_loop_range_folding.mlir
@@ -1,24 +1,24 @@
 // RUN: xdsl-opt -p scf-for-loop-range-folding --split-input-file %s | filecheck %s
 
-// CHECK:       %new_lb = arith.addi %lb, %shift : index
-// CHECK-NEXT:  %new_ub = arith.addi %ub, %shift : index
-// CHECK-NEXT:  %new_lb_1 = arith.muli %new_lb, %mul_shift : index
-// CHECK-NEXT:  %new_ub_1 = arith.muli %new_ub, %mul_shift : index
-// CHECK-NEXT:  %new_step = arith.muli %step, %mul_shift : index
-// CHECK-NEXT:  scf.for %i = %new_lb_1 to %new_ub_1 step %new_step {
+// CHECK:       %shift_idx_lb = arith.addi %lb, %shift : index
+// CHECK-NEXT:  %shift_idx_ub = arith.addi %ub, %shift : index
+// CHECK-NEXT:  %mult_idx_lb = arith.muli %shift_idx_lb, %mul_shift : index
+// CHECK-NEXT:  %mult_idx_ub = arith.muli %shift_idx_ub, %mul_shift : index
+// CHECK-NEXT:  %mult_idx_step = arith.muli %step, %mul_shift : index
+// CHECK-NEXT:  scf.for %i = %mult_idx_lb to %mult_idx_ub step %mult_idx_step {
 // CHECK-NEXT:    "test.op"(%i) : (index) -> ()
 // CHECK-NEXT:  }
 
 %shift, %mul_shift, %lb, %ub, %step = "test.op"() : () -> (index, index, index, index, index)
 scf.for %i = %lb to %ub step %step {
-    %0 = arith.addi %shift, %i : index
-    %1 = arith.muli %0, %mul_shift : index
-    "test.op"(%1) : (index) -> ()
+    %shift_idx = arith.addi %shift, %i : index
+    %mult_idx = arith.muli %shift_idx, %mul_shift : index
+    "test.op"(%mult_idx) : (index) -> ()
 }
 
-// CHECK:       %new_lb_2 = arith.addi %lb, %shift : index
-// CHECK-NEXT:  %new_ub_2 = arith.addi %ub, %shift : index
-// CHECK-NEXT:  scf.for %i_1 = %new_lb_2 to %new_ub_2 step %step {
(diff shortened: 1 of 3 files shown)


tests/filecheck/transforms/scf_for_loop_range_folding.mlir
superlopuh, line 1: could you please replace all memref ops an all arith.constant ops whose results aren't used with test ops?

tests/filecheck/transforms/scf_for_loop_range_folding.mlir
```diff
@@ -0,0 +1,95 @@
+// RUN: xdsl-opt -p scf-for-loop-range-folding --split-input-file %s | filecheck %s
+
+func.func @fold_one_loop(%arg0: memref<?xi32>, %arg1: index, %arg2: index) {
```
superlopuh, line 3: what is the purpose of func op args? maybe could be result of test ops inline also?

tests/filecheck/transforms/scf_for_loop_range_folding.mlir
```diff
@@ -0,0 +1,95 @@
...
+// CHECK-NEXT:  %0 = arith.addi %c0, %arg2 : index
+// CHECK-NEXT:  %1 = arith.addi %arg1, %arg2 : index
+// CHECK-NEXT:  %2 = arith.muli %0, %c4 : index
+// CHECK-NEXT:  %3 = arith.muli %1, %c4 : index
+// CHECK-NEXT:  %4 = arith.muli %c1, %c4 : index
+// CHECK-NEXT:  scf.for %i = %2 to %3 step %4 {
+// CHECK-NEXT:      %5 = memref.load %arg0[%i] : memref<?xi32>
+// CHECK-NEXT:      %6 = arith.muli %5, %5 : i32
+// CHECK-NEXT:      memref.store %6, %arg0[%i] : memref<?xi32>
+// CHECK-NEXT:  }
```
superlopuh, line 29: I've been putting the check lines above the inputs recently, in conformance with the MLIR convention. Could you please make the same shift in this file? I know we're not really consistent here in xDSL

tests/filecheck/transforms/scf_for_loop_range_folding.mlir
mamanain: Yeah, definitely. I just decided to copy tests from mlir. I think they use function arguments to avoid definitions through test.op everywhere. Do you think it's worth changing to test.op?

superlopuh: Probably, and also renaming the values to be a bit more readable, like `%lb`, `%ub` and `%step` for clarity. We definitely don't need the memref though

xdsl/transforms/scf_for_loop_range_folding.py
```diff
@@ -62,8 +62,11 @@ def match_and_rewrite(self, op: scf.For, rewriter: PatternRewriter) -> None:
                             new_step := arith.Muli(op.step, folding_const),
                         ]
                     )
+                    new_step.result.name_hint = "new_step"
                     op.operands[2] = new_step.result
 
+            new_lb.result.name_hint = "new_lb"
+            new_ub.result.name_hint = "new_ub"
```
superlopuh, line 69: This is great, but I have a feeling that it might actually make it easier to read if we keep the name hint as the user's name hint, to make it explicit what calculation we're pulling out

tests/filecheck/transforms/scf_for_loop_range_folding.mlir
```diff
@@ -1,95 +1,56 @@
...
+// CHECK-NEXT:  %new_ub_1 = arith.muli %new_ub, %mul_shift : index
+// CHECK-NEXT:  %new_step = arith.muli %step, %mul_shift : index
+// CHECK-NEXT:  scf.for %i = %new_lb_1 to %new_ub_1 step %new_step {
+// CHECK-NEXT:    "test.op"(%i) : (index) -> ()
+// CHECK-NEXT:  }
+
+%shift, %mul_shift, %lb, %ub, %step = "test.op"() : () -> (index, index, index, index, index)
+scf.for %i = %lb to %ub step %step {
+    %0 = arith.addi %shift, %i : index
+    %1 = arith.muli %0, %mul_shift : index
```
superlopuh, line 15: So wrt my comment below, if we give names to these variables, their names will be visible outside of the for loop in the added arithmetic calculations

superlopuh (approved):

tests/filecheck/transforms/scf_for_loop_range_folding.mlir
mamanain: Could you please check if that's what you meant?

xdsl/transforms/scf_for_loop_range_folding.py
```diff
@@ -46,6 +46,9 @@ def match_and_rewrite(self, op: scf.For, rewriter: PatternRewriter) -> None:
                     return
                 folding_const = user.operands[0]
 
+            new_name_hint = (
+                user.result.name_hint if user.result.name_hint is not None else "new"
+            )
```
superlopuh, line 51: This is cute but I worry that it might be a bit confusing if done extensively, like for the `new_lb_lb` above. I'm now tempted to say that we should not be doing anything here, and let the default name copying  done by the rewriter do the work.

mamanain: I agree, to be honest.

superlopuh: Yep, I would just add the name hints to all variables in the test file, and remove the name hint manipulation in the rewrite

PR #3422, opened 2024-11-11
dialects: (llvm) added overflow flags to trunc

lfrenot:


tobiasgrosser:
There are conflicts to be resolved, also do you want to use the assembly printer?

lfrenot:
This should fix the conflicts
Also, I don't think the assembly printer is usable here because of the overflow flags.

PR #3425, opened 2024-11-11
misc: remove casts in assembly format implementation [NFC]

superlopuh:
I just noticed an abundance of `cast`s, and realized it was partially due to the mutability and hence invariability of `list` being used. When we swap out the `list` for the immutable `tuple` or `Sequence`, Pyright is happier.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-from typing import Any, Literal, cast
+from typing import Literal
 
 from xdsl.dialects.builtin import UnitAttr
 from xdsl.ir import (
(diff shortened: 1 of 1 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -908,22 +899,15 @@ def parse(self, parser: Parser, state: ParsingState) -> None:
...
             ParametrizedAttribute,
         ):
             attr = unique_base.new(unique_base.parse_parameters(parser))
         elif issubclass(unique_base, Data):
-            unique_base = cast(
-                type[Data[Any]],
-                unique_base,
-            )
-            attr = unique_base.new(unique_base.parse_parameter(parser))
+            attr = unique_base.new(unique_base.parse_parameter(parser))  # pyright: ignore[reportUnknownVariableType]
```
alexarice, line 910: Not convinced telling pyright to ignore is better than an explicit cast?

alexarice (approved):

xdsl/irdl/declarative_assembly_format.py
superlopuh: I now feel that telling pyright to ignore `Unknown` specifically is better, as a cast can become outdated with time, and we have no mechanisms to catch this.

PR #3426, opened 2024-11-11
misc: clean up passes getters [NFC]

superlopuh:
Now again in alphabetical order, and with corresponding names.


xdsl/transforms/__init__.py
superlopuh, line 1: I mostly reordered these, renamed the get_ functions when they no longer conformed, and sometimes imported the file name instead of the pass class if ruff split the line

alexarice (approved):

PR #3429, opened 2024-11-12
dialects: (accfg) Update ACCFG dialect to match version in snax-mlir

AntonLydike:
The `accfg` dialect in snax-mlir has diverged slightly from what it is in upstream xDSL. This PR aims to change that so that we can start upstreaming passes from snax-mlir as well.

Some of things that were reconverged:
* Introduction of `launch_fields` instead of a single `launch_addr`
* Introduction of accfg-effect to model side-effects.
* Introduction of custom syntax on the setup op for improved legibility

### DIFF
--- a/xdsl/dialects/accfg.py
+++ b/xdsl/dialects/accfg.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 from collections.abc import Iterable, Sequence
 from enum import Enum
+from typing import cast
 
 from xdsl.dialects.builtin import (
+    AnyIntegerAttr,
     ArrayAttr,
     DictionaryAttr,
     IntegerAttr,
@@ -376,7 +378,9 @@ class AcceleratorOp(IRDLOperation):
 
     launch_fields = prop_def(DictionaryAttr)
 
-    barrier = prop_def(IntegerAttr[IntegerType])  # TODO: this will be reworked in a later version
+    barrier = prop_def(
+        IntegerAttr[IntegerType]
+    )  # TODO: this will be reworked in a later version
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


xdsl/dialects/accfg.py
```diff
@@ -303,16 +281,180 @@ def verify_(self) -> None:
...
+
+    def verify_(self) -> None:
+        for _, val in self.fields.data.items():
+            if not isinstance(val, IntegerAttr):
+                raise VerifyException("fields must only contain IntegerAttr!")
+
+    def field_names(self) -> tuple[str, ...]:
+        return tuple(self.fields.data.keys())
+
+    def field_items(self) -> Iterable[tuple[str, IntegerAttr]]:
```
superlopuh, line 421:
```suggestion
    def field_items(self) -> Iterable[tuple[str, IntegerAttr[Attribute]]]:
```

xdsl/dialects/accfg.py
```diff
@@ -72,117 +105,58 @@ def __init__(self, accelerator: str | StringAttr):
...
     """
 
-    def get_sym_attr_name(self, op: Operation) -> StringAttr | None:
-        assert isinstance(op, AcceleratorOp)
-        return StringAttr(op.name_prop.string_value())
-
+    name = "accfg.launch"
 
-@irdl_op_definition
-class AcceleratorOp(IRDLOperation):
```
AntonLydike, line 86: This op was moved further down the document, sadly the diff isn't very clear here. Can re-arrange if needed

xdsl/dialects/accfg.py
AntonLydike: Doesn't work :(

Do you know what's the best way here?

AntonLydike: Ah, found a fix. Apparently `AnyIntegerAttr` just works

superlopuh (approved):

JosseVanDelm:
Hi @AntonLydike I edited your description a bit thanks for your work!

JosseVanDelm (approved):
Looks good! Thanks!

PR #3430, opened 2024-11-12
misc: use assert_never to enforce exhaustive matches

superlopuh:
Doing some cleanup

### DIFF
--- a/xdsl/dialects/onnx.py
+++ b/xdsl/dialects/onnx.py
@@ -160,16 +160,10 @@ class ElementwiseBinOpBase(IRDLOperation, ABC):
 
     def verify_(self) -> None:
         # Check that the arguments are broadcastable (using Numpy semantics) and that the result type is correct.
-        if (
-            not isinstance(lhs_type := self.lhs.type, TensorType)
-            or not isinstance(rhs_type := self.rhs.type, TensorType)
-            or not isinstance(res_type := self.res.type, TensorType)
-        ):
-            assert False, "onnx elementwise binary operation operands and result must be of type TensorType"
-        lhs_type = cast(TensorType[Attribute], lhs_type)
-        rhs_type = cast(TensorType[Attribute], rhs_type)
-        res_type = cast(TensorType[Attribute], res_type)
-        verify_multidirectional_broadcast_shape(lhs_type, rhs_type, res_type)
+        assert isinstance(lhs_type := self.lhs.type, TensorType)
+        assert isinstance(rhs_type := self.rhs.type, TensorType)
+        assert isinstance(res_type := self.res.type, TensorType)
+        verify_multidirectional_broadcast_shape(lhs_type, rhs_type, res_type)  # pyright: ignore[reportUnknownArgumentType]
 
 
 @irdl_op_definition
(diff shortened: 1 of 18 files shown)


xdsl/dialects/onnx.py
```diff
@@ -276,13 +270,10 @@ def __init__(
...
-            or not isinstance(tensor_c_type := self.tensor_c.type, TensorType)
-            or not isinstance(res_tensor_type := self.res_tensor.type, TensorType)
-        ):
-            assert False, "onnx elementwise operation operands and result must be of type TensorType"
+        assert isinstance(tensor_a_type := self.tensor_a.type, TensorType)
+        assert isinstance(tensor_b_type := self.tensor_b.type, TensorType)
+        assert isinstance(tensor_c_type := self.tensor_c.type, TensorType)
+        assert isinstance(res_tensor_type := self.res_tensor.type, TensorType)
 
         tensor_a_type = cast(TensorType[Attribute], tensor_a_type)
```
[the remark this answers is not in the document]
alexarice: Same in other places

xdsl/transforms/convert_ptr_to_riscv.py
```diff
@@ -91,23 +90,19 @@ def match_and_rewrite(self, op: ptr.LoadOp, rewriter: PatternRewriter, /):
...
+            float_type = cast(AnyFloat, op.res.type)
+            match float_type:
+                case Float32Type():
+                    lw_op = riscv.FLwOp(addr, 0, comment="load float from pointer")
+                case Float64Type():
+                    lw_op = riscv.FLdOp(addr, 0, comment="load double from pointer")
+                case _:
+                    raise DiagnosticException(
+                        f"Unexpected floating point type {float_type}"
+                    )
```
alexarice, line 105: Can the register definitely not take a different type?

xdsl/transforms/lower_affine.py
```diff
@@ -29,34 +29,35 @@ def affine_expr_ops(
...
-            case AffineBinaryOpKind.FloorDiv:
-                op = arith.FloorDivSI(lhs_val, rhs_val)
-            case AffineBinaryOpKind.CeilDiv:
-                op = arith.CeilDivSI(lhs_val, rhs_val)
-
-        return [*lhs_ops, *rhs_ops, op], op.result
-
-    assert False, "Unreachable"
+    match expr:
+        case AffineConstantExpr():
```
alexarice, line 33: Is there a check that this actually works? Seems surprising to me that this does an isinstance check

alexarice (commented):
Seems good, just left a few comments. I'm also not sure what the advantage of `assert_never` over `assert False` is.

xdsl/dialects/onnx.py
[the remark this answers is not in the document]
superlopuh: unfortunately yes, I tried to remove them, but sometimes we actually get the scalar type of the tensor type, and that's a full-on Unknown value, as opposed to an Unknown argument, so can't be pyright-ignored away.

xdsl/transforms/convert_ptr_to_riscv.py
superlopuh: Hmm, this should probably have a different message, as it's more of a not implemented situation, I'll add this.

xdsl/transforms/lower_affine.py
superlopuh: yep plenty of checks, this is the weird syntax python uses for isinstance checks in match statements

alexarice (approved):
I'm still not sure what the advantage of this is but seems ok

superlopuh:
`assert_never` is a Pyright error if the type of the values passed in is not `Never`, so we get type system support that's a bit more explicit. It raises an error at runtime if called, like `assert False`, and is just a bit more explicit about the intent.

PR #3431, opened 2024-11-12
misc: use preferred style for pytest parametrized tests

superlopuh:
I just ran `ruff check -fix --unsafe-fixes`


n-io:
It seems to prefer lists over tuples, I'm not sure I see a value add but have no objections if this is what we wanted.

superlopuh:
I'm not sure why either, but this is the default behaviour, and it seems like a particular version being enforced is better than inconsistency

n-io (approved):

PR #3436, opened 2024-11-13
dialects: (riscv) split out the base classes for custom format and assembly printing

superlopuh:
We currently have a single base class for two kinds of shared functionality: printing the assembly for the target, and printing the custom format. I'd like to decouple them to migrate some RISC-V operations to custom format, which cannot be provided at the same time as custom printing and parsing functions.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -362,7 +362,7 @@ class RISCVAsmOperation(HasRegisterConstraints, IRDLOperation, ABC):
         raise NotImplementedError()
 
 
-class RISCVCustomFormatOperation(RISCVAsmOperation, ABC):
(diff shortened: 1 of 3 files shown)


zero9178 (approved):

xdsl/dialects/riscv.py
```diff
@@ -361,6 +361,12 @@ def get_register_constraints(self) -> RegisterConstraints:
     def assembly_line(self) -> str | None:
         raise NotImplementedError()
 
+
+class RISCVCustomFormatOperation(RISCVAsmOperation, ABC):
```
alexarice, line 365: Is it intentional that this inherits from `RISCVAsmOperation`? Your description makes it sound like these should be disjoint

superlopuh: Ah yes I'll remove this, thank you

PR #3440, opened 2024-11-13
dialects: (stream) add constr to StreamType base class

superlopuh:
This is one of two cases of base classes of attributes that are used in constraints, but don't have a param_def.


xdsl/dialects/stream.py
```diff
@@ -49,37 +47,93 @@ def __init__(self, element_type: _StreamTypeElement):
     def get_element_type(self) -> _StreamTypeElement:
         return self.element_type
 
+    @overload
+    @classmethod
```
alexarice, line 51: Having these as `@staticmethod`s should allow you to specify the generic type properly and avoid having to add `[Attribute]` below. Think a similar thing came up on the previous `constr` additions

superlopuh: Yep I had the same thought just pushed. Forgot that that was the conclusion of the previous iteration's discussion.

alexarice (approved):
Wonderful

xdsl/dialects/stream.py
alexarice: I might go through and fix this in other places

alexarice: Unless that will clash with your next steps

alexarice: The only other one is in stencil and it looks like it'd be a real pain to get rid of

PR #3443, opened 2024-11-13
core: add TypeVarConstraint, and use in irdl_to_attr_constraint

superlopuh:
We currently do not preserve type var information in the ParamDef, meaning that each time that we construct a constraint that specialises a TypeVar used in the definition of an attribute, we reconstruct the definition from scratch. Currently, this means re-parsing the type hints used when defining the attribute, but with a dictionary that maps the TypeVar used to the concrete constraint to use for the attribute's parameter. In order to move away from the type-based ParamDef pattern of defining parametrized attributes, we need to somehow store the information of TypeVars used in the attribute definition. This PR proposes a mechanism based on a new kind of attribute constraint, and uses it to split the existing mechanism into two phases: first constructing the constraint with TypeVars, and then swapping out the TypVars in the result. Upcoming PRs will remove the part where we remove the type parsing for each time that a new constraint for a ParametrizedAttribute is constructed.


superlopuh:
I'm feeling a bit stuck... There are a few problems still remaining, one of which is that I couldn't hack the whole lazy get/set tyoe variable, so we may have to deal with this as a temporary measure. The other problem is with the type system, where it seems like for this to really make sense we would need to ask all clients to enable experimental Pyright type checking. I had a chat with Mathieu and he is fine with this, I'll raise this at the xDSL meeting this afternoon. I've also not yet completed the migration to `param_def` in my topmost branch, so will have to check what other functionality is missing before opening this all up for a proper discussion. I'm giving myself until the end of the year to figure this out, if that's OK with everyone.

math-fehr (approved):
Really cool!

superlopuh:
After some discussions with @math-fehr and some further inspiration I'd like to try making this all a bit simpler and smaller. I'm aiming to make the changes today or this weekend.

docs/marimo/irdl.py
```diff
@@ -355,6 +355,15 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
             for e in attr.data:
                 self.elem_constr.verify(e, constraint_context)
 
+        def mapping_type_vars(self, type_var_mapping: dict[TypeVar, AttrConstraint]):
```
alexarice, line 358: This should have a return type.

docs/marimo/irdl.py
```diff
@@ -355,6 +355,15 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
...
+            # If self.elem_constr has any type variable constraints, map them according to the input dictionary
+            elem_constr = self.elem_constr.mapping_type_vars(type_var_mapping)
+            if elem_constr is self.elem_constr:
```
alexarice: Feels like a premature optimisation to me?

docs/marimo/irdl.py
```diff
@@ -334,7 +334,7 @@ def _(ConstraintContext, IntAttr, VerifyException, attr_constr_coercion):
     from xdsl.ir import Attribute
     from xdsl.irdl import AttrConstraint
     from xdsl.utils.hints import isa
-
+    from typing import TypeVar
```
alexarice: typing_extensions

tests/dialects/test_builtin.py
```diff
@@ -808,3 +819,45 @@ def test_array_constr():
...
+
+def test_array_of_constraint():
+    """Test mapping type variables in ArrayOfConstraint."""
+    array_constraint = ArrayOfConstraint(TypeVarConstraint(_A, BaseAttr(A)))
+
+    assert array_constraint.mapping_type_vars({}) == ArrayOfConstraint(BaseAttr(A))
+    assert array_constraint.mapping_type_vars({_B: BaseAttr(B)}) == ArrayOfConstraint(
+        BaseAttr(A)
+    )
+    assert array_constraint.mapping_type_vars({_A: BaseAttr(B)}) == ArrayOfConstraint(
```
alexarice, line 846: It's not clear to me that this is the behavior we want? Shouldn't this be `BaseAttr(A) & BaseAttr(B)` ?

tests/irdl/test_attribute_definition.py
```diff
@@ -652,13 +662,22 @@ def verify(
         for e in attr.data:
             self.elem_constr.verify(e, constraint_context)
 
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> DataListAttr:
+        mapped_constraint = self.elem_constr.mapping_type_vars(type_var_mapping)
+        if mapped_constraint is self.elem_constr:
```
alexarice: I think I'm not a fan of these optimisations

tests/irdl/test_attribute_definition.py
```diff
@@ -836,6 +855,34 @@ def test_custom_constructor():
...
+                "param",
+                TypeVarConstraint(
+                    type_var=AttributeInvT,
+                    constraint=AnyAttr(),
+                ),
+            )
+        ],
+    )
+
+    assert base(GenericAttr[IntAttr]) == ParamAttrConstraint(
```
alexarice, line 924: This is not what I would have expected, think I'm going through the files in the wrong order.

xdsl/irdl/attributes.py
```diff
@@ -16,14 +16,15 @@
...
     TypeAlias,
-    TypeVar,
     Union,
     cast,
     get_args,
     get_origin,
     get_type_hints,
 )
 
+from typing_extensions import TypeVar
```
alexarice: For a separate PR: can we make this a ruff/pyright error or something?

xdsl/irdl/constraints.py
```diff
@@ -164,6 +163,18 @@ def __and__(
...
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> GenericAttrConstraint[AttributeCovT]:
+        """
+        A helper function to make type vars used in attribute definitions concrete when
+        creating constraints for new attributes or operations.
+        """
+        raise NotImplementedError(
+            "Custom constraints must map type vars in nested constraints, if any."
+        )
```
alexarice, line 177: I'm not sure if this is a comment for this PR or somewhere else, but I've been wondering if it would be nice to have a more general `walk` or `subconstraints` function or something. Also have the need for two things to be able to get a set of all `VarConstraint`s used in a constraint and it would be nice to have a minimal set of things that need to be implemented on the constraint.

alexarice: I'm happy to think about this and iterate later but thought I'd bring it up now.

xdsl/irdl/constraints.py
```diff
@@ -759,6 +868,18 @@ def infer(
         """
         raise ValueError(f"Cannot infer range from constraint {self}")
 
+    @abstractmethod
```
alexarice, line 866: Do we ever forsee `IntConstraints` containing `TypeVars`? Maybe I'm missing something.

xdsl/irdl/constraints.py
```diff
@@ -855,6 +984,14 @@ def infer(
         attr = self.constr.infer(context)
         return (attr,) * length
 
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> GenericRangeConstraint[AttributeCovT]:
+        constr = self.constr.mapping_type_vars(type_var_mapping)
+        if constr is self.constr:
+            return self
+        return RangeOf(constr, length=self.length)
```
alexarice: Either int constraints shouldn't have a `mapping_type_vars` or we should use it here.

tests/irdl/test_attribute_definition.py
alexarice: I never found out where this logic is.

xdsl/irdl/constraints.py
superlopuh: I think that the obvious solution for this is to actually use Attributes to represent constraints, as we can then make stronger assumptions on them, such as them being immutable, introspectable, and others. It would completely eliminate the need to manually specify a walk function, as we already have one for the type rewriting, and it would let us map type vars using that infrastructure. I would rather look into this in a future PR, though.

xdsl/irdl/constraints.py
superlopuh: Yeah this bit I wasn't sure about, feels like we should add this later if we find that it's missing. I'll remove this set of functions.

- [x] remove it

xdsl/irdl/attributes.py
superlopuh: A quick search says no :(

docs/marimo/irdl.py
superlopuh: Fine I'll remove it

docs/marimo/irdl.py
superlopuh: Why?

xdsl/irdl/constraints.py
alexarice: Interesting, so in this system, any subconstraints would actually just be parameters. Is the idea that these attributes would actually be part of a dialect, or would more take the form of "pseudoattributes" which we just use as a data format that we never parse/print etc

docs/marimo/irdl.py
alexarice: it returns something

alexarice: aside: is pyright being ran on marimo notebooks?

superlopuh: Not yet, they have a mypy situation but not Pyright support. Here's the tracking issue:  https://github.com/marimo-team/marimo/issues/1163

docs/marimo/irdl.py
superlopuh: I opened a separate PR to make this PR a bit more focused: https://github.com/xdslproject/xdsl/pull/4487

xdsl/irdl/constraints.py
superlopuh: I think it would make sense to have some built-in constraints and then some be provided by the dialect.

docs/marimo/irdl.py
superlopuh: Done

xdsl/irdl/attributes.py
superlopuh: I was proven wrong.

tests/dialects/test_builtin.py
superlopuh: I expect that the type system should already enforce this, but also the base, unspecialised verification of the attribute.

tests/irdl/test_attribute_definition.py
superlopuh: Not sure what the question is, you mean that you expect `base` to return a `BaseAttr` constraint?

xdsl/irdl/constraints.py
superlopuh: In the process of removing it, I found that `ArrayOfConstraint` actually relies on this being here, as the nested constraint is a range constraint. I'd rather merge this as-is, and see later whether we should also add it to IntConstraints or remove it from RangeConstraint after cleaning up the ArrayOfConstraint.

docs/marimo/irdl.py
superlopuh: Two problems here, one is that marimo doesn't like `from __future__ import annotation`, the other is that this returns an instance of itself, and I'd rather not return Self here. Meanwhile pyright actually would infer this, and constrain it to the right generic superclass since this is a method override. I propose leaving it like this.

xdsl/irdl/constraints.py
superlopuh: IntConstraints currently don't allow for mapping type vars. I'm not sure that it's necessary, but I'm happy to add it in this PR if you'd prefer.

docs/marimo/irdl.py
alexarice: I'll defer to you as I'm less sure on the marimo notebooks

tests/dialects/test_builtin.py
```diff
@@ -808,3 +819,45 @@ def test_array_constr():
     assert constr.infer(ctx) == ArrayAttr([i32, i32])
 
     assert constr.get_bases() == {ArrayAttr}
+
+
+################################################################################
+# Mapping Type Var
```
alexarice, line 825: Ultra nit: this is missing the rhs of the box

tests/dialects/test_builtin.py
alexarice: I don't understand, there's a clear example here of the mapping completely discarding the original constraint, which is not enforced?

tests/irdl/test_attribute_definition.py
alexarice: I'm surprised that `base` wouldn't create a `BaseAttr` yes. Is the logic for this in the `base` function?

tests/dialects/test_builtin.py
alexarice: It seems all the boxes in this file are a bit lidless

tests/irdl/test_attribute_definition.py
```diff
@@ -377,6 +384,11 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
         if attr.data <= 0:
             raise VerifyException(f"Expected positive integer, got {attr.data}.")
 
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> PositiveIntConstr:
+        return self
```
alexarice, line 437: I'm guessing we're not setting a default to enforce people to think about this?

xdsl/irdl/constraints.py
```diff
@@ -244,6 +264,41 @@ def can_infer(self, var_constraint_names: Set[str]) -> bool:
...
+class TypeVarConstraint(AttrConstraint):
+    """
+    Stores the TypeVar instance used to define a generic ParametrizedAttribute.
+    """
+
+    type_var: TypeVar
+    """The instance of the TypeVar used in the definition."""
+
+    constraint: AttrConstraint
+    """The base constraint of the TypeVar."""
```
alexarice: I think this should have a better name/docstring. Is this meant to represent the constraint for the bound of the typevar? Otherwise it feel strange that it is overwritten by `mapping_type_vars`.

xdsl/irdl/constraints.py
```diff
@@ -244,6 +264,41 @@ def can_infer(self, var_constraint_names: Set[str]) -> bool:
...
+    ) -> None:
+        self.constraint.verify(attr, constraint_context)
+
+    def get_bases(self) -> set[type[Attribute]] | None:
+        return self.constraint.get_bases()
+
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> GenericAttrConstraint[Attribute]:
+        return type_var_mapping.get(self.type_var, self.constraint)
```
alexarice: Having `self.constraint` as a default here doesn't feel correct to me. I feel either we should throw an error if we expect all type_vars to be simultaneously mapped, or should have the default be `self` to allow future mappings to set this correctly.

xdsl/irdl/constraints.py
alexarice: I don't see how your comment aligns with mine? I'm not saying that `RangeConstraint` shouldn't have `mapping_type_vars`, just questioning whether `IntConstraint` should. If it definitely should then we should have `self.length.mapping_type_vars(type_var_mapping)` in the `RangeOf` constraint.

xdsl/irdl/constraints.py
```diff
@@ -854,6 +972,13 @@ def infer(
         attr = self.constr.infer(context)
         return (attr,) * length
 
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> RangeOf[AttributeCovT]:
+        return RangeOf(
+            self.constr.mapping_type_vars(type_var_mapping), length=self.length
```
alexarice, line 989:
```suggestion
            self.constr.mapping_type_vars(type_var_mapping), length=self.length.mapping_type_vars(type_var_mapping),
```
This is where I'm referring to.

xdsl/irdl/constraints.py
alexarice: Ah this was my bad, github contracted everything between `class IntConstraint` and here, and I didn't realise this was the one for `RangeConstraint`s

xdsl/irdl/constraints.py
alexarice: Lets leave as is

tests/irdl/test_attribute_definition.py
superlopuh: Yeah exactly. I am optimistic that we'll find a way to improve on this mechanism in the future but for now I think the best we can do is make users aware that this is something they should think about.

xdsl/irdl/constraints.py
superlopuh: I'm happy to throw an error here.

superlopuh: This found a bug, I'll debug later.

tests/irdl/test_attribute_definition.py
superlopuh: Yeah, base is not the same as BaseAttr, and is more of a complement to `isa`, in that it maps the whole type to constraints including the generic parameters.

tests/dialects/test_builtin.py
superlopuh: What I mean is that when the attribute is initialised, the verify is called with the constraint in the paramdef, which will use the base constraint. This means that we don't need to re-verify it here when saying that we want an attribute with more specific arguments.

alexarice: OK, that does make sense

xdsl/irdl/constraints.py
superlopuh: Fixed in #4496

xdsl/irdl/constraints.py
superlopuh: Updated, WDYT?

xdsl/dialects/bufferization.py
```diff
@@ -83,6 +86,11 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
             )
         return self.memref_constraint.verify(memref_type, constraint_context)
 
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> Self:
+        return type(self)(self.memref_constraint.mapping_type_vars(type_var_mapping))
```
math-fehr: Why this `type(self)`? Do you intend to subclass this constraint?

math-fehr (approved):

xdsl/dialects/bufferization.py
superlopuh: Rewritten

PR #3444, opened 2024-11-13
core: introduce param_def API for ParametrizedAttribute declaration

superlopuh:
This change introduces a new API, to bring attribute definition in line with operation definition. My proposal is to allow either API for the next release, in order not to break dependencies. I will open another PR on top of this one that will deprecate the old API, and migrate all of the dialects in xDSL to the new API. After we release a version with the deprecation warning, we can remove the old implementation entirely.

Note stacked PR


xdsl/irdl/attributes.py
```diff
@@ -81,12 +83,56 @@ def generic_constraint_coercion(args: tuple[Any]) -> AttrConstraint:
...
+
+
+@overload
+def param_def(
+    constraint: TypeVar,
+    *,
+    default: None = None,
+    resolver: None = None,
+    init: Literal[False] = False,
+) -> Attribute: ...
```
alexarice, line 109: This doesn't feel ideal? We effectively lose the python typing for the field which was kind of the whole point of having the typevar

xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
```diff
@@ -649,7 +650,7 @@ def return_target_analysis(module: builtin.ModuleOp):
 
 
 class StencilTypeConversion(TypeConversionPattern):
-    @attr_type_rewrite_pattern
+    @attr_constr_rewrite_pattern(StencilTypeConstr)
```
alexarice, line 653: Is this a different change?

alexarice (commented):
I can't see anything particularly wrong but not having the typevar parameters actually use the typevar for their typing feels really bad to me. On the other hand I don't really have a solution

xdsl/irdl/attributes.py
superlopuh: yep and it's currently breaking my code when I migrate xDSL to it. I'll probably change this soon.

xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
superlopuh: I guess it's more related to the previous change, let me revert this and see what happens

superlopuh: Nope it's related to this change, since we use the attr_def instead of the type hints to construct the constraint, we can't do it on the ABC that is StencilTypeConstr since it's not really an attribute. We have to construct the constraint manually instead.

superlopuh:
After chatting with @math-fehr and @alexarice, it seems like a good idea to wait for a week or two to get feedback on https://github.com/microsoft/pyright/issues/10516 before we commit to this API.

superlopuh:
I'm now thinking we should rip it out in one go, and not do the whole deprecation thing.

PR #3445, opened 2024-11-13
transformations: (riscv) Add multiplication by power-of-2 -> SLL canonicalisations

kaylendog:
Implements the following canonicalisations as rewrite patterns to the RISC-V dialect:
- [ ] `x * 2^i = x << i` (WIP)
- [ ] `x * -2^i = -(x << i)`

@superlopuh


xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -70,6 +71,49 @@ def match_and_rewrite(self, op: riscv.MulOp, rewriter: PatternRewriter) -> None:
...
+
+        match (lhs, rhs < 0):
+            case None, False:
+                rewriter.replace_matched_op(
+                    riscv.SlliOp(op.rs1, i, rd=rd, comment=op.comment)
+                )
+            case int(), False:
+                rewriter.replace_matched_op(
+                    riscv.LiOp(lhs << i, rd=rd, comment=op.comment)
+                )
```
superlopuh, line 101: won't this already be done by the pattern that multiplies constant values together? I don't think we need to handle this case here

tests/filecheck/backend/riscv/canonicalize.mlir
```diff
@@ -169,6 +172,9 @@ builtin.module {
 // CHECK-NEXT:   %multiply_immediate_l0 = riscv.mv %c0_1 : (!riscv.reg) -> !riscv.reg<a0>
 // CHECK-NEXT:   "test.op"(%multiply_immediate_l0) : (!riscv.reg<a0>) -> ()
 
+// CHECK-NEXT:   %multiply_immediate_power_two = riscv.slli %i1, 1 : (!riscv.reg) -> !riscv.reg<a0>
```
superlopuh, line 175: The input is in a1 before the rewrite, it should stay there
```suggestion
// CHECK-NEXT:   %multiply_immediate_power_two = riscv.slli %i1, 1 : (!riscv.reg<a1>) -> !riscv.reg<a0>
```

xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -70,6 +71,49 @@ def match_and_rewrite(self, op: riscv.MulOp, rewriter: PatternRewriter) -> None:
...
+        if (rs2 := get_constant_value(op.rs2)) is not None:
+            rhs = rs2.value.data
+
+        rd = cast(riscv.IntRegisterType, op.rd.type)
+
+        if rhs is None:
+            return
+
+        # get shl immediate
+        i = int(log2(abs(rhs)))
```
superlopuh, line 91: this doesn't test that it's a power of two?

PR #3448, opened 2024-11-14
dialects (arm): Initialise ARM dialect and add initial mov operation

emmau678:
Initialise a dialect for ARM instructions, similar to the existing x86 dialect. Add support for an initial mov operation which moves data from the source register to the destination.


tobiasgrosser:
@emmau678, this is great. Consider starting of from a fresh branch with just your new commits. @superlopuh can advise on how.

PR #3449, opened 2024-11-14
dialects (arm): add ARM dialect and initial mov op

emmau678:
Initialise a dialect for ARM instructions, similar to the existing x86 dialect. Add support for an initial mov operation which moves data from the source register to the destination.

### DIFF
--- a/tests/filecheck/dialects/arm/arm_assembly_emission.mlir
+++ b/tests/filecheck/dialects/arm/arm_assembly_emission.mlir
@@ -1,0 +1,7 @@
+// RUN: xdsl-opt -t arm-asm %s | filecheck %s
+
+%0 = arm.get_register : () -> !arm.reg<x0>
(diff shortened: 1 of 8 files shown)


superlopuh (commented):
This is great! I would like to change a few things, and I feel that the most efficient way to do that would be via smaller PRs. I would recommend starting with a PR that only adds the LabelAttr, and a filecheck file called `arm_attrs.mlir` that checks that we can rountrip 
``` mlir
"test.op() {"my_label" = !arm.label<"label_str">} : () -> ()
```
After that is in we can start adding more functionality bit by bit

superlopuh:
I was going to suggest keeping this for the mov op, which will sit on top of the next PR to be opened, but we can always open a new PR for that also

alexarice:
The mov op has been added separately right? Can this be closed?

superlopuh:
Yep, let's close this

PR #3450, opened 2024-11-14
misc: use experimental TypeForm feature of Pyright

superlopuh:
This lets us use better typing when dealing with TypeVars, PRs leveraging this incoming.


compor:
@superlopuh can you link the incoming PRs when they land?

alexarice (approved):
Will this allow us to simplify `isa` and `isattr`?

superlopuh:
I hope so. I'll wait for @math-fehr's review to merge. The main risk with this would be to rely on an experimental feature, so we might get stuck again if they remove support. OTOH, we pin our version of Pyright, so it's not like everything will break, and I haven't seen any negative feedback on the TypeForm PEP so I have a feeling it'll land eventually.

superlopuh:
I'll keep exploring other options but I have a feeling that it might be the best one to have a clean interface for `param_op`...

xdsl/utils/hints.py
```diff
@@ -17,10 +18,13 @@
...
 from xdsl.utils.exceptions import VerifyException
 
+if TYPE_CHECKING:
+    from typing_extensions import TypeForm
+
 _T = TypeVar("_T")
 
 
-def isa(arg: Any, hint: type[_T]) -> TypeGuard[_T]:
+def isa(arg: Any, hint: "TypeForm[_T]") -> TypeGuard[_T]:
```
math-fehr, line 27: What error do you get here?

tests/pattern_rewriter/test_pattern_rewriter.py
```diff
@@ -245,7 +244,7 @@ def test_recursive_rewriter():
     class Rewrite(RewritePattern):
         @op_type_rewrite_pattern
         def match_and_rewrite(self, op: Constant, rewriter: PatternRewriter):
-            if not isa(op_val := op.value, IntegerAttr):
+            if not isinstance(op_val := op.value, IntegerAttr):
```
math-fehr, line 247: What happens in those cases? `isa` is returning an error, or is it just that `isinstance` can do it so we should follow that?

math-fehr (approved):

xdsl/utils/hints.py
superlopuh: TypeForm is defined in Pyright, not typing_extensions, so the error is:
```
E   NameError: name 'TypeForm' is not defined
```

tests/pattern_rewriter/test_pattern_rewriter.py
superlopuh:
```
Argument type is partially unknown
  Argument corresponds to parameter "hint" in function "isa"
  Argument type is "TypeForm[IntegerAttr[Unknown]]"
```

superlopuh: I assume that this isn't the case for newer versions of Python, where generic arguments can have default values, where `IntegerAttr` would be shorthand for `IntegerAttr[IntegerType | IndexType]`

superlopuh:
Actually I'm now realizing that this will mean that any downstream users also need to enable experimental features locally for their xDSL code to type check... I'm not sure that I want to force that on people. Let's see if we find another solution for the current goal of migrating to `param_def`, and reopen this if it's the only option.

superlopuh:
I had an offline discussion with Mathieu, and he convinced me that it would be fine. In the end, type-checking is an optional feature, and is not required for the framework to run correctly for the user, and it's only a flag away.

n-io (approved):
Looks fine from our side!

alexarice:
Will we likely need to turn on some experimental feature flag downstream?

superlopuh:
Yep, same one as in this PR. I'll open another PR for the README change

PR #3451, opened 2024-11-14
dialects: (bufferization) add custom syntax to alloc_tensor

superlopuh:


xdsl/dialects/bufferization.py
```diff
@@ -103,6 +140,51 @@ def __init__(
             result_types=(result_type,),
         )
 
+    @classmethod
```
alexarice, line 143: Does this not work?
```
assembly_format = "`(` $dynamic_sizes `)` ( `copy` `(` $copy^ `)`)? (`size_hint` `=` $size_hint^)? attr-dict `:` type($result)
```

xdsl/dialects/bufferization.py
```diff
@@ -1,4 +1,5 @@
-from typing import Any
+from collections.abc import Sequence
+from typing import Any, Self
```
alexarice, line 2:
```suggestion
from typing import Any
from typing_extensions import Self
```

xdsl/dialects/bufferization.py
superlopuh: It won't print exactly what MLIR prints but I think that's ok, the difference is printing whitespace around the `=`

compor (approved):

alexarice (approved):

PR #3452, opened 2024-11-14
[WIP] transformations: Add expand-copy pass that expands memref.copy as affine.load and affine.store pairs

gabrielrodcanal:
Pass that expands memref.copy as a nested structure of affine for loops with load/store. This is a functional port of the expand-copy pass in Google's HEIR: https://heir.dev/docs/passes/#-expand-copy.  

Not sure if this is interesting for the community, but is something I needed for my own work. If it is, I will provide tests soon.

### DIFF
--- a/xdsl/transforms/experimental/expand_copy.py
+++ b/xdsl/transforms/experimental/expand_copy.py
@@ -9,7 +9,6 @@ from xdsl.dialects.affine import AffineExpr, AffineMap, AffineMapAttr
 from xdsl.ir import BlockArgument
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
-    GreedyRewritePatternApplier,
     PatternRewriter,
     PatternRewriteWalker,
     RewritePattern,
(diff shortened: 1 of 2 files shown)


xdsl/transforms/experimental/expand_copy.py
```diff
@@ -0,0 +1,93 @@
...
+        rewriter.replace_matched_op(top_loop)
+
+
+@dataclass(frozen=True)
+class ExpandCopyPass(ModulePass):
+    name = "expand-copy"
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        expand_copy_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier([ExpandCopy()]),
```
superlopuh, line 89: I don't think there's a need for the greedy applier if it's just one rewrite

xdsl/transforms/experimental/expand_copy.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # Generate the innermost loop that performs element-wise copy and insert in the nesting structure
+        @Builder.region([builtin.IndexType()])
+        def innermost_loop_body(builder: Builder, args: [BlockArgument, ...]):
+            for_indices = list(
+                map(lambda loop: loop.body.block.args[0], loopnest_queue)
+            )
+            for_indices.append(args[0])
+
+            symbols = list(map(lambda dim: AffineExpr.dimension(dim), range(ndims)))
```
superlopuh, line 63: I'd recommend list/tuple comprehensions over map these days, they're noticeably faster

superlopuh (commented):
Sure, why not :) But would definitely be great for it to be well-tested. Out of curiosity, what will you be using it for?

gabrielrodcanal:
Then I will apply your comments and include tests in the following days :).

superlopuh:
Hi @gabrielrodcanal, do you have time to add tests for this? Let's close otherwise

superlopuh:
I think we would benefit from having this, let's reopen this PR when you have time to add the tests

PR #3453, opened 2024-11-14
transformations: port MLIR's empty-tensor-to-alloc-tensor pass

superlopuh:
I wanted to do this before bufferization, so we'll be able to go from start to finish all in Python once that's done.

Note stacked PR.


xdsl/transforms/empty_tensor_to_alloc_tensor.py
```diff
@@ -0,0 +1,36 @@
...
+            bufferization.AllocTensorOp(
+                op.tensor.type,
+                op.dynamic_sizes,
+            )
+        )
+
+
+class EmptyTensorToAllocTensorPass(ModulePass):
+    """
+    tensor.empty ops return a tensor of unspecified contents who's only purpose
```
alexarice, line 25:
```suggestion
    tensor.empty ops return a tensor of unspecified contents whose only purpose
```

alexarice (approved):

compor (approved):

PR #3455, opened 2024-11-15
dialects (arm): initialise ARM dialect

emmau678:
Initialise ARM dialect

### DIFF
--- a/xdsl/dialects/__init__.py
+++ b/xdsl/dialects/__init__.py
@@ -323,6 +323,11 @@ def get_all_dialects() -> dict[str, Callable[[], Dialect]]:
 
         return X86
 
+    def get_arm():
+        from xdsl.dialects.arm import ARM
+
+        return ARM
+
     def get_transform():
         from xdsl.dialects.transform import Transform
 
@@ -392,6 +397,7 @@ def get_all_dialects() -> dict[str, Callable[[], Dialect]]:
         "vector": get_vector,
         "wasm": get_wasm,
         "x86": get_x86,
+        "arm": get_arm,
         "transform": get_transform,
     }
 
--- a/xdsl/dialects/arm/__init__.py
+++ b/xdsl/dialects/arm/__init__.py
@@ -1,0 +1,16 @@
+from xdsl.ir import Dialect
+
+from .register import GeneralRegisterType
+
+"""
+ARM dialect, based on the ISA specification in:
+https://developer.arm.com/documentation/102374/0101/Overview
+"""
+
+ARM = Dialect(
+    "arm",
+    [],
+    [
+        GeneralRegisterType,
+    ],
+)
--- a/xdsl/dialects/arm/register.py
+++ b/xdsl/dialects/arm/register.py
@@ -1,0 +1,127 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import Attribute
+from xdsl.irdl import (
+    irdl_attr_definition,
+)
+from xdsl.parser import AttrParser
+from xdsl.utils.exceptions import VerifyException
+
+
+class ARMRegisterType(RegisterType):
+    """
+    The abstract class for all ARM register types.
+    """
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        if parser.parse_optional_punctuation("<") is not None:
+            name = parser.parse_identifier()
+            parser.parse_punctuation(">")
+            if not name.startswith("x"):  # only including x regs for now
+                assert (
+                    name in cls.abi_index_by_name()
+                ), f"{name}"  # abi_index_by_name raises NotImplementedError (?)
+        else:
+            name = ""
+        return cls._parameters_from_spelling(name)
+
+    def verify(self) -> None:
+        name = self.spelling.data
+        if not self.is_allocated or name.startswith("x"):
+            return
+        if name not in type(self).abi_index_by_name():
+            raise VerifyException(f"{name} not in {self.instruction_set_name()}")
+
+
+ARM_INDEX_BY_NAME = {
+    "x0": 0,
+    "x1": 1,
+    "x2": 2,
+    "x3": 3,
+    "x4": 4,
+    "x5": 5,
+    "x6": 6,
+    "x7": 7,
+    "x8": 8,
+    "x9": 9,
+    "x10": 10,
+    "x11": 11,
+    "x12": 12,
+    "x13": 13,
+    "x14": 14,
+    "x15": 15,
+    "x16": 16,
+    "x17": 17,
+    "x18": 18,
+    "x19": 19,
+    "x20": 20,
+    "x21": 21,
+    "x22": 22,
+    "x23": 23,
+    "x24": 24,
+    "x25": 25,
+    "x26": 26,
+    "x27": 27,
+    "x28": 28,
+    "x29": 29,
+    "x30": 30,
+}
+
+
+@irdl_attr_definition
+class GeneralRegisterType(ARMRegisterType):
+    """
+    A scalar ARM register type representing general registers.
+    """
+
+    name = "arm.reg"
+
+    @classmethod
+    def unallocated(cls) -> GeneralRegisterType:
+        return UNALLOCATED_GENERAL
+
+    @classmethod
+    def instruction_set_name(cls) -> str:
+        return "arm"
+
+    @classmethod
+    def abi_index_by_name(cls) -> dict[str, int]:
+        return ARM_INDEX_BY_NAME
+
+
+UNALLOCATED_GENERAL = GeneralRegisterType("")
+X0 = GeneralRegisterType("x0")
+X1 = GeneralRegisterType("x1")
+X2 = GeneralRegisterType("x2")
+X3 = GeneralRegisterType("x3")
+X4 = GeneralRegisterType("x4")
+X5 = GeneralRegisterType("x5")
+X6 = GeneralRegisterType("x6")
+X7 = GeneralRegisterType("x7")
+X8 = GeneralRegisterType("x8")
+X9 = GeneralRegisterType("x9")
+X10 = GeneralRegisterType("x10")
+X11 = GeneralRegisterType("x11")
+X12 = GeneralRegisterType("x12")
+X13 = GeneralRegisterType("x13")
+X14 = GeneralRegisterType("x14")
+X15 = GeneralRegisterType("x15")
+X16 = GeneralRegisterType("x16")
+X17 = GeneralRegisterType("x17")
+X18 = GeneralRegisterType("x18")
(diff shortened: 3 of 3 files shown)


xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import Attribute
+from xdsl.irdl import (
+    irdl_attr_definition,
```
superlopuh, line 8:
```suggestion
    irdl_attr_definition
```

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,127 @@
...
+)
```
superlopuh, line 9:
```suggestion
from xdsl.irdl import irdl_attr_definition
```

xdsl/dialects/arm/__init__.py
```diff
@@ -0,0 +1,16 @@
+from xdsl.ir import Dialect
+
+from .register import GeneralRegisterType
+
+"""
+ARM dialect, based on the ISA specification in:
+https://developer.arm.com/documentation/102374/0101/Overview
+"""
```
superlopuh, line 8: A string is recognised as a doc string if it's the first thing in the file, can you please move this up?

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,125 @@
...
+from xdsl.utils.exceptions import VerifyException
+
+
+class ARMRegisterType(RegisterType):
+    """
+    The abstract class for all ARM register types.
+    """
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
```
superlopuh, line 18:
```suggestion
class ARMRegisterType(RegisterType, abc.ABC):
    """
    The abstract class for all ARM register types.
    """

class ARMIntRegisterType(ARMRegisterType):
    """
    ARM Integer register class.
    """
    
    name = "reg"
    
    @classmethod
    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
```

superlopuh: We do want to group all ARM registers together, so it's useful to have an abstract base class. This split would make the `ARMRegisterType` actually abstract, and also provide an int register class. I have a feeling that we'll need to do something a bit more involved when we start supporting both X and W registers, but I think it should be OK to tackle that later.

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,125 @@
...
+from collections.abc import Sequence
+
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import Attribute
+from xdsl.irdl import irdl_attr_definition
+from xdsl.parser import AttrParser
+from xdsl.utils.exceptions import VerifyException
+
+
+class ARMRegisterType(RegisterType):
```
superlopuh, line 12:
```suggestion
class ARMRegisterType(RegisterType, abc.ABC):
```

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        if parser.parse_optional_punctuation("<") is not None:
+            name = parser.parse_identifier()
+            parser.parse_punctuation(">")
+            if not name.startswith("x"):  # only including x regs for now
+                assert (
+                    name in cls.abi_index_by_name()
+                ), f"{name}"  # abi_index_by_name raises NotImplementedError (?)
```
superlopuh, line 25: this bit of code doesn't make sense right now, as all the registers below do start with x, so the assert will never trigger.

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,125 @@
...
+    "x26": 26,
+    "x27": 27,
+    "x28": 28,
+    "x29": 29,
+    "x30": 30,
+}
+
+
+@irdl_attr_definition
+class GeneralRegisterType(ARMRegisterType):
```
superlopuh, line 74:
```suggestion
class IntRegisterType(ARMRegisterType):
```

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    A scalar ARM register type representing general registers.
```
superlopuh, line 76:
```suggestion
    A scalar ARM register type representing general-purpose integer registers.
```

xdsl/dialects/arm/register.py
superlopuh: I would say let's not verify any of this for the time being, and allow any string in the register name, we can always add the verification later.

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,125 @@
...
+    @classmethod
+    def instruction_set_name(cls) -> str:
+        return "arm"
+
+    @classmethod
+    def abi_index_by_name(cls) -> dict[str, int]:
+        return ARM_INDEX_BY_NAME
+
+
+UNALLOCATED_GENERAL = GeneralRegisterType("")
```
superlopuh, line 94:
```suggestion
UNALLOCATED_INT = GeneralRegisterType("")
```

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,119 @@
...
+    "x22": 22,
+    "x23": 23,
+    "x24": 24,
+    "x25": 25,
+    "x26": 26,
+    "x27": 27,
+    "x28": 28,
+    "x29": 29,
+    "x30": 30,
+}
```
alexarice, line 64:
```suggestion
ARM_INDEX_BY_NAME = { f"x{i}": i for i in range(0, 31) }
```

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,119 @@
...
+X21 = IntRegisterType("x21")
+X22 = IntRegisterType("x22")
+X23 = IntRegisterType("x23")
+X24 = IntRegisterType("x24")
+X25 = IntRegisterType("x25")
+X26 = IntRegisterType("x26")
+X27 = IntRegisterType("x27")
+X28 = IntRegisterType("x28")
+X29 = IntRegisterType("x29")
+X30 = IntRegisterType("x30")
```
alexarice, line 119: Is this really necessary?

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,119 @@
...
+            name = parser.parse_identifier()
+            parser.parse_punctuation(">")
+            if not name.startswith("x"):  # only including x regs for now
+                assert name in cls.abi_index_by_name(), f"{name}"
+        else:
+            name = ""
+        return cls._parameters_from_spelling(name)
+
+    def verify(self) -> None:
+        return
```
alexarice, line 29: Don't think there's any reason to have an empty verify

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class ARMRegisterType(RegisterType, abc.ABC):
+    """
+    The abstract class for all ARM register types.
+    """
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        if parser.parse_optional_punctuation("<") is not None:
```
alexarice, line 19:
```suggestion
        if parser.parse_optional_punctuation("<"):
```

xdsl/dialects/__init__.py
```diff
@@ -392,6 +397,7 @@ def get_transform():
         "vector": get_vector,
         "wasm": get_wasm,
         "x86": get_x86,
+        "arm": get_arm,
```
alexarice, line 400: Could this go in (approximately) the right place alphabetically

alexarice (commented):
Mostly small comments

xdsl/dialects/arm/register.py
compor: I expect this to go into a namespace like for RISC-V, it is a handy shorthand notation

xdsl/dialects/arm/__init__.py
emmau678: thanks for catching that!

xdsl/dialects/arm/register.py
```diff
@@ -0,0 +1,87 @@
...
+    The abstract class for all ARM register types.
+    """
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        if parser.parse_optional_punctuation("<"):
+            name = parser.parse_identifier()
+            parser.parse_punctuation(">")
+            if not name.startswith("x"):  # only including x regs for now
+                assert name in cls.abi_index_by_name(), f"{name}"
```
alexarice, line 23: Am I missing something? This assert will always fail right?

xdsl/dialects/arm/register.py
compor: hey @emmau678, regardless of my response, please acknowledge the initial concern, so that there is a written record of what is acknowledged and/or dismissed.

Thanks for working on this, really excited to see where it takes us.

emmau678: @compor ok perfect -should I leave this as unresolved? thank you both for the comments and helping me to get up to speed with the code review process

compor: I think it is fine now :) good job!

xdsl/dialects/arm/__init__.py
superlopuh: This is one of the reasons to let the original commenters resolve the comments :)

superlopuh (commented):
Code looks great! Would just need two small tests to make sure that the behaviour is what we expect, one with valid registers, and one without:
```
"test.op"() {unallocated = !arm.reg} : () -> ()
"test.op"() {allocated = !arm.reg<x1>} : () -> ()

// -----

"test.op"() {error = !arm.reg<error>} : () -> ()
```

These can be in the same file, would just need to add --split-input-file and --verify-diagnostics to make sure that it can match successfully

xdsl/dialects/arm/register.py
superlopuh: We played around with this locally, and our verify-diangostics runs after the parsing, so it's not trivial to do here. I also have a feeling that raising an error here might be a bit premature, and that we would probably benefit from building more infrastructure here to see how we deal with the weirdnesses of ARM registers before doing the verification.

xdsl/dialects/__init__.py
```diff
@@ -323,6 +323,11 @@ def get_x86():
 
         return X86
 
+    def get_arm():
```
alexarice, line 326: Could this also be moved to its alphabetical location

alexarice (approved):
Seems good to me (have left 1 small nit)

compor (approved):

PR #3456, opened 2024-11-15
core: new inference system for constraints

alexarice:
I've gone through a few different refactors this week and have eventually settled on this slightly smaller refactor for the constraint system. The key addition is that each constraint now returns a dictionary of "resolvers" (name can maybe be improved) which resolve a variable from the given attribute. This allows inference to be decoupled entirely from verification, and inference should no longer be quite as fragile/maybe slightly quicker.

In this version, the resolvers are not used for verification, as I had originally planned, and instead the previous verification framework is left as is. I believe this change can enable the following changes in the future:
- Allow constraints on properties to resolve variables in a clean way (see #3318)
- Allow the introduction of "length/int constraints" for managing the length of variadics (see #3416)
- Potentially facilitate extensions to assembly format (i.e. directly supplying a variable)
- Resolvers could be used to automate construction of initialiser methods for operation. 

PS. as of time of writing it appears this saves 4 lines of code

### DIFF
--- a/tests/dialects/test_bufferization.py
+++ b/tests/dialects/test_bufferization.py
@@ -45,7 +45,7 @@ def test_tensor_from_memref_inference():
         EqAttrConstraint(UnrankedMemrefType.from_type(f64))
     )
     assert constr3.can_infer(set())
-    assert constr3.infer(dict()) == UnrankedTensorType(f64)
+    assert constr3.infer({}) == UnrankedTensorType(f64)
 
 
 @irdl_op_definition
--- a/tests/dialects/test_bufferization.py
+++ b/tests/dialects/test_bufferization.py
@@ -39,7 +39,7 @@ def test_tensor_from_memref_inference():
         EqAttrConstraint(MemRefType(f64, [10, 20, 30]))
     )
     assert constr2.can_infer(set())
-    assert constr2.infer(dict()) == TensorType(f64, [10, 20, 30])
+    assert constr2.infer({}) == TensorType(f64, [10, 20, 30])
 
     constr3 = TensorFromMemrefConstraint(
         EqAttrConstraint(UnrankedMemrefType.from_type(f64))
--- a/tests/dialects/test_bufferization.py
+++ b/tests/dialects/test_bufferization.py
@@ -22,7 +22,6 @@ from xdsl.dialects.builtin import (
 from xdsl.dialects.test import TestOp
 from xdsl.ir import Attribute
 from xdsl.irdl import (
-    ConstraintContext,
     EqAttrConstraint,
     IRDLOperation,
     VarConstraint,
@@ -40,13 +39,13 @@ def test_tensor_from_memref_inference():
         EqAttrConstraint(MemRefType(f64, [10, 20, 30]))
     )
     assert constr2.can_infer(set())
-    assert constr2.infer(ConstraintContext()) == TensorType(f64, [10, 20, 30])
+    assert constr2.infer(dict()) == TensorType(f64, [10, 20, 30])
 
     constr3 = TensorFromMemrefConstraint(
         EqAttrConstraint(UnrankedMemrefType.from_type(f64))
     )
     assert constr3.can_infer(set())
-    assert constr3.infer(ConstraintContext()) == UnrankedTensorType(f64)
+    assert constr3.infer(dict()) == UnrankedTensorType(f64)
 
 
 @irdl_op_definition
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -62,7 +62,7 @@ from xdsl.irdl import (
 )
 from xdsl.parser import Parser
 from xdsl.printer import Printer
-from xdsl.utils.exceptions import ParseError, PyRDLOpDefinitionError
+from xdsl.utils.exceptions import ParseError, PyRDLOpDefinitionError, VerifyException
 
 ################################################################################
 # Utils for this test file                                                     #
@@ -1748,10 +1748,12 @@ def test_non_verifying_inference():
     %1 = test.one_operand_one_result_nested %0 : i32"""
     )
     with pytest.raises(
-        ParseError,
-        match="Verification error while inferring operation type: ",
+        VerifyException,
+        match="i32 should be of base attribute test.param_one",
     ):
-        check_roundtrip(program, ctx)
+        parser = Parser(ctx, program)
+        while (op := parser.parse_optional_operation()) is not None:
+            op.verify()
 
 
 def test_variadic_length_inference():
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -13,6 +13,7 @@ from xdsl.dialects.builtin import (
     I32,
     BoolAttr,
     IntegerAttr,
+    MemRefType,
     ModuleOp,
     UnitAttr,
 )
@@ -2215,3 +2216,103 @@ def test_renamed_optional_prop(program: str, output: str, generic: str):
     printer.print_op(parsed)
 
     assert generic == stream.getvalue()
+
+
+################################################################################
+#                                Extractors                                    #
+################################################################################
+
+
+@irdl_op_definition
+class AllOfExtractorOp(IRDLOperation):
+    name = "test.all_of_extractor"
+
+    T: ClassVar = VarConstraint("T", AnyAttr())
+    lhs = operand_def(T & MemRefType.constr(element_type=T))
+    rhs = operand_def(T)
+
+    assembly_format = "$lhs `,` $rhs attr-dict `:` type($lhs)"
+
+
+def test_all_of_extraction_fails():
+    ctx = MLContext()
+    ctx.load_op(AllOfExtractorOp)
+    ctx.load_dialect(Test)
+    parser = Parser(
+        ctx,
+        '%0 = "test.op"() : () -> memref<10xindex>\ntest.all_of_extractor %0, %0 : memref<10xindex>',
+    )
+    parser.parse_operation()
+    with pytest.raises(
+        ValueError,
+        match="Value of variable T could not be uniquely extracted.\n"
+        "Possible values are: {index, memref<10xindex>}",
+    ):
+        parser.parse_operation()
+
+
+@irdl_attr_definition
+class DoubleParamAttr(ParametrizedAttribute, TypeAttribute):
+    """An attribute with two unbounded attribute parameters."""
+
+    name = "test.param"
+
+    param1: ParameterDef[Attribute]
+    param2: ParameterDef[Attribute]
+
+
+@irdl_op_definition
+class ParamExtractorOp(IRDLOperation):
+    name = "test.param_extractor"
+
+    T: ClassVar = VarConstraint("T", AnyAttr())
+    lhs = operand_def(ParamAttrConstraint(DoubleParamAttr, (T, T)))
+    rhs = operand_def(T)
+
+    assembly_format = "$lhs `,` $rhs attr-dict `:` type($lhs)"
+
+
+def test_param_extraction_fails():
+    ctx = MLContext()
+    ctx.load_attr(DoubleParamAttr)
+    ctx.load_op(ParamExtractorOp)
+    ctx.load_dialect(Test)
+    parser = Parser(
+        ctx,
+        '%0 = "test.op"() : () -> !test.param<i32,i64>\ntest.param_extractor %0, %0 : !test.param<i32,i64>',
+    )
+    parser.parse_operation()
+    with pytest.raises(
+        ValueError,
+        match="Value of variable T could not be uniquely extracted.\n"
+        "Possible values are: {i32, i64}",
+    ):
+        parser.parse_operation()
+
+
+@irdl_op_definition
+class MultipleOperandExtractorOp(IRDLOperation):
+    name = "test.multiple_operand_extractor"
+
+    T: ClassVar = VarConstraint("T", AnyAttr())
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+
+    assembly_format = "$lhs `,` $rhs attr-dict `:` type($lhs) `,` type($rhs)"
+
+
+def test_multiple_operand_extraction_fails():
+    ctx = MLContext()
+    ctx.load_op(MultipleOperandExtractorOp)
+    ctx.load_dialect(Test)
+    parser = Parser(
+        ctx,
+        '%0, %1 = "test.op"() : () -> (index, i32)\ntest.multiple_operand_extractor %0, %1 : index, i32',
+    )
+    parser.parse_operation()
+    with pytest.raises(
+        ValueError,
+        match="Value of variable T could not be uniquely extracted.\n"
+        "Possible values are: {i32, index}",
+    ):
+        parser.parse_operation()
--- a/xdsl/dialects/bufferization.py
+++ b/xdsl/dialects/bufferization.py
@@ -1,4 +1,4 @@
-from collections.abc import Sequence
+from collections.abc import Sequence, Set
 from dataclasses import dataclass
 from typing import Any, ClassVar
 
@@ -22,6 +22,7 @@ from xdsl.irdl import (
     ConstraintContext,
     GenericAttrConstraint,
     IRDLOperation,
+    ResolveType,
     VarConstraint,
     irdl_op_definition,
     operand_def,
@@ -48,18 +49,16 @@ class TensorFromMemrefConstraint(
         MemRefType[Attribute] | UnrankedMemrefType[Attribute]
     ]
 
-    def can_infer(self, constraint_names: set[str]) -> bool:
-        return self.memref_constraint.can_infer(constraint_names)
+    def can_infer(self, variables: Set[str]) -> bool:
+        return self.memref_constraint.can_infer(variables)
 
-    def infer(self, constraint_context: ConstraintContext) -> Attribute:
-        memref_type = self.memref_constraint.infer(constraint_context)
-        if isa(memref_type, MemRefType[Attribute]):
+    def infer(
+        self, variables: dict[str, ResolveType]
+    ) -> TensorType[Attribute] | UnrankedTensorType[Attribute]:
+        memref_type = self.memref_constraint.infer(variables)
+        if isinstance(memref_type, MemRefType):
             return TensorType(memref_type.element_type, memref_type.shape)
-        assert isa(memref_type, UnrankedMemrefType[Attribute])
         return UnrankedTensorType(memref_type.element_type)
-
-    def get_resolved_variables(self) -> set[str]:
-        return self.memref_constraint.get_resolved_variables()
 
     def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
         if isa(attr, TensorType[Attribute]):
@@ -67,8 +66,8 @@ class TensorFromMemrefConstraint(
             return self.memref_constraint.verify(memref_type, constraint_context)
         if isa(attr, UnrankedTensorType[Attribute]):
             memref_type = UnrankedMemrefType.from_type(attr.element_type)
-            return self.memref_constraint.verify(memref_type, constraint_context)
-        raise VerifyException(f"Expected tensor or unranked tensor type, got {attr}")
+
+        raise VerifyException(f"Expected TensorType or UnrankedTensorType, got {attr}")
 
 
 @irdl_op_definition
--- a/xdsl/dialects/bufferization.py
+++ b/xdsl/dialects/bufferization.py
@@ -63,11 +63,13 @@ class TensorFromMemrefConstraint(
     def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
         if isa(attr, TensorType[Attribute]):
             memref_type = MemRefType(attr.element_type, attr.shape)
-            return self.memref_constraint.verify(memref_type, constraint_context)
-        if isa(attr, UnrankedTensorType[Attribute]):
+        elif isa(attr, UnrankedTensorType[Attribute]):
             memref_type = UnrankedMemrefType.from_type(attr.element_type)
-
-        raise VerifyException(f"Expected TensorType or UnrankedTensorType, got {attr}")
+        else:
+            raise VerifyException(
+                f"Expected tensor or unranked tensor type, got {attr}"
+            )
+        return self.memref_constraint.verify(memref_type, constraint_context)
 
 
 @irdl_op_definition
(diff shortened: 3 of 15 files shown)


xdsl/dialects/bufferization.py
```diff
@@ -41,42 +48,29 @@ class TensorMemrefInferenceConstraint(VarConstraint[Attribute]):
...
-                    f"Unexpected {self.name} - cannot verify shape of attribute {attr}"
-                )
-        elif isinstance(attr, ContainerType):
-            self.constraint.verify(attr, constraint_context)
-            constraint_context.set_variable(self.name, attr)
-        else:
-            raise VerifyException(
-                f"Unexpected {self.name} - attribute must be ContainerType"
-            )
+        if isa(attr, TensorType[Attribute]):
```
superlopuh, line 67:
```suggestion
        if isinstance(attr, TensorType):
```

xdsl/irdl/constraints.py
```diff
@@ -1,17 +1,29 @@
...
+
+ResolveType: TypeAlias = Attribute | Sequence[Attribute]
+
+
+def print_resolve_type(r: ResolveType | None) -> str:
+    if isinstance(r, Attribute):
+        return str(r)
+    elif isinstance(r, Sequence):
+        return str(tuple(str(x) for x in r))
+    return "None"
```
superlopuh, line 25:
```suggestion
def _format_resolve_type(r: ResolveType | None) -> str:
    if isinstance(r, Attribute):
        return str(r)
    elif isinstance(r, Sequence):
        return str(tuple(str(x) for x in r))
    return "None"
```

superlopuh: Also a doc string would be good

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -1748,10 +1748,12 @@ class OneOperandOneResultNestedOp(IRDLOperation):
     %1 = test.one_operand_one_result_nested %0 : i32"""
     )
     with pytest.raises(
-        ParseError,
-        match="Verification error while inferring operation type: ",
+        VerifyException,
+        match="i32 should be of base attribute test.param_one",
```
superlopuh, line 1752: This seems like it conveys the message a bit more directly
```suggestion
        match="unexpected attribute i32, expected test.param_one",
```

xdsl/irdl/constraints.py
```diff
@@ -560,22 +571,24 @@ def verify(
...
-        constraint_context = constraint_context or ConstraintContext()
-        if self.name not in constraint_context.range_variables:
-            raise ValueError(f"Cannot infer attribute from constraint {self}")
-        return constraint_context.get_range_variable(self.name)
+        self,
+        length: int,
+        variables: dict[str, ResolveType],
+    ) -> Sequence[AttributeCovT]:
+        v = variables[self.name]
+        return cast(Sequence[AttributeCovT], v)
```
superlopuh, line 588: Does this work?
```suggestion
        assert not isinstance(V, Attribute)
        return v
```

xdsl/irdl/constraints.py
```diff
@@ -591,19 +604,19 @@ def verify(
...
     def infer(
-        self, length: int, constraint_context: ConstraintContext
-    ) -> list[Attribute]:
-        return [self.constr.infer(constraint_context)] * length
+        self,
+        length: int,
+        variables: dict[str, ResolveType],
+    ) -> Sequence[AttributeCovT]:
+        attr = self.constr.infer(variables)
+        return tuple(attr for _ in range(length))
```
superlopuh, line 616:
```suggestion
        return (attr,) * length
```

superlopuh (commented):
Overall looks good, although I'm still not sure what the motivation for this is. Is it better error messages? Or is there more downstream? Also, my first instinct would be to somehow be a bit more lazy about the resolver situation, would it make sense to make them into lazy model objects, that let the user iterate over the keys, and define each function only once, instead of creating wrappers of wrappers. Not 100% sure if it would work, but might be worth a try in this PR?

xdsl/irdl/constraints.py
```diff
@@ -1,17 +1,29 @@
...
-from typing import Generic, TypeAlias, TypeVar
+from typing import Generic, TypeAlias, TypeVar, cast
 
 from typing_extensions import assert_never
 
 from xdsl.ir import Attribute, AttributeCovT, ParametrizedAttribute
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.runtime_final import is_runtime_final
 
+_AttributeCovT = TypeVar("_AttributeCovT", bound=Attribute, covariant=True)
```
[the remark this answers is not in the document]
alexarice: This is (possibly accidentally) moved instead of being new. It's used in the definition of `__or__` (which uses both this and the existing one.

xdsl/irdl/constraints.py
alexarice: No

alexarice:
I tried to put some motivation in the PR description and there is a bit more discussion in #3318 about needing to verify properties during parsing with the old system (I have rebased that PR onto this PR).

xdsl/irdl/constraints.py
superlopuh: Sad

xdsl/irdl/constraints.py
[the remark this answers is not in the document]
superlopuh: Ah ok sounds good, thank you

tests/irdl/test_declarative_assembly_format.py
superlopuh: hmm actually is this just an existing constraint string? In that case I won't recommend PR creep, it might just be worth changing in the future.

superlopuh:
Thank you, happy to approve once the CI passes

xdsl/irdl/constraints.py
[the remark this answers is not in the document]
alexarice: I've moved it back

xdsl/irdl/constraints.py
alexarice: Seems like this is dead code so I've removed it

tests/irdl/test_declarative_assembly_format.py
alexarice: The key change here is that the exception happens during verification, rather than at parse time

xdsl/irdl/constraints.py
```diff
@@ -72,24 +74,23 @@ def verify(
         """
         ...
 
-    def get_resolved_variables(self) -> set[str]:
+    def get_resolvers(self) -> dict[str, Callable[[AttributeCovT], ResolveType]]:
```
alexarice, line 77: It might be worth making a type alias for this, as it appears in a few places

alexarice:
>  Also, my first instinct would be to somehow be a bit more lazy about the resolver situation, would it make sense to make them into lazy model objects, that let the user iterate over the keys, and define each function only once, instead of creating wrappers of wrappers. Not 100% sure if it would work, but might be worth a try in this PR?

I'm also not sure how happy I am about stacks of wrappers. Another option which could be worth trying is to have a resolver be a list of "accessors" where each accessor is a piece of data (such as the attribute name, or operand index). I haven't tried putting this into action so I'm not sure how messy it would get.

Could you expand on what you mean by "lazy model objects"?

superlopuh:
Something like
```
class AbstractResolver(abc.ABC):
    @abstractmethod
    def iter_variables() -> Iterator[str]: ...

     @abstractmethod
    def resolve(attr: Attribute) -> Attribute | Sequence[Attribute]:
        ...

@dataclass
class WhateverResolver(AbstractResolver):

    a: AbstractResolver
    b: AbstractResolver

    def iter_variables() -> Iterator[str]:
        yield from a.iter_variables()
        yield from b.iter_variables()

    def resolve(attribute: Attribute) -> Attribute | Sequence[Attribute]:
        return a.resolve(attribute) if creative_condition else b.resolve(attribute)
```

alexarice:
I like that idea, I'll give it a go next week

alexarice:
Have merged #3464 here, so this is ready for review again (Think there are no outstanding open comments)

xdsl/irdl/constraints.py
alexarice: Replaced this by object-based resolvers so this is no longer relevant

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -241,20 +243,46 @@ def add_reserved_attrs_to_directive(self, elements: list[FormatDirective]):
...
             if self.seen_result_types[i]:
-                seen_variables |= result_def.constr.get_resolved_variables()
-        return seen_variables
+                resolvers.update(
+                    {
+                        v: self._OperandResultResolver(i, False, r)
+                        for v, r in result_def.constr.get_resolvers().items()
+                    }
+                )
+        return resolvers
```
superlopuh, line 283: What happens if there's an overlap in `v`s? Worth a comment if that's not possible? My instinct here would be to create two dictionaries, probably with two dictionary comprehensions, and then take their union. (Maybe assert that the intersection of keys is empty)

xdsl/dialects/bufferization.py
```diff
@@ -42,42 +47,29 @@ class TensorMemrefInferenceConstraint(VarConstraint[Attribute]):
...
+            memref_type = MemRefType(attr.element_type, attr.shape)
+            return self.memref_var_constr.verify(memref_type, constraint_context)
+        if isa(attr, UnrankedTensorType[Attribute]):
+            memref_type = UnrankedMemrefType.from_type(attr.element_type)
+
+        raise VerifyException(f"Expected TensorType or UnrankedTensorType, got {attr}")
```
superlopuh, line 72: the logic here seems to be changed

xdsl/irdl/constraints.py
```diff
@@ -72,24 +82,23 @@ def verify(
...
         """
-        Get the set of type variables that are always resolved when verifying
-        the constraint.
+        Get a dictionary of variables that can be solved from this attribute.
         """
-        return set()
+        return dict()
 
-    def can_infer(self, constraint_names: set[str]) -> bool:
+    def can_infer(self, variables: Set[str]) -> bool:
```
superlopuh, line 91: How do you feel about this?
```suggestion
    def can_infer(self, `var_constraint_names`: Set[str]) -> bool:
```
I probably should have commented this on the previous PR, my bad

xdsl/irdl/constraints.py
```diff
@@ -339,21 +351,19 @@ def verify(
...
-    def get_resolved_variables(self) -> set[str]:
-        if len(self.attr_constrs) == 0:
-            return set()
-        return set[str].union(
-            *[constr.get_resolved_variables() for constr in self.attr_constrs]
-        )
+    def get_resolvers(self) -> dict[str, Resolver[AttributeCovT]]:
+        d: dict[str, Resolver[AttributeCovT]] = dict()
+        for constr in self.attr_constrs:
+            d |= constr.get_resolvers()
```
superlopuh, line 357: what happens if there are overlapping keys here?

xdsl/dialects/builtin.py
```diff
@@ -1637,6 +1632,49 @@ def constr(
 AnyMemRefTypeConstr = BaseAttr[MemRefType[Attribute]](MemRefType)
 
 
+@dataclass(frozen=True, init=False)
+class TensorOrMemrefOf(
+    GenericAttrConstraint[TensorType[AttributeCovT] | MemRefType[AttributeCovT]]
+):
+    """A type constraint that can be nested once in a vector or a tensor."""
```
superlopuh, line 1639:
```suggestion
    """A type constraint that can be nested once in a memref or a tensor."""
```

superlopuh:
How do you feel about "variable assignment" as opposed to "resolving" variables?

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: If there's an overlap in `v`s then the second one encountered will be used. This would be happen if a variable can be inferred from two different places in the assembly format. Which resolver is used shouldn't make any difference for any operation that successfully verifies.

xdsl/dialects/bufferization.py
alexarice: It is, though it appears I forgot to amend the doc comment. In my opinion the previous constraint was an abuse of the particular workings of the previous system. The one I have written here is a bit more well-behaved with respect to inference, and can be used to achieve the same effect as the old one. I'll update the doc comment.

xdsl/irdl/constraints.py
alexarice: I assume you didn't mean to include the backticks? It seems reasonable to me.

xdsl/irdl/constraints.py
alexarice: Overlapping keys get overwritten

alexarice:
> How do you feel about "variable assignment" as opposed to "resolving" variables?

What would the concrete change be here? Open to different names for some/any of these things, I just went with the first thing that came to me and stuck with it

xdsl/irdl/constraints.py
superlopuh: I didn't indeed

xdsl/irdl/declarative_assembly_format_parser.py
superlopuh: how do you feel about a resolver that asserts at runtime that the two resolved results are the same?

xdsl/irdl/constraints.py
alexarice: There was a small chance it was some obscure python syntax I'd not seen before

superlopuh: it's obscure Swift and Mojo syntax!

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: I imagine this would be easy to do with the new class-based system. I'm not sure how I feel about putting what is effectively verification as a check in parsing.

xdsl/irdl/constraints.py
superlopuh: Is the idea here also that the expectation is that the resolvers will return identical results, so this is not an issue?

alexarice: Yes

xdsl/dialects/bufferization.py
superlopuh: Would it make sense to merge this change as a separate PR, while we discuss the details of this one?

alexarice: Sure, I'll get that ready first thing tomorrow

xdsl/irdl/constraints.py
```diff
@@ -55,6 +56,15 @@ def update(self, other: ConstraintContext):
...
 _AttributeCovT = TypeVar("_AttributeCovT", bound=Attribute, covariant=True)
 
+ResolveType: TypeAlias = Attribute | Sequence[Attribute]
+
+_T = TypeVar("_T")
+
+
+class Resolver(Generic[_T], abc.ABC):
+    @abstractmethod
+    def resolve(self, a: _T) -> ResolveType: ...
```
superlopuh, line 66: This would benefit from a lot more doc commenting, seems like the perfect place for a long-form explanation of why this is useful and how the whole mechanism works. I'm hoping that a plain English explanation will help pick the right names for the functions and function arguments involved

xdsl/irdl/constraints.py
```diff
@@ -55,6 +56,15 @@ def update(self, other: ConstraintContext):
 
 _AttributeCovT = TypeVar("_AttributeCovT", bound=Attribute, covariant=True)
 
+ResolveType: TypeAlias = Attribute | Sequence[Attribute]
```
math-fehr, line 59: Can you add documentation to that alias, so it makes sense what it is?

xdsl/irdl/constraints.py
```diff
@@ -55,6 +56,15 @@ def update(self, other: ConstraintContext):
...
+
+_T = TypeVar("_T")
+
+
+class Resolver(Generic[_T], abc.ABC):
```
math-fehr, line 64: Isn't this an `AttrInference` or something?
I feel using the `Inference` word makes more sense than `Resolve` here

xdsl/irdl/constraints.py
```diff
@@ -1,14 +1,15 @@
 from __future__ import annotations
 
+import abc
 from abc import ABC, abstractmethod
-from collections.abc import Sequence
+from collections.abc import Sequence, Set
```
math-fehr, line 5: Is there a reason you are using `Set` instead of `set`?
AFAIK, we should always prefer `set` since Python 3.something?

xdsl/irdl/constraints.py
```diff
@@ -72,24 +82,23 @@ def verify(
...
+        return dict()
 
-    def can_infer(self, constraint_names: set[str]) -> bool:
+    def can_infer(self, variables: Set[str]) -> bool:
         """
         Check if there is enough information to infer the attribute given the
         constraint variables that are already set.
         """
         # By default, we cannot infer anything.
         return False
```
math-fehr, line 97: Because we have get_resolvers, this function should always be defined?
Is there a reason we could infer something that doesn't have a resolver, or vice versa?

math-fehr (approved):
Looks neat!
I'm trusting Sasha for the review, but on my end that looks good!

xdsl/irdl/constraints.py
alexarice: I was considering something along the lines of `Extractor`. The idea is that it describes how to get the value of a variable from the input `_T` (which could be an attribute, sequence of attributes, ParsingState etc)

xdsl/irdl/constraints.py
alexarice: `Set` is more general and in particular includes `dict.keys()`. Note that `collections.abc.Set =/= typing.Set`

xdsl/irdl/constraints.py
alexarice: Any `EqAttrConstraint` can be inferred but does not resolve any variables. A `ParamAttrConstraint` may be able to provide variable resolutions without being totally inferrable

alexarice: And in general there are lots of things that cannot be inferred, and must be given in the assembly format

xdsl/irdl/constraints.py
math-fehr: Hmm, I see what you mean, I think `Extractor` makes more sense then indeed! Or `VarExtractor` to be more precise?

xdsl/irdl/constraints.py
math-fehr: Oh god, I did not know about `typing.Set` versus `collections.abc.Set`...

alexarice: Makes autocompletion very annoying, especially since `typing.Set` is useless since Python 3.something as you said

xdsl/irdl/constraints.py
math-fehr: Oh yes you're right, my bad.

xdsl/irdl/constraints.py
superlopuh: Also Set and Dict are immutable collections if I remember correctly, it also helps with variance things

alexarice:
I believe all comments have been resolved. 
- I've renamed `Resolvers` to `VarExtractors` and given them a (admittedly small) docstring. 
- I have also introduced a `MergeExtractor` which is used to record that two extractors have been provided for the same variable. This performs a runtime check to determine that each extractor would return the same thing. I have added 3 tests for each case where this is used, using slightly contrived operations.
-  I have rebased on top of the change to the constraint in bufferization.

alexarice:
Forgot to change the `can_infer` method argument name, will do that now

tests/dialects/test_bufferization.py
```diff
@@ -40,13 +39,13 @@ def test_tensor_from_memref_inference():
         EqAttrConstraint(MemRefType(f64, [10, 20, 30]))
     )
     assert constr2.can_infer(set())
-    assert constr2.infer(ConstraintContext()) == TensorType(f64, [10, 20, 30])
+    assert constr2.infer(dict()) == TensorType(f64, [10, 20, 30])
```
superlopuh, line 42: ?
```suggestion
    assert constr2.infer({}) == TensorType(f64, [10, 20, 30])
```

xdsl/irdl/constraints.py
```diff
@@ -55,6 +56,73 @@ def update(self, other: ConstraintContext):
...
+        if len(res) == 1:
+            return res.pop()
+        else:
+            raise ValueError(
+                f"Value of variable {self.variable} could not be uniquely extracted.\n"
+                f"Possible values are: {{{', '.join(sorted(str(x) for x in res))}}}"
+            )
+
+    def flatten(self) -> Iterator[VarExtractor[_T]]:
+        return self.extractors.__iter__()
```
superlopuh, line 100:
```suggestion
        return iter(self.extractors)
```

xdsl/irdl/constraints.py
```diff
@@ -470,23 +557,22 @@ def verify(
...
+    def infer(self, variables: dict[str, ConstraintVariableType]) -> AttributeCovT:
+        return self.constr.infer(variables)
 
 
+@dataclass(frozen=True)
 class GenericRangeConstraint(Generic[AttributeCovT], ABC):
-    """
-    Constrain a range of attributes to a certain value.
-    """
+    """Constrain an range of attributes to certain values."""
```
superlopuh, line 575:
```suggestion
    """Constrain a range of attributes to certain values."""
```

xdsl/irdl/constraints.py
```diff
@@ -500,32 +586,39 @@ def verify(
...
+    ) -> dict[str, VarExtractor[Sequence[AttributeCovT]]]:
         """
-        Get the set of type variables that are always resolved when verifying
-        the constraint.
+        Get a dictionary of constraint variables to extractors for these variables,
+        which provide a method to obtain the value of each constraint variable from
+        the value of the attribute range which is verified by this constraint.
         """
-        return set()
+        return dict()
```
superlopuh, line 597:
```suggestion
        return {}
```

xdsl/irdl/constraints.py
```diff
@@ -72,24 +140,25 @@ def verify(
...
+    def get_variable_extractors(self) -> dict[str, VarExtractor[AttributeCovT]]:
         """
-        Get the set of type variables that are always resolved when verifying
-        the constraint.
+        Get a dictionary of constraint variables to extractors for these variables,
+        which provide a method to obtain the value of each constraint variable from
+        the value of the attribute which is verified by this constraint.
         """
-        return set()
+        return dict()
```
superlopuh, line 149:
```suggestion
        return {}
```

superlopuh (approved):

alexarice:
@superlopuh do you feel the extractor merging is worth the added complexity? My examples in the tests feel a bit contrived

superlopuh:
I feel like it's OK, I'm just imagining the headache that would emerge from trying to debug this, and it all seems worth it.

superlopuh:
I just merged the small conflicting PR, hopefully will be easy to resolve. I think the best order now is for you to merge this and for me to resolve my draft PRs.

PR #3457, opened 2024-11-17
interpreter: (pdl) fix match_operand matching for already matched value

superlopuh:
This is the first PR adding some unit tests for the existing matcher before opening the PRs for the eqsat version.


xdsl/interpreters/pdl.py
```diff
@@ -61,7 +61,7 @@ def match_operand(
         self, ssa_val: SSAValue, pdl_op: pdl.OperandOp, xdsl_val: SSAValue
     ):
         if ssa_val in self.matching_context:
-            return True
```
superlopuh, line 64: I'm not sure why it was implemented in this way before, but it doesn't make sense to me today, am I missing something?

PR #3462, opened 2024-11-18
dialects (arm): add first operation

emmau678:
Add a first op for getting a register and a filecheck
Note: stacked PR

### DIFF
--- a/xdsl/dialects/arm/ops.py
+++ b/xdsl/dialects/arm/ops.py
@@ -1,38 +1,28 @@
-from abc import ABC, abstractmethod
-from typing import Generic, TypeVar
+from abc import ABC
 
 from xdsl.irdl import IRDLOperation, irdl_op_definition, result_def
 
-from .register import ARMRegisterType
-
-R1InvT = TypeVar("R1InvT", bound=ARMRegisterType)
+from .register import IntRegisterType
 
 
-class ARMOp(IRDLOperation, ABC):
+class ARMOperation(IRDLOperation, ABC):
     """
     Base class for operations that can be a part of ARM assembly printing.
     """
 
-    @abstractmethod
-    def assembly_line(self) -> str | None:
-        raise NotImplementedError()
+    ...
 
 
-class GetAnyRegisterOperation(Generic[R1InvT], ARMOp, ABC):
+@irdl_op_definition
+class GetRegisterOp(ARMOperation):
     """
     This instruction allows us to create an SSAValue for a given register name.
     """
 
-    result = result_def(R1InvT)
-
-    def __init__(self, register_type: R1InvT):
-        super().__init__(result_types=[register_type])
-
-    def assembly_line(self) -> str | None:
-        return None
-
-
-@irdl_op_definition
-class GetRegisterOp(GetAnyRegisterOperation[ARMRegisterType]):
     name = "arm.get_register"
+
+    result = result_def(IntRegisterType)
     assembly_format = "attr-dict `:` type($result)"
+
+    def __init__(self, register_type: IntRegisterType):
+        super().__init__(result_types=[register_type])
--- a/xdsl/dialects/arm/ops.py
+++ b/xdsl/dialects/arm/ops.py
@@ -9,7 +9,7 @@ from .register import IntRegisterType
 R1InvT = TypeVar("R1InvT", bound=IntRegisterType)
 
 
-class ARMOp(Operation, ABC):
+class ARMOp(IRDLOperation, ABC):
     """
     Base class for operations that can be a part of ARM assembly printing.
     """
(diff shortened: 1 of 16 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -2474,7 +2474,7 @@ class RM_VbroadcastsdOp(R_RM_Operation[AVXRegisterType, GeneralRegisterType]):
     name = "x86.rm.vbroadcastsd"
 
 
-class GetAnyRegisterOperation(Generic[R1InvT], IRDLOperation, X86Op):
+class GetAnyRegisterOperation(Generic[R1InvT], IRDLOperation, X86Op, ABC):
```
alexarice, line 2477: What's the reason for including this change?

xdsl/dialects/arm/ops.py
```diff
@@ -0,0 +1,38 @@
...
+
+from xdsl.ir import Operation
+from xdsl.irdl import IRDLOperation, irdl_op_definition, result_def
+
+from .register import IntRegisterType
+
+R1InvT = TypeVar("R1InvT", bound=IntRegisterType)
+
+
+class ARMOp(Operation, ABC):
```
alexarice, line 12: It might make sense to have the following?
```suggestion
class ARMOp(IRDLOperation, ABC):
```
Then you can remove `IRDLOperation` below.

xdsl/dialects/arm/ops.py
```diff
@@ -0,0 +1,38 @@
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+from xdsl.ir import Operation
+from xdsl.irdl import IRDLOperation, irdl_op_definition, result_def
+
+from .register import IntRegisterType
+
+R1InvT = TypeVar("R1InvT", bound=IntRegisterType)
```
alexarice, line 9: Are there plans to have more register types? This seems like a premature generalisation at this point

xdsl/dialects/arm/ops.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ARMOp(Operation, ABC):
+    """
+    Base class for operations that can be a part of ARM assembly printing.
+    """
+
+    @abstractmethod
+    def assembly_line(self) -> str | None:
+        raise NotImplementedError()
```
alexarice, line 19: I don't think there's any need to give an implementation for an abstract method? (I could be very wrong on this, perhaps @superlopuh can correct me)

alexarice (commented):
Seems good, I've left a few comments which may need no/limited action

xdsl/dialects/x86/ops.py
compor, line 2477: is this by accident? if not, maybe we add this for this class in the `riscv` dialect too?

compor (approved):

xdsl/dialects/x86/ops.py
compor: I meant piggyback it in this PR too

superlopuh: That was my doing, it was just missing this annotation, and porting the equivalent to ARM meant that it was also missed, it's a small non-functional change

xdsl/dialects/arm/ops.py
superlopuh: My understanding is that this is good practice. In theory, the annotation should raise an error if not overriden, but I'm not sure if that ever breaks (and what happens if the `@abstractmethod` gets lost somehow.

xdsl/dialects/x86/ops.py
superlopuh: Oh yeah that would probably be a good idea.

xdsl/dialects/arm/ops.py
superlopuh: Good point, we probably don't need it right now, let's add it later.

xdsl/dialects/arm/ops.py
```diff
@@ -0,0 +1,39 @@
...
+    This instruction allows us to create an SSAValue for a given register name.
+    """
+
+    result = result_def(R1InvT)
+
+    def __init__(self, register_type: R1InvT):
+        super().__init__(result_types=[register_type])
+
+    def assembly_line(self) -> str | None:
+        return None
```
superlopuh, line 33: If we're starting with specific ops, and generalizing later, it would make sense to move this into the GetRegisterOp for now, and to generalize when we add float registers.

alexarice: If there are immediate plans to add more register types, then I'd be happy with leaving it as is

superlopuh: I'm not sure how immediate they are, and we may well want to put them in a different place when that happens, so it's not a bad idea to make it specific for the time being

superlopuh (approved):

PR #3463, opened 2024-11-18
dialects (arm): add custom syntax

emmau678:
Add custom ARM assembly syntax and generic roundtrip test to filecheck.
Note: stacked PR

### DIFF
--- a/tests/filecheck/dialects/arm/test_ops.mlir
+++ b/tests/filecheck/dialects/arm/test_ops.mlir
@@ -1,5 +1,9 @@
 // RUN: XDSL_ROUNDTRIP
+// RUN: XDSL_GENERIC_ROUNDTRIP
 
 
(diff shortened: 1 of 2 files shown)


compor (approved):

alexarice (approved):
I'd be happy to merge this into the parent PR and review them as one

superlopuh:
@alexarice I don't get it, why not review them separately?

alexarice:
I don't see the value of splitting this into 2 prs where the custom syntax is simple, and think it just creates more work in generating filecheck tests for the intermediate version without custom format.

alexarice:
I admit though that here the work in splitting it had already been done so it was probably the wrong call. Sorry for that.

superlopuh:
No worries, I see what you mean. We actually did a bit of work to make it as simple as it was in this PR, which you can infer by looking at the logic riscv.py file and what mechanism is used for the custom syntax there :)

PR #3464, opened 2024-11-18
Use classes in place of callables for resolvers

alexarice:
This implements something halfway between what I originally had in #3456 and what @superlopuh suggested in https://github.com/xdslproject/xdsl/pull/3456#issuecomment-2479586661 . I think this a strict improvement over the original which uses callables and lambda functions and will merge it into that PR if people agree

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -1,14 +1,15 @@
 from __future__ import annotations
 
+import abc
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Sequence, Set
+from collections.abc import Sequence, Set
 from dataclasses import dataclass, field
 from inspect import isclass
 from typing import Generic, TypeAlias, TypeVar, cast
 
 from typing_extensions import assert_never
 
-from xdsl.ir import Attribute, AttributeCovT, ParametrizedAttribute
+from xdsl.ir import Attribute, AttributeCovT, AttributeInvT, ParametrizedAttribute
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.runtime_final import is_runtime_final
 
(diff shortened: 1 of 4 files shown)


xdsl/irdl/constraints.py
```diff
@@ -55,7 +56,14 @@ def update(self, other: ConstraintContext):
 
 _AttributeCovT = TypeVar("_AttributeCovT", bound=Attribute, covariant=True)
 
-ResolveType: TypeAlias = Attribute | Sequence[Attribute]
+ResolveType = Attribute | Sequence[Attribute]
```
superlopuh, line 59: This is inferred but also doesn't hurt, makes sure that the RHS is purely in type land
```suggestion
ResolveType: TypeAlias = Attribute | Sequence[Attribute]
```

xdsl/irdl/constraints.py
```diff
@@ -55,7 +56,14 @@ def update(self, other: ConstraintContext):
...
+
+T = TypeVar("T")
```
superlopuh, line 61: Any reason to make this public?
```suggestion
_T = TypeVar("_T")
```

superlopuh (approved):
LGTM, except that I'd probably move the resolver classes into the constraint classes, and call them all `_Resolver`

alexarice:
I originally had this exactly as you said with inner classes, and then got worried that this would redefine the classes each time the outer class was used, and then read some stack exchange threads that said that inner classes were stupid, and so I got cold feet about it. I can change it back easily

superlopuh:
I'd be interested in seeing the arguments against inner classes, I'm quite used to them and like them in general

alexarice:
https://stackoverflow.com/questions/719705/what-is-the-purpose-of-pythons-inner-classes this thread had a few things in the comments though I found the argument questionable

PR #3465, opened 2024-11-18
interpreter: (pdl) add test for result matching

superlopuh:
Another PR adding tests to existing PDL infrastructure.

Note stacked PR


jianyicheng (approved):
I see what you meant but it may be helpful to add singe-sentence comment to explain what each assertion is doing. They look all the same and took me a few mins to figure out...

compor (approved):

PR #3470, opened 2024-11-19
dialects: (stream) remove unused stream.read and stream.write ops

superlopuh:
It's been a while since these have been used. This PR makes some more small changes:

 - removes the interpreter functions
 - moves remaining interpreter helper structures to own file in utils


alexarice (approved):
I can't see any immediate issues though don't really know any context for this dialect, so it might be better to wait for the review of someone that knows what is going on

superlopuh:
The context is that I forgot to delete them a while ago :)

PR #3472, opened 2024-11-19
misc: split dialects.utils into multiple files

superlopuh:
It feels like a logical split, and leaves space for some upcoming shared dialects helpers.


xdsl/dialects/utils/format.py
```diff
@@ -1,6 +1,4 @@
-from abc import ABC
 from collections.abc import Iterable, Sequence
```
alexarice, line 1: I'm not sure if `format` is really the best name to characterise these helpers. Have you thought about further splitting them into printer utils and parsing utils?

superlopuh: These tend to be quite coupled to each other, so it made sense to keep them together. I agree that the name isn't great, but I can't think of a better one... We could also go with the irdl and ir route and ban the sub files of the utils, and expose it all via `xdsl.dialects.utils`

superlopuh: I'll try that quickly, will result in smaller diff and probably a better experience

alexarice: I'm not sure I understand the point of the irdl approach, just seems to cause loads of linting errors for me

superlopuh: For some reason, the local development experience is worse than for clients, Pylance automatically inserts the correct import if you pip install it.

pyproject.toml
superlopuh, line 1: sorted this list

math-fehr (approved):
Looks good to me!

PR #3473, opened 2024-11-19
dialects: (stream) simplify constr helper on stream attributes

superlopuh:


xdsl/dialects/stream.py
```diff
@@ -79,66 +60,20 @@ def constr(
...
-        | ParamAttrConstraint[WritableStreamType[_StreamTypeElementConstrT]]
-    ):
-        if element_type is None:
-            return BaseAttr[WritableStreamType[Attribute]](WritableStreamType)
-        return ParamAttrConstraint[WritableStreamType[_StreamTypeElementConstrT]](
-            WritableStreamType, (element_type,)
-        )
+AnyWritableStreamTypeConstr = BaseAttr[WritableStreamType[Attribute]](
+    WritableStreamType
+)
```
superlopuh, line 76: I have a feeling that this will be the pattern to use until we have defaults for TypeVars

PR #3477, opened 2024-11-19
core: (irdl) ParamAttrConstraint can infer recursively

superlopuh:
This seems to work, is there any reason not to do this?

Note: conflicts with #3456

### DIFF
--- a/tests/irdl/test_attr_constraint.py
+++ b/tests/irdl/test_attr_constraint.py
@@ -60,11 +60,13 @@ def test_attr_constraint_get_unique_base(
 
 
 def test_param_attr_constraint_inference():
-    @irdl_attr_definition
-    class WrapAttr(ParametrizedAttribute):
+    class BaseWrapAttr(ParametrizedAttribute):
         name = "wrap"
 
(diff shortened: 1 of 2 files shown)


alexarice (commented):
I don't believe this is sound, `ParamAttrConstraint` only specifies the base class of the attribute and so could theoretically this could infer the wrong subclass?

alexarice:
Note that I'm not sure if being able to capture subclasses is part of the intended semantics of `BaseAttr` and `ParamAttrConstraint`, and this might just be an artifact of using python's `isinstance`

alexarice:
Could do an `is_runtime_final` check?

math-fehr (approved):
I think I either forgot to do this, or at the time `is_runtime_final` didn't exist yet!
So yeah that looks good to me!

math-fehr:
> A similar question to this is whether a `BaseAttr` applied to a type with no parameters should be inferable

Yes it should! What we could do is write an optimization to "convert" a `BaseAttr` with a singleton type to an `EqType`, that would also work!

PR #3480, opened 2024-11-19
dialects: (tensor) add insertop and extractop

francescodaghero:
This PR adds the following operations to the tensor dialect:
- tensor.InsertOp [docs](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert-tensorinsertop)
- tensor.ExtractOp [docs](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorextract-tensorextractop)

that I currently use for a tree ensemble compiler implemented using xdsl ([link](https://github.com/francescodaghero/treeco/tree/main)).

I have tried to mirror the operations available in MLIR, customizing both the print/parse functions, but I am not 100% sure I did it correctly, as this is my first PR.

### DIFF
--- a/tests/filecheck/dialects/tensor/ops.mlir
+++ b/tests/filecheck/dialects/tensor/ops.mlir
@@ -6,8 +6,8 @@
 %dim2 = "tensor.dim"(%tensor, %index) {"hello" = "world"}: (tensor<?x?xf32>, index) -> index
 %cast1 = "tensor.cast"(%tensor) : (tensor<?x?xf32>) -> tensor<4x4xf32>
 %cast2 = "tensor.cast"(%tensor) {"hello" = "world"} : (tensor<?x?xf32>) -> tensor<4x4xf32>
-%extract1 = "tensor.extract"(%tensor, %index, %index1) : (tensor<?x?xf32>, index, index) -> f32
-%insert1 = "tensor.insert"(%extract1, %tensor, %index, %index1) : (f32, tensor<?x?xf32>, index, index) -> tensor<?x?xf32>
+%extract1 = tensor.extract %tensor[%index, %index1] : tensor<?x?xf32>
+%insert1 = tensor.insert %extract1 into %tensor[%index, %index1] : tensor<?x?xf32>
 
 
 // CHECK: builtin.module {
--- a/tests/filecheck/dialects/tensor/ops.mlir
+++ b/tests/filecheck/dialects/tensor/ops.mlir
@@ -1,24 +1,31 @@
 // RUN: XDSL_ROUNDTRIP
 // RUN: XDSL_GENERIC_ROUNDTRIP
 
-%index, %tensor = "test.op"() : () -> (index, tensor<?x?xf32>)
+%index, %index1, %tensor = "test.op"() : () -> (index, index, tensor<?x?xf32>)
 %dim1 = "tensor.dim"(%tensor, %index): (tensor<?x?xf32>, index) -> index
 %dim2 = "tensor.dim"(%tensor, %index) {"hello" = "world"}: (tensor<?x?xf32>, index) -> index
 %cast1 = "tensor.cast"(%tensor) : (tensor<?x?xf32>) -> tensor<4x4xf32>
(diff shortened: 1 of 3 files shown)


superlopuh (commented):
This is great! A few comments:

 - could you please add an interop test with MLIR? We already have tests here: `tests/filecheck/dialects/tensor/ops.mlir`
 - could you please have a go at using `assembly_format` instead? There are many examples throughout the repo, and [here](https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-format) is the MLIR doc on the syntax
 - could you please use `__init__` instead of get, we've transitioned to that as a framework, should be minimal changes, also plenty of examples available

superlopuh:
I'm hoping to add bufferization to xDSL before the end of the month, interesting that you also would benefit from it! I'd love to discuss your project if you have some time, please ping me on Zulip

francescodaghero:
Thanks for your reply,

> This is great! A few comments:
> 
> * could you please add an interop test with MLIR? We already have tests here: `tests/filecheck/dialects/tensor/ops.mlir`

This change should already be in this commit, do I need to modify other parts of the ops.mlir? 

> * could you please have a go at using `assembly_format` instead? There are many examples throughout the repo, and [here](https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-format) is the MLIR doc on the syntax

I see, I did an initial try for insertOp for instance, whose assembly_format should be : 
```python
"$scalar `into` $dest `[` $indices `]` attr-dict `:` type($dest)"
```
(the same as in MLIR), however, this fails with the following error:
```text
# | xdsl.utils.exceptions.PyRDLOpDefinitionError: ('Error during the parsing of the assembly format: ', (Span[63:63](text=''), "type of operand 'scalar' cannot be inferred, consider adding a 'type($scalar)' directive to the custom assembly format"))
```
which makes sense, but I don't know how to solve this issue for xDSL without changing the assembly format w.r.t. the MLIR one.


> * could you please use `__init__` instead of get, we've transitioned to that as a framework, should be minimal changes, also plenty of examples available

Sure, no problem.

superlopuh:
Sorry, I sent the wrong path, I meant this one: `tests/filecheck/mlir-conversion/with-mlir/dialects/tensor/ops.mlir` we should ideally have both.

tests/filecheck/dialects/tensor/ops.mlir
```diff
@@ -1,24 +1,31 @@
...
 // RUN: XDSL_GENERIC_ROUNDTRIP
 
-%index, %tensor = "test.op"() : () -> (index, tensor<?x?xf32>)
+%index, %index1, %tensor = "test.op"() : () -> (index, index, tensor<?x?xf32>)
 %dim1 = "tensor.dim"(%tensor, %index): (tensor<?x?xf32>, index) -> index
 %dim2 = "tensor.dim"(%tensor, %index) {"hello" = "world"}: (tensor<?x?xf32>, index) -> index
 %cast1 = "tensor.cast"(%tensor) : (tensor<?x?xf32>) -> tensor<4x4xf32>
 %cast2 = "tensor.cast"(%tensor) {"hello" = "world"} : (tensor<?x?xf32>) -> tensor<4x4xf32>
+%extract1 = "tensor.extract"(%tensor, %index, %index1) : (tensor<?x?xf32>, index, index) -> f32
+%insert1 = "tensor.insert"(%extract1, %tensor, %index, %index1) : (f32, tensor<?x?xf32>, index, index) -> tensor<?x?xf32>
```
superlopuh, line 10: While you're making changes here, could you please rewrite the inputs in custom syntax? Will be a little easier to read for the next person

superlopuh:
I think I have a way for the custom syntax to work, it might be worth a try once this PR is in:  #3477

superlopuh:
Code LGTM, I'd just like to try the approach enabled by the PR I linked if it is accepted in the next day or two

superlopuh (approved):
Actually let's merge this, I'll fiddle with it on my own time

superlopuh:
Thank you!

PR #3485, opened 2024-11-20
dialects (arm): add assembly printing

emmau678:
Add assembly printing for ARM instructions
Note: stacked PR

### DIFF
--- a/tests/filecheck/dialects/arm/test_ops.mlir
+++ b/tests/filecheck/dialects/arm/test_ops.mlir
@@ -1,5 +1,6 @@
 // RUN: XDSL_ROUNDTRIP
 // RUN: XDSL_GENERIC_ROUNDTRIP
+// RUN: xdsl-opt -t arm-asm %s | filecheck %s --check-prefix=CHECK-ASM
 
 
 // CHECK: %x1 = arm.get_register : !arm.reg<x1>
@@ -8,5 +9,6 @@
 // CHECK: %ds_mov = arm.ds.mov %x1 : (!arm.reg<x1>) -> !arm.reg<x2>
 %ds_mov = arm.ds.mov %x1 : (!arm.reg<x1>) -> !arm.reg<x2>
 
+// CHECK-ASM: mov x2, x1
 // CHECK-GENERIC: %x1 = "arm.get_register"() : () -> !arm.reg<x1>
 // CHECK-GENERIC: %ds_mov = "arm.ds.mov"(%x1) : (!arm.reg<x1>) -> !arm.reg<x2>
--- a/xdsl/dialects/arm/ops.py
+++ b/xdsl/dialects/arm/ops.py
@@ -1,4 +1,4 @@
-from abc import ABC
+from abc import ABC, abstractmethod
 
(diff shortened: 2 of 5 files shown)


pyproject.toml
superlopuh, line 1: can you please revert this change? I already merged it to main

tests/filecheck/dialects/arm/test_ops.mlir
```diff
@@ -8,5 +9,6 @@
 // CHECK: %ds_mov = arm.ds.mov %x1 : (!arm.reg<x1>) -> !arm.reg<x2>
 %ds_mov = arm.ds.mov %x1 : (!arm.reg<x1>) -> !arm.reg<x2>
 
+// CHECK-ASM: mov x2, x1
```
superlopuh, line 12: Let's group the assembly together with the custom format check?
```suggestion
// CHECK-ASM: mov x2, x1
%ds_mov = arm.ds.mov %x1 : (!arm.reg<x1>) -> !arm.reg<x2>

```

superlopuh (approved):

tests/filecheck/dialects/arm/test_ops.mlir
```diff
@@ -1,11 +1,13 @@
 // RUN: XDSL_ROUNDTRIP
 // RUN: XDSL_GENERIC_ROUNDTRIP
+// RUN: xdsl-opt -t arm-asm %s | filecheck %s --check-prefix=CHECK-ASM
 
 
 // CHECK: %x1 = arm.get_register : !arm.reg<x1>
 %x1 = arm.get_register : !arm.reg<x1>
 
 // CHECK: %ds_mov = arm.ds.mov %x1 : (!arm.reg<x1>) -> !arm.reg<x2>
+// CHECK-ASM: mov x2, x1
```
alexarice, line 10: Would it be possible to get a test that includes a comment

xdsl/dialects/arm/ops.py
```diff
@@ -18,14 +19,51 @@ class ARMOperation(IRDLOperation, ABC):
     Base class for operations that can be a part of ARM assembly printing.
     """
 
+    @abstractmethod
+    def assembly_line(self) -> str | None:
+        raise NotImplementedError()
+
+
+class ARMInstruction(ARMOperation, ABC):
```
alexarice, line 27: Will `ARMOperation`s that are not `ARMInstructions` ever print assembly? If not then I imagine that the `assembly_line` method can be removed from `ARMOperation`

superlopuh: Yep! The `get_register` operation participates in assembly printing by explicitly saying that there's nothing to print. There will be more like this.

alexarice: I'm not sure I understand, is explicitly saying there's nothing to print different to printing nothing? Another way to phrase the question above is:
Will there be any `ARMOperation` which is not an `ARMInstruction for which `assembly_line` does not return `None`

superlopuh: In RISC-V, examples of this are CommentOp, and RiscvFunc.FuncOp

PR #3488, opened 2024-11-20
dialects: (builtin) make FloatAttr a TypedAttribute

alexarice:
Pretty much as the title says.

### DIFF
--- a/xdsl/parser/base_parser.py
+++ b/xdsl/parser/base_parser.py
@@ -359,6 +359,36 @@ class BaseParser:
             "Expected integer literal" + context_msg,
         )
 
+    def parse_optional_float(
+        self,
+        *,
+        allow_negative: bool = False,
+    ) -> float | None:
+        """
+        Parse a (possibly negative) float, if present.
+        """
+        is_negative = False
+        if allow_negative:
+            is_negative = self._parse_optional_token(Token.Kind.MINUS) is not None
+
+        if (value := self._parse_optional_token(Token.Kind.FLOAT_LIT)) is not None:
+            value = value.get_float_value()
+            return -value if is_negative else value
+
+    def parse_float(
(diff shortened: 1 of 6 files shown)


alexarice:
I've refactor and rebased on top of #3490

xdsl/parser/base_parser.py
```diff
@@ -359,6 +359,36 @@ def parse_integer(
             "Expected integer literal" + context_msg,
         )
 
+    def parse_optional_float(
+        self,
+        *,
+        allow_negative: bool = False,
```
superlopuh, line 365: I can imagine this being quite confusing, let's drop the default argument for now?

alexarice: It was copied off the above function for integers I think, but sounds like a good idea

alexarice: It seems I copied poorly and the rest have a default value of True, would that be preferred or still no default?

superlopuh: I have a slight preference towards default True

xdsl/parser/base_parser.py
```diff
@@ -376,8 +406,7 @@ def parse_optional_number(
         ) is not None:
             return -value if is_negative else value
 
-        if (value := self._parse_optional_token(Token.Kind.FLOAT_LIT)) is not None:
-            value = value.get_float_value()
+        if (value := self.parse_optional_float()) is not None:
```
superlopuh, line 409: ?
```suggestion
        if (value := self.parse_optional_float(allow_negative=False)) is not None:
```

alexarice: ? indeed

math-fehr (approved):
Nice!

PR #3490, opened 2024-11-20
core: make TypedAttribute printing generic

alexarice:
This currently doesn't really do anything because `IntegerAttr` is the only `TypedAttribute`. But after future PRs this will unify the way in which `TypedAttribute` is printed.

There is a potential worry about circularity here if someone calls `print_attribute` in `print_without_type` but I think this would have caused problems anyway

### DIFF
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -535,7 +535,7 @@ class Printer:
                 )
                 and ty.width.data == 1
             ):
-                self.print_string("false" if attribute.value.data == 0 else "true")
+                self.print_string("true" if attribute.value.data else "false")
                 return
             # Otherwise we fall through to TypedAttribute case
 
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -68,6 +68,7 @@ from xdsl.ir import (
     SpacedOpaqueSyntaxAttribute,
     SSAValue,
     TypeAttribute,
+    TypedAttribute,
 )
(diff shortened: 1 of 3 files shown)


xdsl/printer.py
```diff
@@ -525,6 +526,25 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+                and ty.width.data == 1
+            ):
+                self.print_string("false" if attribute.value.data == 0 else "true")
+                return
+            # Otherwise we fall through to TypedAttribute case
+
+        if isinstance(attribute, TypedAttribute):
+            attribute.print_without_type(self)
+            self.print_string(" : ")
+            self.print_attribute(attribute.parameters[attribute.get_type_index()])
```
alexarice, line 545: This is the line which needs `get_type_index` to be a classmethod

xdsl/printer.py
```diff
@@ -525,6 +526,25 @@ def print_attribute(self, attribute: Attribute) -> None:
...
+        if isinstance(attribute, IntegerAttr):
+            # boolean shorthands
+            if (
+                isinstance(
+                    (ty := attribute.parameters[attribute.get_type_index()]),
+                    IntegerType,
+                )
+                and ty.width.data == 1
+            ):
+                self.print_string("false" if attribute.value.data == 0 else "true")
```
superlopuh, line 538:
```suggestion
                self.print_string("true" if attribute.value.data else "false")
```

superlopuh (approved):

PR #3492, opened 2024-11-20
dialects: (builtin) make DenseIntOrFPElementsAttr generic on element type

alexarice:
Started trying to make `DenseIntOrFPElementsAttr` a `TypedAttribute`.
`TypedAttribute`s force that the `type` parameter must appear as a type var generic. I am not sure why this is the case so it would be good to know what the original motivation for that was. To get this to go through I have disabled this check.

I then ended up adding a typevar for the elements to `DenseIntOrFPElementsAttr` which caused a lot of follow on changes and I am unsure if the result is nicer than what I started with. I think `DenseIntOrFPElementsAttr` can be made into a `TypedAttribute` without this type var.

I also imagine that extensive use of `isattr` is bad?

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -74,6 +74,7 @@ from xdsl.traits import (
     SymbolTable,
 )
 from xdsl.utils.exceptions import DiagnosticException, VerifyException
+from xdsl.utils.hints import isa
 from xdsl.utils.isattr import isattr
 
 if TYPE_CHECKING:
@@ -1715,11 +1716,11 @@ RankedStructure: TypeAlias = (
     VectorType[AttributeCovT] | TensorType[AttributeCovT] | MemRefType[AttributeCovT]
 )
 
+AnyDenseElement: TypeAlias = IntegerType | IndexType | AnyFloat
+
 
 @irdl_attr_definition
-class DenseIntOrFPElementsAttr(
-    ParametrizedAttribute, ContainerType[IntegerType | IndexType | AnyFloat]
-):
+class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
     name = "dense"
     type: ParameterDef[
         RankedStructure[IntegerType]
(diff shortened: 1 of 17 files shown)


alexarice:
It would be good to get any opinions on this. I'm tempted to see what the non-generic version would look like

xdsl/dialects/builtin.py
```diff
@@ -1704,17 +1705,20 @@ def get_element_type(self) -> _UnrankedMemrefTypeElems:
     VectorType[AttributeCovT] | TensorType[AttributeCovT] | MemRefType[AttributeCovT]
 )
 
+AnyDenseElement: TypeAlias = IntegerType | IndexType | AnyFloat
+DenseElementT = TypeVar("DenseElementT", bound=AnyDenseElement, covariant=True)
+_DenseElementT = TypeVar("_DenseElementT", bound=AnyDenseElement)
```
superlopuh, line 1710: Is this the equivalent of sized in MLIR? Would it make sense to add this independently of the other changes?

xdsl/dialects/builtin.py
```diff
@@ -1854,11 +1834,77 @@ def tensor_from_list(
...
+        if len(data) == 0:
+            pass
+        elif data.count(data[0]) == len(data):
+            DenseIntOrFPElementsAttr._print_one_elem(data[0], printer)
+        else:
+            DenseIntOrFPElementsAttr._print_dense_list(data, shape, printer)
+        printer.print_string(">")
+
+
+DenseIntElementsAttr: TypeAlias = DenseIntOrFPElementsAttr[IntegerType]
```
superlopuh, line 1907: This line feels like it could have its own PR, we could probably have a custom constraint to start with to minimise the diff

alexarice:
I think the first step is to make `DenseIntOrFPElementsAttr` a `TypedAttribute[AnyDenseElement]`, without giving it a generic parameter. I think the changes should then be localised to the parsing changes and the removal of the generic check on TypedAttributes

superlopuh:
Yep that also crossed my mind, not sure whether there is a single best first step but it would be nice to have a smaller diff for sure.

alexarice:
I'll split that out and see what it looks like

xdsl/dialects/builtin.py
alexarice: I doubt it is as `IndexType` is not sized right? (Also `IndexType` should be forbidden from `DenseIntOrFPElementsAttr` but that's another matter)

alexarice:
Reworked and stacked this PR. Still not 100% sure this is desirable

xdsl/dialects/builtin.py
superlopuh: I still feel like this could be its own PR :)

alexarice: You want a separate PR just for the type alias?

xdsl/dialects/builtin.py
superlopuh: I think it's quite relevant, and we should forbid it indeed, I thought we already made that change recently.

xdsl/dialects/builtin.py
superlopuh: Yeah that was the idea, it technically is its own API surface change and seems to be a large contribution to this PR's diff, independent of the other changes.

xdsl/dialects/builtin.py
superlopuh: Ah it was DenseArrayBase #3258

xdsl/dialects/builtin.py
superlopuh: Sorry that's what I mean, this change in the diff of this PR could already be done in main directly, and is not dependent on the changes you make here. Or am I misunderstanding?

alexarice: I don't see how to separate the changes. I can't change `DenseIntOrFPElementsAttr` to `DenseIntOrFPElementsAttr[AnyDenseElement]` before this PR.

I could potentially do two steps of changing every `DenseIntOrFPElementsAttr` to `DenseIntOrFPElementsAttr[AnyDenseElement]` and then do a second PR specialising some of these to `DenseIntElementsAttr` = `DenseIntOrFPElementsAttr[IntegerType]`, but I don't think the first diff will be any smaller.

alexarice: Ah I get it now, I could give that a go

xdsl/interpreters/builtin.py
```diff
@@ -85,7 +88,7 @@ def dense_int_or_fp_elements_value(
         attr: Attribute,
         type_attr: builtin.MemRefType[Any],
     ) -> ShapedArray[Any]:
-        assert isinstance(attr, builtin.DenseIntOrFPElementsAttr)
+        assert isattr(attr, base(builtin.DenseIntOrFPElementsAttr[AnyDenseElement]))
```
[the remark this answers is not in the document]
superlopuh: To expand on this thought, it's because `base` does the type parsing every time this is called, whereas Constr caches that, and is a tiny bit more resilient to API changes in the future

superlopuh:
Blocked by #3531

alexarice:
Sure

PR #3495, opened 2024-11-20
dialects: (bufferization) fix materialize_in_destination

### LINKED ISSUE #3489: dialects: (bufferization) MaterializeInDestination wrong operands
https://github.com/xdslproject/xdsl/blob/59e59ad334247498b59e14dc70a481a5957dba59/xdsl/dialects/bufferization.py#L191-L197

The [MaterializeInDestinationOp](https://mlir.llvm.org/docs/Dialects/BufferizationOps/#bufferizationmaterialize_in_destination-bufferizationmaterializeindestinationop) in MLIR allows the dest operand to be of any shaped type (e.g. Tensor or MemRef). It does not seem the case for xDSL, as the current implementation forces TensorType for all operands.

alexarice:
Fixes #3489 

In order to make `materialize_in_destination` compatible with mlir, it needs to be able to take a memref type as the destination. However, in mlir there must be no result if the destination is a memref and so the result here must be optional. This causes a further problem in giving the assembly format which I am yet to solve.

Next steps:
- [x] Add `operands` assembly format directive (#3507)
- [x] Add `results` assembly format directive (#3518)
- [x] Add `functional-type` assembly format directive (I believe this will need some way to tell how many things a directive will print) (#3517)
- [x] Update assembly format for `bufferize.materialize_in_destination`
- [x] Add memref case to mlir integration test

### DIFF
--- a/tests/filecheck/dialects/bufferization/bufferization_ops.mlir
+++ b/tests/filecheck/dialects/bufferization/bufferization_ops.mlir
@@ -1,16 +1,14 @@
 // RUN: XDSL_ROUNDTRIP
 // RUN: XDSL_GENERIC_ROUNDTRIP
 
-// CHECK:       builtin.module {
-
-// CHECK-NEXT:    %i0, %i1 = "test.op"() : () -> (index, index)
-// CHECK-NEXT:    %t0 = "test.op"() : () -> tensor<10x20x30xf64>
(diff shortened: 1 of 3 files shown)


alexarice:
@francescodaghero does this fix the issue?

PR #3500, opened 2024-11-21
API: add InferenceContext

superlopuh:
A wrapper class to encapsulate the inference context. I have a feeling that we might want to split out the inference context into range and non-range variable assignments, this will hide the change from the signature of the `infer` method.

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -61,6 +61,15 @@ ConstraintVariableType: TypeAlias = Attribute | Sequence[Attribute]
 Possible types that a constraint variable can have.
 """
 
+
+@dataclass
+class InferenceContext:
+    variables: dict[str, ConstraintVariableType] = field(default_factory=dict)
+    """
+    A mapping from variable names to the inferred attribute or attribute sequence.
+    """
+
+
 _T = TypeVar("_T")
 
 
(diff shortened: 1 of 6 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -885,7 +886,9 @@ def parse(self, parser: Parser, state: ParsingState) -> None:
         ):
             attr = unique_base.new(unique_base.parse_parameters(parser))
         elif issubclass(unique_base, Data):
-            attr = unique_base.new(unique_base.parse_parameter(parser))  # pyright: ignore[reportUnknownVariableType]
+            attr = unique_base.new(  # pyright: ignore[reportUnknownVariableType]
```
superlopuh, line 889: not sure why ruff formatted this

alexarice (commented):
I don't see how this achieves any encapsulation without making `InferenceContext.variables` private

superlopuh:
It doesn't at all, it just lets us communicate the purpose of the parameter, and potentially change its contents in the future with a smaller diff

xdsl/irdl/constraints.py
```diff
@@ -61,6 +61,15 @@ def update(self, other: ConstraintContext):
 Possible types that a constraint variable can have.
 """
 
+
+@dataclass
```
alexarice, line 65: This can probably be frozen

alexarice (approved):
I feel this is split more than it has to be, but it is the direction we want to head in.

superlopuh:
Marking a class as frozen but exposing mutating properties feels misleading, I'd love to make it actually frozen at some point soon with the help of some data structures that will preserve the mutable structure's speed.

PR #3501, opened 2024-11-21
core: refactor assembly format directives

alexarice:
Refactors the assembly format directives to:
- Remove some unnecessary classes that made the system rigid
- Separate directives that go in a `type` clause from format directives
This will make it easier to add `operands`, `results`, and `functional-type` directives.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-from typing import Literal
+from typing import Literal, TypeAlias
 
 from xdsl.dialects.builtin import UnitAttr
 from xdsl.ir import (
@@ -237,19 +237,11 @@ class FormatProgram:
 
 
 @dataclass(frozen=True)
-class FormatDirective(ABC):
-    """A format directive for operation format."""
-
-    @abstractmethod
-    def parse(self, parser: Parser, state: ParsingState) -> None: ...
-
-    @abstractmethod
-    def print(
-        self, printer: Printer, state: PrintingState, op: IRDLOperation
-    ) -> None: ...
+class Directive(ABC):
+    """An assembly format directive"""
 
 
-class AnchorableDirective(FormatDirective, ABC):
+class AnchorableDirective(Directive, ABC):
     """
     Base class for Directive usable as anchors to optional groups.
     """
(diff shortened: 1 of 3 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -271,79 +275,183 @@ class OptionallyParsableDirective(FormatDirective, ABC):
...
+    """The variable name. This is only used for error message reporting."""
+    index: int
+    """Index of the variable(operand or result) definition."""
 
 
-class VariadicTypeDirective(VariadicLikeTypeDirective, VariadicVariable, ABC):
-    pass
+class VariadicVariable(VariableDirective, AnchorableDirective, ABC):
+    def is_present(self, op: IRDLOperation) -> bool:
+        return len(getattr(op, self.name)) > 0
```
superlopuh, line 449:
```suggestion
        return bool(getattr(op, self.name))
```

superlopuh: but then again I wonder whether we should be going via the def instead?

alexarice: We have the index of the variable so should be able to do `op.operands[self.index]`. This should be more type safe too. Should I work that into this PR?

alexarice: You couldn't do the above without more refactoring as `VariadicVariable` doesn't know which type of variable it is

superlopuh: Ah too bad, although maybe that would also be worth fixing? I'm happy to punt this side quest until someone stumbles upon it and feels motivated

alexarice: FWIW this is moved code rather than new code (I don't really see this as an excuse though)

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -204,25 +203,19 @@ def verify_directives(self, elements: list[FormatDirective]):
...
-                            "A variadic operand variable cannot be followed by another variadic operand variable."
-                        )
-                    elif isinstance(a, RegionDirective) and isinstance(
-                        b, RegionDirective
-                    ):
-                        self.raise_error(
-                            "A variadic region variable cannot be followed by another variadic region variable."
-                        )
-                case AttrDictDirective(), RegionDirective() if not (a.with_keyword):
+                case VariadicOperandDirective(), VariadicOperandDirective():
```
superlopuh, line 206: seems off

superlopuh: ah no my bad the diff wasn't showing the input, looks good

superlopuh (approved):
LGTM!

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -204,25 +203,19 @@ def verify_directives(self, elements: list[FormatDirective]):
...
                     self.raise_error(
-                        "An `attr-dict' directive without keyword cannot be directly followed by a region variable as it is ambiguous."
+                        "A variadic operand variable cannot be followed by another variadic operand variable."
```
kimxworrall, line 208: Glad you've cleaned this up a bit.

kimxworrall (approved):

PR #3503, opened 2024-11-22
misc: use nullable getters in ConstraintContext

superlopuh:
Paring this PR down until it's clear that ScopedDicts are what we want to use for constraint contexts, reducing changes to just getters being nullable.

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -49,8 +49,8 @@ class ConstraintContext:
         scoped_context = ScopedConstraintContext(inner_variables, inner_range_variables)
         yield scoped_context
         if scoped_context.commit_scope:
-            self._variables = inner_variables
-            self._range_variables = inner_range_variables
+            self._variables.local_scope |= inner_variables.local_scope
+            self._range_variables.local_scope |= inner_range_variables.local_scope
 
 
 @dataclass
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -12,6 +12,7 @@ from typing_extensions import assert_never
 from xdsl.ir import Attribute, AttributeCovT, AttributeInvT, ParametrizedAttribute
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.runtime_final import is_runtime_final
+from xdsl.utils.scoped_dict import ScopedDict
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


xdsl/irdl/constraints.py
```diff
@@ -20,38 +21,35 @@ class ConstraintContext:
     Contains the assignment of constraint variables.
     """
 
-    _variables: dict[str, Attribute] = field(default_factory=dict)
+    _variables: ScopedDict[str, Attribute] = field(default_factory=lambda: ScopedDict())
```
alexarice, line 24:
```suggestion
    _variables: ScopedDict[str, Attribute] = field(default_factory=ScopedDict)
```
Does this work?

xdsl/irdl/constraints.py
```diff
@@ -20,38 +21,35 @@ class ConstraintContext:
...
-    @property
-    def variables(self) -> Sequence[str]:
-        return tuple(self._variables.keys())
+    def push_variable_scope(self, name: str | None = None):
+        self._variables = ScopedDict(self._variables, name=name)
+        self._range_variables = ScopedDict(self._range_variables, name=name)
 
-    @property
-    def range_variables(self) -> Sequence[str]:
-        return tuple(self._range_variables.keys())
```
alexarice, line 47: Are these just dead code? Great that we're removing them if so

xdsl/irdl/constraints.py
```diff
@@ -380,15 +379,14 @@ def verify(
     ) -> None:
         constraint_context = constraint_context or ConstraintContext()
         for attr_constr in self.attr_constrs:
-            # Copy the constraint to ensure that if the constraint fails, the
-            # constraint context is not modified.
-            constraint_context_copy = constraint_context.copy()
+            # Push a new scope that will be discarded if the constraint fails to verify,
+            # and kept if verification succeeds
+            constraint_context.push_variable_scope()
```
alexarice, line 384: Could we make this a contextmanager?

xdsl/irdl/constraints.py
```diff
@@ -20,38 +21,35 @@ class ConstraintContext:
     Contains the assignment of constraint variables.
     """
 
-    _variables: dict[str, Attribute] = field(default_factory=dict)
+    _variables: ScopedDict[str, Attribute] = field(default_factory=lambda: ScopedDict())
     """The assignment of constraint variables."""
 
-    _range_variables: dict[str, tuple[Attribute, ...]] = field(default_factory=dict)
+    _range_variables: ScopedDict[str, tuple[Attribute, ...]] = field(
+        default_factory=lambda: ScopedDict()
```
alexarice, line 28:
```suggestion
        default_factory=ScopedDict
```

xdsl/irdl/constraints.py
superlopuh: Unfortunately Pyright isn't happy with this, even though it would work, I have a feeling they have some special handling of the builtin data structures

xdsl/irdl/constraints.py
superlopuh: they are after my change, it feels quite silly that we were constructing and throwing away the tuples so many times. I am very much looking forward to ASV being set up properly to track the impact of these sorts of changes

xdsl/irdl/constraints.py
superlopuh: I tried that, and the scope has to stay once the thing succeeds, so it's not as symmetric as the context manager needs

xdsl/irdl/constraints.py
superlopuh: similarly here, Pyright doesn't like it :(

xdsl/irdl/constraints.py
alexarice: Unfortunate

xdsl/irdl/constraints.py
alexarice: I'd forgotten about that point. In that case this seems reasonable. The only other thing worth considering is that if an AnyOf succeeds then the scope could be "flattened" onto the one below. I imagine this would almost never improve performance though.

superlopuh: I think eventually we might want to implement a more advanced data structure with checkpointing that can do this more efficiently by default, but I don't think there's a rush

alexarice (approved):

math-fehr (approved):
Uncontroversial

alexarice (approved):

PR #3504, opened 2024-11-22
core: remove generic and generic check from TypedAttribute

alexarice:
As discussed in #3492, this check feels unnecessary and is unwanted for making `DenseIntOrFPElementsAttr` a `TypedAttribute`

### DIFF
--- a/xdsl/irdl/attributes.py
+++ b/xdsl/irdl/attributes.py
@@ -24,7 +24,6 @@ from typing import (
 
 from xdsl.ir import (
     Attribute,
-    AttributeCovT,
     AttributeInvT,
     Data,
     ParametrizedAttribute,
@@ -34,7 +33,6 @@ from xdsl.utils.exceptions import PyRDLAttrDefinitionError, VerifyException
 from xdsl.utils.hints import (
     PropertyType,
     get_type_var_from_generic_class,
-    get_type_var_mapping,
 )
 from xdsl.utils.runtime_final import runtime_final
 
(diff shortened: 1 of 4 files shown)


superlopuh (commented):
Do you know why it was added in the first place? I'm wondering why it wasn't just using OO, is it for precise placement of the `type` variable in the list of parameters? It feels like that should be possible by redeclaring the type property at the desired index. Do you feel like that would be worth a go? It would also decouple the dense stuff, I have a feeling.

alexarice:
I'm not really sure why it was there, and @math-fehr wasn't sure either. I kind of feel that enforcing that the type appear as a parameter at all rather than just as an interface method on `TypedAttribute` is overkill, but I guess it was there for irdl things. 

I'm actually really struggling to get all the variances correct on dense PR atm so any suggestions on sorting this stuff would be good.

superlopuh:
Would you like me to have a go at my own suggestion?

alexarice:
Go for it. I'm starting to wonder if the generic parameter on `TypedAttribute` can be removed completely

superlopuh:
My attempt was a failure, let's go back to this idea. Is there any way to keep the unit test deleted here? So we keep the fact that the type is indeed the same as the `TypedAttribute` parameter, but check it differently?

alexarice:
> Is there any way to keep the unit test deleted here?

 What do you mean?

superlopuh:
Like to make the check work differently, instead of looking for the parameter of the `Generic`, look for the parameter of `TypedAttribute1?

alexarice:
I would prefer just to remove the generic parameter on `TypedAttribute`

xdsl/irdl/attributes.py
```diff
@@ -233,8 +212,14 @@ def irdl_param_attr_definition(cls: _PAttrTT) -> _PAttrTT:
...
-        parameter_names: tuple[str] = tuple(zip(*attr_def.parameters))[0]
-        type_index = parameter_names.index("type")
+        type_indexes = tuple(
+            i for i, (p, _) in enumerate(attr_def.parameters) if p == "type"
+        )
+        if not type_indexes:
+            raise PyRDLAttrDefinitionError(
+                f"TypedAttribute {cls.__name__} should have a 'type' parameter."
+            )
+        type_index = type_indexes[0]
```
math-fehr, line 222: Should this logic be put as a function in `TypedAttribute` instead?
By having an accessor to `type` directly with something like `get_type` for instance?

math-fehr (approved):
I just had a comment but otherwise nice!

xdsl/irdl/attributes.py
alexarice: Are you suggesting to remove `get_type_index` entirely?

alexarice: The issue is that I think you want this check to happen at the definition of the attribute, rather than some verify check on an instance of that attribute

math-fehr: actually nevermind I'm an idiot, this is all good!

alexarice:
@superlopuh you happy with this (and the next PR)?

superlopuh:
Oh this is a case where I think I'm only happy with this together with the next PR, as after this PR we can declare something as typed with a certain type, but remove the code that verifies this at runtime. If TypedAttribute is no longer generic then this PR is dead code elimination.

alexarice:
Should I merge the 2 PRs?

superlopuh:
I would recommend mergint them in one go, yep

alexarice:
Great, I'm just merging main into the second one to update `FloatAttr`

superlopuh (approved):

PR #3505, opened 2024-11-22
core: make TypedAttribute not generic

alexarice:
The bit that used the generic wasn't type-sound anyway. 
Stacked PR + need to rebase past making float a typed attribute


math-fehr (approved):
Yoink

PR #3507, opened 2024-11-22
core: Add operands directive

alexarice:
~Started writing an operands directive but the parse errors appear in the wrong place so might refactor everything again.~ I refactored everything again (#3516)

Also needs more tests.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-from typing import Literal
+from typing import Literal, TypeVar
 
 from xdsl.dialects.builtin import UnitAttr
 from xdsl.ir import (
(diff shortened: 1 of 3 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -292,10 +292,11 @@ class VariadicLikeFormatDirective(
     Such directives can not be followed by comma literals.
     """
 
-    def set_empty(self, state: ParsingState):
+    def set_empty(self, state: ParsingState) -> str | None:
         """
         Set the appropriate field of the parsing state to be empty.
         Used when a variable appears in an optional group which is not parsed.
+        Returns a string on error.
```
superlopuh, line 299: why not raise error on error?

alexarice: The errors are handled by the `Parser` object. This is one of the reasons I think this needs a bit more work.

alexarice: I think the `set_type`/`get_type` interface may be going a bit too far, as it makes it difficult to throw an error in the correct place. The original reason I wanted it was so that the `functional_type` directive knows how many types need to be printed so that it knows whether to put brackets round.

Perhaps keeping `get_type` but moving the parsing into each `TypeableDirective` is the correct interface.

alexarice:
Changed based on top of a new slight refactor (basically did what I mentioned in the comment above). Still a draft as this needs more tests but the actual functionality should be ready to review

kimxworrall (approved):
Looks good!

alexarice:
I've added many tests and think this is ready

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -600,6 +600,98 @@ def set_types_empty(self, state: ParsingState) -> None:
...
+class OperandsDirective(VariadicOperandDirective, VariadicTypeableDirective):
+    """
+    An operands directive, with the following format:
+      operands-directive ::= operands
+    Prints each operand of the operation, inserting a comma between each.
+    The variadic_index represents the position of a (single) variadic variable,
+    with the boolean representing whether it is optional
+    """
+
+    variadic_index: tuple[bool, int] | None
```
superlopuh, line 616:
```suggestion
    """

    variadic_index: tuple[bool, int] | None
    """
    Represents the position of a (single) variadic variable, with the boolean
    representing whether it is optional
    """
```

superlopuh (approved):

xdsl/irdl/declarative_assembly_format.py
superlopuh: I still think that errors are probably the way to go here, let's have a go at this after this stack of PRs is in?

alexarice: This diff is outdated. It now throws errors instead of returning a string. The preceding PR to this implements the suggestion I made in the comment above.

PR #3515, opened 2024-11-25
dialects (arm): add mul instruction with 2 source registers

emmau678:
Add mul instruction (https://developer.arm.com/documentation/ddi0597/2024-06/Base-Instructions/MUL--MULS--Multiply-?lang=en).

The mul instruction supports either 2 or 3 operands (in the case of 2, the destination is also treated as the 2nd source). This PR only handles the case where 3 operands are provided, 2 source registers and 1 destination.

### DIFF
--- a/xdsl/dialects/arm/ops.py
+++ b/xdsl/dialects/arm/ops.py
@@ -114,3 +114,43 @@ class GetRegisterOp(ARMOperation):
 
     def assembly_line(self):
         return None
+
+
+@irdl_op_definition
+class DSSMulOp(ARMInstruction):
+    """
+    Multiplies the values in s1 and s2 and stores the result in d.
+
+    https://developer.arm.com/documentation/ddi0597/2024-06/Base-Instructions/MUL--MULS--Multiply-?lang=en
+    """
+
+    name = "arm.dss.mul"
+
+    d = result_def(IntRegisterType)
+    s1 = operand_def(IntRegisterType)
+    s2 = operand_def(IntRegisterType)
+    assembly_format = (
+        "$s1 `,` $s2 attr-dict `:` `(` type($s1) `,` type($s2) `)` `->` type($d)"
+    )
+
+    def __init__(
+        self,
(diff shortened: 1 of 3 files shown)


alexarice (approved):
Looks good.

What's the plan for the 2 operand case? I believe the work I'm doing to fix the bufferization operation would allow you to have an optional result here, though that might mess with your naming scheme.

emmau678:
> Looks good.
> 
> What's the plan for the 2 operand case? I believe the work I'm doing to fix the bufferization operation would allow you to have an optional result here, though that might mess with your naming scheme.

thanks!
I was originally just going to write a separate op for it (i.e. DSMulOp) but maybe what you've suggested would be better. I guess if I do that I'll just remove the DSS part of the name. perhaps @superlopuh would have some thoughts on it

alexarice:
Either would probably work. Which would be better in the long run is hard to tell and probably depends on how you plan to use these operations.

xdsl/dialects/arm/ops.py
```diff
@@ -114,3 +114,43 @@ def __init__(self, register_type: IntRegisterType):
...
+    d = result_def(IntRegisterType)
+    s1 = operand_def(IntRegisterType)
+    s2 = operand_def(IntRegisterType)
+    assembly_format = (
+        "$s1 `,` $s2 attr-dict `:` `(` type($s1) `,` type($s2) `)` `->` type($d)"
+    )
+
+    def __init__(
+        self,
+        d: IntRegisterType,
```
superlopuh, line 138: This is where the named argument is better than positional, as it would make sense for me to make this the unallocated type by default, as in the RISC-V equivalent.

superlopuh (approved):

superlopuh:
I would expect that the two operand case is some sort of assembly shorthand, we could add some custom logic to print it that way when the second argument happens to be the same register as the result, but I don't think we want that reflected in the IR.

PR #3516, opened 2024-11-25
core: Move assembly format type parsing to directives

alexarice:
Went slightly too far on the previous refactor. I think this level of abstraction works well for allowing parse errors to be put in the correct place. PRs for operands, results, and functional-type directives to follow based on this PR.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-from typing import Literal, TypeAlias
+from typing import Literal
 
 from xdsl.dialects.builtin import UnitAttr
 from xdsl.ir import (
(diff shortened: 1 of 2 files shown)


superlopuh (approved):

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -306,39 +306,22 @@ class TypeableDirective(Directive, ABC):
...
-
-
-class VariadicTypeableDirective(AnchorableDirective, ABC):
-    """
-    Directives which can set or get multiple types.
-    """
-
-    @abstractmethod
-    def set_types(self, types: Sequence[Attribute], state: ParsingState) -> None: ...
+    def parse_single_type(self, parser: Parser, state: ParsingState) -> None: ...
```
alexarice, line 309: I'm not sure on the names `parse_single_type` and `parse_many_types` but couldn't think of anything better. I didn't want to go with `parse_type` and `parse_types` as the signatures are quite similar and this could cause confusion.

superlopuh: They're not fantastically beautiful but they are quite clear, which is the most important thing. One thing to consider would be `parse_type`/`parse_type_list`

alexarice: I don't think I feel strongly either way. Happy to change it if you'd like me to.

PR #3518, opened 2024-11-25
core: Add results directive

alexarice:
Adds a results directive to the assembly format. Very similar to the operands directive (#3507) but is not a `FormatDirective`, so must be used as part of a `type` or `functional-type` directive. I wonder if it's worth making a shared subclass for this and the operands directive to deduplicate some code. If so I can rebase on top of that PR.

Stacked on #3516. Draft as this needs more tests


alexarice:
Rebased on top of operands-directive PR, made a shared superclass for the shared functionality, and added a tonne of tests.

superlopuh (approved):

alexarice:
@superlopuh were you happy with the preceding PRs this sits on top of?

PR #3520, opened 2024-11-26
bug: (lower-csl-stencil) Zero-out accumulator for full reduction access

n-io:
This resulted in `nan`s after a few iterations.


dk949 (approved):
Looks great

emmau678 (approved):
looks good to me too

PR #3521, opened 2024-11-26
dialects (arm): add return instruction (bx lr)

emmau678:
Add a return instruction which is equivalent to ```bx lr```


alexarice (approved):

xdsl/dialects/arm/ops.py
```diff
@@ -154,3 +156,38 @@ def __init__(
...
+            attributes={
+                "comment": comment,
+            },
+        )
+
+    def assembly_line_args(self):
+        return ()
+
+    def assembly_instruction_name(self) -> str:
+        return "bx lr"
```
superlopuh, line 193: this should go in a new `arm_func` dialect, not `arm`, as the name of the op does not match the name of the assembly operation, and neither does the semantics

xdsl/dialects/arm_func/__init__.py
superlopuh, line 1: I would recommend just having one file for this, like for riscv

emmau678: I'm not sure I understand the reason for this - I thought riscv and riscv_func were initialised separately in their individual files (riscv.py and riscv_func.py). I'm not sure why we would merge __init__ files for arm and arm_func if they're separate dialects and then which of them should have the __init__?

superlopuh: I'm not sure I understand your comment, arm and arm_func are separate dialects, why merge init files for different dialects? I'm suggesting to just have a `arm_func.py` file in the dialects directory, like for other dialects. ARM and x86 are exceptions here because they're so huge, and I would like to refactor riscv in the future to also be split for the same reason.

emmau678: ah ok. I thought you meant to have one ```__init__``` file. I have updated it now to have a single ```arm_func.py``` file. thanks!

superlopuh (approved):
This looks great!

PR #3522, opened 2024-11-26
API: add `Op` suffix to all operations in xDSL

superlopuh:
I recently received feedback that one of the confusing aspects of xDSL is that it's not consistent in terms of operation class naming, unlike MLIR. This PR changes this. To make these changes, I added a runtime check to Operation's post_init, asserted that it ended in "Op", and changed the class names to make everything pass, then reverted that change, so this PR converts all the classes that are initialized at some point in our tests to this naming convention. We might want to add that check in the future, but I wanted to discuss this in a separate PR.

This was the check I used:
``` python
        assert (
            "cmath" in self.name or self.__class__.__name__[-2:] == "Op"
        ), self.__class__.__name__
```


alexarice:
Are we avoiding a pytest test that loads all dialects and checks the name of each operation in said dialect purely for performance reasons?

math-fehr:
> Are we avoiding a pytest test that loads all dialects and checks the name of each operation in said dialect purely for performance reasons?

This is actually a really nice idea, I think we should do that!

math-fehr (approved):
I definitely didn't read the changes, so I'll trust you on this.
I think we should talk on how to test it on another PR

alexarice:
Am I missing something? Doesn't the following work (possibly with an exclusion for `cmath` needed)
```
def test_op_names():
    dialects = get_all_dialects()
    for dialect_factory in dialects.values():
        dialect = dialect_factory()
        for op in dialect.operations:
            assert op.__name__[-2:] == "Op", f"{op.__name__} should end with 'Op'"
```

superlopuh:
It works, it just takes 1.5 seconds, I'd rather do this in another PR if that's ok with you, so we can discuss the merits of such a check there?

alexarice (approved):

alexarice (commented):
Just tried the test myself, takes 0.41 seconds locally and also found a lot of missing renamings in the experimental folder. I can push these if you like

alexarice:
Happy to discuss testing this separately in any case

superlopuh:
That's faster than I expected! I'll open a PR with this quickly now.

PR #3525, opened 2024-11-27
core: Fix extractor logic for optional and default-valued properties

alexarice:
The extractor used in the assembly parser assumed that all properties/attributes were present. We should instead only propagate the extractors from the property constraint if the property is non-optional or there is a default value. The default_value also needs to be given to the extractor for when the property is not present.

Adds 2 tests to demonstrate this.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format_parser.py
+++ b/xdsl/irdl/declarative_assembly_format_parser.py
@@ -259,12 +259,14 @@ class FormatParser(BaseParser):
         name: str
         is_prop: bool
         inner: VarExtractor[Attribute]
+        default_value: Attribute | None
 
         def extract_var(self, a: ParsingState) -> ConstraintVariableType:
             if self.is_prop:
-                attr = a.properties[self.name]
+                attr = a.properties.get(self.name, self.default_value)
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -259,12 +259,14 @@ class _AttrExtractor(VarExtractor[ParsingState]):
         name: str
         is_prop: bool
         inner: VarExtractor[Attribute]
+        default_value: Attribute | None
```
superlopuh, line 262: what are cases where this is `None`?

superlopuh: Worth a comment?

alexarice: Sure

alexarice: Wrote a comment, thought about it for 10 minutes, deleted it and wrote a different comment. If it looks OK I'll merge

math-fehr (approved):
Nice!

Issue #3527, opened 2024-11-27
misc: generate class names that satisfy our coding convention in irdl.mlir file loading


superlopuh:
Currently, the cmath dialect, defined as an irdl file, is imported into Python with the class names matching the operation names. We should make using irdl files as smooth a process as possible, and matching coding conventions is an important aspect of this. So we should import `cmath.norm` not as `class norm(IRDLOperation)`, but `class NormOp(IRDLOperation)`.

PR #3528, opened 2024-11-27
tool: (irdl-to-pyrdl) Generate better class name

alexarice:
Makes operation class names generated by `irdl_to_pyrdl` be in CamelCase and end with Op.

Unfortunately it seems this is not the tool that deals with `cmath.irdl` :facepalm: 

Related to #3527


xdsl/dialects/irdl/irdl_to_pyrdl.py
```diff
@@ -33,11 +33,15 @@ class {op.sym_name.data}(ParametrizedAttribute{type_addition}):
     return res
 
 
+def get_op_name(op: OperationOp) -> str:
```
superlopuh, line 36:
```suggestion
def class_name_from_op_name(op: OperationOp) -> str:
```

superlopuh: I would recommend adding a separate unit test for this, and to handle the case `dialect.part1.part2_bla`

alexarice: I want to move it somewhere more generic to reuse in the next PR, is `dialects/irdl/utils.py` a decent place?

alexarice: Can a symbol name have a `.` in it?

alexarice: Seems so

superlopuh (approved):

math-fehr (approved):
Cool!

alexarice:
Added suggestion from #3529 (sorry this got so confusing)

superlopuh (approved):

PR #3529, opened 2024-11-27
interpreters: (irdl) generate better op name

### LINKED ISSUE #3527: misc: generate class names that satisfy our coding convention in irdl.mlir file loading

alexarice:
Fixes #3527
Stacked but can be unstacked

### DIFF
--- a/tests/dialects/test_dialects.py
+++ b/tests/dialects/test_dialects.py
@@ -11,7 +11,6 @@ def test_op_class_names():
         (op.name, op.__name__)
         for dialect_factory in all_dialects.values()
         for op in dialect_factory().operations
-        if "cmath" not in op.name
         if op.__name__[-2:] != "Op"
     )
 
--- a/tests/filecheck/dialects/irdl/cmath_irdl_stub.py
+++ b/tests/filecheck/dialects/irdl/cmath_irdl_stub.py
@@ -52,12 +52,12 @@ if __name__ == "__main__":
 # CHECK-NEXT:      p0 : "Float32Type | Float64Type"
 # CHECK-EMPTY:
 # CHECK-EMPTY:
-# CHECK-NEXT:  class norm(IRDLOperation):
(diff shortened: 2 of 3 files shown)


superlopuh:
Oh and let's fix the test also to remove the cmath check

alexarice:
> Let's unstack

Might be easier just to deal with one PR at a time on second thoughts

superlopuh:
Your call

math-fehr (approved):
Nice!
Actually I'm wondering if we should not move `get_op_name` to  `OperationOp` directly?
Just to make it more easily discoverable, since it is only a few lines of code anyway

superlopuh:
I like that a lot, although in that case I'd call it something like `get_py_class_name`

superlopuh:
It definitely seems easier to unstack at this point, as the location and name of the op may need to be changed for both PRs...

alexarice:
I'm not sure I understand unstacking. To me it makes sense to fix up the other PR, get it merged, then rebase/remake this one

alexarice:
By unstack did you mean make the 2 PRs into 1 PR?

superlopuh:
Yep sorry I thought that's what you meant.

alexarice:
For some reason I thought you meant to make 2 completely separate PRs which both add the new function separately, sorry again for the confusion

alexarice:
The diff here isn't correct, I'll sort it out

superlopuh:
That's my bad for assuming what you meant!

alexarice:
Diff looks better now

superlopuh (approved):

PR #3530, opened 2024-11-27
core: remove unnecessary whitespace printing in assembly format

alexarice:
Variadic arguments were printing whitespace even if they had nothing to print after the whitespace, resulting in extra unnecessary whitespace appearing. This change moves the whitespace printing after the check to see if there's anything to print.


superlopuh (approved):

kimxworrall (approved):

math-fehr (approved):
very nice!

PR #3531, opened 2024-11-27
[WIP] change dense attr representation to bytes

jorendumoulin:


xdsl/dialects/builtin.py
```diff
@@ -1725,132 +1726,61 @@ def get_element_type(self) -> _UnrankedMemrefTypeElems:
...
+    @staticmethod
+    def get_format_str(
+        type: RankedStructure[IntegerType | IndexType | AnyFloat],
+    ) -> str:
+        data_len = prod(type.get_shape())
+        element_type = type.element_type
+
+        if isinstance(element_type, IndexType):
+            # store IndexType as long long in bytes object
+            format_str = "{}q"
```
superlopuh, line 1741: I think we might want to remove support for Index before this PR, as the size of index is a function of the target, not the host computer.

xdsl/dialects/builtin.py
```diff
@@ -1725,132 +1726,61 @@ def get_element_type(self) -> _UnrankedMemrefTypeElems:
 @irdl_attr_definition
 class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
     name = "dense"
-    type: ParameterDef[
-        RankedStructure[IntegerType]
-        | RankedStructure[IndexType]
-        | RankedStructure[AnyFloat]
-    ]
-    data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
+    type: ParameterDef[RankedStructure[IntegerType | IndexType | AnyFloat]]
```
superlopuh, line 1729: this seems to be an independent change?

xdsl/dialects/builtin.py
```diff
@@ -1725,132 +1726,61 @@ def get_element_type(self) -> _UnrankedMemrefTypeElems:
...
+        data_len = prod(type.get_shape())
+        element_type = type.element_type
+
+        if isinstance(element_type, IndexType):
+            # store IndexType as long long in bytes object
+            format_str = "{}q"
+        elif isinstance(element_type, IntegerType):
+            if element_type.size == 1:
+                # 1 byte signed char
+                format_str = "{}b"
```
superlopuh, line 1745: it would be great to share code with the `ptr.py` file, if possible, potentially moving shared utilities to a third location.

xdsl/dialects/builtin.py
```diff
@@ -1725,132 +1726,61 @@ def get_element_type(self) -> _UnrankedMemrefTypeElems:
...
 class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
     name = "dense"
-    type: ParameterDef[
-        RankedStructure[IntegerType]
-        | RankedStructure[IndexType]
-        | RankedStructure[AnyFloat]
-    ]
-    data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
+    type: ParameterDef[RankedStructure[IntegerType | IndexType | AnyFloat]]
+    data: ParameterDef[BytesAttr]
```
superlopuh, line 1730: I would recommend first migrating all users of this property to another property/getter that returns a sequence of AnyIntegerAttr or a sequence of AnyFloatAttr, to minimise the diff in other files (probably also a property/getter that returns a sequence of ints or a sequence of floats)

xdsl/dialects/builtin.py
```diff
@@ -1725,132 +1726,61 @@ def get_element_type(self) -> _UnrankedMemrefTypeElems:
 @irdl_attr_definition
 class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
     name = "dense"
-    type: ParameterDef[
-        RankedStructure[IntegerType]
-        | RankedStructure[IndexType]
-        | RankedStructure[AnyFloat]
-    ]
-    data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
```
superlopuh, line 1733: Wouldn't it make sense to migrate ArrayAttr to be bytes-based instead of DenseIntOrFPElementsAttr, BTW?

jorendumoulin:
thank you for the comments! I'll first do some more incremental PRs before making the switch to `bytes`!

xdsl/dialects/builtin.py
superlopuh: I'm not 100% sure of this BTW, but would be good to play with it and maybe chat about findings. It would be nice to have some sort of generic way to have a dense array of things, but maybe that should just be this class, and array should be left alone to explicitly be a tuple of attributes and nothing more.

PR #3533, opened 2024-11-28
dialects: (builtin): remove IndexType support for DenseIntOrFPElementsAttr

jorendumoulin:
This removes support for creating dense attributes with `IndexType` values.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1719,17 +1719,13 @@ RankedStructure: TypeAlias = (
     VectorType[AttributeCovT] | TensorType[AttributeCovT] | MemRefType[AttributeCovT]
 )
 
-AnyDenseElement: TypeAlias = IntegerType | IndexType | AnyFloat
+AnyDenseElement: TypeAlias = IntegerType | AnyFloat
 
 
 @irdl_attr_definition
 class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
     name = "dense"
-    type: ParameterDef[
-        RankedStructure[IntegerType]
-        | RankedStructure[IndexType]
-        | RankedStructure[AnyFloat]
-    ]
+    type: ParameterDef[RankedStructure[IntegerType] | RankedStructure[AnyFloat]]
     data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
 
     # The type stores the shape data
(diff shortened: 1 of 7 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1801,61 +1783,28 @@ def create_dense_float(
...
-    def from_list(
-        type: (
-            RankedStructure[AnyFloat | IntegerType | IndexType]
-            | RankedStructure[AnyFloat]
-            | RankedStructure[IntegerType]
-            | RankedStructure[IndexType]
-        ),
-        data: Sequence[int | float] | Sequence[AnyFloatAttr],
-    ) -> DenseIntOrFPElementsAttr: ...
-
```
jorendumoulin, line 1831: I think these overloads achieve nothing?

superlopuh: It's very possible that they've drifted

xdsl/parser/attribute_parser.py
```diff
@@ -864,6 +858,7 @@ def parse_dense_int_or_fp_elements_attr(
                 assert len(data_values) == 1, "Fatal error in parser"
                 data_values *= type_num_values
 
+        assert type
```
superlopuh, line 861: would be good to remove before merging

superlopuh:
We should just replace the failing tests' types to not use `index`, which would also fail in MLIR IIRC

superlopuh (approved):

superlopuh:
Joren, I'm sorry, I just played locally with it and it looks like MLIR does support index here... I'm not sure how it actually works, though, because the size of index is not known in general. It would be good to replicate the behaviour of MLIR, they probably do what you initially suggested and just use a long long in until it's lowered to a concrete bitwidth.

jorendumoulin:
Hold up!

I don't know why I hadn't checked this before, but this seems to be perfectly valid in MLIR:

```
module {
  memref.global "public" @g : memref<5xindex> = dense<[0, 1, 2, 3, 4]>
}
```

I'm not sure anymore we want to remove support for this?

jorendumoulin:
> Joren, I'm sorry, I just played locally with it and it looks like MLIR does support index here... I'm not sure how it actually works, though, because the size of index is not known in general. It would be good to replicate the behaviour of MLIR, they probably do what you initially suggested and just use a long long in until it's lowered to a concrete bitwidth.

Yes I just noticed as well! So probably for this PR just close it entirely?

math-fehr:
A bit of information, `index` is always less than 64 bits, so they probably store values on 64 bits then trunc them later on.

PR #3534, opened 2024-11-28
dialects: (builtin) shape for DenseIntORFPElementsAttr is always known

jorendumoulin:
As they only work for ranked structures, the shape will always be not None


math-fehr (approved):
Nice!

PR #3535, opened 2024-11-28
dialects: (builtin) DenseIntOrFPElementsAttr: make data private and use getters instead

jorendumoulin:
To prepare for the change of internal data representation, this PR makes the actual data of the attribute a private attribute. Everthing is accessed with the getters `get_attrs` (IntegerAttr, ...) and `get_values` (int, float).

The check whether the dense is set by a scalar is pretty common, so i included an additional helper function for that as well

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1888,7 +1888,7 @@ class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
         """
         return self._data.data
 
-    def is_scalar(self) -> bool:
+    def is_splat(self) -> bool:
         """
         Return whethere or not this dense attribute is defined entirely
         by a single value (splat).
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1730,7 +1730,7 @@ class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
         | RankedStructure[IndexType]
         | RankedStructure[AnyFloat]
     ]
-    _data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
+    data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
 
     # The type stores the shape data
     def get_shape(self) -> tuple[int, ...] | None:
@@ -1755,7 +1755,7 @@ class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
             n *= dim
 
         # Product of dimensions needs to equal length
-        return n == len(self._data.data)
+        return n == len(self.data.data)
 
     @staticmethod
     def create_dense_index(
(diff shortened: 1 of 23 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1875,6 +1875,26 @@ def tensor_from_list(
...
+        return tuple(el.value.data for el in self._data.data)
+
+    def get_attrs(self) -> Sequence[AnyIntegerAttr] | Sequence[AnyFloatAttr]:
+        """
+        Return all elements of the dense attribute in their relevant
+        attribute representation (IntegerAttr / FloatAttr)
+        """
+        return self._data.data
+
+    def is_scalar(self) -> bool:
```
math-fehr, line 1891: Can we rename this to `is_splat`?
`is_scalar` would mean that there is a single element in the tensor, where here we want to say that all elements are the same.

math-fehr (approved):
Seems sensible to me, I would just rename `is_scalar` to `is_splat` though

xdsl/dialects/builtin.py
jorendumoulin: Okay, thanks!

xdsl/transforms/csl_stencil_bufferize.py
```diff
@@ -370,9 +375,11 @@ def match_and_rewrite(self, op: arith.ConstantOp, rewriter: PatternRewriter, /):
             return
         assert isinstance(op.value, DenseIntOrFPElementsAttr)
         assert isa(op.value.type, TensorType[Attribute])
-        typ = DenseIntOrFPElementsAttr(
-            [tensor_to_memref_type(op.value.type), op.value.data]
+        memref_type = cast(
+            RankedStructure[AnyFloat | IntegerType | IndexType],
+            tensor_to_memref_type(op.value.type),
         )
+        typ = DenseIntOrFPElementsAttr.from_list(memref_type, op.value.get_values())
```
superlopuh, line 382: why this change? it should still work, as-is, no?

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -425,7 +428,15 @@ def match_and_rewrite(self, op: ConstantOp, rewriter: PatternRewriter, /):
...
-                        ConstantOp(DenseIntOrFPElementsAttr([typ, op.value.data]))
+                        ConstantOp(
+                            DenseIntOrFPElementsAttr.from_list(
+                                cast(
+                                    RankedStructure[AnyFloat | IntegerType | IndexType],
+                                    typ,
+                                ),
+                                op.value.get_values(),
+                            )
+                        )
```
superlopuh, line 439: same here, let's preserve this behaviour in this PR

xdsl/transforms/linalg_to_csl.py
```diff
@@ -45,9 +45,9 @@ def get_scalar_const(op: SSAValue) -> AnyFloatAttr | AnyIntegerAttr | None:
         isinstance(op, OpResult)
         and isinstance(op.op, arith.ConstantOp)
         and isa(val := op.op.value, DenseIntOrFPElementsAttr)
-        and val.data.data.count(val.data.data[0]) == len(val.data.data)
+        and val.is_splat()
     ):
-        return val.data.data[0]
+        return val.get_attrs()[0]
```
superlopuh, line 50: very nice

xdsl/dialects/builtin.py
```diff
@@ -1730,7 +1730,7 @@ class DenseIntOrFPElementsAttr(TypedAttribute, ContainerType[AnyDenseElement]):
         | RankedStructure[IndexType]
         | RankedStructure[AnyFloat]
     ]
-    data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
+    _data: ParameterDef[ArrayAttr[AnyIntegerAttr] | ArrayAttr[AnyFloatAttr]]
```
superlopuh, line 1733: let's unprivate it now that all the uses are migrated?

xdsl/transforms/csl_stencil_bufferize.py
jorendumoulin: yes, but then you have to access the private _data.
I can unprivatize data and make exceptions for these more trivial cases

superlopuh (approved):

PR #3538, opened 2024-11-29
transforms: use rewriter and listener in convert-stencil-to-csl-stencil

math-fehr:
Stacked PRs:
 * #3540
 * #3539
 * __->__#3538
 * #3537

--- --- ---

### transforms: use rewriter and listener in convert-stencil-to-csl-stencil

The pass was not propagating the listener from the PatternRewriter, and
thus some operations were modified without notifying the rewrite
worklist.


math-fehr:
@n-io, can you confirm that the `coeffs` attribute in `csl_stencil.apply` do not have an order?
This PR changed the order in which the coefficients are stored, but my understanding is that this is irrelevant?

n-io:
Yes, the order is irrelevant since the stencil index is stored with the coefficient value.

n-io (approved):

xdsl/transforms/convert_stencil_to_csl_stencil.py
```diff
@@ -505,12 +506,13 @@ def match_and_rewrite(self, op: stencil.ApplyOp, rewriter: PatternRewriter, /):
         # add operations from list to receive_chunk, use translation table to rebuild operands
         for o in chunk_region_ops:
             if isinstance(o, stencil.ReturnOp | csl_stencil.YieldOp):
+                rewriter.erase_op(o)
```
alexarice, line 509: Why is this new `erase_op` necessary?

alexarice (approved):
Only added one comment. Do you also know what caused the test diff?

xdsl/transforms/convert_stencil_to_csl_stencil.py
compor: good question, if I had to guess, I'd say it seems like defensive programming to me as the yield is added anyhow below. Unless there's something else @math-fehr ?

compor (approved):

PR #3540, opened 2024-11-29
core: Make PatternRewriter a Builder

math-fehr:
Stacked PRs:
 * __->__#3540
 * #3539
 * #3538
 * #3537

--- --- ---

### core: Make PatternRewriter a Builder

This allows us to now use the ImplicitBuilder on PatternRewriter,
or to use `insert` using an InsertPoint.

### DIFF
--- a/xdsl/backend/riscv/riscv_scf_to_asm.py
+++ b/xdsl/backend/riscv/riscv_scf_to_asm.py
@@ -69,15 +69,15 @@ class LowerRiscvScfToLabels(RewritePattern):
         yield_op = body.last_op
         assert isinstance(yield_op, riscv_scf.YieldOp)
 
-        body.insert_ops_after(
+        rewriter.insert_op(
             [
                 riscv.AddOp(get_loop_var, op.step, rd=loop_var_reg),
                 riscv.BltOp(get_loop_var, op.ub, scf_body),
                 riscv.LabelOp(scf_body_end),
             ],
-            yield_op,
+            InsertPoint.after(yield_op),
         )
-        body.erase_op(yield_op)
+        rewriter.erase_op(yield_op)
 
         # We know that the body is not empty now.
         assert body.first_op is not None
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -10,7 +10,7 @@ from typing import TypeVar, Union, final, get_args, get_origin
 
 from typing_extensions import deprecated
 
-from xdsl.builder import BuilderListener
+from xdsl.builder import Builder, BuilderListener
 from xdsl.dialects.builtin import ArrayAttr, DictionaryAttr, ModuleOp
 from xdsl.ir import (
     Attribute,
@@ -77,8 +77,8 @@ class PatternRewriterListener(BuilderListener):
             )
 
 
-@dataclass(eq=False)
-class PatternRewriter(PatternRewriterListener):
+@dataclass(eq=False, init=False)
+class PatternRewriter(Builder, PatternRewriterListener):
     """
     A rewriter used during pattern matching.
     Once an operation is matched, this rewriter is used to apply
@@ -90,6 +90,11 @@ class PatternRewriter(PatternRewriterListener):
 
     has_done_action: bool = field(default=False, init=False)
     """Has the rewriter done any action during the current match."""
+
+    def __init__(self, current_operation: Operation):
+        PatternRewriterListener.__init__(self)
+        self.current_operation = current_operation
+        Builder.__init__(self, InsertPoint.before(current_operation))
 
     def insert_op(
         self, op: Operation | Sequence[Operation], insertion_point: InsertPoint
@@ -833,6 +838,7 @@ class PatternRewriteWalker:
             # Reset the rewriter on `op`
             rewriter.has_done_action = False
             rewriter.current_operation = op
+            rewriter.insertion_point = InsertPoint.before(op)
 
             # Apply the pattern on the operation
             try:
--- a/xdsl/pattern_rewriter.py
+++ b/xdsl/pattern_rewriter.py
@@ -726,7 +726,11 @@ class PatternRewriteWalker:
         """Handle removal of an operation."""
         if self.apply_recursively:
             self._add_operands_to_worklist(op.operands)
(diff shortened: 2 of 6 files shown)


superlopuh (approved):
Fantastico

xdsl/pattern_rewriter.py
```diff
@@ -91,6 +91,11 @@ class PatternRewriter(PatternRewriterListener):
     has_done_action: bool = field(default=False, init=False)
     """Has the rewriter done any action during the current match."""
 
+    def __init__(self, current_operation: Operation):
```
alexarice, line 94: You could possibly do this with a `__post_init__`

xdsl/pattern_rewriter.py
```diff
@@ -91,6 +91,11 @@ class PatternRewriter(PatternRewriterListener):
     has_done_action: bool = field(default=False, init=False)
```
alexarice, line 91:
```suggestion
    has_done_action: bool
```
I think `field` is dead code if you add an init?

xdsl/pattern_rewriter.py
```diff
@@ -726,7 +731,11 @@ def _handle_operation_removal(self, op: Operation) -> None:
         """Handle removal of an operation."""
         if self.apply_recursively:
             self._add_operands_to_worklist(op.operands)
-        self._worklist.remove(op)
+        if op.regions:
```
alexarice, line 734: Is this a separate change to making `PatternRewriter` a `Builder`?

xdsl/pattern_rewriter.py
```diff
@@ -726,7 +731,11 @@ def _handle_operation_removal(self, op: Operation) -> None:
...
+            for sub_op in op.walk():
+                self._worklist.remove(sub_op)
+        else:
+            self._worklist.remove(op)
```
alexarice, line 738:
```suggestion
        self._worklist.remove(op)
```
Surely the operation itself still needs to be removed when it has regions?

xdsl/rewriter.py
```diff
@@ -69,10 +69,10 @@ def erase_op(op: Operation, safe_erase: bool = True):
...
         Otherwise, replace its uses with ErasedSSAValue.
         """
-        assert op.parent is not None, "Cannot erase an operation that has no parents"
-
-        block = op.parent
-        block.erase_op(op, safe_erase=safe_erase)
+        if (block := op.parent) is not None:
+            block.erase_op(op, safe_erase=safe_erase)
+        else:
+            op.erase(safe_erase=safe_erase)
```
alexarice, line 75: Why is it now ok to erase a operation with no parents?

xdsl/transforms/convert_scf_to_openmp.py
```diff
@@ -43,7 +44,8 @@ def match_and_rewrite(self, loop: scf.ParallelOp, rewriter: PatternRewriter, /):
             regions=[Region(Block())],
             operands=[[], [], [], [], [], []],
         )
-        with ImplicitBuilder(parallel.region):
+        rewriter.insertion_point = InsertPoint.at_end(parallel.region.block)
+        with ImplicitBuilder(rewriter):
```
alexarice, line 48: I don't think I'm a fan of this API. Does the user need to reset the insertion point of the rewriter after doing this?

alexarice (commented):
I left some comments but they may be more due to my ignorance than actual problems with the change.

xdsl/backend/riscv/riscv_scf_to_asm.py
```diff
@@ -69,15 +69,15 @@ def match_and_rewrite(self, op: riscv_scf.ForOp, rewriter: PatternRewriter, /):
         yield_op = body.last_op
         assert isinstance(yield_op, riscv_scf.YieldOp)
 
-        body.insert_ops_after(
+        rewriter.insert_op(
```
compor, line 72: What is the benefit of this change?

compor (approved):

xdsl/rewriter.py
superlopuh, line 75: TBH this feels like it could be its own PR, with an accompanying test

xdsl/pattern_rewriter.py
superlopuh: The `self.current_operation = current_operation` seems like strictly `__init__` code to me

math-fehr: I cannot actually do a `__post_init__`, because if I would use the generated `__init__`, this would require the arguments for `PatternRewriterListener` and `Builder`. But the `Builder` arguments gets constructed in the `__init__` I wrote here.

xdsl/pattern_rewriter.py
math-fehr: Nope, the `default` still get added. I removed the `init=False` though, as we don't have an `__init__` here.

xdsl/pattern_rewriter.py
math-fehr: I will rebase this branch on a new one, this was indeed a bug before that can be refactored.

xdsl/rewriter.py
math-fehr: I'll move this to a separate PR as well, the idea is that sometimes we detach an operation and then destroy it.

xdsl/transforms/convert_scf_to_openmp.py
math-fehr: I agree the API is not optimal.
Currently, what we need to use `ImplicitBuilder` is:
* set the insertion point on the rewriter
* Create the ImplicitBuilder using the rewriter

We don't need to reset it after though.

What would you think of `ImplicitBuilder(rewriter, at=InsertPoint.at_end(...))`, which would set the insertion point just for that region of code?

xdsl/backend/riscv/riscv_scf_to_asm.py
math-fehr: If we don't notify the `PatternRewriter` that these operations are added, they are not matched anymore by the `PatternRewriter`.

math-fehr:
Since some time passed, and to simplify splitting this PR, I created a new stack of PRs starting at math-fehr/stack/5
It contains the same things but splitted on 3 different PRs, with more tests.

PR #3542, opened 2024-11-29
dialects: (builtin) don't use data in DenseArrayBase

superlopuh:


xdsl/dialects/gpu.py
```diff
@@ -405,13 +403,13 @@ def __init__(
...
+        if known_block_size is not None:
+            attributes["gpu.known_block_size"] = DenseArrayBase.create_dense_int(
+                i32, known_block_size
             )
-        if knwown_grid_size is not None:
-            attributes["gpu.known_grid_size"] = ArrayAttr(
-                IntegerAttr(i, i32) for i in knwown_grid_size
+        if known_grid_size is not None:
+            attributes["gpu.known_grid_size"] = DenseArrayBase.create_dense_int(
+                i32, known_grid_size
```
superlopuh, line 412: drive-by fix, this is a good reminder to avoid this pattern of building attributes up, as it erases all type information

xdsl/dialects/builtin.py
```diff
@@ -1022,16 +1022,30 @@ def from_list(
...
 
-    def as_tuple(self) -> tuple[int, ...] | tuple[float, ...]:
+    def iter_values(self) -> Iterator[float] | Iterator[int]:
+        """
+        Iterate over "raw" data. This will not apply the datatype restrictions that the
+        array element type would suggest!
+
+        e.g. given a dense<i8: 99999999, 255, 256>, get_values()
+        would return 1234567, 255, 256 and not 135, 255, 0 (mod 256)
+        """
```
jorendumoulin, line 1032: should this check not rather apply to the `__init__` of DenseArrayBase rather than the iterator of it?

xdsl/dialects/builtin.py
```diff
@@ -1022,16 +1022,30 @@ def from_list(
...
+        return (attr.data for attr in self.data.data)
```
jorendumoulin, line 1033: ?
```suggestion
        return iter(self.data.data)
```

xdsl/dialects/builtin.py
jorendumoulin: I would feel the creation of an illegal `dense<i8: 99999999, 255, 256>` is the issue here, not the fact i want to iterate over it

xdsl/dialects/builtin.py
```diff
@@ -1022,16 +1022,30 @@ def from_list(
...
             raise TypeError(f"Unsupported element type {data_type}")
 
-    def as_tuple(self) -> tuple[int, ...] | tuple[float, ...]:
+    def iter_values(self) -> Iterator[float] | Iterator[int]:
+        """
+        Iterate over "raw" data. This will not apply the datatype restrictions that the
+        array element type would suggest!
+
+        e.g. given a dense<i8: 99999999, 255, 256>, get_values()
+        would return 1234567, 255, 256 and not 135, 255, 0 (mod 256)
```
jorendumoulin, line 1031:
```suggestion
        would return 99999999, 255, 256 and not 135, 255, 0 (mod 256)
```

xdsl/dialects/cf.py
```diff
@@ -239,7 +239,7 @@ def case_operand(self) -> tuple[VarOperand, ...]:
 
         def_sizes = cast(
             tuple[int, ...],
-            tuple(size_attr.data for size_attr in self.case_operand_segments.data.data),
+            tuple(size for size in self.case_operand_segments.iter_values()),
```
jorendumoulin, line 242:
```suggestion
            self.case_operand_segments.get_values(),
```

xdsl/dialects/builtin.py
superlopuh: TBH this will no longer hold after we move to the bytes representation, so I'm not sure that it's worth the effort...

superlopuh: On the other hand, we should probably notify users when their input integers are too large for the data type

jorendumoulin: okay, `struct.pack` throws the error if you try to convert a too large value to bytes anyway

superlopuh: Ah, good for us to check it first then

jorendumoulin (approved):

PR #3544, opened 2024-11-29
dialects: (builtin) verify that the elements of dense satisfy type

superlopuh:
Instead of scary doc string, warn the user at the point of DenseArrayBase parsing/creation.


jorendumoulin (approved):
Nice!

When converting to bytes type, we'll probably have to adjust the ranges of the Signedness values to the ones of `struct.pack`, or add some preprocessing logic.
```
>>> pack('@b', 128)
  File "<stdin>", line 1, in <module>
struct.error: 'b' format requires -128 <= number <= 127
```

math-fehr (approved):
Nice!

PR #3546, opened 2024-11-29
dialects (arm): add func op

emmau678:
Add funcOp for ARM, similar to that already in RISCV.

### DIFF
--- a/xdsl/dialects/arm_func.py
+++ b/xdsl/dialects/arm_func.py
@@ -1,26 +1,137 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+
 from xdsl.dialects import arm
-from xdsl.dialects.builtin import StringAttr
-from xdsl.ir import Dialect
+from xdsl.dialects.builtin import FunctionType, StringAttr
+from xdsl.dialects.utils import (
+    parse_func_op_like,
+    print_func_op_like,
+)
+from xdsl.ir import Attribute, Dialect, Operation, Region
 from xdsl.irdl import (
+    attr_def,
     irdl_op_definition,
+    opt_attr_def,
+    region_def,
     traits_def,
 )
-from xdsl.traits import IsTerminator
+from xdsl.parser import Parser
+from xdsl.printer import Printer
+from xdsl.traits import (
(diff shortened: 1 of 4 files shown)


emmau678:
thanks, could you please check if I've done that correctly when you have a chance? I just added an extra assertion to the existing pytest

xdsl/dialects/arm_func.py
```diff
@@ -40,12 +151,13 @@ def assembly_line_args(self):
         return ()
 
     def assembly_instruction_name(self) -> str:
-        return "bx lr"
+        return "ret"
```
alexarice, line 154: These feel like a separate change?

xdsl/dialects/arm_func.py
```diff
@@ -1,26 +1,137 @@
...
+            raise NotImplementedError("arg_attrs not implemented in riscv_func")
+        func = FuncOp(name, region, (input_types, return_types), visibility)
+        if extra_attrs is not None:
+            func.attributes |= extra_attrs.data
+        return func
+
+    def print(self, printer: Printer):
+        if self.sym_visibility:
+            visibility = self.sym_visibility.data
+            printer.print(f" {visibility}")
```
alexarice, line 106:
```suggestion
            printer.print_string(f" {visibility}")
```

alexarice (commented):
Is there a good reason for allowing a region with no blocks? As far as I can tell it seems this has no representation in the assembly.

xdsl/dialects/arm_func.py
alexarice: "these" being this and the comment above.

superlopuh:
A function operation with a region with no blocks is a function declaration.

alexarice:
So it's necessary in mlir just to declare the symbol?

superlopuh:
I'm not 100% sure about this, but I think MLIR verifies that you only call functions that are declared in the module, so it's a bit of a consistency check. Also, I think that the external functions may have an assembly representation, for use during linking, but these also seem optional.

alexarice (approved):

PR #3547, opened 2024-11-30
dialects: (builtin) add PackableType and StructPackableType

superlopuh:
Introduces three ABCs, that allow users to encode and decode values to buffers. There are a few things I'm not sure about in this PR, and would love some help with:

1. Should the PyBackedType be rolled into PackableType? I kind of like the separation but I'm not sure it's necessary
2. Would it be OK to merge with things as-is, as they seem to work, or should we aim for a more complete implementation for different bitwidths for both ints and floats before merging? Maybe just the Ints?
3. Is deferring the normalisation of ints to a future PR OK? We've been OK this far, but actually it's a little problematic that `IntegerAttr(-128, i8) != IntegerAttr(255, i8)`

There are two more PRs incoming, one that swaps out the xtype in `ptr.py` to use xDSL types instead, and one that replaces DenseArrayBase's data with bytes, using the code in this PR, so it all works for the in-tree dialects and tests.

### DIFF
--- a/tests/dialects/test_builtin.py
+++ b/tests/dialects/test_builtin.py
@@ -1,3 +1,4 @@
+import math
 import re
 from collections.abc import Sequence
 
@@ -31,6 +32,7 @@ from xdsl.dialects.builtin import (
     VectorRankConstraint,
     VectorType,
     f32,
+    f64,
     i8,
     i32,
     i64,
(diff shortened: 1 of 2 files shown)


superlopuh:
mlir-opt normalises signless integers to the signed range, I'll do a PR first to do the same:

Input:
``` mlir
"test.op"() {hello=array<i8: -128, -127, -1, 0, 1, 255>}: () -> ()
```

Output:
``` mlir
"test.op"() {hello = array<i8: -128, -127, -1, 0, 1, -1>} : () -> ()
```

jorendumoulin:
Overall this looks very good to me!
I'm just not sure about the `PyBackedType`. To me it's a bit ambiguous on how it relates to the `Data` attribute:

```python
@dataclass(frozen=True)
class Data(Generic[DataElement], Attribute, ABC):
    """An attribute represented by a Python structure."""
```
This is also a an attribute that is backed by a python type. I also don't  directly see a practical use of having the function `py_type`? Is specifying the generic type not enough?

superlopuh:
Good points, I'll merge them.

superlopuh:
I forgot I had opened this

PR #3550, opened 2024-12-02
dialects: (builtin) move value range helper definitions to utils

superlopuh:
This flips the dependency of the two files, I'd like to add some more helpers to the utils one for use in `builtin.py`.

### DIFF
--- a/xdsl/utils/comparisons.py
+++ b/xdsl/utils/comparisons.py
@@ -56,16 +56,16 @@ def signed_upper_bound(bitwidth: int) -> int:
 
 def unsigned_value_range(bitwidth: int) -> tuple[int, int]:
     """
-    For a given bitwidth, returns the range `[min, max+1)`, where min and max are the
-    smallest and largest representable values.
+    For a given bitwidth, returns a tuple `(min, max)`, such that unsigned integers of
+    this bitwidth are in the range [`min`, `max`).
     """
     return 0, unsigned_upper_bound(bitwidth)
 
 
 def signed_value_range(bitwidth: int) -> tuple[int, int]:
     """
-    For a given bitwidth, returns the range `[min, max+1)`, where min and max are the
-    smallest and largest representable values.
+    For a given bitwidth, returns a tuple `(min, max)`, such that signed integers of
+    this bitwidth are in the range [`min`, `max`).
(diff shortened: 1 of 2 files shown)


xdsl/utils/comparisons.py
```diff
@@ -32,31 +32,59 @@
...
     The maximum representable value + 1.
     """
-    _, ub = Signedness.SIGNED.value_range(bitwidth)
-    return ub
+    return 1 << (bitwidth - 1)
+
+
+def unsigned_value_range(bitwidth: int) -> tuple[int, int]:
+    """
+    For a given bitwidth, returns (min, max+1), where min and max are the smallest and
```
jorendumoulin, line 59: why not just `(min, max)`?

superlopuh: It slightly complicates some calculations, which is not a big deal, but a little annoying. I also find that this inclusive-lb-exclusive-ub form is more intuitive.

alexarice: I'm not sure if this is a mathematician thing but this can also be written by saying that the integer takes values in the range `[min, max)`

jorendumoulin (approved):

xdsl/utils/comparisons.py
superlopuh: this is a me being ambiguous thing, I meant for the `()` to be tuple begin end markers, I'll rephrase

superlopuh: How's the new version?

xdsl/utils/comparisons.py
```diff
@@ -32,36 +32,66 @@
...
+    """
+    min_value = signed_lower_bound(bitwidth)
+    max_value = signed_upper_bound(bitwidth)
+
+    return min_value, max_value
+
+
+def signless_value_range(bitwidth: int) -> tuple[int, int]:
+    """
+    For a given bitwidth, returns the range `[min, max+1)`, where min and max are the
```
alexarice, line 78: I think my suggestion has confused things. How about:
"For a given bitwidth, returns a tuple `(min, max)`, such that signless integers of this bitwidth are in the range `[min, max)`"

PR #3551, opened 2024-12-02
core: fix (and test) operands/results directives with no variadics

alexarice:
Managed to miss the test for when there are no variadic/optional operands or results, and of course it turns out that this case is completely broken.

Actual functionality change is 3 characters.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -629,7 +629,7 @@ class OperandsOrResultDirective(VariadicTypeableDirective, ABC):
         if self.variadic_index is None:
             if len(set_to) != len(field):
                 return f"Expected {len(field)} {field_name} but found {len(set_to)}"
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -629,7 +629,7 @@ def _set_using_variadic_index(
         if self.variadic_index is None:
             if len(set_to) != len(field):
                 return f"Expected {len(field)} {field_name} but found {len(set_to)}"
-            field = [o for o in set_to]  # Copy needed as list is not covariant
+            field[:] = [o for o in set_to]  # Copy needed as list is not covariant
```
alexarice, line 632: Perhaps the copy is not needed any more, I'll investigate

alexarice: Seems it isn't, will remove

superlopuh (approved):

PR #3552, opened 2024-12-02
core: fix bitwidth 0 value range

superlopuh:
I'm not 100% sure of my understanding of the contract of i0 and other width-0 integer types, but some logic I'm going to open a PR for soon depends on this not crashing and being explicitly stated.

### DIFF
--- a/xdsl/utils/comparisons.py
+++ b/xdsl/utils/comparisons.py
@@ -51,7 +51,7 @@ def signed_upper_bound(bitwidth: int) -> int:
     """
     The maximum representable value + 1.
     """
-    return 1 << (bitwidth - 1)
+    return 1 << max(bitwidth - 1, 0)
 
 
 def unsigned_value_range(bitwidth: int) -> tuple[int, int]:
--- a/tests/utils/test_comparisons.py
+++ b/tests/utils/test_comparisons.py
@@ -1,9 +1,12 @@
 from xdsl.utils.comparisons import (
     signed_lower_bound,
     signed_upper_bound,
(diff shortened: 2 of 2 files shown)


xdsl/utils/comparisons.py
```diff
@@ -51,7 +51,7 @@ def signed_upper_bound(bitwidth: int) -> int:
     """
     The maximum representable value + 1.
     """
-    return 1 << (bitwidth - 1)
+    return 1 << max(bitwidth - 1, 0)
```
alexarice, line 54: I don't like this but it does look correct

alexarice (commented):
MLIR doesn't support 0-width types. I'm not necessarily opposed to having them but it seems they don't work particularly nicely with the signed max value calculation. What's the motivation for having these?

math-fehr (approved):

Moxinilian:
What do you mean "unfortunately"?! i0 support is super important. There are many things you can’t do without it.

The consistent, less arbitrary way to compute this is to do `(1 << bitwidth) >> 1`, by the way.

superlopuh:
What am I missing?
```
>>> bitwidth = 0
>>> 1 << max(bitwidth - 1, 0)
1
>>> (1 << bitwidth) >> 1
0
```

Moxinilian:
Sorry, I was confusing this with the minimum signed value (`-((1 << n) >> 1`). For the max unsigned value, the branchless approach would be `((2 << n) >> 1) - 1`, or more simply `2 ** n - 1`.

PR #3554, opened 2024-12-02
dialects: (builtin) normalize signless values on init

superlopuh:
Follows MLIR behaviour.


jorendumoulin (approved):
Nice!

xdsl/dialects/builtin.py
```diff
@@ -976,10 +996,17 @@ def create_dense_int(
...
+            attr_list = tuple(IntAttr(d) for d in cast(Sequence[int], data))
         else:
             attr_list = cast(Sequence[IntAttr], data)
 
+        if data_type.signedness.data == Signedness.SIGNLESS:
+            signed_ub = signed_upper_bound(data_type.bitwidth)
+            if any(signed_ub <= attr.data for attr in attr_list):
+                attr_list = tuple(
+                    IntAttr(data_type.normalize_value(attr.data)) for attr in attr_list
+                )
```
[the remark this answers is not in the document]
superlopuh: What do you mean by doing it every time?

tests/filecheck/parser-printer/builtin_attrs.mlir
```diff
@@ -174,7 +174,7 @@
                       value1 = dense<[true, false]> : tensor<2xi1>,
                       sym_name = "dense_bool_attr"} : () -> ()
 
-  // CHECK: "value1" = dense<[1, 0]> : tensor<2xi1>
+  // CHECK: "value1" = dense<[-1, 0]> : tensor<2xi1>
```
superlopuh, line 177: this feels weird, I would expect us to print `true, false`, looking into this now

tests/irdl/test_declarative_assembly_format.py
superlopuh, line 1: same here, this feels off to me, why aren't we printing true and false? looking now

tests/filecheck/parser-printer/builtin_attrs.mlir
alexarice: This will likely be because these are being printed with `print_without_type`, which I expect has no `true`/`false` special casing

alexarice: or the data is being printed directly

tests/dialects/test_builtin.py
```diff
@@ -58,6 +60,51 @@ def test_IntegerType_size():
     assert IntegerType(64).size == 8
 
 
+def test_IntegerType_normalize():
+    s8 = IntegerType(8, Signedness.SIGNED)
```
alexarice, line 64: Extreme nit but I believe `si` is usually used in mlir as a prefix for signed integers

tests/dialects/test_builtin.py
```diff
@@ -58,6 +60,51 @@ def test_IntegerType_size():
...
+    u8 = IntegerType(8, Signedness.UNSIGNED)
+
+    assert i8.normalize_value(-1) == -1
+    assert i8.normalize_value(1) == 1
+    assert i8.normalize_value(255) == -1
+
+    assert s8.normalize_value(-1) == -1
+    assert s8.normalize_value(1) == 1
+    assert s8.normalize_value(255) == 255
```
alexarice, line 73: I'm not sure I understand the semantics here, do we not change values that fall outside the range?

tests/dialects/test_builtin.py
```diff
@@ -58,6 +60,51 @@ def test_IntegerType_size():
...
+    assert u8.normalize_value(-1) == -1
+    assert u8.normalize_value(1) == 1
+    assert u8.normalize_value(255) == 255
+
+
+def test_IntegerAttr_normalize():
+    """
+    Test that the value within the accepted signless range is normalized to signed
+    range.
+
```
alexarice, line 84: Empty line

xdsl/dialects/builtin.py
```diff
@@ -395,6 +396,23 @@ def verify_value(self, value: int):
...
+        same bitwidth.
+        Values outside the valid range are left as-is.
+        """
+        if self.signedness.data != Signedness.SIGNLESS:
+            return value
+
+        signed_ub = signed_upper_bound(self.bitwidth)
+        unsigned_ub = signed_ub << 1
+
+        if signed_ub <= value < unsigned_ub:
```
alexarice, line 411:
```suggestion
        if signed_ub < value < unsigned_ub:
```
signed_ub is exclusive right?

xdsl/dialects/builtin.py
```diff
@@ -395,6 +396,23 @@ def verify_value(self, value: int):
...
+        """
+        Normalize signless values within the expected range to the signed range for the
+        same bitwidth.
+        Values outside the valid range are left as-is.
+        """
+        if self.signedness.data != Signedness.SIGNLESS:
+            return value
+
+        signed_ub = signed_upper_bound(self.bitwidth)
+        unsigned_ub = signed_ub << 1
```
alexarice, line 409:
```suggestion
        unsigned_ub = unsigned_upper_bound(self.bitwidth)
```
Why not use this? There will be complications with `i0` right?

xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -142,9 +142,12 @@ def match_and_rewrite(self, op: cf.ConditionalBranchOp, rewriter: PatternRewrite
...
 
-        if cond == 1:
+        if cond is None:
+            return
+
+        if cond:
             rewriter.replace_matched_op(cf.BranchOp(op.then_block, *op.then_arguments))
-        elif cond == 0:
+        else:
             rewriter.replace_matched_op(cf.BranchOp(op.else_block, *op.else_arguments))
```
alexarice, line 151: I can't work out how to attach the comment in the correct place but line 374 of this file will now be incorrect

xdsl/dialects/builtin.py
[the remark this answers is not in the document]
alexarice: I agree, isn't `normalize_value` doing the same checks anyway?

alexarice (commented):
I think this likely subtly changes a number of existing passes, which compare integers by grabbing the data out of an `IntegerAttr`. I've caught one in the `cf` canonicalization patterns but I'm not what can be done to catch this in general.

tests/dialects/test_builtin.py
superlopuh: oh this is great, I couldn't remember it, and it's not in the IntegerAttr doc, of course

tests/dialects/test_builtin.py
superlopuh: Yep, and your comment means that I should have documented this better: the idea is that we want to first preserve the behaviour for invalid integers, to raise a VerifyException with the invalid value. If the valid value is ambiguous due to signlessness, only then normalize to the signed range. I'll add this somewhere.
- [ ] add more explanation to doc comment

xdsl/dialects/builtin.py
superlopuh: it's exclusive on the other side, meaning that a value that is ub is not within the accepted range, and should be changed

xdsl/dialects/builtin.py
superlopuh: not really, this was a micro-optimisation, but I guess not really very important

xdsl/dialects/builtin.py
alexarice: Of course

superlopuh:
I'm converting to draft as this seems like quite large a change, I'll split this up further by the seemingly unrelated things that are failing.

tests/filecheck/backend/csl/print_csl.mlir
```diff
@@ -649,7 +649,7 @@ csl.func @builtins() {
 // CHECK-NEXT:   return;
 // CHECK-NEXT: }
 // CHECK-NEXT: comptime {
-// CHECK-NEXT:   @bind_control_task(control_task, @get_control_task_id(42));
+// CHECK-NEXT:   @bind_control_task(control_task, @get_control_task_id(-22));
```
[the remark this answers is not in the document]
n-io: Not quite sure about this change? It should prob not print negative values.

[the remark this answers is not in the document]
superlopuh: OK I'll attempt to fix the printer then, as the signless before this change may already be in the negative range.

xdsl/dialects/builtin.py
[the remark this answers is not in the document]
superlopuh: I generally dislike the idea of constructors being written in a way that doesn't allow the user to pass in correct data that is not copied/transformed in some way. I'll rewrite this part to conditionally normalize but return the original IntAttr if there's nothing to change.

math-fehr (approved):

PR #3555, opened 2024-12-02
dialects: (builtin) fix true/false printing

superlopuh:
We currently don't consistently print true and false for i1 values, this PR fixes that in the cases I could find.


xdsl/dialects/builtin.py
```diff
@@ -513,7 +528,7 @@ def parse_with_type(
         return IntegerAttr(parser.parse_integer(allow_boolean=(type == i1)), type)
 
     def print_without_type(self, printer: Printer):
-        return printer.print(self.value.data)
+        self.type.print_value_without_type(self.value.data, printer)
```
alexarice, line 531: The polymorphism feels unnecessary to me

alexarice (approved):

xdsl/dialects/builtin.py
superlopuh: What do you mean?

superlopuh: I considered it and then felt like doing Python runtime polymorphism is cleaner, if only because we have types like DenseArrayBase, and DenseIntOrFPAttr that will soon have a type and then lots of data, and it seems a little simpler to have the functionality available on the type directly.

PR #3557, opened 2024-12-02
dialects: (arith) add muli canonicalization

alexarice:
Adds two canonicalization patterns to `arith.muli`

The first checks if the left hand side is a constant 1, and removes the multiplication if so.
The second reverses the order of arguments if the right hand side is a constant and the left isn't, and folds the operation if both are constants.

This likely has some merge conflicts with the stuff @superlopuh is doing.

I also had to modify a test for the interactive mode, would appreciate a review on that part.

### DIFF
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -101,8 +101,8 @@ func.func @test_const_var_const() {
 // CHECK: "test.op"(%a, %a) {"identity multiplication check"} : (i32, i32) -> ()
 "test.op"(%one_times, %times_one) {"identity multiplication check"} : (i32, i32) -> ()
 
-// CHECK: %times_by_const = arith.muli %c2, %a : i32
-%times_by_const = arith.muli %a, %c2 : i32
+// CHECK: %times_by_const = arith.muli %a, %c2 : i32
+%times_by_const = arith.muli %c2, %a : i32
 "test.op"(%times_by_const) : (i32) -> ()
 
 // CHECK: %foldable_times = arith.constant 4 : i32
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -90,3 +90,21 @@ func.func @test_const_var_const() {
 %z = arith.select %x, %y, %y : i64
 
 "test.op"(%z) : (i64) -> ()
+
+%c1 = arith.constant 1 : i32
+%c2 = arith.constant 2 : i32
+%a = "test.op"() : () -> (i32)
+
+%one_times = arith.muli %c1, %a : i32
+%times_one = arith.muli %a, %c1 : i32
+
+// CHECK: "test.op"(%a, %a) {"identity multiplication check"} : (i32, i32) -> ()
+"test.op"(%one_times, %times_one) {"identity multiplication check"} : (i32, i32) -> ()
+
+// CHECK: %times_by_const = arith.muli %c2, %a : i32
+%times_by_const = arith.muli %a, %c2 : i32
+"test.op"(%times_by_const) : (i32) -> ()
+
+// CHECK: %foldable_times = arith.constant 4 : i32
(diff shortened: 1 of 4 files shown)


xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -163,3 +164,30 @@ class SelectSamePattern(RewritePattern):
...
+        rewriter.replace_matched_op((), (op.rhs,))
+
+
+class MuliConstantProp(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: arith.MuliOp, rewriter: PatternRewriter):
+        if (rhs := const_evaluate_operand(op.rhs)) is None:
+            return
+        if (lhs := const_evaluate_operand(op.lhs)) is None:
+            # Swap inputs if rhs is constant and lhs is not
```
superlopuh, line 186: add filecheck for this?

superlopuh (approved):
LGTM, and I don't expect it to conflict

tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -90,3 +90,21 @@ func.func @test_const_var_const() {
...
+%a = "test.op"() : () -> (i32)
+
+%one_times = arith.muli %c1, %a : i32
+%times_one = arith.muli %a, %c1 : i32
+
+// CHECK: "test.op"(%a, %a) {"identity multiplication check"} : (i32, i32) -> ()
+"test.op"(%one_times, %times_one) {"identity multiplication check"} : (i32, i32) -> ()
+
+// CHECK: %times_by_const = arith.muli %c2, %a : i32
+%times_by_const = arith.muli %a, %c2 : i32
```
alexarice, line 105: The filecheck is here

tests/interactive/test_app.py
```diff
@@ -283,8 +289,14 @@ def callback(x: str):
...
         # assert after "Condense Button" is clicked that the state and condensed_pass list change accordingly
         assert app.condense_mode is True
+        rewrites = get_all_possible_rewrites(
+            expected_module,
+            individual_rewrite.REWRITE_BY_NAMES,
+        )
         assert app.available_pass_list == get_condensed_pass_list(
             expected_module, app.all_passes
+        ) + convert_indexed_individual_rewrites_to_available_pass(
+            rewrites, expected_module
```
alexarice, line 299: @superlopuh does this seem reasonable? I'm not sure I could think of a better way to do it.

tests/filecheck/dialects/arith/canonicalize.mlir
superlopuh: Ah nice I hadn't spotted it, thank you.

tests/interactive/test_app.py
superlopuh: Yep I think so, this part of the test should verify that the signals get passed around successfully, so we just need to check the logic here.

n-io:
> The second reverses the order of arguments if the right hand side is a constant and the left isn't, and folds the operation if both are constants.

Note, that upstream places constants in the rhs.

n-io (commented):
> The second reverses the order of arguments if the right hand side is a constant and the left isn't, and folds the operation if both are constants.

Note, that upstream places constants in the rhs.

n-io:
If you're looking into upstream canonicalisation patterns, you could also consider adding `constant * constant`.

alexarice:
Great to know, I was mainly following the semi-convention set by `AddImmediateZero` in this file, but I'm happy to go with the other way round. This PR already deals with `constant * constant`

alexarice:
Have switched it so that constants are pushed to the right

superlopuh:
Would it be worth checking for things like overflow? CC @math-fehr are there flags that the pattern might check for before doing canonicalisation this way?

alexarice:
It has the commutative trait which I assumed means that it is always safe to swap the arguments

superlopuh (approved):

superlopuh:
Yep actually my bad this should always be fine.

n-io (approved):

PR #3558, opened 2024-12-02
dialects: (csl) handle negative-valued task ID representation

superlopuh:
i6 and friends are signless, meaning that they are bitpatterns that don't have an inherent value. When printing CSL, we currently take the raw data of the IntegerAttr, but it seems like that's not valid for task identifiers, which are positive. This PR changes the printer to always print the unsigned version of the number, and adds a test case to the CSL filecheck.

### DIFF
--- a/tests/filecheck/backend/csl/print_csl.mlir
+++ b/tests/filecheck/backend/csl/print_csl.mlir
@@ -167,19 +167,15 @@
   }
 
 
-  csl.task @data_task(%arg: f32) attributes {kind = #csl<task_kind data>, id = 0 : i5} {
+  csl.task @data_task(%arg: f32) attributes {kind = #csl<task_kind data>, id = 0 : ui5} {
     csl.return
   }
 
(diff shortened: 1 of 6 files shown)


dk949:
I think this makes sense, even if it is really odd to see a negative task ID in the IR.

As a sidenote, it would've probably been better to have made this an `i16` and dealt with the range in the `verify`

superlopuh:
There are probably a few things to do, and making it an unsigned integer is also an option, especially if you'll not do any more arithmetic with it once you pin it down. If you'd like, I can do that in this PR instead?

superlopuh:
I decided to try it out, happy to revert to previous version if you'd like to keep the IDs signless.

n-io (approved):
Looks good, @dk949 ?

dk949 (approved):
I like this, it exactly expresses the constraints on these values in the types.

emmau678 (approved):

PR #3559, opened 2024-12-02
core: allow result variadic inference

alexarice:
Sort of a revert of #3416, but I believe this fixes the underlying problem.

Changes range constraint inference so that passing in a length is optional. The rationale for this is that the length of operands is always known, but the length of results is never known (as they can always be omitted). Similarly updates the `can_infer` function.

Probably needs a test of something. :shrug:

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -677,17 +677,21 @@ class GenericRangeConstraint(Generic[AttributeCovT], ABC):
         """
         return {}
 
-    def can_infer(self, var_constraint_names: Set[str]) -> bool:
+    def can_infer(self, var_constraint_names: Set[str], length_known: bool) -> bool:
         """
         Check if there is enough information to infer the attribute given the
-        constraint variables that are already set.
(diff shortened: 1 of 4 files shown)


xdsl/irdl/constraints.py
```diff
@@ -677,17 +677,21 @@ def get_variable_extractors(
         """
         return {}
 
-    def can_infer(self, var_constraint_names: Set[str]) -> bool:
+    def can_infer(self, var_constraint_names: Set[str], length_known: bool) -> bool:
```
superlopuh, line 680: At the call site, the `True` or `False` don't communicate much in terms of what's going on, how do you feel about making the argument mandatory?
```suggestion
    def can_infer(self, var_constraint_names: Set[str], *, length_known: bool) -> bool:
```

superlopuh (approved):
Very very nice

alexarice:
I'm think I'm going to add a test that shows that trying to infer a `var_result_def(IntegerType(64))` fails, then this should be good to go

superlopuh (approved):

PR #3560, opened 2024-12-02
dialects: (bufferization) add CloneOp

jorendumoulin:


xdsl/dialects/bufferization.py
```diff
@@ -135,6 +135,18 @@ def __init__(
...
+class CloneOp(IRDLOperation):
+    name = "bufferization.clone"
+
+    T: ClassVar = VarConstraint("T", AnyMemRefTypeConstr | AnyUnrankedMemrefTypeConstr)
+
+    input = operand_def(T)
+    output = result_def(T)
+
+    assembly_format = "$input attr-dict `:` type($input) `to` type($output)"
+
```
alexarice, line 148: Would it be worth adding an init method?

alexarice (approved):

xdsl/dialects/bufferization.py
```diff
@@ -135,6 +135,22 @@ def __init__(
...
+    def __init__(self, input: SSAValue | Operation):
+        result_type = SSAValue.get(input).type
+        super().__init__(operands=(input,), result_types=(result_type,))
```
alexarice, line 151: This looks fine but in future it might be better to wait for a second review after adding more code

PR #3563, opened 2024-12-03
dialects: (arith) add addi constant propagation

alexarice:
Moves constants to the right for an addi operation, similar to what #3557 did for muli. Also replaces `AddImmediateZero` by `AddiIdentityRight`, which removes constant zeros on the right of an addition instead of the left of an addition like the previous pattern did.

Updates interactive tests (and driveby fixes the constant zero being given the SSA value name %two in this test).

### DIFF
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -115,7 +115,7 @@ func.func @test_const_var_const() {
 %plus_zero = arith.addi %a, %c0 : i32
 
 // CHECK: "test.op"(%a, %a) {"identity addition check"} : (i32, i32) -> ()
-"test.op"(%one_times, %times_one) {"identity addition check"} : (i32, i32) -> ()
+"test.op"(%zero_plus, %plus_zero) {"identity addition check"} : (i32, i32) -> ()
 
 // CHECK: %plus_const = arith.addi %a, %c2 : i32
 %plus_const = arith.addi %c2, %a : i32
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -108,3 +108,19 @@ func.func @test_const_var_const() {
 // CHECK: %foldable_times = arith.constant 4 : i32
 %foldable_times = arith.muli %c2, %c2 : i32
 "test.op"(%foldable_times) : (i32) -> ()
+
+%c0 = arith.constant 0 : i32
+
+%zero_plus = arith.addi %c0, %a : i32
+%plus_zero = arith.addi %a, %c0 : i32
+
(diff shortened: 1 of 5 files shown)


n-io (approved):

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -9,15 +9,30 @@
...
+        if (lhs := const_evaluate_operand(op.lhs)) is None:
+            return
+        if (rhs := const_evaluate_operand(op.rhs)) is None:
+            # Swap inputs if lhs is constant and rhs is not
+            rewriter.replace_matched_op(arith.AddiOp(op.rhs, op.lhs))
+            return
+
+        assert isinstance(op.result.type, IntegerType | IndexType)
+        rewriter.replace_matched_op(
+            arith.ConstantOp.from_int_and_width(lhs + rhs, op.result.type)
```
superlopuh, line 34: overflow?

superlopuh: I guess we'll crash in the short term and will be able to deal with it then. I just thought I'd raise it now

alexarice: This probably effects the previous PR too

alexarice: I think this is the equivalent part of mlir https://github.com/llvm/llvm-project/blob/9ad09b2930ef2e95bf8772c91f623881d1c14733/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L325

I can't see any mention of overflow flags or wrapping.

tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -108,3 +108,19 @@ func.func @test_const_var_const() {
...
+
+%zero_plus = arith.addi %c0, %a : i32
+%plus_zero = arith.addi %a, %c0 : i32
+
+// CHECK: "test.op"(%a, %a) {"identity addition check"} : (i32, i32) -> ()
+"test.op"(%one_times, %times_one) {"identity addition check"} : (i32, i32) -> ()
+
+// CHECK: %plus_const = arith.addi %a, %c2 : i32
+%plus_const = arith.addi %c2, %a : i32
+"test.op"(%plus_const) : (i32) -> ()
```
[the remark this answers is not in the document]
superlopuh: Unless I'm missing something, it looks like this is a copy/paste from above, but is not necessary

superlopuh (approved):

tests/filecheck/dialects/arith/canonicalize.mlir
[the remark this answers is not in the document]
alexarice: It was also a copy paste

alexarice:
Have fixed the test and resolved the conflict

compor (approved):

PR #3574, opened 2024-12-05
interpreter: (pdl) add an initial version of EqSat PDLMatcher

jianyicheng:
This is an initial attempt of adding a PDL matcher for EqSat e-class ops.


tests/interpreters/test_eqsat_pdl_interpreter.py
```diff
@@ -0,0 +1,393 @@
...
+    assert matcher.match_type(ssa_value, pdl_op, xdsl_value)
+    assert matcher.matching_context == {ssa_value: xdsl_value}
+
+    # Same value
+    assert matcher.match_type(ssa_value, pdl_op, xdsl_value)
+    assert matcher.matching_context == {ssa_value: xdsl_value}
+
+    # Other value
+    assert not matcher.match_type(ssa_value, pdl_op, StringAttr("b"))
+    assert matcher.matching_context == {ssa_value: xdsl_value}
```
superlopuh, line 32: Since only match_operand is overridden, it seems like all the other unit tests are not needed, since the other functions are already tested? Could you please remove all the tests other than `test_match_operand`, and add the case to check that `NotImplementedError` is raised as expected?

jianyicheng: Sorry, just saw your comments - I made some changes and it now passed the tests... what do you think?

superlopuh: Looks great!

superlopuh (approved):
This is perfect! Thank you

PR #3576, opened 2024-12-06
use shrinkray to shrink a test case

superlopuh:
CC @math-fehr @n-io @bollu

I thought I'd use the power of automation to find a smaller test case for one of the rewrite patterns, this case is not very readable, but can be formatted to be more readable!

Command used to generate the reduced test case
`shrinkray hello.sh tests/filecheck/transforms/stencil-inlining.mlir`


superlopuh:
@math-fehr, do you think it would be worth adding shrinkray as an optional dependency for this sort of exercise?

tests/filecheck/transforms/stencil-inlining.mlir.reduced
superlopuh, line 1: Original file: https://github.com/xdslproject/xdsl/blob/994f9b003a11cc2466e36154baaf393e4be632ae/tests/filecheck/transforms/stencil-inlining.mlir

PR #3581, opened 2024-12-06
dialects: (builtin) add PackableType and StructPackableType

superlopuh:
These helper functions will help us deal with dense representations of values.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import math
+import struct
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass
@@ -85,8 +86,11 @@ from xdsl.utils.hints import isa
 from xdsl.utils.isattr import isattr
 
 if TYPE_CHECKING:
+    from _typeshed import ReadableBuffer, WriteableBuffer
+
     from xdsl.parser import AttrParser, Parser
     from xdsl.printer import Printer
+
 
 DYNAMIC_INDEX = -1
 """
@@ -347,10 +351,10 @@ class SignednessAttr(Data[Signedness]):
 
 class FixedBitwidthType(TypeAttribute, ABC):
     """
-    A type attribute with a defined bitwidth
+    A type attribute whose runtime bitwidth is target-independent.
     """
 
-    name = "abstract.bitwidth_type"
+    name = "abstract.fixed_bitwidth_type"
 
     @property
     @abstractmethod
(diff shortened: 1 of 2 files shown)


xdsl/dialects/builtin.py
```diff
@@ -368,8 +372,76 @@ def size(self) -> int:
...
+        https://docs.python.org/3/library/struct.html
+        """
+        raise NotImplementedError()
+
+    def iter_unpack(self, buffer: ReadableBuffer, /) -> Iterator[_PyT]:
+        return (values[0] for values in struct.iter_unpack(self.format, buffer))
+
+    def unpack(self, buffer: ReadableBuffer, num: int, /) -> tuple[_PyT, ...]:
+        fmt = self.format[0] + str(num) + self.format[1:]
+        return struct.unpack(fmt, buffer)
```
jorendumoulin, line 433:
```suggestion
    def unpack(self, buffer: ReadableBuffer, /) -> tuple[_PyT, ...]:
        nb_elements = len(buffer) // struct.calcsize(self.format())
        fmt = self.format[0] + str(nb_elements) + self.format[1:]
        return struct.unpack(fmt, buffer)
```
This seems nicer as to not having to supply the number of elements everytime?

xdsl/dialects/builtin.py
```diff
@@ -368,8 +372,76 @@ def size(self) -> int:
         return (self.bitwidth + 7) >> 3
 
 
+_PyT = TypeVar("_PyT")
+
+
+class PackableType(Generic[_PyT], FixedBitwidthType, ABC):
```
jorendumoulin, line 378: Do we want to combine the `FixedBitwidthType` and `PackableType`? Maybe we are mixing host bitdwidth with target bitwidth? The target system could for example have support for `i4` types, but will be packed as `i8` at compile time.

Maybe this is too niche, I'm just not to sure about this one, the two concepts just seem to have little in common

Or `f80` or `f128` for example, they are not packable, but do have a fixed bitwidth

jorendumoulin: Starting to change my mind about this, if something has a fixedbitwidthtype it should be packable, just not necesarrily packable with that specified fixed bitwidth. I guess it thus makes sense to keep them together

compor (approved):

xdsl/dialects/builtin.py
superlopuh: This is to allow for unpacking a subset of the buffer. We don't need it when getting all the elements of a list, but it's useful in the interpreter, where we bay have a view into a larger bit of memory, and only want some known slice instead of the whole thing.

jorendumoulin (approved):

compor (approved):

PR #3583, opened 2024-12-06
dialects: (arith) add SignlessIntegerBinaryOperation canonicalization

alexarice:
Got bored of implementing various arith canonicalizations that all looked the same, so I've done a bunch of them generically in one go.

I've missed out a fair few methods (shifting operations in particular stand out) where I wasn't sure if the python operation agreed with the mlir operation.

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -198,6 +198,13 @@ class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
 
     @staticmethod
     def py_operation(lhs: int, rhs: int) -> int | None:
+        """
+        Performs a python function corresponding to this operation.
+
+        If `i := py_operation(lhs, rhs)` is an int, then this operation can be
+        canonicalized to a constant with value `i` when the inputs are constants
+        with values `lhs` and `rhs`.
+        """
         return None
 
     @staticmethod
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -167,13 +167,18 @@ class ConstantOp(IRDLOperation):
 
     @staticmethod
     def from_int_and_width(
-        value: int | IntAttr, value_type: int | IntegerType | IndexType
+        value: int | IntAttr,
+        value_type: int | IntegerType | IndexType,
+        *,
+        truncate_bits: bool = False,
     ) -> ConstantOp:
         if isinstance(value_type, int):
             value_type = IntegerType(value_type)
         return ConstantOp.create(
             result_types=[value_type],
-            properties={"value": IntegerAttr(value, value_type)},
+            properties={
+                "value": IntegerAttr(value, value_type, truncate_bits=truncate_bits)
+            },
         )
 
 
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -45,6 +45,7 @@ from xdsl.parser import Parser
 from xdsl.pattern_rewriter import RewritePattern
 from xdsl.printer import Printer
 from xdsl.traits import (
+    Commutative,
     ConditionallySpeculatable,
     ConstantLike,
     HasCanonicalizationPatternsTrait,
@@ -190,6 +191,29 @@ class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
 
     assembly_format = "$lhs `,` $rhs attr-dict `:` type($result)"
 
+    @staticmethod
+    def py_operation(lhs: int, rhs: int) -> int | None:
+        return None
+
+    @staticmethod
+    def is_right_zero(attr: AnyIntegerAttr) -> bool:
+        """
+        Returns True only when 'attr' is a right zero for the operation
+        https://en.wikipedia.org/wiki/Absorbing_element
+
+        Note that this depends on the operation and does *not* imply that
+        attr.value.data == 0
+        """
+        return False
+
+    @staticmethod
+    def is_right_unit(attr: AnyIntegerAttr) -> bool:
+        """
+        Return True only when 'attr' is a right unit/identity for the operation
+        https://en.wikipedia.org/wiki/Identity_element
+        """
+        return False
+
     def __init__(
         self,
         operand1: Operation | SSAValue,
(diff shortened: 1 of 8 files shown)


superlopuh (approved):

xdsl/dialects/arith.py
```diff
@@ -190,6 +191,18 @@ class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
...
+    def py_operation(lhs: int, rhs: int) -> int | None:
+        return None
+
+    @staticmethod
+    def is_right_zero(attr: AnyIntegerAttr) -> bool:
+        return False
+
+    @staticmethod
+    def is_right_unit(attr: AnyIntegerAttr) -> bool:
+        return False
```
math-fehr, line 204: Can you add documentation on that function?
Why is there a `right` in the name? It doesn't seem that it takes the left/right side into account?

xdsl/dialects/arith.py
```diff
@@ -395,19 +425,27 @@ def infer_overflow_type(input_type: Attribute) -> Attribute:
...
+        SignlessIntegerBinaryOperationHasCanonicalizationPatternsTrait(),
+    )
+
+    @staticmethod
+    def py_operation(lhs: int, rhs: int) -> int | None:
+        return lhs * rhs
+
+    @staticmethod
+    def is_right_unit(attr: AnyIntegerAttr) -> bool:
+        return attr == IntegerAttr(1, attr.type)
```
math-fehr, line 444: Why is this one taking the type into account, but the one in Addi just checks the value?

math-fehr (approved):
Nice!

math-fehr:
Btw, at some point we should define an interface for `is_zero` and `is_unit`.
Not the scope of this PR though

superlopuh:
`is_unit` is relative to the operation, though

xdsl/dialects/arith.py
alexarice: `i` being a right unit means that `x . i = x` but we don't necessarily have `i . x = x`. See the shift operations for example

xdsl/dialects/arith.py
alexarice: The one here is just to work around that the unit for `i1` multiplication is `-1` and not `1`. This doesn't matter in the 0 case which `Addi` has

alexarice: Thinking about it I'm not 100% sure this will work for `si1`

alexarice:
`is_right_zero` is also relative to the operation.

alexarice:
I've realised that being a zero for a function might not be common terminology for non-mathematicians. Would it be preferable to rename this to `is_right_absorbing` or `is_right_annihilating` instead?

superlopuh:
Both of those seem more confusing 😅

xdsl/dialects/arith.py
```diff
@@ -190,6 +191,29 @@ class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
...
+        Note that this depends on the operation and does *not* imply that
+        attr.value.data == 0
+        """
+        return False
+
+    @staticmethod
+    def is_right_unit(attr: AnyIntegerAttr) -> bool:
+        """
+        Return True only when 'attr' is a right unit/identity for the operation
+        https://en.wikipedia.org/wiki/Identity_element
```
superlopuh, line 213: ooh, if there are wikipedia pages with those titles then maybe using those for function names is fair game

alexarice:
This suffers from a similar problem to assert #3585 . I'll try to work out the proper overflow semantics

alexarice:
I've rebased, added an `i1` test, and truncated the result of constant propagation to fix this case and keep in line with what mlir does.

xdsl/dialects/arith.py
alexarice: @superlopuh is there a better way to do this check?

xdsl/dialects/arith.py
```diff
@@ -166,13 +167,18 @@ def __init__(
...
+        truncate_bits: bool = False,
     ) -> ConstantOp:
         if isinstance(value_type, int):
             value_type = IntegerType(value_type)
         return ConstantOp.create(
             result_types=[value_type],
-            properties={"value": IntegerAttr(value, value_type)},
+            properties={
+                "value": IntegerAttr(value, value_type, truncate_bits=truncate_bits)
+            },
```
superlopuh, line 181: how do you feel about a quick separate PR for this guy?

alexarice: #3635

alexarice:
Should be ready for review again

alexarice:
@superlopuh @compor @math-fehr I realise you've all approved but just want to check that you're all happy for this to be merged as I think a couple of things have changed since review

compor:
> @superlopuh @compor @math-fehr I realise you've all approved but just want to check that you're all happy for this to be merged as I think a couple of things have changed since review

having another look now

xdsl/dialects/arith.py
```diff
@@ -195,6 +196,29 @@ class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
 
     assembly_format = "$lhs `,` $rhs attr-dict `:` type($result)"
 
+    @staticmethod
+    def py_operation(lhs: int, rhs: int) -> int | None:
```
compor, line 200: Could you please add a simple doc string on what this does/is for? I find the name very unintuitive, although I understand it's hard (this looks like an `eval` as we have for affine maps?)

xdsl/dialects/arith.py
compor: this is what we seem to be doing in other places too

compor (approved):

xdsl/dialects/arith.py
superlopuh: I can't think of one

superlopuh (approved):

xdsl/dialects/arith.py
superlopuh: ah wait a second could this be used in the interpreter also?

alexarice: Have added a docstring. Essentially this should return an integer when we know a canonicalization can happen. Unsure if this can/should be used in the interpreter

PR #3585, opened 2024-12-06
dialects: (builtin) add IntegerAttr truncation and use in individual rewrite

alexarice:
`AdditionOfSameVariablesToMultiplyByTwo` has a corner case where the bitwidth is 1. This small change fixes this issue.


alexarice:
@superlopuh should this be dealt with by normalization?

superlopuh:
This is a classic, I believe CC @math-fehr . I think this particular case will be dealt with with normalization

superlopuh:
It would be good to have a filecheck for this

math-fehr (commented):
Question, why was the previous implementation wrong?
With an `i1`, the constant 2 should be folded to `0` right?
So `x + x` gets folded to `x * 0`, which is correct AFAIK?

alexarice:
> Question, why was the previous implementation wrong? With an `i1`, the constant 2 should be folded to `0` right? So `x + x` gets folded to `x * 0`, which is correct AFAIK?

It doesn't get converted to 0, it just adds an `i1` with `2` inside which then fails a verify check

alexarice:
@math-fehr @superlopuh any idea how to proceed here? Would it be better to always wrap integers to their range, or at least have an IntegerAttr initialiser that does this?

xdsl/transforms/individual_rewrite.py
```diff
@@ -17,10 +17,11 @@ class AdditionOfSameVariablesToMultiplyByTwo(RewritePattern):
...
-            assert isinstance(op.lhs.type, IntegerType | IndexType)
+            assert isinstance(type := op.lhs.type, IntegerType | IndexType)
+            value = 0 if isinstance(type, IntegerType) and type.bitwidth <= 1 else 2
             rewriter.replace_matched_op(
                 [
-                    li_op := arith.ConstantOp(IntegerAttr(2, op.lhs.type)),
+                    li_op := arith.ConstantOp(IntegerAttr(value, type)),
                     arith.MuliOp(op.lhs, li_op),
                 ]
             )
```
superlopuh, line 27: isn't the whole thing just 0 if the bitwidth is 1? So could branch earlier and explicitly have it as a special case

alexarice: Could do

superlopuh:
Another possible solution is to add a truncating constructor to accept any size of input

alexarice:
This is what I meant by having a wrapping IntegerAttr initialiser

math-fehr:
I would agree with the wrapping IntegerAttr init

alexarice:
I've pushed a version which adds an option to truncate an IntegerAttr. If this seems reasonable I can potentially split that into a separate PR or at least add some unit tests.

xdsl/dialects/builtin.py
```diff
@@ -397,21 +397,26 @@ def verify_value(self, value: int):
...
         """
         min_value, max_value = self.value_range()
         if not (min_value <= value.data < max_value):
-            return None
+            if not truncate_bits:
+                return None
+            value = IntAttr(value.data & (2**self.bitwidth - 1))
 
-        if self.signedness.data == Signedness.SIGNLESS:
+        if self.signedness.data != Signedness.UNSIGNED:
```
superlopuh, line 419: why this change?

superlopuh (commented):
I'm not sure that these are entirely composable things, and signedness complicates things. I'm not sure how to best do this in the general case, handling all possible signednesses and scenarios. Would it make sense to change this PR to only handle the i1 case in the pattern, and then start a separate discussion on how to truncate integers?

xdsl/dialects/builtin.py
alexarice: Doing the bitwise and seems to wrap everything into the range [0,2**bitwidth), so for the signed case we need to normalise it back to the correct range. This seemed like the easiest way to do this but it could be refactored

alexarice:
I don't really see how this isn't composable into 
1. Add optional truncation for IntegerAttr (with added tests)
2. Enable the flag on this test.

I am also not sure I understand the concerns around signedness

superlopuh (commented):
I played around with the code locally, and I get how it works a bit better now, happy to go with this direction. Could you please add unit tests for the helper, and a filechec test for the canonicalization pattern?

alexarice:
I've realised I can probably replace the `& (2**8 - 1)` with `% 2**8`, which acts exactly the same as far as I can tell and is way more readable

tests/filecheck/transforms/individual_rewrite/add-same.mlir
```diff
@@ -0,0 +1,18 @@
...
+%1 = arith.addi %0, %0 : i32
+
+// -----
+
+// CHECK:      %[[#var:]] = "test.op"() : () -> i1
+// CHECK-NEXT: %[[#zero:]] = arith.constant false
+// CHECK-NEXT: %{{.*}} = arith.muli %[[#var]], %[[#zero]] : i1
+
+%0 = "test.op"() : () -> (i1)
+%1 = arith.addi %0, %0 : i1
```
superlopuh, line 18: nit: are the regexes here really more readable than giving names to ssa values and checking for those directly?

superlopuh (approved):

tests/filecheck/transforms/individual_rewrite/add-same.mlir
alexarice: I've removed 1, but can't/haven't named the constants

superlopuh: If you name the inputs, the constants they are replaced with inherit their hints, but this is more of a nit at this point

xdsl/dialects/builtin.py
```diff
@@ -397,21 +397,26 @@ def verify_value(self, value: int):
                 f"values in the range [{min_value}, {max_value})"
             )
 
-    def normalized_value(self, value: IntAttr) -> IntAttr | None:
+    def normalized_value(
+        self, value: IntAttr, *, truncate_bits: bool = False
```
math-fehr, line 401: I think a name like `wrapping` would be clearer than `truncate_bits`.

math-fehr (approved):
Nice! Looks good to me!

xdsl/dialects/builtin.py
superlopuh: Hmm, what's it wrapping?

math-fehr: Yes, and truncating has already a meaning in LLVM/arith, which is to go from let's say i32 to i16, which is not really what we are doing here

superlopuh: But wrapping is a feature of the operation, not of a value normalization step. I'm happy for an addition to be wrapping, but it seems a bit out of context here.

superlopuh: Here's the relevant API that I have most experience with https://developer.apple.com/documentation/swift/int/init(truncating:)

tests/filecheck/transforms/individual_rewrite/add-same.mlir
alexarice: I don't want the test to be checking that the constant has the same name as the original operation, and so would prefer to use the regex

xdsl/dialects/builtin.py
alexarice: Did we come to a conclusion?

superlopuh: We can hash it out on the zulip if you want, I'm happy to be outvoted if you both think wrapping is more intuitive here

PR #3587, opened 2024-12-06
dialects: (builtin) replace DenseArrayBase's data array with bytes

superlopuh:


tests/dialects/test_builtin.py
```diff
@@ -439,7 +422,7 @@ def test_complex_init():
 
 def test_dense_as_tuple():
     floats = DenseArrayBase.from_list(f32, [3.14159, 2.71828])
```
math-fehr, line 424: So here we loose precision (because we go to 32 bits), and that's why it's like this?
Just checking if I understand

xdsl/dialects/builtin.py
```diff
@@ -1107,58 +1107,55 @@ class DenseArrayBase(ParametrizedAttribute):
...
-                        "dense array of integer element type should only contain "
-                        "integers"
-                    )
-        else:
-            for d in self.data.data:
-                if isinstance(d, IntAttr):
-                    raise VerifyException(
-                        "dense array of float element type should only contain floats"
-                    )
+        # TODO: verify length of bytes
```
math-fehr, line 1113: Is that hard to do?

math-fehr (approved):
Nice, did you see some speed ups or not at all?

xdsl/dialects/builtin.py
superlopuh: Sorry forgot to do this before pushing

tests/dialects/test_builtin.py
superlopuh: That's my understanding.

superlopuh:
I didn't run any targeted perf tests, I doubt that it would make a big impact since we rarely have large arrays, unlike large dense attrs

PR #3588, opened 2024-12-07
misc: avoid max in signed upper bound calculation

superlopuh:
This was gnawing at me.

### DIFF
--- a/xdsl/utils/comparisons.py
+++ b/xdsl/utils/comparisons.py
@@ -51,7 +51,9 @@ def signed_upper_bound(bitwidth: int) -> int:
     """
     The maximum representable value + 1.
     """
-    return 1 << max(bitwidth - 1, 0)
+    # 1 if input is 0, otherwise unchanged
+    bitwidth = bitwidth | (not bitwidth)
+    return (1 << bitwidth) >> 1
 
 
 def unsigned_value_range(bitwidth: int) -> tuple[int, int]:
--- a/xdsl/utils/comparisons.py
+++ b/xdsl/utils/comparisons.py
@@ -51,9 +51,7 @@ def signed_upper_bound(bitwidth: int) -> int:
     """
     The maximum representable value + 1.
     """
-    # 1 if input is 0, otherwise unchanged
-    bitwidth = bitwidth | (not bitwidth)
-    return (1 << bitwidth) >> 1
+    return (2 << bitwidth) >> 1
 
 
 def unsigned_value_range(bitwidth: int) -> tuple[int, int]:
--- a/xdsl/utils/comparisons.py
+++ b/xdsl/utils/comparisons.py
@@ -51,7 +51,7 @@ def signed_upper_bound(bitwidth: int) -> int:
     """
     The maximum representable value + 1.
     """
-    return (2 << bitwidth) >> 1
+    return (1 << bitwidth) >> 1 if bitwidth else 1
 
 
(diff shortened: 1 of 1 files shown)


xdsl/utils/comparisons.py
```diff
@@ -51,7 +51,9 @@ def signed_upper_bound(bitwidth: int) -> int:
     """
     The maximum representable value + 1.
     """
-    return 1 << max(bitwidth - 1, 0)
+    # 1 if input is 0, otherwise unchanged
+    bitwidth = bitwidth | (not bitwidth)
+    return (1 << bitwidth) >> 1
```
math-fehr, line 56: I actually do find this a bit less readable, is the change only to make it more readable?
Why not having an if/else at that point?

superlopuh: More to avoid the dynamic `max` call, which can be overridden. I just sort of felt like doing some bit fiddling things and overoptimising :)

math-fehr: Is it actually optimiszing? Like in a compiled language I would say yes, but here you have more operations, which roughly means that it will be slower?
I actually don't know which one is faster just from looking at it at least

superlopuh: my understanding is that the proposed version only uses Python bytecode operations that cannot be overridden on int, so it should be faster

Moxinilian:
Why not just `2**bitwidth`?

superlopuh:
```
>>> bitwidth = 0
>>> 2**bitwidth
1
>>> (1 << bitwidth) >> 1 if bitwidth else 1
1
>>> bitwidth = 1
>>> 2**bitwidth
2
>>> (1 << bitwidth) >> 1 if bitwidth else 1
1
```

Moxinilian:
How is 1 the maximum representable value of i1 plus one?

Moxinilian:
Ah, it’s signed. Maybe the docstring should be updated, it was confusing to me. I have a feeling this all would be more natural to express with the maximum value not plus one, as it would stay within its bitrange. Whatever.

xdsl/utils/comparisons.py
math-fehr: But you still have to have function calls for `|` and `not`, as Python won't know that `bitwidth` is an `int` right?
Overall, I feel that if we are doing bit tricks, we want to be sure they are worth it, instead of doing them because we think they are (because they complexify the code no?)

superlopuh: I agree overall. I guess a part of the question is whether you find the new code more or less readable than the one in main. Note that it's not the same as this conversation chain. I think it's a bit simpler to have an explicit condition for when the bitwidth is 0, because that's the thing that makes this code weird. The case for non-0 seems simpler than having a `max` function call. I expect it to be marginally faster but I agree that it's not the most important thing here.

alexarice: I would personally find `1 << (bitwidth - 1)` more readable than `(1 << bitwidth) >> 1`. I don't particular mind either way about `max` vs `if`

superlopuh: I can get behind `1 << (bitwidth - 1) if bitwidth else 1`, @math-fehr, do you think that's more readable than `1 << max(bitwidth - 1, 0)`?

superlopuh: I just don't like the max TBH

superlopuh:
I initially thought this would have been met with enthusiasm, but if not I'm happy to repoen this only once it shows up on a trace

PR #3590, opened 2024-12-07
CI: set maxIndividualTestTime in lit cfg

superlopuh:
I copied the approach in https://github.com/llvm/llvm-project/blob/69227a11fe1d00b2fc6f93b2a985abbe888d0876/bolt/test/lit.cfg.py#L55


math-fehr (approved):

superlopuh:
If we need more time then let's add more time, we do want a timeout though because some things have taken infinity time in the past.

superlopuh:
BTW I think we can safely remove the timeout in the makefile, I forgot to do that in this PR

compor:
yeah, I know, it's inconsistently used in the current rules. The cmd line option overrides the lit configuration, so I'm happy to use it as such.

I'll fix it in a wee PR.

PR #3591, opened 2024-12-08
core: move logic specific to the MLIR grammar from lexer to mlir_lexer

superlopuh:
This decouples the base lexing infrastructure from MLIR's syntax allowing us to to reuse it for other grammars, such as Toy, WASM, Stim, and other downstream projects.


compor (approved):

math-fehr (approved):
Can you please rewrite Kind to TokenKind

PR #3595, opened 2024-12-08
core: split out non-MLIR-specific parser functionality to GenericParser

superlopuh:
Lets us use the generic Lexer infrastructure for other Parsers.


math-fehr (approved):
Nice

compor (approved):
I saw this part of the PR stack this morning, but I forgot to finalize the review.

It might be useful to "track the PR stack" as @math-fehr did in #3538 :)

superlopuh:
Mathieu used a tool for that, which seems to do a part of the work for you, I'm happy to try it out when I next feel like building a tower :)

PR #3599, opened 2024-12-09
dialects: (builtin) add CompileTimeFixedBitwidthType

jorendumoulin:
This PR adds the `CompileTimeFixedBitwidthType` to represent values that have a fixed bitwidth and are thus packable, but the bitwidth may not be known until compile time.

The new dependency graph now looks like the following: https://is.gd/q5N7Dx

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -556,7 +556,7 @@ class LocationAttr(ParametrizedAttribute):
 
 
 @irdl_attr_definition
-class IndexType(ParametrizedAttribute):
+class IndexType(ParametrizedAttribute, StructPackableType[int]):
     name = "index"
 
     def print_value_without_type(self, value: int, printer: Printer):
@@ -565,6 +565,13 @@ class IndexType(ParametrizedAttribute):
         """
         printer.print_string(f"{value}")
 
+    @property
+    def format(self) -> str:
+        return "<q"
+
+    @property
+    def bitwidth(self) -> int:
+        raise RuntimeError("IndexType bitwidth not known at compile-time")
 
 IndexTypeConstr = BaseAttr(IndexType)
 
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -349,7 +349,15 @@ class SignednessAttr(Data[Signedness]):
                 raise ValueError(f"Invalid signedness {data}")
 
 
-class FixedBitwidthType(TypeAttribute, ABC):
+class CompileTimeFixedBitwidthType(TypeAttribute, ABC):
+    """
+    A type attribute whose runtime bitwidth is fixed, but may be target-dependent.
+    """
+
+    name = "abstract.compile_time_fixed_bitwidth_type"
+
+
+class FixedBitwidthType(CompileTimeFixedBitwidthType, ABC):
     """
     A type attribute whose runtime bitwidth is target-independent.
     """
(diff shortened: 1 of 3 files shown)


alexarice (approved):

superlopuh:
IndexType is not FixedBitwidth, since its bitwidth is target-dependent

superlopuh:
We need an adjustment to the hierarchy, where we need to distinguish between packability, and bitwidth fixed at compile-time vs runtime

superlopuh (changes requested):
Probably worth checking with @math-fehr on this also, but my understanding is that we're missing a type that says that the bitwidth is fixed during compilation time but it may differ from the bitwidth at runtime until lowered to target-dependent code.

superlopuh:
I propose going from left to right here: https://dreampuf.github.io/GraphvizOnline/?engine=dot#digraph%20G%20%7B%0A%0A%20%20FixedBitwidthType%20-%3E%20PackableType%3B%0A%20%20PackableType%20-%3E%20StructPackable%3B%0A%20%20StructPackable%20-%3E%20IntegerType%3B%0A%20%20StructPackable%20-%3E%20FPType%3B%0A%20%20%0A%20%20_CompileTimeFixedBitwidthType%20-%3E%20_FixedBitwidthType%3B%0A%20%20_CompileTimeFixedBitwidthType%20-%3E%20_PackableType%3B%0A%20%20%0A%20%20_PackableType%20-%3E%20_StructPackable%3B%0A%20%20_StructPackable%20-%3E%20_IntegerType%3B%0A%20%20_StructPackable%20-%3E%20_FPType%3B%0A%20%20_FixedBitwidthType%20-%3E%20_IntegerType%3B%0A%20%20_FixedBitwidthType%20-%3E%20_FPType%3B%0A%20%20_StructPackable%20%20-%3E%20_IndexType%3B%0A%7D

jorendumoulin:
> I propose going from left to right here: ...

Looks good to me, but what is the practical use of the `_CompileTimeFixedBitwidthType` class? The world seems a bit simpler by just separating `Packable` and `FixedBitwidth`, as I personally tend to steer away from boilerplate code without a clear usage. I am fine either way though.

superlopuh:
Because we need to distinguish between things that can go in DenseArrayBase and DenseIntOrFPElementsAttr, and in general for bitwidth computation of things that reside in memory on the target

jorendumoulin:
I updated things to your suggestions, let me know what you think!

xdsl/dialects/builtin.py
```diff
@@ -349,7 +349,15 @@ def print_parameter(self, printer: Printer) -> None:
...
-class FixedBitwidthType(TypeAttribute, ABC):
+class CompileTimeFixedBitwidthType(TypeAttribute, ABC):
+    """
+    A type attribute whose runtime bitwidth is fixed, but may be target-dependent.
+    """
+
+    name = "abstract.compile_time_fixed_bitwidth_type"
+
+
+class FixedBitwidthType(CompileTimeFixedBitwidthType, ABC):
```
superlopuh, line 360: How do you feel about putting an abstract property here?

Maybe actually not packed-dependent name, like `compile_time_size`?

```suggestion
class CompileTimeFixedBitwidthType(TypeAttribute, ABC):
    """
    A type attribute whose runtime bitwidth is fixed, but may be target-dependent.
    """

    name = "abstract.compile_time_fixed_bitwidth_type"

    @property
    @abstractmethod
    def packed_size(self) -> int:
        """
        Contiguous memory footprint of packed value in bytes.
        """
        raise NotImplementedError()

class FixedBitwidthType(CompileTimeFixedBitwidthType, ABC):
```

math-fehr:
Yeah, you should always be able to pack an `index` into an `i64`, but you don't know its size at compile-time (only when you do target-dependent lowerings)

xdsl/dialects/builtin.py
jorendumoulin: yep, this makes sense!

PR #3600, opened 2024-12-09
core: split print_float into print_float and print_float_attr

jorendumoulin:
To allow for printing floats that are not `FloatAttrs` but the type is known (such as for bytes-backed `DenseIntOrFPElementsAttrs`), this changes the `print_float` implementation to accept raw floats with a known type.


tobiasgrosser:
Is there a way we can add a test case to this PR?

alexarice (approved):
I think there are some other places where we print floats manually but did know the type, however it'd be fine to catch these later.

jorendumoulin:
> Is there a way we can add a test case to this PR?

I have added some tests!

superlopuh (approved):

PR #3602, opened 2024-12-09
transformations: (apply-individual-rewrite) give more context in error message [NFC]

superlopuh:
Helps debug

### DIFF
--- a/xdsl/transforms/individual_rewrite.py
+++ b/xdsl/transforms/individual_rewrite.py
@@ -124,4 +124,6 @@ class ApplyIndividualRewritePass(ModulePass):
 
         pattern.match_and_rewrite(matched_operation, rewriter)
         if not rewriter.has_done_action:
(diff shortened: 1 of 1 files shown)


alexarice (approved):

xdsl/transforms/individual_rewrite.py
```diff
@@ -124,4 +124,6 @@ def apply(self, ctx: MLContext, op: ModuleOp) -> None:
 
         pattern.match_and_rewrite(matched_operation, rewriter)
         if not rewriter.has_done_action:
-            raise ValueError("Invalid rewrite at current location.")
+            raise ValueError(
+                f"Invalid rewrite ({pattern}) for operation ({matched_operation}) at location {self.matched_operation_index}."
```
alexarice, line 128: I was too quick to approve, this should probably use `self.pattern_name`

PR #3604, opened 2024-12-09
core (traits): Fix implicit terminator when last op is unregistered

### LINKED ISSUE #3325: scf.for implicit terminators
The implicit terminator infrastructure for `scf.for` does not seem to work correctly at the moment. See #3303

compor:
This PR:

- Fixes the implicit addition of a terminator when the last operation is an (allowed) unregisted operation
- Adds a filecheck test for the above

Fixes #3325


tests/filecheck/dialects/scf/unregistered.mlir
```diff
@@ -1,22 +1,24 @@
...
-// CHECK-NEXT:   func.return
-// CHECK-NEXT: }
 
-func.func @for_unregistered() {
-  %lb = arith.constant 0 : index
-  %ub = arith.constant 42 : index
-  %s = arith.constant 3 : index
-  scf.for %iv = %lb to %ub step %s {
+%0 = arith.constant 0 : index
+scf.for %iv = %0 to %0 step %0 {
```
compor, line 9: Simplified the already existing test case to be inline with the new one added below.

alexarice (approved):
Seems good

PR #3606, opened 2024-12-09
interactive: do not auto_expand passes_list

alexarice:
I was really struggling to debug this, but from what I can understand the `child_pass_pipeline` was ending up with duplicate passes because `self.pass_pipeline` shouldn't have been added. 
I'm not 100% this is the correct fix as I didn't manage to work out what everything did and why.

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -537,3 +537,85 @@ async def test_dark_mode():
         await pilot.press("d")
 
         assert app.theme == "textual-dark"
+
+
+@pytest.mark.asyncio
+async def test_apply_individual_rewrite():
+    """Tests that using the tree to apply an individual rewrite works"""
+
+    async with InputApp(tuple(get_all_dialects().items()), ()).run_test() as pilot:
+        app = cast(InputApp, pilot.app)
+        # clear preloaded code and unselect preselected pass
+        app.input_text_area.clear()
+
+        await pilot.pause()
+        # Testing a pass
+        app.input_text_area.insert(
+            """
+        func.func @hello(%n : i32) -> i32 {
+  %c0 = arith.constant 0 : i32
+  %res = arith.addi %c0, %n : i32
+  func.return %res : i32
+}
+        """
+        )
+        app.passes_tree.root.expand()
+        await pilot.pause()
+
+        node = None
+        for n in app.passes_tree.root.children:
+            if (
+                n.data is not None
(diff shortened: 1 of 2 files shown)


alexarice:
I'm now pretty sure this change is wrong, I'm going to try to properly figure out what's going on

compor:
@alexarice how do you trigger the crash?

alexarice:
1. run `xdsl-gui`
2. Click dropdown in pass list
3. scroll to bottom
4. click the individual pass

alexarice:
I think I've figured it out. In `expand_tree_node` we try to get the next passes that can be applied, to do this we need the pass list to get to this node of the tree which is gotten by appending:
- `self.pass_pipeline`: The selected passes
- `root_to_child_pass_list`: Passes from the root of the tree (exclusive) to the node to be expanded (exclusive)
- The expanded node
I believe the issue is that when the `root_to_child_pass_list` is empty, the "expanded node" is already in the pass pipeline and gets duplicated

compor:
Very plausible, and that point caught my attention as suspicious.
It would be good to have someway to test this, but I can't see how to do so easily without refactoring this bit.

alexarice:
Completely different fix: the passes_list automatically expanded nodes that were selected. As selection is overriden to reset the tree I think this was breaking stuff.

alexarice:
I still haven't quite got my head around the tree constructions above, but with `auto_expand = False` all the problems go away. I have no idea how to test this (I guess a test that "clicks on" some individual rewrites might be a good idea)

compor:
yeah, where "clicks on" would be somehow disentangled from the UI functionality in a typical MVC, so we could actually test it?

I'll be OK accepting this if you open an issue tracking the suggested improvement, so we can move on.

compor (approved):

compor:
> (I guess a test that "clicks on" some individual rewrites might be a good idea)

superlopuh:
So you're suggesting skipping adding a unit test?

alexarice:
I'm doing the unit test now

compor:
@alexarice I wasn't sure by your phrasing how laborious that was; if there is a unit test, I'll take it.

alexarice:
I've added a test that selects some individual rewrites. It's not the cleanest but it crashes without this change and works with it

compor (approved):
laborious, thanks

superlopuh (approved):
Amazing, thank you

PR #3607, opened 2024-12-09
dialects: (builtin) mimic mlir floating point precision for printing and parsing

jorendumoulin:
I'm running into some precision issues when printing and parsing floats after they have been packed/unpacked.
This PR tries to resolve the issues with printing and parsing, trying to mimic MLIR behaviour as much as possible.


xdsl/printer.py
```diff
@@ -481,11 +481,24 @@ def print_float(self, value: float, type: AnyFloat):
                 )
         else:
             # to mirror mlir-opt, attempt to print scientific notation iff the value parses losslessly
-            float_str = f"{value:.6e}"
-            if float(float_str) == value:
+            float_str = f"{value:.5e}"
+            index = float_str.find("e")
+            float_str = float_str[:index] + "0" + float_str[index:]
+
```
jorendumoulin, line 487: mlir does this interesting thing where it rounds to 5 digits after the comma but still prints the 6th 0

xdsl/printer.py
jorendumoulin, line 1: this could hurt performance for a large amount of floats, which can possibly be resolved by just dumping the hex value of the bytes representation as it is done for arrays > ~100ish elements in MLIR:

```
module {
  %cst = arith.constant dense<"0xC3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43C3F508409EEF0842F5DBC141E73B6A43C76B6A43"> : tensor<180xf32>
}
```

jorendumoulin:
Before I start fixing the all the filechecks that are affected by this, could I get your opinion on whether this should be implemented like this? @superlopuh @alexarice @compor @math-fehr

alexarice:
My understanding is that the current float parsing/printing was changed recently so that it always roundtripped correctly. Was there a bug in this implementation? I'm not 100% it's worth changing how all the floats are represented. (As a side note I'm not exactly sure if it's even desirable to use scientific notation everywhere for floats, but it's possible I'm missing something.)

jorendumoulin:
> My understanding is that the current float parsing/printing was changed recently so that it always roundtripped correctly. Was there a bug in this implementation? I'm not 100% it's worth changing how all the floats are represented. (As a side note I'm not exactly sure if it's even desirable to use scientific notation everywhere for floats, but it's possible I'm missing something.)

Not really a bug, just diverging between behaviour between mlir/xdsl:

for an input

```
arith.constant 1.1 : f16
arith.constant 3.1415 : f32
arith.constant 3.141592 : f32
arith.constant 3.1415 : f64
arith.constant 3.141592 : f64
```

xdsl right now returns:

```
builtin.module {
  %0 = arith.constant 1.100000e+00 : f16
  %1 = arith.constant 3.141500e+00 : f32
  %2 = arith.constant 3.141592e+00 : f32
  %3 = arith.constant 3.141500e+00 : f64
  %4 = arith.constant 3.141592e+00 : f64
}
```

and mlir (and xdsl with this PR):
```
builtin.module {
  %0 = arith.constant 1.099610e+00 : f16
  %1 = arith.constant 3.141500e+00 : f32
  %2 = arith.constant 3.14159203 : f32
  %3 = arith.constant 3.141500e+00 : f64
  %4 = arith.constant 3.1415920000000002 : f64
}
```

The main problem with the version xDSL i have right now is when things are backed with bytes storage vs as FloatAttr, for example for `2.1: f32`:
```
builtin.module {
  %0 = arith.constant {"test" = array<f32: 2.0999999046325684>} 0 : i64
  %1 = arith.constant 2.100000e+00 : f32
}
```

compor:
I don't have any argument to shoot this down ATM, and I'm in general in favor of such convergence.

superlopuh:
Yep this seems like a great change

xdsl/dialects/builtin.py
```diff
@@ -825,6 +821,17 @@ def __init__(
...
             else:
                 raise ValueError(f"Invalid bitwidth: {type}")
+
+        value: float = data.data if isinstance(data, FloatData) else data
+        try:
+            # constrain value to precision of floating point type
+            value = type.unpack(type.pack([value]), 1)[0]
+        except NotImplementedError:
+            # allow full precision
+            pass
```
superlopuh, line 831: This is a little scary, what do you think about instead manually check for the supported values here?

xdsl/dialects/builtin.py
```diff
@@ -825,6 +821,15 @@ def __init__(
                 type = Float128Type()
             else:
                 raise ValueError(f"Invalid bitwidth: {type}")
+
+        value: float = data.data if isinstance(data, FloatData) else data
+        # for supported types, constrain value to precision of floating point type
+        # else, allow full python float precision
+        if isinstance(type, Float64Type | Float32Type | Float16Type):
+            value = type.unpack(type.pack([value]), 1)[0]
```
superlopuh, line 829: supernit:
```suggestion
            value = type.unpack(type.pack((value,)), 1)[0]
```

superlopuh (approved):
Would be good to double-check with @math-fehr

math-fehr (approved):
Looks good to me!

PR #3608, opened 2024-12-09
misc (Makefile): Group `lit` options and allow them to be overriden during `make` invocation

compor:
This PR:

- Groups `lit` options as a Makefile variable that can be overridden during `make` invocation,
  e.g., `make filecheck LIT_OPTIONS="-v --order=smart --timeout=30`
- The default timeout value is 20secs and it is currently provided in `lit.cfg` since #3590
- the `-vv` option is deprecated, `-v` does what we want.

Based on the discussion [here](https://github.com/xdslproject/xdsl/pull/3590#issuecomment-2527765677).


math-fehr (approved):
Cool!

PR #3610, opened 2024-12-09
misc: add ShrinkException for easy test case shrinking

superlopuh:
I think this might come in handy in the future.

I opened a feature request to make it more ergonomic in the future: https://github.com/DRMacIver/shrinkray/issues/11


alexarice (commented):
It's not really clear to me what this change does without an example. Are the printed strings magic keywords that shrinkray understands? Does it just use the program code output?

compor (commented):
Is this eager optimization (or debugging)?

It seems interesting, but I have the same questions as @alexarice.

Does it work like `llvm-reduce` or `bugpoint` but on the Python exception stack level?
Did it help you narrow down the above? Somehow, I've never had an issue with Python's exception unwinding.

superlopuh:
I'll add more documentation, here's an example, it shrunk the test case in the PR it sat on top of to the file in the PR:
https://github.com/xdslproject/xdsl/pull/3576

compor:
Thanks, I'll play with that and come back.

webmiche:
What are you planning to use this for? To me, this is a tool that is primarily useful for bug reporting after finding a bug through some automated way (i.e., some form of fuzzing). I can see that supporting something to reduce testcases directly within xdsl might be useful for some people, but I am unsure whether this quite specific use case actually justifies putting this into xdsl-opt. Can't people just use shrinkray on their own and then report the bug triggering test case back to us? Does this really have to be deeply integrated into xdsl to be useable?

superlopuh:
So it's more of a convenience for when someone wants to use it. The shrinkray interface is you feed it a sccript that exits with 0 if the condition was met, and non-0 if it was not met. This is not necessarily easy to orchestrate for reducing a test-case, and I had to do a bit of leg work to make it work locally. With this change in, the effort required to create a script that shrinks correctly is minimised.

webmiche:
I see. Does this make shrinkray a dependency? Or do only users that want to use this feature need it installed?

I do not have a strong opinion on this. To me, it feels like a bit of an overfit for a specific use case, but if you believe this to be an important/nice use case, I am fine with it.

superlopuh:
I just think it's magic, and I'd love to be able to perform magic on other people's computers when they ask me how to reduce a test case :)

xdsl/utils/exceptions.py
```diff
@@ -48,6 +48,23 @@ class InvalidIRException(Exception):
...
+
+    When using DRMacIver's [Shrink Ray](https://github.com/DRMacIver/shrinkray) library,
+    this exception can be used to reduce test cases. For example, to find a smaller
+    version of a case that has some behavior you are interested in, raise this exception
+    on the line of code you want to hit, and pass the `--shrink` argument to `xdsl-opt`.
+
+    To shrink a test case that raises a ShrinkException when called like this:
+    `xdsl-opt input_file.mlir -p my,pass,pipeline`, it needs to be changed to:
+    `shrinkray "xdsl-opt -p my,pass,pipeline --shrink" input_file.mlir`.
+    """
```
compor, line 63:
```suggestion
    Exception for test case reduction when used in conjunction with the [Shrink Ray](https://github.com/DRMacIver/shrinkray) reducer.
    
    To find a reduced version of a test case, raise this exception on the line of code you want to hit,
    and pass the `--shrink` argument to `xdsl-opt`, by changing its invocation from:
    `xdsl-opt input_file.mlir -p my,pass,pipeline`
    to:
    `shrinkray "xdsl-opt -p my,pass,pipeline --shrink" input_file.mlir`.
    """
```

compor (approved):

PR #3613, opened 2024-12-10
API: split out MLIR-independent printing into BasePrinter and use in STIM [NFC]

superlopuh:
I'd also like to use the indentation logic in ASL, WGSL, and the assembly dialects, PRs incoming.


xdsl/dialects/stim/stim_printer_parser.py
```diff
@@ -1,25 +1,20 @@
...
+from xdsl.ir import Attribute, Operation
+from xdsl.utils.base_printer import BasePrinter
 
-    def print_string(self, text: str) -> None:
-        print(text, end="", file=self.stream)
 
+@dataclass(eq=False, repr=False)
+class StimPrinter(BasePrinter):
     @contextmanager
     def in_braces(self):
```
alexarice, line 19: Possibly for a future PR, but I wonder if context managers for each delimiter pair should be added to `BasePrinter`. It's always annoyed me that there is `in_angle_brackets` (at least in the parser) but not the equivalents for braces, square brackets, and parentheses.

alexarice (approved):
Seems like a great change to me

xdsl/dialects/stim/stim_printer_parser.py
superlopuh: Hundo p, just trying to make the API changes small, so we can have a focused discussion about the API in a separate PR

xdsl/utils/base_printer.py
```diff
@@ -0,0 +1,134 @@
+from __future__ import annotations
+
+from collections.abc import Callable, Iterable
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from typing import IO, Any, TypeVar
+
+indentNumSpaces = 2
```
math-fehr, line 8: O think this should stay in the printer no? Or as an option?

math-fehr (approved):
Looks good to me'

PR #3617, opened 2024-12-10
misc (Makefile): Update coverage handling in Makefile

compor:
This PR:

- Moves coverage options from the `Makefile` to `.coveragerc` and adds the [`ignore_errors`](https://coverage.readthedocs.io/en/latest/config.html#report-ignore-errors) option
- Adds a `coverage-clean` target and lets the coverage module to erase the produced files from a coverage run
- Removes the `TESTS_COVERAGE_FILE` env variable as this is only used in a target to avoid name collision of coverage files; for all purposes using the common prefix as enforced by `COVERAGE_FILE` is enough to make things work uniformly across targets

I would have loved to [move](https://coverage.readthedocs.io/en/latest/config.html#sample-file) the `.coveragerc` configuration to `pyproject.toml`, but it seems that `pytest-cov` does not support it yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,6 @@ VENV_DIR ?= venv
 # use a default prefix for coverage data files
 COVERAGE_FILE ?= .coverage
 
-# use different coverage data file per coverage run, otherwise combine complains
-TESTS_COVERAGE_FILE = ${COVERAGE_FILE}.tests
-
 # default lit options
 LIT_OPTIONS ?= -v --order=smart
 
@@ -19,8 +16,9 @@ LIT_OPTIONS ?= -v --order=smart
 # these targets don't produce files:
 .PHONY: ${VENV_DIR}/ venv clean clean-caches filecheck pytest pytest-nb tests-toy tests
 .PHONY: rerun-notebooks precommit-install precommit pyright tests-marimo
-.PHONY: coverage coverage-tests coverage-filecheck-tests
-.PHONY: coverage-report-html coverage-report-md
+.PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report coverage-clean
+.PHONY: pytest-toy-nb
+.PHONY: tests-marimo-onnx
 
 # set up the venv with all dependencies for development
 ${VENV_DIR}/: requirements.txt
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,8 +16,9 @@ LIT_OPTIONS ?= -v --order=smart
 # these targets don't produce files:
 .PHONY: ${VENV_DIR}/ venv clean clean-caches filecheck pytest pytest-nb tests-toy tests
 .PHONY: rerun-notebooks precommit-install precommit pyright tests-marimo
-.PHONY: coverage coverage-tests coverage-filecheck-tests
-.PHONY: coverage-report-html coverage-report-md
+.PHONY: coverage coverage-tests coverage-filecheck-tests coverage-report-html coverage-report coverage-clean
+.PHONY: pytest-toy-nb
+.PHONY: tests-marimo-onnx
```
compor, line 21: This is piggybacked, just to gather all phony targets in the same place

Makefile
```diff
@@ -129,8 +124,8 @@ precommit:
 	pre-commit run --all
 
 # run pyright on all files in the current git commit
+# make sure to generate the python typing stubs before running pyright
```
compor, line 127: another minor cleanup moving the comment out

Makefile
```diff
@@ -150,6 +146,9 @@ coverage-filecheck-tests:
 coverage-report-html:
 	coverage html
 
-# generate markdown coverage report
-coverage-report-md:
-	coverage report --format=markdown
+# generate coverage report
+coverage-report:
```
compor, line 150: This is now by default exporting markdown as per `.coveragerc` settings

superlopuh (approved):

superlopuh:
does that mean that coverage actually works for you?

compor:
> does that mean that coverage actually works for you?

It does, doesn't it for you?

My main issue was the if you specified a difference `COVERAGE_FILE` some rules didn't use it, specifically the cleanup.

superlopuh:
I haven't really been able to leverage it locally, and the online version is broken for xDSL. I'd love to fix this, as I think it's actually quite useful, but it's never high enough in my priority list.

compor:
> I haven't really been able to leverage it locally, and the online version is broken for xDSL. I'd love to fix this, as I think it's actually quite useful, but it's never high enough in my priority list.

Hmm, I use it a lot locally, and helps me a lot with testing, so I'd love to fix it.
Maybe we can chat on what doesn't work for you.
I'm not sure what you mean by the online version (CI?).
If so, maybe it can be 2 birds with 1 stone kind of thing.

superlopuh:
I mean the comment above, from codecov, has been broken for something like a year

PR #3517, opened 2024-11-25
core: add functional-type directive

alexarice:
Adds the functional-type directive, which takes two typeable directives as input.

The key difficulty here was dealing with the results, which are usually surrounded by parentheses but don't have to be if there is exactly one result.

Still a draft as this needs more tests, also stacked on #3516.

Issue #3497, opened 2024-11-21
dialects: (func) "func.call" does not verify function signature


watermelonwolverine:
"func.call" does not verify the argument types and result types against function signature.

PR #3618, opened 2024-12-10
dialects (func): Use assembly format in `func.call` operation

compor:
This PR:

- uses the declarative assembly format for `func.call` made possible by `functional-type` #3517

### DIFF
--- a/xdsl/dialects/func.py
+++ b/xdsl/dialects/func.py
@@ -11,9 +11,7 @@ from xdsl.dialects.builtin import (
     SymbolRefAttr,
 )
 from xdsl.dialects.utils import (
-    parse_call_op_like,
     parse_func_op_like,
-    print_call_op_like,
     print_func_op_like,
 )
 from xdsl.ir import (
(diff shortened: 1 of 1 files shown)


superlopuh (approved):
😍

xdsl/dialects/func.py
```diff
@@ -11,9 +11,7 @@
     SymbolRefAttr,
 )
 from xdsl.dialects.utils import (
-    parse_call_op_like,
```
superlopuh, line 14: Let's delete these helpers also?

xdsl/dialects/func.py
```diff
@@ -274,6 +272,10 @@ class CallOp(IRDLOperation):
     callee = prop_def(FlatSymbolRefAttrConstr)
     res = var_result_def()
 
+    assembly_format = (
+        "$callee `(` $arguments `)` attr-dict `:` functional-type($arguments, $res)"
+    )
+
     # TODO how do we verify that the types are correct?
```
superlopuh, line 279:
```suggestion
```

compor: My understanding was that this comment is essential #3497, no?
I was planning to tackle that next, but not sure when.

xdsl/dialects/func.py
compor: ehm, I have?

xdsl/dialects/func.py
superlopuh: ooh then yes

xdsl/dialects/func.py
superlopuh: I don't see a diff in `xdsl.dialects.utils`

compor: they are still used in `riscv_func`, so maybe after this :)

PR #3621, opened 2024-12-11
misc: add a DisjointSet data structure

superlopuh:
This is useful in a few places, notably in bufferization.

### DIFF
--- a/xdsl/utils/disjoint_set.py
+++ b/xdsl/utils/disjoint_set.py
@@ -97,6 +97,9 @@ class IntDisjointSet:
         # Note: We don't need to update _count[new_child] since it's no longer a root
         return True
 
+    def are_connected(self, lhs: int, rhs: int) -> bool:
+        return self.find(lhs) == self.find(rhs)
+
 
 _T = TypeVar("_T", bound=Hashable)
 
@@ -138,10 +141,8 @@ class DisjointSet(Generic[_T]):
         """
         Find the representative value for the set containing the given value.
 
-        Args:
-            value: The value to find the set representative for
-        Returns:
-            The representative value for the set
+        Returns the representative value for the set.
+
         Raises:
             KeyError: If the value is not in the disjoint set
         """
@@ -152,12 +153,20 @@ class DisjointSet(Generic[_T]):
         """
         Merge the sets containing the two given values if they are different.
 
-        Args:
-            lhs: First value
-            rhs: Second value
-        Returns:
-            True if the sets were merged, False if they were already the same set
+        Returns `True` if the sets were merged, `False` if they were already the same set.
+
         Raises:
             KeyError: If either value is not in the disjoint set
         """
         return self._base.union(self._index_by_value[lhs], self._index_by_value[rhs])
+
+    def are_connected(self, lhs: _T, rhs: _T) -> bool:
+        """
+        Returns `True` if the values are in the same set.
+
+        Raises:
+            KeyError: If either value is not in the disjoint set
+        """
+        return self._base.are_connected(
+            self._index_by_value[lhs], self._index_by_value[rhs]
+        )
--- a/xdsl/utils/disjoint_set.py
+++ b/xdsl/utils/disjoint_set.py
@@ -40,13 +40,6 @@ class IntDisjointSet:
         """Number of nodes in this structure."""
         return len(self._parent)
 
-    def set_count(self) -> int:
-        """
-        Number of sets in this structure.
-        Note: This is O(n) as it needs to scan all parents.
-        """
-        return len(set(self._parent))
-
     def add(self) -> int:
         """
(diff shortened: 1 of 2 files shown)


compor (approved):
All good, but a few thoughts:

- Whenever I see typical DS or algorithms like this, it makes me think why not use an external library?
I know we want to keep dependencies slim, but we have other things that are way less stable.
At the end of the day we are a compiler framework, and that's part of the territory.
Using a robust framework like scipy or networkx, might save us the time spent in undergrad course territory.

Is this also to be used in the EqSat stuff?

Some maybe more actionable suggestions:
1) Add a `connected` method that takes 2 or more elements and checks if they are in the same union set? I'm a fan of Robert Sedgewick's API.

2) Maybe it's time to organize `utils` a bit and add an `algo` and/or `ds` (or `adt`) subdir?

xdsl/utils/disjoint_set.py
```diff
@@ -0,0 +1,170 @@
...
+    def value_count(self) -> int:
+        """Number of nodes in this structure."""
+        return len(self._parent)
+
+    def set_count(self) -> int:
+        """
+        Number of sets in this structure.
+        Note: This is O(n) as it needs to scan all parents.
+        """
+        return len(set(self._parent))
```
alexarice, line 48: Parent pointers are not eagerly updated so I don't think this is correct? If this operation is needed then it's likely better to keep track of the number of sets in a variable

xdsl/utils/disjoint_set.py
```diff
@@ -0,0 +1,170 @@
...
+Generic implementation of a disjoint set data structure.
+
+https://en.wikipedia.org/wiki/Disjoint-set_data_structure
+"""
+
+from collections.abc import Hashable, Sequence
+from typing import Generic, TypeVar
+
+
+class IntDisjointSet:
```
alexarice, line 11: Is the IntDisjointSet independently useful? Otherwise this feels like an unnecessary step to making the structure below.

alexarice (commented):
I agree with @compor. It's not clear to me why basic data structures should be reimplemented

xdsl/utils/disjoint_set.py
superlopuh: Good point I'll remove it

superlopuh:
Chris:
1. I'll add it. I'm not familiar with Sedgwick's API, could you please send a link to it?
2. sounds good but should probably be its own PR

To everyone who would prefer to add a dependency please let me know where I can find one with this generic API for hashable things :)

alexarice:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.DisjointSet.html ?

compor:
> Chris:
> 
>     1. I'll add it. I'm not familiar with Sedgwick's API, could you please send a link to it?
> 
>     2. sounds good but should probably be its own PR
> 
> 
> To everyone who would prefer to add a dependency please let me know where I can find one with this generic API for hashable things :)

https://sedgewick.io/wp-content/uploads/2022/04/Algs01-UnionFind.pdf

He uses Java, but that's besides the point. It's a long-running Algo+DS course.

xdsl/utils/disjoint_set.py
compor: Yeah, good point. I'm not sure why the generality here.
Also, if this is for the equality saturation tasks, I'm not familiar how it is used there; maybe future uses warrant this?

superlopuh:
Scipy is a huge dependency to impose on all clients of xDSL

superlopuh:
Not to mention that it's not Pure python

superlopuh:
OTOH I'm happy to mirror scipy's API since users might already be familiar

compor:
granted, scipy is a big dep and most significantly, not pure Python. AFAIK, networkx is pure python?

compor:
Maybe that's a discussion to have at the next meeting or the one after.

xdsl/utils/disjoint_set.py
superlopuh: I feel like it's a nice separation if only for testing purposes, it neatly encapsulates the core logic, and adds negligeable overhead. I expect that no-one's going to use it, but I also see no harm in the separation.

For equality saturation, my understanding is that a persistent version of this tends to be used, and I also expect that a separation between the core logic and the hashable/generic helpers would be useful both for testing and readability.

compor:
https://networkx.org/documentation/latest/_modules/networkx/utils/union_find.html#UnionFind.union

> Based on the meeting:
> 
>     * We don't want to use scipy as maintaining pure python is a priority
> 
>     * We would potentially be up for using an external library providing a pure python implementation, though have not found one for this (@compor claims networkx has one)
> 
>     * We should just use this PR for now/ever

PR #3622, opened 2024-12-11
dialects (riscv_func): Use assembly format for `riscv_func.call` operation

compor:
This PR:

- uses the declarative assembly format for `riscv_func.call` made possible by `functional-type` #3517 
- removes unused utility methods for generic printing/parsing of the above since there are no users for them as a result of the above

Also discussed in #3618

### DIFF
--- a/xdsl/dialects/riscv_func.py
+++ b/xdsl/dialects/riscv_func.py
@@ -80,7 +80,9 @@ class CallOp(riscv.RISCVInstruction):
     callee = attr_def(SymbolRefAttr)
     ress = var_result_def(riscv.RISCVRegisterType)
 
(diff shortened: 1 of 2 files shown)


alexarice (approved):
:+1:

xdsl/dialects/riscv_func.py
```diff
@@ -83,6 +80,8 @@ class CallOp(riscv.RISCVInstruction):
     callee = attr_def(SymbolRefAttr)
     ress = var_result_def(riscv.RISCVRegisterType)
 
+    assembly_format = "$callee `(` $args`)` attr-dict `:` functional-type($args, $ress)"
```
alexarice, line 83:
```suggestion
    assembly_format = "$callee `(` $args `)` attr-dict `:` functional-type($args, $ress)"
```
?

PR #3623, opened 2024-12-11
dialects: (builtin) change data representation of DenseIntOrFPElements to use bytes

jorendumoulin:


xdsl/dialects/builtin.py
```diff
@@ -1939,51 +1939,64 @@ def shape_is_complete(self) -> bool:
             n *= dim
 
         # Product of dimensions needs to equal length
-        return n == len(self.data.data)
+        return n == len(self.get_values())
```
superlopuh, line 1942: Would this work?
```suggestion
        return n == len(self.data.data) // self.type.packed_size
```

superlopuh: Otherwise O(n)...

docs/Toy/examples/scalar.toy
```diff
@@ -6,8 +6,8 @@ def main() {
 }
 
 # CHECK:         "toy.func"() ({
-# CHECK-NEXT:      %0 = "toy.constant"() {"value" = dense<5.500000e+00> : tensor<f64>} : () -> tensor<f64>
-# CHECK-NEXT:      %1 = "toy.reshape"(%0) : (tensor<f64>) -> tensor<2x2xf64>
```
superlopuh, line 10: this is sus, we can't reshape from 1xf64 to 2x2xf64, I must have messed up somewhere, I'll try to fix this quickly

superlopuh: I had a look, and this change is explicitly incorrect, at least according to MLIR:

```sh
$ toyc-ch2 docs/Toy/examples/scalar.toy -emit=mlir
> module {
>   toy.func @main() {
>     %0 = toy.constant dense<5.500000e+00> : tensor<f64>
>     %1 = toy.reshape(%0 : tensor<f64>) to tensor<2x2xf64>
>     toy.print %1 : tensor<2x2xf64>
>     toy.return
>   }
> }
```

superlopuh: https://mlir.llvm.org/docs/Dialects/Builtin/#example-6

superlopuh: Do you know which of the changes below made this change as consequence? We should add an explicit unit test for the builder to preserve this behaviour, it's a bad sign that Toy tests caught it

tests/dialects/test_builtin.py
```diff
@@ -246,7 +246,7 @@ def test_DenseIntOrFPElementsAttr_fp_type_conversion():
 def test_DenseIntOrFPElementsAttr_from_list():
     attr = DenseIntOrFPElementsAttr.tensor_from_list([5.5], f32, [])
 
-    assert attr.type == AnyTensorType(f32, [])
+    assert attr.type == AnyTensorType(f32, [1])
```
alexarice, line 249: My understanding is that this change shouldn't change the external functionality (except for speed). Should this have been changed?

tests/interpreters/test_ml_program_interpreter.py
```diff
@@ -35,7 +35,7 @@ def test_ml_program_global_load_constant():
     interpreter.register_implementations(MLProgramFunctions())
 
     (result,) = interpreter.run_op(fetch, ())
-    assert result == ShapedArray(TypedPtr.new_int32([4]), [4])
+    assert result == ShapedArray(TypedPtr.new_int32([4] * 4), [4])
```
alexarice, line 38: Same here, why is the interface changing?

jorendumoulin:
Oh, I was completely unaware of the zero-rank tensor, sorry! I wrongfully assumed some of the tests were just a bit quirky.

What still bothers me is the following test:
https://github.com/xdslproject/xdsl/blob/8a721ad5092421a3b8d3552ac05d05f599bb8cb4/tests/dialects/test_builtin.py#L220
this shouldn't be legal, right? storing two values in a zero-rank tensor?

Does the following make sense as expected behaviour?
```python

    # legal zero-rank tensor
    attr = DenseIntOrFPElementsAttr.tensor_from_list([4], f32, [])
    assert attr.type == AnyTensorType(f32, [])

    # illegal zero-rank tensor
    with pytest.raises(ValueError, match="A zero-rank tensor can only hold 1 value but 2 were given."):
        DenseIntOrFPElementsAttr.tensor_from_list([4, 5], f32, [])

    # legal normal tensor
    attr = DenseIntOrFPElementsAttr.tensor_from_list([4, 5], f32, [2])
    assert attr.type == AnyTensorType(f32, [2])

```

docs/Toy/examples/scalar.toy
jorendumoulin: I added explicit support for zero-rank tensors, so this change is reverted!

tests/dialects/test_builtin.py
jorendumoulin: this change is reverted with the support for zero rank tensors

tests/interpreters/test_ml_program_interpreter.py
jorendumoulin: In this case I do think the test was a bit fishy.
If you ask the values of a `1x4` array, would you not expect for the array to have 4 values (regardless of whether it is initialized as a splat)?
The previous implementation allowed for the nb of stored elements to differ from the nb of elements defined by the shape. This new one doesn't which is why

docs/Toy/examples/scalar.toy
jorendumoulin: I will also add some more extensive filetests tomorrow!

xdsl/dialects/builtin.py
```diff
@@ -1927,6 +1927,10 @@ def get_shape(self) -> tuple[int, ...]:
     def get_element_type(self) -> IntegerType | IndexType | AnyFloat:
         return self.type.get_element_type()
 
+    @property
+    def nb_elements(self) -> int:
+        return prod(self.get_shape())
+
```
superlopuh, line 1933: The code changes are now looking great. Just to make sure that I didn't miss anything, could you please split this PR into one that introduces this helper and a verification to check that the number of elements contained matches the size of the shape? The goal here is for this PR to change only the underlying implementation and none of the API.

Also, I think this could just be `__len__`

xdsl/dialects/builtin.py
```diff
@@ -2023,25 +2040,42 @@ def from_list(
...
             new_type = cast(RankedStructure[IndexType], type)
-            new_data = cast(Sequence[int] | Sequence[IntegerAttr[IndexType]], data)
+            new_data = cast(Sequence[int] | Sequence[IntegerAttr[IndexType]], new_data)
             return DenseIntOrFPElementsAttr.create_dense_index(new_type, new_data)
 
     @staticmethod
     def vector_from_list(
         data: Sequence[int] | Sequence[float],
         data_type: IntegerType | IndexType | AnyFloat,
+        shape: Sequence[int],
```
superlopuh, line 2076: could this not be optional, and inferred from the length of the list?

xdsl/dialects/builtin.py
jorendumoulin: okay, I will do that, thank you for the review!

xdsl/dialects/builtin.py
jorendumoulin: then the difference between `vector<1xf64>` and `vector<f64>` becomes indistinguishable

superlopuh: yep I would say you could add `()` explicitly for the scaler

jorendumoulin: oh, it is also just required to differentiate between `vector<4xf64>` and `vector<2x2xf64>`, vectors can be multi-dimensional

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -427,8 +435,15 @@ def match_and_rewrite(self, op: ConstantOp, rewriter: PatternRewriter, /):
             if typ := get_required_result_type(op):
                 if needs_update_shape(op.result.type, typ):
                     assert isinstance(op.value, DenseIntOrFPElementsAttr)
+                    dense_type = cast(
+                        RankedStructure[IntegerType | IndexType | AnyFloat], typ
+                    )
                     rewriter.replace_matched_op(
-                        ConstantOp(DenseIntOrFPElementsAttr([typ, op.value.data]))
```
superlopuh, line 431: Hmm, this should continue working, no?

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
```diff
@@ -261,7 +267,9 @@ def is_tensorized(
     return len(typ.get_shape()) == 2 and isinstance(typ.get_element_type(), TensorType)
 
 
-def is_tensor(typ: Attribute) -> TypeGuard[TensorType[Attribute]]:
+def is_tensor(
+    typ: Attribute,
+) -> TypeGuard[TensorType[IndexType | IntegerType | AnyFloat]]:
```
superlopuh, line 272: why this change?

xdsl/dialects/builtin.py
```diff
@@ -2102,43 +2115,52 @@ def get_values(self) -> Sequence[int] | Sequence[float]:
...
+        elif isinstance(self.type.element_type, IntegerType):
+            return tuple(
+                IntegerAttr(value, self.type.element_type)
+                for value in cast(tuple[int, ...], self.get_values())
+            )
+        else:  # index
+            return tuple(
+                IntegerAttr(value, self.type.element_type)
+                for value in cast(tuple[int, ...], self.get_values())
+            )
```
superlopuh, line 2139: This would be a great helper in a separate PR, I think, and also provided for DenseArrayBase. I wonder if we should change `iter/unpack` to `iter_values/unpack_values` and `iter_attrs/unpack_attrs`

tests/interpreters/test_ml_program_interpreter.py
jorendumoulin: resolved in #3637

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
jorendumoulin: yes, indeed

xdsl/transforms/experimental/stencil_tensorize_z_dimension.py
jorendumoulin: sorry if this is incorrect, as i'm not fully understanding what this pass does, but it looked like `is_tensor` check is used whether an operand is a tensor and can be rewritten as `DenseIntOrFPElementsAttr`, but `Tensortype[Attribute]` is not good enough of a constraint for that, as the dense type can only store tensors of these values.

I'm not sure why pyright didn't complain about this before, I'll try to investigate

jorendumoulin: Oh it didn't complain before because there is no check when you don't use the custom constructors

superlopuh: Ah nice, I see

compor (approved):

xdsl/dialects/builtin.py
superlopuh: #3706

jorendumoulin:
> Just for the sake of ideal PRs, it would be great to split out all the changes that aren't in builtin.py to a separate PR, so that the underlying storage change is transparent, if that's not too tedious!

No problem! See #3715 , #3716, #3717, #3718

jorendumoulin:
This is now rebased, only remaining with changes in the class definition itself!

superlopuh (approved):
🚀🚀

PR #3627, opened 2024-12-11
misc: use base printer infrastructure in WGSL printing

superlopuh:
Changing to the new parser infrastructure, small diff for now, but should be easier to use helpers if this ever gets picked up in the future.

CC @PapyChacal @cigarichard


tests/backend/wgsl/test_wgsl_printer.py
```diff
@@ -10,177 +10,155 @@
...
-    file = StringIO("")
+    stream = StringIO()
+    printer = WGSLPrinter(stream=stream)
 
     cst = arith.ConstantOp(IntegerAttr(42, IndexType()))
-
-    printer = WGSLPrinter()
-    printer.print(cst, file)
-
-    assert "let v0 : u32 = 42u;" in file.getvalue()
```
n-io, line 64: Is this assert still required?

n-io (approved):

PR #3628, opened 2024-12-11
CI: install wgpu and onnx in mlir ci

superlopuh:
In our migration to uv, we missed these

### DIFF
--- a/.github/workflows/ci-mlir.yml
+++ b/.github/workflows/ci-mlir.yml
@@ -50,7 +50,7 @@ jobs:
       run: |
         # Change directory so that xdsl-opt can be found during installation.
         cd xdsl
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-mlir.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       run: |
         # Change directory so that xdsl-opt can be found during installation.
         cd xdsl
-        make venv
+        UV_PROJECT_ENVIRONMENT=.venv uv sync --extra gui --extra dev --extra jax --extra riscv --extra wgpu --extra onnx
```
EdmundGoodman, line 53: Instead of this, could we still go through the Makefile to ensure everything stays aligned

```suggestion
        VENV_EXTRAS="--extra gui --extra dev --extra jax --extra riscv --extra wgpu --extra onnx" make venv
```

compor (approved):

PR #3629, opened 2024-12-11
misc: use .venv by default instead of venv

superlopuh:
I didn't want to have this discussion before we merged uv, but it definitely feels like it really prefers `.venv`, and makes lots of things easier if this name is used. How do we feel about using this name by default? It can still easily be overridden locally, but will change to .venv by default for everyone.


alexarice (approved):

Makefile
```diff
@@ -5,7 +5,7 @@ MAKEFLAGS += --no-builtin-variables
 COVERAGE_FILE ?= .coverage
 
 # allow overriding the name of the venv directory
-VENV_DIR ?= venv
+VENV_DIR ?= .venv
```
EdmundGoodman, line 8: The `venv` make target on line 32 should also be renamed to `.venv`, as that is now what `uv sync` will create

EdmundGoodman (commented):
Just a note that this may break some dev phonies in the Makefile for existing users with a `venv/` already generated, as it will create a new `.venv` directory when they are run, but not install the `VENV_EXTRAS` into them unless make `venv` is run first.

This is not necessarily a big problem as they can just run `make .venv/ && rm -rf venv`, but might confuse some people?

Makefile
superlopuh: I tested it locally and `make venv` still creates a `.venv`

compor:
> Just a note that this may break some dev phonies in the Makefile for existing users with a `venv/` already generated, as it will create a new `.venv` directory when they are run, but not install the `VENV_EXTRAS` into them unless make `venv` is run first.
> 
> This is not necessarily a big problem as they can just run `make .venv/ && rm -rf venv`, but might confuse some people?

we can't account for everything, this is what I did locally as well.

Makefile
compor: so do we want this suggestion?

compor (approved):

PR #3630, opened 2024-12-12
misc: use activated venv dir if any, and add back project venv export

superlopuh:
This change lets us use the activated venv, if any, for all our make commands. In order to make it work, I had to add back the export, because otherwise the inner uv commands would complain about the fact that the current venv is not the same as the project venv, which is .venv if not overridden.

@compor, what was the actual issue you saw that led you to remove it?


Makefile
```diff
@@ -5,8 +5,11 @@ MAKEFLAGS += --no-builtin-variables
 COVERAGE_FILE ?= .coverage
 
 # allow overriding the name of the venv directory
-VENV_DIR ?= .venv
-UV_PROJECT_ENVIRONMENT=${VENV_DIR}
+DEFAULT_XDSL_VENV_DIR ?= .venv
+
+# use activated venv if any
+VENV_DIR ?= $(if $(VIRTUAL_ENV),$(VIRTUAL_ENV),$(DEFAULT_XDSL_VENV_DIR))
+export UV_PROJECT_ENVIRONMENT=${VENV_DIR}
```
EdmundGoodman, line 12: Per https://github.com/xdslproject/xdsl/pull/3626/files , should this instead be the following?

```suggestion
UV_PROJECT_ENVIRONMENT=${VENV_DIR}
```

superlopuh: It's an explicit revert of that change, I don't understand why it was made, and it seems like it's required if we want uv to work for non `.venv` env names (https://github.com/astral-sh/uv/issues/9834#issuecomment-2539032524)

compor, line 12: Can we not simplify as

```makefile
# allow overriding the name of the venv directory
VENV_DIR ?= .venv

# use activated venv if any
export UV_PROJECT_ENVIRONMENT=$(if $(VIRTUAL_ENV),$(VIRTUAL_ENV),$(VENV_DIR))
```

Where `VENV_DIR` is just ignored if we are already in an active venv?

compor (approved):

Makefile
compor: This keeps the same make usage as we had been using prior to all the changes to accommodate `uv`.
But I don't feel strong about it.

superlopuh: I'll try this

compor (approved):

EdmundGoodman (approved):

PR #3631, opened 2024-12-12
transformations: (apply-individual-rewrite) lazily find canonicalization patterns

superlopuh:
It's time to make this part of the codebase a bit more extensible and robust, as we will be reusing it for some scheduling exploration. As a first step, this PR changes the rewrite infrastructure to get the available patterns on the operation lazily, instead of by scanning all ops of all dialects.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2390,7 +2390,7 @@ wheels = [
 
 [[package]]
 name = "xdsl"
-version = "0.25.0+37.g414bcb0e.dirty"
+version = "0.25.0+89.gb85bb1a8"
 source = { editable = "." }
 dependencies = [
     { name = "immutabledict" },
--- a/uv.lock
+++ b/uv.lock
@@ -2390,7 +2390,7 @@ wheels = [
 
 [[package]]
 name = "xdsl"
-version = "0.25.0+89.gb85bb1a8"
+version = "0.25.0+37.g414bcb0e.dirty"
 source = { editable = "." }
 dependencies = [
     { name = "immutabledict" },
--- a/xdsl/transforms/individual_rewrite.py
+++ b/xdsl/transforms/individual_rewrite.py
@@ -1,7 +1,7 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from xdsl.context import MLContext
-from xdsl.dialects import arith, get_all_dialects
+from xdsl.dialects import arith
 from xdsl.dialects.builtin import IndexType, IntegerAttr, IntegerType, ModuleOp
 from xdsl.ir import Operation
 from xdsl.passes import ModulePass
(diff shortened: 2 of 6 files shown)


xdsl/transforms/individual_rewrite.py
```diff
@@ -94,30 +90,34 @@ class ApplyIndividualRewritePass(ModulePass):
 
     name = "apply-individual-rewrite"
 
-    matched_operation_index: int | None = None
-    operation_name: str | None = None
-    pattern_name: str | None = None
```
superlopuh, line 99: none of these were really optional

uv.lock
```diff
@@ -2390,7 +2390,7 @@ wheels = [
 
 [[package]]
 name = "xdsl"
-version = "0.25.0+89.gb85bb1a8"
+version = "0.25.0+37.g414bcb0e.dirty"
```
alexarice, line 2393: This looks suspicious

xdsl/transforms/individual_rewrite.py
```diff
@@ -44,43 +44,26 @@ def match_and_rewrite(self, op: arith.DivUIOp, rewriter: PatternRewriter) -> Non
             rewriter.replace_matched_op([], [mul_op.lhs])
 
 
-INDIVIDUAL_REWRITE_PATTERNS_BY_OP_CLASS: dict[
-    type[Operation], tuple[RewritePattern, ...]
-] = {
-    arith.AddiOp: (AdditionOfSameVariablesToMultiplyByTwo(),),
-    arith.DivUIOp: (DivisionOfSameVariableToOne(),),
+INDIVIDUAL_REWRITE_PATTERNS_BY_NAME: dict[str, dict[str, RewritePattern]] = {
```
alexarice, line 47: I don't understand why this change fits into this PR? It's also unclear to me that matching on a string is better than matching on a type

alexarice: Further, why change this to not being a tuple?

alexarice: On further thought, why is this necessary at all. Can these two just be made into canonicalization patterns?

xdsl/transforms/individual_rewrite.py
```diff
@@ -94,30 +77,34 @@ class ApplyIndividualRewritePass(ModulePass):
 
     name = "apply-individual-rewrite"
 
-    matched_operation_index: int | None = None
-    operation_name: str | None = None
-    pattern_name: str | None = None
+    matched_operation_index: int = field()
+    operation_name: str = field()
+    pattern_name: str = field()
```
alexarice, line 82: I had been thinking that it might be more ergonomic here to just store the pattern directly here, instead of storing the name of the pattern and then looking it up, as we currently duplicate the lookup. Thoughts?

xdsl/transforms/individual_rewrite.py
```diff
@@ -44,43 +44,26 @@ def match_and_rewrite(self, op: arith.DivUIOp, rewriter: PatternRewriter) -> Non
...
-    for op in set(INDIVIDUAL_REWRITE_PATTERNS_BY_OP_CLASS)
-    | set(CANONICALIZATION_PATTERNS_BY_OP_CLASS)
-}
-"""
-Returns a dictionary representing all possible rewrites. Keys are operation names, and
-values are dictionaries. In the inner dictionary, the keys are names of patterns
-associated with each operation, and the values are the corresponding RewritePattern
-instances.
-"""
+def _get_pattern(op: Operation, pattern_name: str) -> RewritePattern | None:
```
alexarice, line 60:
```suggestion
def _get_canonicalization_pattern(op: Operation, pattern_name: str) -> RewritePattern | None:
```
I thought this was including the individual rewrites at first

alexarice (approved):
Great change, this is one of the parts of the interactive gui which doesn't translate well downstream. It's possible this change fixes that issue.

uv.lock
superlopuh: my bad

xdsl/transforms/individual_rewrite.py
superlopuh: It's at least not a regression

xdsl/transforms/individual_rewrite.py
alexarice: fair enough, that makes a lot of sense

PR #3635, opened 2024-12-13
dialects: (arith) add truncation option to ConstantOp

alexarice:
split off from #3583

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2015,27 +2015,27 @@ wheels = [
 
 [[package]]
 name = "ruff"
-version = "0.8.2"
+version = "0.8.3"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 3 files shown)


uv.lock
superlopuh, line 1: uh oh, let's separate this out?

superlopuh: This is my bad, I think the new workflow should be to make this change a part of the dependabot PR, I'll do this quickly

xdsl/dialects/arith.py
```diff
@@ -166,13 +166,18 @@ def __init__(
...
+        truncate_bits: bool = False,
     ) -> ConstantOp:
         if isinstance(value_type, int):
             value_type = IntegerType(value_type)
         return ConstantOp.create(
             result_types=[value_type],
-            properties={"value": IntegerAttr(value, value_type)},
+            properties={
+                "value": IntegerAttr(value, value_type, truncate_bits=truncate_bits)
+            },
```
superlopuh, line 180: could you please add a unit test for this?

uv.lock
superlopuh: Ah no I merged textual, you have a newer version of ruff than our requirements

alexarice: This is not ideal that switching between branches messes up lock files

superlopuh (approved):
TY

uv.lock
superlopuh: In theory, this is exactly what UV is supposed to enable, switching branches with different dependencies and updating those silently

PR #3636, opened 2024-12-13
dependencies: update textual lockfile

superlopuh:


alexarice:
Is textual 0.89 to textual 1.0 not a sizable change?

superlopuh:
Not really, it's a symbolic change. Also we already updated our requirements, this is an update to the lockfile

alexarice (approved):

PR #3638, opened 2024-12-13
dependencies: update ruff

superlopuh:


jorendumoulin (approved):

uv.lock
```diff
@@ -2390,7 +2390,7 @@ wheels = [
 
 [[package]]
 name = "xdsl"
-version = "0.25.0+43.g10221499"
+version = "0.25.0+47.g46c79300"
```
alexarice, line 2393: Unrelated, but does it make sense for xdsl itself to be in the lock file? Is this just how uv manages things?

alexarice (approved):

PR #3642, opened 2024-12-16
interactive: use individual rewrite pass directly instead of helpers

superlopuh:
It doesn't seem that this abstraction currently adds anything, both the available pass and pass itself are already immutable and can be reasoned about directly.


alexarice (approved):
Seems reasonable, though I'm not sure what the `AvailablePass` or old infrastructure was there to achieve

dshaaban01 (commented):
Nice

superlopuh:
There was a time when passes weren't guaranteed to be immutable, so there was a good reason to separate the app things from the rewrite things, now we can bring them a bit closer together. I think I'll slowly refactor this now that I'm working on these bits of the code more.

PR #3643, opened 2024-12-16
dialects (llvm): Add dense array constraint for the `position` attribute of `llvm.extractvalue` and `llvm.insertvalue` operations

### LINKED ISSUE #3155: In xDSL the llvm.InsertValueOp and llvm.ExtractValueOp ops don't have the same restrictions on the element_type of the DenseArray attribute as in MLIR
In MLIR both only accept `i64` as element types. In xDSL there are not restrictions

compor:
This PR:

- Adds a dense array constraint for the `position` attribute of `llvm.extractvalue` and `llvm.insertvalue` operations, restricting to `i64` as in MLIR.
- Filecheck tests of the above

Resolves #3155

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1248,6 +1248,13 @@ class DenseArrayBase(ParametrizedAttribute):
         return len(self.data.data) // self.elt_type.size
 
 
+DenseI64ArrayConstraint = ParamAttrConstraint(DenseArrayBase, [i64, BytesAttr])
+"""Type constraint for DenseArrays containing integers of i64 integers."""
+
+DenseI64ArrayType: TypeAlias = Annotated[DenseArrayBase, DenseI64ArrayConstraint]
+"""Type alias constrained to dense array of i64 integers."""
+
+
 @irdl_attr_definition
 class FunctionType(ParametrizedAttribute, TypeAttribute):
     name = "fun"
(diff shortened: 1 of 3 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1248,6 +1248,13 @@ def __len__(self) -> int:
         return len(self.data.data) // self.elt_type.size
 
 
+DenseI64ArrayConstraint = ParamAttrConstraint(DenseArrayBase, [i64, BytesAttr])
```
compor, line 1251: Should this be `DenseI64ArrayConstr`? Not sure if we have a naming convention here.

superlopuh: I prefer Constr, it's what I've used so far

xdsl/dialects/builtin.py
```diff
@@ -1248,6 +1248,13 @@ def __len__(self) -> int:
...
+"""Type constraint for DenseArrays containing integers of i64 integers."""
+
+DenseI64ArrayType: TypeAlias = Annotated[DenseArrayBase, DenseI64ArrayConstraint]
```
superlopuh, line 1254:
```suggestion
DenseI64Array: TypeAlias = Annotated[DenseArrayBase, DenseI64ArrayConstraint]
```

superlopuh: But also this seems not to be used?

compor: We do this with other types, so I wasn't sure.
Delete?

superlopuh (approved):

PR #3644, opened 2024-12-16
CI: initialize mlir-aie CI

jorendumoulin:
This PR initalizes the CI workflow for mlir-aie based testing.

There are no tests yet, so nothing is tested except for printing out the `mlir-aie` version to see if we can use it correctly.

The project does not really have a versioning strategy, but rather regularly updates its llvm commit, and pushes prebuilt binaries daily with the release version `0.0.1.<date>+<commit>`.

As the mlir version of mlir-aie will not necessarily correspond to the mlir verison of xdsl, i suggest only testing mlir-aie specific dialects with this ci (which I imagine will reside in `/tests/filecheck/with-mlir-aie/dialects/aie.mlir`)


superlopuh:
Maybe it's time to discuss moving the aie things to a new project in xdslproject called xdsl-aie or something like that

jorendumoulin:
I am happy to! I am now just basing myself on my poll here https://xdsl.zulipchat.com/#narrow/channel/364199-Beginners/topic/AMD.20AIE.3F

superlopuh:
In a way, the extra CI changes the inputs to the poll somewhat. It might be worth picking it up there, but I think that it makes sense to have subprojects like this one in a separate project. I'm still happy to volunteer to update the APIs when we change them.

compor:
We have done both recently (in and downstream of the repo), so not sure what the actual criteria are.
I'd be happy to support this wherever it lives.
Maybe we can continue as we are with the requirement that AIE stuff moves to `main` and have a separate CI repo for that? Not great I have to admit, but if the CI support is the sticky issue we can circumvent for now?

jorendumoulin:
development moved to xdslproject/xdsl-aie#4

PR #3646, opened 2024-12-16
CI: trigger lockfile CI automatically on dependabot PRs

superlopuh:

PR #3647, opened 2024-12-16
CI: add lockfile update action to manually trigger for dependabot PRs

superlopuh:
I tested it [here](https://github.com/xdslproject/xdsl/pull/3646) and it works if you do the dumb thing and trigger on push, but it seems to not run actions on unsigned commits by default. I'm hoping that if we trigger the action manually it'll inherit the clicker's permissions and re-run the action. In either case, I think whatever we adopt should supercede the cron job one as we want to never be out of sync between the pyproject and lockfile.


alexarice (commented):
So this needs to be manually run on dependabot prs?

superlopuh:
For now, yes

jorendumoulin (approved):

superlopuh:
Stop-gap until this beast is merged: https://github.com/dependabot/dependabot-core/pull/10040

PR #3653, opened 2024-12-18
transforms: (memref-to-dsd) Support 1d subview of nd memref

n-io:


xdsl/transforms/lower_csl_stencil.py
```diff
@@ -202,24 +202,25 @@ def match_and_rewrite(self, op: csl_stencil.ApplyOp, rewriter: PatternRewriter,
         # ensure we send only core data
         assert isa(op.accumulator.type, memref.MemRefType[Attribute])
         assert isa(op.field.type, memref.MemRefType[Attribute])
+        # the accumulator might have additional dims when used for holding prefetched data
+        send_buf_shape = op.accumulator.type.get_shape()[
+            -len(op.field.type.get_shape()) :
+        ]
```
dk949, line 208: Do we always just discard the higher dims?

xdsl/transforms/memref_to_dsd.py
```diff
@@ -119,7 +123,48 @@ class LowerSubviewOpPass(RewritePattern):
...
-        assert isa(op.source.type, MemRefType[Attribute])
+        assert isa(op.source.type, AnyMemRefType)
+        assert isa(op.result.type, AnyMemRefType)
+
+        if len(op.result.type.get_shape()) == 1 and len(op.source.type.get_shape()) > 1:
+            # 1d subview onto a nd memref
+            sizes = op.static_sizes.get_values()
+            counter_sizes = collections.Counter(sizes)
+            if 1 in counter_sizes:
+                counter_sizes.pop(1)
```
dk949, line 134: You can do `counter_sizes.pop(1, None)` to remove `1` if it's present and do nothing otherwise. Though this may be less readable

xdsl/transforms/memref_to_dsd.py
```diff
@@ -119,7 +123,48 @@ class LowerSubviewOpPass(RewritePattern):
...
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: memref.SubviewOp, rewriter: PatternRewriter, /):
-        assert isa(op.source.type, MemRefType[Attribute])
+        assert isa(op.source.type, AnyMemRefType)
+        assert isa(op.result.type, AnyMemRefType)
+
+        if len(op.result.type.get_shape()) == 1 and len(op.source.type.get_shape()) > 1:
+            # 1d subview onto a nd memref
+            sizes = op.static_sizes.get_values()
+            counter_sizes = collections.Counter(sizes)
```
dk949, line 132: I don't really understand why the `Counter` is needed here

dk949 (commented):
I'm a little confused on the `memref_to_dsd` part

xdsl/transforms/lower_csl_stencil.py
n-io: Yes, for prefetching it's set up as n times what the accumulator would normally be.

xdsl/transforms/memref_to_dsd.py
n-io: Good point!

xdsl/transforms/memref_to_dsd.py
n-io: To assert that all but one values are 1, to find the value that isn't 1, and later to find its index. Essentially, to find the active dimension. Plus some asserting that only one dimension is active, as we're creating a 1-d subview on an n-d memref.

dk949 (approved):
Looks great!

PR #3657, opened 2024-12-19
dialects: (csl) Switch dsds to use affine maps

n-io:


xdsl/transforms/canonicalization_patterns/csl.py
```diff
@@ -81,21 +93,27 @@ def match_and_rewrite(self, op: csl.GetMemDsdOp, rewriter: PatternRewriter) -> N
             stride_op := next(iter(op.result.uses)).operation, csl.SetDsdStrideOp
         ):
             return
-        # only works on 1d
-        if op.offsets and len(op.offsets) > 1:
+        # only works on 1d and default (unspecified) tensor_access
+        if len(op.sizes) > 1 or op.tensor_access:
             return
```
dk949, line 99: I think it'd be better to have an `is_mem1d` member function rather than relying on the internals of the operation like this

xdsl/transforms/canonicalization_patterns/csl.py
```diff
@@ -81,21 +93,27 @@ def match_and_rewrite(self, op: csl.GetMemDsdOp, rewriter: PatternRewriter) -> N
...
         if (
             isinstance(stride_op.stride, OpResult)
             and isinstance(cnst := stride_op.stride.op, arith.ConstantOp)
-            and isattr(cnst.value, AnyIntegerAttrConstr)
+            and isa(attr_val := cnst.value, AnyIntegerAttr)
         ):
+            tensor_access = AffineMap.from_callable(
+                lambda x: (x * attr_val.value.data,)
+            )
             rewriter.replace_matched_op(
```
dk949, line 109: Is there a way to not need to suppress all the warnings?

If not, could you please comment why they're suppressed?

dk949 (commented):
Looks great!

Could you also please add a test case (in the backend filecheck) to test for `mem1d_dsd` accessing a 2d array?

dk949 (approved):
Thanks!

emmau678 (approved):

PR #3658, opened 2024-12-19
interactive: add mlir-opt pass to condensed list

dshaaban01:
This actually adds the mlir-opt pass to the condensed list (see commented line 65), this previous implementation did not do that - it did nothing, in fact.


superlopuh:
Can you please change the title to reflect the error, instead of describing it as small?

superlopuh:
Also, could you please set the label?

dshaaban01:
Done.

compor (approved):

PR #3662, opened 2024-12-20
dialects: (scf/scf_for) Factor out print and parse methods for For-like operations

compor:
This PR:

- Factors out print and parse methods for For-like operations (currently `scf.for`, `riscv_scf.for` and `riscv_scf.rof`.

### DIFF
--- a/tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
+++ b/tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
@@ -6,7 +6,7 @@ builtin.module {
         %zero = riscv.li 0 : !riscv.reg<a2>
         %step = riscv.li 4 : !riscv.reg<a3>
         %forty = riscv.li 40 : !riscv.reg<a4>
-        riscv_scf.for %offset : !riscv.reg<a5> = %zero to %forty step %step {
+        riscv_scf.for %offset = %zero to %forty step %step : !riscv.reg<a5> {
             %srcptr = riscv.add %src, %offset : (!riscv.reg<a0>, !riscv.reg<a5>) -> !riscv.reg<a6>
             %dstptr = riscv.add %dst, %offset : (!riscv.reg<a1>, !riscv.reg<a5>) -> !riscv.reg<a7>
             %val = riscv.lw %srcptr, 0 : (!riscv.reg<a6>) -> !riscv.reg<t0>
@@ -47,7 +47,7 @@ builtin.module {
     riscv_func.func @sum_range(%0 : !riscv.reg<a0>, %1 : !riscv.reg<a1>) {
         %2 = riscv.li 1 : !riscv.reg<a2>
         %3 = riscv.li 0 : !riscv.reg<a3>
-        %4 = riscv_scf.for %5 : !riscv.reg<a4> = %0 to %1 step %2 iter_args(%6 = %3) -> (!riscv.reg<a3>) {
+        %4 = riscv_scf.for %5 = %0 to %1 step %2 iter_args(%6 = %3) -> (!riscv.reg<a3>) : !riscv.reg<a4> {
             %7 = riscv.add %5, %6 : (!riscv.reg<a4>, !riscv.reg<a3>) -> !riscv.reg<a3>
             riscv_scf.yield %7 : !riscv.reg<a3>
         }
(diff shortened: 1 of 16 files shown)


tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
```diff
@@ -6,7 +6,7 @@ builtin.module {
         %zero = riscv.li 0 : !riscv.reg<a2>
         %step = riscv.li 4 : !riscv.reg<a3>
         %forty = riscv.li 40 : !riscv.reg<a4>
-        riscv_scf.for %offset : !riscv.reg<a5> = %zero to %forty step %step {
+        riscv_scf.for %offset = %zero to %forty step %step : !riscv.reg<a5> {
```
compor, line 9: This brings the format inline with `scf.for` where the induction variable type is printed last when it does not default to `index` type.

xdsl/dialects/utils/format.py
```diff
@@ -15,12 +16,28 @@
...
+
+def parse_assignment(
+    parser: Parser,
+) -> tuple[Parser.UnresolvedArgument, UnresolvedOperand]:
+    arg = parser.parse_argument(expect_type=False)
+    parser.parse_characters("=")
+    val = parser.parse_unresolved_operand()
+    return arg, val
+
+
```
compor, line 40: just moved up

xdsl/dialects/utils/format.py
```diff
@@ -45,6 +173,10 @@ def print_func_op_like(
     arg_attrs: ArrayAttr[DictionaryAttr] | None = None,
     reserved_attr_names: Sequence[str],
 ):
+    """
```
compor, line 176: Added missing docstring while at it.

xdsl/dialects/utils/format.py
```diff
@@ -35,6 +52,117 @@ def __init__(self, *operands: SSAValue | Operation):
...
 
+def print_for_op_like(
+    printer: Printer,
+    lower_bound: SSAValue,
+    upper_bound: SSAValue,
+    step: SSAValue,
+    iter_args: Sequence[SSAValue],
+    body: Region,
+    bound_words: Sequence[str] = ["to"],
+    indvar_type: type[TypeAttribute] | None = None,
```
alexarice, line 63: It definitely wasn't clear to me what this did just from the name, perhaps `default_indvar_type` might be slightly better? (I see a description is in the docstring.

alexarice (approved):
This is potentially an interesting target for assembly format. I don't think we have any support for region arguments in assembly_format/pyrdl yet

xdsl/dialects/utils/format.py
compor: yeah, I debated back and forth on the name; I'll amend.

tests/filecheck/backend/convert_riscv_scf_to_riscv_cf.mlir
superlopuh: wait but this is different, the offset is not the step

alexarice: The same thing happens with `scf.for`

superlopuh: I'm worried this will be confusing. This is less confusing in MLIR, as all the types are the same, but in regalloc'ed riscv_scf, each of lb, ub, and step may be in different registers. Also actually we may want the lowering to riscv_cf to add a move from lb to induction, meaning that each of these would have a separate type. We may want to discuss that first before changing the API here?

compor: I understand the confusing bit, but I fail to see what it has to do lowering or optimizing it.
This is just the print format, nothing has changed in the underlying representation of the for operation.
What am I missing?

As a stop gap, the location of the induction variable printing can also be an option.

superlopuh: what's changed is that the type is currently only for the induction variable, and is printed near it. the proposed change implies that it's either the type of all the variables, or the step one specifically

compor: After revisiting this, I agree that is confusing, and I'll amend to keep the location where the induction var type is printed, but I'll try to make it configurable.

tests/filecheck/projects/riscv-backend-paper/bottom_up_f64.mlir
superlopuh, line 1: Is this intentional?

superlopuh (approved):

PR #3664, opened 2024-12-22
dialects: (builtin) Parse @ module names 

mamanain:
It seems that modules in xdsl already support sym_name attribute but we don't parse @-starting names.


superlopuh (commented):
This looks great, although this is a good opportunity to make the module name a property, and it'll simplify the format logic:

```
❯ echo "module @hello {}" | mlir-opt --mlir-print-op-generic
"builtin.module"() <{sym_name = "hello"}> ({
^bb0:
}) : () -> ()
```

superlopuh:
Does this work?

``builtin.module` ($sym_name^)? attr-dict-with-keyword $bodyRegion`

mamanain:
> Does this work?
> 
> `` `builtin.module` ($sym_name^)? attr-dict-with-keyword $bodyRegion ``

You mean does the combination of module name and attribute works? It does, added a test for it.

xdsl/dialects/builtin.py
```diff
@@ -1643,21 +1643,30 @@ def __init__(
         self,
         ops: list[Operation] | Region,
         attributes: Mapping[str, Attribute] | None = None,
+        properties: dict[str, Attribute] | None = None,
```
math-fehr, line 1646: Ad properties are not extensible, and would only contain the sym_name, I would just pass a `StringAttr | None` instead, and create the properties dictionary in the __init__

math-fehr (commented):
Looks good, I would just change a bit the __init__ with my comment

xdsl/dialects/builtin.py
mamanain: Makes sense. Done, thanks.

xdsl/dialects/builtin.py
```diff
@@ -1667,9 +1671,14 @@ def parse(cls, parser: Parser) -> ModuleOp:
...
             region.add_block(Block())
 
-        return ModuleOp(region, attributes)
+        return ModuleOp(region, attributes, module_name)
 
     def print(self, printer: Printer) -> None:
+        if "sym_name" in self.properties and isinstance(
+            self.properties["sym_name"], StringAttr
+        ):
+            printer.print(f" @{self.properties['sym_name'].data}")
```
superlopuh, line 1680:
```suggestion
        if self.sym_name is not None:
            printer.print(f" @{self.sym_name.data}")
```

superlopuh (approved):
Thank you!

superlopuh:
I meant to use the assembly_format, but there's currently a circular dependency of the assembly format parser on builtin dialect so let's not go there.

PR #3666, opened 2024-12-22
interactive: when generating new MLContext, set allow_unregistered attribute to default (=false)

dshaaban01:
This PR sets the MLContext, `allow_unregistered` attribute to its default, which is false, instead of having it at true. 

This was causing a weird parsing error that did not allow the `mlir-opt` pass to run --> it was parsing `module {}` as UnregisteredOp. 

This fixes it.

### DIFF
--- a/xdsl/interactive/passes.py
+++ b/xdsl/interactive/passes.py
@@ -26,7 +26,7 @@ def get_new_registered_context(
     """
     Generates a new MLContext, registers it and returns it.
     """
-    ctx = MLContext(True)
+    ctx = MLContext()
     for dialect_name, dialect_factory in all_dialects:
         ctx.register_dialect(dialect_name, dialect_factory)
     return ctx
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -53,13 +53,13 @@ async def test_inputs():
         await pilot.pause()
         assert (
             app.output_text_area.text
-            == "<unknown>:1:5\ndkjfd\n     ^\n     Operation builtin.unregistered does not have a custom format.\n"
+            == "<unknown>:1:5\ndkjfd\n     ^\n     unregistered operation dkjfd!\n"
         )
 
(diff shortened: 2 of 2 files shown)


xdsl/interactive/passes.py
```diff
@@ -26,7 +26,7 @@ def get_new_registered_context(
     """
     Generates a new MLContext, registers it and returns it.
     """
-    ctx = MLContext(True)
+    ctx = MLContext()
```
alexarice, line 29: If we want to go ahead with this change then could we at least have allow_unregistered_dialect passed in as a command line argument?

alexarice (commented):
Why does this break mlir-opt pass? It feels like this isn't actually fixing whatever the real issue is and is at worst a regression.

superlopuh (commented):
I don't think this is the right fix. The problem we discussed is that `module {}`, which is a valid spelling of `builtin.module {}` in MLIR, is parsed as an unregistered operation, even though it's always registered.

superlopuh:
In either case, this PR is missing a filecheck test for `module {}`

alexarice:
Ah, is this an MLIR bug and not an xdsl bug?

dshaaban01:
> I don't think this is the right fix. The problem we discussed is that `module {}`, which is a valid spelling of `builtin.module {}` in MLIR, is parsed as an unregistered operation, even though it's always registered.

But I never tried to parse `module{}`... The module I tried to parse was as follows:

```
builtin.module {
    func.func @hello(%n : i32) -> i32 {
    %b = arith.addi %n, %n : i32
    %c = arith.muli %b, %b : i32
    func.return %b : i32
    }
}
```

dshaaban01:
https://github.com/xdslproject/xdsl/issues/3668

Made an issue.

dshaaban01:
Closing this PR as @alexarice found the bug and opened a new PR to fix it.

PR #3667, opened 2024-12-22
interactive: mlir-gui version 1

dshaaban01:
This is a first attempt at creating an mlir-gui. I would like to upstream it (still figuring out whether to do that in this repo or in the mlir repo). Open to feedback/advice from anyone interested :)

It is very similar to xdsl-gui, with some changes: 
- only executes `mlir-opt --passes` , no xdsl-passes
- displays the list of available mlir-opt passes via parsing the output of `mlir-opt --help' (rather ugly way to get the mlir passes, if I upstream it, will make it more fancy)
- condense button not there anymore
- requires mlir-opt to be on your PATH
- call `mlir-gui` in terminal to get it :)

Screenshot 2024-12-22 at 16 34 11

Some open questions: 
- should it be connected to xdsl-gui with a button (users can flip between xdsl-gui app and mlir-gui app)? 
- does it belong in the xdsl repo?

Note: It is not perfect and has some errors if you search hard enough (especially with pass selection)! Still needs work, this is just version 1 :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,8 @@ xdsl-gui = "xdsl.interactive.app:main"
 xdsl-stubgen = "xdsl.utils.dialect_stub:make_all_stubs"
 xdsl-tblgen = "xdsl.tools.tblgen_to_py:main"
 
+mlir-gui = "xdsl.interactive.mlirapp:main"
+
 [tool.setuptools]
 platforms = ["Linux", "Mac OS-X", "Unix"]
 zip-safe = false
--- a/uv.lock
+++ b/uv.lock
@@ -2457,7 +2457,7 @@ wheels = [
 
 [[package]]
 name = "xdsl"
-version = "0+dynamic"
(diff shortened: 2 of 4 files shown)


superlopuh:
Hi @dshaaban01, any updates on this? If we want to merge it to the xDSL repo, I think it would be great to find a way to integrate it into xdsl-gui, hopefully via logic encapsulated on the MLIROptPass itself.

dshaaban01:
> Hi @dshaaban01, any updates on this? If we want to merge it to the xDSL repo, I think it would be great to find a way to integrate it into xdsl-gui, hopefully via logic encapsulated on the MLIROptPass itself.

Hi. I agree. I will work on this in my spare time - I am overloaded with other things until mid April, but will try to work on this sporadically until then, and then focus on it end of April/beginning of May. Hope this works. Will close this PR for now and will reopen smaller PR's to begin integration when the time comes.

PR #3669, opened 2024-12-22
core: Parse `module` correctly when `allow-unregistered-dialect` is set

### LINKED ISSUE #3668: core: `module{}` parsed as an unregistered operation when `MLContext(allow_unregistered=True)`
`module {}`, which is a valid spelling of builtin.module {} in MLIR, is parsed as an unregistered operation, even though it's always registered. This happens in particular when `MLContext(allow_unregistered=True)` instead of it's default, which is False.

alexarice:
Fixes #3668

### DIFF
--- a/tests/test_mlcontext.py
+++ b/tests/test_mlcontext.py
@@ -60,6 +60,40 @@ def test_get_op_unregistered():
     assert issubclass(ctx.get_op("test.dummy2"), UnregisteredOp)
 
 
+def test_get_op_with_dialect_stack():
+    """Test `get_op` and `get_optional_op` methods."""
+    ctx = MLContext()
+    ctx.load_op(DummyOp)
+
+    assert ctx.get_op("dummy", dialect_stack=("test",)) == DummyOp
+    with pytest.raises(Exception):
+        _ = ctx.get_op("dummy2", dialect_stack=("test",))
+
+    assert ctx.get_optional_op("dummy", dialect_stack=("test",)) == DummyOp
+    assert ctx.get_optional_op("dummy2", dialect_stack=("test",)) is None
+
+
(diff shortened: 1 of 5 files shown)


dshaaban01 (approved):
Amazing🥳 thank you!

superlopuh (commented):
This LGTM, but I'm wondering whether we could simplify things by passing the dialect stack to `get_optional_op` directly, avoiding the creation of extra ops?

alexarice:
> This LGTM, but I'm wondering whether we could simplify things by passing the dialect stack to `get_optional_op` directly, avoiding the creation of extra ops?

Sounds good, I knew there was something I didn't like about my solution

alexarice:
I've rewritten it with @superlopuh suggestion. This didn't quite fix the needless creation of unregistered ops so I've moved some things to a helper function. Also added 2 unit tests.

math-fehr (approved):
Nice!

tests/test_mlcontext.py
```diff
@@ -60,6 +60,40 @@ def test_get_op_unregistered():
...
+    """
+    Test `get_op` and `get_optional_op`
+    methods with the `allow_unregistered` flag.
+    """
+    ctx = MLContext(allow_unregistered=True)
+    ctx.load_op(DummyOp)
+
+    assert ctx.get_optional_op("dummy", dialect_stack=("test",)) == DummyOp
+    op_type = ctx.get_optional_op("dummy2", dialect_stack=("test",))
+    print(op_type)
```
superlopuh, line 86: Could you please open a mini PR removing this?

alexarice: #3673

PR #3676, opened 2024-12-25
testing: split bottom_up into f32 and f64 kernels

superlopuh:
There's still a bit of work to do on the f32 front, this PR just splits the test file, and removes erroneous f32 mentions from the f64 kernels.


mamanain (approved):
makes sense

PR #3677, opened 2024-12-25
transformations: (memref-streamify) don't streamify 0D memrefs

superlopuh:
There's no point in streaming 0D memrefs, it costs more to set up the streaming registers. In the future we might want to canonicalize away the 0D memrefs on memref_stream operations to loads/stores and generic on scalars, which will be lowered to registers.

Note stacked PR.


mamanain:
Don't you want to deal with this during linalg -> stream_memref lowering? Just replace it with load/op/store there.

superlopuh:
What's the advantage of doing it when transforming`linalg` to `memref_stream`? I would rather keep that pass as simple as possible, and do all the transformations/legalizations in downstream passes. We can transform the `memref<f64>` inputs to scalars in a separate canonicalization pass.

mamanain (approved):
LGTM

PR #3679, opened 2024-12-26
transforms: Jax buffer donation

mamanain:
Reuse donated buffers + remove return values if we can load data to buffer.

### DIFF
--- a/xdsl/transforms/jax_use_donated_arguments.py
+++ b/xdsl/transforms/jax_use_donated_arguments.py
@@ -1,4 +1,3 @@
-import copy
 import itertools
 from collections.abc import Sequence
 from dataclasses import dataclass
@@ -6,7 +5,7 @@ from dataclasses import dataclass
 from xdsl.context import MLContext
 from xdsl.dialects import builtin
 from xdsl.dialects.bufferization import MaterializeInDestinationOp
-from xdsl.dialects.builtin import FunctionType, TensorType
+from xdsl.dialects.builtin import ArrayAttr, DictionaryAttr, FunctionType, TensorType
 from xdsl.dialects.func import FuncOp, ReturnOp
 from xdsl.ir import BlockArgument, Operation, SSAValue
 from xdsl.passes import ModulePass
@@ -87,7 +86,7 @@ class SubstituteDonatedTensors(RewritePattern):
         if not donated_input_by_output:
             return
 
-        ordered_buffered_outputs = list(
+        ordered_buffered_outputs = tuple(
             filter(lambda out: out in donated_input_by_output, op.arguments)
         )
         new_ops: list[Operation] = [
@@ -100,7 +99,7 @@ class SubstituteDonatedTensors(RewritePattern):
             out: mater_ops.results[0]
             for out, mater_ops in zip(ordered_buffered_outputs, new_ops, strict=True)
         }
-        new_outputs = [new_output_mapping.get(out, out) for out in op.arguments]
+        new_outputs = tuple(new_output_mapping.get(out, out) for out in op.arguments)
 
         return_types = tuple(func_op.function_type.outputs.data)
         if self.remove_matched_outputs:
@@ -118,13 +117,15 @@ class SubstituteDonatedTensors(RewritePattern):
 
         # remove the donation attribute to avoid their reuse if we run the pass multiple times on the same function
         used_donated_arguments = set(donated_input_by_output.values())
-        new_input_attrs = copy.deepcopy(func_op.arg_attrs)
+        new_input_attrs = [attr.data.copy() for attr in func_op.arg_attrs]
 
         for inp, new_attr in zip(func_op.args, new_input_attrs, strict=True):
             if inp in used_donated_arguments:
-                del new_attr.data["tf.aliasing_output"]
+                del new_attr["tf.aliasing_output"]
 
-        func_op.arg_attrs = new_input_attrs
+        func_op.arg_attrs = ArrayAttr(
+            [DictionaryAttr(attr) for attr in new_input_attrs]
+        )
 
 
 @dataclass(frozen=True)
--- a/xdsl/transforms/jax_use_donated_arguments.py
+++ b/xdsl/transforms/jax_use_donated_arguments.py
@@ -72,13 +72,20 @@ class SubstituteDonatedTensors(RewritePattern):
         func_op = op.parent_op()
         assert isinstance(func_op, FuncOp)
 
-        if func_op.arg_attrs is None:
+        if func_op.arg_attrs is None or len(func_op.body.blocks) > 1:
             return
 
-        donated_inputs = tuple(
-            inp
-            for inp, attr in zip(func_op.args, func_op.arg_attrs, strict=True)
-            if isinstance(inp.type, TensorType) and "tf.aliasing_output" in attr.data
+        donated_inputs_idx: list[int]
+        donated_inputs: Sequence[BlockArgument]
+        donated_inputs_idx, donated_inputs = zip(
+            *[
+                (arg_idx, inp)
+                for arg_idx, (inp, attr) in enumerate(
+                    zip(func_op.args, func_op.arg_attrs, strict=True)
+                )
+                if isinstance(inp.type, TensorType)
+                and "tf.aliasing_output" in attr.data
+            ]
         )
 
         if not donated_inputs:
@@ -94,9 +101,9 @@ class SubstituteDonatedTensors(RewritePattern):
 
         return_types = tuple(func_op.function_type.outputs.data)
         if self.remove_matched_outputs:
-            kept_outputs_mask = [
+            kept_outputs_mask = tuple(
                 i not in output_input_mapping for i in range(len(op.arguments))
-            ]
+            )
             return_types = list(itertools.compress(return_types, kept_outputs_mask))
             new_outputs = list(itertools.compress(new_outputs, kept_outputs_mask))
 
@@ -105,6 +112,12 @@ class SubstituteDonatedTensors(RewritePattern):
             func_op.function_type.inputs.data, return_types
         )
         rewriter.replace_matched_op(new_ops)
+
+        # remove the donation attribute to avoid their reuse if we run the pass multiple times on the same function
+        for input_idx in output_input_mapping.values():
+            del func_op.arg_attrs.data[donated_inputs_idx[input_idx]].data[
+                "tf.aliasing_output"
+            ]
 
 
 @dataclass(frozen=True)
--- a/xdsl/transforms/jax_use_donated_arguments.py
+++ b/xdsl/transforms/jax_use_donated_arguments.py
@@ -6,7 +6,8 @@ from xdsl.dialects import builtin
 from xdsl.dialects.bufferization import MaterializeInDestinationOp
 from xdsl.dialects.builtin import FunctionType, TensorType
 from xdsl.dialects.func import FuncOp, ReturnOp
-from xdsl.ir import Operation, SSAValue
+from xdsl.ir import Attribute, BlockArgument, Operation, SSAValue
+from xdsl.irdl import VarOperand
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
     PatternRewriter,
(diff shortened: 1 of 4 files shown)


xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,92 @@
...
+        if func_op.arg_attrs is None:
+            return
+
+        donated_inputs = [
+            inp
+            for inp, attr in zip(func_op.args, func_op.arg_attrs, strict=True)
+            if isinstance(inp.type, TensorType) and "tf.aliasing_output" in attr.data
+        ]
+
+        if len(donated_inputs) == 0:
```
superlopuh, line 37:
```suggestion
        if not donated_inputs:
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,92 @@
...
+
+@dataclass(frozen=True)
+class JaxUseDonatedArguments(ModulePass):
+    name = "jax-use-donated-arguments"
+
+    remove_matched_outputs: bool = False
+
+    def apply(self, ctx: MLContext, op: builtin.ModuleOp) -> None:
+        the_one_pass = PatternRewriteWalker(
+            GreedyRewritePatternApplier(
```
superlopuh, line 85: No need for greedy applier with one pattern

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,92 @@
...
+                [SubstituteDonatedTensors(self.remove_matched_outputs)]
+            ),
+            apply_recursively=False,
+            walk_reverse=True,
+            walk_regions_first=True,
+        )
+        the_one_pass.rewrite_module(op)
```
superlopuh, line 92: Can just do this inline without the variable definition

superlopuh (commented):
Could probably simplify a part of the logic with itertools.compress

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    used_inputs_idx: set[int] = set()
+    output_input_mapping: dict[int, int] = {}
+
+    for outputs_idx, output in enumerate(outputs):
+        if isinstance(output.owner, MaterializeInDestinationOp) and isinstance(
+            output.owner.dest, BlockArgument
+        ):
+            # the return value is already bufferized in a donated buffer
+            continue
```
superlopuh, line 37: is this still necessary? With your changes, it should be safe to make the assumption that the result can be donated if the attribute is present.

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,144 @@
...
+        func_op.function_type = FunctionType.from_lists(
+            func_op.function_type.inputs.data, return_types
+        )
+        rewriter.replace_matched_op(new_ops)
+
+        # remove the donation attribute to avoid their reuse if we run the pass multiple times on the same function
+        for input_idx in output_input_mapping.values():
+            del func_op.arg_attrs.data[donated_inputs_idx[input_idx]].data[
+                "tf.aliasing_output"
+            ]
```
superlopuh, line 129: uh oh, this shouldn't be possible, attributes should be immutable, and this will mess with some guarantees around hashability and such. We should construct a new arg_attrs here and set it instead of mutating the existing one.

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,144 @@
...
+        donated_inputs_idx, donated_inputs = zip(
+            *[
+                (arg_idx, inp)
+                for arg_idx, (inp, attr) in enumerate(
+                    zip(func_op.args, func_op.arg_attrs, strict=True)
+                )
+                if isinstance(inp.type, TensorType)
+                and "tf.aliasing_output" in attr.data
+            ]
+        )
```
superlopuh, line 98:
```suggestion
        donated_input_mask = tuple(
            isinstance(inp.type, TensorType) and "tf.aliasing_output" in attr.data
            for arg_idx, (inp, attr) in enumerate(
                zip(func_op.args, func_op.arg_attrs, strict=True)
            )
        )
        
        donated_inputs = tuple(itertools.compress(func_op.args, donated_input_mask)
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,144 @@
...
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+
+
+def map_outputs_to_inputs(
+    donated_inputs: Sequence[BlockArgument], outputs: VarOperand
+) -> dict[int, int]:
```
superlopuh, line 23: It's not super clear by reading the function name what maps to what. I'm tepted to try mapping the ssa values directly so the type communicates in which direction the mapping goes, maybe something like this?
```suggestion
def donated_input_by_output(
    donatable_inputs: Sequence[BlockArgument], outputs: Sequence[SSAValue]
) -> dict[SSAValue, BlockArgument]:
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Find suitable donated buffers for each of returned variables.
+    Each buffer can be used only once.
+    Types of the buffer and the variable should match.
+    """
+    used_inputs_idx: set[int] = set()
+    output_input_mapping: dict[int, int] = {}
```
superlopuh, line 30: How do you feel about this implementation?
``` python
donatable_inputs_by_type = {k: tuple(g) for k, g in itertools.groupby(donated_inputs, lambda i: i.type)}
outputs_by_type = {k: tuple(g) for k, g in itertools.groupby(donated_inputs, lambda i: i.type)}
mapping_by_type = {
    k: tuple(zip(donatable_inputs_by_type[k], outputs_by_type[k])) 
    for k in donatable_inputs_by_type.keys() & outputs_by_type.keys()
}
return {
    o: i
    for mapping in mapping_by_type.values()
    for (i, o) in mapping
}
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,144 @@
...
+    for output_idx, input_idx in sorted(output_input_mapping.items()):
+        materialize_ops.append(
+            MaterializeInDestinationOp(
+                operands=[outputs[output_idx], donated_inputs[input_idx]],
+                result_types=[outputs[output_idx].type],
+            )
+        )
+        new_outputs[output_idx] = materialize_ops[-1].results[0]
+
+    return new_outputs, materialize_ops
```
superlopuh, line 68: If the mapping is changed to be of type `dict[SSAValue, BlockArgument]`, then each of the results of this helper can be written in a one-liner, and probably better inlined then. The rule of thumb for me is that a helper is useful to isolate some mutation or tricky state, and if we can eliminate the mutation or tricky state entirely then that seems like a good move. Sorry to yo-yo on this, sometimes the complexity only goes away after some iteration!

xdsl/transforms/jax_use_donated_arguments.py
mamanain: It still is we want to avoid doubly buffering outputs - if we have multiple donated buffers of the same type and output has already been buffered on the first run it will buffered again on the second run. Here is a [test](https://github.com/xdslproject/xdsl/pull/3679/files#diff-d7531353fdbb9dbfec680ebe4b414d0d1b6fd91d79107c03d646c0c64867a405R45) for it.

superlopuh: Ah yes excellent point

xdsl/transforms/jax_use_donated_arguments.py
mamanain: I don't think it works. groupby groups only consecutive elements so if arguments are not consecutive it will lead to only parts of them being processed.

example:
```
>>> a = ["A", "B", "AB", "C", "CD"]
>>> {k: tuple(g) for k, g in itertools.groupby(a, len)}
{1: ('C',), 2: ('CD',)}
```

superlopuh: hmm, do you think a fixed version would work, though? The main idea here is to avoid the double for loops, and to match things by type

mamanain: And I personally think that in this case for loops would be more readable compared to functional stuff. But maybe I'm just not python enough.

mamanain: But maybe it's ok. I'll just insert sorting by type in your implementation.

mamanain:
@superlopuh do you think this is more readable?

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,142 @@
...
+
+        donated_input_by_output = map_donated_input_by_output(
+            donated_inputs, op.arguments
+        )
+        if not donated_input_by_output:
+            return
+
+        ordered_buffered_outputs = list(
+            filter(lambda out: out in donated_input_by_output, op.arguments)
+        )
```
superlopuh, line 92:
```suggestion
        ordered_buffered_outputs = tuple(
            arg for arg in op.arguments if arg in donated_input_by_output
        )
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,142 @@
...
+        new_ops: list[Operation] = [
+            MaterializeInDestinationOp(
+                operands=[out, donated_input_by_output[out]], result_types=[out.type]
+            )
+            for out in ordered_buffered_outputs
+        ]
```
superlopuh, line 98:
```suggestion
        new_ops = tuple(
            MaterializeInDestinationOp(
                operands=[out, donated_input_by_output[out]], result_types=[out.type]
            )
            for out in ordered_buffered_outputs
        )
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,142 @@
...
+        new_output_mapping = {
+            out: mater_ops.results[0]
+            for out, mater_ops in zip(ordered_buffered_outputs, new_ops, strict=True)
+        }
+        new_outputs = [new_output_mapping.get(out, out) for out in op.arguments]
```
superlopuh, line 103:
```suggestion
        new_outputs = tuple(new_output_mapping.get(out, out) for out in op.arguments)
```

xdsl/transforms/jax_use_donated_arguments.py
mamanain: But I append the return op to it

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # remove the donation attribute to avoid their reuse if we run the pass multiple times on the same function
+        used_donated_arguments = set(donated_input_by_output.values())
+        new_input_attrs = copy.deepcopy(func_op.arg_attrs)
+
+        for inp, new_attr in zip(func_op.args, new_input_attrs, strict=True):
+            if inp in used_donated_arguments:
+                del new_attr.data["tf.aliasing_output"]
+
+        func_op.arg_attrs = new_input_attrs
```
superlopuh, line 127: This has the same problem, we must never mutate attributes. The move is probably to get the data dict, copy it, and then mutate it, finally creating the new attr array. I'll quickly open a PR to migrate DictionaryAttr to immutable dict, as we can have other problems from this being done in general.

xdsl/transforms/jax_use_donated_arguments.py
superlopuh: I see, it still feels a little neater to make this immutable, and insert these ops before the existing return, and then replace the existing return with the new return. This feels more like a nit to me, feel free to ignore also.

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,143 @@
...
+    Each buffer can be used only once.
+    Types of the buffer and the variable should match.
+    """
+
+    donatable_inputs_by_type = {
+        type: tuple(group)
+        for type, group in itertools.groupby(
+            sorted(donatable_inputs, key=lambda i: str(i.type)), lambda i: i.type
+        )
+    }
```
superlopuh, line 34: Yeah I think with the groupby deficiency something like this might be a little easier to read:
```suggestion
    donatable_inputs_by_type = defaultdict(list)
    for value in donatable_inputs:
        donatable_inputs_by_type[value.type].append(value)
    }
```

xdsl/transforms/jax_use_donated_arguments.py
```diff
@@ -0,0 +1,143 @@
...
+
+        donated_input_by_output = map_donated_input_by_output(
+            donated_inputs, op.arguments
+        )
+        if not donated_input_by_output:
+            return
+
+        ordered_buffered_outputs = tuple(
+            filter(lambda out: out in donated_input_by_output, op.arguments)
+        )
```
superlopuh, line 91: why the filter? filter and map are slower and imo less readable than comprehensions
```suggestion
        ordered_buffered_outputs = tuple(
            arg for arg in donated_input_by_output if arg in donated_input_by_output
        )
```

xdsl/transforms/jax_use_donated_arguments.py
superlopuh: oh also might as well iterate on `.items()` instead of fetching the value for key

superlopuh: Also having the type of this be exclusively `MaterializeInDestinationOp` will simplify the expression below, getting the operand and result directly instead of through the `Operation` interface.

xdsl/transforms/jax_use_donated_arguments.py
superlopuh: #3690

superlopuh (approved):
Thank you very much! Once the immutabledict things are sorted I think this is good to merge. I look forward to using this!

PR #3681, opened 2024-12-30
core: Allow to erase orphan operations

math-fehr:
Stacked PRs:
 * #3683
 * #3682
 * __->__#3681

--- --- ---

### DIFF
--- a/xdsl/rewriter.py
+++ b/xdsl/rewriter.py
@@ -69,10 +69,10 @@ class Rewriter:
         If safe_erase is True, check that the operation has no uses.
         Otherwise, replace its uses with ErasedSSAValue.
         """
-        assert op.parent is not None, "Cannot erase an operation that has no parents"
-
-        block = op.parent
-        block.erase_op(op, safe_erase=safe_erase)
+        if (block := op.parent) is not None:
+            block.erase_op(op, safe_erase=safe_erase)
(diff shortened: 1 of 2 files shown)


xdsl/rewriter.py
```diff
@@ -69,10 +69,10 @@ def erase_op(op: Operation, safe_erase: bool = True):
...
         Otherwise, replace its uses with ErasedSSAValue.
         """
-        assert op.parent is not None, "Cannot erase an operation that has no parents"
-
-        block = op.parent
-        block.erase_op(op, safe_erase=safe_erase)
+        if (block := op.parent) is not None:
+            block.erase_op(op, safe_erase=safe_erase)
+        else:
+            op.erase(safe_erase=safe_erase)
```
superlopuh, line 75: nit: swap around these to so the test is just `is` and not `is not`?

superlopuh (approved):

xdsl/rewriter.py
math-fehr: I can easily change it, but does this change anything?
Why would one be different?

superlopuh: IDK I feel like it's a bit easier to think about conditionals in the positive than in the negative, but it's a tiny personal preference.

PR #3687, opened 2025-01-02
transformations: (mlir-opt) separate mlir-opt pass errors

superlopuh:
We currently have two statements that can raise an error when calling out to `mlir-opt`, one for errors the executable raises, the other for parsing in xDSL, these should be reported separately.


alexarice (approved):
Great

Issue #3684, opened 2024-12-30
core: (Parsing) The int value of a BoolAttr that is `true` is `-1` instead of `1`


watermelonwolverine:
I'm not sure that this is a bug. Reading through the code and the comments it seems to me that true should equal to 1 and false should equal to 0 but this is not the case. During parsing true is initially read as 1 but it is later normalized to -1 by `IntegerType.normalized_value`. 

Might this an issue with normalization of i1 values?

superlopuh:
This is how signed integers work, and should maybe be documented better, but the largest negative number is 1 bigger than the largest positive number in signed integers, meaning if the positive range goes up to 0, the negative range goes down to -1.

watermelonwolverine:
The documentation in https://github.com/xdslproject/xdsl/blob/346e1d4d95a8a2eefc130d0ef1c93c7159b41765/xdsl/parser/base_parser.py#L42 and https://github.com/xdslproject/xdsl/blob/346e1d4d95a8a2eefc130d0ef1c93c7159b41765/xdsl/parser/base_parser.py#L72 need to be fixed.

I would also suggest adding something like
```python
    def to_bool(self) -> bool:
        return bool(self.value.data)
```
to IntegerAttr as a counterpart to `from_bool`. This looks too simple to be useful but I had several bugs in my project after upgrading to 0.25.0 because I worked on the assumption that true always equals 1 and false always equals 0 so I hastily implemented the checks based on that assumption. It's nice not having to make this assumption.

PR #3689, opened 2025-01-02
dialects: (builtin) Fix documentation for parsing of bools, and add conversion to Python bool

watermelonwolverine:
Updated documentation in attribute_parser and base_parser regarding the integer values of bools.
Added `to_bool` to IntegerAttr
See issue #3684


xdsl/dialects/builtin.py
```diff
@@ -705,6 +705,11 @@ def constr(
             ),
         )
 
+    def to_bool(self) -> bool:
+        # Do assertion, yes, no?
+        assert self.type == i1
+        return bool(self.value.data)
```
superlopuh, line 711:
```suggestion
    def __bool__(self) -> bool:
        """Returns True if value is non-zero."""
        return bool(self.value.data)
```

tests/dialects/test_builtin.py
```diff
@@ -167,6 +167,14 @@ def test_IntegerAttr_normalize():
         IntegerAttr(256, 8)
 
 
+def test_IntegerAttr_to_bool():
+    assert IntegerAttr.from_bool(False).to_bool() is False
+    assert IntegerAttr.from_bool(True).to_bool() is True
+
+    with pytest.raises(AssertionError):
+        IntegerAttr(1, i8).to_bool()
+
```
superlopuh, line 176: Yeah I don't think we need to be this strict, I'm happy to follow C/Python's truthiness as being non-zero.

superlopuh:
Why not `__bool__`?

superlopuh:
This is an antipattern and should never be used, for precisely this reason :) I would like to move towards more functions using attributes for calculations instead of python integers, and this is a valid use for them, I would argue that we want this functionality for them also. I would recommend making the change, and then waiting for @math-fehr and @alexarice's reviews before merging.

watermelonwolverine:
> This is an antipattern and should never be used, for precisely this reason :) I would like to move towards more functions using attributes for calculations instead of python integers, and this is a valid use for them, I would argue that we want this functionality for them also. I would recommend making the change, and then waiting for @math-fehr and @alexarice's reviews before merging.

A test failed because there was such an if. This could potentially break a lot of code of people who use xDSL.

alexarice:
Whether or not it is an antipattern, I feel that we have used conversion to bool in place of "is not None" in a lot of places and I'm not sure there's a good way to find all the places.

math-fehr:
I would personally not overload __bool__, and would probably name the function `is_nonnull` instead?

xdsl/dialects/builtin.py
```diff
@@ -294,6 +294,10 @@ def print_parameter(self, printer: Printer) -> None:
         with printer.in_angle_brackets():
             printer.print_string(f"{self.data}")
 
+    def __bool__(self) -> bool:
+        """Returns True if value is non-zero."""
+        return bool(self.data)
+
```
superlopuh, line 300: would be good to add unit tests for this for good measure

superlopuh:
@alexarice I've been pretty consistent in my PR reviews to discourage this approach, and I feel that we're actually in a good place here, modulo the two uses of it in csl.

@math-fehr, why not? I don't see why we shouldn't be able to use integer attrs as idiomatic Python integers, whether it's IntAttr or IntegerAttr.

alexarice:
My memory is that we've changed a few `if X is not None` to `if X` in PRs, but maybe I'm misremembering. I still feel that bugs could creep in but am happy for the change to go ahead if this is the direction we want to go in.

superlopuh:
If I had anything to do with it it would have been the other way around, which is hopefully why all the tests pass with this change.

watermelonwolverine:
One thing you might want to consider when going with `__bool__` is that there are some non-obvious pitfalls, for example you can't use:

```python
if bool_attr is True:
    # Do stuff
```

or

```python
if bool_attr is False:
    # Do stuff
```

You could argue that this code is simply wrong but it's a code a beginner might write (or someone who likes to make things very explicit). Basically the programmers would have to know when they can use a IntegerAttr like a bool and in what situations they shouldn't. Now, whether you really want to cater to beginner programmers is another question.

As another side note: What about adding a function `is_present` that checks if an optional attribute, region, operand is present? This would make things very explicit and would even open the door to changing the underlying implementation of optionals at some later point. I.e. using something other than `None`. At the moment it would simple check if the given value is not `None`.

superlopuh:
Pyright should flag that as the types have no overlap, so I don't think it's a significant concern.

tests/dialects/test_builtin.py
```diff
@@ -167,6 +167,16 @@ def test_IntegerAttr_normalize():
         IntegerAttr(256, 8)
 
 
+def test_IntAttr___bool__():
+    assert not IntAttr(0)
+    assert IntAttr(1)
+
+
+def test_IntegerAttr___bool__():
+    assert not IntegerAttr.from_bool(False)
```
alexarice, line 176:
```suggestion
    assert not BoolAttr.from_bool(False)
```
Super nit and probably doesn't really matter

tests/dialects/test_builtin.py
```diff
@@ -167,6 +167,16 @@ def test_IntegerAttr_normalize():
...
+    assert IntegerAttr.from_bool(True)
```
alexarice, line 177:
```suggestion
    assert BoolAttr.from_bool(True)
```

xdsl/parser/base_parser.py
```diff
@@ -45,7 +45,7 @@ def parse_optional_integer(
         """
         Parse an (possible negative) integer. The integer can either be
         decimal or hexadecimal.
-        Optionally allow parsing of 'true' or 'false' into 1 and 0.
+        Optionally allow parsing of 'true' or 'false' into -1 and 0.
```
alexarice, line 48: This does not agree with the implementation below

xdsl/parser/base_parser.py
```diff
@@ -78,7 +78,7 @@ def parse_integer(
         """
         Parse an (possible negative) integer. The integer can
         either be decimal or hexadecimal.
-        Optionally allow parsing of 'true' or 'false' into 1 and 0.
+        Optionally allow parsing of 'true' or 'false' into -1 and 0.
```
alexarice, line 81: Same here

xdsl/parser/base_parser.py
```diff
@@ -121,7 +121,7 @@ def parse_optional_number(
     ) -> int | float | None:
         """
         Parse a (possibly negative) integer or float literal, if present.
-        Can optionally parse 'true' or 'false' into 1 and 0.
+        Can optionally parse 'true' or 'false' into -1 and 0.
```
alexarice, line 124: Same here

xdsl/parser/base_parser.py
```diff
@@ -149,7 +149,7 @@ def parse_number(
     ) -> int | float:
         """
         Parse a (possibly negative) integer or float literal.
-        Can optionally parse 'true' or 'false' into 1 and 0.
+        Can optionally parse 'true' or 'false' into -1 and 0.
```
alexarice, line 152: and here

alexarice (commented):
Some small comments

xdsl/parser/base_parser.py
superlopuh: It does though

superlopuh: Ah actually not here

superlopuh: You're right, all of the functions with the changed documentation here return the 1/0 values, it's only when creating the attribute that the value is different. This should  be reflected in documentation for the attribute, not the parser.

superlopuh:
@watermelonwolverine Do you have some time to address Alex's comments?

superlopuh (approved):
Thank you!

compor (approved):
awesome, thank you for this!

PR #3690, opened 2025-01-02
dialects: (builtin) make DictionaryAttr data an immutabledict

superlopuh:
It's not safe to mutate attributes, and one of the most frequently used attributes currently has a mutable core. This PR replaces the dict with an immutabledict, that we already depend on.

### DIFF
--- a/xdsl/dialects/scf.py
+++ b/xdsl/dialects/scf.py
@@ -150,7 +150,7 @@ class WhileOp(IRDLOperation):
         op = cls(arguments, function_type.outputs.data, before_region, after_region)
 
         if attrs is not None:
-            op.attributes = attrs.data
+            op.attributes |= attrs.data
 
         return op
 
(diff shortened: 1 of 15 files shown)


alexarice (approved):
Seems fine though might be nicer to overload the constructor for `DictionaryAttr` to be able to take a dictionary or immutabledict, which should reduce the diff a bit.

xdsl/dialects/scf.py
```diff
@@ -150,7 +150,7 @@ def parse_assignment():
         op = cls(arguments, function_type.outputs.data, before_region, after_region)
 
         if attrs is not None:
-            op.attributes = attrs.data
+            op.attributes |= attrs.data
```
math-fehr, line 153: Is this related to this PR?

math-fehr (approved):

xdsl/dialects/scf.py
superlopuh: Yeah, as attrs.data is now an immutabledict, this could've been `= dict(...)` but this seems cleaner

PR #3694, opened 2025-01-03
interpreter: (riscv_cf) register riscv_cf functions

superlopuh:
Forgot to add this previously.


xdsl/interpreters/__init__.py
```diff
@@ -46,6 +47,7 @@ def register_implementations(
     interpreter.register_implementations(riscv_func.RiscvFuncFunctions())
     interpreter.register_implementations(riscv_libc.RiscvLibcFunctions())
     interpreter.register_implementations(riscv_scf.RiscvScfFunctions())
+    interpreter.register_implementations(riscv_cf.RiscvCfFunctions())
```
alexarice, line 50: Nit: The rest are in alphabetical order

alexarice (approved):

alexarice:
Further thought, would it be good to have a filecheck test for each set of interpreter functions?

superlopuh:
Yes but I haven't figured out how to do it in a way that lets us check the individual behaviour of each function, and not the code as a whole

PR #3698, opened 2025-01-06
core: Let inferrable constraints solve variables

alexarice:
Allows an inferrable type to solve other variables, and further allows range variable constraints to be solved if the underlying constraint is inferrable. Includes a test with the motivating example.

Currently this is restricted to when the underlying constraint can be inferred using no variables, and so is likely only useful for the scenario in the test. Anything more involved would be a fair bit more complex.

I also have a branch somewhere with a draft implementation of "length variables", which would probably achieve a similar thing to this PR (and could potentially be extended to be more powerful). Would be good to hear people's thoughts on whether this would be preferable instead of this (slightly hacky) solution.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format_parser.py
+++ b/xdsl/irdl/declarative_assembly_format_parser.py
@@ -28,6 +28,7 @@ from xdsl.irdl import (
     OptSuccessorDef,
     ParamAttrConstraint,
     ParsePropInAttrDict,
+    RangeConstraint,
     VarExtractor,
     VariadicDef,
     VarOperandDef,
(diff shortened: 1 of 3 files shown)


xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -247,6 +248,25 @@ def extract_var(self, a: ParsingState) -> ConstraintVariableType:
                 types = (types,)
             return self.inner.extract_var(types)
 
+    @dataclass(frozen=True)
+    class _OperandResultInferenceExtractor(VarExtractor[ParsingState]):
+        idx: int
+        is_operand: bool
+        inner: VarExtractor[Sequence[Attribute]]
+        constr: RangeConstraint
```
superlopuh, line 256:
```suggestion
    class _OperandResultInferenceExtractor(VarExtractor[ParsingState]):
        idx: int
        KW_ONLY
        is_operand: bool
        inner: VarExtractor[Sequence[Attribute]]
        constr: RangeConstraint
```

superlopuh: should help read the code below

alexarice: I've never seen this before

superlopuh: https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY

superlopuh: I misremembered the API

alexarice: I truly hate that API but agree it would make the code better

superlopuh:
Overall I think it looks good, and it feels like it would be worth having the support. It feels like it would be worth adding some unit tests with explicit NotImplemented catches with corresponding issues to highlight what's missing.

alexarice:
> Overall I think it looks good, and it feels like it would be worth having the support. It feels like it would be worth adding some unit tests with explicit NotImplemented catches with corresponding issues to highlight what's missing.

I'm going to get the length constraint version into shape and can compare to that.

math-fehr (approved):

alexarice:
Closing in favour of #3699

PR #3699, opened 2025-01-06
core: Introduce integer constraints

alexarice:
This would probably make #3698 obsolete, as it covers the one use case of that PR that I can think of. This is a bit more of an involved change but I feel that there are some interesting a powerful things you could do with this in the future.

Would likely need a few more tests if we want to go with this.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -17,6 +17,7 @@ from xdsl.dialects.builtin import (
     FloatAttr,
     IndexType,
     IntegerAttr,
+    IntegerType,
     MemRefType,
     ModuleOp,
     UnitAttr,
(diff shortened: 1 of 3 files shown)


math-fehr (approved):
Sorry for the delay, that looks good to me!
Hopefully we can encode some constraints for variadics with it ;)

alexarice:
@math-fehr the idea was that this is an alternative to #3698, did you have any preferences between the two.

kimxworrall:
I vote this one as it allows you to have different input and output types in the example given.

compor (approved):
I prefer this one

alexarice:
I'll merge this PR in a couple of hours unless anyone else has comments

math-fehr:
Sorry, I am overwhelmed work-wise recently, so I missed your question!
I do prefer that one :)

PR #3703, opened 2025-01-06
core: Add BlockInsertPoint to simplify the builder API

math-fehr:
Stacked PRs:
 * #3705
 * #3704
 * __->__#3703

--- --- ---

### core: Add BlockInsertPoint to simplify the builder API

`BlockInsertPoint` acts the same as `InsertPoint`, but for blocks.
Its equivalent in MLIR is `Block::iterator`.

### DIFF
--- a/xdsl/rewriter.py
+++ b/xdsl/rewriter.py
@@ -58,6 +58,56 @@ class InsertPoint:
         return InsertPoint(block)
 
 
+@dataclass(frozen=True)
+class BlockInsertPoint:
+    """
+    An insert point for a block.
+    It is either a point before a block, or after a block.
+    """
+
+    region: Region
+    """The region where the insertion point is in."""
+
+    insert_before: Block | None = field(default=None)
+    """
+    The insertion point is right before this block.
+    If the block is None, the insertion point is at the end of the region.
+    """
+
+    def __post_init__(self) -> None:
+        # Check that the insertion point is valid.
+        # An insertion point can only be invalid if `insert_before` is a `Block`,
+        # and its parent is not `region`.
+        if self.insert_before is not None:
+            if self.insert_before.parent is not self.region:
+                raise ValueError("Insertion point must be in the builder's `region`")
+
+    @staticmethod
+    def before(block: Block) -> BlockInsertPoint:
+        """Gets the insertion point before a block."""
+        if (region := block.parent) is None:
+            raise ValueError("Block insertion point must have a parent region")
+        return BlockInsertPoint(region, block)
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


xdsl/rewriter.py
```diff
@@ -58,6 +58,56 @@ def at_end(block: Block) -> InsertPoint:
...
+    If the block is None, the insertion point is at the end of the region.
+    """
+
+    def __post_init__(self) -> None:
+        # Check that the insertion point is valid.
+        # An insertion point can only be invalid if `insert_before` is a `Block`,
+        # and its parent is not `region`.
+        if self.insert_before is not None:
+            if self.insert_before.parent is not self.region:
+                raise ValueError("Insertion point must be in the builder's `region`")
```
superlopuh, line 83: unit test for this?

xdsl/rewriter.py
```diff
@@ -58,6 +58,56 @@ def at_end(block: Block) -> InsertPoint:
...
+
+    @staticmethod
+    def before(block: Block) -> BlockInsertPoint:
+        """Gets the insertion point before a block."""
+        if (region := block.parent) is None:
+            raise ValueError("Block insertion point must have a parent region")
```
superlopuh, line 89: unit test for this?

alexarice:
Could `InsertPoint` and `BlockInsertPoint` be a generic class instead? They superficially look very similar except for some types.

compor (approved):

compor:
LGTM modulo the missing tests.

math-fehr:
> Could `InsertPoint` and `BlockInsertPoint` be a generic class instead? They superficially look very similar except for some types.

I can make them generic, I am just a bit scared of what the documentation will look like.
As you want, I can make it generic now, or later as well

alexarice:
Maybe just leave it as is. The documentation is important

alexarice (approved):
Couple of points that don't need to be dealt with in this PR:
- Should InsertPoint be renamed to OpInsertPoint?
- Is it possible/preferable to make the initializer private and force the use of the static methods to construct both `InsertPoint`s and `BlockInsertPoint`s, letting the runtime check be dropped?

math-fehr:
>  Should InsertPoint be renamed to OpInsertPoint?

Yes, I think we could rename `InsertPoint` to `OpInsertPoint` later on (even if MLIR names it `InsertPoint`).

>  Is it possible/preferable to make the initializer private and force the use of the static methods to construct both `InsertPoint`s and `BlockInsertPoint`s, letting the runtime check be dropped?

I never found an actual way to make Python constructors private, which otherwise yes that would make things much better.

PR #3709, opened 2025-01-06
interpreter: add type property to ShapedArray

superlopuh:
Previously not exposed.

### DIFF
--- a/xdsl/interpreters/shaped_array.py
+++ b/xdsl/interpreters/shaped_array.py
@@ -26,7 +26,7 @@ class ShapedArray(Generic[_T]):
     shape: list[int]
 
     @property
-    def type(self) -> PackableType[_T]:
+    def element_type(self) -> PackableType[_T]:
         return self._data.xtype
 
     @property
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/shaped_array.py
```diff
@@ -25,6 +25,10 @@ class ShapedArray(Generic[_T]):
     _data: TypedPtr[_T]
     shape: list[int]
 
+    @property
+    def type(self) -> PackableType[_T]:
```
jorendumoulin, line 29: is this not get_element_type?

compor: Good point, that's that we use in `builtin.py` and the camelcase version of this is used in MLIR (e.g., `TypeUtilities.cpp` and friends)

compor (approved):

xdsl/interpreters/shaped_array.py
superlopuh: This is the interpreter utility class, not the attribute. I guess we might want type to return some kind of ShapedType in the future, I'll rename this element_type for now.

PR #3710, opened 2025-01-07
core: print attribute names without double quotes for bare identifiers

jorendumoulin:
This PR makes sure that attribute names that are bare identifiers (matching `bare_identifier_regex`) are printed without double quotes, as it is done in MLIR.

The actual change is quite minimal (in `xdsl/printer.py`), all other diffs come from all of the filechecks and other tests that change.

most changes were made automatically with [this script](https://gist.github.com/jorendumoulin/855686adb8197ee207fb3224a480cfd6), using the following conservative regex:
`r'(\/\/.*\{( *|[^[}\n]*, *))(\"[a-zA-Z_][a-zA-Z0-9_$.]*\")'`

the remaining ones were done using a vim search and replace with manual confirmation
`:%s/"\([a-zA-Z_][a-zA-Z0-9_$.]*\)"/\1/gc`


tests/filecheck/mlir-conversion/with-mlir/affinet_set.mlir
compor, line 1: great, I'm not sure if we want a specific (failing?) interoperability test when the attribute name is not a bare identifier, but we don't seem to have anything other similar I think.

jorendumoulin: I am not sure what you mean by this. `set` is a bare identifier, according to the rules at https://github.com/llvm/llvm-project/blob/ca603d2536f039194141bf3a01e9ee7f60e37406/mlir/lib/IR/AsmPrinter.cpp#L2226
and defined in xDSL by `bare_identifier_regex`

alexarice (commented):
This is great.
Would it be possible to get a simple mlir integration test that ensures that xdsl and mlir print some simple attributes in the same way?

tests/filecheck/mlir-conversion/with-mlir/affinet_set.mlir
compor: I was referring to integration tests as @alexarice mentions below, not specifically this attribute, it's just happened to be the only one changed that interoperates with `mlir-opt`. Sorry if it wasn't clear.

jorendumoulin: Okay! I have added some interop tests in my last commit

tests/filecheck/mlir-conversion/with-mlir/parser-printer/attribute_names.mlir
```diff
@@ -0,0 +1,50 @@
+// RUN: xdsl-opt %s | filecheck %s
```
alexarice, line 1:
```suggestion
// RUN: XDSL_ROUNDTRIP
```

tests/filecheck/mlir-conversion/with-mlir/parser-printer/attribute_names.mlir
```diff
@@ -0,0 +1,50 @@
...
+// RUN: mlir-opt %s | filecheck %s
+
+// CHECK: module
+
+"test.op"() {"name"} : () -> ()
+// CHECK-NOT: "name"
+// CHECK-NEXT: name
```
alexarice, line 8: Perhaps include the braces here?

alexarice (approved):
Very small nits. Would it also be possible to add an example which is not a unit attr to the test just so we have full coverage.

tests/filecheck/mlir-conversion/with-mlir/parser-printer/attribute_names.mlir
alexarice: Occured to me that this might not actually do the same thing. Perhaps ignore this suggestion

tests/filecheck/mlir-conversion/with-mlir/parser-printer/attribute_names.mlir
jorendumoulin: ah yes that's much easier

superlopuh (approved):
I'm very glad this is finally done, thank you very much for the initiative!

alexarice (approved):

PR #3712, opened 2025-01-07
documentation: add helpers for marimo things to be shared across notebooks

superlopuh:
We want to have more than one notebook, and a consistent way of showing things in them. I think a helpers file would be a good way to do that.

The main question is where we want the code to live, I think `xdsl.utils.marimo` makes sense, but I'm open to suggestions.

### DIFF
--- a/docs/marimo/linalg_snitch.py
+++ b/docs/marimo/linalg_snitch.py
@@ -11,6 +11,12 @@ def _():
 
 
 @app.cell
+def _():
+    import xdsl.utils.marimo as xmo
+    return (xmo,)
+
+
+@app.cell
 def _(mo):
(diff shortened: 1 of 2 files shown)


math-fehr:
xdsl.docs.marimo?
Is there a reason we want to put them all in the same place btw? Compared to order them by doc, tutorial, etc..?

superlopuh:
Hmm, it's really just reusable UI elements, not the full notebooks. The notebooks will still live in the `docs` folder, and you're right that they should be grouped together in some way. This seems like an orthogonal proposal, unless I'm misunderstanding you.

math-fehr (approved):

PR #3719, opened 2025-01-07
documentation: (marimo) move pipeline_html helper to utils

superlopuh:
Creates a unified way of presenting the intermediate stages of compilation in the marimo notebooks, together with optional descriptions of what's happening. I'm not super happy with the way it's presented right now, but we can iterate on it once it's in, and all the notebooks will benefit from it at the same time. I'm relatively sure that this API is all we need, but we can change it later also.

### DIFF
--- a/xdsl/utils/marimo.py
+++ b/xdsl/utils/marimo.py
@@ -33,8 +33,8 @@ def _spec_str(p: ModulePass) -> str:
 
 
 def pipeline_html(
-    ctx: MLContext, passes: Sequence[tuple[mo.Html, ModulePass]], module: ModuleOp
-) -> tuple[ModuleOp, mo.Html]:
+    ctx: MLContext, module: ModuleOp, passes: Sequence[tuple[mo.Html, ModulePass]]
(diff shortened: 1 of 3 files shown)


xdsl/utils/marimo.py
```diff
@@ -36,10 +36,15 @@ def pipeline_html(
     ctx: MLContext, module: ModuleOp, passes: Sequence[tuple[mo.Html, ModulePass]]
 ) -> tuple[MLContext, ModuleOp, mo.Html]:
     """
-    Returns a tuple of the resulting module after applying the passes, and the
+    Returns a tuple of the resulting context and module after applying the passes, and the
     Marimo-optimised representation of the modules throughout compilation.
```
compor, line 40: what does "marimo-optimized" mean?

compor (approved):

xdsl/utils/marimo.py
superlopuh: I meant "prettier to view than `str(bla)`", but I was conscious of it not being a great way to phrase this idea. I couldn't find a term that they used themselves for this concept that isn't just the type of the return value.

PR #3721, opened 2025-01-07
transformations: (memref-stream-interleave) always take bigger factor

superlopuh:
When I initially implemented this pass, I assumed that it was better to err on the side of caution, to take the smallest factor above 3. In practice, this isn't the best approach, as a factor of 4 is only good in the scenario with no function cache misses, and it's a safer bet to take bigger values if possible. We also learned from the Snitch backend paper experiments that the register pressure never goes close to the limit for the kernels we care about, so this should not be that risky.

My understanding is that this change is what @zero9178 implemented in Quidditch already.

I'm currently working on automatically pruning the space of possible values, and merging this PR would help keep the logic relatively simple, while sharing some helpers between this heurist-based pass, and the new compile-and-evaluate-based-pass.

### DIFF
--- a/xdsl/transforms/memref_stream_interleave.py
+++ b/xdsl/transforms/memref_stream_interleave.py
@@ -150,7 +150,7 @@ class MemrefStreamInterleavePass(ModulePass):
     The pass will select the largest factor of the corresponding bound smaller than
     `pipeline-depth * 2`.
     The search range is bound by `pipeline-depth * 2` as very large interleaving factors
-    would cause too much register pressure, potentially running out of registers.
+    can increase register pressure and potentially exhaust all available registers.
     In the future, it would be good to take the number of available registers into account
     when choosing a search range, as well as inspecting the generic body for
     read-after-write dependencies.
--- a/xdsl/transforms/memref_stream_interleave.py
+++ b/xdsl/transforms/memref_stream_interleave.py
@@ -51,18 +51,10 @@ class PipelineGenericPattern(RewritePattern):
 
         interleave_factor = 1
         # Search factors until the next number divisible by pipeline_depth
-        for potential_factor in range(self.pipeline_depth, self.pipeline_depth * 2):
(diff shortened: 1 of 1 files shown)


zero9178 (approved):
LGTM!

Yes this was roughly what I used as well if I recall correctly. The only thing to keep in mind is the much higher register pressure, but as you said, for the kernels we care about we are unlikely to run out of registers. 
Something to be aware about in the future however. Maybe this should be a note in the pass description?

superlopuh:
Would you suggest something scarier than this?

> The search range is bound by `pipeline-depth * 2` as very large interleaving factors would cause too much register pressure, potentially running out of registers.

zero9178:
> Would you suggest something scarier than this?
> 
> > The search range is bound by `pipeline-depth * 2` as very large interleaving factors would cause too much register pressure, potentially running out of registers.

Looks good to me :)

xdsl/transforms/memref_stream_interleave.py
```diff
@@ -155,8 +147,8 @@ class MemrefStreamInterleavePass(ModulePass):
     Tiles the innermost parallel dimension of a `memref_stream.generic`.
     If specified, the `pipeline-depth` parameter specifies the number of operations in the
     resulting body that should be executed concurrently.
-    The pass will search through possible factors, starting with `pipeline-depth`, going
-    up to `pipeline-depth * 2`, and then down to 1, stopping at the first one.
+    The pass will select the largest factor of the corresponding bound smaller than
+    `pipeline-depth * 2`.
     The search range is bound by `pipeline-depth * 2` as very large interleaving factors
     would cause too much register pressure, potentially running out of registers.
```
compor, line 153:
```suggestion
    The search range is bound by `pipeline-depth * 2` as very large interleaving factors
    can potentially increase register pressure or even exhaust all available registers.
```

I know this isn't directly related, but the wording has been annoying me for a bit.
Take it or leave it :)

compor (approved):

PR #3724, opened 2025-01-08
[WIP] experimental memref_stream unroll-and-jam pass using interpreter-based cost model

superlopuh:
Most of the juicy stuff here is in the autotune.py marimo notebook, best experienced by checking out this branch and running `uv run marimo edit docs/marimo`.

The main idea here is to have a system that proposes a bunch of rewrites, and then evaluates each of these rewrites based on an additional lowering + interpreter tracing.


qaco:
In the marimo notebook, when I click on ```autotune.py```, I have the following error : ```"f-string expression part cannot include a backslash (autotune.py, line 516)"```

qaco:
It's nice. I like the "lazy pass" pattern. And I like the fact that the unroll and jam is confined at linalg level (rather than materializing loops).

In the case of this cost model, it's minimizing the number of jumps/blt (thanks to unrolling) that reduces the cost, right?

superlopuh:
Yep, basically. It should be relatively easy to make the cost model a bit smarter and to add latency of instructions.

superlopuh:
Let's get back to this when we next try autotuning.

PR #3725, opened 2025-01-08
interactive: AvailablePass now has either a pass type or instance, no spec

superlopuh:
Same deal as #3723, we used to need the spec due to potential mutability, now don't need it.

### DIFF
--- a/tests/interactive/test_get_all_available_passes.py
+++ b/tests/interactive/test_get_all_available_passes.py
@@ -14,7 +14,6 @@ from xdsl.pattern_rewriter import (
     op_type_rewrite_pattern,
 )
 from xdsl.transforms.individual_rewrite import ApplyIndividualRewritePass
-from xdsl.utils.parse_pipeline import PipelinePassSpec
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


xdsl/interactive/app.py
```diff
@@ -347,10 +347,12 @@ def expand_node(
...
         expanded_pass.remove_children()
 
-        for pass_name, value, value_spec in child_pass_list:
+        for pass_name, value in child_pass_list:
             expanded_pass.add(
                 label=pass_name,
-                data=(value, value_spec),
+                data=(type(value), value.pipeline_pass_spec())
+                if isinstance(value, ModulePass)
+                else (value, None),
```
superlopuh, line 355: This is a newly introduced ugly bit due to the tree node data still being in terms of pass type and spec, next PR will remove this.

n-io (approved):

tests/interactive/test_get_all_available_passes.py
```diff
@@ -82,25 +81,15 @@ def test_get_all_available_passes():
         (
             AvailablePass(
                 display_name="bc",
-                module_pass=BCPass,
-                pass_spec=None,
+                module_pass=BCPass(),
```
n-io, line 84: nit: It looks like AvailablePass can still be built with a type?

compor (approved):

dshaaban01 (approved):

tests/interactive/test_get_all_available_passes.py
superlopuh: Yep, this is used when "Condense Mode" is not on, and for mlir-opt pass as we want to present the pass as an option to the user without instantiating it

PR #3726, opened 2025-01-08
interactive: App tree holds either a pass type or instance, no spec

superlopuh:
Last PR of the batch, with this the app only deals with pass types or instances, no more specs.


dshaaban01 (approved):
Looks good

PR #3728, opened 2025-01-08
dialects: (builtin) add support for IntegerType packing with arbitrary bitwidth up to 64

jorendumoulin:


alexarice:
While the bug I hit was with `i2`, I'm unsure why we can't support at least all `IntegerType`s up to 64?

alexarice:
Can we not just store each integer in `bitwidth // 8` bytes?

jorendumoulin:
yes, I think that makes sense. for larger than 64 bits we would need something else than struct.pack however

jorendumoulin:
> Can we not just store each integer in `bitwidth // 8` bytes?

it should be ceil(bitwidth/8) though

tests/dialects/test_builtin.py
```diff
@@ -73,13 +73,14 @@ def test_FloatType_formats():
...
-    with pytest.raises(NotImplementedError):
-        IntegerType(2).format
     assert IntegerType(1).format == "<b"
+    assert IntegerType(2).format == "<b"
     assert IntegerType(8).format == "<b"
     assert IntegerType(16).format == "<h"
     assert IntegerType(32).format == "<i"
     assert IntegerType(64).format == "<q"
+    with pytest.raises(NotImplementedError):
+        IntegerType(65).format
```
superlopuh, line 83: Let's add these sanity checks?
```suggestion
    assert IntegerType(1).format == "<b"
    assert IntegerType(2).format == "<b"
    assert IntegerType(8).format == "<b"
    assert IntegerType(15).format == "<h"
    assert IntegerType(16).format == "<h"
    assert IntegerType(31).format == "<i"
    assert IntegerType(32).format == "<i"
    assert IntegerType(63).format == "<q"
    assert IntegerType(64).format == "<q"
    with pytest.raises(NotImplementedError):
        IntegerType(65).format
```

xdsl/dialects/builtin.py
```diff
@@ -532,14 +532,14 @@ def print_value_without_type(self, value: int, printer: Printer):
 
     @property
     def format(self) -> str:
-        match self.bitwidth:
-            case 1 | 8:
+        match -(self.bitwidth // -8):  # ceildiv
```
superlopuh, line 535: ?
```suggestion
        match (1 + self.bitwidth) >> 3:
```

alexarice: This isn't correct

alexarice: `(7 + self.bitwidth) >> 3` works I think

jorendumoulin: `(self.bitwidth + 7) >> 3` is also used on L389 to compute the byte size of a `FixedBitWidthType`, so i think that one makes sense

superlopuh (approved):

PR #3731, opened 2025-01-09
misc: add __init__.py file to wgpu backend folder

superlopuh:
A downstream client of xDSL currently can't import it.


alexarice (approved):
Is checking for these files precommitable/ci-able?

PR #3732, opened 2025-01-09
dependencies: (wgpu) remove wgpu dependency

superlopuh:
The code is moved to https://github.com/xdslproject/xdsl-webgpu/

Clients wanting to execute their gpu kernels using wgpu will need to install `xdsl-webgpu` instead.


alexarice (approved):
No results from `rg wgpu` :+1:

PR #3745, opened 2025-01-13
dialects (arm): add LabelAttr

emmau678:
Add LabelAttr to ARM dialect.

### DIFF
--- a/tests/test_assembly_arg_str.py
+++ b/tests/test_assembly_arg_str.py
@@ -1,6 +1,10 @@
+import pytest
+
 from xdsl.dialects.arm.assembly import assembly_arg_str
 from xdsl.dialects.arm.attrs import LabelAttr
 from xdsl.dialects.arm.register import IntRegisterType
+from xdsl.dialects.test import TestType
+from xdsl.utils.test_value import TestSSAValue
 
 
 def test_assembly_arg_str_ARMRegister():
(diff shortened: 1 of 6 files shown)


xdsl/dialects/arm/assembly.py
```diff
@@ -2,11 +2,31 @@
...
+    @classmethod
+    def parse_parameter(cls, parser: AttrParser) -> str:
+        with parser.in_angle_brackets():
+            return parser.parse_str_literal()
+
+    def print_parameter(self, printer: Printer) -> None:
+        with printer.in_angle_brackets():
+            printer.print_string_literal(self.data)
+
+
```
superlopuh, line 29: this should be in the dialect

superlopuh: Maybe we need a new `.attrs` file with the attribute definitions

emmau678: I have moved it to a separate .attrs file. thank you

xdsl/dialects/arm/ops.py
superlopuh, line 1: let's do this one in a new PR?

xdsl/dialects/arm/attrs.py
superlopuh, line 1: could you please add labels to arm assembly printing as one of `AssemblyInstructionArg`, and add a small unit test for the `assembly_arg_str` function?

xdsl/dialects/arm/ops.py
emmau678: ok, I have removed it from this one.

compor:
Maybe update the PR title and change to non-draft when ready? This is only going to have the attribute right?

tests/test_assembly_arg_str.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_assembly_arg_str_ARMRegister():
+    arg = IntRegisterType("x0")
+    assert assembly_arg_str(arg) == "x0"
+
+
+def test_assembly_arg_str_LabelAttr():
+    arg = LabelAttr("main")
+    assert assembly_arg_str(arg) == "main"
```
emmau678, line 13: not really sure how I would test for other SSAValue types and invalid ones

compor: Maybe use `TestSSAValue` with a `TestType` type? if you search for these in the repo, you'll find examples.
You're trying to cover the `else` when the SSAValue is not of `ARMRegisterType` right?

emmau678: yes, that's exactly what I was looking for, thank you @compor. I have updated the test now

compor (approved):

PR #3748, opened 2025-01-13
CI: Create missing virtual enviroment in the publish job

math-fehr:
The virtual environment was missing from the CI

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -16,12 +16,10 @@ jobs:
     - name: Set up Python
       run: uv python install 3.11
 
-    - name: Create the virtual environment
+    - name: Build and publish
       run: |
-        # Change directory so that xdsl-opt can be found during installation.
-        cd xdsl
-        VENV_EXTRAS="--all-extras" make venv
-
+        uvx --with setuptools --with wheel setup.py sdist bdist_wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/pythonpublish.yml
```diff
@@ -16,6 +16,12 @@ jobs:
     - name: Set up Python
       run: uv python install 3.11
 
+    - name: Create the virtual environment
+      run: |
+        # Change directory so that xdsl-opt can be found during installation.
+        cd xdsl
+        VENV_EXTRAS="--all-extras" make venv
+
```
superlopuh, line 24:
```suggestion
    - name: Build and publish
      run: |
        uvx --with setuptools --with wheel setup.py sdist bdist_wheel
        uvx twine upload -u __token__ -p ${{ secrets.PYPI_XDSL_TOKEN }} dist/*
```
I had a quick read of the docs here and it looks like the lines below can be done like this, the point is to not use the same venv as the xdsl one for these one-off scripts. BTW, do we want to reuse the uv cache for this?

https://docs.astral.sh/uv/concepts/tools/#the-path

math-fehr: I wouldn't use the cache here, we run this once every few weeks or so, so we don't need it?

superlopuh (approved):

PR #3751, opened 2025-01-14
core: (traits) Add `SameOperandsAndResultType` trait

compor:
This PR:

- Adds the `SameOperandsAndResultType` trait
- Tests of the above

A [few](https://github.com/llvm/llvm-project/blob/539b15b41a6a01017c0a555e89b7d2b62ba194d2/mlir/test/IR/traits.mlir#L41) examples of this being tested in upstream MLIR.

### DIFF
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -12,16 +12,26 @@ import pytest
 
 from xdsl.dialects import test
 from xdsl.dialects.builtin import (
+    DYNAMIC_INDEX,
     AnyIntegerAttr,
+    AnyMemRefTypeConstr,
+    AnyTensorTypeConstr,
+    AnyUnrankedMemrefTypeConstr,
+    AnyUnrankedTensorTypeConstr,
     IntegerAttr,
     IntegerType,
+    MemRefType,
+    NoneAttr,
     StringAttr,
     SymbolRefAttr,
+    TensorType,
+    UnrankedTensorType,
     i1,
     i32,
     i64,
 )
-from xdsl.ir import Operation, OpTrait, OpTraits
+from xdsl.ir import Attribute, Operation, OpTrait, OpTraits
+from xdsl.ir.core import SSAValue
 from xdsl.irdl import (
     Block,
     IRDLOperation,
@@ -37,6 +47,7 @@ from xdsl.irdl import (
     result_def,
     traits_def,
 )
+from xdsl.irdl.operations import var_operand_def, var_result_def
 from xdsl.traits import (
     AlwaysSpeculatable,
     ConditionallySpeculatable,
@@ -44,6 +55,7 @@ from xdsl.traits import (
     HasParent,
     OptionalSymbolOpInterface,
     RecursivelySpeculatable,
+    SameOperandsAndResultType,
     SymbolOpInterface,
     SymbolTable,
     is_speculatable,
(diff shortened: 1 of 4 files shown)


math-fehr:
Quick question. Looking at it, it seems that this is more complex than just the `SameOperandsAndResultType` in MLIR, right?
MLIR only checks that the operands and results are the same, but here you seem to check across containers? What does the check do exactly? I would probably rename the trait to not confuse it with the MLIR one. (if I'm correct to what MLIR does)

compor:
> Quick question. Looking at it, it seems that this is more complex than just the `SameOperandsAndResultType` in MLIR, right? MLIR only checks that the operands and results are the same, but here you seem to check across containers? What does the check do exactly? I would probably rename the trait to not confuse it with the MLIR one. (if I'm correct to what MLIR does)

Is it? I thought it also checks within the container type?
https://github.com/llvm/llvm-project/blob/8d9dcd111e5ced8135387917859dd64d67886be0/mlir/lib/IR/Operation.cpp#L1100

math-fehr:
Okay my bad then, I indeed misunderstood what MLIR was doing!

math-fehr:
Thanks for the pointer!

alexarice:
Does this achieve anything that you can't do with range constraints? Is it just so you can write the same thing as you can in mlir?

alexarice:
I think i've misunderstood what this does then, I'll have a proper look tomorrow

tests/test_traits.py
```diff
@@ -549,6 +561,440 @@ class SupeculatabilityTestOp(IRDLOperation):
...
+            * operands_num,
+        ],
+        result_types=[[MemRefType(result_element_type, result_shape)] * results_num],
+    )
+
+    with pytest.raises(
+        VerifyException,
+        match="requires the same type for all operands and results",
+    ):
+        op.verify()
```
superlopuh, line 755: these seem duplicated, unless I'm missing something

xdsl/utils/type.py
```diff
@@ -0,0 +1,41 @@
...
+def have_compatible_shape(lhs_type: Attribute, rhs_type: Attribute) -> bool:
+    is_lhs_shaped = isinstance(lhs_type, ContainerType)
+    is_rhs_shaped = isinstance(rhs_type, ContainerType)
+
+    # both are scalars
+    if not is_lhs_shaped and not is_rhs_shaped:
+        return True
+
+    # one is scalar and the other shaped
+    if (is_lhs_shaped and not is_rhs_shaped) or (not is_lhs_shaped and is_rhs_shaped):
```
superlopuh, line 27:
```suggestion
    if is_lhs_shaped != is_rhs_shaped:
```

xdsl/traits.py
```diff
@@ -729,3 +729,60 @@ def get_insn(self, op: Operation) -> str:
...
+            raise VerifyException(
+                f"'{op.name}' requires at least one result or operand"
+            )
+
+        res_type0 = get_element_type_or_self(op.result_types[0])
+
+        def get_encoding(maybe_shaped_type: Attribute) -> Attribute:
+            if isinstance(maybe_shaped_type, TensorType):
+                return maybe_shaped_type.encoding
+            return NoneAttr()
```
superlopuh, line 752: I would recommend moving this to a static method

superlopuh (approved):

xdsl/traits.py
compor: hmm, yeah, static method or put it under `utils/type.py`?

superlopuh: Your call, in either case if it's public would be nice to add a unit test for it also

alexarice:
Ah, I see what this does now, I feel the name is confusing but I guess that's what mlir uses

xdsl/utils/type.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_element_type_or_self(maybe_shaped_type: Attribute) -> Attribute:
+    if isinstance(maybe_shaped_type, ContainerType):
+        container_type = cast(ContainerType[Any], maybe_shaped_type)
+        return container_type.get_element_type()
+    return maybe_shaped_type
+
+
+def have_compatible_shape(lhs_type: Attribute, rhs_type: Attribute) -> bool:
+    is_lhs_shaped = isinstance(lhs_type, ContainerType)
```
alexarice, line 19: would prefer `is_lhs_container` or invert it and have `is_lhs_scalar`

alexarice:
> My feeling is that `have_compatible_shape` is either impossible, or very very annoying to write in xDSL. So here, range constraints are not enough AFAIK

We should figure out how to do shapes properly with constraints/irdl at some point

tests/test_traits.py
compor: One tests a `TensorType`, that could potentially have an encoding) and the other a `MemRefType` that doesn't have an encoding.
I'll add a comment.

superlopuh: Ah yes I didn't see the difference, maybe could be a parametrized pytest?

xdsl/utils/type.py
compor: Horn clause fail, thanks!

xdsl/traits.py
```diff
@@ -729,3 +729,60 @@ def get_insn(self, op: Operation) -> str:
...
+            if res_type0 != oprnd_type_elem or not have_compatible_shape(
+                op.result_types[0], oprnd_type
+            ):
+                raise VerifyException(
+                    f"'{op.name} requires the same type for all operands and results"
+                )
+
+            elem_encoding = NoneAttr()
+            if isinstance(oprnd_type, TensorType):
+                elem_encoding = oprnd_type.encoding
```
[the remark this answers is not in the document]
compor: forgot to refactor, thanks!

tests/test_traits.py
compor: I'll give it a go

compor: I couldn't do it by adding a `parametrize` on just the container type (`TensorType`, `MemRefType`, etc.) and taking advantage of the Cartesian product with the ctor arguments because they are generics, so just denoting the class type as `TensorType[Any]` did not let me instantiate it in the test method.

Not sure if I'm missing something, but I couldn't find anything similar in the rest of the codebase.
The closest I found were the tests for bufferization and they just instantiate the type in the `parametrize` decorator.

superlopuh (approved):

PR #3754, opened 2025-01-15
documentation: add mkdocs-based documentation

superlopuh:
It is time to add a docs website to xDSL. I'd like for our docs to be as pretty as [marimo's](https://docs.marimo.io/api/), with similar interactive elements and pretty styling. I'm not super confident in web-things, so would love some help with styling and deploying to either docs.xdsl.dev or xdsl.dev/docs using GH pages.

This PR adds mkdocs, adds some make commands to build and serve the docs locally, and adds some barebones docs for the core.py file in xDSL.
In the future, it would be nice to do something like MLIR's automatic building of docs for dialects and other parts of the codebase that are more structured.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,3 +163,13 @@ coverage-report: uv-installed
 .PHONY: coverage-clean
 coverage-clean: uv-installed
 	uv run coverage erase
+
+# docs
+
+.PHONY: docs-serve
+docs-serve: uv-installed
+	uv run mkdocs serve
+
+.PHONY: docs-build
+docs-build: uv-installed
+	uv run mkdocs build
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ VENV_DIR ?= .venv
 export UV_PROJECT_ENVIRONMENT=$(if $(VIRTUAL_ENV),$(VIRTUAL_ENV),$(VENV_DIR))
 
 # allow overriding which extras are installed
-VENV_EXTRAS ?= --extra gui --extra dev --extra jax --extra riscv
+VENV_EXTRAS ?= --extra gui --extra dev --extra jax --extra riscv --extra docs
 
 # default lit options
 LIT_OPTIONS ?= -v --order=smart
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1,0 +1,5 @@
+# Api Reference
+
+| Module               | Name                     |
+|----------------------|--------------------------|
+| [`ir`](ir/index.md)  | IR Data Structures       |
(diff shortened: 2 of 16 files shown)


tobiasgrosser:
Is Mariko using mkdocs?

superlopuh:
Yep, I took some of the modules they use, but not all of them, once we get past the basics of API reference we might want to look a bit more closely at bits of the docs they do particularly well to try and replicate. I'd especially love embedding mini marimo notebooks to let users play with the APIs on the relevant doc page directly like in the [slider doc](https://docs.marimo.io/api/inputs/range_slider/)

tobiasgrosser:
That looks amazing, indeed.

EdmundGoodman:
Looks awesome!

I think deployment shouldn't be too painful as `mkdocs` has a subcommand for github pages deployment which can just be run in an action and everything should just work... (Something like these <https://squidfunk.github.io/mkdocs-material/publishing-your-site/#with-github-actions> / <https://github.com/EdmundGoodman/hpc-multibench/blob/main/.github/workflows/docs.yml>)

Am happy to help out but have lots of deadlines coming up in the next two weeks so can't do anything much right now

alexarice:
Are the `.md` files auto generated?

superlopuh:
Nope I hand-wrote them. I'm not sure how much of them makes sense to auto-generate. Obviously we want to do that for the dialects, but I don't know about the rest of the project. Marimo seems to use some helpers to make the actual doc generation easier, I'm not yet sure how they work and whether we want to do something similar.

alexarice:
I have no issues with any of the styling or web stuff

alexarice:
I see that if I add a docstring to the `Dialect` class, then it does appear in this documentation, which is great. I guess the core of my issue is that I feel I want the default to be that something like `dialect.md` is made, instead of having to make it myself for every dialect.

superlopuh:
I found a thing that can do this, but would rather set it up later, as our project structure promises to be messy enough that this would probably need some special casing and polish to not be a huge mess.

superlopuh:
https://mkdocstrings.github.io/recipes/

alexarice (approved):
Think this will be great, there's slightly more discussion on zulip

alexarice:
Small nit, can acronyms be capitalised here?
image

superlopuh:
Yeah that was annoying to me too, I think we'll want to lowercase the autogen things, and manually specify titles for the manual things. I propose we have two weeks of sprinting the docs then can have a session filing issues to improve minor things like this.

math-fehr (approved):
That's a great start yep!

PR #3757, opened 2025-01-16
CI: remove version override in uv

superlopuh:
UV now supports dynamic versions, by ommitting them from the lockfile! No more need for our workaround.

https://github.com/astral-sh/uv/pull/10622

After this PR is merged, all xDSL contributors must update to uv version 0.5.20 or higher.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2459,7 +2459,6 @@ wheels = [
 
 [[package]]
 name = "xdsl"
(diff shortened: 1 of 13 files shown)


uv.lock
superlopuh, line 1: No more version in our lockfile

math-fehr (approved):

PR #3760, opened 2025-01-17
[WIP]: Initialise arm_cf and begin beq implementation

emmau678:
Add arm_cf dialect and implement beq instruction

### DIFF
--- a/tests/filecheck/dialects/arm_cf/test_ops.mlir
+++ b/tests/filecheck/dialects/arm_cf/test_ops.mlir
@@ -1,0 +1,19 @@
+// RUN: XDSL_ROUNDTRIP
+// RUN: XDSL_GENERIC_ROUNDTRIP
+// RUN: xdsl-opt -t arm-asm %s | filecheck %s --check-prefix=CHECK-ASM
+
+// CHECK: %x1 = arm.get_register : !arm.reg<x1>
+%x1 = arm.get_register : !arm.reg<x1>
+
+// CHECK: %x2 = arm.get_register : !arm.reg<x2>
+%x2 = arm.get_register : !arm.reg<x2>
+
+// CHECK: arm.label "testlabel" {comment = "this is a label"}
+// CHECK-ASM: testlabel:                                       # this is a label
+arm.label "testlabel" {comment = "this is a label"}
+
+// CHECK-NEXT: arm_cf.beq %x1, %x2, ^testlabel
+// CHECK-ASM: arm_cf.beq x1, x2, ^testlabel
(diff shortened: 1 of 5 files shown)


emmau678:
I've been stuck on this error for a while now so just pushing this as WIP in case anyone could help speed things up: ```KeyError: 'testlabel'```. Advice would be much appreciated if anyone can see where I've gone wrong.

tests/filecheck/dialects/arm_cf/test_ops.mlir
```diff
@@ -0,0 +1,19 @@
...
+// CHECK: %x2 = arm.get_register : !arm.reg<x2>
+%x2 = arm.get_register : !arm.reg<x2>
+
+// CHECK: arm.label "testlabel" {comment = "this is a label"}
+// CHECK-ASM: testlabel:                                       # this is a label
+arm.label "testlabel" {comment = "this is a label"}
+
+// CHECK-NEXT: arm_cf.beq %x1, %x2, ^testlabel
+// CHECK-ASM: arm_cf.beq x1, x2, ^testlabel
+arm_cf.beq %x1, %x2, ^testlabel : (!arm.reg<x1>, !arm.reg<x2>)
```
alexarice, line 17: I think the issue is you've that you've referred to a block that doesn't exist

xdsl/dialects/arm_cf.py
```diff
@@ -0,0 +1,65 @@
...
+            attributes={
+                "comment": comment,
+            },
+            successors=(then_block),
+        )
+
+    def assembly_line_args(self):
+        then_label = self.then_block.first_op
+        assert isinstance(then_label, arm.ops.LabelOp)
+        return (self.s1, self.s2, then_label.label.data)
```
superlopuh, line 57: It would be good to allow both str and StringAttr here
```suggestion
        return (self.s1, self.s2, then_label.label)
```

tests/filecheck/dialects/arm_cf/test_ops.mlir
superlopuh: This is a good opportunity to fix the error message in a separate PR!

emmau678: PR for fixing error message: https://github.com/xdslproject/xdsl/pull/3762

xdsl/dialects/arm_cf.py
superlopuh: It would be great to address this before merging, I'm actually leaning towards not allowing strings here at all, requiring the label attribute directly.

PR #3761, opened 2025-01-17
dialects: (math) Use `SameOperandsAndResultType` trait

compor:
This PR:

- Uses the `SameOperandsAndResultType` trait in the `math` dialect
- Simplifies relevant test cases which were not using the `return_type` parameter

### DIFF
--- a/tests/dialects/test_math.py
+++ b/tests/dialects/test_math.py
@@ -1,5 +1,3 @@
-from typing import TypeVar
-
 import pytest
 
 from xdsl.dialects.arith import ConstantOp, FloatingPointLikeBinaryOperation
@@ -45,11 +43,8 @@ from xdsl.dialects.experimental.math import (
     TruncOp,
 )
 from xdsl.dialects.test import TestOp
-from xdsl.ir import Attribute
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.test_value import TestSSAValue
-
-_BinOpArgT = TypeVar("_BinOpArgT", bound=Attribute)
 
 
 class Test_float_math_binary_construction:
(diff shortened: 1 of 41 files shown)


xdsl/dialects/experimental/math.py
```diff
@@ -32,13 +39,13 @@ class AbsFOp(IRDLOperation):
...
     result = result_def(AnyFloatConstr)
 
-    traits = traits_def(Pure())
+    traits = traits_def(Pure(), SameOperandsAndResultType())
 
     def __init__(
         self, operand: Operation | SSAValue, fastmath: FastMathFlagsAttr | None = None
     ):
         operand = SSAValue.get(operand)
-        return super().__init__(
```
compor, line 41: Piggybacked this change as pylint was complaining.

compor:
No sure I understand the CI lockfile fail, but I'll have a look later

superlopuh (approved):
Looks great! What's blocking us from moving math out of experimental?

compor:
> Looks great! What's blocking us from moving math out of experimental?

I don't think anything, really, that was my goal with this.
I was just thinking a separate NFC PR for the move, but we can avoid the churn and do this here if you want.

tests/dialects/test_math.py
```diff
@@ -45,12 +43,9 @@
     TruncOp,
 )
 from xdsl.dialects.test import TestOp
-from xdsl.ir import Attribute
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.test_value import TestSSAValue
 
-_BinOpArgT = TypeVar("_BinOpArgT", bound=Attribute)
```
compor, line 52: This is unused. Not sure what the initial outlook here was.

superlopuh:
Nice, I think separate PR is good.

PR #3762, opened 2025-01-17
core: Add check for blocks referenced without implementation

emmau678:
Make the error message more specific where an unknown successor is encountered.

### DIFF
--- a/tests/filecheck/parser-printer/parse_error.mlir
+++ b/tests/filecheck/parser-printer/parse_error.mlir
@@ -14,3 +14,11 @@ test.op : () -> ()
 // CHECK-NEXT: test.op : () -> ()
 // CHECK-NEXT:         ^
 // CHECK-NEXT:         Operation test.op does not have a custom format.
+
+// -----
+
+module {
+    "test.op"() [^unknown_successor]: () -> ()
+}
+
+// CHECK: Unknown location of span region ends with missing block declarations for block(s) unknown_successor.
--- a/tests/filecheck/parser-printer/parse_error.mlir
+++ b/tests/filecheck/parser-printer/parse_error.mlir
@@ -21,4 +21,4 @@ module {
     "test.op"() [^unknown_successor]: () -> ()
 }
 
-// CHECK: Unknown location of span region ends with missing block declarations for block(s) unknown_successor.
+// CHECK: reference to block "^unknown_successor" without implementation
--- a/xdsl/parser/core.py
+++ b/xdsl/parser/core.py
@@ -145,10 +145,7 @@ class Parser(AttrParser):
             value_names = ", ".join(
                 "%" + name for name in self.forward_ssa_references.keys()
             )
-            if len(self.forward_ssa_references.keys()) > 1:
-                self.raise_error(f"values {value_names} were used but not defined")
-            else:
-                self.raise_error(f"value {value_names} was used but not defined")
+            self.raise_error(f"values used but not defined: [{value_names}]")
 
         return module_op
 
--- a/xdsl/parser/core.py
+++ b/xdsl/parser/core.py
@@ -100,7 +100,7 @@ class Parser(AttrParser):
         super().__init__(ParserState(MLIRLexer(Input(input, name))), ctx)
         self.ssa_values = dict()
         self.blocks = dict()
-        self.forward_block_references = dict()
+        self.forward_block_references = defaultdict(list)
         self.forward_ssa_references = dict()
 
     def parse_module(self, allow_implicit_module: bool = True) -> ModuleOp:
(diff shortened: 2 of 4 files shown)


tests/filecheck/parser-printer/parse_error.mlir
```diff
@@ -14,3 +14,11 @@ test.op : () -> ()
...
 // CHECK-NEXT:         ^
 // CHECK-NEXT:         Operation test.op does not have a custom format.
+
+// -----
+
+module {
+    "test.op"() [^unknown_successor]: () -> ()
+}
+
+// CHECK: Unknown location of span region ends with missing block declarations for block(s) unknown_successor.
```
compor, line 24: "span" is parser terminology, no? Same with "unkown location".

The block name could be in quotes - maybe have a look how we do it in other block references within error messages.

alexarice: This error message is definitely not correct.

alexarice (commented):
Can't quite debug what's gone wrong but the error message should be at the end of the region and should point to each time a block is used without being defined.

superlopuh:
@emmau678 will you have time this week to address the comments on this PR?

emmau678:
> @emmau678 will you have time this week to address the comments on this PR?

I will attempt it today, I’m finding it a bit hard to identify how to fix it though.

superlopuh:
Please message with questions if anything isn't clear, it's much better to iterate quickly, otherwise we don't know if our comment was too vague. It would be great to reply to the comments with either the answer or a clarifying question, for example.

tests/filecheck/parser-printer/parse_error.mlir
emmau678: why is it a problem to use parser terminology here (since it's a parse error)?

emmau678: how would I know what the correct error might be? (this one is the one that's currently being printed when the code is run)

xdsl/parser/core.py
```diff
@@ -100,7 +100,7 @@ def __init__(
         super().__init__(ParserState(MLIRLexer(Input(input, name))), ctx)
         self.ssa_values = dict()
         self.blocks = dict()
-        self.forward_block_references = dict()
+        self.forward_block_references = defaultdict(list)
```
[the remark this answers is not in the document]
emmau678: it creates automatic values for non-existent keys so you don't get a key error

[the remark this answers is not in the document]
alexarice: This seems reasonably now that I understand the code a bit more.

xdsl/parser/core.py
```diff
@@ -564,7 +564,7 @@ def parse_optional_region(
             region.add_block(block)
 
         # Finally, check that all forward block references have been resolved.
-        if len(self.forward_block_references) > 0:
+        if self.forward_block_references:
             pos = self.lexer.pos
             raise MultipleSpansParseError(
                 Span(pos, pos + 1, self.lexer.input),
```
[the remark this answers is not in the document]
emmau678: ok, thank you @alexarice. I will make some updates this afternoon (I am mostly busy until then)

superlopuh:
let's take a quick look at this next week? Am I right in remembering that the blocker for this was merged in the last PR?

compor:
> let's take a quick look at this next week? Am I right in remembering that the blocker for this was merged in the last PR?

I think a better name for this PR and clear test/use case would be great, as I now don't even remember what was the initial intention with this.

xdsl/utils/exceptions.py
```diff
@@ -110,6 +110,12 @@ def __str__(self) -> str:
             res += span.print_with_context(msg) + "\n"
         return res
 
+    def __str__(self) -> str:
+        res = ""
+        for span, msg in self.refs:
+            res += span.print_with_context(msg) + "\n"
+        return res
+
```
superlopuh, line 118: seems like this should just be reverted

xdsl/parser/core.py
```diff
@@ -145,7 +145,7 @@ def parse_module(self, allow_implicit_module: bool = True) -> ModuleOp:
             value_names = ", ".join(
                 "%" + name for name in self.forward_ssa_references.keys()
             )
-            if len(self.forward_block_references.keys()) > 1:
+            if len(self.forward_ssa_references.keys()) > 1:
```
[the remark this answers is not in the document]
superlopuh: This is a regression wrt the change merged in #3851

tests/filecheck/parser-printer/parse_error.mlir
superlopuh: I'm surprised that this is the error printed as the location should now be known.

superlopuh (approved):
Thank you for fixing this, and all the bits of infrastructure that also needed fixing!

compor (approved):

PR #3768, opened 2025-01-20
dialects: (math) Move `math` dialect out of experimental [NFC]

compor:
This PR:

- Move `math` dialect out of the experimental module (as a result of #3761)


superlopuh (approved):
🎓

PR #3769, opened 2025-01-20
installation: (nix) update flake

alexarice:
Updates ruff to 0.5.21 in flake


compor:
What's the etiquette for nix stuff? I don't run it, but do you need a nix user to run this or are you happy with a random green tick if none is around?

superlopuh (approved):

PR #3771, opened 2025-01-20
documentation: code block in doc comments

superlopuh:
Makes doc comments prettier and improves rendering in online docs (upcoming).

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -1704,7 +1704,9 @@ class AddOp(RdRsRsOperation[IntRegisterType, IntRegisterType, IntRegisterType]):
     Adds the registers rs1 and rs2 and stores the result in rd.
     Arithmetic overflow is ignored and the result is simply the low XLEN bits of the result.
 
+    ```
     x[rd] = x[rs1] + x[rs2]
+    ```
 
     https://msyksphinz-self.github.io/riscv-isadoc/html/rvi.html#add
     """
@@ -1946,7 +1948,7 @@ class JalrOp(RdRsImmJumpOperation):
     """
     Jump to address and place return address in rd.
 
-    ```
+    ``` C
     t = pc+4
     pc = (x[rs1] + sext(offset)) & ~1
     x[rd] = t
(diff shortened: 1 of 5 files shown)


xdsl/dialects/riscv.py
```diff
@@ -1704,7 +1704,9 @@ class AddOp(RdRsRsOperation[IntRegisterType, IntRegisterType, IntRegisterType]):
     Adds the registers rs1 and rs2 and stores the result in rd.
     Arithmetic overflow is ignored and the result is simply the low XLEN bits of the result.
 
+    ```
```
alexarice, line 1707: Is this code?

xdsl/dialects/riscv.py
```diff
@@ -1946,7 +1948,7 @@ class JalrOp(RdRsImmJumpOperation):
     """
     Jump to address and place return address in rd.
 
-    ```
+    ``` C
```
alexarice, line 1951: Does this not have to be 
```
```C
```
To work correctly?

xdsl/dialects/riscv.py
alexarice: If it is should it also be C code as below?

xdsl/dialects/x86/ops.py
```diff
@@ -1293,7 +1353,8 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
 
 class M_M_Operation(Generic[R1InvT], IRDLOperation, X86Instruction, ABC):
     """
-    A base class for x86 operations with a memory reference that's both a source and a destination
+    A base class for x86 operations with a memory reference that's both a source and a
```
alexarice, line 1356: Unrelated formatting change?

alexarice (approved):

xdsl/dialects/riscv.py
superlopuh: The code below is C pseudocode, it's not currently syntax coloured, but at least the syntax colouring there would be correct. Here there's not really a close in syntax thing that would make sense IMO

xdsl/dialects/riscv.py
superlopuh: hmm, not in the markdown editors I use but you might be right

xdsl/dialects/x86/ops.py
superlopuh: Yep I accidentally included it but it makes the lines fit into 88 columns so I think still a good thing to merge.

compor (approved):
I believe you, but it would be nice to get a screenshot?

superlopuh:
I added one to the other PR, soon enough we'll have docs for our PRs so you won't have to take my word for it!

PR #3781, opened 2025-01-23
dialects: (llvm) Add a bunch of float methods

### LINKED ISSUE #3780: Implement missing ops for translating sympy to llvm
Missing ops:

- [ ] `llvm.sitofp` (can use `arith.sitofp`)
- [x] `llvm.fdiv` (can use `arith.divf`)
- [x] `llvm.sdiv` (can use `arith.divsi`)
- [ ] `llvm.fmul` (can use `arith.mulf`)
- [ ] `llvm.fadd` (can use `arith.addf`)
- [ ] `llvm.fmul`(can use `arith.mulf`)
- [ ] `llvm.fdiv` (can use `arith.divf`)
- [ ] `llvm.fpext` no replacement
- [X] `llvm.bitcast` no replacement(?)

Will add more as I find more, will cross out as I merge the PRs

AntonLydike:
Adds the following llvm floating point operations:

- `llvm.fadd`
- `llvm.fsub`
- `llvm.fmul`
- `llvm.fdiv`
- `llvm.frem`
- `llvm.sitofp`
- `llvm.fpext`

Closes #3780 

I also took the liberty to sort the dialect lists alphabetically

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/arithmetic.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/arithmetic.mlir
@@ -1,47 +1,68 @@
 // RUN: mlir-opt %s --mlir-print-op-generic | xdsl-opt | filecheck %s
 
 builtin.module {
-    %arg0, %arg1 = "test.op"() : () -> (i32, i32)
+    %arg0, %arg1, %f1 = "test.op"() : () -> (i32, i32, f32)
+    // CHECK: [[arg0:%\d+]], [[arg1:%\d+]], [[f1:%\d+]]
 
     %add = llvm.add %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.add %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.add [[arg0]], [[arg1]] : i32
 
     %add2 = llvm.add %arg0, %arg1 {"nsw"} : i32
-    // CHECK: %{{.*}} = llvm.add %{{.*}}, %{{.*}} {nsw} : i32
+    // CHECK: llvm.add [[arg0]], [[arg1]] {nsw} : i32
 
     %sub = llvm.sub %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.sub %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.sub [[arg0]], [[arg1]] : i32
 
     %mul = llvm.mul %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.mul %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.mul [[arg0]], [[arg1]] : i32
 
     %udiv = llvm.udiv %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.udiv %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.udiv [[arg0]], [[arg1]] : i32
 
     %sdiv = llvm.sdiv %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.sdiv %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.sdiv [[arg0]], [[arg1]] : i32
 
     %urem = llvm.urem %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.urem %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.urem [[arg0]], [[arg1]] : i32
 
     %srem = llvm.srem %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.srem %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.srem [[arg0]], [[arg1]] : i32
 
     %and = llvm.and %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.and %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.and [[arg0]], [[arg1]] : i32
 
     %or = llvm.or %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.or %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.or [[arg0]], [[arg1]] : i32
 
     %xor = llvm.xor %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.xor %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.xor [[arg0]], [[arg1]] : i32
 
     %shl = llvm.shl %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.shl %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.shl [[arg0]], [[arg1]] : i32
 
     %lshr = llvm.lshr %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.lshr %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.lshr [[arg0]], [[arg1]] : i32
 
     %ashr = llvm.ashr %arg0, %arg1 : i32
-    // CHECK: %{{.*}} = llvm.ashr %{{.*}}, %{{.*}} : i32
+    // CHECK: llvm.ashr [[arg0]], [[arg1]] : i32
+
+    // float arith:
+
+    %fmul = llvm.fmul %f1, %f1 : f32
+    // CHECK: llvm.fmul [[f1]], [[f1]] : f32
+
+    %fmul_fast = llvm.fmul %f1, %f1 {test = true, fastmathFlags = #llvm.fastmath<fast>} : f32
+    // CHECK: llvm.fmul [[f1]], [[f1]] {test = true, fastmathFlags = #llvm.fastmath<fast>} : f32
+
+    %fdiv = llvm.fdiv %f1, %f1 : f32
+    // CHECK: llvm.fdiv [[f1]], [[f1]] : f32
+
+    %fadd = llvm.fadd %f1, %f1 : f32
+    // CHECK: llvm.fadd [[f1]], [[f1]] : f32
+
+    %fsub = llvm.fsub %f1, %f1 : f32
+    // CHECK: llvm.fsub [[f1]], [[f1]] : f32
+
+    %frem = llvm.frem %f1, %f1 : f32
+    // CHECK: llvm.frem [[f1]], [[f1]] : f32
 }
--- a/xdsl/dialects/llvm.py
+++ b/xdsl/dialects/llvm.py
@@ -8,6 +8,7 @@ from typing import ClassVar
 
 from xdsl.dialects.builtin import (
     I64,
+    AnyFloatConstr,
     AnyIntegerAttr,
     ArrayAttr,
     ContainerType,
@@ -25,7 +26,7 @@ from xdsl.dialects.builtin import (
     i32,
     i64,
 )
-from xdsl.dialects.utils import FastMathAttrBase
+from xdsl.dialects.utils import FastMathAttrBase, FastMathFlag
 from xdsl.ir import (
     Attribute,
     BitEnumAttribute,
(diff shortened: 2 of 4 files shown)


xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1734,132 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
         )
 
 
+class AbstractFloatArithOp(IRDLOperation, ABC):
+    T: ClassVar = VarConstraint("T", BaseAttr(AnyFloat))
```
alexarice, line 1738:
```suggestion
    T: ClassVar = VarConstraint("T", AnyFloatConstr)
```
I think this exists for this reason

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1734,132 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+    res = result_def(T)
+
+    fastmathFlags = opt_attr_def(FastMathAttr)
+
+    traits = traits_def(NoMemoryEffect())
```
alexarice, line 1746: Any reason for these not being `Pure`

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1734,132 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
 
 
+class AbstractFloatArithOp(IRDLOperation, ABC):
+    T: ClassVar = VarConstraint("T", BaseAttr(AnyFloat))
+
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+    res = result_def(T)
+
+    fastmathFlags = opt_attr_def(FastMathAttr)
```
alexarice, line 1744: I just checked and this should be a property, non-optional, and have a default value

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1734,132 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+        rhs: SSAValue | Operation,
+        fast_math: FastMathAttr | FastMathFlag | None,
+    ):
+        if isinstance(fast_math, FastMathFlag | str):
+            fast_math = FastMathAttr(fast_math)
+
+        super().__init__(
+            operands=[lhs, rhs],
+            result_types=[SSAValue.get(lhs).type],
+            attributes={"fastmathFlags": fast_math},
```
alexarice, line 1762:
```suggestion
            properties={"fastmathFlags": fast_math},
```

alexarice:
Do we have an mlir integration test for llvm ops? If not I feel it would be useful.

xdsl/dialects/llvm.py
AntonLydike: It's marked as an attribute on [the docs](https://mlir.llvm.org/docs/Dialects/LLVM/#llvmfmul-llvmfmulop) though? And the format it prints in is also `{fastmathFlags = ...}` without the prop brackets...

alexarice: I'm sure we've had this exact conversation before, it literally prints with `<{...}>`
```
"builtin.module"() ({
  %0:2 = "test.op"() : () -> (f32, f32)
  %1 = "llvm.fmul"(%0#0, %0#1) <{fastmathFlags = #llvm.fastmath<none>}> : (f32, f32) -> f32
}) : () -> ()
```

alexarice: Everything is marked as an attribute on the docs, they never switched the docs over to properties

AntonLydike:
> Do we have an mlir integration test for llvm ops? If not I feel it would be useful.

Added!

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1736,159 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+        attrs: dict[str, Attribute] | None = None,
+    ):
+        if isinstance(fast_math, FastMathFlag | str | None):
+            fast_math = FastMathAttr(fast_math)
+
+        super().__init__(
+            operands=[lhs, rhs],
+            result_types=[SSAValue.get(lhs).type],
+            properties={"fastmathFlags": fast_math},
+            attributes=attrs if attrs is not None else {},
```
alexarice, line 1767:
```suggestion
            attributes=attrs,
```
Does this work?

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1736,159 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+        )
+
+    def print(self, printer: Printer):
```
alexarice, line 1770: I don't think this is necessary, the assembly format from before should still work

xdsl/dialects/llvm.py
AntonLydike: Fixed!

xdsl/dialects/llvm.py
AntonLydike: `xdsl.utils.exceptions.PyRDLOpDefinitionError: ('Error during the parsing of the assembly format: ', (Span[38:38](text=''), "fastmathFlags properties are missing from the declarative format. If this is intentional, consider using 'ParsePropInAttrDict' IRDL option."))`

alexarice: Have you considered using the 'ParsePropInAttrDict' IRDL option?

alexarice (approved):

xdsl/dialects/llvm.py
AntonLydike: Ah, no it doesnt! Because `fastmath=none` flags are not supposed to be printed at all!

alexarice (changes requested):
This should be working with assembly format and I'd prefer to get the bugged fixed rather than introducing custom parsing for something that shouldn't need it

xdsl/dialects/llvm.py
AntonLydike: Hmmmm, yeah maybe we could hide attributes if they are the default value

AntonLydike:
> This should be working with assembly format and I'd prefer to get the bugged fixed rather than introducing custom parsing for something that shouldn't need it

But it doesn't. 

And fixing the assembly format is outside the scope of this PR.

I'm not able to fix the assembly format today.

xdsl/dialects/llvm.py
alexarice: This should have been happening anyway, but I think I must have missed a case somehow

alexarice: #3783

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1737,139 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
         )
 
 
+floatingPointLike = AnyOf([Float16Type, Float32Type, Float64Type])
```
alexarice, line 1740:
```suggestion
LLVMFloatConstr = AnyOf([Float16Type, Float32Type, Float64Type])
```

alexarice: More in line with naming for other constraints

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1737,139 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+    name = "llvm.frem"
+
+
+@irdl_op_definition
+class BitcastOp(GenericCastOp):
+    name = "llvm.bitcast"
+
+
+@irdl_op_definition
+class SIToFPOp(GenericCastOp):
```
alexarice, line 1807: These are currently very unconstrained but this isn't technically wrong

alexarice (approved):
Will leave the parent PR up to give people time to review.

xdsl/dialects/llvm.py
alexarice: Note the mlir op allows vectors of floats it seems

xdsl/dialects/llvm.py
AntonLydike: I think it kinda is, you can't cast from a float with this. Added a verify check.

xdsl/dialects/llvm.py
AntonLydike: Switched to `AnyFloatConstr`

xdsl/dialects/llvm.py
alexarice: The custom verifier is `verify_`

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1735,142 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+    name = "llvm.sitofp"
+
+    def verify(self, verify_nested_ops: bool = True) -> None:
+        if not isinstance(self.arg.type, IntegerType):
+            raise VerifyException("Expected integer type as input")
+        if not isinstance(self.result.type, AnyFloat):
+            raise VerifyException("Expected output to be a float")
```
alexarice, line 1809: I don't really understand why you would check input and output types in a custom verifier

alexarice (commented):
I'd prefer to either revert to how it was before or check the input and output types with constraints

alexarice:
Could you stop resolving discussions please, makes it hard to check what's happened

xdsl/dialects/llvm.py
AntonLydike: because then I can subclass the main conversion thingy class and keep the remaining bits.

xdsl/dialects/llvm.py
AntonLydike: Damn, `LinkageAttr` also uses the wrong verifier then. Just copied from there as I'm never sure which one to use :upside_down_face:

alexarice: It's different for attributes

xdsl/dialects/llvm.py
```diff
@@ -1735,56 +1735,142 @@ def __init__(self, val: Operation | SSAValue, res_type: Attribute):
...
+class AbstractFloatArithOp(IRDLOperation, ABC):
+    T: ClassVar = VarConstraint("T", AnyFloatConstr)
+
+    lhs = operand_def(T)
+    rhs = operand_def(T)
+    res = result_def(T)
+
+    fastmathFlags = prop_def(FastMathAttr, default_value=FastMathAttr(None))
+
+    traits = traits_def(Pure())
```
compor, line 1747: Could you please use the`SameOperandsAndResultType` trait where applicable?
It shouldn't be here, as not all ops that inherit from this base should comply to it.

tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/arithmetic.mlir
```diff
@@ -44,4 +44,24 @@ builtin.module {
 
     %ashr = llvm.ashr %arg0, %arg1 : i32
     // CHECK: %{{.*}} = llvm.ashr %{{.*}}, %{{.*}} : i32
+
+    // float arith:
+
+    %fmul = llvm.fmul %f1, %f1 : f32
+    // CHECK: %{{\d+}} = llvm.fmul %2, %2 : f32
```
compor, line 51: Any reason for not using a pattern capture here (i.e., `{{.*}}` instead of `%2`)?

superlopuh: hmm, why are they not `%f1` in the first place?

superlopuh: Ah because of `mlir-opt`

xdsl/dialects/llvm.py
alexarice: This is now merged

tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/arithmetic.mlir
alexarice: Likely better to use a numeric capture if you really want to avoid hard coding this
```suggestion
    // CHECK: %{{\d+}} = llvm.fmul %[[#in:]], %[[#in]] : f32
```

compor: yes, that's what I meant :)

lfrenot (approved):

alexarice (approved):

PR #3782, opened 2025-01-23
documentation: update ImplicitBuilder doc comment

### LINKED ISSUE #3779: Documentation on implicit builder is out of date?
The implicit builder docs state:

```

class ImplicitBuilder(contextlib.AbstractContextManager[tuple[BlockArgument, ...]]):
    """
    Stores the current implicit builder context, consisting of the stack of builders in
    the current thread, and the current builder.

    Operations created within a `with` block of an implicit builder will be added to it.
    If there are nested implicit builder blocks, the operation will be added to the
    innermost one. Operations cannot be added to multiple blocks, and any attempt to do so
    will result in an exception.

    Example:

    ``` python
    from xdsl.dialects import arith

    block = Block()
    builder = Builder(block)

    with builder.implicit():
        arith.Constant.from_int_and_width(5, 32)

    assert len(block.ops) == 1
    assert isinstance(block.ops.first, arith.Constant)
    ```
    """
```


But trying to use `Bulder(block).implicit()` throws:

```
  File "/home/anton/projects/sympy2xdsl/test/filecheck/basic.py", line 13, in <module>
    with builder.implicit():
         ^^^^^^^^^^^^^^^^
AttributeError: 'Builder' object has no attribute 'implicit'
```

(@superlopuh )

superlopuh:
fixes #3779

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -331,10 +331,7 @@ class ImplicitBuilder(contextlib.AbstractContextManager[tuple[BlockArgument, ...
     ``` python
     from xdsl.dialects import arith
 
-    block = Block()
-    builder = Builder(block)
-
-    with builder.implicit():
+    with ImplicitBuilder(block):
         arith.Constant.from_int_and_width(5, 32)
 
     assert len(block.ops) == 1


alexarice (approved):
Worth noting that you can still pass a builder into `ImplicitBuilder`

Issue #3739, opened 2025-01-13
dialects: (math) Several ops do not accepts vector or tensor types, but their MLIR counterparts do


watermelonwolverine:
Example: The `math.fma` and `math.op` do not accept `vector<3xf64>`, but their MLIR counterparts do.

superlopuh:
Would be great to update these, indeed!

Issue #2739, opened 2024-06-17
Missing custom syntax for `math.exp`


zero9178:
See https://mlir.llvm.org/docs/Dialects/MathOps/#mathexp-mathexpop for the syntax

compor:
resiolved in #3786

PR #3786, opened 2025-01-24
dialects: (math) Add support for custom format and container types to `math` dialect ops

compor:
This PR:

- Adds support for custom format
- Adds support for container types in the operands and results of the relevant ops (e.g., `tensor`)
- Tests (including interoperability) of the above

Resolves: #3739, #2739


xdsl/dialects/math.py
```diff
@@ -487,36 +436,20 @@ class ExpOp(IRDLOperation):
...
-        attributes = {"fastmath": fastmath}
-
-        operand = SSAValue.get(operand)
-        super().__init__(
-            attributes=attributes, operands=[operand], result_types=[operand.type]
-        )
-
 
 @irdl_op_definition
 class FPowIOp(IRDLOperation):
```
compor, line 444: Not conforming the base classes. Also the only one that `SameOperandsAndResultType` trait does not apply.

xdsl/dialects/math.py
```diff
@@ -568,29 +503,13 @@ class FloorOp(IRDLOperation):
...
-        attributes = {"fastmath": fastmath}
-
-        operand = SSAValue.get(operand)
-        super().__init__(
-            attributes=attributes, operands=[operand], result_types=[operand.type]
-        )
-
 
 @irdl_op_definition
 class FmaOp(IRDLOperation):
```
compor, line 511: Not conforming to any of the base classes.

xdsl/dialects/math.py
```diff
@@ -67,22 +193,13 @@ class AbsIOp(IRDLOperation):
...
-        super().__init__(operands=[operand], result_types=[operand.type])
-
 
 @irdl_op_definition
-class Atan2Op(IRDLOperation):
+class Atan2Op(FloatingPointLikeBinaryMathOperationWithFastMath):
     """
-    Syntax:
-    operation ::= ssa-id `=` `math.atan2` ssa-use `,` ssa-use `:` type
-
```
compor, line 85: Not fond of adding the syntax in the doc; this was done only partially for some of the `math` ops, and we don't do it anywhere else in the codebase.

xdsl/dialects/math.py
```diff
@@ -529,12 +462,18 @@ class FPowIOp(IRDLOperation):
     """
 
     name = "math.fpowi"
+
+    T1: ClassVar = VarConstraint("T1", floatingPointLike)
+    T2: ClassVar = VarConstraint("T2", signlessIntegerLike)
```
superlopuh, line 467: seems unnecessary, although I guess it doesn't hurt much

superlopuh (approved):

xdsl/dialects/math.py
alexarice: I agree, there's no reason to have this variable

xdsl/dialects/math.py
```diff
@@ -529,12 +462,18 @@ class FPowIOp(IRDLOperation):
...
+
     fastmath = opt_prop_def(FastMathFlagsAttr)
```
alexarice, line 469: A lot of the other fastmath flags in mlir are default valued props instead of optional props. Is the math dialect different?

xdsl/dialects/math.py
compor: I'll amend, it's a leftover from reusing the base classes.

xdsl/dialects/math.py
compor: What do you mean by "a lot"? Our arith and llvm dialect uses of these differ, but the net result is the same as absence means `none`.

I'll check MLIR later and amend in a separate PR just for the flags as required.
I also wanted to check the relevant interface, so another incentive to do so.

xdsl/dialects/math.py
compor: It seems like the assembly format cannot infer the type of `result`

>  type of result 'result' cannot be inferred, consider adding a 'type($result)' directive to the custom assembly format

Hence, I ended up keep one of them.

PR #3797, opened 2025-01-28
core: (constraints) add IntAttrConstraint

alexarice:
Adds a constraint that applies an integer constraint to the value of an integer attribute.

See: https://github.com/xdslproject/inconspiquous/blob/70f923cfe577b7eb48ece03d6f9a597c296cb698/inconspiquous/dialects/gate.py#L237
and https://github.com/xdslproject/inconspiquous/blob/70f923cfe577b7eb48ece03d6f9a597c296cb698/inconspiquous/dialects/qssa.py#L78

for an example of using this. 

I'm unsure if this is the best place for this to live, or if it should be upstreamed at all.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -3163,25 +3163,26 @@ def test_multiple_operand_extraction_fails():
 ################################################################################
 
 
+@irdl_op_definition
+class IntAttrExtractOp(IRDLOperation):
+    name = "test.int_attr_extract"
+
+    _I: ClassVar = IntVarConstraint("I", AnyInt())
+
+    prop = prop_def(
+        IntegerAttr.constr(value=IntAttrConstraint(_I), type=eq(IndexType()))
+    )
+
+    outs = var_result_def(RangeOf(eq(IndexType()), length=_I))
+
+    assembly_format = "$prop attr-dict"
+
+
 @pytest.mark.parametrize(
     "program",
     ["%0 = test.int_attr_extract 1", "%0, %1 = test.int_attr_extract 2"],
 )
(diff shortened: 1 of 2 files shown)


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -3155,3 +3156,155 @@ def test_multiple_operand_extraction_fails():
...
+################################################################################
+
+
+@pytest.mark.parametrize(
+    "program",
+    ["%0 = test.int_attr_extract 1", "%0, %1 = test.int_attr_extract 2"],
+)
+def test_int_attr_extraction(program: str):
+    @irdl_op_definition
+    class IntAttrExtractOp(IRDLOperation):
```
compor, line 3172: Mind pulling this and one below out and use a single definition? Fewer lines of code to eyeball over.

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -3155,3 +3156,155 @@ def test_multiple_operand_extraction_fails():
...
+        "test.int_attr_verify 1, %0",
+        "test.int_attr_verify 2, %0, %1",
+        "test.int_attr_verify 1 and 1, %0",
+        "test.int_attr_verify 2 and 2, %0, %1",
+    ],
+)
+def test_int_attr_verify(program: str):
+    @irdl_op_definition
+    class IntAttrExtractOp(IRDLOperation):
+        name = "test.int_attr_verify"
```
compor, line 3238: Same here, plus a name for the class and its name that match (here one uses 'extract' and the other uses 'verify'

compor (approved):

superlopuh (approved):
Very cool

PR #3802, opened 2025-01-29
dialects: (arith) Fix `arith.constant` to return only signless integers

### LINKED ISSUE #1101: dialects: (arith) `arith.constant` verifies non-signless types.
It seems that `xdsl-opt` allows us to create signed and unsigned output whereas this is not allowed in `mlir-opt`

compor:
This PR:

- Fixes `arith.constant` to return only signless integers
- Adds tests of the above

Resolves #1101

### Update: The below are fixed in #3966

cc @n-io @dk949 
A side effect of this is that some of CSL backend tests are broken, as they use `arith.constant` to generate signed and unsigned integers for various bit widths.

I've fixed `print_csl.mlir` by adding `csl.mlir.signedness_cast`s which creates benign(?) artifacts of `@as(i32, 100)` vs plain `100`. I'm hoping these are essentially optimized away by the CSL compiler/SDK, but I couldn't find anything related by a cursory look in the docs.

More involved fixes would be need in the CSL canonicalization patterns, as they explicitly check for the existence of `arith.constant` ops.

@superlopuh my suggestion would be for the CSL ppl to create a separate PR to fix these before this going in, not sure if you agree.

Failing filecheck tests:

```
dialects/csl/csl-canonicalize.mlir
transforms/csl-stencil-handle-async-flow.mlir
transforms/memref-to-dsd.mlir
```


tests/filecheck/dialects/arith/arith_invalid.mlir
```diff
@@ -1,29 +1,27 @@
 // RUN: xdsl-opt %s --split-input-file --verify-diagnostics | filecheck %s
 
-"builtin.module"() ({
-
```
compor, line 4: Removed these as the builtin module is implied by default

tests/filecheck/dialects/arith/arith_invalid.mlir
```diff
@@ -1,29 +1,27 @@
...
 
-}) : () -> ()
+%i32 = "test.op"() : () -> i32
+%res = "arith.index_cast"(%i32) : (i32) -> i32
+// CHECK: 'arith.index_cast' op operand type 'i32' and result type 'i32' are cast incompatible
 
 // -----
 
-"builtin.module"() ({
+%c = arith.constant 1 : si32
```
compor, line 21: Diff seems a bit confused here, but these are the 2 new invalid test cases.

superlopuh:
I'm a bit confused by the "fix", why is this desirable?

compor:
> I'm a bit confused by the "fix", why is this desirable?

What fix are you referring to? The arith or the CSL part?

compor:
Yeah, that's right, I think the docs should say "signless integer-like" etc. I believe that's why it slipped.

But thinking about it, arith is a high enough level of abstraction to deal with signless only integers, so it makes sense within the MLIR rationale, just poor documentation.

> The second, I can't find this constraint in the description here: https://mlir.llvm.org/docs/Dialects/ArithOps/#arithconstant-arithconstantop
> 
> I tried it on godbolt and you're right: https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AtaeMkvrICeAZUboAwqloBXFgz32AMngaYAcm4ARpjEIABMpAAOqAqENgxOru56MXHWAj5%2BgSwhYZEWmFYJQgRMxARJbh5cpIXFAqXlBFkBwaERdWUVVSm1Ct0tvm25HeEAlHWoLsTI7BzlhAgAdGgMA4YEANQaWyBbcQDM4QCkGgCCHJO0nACsvB4cWqSonACyXgCSAEoH07OYLYncKHHikAiaK6TADWIEOh2W8KRyJRADZ9JxJA9IS9OLwFCANODIZM4LAkGgWFE6KFyJRKdT6GETGZUFECFxURoiTRaARQgSIEEcUFfOUAJ6cMGi5jEcUAeSC2iKEO4vEpbEE8oYtElT14WCCLmADjEtAJatIWBYhmA4n1VrMKrwADdMBbnphVEUXPypbxfPybg7TEFiBKnFgcQRiHgWP7SG7iEFYpgACKYG1GUxGEl8AzABQANTwmAA7vKoowE/xBCIxOwpDJBIoVOoHbpagZcyyyKGCZBJmyMus8S8k7GsAOIJN6iO7AxHM5qiBbrVvMMcnkQNI0vEBL0PGvorF9wxWluOtI5yVBofV/1MJYR00Khf2mFr3flyljxs35uH47rO/xzBI1x3NiDqvBwWwfD8WzDlsnLLBoqFbBAuCECQQIglw4y8KqWjjDCcIIiiFFIuiwZYqQjzPDB%2BKEsS%2BqkjAiAgNMBBRL6dIQAyNLEP4rDzPBvxIShaGevgRCTnotbCKI4hNgprZqDinakGW4ZRP6EEcPcdE4jB8q%2Bjx2yoFQcFfOJ7LIaiqHoRAThUoJuGggRLHEaR8KIpRFEYhwtH0bwjF1MxRFQoF4RQQxY6RSRiYCgkO5AA

xdsl/dialects/arith.py
```diff
@@ -132,7 +133,8 @@ class ConstantOp(IRDLOperation):
     result = result_def(_T)
     value = prop_def(
         TypedAttributeConstraint(
-            AnyIntegerAttrConstr
+            IntegerAttr.constr(type=IndexTypeConstr)
+            | IntegerAttr.constr(type=SignlessIntegerConstraint)
```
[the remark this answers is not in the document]
alexarice: Side note: we should rename `SignlessIntegreConstraint` to `SignlessIntegerConstr`

tests/filecheck/dialects/arith/arith_invalid.mlir
superlopuh: Maybe tiny PR to dedent first I'll insta approve?

compor:
@n-io @dk949 any thoughts on this?

tests/filecheck/backend/csl/print_csl.mlir
```diff
@@ -580,12 +593,12 @@ csl.func @builtins() {
 // CHECK-NEXT: }
 // CHECK-NEXT: {{ *}}
 // CHECK-NEXT: fn casts() void {
-// CHECK-NEXT:   const castIndex : i32 = @as(i32, 0);
+// CHECK-NEXT:   const castIndex : i32 = @as(i32, @as(u16, 0));
```
compor, line 596: @emmau678 do you mind taking a look wrt this change? Is it benign, easily optimized away?

emmau678: To the best of my knowledge, yes it is benign. I ran a quick test in CSL and this much compiles and passes:
```
  const r1: i32 = @as(i32, 0);
  const r2: i32 = @as(i32, @as(u16, 0));
  if (r1 == r2) {
    prt.print_string("equal\n");
  }
```

emmau678: The @as builtin accepts a numeric value as the second arg (and csl numeric types include signed and unsigned ints) so I don't think it matters in which format you pass the zero.

compor: I assume it is the same with things like:

```
.tensor_access = | d0 | { @as(i32, 200) } -> A[ d0 ]
```
vs

```
.tensor_access = | d0 | { 200 } -> A[ d0 ]
```

and

```
const r2: i32 = @as(i32, 42);
```

vs

```
const r2: i32 = @as(i32, @as(i32, 42));
```

emmau678: yes (as far as I can tell) - this all compiles

compor: Thank you very much for verifying

n-io: Yes, this is indeed all benign.

n-io (approved):

superlopuh (approved):

PR #3805, opened 2025-01-29
dialects: (arith) Add support for bitcast operation

jakedves:
Implementation of MLIR's `arith.bitcast` op. Test cases are passing but type checking on `t1` and `t2` variables isn't

### DIFF
--- a/tests/dialects/test_arith.py
+++ b/tests/dialects/test_arith.py
@@ -7,6 +7,7 @@ from xdsl.dialects.arith import (
     AddiOp,
     AddUIExtendedOp,
     AndIOp,
+    BitcastOp,
     CeilDivSIOp,
     CeilDivUIOp,
     CmpfOp,
@@ -55,6 +56,7 @@ from xdsl.dialects.builtin import (
     IndexType,
     IntegerAttr,
     IntegerType,
+    Signedness,
     TensorType,
     VectorType,
     f32,
@@ -247,6 +249,54 @@ def test_select_op():
     # wanting to verify it actually selected the correct operand, but not sure if in correct scope
     assert select_t_op.result.type == t.result.type
     assert select_f_op.result.type == f.result.type
+
+
+@pytest.mark.parametrize(
+    "in_type, out_type, should_verify",
+    [
+        (i1, IntegerType(1, signedness=Signedness.UNSIGNED), True),
+        (i32, f32, True),
+        (i64, f64, True),
+        (i32, i32, True),
+        (IndexType(), i1, True),
+        (i1, IndexType(), True),
+        (f32, IndexType(), True),
+        (IndexType(), f64, True),
+        (VectorType(i64, [3]), VectorType(f64, [3]), True),
+        (VectorType(f32, [3]), VectorType(i32, [3]), True),
+        # false cases
+        (i1, i32, False),
+        (i32, i64, False),
+        (i64, i32, False),
+        (f32, i64, False),
+        (f32, f64, False),
+        (VectorType(i32, [5]), i32, False),
+        (i64, VectorType(i64, [5]), False),
+        (VectorType(i32, [5]), VectorType(f32, [6]), False),
+        (VectorType(i32, [5]), VectorType(f64, [5]), False),
+    ],
+)
+def test_bitcast_op(in_type: Attribute, out_type: Attribute, should_verify: bool):
+    in_arg = TestSSAValue(in_type)
+    cast = BitcastOp(in_arg, out_type)
+
+    if should_verify:
+        cast.verify_()
+        assert cast.result.type == out_type
+        return
+
+    # expecting test to fail
+    with pytest.raises(TypeError) as e:
+        cast.verify_()
+
+    err_msg1 = "Expected operand and result types to be signless-integer-or-float-like"
+    err_msg2 = "'arith.bitcast' operand and result types must have equal bitwidths"
+    err_msg3 = "'arith.bitcast' operand and result must both be containers or scalars"
+    err_msg4 = (
+        "'arith.bitcast' operand and result type elements must have equal bitwidths"
+    )
+    err_msg5 = "'arith.bitcast' operand and result types must have the same shape"
+    assert e.value.args[0] in [err_msg1, err_msg2, err_msg3, err_msg4, err_msg5]
 
 
 def test_index_cast_op():
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -10,7 +10,9 @@ from xdsl.dialects.builtin import (
     AnyIntegerAttr,
     AnyIntegerAttrConstr,
     ContainerOf,
+    ContainerType,
     DenseIntOrFPElementsAttr,
+    FixedBitwidthType,
     Float16Type,
     Float32Type,
     Float64Type,
@@ -19,6 +21,7 @@ from xdsl.dialects.builtin import (
     IntAttr,
     IntegerAttr,
     IntegerType,
+    ShapedType,
     TensorType,
     UnrankedTensorType,
     VectorType,
@@ -1224,6 +1227,67 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
 
 
 @irdl_op_definition
+class BitcastOp(IRDLOperation):
+    name = "arith.bitcast"
+
+    input = operand_def(signlessIntegerLike | floatingPointLike)
+    result = result_def(signlessIntegerLike | floatingPointLike)
+
+    assembly_format = "$input attr-dict `:` type($input) `to` type($result)"
+
+    def __init__(self, in_arg: SSAValue | Operation, target_type: Attribute):
+        super().__init__(operands=[in_arg], result_types=[target_type])
+
+    def verify_(self) -> None:
+        # check if we have a ContainerType
+        in_type = self.input.type
+        res_type = self.result.type
+
+        if isinstance(in_type, ShapedType) and isinstance(res_type, ShapedType):
+            if in_type.get_shape() != res_type.get_shape():
+                raise TypeError(
+                    "'arith.bitcast' operand and result types must have the same shape"
+                )
+
+            assert isinstance(in_type, ContainerType)
+            assert isinstance(res_type, ContainerType)
+
+            t1: Attribute = in_type.get_element_type()
+            t2: Attribute = res_type.get_element_type()
+            if BitcastOp._equal_bitwidths(t1, t2):
+                return
+
+            raise TypeError(
+                "'arith.bitcast' operand and result type elements must have equal bitwidths"
+            )
+
+        if isinstance(in_type, ContainerType) or isinstance(res_type, ContainerType):
+            raise TypeError(
+                "'arith.bitcast' operand and result must both be containers or scalars"
+            )
+
+        # at this point we know we have two scalar types
+        if not BitcastOp._equal_bitwidths(in_type, res_type):
+            raise TypeError(
+                "'arith.bitcast' operand and result types must have equal bitwidths"
+            )
+
+    @staticmethod
+    def _equal_bitwidths(type_a: Attribute, type_b: Attribute) -> bool:
+        if isinstance(type_a, IndexType) or isinstance(type_b, IndexType):
+            return True
+
+        if isinstance(type_a, FixedBitwidthType) and isinstance(
+            type_b, FixedBitwidthType
+        ):
+            return type_a.bitwidth == type_b.bitwidth
+
+        raise TypeError(
+            "Expected operand and result types to be signless-integer-or-float-like"
+        )
+
+
+@irdl_op_definition
 class IndexCastOp(IRDLOperation):
     name = "arith.index_cast"
 
@@ -1420,6 +1484,7 @@ Arith = Dialect(
         MaximumfOp,
         MaxnumfOp,
         # Casts
+        BitcastOp,
         IndexCastOp,
         FPToSIOp,
         SIToFPOp,


xdsl/dialects/arith.py
```diff
@@ -1223,6 +1226,67 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+                raise TypeError(
+                    "'arith.bitcast' operand and result types must have the same shape"
+                )
+
+            assert isinstance(in_type, ContainerType)
+            assert isinstance(res_type, ContainerType)
+
+            t1: Attribute = in_type.get_element_type()
+            t2: Attribute = res_type.get_element_type()
+            if BitcastOp._equal_bitwidths(t1, t2):
```
jakedves, line 1257: This line isn't type checking as types `t1` and `t2` are partially Unknown `(Unknown | Attribute)`

xdsl/dialects/arith.py
```diff
@@ -1223,6 +1227,64 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+            )
+
+        # at this point we know we have two scalar types
+        if not BitcastOp._equal_bitwidths(in_type, res_type):
+            raise VerifyException(
+                "'arith.bitcast' operand and result types must have equal bitwidths"
+            )
+
+    @staticmethod
+    def _equal_bitwidths(type_a: Attribute, type_b: Attribute) -> bool:
```
compor, line 1274: Is this something that could be reused? If so, maybe pull out as a utility method?

xdsl/dialects/arith.py
```diff
@@ -1223,6 +1227,64 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+            return True
+
+        if isinstance(type_a, FixedBitwidthType) and isinstance(
+            type_b, FixedBitwidthType
+        ):
+            return type_a.bitwidth == type_b.bitwidth
+
+        raise VerifyException(
+            "Expected operand and result types to be signless-integer-or-float-like"
+        )
```
compor, line 1285: I find it a bit confusing that this returns bool and can also throw an exception.
Maybe leave the exception raising to clients and just plainly return true/false here.

tests/dialects/test_arith.py
```diff
@@ -249,6 +251,54 @@ def test_select_op():
...
+    with pytest.raises(VerifyException) as e:
+        cast.verify_()
+
+    err_msg1 = "Expected operand and result types to be signless-integer-or-float-like"
+    err_msg2 = "'arith.bitcast' operand and result types must have equal bitwidths"
+    err_msg3 = "'arith.bitcast' operand and result must both be containers or scalars"
+    err_msg4 = (
+        "'arith.bitcast' operand and result type elements must have equal bitwidths"
+    )
+    err_msg5 = "'arith.bitcast' operand and result types must have the same shape"
```
compor, line 298: Please make these messages part of the `parametrize` decorator to match each individual case

xdsl/dialects/arith.py
```diff
@@ -1223,6 +1227,64 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+    input = operand_def(signlessIntegerLike | floatingPointLike)
+    result = result_def(signlessIntegerLike | floatingPointLike)
+
+    assembly_format = "$input attr-dict `:` type($input) `to` type($result)"
+
+    def __init__(self, in_arg: SSAValue | Operation, target_type: Attribute):
+        super().__init__(operands=[in_arg], result_types=[target_type])
+
+    def verify_(self) -> None:
+        # check if we have a ContainerType
```
compor, line 1243: this comment doesn't add much and it's a bit confusing since we first check for shaped types

xdsl/dialects/arith.py
compor: What I meant here is to maybe take inspiration from `get_element_type_or_self` and maybe just reuse this

```py
container_type = cast(ContainerType[Any], maybe_shaped_type)
container_type.get_element_type()
```

in order to incorporate the `ContainerType` check once in this verification method.

compor (commented):
Very good, thanks for this. I've left a few comments.

xdsl/dialects/arith.py
```diff
@@ -1223,6 +1227,64 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+        in_type = self.input.type
+        res_type = self.result.type
+
+        if isinstance(in_type, ShapedType) and isinstance(res_type, ShapedType):
+            if in_type.get_shape() != res_type.get_shape():
+                raise VerifyException(
+                    "'arith.bitcast' operand and result types must have the same shape"
+                )
+
+            t1 = get_element_type_or_self(in_type)
```
alexarice, line 1253: Could this check not go right at the start of the function? This would save you checking again in the scalar case at the bottom

xdsl/dialects/arith.py
```diff
@@ -1223,6 +1227,64 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+    def __init__(self, in_arg: SSAValue | Operation, target_type: Attribute):
+        super().__init__(operands=[in_arg], result_types=[target_type])
+
+    def verify_(self) -> None:
+        # check if we have a ContainerType
+        in_type = self.input.type
+        res_type = self.result.type
+
+        if isinstance(in_type, ShapedType) and isinstance(res_type, ShapedType):
+            if in_type.get_shape() != res_type.get_shape():
```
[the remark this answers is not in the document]
compor: Good point, the logic in upstream MLIR is enforced by the `SameOperandsAndResultShape` which reuses heavily parts of the `SameOperandsAndResultType` trait that we got in xDSL recently.
Actually, this operation observes the `SameOperandsAndResultShape`, but since we don't have that yet, maybe reuse part of that infrastructure in the custom verifier here for now as @alexarice suggests/questions.

jakedves:
I've made all the changes suggested besides moving that function to `xdsl.utils.type` because it handles an edge case (`IndexType`) that I don't think would be useful/apply to other ops. Renamed it to make it clearer that its not generic if that's alright?

tests/dialects/test_arith.py
```diff
@@ -249,6 +252,63 @@ def test_select_op():
...
+    def test_bitcast_incorrect(
+        self, in_type: Attribute, out_type: Attribute, err_msg: str
+    ):
+        in_arg = TestSSAValue(in_type)
+        cast = BitcastOp(in_arg, out_type)
+
+        with pytest.raises(VerifyException) as e:
+            cast.verify_()
+
+        assert e.value.args[0] == err_msg
```
superlopuh, line 309:
```suggestion
        with pytest.raises(VerifyException, match=err_msg) as e:
            cast.verify_()
```

compor (commented):
A few nits!

tests/dialects/test_arith.py
```diff
@@ -249,6 +252,63 @@ def test_select_op():
...
+            (VectorType(f32, [3]), VectorType(i32, [3])),
+            (MemRefType(i32, [5]), MemRefType(f32, [5])),
+        ],
+    )
+    def test_bitcast_op(self, in_type: Attribute, out_type: Attribute):
+        in_arg = TestSSAValue(in_type)
+        cast = BitcastOp(in_arg, out_type)
+
+        cast.verify_()
+        assert cast.result.type == out_type
```
superlopuh, line 282: why put this in a test class?

superlopuh: And below

superlopuh: Ah, to use the constant strings above... I think they'd be fine at the file level also, with UPPER_SNAKE_CASE

xdsl/dialects/arith.py
```diff
@@ -1223,6 +1227,55 @@ class MinnumfOp(FloatingPointLikeBinaryOperation):
...
+    def _types_bitcastable(type_a: Attribute, type_b: Attribute) -> bool:
+        if isinstance(type_a, IndexType) or isinstance(type_b, IndexType):
+            return True
+
+        if isinstance(type_a, FixedBitwidthType) and isinstance(
+            type_b, FixedBitwidthType
+        ):
+            return type_a.bitwidth == type_b.bitwidth
+
+        return False
```
superlopuh, line 1276: IMO this is a good place for it for now, we can always move it out if it becomes useful elsewhere. Out of curiosity, where is the equivalent logic in MLIR?

superlopuh (approved):

xdsl/dialects/arith.py
jakedves: I would say [this method](https://github.com/llvm/llvm-project/blob/03ad7edbb652f17382d71e345492534202c437c9/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L1739) - not sure how it handles `IndexType`

superlopuh:
@jakedves, as a convention, we tend to let the original commenter resolve the messages, it makes it easier to see how the change was addressed, and whether the conversation is really resolved.

xdsl/dialects/arith.py
compor: @superlopuh actually that's a good point. Do we need interoperability tests for this one? we should, shouldn't we?

compor: btw, I don't think `IntegerType` also represents `IndexType`s and trying this locally with `mlir-opt` throws an assertion for me

>  unsigned int mlir::Type::getIntOrFloatBitWidth() const: Assertion `isIntOrFloat() && "only integers and floats have a bitwidth"' failed.

jakedves: Sounds good to me, does this mean that I shouldn't accept `IndexType` in the `operand/result_def` (i.e. replace `signlessIntegerLike` with `IntegerType`)? Or verify exception when an `IndexType` gets passed in?

The [MLIR docs](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithbitcast-arithbitcastop) accept `signless-integer-or-float-like` which  I assumed would be the `signlessIntegerLike = ContainerOf(AnyOf([IntegerType, IndexType]))` constant defined at the top of `arith`

superlopuh: After some quick searching, my understanding is that MLIR actually does allow index... 

```
def BitcastTypeConstraint : TypeConstraint<Or<[
        SignlessIntegerOrFloatLike.predicate,
        MemRefOf<[AnySignlessInteger, Index, AnyFloat]>.predicate]>,
    "signless-integer-or-float-like or memref of signless-integer or float">;

// Type constraint for signless-integer-like or float-like types.
def SignlessIntegerOrFloatLike : TypeConstraint<Or<[
    SignlessIntegerLike.predicate, FloatLike.predicate]>,
    "signless-integer-like or floating-point-like">;

// Type constraint for signless-integer-like types: signless integers, indices,
// vectors of signless integers or indices, tensors of signless integers.
def SignlessIntegerLike : TypeOrValueSemanticsContainer<
    AnySignlessIntegerOrIndex, "signless-integer-like">;

// Any signless integer type or index type.
def AnySignlessIntegerOrIndex : Type<CPred<"$_self.isSignlessIntOrIndex()">,
                                     "signless integer or index">;
```

superlopuh: This seems to be only for itof and ftoi casts

superlopuh: This actually seems like the implementation of the helper above. God bless CPP.
```cpp
static ::llvm::LogicalResult __mlir_ods_local_type_constraint_ArithOps3(
    ::mlir::Operation *op, ::mlir::Type type, ::llvm::StringRef valueKind,
    unsigned valueIndex) {
  if (!(((((type.isSignlessIntOrIndex())) || (((type.hasTrait<::mlir::ValueSemantics>())) && ([](::mlir::Type elementType) { return (elementType.isSignlessIntOrIndex()); }(::llvm::cast<::mlir::ShapedType>(type).getElementType())))) || (((::llvm::isa<::mlir::FloatType>(type))) || (((type.hasTrait<::mlir::ValueSemantics>())) && ([](::mlir::Type elementType) { return (::llvm::isa<::mlir::FloatType>(elementType)); }(::llvm::cast<::mlir::ShapedType>(type).getElementType()))))) || (((::llvm::isa<::mlir::MemRefType>(type))) && ([](::mlir::Type elementType) { return ((elementType.isSignlessInteger())) || ((::llvm::isa<::mlir::IndexType>(elementType))) || ((::llvm::isa<::mlir::FloatType>(elementType))); }(::llvm::cast<::mlir::ShapedType>(type).getElementType()))))) {
    return op->emitOpError(valueKind) << " #" << valueIndex
        << " must be signless-integer-or-float-like or memref of signless-integer or float, but got " << type;
  }
  return ::mlir::success();
}
```

superlopuh (approved):

xdsl/dialects/arith.py
superlopuh: After my spelunking, my impression is that this implementation is consistent with MLIR, even if I don't quite get how it's supposed to work.

compor: Hmm, sorry for coming back to this; I'm not sure where you're looking, but isn't the relevant implementation `bool arith::BitcastOp::areCastCompatible(TypeRange inputs, TypeRange outputs)` in ArithOps.cpp?

It does invoke the helper above, but it fails when trying to retrieve the bit width, since `index` types have none?

> unsigned int mlir::Type::getIntOrFloatBitWidth() const: Assertion `isIntOrFloat() && "only integers and floats have a bitwidth"' failed. 

This works with `mlir-opt`
```mlir
%arg1 = "test.op"() : () -> f32
%r2 = arith.bitcast %arg1 : f32 to i32
````

while this doesn't
```mlir
%arg0 = "test.op"() : () -> index
%r1 = arith.bitcast %arg0 : index to i32
```

Am I missing something obvious?

tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_bcast.mlir
```diff
@@ -0,0 +1,20 @@
...
+  %6 = "arith.bitcast"(%5) : (f64) -> i64
+}) : ()->()
+
+// CHECK:        %c1_i32 = "arith.constant"() <{value = 1 : i32}> : () -> i32
+// CHECK-NEXT:   %0 = "arith.bitcast"(%c1_i32) : (i32) -> i32
+// CHECK-NEXT:   %1 = "arith.bitcast"(%c1_i32) : (i32) -> f32
+// CHECK-NEXT:   %2 = "arith.bitcast"(%1) : (f32) -> i32
+// CHECK-NEXT:   %c1_i64 = "arith.constant"() <{value = 1 : i64}> : () -> i64
+// CHECK-NEXT:   %3 = "arith.bitcast"(%c1_i64) : (i64) -> f64
+// CHECK-NEXT:   %4 = "arith.bitcast"(%3) : (f64) -> i64
```
superlopuh, line 20: Nit: we've been slowly transitioning to check lines above inputs
```suggestion
// CHECK:        %c1_i32 = "arith.constant"() <{value = 1 : i32}> : () -> i32
// CHECK-NEXT:   %0 = "arith.bitcast"(%c1_i32) : (i32) -> i32
// CHECK-NEXT:   %1 = "arith.bitcast"(%c1_i32) : (i32) -> f32
// CHECK-NEXT:   %2 = "arith.bitcast"(%1) : (f32) -> i32
// CHECK-NEXT:   %c1_i64 = "arith.constant"() <{value = 1 : i64}> : () -> i64
// CHECK-NEXT:   %3 = "arith.bitcast"(%c1_i64) : (i64) -> f64
// CHECK-NEXT:   %4 = "arith.bitcast"(%3) : (f64) -> i64

%0 = "arith.constant"() {"value" = 1 : i32} : () -> i32
%1 = "arith.bitcast"(%0) : (i32) -> i32
%2 = "arith.bitcast"(%0) : (i32) -> f32
%3 = "arith.bitcast"(%2) : (f32) -> i32

%4 = "arith.constant"() {"value" = 1 : i64} : () -> i64
%5 = "arith.bitcast"(%4) : (i64) -> f64
%6 = "arith.bitcast"(%5) : (f64) -> i64
```

Also, why the generic format? It should work just as well in custom format, and be more readable

xdsl/dialects/arith.py
jakedves: @compor something like [this](https://github.com/xdslproject/xdsl/pull/3805/commits/4c560260d035d16c422cae90805c6901dacae17f)? Then remove the case of `IndexType` manually for operand/result types?

tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_bcast.mlir
jakedves: No reason, I was just following what other files had done

compor: Could we also please use `test.op` instead of `arith.constant`s while allowing unregistered dialects with ` mlir-opt --allow-unregistered-dialect`?

superlopuh: I would vote for custom format, and I agree with the test.op thing, although in my understanding there's no need for that flag

jakedves: I needed that flag on my (quite possibly broken) machine

tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_bcast.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: mlir-opt %s --allow-unregistered-dialects | xdsl-opt | filecheck %s
+
```
compor, line 2:
```suggestion
// RUN: mlir-opt %s --allow-unregistered-dialect | xdsl-opt | filecheck %s

```

compor (approved):
LGTM module a green CI run (the `mlir-opt` flag has a typo)

superlopuh:
Thank you for this!

PR #3809, opened 2025-01-30
tests: Remove module from `arith` filecheck tests

compor:
This PR to de-indents tests by removing the implicitly added `builtin.module`.

@superlopuh I applied my script to similar files within that test subfolder while at it.

As discussed [here](https://github.com/xdslproject/xdsl/pull/3802#discussion_r1934106464).


superlopuh (approved):
Thank you!

PR #3813, opened 2025-01-31
testing: move frontend desymref test to custom format

superlopuh:
It's not very clear from the diff, but here are the changes:

1. remove the outer builtin.module and dedent
2. move check lines above inputs
3. move everything to custom syntax


compor (approved):
Thanks, we need this in many places.

PR #3815, opened 2025-02-01
misc: simplify setup.py

superlopuh:
Just a bit of simplification. I tried to remove the whole thing but versioneer breaks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ xdsl-tblgen = "xdsl.tools.tblgen_to_py:main"
 [tool.setuptools]
 platforms = ["Linux", "Mac OS-X", "Unix"]
 zip-safe = false
+packages = ["xdsl"]
 
 [tool.setuptools.package-data]
 xdsl = ["**/*.irdl", "py.typed", "interactive/*.tcss", "_version.py"]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,15 +1,8 @@
...
 
-version = versioneer.get_version()
-
-
 setup(
-    version=version,
-    cmdclass=cast(Mapping[str, type[Command]], versioneer.get_cmdclass()),
-    packages=find_packages(),
+    version=versioneer.get_version(),
+    cmdclass=versioneer.get_cmdclass(),
```
superlopuh, line 7: this is well typed now

pyproject.toml
```diff
@@ -57,6 +57,7 @@ xdsl-tblgen = "xdsl.tools.tblgen_to_py:main"
 [tool.setuptools]
 platforms = ["Linux", "Mac OS-X", "Unix"]
 zip-safe = false
+packages = ["xdsl"]
```
superlopuh, line 60: we only have one package, no need to auto-discover things

compor (approved):

math-fehr (approved):

alexarice:
This has completely broken downstream

PR #3821, opened 2025-02-02
misc: use default-valued type elements for constrs

superlopuh:
Turns out that we can use typing_etensions for default values, leeting us have simpler helpers that leverage more advanced type system things.


math-fehr (approved):
That's quite nice indeed!

PR #3824, opened 2025-02-03
dialects: (arith) Use the `SameOperandsAndResultType` trait in `arith` dialect

compor:
This PR:

- Uses the `SameOperandsAndResultsType` trait in the relevant `arith` ops
- Cleans up unused test infrastructure for the return type of the above


tests/dialects/test_arith.py
```diff
@@ -101,18 +97,15 @@ class Test_integer_arith_construction:
             ShRSIOp,
         ],
     )
-    @pytest.mark.parametrize("return_type", [None, operand_type])
```
compor, line 104: This was never used. Now unnecessary.

tests/dialects/test_arith.py
```diff
@@ -69,10 +67,8 @@
 from xdsl.utils.isattr import isattr
 from xdsl.utils.test_value import TestSSAValue
 
-_BinOpArgT = TypeVar("_BinOpArgT", bound=Attribute)
-
```
compor, line 73: Not used anywhere.

xdsl/dialects/arith.py
```diff
@@ -59,7 +60,6 @@
 signlessIntegerLike = ContainerOf(AnyOf([IntegerType, IndexType]))
 floatingPointLike = ContainerOf(AnyOf([Float16Type, Float32Type, Float64Type]))
 
-_FloatTypeT = TypeVar("_FloatTypeT", bound=AnyFloat)
```
compor, line 62: Not used anywhere.

tests/dialects/test_arith.py
```diff
@@ -69,10 +67,8 @@
...
 
-class Test_integer_arith_construction:
+class test_integer_arith_construction:
     operand_type = i32
```
compor, line 72: minor renaming.

alexarice: Don't we normally go with `TestIntegerArithConstruction` for classes

alexarice (commented):
I'm not sure this does much, we already enforce (too strictly) that the operands and results must be the same if my understanding is correct

tests/dialects/test_arith.py
compor: Do we? I hadn't noticed. Let me see and I'll update it.

alexarice:
Right, I would have made the same comment on the other one if I had got round to reviewing it. I believe the operation is already more constrained by the current constraints.

compor:
> Right, I would have made the same comment on the other one if I had got round to reviewing it. I believe the operation is already more constrained by the current constraints.

Sorry I don't understand, even for shapes?

alexarice:
Maybe a better question is do you have a test case which has different behaviour with this change.

compor:
> Maybe a better question is do you have a test case which has different behaviour with this change?

Maybe I wasn't clear, but my annoyance for what I'm trying out here is that it is overly constraint, especially with dynamic shapes:

```mlir
  %0 = "test.op"() : () -> tensor<1x?xi64>
  %1 = "test.op"() : () -> tensor<1x3xi64>
  %2 = "arith.addi"(%0, %1) <{overflowFlags = #arith.overflow<none>}> : (tensor<1x?xi64>, tensor<1x3xi64>) -> tensor<1x?xi64>
```

This passes from `mlir-opt` but not us ATM.

My idea was to add the attribute and then relax this.
I didn't do it in `math` dialect, because I don't care (yet?).

superlopuh:
This is a great example, could it be added to this PR?

superlopuh:
Might as well relax in one go, no?

compor:
> Might as well relax in one go, no?

Yeah, sure, wdyt @alexarice ?

alexarice:
I agree, with the caveat that I'm almost certain this will mess up the assembly format in a way I don't know how to fix

alexarice:
what would the custom format look like for your example, as `arith.addi` only have one type in its assembly format

compor:
> what would the custom format look like for your example, as `arith.addi` only have one type in its assembly format

```mlir
%0 = "test.op"() : () -> tensor<1x?xi64>
%1 = "test.op"() : () -> tensor<1x3xi64>
%2 = arith.addi %0, %1 : tensor<1x?xi64
```

We get a different error then, because it assumes that `%1` should be of type `tensor<1x?xi64>`

>  operand is used with type tensor<1x?xi64>, but has been previously used or defined with type tensor<1x3xi64>

which I find even more confusing.

alexarice:
I just tried it myself, `mlir-opt` can't roundtrip `arith.addi`. How cursed.

```
 ~/p/mlir   master ±  mlir-opt ../mlir/test.mlir
module {
  %0 = "test.op"() : () -> tensor<1x?xi64>
  %1 = "test.op"() : () -> tensor<1x3xi64>
  %2 = arith.addi %0, %1 : tensor<1x?xi64>
}

 ~/p/mlir   master ±  mlir-opt ../mlir/test.mlir --mlir-print-op-generic
"builtin.module"() ({
  %0 = "test.op"() : () -> tensor<1x?xi64>
  %1 = "test.op"() : () -> tensor<1x3xi64>
  %2 = "arith.addi"(%0, %1) <{overflowFlags = #arith.overflow<none>}> : (tensor<1x?xi64>, tensor<1x3xi64>) -> tensor<1x?xi64>
}) : () -> ()

 ~/p/mlir   master ±  mlir-opt ../mlir/test.mlir | mlir-opt --mlir-print-op-generic
<stdin>:4:23: error: use of value '%1' expects different type than prior uses: 'tensor<1x?xi64>' vs 'tensor<1x3xi64>'
  %2 = arith.addi %0, %1 : tensor<1x?xi64>
                      ^
<stdin>:3:3: note: prior use here
  %1 = "test.op"() : () -> tensor<1x3xi64>
  ^
```

compor:
Cursed indeed.

I have no idea how to assist type inference for the operands `lhs`/`rhs` for this without reverting to custom parser/printer.

compor:
I can revert to handwritten parsers/printers and open an issue with a MWE for this

alexarice:
The least cursed thing to do might be to fallback to printing generic format if the input and output types don't exactly match

compor:
> The least cursed thing to do might be to fallback to printing generic format if the input and output types don't exactly match

that's a bit too clever and mixes too many aspects in printing,  no?

compor:
> Is this waiting for any input. It seemed to me that the result of last week is that we possibly don't want to support this, even if MLIR does, as it is likely a bug/point of high contention that MLIR supports this in the first place.

I think it is still good to add a test case to follow/note the MLIR behaviour, hence why I haven't closed or updated this.
Is this blocking you in any way?

alexarice:
It is not blocking me in any way

PR #3826, opened 2025-02-04
core: make all constr methods classmethods

alexarice:
Makes every constr method (found with ripgrep `def\ constr`) a classmethod instead of a staticmethod. This is in preparation for adding a default `constr` method to `Attribute` in a future PR.


compor (approved):
In similar vein as in #3821 I assume?

superlopuh (approved):

alexarice:
My plan was to start chipping away at these after #3828

superlopuh:
kk

PR #3828, opened 2025-02-04
core: add constr method to Attribute

alexarice:
Stacked on top of #3826 
Adds a `constr` method to `Attribute` which allows the easy construction of `BaseAttr` constraints. The hope here is to move towards having a single interface for generating constraints, instead of interchangably constructing the constraints manually, using `constr` methods, and using `AnyXConstr` type aliases.

### DIFF
--- a/xdsl/transforms/memref_to_dsd.py
+++ b/xdsl/transforms/memref_to_dsd.py
@@ -21,6 +21,7 @@ from xdsl.dialects.builtin import (
     StridedLayoutAttr,
     UnrealizedConversionCastOp,
 )
+from xdsl.dialects.csl.csl import ZerosOpAttr
 from xdsl.ir import Attribute, Operation, OpResult, SSAValue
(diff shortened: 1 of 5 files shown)


xdsl/transforms/memref_to_dsd.py
```diff
@@ -42,7 +43,7 @@ class LowerAllocOpPass(RewritePattern):
     def match_and_rewrite(self, op: memref.AllocOp, rewriter: PatternRewriter, /):
         assert isattr(
             memref_type := op.memref.type,
-            MemRefType.constr(element_type=csl.ZerosOpAttrConstr),
+            MemRefType[ZerosOpAttr].constr(element_type=csl.ZerosOpAttrConstr),
```
alexarice, line 46: This is a regression from this change. I believe there is no way to convince pyright of the various types involved without this annotation. See https://github.com/microsoft/pyright/issues/7721

superlopuh (approved):

compor (approved):

PR #3829, opened 2025-02-04
core: add PassFailedException

alexarice:
Adds a `PassFailedException` as a subclass of `DiagnosticException`. Can be used to signify the failure of a pass in a way that can be caught by filecheck.

### DIFF
--- a/xdsl/utils/exceptions.py
+++ b/xdsl/utils/exceptions.py
@@ -23,6 +23,10 @@ class VerifyException(DiagnosticException):
     pass
 
 
+class PassFailedException(DiagnosticException):
+    pass
+
+
 class PyRDLError(Exception):
     pass
 


math-fehr (approved):

xdsl/utils/exceptions.py
```diff
@@ -23,6 +23,10 @@ class VerifyException(DiagnosticException):
     pass
 
 
+class PassFailedException(DiagnosticException):
```
compor, line 26: Can we please add a docstring? I know it's "captain obvious", but maybe when and where this is supposed to appear/be used? Like in the `InterpretationError` below?

compor (approved):

superlopuh (approved):

xdsl/utils/exceptions.py
```diff
@@ -24,6 +24,11 @@ class VerifyException(DiagnosticException):
 
 
 class PassFailedException(DiagnosticException):
+    """
+    A diagnostic error which can be raised during the execution of a pass, used to
+    signify the the pass did not succeed.
+    """
```
compor, line 30:
```suggestion
    signify that the pass did not succeed.
    """
```
typo

compor: Story of my life

PR #3830, opened 2025-02-04
core: Disallow duplicate keys in attribute dictionaries

compor:
This PR:

- Disallows duplicate keys in properties and attribute dictionaries
- Tests of the above

Bumped into this by accident and the tests are similar to https://github.com/llvm/llvm-project/blob/46b1543dc04970719caab0d4f9f65699fea6adbc/mlir/test/IR/invalid-builtin-attributes.mlir#L485

I don't venture often in the parser, so any feedback is welcome.

### DIFF
--- a/tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
+++ b/tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
@@ -7,4 +7,3 @@
 
 "test.op"() <{key1, key1}> : () -> ()
 // CHECK: Duplicate key 'key1' in properties dictionary 
-
--- a/tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
+++ b/tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
@@ -1,9 +1,9 @@
 // RUN: xdsl-opt %s --parsing-diagnostics --split-input-file | filecheck %s
 
-"test.op"() {key1, key1} : () -> ()
 // CHECK: Duplicate key 'key1' in dictionary attribute
+"test.op"() {key1, key1} : () -> ()
 
 // -----
 
-"test.op"() <{key1, key1}> : () -> ()
 // CHECK: Duplicate key 'key1' in properties dictionary 
+"test.op"() <{key1, key1}> : () -> ()
--- a/tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
+++ b/tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
@@ -1,6 +1,11 @@
 // RUN: xdsl-opt %s --parsing-diagnostics --split-input-file | filecheck %s
 
 // CHECK: Duplicate key 'key1' in dictionary attribute
+"test.op"() {a = {key1, key1}} : () -> ()
+
+// -----
+
+// CHECK: Duplicate key 'key1' in dictionary attribute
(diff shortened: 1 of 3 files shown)


tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: xdsl-opt %s --parsing-diagnostics --split-input-file | filecheck %s
+
+"test.op"() {key1, key1} : () -> ()
```
alexarice, line 3: I believe we aim for putting the `CHECK` directive on the line above the actual code now. (This isn't followed everywhere but may as well try to have a convention)

tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
```diff
@@ -0,0 +1,9 @@
...
+// CHECK: Duplicate key 'key1' in dictionary attribute
```
alexarice, line 4: Nit: It would be nice if we could have the error be "Duplicate key 'key1' in attributes dictionary" though I appreciate this makes the code more complicated

tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: xdsl-opt %s --parsing-diagnostics --split-input-file | filecheck %s
```
alexarice, line 1: Could we get a test for a dictionary attribute?

alexarice (approved):
Seems good, left a couple of minor comments.

tests/filecheck/parser-printer/duplicate_attribute_keys.mlir
compor: MLIR doesn't do that and I'd rather keep it simple if you don't mind as printing the key name makes the location obvious.

alexarice: fair enough

xdsl/parser/core.py
```diff
@@ -792,13 +793,18 @@ def parse_optional_properties_dict(self) -> dict[str, Attribute]:
...
+        if self.parse_optional_punctuation("<") is None:
+            return dict()
+
+        entries = self.parse_comma_separated_list(
+            self.Delimiter.BRACES, self._parse_attribute_entry
+        )
+        self.parse_punctuation(">")
+
+        if (key := self._find_duplicated_key(entries)) is not None:
+            self.raise_error(f"Duplicate key '{key}' in properties dictionary")
```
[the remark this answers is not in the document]
compor: yeah, I wanted to, but `parse_optional_comma_separated_list` is part of the public API. I could pass an optional argument for the spans, but even then I'd need to modify the `_parse_attribute_entry` to keep them around as it's the callback used by `parse_optional_comma_separated_list` in this case.

I'm not sure I have the time to sink in parsing/printing ATM, unless it's something obvious I'm missing.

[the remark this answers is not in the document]
alexarice: It's not obvious, and would surely be messy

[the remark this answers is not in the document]
compor: You're not really selling it ;-)

math-fehr (approved):
<3

PR #3831, opened 2025-02-04
dialects: (builtin) remove AnyMemRefType

alexarice:
`MemRefType` now has a default generic, so `AnyMemRefType` is unnecessary.


compor (approved):
nice cleanup

PR #3833, opened 2025-02-04
dialects: (builtin/memref) rename Memref to MemRef everywhere

### LINKED ISSUE #3474: dialects: (memref) standardise naming of memref-related types to `MemRef` capitalization
In `MemRefType` the `R` is capitalised but it is not in `UnrankedMemrefType`.

alexarice:
It's time.

Makes the spelling consistent between `MemRefType` and `UnrankedMemrefType`.

Fixes #3474


docs/Toy/toy/__main__.py
```diff
@@ -104,7 +104,7 @@ def main(path: Path, emit: str, ir: bool, print_generic: bool):
         interpreter.register_implementations(AffineFunctions())
     if emit in ("affine", "scf"):
         interpreter.register_implementations(ArithFunctions())
-        interpreter.register_implementations(MemrefFunctions())
```
compor, line 107: Sorry to spoil the fun, but does this include internal methods like the above?
Isn't that a bit of an overkill and almost impossible to police in the future?

I thought we were only going to touch the type definition in `builtin`.

alexarice: I'm not sure I understand. I can restrict this to the type in `builtin` and the dialect name, but I'm not sure I see a reason against including things like the above.

alexarice: I've also already had the fun of emacs nearly crashing while trying to process the commit

math-fehr: I'm split here.
On the one hand, I agree with @compor and changing everything will make it a bit annoying downstream.
On the other hand, having something consistent is quite nice.

In any case, thanks a lot for this, I'll approve it for either choice you make

superlopuh: I think this is one of the less controversial changes in this PR, it's the dialect name, so seems like it should follow capitalization

superlopuh (approved):

PR #3835, opened 2025-02-05
dialects: Add more defaults to TypeVar

math-fehr:
Stacked PRs:
 * #3836
 * __->__#3835

--- --- ---

### dialects: Add more defaults to TypeVar

This includes default to AttributeInvT and AttributeCovT.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1818,7 +1818,7 @@ _MemRefTypeElement = TypeVar(
     "_MemRefTypeElement", bound=Attribute, covariant=True, default=Attribute
 )
 _UnrankedMemRefTypeElems = TypeVar(
-    "_UnrankedMemRefTypeElems", bound=Attribute, covariant=True
+    "_UnrankedMemRefTypeElems", bound=Attribute, covariant=True, default=Attribute
 )
 _UnrankedMemRefTypeElemsInit = TypeVar("_UnrankedMemRefTypeElemsInit", bound=Attribute)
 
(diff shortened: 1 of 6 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1031,23 +1031,28 @@ def __init__(self, types: list[Attribute] | ArrayAttr[Attribute]) -> None:
...
 
 
+_VectorTypeElement = TypeVar(
+    "_VectorTypeElement", bound=Attribute, covariant=True, default=Attribute
+)
+
+
 @irdl_attr_definition
 class VectorType(
-    Generic[AttributeCovT],
```
superlopuh, line 1036: Why would `AttributeCovT` not default to Attribute? There may be a downside I'm not appreciating, it just feels like it might make sense as an opt-out thing. My intuition is that since Attributes are immutable it's always safe to assume that the generic argument is covariant up to Attribute, and doesn't have to be Unknown

math-fehr: We can do that if you prefer, I just didn't want to introduce a huge debate on that.

superlopuh: I vote for this approach if you're also happy with it.

math-fehr: Should be done now

math-fehr:
As mentionned in https://github.com/xdslproject/xdsl/pull/3836, I'll merge both PRs at the same time, so that I won't write meaningless code just to fix the CI that will be fixed in the next PR

PR #3836, opened 2025-02-05
core: Add shorthand type for types that can be converted to IRDL

math-fehr:
Stacked PRs:
 * __->__#3836
 * #3835

--- --- ---

### core: Add shorthand type for types that can be converted to IRDL

There was a lot of inconsistencies in the typing of functions that are
using the type to constraint conversion. Having a shorthand makes this
more explicit.

### DIFF
--- a/xdsl/irdl/attributes.py
+++ b/xdsl/irdl/attributes.py
@@ -11,9 +11,11 @@ from dataclasses import dataclass
 from inspect import isclass
 from types import FunctionType, GenericAlias, UnionType
 from typing import (
+    TYPE_CHECKING,
     Annotated,
     Any,
     Generic,
+    TypeAlias,
     TypeVar,
     Union,
     cast,
@@ -21,6 +23,10 @@ from typing import (
     get_origin,
     get_type_hints,
 )
+
+if TYPE_CHECKING:
+    from typing_extensions import TypeForm
+
 
 from xdsl.ir import (
     Attribute,
@@ -256,8 +262,30 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
     )
 
 
+IRDLGenericAttrConstraint: TypeAlias = (
+    GenericAttrConstraint[AttributeInvT]
+    | Attribute
+    | type[AttributeInvT]
+    | "TypeForm[AttributeInvT]"
+    | ConstraintVar
+    | TypeVar
+)
+"""
+Attribute constraints represented using the IRDL python frontend. Attribute constraints
+can either be:
+- An instance of `AttrConstraint` representing a constraint on an attribute.
+- An instance of `Attribute` representing an equality constraint on an attribute.
+- A type representing a specific attribute class.
+- A TypeForm that can represent both unions and generic attributes.
+- A `ConstraintVar` representing a constraint variable.
+"""
+
+IRDLAttrConstraint = IRDLGenericAttrConstraint[Attribute]
+"""See `IRDLGenericAttrConstraint`."""
+
+
 def irdl_list_to_attr_constraint(
-    pyrdl_constraints: Sequence[Any],
+    pyrdl_constraints: Sequence[IRDLAttrConstraint],
     *,
     allow_type_var: bool = False,
     type_var_mapping: dict[TypeVar, AttrConstraint] | None = None,
@@ -299,7 +327,7 @@ def irdl_list_to_attr_constraint(
 
 
 def irdl_to_attr_constraint(
-    irdl: Any,
+    irdl: IRDLAttrConstraint,
     *,
     allow_type_var: bool = False,
     type_var_mapping: dict[TypeVar, AttrConstraint] | None = None,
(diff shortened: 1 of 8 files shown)


xdsl/utils/hints.py
```diff
@@ -93,7 +93,9 @@ def isa(arg: Any, hint: "TypeForm[_T]") -> TypeGuard[_T]:
     from xdsl.irdl import GenericData, irdl_to_attr_constraint
 
     if (origin is not None) and issubclass(origin, GenericData | ParametrizedAttribute):
-        constraint = irdl_to_attr_constraint(hint)
+        constraint = irdl_to_attr_constraint(
+            hint  # pyright: ignore[reportArgumentType]
```
superlopuh, line 97: why this? what's the error?

xdsl/irdl/attributes.py
```diff
@@ -256,8 +262,30 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
     )
 
 
+IRDLGenericAttributeConstraint: TypeAlias = (
+    GenericAttrConstraint[AttributeInvT]
+    | Attribute
+    | type[AttributeInvT]
+    | "TypeForm[AttributeInvT]"
+    | ConstraintVar
+    | TypeVar
```
alexarice, line 271: Adding `typing_extensions.TypeVar`  here would probably fix your previous PR.

xdsl/irdl/attributes.py
```diff
@@ -256,8 +262,30 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
     )
 
 
+IRDLGenericAttributeConstraint: TypeAlias = (
```
alexarice, line 265:
```suggestion
IRDLGenericAttrConstraint: TypeAlias = (
```
To align with `GenericAttrConstraint`?

xdsl/irdl/attributes.py
```diff
@@ -256,8 +262,30 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
...
+    | Attribute
+    | type[AttributeInvT]
+    | "TypeForm[AttributeInvT]"
+    | ConstraintVar
+    | TypeVar
+)
+"""
+Attribute constraints represented using the IRDL python frontend. Attribute constraints
+can either be:
+- An instance of `AttrConstraint` representing a constraint on an attribute.
```
alexarice, line 276: Would it be worth documenting the effect of an attribute constraint in a variadic? i.e. using `var_operand_def(my_attr_constr)` gives you a `RangeOf` constraint where each element of the range is constrained by `my_attr_constr`

xdsl/irdl/operations.py
```diff
@@ -615,7 +598,7 @@ class _SuccessorFieldDef(_OpDefField[SuccessorDef]):
 
 
 def result_def(
-    constraint: (AttrConstraint | Attribute | type[Attribute] | TypeVar) = Attribute,
+    constraint: IRDLAttributeConstraint = Attribute,
```
alexarice, line 601: why are some of these generic and some not?

alexarice: I see that this isn't actually changed so maybe lets not address this here

alexarice (approved):
Pyright errors seem to suggest a missing default value on the `UnrankedMemrefType` (Will need to rebase past `Memref` -> `MemRef` (sorry)) generic

xdsl/irdl/attributes.py
math-fehr: If that's okay, I'll not fix the last PR CI, and merge both commits at once.
Otherwise there are a lot of places to add `typing_extensions.TypeVar` that is going to be replaced by this commit anyway!

xdsl/irdl/attributes.py
math-fehr: Why would this be documented here? I feel this kind of documentation should be on the `var_operand_def` and al rather than there, or on attribute constraints directly?

xdsl/irdl/operations.py
math-fehr: To answer this, when you do `op.my_attr`, the attribute will always be correct.
But if you do `op.my_operand`, the operand might not be of the correct type because you are in the middle of a rewriting.
That's why one should keep the generic types while the other not.

But that will change later once someone implement conversion patterns.

xdsl/irdl/attributes.py
alexarice: Looking at it again that seems fair

xdsl/irdl/attributes.py
alexarice: Maybe we should just use `typing_extensions.TypeVar` everywhere

math-fehr: That's something I'm not 100% sure, because that's asking everyone to use typing_extensions.
I think we can just support both, I can add support in another PR

alexarice: Ah, I see what you meant by your message now

superlopuh (approved):

PR #3837, opened 2025-02-05
dialects: (linalg) enable generic printing in mlir conversion filecheck

jorendumoulin:
This PR adds generic printing to the mlir conversion filecheck

This will check whether the issues posed in #2959 are correctly resolved, by checking if mlir correctly parses the generic output of xdsl

There are still reminaing issues, solved in the following PRs:
(stacked on: ) #3838, #3839, #3840, #3841

PR #3838, opened 2025-02-05
dialects: (linalg) add hidden region to transpose op

jorendumoulin:
This adds a hidden region to the linalg.transpose op to ensure correct generic printing
Also changes permutation to a property instead of attribute.

This resolves the transpose op in #2959 

This has now been checked manually, and will be put in ci with #3837 (but for that 3 other ops need to be fixed, PRs incoming...)


tests/filecheck/mlir-conversion/with-mlir/dialects/linalg/invalid_ops.mlir
jorendumoulin, line 1: this filecheck had invalid generic format, so i converted it to use the custom format

superlopuh (approved):

compor (approved):

PR #3847, opened 2025-02-05
installation: fix package discovery

alexarice:
Partial revert of 27bf82162f438557520caf3500f8467dae4a016b, which broke downstream installation.

PR #3848, opened 2025-02-05
installation: fix packages discovery

superlopuh:
Fixes regression in #3815, turns out that it expects all the packages in the folder, not just the root.


superlopuh:
Closing in favour of #3847

PR #3850, opened 2025-02-05
core: Fix bug in extractor when having nested variables

math-fehr:
When having an attribute constraint variable `T` and a variable `U` that could infer the
variable `T` (for instance `U = Vector<T>`), the constraint `U` would not realize that it can infer `T`.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -2399,6 +2399,44 @@ def test_nested_inference():
     check_roundtrip(program, ctx)
 
 
+def test_nested_inference_variable():
+    """Check that Param<T> infers correctly T when Param<T> is nested in a variable."""
+
+    @irdl_attr_definition
+    class ParamOne(ParametrizedAttribute, TypeAttribute, Generic[_T]):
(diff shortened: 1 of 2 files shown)


alexarice (approved):
Seems fine but do you have a use case for this? I didn't include this as couldn't think of an operation that would actually need it.

superlopuh (approved):

math-fehr:
> Seems fine but do you have a use case for this? I didn't include this as couldn't think of an operation that would actually need it.

Yep, I got this in downstream code:
```
@irdl_op_definition
class StoreOp(IRDLOperation):
    name = "smt.array.store"

    _DOMAIN: ClassVar = VarConstraint("_DOMAIN", AnyAttr())
    _RANGE: ClassVar = VarConstraint("_RANGE", AnyAttr())
    _ARRAY: ClassVar = VarConstraint("_ARRAY", ArrayType.constr(_DOMAIN, _RANGE))

    array: Operand = operand_def(_ARRAY)
    index: Operand = operand_def(_DOMAIN)
    value: Operand = operand_def(_RANGE)

    res: OpResult = result_def(_ARRAY)
```

PR #3851, opened 2025-02-06
core: Fix check for forward refs by checking for ssa refs instead of block refs

emmau678:
Change check for forward block refs to forward SSA refs for correctness of the plural case


emmau678:
sorry - something seems to have gone wrong with rebase, will check it now

tests/filecheck/parser-printer/graph_region.mlir
```diff
@@ -53,6 +53,16 @@ builtin.module {
 
 // -----
 
+// A graph region that refers to values that are not defined in the module.
+
+builtin.module {
+    %0 = "test.termop"(%1, %2) : (i32, i32) -> i32
+}
+
+// CHECK: values %1, %2 were used but not defined
```
alexarice, line 62: Can the check statement go before the code? I'm also not sure the comment is necessary, the error message describes what is going on.

xdsl/parser/core.py
```diff
@@ -145,7 +145,7 @@ def parse_module(self, allow_implicit_module: bool = True) -> ModuleOp:
             value_names = ", ".join(
                 "%" + name for name in self.forward_ssa_references.keys()
             )
-            if len(self.forward_block_references.keys()) > 1:
+            if len(self.forward_ssa_references.keys()) > 1:
```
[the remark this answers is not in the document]
emmau678: ok thanks @alexarice, I updated the description

tests/filecheck/parser-printer/graph_region.mlir
```diff
@@ -73,7 +83,7 @@ builtin.module {
         "test.op"() : () -> ()
 }
 
-// CHECK:       /graph_region.mlir:72:4
+// CHECK:       /graph_region.mlir:82:4
```
alexarice, line 86: This is the fun I was hoping to avoid

tests/filecheck/parser-printer/graph_region.mlir
emmau678: moved it, thanks

alexarice: Unless github is being weird, it seems you moved a different one?

xdsl/parser/core.py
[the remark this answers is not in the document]
alexarice: I feel this what we do in most places. I was surprised that it tries to get the plural correct here (yet gets it wrong anyway)

compor:
Also, can we update the PR title? Shouldn't it be "core: Fix ..."

compor:
I approved by accident!

tests/filecheck/parser-printer/graph_region.mlir
```diff
@@ -64,20 +74,18 @@ builtin.module {
...
 
-// A block defined twice
+// CHECK:       /graph_region.mlir:89:4
+// CHECK-NEXT:      ^blockA:
+// CHECK-NEXT:      ^^^^^^^
+// CHECK-NEXT:      re-declaration of block 'blockA'
+// CHECK-NEXT:  originally declared here:
+// CHECK-NEXT:  /graph_region.mlir:13:4
+// CHECK-NEXT:      ^blockA:
+// CHECK-NEXT:      ^^^^^^^
```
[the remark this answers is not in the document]
emmau678: yes, I moved it by mistake but have reverted this now

xdsl/parser/core.py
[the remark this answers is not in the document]
superlopuh: TBH yeah it seems like a reasonable simplifying change. Maybe even something like this:

> values used but not defined: [%1, %2]
> values used but not defined: [%1]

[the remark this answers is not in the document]
compor: easier to grep

tests/filecheck/parser-printer/graph_region.mlir
superlopuh: I still think it's worth it :)

xdsl/parser/core.py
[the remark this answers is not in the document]
superlopuh: @emmau678 let's update to do something like this? maybe `values [%1, %2] used but not defined`?

tests/filecheck/parser-printer/graph_region.mlir
emmau678: I think they are similar except the first one only reports one value, so it's coming from a different if branch in core.py

xdsl/parser/core.py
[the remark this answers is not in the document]
emmau678: sorry had not seen this last comment

xdsl/parser/core.py
```diff
@@ -145,10 +145,10 @@ def parse_module(self, allow_implicit_module: bool = True) -> ModuleOp:
             value_names = ", ".join(
                 "%" + name for name in self.forward_ssa_references.keys()
             )
-            if len(self.forward_block_references.keys()) > 1:
-                self.raise_error(f"values {value_names} were used but not defined")
+            if len(self.forward_ssa_references.keys()) > 1:
```
superlopuh, line 148:
```suggestion
            self.raise_error(f"value used but not defined: [{value_names}]")
```

superlopuh: let's collapse this guy to simplify the code

tests/filecheck/parser-printer/graph_region.mlir
```diff
@@ -43,13 +43,21 @@ builtin.module {
 
 // -----
 
-// A graph region that refers to a value that is not defined in the module.
+// A graph region that refers to values that are not defined in the module.
+
+// CHECK: value used but not defined: [%1]
```
superlopuh, line 48:
```suggestion
// CHECK: values used but not defined: [%1]
```

xdsl/parser/core.py
superlopuh: the lines below are then not needed

compor (approved):

superlopuh (approved):

superlopuh:
It's more or less when all the points raised have been addressed (not necessarily all comments resolved) and you have at least one approval. In this case, it would have been ok to merge after your most recent commit, since Chris had already approved it and you had done exactly what I suggested so no need to wait for my approval since it's kind of implied. And it's always OK to disagree with my suggestion and just reply saying why you think something else should be done :)

emmau678:
> It's more or less when all the points raised have been addressed (not necessarily all comments resolved) and you have at least one approval. In this case, it would have been ok to merge after your most recent commit, since Chris had already approved it and you had done exactly what I suggested so no need to wait for my approval since it's kind of implied. And it's always OK to disagree with my suggestion and just reply saying why you think something else should be done :)

that's perfect, thank you. I'm happy that I've done everything so I'll merge it

PR #3854, opened 2025-02-06
documentation: fix l1 scf example

superlopuh:
A bug slipped in. In the future we will want to look for ways to test these snippets, maybe, but that's not trivial today.

CC @tobiasgrosser

### DIFF
--- a/docs/marimo/mlir_ir.py
+++ b/docs/marimo/mlir_ir.py
@@ -201,8 +201,8 @@ def _(mo):
 @app.cell(hide_code=True)
 def _():
     l1_dist_scf_text = """\
-    func.func @l1_dist(%a : i32, %b : i32) -> (i32) {
-      %slt = arith.cmpi slt, %a, %b : i1
+    func.func @l1_dist(%a : i32, %b : i32) -> i32 {
+      %slt = arith.cmpi slt, %a, %b : i32
       %res = scf.if %slt -> (i32) {
         %b_minus_a = arith.subi %b, %a : i32
         scf.yield %b_minus_a : i32
--- a/docs/marimo/mlir_ir.py
+++ b/docs/marimo/mlir_ir.py
@@ -1,6 +1,6 @@
 import marimo
 
-__generated_with = "0.10.17"
+__generated_with = "0.11.0"
 app = marimo.App(width="medium")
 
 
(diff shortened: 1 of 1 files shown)


docs/marimo/mlir_ir.py
```diff
@@ -201,8 +201,8 @@ def _(mo):
 @app.cell(hide_code=True)
 def _():
     l1_dist_scf_text = """\
-    func.func @l1_dist(%a : i32, %b : i32, %c : i32) -> (i32) {
-      %slt = arith.cmpi slt, %lhs, %rhs : i1
+    func.func @l1_dist(%a : i32, %b : i32) -> (i32) {
+      %slt = arith.cmpi slt, %a, %b : i1
```
alexarice, line 205:
```suggestion
      %slt = arith.cmpi slt, %a, %b : i32
```

docs/marimo/mlir_ir.py
```diff
@@ -201,8 +201,8 @@ def _(mo):
 @app.cell(hide_code=True)
 def _():
     l1_dist_scf_text = """\
-    func.func @l1_dist(%a : i32, %b : i32, %c : i32) -> (i32) {
-      %slt = arith.cmpi slt, %lhs, %rhs : i1
+    func.func @l1_dist(%a : i32, %b : i32) -> (i32) {
```
alexarice, line 204:
```suggestion
    func.func @l1_dist(%a : i32, %b : i32) -> i32 {
```

docs/marimo/mlir_ir.py
```diff
@@ -1,6 +1,6 @@
 import marimo
 
-__generated_with = "0.10.17"
+__generated_with = "0.11.0"
```
alexarice, line 3: I have not opened this with marimo, but will assume this change doesn't break stuff

alexarice (approved):

docs/marimo/mlir_ir.py
superlopuh: yeah it's just an automatic version bump on save

superlopuh: I'm not sure if there's an easy way to do this in the CI when updating marimo with dependabot

Issue #3569, opened 2024-12-04
interpreter: (pdl) fix multiple inserted operations in rewrite pattern interpreter


jianyicheng:
(a + b) + c -> a + (b + c)

superlopuh:
I would recommend doing a different approach to the one here: https://github.com/xdslproject/xdsl/pull/3402/files#diff-571b0c5cfc943738c6c960434941019e3e0fdc733da7c96dc242bd727e182ff6R402-R415

Instead, accumulating operations created in the pass to a work list, and adding them in the end, instead of getting them from a use-def chain at the end.

superlopuh:
@cowardsa to provide example

superlopuh:
I'm not sure I understand your question, but my understanding is that the expected PDL pattern is only the last line, not the previous one, as the newresult will have the same value as result, and newaccumop will not have the same value as accumop.
Like if you have `x = (a + b); y = x + c; y = x + 2`, your rewrite would also rewrite `x` to `b + c`. But our current interpreter has a bug where the operations that are created in the `pdl.rewrite` are not inserted.

PR #3862, opened 2025-02-06
[WIP] interpreter (pdl): rewrites with multiple inserted operations

jumerckx:
ref: #3569 
Not yet in a mergeable state but I got a test case working.

### DIFF
--- a/tests/filecheck/transforms/apply-eqsat-pdl/apply_eqsat_pdl_multi_result.mlir
+++ b/tests/filecheck/transforms/apply-eqsat-pdl/apply_eqsat_pdl_multi_result.mlir
@@ -1,0 +1,66 @@
+builtin.module {
+  pdl.pattern : benefit(1) {
+    %x = pdl.operand
+    %y = pdl.operand
+    %z = pdl.operand
+    %type = pdl.type
+    %accumop = pdl.operation "arith.addi" (%x, %y : !pdl.value, !pdl.value) -> (%type : !pdl.type)
+    %accum = pdl.result 0 of %accumop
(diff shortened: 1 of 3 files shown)


compor:
good stuff @jumerckx I'll just rename it to WIP and you can undo that when this graduates from being a draft :)

superlopuh:
Let's discuss this during the meeting, but the issue addressed here is about pdl, not eqsat-pdl. After some discussions with @math-fehr and @qaco, it looks like the thing to do is just to insert all operations created inside the rewrite op. After this works in the pdl interpreter, we should take a look at the eqsat-pdl side of things. I'm happy to do the first thing next week, but it might be a good starter task on the vanilla pdl side if you're up for it @jumerckx

jumerckx:
> it looks like the thing to do is just to insert all operations created inside the rewrite op

Right, I believe this is essentially what I ended up doing [here](https://github.com/xdslproject/xdsl/blob/d6a2c76790a43e7314bed9451e316df43b21f29a/xdsl/interpreters/eqsat_pdl.py#L258C1-L260C1) based on your comments in the project tracker as well.
Seems straightforward to do the same for the non-eqsat pdl 🤞

> I'm happy to do the first thing next week, but it might be a good starter task on the vanilla pdl side if you're up for it @jumerckx

Thanks, I should be able to look at it the coming week as well, unless you need it fast. I've been busier than hoped for which is why I haven't pushed more yet, but I very much intend to keep going :).

superlopuh:
Not quite, the idea is just to insert the operation in the `pdl.operation` implementation, no need to track with state on the interpreter functions.

superlopuh:
There's no rush!

superlopuh:
Fixed with #3949

PR #3864, opened 2025-02-07
backend: (riscv) split out register queue interface from implementation

superlopuh:
In preparation for ARM and x86 backends, this is the first PR towards interface-based, target-agnostic register allocation.

### DIFF
--- a/xdsl/backend/riscv/register_allocation.py
+++ b/xdsl/backend/riscv/register_allocation.py
@@ -6,7 +6,7 @@ from typing import cast
 
 from ordered_set import OrderedSet
 
-from xdsl.backend.riscv.register_queue import RegisterQueue
+from xdsl.backend.register_queue import RegisterQueue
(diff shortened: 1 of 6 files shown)


xdsl/backend/riscv/register_allocation.py
```diff
@@ -22,10 +22,12 @@
...
 
 
-def gather_allocated(func: riscv_func.FuncOp) -> set[RISCVRegisterType]:
+def gather_allocated(
+    func: riscv_func.FuncOp,
+) -> set[IntRegisterType | FloatRegisterType]:
     """Utility method to gather already allocated registers"""
 
-    allocated: set[RISCVRegisterType] = set()
+    allocated: set[IntRegisterType | FloatRegisterType] = set()
```
compor, line 31: Why this change?

superlopuh: I'm not 100% sure it was necessary, but the type variable is invariant, which makes things a little weird, now all the types match exactly.

PR #3866, opened 2025-02-07
backend: (x86) Fix register sets test coverage

### LINKED ISSUE #2353: dialects: (x86) fully cover x86 register class with tests
https://app.codecov.io/gh/xdslproject/xdsl/pull/2333?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=xdslproject

There are two-three pytest tests missing for register names.

compor:
Resolves #2353

### DIFF
--- a/xdsl/dialects/x86/register.py
+++ b/xdsl/dialects/x86/register.py
@@ -21,15 +21,13 @@ class X86RegisterType(RegisterType):
         if parser.parse_optional_punctuation("<") is not None:
             name = parser.parse_identifier()
             parser.parse_punctuation(">")
-            if not name.startswith("e") and not name.startswith("r"):
-                assert name in cls.abi_index_by_name(), f"{name}"
         else:
             name = ""
         return cls._parameters_from_spelling(name)
 
     def verify(self) -> None:
         name = self.spelling.data
-        if not self.is_allocated or name.startswith("e") or name.startswith("r"):
+        if not self.is_allocated:
             return
         if name not in type(self).abi_index_by_name():
             raise VerifyException(f"{name} not in {self.instruction_set_name()}")
--- a/xdsl/tools/command_line_tool.py
+++ b/xdsl/tools/command_line_tool.py
@@ -8,7 +8,7 @@ from xdsl.context import MLContext
 from xdsl.dialects import get_all_dialects
 from xdsl.dialects.builtin import ModuleOp
 from xdsl.parser import Parser
-from xdsl.utils.exceptions import ParseError
+from xdsl.utils.exceptions import DiagnosticException, ParseError
 from xdsl.utils.lexer import Span
 
 
@@ -114,6 +114,11 @@ class CommandLineTool:
             s = e.span
             e.span = Span(s.start, s.end, s.input, start_offset)
             if "parsing_diagnostics" in self.args and self.args.parsing_diagnostics:
+                print(e)
+            else:
+                raise
+        except DiagnosticException as e:
+            if self.args.parsing_diagnostics:
                 print(e)
             else:
                 raise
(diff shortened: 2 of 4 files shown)


xdsl/tools/command_line_tool.py
```diff
@@ -117,5 +117,10 @@ def parse_chunk(
                 print(e)
             else:
                 raise
+        except DiagnosticException as e:
+            if self.args.verify_diagnostics:
+                print(e)
+            else:
+                raise
```
compor, line 124: I couldn't trigger the expected behaviour for the above filechecks without this here. It seems to trigger only upon `apply_passes` invocation, not by just parsing an mlir file.

Worth a separate PR or a workaround I'm missing?

xdsl/dialects/x86/register.py
```diff
@@ -21,15 +21,13 @@ def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
...
-            if not name.startswith("e") and not name.startswith("r"):
-                assert name in cls.abi_index_by_name(), f"{name}"
         else:
             name = ""
         return cls._parameters_from_spelling(name)
 
     def verify(self) -> None:
         name = self.spelling.data
-        if not self.is_allocated or name.startswith("e") or name.startswith("r"):
+        if not self.is_allocated:
```
compor, line 30: This was overly permissive

xdsl/dialects/x86/register.py
```diff
@@ -21,15 +21,13 @@ def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
         if parser.parse_optional_punctuation("<") is not None:
             name = parser.parse_identifier()
             parser.parse_punctuation(">")
-            if not name.startswith("e") and not name.startswith("r"):
-                assert name in cls.abi_index_by_name(), f"{name}"
```
compor, line 25: This aligns to what we do in the `arm` dialect

xdsl/tools/command_line_tool.py
superlopuh: CC @math-fehr @alexarice seems like a bug to me, I would expect the tests above to work

KGrykiel (approved):

superlopuh (approved):
LGTM modulo reply from @math-fehr on the exception thing

xdsl/tools/command_line_tool.py
alexarice: Seems this catch was missing. In my opinion, we should move the catches for `parsing_diagnostics` and `verify_diagnostics` to `xDSLOptMain.run` so that all the logic is in one place. (Not necessarily in this PR)

alexarice: It's currently a bit of a mess. This function `parse_chunk` is a method of `CommandLineTool` but is referencing an argument `parsing-diagnostics` which is only defined on the subclass `xDSLOptMain`

compor: Yeah, the split is a bit all over the place.
I'd suggest opening a separate tiny PR on this, so it unblocks this PR, and we can move on after.

alexarice: @compor Do you want to just merge this then I can try to cook up a quick PR to sort this out on top?

compor: You don't need to cook anything, and I'm not in a hurry, I'd rather wait for @math-fehr on this

alexarice (approved):
Happy for this to be merged and leave the discussion about diagnostic flags to later

math-fehr (approved):
Sorry, I didn't see this!
I think the exception is fine!

compor:
Thanks we can short out the split after

PR #3868, opened 2025-02-07
misc: remove xdsl.irdl.error.error function

superlopuh:
The auto-generated docs website complained about this function with the same name as its module. We actually have a helper for it already on Operation, and can just remove it.


math-fehr (approved):

compor (commented):
👍🏽

PR #3869, opened 2025-02-08
misc: reduce max-line-length ruff setting to 130 [NFC]

superlopuh:
I don't know if we'll ever get to 88, but it feels like it would be good to get the line lengths a little more under control. In this PR, I update the line length, ignore all test files, ignore the line length limit on all assembly format strings, reflow some doc strings, and split up some inline strings in the code.

After this change, there are 840 errors when I set the line length to 88, and 313 if I set it to 100. I think without some sort of AI "fix everything" button it's a waste of time to go shorter, but our current setting of 300 definitely seems to big.

This PR contains about 60 whitespace changes.

### DIFF
--- a/tests/dialects/test_transform.py
+++ b/tests/dialects/test_transform.py
@@ -78,7 +78,7 @@ def test_tileop_init():
             dynamic_sizes=[],
             static_sizes=static_sizes,
         ),
-        """%0, %1, %2 = "transform.structured.tile_using_for"(%3) <{static_sizes = array<i32: 8, 8>}> : (!transform.any_value) -> (!transform.any_op, !transform.any_op, !transform.any_op)""",  # noqa: E501
+        """%0, %1, %2 = "transform.structured.tile_using_for"(%3) <{static_sizes = array<i32: 8, 8>}> : (!transform.any_value) -> (!transform.any_op, !transform.any_op, !transform.any_op)""",
         None,
     )
 
@@ -115,7 +115,7 @@ def test_get_producer_of_operand():
     target = test.TestOp(result_types=[transform.AnyValueType()]).results[0]
     assert_print_op(
         transform.GetProducerOfOperandOp(operand_number=0, target=target),
-        """%0 = "transform.get_producer_of_operand"(%1) <{operand_number = 0 : i64}> : (!transform.any_value) -> !transform.any_op""",  # noqa: E501
+        """%0 = "transform.get_producer_of_operand"(%1) <{operand_number = 0 : i64}> : (!transform.any_value) -> !transform.any_op""",
         None,
     )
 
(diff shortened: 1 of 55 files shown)


tests/test_pyrdl.py
```diff
@@ -225,12 +225,12 @@ def test_allof_verify_multiple_failures():
     """
     constraint = AllOf((LessThan(5), GreaterThan(8)))
 
-    with pytest.raises(VerifyException) as e:
+    with pytest.raises(
+        VerifyException,
+        match=f"The following constraints were not satisfied:\n{IntData(7)} should "
+        f"hold a value less than 5\n{IntData(7)} should hold a value greater than 8",
+    ):
```
superlopuh, line 232: the only functional-ish change

xdsl/dialects/ptr.py
```diff
@@ -1,12 +1,16 @@
-########
-#  This is a port of the 'ptr' dialect proposed in this thread https://discourse.llvm.org/t/rfc-ptr-dialect-modularizing-ptr-ops-in-the-llvm-dialect/75142/21
-#  Main parts of the dialect has already been agreed upon but only the basic datatype is implemented upstream.
-#  When the upstream implementation is merged we should fix our implementation to be consistent with the mlir version.
-#  Current diviations:
-#  (1) Name of the dialect should be changed: ptr_xdsl -> ptr. We currently chose another name to avoid conflicts when loading our dialect to mlir-opt.
-#  (2) There is no ptr.to_ptr operation currently proposed, but there is a memref.to_ptr. We do this so that we can feed the results of convert-memref-to-ptr pass without any conflict.
-########
-
+"""
```
superlopuh, line 1: this was a weird pseudo doc comment I made it an actual doc comment

tests/dialects/test_transform.py
```diff
@@ -78,7 +78,7 @@ def test_tileop_init():
             dynamic_sizes=[],
             static_sizes=static_sizes,
         ),
-        """%0, %1, %2 = "transform.structured.tile_using_for"(%3) <{static_sizes = array<i32: 8, 8>}> : (!transform.any_value) -> (!transform.any_op, !transform.any_op, !transform.any_op)""",
+        """%0, %1, %2 = "transform.structured.tile_using_for"(%3) <{static_sizes = array<i32: 8, 8>}> : (!transform.any_value) -> (!transform.any_op, !transform.any_op, !transform.any_op)""",  # noqa: E501
         None,
     )
```
compor, line 83: Do we need these noqa comments here since we ignore all test file above (`"tests/*" = ["E501"]`)?
When I remove it, `ruff check` doesn't trigger for me unless I also remove the aforementioned line from `pyproject.toml`.

compor (approved):

compor:
I'm happy with 100 chars of line length, although most OSS Python project seem to adhere to the PEP8 88 gospel.

Related also to #1274 (maybe close it with a comment and pointing to this PR?)

tests/dialects/test_transform.py
superlopuh: Ah you're right I'll do a pass

superlopuh:
Let's keep that open for now, and close when/if we get to 100?

compor (approved):

PR #3872, opened 2025-02-10
backend: unify assembly printing helpers in AssemblyPrinter

superlopuh:
Now that we're adding more backend infrastructure, I think it would make sense to unify some of the helpers, and iterate on the logic. This PR introduces an AssemblyPrinter class that holds the helpers previously copy/pasted into each of the riscv, x86, and ARM dialects.

### DIFF
--- a/tests/filecheck/dialects/x86/x86_assembly_emission.mlir
+++ b/tests/filecheck/dialects/x86/x86_assembly_emission.mlir
@@ -148,7 +148,7 @@ x86.directive ".align" "2"
 x86.label "label"
 // CHECK-NEXT: label:
 
-func.func @funcyasm() {
+x86_func.func @funcyasm() {
     %3 = x86.get_register : () -> !x86.reg<rax>
     %4 = x86.get_register : () -> !x86.reg<rdx>
     %rflags = x86.rr.cmp %3, %4 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.rflags<rflags>
--- a/xdsl/backend/assembly_printer.py
+++ b/xdsl/backend/assembly_printer.py
@@ -1,0 +1,66 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+from xdsl.dialects.builtin import ModuleOp, StringAttr
+from xdsl.utils.base_printer import BasePrinter
+
+
+class AssemblyPrinter(BasePrinter):
+    @staticmethod
+    def append_comment(line: str, comment: StringAttr | None) -> str:
+        if comment is None:
+            return line
+
+        padding = " " * max(0, 48 - len(line))
+
+        return f"{line}{padding} # {comment.data}"
+
+    @staticmethod
+    def assembly_line(
+        name: str,
+        arg_str: str,
+        comment: StringAttr | None = None,
+        is_indented: bool = True,
+    ) -> str:
+        code = "    " if is_indented else ""
+        code += name
+        if arg_str:
+            code += f" {arg_str}"
+        code = AssemblyPrinter.append_comment(code, comment)
+        return code
+
+    def print_module(self, module: ModuleOp) -> None:
+        for op in module.body.walk():
+            assert isinstance(op, AssemblyPrintable), f"{op}"
+            op.print_assembly(self)
+
+
+class AssemblyPrintable(ABC):
+    """
+    Base class for operations that can be a part of assembly printing.
+    """
+
+    @abstractmethod
+    def print_assembly(self, printer: AssemblyPrinter) -> None:
+        raise NotImplementedError()
+
+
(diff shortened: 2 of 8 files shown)


tests/filecheck/dialects/x86/x86_assembly_emission.mlir
```diff
@@ -148,7 +148,7 @@ x86.directive ".align" "2"
 x86.label "label"
 // CHECK-NEXT: label:
 
-func.func @funcyasm() {
+x86_func.func @funcyasm() {
```
superlopuh, line 151: the previous x86 assembly printing had a special handling for func, the only functional change to this PR is that now we require func to be lowered to x86_func before assembly printing.

xdsl/backend/assembly_printer.py
```diff
@@ -0,0 +1,66 @@
...
+        code += name
+        if arg_str:
+            code += f" {arg_str}"
+        code = AssemblyPrinter.append_comment(code, comment)
+        return code
+
+    def print_module(self, module: ModuleOp) -> None:
+        for op in module.body.walk():
+            assert isinstance(op, AssemblyPrintable), f"{op}"
+            op.print_assembly(self)
```
superlopuh, line 36: these are just copied from the other files

xdsl/backend/assembly_printer.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class AssemblyPrintable(ABC):
+    """
+    Base class for operations that can be a part of assembly printing.
+    """
+
+    @abstractmethod
+    def print_assembly(self, printer: AssemblyPrinter) -> None:
+        raise NotImplementedError()
```
superlopuh, line 46: currently each operation prints one or zero lines of assembly, but I'm now thinking that it would make sense to make this more flexible. After this gets merged, I'll look into riscv_func, and how we could print the various directives to do with alignment etc just from the func itself.

xdsl/dialects/riscv.py
```diff
@@ -524,26 +514,9 @@ def _assembly_arg_str(arg: AssemblyInstructionArg) -> str:
...
-
-
 def print_assembly(module: ModuleOp, output: IO[str]) -> None:
-    for op in module.body.walk():
-        assert isinstance(op, RISCVAsmOperation), f"{op}"
-        asm = op.assembly_line()
-        if asm is not None:
-            print(asm, file=output)
+    printer = AssemblyPrinter(stream=output)
+    printer.print_module(module)
```
superlopuh, line 519: I wanted to keep this around not just to minimise diff but also to allow for future RISCVAssemblyPrinter or whatever if we want to add some special helper functionality.

xdsl/dialects/x86/ops.py
```diff
@@ -2582,14 +2582,8 @@ class GetAVXRegisterOp(GetAnyRegisterOperation[AVXRegisterType]):
 
 
 def print_assembly(module: ModuleOp, output: IO[str]) -> None:
-    for op in module.body.walk():
-        if isinstance(op, FuncOp):
```
superlopuh, line 2586: the code change that required the test change above

xdsl/backend/assembly_printer.py
compor: yeah, maybe it would make sense to have some related state held here.

compor (approved):

emmau678 (approved):
looks good to me

xdsl/backend/assembly_printer.py
```diff
@@ -0,0 +1,66 @@
...
+        code = AssemblyPrinter.append_comment(code, comment)
+        return code
+
+    def print_module(self, module: ModuleOp) -> None:
+        for op in module.body.walk():
+            assert isinstance(op, AssemblyPrintable), f"{op}"
+            op.print_assembly(self)
+
+
+class AssemblyPrintable(ABC):
```
alexarice, line 39:
```suggestion
class AssemblyPrintable(Operation, ABC):
```
Is it worth adding this to agree with the docstring below (I'm not sure if this is good practice in general).

xdsl/backend/assembly_printer.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Returns the line that should be printed in assembly, or `None` to not be
+        printed.
+        """
+        raise NotImplementedError()
+
+    def print_assembly(self, printer: AssemblyPrinter) -> None:
+        line = self.assembly_line()
+        if line is not None:
+            printer.print_string(line + "\n")
```
alexarice, line 66:
```suggestion
            printer.print_string(line)
            printer.print_string("\n")
```
avoid string concatenation? Wondering if there's a way we can bypass some of the newline checking with the assumption that the input string is one line here.

alexarice (approved):
Don't know much about the assembly side but seems reasonable.

PR #3874, opened 2025-02-10
installation: specify find

superlopuh:
Reopens #3848, which fixes a regression introduced by #3815, and partially fixed with #3847.

While the issue is fixed on most installations with the other fix PR it breaks on some systems, and when a `site` folder is present after generating the docs website locally.

We tested this change with @georgebisbas today, on whose linux machine xDSL cannot currently be installed, and works after applying the changes in this PR.


georgebisbas (approved):
Yes, this was tested earlier with @superlopuh .  Installing in dev mode (in a venv) after cloning the repo could not import the right modules.

compor (approved):

PR #3875, opened 2025-02-10
documentation: do a pass over MLIR IR notebook

superlopuh:


docs/marimo/mlir_ir.py
```diff
@@ -44,28 +44,7 @@ def _(mo, triangle_text):
...
-        3. The things immediately to the right of the assignments are in the form `first.second`
-
-        These things are the names of operations.
-        These operations are the core building blocks of MLIR IR, and their structure and meaning is indicated by this name.
-        The names are all in two parts, where the first part is the name of a dialect, a kind of namespace for related concepts, and the second makes the operation name unique within the dialect.
-
-        With this in mind, let's zoom in to the first operation.
-        """
-    )
+    mo.md(r"""This notebook explains all the components of the above snippet of code.""")
```
math-fehr, line 47: Why removing this? This is quite useful outside of workshops

math-fehr (approved):

docs/marimo/mlir_ir.py
superlopuh: The broad feedback was that the wall of text is not a great start to the notebook. I'll intersperse these in the relevant parts below.

compor (approved):
I also have the following suggestions for this notebook:

1) "MLIR and xDSL use an encoding of the IR as a textual format for debugging, testing, and storing intermediate representations of programs."

to

"MLIR and xDSL use a textual encoding of the IR for debugging, testing, and storing intermediate representations of programs."

Otherwise it is a bit too close to the "assembly textual format" and at the end of the way formatting is different than textual vs bytecode representation.

2) "The [func dialect](https://mlir.llvm.org/docs/Dialects/Func/) contains building blocks to model functions and function calls."

to 

"The [func dialect](https://mlir.llvm.org/docs/Dialects/Func/) contains building blocks to model function definitions and their calls."

3) "The above function takes two 32-bit integers, and returns them in the opposite order "

Missing a period at the end of this sentence.

4) The cell "Unhide the cell below to see the solution:" is not visible for me? I don't get an option to unhide anything. Same for all appearances of this in the notebook.


5) "Instead, the loop body takes in some number of immutable values and yields the same number of values to use for the next loop. After all the iterations, these values are returned by the operation."

to

"Instead, the loop body takes in a number of immutable values and yields the same number of values to use for the next iteration. When all iterations are complete, these values are returned by the operation."

compor (approved):

PR #3881, opened 2025-02-11
docs: Add IR generation notebook

math-fehr:
Stacked PRs:
 * #3884
 * #3883
 * #3882
 * __->__#3881
 * #3886

--- --- ---

### DIFF
--- a/docs/marimo/ir_gen.py
+++ b/docs/marimo/ir_gen.py
@@ -1,0 +1,655 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "marimo",
+#     "sympy==1.13.3",
+#     "xdsl==0.27.0",
+# ]
+# ///
+
+import marimo
+
+__generated_with = "0.11.0"
+app = marimo.App()
+
+
+@app.cell(hide_code=True)
+def _():
+    import marimo as mo
+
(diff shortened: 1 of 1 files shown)


superlopuh:
It feels like we should update the test script to launch in a way that takes the versions into account.

compor:
also, why are these stacked?

math-fehr:
I stacked them because it's really simple for me to convert 4 commits into 4 PRs.
Otherwise there are no particular reasons

compor:
what level of review are we interested in here? is it just to get them in and then iterate over?

For example, in this notebook I see there following:

2025-02-11-132549_771x191_scrot

i.e., no SymPy examples.

I don't remember this part in the Winter School, but is there supposed to be something there?

math-fehr:
> what level of review are we interested in here? is it just to get them in and then iterate over?
> 
> For example, in this notebook I see there following:
> 
> 2025-02-11-132549_771x191_scrot
> 
> i.e., no SymPy examples.
> 
> I don't remember this part in the Winter School, but is there supposed to be something there? Unless it's the missing deps?

I fixed this here, but overall I don't really have much time to fix all these things because of the PLDI rebuttal. The plan is to just have something we can CI check, and we can clean things bit by bit

PR #3888, opened 2025-02-11
backend: (x86) fix assembly when the offset of a memory access is 0

qaco:


xdsl/dialects/x86/assembly.py
```diff
@@ -78,7 +78,7 @@ def memory_access_str(
     register: AssemblyInstructionArg, offset: IntegerAttr | None
 ) -> str:
     register_str = assembly_arg_str(register)
-    if offset is not None:
```
superlopuh, line 81: I think this was an error in the original code, or maybe it was added before we had default values. I would recommend making the offset never None, and instead not printing it in the assembly format when it's 0

qaco: Yeay I agree

compor (approved):

compor (approved):

xdsl/dialects/x86/ops.py
```diff
@@ -557,15 +557,15 @@ class R_RM_Operation(
 
     r1 = operand_def(R1InvT)
     r2 = operand_def(R2InvT)
-    offset = opt_attr_def(IntegerAttr)
+    offset = attr_def(IntegerAttr)
```
superlopuh, line 560: We actually have functionality to make the attribute have a default value if it's not defined. I would actually recommend doing it this way:
```suggestion
    offset = attr_def(IntegerAttr, default_value=IntegerAttr(0, 64))
```

tests/filecheck/dialects/x86/x86_assembly_emission.mlir
```diff
@@ -41,6 +41,8 @@
 
 %rm_add_no_offset = x86.rm.add %0, %1 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
 // CHECK-NEXT: add rax, [rdx]
+%rm_add_offset_zero = x86.rm.add %0, %1, 0 : (!x86.reg<rax>, !x86.reg<rdx>) -> !x86.reg<rax>
+// CHECK-NEXT: add rax, [rdx]
```
superlopuh, line 45: It seems off to me to allow both of these formats. We should either not print the offset if it's 0, or always print it.

compor: sorry, I am not sure I understand, isn't this what is happening now (i.e., not printing the offset when it's 0)?

superlopuh: The line above does not print it, and this one does

qaco: We should not print the 0, but we should be able to parse it, right ? In my understanding it is the point here.

xdsl/dialects/x86/ops.py
```diff
@@ -62,6 +62,27 @@
...
 
+def parse_memory_access_offset(parser: Parser) -> IntegerAttr:
+    offset_type = IntegerType(64, Signedness.SIGNED)
+    temp = parse_optional_immediate_value(
+        parser,
+        offset_type,
+    )
+    if temp is None:
+        temp = IntegerAttr(0, offset_type)
+    assert isinstance(temp, IntegerAttr)
```
superlopuh, line 73: this is recommended against, the verification is supposed to do this, the function should return an `Attribute`

xdsl/dialects/x86/ops.py
```diff
@@ -62,6 +62,27 @@
...
+    return temp
+
+
+def parse_comma_memory_access_offset(parser: Parser) -> IntegerAttr:
```
superlopuh, line 77: This seems like something that would make more sense inline, let's take a look at this helper together later?

tests/filecheck/dialects/x86/x86_assembly_emission.mlir
superlopuh: Ah that's a good point. In that case the diff in this file LGTM

xdsl/dialects/x86/ops.py
qaco: ```temp``` can be the return value of ```parse_optional_immediate_value``` which is of type ```IntegerAttr[IntegerType | IndexType] | LabelAttr | None```

superlopuh: Sure, this is a subtype of `Attribute | None`, so we can return a value of type `Attribute` without any type errors

xdsl/dialects/x86/ops.py
```diff
@@ -91,6 +91,25 @@ def parse(cls, parser: Parser) -> Self:
...
+            temp = IntegerAttr(0, offset_type)
+        return temp
+
+    @classmethod
+    def parse_comma_memory_access_offset(cls, parser: Parser) -> Attribute:
+        offset_type = IntegerType(64, Signedness.SIGNED)
+        temp = IntegerAttr(0, offset_type)
+        if parser.parse_optional_punctuation(",") is not None:
+            temp = cls.parse_memory_access_offset(parser)
+        return temp
```
superlopuh, line 111: Does this work? Without changing the lines below.
```suggestion
    @classmethod
    def parse_memory_access_offset(cls, parser: Parser) -> Attribute | None:
        return parse_optional_immediate_value(
            parser,
            IntegerType(64, Signedness.SIGNED),
        )

    @classmethod
    def parse_comma_memory_access_offset(cls, parser: Parser) -> Attribute | None:
        if parser.parse_optional_punctuation(",") is not None:
            return cls.parse_memory_access_offset(parser)
```

superlopuh (approved):

PR #3889, opened 2025-02-12
bench: add back benchmark and profiling infrastructure

EdmundGoodman:
## Summary

This PR moves the existing benchmarks and profiling infrastructure from <https://github.com/xdslproject/xdsl-bench> into the main xDSL repo so they can be used more easily by developers. In addition to this, it restructures and extends them to provide more microbenchmarks, component-level benchmarks, and end-to-end benchmarks.

A full description of the motivate and structure is available in `benchmarks/README.md`.

## Future work

- [x] Modify <https://github.com/xdslproject/xdsl-bench> to run against new benchmark repo (blocked by merging this PR)
- [ ] Add further benchmarks, with main areas for development being (to be done as needed within the provided framework)
   - Evaluate additional components, such as the printer and verifier
   - Evaluate larger "hero" end-to-end workloads, such as CIRCT, ASL, or quantum
   - Make synthetic data driving microbenchmarks more representative
- [x] Consider reducing the amount of MLIR duplicated from the test directory, by leveraging the `mlir-opt` installation in CI to generate it on the fly


EdmundGoodman:
> Is there a way to avoid duplicating the files? I was expecting this to be more minimal

If I remove the benchmarks which parse/lex over all test MLIR files instead of individual ones that should cut down the size a lot?

EdmundGoodman:
OK, this looks a bit more sensible in the number of files changed and shouldn't impact the expressiveness of the tests too much.

benchmarks/Makefile
```diff
@@ -0,0 +1,46 @@
...
+MAKEFLAGS += --warn-undefined-variables
+SHELL := bash
+
+# ===== #
+#  ASV  #
+# ===== #
+
+.PHONY: asv
+asv:
+	cd .. && uv run asv run
```
compor, line 11: this `..` feels a bit fragile, especially repeated at every make target.

Can't we use the `--directory` option along with exposing this as a make var along the whole Makefile?

EdmundGoodman: I've just moved the `asv`, `asv-html, `asv-preview`, and `asv-clean` targets up a directory into the main Makefile as that probably makes a bit more sense for where developers would want to run them from anyway?

EdmundGoodman:
I'm not sure why the lockfile CI is failing -- I have tried both re-running `uv lock` locally and remade the venv.

Other than that and any other changes needed, I think this is good to go? I've got the dev branch of xdsl-bench [working](https://github.com/xdslproject/xdsl-bench/actions/runs/13372257098/job/37343307061) pointing at this branch so once merged it should be easy to update that as well.

superlopuh:
Adding 3k lines of code in one PR is not a trivial thing. While it's in xdsl-bench it's all a bit more chill, but there are many more people working on and depending on the xdsl repo, so we should be able to actually review the changes. I also am not sure that this approach is the good one for all the different parts being merged, and I'm not sure that it makes sense to have all the discussions about the different aspects of this PR in one place. For one, I don't understand why we need to make any changes other than adding the tests themselves. I would still expect the users to use the xdsl-bench repo to run the benchmarks, and not do it form the xdsl repo directly. You can see that this is an orthogonal discussion to what benchmarks to add and whether they are representative of xDSL workloads, so it makes sense to at least split them out into separate PRs.

math-fehr:
I think separating the different tests might make this much more easier to review.
Or separating the infrastructure (like the asv file) with the tests, to understand more easily what's going on!

EdmundGoodman:
> Adding 3k lines of code in one PR is not a trivial thing. While it's in xdsl-bench it's all a bit more chill, but there are many more people working on and depending on the xdsl repo, so we should be able to actually review the changes. I also am not sure that this approach is the good one for all the different parts being merged, and I'm not sure that it makes sense to have all the discussions about the different aspects of this PR in one place. For one, I don't understand why we need to make any changes other than adding the tests themselves. I would still expect the users to use the xdsl-bench repo to run the benchmarks, and not do it form the xdsl repo directly. You can see that this is an orthogonal discussion to what benchmarks to add and whether they are representative of xDSL workloads, so it makes sense to at least split them out into separate PRs.

That's a very fair point. I'll close up this PR then maybe we could discuss next steps in the meeting tomorrow?

PR #3900, opened 2025-02-13
misc: fix arguments for riscv_lowering

oluwatimilehin:
@superlopuh  Fixing some potential bugs I encountered while using xDSL as a backend


xdsl/backend/riscv/lowering/convert_func_to_riscv_func.py
```diff
@@ -80,7 +80,7 @@ def match_and_rewrite(self, op: func.CallOp, rewriter: PatternRewriter) -> None:
         new_result_types = list(a_regs(op.results))
         new_op = riscv_func.CallOp(op.callee, moved_operands, new_result_types)
         move_result_ops, moved_results = move_to_unallocated_regs(
-            new_op.results, operand_types
+            new_op.results, new_result_types
```
compor, line 83: this looks like a bug? maybe our funcs were void so far? could you please add a test case for this?
let us know if you need a hand

oluwatimilehin: Hey Chris, how would you recommend I test this? It looks like I need to modify line 6 [here](https://github.com/oluwatimilehin/xdsl/blob/18a0e4da3cae5d72870b3264686cde7f39ae1612/tests/filecheck/backend/riscv/convert_func_to_riscv_func.mlir#L6), but I am not sure how to exercise this change and don't remember how I encountered the bug :(

What do you have in mind?

xdsl/dialects/linalg.py
superlopuh, line 1: this seems like an unrelated change, could you please open a separate PR for this?

xdsl/backend/riscv/lowering/convert_func_to_riscv_func.py
superlopuh: Just modifying it to be `(i32, f32) -> (i64, f64)` should be enough

oluwatimilehin: Do you mean doing this:

```
    func.func @main() {
        %0, %1 = "test.op"() : () -> (i32, f32)
        %2, %3 = func.call @foo(%0, %1) : (i32, f32) -> (i64, f64)
        func. return
    }
```
Which I suppose will require changing `foo()` too to accept an `f32` type:

```
    func.func @foo(%arg0 : i32, %arg1 : i32) -> (i32, i32) {
        %res0, %res1 = "test.op"(%arg0, %arg1) : (i32, i32) -> (i32, i32)
        func.return %res0, %res1 : i32, i32
    }
```
Is that right?

xdsl/dialects/linalg.py
oluwatimilehin: Created #3903

xdsl/backend/riscv/lowering/convert_func_to_riscv_func.py
superlopuh: yeah these would all have to be consistent within the file but that's the essence of the change

compor: thanks for splitting the PRs btw

xdsl/backend/riscv/lowering/convert_func_to_riscv_func.py
```diff
@@ -77,10 +77,13 @@ def match_and_rewrite(self, op: func.CallOp, rewriter: PatternRewriter) -> None:
         move_operand_ops, moved_operands = move_to_a_regs(
             register_operands, operand_types
         )
+
+        new_result_value_types = [result.type for result in op.results]
```
oluwatimilehin, line 81: This fixes a bug here that makes it raise an error for floating point registers: https://github.com/xdslproject/xdsl/blob/6f2c376bf5ee81ae675522d0ce6f7eba6b590edc/xdsl/backend/riscv/lowering/utils.py#L57-L67

`value_type` in the original code is neither of the two.

superlopuh: what is it?

oluwatimilehin: It's a FloatRegisterType.

xdsl/backend/riscv/lowering/convert_func_to_riscv_func.py
```diff
@@ -77,10 +77,13 @@ def match_and_rewrite(self, op: func.CallOp, rewriter: PatternRewriter) -> None:
...
         new_result_types = list(a_regs(op.results))
         new_op = riscv_func.CallOp(op.callee, moved_operands, new_result_types)
+
         move_result_ops, moved_results = move_to_unallocated_regs(
-            new_op.results, operand_types
+            new_op.results, new_result_value_types
```
superlopuh, line 86:
```suggestion
            new_op.results, op.result_types
```

superlopuh: no need for thing above

superlopuh (approved):

superlopuh (approved):

PR #3907, opened 2025-02-14
core: replace MLContext with Context

### LINKED ISSUE #249: xDSL: rename MLContext
The name MLContext was chosen in a time where we were closer to MLIR. We should rename it.

superlopuh:
Just a single find and replace.

Fixes #249

Keeps MLContext as a deprecated subclass of Context for clients.


xdsl/context.py
```diff
@@ -232,5 +232,7 @@ def get_optional_dialect(self, name: str) -> "Dialect | None":
         return None
 
 
-@deprecated("Please use Context instead")
-class MLContext(Context): ...
+class MLContext(Context):
```
alexarice, line 235: What's wrong with @deprecated?

superlopuh: 3.13

superlopuh: Cool username :)

compor (approved):

tests/test_mlcontext.py
compor, line 1: Not sure if we want to rename this file to `test_context.py` while we are at it. no biggie if not.

superlopuh: oh yeah good spot

math-fehr (approved):
o7

PR #3913, opened 2025-02-14
core: custom format operations cannot be UnregisteredOp

### LINKED ISSUE #3671: UnregisteredOp should only be made when parsing generic format
Currently if the parser encounters an unknown operation not in generic format (with `allow-unregistered-dialects` set), it creates an UnregisteredOp, and then complains that `UnregisteredOp` has no custom format. This is a slightly cryptic error, so it may be better to fail out earlier with a clearer error when an unknown operation in custom format is found.

alexarice:
Fixes #3671 

I'm not sure if this is the cleanest way to achieve this, an alternative would be to temporarily set `allow_unregistered` off in the ctx when making a custom format operation.

### DIFF
--- a/xdsl/context.py
+++ b/xdsl/context.py
@@ -125,13 +125,14 @@ class MLContext:
                 return self._get_known_op(name)
 
     def get_optional_op(
-        self, name: str, *, dialect_stack: Sequence[str] = ()
+        self, name: str, *, is_generic: bool = False, dialect_stack: Sequence[str] = ()
     ) -> "type[Operation] | None":
         """
         Get an operation class from its name if it exists or is contained in one of the
         dialects in the dialect stack.
         If the operation is not registered, return None unless unregistered operations
-        are allowed in the context, in which case return an UnregisteredOp.
+        are allowed in the context or the name was found in custom syntax,
+        in which case return an UnregisteredOp.
         """
         # Check if the name is known.
         if op_type := self._get_known_op(name):
@@ -144,7 +145,7 @@ class MLContext:
                 return op_type
 
         # If the context allows unregistered operations then create an UnregisteredOp
-        if self.allow_unregistered:
+        if self.allow_unregistered and is_generic:
             from xdsl.dialects.builtin import UnregisteredOp
 
             op_type = UnregisteredOp.with_name(name)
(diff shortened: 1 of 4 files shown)


xdsl/context.py
```diff
@@ -144,23 +145,25 @@ def get_optional_op(
                 return op_type
 
         # If the context allows unregistered operations then create an UnregisteredOp
-        if self.allow_unregistered:
+        if self.allow_unregistered and is_generic:
```
compor, line 148: the comment above is not inline with the condition

xdsl/parser/core.py
```diff
@@ -736,17 +736,19 @@ def parse_operation(self) -> Operation:
 
         return op
 
-    def _get_op_by_name(self, name: str) -> type[Operation]:
+    def _get_op_by_name(
+        self, name: str, *, is_generic: bool = False
+    ) -> type[Operation]:
```
compor, line 741: I am not sure I understand what is happening here; `is_generic` is not the greatest name and the way this is described, I'd expect the parser to tell us if the op is in generic or custom format and then decide based on that (along with the cmdline option for unregistered dialects)?

compor (commented):
I've tried this locally because I had bumped to it in the past with

```mlir
"test.dummy"() : () -> ()
test.dummy : () -> ()
```

and I still get:

```
test.dummy : () -> ()
           ^
           Operation builtin.unregistered does not have a custom format.
```

with and without `--allow-unregistered-dialect` from `xdsl-opt`.

with ` mlir-opt --allow-unregistered-dialect` I get:

```
unregister.mlir:3:1: error: custom op 'test.dummy' is unknown
test.dummy : () -> ()
^
```

alexarice:
@compor I'm also not particularly happy with the implementation here. How would you feel about mutating `ctx.allow_unregistered` in parser.core when parsing a custom format operation (and mutating it back after)

alexarice:
Maybe at least the error should be changed so that it prints the actual name of the unregistered op (maybe that would be preferable to this whole change)

compor:
> I see what's happening in your example... When it parses the generic `"test.dummy"` it assigns it to an `UnregisteredOp` and then for the custom format it recognises that `test.dummy` is already an `UnregisteredOp`.

Really? I find this surprising. `mlir-opt` behaves the same with and without having parsed the generic format of this op (same error I posted above).

alexarice:
xdsl "registers" test.dummy as an "UnregisteredOp" the first time you use it

compor:
I understand that, but still not what I'd expect

alexarice:
I'm not really sure how to proceed on this. It seems like the problem is a bit deeper than I appreciated.

alexarice:
I'm going to close this for now. I don't like this fix and it's not on my list of priorities to find a better one.

Issue #2390, opened 2024-03-27
dialects: (Arith) arith.const printing and parsing is broken


AntonLydike:
The printing and parsing of `arith.constant` in custom format is broken, see:

```
%0 = arith.constant true
```

will not print the type. Similarly:

```
from xdsl.dialects import arith
from xdsl.dialects.builtin import *
print(arith.Constant(IntegerAttr(0, 32), IntegerType(64)))
# prints: %0 = arith.constant 0 : i32
```

(see the return value being subtly changed to i32, the type of the attribute and not of the value. 

Both of these issues are due to the printing logic printing the value attribute, not the value's value and then the type.

PapyChacal:
I'm having a walk around or issues, and I find this one very interesting!

What put me on that path is seeing `arith.Constant(IntegerAttr(0, 32), IntegerType(64))` in the first place: you're defining a constant of type `i64` with a value of type `i32`, what's going on and how is it supposed to make sense?

Looking at MLIR's `arith.constant`, it only has one attribute, `value`, being a `TypedAttr` (i.e., carrying its type (as an attribute)). Our has both `value` and `type`.
Thing is, MLIR *restrict* `arith.constant`'s `value` attribute to such TypedAttrs. I just recently implemented the equivalent concept in xDSL:
- #2493 
Which would allow us to do the same: just having one typed `value` attribute!

Caveat: I only used TypedAttribute on IntegerAttr for declarative known-type integer syntax support.
This would require to match MLIR on which attributes types are actually typed, to make sure `arith.constant` keeps accepting everything we expect!

compor:
Resolved in #3916

PR #3916, opened 2025-02-14
dialects: (arith) fix `arith.constant` construction

compor:
I bumped into this recently, and it is documented in #2390, but it is not the parsing and printing that is broken.
We just use the `value_type` when provided, but don't adjust the type of the value attribute to it, ending up in having

```mlir
%0 = "arith.constant"() <{value = 0 : i32}> : () -> i64
```

which is invalid IR.

~I'd expect the `TypedAttributeConstraint` to kick in, but even though I tried to play with it, I'm not sure if I'm missing something obvious there. Happy to close this if I did.~


alexarice:
Constraints currently only deal with verification and inference for declarative assembly format

xdsl/dialects/arith.py
```diff
@@ -148,21 +149,55 @@ class ConstantOp(IRDLOperation):
...
         value_type: Attribute | None = None,
     ):
         if value_type is None:
-            value = cast(IntegerAttr | FloatAttr[AnyFloat], value)
-            value_type = value.type
+            if isinstance(value, IntegerAttr | FloatAttr):
+                value = cast(IntegerAttr | FloatAttr[AnyFloat], value)
+                value_type = value.type
+            elif isinstance(value, DenseIntOrFPElementsAttr):
+                value_type = value.type
```
alexarice, line 187: You should use `value.get_type()` to get the type associated to a `TypedAttribute`

xdsl/dialects/arith.py
```diff
@@ -148,21 +149,55 @@ class ConstantOp(IRDLOperation):
...
+
+        if isinstance(value, IntegerAttr):
+            value_type = cast(IntegerType, value_type)
+            value = IntegerAttr(value.value, value_type)
```
alexarice, line 191: I am against doing a silent coercion here. Personally I think we should either throw an error, or not allow a type in the constructor.

tests/dialects/test_arith.py
```diff
@@ -128,6 +129,29 @@ def test_Cmpi_from_mnemonic(self, input: str):
         _ = CmpiOp(self.a, self.b, input)
 
 
+def test_constant_construction():
+    c1 = ConstantOp(IntegerAttr(1, i32))
+    assert c1.value.type == i32
+
+    c2 = ConstantOp(IntegerAttr(1, i32), i64)
+    assert c2.value.type == i64
```
alexarice, line 137: Is there a use case where this behaviour is desirable

alexarice (commented):
I'm in general not in favour of this change.

tests/dialects/test_arith.py
compor: What behaviour exactly?

alexarice: Passing in an typed attribute and a different type, and having the constructor make a new attribute with the new type

xdsl/dialects/arith.py
compor: I am in favour of not allowing the type, as this is what is the most clear IMHO. I didn't do it as I wasn't sure of the initial rationale for this API and this provides the minimum change against it.

As for throwing an error, I don't think we are doing it anywhere in any constructor.

alexarice: I believe the original API allowed passing in python ints? I would be in favour of a single constructor that takes a single attribute and infers the type from that attribute.

tests/dialects/test_arith.py
compor: There are some obscure uses (one in CSL).

xdsl/dialects/arith.py
compor: The original API? I'm confused, the diff is clear, no? I can't see what you're saying. We have the `from_int_and_width` for that.

alexarice: Sorry, I meant original as in "when arith.constant was first added to xDSL" original.

compor: Well, I can't answer this, and at the end of the day that was changed to the current one that uses a typed attribute and a type (for which I mentioned I don't know the rationale behind it, AFAIU when I checked a while back, it was another deprecation of similar helpers to `from_int_and_width`, e.g. `from_attr`).

math-fehr:
I am personally against this constructor, because I really don't understand why this should be a standard constructor on `arith.constant`.
Why do we want this as a constructor rather than asking the users to write `ConstantOp(IntegerAttr(value.value, value_type))`?

xdsl/dialects/arith.py
superlopuh: I think at this point we can just remove value_type from the argument

tests/dialects/test_arith.py
superlopuh: I'm not sure it makes sense to support this

superlopuh:
I agree with Mathieu :)

compor:
> I agree with Mathieu :)

Great, consensus has been reached :) I'll adjust this PR as discussed above

xdsl/dialects/arith.py
compor: There seem to be uses within the stencil part of xDSL that use the general `Attribute` overload and rely on passing a `value_type`; for the rest of the types the API is simpler without requiring a `value_type` provided, as discussed previously here.

xdsl/dialects/arith.py
```diff
@@ -153,16 +153,26 @@ def __init__(
...
+    def __init__(
+        self,
+        value: Attribute,
+        value_type: Attribute,
+    ) -> None: ...
 
     def __init__(
         self,
-        value: IntegerAttr | FloatAttr[AnyFloat] | Attribute,
+        value: IntegerAttr | FloatAttr[AnyFloat] | DenseIntOrFPElementsAttr | Attribute,
```
superlopuh, line 164: In my understanding, the latest changes don't correspond to Mathieu's suggestion. Why allow passing `Attribute` here, if it's not allowed by the constraint on `value`? Wouldn't this cover all possible values to be used as constants?
```suggestion
        value: IntegerAttr | FloatAttr[AnyFloat] | DenseIntOrFPElementsAttr,
```
If so, why let users provide the value_type, if it can always be found on the input?

superlopuh: It's possible that I also misunderstood, but the type definitely seems overly permissive for this style of init that only lets you assign valid things

xdsl/dialects/arith.py
compor: done

tests/dialects/test_arith.py
compor: patched the ones that were left (in experimental)

xdsl/dialects/arith.py
compor: I think it's clear now

superlopuh (approved):
Nice

PR #3917, opened 2025-02-15
backend: (x86) add x86.assembly_section

qaco:


superlopuh (commented):
Oh actually let's chat about this on Monday? I'm not really sure whether the design we came up with for RISC-V is the best, this was a part of the idea behind the printer refactor. Would it make sense instead to print the `.text` as part of the func op printing logic? That way it can stay as top level in the module, simplifying the lowering and symbol lookup.

qaco:
> Oh actually let's chat about this on Monday? I'm not really sure whether the design we came up with for RISC-V is the best, this was a part of the idea behind the printer refactor. Would it make sense instead to print the `.text` as part of the func op printing logic? That way it can stay as top level in the module, simplifying the lowering and symbol lookup.

Ok let's discuss it on monday !

compor:
any news on the discussion? how one should review this? is it ready?

tests/filecheck/dialects/x86/x86_assembly_section.mlir
```diff
@@ -0,0 +1,13 @@
...
+
+x86.assembly_section ".text" {
+  x86.directive ".global" "external"
+  x86_func.func @external() -> ()
+}
+
+// CHECK:      builtin.module {
+// CHECK-NEXT:   x86.assembly_section ".text" {
+// CHECK-NEXT:     x86.directive ".global" "external"
+// CHECK-NEXT:     x86_func.func @external() -> ()
```
compor, line 11: Now that I'm thinking about it, can this be a simple `test.op`? I find it a bit confusing because this is an op that is a `DirectiveOp` and the assembly section also translates to a directive in plain assembly.

compor (approved):

xdsl/dialects/x86/ops.py
```diff
@@ -72,6 +69,59 @@ def assembly_line(self) -> str | None:
...
+        printer.print_string_literal(self.directive.data)
+        printer.print_op_attributes(
+            self.attributes, reserved_attr_names=("directive",), print_keyword=True
+        )
+        printer.print_string(" ")
+        if self.data.block.ops:
+            printer.print_region(self.data)
+
+    def assembly_line(self) -> str | None:
+        return AssemblyPrinter.assembly_line(self.directive.data, "", is_indented=False)
```
superlopuh, line 122: the test for this is missing in this PR

superlopuh:
?

PR #3919, opened 2025-02-16
backend: (x86) add the extended registers (eax, ecx...)

qaco:


xdsl/dialects/x86/register.py
```diff
@@ -43,14 +43,22 @@ def verify(self) -> None:
...
-    "r15": 15,
+    "eax": 8,
+    "ecx": 9,
+    "edx": 10,
+    "ebx": 11,
+    "esp": 12,
+    "ebp": 13,
+    "esi": 14,
+    "edi": 15,
+    "r8": 16,
```
superlopuh, line 54: Is that right? Is there a url to the spec we could add as a comment here?

qaco: Oh, I thought it was arbitrary identifiers (for uniquification of the registers). In fact it describes the numerical encoding of the name of the register, right ? I start looking for the standard. Thank you a lot for your review, I was about to make a huge mistake.

qaco: Done !

superlopuh (approved):
Thank you!

PR #3924, opened 2025-02-17
bug: (x86) always remove the register/memory prefix from the emitted assembly

qaco:
The printer should not emit code like ```rr.add```, but our Python filecheck implementation only checks substrings (even when the dedicated knob is on). I fix the emitter to make it usable but a long-term solution is either to fix the Python filecheck reimplementation or to py-package LLVM filecheck.

### DIFF
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -179,7 +179,7 @@ class X86Instruction(X86AsmOperation):
         By default, the name of the instruction is the same as the name of the operation.
         """
 
-        return Dialect.split_name(self.name)[1]
+        return Dialect.split_name(self.name)[1].split(".")[-1]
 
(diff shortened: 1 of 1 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -179,7 +179,7 @@ def assembly_instruction_name(self) -> str:
         By default, the name of the instruction is the same as the name of the operation.
         """
 
-        return Dialect.split_name(self.name)[1]
+        return Dialect.split_name(self.name)[1].split(".")[-1]
```
superlopuh, line 182:
```suggestion
        return self.name.split(".")[-1]
```

superlopuh: Might want to do this for ARM also

compor (approved):

PR #3927, opened 2025-02-17
backend: (riscv) split out reserved int and float registers in regalloc

superlopuh:
This is the only part of the `RiscvRegisterQueue` that mixes the two register types, this split is necessary to make the more fundamental split into two register queues.


compor (approved):
Where is this heading? I assume 2 independent queues or a queue per register type.

superlopuh:
That will get us closer to structures that we can reuse for the other backends

PR #3929, opened 2025-02-17
backend: add `infinite_register` class method to RegisterType

superlopuh:
It's nice to have a check to make sure that our fake infinite registers don't clash with the real ones.

### DIFF
--- a/xdsl/backend/register_type.py
+++ b/xdsl/backend/register_type.py
@@ -96,9 +96,9 @@ class RegisterType(ParametrizedAttribute, TypeAttribute, ABC):
         """
         Provide the register at the given index in the "infinite" register set.
         """
-        spelling = cls.infinite_register_prefix()
-        res = cls(spelling + str(index))
+        spelling = cls.infinite_register_prefix() + str(index)
+        res = cls(spelling)
         assert isinstance(res.index, NoneAttr), (
-            f"Invalid 'infinite' register name {spelling}"
+            f"Invalid 'infinite' register name: {spelling} clashes with finite register set"
         )
         return res
--- a/xdsl/backend/register_type.py
+++ b/xdsl/backend/register_type.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
+from typing import Self
 
 from xdsl.dialects.builtin import (
     IntAttr,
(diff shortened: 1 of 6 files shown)


xdsl/dialects/x86/register.py
```diff
@@ -146,6 +154,10 @@ def instruction_set_name(cls) -> str:
     def abi_index_by_name(cls) -> dict[str, int]:
         return SSE_INDEX_BY_NAME
 
+    @classmethod
+    def infinite_register_name(cls, index: int):
+        return f"sse_{index}"
```
JosseVanDelm, line 159:
```suggestion
        return f"josse_{index}"
```

xdsl/backend/register_type.py
```diff
@@ -78,3 +80,24 @@ def instruction_set_name(cls) -> str:
...
+    @classmethod
+    def infinite_register(cls, index: int) -> Self:
+        """
+        Provide the register at the given index in the "infinite" register set.
+        """
+        spelling = cls.infinite_register_name(index)
+        res = cls(spelling)
+        assert isinstance(res.index, NoneAttr), (
+            f"Invalid 'infinite' register name {spelling}"
+        )
```
JosseVanDelm, line 102:
```suggestion
        assert isinstance(res.index, NoneAttr), (
            f"Invalid 'infinite' register name: {spelling} clashes with finite register set"
        )
```

JosseVanDelm (approved):
Looks good to me, I have a few suggestions though

compor (approved):

superlopuh:
I changed the logic to just provide the prefix, not the whole thing. One of the questions is do we want `_` suffixes everywhere? How important do you think it is to be consistent here, should we allow `j0` in riscv and `avx2_0` on x86, or should we change to `j_0` in riscv land?

compor:
> I changed the logic to just provide the prefix, not the whole thing. One of the questions is do we want `_` suffixes everywhere? How important do you think it is to be consistent here, should we allow `j0` in riscv and `avx2_0` on x86, or should we change to `j_0` in riscv land?

Not a biggie for me; leaning slightly towards using `_` everywhere, since for the RISC-V side is purely a riscemu thing.
Maybe even slightly easier to search/grep?

superlopuh:
OK then I'll do a quick PR first doing `j_` and `fj_` for riscv then rebase this on top of that

superlopuh:
#3934

PR #3930, opened 2025-02-17
backend: (x86) consistently print the function type of x86_func

qaco:
I assume that when in x86_func, arguments are passed using calling conventions (writing/reading registers and stack). Therefore, the arguments of the first basic block of the function should not be displayed, but rather the type of the function (which is just high-level information annotating the function).


superlopuh:
I'm not sure I get it, what's the advantage of this? I feel like we lose a bit in consistency, so it would be good to make sure it's worth it.

compor (commented):
I'm not sure I understand how the calling convention gives us the liberty to not display the first block's arguments.

If nothing else, I'd make it explicit by printing comments for the retrieval of the function arguments that went on the stack as in:

```asm
mv eax, rsp, 8    # get function arg 8
mv ebx, rsp, 16   # get function arg 9
```

something I forgot to comment in the previous relevant PR.

I don't see the gain here.

PR #3933, opened 2025-02-18
documentation: Fix nits on the first marimo notebook

compor:
Minor nits and rewording of sentences.

### DIFF
--- a/docs/marimo/mlir_ir.py
+++ b/docs/marimo/mlir_ir.py
@@ -13,6 +13,7 @@ def _():
     from xdsl.utils import marimo as xmo
     from xdsl.printer import Printer
     from typing import Any
+
     return Any, Context, Printer, arith, builtin, func, mo, scf, xmo
 
 
@@ -30,7 +31,8 @@ def _(mo):
 
 @app.cell(hide_code=True)
 def _(mo, triangle_text):
-    mo.md(fr"""
+    mo.md(
+        rf"""
     MLIR and xDSL use a textual encoding of the IR for debugging, testing, and storing intermediate representations of programs.
     It can be very useful to take a program at some stage of compilation, and inspect it.
     The textual format makes this easy to do.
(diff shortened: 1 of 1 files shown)


docs/marimo/mlir_ir.py
```diff
@@ -103,7 +107,12 @@ def _(first_text, mo):
 
 @app.cell(hide_code=True)
 def _(exercise_text, first_text_area, mo):
-    _first_info_text = exercise_text(first_text_area.value, "first", ((1, 2), (3, 4)), ("first(1, 2) = ", "first(3, 4) = "))
+    _first_info_text = exercise_text(
+        first_text_area.value,
+        "first",
+        ((1, 2), (3, 4)),
+        ("first(1, 2) = ", "first(3, 4) = "),
+    )
```
compor, line 115: we have ruff formatting exclude notebooks, so it's my local settings I assum.

docs/marimo/mlir_ir.py
```diff
@@ -514,19 +531,17 @@ def _(mo):
 def _(mo):
     mo.md(
         r"""
-        The [`builtin` dialect](https://mlir.llvm.org/docs/Dialects/Builtin/) contains the most commonly-used operation in MLIR/xDSL: `builtin.module`
-
-        A module is a unit of code in xDSL and MLIR.
-        It holds a single region.
+        The [`builtin` dialect](https://mlir.llvm.org/docs/Dialects/Builtin/) contains the most commonly-used operation in MLIR and xDSL: the `builtin.module` operation.
 
+        A module is a unit of code which holds a single region.
```
compor, line 536: Regions are not explained, so my main issue here is adding a definition earlier or omitting all together.
I'm inclined to do that later.

docs/marimo/mlir_ir.py
superlopuh: Are you editing the raw python locally or using marimo edit?

superlopuh (approved):
This is great, thank you

docs/marimo/mlir_ir.py
compor: Raw python locally, do you want me to run with the editor? I'll see if it can reformat these.

superlopuh: Yeah it would be good to run it in the editor, before merging because they have their own formatter that will likely conflict with your changes.

compor: Yeah, there is no "format all cells", I have to go over each cell and do a `ctrl-b`

compor: "Run all cells" seems to do it

PR #3934, opened 2025-02-18
backend: (riscv) rename jx registers to j_x

superlopuh:


compor (approved):
I think you should wait for @JosseVanDelm ;-]

PR #3938, opened 2025-02-19
bug: (x86) fix the lowering of the function signature

qaco:
This PR concerns the registers used to pass function parameters. The previous PR retrieved them using get_register, so the function type was inherently empty (I thought a x86 function was just a label). This new PR provides the registers in question as well as the stack pointer as input to the first basic block.


xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -30,70 +30,72 @@
...
-                isinstance(arg.type, builtin.FixedBitwidthType)
-                and arg.type.bitwidth > 64
-            ):
+            elif isinstance(ty, builtin.FixedBitwidthType) and ty.bitwidth > 64:
                 raise DiagnosticException(
                     "Cannot lower function parameters bigger than 64 bits (not implemented)"
                 )
 
         num_inputs = len(op.function_type.inputs.data)
         num_passing_args = num_inputs if num_inputs <= 6 else 6
```
superlopuh, line 49:
```suggestion
        num_passing_args = min(num_inputs, 6)
```

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -30,70 +30,72 @@
...
+        reg_args_types = arg_passing_registers[0:num_passing_args]
```
superlopuh, line 50:
```suggestion
        reg_args_types = arg_passing_registers[:min(num_inputs,6)]
```

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -30,70 +30,72 @@
...
-                )
-                actual_registers.append(mov_op.result)
-                rewriter.insert_op([get_reg_op, mov_op], insertion_point)
+        params_mapping: list[tuple[BlockArgument, SSAValue]] = []
+
+        # Build the basic block header
+        for i, register_type in enumerate(reg_args_types):
+            arg = first_block.args[i]
+            register = first_block.insert_arg(register_type, i)
+            params_mapping.append((arg, register))
```
superlopuh, line 64: Do you need to save this mapping? It seems like a simple index calculation later, since you can just save the number of args.

compor:
hey @qaco is this ready for review (it's marked as `draft`, that's why I'm asking)? Also, a brief description of the PR would help as the titled is a bit generic, what is the "fix" here? thanks!

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
qaco: Right. Instead of doing it later, I apply the replace/erase in a row with the creation of the parameters.

qaco:
> hey @qaco is this ready for review (it's marked as `draft`, that's why I'm asking)? Also, a brief description of the PR would help as the titled is a bit generic, what is the "fix" here? thanks!

It was actually not ready for review ^^ I've added the description too !

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -30,70 +30,67 @@
...
-                and arg.type.bitwidth > 64
-            ):
+            elif isinstance(ty, builtin.FixedBitwidthType) and ty.bitwidth > 64:
                 raise DiagnosticException(
                     "Cannot lower function parameters bigger than 64 bits (not implemented)"
                 )
 
         num_inputs = len(op.function_type.inputs.data)
-        num_passing_args = num_inputs if num_inputs <= 6 else 6
+        reg_args_types = arg_passing_registers[: min(num_inputs, 6)]
```
compor, line 49: Can this "floating" 6 be made a global var that denotes exactly its purpose throughout the file?

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -30,70 +30,67 @@
...
+        for i in range(num_inputs - 6):
+            arg = first_block.args[6 + 1]
+            assert sp != arg
+            get_reg_op = x86.ops.GetRegisterOp(x86.register.GeneralRegisterType(""))
+            mov_op = x86.RM_MovOp(
+                r1=get_reg_op.result,
+                r2=sp,
+                offset=STACK_SLOT_SIZE_BYTES * (i + 1),
+                result=x86.register.GeneralRegisterType(""),
+            )
```
compor, line 78: Could you add a comment denoting the function argument number being retrieved here? e.g., `# func argument 7`

compor (changes requested):

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
qaco: Done !

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
qaco: Done !

compor: The Python comment is great, but I meant an assembly comment per `mv` from `rsp+offset`

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -24,76 +24,83 @@
 
 return_passing_register = x86.register.RAX
 
+MAX_REG_PASSING_INPUTS = 6
```
superlopuh, line 27: Can you please add a doc comment here?

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
qaco: Haha ok done !

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -24,76 +24,83 @@
...
                 )
 
         num_inputs = len(op.function_type.inputs.data)
-        num_passing_args = num_inputs if num_inputs <= 6 else 6
+        reg_args_types = arg_passing_registers[
+            : min(num_inputs, MAX_REG_PASSING_INPUTS)
+        ]
 
-        insertion_point = InsertPoint.at_start(first_block)
+        new_region = rewriter.move_region_contents_to_new_regions(op.body)
```
superlopuh, line 55: why move to a new region instead of rewriting the existing region?

compor (approved):

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
compor: Maybe also add one for the `STACK_SLOT_SIZE_BYTES` below while you're at it?

superlopuh (approved):
LGTM modulo comments!

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
qaco: Because when you build the new FuncOp, you ```Cannot add region that is already attached on an operation.```

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
qaco: Done !

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
superlopuh: Ah right thank you

xdsl/backend/x86/lowering/convert_func_to_x86_func.py
superlopuh: Doc comments are done by adding a string (either "" or """""" for multiline below the variable), that way the IDE picks them up and displays them when you hover over the symbol name with your cursor

superlopuh: For classes, functions, and .py files it has to be the first expression in the definition

PR #3942, opened 2025-02-21
documentation: fix arith cmpi example in MLIR IR notebook

superlopuh:


math-fehr (approved):

docs/marimo/mlir_ir.py
```diff
@@ -165,19 +165,19 @@ def _(mo):
 def _():
     less_than_text = """\
     func.func @less_than(%a: i32, %b: i32) -> i1 {
-      %slt = arith.cmpi slt, %lhs, %rhs : i1
+      %slt = arith.cmpi slt, %lhs, %rhs : i32
```
compor, line 168: I have to admit, I always need to think about this for a minute to make sure it is the operands and not the bool-like return

superlopuh: Yep same and it confused today's audience also

PR #3943, opened 2025-02-21
backend: (x86) fix missing __init__.py files

superlopuh:
We need a way to track this in the CI.


math-fehr (approved):

compor:
not sure about the CI, but maybe for our ruff config https://docs.astral.sh/ruff/rules/implicit-namespace-package/

superlopuh:
This is fantastic and exactly what we need.

superlopuh:
#3945

PR #3946, opened 2025-02-22
misc: shorten PR template

superlopuh:
I have a feeling that our template is too long, and that people just see a wall of text and zone out, when there are actually only two things to pay attention to: 1. read the guidelines if you haven't, 2. add a motivation. We can easily do the labels, reviewers and assignees as maintainers.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,16 +8,20 @@ assignees: ''
 ---
 
 ### Description
-A clear and concise description of the issue or bug. 
+
+A clear and concise description of the issue or bug.
 
 ### Expected vs Actual Behavior
+
 Include what you expected to happen and what actually happened (e.g., when compared with upstream MLIR).
 
 ### Steps to Reproduce and a Minimal Working Example (MWE)
+
 1. Describe the steps to reproduce the issue (commands, etc.).
 2. If applicable, describe the environment (e.g., operating system, Python version, etc.).
-3. Please make your reproducible example as small as possible.  
+3. Please make your reproducible example as small as possible.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
superlopuh, line 1: I just formatted this file with markdownlint

.github/pull_request_template.md
```diff
@@ -1,17 +1,8 @@
-*Short paragraph summarizing the changes*
-
 ---
+title: ""
```
superlopuh, line 2: markdownlint complains without this bit

compor: do you have the `line-length` option disabled? because it complains for 2 lines for me.

compor (approved):
I'm not sure where this feeling is based, but OK

.github/pull_request_template.md
superlopuh: Locally in your VSCode? No complaints for me with this change, I wonder if we have different global configs

JosseVanDelm (approved):
As a person who does not regularly contribute to xdsl I think it still would be nice to have the process documented somewhere (Contributing.md?). Although I do agree that in its current form it might be a bit scary for someone to contribute who has never met the maintainers(?)
Anyways big fan of conciseness!

.github/pull_request_template.md
compor: yeah, I just have the stock one and just using https://github.com/fannheyward/coc-markdownlint and by default the config is `{}` which has the https://github.com/DavidAnson/markdownlint/blob/main/doc/md013.md line length rule on (80 columns).

I'll disable mine locally.

math-fehr (approved):

PR #3948, opened 2025-02-22
dialects: (transform) move tests to filecheck and fix two ops

superlopuh:
I came across the transform dialect pytest test, and decided to move it to filecheck, in the process finding two erroneous definitions of transform ops. The issue with using pytest for checking of printing and parsing is that it doesn't call verify on the operations. I'm now a bit more confident in our transform dialect reimplementation.


xdsl/dialects/transform.py
```diff
@@ -297,7 +296,7 @@ class GetTypeOp(IRDLOperation):
 
     name = "transform.get_type"
 
-    elemental = prop_def(UnitAttr)
+    elemental = opt_prop_def(UnitAttr)
```
alexarice, line 299: Let's include a test with this attribute present?

tests/dialects/test_transform.py
```diff
@@ -1,250 +0,0 @@
-from conftest import assert_print_op
```
alexarice, line 1: Is this used anywhere else?

alexarice (approved):
Two minor comments then LGTM

JosseVanDelm (approved):
Nice catch!

compor (approved):

tests/dialects/test_transform.py
superlopuh: Too many places, I've got some branches locally that remove this dependency, which is kinda what motivated this PR in the first place.

PR #3949, opened 2025-02-22
interpreter: (pdl) add operations directly in `pdl.operation`

### LINKED ISSUE #3569: interpreter: (pdl) fix multiple inserted operations in rewrite pattern interpreter

jumerckx:
reference: https://github.com/xdslproject/xdsl/pull/3862#issuecomment-2661650520
Operations are now inserted directly when interpreting `pdl.operation`. Insertion happens right before the root of the rewrite op.

Fixes #3569


xdsl/interpreters/pdl.py
```diff
@@ -341,6 +343,18 @@ def run_operation(
...
+        if self.pdl_rewrite_op is None:
+            raise InterpretationError(
+                "pdl_rewrite_op was not provided to PDLRewriteFunctions"
+            )
+        if self.pdl_rewrite_op.root is None:
+            raise InterpretationError("rewrite op does not contain a valid root")
+        root_op = interpreter.get_values((self.pdl_rewrite_op.root,))[0]
+
+        # Insert the new operation before the root operation
+        self.rewriter.insert_op(result_op, InsertPoint.before(root_op))
```
superlopuh, line 356: This should do the trick, hopefully voiding the need for `pdl_rewrite_op`
```suggestion
        self.rewriter.insert_op_before_matched_op(result_op)
```

superlopuh (commented):
This is great! It would be great to add a test here, hopefully something that failed before, and no longer fails with these changes. Ideally a `pdl-interp` lit test, as well as a pytest test. I'm happy to help out with either!

superlopuh (approved):
Thank you!

PR #3944, opened 2025-02-21
CI: fix RUN command in fsm and ltl lit tests

superlopuh:
The good news is that in the near future this will be detected automatically.

PR #3953, opened 2025-02-24
testing: don't cd to git root dir in lit

superlopuh:
This change in configuration was motivated by the lack of detection of missing `__init__.py` files in the project. We now have a ruff check for this (#3945), but in my changing the config, I found a number of other issues stemming from the `cd ../..` in the preamble to the tests. One is that for some reason files with no RUN lines aren't detected with the preamble. I fixed the broken tests that weren't run in a separate PR (#3944), so this is just about making the config change and making minor fixes to remaining files.


tests/filecheck/backend/csl/gaussian-end-to-end.mlir
superlopuh, line 1: CC @emmau678 @n-io @dk949

superlopuh:
I'm a bit stumped by this, it looks like `coverage` becomes flaky, and occasionally doesn't print anything to either of stdout or stderr, about once per test run. I can reproduce this locally, but it's a different file every time.

compor:
`coverage` has a `debug` option, maybe it's worth a try? this sounds like a lock or ordering issue (might be off though)

tests/filecheck/lit.cfg
```diff
@@ -1,25 +1,34 @@
...
+else:
+    rc_file = xdsl_src / "pyproject.toml"
+    data_file = xdsl_src / ".coverage.filecheck"
+    coverage_cmd = f"coverage run --source={xdsl_src} --rcfile={rc_file} --data-file={data_file}"
+    config.test_format = lit.formats.ShTest()
 
+config.substitutions.append(('xdsl-opt', f"{coverage_cmd} {xdsl_opt}"))
+config.substitutions.append(('xdsl-run', f"{coverage_cmd} {xdsl_run}"))
+config.substitutions.append(('irdl-to-pyrdl', f"{coverage_cmd} {irdl_to_pyrdl}"))
+config.substitutions.append(('python', f"{coverage_cmd}"))
```
superlopuh, line 31: Basically I couldn't hack making the change work with coverage enabled, so we now have a bit more logic that depends on whether or not we collect code coverage metrics, such as where the tests are run from, and what the paths to the commands are.

compor: Maybe this is worth a comment?

compor (commented):
I don't mean to create work, but the four unrelated changes could be a separate "cleanup" PR?

Moreover, what is the gain here in relation to the CI test runs?

superlopuh:
They're kind of related in that they were only raised as errors once I removed the cd thing, but I'm happy to split.

compor:
> They're kind of related in that they were only raised as errors once I removed the cd thing, but I'm happy to split.

also, what is the gain here in relation to the CI test runs?

superlopuh:
I'm not sure I understand your question, what do you mean by gain?

compor:
what does this PR do?

I still see that we are changing dirs by `cd` on coverage runs:

```py
config.test_format = lit.formats.ShTest(preamble_commands=[f"cd {xdsl_src}"])
```

but what issue does it resolve in the CI?

superlopuh:
When we don't cd to the root of the directory we get two things: `xdsl` the folder is not in the path, so the lookup has to happen via the locally-installed `xdsl` the module. We then automatically detect that files aren't importable due to missing `__init__.py` (for which we now have a ruff rule but I'm not convinced it works super well), and potentially other behaviours that are different when the package is installed as it would be on a client; the errors such as those fixed in the peeled-off PR are found, like files without RUN commands, empty files, etc.

compor (approved):

PR #3954, opened 2025-02-24
misc: remove xdsl-opt file

superlopuh:
I tested this locally by installing via a different path, seems like this is just unused at this point.

### DIFF
--- a/docs/mlir_interoperation.ipynb
+++ b/docs/mlir_interoperation.ipynb
@@ -215,7 +215,7 @@
     }
    ],
    "source": [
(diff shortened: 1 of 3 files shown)


compor:
I don't understand this?

alexarice (approved):

superlopuh:
We have xdsl_opt_main.py, xdsl_opt.py, and xdsl-opt, feels like we need at most two

compor:
> We have xdsl_opt_main.py, xdsl_opt.py, and xdsl-opt, feels like we need at most two

Oh, I see now!

compor (approved):

PR #3956, opened 2025-02-24
bug: (libxsmm) fix the little matmul use case

qaco:
The use case contained 4 bugs according to computing a matmul:
1- Bad order of vfmadd parameters (AT&T syntax instead of Intel syntax)
2- vfmadd231pd instead of vfmadd231ps (double precision instead of single precision)
3- Incosistency between B dimensions (4x2) and C dimensions (4x4)
4- (performance bug, avoidable dependency) Load all B columns in the same register

### DIFF
--- a/tests/filecheck/projects/libxsmm/x86_matmul_kernel.mlir
+++ b/tests/filecheck/projects/libxsmm/x86_matmul_kernel.mlir
@@ -1,6 +1,6 @@
 // RUN: xdsl-opt -t x86-asm %s | filecheck %s
 
-// C: 4x2xf32 = A: 4x8xf32 * B: 8x2xf32
+// C: 4x4xf32 = A: 4x8xf32 * B: 8x4xf32
 x86_func.func @matmul() {
     // General registers loading; according to the x86 calling
     // conventions, rdi, rsi and rcx always contain the first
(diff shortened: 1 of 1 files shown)


compor (approved):
I am not sure how to review this as (apart from the multiple issues), I have no way of verifying (easily) the above.

There are two aspects:
1) Maybe it is time to invest in an "experiments repo"; I can see this being easily called in from a `main()` stub that provides the inputs. Moreover, I have a machine here that can run those if needed.
2) I can't pretend that I understand the assembly code's performance, just by reading it on a PR, although it is still easily parsable by a human

Based on these take my "Approval" at face value.

emmau678 (approved):
It looks good to me

qaco:
> I am not sure how to review this as (apart from the multiple issues), I have no way of verifying (easily) the above.
> 
> There are two aspects:
> 
>     1. Maybe it is time to invest in an "experiments repo"; I can see this being easily called in from a `main()` stub that provides the inputs. Moreover, I have a machine here that can run those if needed.
> 
>     2. I can't pretend that I understand the assembly code's performance, just by reading it on a PR, although it is still easily parsable by a human
> 
> 
> Based on these take my "Approval" at face value.

Fully agree on your idea to enhance the testing framework. Do you think this "validation" step should be performed in a separate repo or in a subdirectory of xdsl/tests ?

compor:
> Fully agree on your idea to enhance the testing framework. Do you think this "validation" step should be performed in a separate repo or in a subdirectory of xdsl/tests ?

I don't think there is a perfect solution, but we've had success with a separate repo when using a setup that has quite hefty dependencies.

I think the case here is much simpler, although not without overhead (syncing the repos, etc.), but such is life :-)

superlopuh (approved):

PR #3957, opened 2025-02-24
testing: (stencil) remove stencil printing test

superlopuh:
Is there a need for this test any more?

I'm not sure what it's testing, it seems to be more of an example of the API needed to represent a stencil in Python, but I'm not sure whether this is useful anymore. There's a TODO comment to move to documentation, but something tells me it can happily not ever happen if no-one's working on the stencil project any more.

### DIFF
--- a/tests/dialects/test_stencil.py
+++ b/tests/dialects/test_stencil.py
@@ -1,9 +1,7 @@
 import pytest
-from conftest import assert_print_op
 
-from xdsl.builder import Builder, ImplicitBuilder
-from xdsl.dialects import builtin, func
-from xdsl.dialects.arith import AddfOp
+from xdsl.builder import Builder
+from xdsl.dialects import builtin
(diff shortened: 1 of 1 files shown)


georgebisbas (commented):
This was used as testing infra for the stencil dialect, and as an example. Overall it is tested enough I think. Are you removing `def assert_print_op` as well?

superlopuh:
Yeah that's the overall goal. It's the only helper that is shared between multiple tests, and is preventing us from testing xDSL as a library, meaning closer in behaviour to how people would use it when pip installing.

superlopuh:
@PapyChacal @AntonLydike @georgebisbas I'm happy to move this wherever you tell me to, but I'd like to delete this test, please let me know how to proceed.

superlopuh:
I'll take that as an approve :)

PR #3958, opened 2025-02-24
testing: remove func call printing tests

superlopuh:
Do we need this any more? This seems to be a test for the Python API, which we tend not to test too much these days. I'd like to remove the dependency on `assert_print_op`, so it would be good to figure out how we want to change these tests if we want to keep them to avoid the import.

### DIFF
--- a/tests/dialects/test_func.py
+++ b/tests/dialects/test_func.py
@@ -1,5 +1,4 @@
 import pytest
-from conftest import assert_print_op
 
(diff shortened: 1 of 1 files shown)


compor (approved):
I'm in favour of cleaning up, as I feel this kind of testing is part of the printing/parsing infrastructure (either custom parsers/printers or using the assembly format) and not really dialect-specific.

PR #3959, opened 2025-02-24
testing: move coverage config to pyproject.toml

superlopuh:
Just a bit of simplification to the project, moving the config to one place instead of spreading it across multiple files.


pyproject.toml
```diff
@@ -172,3 +172,28 @@ style = "pep440"
...
+[tool.coverage.run]
+branch = true
+omit = [
+    # Autogenerated missed code handles other VCSes.
+    "xdsl/_version.py",
+    "tests/filecheck/frontend/*",
+]
+concurrency = ["multiprocessing"]
+parallel = true
+source = ["xdsl/", "tests/", "docs/"]
```
compor, line 185: didn't we just remove `tests/` in the other PR?

compor (approved):
I had tried this a while back, but coverage didn't support pyproject.toml; cool that they added support

pyproject.toml
superlopuh: Indeed, I've just merged that one, will remove tests from here before I merge

PR #3961, opened 2025-02-25
core: remove indirection in operation cloning

EdmundGoodman:
Toy example of using benchmarking to inform performance optimisation.

In this case, we remove indirection from the `clone_without_regions` method on `Operation` to avoid incurring the cost of creating extraneous `OpOperand` and `Successor` dataclasses, which are then only used to construct iterators on a private property of the original class.

Running `python3 benchmarks/microbenchmarks.py OpCreation.operation_clone timeit`
Before: Test OpCreation.operation_clone ran in: 2.44e-06 ± 3.27e-07s
After : Test OpCreation.operation_clone ran in: 2.15e-06 ± 2.75e-07s
(~10% performance uplift for this limited case of a moderately common feature)

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1104,14 +1104,14 @@ class Operation(IRNode):
             block_mapper = {}
         operands = [
             (value_mapper[operand] if operand in value_mapper else operand)
-            for operand in self.operands
+            for operand in self._operands
         ]
         result_types = self.result_types
         attributes = self.attributes.copy()
         properties = self.properties.copy()
         successors = [
             (block_mapper[successor] if successor in block_mapper else successor)
-            for successor in self.successors
+            for successor in self._successors
         ]
         regions = [Region() for _ in self.regions]
         cloned_op = self.create(


superlopuh (approved):

superlopuh:
Can you please follow the guidelines for PR naming, assign yourself, and add a label? I think `core` is most appropriate here

alexarice (commented):
This should save 1 object creation. Is this really the bottleneck?

EdmundGoodman:
> Sorry, do I have to have the benchmarks checked out? Where do I run `python3 benchmarks/microbenchmarks.py OpCreation.operation_clone timeit`?

For now all the benchmarking is in the xdsl-bench repo (but I'll try to get small PRs bringing features across soon). If you `git clone https://github.com/xdslproject/xdsl-bench` then you should be able to run the commands

> So, the 10% gain brings us to 0.00215ms for this operation, right?

Yes

EdmundGoodman:
> This should save 1 object creation. Is this really the bottleneck?

Below are the traces before (with highlighted bits being the deleted object creations):

Screenshot 2025-02-25 at 11 40 18

and after:

Screenshot 2025-02-25 at 11 40 25

In such as short operation, it's not necessarily the bottleneck, but definitely something we can avoid which will be beneficial for workloads with lots of clone operations.

compor:
> In such as short operation, it's not necessarily the bottleneck, but definitely something we can avoid which will be beneficial for workloads with lots of clone operations. There's also some `__post__init__` stuff which could probably also be improved, but I will need to stare at it a bit more to understand what it is trying to do before I change it.
> 
> More generally, looking at the clone operation in particular was motivated by the micro-benchmarks where it was an outlier in being a lot closer to MLIR, and also because some of Sasha's upcoming workloads are pretty clone-heavy.

How many thousands of clone operations are we looking at?

I'm referring to thousands because that will bring us to a little over 2.15 seconds as opposed to 2.45 seconds (ignoring deviations) for roughly 1000 clone operations IIUC.

superlopuh:
It could easily be millions for the scheduling things we're working on (which is what it would take for it add up to .3s). I'd just like to say that bottleneck is probably not the only way to look at things. We have previously discussed end-to-end workloads, which are important, but it's also worth doing micro-optimisations on individual commmon operations. The issue with optimising for end-to-end things is that you might overfit on current use-cases, and make future use-cases less performant. We're slowly converging on making individual operations in xDSL faster for Edmund's thesis, even if we'll probably also want to optimise end-to-end things for xDSL as a project.

alexarice:
This is on me as I didn't really make my point clear (or make my point at all), but I'm not really convinced that changes like this are worth doing. There's a cost in breaking abstraction barriers and making the code more complex and I'm not sure we had really looked at other options, especially when the cause of the slow down here was identified as `OpOperands` being a dataclass.

superlopuh:
I agree in general but I feel like for this particular change it's just using a private field instead of a public helper method which feels like something that's generally fine to do. I'm also not sure that it was identified as it being a dataclass vs creating an unnecessary intermediate object given that we have access to the real deal in this context.

alexarice:
I think I agree that this case is probably fine, as this is still part of the core API, but I'm not sure I view `OpOperands` as a public helper. It's really part of abstraction layer and bypassing it could lead to wrongly formed IR (it is fine here as we do not modify anything).

superlopuh:
It sounds like we're broadly on the same page, I definitely think that this sort of thing would be quite fun to present one method at a time in a meeting, so we can have a bit more of a high-bandwith conversation about it before it gets merged. This could be in one of the xDSL weekly meetings or in a separate one.

PR #3964, opened 2025-02-25
backend: use negative indices to track infinite registers

superlopuh:
The overall goal is to be able to track registers when allocating by their indices, instead of an instance. On some platforms, 32-bit registers and 64-bit registers overlap, so the type of the register class does not uniquely identify the space of non-aliasing registers. Even in RISC-V, we already have this problem, as the `x` register names alias with the `a`, `t`, `s`, etc register names.

This PR is a first step in that direction, which lets us uniquely specify an index for the "infinite" register set for each class. Subsequent PRs will actually leverage this index.

### DIFF
--- a/tests/dialects/test_riscv.py
+++ b/tests/dialects/test_riscv.py
@@ -6,7 +6,6 @@ from xdsl.dialects.builtin import (
     IntAttr,
     IntegerAttr,
     ModuleOp,
-    NoneAttr,
     Signedness,
     i32,
 )
@@ -35,7 +34,7 @@ def test_add_op():
     assert a2.type.index == IntAttr(12)
 
     # Registers that aren't predefined should not have an index.
-    assert isinstance(riscv.IntRegisterType("j_1").index, NoneAttr)
+    assert riscv.IntRegisterType("j_1").index == IntAttr(-2)
 
 
 def test_csr_op():
--- a/tests/backend/test_register.py
+++ b/tests/backend/test_register.py
@@ -1,31 +1,75 @@
 import pytest
 
 from xdsl.backend.register_type import RegisterType
+from xdsl.dialects.builtin import IntAttr, StringAttr
 from xdsl.irdl import irdl_attr_definition
 
 
-@irdl_attr_definition
-class TestRegister(RegisterType):
-    name = "test.reg"
-
-    def verify(self) -> None: ...
-
-    @classmethod
-    def instruction_set_name(cls) -> str:
-        return "TEST"
-
-    @classmethod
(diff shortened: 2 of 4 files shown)


superlopuh:
There were a few tests I didn't catch here I'll take the time to fix them properly, which might benefit from a bit more of a refactor, moved this to draft for now.

tests/dialects/test_riscv.py
```diff
@@ -35,7 +34,7 @@ def test_add_op():
     assert a2.type.index == IntAttr(12)
 
     # Registers that aren't predefined should not have an index.
-    assert isinstance(riscv.IntRegisterType.infinite_register(1).index, NoneAttr)
+    assert riscv.IntRegisterType.infinite_register(1).index == IntAttr(-2)
```
compor, line 37: Why do I have to keep track that 1 is -2, instead of a straightforward match from 1 to -1, 2 to -2, etc.?
The 0 index is taken, and this infinite register set doesn't overlap with anything else?

What is the point of being subjected to this mild number juggling? What does it save us?

superlopuh: Hmm, do you have to keep track of it? To me it seems like an implementation detail of the storage of negative indices, it doesn't really matter much whether it's -1 or -2, right? It's never shown in the IR, and users are never expected to construct the attribute using the init, as opposed to one of the `from_` helpers.

compor: Even more so then, Why do this? If I visited this codebase for the first time, I'd stumble in comprehension here

superlopuh: OK so if I understand correctly, you would propose for the infinite registers to not start at `j0`, but at `j1`, where the index attribute would hold `-1`?

compor: Actually, did we change this recently, or I misremember (the start of the j registers I mean)? If so, then maybe that is too much flip-flopping

superlopuh: The thing I changed recently was that int and float riscv registers actually check that they're the correct (j/fj) prefix, not really the numbering.

superlopuh: I'm happy to change the 0- or 1-indexed infinite registers, I don't feel strongly either way, we just need some index-based representation.

compor: hmm, I changed my mind, let's just keep it that way, starting from 0 for infinite registers raises less questions and it seems more natural, following actual ISAs.

compor (approved):

PR #3966, opened 2025-02-25
csl: (dialects) Change csl to use signless integers

n-io:
~~Stacked PR for #3802~~

### DIFF
--- a/xdsl/dialects/csl/csl.py
+++ b/xdsl/dialects/csl/csl.py
@@ -1255,7 +1255,7 @@ class IncrementDsdOffsetOp(IRDLOperation):
     name = "csl.increment_dsd_offset"
 
     op = operand_def(DsdType)
-    offset = operand_def(i16_value)
+    offset = operand_def(IntegerType(16))
     elem_type = prop_def(DsdElementTypeConstr)
     result = result_def(DsdType)
 
@@ -1283,7 +1283,7 @@ class SetDsdLengthOp(IRDLOperation):
 
     name = "csl.set_dsd_length"
     op = operand_def(DsdType)
-    length = operand_def(u16_value)
+    length = operand_def(IntegerType(16))
     result = result_def(DsdType)
 
     traits = traits_def(Pure(), SetDsdLengthOpHasCanonicalizationPatternsTrait())
@@ -1311,7 +1311,7 @@ class SetDsdStrideOp(IRDLOperation):
 
     name = "csl.set_dsd_stride"
     op = operand_def(DsdType)
-    stride = operand_def(IntegerType(8, Signedness.SIGNED))
+    stride = operand_def(IntegerType(8))
     result = result_def(DsdType)
 
     traits = traits_def(Pure(), SetDsdStrideOpHasCanonicalizationPatternsTrait())
--- a/xdsl/dialects/csl/csl.py
+++ b/xdsl/dialects/csl/csl.py
@@ -362,6 +362,7 @@ f16_pointer = PtrType(
 f32_pointer = PtrType(
     [Float32Type(), PtrKindAttr(PtrKind.SINGLE), PtrConstAttr(PtrConst.VAR)]
 )
+i8_value = IntegerType(8, Signedness.SIGNED)
 u16_value = IntegerType(16, Signedness.UNSIGNED)
 i16_value = IntegerType(16, Signedness.SIGNED)
 u32_value = IntegerType(32, Signedness.UNSIGNED)
@@ -1313,7 +1314,7 @@ class SetDsdStrideOp(IRDLOperation):
 
     name = "csl.set_dsd_stride"
     op = operand_def(DsdType)
-    stride = operand_def(eq(i8) | eq(IntegerType(8, Signedness.SIGNED)))
+    stride = operand_def(eq(i8) | eq(i8_value))
     result = result_def(DsdType)
 
     traits = traits_def(Pure(), SetDsdStrideOpHasCanonicalizationPatternsTrait())
--- a/xdsl/dialects/csl/csl.py
+++ b/xdsl/dialects/csl/csl.py
@@ -1257,7 +1257,7 @@ class IncrementDsdOffsetOp(IRDLOperation):
     name = "csl.increment_dsd_offset"
 
     op = operand_def(DsdType)
-    offset = operand_def(i16)
+    offset = operand_def(eq(i16) | eq(i16_value))
     elem_type = prop_def(DsdElementTypeConstr)
     result = result_def(DsdType)
 
@@ -1285,7 +1285,7 @@ class SetDsdLengthOp(IRDLOperation):
 
     name = "csl.set_dsd_length"
     op = operand_def(DsdType)
-    length = operand_def(i16)
+    length = operand_def(eq(i16) | eq(u16_value))
     result = result_def(DsdType)
 
     traits = traits_def(Pure(), SetDsdLengthOpHasCanonicalizationPatternsTrait())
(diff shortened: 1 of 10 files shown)


compor:
Thanks for this @n-io; we should move the `print_csl.mlir` fixes here as well, since they are unrelated (strictly) speaking to #3802. Let me know how you want go about it, happy for you do edit my branch and move it over at your convenience

n-io:
> Thanks for this @n-io; we should move the `print_csl.mlir` fixes here as well, since they are unrelated (strictly) speaking to #3802. Let me know how you want go about it, happy for you do edit my branch and move it over at your convenience

This is taking a slightly different approach, though, of making dsd ops signless. The casts proposed in #3802 wouldn't be needed if we decide to go for this.

xdsl/transforms/memref_to_dsd.py
```diff
@@ -227,11 +226,7 @@ def _update_strides(
         static_strides = cast(Sequence[int], subview.static_strides.get_values())
 
         if static_strides[0] == memref.SubviewOp.DYNAMIC_INDEX:
-            ops.append(
-                cast_op := arith.IndexCastOp(
-                    subview.strides[0], IntegerType(8, Signedness.SIGNED)
-                )
-            )
+            ops.append(cast_op := arith.IndexCastOp(subview.strides[0], IntegerType(8)))
```
superlopuh, line 229: i8?

superlopuh:
Would it be possible to retarget this to main and isolate the changes to be csl-related? That way we can merge it first and tighten the arith.constant constraints in a later PR.

dk949:
> @dk949 I'd love your input on the printing of dsd functions (`increment_dsd_offset`, `set_dsd_(length|stride)`). The tests will currently put literals without the `@as()` (🥳). However, I'm not sure what would happen if we pass a var of the wrong type, or how we would address this.

RE: using wrong types for dsd manipulation functions, you can use any type there (any signedness, any width) as long as the number is compile time known. So:

```zig
    const i32_const : i32 = 10;
    const i16_const : i16 = 10;

    const my_dsd : mem1d_dsd = @set_dsd_length(src, i32_const); // works
    const my_dsd : mem1d_dsd = @set_dsd_length(src, i16_const); // works
    const my_dsd : mem1d_dsd = @set_dsd_length(src, @as(i16, 10)); // works

    var i16_var : i16 = 10;
    // const my_dsd : mem1d_dsd = @set_dsd_length(src, i16_var); // does not work
```

dk949 (approved):
This should all work, since we only use `comptime` known values

n-io:
> RE: using wrong types for dsd manipulation functions, you can use any type there (any signedness, any width) as long as the number is compile time known. So:
> 
> ```zig
>     var i16_var : i16 = 10;
>     // const my_dsd : mem1d_dsd = @set_dsd_length(src, i16_var); // does not work
> ```

That's the case I'm worried about.

n-io:
Say we have
```mlir
%base_dsd, %test = "test.op"() : () -> (!csl<dsd mem1d_dsd>, i16) // assume this comes from somewhere sensible
%test_si = csl.mlir.signedness_cast %test : i16 to ui16
%new_dsd0 = "csl.set_dsd_length"(%base_dsd, %test) : (!csl<dsd mem1d_dsd>, i16) -> !csl<dsd mem1d_dsd>
//%new_dsd1 = "csl.set_dsd_length"(%base_dsd, %test_si) : (!csl<dsd mem1d_dsd>, ui16) -> !csl<dsd mem1d_dsd> // will not work
```

resulting in:
```zig
  //unknown op TestOp(%base_dsd, %test = "test.op"() : () -> (!csl<dsd mem1d_dsd>, i16))
  const test_si : u16 = @as(u16, test);
  const new_dsd0 : mem1d_dsd = @set_dsd_length(base_dsd, test);
```

To pass a run-time known value, we'd have to call `@as()`, which it doesn't automatically, but we cannot call `@as()` since it doesn't accept unsigned int in the signature.

xdsl/dialects/csl/csl.py
```diff
@@ -1255,7 +1257,7 @@ class IncrementDsdOffsetOp(IRDLOperation):
     name = "csl.increment_dsd_offset"
 
     op = operand_def(DsdType)
-    offset = operand_def(i16_value)
+    offset = operand_def(eq(i16) | eq(i16_value))
```
superlopuh, line 1260:
```suggestion
    offset = operand_def(base(i16) | eq(i16_value))
```

xdsl/dialects/csl/csl.py
```diff
@@ -1283,7 +1285,7 @@ class SetDsdLengthOp(IRDLOperation):
 
     name = "csl.set_dsd_length"
     op = operand_def(DsdType)
-    length = operand_def(u16_value)
+    length = operand_def(eq(i16) | eq(u16_value))
```
superlopuh, line 1288:
```suggestion
    length = operand_def(base(i16) | eq(u16_value))
```

xdsl/dialects/csl/csl.py
```diff
@@ -1311,7 +1313,7 @@ class SetDsdStrideOp(IRDLOperation):
 
     name = "csl.set_dsd_stride"
     op = operand_def(DsdType)
-    stride = operand_def(IntegerType(8, Signedness.SIGNED))
+    stride = operand_def(eq(i8) | eq(IntegerType(8, Signedness.SIGNED)))
```
superlopuh, line 1316:
```suggestion
    stride = operand_def(base(i8) | base(IntegerType(8, Signedness.SIGNED)))
```

superlopuh: I'm a little surprised these don't raise errors

compor:
> > Thanks for this @n-io; we should move the `print_csl.mlir` fixes here as well, since they are unrelated (strictly) speaking to #3802. Let me know how you want go about it, happy for you do edit my branch and move it over at your convenience
> 
> This is taking a slightly different approach, though, of making dsd ops signless. The casts proposed in #3802 wouldn't be needed if we decide to go for this.

all good!

xdsl/dialects/csl/csl.py
compor: true, and I'm now very confused about our constraints.

xdsl/dialects/csl/csl.py
superlopuh: Sorry please disregard, I got confused by the value thing

xdsl/dialects/csl/csl.py
superlopuh: Actually it's my mistake my suggested change is bad, I'll suggest another

xdsl/dialects/csl/csl.py
superlopuh, line 1260: The value naming is confusing, can you please rename this type to be consistent with `i16` and the way it's printed in the IR?
```suggestion
    offset = operand_def(eq(i16) | eq(si16))
```

xdsl/dialects/csl/csl.py
superlopuh, line 1316: Might as well also declare this type locally for consistency
```suggestion
    stride = operand_def(eq(i8) | eq(si8))
```

dk949 (approved):
Looks great!

Just to confirm, is `csl.mlir.signedness_cast` currently generated by anything, or is it just there for when we need it?

xdsl/dialects/csl/csl.py
n-io: It makes it consistent with the cerebras docs.

n-io:
> Looks great!
> 
> Just to confirm, is `csl.mlir.signedness_cast` currently generated by anything, or is it just there for when we need it?

Good point, I don't see it used anywhere except in hand-written mlir.

Issue #3937, opened 2025-02-19
dialects: (x86) mv op should not take two operands


superlopuh:
The `x86.rm.mov` operation inherits from `R_RM_Operation`, which is for binary operations that update their first register argument in-place. This doesn't make sense for mv, as it moves a value from one register into another, discarding that register's initial value.

compor:
Resolved in #3967

PR #3967, opened 2025-02-25
bug: (x86) memory-to-register mv operation should not take two operands

qaco:
Addresses [this issue](https://github.com/xdslproject/xdsl/issues/3937)

### DIFF
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -609,6 +609,59 @@ class R_RM_Operation(
         return {"offset"}
 
 
+class R_M_Operation(
+    Generic[R1InvT, R2InvT], X86Instruction, X86CustomFormatOperation, ABC
+):
+    """
+    A base class for x86 operations that have one register and one memory access with an optional offset.
+    """
+
+    r1 = operand_def(R1InvT)
+    offset = attr_def(IntegerAttr, default_value=IntegerAttr(0, 64))
+
+    result = result_def(R2InvT)
+
+    def __init__(
+        self,
+        r1: Operation | SSAValue,
+        offset: int | IntegerAttr,
+        *,
+        comment: str | StringAttr | None = None,
+        result: R2InvT,
+    ):
+        if isinstance(offset, int):
+            offset = IntegerAttr(offset, 64)
+        if isinstance(comment, str):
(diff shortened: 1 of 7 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -688,11 +741,11 @@ class RM_XorOp(R_RM_Operation[GeneralRegisterType, GeneralRegisterType]):
 
 
 @irdl_op_definition
-class RM_MovOp(R_RM_Operation[GeneralRegisterType, GeneralRegisterType]):
+class RM_MovOp(R_M_Operation[GeneralRegisterType, GeneralRegisterType]):
     """
-    Copies the value from the memory location pointed to by r2 into r1.
+    Copies the value from the memory location pointed to by operand register into output register.
```
superlopuh, line 746: we've been using "source" and "destination" in riscv and asm, would that make sense as a description in x86 land also?

superlopuh (approved):

xdsl/dialects/x86/ops.py
```diff
@@ -688,11 +741,11 @@ class RM_XorOp(R_RM_Operation[GeneralRegisterType, GeneralRegisterType]):
 
 
 @irdl_op_definition
-class RM_MovOp(R_RM_Operation[GeneralRegisterType, GeneralRegisterType]):
+class RM_MovOp(R_M_Operation[GeneralRegisterType, GeneralRegisterType]):
     """
-    Copies the value from the memory location pointed to by r2 into r1.
+    Copies the value from the memory location pointed to by source register into destination register.
```
compor, line 746:
```suggestion
    Copies the value from the memory location pointed to by source register r1 into destination register r2.
```

xdsl/dialects/x86/ops.py
```diff
@@ -688,11 +741,11 @@ class RM_XorOp(R_RM_Operation[GeneralRegisterType, GeneralRegisterType]):
...
     ```C
-    x[r1] = [x[r2]]
+    x[output] = [x[operand]]
```
compor, line 748: Can you please adjust these based on r1 and r2 as per the rest of the file; I don't understand what `output` or `operand` is as they appear nowhere else in the API?

compor (approved):

PR #3968, opened 2025-02-26
backend: (riscv) make rd unallocated by default

superlopuh:
If I remember correctly, this was done as a kind of defensive API measure to make sure that users of the riscv operations knew to assign to the result register an appropriate value. I no longer think that the benefits of explicit result values outweigh the cost of verbose code everywhere. I'm currently refactoring register types, and I'd like to change the init and parsing situation, where it will no longer make sense for the default init with no parameters to create an unallocated register, so the API for those is about to change towards something less ergonomic, which will make the no default parameter situation worse.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -556,6 +556,34 @@ class RdRsRsOperation(
         return self.rd, self.rs1, self.rs2
 
 
+class RdRsRsIntegerOperation(
+    Generic[RS1InvT, RS2InvT], RdRsRsOperation[IntRegisterType, RS1InvT, RS2InvT], ABC
+):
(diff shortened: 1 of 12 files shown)


xdsl/dialects/riscv.py
```diff
@@ -624,17 +652,13 @@ def __init__(
         self,
         immediate: int | IntegerAttr | str | LabelAttr,
         *,
-        rd: IntRegisterType | str | None = None,
```
superlopuh, line 627: allowing str to be passed here was also a mistake, I think, we never used it in practice, and it just makes the code bloated

qaco (approved):
Nice !

PR #3971, opened 2025-02-26
testing: move assert_print_op definition to test_printer

superlopuh:
This is now the only file that uses this function, and I feel like that's where that function belongs, hope you all agree.


tests/test_printer.py
```diff
@@ -954,3 +954,66 @@ def test_get_printed_name():
     printed = StringIO()
     picked_name = Printer(printed).print_ssa_value(val)
     assert f"%{picked_name}" == printed.getvalue()
+
+
+def assert_print_op(
```
superlopuh, line 959: move the definition and explicitly spell out the arguments it can take

tests/test_printer.py
```diff
@@ -587,7 +587,7 @@ def test_generic_format():
     parser = Parser(ctx, prog)
     module = parser.parse_op()
 
-    assert_print_op(module, expected, None, False)
+    assert_print_op(module, expected, None, print_generic_format=False)
```
alexarice, line 590: Could we also change the `diagnostic` argument, either to default `None` or keyword only (or both).

alexarice (approved):
Seems good to me. Are all the arguments for `assert_print_op` used in one of its uses, because if not we could remove some.

superlopuh:
Just checked, and all the named arguments are used.

PR #3972, opened 2025-02-26
backend: (riscv) use named unallocated registers instead of empty init

superlopuh:
Two related changes:

 - removes the default value of "" for the register name
 - replaces previous unallocated register init calls with no args with the dedicated UNALLOCATED constants

This reduces the users of the init API, which I'm slowly refactoring.

### DIFF
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -176,11 +176,11 @@ class ConvertMemRefStoreOp(RewritePattern):
 
         rewriter.insert_op_before_matched_op(ops)
         match value.type:
-            case riscv.Registers.UNALLOCATED_INT:
+            case riscv.IntRegisterType():
                 new_op = riscv.SwOp(
                     ptr, value, 0, comment=f"store int value to memref of shape {shape}"
                 )
-            case riscv.Registers.UNALLOCATED_FLOAT:
+            case riscv.FloatRegisterType():
                 float_type = cast(AnyFloat, memref_type.element_type)
                 match float_type:
                     case Float32Type():
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -176,11 +176,11 @@ class ConvertMemRefStoreOp(RewritePattern):
 
         rewriter.insert_op_before_matched_op(ops)
         match value.type:
-            case riscv.IntRegisterType():
+            case riscv.Registers.UNALLOCATED_INT:
                 new_op = riscv.SwOp(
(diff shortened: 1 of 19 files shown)


xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -176,11 +176,11 @@ def match_and_rewrite(self, op: memref.StoreOp, rewriter: PatternRewriter):
 
         rewriter.insert_op_before_matched_op(ops)
         match value.type:
-            case riscv.IntRegisterType():
+            case riscv.Registers.UNALLOCATED_INT:
```
alexarice, line 179: This one looks suspect, this was just being used for an isinstance check right?

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -176,11 +176,11 @@ def match_and_rewrite(self, op: memref.StoreOp, rewriter: PatternRewriter):
...
                 new_op = riscv.SwOp(
                     ptr, value, 0, comment=f"store int value to memref of shape {shape}"
                 )
-            case riscv.FloatRegisterType():
+            case riscv.Registers.UNALLOCATED_FLOAT:
```
alexarice, line 183: Same here

xdsl/backend/riscv/lowering/utils.py
```diff
@@ -33,7 +33,7 @@ def cast_to_regs(values: Iterable[SSAValue]) -> tuple[list[Operation], list[SSAV
         if not isinstance(value.type, riscv.RISCVRegisterType):
             register_type = register_type_for_type(value.type)
             cast_op = builtin.UnrealizedConversionCastOp.get(
-                (value,), (register_type(),)
+                (value,), (register_type(""),)
```
alexarice, line 36: Is the point of this PR not to remove `""` as a special register? Why can't the UNALLOCATED_INT register be reused?

alexarice (approved):
I don't know much about riscv but looks reasonable

xdsl/backend/riscv/lowering/utils.py
superlopuh: All register types with an empty name are the unallocated register for their class, and there's no API like `.unallocated_register` to use instead of this right now. I'm now thinking that that might be a good idea but I'd like to do that as a separate PR.

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: The tests passed because the registers were never allocated, which might actually be safe to assume here, but I'll remove this change from this PR

xdsl/backend/riscv/lowering/utils.py
alexarice: Sounds good.

PR #3973, opened 2025-02-26
testing: use importlib import mode in pytest

superlopuh:
This makes the tests behave more like how clients of xDSL would, with the xdsl module in the venv rather than locally available in the current directory.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -166,7 +166,7 @@ exclude = [
 python_files = ["tests/*test_*.py", "docs/*test_*.py"]
 python_classes = "Test_*"
 python_functions = "test_*"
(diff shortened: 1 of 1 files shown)


compor (approved):
what's the benefit of this? better isolation?

superlopuh:
I probably should have included a bit more motivation for this, here is the docs page: https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html

It changes the way that the path is constructed, to not include the current path in PATH, which means that xDSL has to be installed for the pytests to pass. One benefit I only just found out about is that we can have tests with the same name with this change, which is nice because we frequently reuse names of files in xDSL, and use the folder structure to disambiguate, so it should remove some redundancy also.

PR #3974, opened 2025-02-26
backend: add unallocated class method on RegisterType

superlopuh:
Makes the API a little more explicit.

### DIFF
--- a/tests/backend/test_register.py
+++ b/tests/backend/test_register.py
@@ -29,3 +29,8 @@ def test_register_clashes():
         match="Invalid 'infinite' register name: x0 clashes with finite register set",
     ):
         TestRegister.infinite_register(0)
+
+
+def test_unallocated_register():
+    assert not TestRegister.unallocated().is_allocated
+    assert TestRegister("x0").is_allocated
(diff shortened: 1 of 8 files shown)


compor:
between this and the other 2 PRs (#3926, #3972) I'm not sure where this is heading - maybe some discussion on the outlook would make this a bit easier to review.

superlopuh:
Yeah I messed up on that one, that's my bad. I don't remember exactly why I had removed it in the first place, but it wasn't a good move. My current aim is for the register type init to just take the spelling and index, to move some logic to verification, and to add `from_spelling` and `from_index` helpers. While it would be possible for the init to be overloaded to take default values for both it just seems like a bad idea. Sorry for the churn. I'll now actually first do the stack of commits locally to make sure that it works and then open the rest of the PRs to try and minimise the churn for this particular bit of code.

superlopuh:
To separately answer your question on the other PR, one of the recent changes is that more of the operations set the default return register to be unallocated, so the verbosity becomes a bit less of an issue.

JosseVanDelm (approved):
I like it!

PR #3976, opened 2025-02-26
misc: escape code in docstrings

superlopuh:
These are currently stopping the docs website from generating correctly.

### DIFF
--- a/xdsl/dialects/csl/csl.py
+++ b/xdsl/dialects/csl/csl.py
@@ -1893,15 +1893,18 @@ class AddressOfOp(IRDLOperation):
     def _verify_memref_addr(self, val_ty: MemRefType[Attribute], res_ty: PtrType):
         """
         Verify that if the address of a memref is taken, the resulting pointer is either:
-            A single pointer to the array type or
-            A many pointer to the array element type
+        - A single pointer to the array type or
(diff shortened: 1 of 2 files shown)


xdsl/dialects/csl/csl.py
```diff
@@ -1893,15 +1893,18 @@ def __init__(self, value: SSAValue | Operation, result_type: PtrType):
     def _verify_memref_addr(self, val_ty: MemRefType[Attribute], res_ty: PtrType):
         """
         Verify that if the address of a memref is taken, the resulting pointer is either:
-            A single pointer to the array type or
-            A many pointer to the array element type
+        - A single pointer to the array type or
+        - A many pointer to the array element type
+
         E.g.
+        ```mlir
```
dk949, line 1900: Not sure if the docs generator supports it, but this would ideally use the `zig` highlighting, since it's a CSL code block, not MLIR

dk949 (approved):

xdsl/dialects/csl/csl.py
superlopuh: vscode seems to highlight these syntaxes correctly for moth zig and mlir! Not sure about mkdocs

PR #3979, opened 2025-02-27
backend: (x86) insert .global directives when lowering func.func

qaco:


xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -55,6 +55,10 @@ def match_and_rewrite(self, op: func.FuncOp, rewriter: PatternRewriter):
                     "Cannot lower function parameters bigger than 64 bits (not implemented)"
                 )
 
+        if op.sym_visibility == StringAttr("public"):
+            directive_op = x86.DirectiveOp(".global", op.sym_name)
+            rewriter.insert_op_before_matched_op(directive_op)
+
```
compor, line 61: Keep a lowering test that does *not* have a `public` symbol? Not sure if there is a test that does this elsewhere.

qaco: Done !

tests/filecheck/backend/x86/convert_func_to_x86.mlir
```diff
@@ -1,5 +1,19 @@
...
+  func.return %0: i32
+}
+// CHECK-NEXT:    x86_func.func @foo_const(%0 : !x86.reg<rsp>) -> !x86.reg<rax> {
+// CHECK-NEXT:      %1 = "test.op"() : () -> i32
+// CHECK-NEXT:      %2 = builtin.unrealized_conversion_cast %1 : i32 to !x86.reg
+// CHECK-NEXT:      %3 = x86.get_register : () -> !x86.reg<rax>
+// CHECK-NEXT:      %4 = x86.rr.mov %2, %3 : (!x86.reg, !x86.reg<rax>) -> !x86.reg<rax>
+// CHECK-NEXT:      x86_func.ret
+// CHECK-NEXT:    }
+
```
compor, line 16: This is missing the test case separator (for `--split-input-file`) `// -----`.
I think it'd be could if these tests are separate.

PR #3980, opened 2025-02-27
dialects: (stablehlo) Add support for xor and op operations

Hatsunespica:
This PR includes following changes:
1. Add an abstract elementwise binary operations for integer tensor types `IntegerTensorLikeElementwiseBinaryOperation` like https://github.com/xdslproject/xdsl/blob/a072093aa7815d944a6fb97d17d13bcf4777cda4/xdsl/dialects/math.py#L38
2. Replace `AndOp`'s parent class with `IntegerTensorLikeElementwiseBinaryOperation`
3. Add Xor and Or Op 
4. Update on test cases

### DIFF
--- a/xdsl/dialects/stablehlo.py
+++ b/xdsl/dialects/stablehlo.py
@@ -55,12 +55,31 @@ from xdsl.printer import Printer
 from xdsl.traits import IsTerminator
 from xdsl.utils.exceptions import VerifyException
 
+IntegerTensorType: TypeAlias = TensorType[IntegerType]
+
 # region Abstract Base Classes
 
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/stablehlo.py
```diff
@@ -55,6 +55,8 @@
 from xdsl.traits import IsTerminator
 from xdsl.utils.exceptions import VerifyException
 
+IntegerTensorType: TypeAlias = TensorType[IntegerType]
```
superlopuh, line 58: I'm not sure of the utility of this, let's just inline it below?

superlopuh (approved):
Thank you!

PR #3984, opened 2025-02-28
backend: (riscv) make LiOp and float binary ops not take str for rd

superlopuh:
There's no need for the string, the user can create the register themselves, simplifies API.


xdsl/dialects/riscv.py
```diff
@@ -2998,45 +2994,6 @@ def assembly_line_args(self) -> tuple[AssemblyInstructionArg, ...]:
         return self.rd, self.rs1, self.rs2, self.rs3
 
 
-class RdRsRsFloatFloatIntegerOperation(
```
superlopuh, line 3001: this base class was just unused

JosseVanDelm (approved):

PR #3986, opened 2025-02-28
core: Move SSAValue definitions around (NFC)

math-fehr:
Stacked PRs:
 * #3991
 * #3990
 * #3989
 * #3988
 * #3987
 * __->__#3986

--- --- ---

### core: Move SSAValue definitions around (NFC)

Move the definition of SSAValue after the definition of Attributes.


alexarice (approved):
Why though?

PR #3987, opened 2025-02-28
core: Make SSAValue generic

math-fehr:
Stacked PRs:
 * #3991
 * #3989
 * #3988
 * __->__#3987

--- --- ---

### core: Make SSAValue generic

Adds a generic parameter to SSAValue that constrains the type associated
with the SSAValue.
This is useful whenever a function only accepts SSAValue of a given type.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -480,13 +480,13 @@ class IRWithUses(ABC):
 
 
 @dataclass(eq=False)
-class SSAValue(IRWithUses, ABC):
+class SSAValue(Generic[AttributeCovT], IRWithUses, ABC):
     """
     A reference to an SSA variable.
     An SSA variable is either an operation result, or a basic block argument.
     """
 
-    _type: Attribute
+    _type: AttributeCovT
     """Each SSA variable is associated to a type."""
 
     _name: str | None = field(init=False, default=None)
@@ -494,7 +494,7 @@ class SSAValue(IRWithUses, ABC):
     _name_regex: ClassVar[re.Pattern[str]] = re.compile(r"([A-Za-z_$.-][\w$.-]*)")
 
     @property
-    def type(self) -> Attribute:
+    def type(self) -> AttributeCovT:
         return self._type
 
     @property
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -119,6 +119,7 @@ class Printer(BasePrinter):
                 self.print_string(arg)
                 continue
             if isinstance(arg, SSAValue):
+                arg = cast(SSAValue[Attribute], arg)
                 self.print_ssa_value(arg)
                 continue
             if isinstance(arg, Attribute):
--- a/xdsl/utils/test_value.py
+++ b/xdsl/utils/test_value.py
@@ -1,9 +1,11 @@
+from typing import Generic
+
 import pytest
 
-from xdsl.ir import Block, Operation, SSAValue
+from xdsl.ir import AttributeCovT, Block, Operation, SSAValue
 
 
-class TestSSAValue(SSAValue):
+class TestSSAValue(Generic[AttributeCovT], SSAValue[AttributeCovT]):
     @property
     def owner(self) -> Operation | Block:
         pytest.fail("Attempting to get the owner of a `TestSSAValue`")


xdsl/printer.py
```diff
@@ -119,6 +119,7 @@ def print(self, *argv: Any) -> None:
                 self.print_string(arg)
                 continue
             if isinstance(arg, SSAValue):
+                arg = cast(SSAValue[Attribute], arg)
```
alexarice, line 122: This is yucky but tbf this whole function is yucky

tests/test_ssa_value.py
```diff
@@ -59,3 +60,16 @@ def test_invalid_ssa_vals(name: str):
...
+    """
+    Check that SSAValue type hinting works as expected.
+    This is a type hinting test.
+    """
+    value = TestSSAValue(i32)
+
+    def foo(arg: SSAValue[IntegerType]):
+        pass
+
+    foo(value)
```
alexarice, line 75: is there any way we can make a test that shows a type error? I guess not as it would have to rely on pyright

xdsl/ir/core.py
```diff
@@ -478,13 +478,13 @@ def remove_use(self, use: Use):
...
 @dataclass(eq=False)
-class SSAValue(IRWithUses, ABC):
+class SSAValue(Generic[AttributeCovT], IRWithUses, ABC):
     """
     A reference to an SSA variable.
     An SSA variable is either an operation result, or a basic block argument.
     """
 
-    type: Attribute
+    type: AttributeCovT
```
alexarice, line 487: Not sure if this covered further down the stack, but the get method below probably doesn't have tight enough type bounds.

alexarice (approved):
`AttributeCovT` has a default value of `Attribute` right?

xdsl/printer.py
math-fehr: Yeah, the problem is that `isinstance(arg, SSAValue)` returns an `SSAValue[Unknown]`, and we cannot use `isa` in this file if I recall

tests/test_ssa_value.py
math-fehr: Yeah I'm discussing it with Sasha, we are trying to find a solution

xdsl/printer.py
alexarice: I'm not sure I fully grasped why pyright can't realise the type is `SSAValue[Attribute]` here using the lower bound on the type variable.

xdsl/ir/core.py
```diff
@@ -478,13 +478,13 @@ def remove_use(self, use: Use):
 
 
 @dataclass(eq=False)
-class SSAValue(IRWithUses, ABC):
+class SSAValue(Generic[AttributeCovT], IRWithUses, ABC):
```
superlopuh, line 481: My feeling is that this is unsound unless we make SSAValue immutable first, which blocks most of this PR stack.

xdsl/ir/core.py
math-fehr: I don't think it should actually have a better bound.
If you have an `SSAValue[T] | Operation`, you can only get an `SSAValue` back, because you don't know the type of the operation.
If you have an `SSAValue[T]`, then you wouldn't call `get` in the first place I think?

alexarice: Yeah, on second thoughts I think you are correct.

xdsl/ir/core.py
math-fehr: I am currently trying to see how easy it is to make this immutable, I'll probably add 1-2 commits in the stack

xdsl/printer.py
math-fehr: https://github.com/microsoft/pyright/discussions/3891

It *could* work if the field is frozen, but would require too many changes in pyright AFAIU

xdsl/ir/core.py
math-fehr: https://github.com/xdslproject/xdsl/pull/4125 is making them immutable now

superlopuh: Yay!

xdsl/printer.py
superlopuh, line 122:
```suggestion
                _arg: SSAValue[Attribute] = arg
```

superlopuh: Do default typevar types work here?

tests/test_ssa_value.py
superlopuh: I don't think we need this?

xdsl/printer.py
compor: I think it already defaults to `Attribute`, no?

math-fehr: I'm so confused on why this works in the first place actually.
This is technically a cast, right, why does this work?
I'm almost certain this behavior is a bug, as doing `_arg: SSAValue[IntegerType] = arg` works as well.

math-fehr: It doesn't in this case.
The default only works when doing type checking, here `isinstance(arg, SSAValue)` doesn't use the default as `SSAValue` is a runtime value.
So here, `arg` is a `SSAValue[Unknown]`, which is why I casted it.
Sasha found a workaround, but I believe it is a pyright error, I'm reporting it for now.

I would rather keep the cast for now, while reporting the bug on pyright?

tests/test_ssa_value.py
math-fehr: removed

xdsl/printer.py
math-fehr: Reported here: https://github.com/microsoft/pyright/issues/10214

superlopuh: Yeah let's keep the cast

superlopuh: Yeah seems like a bug indeed

PR #3988, opened 2025-02-28
core: Add support for typed SSA values in isa

math-fehr:
Stacked PRs:
 * #3991
 * #3989
 * __->__#3988
 * #3987

--- --- ---

### core: Add support for typed SSA values in isa

`isa(arg, SSAValue[T])` will check that `arg` is an SSA value, and that
its type is `T`.

PR #3990, opened 2025-02-28
dialects: Raise a parsing error in accfg.setup when result type is invalid

math-fehr:
Stacked PRs:
 * #3991
 * #3989
 * #3988
 * #3987
 * #4125
 * __->__#3990

--- --- ---

### dialects: Raise a parsing error in accfg.setup when result type is invalid

As there are a redundancy in the assembly format between the first
parsed string and the result type, this should be a parsing error.

### DIFF
--- a/tests/filecheck/dialects/accfg/invalid.mlir
+++ b/tests/filecheck/dialects/accfg/invalid.mlir
@@ -1,0 +1,7 @@
+// RUN: xdsl-opt --split-input-file --parsing-diagnostics %s | filecheck %s
+
+%one = "test.op"() : () -> i32
+%zero = arith.constant 0 : i32
+%state = accfg.setup "acc1" to ("A" = %one : i32) : !accfg.state<"acc2">
+
+// CHECK: expected !accfg.state<"acc1">, but got !accfg.state<"acc2">
--- a/xdsl/dialects/accfg.py
+++ b/xdsl/dialects/accfg.py
@@ -341,14 +341,19 @@ class SetupOp(IRDLOperation):
             attributes = parser.parse_optional_attr_dict()
 
         parser.parse_punctuation(":")
+        pos = parser.pos
         res_typ = parser.parse_type()
+        if res_typ != StateType(accelerator):
+            parser.raise_error(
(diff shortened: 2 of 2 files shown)


superlopuh (approved):

xdsl/dialects/accfg.py
```diff
@@ -341,14 +341,19 @@ def parse_itm() -> tuple[str, SSAValue]:
             attributes = parser.parse_optional_attr_dict()
 
         parser.parse_punctuation(":")
+        pos = parser.pos
```
alexarice, line 344: I might be blind but is this used anywhere?

alexarice (approved):
Seems reasonable though I'm not really sure why the operation parses/prints a type at this point if it's fully determined by the other inputs.

xdsl/dialects/accfg.py
superlopuh: last three chars of line 348

JosseVanDelm (approved):
Thanks @math-fehr !

JosseVanDelm:
> Seems reasonable though I'm not really sure why the operation parses/prints a type at this point if it's fully determined by the other inputs.

@alexarice, are you suggesting that you would expect to print/parse it as:
```mlir
%state = accfg.setup "acc1" to ("A" = %one : i32)
```

alexarice:
Yes

alexarice:
To expand a bit, I wouldn't necessarily expect the custom format to be a certain way (as in I don't feel it's wrong at the moment), but in this case it doesn't feel like the type adds much information to the user, and doesn't seem to be needed by the system.

tests/filecheck/dialects/accfg/invalid.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: xdsl-opt --split-input-file --parsing-diagnostics %s | filecheck %s
+
+%one = "test.op"() : () -> i32
+%zero = arith.constant 0 : i32
```
compor, line 4: Where is `%zero` used? Am I missing something?

compor (approved):

PR #3992, opened 2025-02-28
[WIP] CI: adopt beta dependabot support for uv

EdmundGoodman:
Dependabot has [added beta support for uv](https://github.com/dependabot/dependabot-core/issues/10478#issuecomment-2691330949). This PR updates the dependabot configuration to use uv instead of pip.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,9 @@
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
+enable-beta-ecosystems: true
 updates:
-  - package-ecosystem: "pip" # See documentation for possible values
+  - package-ecosystem: "uv" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,8 +4,9 @@
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
+enable-beta-ecosystems: true
```
superlopuh, line 7: I'm confused https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#enable-beta-ecosystems-

superlopuh: Ah no please disregard

superlopuh:
This is great but I would give them about a month to fix bugs before merging to xDSL

superlopuh:
Let's reopen this when we're sure that the feature works. I'm subscribed to the issue, they currently don't actually update the lockfile directly.

superlopuh:
https://github.com/dependabot/dependabot-core/issues/10478#issuecomment-2722722972

.github/dependabot.yml
superlopuh, line 7:
```suggestion
```

superlopuh:
Actually it looks like dependabot support is still pretty broken, closing again.

superlopuh:
I don't understand why they didn't use our approach of just re-generating the lockfile using uv...

PR #3993, opened 2025-03-01
documentation: use strict flag for building website and add to CI

superlopuh:
Let's make sure it always builds even if we're not deploying it yet.

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -1,0 +1,38 @@
+# This workflow will install Python dependencies, and build the website
+
+name: CI - Documentation Website
+
+on:
+  # Trigger the workflow on push or pull request,
+  # but only for the master branch
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+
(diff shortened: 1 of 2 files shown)


math-fehr (approved):

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,38 @@
...
+        cache-dependency-glob: "uv.lock"
+
+    - name: Set up Python ${{ matrix.python-version }}
+      run: uv python install ${{ matrix.python-version }}
+
+    - name: Install the project
+      run: VENV_EXTRAS="--all-extras" make venv
+
+    - name: Build documentation
+      run: uv run mkdocs build --strict
```
compor, line 37: is the `--strict` needed here? I thought having it in the [yaml](https://www.mkdocs.org/user-guide/configuration/#strict) file was enough?

superlopuh: I'm not 100% sure, seems fine to keep it in both places to me, but it's a good point

PR #3997, opened 2025-03-02
misc: store context as property on ParsingState

superlopuh:
Makes it more explicit that the variable contains all the state necessary for inference.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -21,7 +21,6 @@ from xdsl.ir import (
     TypedAttribute,
 )
 from xdsl.irdl import (
-    ConstraintVariableType,
     InferenceContext,
     IRDLOperation,
     IRDLOperationInvT,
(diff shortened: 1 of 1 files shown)


alexarice (commented):
I'm unsure if this change is sound with `InferenceContext` being mutable. If it's only something we use internally then it's probably fine

superlopuh:
Aren't they both just as mutable as each other?

alexarice:
True, I for some reason thought we were copying it when making the context but we obviously weren't

alexarice:
Could you remind me of the reason for the `InferenceContext`. Do we expect other things to be in there?

superlopuh:
That's the next PR!

superlopuh:
I'm actually going towards using the ConstraintContext for inference again.

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -51,7 +50,7 @@ class ParsingState:
     successors: list[Successor | None | Sequence[Successor]]
     attributes: dict[str, Attribute]
     properties: dict[str, Attribute]
-    variables: dict[str, ConstraintVariableType]
+    context: InferenceContext
```
alexarice, line 53: I feel `context` is a slightly ambiguous name but I'm getting the feeling that this is part of a bigger change that may make it make more sense

alexarice (approved):

xdsl/irdl/declarative_assembly_format.py
superlopuh: Yeah it's InferenceContext in this PR but will be a ConstraintContext in a future one.

PR #3998, opened 2025-03-02
misc: constrain ConstraintVariableType variadic to be a tuple

superlopuh:
Avoids unnecessary copies and exposes a more efficient membership method.


xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -249,6 +249,8 @@ def extract_var(self, a: ParsingState) -> ConstraintVariableType:
             assert types is not None
             if isinstance(types, Attribute):
                 types = (types,)
+            else:
```
alexarice, line 252: This looks like it introduces a copy rather than getting rid of any. Am I missing something?

superlopuh:
Actually I found a way to avoid this change, closing

PR #3999, opened 2025-03-02
misc: ConstraintContext keys properties are abstract sets

superlopuh:
Avoids unnecessary copies and exposes a more efficient membership method.


superlopuh:
@alexarice This is actually the commit I intended to open the PR for, hopefully the description now makes more sense.

alexarice (approved):
This makes more sense

PR #4000, opened 2025-03-02
core: split InferenceContext dictionaries

superlopuh:
When inferring, we know which dictionary to look into, same for when we construct the InferenceContext, so we might as well split it. PR-1 to using the ConstraintContext during inference.


alexarice (approved):
Looks good, we could consider trying to split up the extractors but that would be a larger change

superlopuh:
You have foretold the future 🔮

PR #3650, opened 2024-12-17
dialects: (vector) Add for `vector.transfer_read` and `vector.transfer_write` operations

watermelonwolverine:
Added stubs for "vector.transfer_read" and "vector.transfer_write".
Verification is missing but that also seems to be a lot of work.

Also fixed a small bug in TensorOrMemrefOf.verify (?)

PR #4006, opened 2025-03-03
core: (affine) Add Affine.binary

watermelonwolverine:
Cherry picked from #3650. Prerequisite for implementing `vector.transfer_read` and `vector.transfer_write` ops


watermelonwolverine:
While implementing this I realized that when composing two affine maps the expressions are not automatically simplified/constant folded. 
So, a resulting `AffineBinaryOpExpr` where both lhs and rhs are `AffineConstantExpr` is not simplified to a AffineConstantExpr.
That's why I added the `AffineExpr.as_constant` function. Is this by design or should the constant folding happen while composing?

watermelonwolverine:
MLIR simplifies while composing. This is the call stack, the simplification happens at the last call:

```
AffineExpr AffineExpr::compose(AffineMap map)
AffineExpr AffineExpr::replaceDimsAndSymbols(ArrayRef<AffineExpr> dimReplacements,ArrayRef<AffineExpr> symReplacements)
AffineExpr mlir::getAffineBinaryOpExpr(AffineExprKind kind, AffineExpr lhs, AffineExpr rhs)
AffineExpr AffineExpr::operator+(AffineExpr other)
static AffineExpr simplifyAdd(AffineExpr lhs, AffineExpr rhs)
```

watermelonwolverine:
Added `AffineExpr.binary` which mimics `AffineExpr mlir::getAffineBinaryOpExpr` from `AffineExpr.cpp`.
Now it should be pretty close to how MLIR is doing it

xdsl/ir/affine/affine_map.py
```diff
@@ -175,6 +175,39 @@ def compose(self, other: AffineMap) -> AffineMap:
...
+        """
+
+        if self.num_symbols != 0:
+            raise ValueError(
+                "Cannot compose a AffineMap with values if the map has symbols."
+            )
+
+        expressions: list[AffineExpr] = []
+        for value in values:
+            expressions.append(AffineExpr.constant(value))
```
superlopuh, line 200:
```suggestion
        expressions = tuple(AffineExpr.constant(value) for value in values)
```

xdsl/ir/affine/affine_map.py
```diff
@@ -175,6 +175,39 @@ def compose(self, other: AffineMap) -> AffineMap:
...
+        result_map = self.compose(AffineMap(0, 0, tuple(expressions)))
+
+        result: list[int] = list()
+
+        for res in result_map.results:
+            assert isinstance(res, AffineConstantExpr)
+            result.append(res.value)
+
+        return tuple(result)
```
superlopuh, line 209:
```suggestion
        result = tuple(res.value for res in result_map.results)
        return result
```

xdsl/ir/affine/affine_map.py
```diff
@@ -175,6 +175,39 @@ def compose(self, other: AffineMap) -> AffineMap:
             results=results,
         )
 
+    def compose_with_values(self, values: Sequence[int]) -> tuple[int, ...]:
```
superlopuh, line 178: Is this not just `eval`?

watermelonwolverine: Huh.. I think you might be right :sweat_smile: 
MLIR does have a `SmallVector<int64_t, 4> AffineMap::compose(ArrayRef<int64_t> values)` function but that is also in essence an eval function

xdsl/ir/affine/affine_expr.py
```diff
@@ -37,6 +37,31 @@ def dimension(position: int) -> AffineExpr:
     def symbol(position: int) -> AffineExpr:
         return AffineSymExpr(position)
 
+    @staticmethod
+    def binary(
```
superlopuh, line 41: I've been thinking about this, I think you've found an important deficiency in the API. I have a feeling that this API isn't quite right, but it addresses an important problem. How do you feel about adding a method `AffineBinaryOpKind.simplified(lhs: AffineExpr, rhs: AffineExpr)` that would perform the simplification in a generic way? That helper could then be used in the `__` operators on AffineExpr, so the logic would already be tested.

watermelonwolverine: I'm not quite sure I understand what you mean. Do you mean adding something like this to `AffineBinaryOpExpr` (or did you really mean `AffineBinaryOpKind`?:

```python
    @staticmethod
    def simplified(
        kind: AffineBinaryOpKind,
        lhs: AffineExpr,
        rhs: AffineExpr,
    ) -> AffineExpr:
        match kind:
            case AffineBinaryOpKind.Add:
                return AffineBinaryOpExpr.simplified_add(lhs, rhs)
            case AffineBinaryOpKind.Mul:
                return AffineBinaryOpExpr.simplified_mul(lhs, rhs)
            case AffineBinaryOpKind.Mod:
                return AffineBinaryOpExpr.simplified_mod(lhs, rhs)
            case AffineBinaryOpKind.FloorDiv:
                return AffineBinaryOpExpr.simplified_floor_div(lhs, rhs)
            case AffineBinaryOpKind.CeilDiv:
                return AffineBinaryOpExpr.simplified_ceil_div(lhs, rhs)
```

That would work, but really wouldn't do much other than moving and consolidating all the logic that basically already exists (like `_simplify_add`) from `AffineExpr` to `AffineBinaryOpExpr`/`AffineBinaryOpKind`. The way it is done right now is actually quite generic and is also pretty similar to how MLIR does it as far as I can tell. Can you help me understand what the goal is? Should AffineExpr.binary not return anything but a `AffineBinaryOpExpr`?

superlopuh: I see, it is indeed very much like `getAffineBinaryOpExpr`, let's iterate on the code style then.

xdsl/ir/affine/affine_expr.py
```diff
@@ -37,6 +37,31 @@ def dimension(position: int) -> AffineExpr:
...
+        lhs: AffineExpr,
+        rhs: AffineExpr,
+    ) -> AffineExpr:
+        """
+        The difference between creating an `AffineBinaryOpExpr` with this function
+        and creating an `AffineBinaryOpExpr` using its constructor
+        is that simplifications are automatically applied in this function.
+        """
+
+        if kind == AffineBinaryOpKind.Add:
```
superlopuh, line 52: can you make this a match statement please?

xdsl/ir/affine/affine_expr.py
```diff
@@ -37,6 +37,31 @@ def dimension(position: int) -> AffineExpr:
...
+        elif kind == AffineBinaryOpKind.Mul:
+            return lhs * rhs
+        elif kind == AffineBinaryOpKind.Mod:
+            return lhs % rhs
+        elif kind == AffineBinaryOpKind.FloorDiv:
+            return lhs // rhs
+        elif kind == AffineBinaryOpKind.CeilDiv:
+            return lhs.ceil_div(rhs)
+
+        raise ValueError("Unreachable")
```
superlopuh, line 63:
```suggestion
        assert_never(kind)
```

tests/test_affine_builtins.py
```diff
@@ -212,8 +232,18 @@ def test_compress_dims():
...
+    six = AffineExpr.constant(6)
+
+    # Should return AffineConstExpr when both lhs and rhs are AffineConstantExpr
+    assert AffineExpr.binary(AffineBinaryOpKind.Add, one, one) == two
+    assert AffineExpr.binary(AffineBinaryOpKind.Mul, two, three) == six
+    assert AffineExpr.binary(AffineBinaryOpKind.Mod, five, two) == one
+    assert AffineExpr.binary(AffineBinaryOpKind.FloorDiv, five, two) == two
+    assert AffineExpr.binary(AffineBinaryOpKind.CeilDiv, five, two) == three
+
+    # TODO test other simplifications like dim + const + const = dim + const
```
superlopuh, line 249: no need for this TODO

superlopuh (approved):

xdsl/ir/affine/affine_expr.py
```diff
@@ -37,6 +39,32 @@ def dimension(position: int) -> AffineExpr:
...
+    def binary(
+        kind: AffineBinaryOpKind,
+        lhs: AffineExpr,
+        rhs: AffineExpr,
+    ) -> AffineExpr:
+        """
+        The difference between creating an `AffineBinaryOpExpr` with this function
+        and creating an `AffineBinaryOpExpr` using its constructor
+        is that simplifications are automatically applied in this function.
+        """
```
compor, line 52: Nit: can you please first state what this method does, and then elaborate on the differences with the aforementioned constructor?

The text you have already there for the latter is fine, but for the uninitiated it might be good mentioning when does this simplification take place (e.g., is it when calling `eval` for example?).

compor (approved):

PR #4009, opened 2025-03-03
core: check next char before whitespace regex in MLIRLexer

superlopuh:
`make filecheck` goes from 16s to 13s on my computer, I look forward to seeing how it impacts things on xdsl-bench.

### DIFF
--- a/xdsl/utils/mlir_lexer.py
+++ b/xdsl/utils/mlir_lexer.py
@@ -258,12 +258,17 @@ class MLIRLexer(Lexer[MLIRTokenKind]):
         return match
 
     _whitespace_regex = re.compile(r"((//[^\n]*(\n)?)|(\s+))*", re.ASCII)
+    _first_whitespace_chars = set((ord("\n"), ord(" "), ord("\t"), ord("\r"), ord("/")))
 
     def _consume_whitespace(self) -> None:
         """
         Consume whitespace and comments.
         """
-        self._consume_regex(self._whitespace_regex)
+        first = self.input.slice(self.pos, self.pos + 1)
+        if first is None:
+            return
+        if ord(first) in self._first_whitespace_chars:
+            self._consume_regex(self._whitespace_regex)
 
     def _form_token(self, kind: MLIRTokenKind, start_pos: Position) -> MLIRToken:
         """


xdsl/utils/mlir_lexer.py
```diff
@@ -258,12 +258,17 @@ def _consume_regex(self, regex: re.Pattern[str]) -> re.Match[str] | None:
...
 
     def _consume_whitespace(self) -> None:
         """
         Consume whitespace and comments.
         """
-        self._consume_regex(self._whitespace_regex)
+        first = self.input.slice(self.pos, self.pos + 1)
+        if first is None:
+            return
+        if ord(first) in self._first_whitespace_chars:
```
alexarice, line 270: Is the `ord` necessary? Does it make it significantly faster than directly checking character equality?

alexarice (commented):
I feel at this stage it might be simpler to hand write this lexer case (without a regex)

xdsl/utils/mlir_lexer.py
superlopuh: I tried make filecheck with char in string, char in set, ord in set, and ord in set won.

EdmundGoodman:
This looks good to me, but I think it is worth noting the benefit will depend on the workload it is running against. For example, for lexing 1000 additions it roughly doubles the runtime of the modified function in the miss case and roughly halves under the hit case under the tracer -- but this amounts to adding ~1000ns vs saving ~500ns, so the hit case needs to be fairly frequent.

Before (958ns):
image

After (hit case, 583ns):
image

After (miss case, 1us 708ns):
image

superlopuh:
It's worth a go, but I expect it to be much slower. I'll try speeding up the first char check.

alexarice:
Perhaps this is naive but have we checked if using https://pypi.org/project/regex/ speeds anything up? It claims to be fully api compatible so in theory is just a find replace

EdmundGoodman:
> Perhaps this is naive but have we checked if using https://pypi.org/project/regex/ speeds anything up? It claims to be fully api compatible so in theory is just a find replace

I think the prevailing opinion is it is more feature complete but less fast (will find a reference), but worth benchmarking at least

alexarice:
In this specific case I'm starting to think more strongly that a bunch of if statements might be optimal

alexarice:
> Also could look at something more niche like https://github.com/itsmeadarsh2008/flpc?

Could give this a go but my intuition would be that the FFI hit surely outweighs the performance gain for this workload of matching small patterns at the start of a string

superlopuh:
I propose to revisit this PR after we've merged or rejected #4010

superlopuh:
I'll close this in the meantime, it might not be as much of a bottleneck once we shift to bytes.

PR #4010, opened 2025-03-03
core: use bytes internally in Input and Lexer

superlopuh:
I'm not sure why it's faster but `make filecheck` runs for 13s on my computer instead of 16s.

### DIFF
--- a/docs/Toy/toy/frontend/lexer.py
+++ b/docs/Toy/toy/frontend/lexer.py
@@ -49,7 +49,6 @@ SINGLE_CHAR_TOKENS = {
     "/": ToyTokenKind.OPERATOR,
 }
 
-IDENTIFIER_CHARS = re.compile(r"[\w]|[\d]|_")
 SPECIAL_CHARS = set(",")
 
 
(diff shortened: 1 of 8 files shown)


superlopuh:
MLIR can have unicode and non-unicode, meaning raw bytes of data, so `str` is technically incorrect in my understanding. As long as we use the positions as returned by regex matching we should be OK as the regular expressions are unicode compliant, in my understanding.

alexarice:
My concern was more if `bytes` was enforcing ascii, but I did a small test and everything seems to be working with or without this PR

alexarice:
Actually there's a slight regression: when trying to parse the string
```
"test.op"() {value₁ = "αβγ₁"} : () -> ()
```
you get a `UnicodeDecodeError` with this change, whereas you got a parse error before.

superlopuh:
Fair enough, I'll look into this a bit more later this week. I expect that it's a matter of adding more tests and having better regular expressions that match MLIR more closely, as opposed to sticking to `str`, but we'll see.

alexarice:
I also cannot reproduce these timing results

superlopuh:
OK then I'll close this. Thanks for running these locally!

PR #4020, opened 2025-03-04
backend: add full init to RegisterType

superlopuh:
A checkpoint in what I've been slowly working towards, we need to have a way to explicitly create the register we need, and not just from a string.

Verifier PR to follow.


xdsl/backend/register_type.py
```diff
@@ -28,26 +28,32 @@ class RegisterType(ParametrizedAttribute, TypeAttribute, ABC):
...
         Returns the parameter list required to construct a register instance from the given spelling.
         """
-        index = cls.abi_index_by_name().get(spelling)
+        index = cls.abi_index_by_name().get(spelling.data)
         index_attr = NoneAttr() if index is None else IntAttr(index)
-        return index_attr, StringAttr(spelling)
+        return index_attr, spelling
+
+    @classmethod
+    def from_spelling(cls, spelling: StringAttr | str) -> Self:
```
compor, line 53: This name has been bothering me in the last few PRs, because in my mind spelling redirects to orthography.
Moreover, there is no mention in the RISC-V or Arm ISA specification manuals, so it is completely foreign.

Can't we call this simply and plainly `from_name`? I mean, we also have `register_name` at some point, and it won't affect verification even if I give an erroneous register name.

Lastly, for example, in the RISC-V case you have the register number and the register name (e.g., `x8` and `s0`/`fp`).
The index will cover the register number part, while for `xN` this is just a mnemonic name of that, and the `sN` is the corresponding ABI name.

In any case, my point is that these are just names, and we should just call them that.

compor: Hopefully, this is just a find/replace kind of thing.

compor (approved):

xdsl/backend/register_type.py
superlopuh: I completely agree, with the caveat that `name` is already taken. Let's discuss this on Zulip.

tests/backend/test_register.py
qaco, line 1: It means that, as before, the one and only difference between an allocated register and an unallocated one is that the former has a name/spelling?

tests/dialects/test_x86.py
```diff
@@ -5,23 +5,23 @@
 
 
 def test_unallocated_register():
-    unallocated = x86.register.GeneralRegisterType("")
+    unallocated = x86.register.GeneralRegisterType.from_spelling("")
```
qaco, line 8: It means that, as before, the one and only difference between an allocated register and an unallocated one is that the former has a name/spelling?

superlopuh: Unallocated is always "" and None for name and index, and allocated is always != "", and not None for name and index.

PR #4030, opened 2025-03-06
backend: (riscv) allow x0... and f0... register names

superlopuh:
RISC-V has at least two ways to indicate the standard integer and floating-point registers, with their ABI names and indexed (x0, x1, ..., f0, f1, ...). With this PR, we support both in the assembly format again.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -92,7 +92,7 @@ class RISCVRegisterType(RegisterType):
         raise NotImplementedError()
 
 
-RV32I_INDEX_BY_NAME = {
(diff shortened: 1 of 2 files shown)


xdsl/dialects/riscv.py
```diff
@@ -92,7 +92,7 @@ def a_register(cls, index: int) -> Self:
         raise NotImplementedError()
 
 
-RV32I_INDEX_BY_NAME = {
+_RV32I_NAMED_INDEX_BY_NAME = {
```
compor, line 95:
```suggestion
_RV32I_ABI_INDEX_BY_NAME = {
```
?
similarly below?

PR #4032, opened 2025-03-06
interactive: return structured tuple in get_pass_argument_names_and_types

### LINKED ISSUE #1845: interactive: minor type fix in get_pass_argument_names_and_types() in passes.py
`get_pass_argument_names_and_types() `function in passes.py takes a ModulePassT and outputs a string containing the names of the pass arguments and their types. For Literal's, we would like it to output the actual type instead of its element. e.g. for the attribute declaration literal:`literal : Literal["yes", "no", "maybe"] = "no"`, the function currently outputs `literal=no`, but we would like to instead output the string `literal=Literal["no"]`, as `no` is not a type, while `Literal["no"]` is.

superlopuh:
We currently return a string, which doesn't allow the caller to choose whether and how to display the options.
Plenty of work to do still when the interactive project picks back up, but I feel like it's a step in the right direction.

Fixes #1845

### DIFF
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -3,7 +3,15 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator
 from dataclasses import Field, dataclass, field
 from types import NoneType, UnionType
-from typing import Any, ClassVar, TypeVar, Union, get_args, get_origin
+from typing import (
+    Any,
+    ClassVar,
+    NamedTuple,
+    TypeVar,
+    Union,
+    get_args,
+    get_origin,
+)
 
 from xdsl.context import Context
 from xdsl.dialects import builtin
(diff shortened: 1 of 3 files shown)


dshaaban01 (approved):
Nice

xdsl/passes.py
```diff
@@ -151,23 +159,30 @@ def pipeline_pass_spec(self, *, include_default: bool = False) -> PipelinePassSp
         return PipelinePassSpec(self.name, args)
 
 
-# Git Issue: https://github.com/xdslproject/xdsl/issues/1845
-def get_pass_argument_names_and_types(arg: type[ModulePassT]) -> str:
+class PassArgOption(NamedTuple):
```
compor, line 162:
```suggestion
class PassOption(NamedTuple):
```

I know we use "arg", but the "LLVM lingo" is option (e.g., `cl::opt<>` from `CommandLine.h`).
Maybe let's start using it like that and we can rename the other parts later (e.g., `get_pass_option_info`).

xdsl/passes.py
```diff
@@ -151,23 +159,30 @@ def pipeline_pass_spec(self, *, include_default: bool = False) -> PipelinePassSp
...
+    """ """
```
compor, line 163:
```suggestion
    """Information on the options of a pass."""
```
or similar?

xdsl/passes.py
```diff
@@ -151,23 +159,30 @@ def pipeline_pass_spec(self, *, include_default: bool = False) -> PipelinePassSp
...
+
+    name: str
+    type_str: str
+    default_value_str: str | None = None
```
compor, line 167:
```suggestion
    value_type: str
    default_value: str | None = None
```
I know these are for object reconstruction, but `_str` doesn't add much info?

xdsl/passes.py
compor: Or even `PassOptionInfo`? I feel it leaves room to grow.

compor (approved):

xdsl/passes.py
superlopuh: I like PassOptionInfo more as it's more the information about what is expected, rather than what is passed.

xdsl/passes.py
superlopuh: You're right. I was initially hoping to unstringify it all the way, and pass structured information around, but it seemed like too big a change. I think it makes sense to give things nice names even if we might change the types later.

PR #4035, opened 2025-03-07
misc: (makefile) Run all marimo tests and report at the end

### LINKED ISSUE #2841: testing: marimo tests should run all tests and report all that fail, not stop at first one

compor:
This PR: 

- Executes all marimo tests and only reports failures at the end.

I encountered this while porting more than one notebook that was in a state of flux, which was a bit annoying.

Resolves #2841

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,9 @@ LIT_OPTIONS ?= -v --order=smart
 # make tasks run all commands in a single shell
 .ONESHELL:
 
+# use bash as the shell
+SHELL := /bin/bash
+
 .PHONY: uv-installed
 uv-installed:
 	@command -v uv &> /dev/null ||\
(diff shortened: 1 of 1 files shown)


compor:
Please give it a go locally if possible.

Makefile
```diff
@@ -19,6 +19,9 @@ LIT_OPTIONS ?= -v --order=smart
 # make tasks run all commands in a single shell
 .ONESHELL:
 
+# use bash as the shell
+SHELL := /bin/bash
```
compor, line 23: forces `make` to use `bash` instead of `sh`

superlopuh (approved):

superlopuh:
Fails locally on my mac: 
```
make: declare: No such file or directory
make: *** [tests-marimo] Error 1
```

compor:
`declare` is bash syntax, I'll reiterate to avoid it then; no access to mac

PR #4037, opened 2025-03-08
transformations: Make loop folding consistent with mlir

mamanain:
There was a difference in logic between this pass in mlir and xdsl and xdsl's version made it hard to flatten ptr loops. Made the logic consistent.


tests/filecheck/transforms/ptr_loop_folding.mlir
superlopuh, line 1: Let's not merge this end-to-end test

superlopuh (approved):
Nice!

PR #4008, opened 2025-03-03
[WIP] bench: development tracking branch for benchmarking

EdmundGoodman:
A draft PR to track the ongoing development of xDSL benchmarking. This branch is not intended to be merged, but instead to provide visibility. Benchmarks will be merged to the main branch in small increments in separate PRs.

PR #4041, opened 2025-03-10
bench: add utilities for profiling and tracing benchmarks

EdmundGoodman:
Add the `bench_utils` file to provide a simple command-line interface for developers to manually run and profile benchmarks outside of ASV.

### DIFF
--- a/benchmarks/bench_utils.py
+++ b/benchmarks/bench_utils.py
@@ -1,0 +1,250 @@
+#!/usr/bin/env python3
+"""Utilities for profiling ASV benchmarks with a variety of tools."""
+
+import cProfile
+import gc
+import subprocess
+import time
+from argparse import ArgumentParser, Namespace
+from collections.abc import Callable
+from pathlib import Path
+from statistics import mean, median, stdev
+from typing import Any, cast
+
+DEFAULT_OUTPUT_DIRECTORY = Path(__file__).parent / "profiles"
+PROFILERS = ("run", "timeit", "snakeviz", "viztracer", "flameprof")
+
+
+def warmed_timeit(
+    func: Callable[[], Any],
+    setup: Callable[[], Any] | None = None,
+    number: int = 100,
+    warmup: int = 3,
+) -> tuple[float, float, float]:
+    """Time the contents of a class method with setup and warmup.
+
+    Derived from <https://github.com/python/cpython/blob/3.13/Lib/timeit.py>.
+    """
+
+    class EmptyBenchmarkClass:
+        """A benchmark class for the empty function."""
+
+        def empty(self) -> None:
+            """An empty function call."""
+            pass
+
+    benchmark_class_empty = EmptyBenchmarkClass().empty
+
+    gcold = gc.isenabled()
+    gc.disable()
+
+    times = [0.0 for _ in range(number)]
+    offset = [0.0 for _ in range(number)]
+
+    for _ in range(warmup):
+        if setup is not None:
+            setup()
+        func()
+
+    for i in range(number):
+        if setup is not None:
+            setup()
+
+        offset_start = time.perf_counter()
+        benchmark_class_empty()
+        offset_end = time.perf_counter()
+        offset[i] = offset_end - offset_start
+
+        func_start = time.perf_counter()
+        func()
+        func_end = time.perf_counter()
+        times[i] = func_end - func_start
+
+    if gcold:
+        gc.enable()
+
+    return (
+        mean(times) - mean(offset),
+        median(times) - mean(offset),
+        stdev(times) + stdev(offset),
+    )
+
+
+def parse_arguments(benchmark_names: list[str]) -> ArgumentParser:
+    """Parse the arguments for the profiler tool."""
+    parser = ArgumentParser()
+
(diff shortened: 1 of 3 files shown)


benchmarks/bench_utils.py
```diff
@@ -0,0 +1,250 @@
...
+        if setup is not None:
+            setup()
+        func()
+
+    for i in range(number):
+        if setup is not None:
+            setup()
+
+        offset_start = time.perf_counter()
+        benchmark_class_empty()
```
superlopuh, line 54: Can you leave a comment somewhere in the code to say why this is useful?

benchmarks/bench_utils.py
```diff
@@ -0,0 +1,250 @@
...
+PROFILERS = ("run", "timeit", "snakeviz", "viztracer", "flameprof")
+
+
+def warmed_timeit(
+    func: Callable[[], Any],
+    setup: Callable[[], Any] | None = None,
+    number: int = 100,
+    warmup: int = 3,
+) -> tuple[float, float, float]:
+    """Time the contents of a class method with setup and warmup.
```
superlopuh, line 24:
```suggestion
    """
    Time the contents of a class method with setup and warmup.
```

benchmarks/bench_utils.py
superlopuh: Could be here, or on the class, or in both places

benchmarks/bench_utils.py
```diff
@@ -0,0 +1,250 @@
...
+            """An empty function call."""
+            pass
+
+    benchmark_class_empty = EmptyBenchmarkClass().empty
+
+    gcold = gc.isenabled()
+    gc.disable()
+
+    times = [0.0 for _ in range(number)]
+    offset = [0.0 for _ in range(number)]
```
superlopuh, line 42:
```suggestion
    times = [0.0] * number
    offset = [0.0] * number
```

benchmarks/bench_utils.py
```diff
@@ -0,0 +1,250 @@
...
+        times[i] = func_end - func_start
+
+    if gcold:
+        gc.enable()
+
+    return (
+        mean(times) - mean(offset),
+        median(times) - mean(offset),
+        stdev(times) + stdev(offset),
+    )
```
superlopuh, line 70: definitely worth a comment

benchmarks/bench_utils.py
```diff
@@ -0,0 +1,250 @@
...
+        "-q", "--quiet", action="store_true", help="don't show the profiler's UI"
+    )
+
+    return parser
+
+
+def get_benchmark_runs(
+    args: Namespace,
+    benchmarks: dict[
+        str, Callable[[], Any] | tuple[Callable[[], Any], Callable[[], Any]]
```
superlopuh, line 106: Worth a typealias?

superlopuh (approved):
Very cool

superlopuh:
Small note: our convention is to let the original commenter resolve the comments, makes it easier to track whether it's been addressed in the way that was expected.

benchmarks/bench_utils.py
EdmundGoodman: Added type alias -- this also makes the invocation in benchmark files neater which is nice!

benchmarks/bench_utils.py
```diff
@@ -0,0 +1,253 @@
...
+import gc
+import subprocess
+import time
+from argparse import ArgumentParser, Namespace
+from collections.abc import Callable
+from pathlib import Path
+from statistics import mean, median, stdev
+from typing import Any, NamedTuple, cast
+
+DEFAULT_OUTPUT_DIRECTORY = Path(__file__).parent / "profiles"
```
compor, line 14: Isn't more natural to use the current working directory (i.e., `os.get_cwd()`)? it might be convenient for the CI runs ATM though.

compor (approved):

benchmarks/bench_utils.py
EdmundGoodman: I did this because it avoids making multiple `profiles` directories if you run the script in different places, to avoid making lots of big directories and ensuring they are correctly gitignored.

If it is confusing behaviour, am happy to change though.

compor: The gitignore can be a regex, so I don't see an issue there. As for the size and multiple directories, I usually have full control over the generated files (e.g., no overwrites if they can be avoided, etc.).

This is not an issue, let's get this in, and I'll play with it more in due time.

benchmarks/bench_utils.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def parse_arguments(benchmark_names: list[str]) -> ArgumentParser:
+    """Parse the arguments for the profiler tool."""
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        "test",
+        choices=benchmark_names + ["all"],
+        help="the name of the benchmark to run",
```
math-fehr, line 98:
```suggestion
        help="the name of the benchmark to run, `all` to run all benchmarks",
```

math-fehr (approved):
That's amazing, thanks!

PR #4042, opened 2025-03-10
backend: add abi_name_by_index helper to RegisterType

superlopuh:
This will let us create registers from the index directly.


tests/dialects/test_riscv.py
```diff
@@ -280,3 +280,77 @@ def test_get_constant_value():
     zero_op = riscv.GetRegisterOp(riscv.Registers.ZERO)
     zero_val = get_constant_value(zero_op.res)
     assert zero_val == IntegerAttr.from_int_and_width(0, 32)
+
+
+def test_int_name_by_index():
```
compor, line 285:
```suggestion
def test_int_abi_name_by_index():
```
?
Pleased ignore if there is no plan to add other name mappings (e.g., the register name, etc.).

In general, shouldn't the "abi" qualify all uses where that is the name expected?
I just find it a bit confusing because we also have ` abi_index_by_name()` right?
Shouldn't that be `index_by_abi_name()` instead?

If so, maybe something for another PR? Not sure how you feel about it.

compor (approved):

tests/dialects/test_riscv.py
superlopuh: `index_by_abi_name` would be a bit misleading, because we want both `x8` and `fp` to map to 8 there, so not just the ABI name but any other valid name. I've renamed it to just `index_by_name` in this PR, can revert or rename later.

compor: well, in any case "abi" is just confusing then.

superlopuh: I agree but I don't have a nicer name for it... Suggestions welcome!

PR #4044, opened 2025-03-11
backend: add from_index method to register

superlopuh:
The last PR I have planned for the register types themselves, after this onto the register queue, using this indices to keep track of which registers are reserved or available instead of names.

### DIFF
--- a/tests/backend/test_register.py
+++ b/tests/backend/test_register.py
@@ -112,3 +112,10 @@ def test_invalid_index():
 
 def test_name_by_index():
     assert TestRegister.abi_name_by_index() == {0: "a0", 1: "a1"}
+
+
+def test_from_index():
+    assert TestRegister.from_index(0) == TestRegister.from_name("a0")
+    assert TestRegister.from_index(1) == TestRegister.from_name("a1")
+    assert TestRegister.from_index(-1) == TestRegister.infinite_register(0)
+    assert TestRegister.from_index(-2) == TestRegister.infinite_register(1)
--- a/tests/dialects/test_riscv.py
+++ b/tests/dialects/test_riscv.py
@@ -319,6 +319,20 @@ def test_int_abi_name_by_index():
     }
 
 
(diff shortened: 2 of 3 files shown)


tests/backend/test_register.py
```diff
@@ -112,3 +112,10 @@ def test_invalid_index():
 
 def test_name_by_index():
     assert TestRegister.abi_name_by_index() == {0: "a0", 1: "a1"}
+
+
+def test_from_index():
+    assert TestRegister.from_index(0) == TestRegister.from_name("a0")
+    assert TestRegister.from_index(1) == TestRegister.from_name("a1")
```
compor, line 119: Would it make sense to also add `TestRegister.from_name("x0")` etc. here?

compor (approved):

tests/backend/test_register.py
superlopuh: Ah but these cannot be created from an index alone

compor: not sure I understand.

superlopuh: Like there is no value `x` for which `TestRegister.from_index(x)` is equal to `TestRegister.from_name("x0")`

compor: I meant smt like

```py
assert TestRegister.from_index(0) == TestRegister.from_name("a0") == TestRegister.from_name("x0")
```

superlopuh: Ah but those are different registers! `TestRegister.from_name("a0") == TestRegister(0, "a0") != TestRegister(0, "x0") == TestRegister.from_name("x0")`

compor: Yeah! Sorry, I meant `x10` in this case! Maybe I should lie down ;-]

superlopuh: I feel like this is somewhat important to discuss at some point just so we're on the same page, maybe another day :)

PR #4045, opened 2025-03-11
CI: fix tests-marimo on mac

superlopuh:
The changes from  #4035 didn't work on my computer, not sure if it's because bash wasn't picked up properly, or a difference in `make`, but this seems to work.

I've also removed the stdout output again from Python.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,22 +91,25 @@ tests-toy: filecheck-toy pytest-toy pytest-toy-nb
 
 .PHONY: tests-marimo
 tests-marimo: uv-installed
-	@ERROR_LOG=$(shell mktemp)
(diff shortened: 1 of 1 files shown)


math-fehr (approved):
I'm not that knowledgeable about that, but if that works that works.
Do we need to do that for other make targets?

superlopuh:
This is the only one that did fancy temp file things

compor (approved):
It works locally, not sure what is off with mac bash support (seems like explicit array variable support to me, which is unfortunate).

nit: if you create a random temp file using the PID, it is usually not necessary to `rm` explicitly, especially if it lives in `/tmp/`

PR #4053, opened 2025-03-13
dialects (arm): Add mixed vector/scalar fmul op

emmau678:
Note: Stacked PR.

Add fmul op for NEON vector registers


xdsl/dialects/arm_neon.py
```diff
@@ -62,9 +74,124 @@ def infinite_register_prefix(cls):
...
+class ARMNEONInstruction(ARMInstruction, ABC):
+    """
+    Base class for operations that can be a part of x86 assembly printing. Must
+    represent an instruction in the x86 instruction set.
+    The name of the operation will be used as the x86 assembly instruction name.
+    """
+
+    arrangement = attr_def(StringAttr)
+    """
+    An optional comment that will be printed along with the instruction.
```
superlopuh, line 87: ?

xdsl/dialects/arm_neon.py
```diff
@@ -62,9 +74,124 @@ def infinite_register_prefix(cls):
 V30 = NEONRegisterType.from_name("v30")
 V31 = NEONRegisterType.from_name("v31")
 
+
+class ARMNEONInstruction(ARMInstruction, ABC):
+    """
+    Base class for operations that can be a part of x86 assembly printing. Must
```
superlopuh, line 80: ?

xdsl/dialects/arm_neon.py
```diff
@@ -62,9 +74,124 @@ def infinite_register_prefix(cls):
...
+      - "4S"  → 4 single-precision floats
+      - "2D"  → 2 double-precision floats
+    https://developer.arm.com/documentation/ddi0602/2024-12/SIMD-FP-Instructions/FMUL--vector---Floating-point-multiply--vector--?lang=en#T_option__4
+    """
+
+    name = "arm_neon.dss.fmulvec"
+    d = result_def(NEONRegisterType)
+    s1 = operand_def(NEONRegisterType)
+    s2 = operand_def(NEONRegisterType)
+    arrangement = attr_def(StringAttr)
```
superlopuh, line 141: seems like a redefinition

xdsl/dialects/arm_neon.py
```diff
@@ -62,9 +74,124 @@ def infinite_register_prefix(cls):
...
+        self,
+        s1: Operation | SSAValue,
+        s2: Operation | SSAValue,
+        *,
+        d: NEONRegisterType,
+        arrangement: str | StringAttr,
+        comment: str | StringAttr | None = None,
+    ):
+        if isinstance(arrangement, str):
+            valid_arrangements = {"4H", "8H", "2S", "4S", "2D"}
```
superlopuh, line 157: I thought the bitwidth specifiers were per register?

xdsl/dialects/arm_neon.py
```diff
@@ -62,9 +74,124 @@ def infinite_register_prefix(cls):
...
+        arg_str = ", ".join(
+            f"{assembly_arg_str(arg)}.{self.arrangement.data}"
+            for arg in self.assembly_line_args()
+            if arg is not None
+        )
+        return AssemblyPrinter.assembly_line(instruction_name, arg_str, self.comment)
+
+
+@irdl_op_definition
+class GetRegisterOp(ARMOperation):
```
superlopuh, line 105: why add this?

emmau678: because I have the neon register type in arm_neon (so I need an arm_neon.get_register op for the filecheck)

xdsl/dialects/arm_neon.py
emmau678: I'm not sure if I understand this correctly - they are specified with the register yes, but the same register can be used with different arrangement specifiers, as I understand

qaco: If I understand well, the arrangement specifier describes the registers, but is carried by the instruction (whereas in x86, the arrangement specifier is essentially carried by the registers' names) ?

emmau678: yes that’s my understanding too

xdsl/dialects/arm_neon.py
superlopuh: "test.op" should do the trick for the time being, no?

superlopuh: Ah it's for assembly printing

xdsl/dialects/arm_neon.py
superlopuh: What about `SADDL2 V0.2D, V1.4S, V2.4S` from page 102 of https://cs140e.sergio.bz/docs/ARMv8-A-Programmer-Guide.pdf ?

superlopuh: I now think it's worth not overthinking it, we can fix things later. It would be great to add more documentation around the place to explain the design for future readers.

qaco: Yeah I think it's fine. What we basically need is (to generate) runnable code implementing tiled matrix multiplications. We can refine it incrementally.

qaco (approved):

emmau678:
I'm now thinking this actually does not fully reflect what I need for my asm implementation yet. Instead of the vector/vector instruction, it should be the vector/scalar version so I will try to update this

xdsl/dialects/arm_neon.py
```diff
@@ -1,11 +1,27 @@
...
+    base,
+    irdl_attr_definition,
+    irdl_op_definition,
+    operand_def,
+    result_def,
+)
 
 ARM_NEON_INDEX_BY_NAME = {f"v{i}": i for i in range(0, 32)}
 
+Imm8Attr = IntegerAttr[Annotated[IntegerType, i8]]
```
superlopuh, line 23: I'm not sure we want this typealias, might as well just to `IntegerAttr[I8]` in the operation definition

xdsl/dialects/arm_neon.py
```diff
@@ -62,9 +78,125 @@ def infinite_register_prefix(cls):
...
+
+    The arrangement specifier for NEON instructions determines element size and count:
+      - "4H"  → 4 half-precision floats
+      - "8H"  → 8 half-precision floats
+      - "2S"  → 2 single-precision floats
+      - "4S"  → 4 single-precision floats
+      - "2D"  → 2 double-precision floats
+    """
+
+    arrangement = attr_def(StringAttr)
```
superlopuh, line 91: This should be an `EnumAttribute`, maybe we can do this as a first PR?

superlopuh: Or rather before this one, and after the register type

xdsl/dialects/arm_neon.py
```diff
@@ -75,30 +82,63 @@ def infinite_register_prefix(cls):
 V31 = NEONRegisterType.from_name("v31")
 
 
+class NeonArrangement(StrEnum):
```
qaco, line 85: Nice !

tests/dialects/arm/test_assembly_arg_str.py
```diff
@@ -1,5 +1,18 @@
...
 from xdsl.dialects.arm.register import IntRegisterType
+from xdsl.dialects.arm_neon import NeonArrangement, NeonArrangementAttr
+from xdsl.printer import Printer
 
 
 def test_assembly_arg_str_ARMRegister():
     assert IntRegisterType.from_name("x0").assembly_str() == "x0"
+
+
+def test_gello():
```
superlopuh, line 12: I'm confused, isn't this from another PR?

PR #4054, opened 2025-03-13
fix: handle-async-flow to work with no `iter_args`

dk949:
Most of `csl-stencil-handle-async-flow` works by iterating thought `iter_args`, so it still works if there are none.

A guard is needed for one place where `iter_args` is indexed directly.

@n-io I've compiled this all the way to CSL, and it seems to produce sensible results.

This is needed to put a loop around the `uvbke` benchmark.


tests/filecheck/transforms/csl-stencil-handle-async-flow-no-args.mlir
```diff
@@ -0,0 +1,199 @@
...
+  ^0(%arg35 : i16, %arg36 : i16, %arg37 : i16, %arg38 : i16, %arg39 : i16, %arg40 : i16, %arg41 : i16, %arg42 : i16, %arg43 : i16):
+    %0 = arith.constant 1 : i16
+    %1 = arith.constant 0 : i16
+    %2 = "csl.get_color"(%1) : (i16) -> !csl.color
+    %3 = "csl_wrapper.import"(%arg37, %arg38, %2) <{fields = ["width", "height", "LAUNCH"], module = "<memcpy/get_params>"}> : (i16, i16, !csl.color) -> !csl.imported_module
+    %4 = "csl_wrapper.import"(%arg40, %arg37, %arg38) <{fields = ["pattern", "peWidth", "peHeight"], module = "routes.csl"}> : (i16, i16, i16) -> !csl.imported_module
+    %5 = "csl.member_call"(%4, %arg35, %arg36, %arg37, %arg38, %arg40) <{field = "computeAllRoutes"}> : (!csl.imported_module, i16, i16, i16, i16, i16) -> !csl.comptime_struct
+    %6 = "csl.member_call"(%3, %arg35) <{field = "get_params"}> : (!csl.imported_module, i16) -> !csl.comptime_struct
```
superlopuh, line 12: Is there any way to reduce this test case to detect exactly your change? I would recommend trying to shrink it with shrinkray by catching the out of bounds error and re-raising a ShrinkException

superlopuh: That's the lazy way to do it, anyway, I'm sure it's also reducible by hand.

n-io:
> This is needed to put a loop around the `uvbke` benchmark.

The version I have for `uvbke` did not run in a loop, did you add it or are you looking at a different version?

dk949:
I've added one for benchmarking purposes.

n-io:
Sure, having `iter_args` on the benchmark would prob make sense as well, though I'm not opposed to merging this.

n-io (approved):
Happy to approve, though test case could possibly be reduced.

n-io (approved):

n-io:
You could prob delete most of the layout module as well

superlopuh:
Would you like help shrinking the test case further?

dk949:
> Would you like help shrinking the test case further?

If you think it should be shrunk further, that would be very helpful

superlopuh:
Let's do a call some time this week to do this live together? I want to make the shrinking flow as easy as possible for users/contributors, this is a good opportunity to test it together, and see how the documentation could be improved.

superlopuh:
This is the current doc: https://xdsl.readthedocs.io/en/latest/reference/utils/exceptions/#xdsl.utils.exceptions.ShrinkException

n-io:
Have a look at the first test case in https://github.com/xdslproject/xdsl/blob/main/tests/filecheck/transforms/lower-csl-wrapper.mlir for a minimal wrapper_module op

tests/filecheck/transforms/csl-stencil-handle-async-flow-no-args.mlir
```diff
@@ -0,0 +1,118 @@
...
+    %10 = arith.cmpi slt, %arg35, %7 : i16
+    %11 = arith.cmpi slt, %arg36, %7 : i16
+    %12 = arith.cmpi slt, %8, %arg40 : i16
+    %13 = arith.cmpi slt, %9, %arg40 : i16
+    %14 = arith.ori %10, %11 : i1
+    %15 = arith.ori %14, %12 : i1
+    %16 = arith.ori %15, %13 : i1
+    "csl_wrapper.yield"(%6, %5, %16) <{fields = ["memcpy_params", "stencil_comms_params", "isBorderRegionPE"]}> : (!csl.comptime_struct, !csl.comptime_struct, i1) -> ()
+  }, {
+  ^1(%arg0 : i16, %arg1 : i16, %arg2 : i16, %arg3 : i16, %arg4 : i16, %arg5 : i16, %arg6 : i16, %arg7 : !csl.comptime_struct, %arg8 : !csl.comptime_struct, %arg9 : i1):
```
n-io, line 25:
```suggestion
  "csl_wrapper.module"() <{height = 4 : i16, params = [], program_name = "loop_kernel", width = 4 : i16}> ({
  ^0(%arg35 : i16, %arg36 : i16, %arg37 : i16, %arg38 : i16):
    %memparams = "test.op"() : () -> !csl.comptime_struct
    "csl_wrapper.yield"(%memparams) <{"fields" = ["memcpy_params"]}> : (!csl.comptime_struct) -> ()
  }, {
  ^1(%arg0 : i16, %arg1 : i16, %arg8 : !csl.comptime_struct):
```

This *might work (though would need updating the tests)

superlopuh (approved):
Thank you!

PR #4056, opened 2025-03-14
dependencies: use distributed version of llvm

superlopuh:


AntonLydike (approved):
:scream:

compor (approved):

math-fehr (approved):
Very cool!

JosseVanDelm (approved):
Looks good!
One question: Why not track the Dockerfile in this repository?

superlopuh:
I wasn't sure what the right way to do it was, we can always move it, but I wanted to play with it first before moving it here. Having it in the other repo lets us have different version tags (I imagine that's not impossible to do in this repo), and lets us update the docker before updating the CI script here (I imagine that's also not impossible to do in this repo, by publishing manually instead of a CI action).

PR #4060, opened 2025-03-14
backend: (riscv) RiscvRegisterQueue makes empty queue, add helper

superlopuh:
As a step towards a generic register queue, I wanted to decouple the init from the ABI default available and reserved registers.


compor (approved):
LGTM, in my mind these live in a separate dialect, but this is a good interim step.

PR #4061, opened 2025-03-14
backend: (arm) make AssemblyInstructionArg an ABC

superlopuh:
Looking at https://github.com/xdslproject/xdsl/pull/4053/files#diff-c827fb3cdc1ccc5298a4bb1facd25b57df2e97091cc64e6961f6a05e3f147151R180-R187, it feels like our current API for printing assembly doesn't scale to the complexity of NEON registers. What do you think of changing things around a bit, and instead of having assembly line arguments being a union to be switched over adding an abstract base class to call into? That way when things get weird we can just make a new helper class to format things in the way we want, whether it's `V0.4S` or `V0.S[2]`?


compor (approved):
I think this is the right direction for this.

I was hoping to be able to delegate even more to each assembly component (e.g., even the instruction itself) using abstract interfaces, but this wil have to do for now.

superlopuh:
Yep eventually that would make sense, especially if we want to be able to swap out the format of the assembly syntax, but we can slowly move in that direction.

PR #4062, opened 2025-03-14
backend: (riscv) use register indices to track available registers

superlopuh:
One step closer to generic register queue.


qaco:
Why is it better to use register indices than actual Python objects ?

compor (commented):
post-merge review: loooks great, thanks!

PR #4066, opened 2025-03-17
bench: add lexing benchmarks and their workloads

EdmundGoodman:
Add benchmarks for the lexer. The main complexity is the MLIR workload files to lex, which can then be re-used, for example by the parser/pattern rewriter/printer. Code is added to generate arbitrary constant folding workloads, and real-world use files are included from https://github.com/xdslproject/xdsl-bench/pull/1.


benchmarks/workloads.py
```diff
@@ -0,0 +1,61 @@
...
+        random.seed(0)
+        ops: list[str] = []
+        ops.append(
+            '%0 = "arith.constant"() {"value" = '
+            f"{random.randint(1, 1000)} : i32}} : () -> i32"
+        )
+        for i in range(1, size + 1):
+            if i % 2 == 0:
+                ops.append(
+                    f'%{i} = "arith.addi"(%{i - 1}, %{i - 2}) : (i32, i32) -> i32'
```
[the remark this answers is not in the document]
EdmundGoodman: This is the input to the lexer (and parser), so needs to be a string -- but this is a good point for upcoming rewriting/printing benchmarks, so I will add similar functionality later to build IR instead as well.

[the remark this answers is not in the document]
superlopuh: My thought was that we could just print and re-parse, but I don't feel strongly about this. I am pretty sure that we don't want to commit this, though.

benchmarks/workloads.py
```diff
@@ -0,0 +1,77 @@
...
+        ]
+        return WorkloadBuilder.wrap_module(ops)
+
+    @classmethod
+    def large_dense_attr_hex(cls, x: int = 1024, y: int = 1024) -> str:
+        """Get the MLIR hex representation of a large dense attr."""
+        assert x >= 0
+        assert y >= 0
+        random.seed(RANDOM_SEED)
+        dense_attr_hex = "".join([random.choice(HEX_CHARS) for _ in range(x * y * 2)])
```
superlopuh, line 69: why times two?

superlopuh (approved):
Looks great!

benchmarks/workloads.py
superlopuh, line 69:
```suggestion
        dense_attr_hex = "".join(random.choice(HEX_CHARS) for _ in range(x * y * 2))
```

EdmundGoodman: Two hex chars per byte, with each value in the matrix being one byte [-128,127]

superlopuh: Worth a comment

compor:
post-merge review: 👍🏽 thank you

PR #4070, opened 2025-03-18
misc: (eqsat) eqsat-serialize-egraph pass

jumerckx:
This can be used to dump a json representation of a piece of IR using the eqsat dialect. The json can then be used to visualize the e-graph here https://egraphs-good.github.io/egraph-visualizer/

Serialisation is currently implemented as a pass but should perhaps instead be a target? Currently, running the pass dumps json as well as the original IR, meaning that you can't just pipe the output somewhere directly.


xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,69 @@
...
+                            "op": name,
+                            "eclass": f"eclass_{i_eclass}",
+                            "children": children,
+                        }
+                    )
+                    i_enode += 1
+                i_eclass += 1
+        print(
+            json.dumps({"nodes": {f"enode_{i}": node for i, node in enumerate(nodes)}})
+        )
```
superlopuh, line 60: Let's decouple the dictionary construction from json printing? It would be good to have two functions

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,69 @@
+import json
+
+from xdsl.context import Context
+from xdsl.dialects import builtin, eqsat, func
+from xdsl.ir import BlockArgument, Operation
+from xdsl.passes import ModulePass
+
+
+def serialize_func_to_egraph(f_op: func.FuncOp):
```
superlopuh, line 9: Could you walk the whole IR backwards, instead of just per function? Should give the same results, and be more generic.

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,69 @@
...
+    body = body.first_block
+
+    i_eclass = 0
+    i_enode = 0
+    eclass_to_id: dict[eqsat.EClassOp, int] = {}
+    op_to_id = {}
+    nodes: list[dict[str, str | list[str]]] = []
+
+    if body:
+        for op in body.ops:
```
superlopuh, line 24: `for op in module.walk(reverse=True)`

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,69 @@
...
+            if isinstance(op, eqsat.EClassOp):
+                eclass_to_id[op] = (
+                    i_enode  # pick the first e-node as the representative
+                )
+                assert len(op.operands) > 0, "EClassOp must have operands"
```
superlopuh, line 29:
```suggestion
                assert op.operands, "EClassOp must have operands"
```
In Python, checking that the length is not 0 is an antipattern, as an empty check is assumed to be constant time, but length calculations could be O(n)

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,69 @@
...
+                for operand in op.operands:
+                    node = operand.owner
+                    children: list[str] = []
+                    if isinstance(node, Operation):
+                        op_to_id[node] = i_enode
+                        if len(node.operands) > 0:
```
superlopuh, line 35:
```suggestion
                        if node.operands:
```

superlopuh (commented):
Can you please add some filecheck tests?

superlopuh:
I'm still not sure about pass vs target, it feels like a pass could be useful to insert multiple over time to do pretty animations of the saturation. I would just add an optional filename argument to print to that instead of stdout.

superlopuh:
We use a convention where the original commenter resolves the comments on PRs, it makes it easier to track the conversations, and to make sure that they have been addressed as intended.

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,69 @@
...
+    op_to_id = {}
+    nodes: list[dict[str, str | list[str]]] = []
+
+    if body:
+        for op in body.ops:
+            if isinstance(op, eqsat.EClassOp):
+                eclass_to_id[op] = (
+                    i_enode  # pick the first e-node as the representative
+                )
+                assert op.operands, "EClassOp must have operands"
```
compor, line 29: Just a quick question: isn't this enforced by the verifier during loading/IR construction?

compor (commented):
I don't mean to bog down WIP/experimental stuff, but technically this is not a transform.

Shouldn't this "live" under `xdsl/tools/`, especially if you're planning to have an import capability at some point (I haven't followed closely the discussions)?

superlopuh:
IDK MLIR has a bunch of passes that print debug things, I think it's ok for us to add some also, arguably our regalloc things should have been similar.

compor:
Well, regalloc mutates the IR.

If this is for debugging purposes, then sure it is convenient to have it hooked up to the pipeline. I was just not sure if this is a one-off thing, and then you go on with your pipeline in another tool (there were discussions on Zulip on this as an exchange format)

superlopuh:
The way I see it this would be a cool debugging tool, to insert a number of these passes in the pipeline to see how the egraph evolves over time, as we apply different passes, say every 100 pattern applications during saturation.

compor:
> The way I see it this would be a cool debugging tool, to insert a number of these passes in the pipeline to see how the egraph evolves over time, as we apply different passes, say every 100 pattern applications during saturation.

No worries. I get it, I've done something similar in the past for dependence and dominance graphs, in order to be able to import them in NetworkX to test isomorphism, etc.

BTW, no need to for me to approve this, my questions were mostly curiosity-driven.

xdsl/transforms/eqsat_serialize_egraph.py
jumerckx: You mean, do the walk over the module and handle every FuncOp like I did here, or something else?
(also, why reverse?)

jumerckx:
I personally don't have a strong opinion on transform vs tool, although philosophically, a tool makes more sense to me.

> The way I see it this would be a cool debugging tool, to insert a number of these passes in the pipeline to see how the egraph evolves over time, as we apply different passes, say every 100 pattern applications during saturation.

What would potential usage of this look like? Maybe this can be exposed as an option for the equality saturation, i.e. something like
```
xdsl-opt input.mlir -p apply-eqsat{serialize-every=100}
```
(syntax might be wrong)
In that case, serialization doesn't need to be a proper pass?

xdsl/transforms/eqsat_serialize_egraph.py
jumerckx: The reason I only considered funcs is that I don't think e.g. scf.for bodies should be introspected and serialized.

superlopuh: Backwards because then you only need to process each operation once, and I don't think that this is the right place to put logic to restrict what is introspected, we might as well blindly trust the use-def structure.

superlopuh:
That could also work, but I wouldn't put too much thought into this now, we can always refactor later. For writing a first test the version in this PR is the way to go IMO.

jumerckx:
Converted to draft as I first want to add an `eqsat.egraph` op that makes it easier to do this serialization.

jumerckx:
Opened back up as `eqsat.egraph` will not be used all the time anyway. I've reworked the code to run over all operations as suggested. This now also handles cycles.

(Enjoy your holidays, Sasha, no hurry on a review!)

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,75 @@
...
+                firstoperand = operand.owner.operands[0]
+                if isinstance(firstoperand.owner, Operation):
+                    canonical_enode = firstoperand.owner
+                else:
+                    canonical_enode = cast(BlockArgument, firstoperand)
+                children.append(enode_to_id[canonical_enode])
+        if len(op.operands) > 0:
+            name = op.name
+        else:
+            name = str(op).split("=")[1].split(":")[0].strip()
```
superlopuh, line 62: Neat, but worth a comment
```suggestion
        if op.operands:
            name = op.name
        else:
            name = str(op).split("=")[1].split(":")[0].strip()
```

jumerckx: Upon further thought, the string representation of an operation doesn't always include a colon followed by the type. Perhaps we can just keep the type in the name as well?

xdsl/transforms/eqsat_serialize_egraph.py
```diff
@@ -0,0 +1,75 @@
...
+from collections import defaultdict
+from typing import Any, cast
+
+from xdsl.context import Context
+from xdsl.dialects import builtin, eqsat
+from xdsl.ir import BlockArgument, Operation
+from xdsl.passes import ModulePass
+
+
+class IDGenerator:
```
superlopuh, line 11:
```suggestion
class _IDGenerator:
	"""Helper to generate incrementing identifiers with a custom prefix."""
```

jumerckx:
Thanks! Comments have been addressed.

superlopuh (approved):

superlopuh:
Very nice

PR #4071, opened 2025-03-18
CI: add Dockerfile and CI action for preinstalled LLVM

superlopuh:
The main idea here is to provide a base image that would have the necessary LLVM and MLIR binaries to test the interop. 

The toolchain action is lifted from the Quidditch with small modifications.

Having an image like this that we can control would also let us have a base image for GH Codespaces with MLIR preinstalled.

Do you all think that this is a good idea? Is there a better way to set this up?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu:latest
+
+# Install xz-utils
+RUN apt-get update && apt-get install -y \
+    wget xz-utils \
+    && rm -rf /var/lib/apt/lists/*
+
+# Download and extract the official LLVM 19.1.7 binary
+WORKDIR /tmp
+RUN wget https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.7/LLVM-19.1.7-Linux-X64.tar.xz && \
+    tar xf LLVM-19.1.7-Linux-X64.tar.xz -C /usr/local --strip-components=1 && \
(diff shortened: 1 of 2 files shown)


Dockerfile
superlopuh, line 1: I have tested that this works locally with

```sh
docker build -t my-ubuntu-image . --platform linux/amd64
docker run -it my-ubuntu-image
```

superlopuh:
Oh another question is whether this should live in a different repository

superlopuh:
I'm not going to fix the CI before positive feedback as I'm relatively sure it'll actually publish the image once fixed.

Dockerfile
compor: I'll give it a go locally once I find some time.

Another thing that might be useful is https://github.com/nektos/act I've used in the past with moderate success (not sure about macos support)

compor:
Would you mind adding the goals as discussed from the last xDSL meeting for some more context especially for people not being there? (i.e., what does "that we can control" mean exactly?)

compor:
> Oh another question is whether this should live in a different repository

Looks fine in the same repo to me for now

superlopuh:
While everyone replies to this PR, I'll actually publish the same docker image in opencompl, just to test that it can actually work in our CI

superlopuh:
Put this in a different repo: https://github.com/xdslproject/llvm-docker/, with image name `xdslproject/llvm`

PR #4074, opened 2025-03-19
dependencies: make pre-commit version less strict

superlopuh:
Inspired by #4072, I don't think we need to be quite so precise on the pre-commit version.

We might want to revisit the other dependencies.


EdmundGoodman (approved):

compor:
post-merge review: 👍🏽

PR #4078, opened 2025-03-19
core: better spans for assembly format error messages

superlopuh:
I noticed that we were giving spans of the token after the variable name, this change includes both the $ token and the name.


alexarice (approved):
Seems reasonable, I'm not a huge fan of manually specifying spans for error messages but it seems necessary in this case.

compor (commented):
yucky

superlopuh:
Can you expand, please?

PR #4079, opened 2025-03-19
bench: pattern rewriting microbenchmarks

EdmundGoodman:
Add microbenchmarks for pattern rewriting. Using `pyinstrument`, we can get a breakdown of the worst offending functions (shown below, with sections collapsed to highlight the interesting bits)

image

From this, we write a number of microbenchmarks to exercise these bottlenecking functions for pattern rewriting in general and the `RemoveUnusedOperations`, `CanonicalizationRewritePattern`, and `region_dce` functions of constant folding.


benchmarks/rewriting.py
```diff
@@ -0,0 +1,401 @@
...
+        """Time `Operation.get_traits_of_type`."""
+        self.add_op.get_traits_of_type(MemoryEffect)
+
+
+class RegionDCE(RewritingMicrobenchmarks):
+    """Microbenchmarks for region dead-code elimination of constant folding."""
+
+    def time_post_order_iterator(self) -> None:
+        """Time `PostOrderIterator`."""
+        PostOrderIterator(self.first_block)
```
superlopuh, line 262: this doesn't do anything

benchmarks/rewriting.py
```diff
@@ -0,0 +1,401 @@
...
+        """Time `Worklist.pop`."""
+        self.worklist.pop()
+
+    def time_get_trait(self) -> None:
+        """Time `Operation.get_trait`."""
+        self.add_op.get_trait(HasCanonicalizationPatternsTrait)
+
+    def time_insert_point_before(self) -> None:
+        """Time `InsertPoint.before`."""
+        InsertPoint.before(self.add_op)
```
superlopuh, line 170: this doesn't do anything

benchmarks/rewriting.py
```diff
@@ -0,0 +1,401 @@
...
+    parser = XdslParser(context, contents)
+    return parser.parse_module()
+
+
+class ConstantFolding:
+    """Benchmark rewriting in xDSL."""
+
+    WORKLOAD_CONSTANT_20 = parse_module(CTX, WorkloadBuilder.constant_folding(20))
+    WORKLOAD_CONSTANT_100 = parse_module(CTX, WorkloadBuilder.constant_folding(100))
+    WORKLOAD_CONSTANT_1000 = parse_module(CTX, WorkloadBuilder.constant_folding(1_000))
```
compor, line 52: Does this have to be hardcoded here? Can it be dynamic (i.e., configured as such down in `main`?).
I have the feeling that I've asked about this before, but I might be wrong :)

benchmarks/rewriting.py
```diff
@@ -0,0 +1,401 @@
...
+    WORKLOAD_CONSTANT_20 = parse_module(CTX, WorkloadBuilder.constant_folding(20))
+    region: Region
+    const_0: ConstantOp
+    const_1: ConstantOp
+    add_op: AddiOp
+    sub_op: SubiOp
+    integer_type: IntegerType
+
+    def setup(self) -> None:
+        """Setup the benchmarks."""
```
compor, line 101: Would you mind adding the sketch for this IR is some sort of pseudocode?

compor (approved):
Does this give "self" time/percentage similar to `gprof` for example?

superlopuh:
Good to merge?

benchmarks/rewriting.py
EdmundGoodman: Add docstring explaining -- it times instantiating the `InsertPoint` rather than using it.

EdmundGoodman:
> Does this give "self" time/percentage similar to `gprof` for example?

`pyinstrument` does, as pictured below:
image

Other tools like `viztracer` don't appear to, and the custom timing logic is designed to measure and subtract this overhead.

benchmarks/rewriting.py
EdmundGoodman: Added to the `WorkloadBuilder` class.

compor (approved):

EdmundGoodman:
> Good to merge?

I think so? If you approve it, I'll press the button

benchmarks/rewriting.py
```diff
@@ -0,0 +1,491 @@
...
+        the traits iterator. This is used frequently in pattern rewriting, both
+        for matching and for the rewriting logic.
+        """
+        self.add_op.get_trait(HasCanonicalizationPatternsTrait)
+
+    def time_insert_point_before(self) -> None:
+        """Time `InsertPoint.before`.
+
+        Exercise creating an `InsertPoint` from the block's parent block. This
+        is used to move indicate the location to rewrite.
```
superlopuh, line 166:
```suggestion
        Measure time to create an `InsertPoint` from the block's parent block.
```

superlopuh (approved):

PR #4088, opened 2025-03-21
frontend: Add support for bigint dialect in PyAST frontend

EdmundGoodman:
Add support for bigint dialect in PyAST frontend

### DIFF
--- a/tests/frontend/pyast/test_bigint.py
+++ b/tests/frontend/pyast/test_bigint.py
@@ -1,0 +1,23 @@
+from textwrap import dedent
+
+from xdsl.frontend.pyast.context import CodeContext
+from xdsl.frontend.pyast.program import FrontendProgram
+
+
+def test_bigint_identity() -> None:
+    p = FrontendProgram()
+    with CodeContext(p):
+
+        def foo(x: int) -> int:  # pyright: ignore[reportUnusedFunction]
+            return x
+
+    p.compile()
+
+    expected_result = dedent("""
+    builtin.module {
+      func.func @foo(%0 : bigint.bigint) -> bigint.bigint {
+        func.return %0 : bigint.bigint
+      }
+    }
+    """).strip()
+    assert expected_result == p.textual_format()
--- a/tests/frontend/pyast/test_readme_examples.py
+++ b/tests/frontend/pyast/test_readme_examples.py
@@ -1,0 +1,26 @@
+from textwrap import dedent
+
+from xdsl.frontend.pyast.context import CodeContext
+from xdsl.frontend.pyast.dialects.builtin import i32
+from xdsl.frontend.pyast.program import FrontendProgram
+
+
+def test_readme_example() -> None:
+    p = FrontendProgram()
+    with CodeContext(p):
+
+        def foo(x: i32, y: i32, z: i32) -> i32:  # pyright: ignore[reportUnusedFunction]
+            return x + y * z
+
+    p.compile()
+
+    expected_result = dedent("""
+    builtin.module {
+      func.func @foo(%0 : i32, %1 : i32, %2 : i32) -> i32 {
+        %3 = arith.muli %1, %2 : i32
+        %4 = arith.addi %0, %3 : i32
+        func.return %4 : i32
+      }
+    }
+    """).strip()
+    assert expected_result == p.textual_format()
--- a/xdsl/dialects/bigint.py
+++ b/xdsl/dialects/bigint.py
@@ -1,0 +1,18 @@
+from xdsl.ir import Dialect, ParametrizedAttribute, TypeAttribute
+from xdsl.irdl import irdl_attr_definition
+
+
+@irdl_attr_definition
+class BigIntegerType(ParametrizedAttribute, TypeAttribute):
+    name = "bigint.bigint"
+
+
+bigint = BigIntegerType()
+
+BigInt = Dialect(
+    "bigint",
+    [],
+    [
+        BigIntegerType,
+    ],
+)
--- a/xdsl/dialects/bigint.py
+++ b/xdsl/dialects/bigint.py
@@ -1,9 +1,15 @@
+"""Dialect for arbitrary-precision integers."""
+
 from xdsl.ir import Dialect, ParametrizedAttribute, TypeAttribute
 from xdsl.irdl import irdl_attr_definition
 
 
 @irdl_attr_definition
 class BigIntegerType(ParametrizedAttribute, TypeAttribute):
+    """
+    Type for arbitrary-precision integers (bigints), such as those in Python.
+    """
+
     name = "bigint.bigint"
 
 
--- a/tests/filecheck/dialects/bigint/attrs.mlir
+++ b/tests/filecheck/dialects/bigint/attrs.mlir
@@ -1,12 +1,7 @@
 // RUN: XDSL_ROUNDTRIP
 
-"test.op"() {
-    bigint = !bigint.bigint,
-} : () -> ()
-
-%bigint = "test.op"() : () -> (!bigint.bigint)
+%bigint = "test.op"() : () -> !bigint.bigint
 
 // CHECK:       builtin.module {
-// CHECK-NEXT:    "test.op"() {bigint = !bigint.bigint} : () -> ()
-// CHECK-NEXT:    %bigint = "test.op"() : () -> (!bigint.bigint)
+// CHECK-NEXT:    %bigint = "test.op"() : () -> !bigint.bigint
 // CHECK-NEXT:  }
--- a/tests/filecheck/dialects/bigint/attrs.mlir
+++ b/tests/filecheck/dialects/bigint/attrs.mlir
@@ -1,0 +1,12 @@
+// RUN: XDSL_ROUNDTRIP
+
+"test.op"() {
+    bigint = !bigint.bigint,
+} : () -> ()
+
+%bigint = "test.op"() : () -> (!bigint.bigint)
+
+// CHECK:       builtin.module {
+// CHECK-NEXT:    "test.op"() {bigint = !bigint.bigint} : () -> ()
+// CHECK-NEXT:    %bigint = "test.op"() : () -> (!bigint.bigint)
+// CHECK-NEXT:  }
--- a/xdsl/dialects/__init__.py
+++ b/xdsl/dialects/__init__.py
@@ -43,6 +43,11 @@ def get_all_dialects() -> dict[str, Callable[[], Dialect]]:
 
         return ARM_NEON
 
+    def get_bigint():
+        from xdsl.dialects.bigint import BigInt
+
+        return BigInt
+
     def get_bufferization():
         from xdsl.dialects.bufferization import Bufferization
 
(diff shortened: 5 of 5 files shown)


xdsl/frontend/pyast/type_conversion.py
```diff
@@ -70,9 +71,16 @@ def _convert_name(self, type_hint: ast.Name) -> Attribute:
         if type_name in self.name_to_xdsl_type_map:
             return self.name_to_xdsl_type_map[type_name]
 
+        # Individually handle the bigint dialect
+        if type_name == "int":
+            type_class = BigIntegerType
+            xdsl_type = type_class.to_xdsl()()
+            self._cache_type(type_class, xdsl_type, type_name)
+            return xdsl_type
+
```
EdmundGoodman, line 80: To implement this functionality, we add a special case for `int` type hints. This leaves the original logic for the `frontend.dialect.builtin` stubs intact -- but at some point we probably want to remove these as they are not the programming model they want and add a lot of complexity. However, their removal might break downstream code.

As such, a possible approach is to make a new `python` frontend with for our new goals, re-using the minimal subset of code we need from `pyast` as we go. This then means that we don't need to worry so much about backwards compatibility and can develop at a higher velocity.

xdsl/dialects/bigint.py
superlopuh, line 1: This is a great start! I'm thinking that we can use this PR to track the changes that need to be made in the frontend to make the API not so weird. For example, what is the use of this? Can we remove it somehow?

tests/frontend/pyast/test_readme_examples.py
superlopuh, line 1: can you please make this a filecheck test instead?

tests/frontend/pyast/test_bigint.py
superlopuh, line 1: Please make this test a filecheck test instead

xdsl/frontend/pyast/type_conversion.py
superlopuh: You now know the code better than I do, and maybe duplicating is the best approach, but I would love to avoid duplication if at all possible. I'm sure we can come up with a plan to migrate things incrementally, and move quite quickly. Rewriting things often seems easier than it ends up being, and migration in this case is made easier for us than usual, as the only thing we want to try and do is avoid breaking the tests, we don't even need to avoid breaking the API, as it currently has no users.

EdmundGoodman:
New goal:

```python
p = FrontendProgram()
p.register_type(int, BigInt)
with CodeContext(p):
    def foo(x: int) -> int:  # pyright: ignore[reportUnusedFunction]
        return x

p.compile()
print(p.textual_format())
```

and get the following result:

```python
builtin.module {
  func.func @foo(%0 : !bigint.bigint) -> !bigint.bigint {
    func.return %0 : !bigint.bigint
  }
}
```

The next step is to add operations on these bigint types.

EdmundGoodman:
After this is done:

- [ ] Operations on objects, failing if it can't generate the function `p.register_func(int.__add__, bigint.AddOp)`

EdmundGoodman:
> New goal:
> 
> ```python
> p = FrontendProgram()
> p.register_type(int, BigInt)
> with CodeContext(p):
>     def foo(x: int) -> int:  # pyright: ignore[reportUnusedFunction]
>         return x
> 
> p.compile()
> print(p.textual_format())
> ```
> 
> and get the following result:
> 
> ```python
> builtin.module {
>   func.func @foo(%0 : !bigint.bigint) -> !bigint.bigint {
>     func.return %0 : !bigint.bigint
>   }
> }
> ```
> 
> The next step is to add operations on these bigint types.

This functionality is now added. A `type_registry` on the program is maintained, which is then used to resolve function definition type hints instead of `TypeConverter.convert_type_hint`. However, this breaks the existing filecheck tests as they rely on the existing mechanism of implicit type registration from the frame's globals. Filecheck tests exercising the new functionality are in the `.../frontend/refactor` directory.

Next steps (in new PRs based off this branch for clarity):
1. Implement `register_func` to support arbitrary expressions
2. Update old filecheck tests to use new explicit `register_type` definitions
3. Delete depracated code, for example after `register_func` is implemented, `TypeConverter` can be removed

xdsl/frontend/pyast/code_generation.py
```diff
@@ -44,6 +49,9 @@ class CodeGenerationVisitor(ast.NodeVisitor):
     type_converter: TypeConverter
     """Used for type conversion during code generation."""
 
+    type_registry: dict[TypeName, SourceIrTypePair] = field(default_factory=dict)
+    """Mappings between source code and ir type, indexed by name."""
```
compor, line 53:
```suggestion
    """Mappings between source code and IR type, indexed by name."""
```

xdsl/frontend/pyast/program.py
```diff
@@ -31,9 +36,19 @@ class FrontendProgram:
     xdsl_program: ModuleOp | None = field(default=None)
     """Generated xDSL program when AST is compiled."""
 
+    type_registry: dict[TypeName, SourceIrTypePair] = field(default_factory=dict)
+    """Mappings between source code and ir type, indexed by name."""
```
compor, line 40:
```suggestion
    """Mappings between source code and IR type, indexed by name."""
```

xdsl/frontend/pyast/type_conversion.py
```diff
@@ -12,11 +13,18 @@
     _FrontendType,  # pyright: ignore[reportPrivateUsage]
 )
 from xdsl.frontend.pyast.exception import CodeGenerationException
-from xdsl.ir import Attribute
+from xdsl.ir import Attribute, TypeAttribute
 
 TypeName: TypeAlias = str
 
 
+class SourceIrTypePair(NamedTuple):
```
compor, line 21:
```suggestion
class SourceIRTypePair(NamedTuple):
```

I wasn't pinged, so just caught my eye while scrolling through in a dash.
We have named things in core as `IRNode` etc., hence my nit above.

tests/filecheck/frontend/dialects/readme.py
superlopuh, line 1: Maybe we don't need this? The functionality is already tested

xdsl/frontend/pyast/type_conversion.py
```diff
@@ -12,11 +13,18 @@
...
+from xdsl.ir import Attribute, TypeAttribute
 
 TypeName: TypeAlias = str
 
 
+class SourceIRTypePair(NamedTuple):
+    """Pair of types for source code and its generated IR."""
+
+    source: type
+    ir: type[TypeAttribute]
```
superlopuh, line 25: can you add doc strings please?

xdsl/frontend/pyast/type_conversion.py
```diff
@@ -12,11 +13,18 @@
...
 )
 from xdsl.frontend.pyast.exception import CodeGenerationException
-from xdsl.ir import Attribute
+from xdsl.ir import Attribute, TypeAttribute
 
 TypeName: TypeAlias = str
 
 
+class SourceIRTypePair(NamedTuple):
+    """Pair of types for source code and its generated IR."""
```
superlopuh, line 22: can you add a little example, like float mapping to Float32Type?

xdsl/frontend/pyast/program.py
```diff
@@ -31,9 +36,19 @@ class FrontendProgram:
...
+    """Mappings between source code and IR type, indexed by name."""
+
     file: str | None = field(default=None)
     """Path to the file that contains the program."""
 
+    def register_type(self, source_type: type, ir_type: type[TypeAttribute]) -> None:
+        """Associate a type in the source code with its type in the IR."""
+        type_name = source_type.__name__
+        if type_name in self.type_registry:
+            raise FrontendProgramException(f"Cannot re-register type '{source_type}'")
```
superlopuh, line 49: test for this?

tests/filecheck/frontend/refactor/bigint.py
```diff
@@ -0,0 +1,20 @@
...
+
+from xdsl.dialects.bigint import BigIntegerType
+from xdsl.frontend.pyast.context import CodeContext
+from xdsl.frontend.pyast.program import FrontendProgram
+
+p = FrontendProgram()
+p.register_type(int, BigIntegerType)
+with CodeContext(p):
+    # CHECK:      builtin.module {
+    # CHECK-NEXT:   func.func @foo(%0 : !bigint.bigint) -> !bigint.bigint {
```
superlopuh, line 11: Can you write down somewhere a reminder to add the python names as hints to the ssa values? I have a feeling this is actually important for debuggability and not just a nice to have

xdsl/frontend/pyast/code_generation.py
```diff
@@ -470,14 +483,24 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
...
+                xdsl_type = self.type_registry[arg.annotation.id].ir()
+            else:
+                xdsl_type = self.type_converter.convert_type_hint(arg.annotation)
             argument_types.append(xdsl_type)
 
         return_types: list[Attribute] = []
         if node.returns is not None:
-            xdsl_type = self.type_converter.convert_type_hint(node.returns)
+            if not isinstance(node.returns, ast.Name):
+                raise CodeGenerationException(
```
superlopuh, line 497: test for this?

xdsl/frontend/pyast/code_generation.py
```diff
@@ -470,14 +483,24 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
...
+                    self.file, node.returns.lineno, node.returns.col_offset, ""
```
superlopuh, line 498: ""?

EdmundGoodman: This is copied from the original code, but will fix it now!

tests/filecheck/frontend/dialects/invalid.py
```diff
@@ -118,3 +120,27 @@ def test_not_supported_loop_IV():
...
+
+
+try:
+    with CodeContext(p):
+
+        def test_complex_return_annotation() -> int | None:
+            return
+
+    p.compile(desymref=False)
+    print(p.textual_format())
```
superlopuh, line 144:
```suggestion
    exit(1)
```

tests/filecheck/frontend/dialects/invalid.py
```diff
@@ -118,3 +120,27 @@ def test_not_supported_loop_IV():
...
+except CodeGenerationException as e:
+    print(e.msg)
```
superlopuh, line 146: add check please

superlopuh: and above

tests/filecheck/frontend/dialects/invalid.py
superlopuh: and above

tests/filecheck/frontend/programs/invalid.py
```diff
@@ -239,3 +240,10 @@ def foo() -> i32:
     print(p.textual_format())
 except FrontendProgramException as e:
     print(e.msg)
+
+
+try:
+    p.register_type(int, BigIntegerType)
+    p.register_type(int, BigIntegerType)
+except FrontendProgramException as e:
+    print(e.msg)
```
superlopuh, line 249: can you add a check here, please?

superlopuh (approved):

PR #4095, opened 2025-03-24
misc: add python dialect codegen file with minimal functionality

superlopuh:
This adds a minimal Python codegen stub file, to be filled in later by @mamanain and me. This will be used first by the torch dialect codegen file, and later by the tablegen and potentially other use-cases.


mamanain (approved):
Beautiful

PR #4101, opened 2025-03-25
interpreter: Use symboltable lookup in the interpreter

jumerckx:
This is to support nested `SymbolRefAttr`s.


jumerckx:
With this change, xdsl-run doesn't find the main entry point when it is nested deeper than the top-level module. For example, here:
```mlir
builtin.module {
  riscv.assembly_section ".data" {
  ...
  }
  riscv.assembly_section ".text" {
    riscv.directive ".globl" "main"
    riscv.directive ".p2align" "2"
    riscv_func.func @main() {
    ...
    }
  }
}

```
Potential solution is to have a special `find_main_entrypoint` that goes over all symbols and finds the unique `main` symbol?

xdsl/interpreter.py
```diff
@@ -726,11 +733,11 @@ def cast_value(self, o: Attribute, r: Attribute, value: Any) -> Any:
     def value_for_attribute(self, attr: Attribute, type_attr: Attribute) -> Any:
         return self._impls.attr_value(self, attr, type_attr)
 
-    def get_op_for_symbol(self, symbol: str) -> Operation:
-        if symbol in self.symbol_table:
-            return self.symbol_table[symbol]
-        else:
-            raise InterpretationError(f'Could not find symbol "{symbol}"')
+    def get_op_for_symbol(self, symbol: str | SymbolRefAttr) -> Operation:
+        op = SymbolTable.lookup_symbol(self.module, symbol)
```
compor, line 737: This is not exactly the same lookup, as the interpreter maintains a lookup via the `symbol_table` property.
When using the `SymbolTable.lookup_symbol` you indicate the operation from where to start the search and move "outwards"; here this seems to go straight to the containing module.

Not sure why something like this wouldn't do the trick? (haven't tested it):

```py
if isinstance(symbol, SymbolRefAttr):
    symbol = symbol.data
if symbol in self.symbol_table:
    return self.symbol_table[symbol]
...
```

compor: Hmm, I see; that lookup as you described happens via the `SymbolTable::lookup`; to me then sounds like the intepreter lookup needs a fix

jumerckx: Is there a way to get the current operation being handled by the interpreter, some sort of program counter. If not I can add it and use that value instead of `self.module` to seed the lookup?

superlopuh: This definitely seems like an upgrade in functionality and we need to figure out how to fix riscv to make it work properly with these symbols. The existing symbol table helper should be deleted IMO and just replaced with this, the clients might as well call it directly, and give the relevant root op. I can see how to refactor RISC-V here to work better, and that seems like a pre-requisite to merging this.

tests/filecheck/projects/riscv-backend-paper/add_snitch_stream.mlir
superlopuh, line 1: just a dedent

superlopuh:
@jumerckx could you please add two minimal tests for the lines missing code coverage before you merge?

superlopuh:
Oh actually I just realized that we should add a test for the reason we're adding this functionality, which is a nested symbol. Should be GTG after that.

jumerckx:
I added a test but it also depends on func dialect.
In the process I also got rid of the conversion from callee symbol to string in func.CallOp's run method.

PR #4103, opened 2025-03-25
dialects: (csl-module-wrapper) Add target arch param

n-io:
- Add a `target` field to the csl module wrapper
- Update coeffs API call and the full stencil access reduction
- Remove bool flag from coeffs API

### DIFF
--- a/xdsl/dialects/csl/csl.py
+++ b/xdsl/dialects/csl/csl.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 from abc import ABC
 from collections.abc import Sequence
 from dataclasses import KW_ONLY, dataclass, field
-from typing import Annotated, ClassVar, TypeAlias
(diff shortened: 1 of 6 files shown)


xdsl/dialects/csl/csl.py
```diff
@@ -92,6 +92,8 @@
 from xdsl.utils.isattr import isattr
 from xdsl.utils.str_enum import StrEnum
 
+target = Literal["wse2", "wse3"]
```
dk949, line 95: Nit: could you change this to `Target`, because it's a python type?

dk949 (commented):
Just a minor nitpick from me, otherwise looks good.

xdsl/dialects/csl/csl.py
n-io: Thanks, yes. I was contemplating all CAPS but didn't like it.

PR #4106, opened 2025-03-26
backend: add emit_section helper to AssemblyPrinter

superlopuh:
This lets us avoid a bit of duplication in the emitted assembly, things that belong in the same section now stay in one section as long as they are contiguous.


qaco:
Cool (post-merge but still)

PR #4107, opened 2025-03-26
core: add flag for whether op attributes were printed

superlopuh:
Makes it easier to track state, and lets us avoid the ugly `}{` in the function-like op printing.


xdsl/dialects/utils/format.py
superlopuh, line 1: after some more iterations, I now see that this change is independent of the other one, I'll split it out into its own PR

alexarice (approved):

PR #4112, opened 2025-03-27
dialects: (riscv_func) print visibility in riscv_func.func

superlopuh:
This PR adds logic to print visibility when present on riscv_func.func. The next PR will change the lowering from func to preserve the visibility, currently it's being erased.


superlopuh:
Actually I guess this also could be split into two PRs I'll trim this one down for easier reviewing.

compor (approved):

PR #4113, opened 2025-03-27
dialects: (arm) add helper to print register SSAValues in ARM assembly

superlopuh:
Should make the code a bit more readable and less tedious to extend.

### DIFF
--- a/xdsl/dialects/arm/assembly.py
+++ b/xdsl/dialects/arm/assembly.py
@@ -1,5 +1,8 @@
 import abc
 
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import SSAValue
+
 
 class AssemblyInstructionArg(abc.ABC):
     """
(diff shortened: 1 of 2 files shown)


compor (approved):

emmau678 (approved):

xdsl/dialects/arm/assembly.py
```diff
@@ -9,3 +12,19 @@ class AssemblyInstructionArg(abc.ABC):
...
+    """
+
+    value: SSAValue
+
+    def __init__(self, value: SSAValue) -> None:
+        self.value = value
+
+    def assembly_str(self) -> str:
+        assert isinstance(self.value.type, RegisterType)
+        return self.value.type.register_name.data
```
emmau678, line 30: why is this in the arm assembly file if it is supposed to be used with RegisterType and not ArmRegisterType?

emmau678, line 30: also I'm not too clear on why we would have a concrete class like this instead of just an accessor method inside the register class

superlopuh: Two reasons, one is that I'd like to move this file to somewhere to be shared by all dialects, but also because the ArmRegisterType imports this file, and circular imports aren't allowed. We could have moved this class into that file, I don't feel super strongly about which is better.

PR #4114, opened 2025-03-27
transforms: Add csl-stencil-set-global-coeffs pass

n-io:


xdsl/transforms/csl_stencil_set_global_coeffs.py
```diff
@@ -0,0 +1,189 @@
...
+from xdsl.pattern_rewriter import (
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.rewriter import InsertPoint
+
+
+def get_dir_and_distance(
```
n-io, line 24: This fn was moved here from lower-csl-stencil

xdsl/transforms/csl_stencil_set_global_coeffs.py
```diff
@@ -0,0 +1,189 @@
...
+        d = csl.Direction.SOUTH
+    else:
+        raise ValueError(
+            "Invalid offset, expecting 2-dimensional star-shape neighbor access"
+        )
+    max_distance = abs(max(offset, key=abs))
+    return d, max_distance
+
+
+def get_dir_and_distance_ops(
```
n-io, line 54: This fn was moved here from lower-csl-stencil

xdsl/transforms/csl_stencil_set_global_coeffs.py
```diff
@@ -0,0 +1,189 @@
...
+                        global_coeffs = sorted(
+                            apply.coeffs.data, key=lambda x: x.offset
+                        )
+                elif global_coeffs != (
+                    sorted(apply.coeffs.data, key=lambda x: x.offset)
+                    if apply.coeffs
+                    else []
+                ):
+                    return
+                applies.append(apply)
```
dk949, line 149: This was quite confusing on first parse. Not sure how it can be improved much, a comment would help though

xdsl/transforms/csl_stencil_set_global_coeffs.py
```diff
@@ -0,0 +1,189 @@
...
+        if not global_coeffs or not applies:
+            return
+
+        op_in_main_fn = applies[0]
+        main_fn = None
+        while (
+            op_in_main_fn
+            and (main_fn := op_in_main_fn.parent_op())
+            and not isinstance(main_fn, csl.FuncOp)
+            and not isinstance(main_fn.parent_op(), csl_wrapper.ModuleOp)
```
dk949, line 159: Do we still need to check that parent is a `ModuleOp` if we already know it's a `FuncOp`?

xdsl/transforms/csl_stencil_set_global_coeffs.py
```diff
@@ -0,0 +1,189 @@
...
+    Generates calls to the stencil_comms API to set coefficients.
+
+    The API currently supports only f32 coeffs.
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: csl_wrapper.ModuleOp, rewriter: PatternRewriter, /):
+        applies: list[csl_stencil.ApplyOp] = []
+        global_coeffs = []
+        # first_apply = None
```
dk949, line 135: Commented code :)

dk949 (commented):
Makes sense overall, but I did get stuck when trying to understand the apply op filtering logic in `xdsl/transforms/csl_stencil_set_global_coeffs.py`. 

Some comments would be appreciated :) (especially wrt. sorting the coeffs being needed to handle the reversed case).

xdsl/transforms/csl_stencil_set_global_coeffs.py
n-io: Just to check we're in a top-level function, i.e. one that is not nested. And to check it's got the right name.

xdsl/transforms/csl_stencil_set_global_coeffs.py
n-io: Thanks, removed.

dk949 (approved):

PR #4116, opened 2025-03-27
dialects (arm): add arrangement specifier attribute

emmau678:
---
title: dialects (arm): add arrangement specifier attribute

---

Adding an arrangement specifier attribute for ARM NEON instructions, which specifies the number and size of vector lanes. We assume full 128-bit registers.

### DIFF
--- a/tests/dialects/arm/test_assembly_arg_str.py
+++ b/tests/dialects/arm/test_assembly_arg_str.py
@@ -1,5 +1,18 @@
+from io import StringIO
+
 from xdsl.dialects.arm.register import IntRegisterType
+from xdsl.dialects.arm_neon import NeonArrangement, NeonArrangementAttr
+from xdsl.printer import Printer
 
 
 def test_assembly_arg_str_ARMRegister():
     assert IntRegisterType.from_name("x0").assembly_str() == "x0"
+
+
+def test_gello():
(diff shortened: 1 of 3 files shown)


tests/dialects/arm/test_assembly_arg_str.py
superlopuh, line 1: Let's move this test to a filecheck test like `tests/filecheck/dialects/bigint/attrs.mlir`? I would expect a file split in two, one with the three variants that work, and one with `#arm_neon<arrangement INVALID>` that raises an error of some sort.

emmau678: added in latest commit

xdsl/dialects/arm_neon.py
```diff
@@ -62,10 +69,39 @@ def infinite_register_prefix(cls):
...
+      - D  → 2 double-precision floats
+      - S  → 4 single-precision floats
+      - H  → 8 half-precision floats
+    """
+
+    D = "D"
+    S = "S"
+    H = "H"
+
+    def map_to_num_els(self):
```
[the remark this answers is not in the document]
superlopuh: Now that I think about it, it should probably also be a `@property`, and the map should be defined outside of the function, but this is all more of a nit.

PR #4117, opened 2025-03-27
dialects (arm): Add vector with arrangement

emmau678:
Add representation of vector with arrangement specifier for asm printing


superlopuh (approved):
Perfect

PR #4122, opened 2025-03-28
dialects (arm): Add NEON FMUL op (mixed scalar/vector)

emmau678:
Add Floating-point multiply operation to ARM (NEON) dialect (mixed: first source operand is a vector, second is a scalar, destination is a vector).

This instruction multiplies each of the floating-point values in the first source operand by the second source operand and writes the resulting values to the corresponding lanes of the destination.

### DIFF
--- a/tests/filecheck/dialects/arm_neon/test_ops.mlir
+++ b/tests/filecheck/dialects/arm_neon/test_ops.mlir
@@ -1,0 +1,9 @@
+// RUN: XDSL_ROUNDTRIP
+// RUN: xdsl-opt -t arm-asm %s | filecheck %s --check-prefix=CHECK-ASM
+// CHECK: %v1 = arm_neon.get_register : !arm_neon.reg<v1>
+%v1 = arm_neon.get_register : !arm_neon.reg<v1>
+// CHECK: %v2 = arm_neon.get_register : !arm_neon.reg<v2>
+%v2 = arm_neon.get_register : !arm_neon.reg<v2>
+// CHECK: %dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2 {arrangement = #arm_neon<arrangement S>, comment = "floating-point vector multiply v1 by v2", scalar_idx = 0 : i8} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
+// CHECK-ASM: fmul v3.4S, v1.4S, v2.S[0] # floating-point vector multiply v1 by v2
+%dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2 {"arrangement" = #arm_neon<arrangement S>, "comment" = "floating-point vector multiply v1 by v2", "scalar_idx" = 0 : i8} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
--- a/tests/filecheck/dialects/arm_neon/test_ops.mlir
+++ b/tests/filecheck/dialects/arm_neon/test_ops.mlir
@@ -4,6 +4,6 @@
 %v1 = arm_neon.get_register : !arm_neon.reg<v1>
 // CHECK: %v2 = arm_neon.get_register : !arm_neon.reg<v2>
 %v2 = arm_neon.get_register : !arm_neon.reg<v2>
-// CHECK: %dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2 {arrangement = #arm_neon<arrangement S>, comment = "floating-point vector multiply v1 by v2", scalar_idx = 0 : i8} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
+// CHECK: %dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2[0] S {comment = "floating-point vector multiply v1 by v2"} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
 // CHECK-ASM: fmul v3.4S, v1.4S, v2.S[0] # floating-point vector multiply v1 by v2
-%dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2 {"arrangement" = #arm_neon<arrangement S>, "comment" = "floating-point vector multiply v1 by v2", "scalar_idx" = 0 : i8} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
+%dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2[0] S {"comment" = "floating-point vector multiply v1 by v2"} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
--- a/xdsl/dialects/arm_neon.py
+++ b/xdsl/dialects/arm_neon.py
@@ -161,7 +161,7 @@ class DSSFMulVecScalarOp(ARMInstruction):
     Vd, Vn, Vm specify the regs. The <T> specifier determines element arrangement (size and count).
     The <idx> specifier determines the index of Vm at which the second source operand (scalar) can be found,
     preceded by a size specifier.
-    
+
     See external [documentation](https://developer.arm.com/documentation/ddi0602/2024-12/SIMD-FP-Instructions/FMUL--vector---Floating-point-multiply--vector--?lang=en#T_option__4).
     """
 
--- a/xdsl/dialects/arm_neon.py
+++ b/xdsl/dialects/arm_neon.py
@@ -168,9 +168,7 @@ class DSSFMulVecScalarOp(ARMInstruction):
     scalar_idx = attr_def(IntegerAttr[i8])
     arrangement = attr_def(NeonArrangementAttr)
 
-    assembly_format = (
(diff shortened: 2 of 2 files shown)


xdsl/dialects/arm_neon.py
```diff
@@ -123,9 +130,95 @@ def assembly_str(self):
...
+
+    def __init__(self, register_type: NEONRegisterType):
+        super().__init__(result_types=[register_type])
+
+    def assembly_line(self):
+        return None
+
+
+@irdl_op_definition
+class DSSFMulVecScalarOp(ARMInstruction):
```
qaco, line 152: So if I understand correctly: The fmul instruction is polymorphic and can consume vectors of arbitrary size or even scalars. Is that right? The ARM manual page is not very clear on this point.

In the case implemented here, the fmul on a vector and a scalar allows saving the vector broadcast instruction that is required in AVX to multiply each element of a vector by the same scalar (like with variadic load/store, I wonder if it simply optimizes the front-end or if it also relieves pressure on the back-end). Anyway, this is great!

This inspires two additional questions:
+ Does fma have the same property than fmul?
+ Does this mechanism work even when the second parameter is directly a scalar register (and not an indexed vector register)?

tests/filecheck/dialects/arm_neon/test_ops.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: XDSL_ROUNDTRIP
+// RUN: xdsl-opt -t arm-asm %s | filecheck %s --check-prefix=CHECK-ASM
+// CHECK: %v1 = arm_neon.get_register : !arm_neon.reg<v1>
+%v1 = arm_neon.get_register : !arm_neon.reg<v1>
+// CHECK: %v2 = arm_neon.get_register : !arm_neon.reg<v2>
+%v2 = arm_neon.get_register : !arm_neon.reg<v2>
+// CHECK: %dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2 {arrangement = #arm_neon<arrangement S>, comment = "floating-point vector multiply v1 by v2", scalar_idx = 0 : i8} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
+// CHECK-ASM: fmul v3.4S, v1.4S, v2.S[0] # floating-point vector multiply v1 by v2
+%dss_fmulvec = arm_neon.dss.fmulvec %v1, %v2 {"arrangement" = #arm_neon<arrangement S>, "comment" = "floating-point vector multiply v1 by v2", "scalar_idx" = 0 : i8} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
```
superlopuh, line 9: Would something like this be possible? feels like it would be nice to have this crucial info a bit more prominent in the custom syntax (`arrangement S` instead of just `S` would also work if that's easier`

```suggestion
%dss_fmulvec = arm_neon.dss.fmulvec S %v1, %v2[0] {"comment" = "floating-point vector multiply v1 by v2",} : (!arm_neon.reg<v1>, !arm_neon.reg<v2>) -> !arm_neon.reg<v3>
```

xdsl/dialects/arm_neon.py
```diff
@@ -123,9 +130,95 @@ def assembly_str(self):
...
+        )
+
+    def assembly_instruction_name(self) -> str:
+        return "fmul"
+
+    def assembly_line_args(self):
+        assert isinstance(self.d.type, NEONRegisterType)
+        assert isinstance(self.s1.type, NEONRegisterType)
+        assert isinstance(self.s2.type, NEONRegisterType)
+        if isinstance(self.arrangement, NeonArrangement):
```
superlopuh, line 204: is this necessary?

superlopuh: It's not possible, I think, and we definitely shouldn't be setting attributes in this helper

xdsl/dialects/arm_neon.py
```diff
@@ -123,9 +130,95 @@ def assembly_str(self):
...
+            result_types=(d,),
+        )
+
+    def assembly_instruction_name(self) -> str:
+        return "fmul"
+
+    def assembly_line_args(self):
+        assert isinstance(self.d.type, NEONRegisterType)
+        assert isinstance(self.s1.type, NEONRegisterType)
+        assert isinstance(self.s2.type, NEONRegisterType)
```
superlopuh, line 203: I was hoping that the `VectorWithArrangement` class would do this for us, to avoid extra calls inside the helper

xdsl/dialects/arm_neon.py
emmau678: thanks for the review! This is my understanding of these points, although the documentation is difficult to navigate (and sometimes even to find):

> The fmul instruction itself is polymorphic and can consume vectors of arbitrary size or even scalars. Is that right?
There are a couple of types of fmul that can be done using the same mnemonic, yes: FMUL and FNMUL can be scalar, vector, or mixed (see [VFP vector and scalar operations](https://developer.arm.com/documentation/dui0204/g/neon-and-vfp-programming/vfp-vector-and-scalar-operations?lang=en)). The vectors aren't arbitrarily sized, but instead defined by the arrangement specifiers, which tell us the size and number of the vector lanes within the registers. In this implementation we assume full use of 128-bit registers.

> Does fma have the same property than fmul?
Yes

> Does this mechanism work even when the second parameter is directly a scalar register (and not an indexed vector register)?
The docs I can find indicate yes, but I haven't tested it

tests/filecheck/dialects/arm_neon/test_ops.mlir
emmau678: sure, please see next commit

xdsl/dialects/arm_neon.py
qaco: Great, so for the efficient implementation of matmuls, I guess the scalar/vector fma is the next step!

xdsl/dialects/arm_neon.py
emmau678: it definitely doesn't work out-of-the-box, I've been trying a few things but haven't found a way to make it work yet.
```
Argument of type "Attribute" cannot be assigned to parameter "reg" of type "NEONRegisterType | SSAValue" in function "__init__"
  Type "Attribute" is not assignable to type "NEONRegisterType | SSAValue"
    "Attribute" is not assignable to "NEONRegisterType"
    "Attribute" is not assignable to "SSAValue"
    ```

Comment (superlopuh) [reply xdsl/dialects/arm_neon.py:201]: it seems like it's complaining about `self.d.type` instead of `self.d`

File: xdsl/dialects/arm_neon.py [hunk 5]
```diff
@@ -106,7 +113,7 @@ class VectorWithArrangement(AssemblyInstructionArg):
 
     def __init__(
         self,
-        reg: NEONRegisterType,
+        reg: NEONRegisterType | SSAValue,
```
superlopuh, line 116: My feeling is that `self.reg` should be a NEONRegisterType, and that the init should do a `if isinstance(reg, SSAValue): assert isinstance(reg.type, NEONRegisterType); reg = reg.type`

xdsl/dialects/arm_neon.py
emmau678: if I change ```self.d.type``` to ```self.d``` in the argument to VectorWithArrangement, the assertion fails with: ```AssertionError: Expected NEONRegisterType, but got OpResult```

xdsl/dialects/arm_neon.py
emmau678: updated with these changes. I'm still getting red squiggly line warnings in the ide but the tests are passing

emmau678: ok the pyright checks are catching it in CI, not surprising I suppose

xdsl/dialects/arm_neon.py
```diff
@@ -111,21 +118,102 @@ def __init__(
         *,
         index: int | None = None,
     ):
+        if isinstance(reg, SSAValue):
+            assert isinstance(reg.type, NEONRegisterType)
+            reg = reg.type
```
superlopuh, line 123: I can't comment above, but changing 110 to `reg: NEONRegisterType` should do the trick

xdsl/dialects/arm_neon.py
```diff
@@ -111,21 +118,102 @@ def __init__(
...
+class DSSFMulVecScalarOp(ARMInstruction):
+    """
+    Floating-point multiply (mixed: first source operand is a vector, second is a scalar. Destination is a vector)
+    This instruction multiplies each of the floating-point values in the first source operand by the
+    second source operand and writes the resulting values to the corresponding lanes of the destination.
+    Encoding: FMUL <Vd>.<T>, <Vn>.<T>, <Vm>.<idx>.
+    Vd, Vn, Vm specify the regs. The <T> specifier determines element arrangement (size and count).
+    The <idx> specifier determines the index of Vm at which the second source operand (scalar) can be found,
+    preceded by a size specifier.
+    https://developer.arm.com/documentation/ddi0602/2024-12/SIMD-FP-Instructions/FMUL--vector---Floating-point-multiply--vector--?lang=en#T_option__4
```
compor, line 164:
```suggestion
    
    See external [documentation](https://developer.arm.com/documentation/ddi0602/2024-12/SIMD-FP-Instructions/FMUL--vector---Floating-point-multiply--vector--?lang=en#T_option__4).
```

superlopuh (approved):

PR #4124, opened 2025-03-29
dialects: (builtin) handle unsigned integer packing

superlopuh:


jorendumoulin (approved):
Nice!

isn't unsigned just always `upper(format_str)` ?

superlopuh:
Yep but that felt a little heavy-handed, for some reason...

PR #3991, opened 2025-02-28
core: Make IRDLOperation results typed

math-fehr:
Stacked PRs:
 * __->__#3991
 * #3989
 * #3988
 * #3987

--- --- ---

### core: Make IRDLOperation results typed

Now, defining a result with `resname = result_def(T)` will make
`resname` have the type `SSAValue[T]`.

This removes a lot of `cast` and `isa`/`isattr` in our codebase

PR #4125, opened 2025-03-29
core: Make SSAValue type immutable

math-fehr:
Stacked PRs:
 * #3991
 * #3989
 * #3988
 * #3987
 * __->__#4125

--- --- ---

### core: Make SSAValue type immutable

SSAValue can no longer change type after creation.
This ensures that we can have a generic type in values,
as otherwise this would be invalidated by changing the type.

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -42,6 +42,7 @@ from xdsl.irdl import (
     var_result_def,
 )
 from xdsl.pattern_rewriter import RewritePattern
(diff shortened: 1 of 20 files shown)


math-fehr:
Sorry for the delay, this PR makes SSAValue immutable so we can safely make it generic on the next PR.
The only non-trivial change is in rewriter, pattern_rewriter, and register_allocation

superlopuh (approved):
🚀🚀🚀

compor (approved):

PR #4130, opened 2025-03-30
misc: (makefile) remove directive defining bash location

### LINKED ISSUE #4123: misc: (makefile) tests can't be run without "/bin/bash"
The test makefile now require bash to be located at "/bin/bash"

compor:
Resolves #4123

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,6 @@ LIT_OPTIONS ?= -v --order=smart
 # make tasks run all commands in a single shell
 .ONESHELL:
 
-# use bash as the shell
-SHELL := /bin/bash
-
 .PHONY: uv-installed
 uv-installed:
 	@command -v uv &> /dev/null ||\


superlopuh (approved):
Tested this on my Mac

compor:
> Tested this on my Mac
I also tried this after moving `bash` to another location in my `PATH`

I'll wait for @alexarice , since I don't use NixOS.

alexarice (approved):
Works locally, do you remember why this was added to begin with? I think `SHELL=bash` also works for me, as should `/bin/sh`

compor:
> Works locally, do you remember why this was added to begin with? I think `SHELL=bash` also works for me, as should `/bin/sh`

I think it was added while experimenting with the `test-marimo` rule, so my suspicion it was a relic from that.
The rule was broken for the macOS bash, and then probably escaped us from the fix.

PR #4131, opened 2025-03-30
CI: add some more Pyright rules and add comments

superlopuh:
I thought I'd add some more Pyright rules and comment some of the ones we have to make the reason clearer.


math-fehr (approved):
Nice!

PR #4155, opened 2025-04-01
tool: (xdsl-tblgen) add tablegen file to build test json from

alexarice:
Adds a tablegen file with instructions for regenerating the test json. Should hopefully make it easier to modify this test in the future.

Stacked PR


alexarice:
Open question: How important is pretty printing the json. In my mind it doesn't matter too much as in the ideal world the user should never look at it. If we did want to pretty print it however, does anyone have a suggestion for a common tool for doing so?

alexarice:
I can't see these files in the build output, and even if they were I think current mlir integration tests just rely on mlir-opt (and maybe mlir-opt-runner) being in path

tests/xdsl_tblgen/test_tblgen.py
```diff
@@ -11,5 +11,4 @@ def test_run_tblgen_to_py():
     with open("tests/xdsl_tblgen/test.py") as f:
         expected = f.read()
 
-        assert len(out_str.strip()) == len(expected.strip())
```
superlopuh, line 14: 🙏

tests/xdsl_tblgen/test.json
```diff
@@ -1,1015 +1 @@
...
...uildableType", "I"], "bitwidth": 32, "summary": "32-bit signless integer"}, "UI8": {"!name": "UI8", "!superclasses": ["Constraint", "TypeConstraint", "Type", "BuildableType", "UI"], "bitwidth": 8, "summary": "8-bit unsigned integer"}, "Test_SingletonBType": {"!name": "Test_SingletonBType", "!superclasses": ["Constraint", "TypeConstraint", "Type", "DialectType", "AttrOrTypeDef", "TypeDef", "Test_Type"], "assemblyFormat": null, "dialect": {"def": "Test_Dialect", "kind": "def", "printable": "Test_Dialect"}, "summary": "", "typeName": "test.singleton_b"}, "Test_SingletonAType": {"!name": "Test_SingletonAType", "!superclasses": ["Constraint", "TypeConstraint", "Type", "DialectType", "AttrOrTypeDef", "TypeDef", "Test_Type"], "assemblyFormat": null, "dialect": {"def": "Test_Dialect", "kind": "def", "printable": "Test_Dialect"}, "summary": "", "typeName": "test.singleton_a"}, "!instanceof": {"TypeDef": ["Test_SingletonAType", "Test_SingletonBType", "Test_SingletonCType"], "AttrDef": ["Test_TestAttr"], "Op": ["Test_AndOp", "Test_AnyOp", "Test_AttributesOp", "Test_ConfinedOp", "Test_Integers", "Test_OrOp", "Test_SummaryOp", "Test_TypesOp", "Test_VariadicityOp"], "Dialect": ["Test_Dialect"]}}
```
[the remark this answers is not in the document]
superlopuh: Piping it through jq would be nice to make it more readable, and for easier diffing

superlopuh (approved):

alexarice:
Pretty printed the json and updated the instructions

alexarice:
So apparently the json culling part of the program is non-deterministic, so that's fun. I'll try to fix it tomorrow but imagine I need an ordered dict somewhere (or sort the output keys)

tests/xdsl_tblgen/test.td
```diff
@@ -0,0 +1,87 @@
+// Tablegen file used to generate `test.json`, which can be generated with the following command:
+// ```
+// llvm-tblgen -D test test.td -I $LLVM/mlir/include --dump-json | xdsl-tblgen -c | jq > test.json
```
superlopuh, line 3: ?
```suggestion
// llvm-tblgen -D test test.td -I $LLVM/mlir/include --dump-json | xdsl-tblgen -c | jq --sort-keys > test.json
```

alexarice: I feel this fixes the symptom rather than the cause, so will investigate what is making `xdsl-tblgen -c` non-deterministic

alexarice:
The non-determinism is caused by iterating over a python set type. There seems to be two options:
- Use a `dict[A, None]` as an ordered `set[A]`. Python unfortunately doesn't seem to have an ordered set
- sort the keys when dumping the json

Any preference between these? @compor @superlopuh

alexarice:
I've pushed the sorted keys version for now

xdsl/tools/xdsl_tblgen.py
```diff
@@ -521,7 +521,7 @@ def cull_field(js_in: dict[str, Any]) -> dict[str, Any]:
         key: js["!instanceof"][key] for key in ("TypeDef", "AttrDef", "Op", "Dialect")
     }
 
-    print(json.dumps(culled), file=output_file)
+    print(json.dumps(culled, sort_keys=True), file=output_file)
```
superlopuh, line 524: oh, we can avoid the jq thing completely, right?
```suggestion
    print(json.dumps(culled, sort_keys=True, indent=4), file=output_file)
```

alexarice: Perhaps, I'll try it after seminars

alexarice:
Went for indent=2 as that aligns with what `jq` did anyway.

alexarice:
@compor @superlopuh happy for this to be merged?

superlopuh:
yep!

math-fehr (approved):
Nice!

PR #4159, opened 2025-04-01
backend: vector loads to x86 assembly

qaco:
These changes implement the lowering of vector.load to x86 vector instructions.

### DIFF
--- a/tests/filecheck/backend/x86/convert_vector_load_to_x86.mlir
+++ b/tests/filecheck/backend/x86/convert_vector_load_to_x86.mlir
@@ -1,0 +1,27 @@
+// RUN: xdsl-opt -p 'convert-vectors-to-x86{arch=avx2}' --verify-diagnostics --split-input-file  %s | filecheck %s
+
+%m = "test.op"(): () -> memref<16xf32>
+%i = arith.constant 0: index
+%v = vector.load %m[%i]: memref<16xf32>, vector<8xf32>
+
+// CHECK:       builtin.module {
+// CHECK-NEXT:    %m = "test.op"() : () -> memref<16xf32>
+// CHECK-NEXT:    %i = arith.constant 0 : index
+// CHECK-NEXT:    %v = memref.subview %m[%i] [8] [1] : memref<16xf32> to memref<8xf32>
+// CHECK-NEXT:    %v_1 = builtin.unrealized_conversion_cast %v : memref<8xf32> to !x86.reg
+// CHECK-NEXT:    %v_2 = x86.rm.vmovups %v_1, 0 : (!x86.reg) -> !x86.avx2reg
+// CHECK-NEXT:  }
+
+// -----
+
+// CHECK: Half-precision vector load is not implemented yet.
+%m1 = "test.op"(): () -> memref<32xf16>
+%i1 = arith.constant 0: index
+%v1 = vector.load %m1[%i1]: memref<32xf16>, vector<16xf16>
+
+// -----
+
+// CHECK: Double precision vector load is not implemented yet.
+%m1 = "test.op"(): () -> memref<8xf64>
+%i1 = arith.constant 0: index
+%v1 = vector.load %m1[%i1]: memref<8xf64>, vector<4xf64>
--- a/xdsl/transforms/__init__.py
+++ b/xdsl/transforms/__init__.py
@@ -168,6 +168,11 @@ def get_all_passes() -> dict[str, Callable[[], type[ModulePass]]]:
 
         return varith_transformations.ConvertVarithToArithPass
 
+    def get_convert_vectors_to_x86():
+        from xdsl.transforms import convert_vectors_to_x86
+
+        return convert_vectors_to_x86.ConvertVectorsToX86Pass
+
     def get_jax_use_donated_arguments():
         from xdsl.transforms import jax_use_donated_arguments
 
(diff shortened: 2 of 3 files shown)


tests/filecheck/backend/x86/convert_vector_load_to_x86.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: xdsl-opt -p 'convert-vectors-to-x86{arch=avx2}' --verify-diagnostics --split-input-file  %s | filecheck %s
+
+%m = "test.op"(): () -> memref<16xf32>
+%i = arith.constant 0: index
+%v = vector.load %m[%i]: memref<16xf32>, vector<8xf32>
+
+// CHECK:       builtin.module {
+// CHECK-NEXT:    %m = "test.op"() : () -> memref<16xf32>
+// CHECK-NEXT:    %i = arith.constant 0 : index
+// CHECK-NEXT:    %v = memref.subview %m[%i] [8] [1] : memref<16xf32> to memref<8xf32>
```
superlopuh, line 10: Now I see why the subview op came into this! I'm already living on planet ptr in my head, it feels like the best way to go here also. We have a partial implementation of [the future of ptr in MLIR](https://github.com/llvm/llvm-project/pull/73057/files#diff-87703bf35afbecd86f9399ef2e6f4d81f0f3cd25dccd3cba881d9ad253d37006L2065), and would probably benefit from using it here. The idea would be to add a `vector-to-ptr` lowering that could be used by all our backends, and avoid the memref subview here, doing pointer arithmetic instead. Doing things in ptr + arith let us use existing patterns for things like cse and licm for pointer arithmetic, hopefully moving more of the optimisations to a shared, target-independent, stage of the compilation pipeline.

superlopuh: Then we would just need a `lower-vector-to-x86` that would inspect the type of the load and pick the right x86 op accordingly

qaco: Why is it cool to use directly pointer arithmetic rathen than a subview (which is a more structured representation) ? At some point, the subview will be lowered to a raw pointer, no ?

superlopuh: Yep, the question is when in the lowering pipeline do we lower to pointers. From my work on the riscv dialects, doing the kinds of optimisations that we might want to do with constant folding is and loop-invariant code motion is better done earlier than later. Also, doing both the instruction selection and pointer arithmetic in one pattern feels like too much coupling (and work). We haven't finished the migration in the riscv pipeline yet (PR is up but is sort of on pause), but I find that approach much nicer than the one-shot memref to riscv path we initially used. My understanding is that this is why they are adding the ptr dialect to MLIR as well.

qaco: Let's say I have ```v = vector.load m[i,j]```, m being a memref and i,j being indices (and v being the vector). What is the outcome of ```vector-to-ptr``` on this operator in your vision ?

superlopuh: Something like:
```
func (m: memref<100x100xf32>, i: index, j: index) -> vector<4xf32> {
   // v = vector.load m[i,j] : vector<4xf32>
   base = ptr.to_ptr : ptr.ptr
   c100 = arith.constant 100 : index
   row_offset = c100 * i
   el_offset = row_offset + j
   ptr = ptr.add base, el_offset
   v = ptr.read : vector<4xf32>
}
```

superlopuh: each of these things will lower to either a single x86 instruction or noop

qaco: So in your vision, when it comes to generate x86 instructions, the semantic information about the data layout (shape, element type, stride...) is carried by the instruction ```ptr.read```, instead of being carried by the data itself (the SSA variable) ?

superlopuh: Can they be leveraged in x86 instructions?

qaco: I don't know, I visualized the reverse (first vectors to x86, then memrefs to pointers, possibly directly x86 pointers). I just reword in order to make sure I understand well ! A good point for the approach you describe is the fact it does not need to insert ```unrealized_conversion_cast```.  Btw, have we an already existing ```memref-to-ptr``` pass in xDSL ?

superlopuh: Yep! https://github.com/xdslproject/xdsl/blob/main/xdsl/transforms/convert_memref_to_ptr.py I found that our memref to riscv lowering was too big a jump, so if you're happy with it then I think ptr is the way to go everywhere.

qaco: Let's go !

compor: so, are you guys closing this then?

qaco: Yes I do it !

PR #4158, opened 2025-04-01
tool: (xdsl-tblgen) fix assembly format

alexarice:
Fixes assembly format printing for xdsl-tblgen by wrapping the string in triple quotes. This is necessary to deal with newlines, an alternative would be to replace all the new lines in the assembly format with spaces, or check if the assembly format has new lines before deciding whether to use single or triple quotes.

@Jimmy2027, I believe this should fix the issues addressed in https://github.com/xdslproject/xdsl/pull/4149 now, would you be able to test this commit to see if it can parse the dialect you were testing.

@erick-xanadu Could you confirm this partially fixes #3787 . At least locally this seem to not crash once the second dialect is removed.

PR #4160, opened 2025-04-01
tool: (xdsl-tblgen) fix AnyAttrOf generation

Jimmy2027:
fixes issue mentioned in #4158, without this the AnyAttrOf generation is broken:
```python
@irdl_op_definition
class EmitC_ConstantOp(IRDLOperation):
    """Constant operation"""

    name = "emitc.constant"

    value = prop_def(
    AnyOf(
        BaseAttr(EmitC_OpaqueAttr),AnyAttr()
        )
    )
    )

    v12 = result_def(AnyAttr())
```

with this change the [emitc](https://mlir.llvm.org/docs/Dialects/EmitC/) dialect can be generated without errors.


Jimmy2027:
@alexarice another issue is the generation of the [emitC.GetGlobalOp](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/EmitC/IR/EmitC.td#L1235)
it is generated like the following:
```python
@irdl_op_definition
class EmitC_GetGlobalOp(IRDLOperation):
    """Obtain access to a global variable"""

    name = prop_def(AnyAttr())

    assembly_format = """$name `:` type($result) attr-dict"""

    result = result_def(
    AnyOf(
        (
            BaseAttr(EmitC_ArrayType),BaseAttr(EmitC_LValueType)
        )
    )
    )

```

the "name" attribute has a conflict with the name parameter used by xDSL. 
If that happens it probably needs to be renamed to something like "symbol_name".
Something like this could work:
```python
@irdl_op_definition
class EmitC_GetGlobalOp(IRDLOperation):
    """Obtain access to a global variable"""

    name = "emitc.get_global"

    sym_name = prop_def(BaseAttr(SymbolNameAttr))

    result = result_def(AnyOf((BaseAttr(EmitC_ArrayType), BaseAttr(EmitC_LValueType))))

    assembly_format = """$sym_name `:` type($result) attr-dict"""

 ```

However when using this, the sym_name attribute is not printed correctly, it is printed as "#symbol_name<"my_data">":
` %0 = emitc.get_global #symbol_name<"my_data"> : !emitc.array<4xi8>` instead of `%0 = emitc.get_global @my_data : !emitc.array<4xi8>`
Does this need a custom print function to work correctly?

**Review by alexarice — COMMENTED** [review]
This looks great, thanks for the fix. We should also add a test that uses `AnyAttrOf` but that can wait until the PR stack is simplified a bit

**Comment by alexarice** [comment]
For the `name` property, we will want to change it to:
```python
name_ = prop_def(AnyAttr(), prop_name="name")
```
we've done similar for names that clash with python keywords (e.g. `in`) but hadn't encountered this case yet.

alexarice:
How this will/should interact with the assembly format is another question, I hope that the assembly format will use the `prop_name` but am unsure.

Jimmy2027:
okay the `name_` should actually be defined as `SymbolRefAttr` in this case:
```python
name_ = prop_def(BaseAttr(SymbolRefAttr), prop_name="name")
```
There is actually no printing defined for `SymbolNameAttr`, is that intended?
https://github.com/xdslproject/xdsl/blob/main/xdsl/printer.py#L447

Jimmy2027:
yup sounds good :)

alexarice:
I'm not sure why this got closed by me merging the parent commit, and github won't let me reopen it.

Jimmy2027:
no worries, I created a new PR: #4165

PR #4161, opened 2025-04-01
transformations: no assembly section when lowering to riscv_func

superlopuh:
Now that the riscv_func.funcs can emit their own .text section as necessary, we don't need the explicit assembly_section structure. With this change, we can look up the riscv func using the standard SymbolOpInterface, like func.funv.

CC @jumerckx - With this the functions will be ported, but I need to do a similar refactor for global data for the switch to symbols in the interpreter to work.

### DIFF
--- a/tests/interactive/test_app.py
+++ b/tests/interactive/test_app.py
@@ -178,21 +178,20 @@ async def test_buttons():
         await pilot.pause()
         assert (
             app.output_text_area.text
-            == """builtin.module {
-  riscv.assembly_section ".text" {
-    riscv_func.func public @hello(%n : !riscv.reg<a0>) -> !riscv.reg<a0> attributes {p2align = 2 : i8} {
-      %0 = riscv.mv %n : (!riscv.reg<a0>) -> !riscv.reg
-      %n_1 = builtin.unrealized_conversion_cast %0 : !riscv.reg to index
-      %two = riscv.li 2 : !riscv.reg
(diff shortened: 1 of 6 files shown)


tests/interactive/test_app.py
```diff
@@ -178,21 +178,20 @@ async def test_buttons():
...
-      %two_1 = builtin.unrealized_conversion_cast %two : !riscv.reg to index
-      %res = builtin.unrealized_conversion_cast %n_1 : index to !riscv.reg
-      %res_1 = builtin.unrealized_conversion_cast %two_1 : index to !riscv.reg
-      %res_2 = riscv.mul %res, %res_1 : (!riscv.reg, !riscv.reg) -> !riscv.reg
-      %res_3 = builtin.unrealized_conversion_cast %res_2 : !riscv.reg to index
-      %1 = builtin.unrealized_conversion_cast %res_3 : index to !riscv.reg
-      %2 = riscv.mv %1 : (!riscv.reg) -> !riscv.reg<a0>
-      riscv_func.return %2 : !riscv.reg<a0>
-    }
+            == """\
```
compor, line 181: super-annoying these changes

compor (approved):

tests/interactive/test_app.py
superlopuh: I agree that it's a pain but my opinion is that it is worth it

compor: apparently there is an issue to decouple this from the interactive stuff, I'll try to get to it some time soon

PR #4162, opened 2025-04-02
dialects: (pdl_interp) Add initial support for  pdl_interp

jumerckx:
The [pdl_interp dialect](https://mlir.llvm.org/docs/Dialects/PDLInterpOps/) provides a lower-level abstraction compared to pdl for executing pattern matching and rewriting. Lowering from pdl to pdl_interp is not considered in this pr.
Not yet all operations have been implemented, only those that occured in simple tests.


compor:
I wonder if this can be broken down a bit to a series of PRs, especially if you're interested in more reviews (i.e., this is not experimental as far as I can tell)

jumerckx:
> I wonder if this can be broken down a bit to a series of PRs

Ok! How do you propose I do the split, dialect definition separate from interpreter, or more splits than that?

superlopuh:
I would recommend splitting this up like this:

1. all pdl_interp ops where we can handle their assembly format (+ XDSL_ROUNDTRIP test + mlir-opt interop test)
2. the rest of the pdl_interp ops with custom print/parse functions (+ same tests)
3. interpreter impls (might want multiple PRs for this, depending on how complex they are, the file is quite big rn)
4. pdl-interp-apply pass

I would probably make this the pdl-interp-apply pass PR, mark it as a draft for now, and re-open when 1-3 are in

jumerckx:
> I would recommend splitting this up like this:
> 1. all pdl_interp ops where we can handle their assembly format (+ XDSL_ROUNDTRIP test + mlir-opt interop test)

https://github.com/xdslproject/xdsl/pull/4176

I assume you mean to open those prs one after the other, each time on top of main?

superlopuh:
Yeah, it's possible to open them all at the same time, but I find that to be more time consuming and error-prone, by opening them one at a time you make sure that people review the right thing, and you avoid noise by not propagating changes up the stack if something has to change in the PR closest to main.

jumerckx:
> 2. the rest of the pdl_interp ops with custom print/parse functions (+ same tests)

https://github.com/xdslproject/xdsl/pull/4180

xdsl/interpreters/pdl_interp.py
```diff
@@ -0,0 +1,355 @@
...
+        return Successor(successor, ()), ()
+
+    @impl_terminator(pdl_interp.AreEqualOp)
+    def run_areequal(
+        self,
+        interpreter: Interpreter,
+        op: pdl_interp.AreEqualOp,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        assert len(args) >= 2
```
cowardsa, line 230: What happens if len(args) > 2? This only checks equality of arg[0] and arg[1]

xdsl/interpreters/pdl_interp.py
```diff
@@ -0,0 +1,355 @@
...
+        return (args[0].results[op.index.value.data],)
+
+    @impl(pdl_interp.GetResultsOp)
+    def run_getresults(
+        self,
+        interpreter: Interpreter,
+        op: pdl_interp.GetResultsOp,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        assert len(args) > 0
```
cowardsa, line 77: Should this len(arg) be equal to 1?

cowardsa (commented):
Several asserts seem to allow for more arguments then ignore the additional arguments - comments inline, primarily in interpreters/pdl_interp.py

xdsl/interpreters/pdl_interp.py
jumerckx: Yes, same goes everywhere else. I think I was confused by the ranges in pdl, but those are still a single python values to the interpreter.
I've added the change in this pr: https://github.com/xdslproject/xdsl/pull/4247
Thanks!

PR #4163, opened 2025-04-02
core: Lazily include ImplicitBuilder in Operation __post_init__

EdmundGoodman:
Refactor `ImplicitBuilder` to lazily add its modified `__post_init__` only when it is in use within a context.

The `ImplicitBuilder` implementation can be further improved by storing the stack as a linked list in object properties instead of using the singleton `_ImplicitBuilderStack`, but this is a seperate issue to the lazy enablement of `ImplicitBuilder`s on operations.


EdmundGoodman:
Trace on constant folding before:
image

Trace on constant folding after:
image

This saves about 800ns per operation instantiation outside implicit builders.

xdsl/builder.py
```diff
@@ -365,6 +367,9 @@ def __exit__(
         __traceback: TracebackType | None,
     ) -> bool | None:
         type(self)._stack.pop(self._builder)
+        if not type(self)._stack.stack:
+            assert (old_post_init := type(self)._old_post_init)
+            Operation.__post_init__ = old_post_init  # pyright: ignore[reportAttributeAccessIssue]
```
EdmundGoodman, line 372: As far as I can tell, `Callable[[Operation], None]` is the correct type hint for a variable storing a method of an object with no arguments other than `self` and no return values. However, pyright still complains. Any ideas on better fixes appreciated :)

xdsl/builder.py
```diff
@@ -340,7 +340,7 @@ class ImplicitBuilder(contextlib.AbstractContextManager[tuple[BlockArgument, ...
     """
 
     _stack: ClassVar[_ImplicitBuilderStack] = _ImplicitBuilderStack()
```
superlopuh, line 342: why is there still a stack? I expected an optional parent object to replace the current one with on exit

superlopuh: We can improve it in a different PR

tests/test_implicit_builder.py
superlopuh, line 1: mostly a nit, but I would find a multi-line string much easier to read here, and I'm not sure we benefit much from the name hint

tests/test_implicit_builder.py
```diff
@@ -0,0 +1,60 @@
...
+
+@pytest.fixture
+def dialect_context() -> Context:
+    """Fixture that creates a Context with Arith and Builtin dialects loaded.
+
+    Returns:
+        Context: An MLIR Context with required dialects.
+    """
+    ctx = Context()
+    ctx.load_dialect(arith.Arith)
```
superlopuh, line 18: not used?

superlopuh (approved):
Nice

tests/test_implicit_builder.py
EdmundGoodman: Removed

tests/test_implicit_builder.py
EdmundGoodman: Changed to multi-line string and removed name hint

xdsl/builder.py
superlopuh: It's hard to help here without the error message :)

EdmundGoodman: Sorry - should've given more context. This is all pyright gives:

```
Cannot assign to attribute "__post_init__" for class "type[Operation]"
  Type "(Operation) -> None" is not assignable to type "(self: Operation) -> None"
    Missing keyword parameter "self"
      Position-only parameter mismatch; parameter "self" is not position-only
      Position-only parameter mismatch; expected 1 but received 0Pylance[reportAttributeAccessIssue](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportAttributeAccessIssue)
(variable) def old_post_init(Operation) -> None
```

superlopuh: Interesting, seems vaguely bogus to me

superlopuh: I think this pyright ignore is justified

math-fehr: Until we require Python 3.10, I think you need to use a Protocol for this:
```
class OperationMethod(Protocol):
    def __call__(self: Operation) -> None:
        ...
```

math-fehr (approved):
Looks good to me, from what I understand this will give us some performance back?

EdmundGoodman:
> Looks good to me, from what I understand this will give us some performance back?

Yes, all operation creation outside implicit builders will be faster (at a guesstimate by about 15% given its halving the time taken for `__post_init__` which is itself about a third of operation creation)

xdsl/builder.py
math-fehr: > Interesting, seems vaguely bogus to me

It's not actually bogus, it's just that `Callable` cannot express named arguments, so you wouldn't be able to typecheck `func(self=...)`

superlopuh: I'm not sure that this is worth it TBH

compor (approved):

alexarice (commented):
Do we have benchmarks backing up that this makes an observable difference in runtime? As far as I can see this saves a single integer equality check.

alexarice:
I actually have a further question, at the moment does `_override_operation_post_init()` get run every time `xdsl.builder` is imported? i.e. are we currently making a massive stack of `__post_init__` functions?

alexarice:
I didn't actually see the benchmarks above, but I would be interested in seeing how just having the `ImplicitBuilder` check be static in `Operation.__post_init__` (or even `Operation.__init__` assuming this always gets called (?)) compares.

superlopuh:
Why not just merge this PR and see it in the nightly benchmark runs?

superlopuh:
I would say that now that this has been measured and discussed we can leave the experiments with different ImplicitBuilder checks for later. I feel positively about not having ImplicitBuilder mechanisms imposed on all Operation creations, as it's a non-essential mechanism.

alexarice:
Fine, I give in.

PR #4165, opened 2025-04-02
tool: (xdsl-tblgen) fix AnyAttrOf generation

Jimmy2027:
re-opened #4160:
fixes issue mentioned in #4158, without this the AnyAttrOf generation is broken:
```python
@irdl_op_definition
class EmitC_ConstantOp(IRDLOperation):
    """Constant operation"""

    name = "emitc.constant"

    value = prop_def(
    AnyOf(
        BaseAttr(EmitC_OpaqueAttr),AnyAttr()
        )
    )
    )

    v12 = result_def(AnyAttr())
```

with this change the [emitc](https://mlir.llvm.org/docs/Dialects/EmitC/) dialect can be generated without errors.

### DIFF
--- a/xdsl/tools/xdsl_tblgen.py
+++ b/xdsl/tools/xdsl_tblgen.py
@@ -339,8 +339,7 @@ class TblgenLoader:
         if "AnyAttrOf" in rec.superclasses:
             return textwrap.dedent(f"""
             AnyOf(
-                {",".join(self._resolve_prop_constraint(x["def"]) for x in rec["allowedAttributes"])}
-                )
+                [{",".join(self._resolve_prop_constraint(x["def"]) for x in rec["allowedAttributes"])}]
             )
             """)
 
@@ -511,6 +510,7 @@ def cull_json(output_file: IO[str] | None, loader: TblgenLoader):
         "baseAttr",
         "def",
         "name",
+        "allowedAttributes",
     }
 
     def cull_field(js_in: dict[str, Any]) -> dict[str, Any]:
--- a/tests/xdsl_tblgen/test.json
+++ b/tests/xdsl_tblgen/test.json
@@ -8,6 +8,8 @@
     ],
     "Op": [
       "Test_AndOp",
+      "Test_AnyAttrOfOp",
+      "Test_AnyAttrOfSingleOp",
       "Test_AnyOp",
       "Test_AssemblyFormat",
       "Test_AssemblyFormatLong",
@@ -118,6 +120,24 @@
     "bitwidth": 32,
     "summary": "32-bit signless integer"
   },
+  "I32Attr": {
+    "!name": "I32Attr",
+    "!superclasses": [
+      "Constraint",
+      "AttrConstraint",
+      "Attr",
+      "TypedAttrBase",
+      "SignlessIntegerAttrBase",
+      "TypedSignlessIntegerAttrBase"
+    ],
+    "baseAttr": null,
+    "summary": "32-bit signless integer attribute",
+    "valueType": {
+      "def": "I32",
+      "kind": "def",
+      "printable": "I32"
+    }
+  },
   "Index": {
     "!name": "Index",
     "!superclasses": [
@@ -150,6 +170,22 @@
     "bitwidth": 64,
     "summary": "64-bit signed integer"
   },
+  "StrAttr": {
+    "!name": "StrAttr",
+    "!superclasses": [
+      "Constraint",
+      "AttrConstraint",
+      "Attr",
+      "StringBasedAttr"
+    ],
+    "baseAttr": null,
+    "summary": "string attribute",
+    "valueType": {
+      "def": "NoneType",
+      "kind": "def",
+      "printable": "NoneType"
+    }
+  },
   "Test_AndOp": {
     "!name": "Test_AndOp",
     "!superclasses": [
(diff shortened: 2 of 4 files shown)


xdsl/tools/xdsl_tblgen.py
```diff
@@ -339,8 +339,7 @@ def _resolve_prop_constraint(self, rec: TblgenRecord | str) -> str:
         if "AnyAttrOf" in rec.superclasses:
             return textwrap.dedent(f"""
             AnyOf(
-                {",".join(self._resolve_prop_constraint(x["def"]) for x in rec["allowedAttributes"])}
-                )
+                ({",".join(self._resolve_prop_constraint(x["def"]) for x in rec["allowedAttributes"])})
```
alexarice, line 342: Will this work if there is only 1 entry? (I realise that this is a degenerate case and the corresponding `AnyOf` might also be broken

alexarice (commented):
Thanks for the fix, is it possible to add an entry to `test.td` which uses `AnyAttrOf` and if possible regenerate `test.json` with the command at the top of `test.td` (I'm happy to do this if you don't have easy access to `llvm-tblgen` etc)

Jimmy2027:
> Thanks for the fix, is it possible to add an entry to `test.td` which uses `AnyAttrOf` and if possible regenerate `test.json` with the command at the top of `test.td` (I'm happy to do this if you don't have easy access to `llvm-tblgen` etc)

sure will add that tomorrow

xdsl/tools/xdsl_tblgen.py
Jimmy2027: I think that should also work, will try with a testcase

superlopuh: I would expect that it would with a list and not with a tuple, but it makes more sense to me to do the len check and skip the AnyOf if there's only one element anyway.

Jimmy2027:
I was trying to re-use this test op from the llvm repo:
https://github.com/llvm/llvm-project/blob/73e8d67a200beaa554a72cdd50e4d1a5a55caf69/mlir/test/lib/Dialect/Test/TestOps.td#L174

But this fails because the generated test.json does not contain the "allowedAttributes" key:
https://github.com/xdslproject/xdsl/pull/4165/files#diff-8f5a00af4ab8bf84c22fbc9b2fc1850cfb4b8359f14f6d1a13c997c932dee368R252

I'm not sure if I'm doing something wrong here of if this is intended

alexarice:
It seems `AnyAttrOf` doesn't actually save its `allowedAttrs` in MLIR. @math-fehr is this a bug?

math-fehr:
> `allowedAttrs`

In what sense?
Doesn't it saves them in `allowedAttributes` ?

alexarice:
https://github.com/llvm/llvm-project/blob/73e8d67a200beaa554a72cdd50e4d1a5a55caf69/mlir/include/mlir/IR/CommonAttrConstraints.td#L177

nope

math-fehr:
> https://github.com/llvm/llvm-project/blob/73e8d67a200beaa554a72cdd50e4d1a5a55caf69/mlir/include/mlir/IR/CommonAttrConstraints.td#L177
> 
> nope

But what does this do then? https://github.com/llvm/llvm-project/blob/73e8d67a200beaa554a72cdd50e4d1a5a55caf69/mlir/include/mlir/IR/CommonAttrConstraints.td#L187C1-L187C49

alexarice:
I might be an idiot (don't look at the git blame for that). In that case I have no idea why that information is not in the json file

Jimmy2027:
I'm pretty sure the issue is with the new entry in `test.td` I added. The conversion of the emitC dialect works fine (the resulting json contains the `allowedAttributes` key for the `AnyAttrOf` attributes) when running this:
`llvm-tblgen --dump-json mlir/Dialect/EmitC/IR/EmitC.td`

Jimmy2027:
hmm well I re-ran the `llvm-tblgen` command and now the test.json contains the `allowedAttributes`... along with many other changes. Not sure what changed, maybe I rebuilt llvm in the meantime, now I am on `LLVM version 21.0.0git41be5bbb`. If that works for you I can update `test.json` on main and put the changes in a separate commit to keep the changes related to the `AnyAttrOf` issues separate.

math-fehr:
> I might be an idiot (don't look at the git blame for that). In that case I have no idea why that information is not in the json file

Lmao, I won't blame you for not remembering having writting this.
Maybe was it because the MLIR version used was before your change? No clue, but if it works now everything is all good!

alexarice:
@math-fehr I have a feeling I might have been looking at a local checkout of llvm without this change then just blindly found the same line on github.

@Jimmy2027 we're on 20.1.1 but I think that should have this change.

alexarice:
`test.json` should be run through `xdsl-tblgen -c` to avoid adding an 18000 line json to the repository. I have a feeling this might be the actual problem, with this stage removing the `allowedAttributes` field

alexarice:
Yep `allowedAttributes` should be added to the `required_attributes` dictionary.

Jimmy2027:
ah my bad, didn't remember the command I ran last week, I can fix that here as well

alexarice (approved):
Looks good to me, sorry for all the back and forth on this PR and thanks for the fix.

Jimmy2027:
no worries :) was actually quite insightful for me!

xdsl/tools/xdsl_tblgen.py
Jimmy2027: for the AnyAttrOf with 1 entry, I'm not sure if adding a hotfix in the conversion to json that removes the AnyOf is not confusing. Probably better to fix the TableGen file instead

PR #4167, opened 2025-04-03
dialects (arm): Add NEON fmla instruction (mixed: vector/scalar)

emmau678:
Floating-point fused Multiply-Add to accumulator (mixed: first source operand is a vector, second is a scalar. Destination is a vector)

This instruction multiplies the values in the first source operand by the second source operand, adds the accumulated value from the destination operand, and writes the resulting values to the destination.

This instruction is similar to FMUL and so this PR follows the same approach as https://github.com/xdslproject/xdsl/pull/4122.


superlopuh (approved):
Flawless 🚀

compor (approved):

PR #4169, opened 2025-04-03
misc: Make dialect object printing optional

mamanain:
We don't need to print it in the torch dialect.


compor:
I haven't followed from the start what are the conventions here, and there is no discussion on Zulip that I can find.

Are we shoving features in without tests (i.e., `tests/filecheck/utils/codegen/simple.py`)?
What is the process/etiquette? Is it for just @superlopuh's eyes?

mamanain:
@compor I just YOLO sent this but we discussed with Sasha that it's better to do it in another way. I think I'll close for now. I'm just trying to complete couple of features until the call on dynamic networks tomorrow.

compor:
If it is YOLO mode, that's fine; that's why I'm asking. I'll move out of the "way" then :-)

PR #4172, opened 2025-04-03
dialects: (bigint) Add operations for bigint dialect

EdmundGoodman:
Add operations for bigint dialect

### DIFF
--- a/xdsl/dialects/bigint.py
+++ b/xdsl/dialects/bigint.py
@@ -1,7 +1,25 @@
 """Dialect for arbitrary-precision integers."""
 
-from xdsl.ir import Dialect, ParametrizedAttribute, TypeAttribute
-from xdsl.irdl import irdl_attr_definition
+import abc
+
+from xdsl.dialects.builtin import f64, i1
+from xdsl.ir import (
+    Attribute,
+    Dialect,
+    Operation,
+    ParametrizedAttribute,
+    SSAValue,
+    TypeAttribute,
+)
+from xdsl.irdl import (
+    IRDLOperation,
+    irdl_attr_definition,
+    irdl_op_definition,
+    operand_def,
+    result_def,
+    traits_def,
+)
+from xdsl.traits import Commutative, Pure, SameOperandsAndResultType
 
 
 @irdl_attr_definition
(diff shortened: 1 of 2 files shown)


xdsl/dialects/bigint.py
```diff
@@ -1,7 +1,25 @@
 """Dialect for arbitrary-precision integers."""
```
superlopuh, line 1: So it turns out that LLVM's APInt or arbitrary-precision integer is different in that it holds both the value and the bitwidth (so you can have an infinite number of representations of the value 0, one for each bitwidth, if I understand correctly). In that way, this dialect is not for arbitrary-precision integers, but specifically as a way to model methods of Python's int class. This is especially the case if we add a method per `<`, `<=`, `>`, `>=`, which are otherwise redundant. It feels worth expanding on the description here to explain all this.

superlopuh (commented):
Overall LGTM modulo doc comment, @math-fehr said he wanted to review also so let's wait for him

xdsl/dialects/bigint.py
```diff
@@ -15,9 +33,262 @@ class BigIntegerType(ParametrizedAttribute, TypeAttribute):
...
+    name = "bigint.floordiv"
+
+    traits = traits_def(
+        Pure(),
+        SameOperandsAndResultType(),
+    )
+
+
+@irdl_op_definition
+class ModOp(BinaryOperation):
```
math-fehr, line 99: Can you mention that the sign of the result takes the sign of the divisor?
This is different than in other languages such as Javascript for instance.

xdsl/dialects/bigint.py
```diff
@@ -15,9 +33,262 @@ class BigIntegerType(ParametrizedAttribute, TypeAttribute):
...
+
+    traits = traits_def(
+        Pure(),
+        Commutative(),
+        SameOperandsAndResultType(),
+    )
+
+
+@irdl_op_definition
+class DivOp(IRDLOperation):
```
math-fehr, line 172: If this is supposed to represent Python division, could you rename it to `FloorDivOp`, as it rounds to -infinity?
Just to make this explicit, as it's easy to forget.

math-fehr (approved):
Very cool!

xdsl/dialects/bigint.py
compor: Regarding `APInt` that is correct, any equality checks the bit width value first.

Regarding this PR, I think more documentation would benefit those that are not tagged in these PRs or don't fully understand the utility of this/where this is going. IMHO, more input the better it is, even if this can be provided sporadically.

xdsl/dialects/bigint.py
EdmundGoodman: There is already a `FloorDiv` operation which implements `__floordiv__` / `//` returning an integer, whereas this one does `__div__` / `/` and returns a float64 -- so I've just added a docstring to make it more clear.

math-fehr: Oh, my bad I indeed didn't see the result type!

math-fehr: I am wondering if we should change the name of it to avoid, but none come to my mind besides maybe `DivToFOp`?

superlopuh: `DivFOp`?

superlopuh: But TBH I think keeping the name same as the Python name also makes sense to me

PR #4173, opened 2025-04-03
misc: Print properties, traits and attributes in codegen

mamanain:
We need to support traits for the torch dialect, decided to do couple more fields while I'm at it.

### DIFF
--- a/tests/filecheck/utils/codegen/simple.py
+++ b/tests/filecheck/utils/codegen/simple.py
@@ -6,6 +6,7 @@ from xdsl.dialects.builtin import (
     Float32Type,
     IndexType,
     IntAttr,
+    IntegerAttr,
     IntegerType,
     NoneType,
     Signedness,
@@ -22,6 +23,7 @@ from xdsl.irdl import (
     OptOperandDef,
     ParamAttrConstraint,
     ParamAttrDef,
+    PropertyDef,
     ResultDef,
     SameVariadicOperandSize,
     VarOperandDef,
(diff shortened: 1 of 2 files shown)


tests/filecheck/utils/codegen/simple.py
```diff
@@ -236,11 +276,36 @@
...
+# CHECK-NEXT:      name = "test.traits"
+# CHECK-EMPTY:
+# CHECK-NEXT:      traits = traits_def(ConstantLike(), Pure())
+
+# CHECK:       @irdl_op_definition
+# CHECK-NEXT:  class Test_AttributesOp(IRDLOperation):
+# CHECK-NEXT:      name = "test.attributes"
+# CHECK-EMPTY:
+# CHECK-NEXT:      some_attr = attr_def(AnyAttr(), attr_name="attr")
+# CHECK-NEXT:      operandSegmentSizes = attr_def(BaseAttr(DenseArrayBase))
```
superlopuh, line 307: this feels weird to me, why add it as a property?

mamanain: wym? what is the alternative?

superlopuh: `irdl_options = [AttrSizedOperandSegments()]`

superlopuh: In fact I would expect this to raise an error because it's supposed to be handled automatically by the irdl annotation CC @math-fehr

mamanain: I'm sorry, how are attributes related to `irdl_options`? I think I still don't get it.

superlopuh: this attribute is created automatically when that option is specified, for example here:
https://github.com/xdslproject/xdsl/blob/main/tests/irdl/test_declarative_assembly_format.py#L898

So to specify is manually is something weird waiting to happen, as it could clash with the option. I would expect some stage in the system to let the user know that they're doing something error-prone, and forbid it. It feels off to me to add it as a test case here, in any case.

mamanain: Oh, ok. I can just remove it from the test then. I just wanted some attribute to plug in here.

superlopuh: I think one attribute without the mapping would be great, I'd recommend something simple :)

PR #4180, opened 2025-04-04
dialects: (pdl-interp) Add manual parse and print functions

jumerckx:
For `pdl_interp.func`, `pdl_interp.create_operation`, and `pdl_interp.get_results`, since these have issues when using the `assembly_format` directly.


tests/filecheck/mlir-conversion/with-mlir/dialects/pdl_interp/pdl_interp_ops.mlir
```diff
@@ -46,77 +43,74 @@
...
     %3 = pdl_interp.get_result 0 of %2
-    %4 = "pdl_interp.get_results"(%2) : (!pdl.operation) -> !pdl.range<value>
+    %4 = pdl_interp.get_results of %2 : !pdl.range<value>
     pdl_interp.replace %arg4 with (%4 : !pdl.range<value>)
     pdl_interp.finalize
-  }) : () -> ()
+  }
 }
 
-//CHECK-NEXT:   builtin.module {
```
jumerckx, line 64: Seems like I'd done it the wrong way 😅

xdsl/dialects/pdl_interp.py
```diff
@@ -273,6 +276,27 @@ def __init__(
...
+    def parse(cls, parser: Parser) -> GetResultsOp:
+        index = parser.parse_optional_integer()
+        if index is not None:
+            index = IntegerAttr.from_int_and_width(index, 32)
+        parser.parse_characters("of")
+        input_op = parser.parse_operand()
+        parser.parse_punctuation(":")
+        result_type = parser.parse_type()
+        assert isa(result_type, ValueType) or isa(result_type, RangeType[ValueType])
+        return GetResultsOp(index, input_op, result_type)
```
superlopuh, line 289: This is the wrong place to do type checking, we expect this to happen during verification. I'd recommend something like this:
```suggestion
        return GetResultsOp.build(operands=(input_op,), properties={"index": index}, result_types=(result_type,))
```

superlopuh: Another approach would be to raise a parsing error
```suggestion
        pos = parser.pos
	result_type = parser.parse_type()
	if not isa(result_type, ValueType | RangeType[ValueType]):
	    parser.raise_error(f"Expected value type or range of value types, got {result_type}", at_position: pos)
        return GetResultsOp(index, input_op, result_type)
```

xdsl/dialects/pdl_interp.py
```diff
@@ -560,6 +584,105 @@ def __init__(
             },
         )
 
+    @classmethod
+    def parse(cls, parser: Parser) -> CreateOperationOp:
+        def _parse_attribute() -> tuple[StringAttr, SSAValue]:
+            attrname = parser.parse_attribute()
+            assert isinstance(attrname, StringAttr)
```
superlopuh, line 591: similarly for the asserts in this function, either we should let the verification happen after the whole file is parsed, or we should raise a parse error

xdsl/dialects/pdl_interp.py
jumerckx: Sorry, took a while. I opted to not add the parsing errors.

xdsl/dialects/pdl_interp.py
```diff
@@ -273,6 +276,30 @@ def __init__(
...
+        )
+
+    def print(self, printer: Printer):
+        if self.index is not None:
+            printer.print(" ")
+            printer.print(self.index.value.data)
+        printer.print(" of ")
+        printer.print_operand(self.input_op)
+        printer.print(" : ")
+        printer.print(self.value.type)
```
superlopuh, line 301: More of a nit, but for the future I'd recommend specific APIs that aren't doing dispatch based on type:
```suggestion
        if self.index is not None:
            printer.print_string(" ", indent=0)
            printer.print_operand(self.index.value.data)
        printer.print_string(" of ", indent=0)
        printer.print_operand(self.input_op)
        printer.print_string(" : ", indent=0)
        printer.print_attribute(self.value.type)
```
The indent=0 removes the check for newlines in the input string and prints it as-is, instead of calculating indentation levels.

xdsl/dialects/pdl_interp.py
```diff
@@ -560,6 +587,110 @@ def __init__(
...
+                    delimiter=Parser.Delimiter.NONE,
+                    parse=lambda: parser.parse_operand(),
+                )
+                parser.parse_punctuation(":")
+                parser.parse_comma_separated_list(
+                    delimiter=Parser.Delimiter.NONE,
+                    parse=lambda: parser.parse_type(),
+                )
+                parser.parse_punctuation(")")
+            return values
```
superlopuh, line 612: this pattern redefines the helper functions each time that an operation of this class is parsed, I'd recommend moving them out to private `@static` methods on the type instead, to avoid redundant work.

xdsl/dialects/pdl_interp.py
```diff
@@ -560,6 +587,110 @@ def __init__(
...
+        name = parser.parse_attribute()
+
+        input_operands = _parse_input_list()
+
+        input_attribute_names = None
+        input_attributes = None
+        attributes = parser.parse_optional_comma_separated_list(
+            delimiter=Parser.Delimiter.BRACES, parse=_parse_attribute
+        )
+        if attributes:
```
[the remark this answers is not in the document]
superlopuh: As a code style, we check explicitly for `is not None` to avoid the ambiguity of falseyness of the non-None object.

xdsl/dialects/pdl_interp.py
```diff
@@ -560,6 +587,110 @@ def __init__(
...
+            printer.print(" = ")
+            printer.print_operand(value[1])
+
+        printer.print(" ")
+        printer.print(self.constraint_name)
+        if self.input_operands:
+            _print_input_list(self.input_operands)
+        else:
+            printer.print("() ")
+        # printer.print(" ") # TODO: should we mimick the reference by including an extra space?
```
superlopuh, line 676: what does this mean? Does MLIR print an extra space here? I would drop it.

jumerckx:
Thanks a lot for the review, Sasha! I believe all comments have been addressed.

superlopuh (approved):
Nice!

PR #4184, opened 2025-04-04
dialects (arm): add destination reg as operand to fmla instruction

emmau678:
Modify the ```fmla``` op to include the destination register (accumulator) as an operand. This is needed to represent dependencies between a series of these ops in the matmul kernels and therefore will be required for register allocation.


compor (approved):
the doc building has choked, but it's unrelated

xdsl/dialects/arm_neon.py
```diff
@@ -252,20 +252,25 @@ class DSSFmlaVecScalarOp(ARMInstruction):
     """
 
     name = "arm_neon.dss.fmla"
-    d = result_def(NEONRegisterType)
+    res = result_def(NEONRegisterType)
+    d = operand_def(NEONRegisterType)
```
superlopuh, line 256: These have to be the same type. I would add this as a constraint and to the documentation, as well as a test verifying this

superlopuh: See `RdRsAccumulatingFloatOperation`

emmau678: should this be a unit test or filecheck?

compor: ah right, I missed this. I think filecheck might be a bit easier

superlopuh (approved):

PR #4185, opened 2025-04-04
core: Evaluate string annotations in rewrite patterns

math-fehr:
The op_type_rewrite_pattern decorator fails to apply on methods under certain conditions downstream due to the fact that the type annotations are not evaluated, and passed as strings.

Adding `eval_str=True` in the signature inspector will ensure that the type annotations are evaluated.


math-fehr:
I couldn't actually reproduce this in tests, but downstream it does have an effect when `from __future__ import annotations` is present.

superlopuh (approved):

PR #4186, opened 2025-04-04
core: Allow nesting of optional groups in declarative assembly format

math-fehr:
Stacked PRs:
 * #4187
 * __->__#4186

--- --- ---

### core: Allow nesting of optional groups in declarative assembly format

Previously, nesting optional groups would raise asserts in the
declarative assembly format generated parser, as nested groups would not
set their operands and results as empty.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -21,6 +21,7 @@ from xdsl.dialects.builtin import (
     IntegerType,
     MemRefType,
     ModuleOp,
+    StringAttr,
     UnitAttr,
 )
 from xdsl.dialects.test import Test, TestType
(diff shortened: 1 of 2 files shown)


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -2670,6 +2671,44 @@ class OptionalGroupOp(IRDLOperation):
     check_equivalence(program, generic_program, ctx)
 
 
+@pytest.mark.parametrize(
+    "program, generic_program",
+    [
+        (
+            '%0 = "test.op"() : () -> i32\ntest.optional_optional_group with "prop" %0',
```
alexarice, line 2678: I believe you can omit the test.op and just have the `%0` be unbound

alexarice (approved):
Looks good, sorry for missing this on the original refactor

superlopuh (approved):

tests/irdl/test_declarative_assembly_format.py
math-fehr: Nope, tried it and it gave me an error.

PR #4187, opened 2025-04-04
dialects: Move PDL parser and printer to assembly format

math-fehr:
Stacked PRs:
 * __->__#4187

--- --- ---

### dialects: Move PDL parser and printer to assembly format

This moves the manually written PDL parser and printer methods to the
nice assembly format. The format was taken from the PDL dialect
MLIR documentation.

### DIFF
--- a/xdsl/dialects/pdl.py
+++ b/xdsl/dialects/pdl.py
@@ -24,7 +24,6 @@ from xdsl.ir import (
 from xdsl.irdl import (
     AttrSizedOperandSegments,
     IRDLOperation,
-    Operand,
     ParameterDef,
     base,
     irdl_attr_definition,
@@ -269,6 +268,8 @@ class AttributeOp(IRDLOperation):
     value_type = opt_operand_def(TypeType)
     output = result_def(AttributeType)
 
+    assembly_format = "(`:` $value_type^)? (`=` $value^)? attr-dict-with-keyword"
+
     def verify_(self):
         if self.value is not None and self.value_type is not None:
             raise VerifyException(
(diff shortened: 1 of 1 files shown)


xdsl/dialects/pdl.py
```diff
@@ -721,6 +680,11 @@ class ReplaceOp(IRDLOperation):
 
     irdl_options = [AttrSizedOperandSegments()]
 
+    assembly_format = (
+        "$op_value `with` ` ` "
+        "(`(` $repl_values^ `:` type($repl_values) `)`)? ($repl_operation^)? attr-dict"
```
alexarice, line 685: Don't believe `repl_operation` needs to be in an optional group but understand if you want to keep this the same as MLIR

xdsl/dialects/pdl.py
```diff
@@ -853,7 +788,7 @@ class RewriteOp(IRDLOperation):
     """
 
     name = "pdl.rewrite"
-    root = opt_operand_def(OperationType)
+    root = opt_operand_def(OperationType())
```
alexarice, line 791: I thought we changed this to not matter?

alexarice (approved):
Nice

superlopuh (approved):

xdsl/dialects/pdl.py
math-fehr: Oh yeah, I really just copy pasted it so I didn't look at it properly. I'll change it, it doesn't *need* to be exactly the same

xdsl/dialects/pdl.py
math-fehr: Removed the change, thanks I forgot about it

PR #4188, opened 2025-04-04
misc: remove quotes around "operandSegmentSizes" in filecheck

superlopuh:
Just a find and replace.


math-fehr (approved):
I'm approving it because I don't really mind, but I would feel that these kind of changes probably create more conflicts/annoyance that they solve problems?
For instance, some people might wonder why some operations have some properties with strings and some without.

superlopuh:
I can open another PR making things consistent, this particular property was getting in the way of a find/replace refactor working properly.

PR #4191, opened 2025-04-05
dialects: Add vector.insert and vector.extract operations

math-fehr:
These operations are defined in MLIR and are meant to replace vector.insertelement and vector.extractelement.


math-fehr:
I reused the MLIR test files for the filecheck files.

xdsl/dialects/vector.py
```diff
@@ -282,6 +291,148 @@ def get(mask_operands: list[Operation | SSAValue]) -> CreateMaskOp:
...
+        result.append(pos)
+    return result
+
+
+@irdl_op_definition
+class ExtractOp(IRDLOperation):
+    name = "vector.extract"
+
+    _T = Annotated[Attribute, ConstraintVar("T")]
+    _V = Annotated[VectorType[_T], ConstraintVar("V")]
```
superlopuh, line 319: Can you make these ConstraintVarConstrs instead, please? We're migrating away from this API since it’s not officially supported by Pyright

superlopuh:
This would be much easier to review as a pair of PRs, one for ops with custom syntax, and one for verifiers

tests/filecheck/mlir-conversion/with-mlir/dialects/vector/insert.mlir
```diff
@@ -0,0 +1,36 @@
+// RUN: xdsl-opt --print-op-generic %s | mlir-opt --mlir-print-op-generic --allow-unregistered-dialect | xdsl-opt | filecheck %s
```
superlopuh, line 1: Let's add a custom syntax chack also?

xdsl/dialects/vector.py
```diff
@@ -282,6 +291,148 @@ def get(mask_operands: list[Operation | SSAValue]) -> CreateMaskOp:
         )
 
 
+def _get_mixed_position(
```
superlopuh, line 294: I feel like we have a few of these around, maybe could warrant a helper to be shared across dialect? I remember being asked about the right way to do this a couple of times.

xdsl/dialects/vector.py
```diff
@@ -282,6 +291,148 @@ def get(mask_operands: list[Operation | SSAValue]) -> CreateMaskOp:
...
+    next_dynamic_index = 0
+    result: list[SSAValue | int] = []
+    for pos in static_position:
+        assert isinstance(pos, int)
+        if pos == DYNAMIC_INDEX:
+            result.append(dynamic_position[next_dynamic_index])
+            next_dynamic_index += 1
+            continue
+        result.append(pos)
+    return result
```
superlopuh, line 311: This is more of a nit, but I feel compelled to rewrite this a little:
```suggestion
    result = list[SSAValue | int](static_position)
    dynamic_position_indices = tuple(i for i, pos in enumerate(static_position) if pos == DYNAMIC_INDEX)
    for i, pos in zip(dynamic_position_indices, dynamic_position, strict=True):
        result[i] = pos
    return result
```

math-fehr:
Splitted

PR #4192, opened 2025-04-07
transformations: vector load to ptr

qaco:
Lowering of ```vector.load``` to the ```ptr``` dialect. The memory indexing is done in the ```memref``` world (using ```subview```) rather than in the ```ptr``` world. This way, the low-level pointer arithmetic, which is error-prone and implementation-dependent, is confined to the lowering passes of the ```memref``` dialect.

### DIFF
--- a/tests/filecheck/transforms/convert_vector_to_ptr.mlir
+++ b/tests/filecheck/transforms/convert_vector_to_ptr.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt -p convert-vector-to-ptr %s | filecheck %s
+// RUN: xdsl-opt -p convert-vector-to-ptr --split-input-file %s | filecheck %s
 
 %m = "test.op"(): () -> memref<16xf32>
 %i = arith.constant 0: index
@@ -7,8 +7,25 @@
 // CHECK:       builtin.module {
 // CHECK-NEXT:    %m = "test.op"() : () -> memref<16xf32>
 // CHECK-NEXT:    %i = arith.constant 0 : index
-// CHECK-NEXT:    %v = affine.apply affine_map<(d0) -> (d0)> (%i)
+// CHECK-NEXT:    %v = affine.apply affine_map<(d0) -> ((d0 * 4))> (%i)
 // CHECK-NEXT:    %v_1 = ptr_xdsl.to_ptr %m : memref<16xf32> -> !ptr_xdsl.ptr
 // CHECK-NEXT:    %v_2 = ptr_xdsl.ptradd %v_1, %v : (!ptr_xdsl.ptr, index) -> !ptr_xdsl.ptr
 // CHECK-NEXT:    %v_3 = ptr_xdsl.load %v_2 : !ptr_xdsl.ptr -> vector<8xf32>
 // CHECK-NEXT:  }
+
+// -----
+
+%m0 = "test.op"(): () -> memref<2x8xf32>
+%i0 = arith.constant 0: index
+%j0 = arith.constant 0: index
+%v0 = vector.load %m0[%i0,%j0]: memref<2x8xf32>, vector<8xf32>
+
+// CHECK:      builtin.module {
+// CHECK-NEXT:   %m0 = "test.op"() : () -> memref<2x8xf32>
+// CHECK-NEXT:   %i0 = arith.constant 0 : index
+// CHECK-NEXT:   %j0 = arith.constant 0 : index
+// CHECK-NEXT:   %v0 = affine.apply affine_map<(d0, d1) -> (((d0 * 32) + (d1 * 4)))> (%i0, %j0)
+// CHECK-NEXT:   %v0_1 = ptr_xdsl.to_ptr %m0 : memref<2x8xf32> -> !ptr_xdsl.ptr
+// CHECK-NEXT:   %v0_2 = ptr_xdsl.ptradd %v0_1, %v0 : (!ptr_xdsl.ptr, index) -> !ptr_xdsl.ptr
+// CHECK-NEXT:   %v0_3 = ptr_xdsl.load %v0_2 : !ptr_xdsl.ptr -> vector<8xf32>
+// CHECK-NEXT: }
--- a/tests/filecheck/transforms/convert_vector_to_ptr.mlir
+++ b/tests/filecheck/transforms/convert_vector_to_ptr.mlir
@@ -8,6 +8,6 @@
 // CHECK-NEXT:    %m = "test.op"() : () -> memref<16xf32>
 // CHECK-NEXT:    %i = arith.constant 0 : index
 // CHECK-NEXT:    %v = memref.subview %m[%i] [8] [1] : memref<16xf32> to memref<8xf32>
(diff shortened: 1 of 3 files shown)


tests/filecheck/transforms/convert_vector_to_ptr.mlir
```diff
@@ -0,0 +1,13 @@
...
+
+%m = "test.op"(): () -> memref<16xf32>
+%i = arith.constant 0: index
+%v = vector.load %m[%i]: memref<16xf32>, vector<8xf32>
+
+// CHECK:       builtin.module {
+// CHECK-NEXT:    %m = "test.op"() : () -> memref<16xf32>
+// CHECK-NEXT:    %i = arith.constant 0 : index
+// CHECK-NEXT:    %v = memref.subview %m[%i] [8] [1] : memref<16xf32> to memref<8xf32>
+// CHECK-NEXT:    %v_1 = builtin.unrealized_conversion_cast %v : memref<8xf32> to !ptr_xdsl.ptr
```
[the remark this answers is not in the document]
superlopuh: sorry `ptr.to_ptr`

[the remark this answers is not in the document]
qaco: Oh yes thank you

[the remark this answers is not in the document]
qaco: Done !

xdsl/transforms/convert_vector_to_ptr.py
```diff
@@ -0,0 +1,65 @@
...
+        subview_type = memref.MemRefType(
+            element_type=element_type, shape=vector_int_shape
+        )
+        subview_op = memref.SubviewOp.get(
+            source=memory,
+            result_type=subview_type,
+            offsets=op.indices,
+            strides=static_strides,
+            sizes=vector_int_shape,
+        )
```
superlopuh, line 47: Won't this break for dynamic strides? Feels like we should either handle it or explicitly raise an error if any of the strides are None. We should definitely raise an error if strides is not None. On the other hand, I have a feeling that we should be using the more general API of affine maps, and calculating the offset from that, as the affine map is guaranteed to exist, and strides are not.

xdsl/transforms/convert_vector_to_ptr.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Build a pointer from the subview
+        cast_op = ptr.ToPtrOp(operands=[subview_op], result_types=[ptr.PtrType()])
+        # Load a vector from the pointer
+        load_op = ptr.LoadOp(operands=cast_op.results, result_types=[vector_ty])
```
superlopuh, line 52: this is a little tangential, but it would be great to add custom inits for both these ops. We'll add it at some point in the future, but then it'll break this rewrite, might as well just fix now. This is if you have the bandwidth, of course.

superlopuh: If you do, could you please add them in separate PRs?

xdsl/transforms/convert_vector_to_ptr.py
```diff
@@ -0,0 +1,65 @@
...
+        rewriter.replace_matched_op([subview_op, cast_op, load_op])
+
+
+@dataclass(frozen=True)
+class ConvertVectorToPtrPass(ModulePass):
+    name = "convert-vector-to-ptr"
+
+    def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
+        PatternRewriteWalker(
+            GreedyRewritePatternApplier([VectorLoadToPtr()]),
```
superlopuh, line 63: There's no need for a greedy applier for a single pattern
```suggestion
            VectorLoadToPtr(),
```

qaco: But other patterns are coming, right ?

superlopuh: I wasn't sure if you were adding them soon, if you think you are then let's add a trailing comma so that they each get added with a single line change?

xdsl/transforms/convert_vector_to_ptr.py
superlopuh: By default, the `__init__` for IRDLOperation is defined to be a sort of generic builder, where you pass some untyped parameters, and then it uses the operation definition to assign to the right places. We've been adding custom `__init__` methods to make the APIs a little neater. There are now lots of examples of custom `__init__` methods around the place, like in the backend dialects, arith, and so on, although some dialects don't yet have it. One day we'll generate those automatically but we can't do it yet :)

xdsl/transforms/convert_vector_to_ptr.py
qaco: Ok !

xdsl/transforms/convert_vector_to_ptr.py
qaco: Ok !

xdsl/transforms/convert_vector_to_ptr.py
superlopuh: I don't understand this question, are you proposing to ignore the strides information? I am proposing not to ignore it, and handle the case more generally if possible, using an affine.apply operation to calculate the offset for a given map, if there is one.

qaco: Ok thank you !

xdsl/transforms/convert_vector_to_ptr.py
qaco: I'll do that after finishing the vector -> ptr -> x86 flow, if that's okay with you! Just so I don't get sidetracked!

xdsl/transforms/convert_vector_to_ptr.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Build an affine.apply to compute the linearized offset
+        layout_map = memory_ty.layout
+        if isinstance(layout_map, NoneAttr):
+            layout_map = AffineMapAttr(affine.AffineMap.identity(len(op.indices)))
+        assert isinstance(layout_map, AffineMapAttr)
+        apply_op = affine.ApplyOp(
+            map_operands=op.indices,
+            affine_map=layout_map,
+        )
```
superlopuh, line 42: We actually have a helper for this:
```suggestion
        layout_map = memory_ty.get_affine_map_in_bytes()
        apply_op = affine.ApplyOp(
            map_operands=op.indices,
            affine_map=AffineMapAttr(layout_map),
        )
```
It would be good to add a test for multiple indices, I don't think that the current code that creates an identity map is correct for multiple dimensions.

qaco: Done

superlopuh (approved):
This is great! It would be nice to add a test with a different data type and vector size just to see that the affine map is generated correctly, but that's more of a nit.

qaco:
> This is great! It would be nice to add a test with a different data type and vector size just to see that the affine map is generated correctly, but that's more of a nit.

Done !

superlopuh (approved):

PR #4193, opened 2025-04-07
frontend: (pyast) Add support for bigint comparison operations in the PyAST frontend

EdmundGoodman:
Add support for bigint comparison operations in the PyAST frontend.


xdsl/frontend/pyast/type_conversion.py
```diff
@@ -191,13 +194,25 @@ def get_ir_type(
...
         if source_type_name not in self.type_names:
             return None
         source_type = self.type_names[source_type_name]
         if source_type not in self.type_registry:
             return None
-        return self.type_registry[source_type]()
+        ir_type_class = self.type_registry[source_type]
+        if metadata := get_args(ir_type_class):
+            assert len(metadata) >= 2
+            assert isinstance(metadata[1], TypeAttribute)
```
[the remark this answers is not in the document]
EdmundGoodman: This is guaranteed by type hints and asserted only for completeness

tests/filecheck/frontend/refactor/operations.py
```diff
@@ -46,3 +46,22 @@ def bar(x: float, y: float, z: float) -> float:
 
 p2.compile()
 print(p2.textual_format())
+
+
+p3 = FrontendProgram()
+p3.register_type(int, BigIntegerType)
+p3.register_type(bool, I1)
```
superlopuh, line 53: Why is this I1 and not i1?

superlopuh: I guess more generally why do we pass classes and not objects here?

EdmundGoodman: I think this was an arbitrary design decision, probably because then both arguments are Python `type`s. However, since the concrete class `i1` itself represents a type and it makes code neater I guess it makes sense to change it

superlopuh: Small separate PR?

superlopuh:
This seems to now have the same changes as the other PR, is that on purpose?

EdmundGoodman:
> This seems to now have the same changes as the other PR, is that on purpose?

Almost all of this PR is changing the type registration, which is now in #4211. When that is merged, this will only have a single change wrt main which is adding comparison support

PR #4195, opened 2025-04-07
backend: remove instruction set name from register types

superlopuh:
With this change, a register is uniquely identified by its register set name and index, allowing us to do things like disambiguating registers with multiple register names within a register set, as well as reason about overlapping registers in some simple cases.

### DIFF
--- a/xdsl/dialects/x86/register.py
+++ b/xdsl/dialects/x86/register.py
@@ -56,7 +56,7 @@ class GeneralRegisterType(X86RegisterType):
     name = "x86.reg"
 
     @classmethod
-    def instruction_set_name(cls) -> str:
+    def register_set_name(cls) -> str:
         return "x86"
 
     @classmethod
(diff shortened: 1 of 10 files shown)


xdsl/dialects/x86/register.py
```diff
@@ -110,8 +110,8 @@ class RFLAGSRegisterType(X86RegisterType):
     name = "x86.rflags"
 
     @classmethod
-    def instruction_set_name(cls) -> str:
-        return "x86"
+    def register_set_name(cls) -> str:
+        return "x86_rflags"
```
alexarice, line 114: are these standard official names for these things? I'm not really sure what this method is used for so not sure how to review

superlopuh: I have no idea if they even have official names, as far as I'm concerned the most important thing is that the names don't overlap if the registers don't overlap.

alexarice: Why wouldn't the operation name be sufficient for this?

superlopuh: I couldn't remember, so let's remove it, and add it back later when we need to.

alexarice (approved):
Seems reasonable to me

PR #4196, opened 2025-04-07
dialects: (vector) vector.fma expects vectors of floats

superlopuh:
Just tried to run our dialect tests through mlir-opt and failed, this constraint brings us in line with MLIR.


xdsl/dialects/vector.py
```diff
@@ -128,7 +130,9 @@ def get(source: Operation | SSAValue) -> BroadcastOp:
 class FMAOp(IRDLOperation):
     name = "vector.fma"
 
-    T: ClassVar = VarConstraint("T", base(VectorType))
+    T: ClassVar = VarConstraint(
+        "T", ParamAttrConstraint(VectorType, (AnyAttr(), AnyFloatConstr, AnyAttr()))
+    )
```
superlopuh, line 135: Probably best for @math-fehr to merge this first, then use that here/

https://github.com/xdslproject/xdsl/pull/4191/files#diff-b8dee9509fcf97e7a54db20b56724fbe2905745d98c46d0e54547672203e198aR1093-R1108

math-fehr (approved):

tests/filecheck/dialects/vector/vector_fma_invalid.mlir
```diff
@@ -0,0 +1,49 @@
+// RUN: xdsl-opt %s --verify-diagnostics --split-input-file | filecheck %s
```
alexarice, line 1: Is this file just the merging of the deleted files?

alexarice (commented):
If the purpose of this was to bring us in line with mlir-opt should we not have a interop test? (Maybe I've misunderstood the purpose of this PR)

tests/filecheck/dialects/vector/vector_fma_invalid.mlir
superlopuh: Yep and adding one new one for the fact that it's a float

superlopuh:
You're right that we should have at least one test for mlir interop, I added one, but it's hard to have a test for the same exclusion of types, so it's just a single roundtrip test for 2xf32.

PR #4204, opened 2025-04-08
core: TestSSAValue depends on pytest, which is an optional depedency

### LINKED ISSUE #4203: core: `TestSSAValue` depends on `pytest`, which is an optional depedency
### Description

Importing `xdsl.dialects.arith` with only the base dependencies results in a `ModuleNotFoundError`, as a result of the import chain: `xdsl.dialects.arith` -> `xdsl.pattern_rewriter` -> `xdsl.builder` -> `xdsl.rewriter` -> `xdsl.utils.test_value` -> `pytest`.

This probably wasn't caught by unit tests as they necessarily have `pytest` installed!

### Expected vs Actual Behaviour

When only the core dependencies are installed, you cannot import `xdsl.dialects.arith`. You should be able to import `xdsl.dialects.arith`.

### Steps to Reproduce and a Minimal Working Example (MWE)

Running:

```
uv sync  # This will remove all extra dependencies from the venv
uv run python3 -c "import xdsl.dialects.arith"
```

Yields the error:

```
  File "<string>", line 1, in <module>
  File "/Users/edjg/Desktop/thesis/code/xdsl-bench/xdsl/xdsl/dialects/arith.py", line 47, in <module>
    from xdsl.pattern_rewriter import RewritePattern
  File "/Users/edjg/Desktop/thesis/code/xdsl-bench/xdsl/xdsl/pattern_rewriter.py", line 13, in <module>
    from xdsl.builder import Builder, BuilderListener
  File "/Users/edjg/Desktop/thesis/code/xdsl-bench/xdsl/xdsl/builder.py", line 14, in <module>
    from xdsl.rewriter import BlockInsertPoint, InsertPoint, Rewriter
  File "/Users/edjg/Desktop/thesis/code/xdsl-bench/xdsl/xdsl/rewriter.py", line 17, in <module>
  File "/Users/edjg/Desktop/thesis/code/xdsl-bench/xdsl/xdsl/utils/test_value.py", line 3, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

### Additional Information

This is one of the causes of the failing CI jobs in xdsl-bench, as the `LoadDialects` microbenchmark loads the arith dialect.

EdmundGoodman:
Resolves #4203.

I'm not sure if this is a good permanent solution as it removes an `assert` statement and I don't have full context on what the original code was doing, but at the very least this makes it possible to `import xdsl.dialects.arith` with only the core dependencies.

### DIFF
--- a/xdsl/rewriter.py
+++ b/xdsl/rewriter.py
@@ -14,7 +14,6 @@ from xdsl.ir import (
     Region,
     SSAValue,
 )
-from xdsl.utils.test_value import TestSSAValue
 
 
 @dataclass(frozen=True)
@@ -202,8 +201,7 @@ class Rewriter:
                 *args[index + 1 :],
             )
         else:
-            assert isinstance(val, TestSSAValue)
-            new_value = TestSSAValue(new_type)
+            new_value = val.__class__(new_type)
 
         new_value.name_hint = val.name_hint
         val.replace_by(new_value)


superlopuh (approved):

superlopuh:
Actually no, let's import pytest in the pytest.fail instead?

xdsl/rewriter.py
```diff
@@ -14,7 +14,6 @@
     Region,
     SSAValue,
 )
-from xdsl.utils.test_value import TestSSAValue
```
superlopuh, line 17: Importing this should be legal even without pytest imported

EdmundGoodman: Have moved the import inline to fix this. A further question is should `TestSSAValue` even live in the main xdsl code, or could it be moved into the `test/` directory, people could still invoke `.owner` on it reasonably in normal usage?

superlopuh (approved):

xdsl/rewriter.py
superlopuh: It shouldn't occur in normal usage, and it will likely be useful for testing in downstream repos.

EdmundGoodman:
Merging now to fix this promptly, but think we should return to the question of moving `TestSSAValue` to `test/` later.

xdsl/rewriter.py
superlopuh: @math-fehr do you agree with this change as the fix? Should probably merge before the next release.

superlopuh:
Why the urgency?

EdmundGoodman:
> Why the urgency?

It breaks anyone downstream trying to use xdsl?

superlopuh:
Only people pointing to specific commits or main branch, not sure who does that. Merging it now won't change anything for anyone installing a version from pypi.

PR #4207, opened 2025-04-08
core: add is_projected_permutation to AffineMap

superlopuh:


xdsl/ir/affine/affine_map.py
```diff
@@ -331,6 +331,49 @@ def is_minor_identity(self) -> bool:
...
+        Examples:
+        ```
+        no_zeros = (d0, d1, d2) -> (d1, d0)
+        with_zeros = (d0, d1, d2) -> (d1, 0, d0)
+        ```
+
+        Equivalent to `isProjectedPermutation` in MLIR.
+        """
+        if self.num_symbols:
+            return False
```
EdmundGoodman, line 349: Missing test coverage.

EdmundGoodman (changes requested):

EdmundGoodman (approved):

PR #4212, opened 2025-04-09
bench: add parsing benchmarks using existing workloads

EdmundGoodman:
Add parsing benchmarks using existing workloads, peeled from #4008

### DIFF
--- a/benchmarks/workloads.py
+++ b/benchmarks/workloads.py
@@ -104,8 +104,11 @@ class WorkloadBuilder:
         assert x >= 0
         assert y >= 0
         random.seed(RANDOM_SEED)
-        # Each dense attr item is a byte = 2 hex chars
-        dense_attr_hex = "".join(random.choice(HEX_CHARS) for _ in range(x * y * 2))
+        # In order to guarantee the hex value encodes a valid i8 tensor without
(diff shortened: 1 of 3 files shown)


benchmarks/workloads.py
```diff
@@ -104,8 +104,11 @@ def large_dense_attr_hex(cls, x: int = 1024, y: int = 1024) -> str:
         assert x >= 0
         assert y >= 0
         random.seed(RANDOM_SEED)
-        # Each dense attr item is a byte = 2 hex chars
-        dense_attr_hex = "".join(random.choice(HEX_CHARS) for _ in range(x * y * 2))
+        # In order to guarantee the hex value encodes a valid i8 tensor without
+        # significant logic coupled to the hex implementation, we set all values
+        # to zero. Each dense attr item is a byte is 2 hex chars, so we need
+        # x * y * 2 characters.
+        dense_attr_hex = "0" * x * y * 2
```
superlopuh, line 111: I would recommend anything other than 0, as that is equivalent to `dense<0> : tensor<xxyxi8>`, and while we don't have this transformation today, we might in the future. Something like that many ones should work fine.

PR #4215, opened 2025-04-09
bench: add microbenchmarks for comparison with "How slow is MLIR?"

EdmundGoodman:
Add microbenchmarks for comparison with "How slow is MLIR?", peeled from #4008


benchmarks/microbenchmarks.py
```diff
@@ -1,37 +1,151 @@
...
+        to compare `isinstance` with `dyn_cast` in context.
+        """
+        assert isinstance(Extensibility.HAS_TRAIT_A_OP, HasTraitAOp)
+
+    def time_trait_check(self) -> None:
+        """Time checking the trait of an operation.
+
+        Comparison with `assert( op->hasTrait<TraitT>(op) )`.
+        """
+        assert Extensibility.HAS_TRAIT_A_OP.has_trait(TraitA)
```
math-fehr, line 92: Can we also have a test like this where we have 10/20 traits in an operation?
AFAIK, that's where things become costly, even in MLIR

benchmarks/microbenchmarks.py
```diff
@@ -1,37 +1,151 @@
...
+
+    CONSTANT_OPERATION = EmptyOp()
+
+    def time_operation_create(self) -> None:
+        """Time creating an empty operation.
+
+        Comparison with `OperationState opState(unknownLoc, "testbench.empty");
+        Operation::create(opState)`.
+        """
+        EmptyOp.create()
```
math-fehr, line 113: Can we have another one where we pass a few operands / result types / regions?

math-fehr (approved):
Very nice!
I just suggested a few other benchmarks, but feel free to ignore them or add them later if you want

benchmarks/microbenchmarks.py
superlopuh: I would recommend a separate PR for this

superlopuh (approved):

benchmarks/microbenchmarks.py
math-fehr: Fine by me

PR #4217, opened 2025-04-09
bench: add bytecode disassembly tool

EdmundGoodman:
Add bytecode disassembly tool, peeled from #4008


benchmarks/bytecode/visualise.py
```diff
@@ -0,0 +1,180 @@
...
+            stack.append(func_name)
+
+        elif event == "return":
+            stack.pop()
+        return trace_calls
+
+    sys.settrace(trace_calls)
+    func()
+    sys.settrace(None)
+    return list(zip(names, objs, depths))
```
superlopuh, line 157: ?
```suggestion
    return list(zip(names, objs, depths, strict=True))
```

superlopuh (approved):
Very cool

math-fehr (approved):

PR #4222, opened 2025-04-10
testing: replace TestSSAValue class with create_ssa_value helper

superlopuh:
As discussed on Zulip, this gets rid of a weird special SSAValue that messed up some core infrastructure.

### DIFF
--- a/tests/test_ssa_value.py
+++ b/tests/test_ssa_value.py
@@ -4,7 +4,7 @@ from xdsl.dialects.builtin import StringAttr, i32, i64
 from xdsl.ir import Block, BlockArgument, Operation, SSAValue
 from xdsl.irdl import IRDLOperation, irdl_op_definition, result_def
 from xdsl.rewriter import Rewriter
-from xdsl.utils.test_value import TestSSAValue
+from xdsl.utils.test_value import create_ssa_value
 
 
 @irdl_op_definition
@@ -65,7 +65,7 @@ def test_invalid_ssa_vals(name: str):
 
 def test_rewrite_type():
     """We can rewrite the type of a test SSA value."""
-    val = TestSSAValue(i32)
+    val = create_ssa_value(i32)
     rewriter = Rewriter()
     new_val = rewriter.replace_value_with_new_type(val, i64)
     assert val.type == i32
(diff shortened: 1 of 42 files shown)


xdsl/utils/test_value.py
```diff
@@ -1,9 +1,7 @@
...
+from xdsl.ir import AttributeCovT, OpResult
 
-from xdsl.ir import AttributeCovT, Block, Operation, SSAValue
 
-
-class TestSSAValue(Generic[AttributeCovT], SSAValue[AttributeCovT]):
-    @property
-    def owner(self) -> Operation | Block:
-        raise ValueError("Attempting to get the owner of a `TestSSAValue`")
+def TestSSAValue(t: AttributeCovT) -> OpResult[AttributeCovT]:
```
alexarice, line 5: I get minimising the diff, but CamelCase for a function feels wrong

superlopuh: Yeah I was thinking of moving it over either in this PR or a subsequent one if everyone was happy with the mechanism. @math-fehr ?

alexarice: I'm otherwise happy with the idea of "testSSAValue" (or similar name) just being a helper to create a test op

tests/test_ssa_value.py
```diff
@@ -62,9 +63,31 @@ def test_invalid_ssa_vals(name: str):
...
+
+    rewriter = Rewriter()
+
     with pytest.raises(
-        ValueError, match="Attempting to get the owner of a `TestSSAValue`"
+        ValueError,
+        match="Expected OpResult or BlockArgument, got InvalidValue",
     ):
-        val.owner
+        rewriter.replace_value_with_new_type(val, i32)
```
math-fehr, line 93: I don't think this test is relevant.
Or at least, we have so much piece of infrastructure that relies on the fact that a value is an operand or a block argument, that we should test these then (but I don't believe it's relevant).

xdsl/utils/test_value.py
math-fehr: I'm fine in keeping it for a bit, but yeah we should change it to `get_test_ssa_value` or something.

math-fehr (approved):

tests/test_ssa_value.py
superlopuh: I don't understand your suggestion, should we keep it or remove it?

math-fehr (approved):
Thanks for following up!

PR #4223, opened 2025-04-10
dialects: (riscv) canonicalization pattern added (x<<0=x)

yamaguchi1024:


superlopuh:
We use ruff to format code in xDSL, could you please run `make precommit-install` in xDSL, this will make the formatting automatic at commit time?

yamaguchi1024:
ok done!

tests/filecheck/backend/riscv/canonicalize.mlir
```diff
@@ -123,6 +123,9 @@ builtin.module {
   %xor_bitwise_zero_r0 = riscv.xor %c0, %c1 : (!riscv.reg, !riscv.reg) -> !riscv.reg<a0>
   "test.op"(%xor_bitwise_zero_r0) : (!riscv.reg<a0>) -> ()
 
+  %shift_left_zero_r0 = riscv.slli %i2, 0 : (!riscv.reg) -> !riscv.reg
```
alexarice, line 126: I'm not a riscv expert but would it be good to make the input and output different registers as in the other examples?

superlopuh (approved):

tests/filecheck/backend/riscv/canonicalize.mlir
superlopuh: Ah I remembered now why the other tests have the result register specified. The default value for the return type when creating a riscv op is not allocated, and we want to test that the result value being allocated is preserved. Could you please make a similar change here?

superlopuh:
This is good to merge, but would be nice to both add the return register assignment back and to do a fly-by fix of the broken docstring you spotted yesterday since it's a tiny non-functional change in the same area.

yamaguchi1024:
Thanks, fixed!

superlopuh (approved):
Thank you!

PR #4224, opened 2025-04-10
misc: move some typealiases up in interpreter.py

superlopuh:
The recent Pyright update (#4221) requires the default factories to be fully typed, which is a minor change in most of the framework, but messes things up in interpreter.py due to the typealiases for the more complex signatures. I thought I'd split it out for an easier review.

### DIFF
--- a/xdsl/interpreter.py
+++ b/xdsl/interpreter.py
@@ -103,29 +103,15 @@ AttrImpl: TypeAlias = Callable[
 ]
 
 _ImplDict: TypeAlias = dict[type[Operation], OpImpl["InterpreterFunctions", Operation]]
-_IMPL_DICT: TypeAlias = dict[
-    type[Operation],
-    tuple["InterpreterFunctions", OpImpl["InterpreterFunctions", Operation]],
-]
 
 _CastImplDict: TypeAlias = dict[
     tuple[type[Attribute], type[Attribute]],
     CastImpl["InterpreterFunctions", Attribute, Attribute],
 ]
-_CAST_IMPL_DICT: TypeAlias = dict[
-    tuple[type[Attribute], type[Attribute]],
-    tuple[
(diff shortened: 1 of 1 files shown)


xdsl/interpreter.py
```diff
@@ -30,16 +30,140 @@
...
-_IMPL_DICT = "__impl_dict"
-_CAST_IMPL_DICT = "__cast_impl_dict"
-_ATTR_IMPL_DICT = "__attr_impl_dict"
-_EXT_FUNC_DICT = "__external_func_dict"
-_CALLABLE_IMPL_DICT = "__callable_impl_dict"
+PythonValues: TypeAlias = tuple[Any, ...]
+"""
+A tuple of result values corresponding to the results of the operation being
+interpreted.
+"""
```
superlopuh, line 37: added some doc strings

superlopuh: Otherwise just cut/paste from bottom of file, with the addition of `_IMPL_DICT` and friends, renamed the keys to have a KEY suffix.

alexarice:
Can I be really annoying and ask that the rename/refactor and actual change are separated?

alexarice:
Even if it's just into separate commits instead of separate PRs

superlopuh:
I'm not 100% sure of what you mean but I'll try to split it up and see if you like it :)

alexarice:
I feel like there's three things happening here:
- Renaming some type aliases
- Moving things to the top of the file
- Adding extra annotations to the default factories

I think I'd like to see at least the last of these on its own

xdsl/interpreter.py
superlopuh: I have removed `_IMPL_DICT` and friends, now only change are copy/paste up, change names of keys to add `_KEY`, and add some doc strings.

alexarice (approved):

PR #4230, opened 2025-04-11
dialects: (x86) Change RR_MovOp and RI_MovOp to have only one input

manueljacob:
Before this change, RR_MovOp and RI_MovOp were subclasses of R_RR_Operation and R_RI_Operation, respectively. These classes are meant for instructions for which, in x86 Intel syntax, the first operand refers to the register which is both the first input register and the output register, and the second operand refers to the second input register or immediate.

The mov instruction with 32-bit or 64-bit registers has only one input because the result register is fully overwritten (in the 32-bit case, the upper half is zeroed). In these cases, there is only one input. With 8-bit or 16-bit registers, only part of the result register is overwritten, so the content of the result register depends on the previous content of it. Partial registers except 32-bit ones are not yet implemented in the x86 dialect, so for now the latter case can be ignored. Therefore, the mov instruction is changed to be modelled with one input register or immediate, and one result register. This change is in line with #3967. There are no existing abstract classes for these kinds of instructions, so RR_MovOp and RI_MovOp were changed to directly subclass X86Instruction.

If the result register is a 64-bit register, R_RI_Operation should be able to take 64-bit immediates. However, supporting this is currently hard because unallocated registers don’t track the size of the register. As R_RI_Operation didn’t support 64-bit immediates, either, this is not a regression. (As a side-note, mov is the only x86 instruction able to take 64-bit immediates.)

### DIFF
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -53,6 +53,7 @@ from .register import (
     RAX,
     RDX,
     RSP,
+    UNALLOCATED_GENERAL,
     GeneralRegisterType,
     RFLAGSRegisterType,
     X86RegisterType,
@@ -314,17 +315,42 @@ class RR_XorOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
 
 
 @irdl_op_definition
-class RR_MovOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
+class RR_MovOp(X86Instruction, X86CustomFormatOperation):
     """
-    Copies the value of r1 into r2.
+    Copies the value of source into result.
     ```C
-    x[r1] = x[r2]
+    x[result] = x[source]
     ```
 
     See external [documentation](https://www.felixcloutier.com/x86/mov).
     """
 
     name = "x86.rr.mov"
+
+    source = operand_def(GeneralRegisterType)
+
+    result = result_def(GeneralRegisterType)
+
+    def __init__(
+        self,
+        source: Operation | SSAValue,
+        *,
+        comment: str | StringAttr | None = None,
+        result: GeneralRegisterType = UNALLOCATED_GENERAL,
+    ):
+        if isinstance(comment, str):
+            comment = StringAttr(comment)
+
+        super().__init__(
+            operands=[source],
+            attributes={
+                "comment": comment,
+            },
+            result_types=[result],
+        )
+
+    def assembly_line_args(self) -> tuple[AssemblyInstructionArg | None, ...]:
+        return self.result, self.source
 
 
 @irdl_op_definition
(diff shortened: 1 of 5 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -314,18 +315,43 @@ class RR_XorOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
 
 
 @irdl_op_definition
-class RR_MovOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
+class RR_MovOp(X86Instruction, X86CustomFormatOperation):
```
superlopuh, line 318: I would recommend against this change. The reason for the helpers is to make the code for each instruction a bit more readable. The base class naming convention in this file is:

`R` * number of destination registers `_` `R` * number of source registers `_` I * number of immmediate values `_` `Operation`

So this line should read:
```suggestion
class R_R_MovOp(R_R_Operation[GeneralRegisterType, GeneralRegisterType):
```

I can't see in the GH UI whether this exists already, could you please add it if not? Similarly for the move immediate operation below.

manueljacob: As noted in the commit message, I found no suitable base classes for these instructions. I do agree, though, that factoring out the boilerplate would make things more readable.

`R_R_Operation` exists, but it’s “a base class for x86 operations that have one register acting as both source and destination”. But here we need two registers, one destination and one source. Do you think there’s a way to disambiguate these two cases within the current naming scheme?

superlopuh: I think that's just bad phrasing, it's defined as having one operand and one result, so the doc comment should be changed, but it seems like the definition is good.

manueljacob: The `R_R_Operation` is implemented as described in the class docstring, as can be seen here: https://github.com/xdslproject/xdsl/blob/841db977a0d25e9a4ec6a09b9311ce099a9bfdcf/xdsl/dialects/x86/ops.py#L431

superlopuh: I see, it looks like this scheme is not quite pulling its weight then, and that these definitions are missing an explicit constraint that enforces that the registers are the same. I would probably rename the existing op `R_R_Inout_Operation`, and add an `R_R_Operation` for the case where the registers aren't the same.

superlopuh:
Hi @manueljacob, I needed this fix for another project, so I opened a new PR: #4427. I've been fixing up the dialect and the naming convention, and generally moving things around, so this branch got a little out of sync. Would appreciate your review if you have time. The main difference with the approach in this PR is that it tries to keep almost all the logic in the abstract base classes, so that the assembly ops themselves just specify the name. It also has some slightly different naming conventions.

manueljacob:
Sorry, my work on this got interrupted by EuroLLVM and after that I needed to catch up with a lot of other stuff.

manueljacob:
It seems like this is completely superseded by #4427.

superlopuh:
No worries, sorry we didn't manage to resolve this by iterating together. Hopefully you'll find the improvements we're making in the x86 dialects useful.

PR #4240, opened 2025-04-14
[WIP] bench: reference version to optimise against

EdmundGoodman:
A reference version to optimise against, as mainline xDSL is very transient, so hard to make fair comparisons against. This reference version is [ce37678](https://github.com/xdslproject/xdsl/commit/ce37678ff1b787c7bb16d4b48d39fa9c2f313950) (having all the benchmarks merged) with the two already merged optimisations #3961 and #4163 reverted.


EdmundGoodman:
Closed, as intended for use only as a reference point for measurements, with no intention to roll back changes in main

PR #4245, opened 2025-04-15
dialects: (tensor) add ExpandShapeOp

Jimmy2027:
wip to add the tensor ExpandShapeOp.

I'm not sure what the best way to do this is. I ran xdsl-tblgen to create a new tensor.py but the diff to the original is quite big and lots of functionality seems missing.

I'm currently stuck on trying to parse something like this:
```mlir
    %expanded = tensor.expand_shape %4 [[0], [1, 2, 3]] output_shape [%6, 1, 1, 5] : tensor<?x5xi32> into tensor<?x1x1x5xi32>
```
I couldn't find a method in the BaseParser that could parse either an int or a reference.


superlopuh:
What do you mean by reference? We recently added helpers to parse a list of mixed indices and ssa values, `parse_dynamic_index_list_without_types`, maybe that will be useful here?

Jimmy2027:
thanks for the reference, will have a look! With reference I mean a SSA value, e.g. `[%6, 1, 1, 5]`

Jimmy2027:
this works for my use case like this, is it good enough to be merged in?

Jimmy2027:
hey @superlopuh, I forgot about this, sorry 😅 Will clean this up, add tests and then mark it as ready

xdsl/dialects/tensor.py
```diff
@@ -293,6 +298,97 @@ def verify_(self) -> None:
...
+
+        shape_attr = DenseArrayBase.create_dense_int(i64, static_shape)
+
+        reassociation = cast(ReassociationAttr, reassociation)
+        result_type = cast(TensorType[Attribute], result_type)
+
+        return cls(src, dyn_shape, reassociation, shape_attr, result_type)
+
+    def print(self, printer: Printer):
+        printer.print(" ")
```
superlopuh: We're currently removing uses of this function, please use print_string instead
```suggestion
        printer.print_string(" ")
```

xdsl/dialects/tensor.py
```diff
@@ -293,6 +298,97 @@ def verify_(self) -> None:
...
+        parser.parse_characters("into")
+        result_type = parser.parse_type()
+        src = parser.resolve_operand(src_operand, src_type)
+
+        shape_attr = DenseArrayBase.create_dense_int(i64, static_shape)
+
+        reassociation = cast(ReassociationAttr, reassociation)
+        result_type = cast(TensorType[Attribute], result_type)
+
+        return cls(src, dyn_shape, reassociation, shape_attr, result_type)
```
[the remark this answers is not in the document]
Jimmy2027: something like this?
https://github.com/xdslproject/xdsl/pull/4245/commits/5300d691495d2d671d404b365a7e3bb8c43d4555

Jimmy2027:
still working on the verification, will mark as ready once I'm done

Jimmy2027:
tests were taken over from https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Tensor/invalid.mlir#L351C17-L351C18

xdsl/dialects/tensor.py
```diff
@@ -41,37 +43,6 @@
 from xdsl.printer import Printer
 from xdsl.traits import NoMemoryEffect
 from xdsl.utils.exceptions import VerifyException
-from xdsl.utils.hints import isa
-
-
-@dataclass(frozen=True)
-class ContiguousArrayOfIntArray(AttrConstraint):
```
superlopuh: What's the motivation for moving this? Is it in a utils file in MLIR? Do you envision this being useful for other dialects?

Jimmy2027: yeah I don't see why not, the [MLIR counterpart](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Utils/ReshapeOpsUtils.h) is also shared

superlopuh: OK let's do that in a separate PR then? WIll make it easier to review. I think in that case it might be good to have more documentation both for the file and each function.

Jimmy2027: sure, will do!

Jimmy2027:
converted to draft to get #4714 in first

superlopuh (approved):

superlopuh:
Sorry forgot about this one

PR #4248, opened 2025-04-16
dialects: (linalg) Select op

n-io:
Adds the `linalg.select` op


superlopuh (commented):
Looks great can you please add an MLIR interop test?

superlopuh (approved):
Thank you!

PR #4251, opened 2025-04-17
canonicalize: (arith) Fold cmpf and select on same operands

n-io:
I have encountered a pattern where `arith.cmpf` is immediately followed by `arith.select` on the same operands:

```
    %cst_0 = arith.constant 3.000000e+00 : f64
    %0 = "test.op"() : () -> f64
    %1 = arith.cmpf ogt, %0, %cst_0 : f64
    %2 = arith.select %1, %0, %cst_0 : f64
    "test.op"(%2) : (f64) -> ()
```
This could be simplified to:
```
    %cst_0 = arith.constant 3.000000e+00 : f64
    %0 = "test.op"() : () -> f64
    %1 = arith.maximumf %0, %cst_0 : f64
    "test.op"(%1) : (f64) -> ()
```

The ordered comparisons olt/ole and ogt/oge will produce `minimumf` and `maximumf` ([link](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithmaximumf-arithmaximumfop)), respectively. Similarly, the unordered comparisons ult/ule and ugt/uge will produce a `minnumf` and `maxnumf`, respectively. 

Here, ogt and olt mean "ordered greater than" and "ordered less than", respectively.

### DIFF
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -159,3 +159,29 @@ func.func @test_const_var_const() {
 %11 = arith.constant true
 %12 = arith.addi %11, %11 : i1
 "test.op"(%12) : (i1) -> ()
+
+func.func @test_fold_cmpf_select() {
+  %cst = arith.constant 4.000000e+00 : f64
+  %cst_0 = arith.constant 3.000000e+00 : f64
+  %13 = "test.op"() : () -> f64
+  %14 = arith.cmpf ogt, %13, %cst_0 : f64
+  %15 = arith.select %14, %13, %cst_0 : f64
+  %16 = arith.cmpf olt, %15, %cst : f64
+  %17 = arith.select %16, %15, %cst : f64
+  %18 = arith.cmpf ule, %17, %cst : f64
+  %19 = arith.select %18, %17, %cst : f64
+  %20 = arith.cmpf ule, %19, %cst : f64
+  %21 = arith.select %20, %19, %cst : f64
+  "test.op"(%21) : (f64) -> ()
+  return
+
+  // CHECK-LABEL: @test_fold_cmpf_select
+  // CHECK-NEXT:  %cst = arith.constant 4.000000e+00 : f64
+  // CHECK-NEXT:  %cst_1 = arith.constant 3.000000e+00 : f64
+  // CHECK-NEXT:  %12 = "test.op"() : () -> f64
+  // CHECK-NEXT:  %13 = arith.maximumf %12, %cst_1 : f64
+  // CHECK-NEXT:  %14 = arith.minimumf %13, %cst : f64
+  // CHECK-NEXT:  %15 = arith.minnumf %14, %cst : f64
+  // CHECK-NEXT:  %16 = arith.minnumf %15, %cst : f64
+  // CHECK-NEXT:  "test.op"(%16) : (f64) -> ()
+}
--- a/tests/filecheck/dialects/arith/canonicalize.mlir
+++ b/tests/filecheck/dialects/arith/canonicalize.mlir
@@ -161,9 +161,7 @@ func.func @test_const_var_const() {
 "test.op"(%12) : (i1) -> ()
 
 func.func @test_fold_cmpf_select() {
-  %cst = arith.constant 4.000000e+00 : f64
-  %cst_0 = arith.constant 3.000000e+00 : f64
-  %13 = "test.op"() : () -> f64
+  %cst, %cst_0, %13 = "test.op"() : () -> (f64, f64, f64)
   %14 = arith.cmpf ogt, %13, %cst_0 : f64
   %15 = arith.select %14, %13, %cst_0 : f64
   %16 = arith.cmpf olt, %15, %cst : f64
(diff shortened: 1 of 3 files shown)


alexarice:
I worry that this is unsound in the presence of NAN

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -189,6 +189,41 @@ def match_and_rewrite(self, op: arith.SelectOp, rewriter: PatternRewriter):
...
 
+class CmpfOpFoldSelectPattern(RewritePattern):
+    """
+    %1 = arith.cmpf  ogt, %0, %cst : f64
+    %2 = arith.select %1, %0, %cst : f64
+    """
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: arith.CmpfOp, rewriter: PatternRewriter):
+        if len(op.result.uses) != 1:
```
alexarice, line 200: Is it not better to continue anyway in this case?

xdsl/transforms/canonicalization_patterns/arith.py
```diff
@@ -189,6 +189,41 @@ def match_and_rewrite(self, op: arith.SelectOp, rewriter: PatternRewriter):
...
+            case 9 | 10:
+                # ugt | uge
+                target = arith.MaxnumfOp
+            case 11 | 12:
+                # ult | ule
+                target = arith.MinnumfOp
+            case _:
+                return
+        rewriter.replace_op(select, target(select.lhs, select.rhs, op.fastmath))
+        rewriter.erase_matched_op()
```
alexarice, line 224: This should be taken care of by DCE, which is run by the canonicalizer

tests/filecheck/dialects/arith/canonicalize.mlir
```diff
@@ -159,3 +159,29 @@ func.func @test_const_var_const() {
 %11 = arith.constant true
 %12 = arith.addi %11, %11 : i1
 "test.op"(%12) : (i1) -> ()
+
+func.func @test_fold_cmpf_select() {
+  %cst = arith.constant 4.000000e+00 : f64
+  %cst_0 = arith.constant 3.000000e+00 : f64
```
alexarice, line 165: Is there a reason to test on constants instead of opaque values from "test.op"?

alexarice (commented):
I'm less sure on the NAN thing now, didn't realise there are separate minimum/minnum operations etc, but would be good to get some confirmation of this as I couldn't see anything in the documentation about this

xdsl/transforms/canonicalization_patterns/arith.py
n-io: removed line

xdsl/transforms/canonicalization_patterns/arith.py
n-io: Hmm not sure, it's a neat way of getting the select I'm interested in. Maybe this should be a canonicalisation pattern on the select op in the first place? It's just that the logic is rather cmpf-specific

n-io:
> I'm less sure on the NAN thing now, didn't realise there are separate minimum/minnum operations etc, but would be good to get some confirmation of this as I couldn't see anything in the documentation about this

This is a good starting point:
https://stackoverflow.com/a/40352836

But really, it might be a bit complex The [LLVM fcmp docs](https://llvm.org/docs/LangRef.html#fcmp-instruction) state, that:
* ordered cmpf: If either operand is NaN, the comparison will yield false
* unordered cmpf: If either operand is a NaN, the comparison will yield true

For a select following a cmpf, this would mean it would return the second or first operand, respectively, if either operand is NaN. This is mildly different from e.g. the minnumf/minimumf behaviour.

I suggest that we perform this canonicalisation iff the no-nans fast-math flag is set ([link to docs](https://llvm.org/docs/LangRef.html#fast-math-flags)) and always rewrite to the maximumf/miminumf functions only.

alexarice:
I'm happy with this version if we're sure everything respects NaNs properly, I just couldn't grasp whether this was true from the mlir documentation

n-io:
Changed as discussed, the canonicalisation now requires no-nan fast math flag, always outputs maximumf/miminumf, and has been moved to the SelectOp, such that the single-use check has disappeared.

math-fehr (approved):
I don't know how correct this is, so I will trust you on this!
It seems correct from far away at least ;)

superlopuh (commented):
FWIW `mlir-opt` does not perform this canonicalization `mlir-opt tests/filecheck/dialects/arith/canonicalize.mlir --allow-unregistered-dialect --canonicalize`, which makes me suspect that this is incorrect...

superlopuh:
@math-fehr Given that these ops have LLVM equivalents, would it be possible to use Alive2 to check this?

superlopuh:
Alive says that the before and after have different behaviour when comparing +0 and -0: https://alive2.llvm.org/ce/z/wEw_iC. In my understanding, floating point comparison of +0 and -0 will always yield equal, and floating point min/max will always yield the 0 with the appropriate sign, so this canonicalization is incorrect.

georgebisbas (approved):

n-io:
> FWIW `mlir-opt` does not perform this canonicalization `mlir-opt tests/filecheck/dialects/arith/canonicalize.mlir --allow-unregistered-dialect --canonicalize`, which makes me suspect that this is incorrect...

There is a number of canonicalizations that `mlir-opt` does not perform, for instance, reassociating arithmetic expressions on constants when the reassoc fast-math flag is set.

> Alive says that the before and after have different behaviour when comparing +0 and -0: https://alive2.llvm.org/ce/z/wEw_iC. In my understanding, floating point comparison of +0 and -0 will always yield equal, and floating point min/max will always yield the 0 with the appropriate sign, so this canonicalization is incorrect.

Great observation, I propose requiring the no-signed-zeros fast-math flag in order for this rewrite to work.

tests/filecheck/dialects/arith/canonicalize.mlir
superlopuh, line 1: Could you please add names of ssa values and a negative test case to make sure we only rewrite if the flag is present?

superlopuh (approved):
Thank you!

PR #4253, opened 2025-04-17
dialects: (linalg) MinOp and MaxOp

n-io:


xdsl/dialects/linalg.py
```diff
@@ -717,6 +717,92 @@ def verify_(self) -> None:
...
+        attributes: dict[str, Attribute] | None = None,
+    ):
+        if res is None:
+            result_types = tuple(output.type for output in outputs)
+        else:
+            result_types = res
+
+        arg_types = self.body_arg_types((*inputs, *outputs))
+        maxop = (
+            arith.MaximumfOp
```
alexarice, line 744: Is this a typo? Why the if b then X else X?

xdsl/dialects/linalg.py
```diff
@@ -717,6 +717,92 @@ def verify_(self) -> None:
...
+    ):
+        if res is None:
+            result_types = tuple(output.type for output in outputs)
+        else:
+            result_types = res
+
+        arg_types = self.body_arg_types((*inputs, *outputs))
+        minop = (
+            arith.MinimumfOp
+            if isinstance(arg_types[-1], AnyFloat)
```
alexarice, line 788: Same here

alexarice (approved):
I'm not hugely familiar with `linalg` but looks reasonable - two small comments

xdsl/dialects/linalg.py
n-io: Thanks, great spot! Upstream appears to use the maxsi/minsi ops internally.

PR #4255, opened 2025-04-17
transformations: (linalg-to-csl) Translate linalg.max

n-io:
It turns out, there are functions for max but not for min. Is there an undocumented one?


dk949 (approved):
`@*min*` appears to be missing, it's not documented and the compiler doesn't support it.

PR #4257, opened 2025-04-19
interpreter: (pdl_interp) Implement some `check*` operations

jumerckx:


xdsl/interpreters/pdl_interp.py
```diff
@@ -131,6 +133,107 @@ def run_getdefiningop(
         )
         return (args[0].owner,)
 
+    @impl_terminator(pdl_interp.CheckOperationNameOp)
+    def run_checkoperationname(
```
superlopuh, line 137: For the longer names the snake_case is much easier to read IMO
```suggestion
    def run_check_operation_name(
```

xdsl/interpreters/pdl_interp.py
```diff
@@ -131,6 +133,107 @@ def run_getdefiningop(
...
+        self,
+        interpreter: Interpreter,
+        op: pdl_interp.CheckOperationNameOp,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        assert len(args) > 0
```
superlopuh, line 143: Nit: as a code style we try to stick to checking whether a collection is truthy, rather than first calculating the length, which could be O(n), then comparing with 0
```suggestion
        assert args
```

superlopuh: and below

tests/interpreters/test_pdl_interp_interpreter.py
```diff
@@ -107,6 +107,268 @@ def test_getters():
...
+        "test.op", create_ssa_value(pdl.OperationType()), truedest, falsedest
+    )
+
+    trueresult = pdl_interp_functions.run_checkoperationname(
+        interpreter, check_name_op, (op,)
+    )
+
+    assert isinstance(trueresult.terminator_value, Successor)
+    assert isinstance(trueresult.terminator_value.block, Block)
+    assert trueresult.terminator_value.block == truedest
```
superlopuh, line 133: nit:
```suggestion
    assert trueresult.terminator_value.block is truedest
```

superlopuh: and below

xdsl/interpreters/pdl_interp.py
jumerckx: Right, sorry, you'd told me this before.
In this case the checks should actually check for a specific length, mostly `== 1`. Updated as such.

jumerckx:
I think I addressed the comments

superlopuh (approved):
Looks great! As a final nit, it would be great to convert the other `run_*` names to snake_case, but that's up to you.

PR #4258, opened 2025-04-19
interpreter: (pdl_interp) Implement `ReplaceOp`, `FuncOp`, `RecordMatchOp`, and `FinalizeOp`.

jumerckx:
`RecordMatchOp` and `FinalizeOp` don't really have unit tests currently, these don't produce values that can be used in further interpretation, and these will get tested by apply_pdl_interp pass tests. Does that suffice?
I will still add a test for `ReplaceOp` in this pr.


xdsl/interpreters/pdl_interp.py
```diff
@@ -234,6 +237,33 @@ def run_are_equal(
...
+        input_op = args[0]
+        assert isinstance(input_op, Operation)
+
+        # Get replacement values (if any)
+        repl_values: list[SSAValue] = []
+        for i in range(0, len(args) - 1):
+            if isa(op.repl_values.types[i], ValueType):
+                repl_values.append(args[i + 1])
+            elif isa(op.repl_values.types[i], RangeType[ValueType]):
+                repl_values.extend(args[i + 1])
```
superlopuh, line 257: Let's add a test for this case?

jumerckx:
Comments have been addressed.

PR #4261, opened 2025-04-22
CI: remove references to setup.py

superlopuh:
I forgot to remove all mentions of setup.py when removing the file, leading us to not actually publishing xDSL for a few weeks. This PR replaces commands running setup.py with `uv build`, which is the recommended modern approach.


superlopuh:
It worked!

PR #4263, opened 2025-04-22
dialects: (vector) add transfer_read and transfer_write

superlopuh:
This is a partial implementation of the ops, which just defines them and the custom format. The idea is to merge them first, then the verification #3650.

CC @watermelonwolverine


xdsl/dialects/vector.py
```diff
@@ -757,6 +761,286 @@ def _print_attrs(self, printer: Printer):
...
+            printer.print_ssa_value(self.mask)
+        self._print_attrs(printer)
+        printer.print_string(" : ", indent=0)
+        printer.print_attribute(self.source.type)
+        printer.print_string(", ", indent=0)
+        printer.print_attribute(self.result.type)
+
+    @classmethod
+    def parse(cls, parser: Parser) -> TransferReadOp:
+        source = parser.parse_unresolved_operand()
```
math-fehr, line 815: I know we have to do this because of MLIR, but eww

xdsl/dialects/vector.py
```diff
@@ -757,6 +761,286 @@ def _print_attrs(self, printer: Printer):
...
+
+        types_pos = parser.pos
+        parser.parse_punctuation(":")
+        vector_type = parser.parse_type()
+        parser.parse_punctuation(",")
+        shaped_type = parser.parse_type()
+
+        vector = parser.resolve_operand(vector, vector_type)
+        source = parser.resolve_operand(source, shaped_type)
+
```
math-fehr, line 978: How much can the next part be shared between the two operations?

math-fehr (approved):
I think we could share more the implementation of the two functions, but otherwise LGTM!

xdsl/dialects/vector.py
superlopuh: I factored it out

PR #4264, opened 2025-04-22
dialects: (vector) add helpers for transfer ops

superlopuh:
PR 1/2, adding helpers, actual operations will be merged in #4263.

### DIFF
--- a/xdsl/dialects/vector.py
+++ b/xdsl/dialects/vector.py
@@ -664,7 +664,7 @@ class VectorTransferOperation(IRDLOperation, ABC):
     rank; the only exception is when a vector is flattened as part of the
     transfer (see `permutation_map`).
 
-    Mirrors VectorTransferOpInterface from https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/VectorInterfaces.td
+    Mirrors VectorTransferOpInterface from [MLIR](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/VectorInterfaces.td)
     """
 
     permutation_map = prop_def(AffineMapAttr)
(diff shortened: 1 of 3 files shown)


xdsl/dialects/vector.py
```diff
@@ -630,6 +637,126 @@ def __init__(
...
+        : memref<?x?xf32>, vector<128xf32>
+    ```
+
+    The "vector transfer rank" is the number of dimensions that participate in
+    the transfer and broadcasts, and matches the number of results in the
+    permutation map. In most cases, the vector rank matches the vector transfer
+    rank; the only exception is when a vector is flattened as part of the
+    transfer (see `getPermutationMap`).
+
+    Mirrors VectorTransferOpInterface from https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/VectorInterfaces.td
```
math-fehr, line 667:
```suggestion
    Mirrors VectorTransferOpInterface from 
    https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/VectorInterfaces.td
```

xdsl/dialects/vector.py
```diff
@@ -630,6 +637,126 @@ def __init__(
...
+    %0 = vector.transfer_read %arg0[%c3, %c3], %f0
+        {permutation_map = affine_map<(d0, d1) -> (d0)>}
+        : memref<?x?xf32>, vector<128xf32>
+    ```
+
+    The "vector transfer rank" is the number of dimensions that participate in
+    the transfer and broadcasts, and matches the number of results in the
+    permutation map. In most cases, the vector rank matches the vector transfer
+    rank; the only exception is when a vector is flattened as part of the
+    transfer (see `getPermutationMap`).
```
math-fehr, line 665: Probably the name is different here?

math-fehr (approved):

xdsl/dialects/vector.py
superlopuh: Oh yeah I need to change this to a markdown link anyway

PR #4254, opened 2025-04-17
installation: (pre-commit) update versions

alexarice:
Updates ruff precommit version and pre-commit-hooks version. Shouldn't renovate do this?

PR #4267, opened 2025-04-23
installation: (renovate) add pre-commit

alexarice:
As suggested in #4254, trying to make renovate update the pre-commit files. Not sure if this is the correct way to do it.


math-fehr (approved):
Looks like what's needed!

PR #4268, opened 2025-04-23
transforms: (linalg-to-csl) Workaround for min

n-io:


dk949 (approved):
Looks good. Mathias confirms that's how you'd do it on the CS-2 👍

PR #4278, opened 2025-04-24
transformations: (stencil-shape-minimise) Support missing dims

n-io:
This PR adds support for stencil types of different dimensionality.

Previously, it used to determine a shape it considers minimal, propagating that throughout the program. Instead, it now builds a separate minimal shape per number of dimensions (e.g. separate for 1-d, 2-d, 3-d, etc).

### DIFF
--- a/xdsl/transforms/stencil_shape_minimize.py
+++ b/xdsl/transforms/stencil_shape_minimize.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass, field
 
 from xdsl.context import Context
 from xdsl.dialects import builtin, func, stencil
+from xdsl.dialects.stencil import StencilBoundsAttr
 from xdsl.ir import Attribute
 from xdsl.passes import ModulePass
 from xdsl.pattern_rewriter import (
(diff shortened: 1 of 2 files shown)


n-io:
The build should work once #4273 is merged

xdsl/transforms/stencil_shape_minimize.py
```diff
@@ -104,15 +112,28 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
         )
         if not bounds:
             return
-        shape: stencil.StencilBoundsAttr = bounds.pop()
+        dim_shapes = dict[int, StencilBoundsAttr]()
         for b in bounds:
-            shape = shape | b
+            dims = len(b.ub)
```
georgebisbas, line 117: I think you can add a docstring/comment on how this case emerges, otherwise GTG.

georgebisbas (approved):

xdsl/transforms/stencil_shape_minimize.py
n-io: Added comment as suggested.

PR #4279, opened 2025-04-24
transformations: (pdl_interp) Add `pdl-interp-apply` pass

jumerckx:
The final pr of the initial pdl-interp implementation, supersedes https://github.com/xdslproject/xdsl/pull/4162.

PR #4287, opened 2025-04-25
interpreter: (pdl_interp) Change rewriter check in `run_create_operation`

jumerckx:


superlopuh:
What's the motivation for this?

jumerckx:
In #4279, I implement `match_and_rewrite`, similar to how things are done for the pdl interpreter:
```py
    def match_and_rewrite(self, xdsl_op: Operation, rewriter: PatternRewriter) -> None:
        # Setup the rewriter
        self.functions.rewriter = rewriter

        # Call the matcher function on the operation
        self.interpreter.call_op(self.matcher, (xdsl_op,))

        # # Reset rewriter
        # self.functions.clear_rewriter()
```
Since the rewriter is provided here, I figured it doesn't make sense to manage the rewriter inside the interpreter functions.

superlopuh:
I'd just like to make sure that the state of the interpreter is modeled well. The PatternRewriter is useful to track changes to operations and to keep a queue of operations to process. Is that needed given that we have our own state machine processing things? Maybe we're better off just passing a Rewriter instead? Then we can track the current operation in some other bit of state.

superlopuh:
Actually it looks like we don't need to track state such as current operation in pdl_interp? Is that correct?

jumerckx:
> The PatternRewriter is useful to track changes to operations and to keep a queue of operations to process.

We still need to have an outer loop calling the matcher on each separate operation.

> Actually it looks like we don't need to track state such as current operation in pdl_interp? Is that correct?

I think the one thing we need to track is something like `has_done_action` in order to know when an e-graph is saturated. 

I have to look closer at how all the different rewriting classes fit together but you're probably right that we can do with less than a full PatternRewriter.

jumerckx:
That being said, it seems to me like either way, having `run_create_operation` not handle rewriter creation makes sense?

superlopuh:
I think we're on the same page, what I was trying to say was that there doesn't seem to be a benefit of using that class at all. Let's try to only store `has_done_action` as state and use a `Rewriter` instead? I think it's fine to use it, it doesn't need to even be initialized in `run_create_operation`.

PR #4288, opened 2025-04-25
core: (parser) Fix VariadicSuccessorVariable assembly format

jumerckx:
This changes xDSL's interpretation of `VariadicSuccessorVariable` in the assemblyformat string to match MLIR.
Concretely, it now parses and prints a list of the form `^bb1, ^bb2, ^bb3` instead of without the commas.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1011,7 +1011,9 @@ class VariadicSuccessorVariable(VariadicVariable, SuccessorDirective):
         current_successor = parser.parse_optional_successor()
         while current_successor is not None:
             successors.append(current_successor)
-            current_successor = parser.parse_optional_successor()
+            if not parser.parse_optional_punctuation(","):
+                break
+            current_successor = parser.parse_successor()
 
         state.successors[self.index] = successors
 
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1023,7 +1025,7 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
             return
         if state.should_emit_space or not state.last_was_punctuation:
             printer.print(" ")
-        printer.print_list(successor, printer.print_block_name, delimiter=" ")
+        printer.print_list(successor, printer.print_block_name, delimiter=", ")
```
alexarice, line 1028:
```suggestion
        printer.print_list(successor, printer.print_block_name)
```
this is the default

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1011,7 +1011,9 @@ def parse(self, parser: Parser, state: ParsingState) -> bool:
         current_successor = parser.parse_optional_successor()
         while current_successor is not None:
             successors.append(current_successor)
-            current_successor = parser.parse_optional_successor()
+            if not parser.parse_optional_punctuation(","):
```
alexarice, line 1014: I feel there's already a parser function for comma separated lists

alexarice (approved):
Some minor comments but seems good to me

xdsl/irdl/declarative_assembly_format.py
alexarice: See `VariadicOperandVariable`. To be honest it doesn't seem like a great improvement over what is written here.

jumerckx:
Thanks for the review, comments should've been addressed now!

superlopuh (approved):

PR #4289, opened 2025-04-26
transformations: (transform) Add initial transform interpreter pass

efferifick:


xdsl/transforms/transform_interpreter.py
```diff
@@ -0,0 +1,43 @@
...
+from xdsl.passes import Context, ModulePass
+from xdsl.utils.exceptions import PassFailedException
+
+
+@dataclass(frozen=True)
+class TransformInterpreterPass(ModulePass):
+    """Transform dialect interpreter"""
+
+    # TODO: Add the rest of options to the transform interpreter
+    # and add semantics for all of the other transform operations
```
superlopuh, line 17: let's remove this comment, we avoid TODOs and track these with issues instead

xdsl/transforms/transform_interpreter.py
```diff
@@ -0,0 +1,43 @@
...
+
+    name = "transform-interpreter"
+
+    entry_point: str = "__transform_main"
```
superlopuh, line 21: if we provide a way to manually specify the entry point, we should add tests that leverage it, as well as a test for the case where the entry point is not found

xdsl/transforms/transform_interpreter.py
```diff
@@ -0,0 +1,43 @@
...
+    ) -> transform.NamedSequenceOp:
+        for op in root.walk():
+            if (
+                isinstance(op, transform.NamedSequenceOp)
+                and op.sym_name.data == entry_point
+            ):
+                return op
+        raise PassFailedException(
+            f"{root} could not find a nested named sequence with name: {entry_point}"
+        )
```
superlopuh, line 35: it feels like we should be able to use the SymbolTable mechanism here instead of walking the whole module, or something that more directly looks first for the right module, then looks up the function by name

tests/filecheck/dialects/transform/transform_interpreter.mlir
```diff
@@ -0,0 +1,17 @@
+// RUN: xdsl-opt %s -p transform-interpreter | filecheck %s
+
+
+module {
+
+  "builtin.module"() ({
+    "transform.named_sequence"() <{arg_attrs = [{transform.readonly}], function_type = (!transform.any_op) -> (), sym_name = "__transform_main"}> ({
```
superlopuh, line 7: why not custom syntax here?

superlopuh: I'd rather the tests were using custom syntax if possible, adding it in a separate PR would be nice to make the tests more readable.

xdsl/transforms/transform_interpreter.py
efferifick: This mimics the implementation in MLIR. Can you specify a nested module where the nested module has no symbol? I.e.,

```
module {
  module {
    transform.named_sequence @__main() { // <--- this one?
    }
  }
}
```

tests/filecheck/dialects/transform/transform_interpreter.mlir
efferifick: The named sequence op [does not yet have a custom syntax.](https://github.com/xdslproject/xdsl/blob/main/xdsl/dialects/transform.py#L737-L782) I can change the other ones, but I rather have small commits than a big one that does many changes, so the custom syntax can be a different commit.

xdsl/transforms/transform_interpreter.py
efferifick: https://github.com/xdslproject/xdsl/pull/4289/commits/1b8391835ca362b8f2c2212cc1aaf5b613830c63

xdsl/transforms/transform_interpreter.py
efferifick: I cannot test with filecheck with expected-error@below. But I added a test with another entry point and a pytest for the exception.

xdsl/transforms/transform_interpreter.py
superlopuh: Oh interesting I'm not sure why they did it that way but I'm happy to mirror their mechanism.

tests/transforms/test_transform_interpreter.py
```diff
@@ -0,0 +1,13 @@
...
+from xdsl.transforms.transform_interpreter import TransformInterpreterPass
+from xdsl.utils.exceptions import PassFailedException
+
+
+def test_entry_point_not_found():
+    root = builtin.ModuleOp(ops=[])
+    entry_point = "symbol-name"
+    msg = f"could not find a nested named sequence with name: {entry_point}"
+    with pytest.raises(PassFailedException, match=msg):
+        TransformInterpreterPass.find_transform_entry_point(root, "symbol-name")
```
superlopuh, line 13: nit:
```suggestion
    with pytest.raises(PassFailedException, match="could not find a nested named sequence with name: symbol-name"):
        TransformInterpreterPass.find_transform_entry_point(root, "symbol-name")
```

superlopuh (approved):

tests/transforms/test_transform_interpreter.py
efferifick: https://github.com/xdslproject/xdsl/pull/4289/commits/d00bb653a1d773e21e7423c1cba4742f6138bb25

PR #4290, opened 2025-04-27
dialects: (transform) Add custom syntax for `transform.named_sequence`

efferifick:
Also possibly change the name of the operands and properties to match funcOp


superlopuh (commented):
Could you please add a test checking syntax compatibility with MLIR?

superlopuh (approved):
Beautiful

PR #4292, opened 2025-04-28
dialects: (builtin) add generic parameter to DenseIntOrFPElementsAttr

alexarice:
A much simpler version of #3492 using the default valued typevars. The advantage of this is that you can now represent for example `DenseIntOrFPElementsAttr[AnyFloat]` as a type.

I'm not totally happy with this as there doesn't seem to be a good way to write the extraction methods (`get_attrs`/`iter_attrs`) in a way which actually respects the generic type, so any suggestions would be welcome. 

This PR does not attempt to specialise the use of `DenseIntOrFPElementsAttr` anywhere. I also didn't touch the builder methods `tensor_from_list` or `vector_from_list` or `from_list` as I personally believe it might be possible/best to remove the methods in the future, favouring `create_dense_{float/int/index}`.

### DIFF
--- a/xdsl/interpreters/builtin.py
+++ b/xdsl/interpreters/builtin.py
@@ -86,7 +86,7 @@ class BuiltinFunctions(InterpreterFunctions):
         attr: Attribute,
         type_attr: builtin.MemRefType[Any],
     ) -> ShapedArray[Any]:
-        assert isinstance(attr, builtin.DenseIntOrFPElementsAttr)
+        assert isa(attr, builtin.DenseIntOrFPElementsAttr)
         shape = attr.get_shape()
         data = attr.get_values()
         data_ptr = ptr.TypedPtr[Any].new(
--- a/xdsl/backend/csl/print_csl.py
+++ b/xdsl/backend/csl/print_csl.py
@@ -454,8 +454,8 @@ class CslPrintContext:
                 return str(val.data)
             case StringAttr() as s:
                 return f'"{s.data}"'
-            case DenseIntOrFPElementsAttr(type=typ):
-                return f"{self.mlir_type_to_csl_type(typ)} {{ {', '.join(self.attribute_value_to_str(a) for a in attr.iter_attrs())} }}"  # noqa: E501
+            case DenseIntOrFPElementsAttr():
+                return f"{self.mlir_type_to_csl_type(attr.get_type())} {{ {', '.join(self.attribute_value_to_str(a) for a in attr.iter_attrs())} }}"  # noqa: E501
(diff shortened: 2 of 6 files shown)


jorendumoulin (approved):
This would be very nice!

How about using overloads for the `get_attrs`?
```
    @overload
    def get_attrs(self: DenseIntOrFPElementsAttr[IntegerType]) -> Sequence[IntegerAttr[IntegerType]]: ...
    @overload
    def get_attrs(self: DenseIntOrFPElementsAttr[IndexType]) -> Sequence[IntegerAttr[IndexType]]: ...
    @overload
    def get_attrs(self: DenseIntOrFPElementsAttr[AnyFloat]) -> Sequence[FloatAttr[AnyFloat]]: ...
    @overload # in case generic type is not known
    def get_attrs(self: DenseIntOrFPElementsAttr[AnyDenseElement]) -> Sequence[IntegerAttr | FloatAttr]: ...
    def get_attrs(self) -> Sequence[IntegerAttr | FloatAttr]:
 ```
 
 The `_from_list` constructors have always felt a bit weird to me, so imo it makes sense to migrate away from them.

**Comment by alexarice** [comment]
I'll give the overloads a try, feel I don't have a good grasp of when they'll work in python

**Comment by alexarice** [comment]
The overloads seem to work great, though I think it might be best to make that into a new PR

**Comment by alexarice** [comment]
I've tried the overloads a bit more and there's some really weird stuff happening. If you have a `DenseIntOrFPElementAttr[Unknown]` and call `get_attrs` on it then the first overload triggers for some reason and you get a `Sequence[IntegerAttr[IntegerType]]` which isn't ideal.

File: xdsl/interpreters/builtin.py [hunk 1]
```diff
@@ -86,7 +86,7 @@ def dense_int_or_fp_elements_value(
         attr: Attribute,
         type_attr: builtin.MemRefType[Any],
     ) -> ShapedArray[Any]:
-        assert isinstance(attr, builtin.DenseIntOrFPElementsAttr)
+        assert isa(attr, builtin.DenseIntOrFPElementsAttr)
```
superlopuh, line 89: is this required? does `isattr` work instead by any chance?

superlopuh (approved):
LGTM

xdsl/interpreters/builtin.py
alexarice: `isattr` would just call `isa` right?

superlopuh: Yeah on second thought this is fine.

PR #4294, opened 2025-04-28
dialects: (irdl) add name properties

alexarice:
Updates irdl to the version in MLIR master.

I think we could benefit from some mlir interop tests here, but xdsl and mlir seems to treat symbols slightly differently which could make this difficult.

### DIFF
--- a/tests/dialects/test_irdl.py
+++ b/tests/dialects/test_irdl.py
@@ -3,7 +3,7 @@ from typing import ClassVar
 
 import pytest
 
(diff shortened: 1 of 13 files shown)


tests/dialects/test_irdl.py
```diff
@@ -47,44 +43,6 @@ def test_named_region_op_init(
     assert len(op.body.blocks) == 1
 
 
-def test_parameter_op_init():
```
alexarice, line 50: Simplified the init method and so removed the old test for it

math-fehr (approved):
👍

PR #4296, opened 2025-04-28
bench: Fix issues in benchmark warmed timeit for computers with low precision timers

EdmundGoodman:
The highest resolution timer on Apple Silicon Macs is `mach_time` at around 40ns, and other computers may have even lower resolution timers. Some microbenchmarks' runtimes are within an order of magnitude of this resolution, reducing the reliability of the results.

This change to the warmed timeit function adds batching for very short microbenchmarks to ensure that their runtimes are 25x greater than this resolution for reliable results.

### DIFF
--- a/benchmarks/bench_utils.py
+++ b/benchmarks/bench_utils.py
@@ -3,6 +3,7 @@
 
 import cProfile
 import gc
+import math
 import subprocess
 import time
 from argparse import ArgumentParser, Namespace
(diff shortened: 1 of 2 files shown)


EdmundGoodman:
Having done a bit more reading around for the related work section, it looks like `pyperf` might support this functionality. Will update PR to see if it can be drop-in replaced.

EdmundGoodman:
Superseded by changes batched in https://github.com/xdslproject/xdsl/pull/4522

superlopuh:
I'm confused, this PR seems more focused, why batch all the changes together?

PR #4297, opened 2025-04-28
bench: Make the `has_trait` microbenchmark comparable to the how slow is MLIR version

EdmundGoodman:
The "How Slow is MLIR?" benchmark for the `hasTrait` method checks the `SingleBlock` trait (the fourth of eight) on the `OpWithRegion` operation. For fair comparison, the xDSL version must also measure getting the fourth trait of eight from an operation. This PR makes this true in the microbenchmark implementation.


benchmarks/microbenchmarks.py
```diff
@@ -35,9 +35,12 @@ class HasTraitAOp(IRDLOperation):
...
 
 
-NUM_CONSTRUCTED_TRAITS = 100
+NUM_CONSTRUCTED_TRAITS = 8
 for i in range(NUM_CONSTRUCTED_TRAITS):
-    exec(f"class Trait{i}(OpTrait): pass")
+    exec(f"""\
+class Trait{i}(OpTrait):
+    pass
+""")
```
math-fehr, line 43: I think instead of calling `exec` to create a new class, you can probably have something like:
```
traits: dict[int, type(OpTrait)] = {}
def foo(i: int):
  if i in traits:
    return traits[i]
  else:
    class Trait(OpTrait):
      pass
    return Trait
```
This should work right?

math-fehr (approved):
Looks good!
How far are we compared to MLIR there?

benchmarks/microbenchmarks.py
EdmundGoodman: Thanks -- that's definitely nicer!

PR #4298, opened 2025-04-28
bench: Add profiler for python bytecode

EdmundGoodman:
For in-detail examination of constant folding workloads, we want to be able to examine the bytecode operations being run by CPython interpreter. The existing visualisation tool is flawed as it cannot extract the bytecode from functions with complex invocations (decorated, meta-programmed, ...), but more fundamentally it shows the bytecode of each function invoked, as opposed to the stream of bytecode operations invoked.

As such, this PR introduces a new bytecode profiling tool. This tool shows the stream of bytecode operations executed by the interpreter, in the context of the lines and functions from which they are dispatched. In addition to this, it supports performance profiling functionality to examine the amount of time spent on each opcode in the program.


benchmarks/bytecode/bytecode_profiler.py
```diff
@@ -0,0 +1,593 @@
+"""A bytecode performance profiler."""
+# TODO: Investigate specialising adaptive interpreter
+# TODO: `OpcodeEvent` could just store the instruction and make existing variables properties?
+# TODO: Split out into separate project
+# TODO: Make fun textual UI
```
superlopuh, line 5: I would avoid TODOs. I would recommend the following approach for each task:

1. we definitely want it, and don't want to merge in this PR, add an issue
2. we might want it, not worth an issue nor a TODO

There's a risk of issue/todo fatigue, where it gets difficult to actually find the important things to tackle in the sea of things to do.

benchmarks/bytecode/bytecode_profiler.py
```diff
@@ -0,0 +1,593 @@
...
+                    func(*args, **kwargs)
+                    finish_timestamp = perf_counter()
+                except Exception as _exc:
+                    finish_timestamp = perf_counter()
+                assert start_timestamp is not None
+                self._uninstrumented_time += finish_timestamp - start_timestamp
+            if self.num_repeats != 0:
+                self._uninstrumented_time /= self.num_repeats
+
+            # Trace the function collecting the event data to corrollate
```
superlopuh, line 340:
```suggestion
            # Trace the function collecting the event data to correllate
```

benchmarks/bytecode/bytecode_profiler.py
```diff
@@ -0,0 +1,593 @@
...
+    def calculate_elapsed_times(self) -> list[float]:
+        """Calculate the elapsed time for each event."""
+        timestamp_repeats: list[list[EventTimestamp]] = []
+        opcode_events = [
+            e
+            for e in self._events
+            if isinstance(e, OpcodeEvent) or isinstance(e, StopEvent)
+        ]
+        for i in range(0, len(self._timestamps), len(opcode_events)):
+            timestamp_repeats.append(self._timestamps[i : i + len(opcode_events)])
```
superlopuh, line 415:
```suggestion
        timestamp_repeats = tuple(itertools.batched(len(opcode_events)))
```

benchmarks/bytecode/bytecode_profiler.py
```diff
@@ -0,0 +1,593 @@
...
+
+        elapsed_time_repeats: list[list[float]] = []
+        for timestamps in timestamp_repeats:
+            elapsed_times: list[float] = [0.0 for _ in range(len(timestamps))]
+            for i, timestamp in enumerate(timestamps):
+                if i == 0:
```
superlopuh, line 421:
```suggestion
                if not i:
```

benchmarks/bytecode/bytecode_profiler.py
```diff
@@ -0,0 +1,593 @@
...
+        raise_exception()
+        inner_function(5.0)
+
+    nop_sled = BytecodeProfiler.get_nop_sled(50)
+
+    def go():
+        example_function()
+
+    profile_bytecode(go)
+    # print_bytecode(go)
```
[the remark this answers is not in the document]
superlopuh: If we can't it would be nice to have a comment to indicate how to execute this and what the expected result should be.

benchmarks/bytecode/bytecode_profiler.py
EdmundGoodman: Needs to be:
```
timestamp_repeats = tuple(
    itertools.batched(self._timestamps, len(opcode_events))
)
```
but otherwise agreed

benchmarks/bytecode/bytecode_profiler.py
EdmundGoodman: To be honest I'm not a big fan of relying on the falsiness of integers here, I think it just makes it harder to read the intention that is we want to skip the first item

benchmarks/bytecode/bytecode_profiler.py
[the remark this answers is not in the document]
EdmundGoodman: Will change this out -- this was just my own debugging code I should've got rid of before marking the PR as ready. I'll add some documentation but I think leave testing out for now, especially as it is likely transient on the bytecode spec across python versions

superlopuh (approved):

EdmundGoodman:
Superseded by https://github.com/xdslproject/xdsl/pull/4495

PR #4299, opened 2025-04-28
bench: Add support for memory profiling tool

EdmundGoodman:
Add support for leveraging `memory_profiler` to examine the memory usage of microbenchmarks.


math-fehr (approved):

EdmundGoodman:
Looks like pyright might have version bumped, breaking some ignores in a file this PR doesn't touch? Probably worth fixing that separately, then can merge this after?

superlopuh:
@EdmundGoodman let's iterate and close this on Monday?

EdmundGoodman:
Closed as unnecessary -- memory profiling is not a current use case of the benchmarking infrastructure. May be re-opened in future if dependency is useful.

superlopuh:
I'm not entirely sure of the downside of merging this even if it's not used in your dissertation, seems like a useful tool to have available?

PR #4241, opened 2025-04-14
[WIP] bench: specialised pattern rewriting for integer addition constant folding

EdmundGoodman:
The most minimal implementation of pattern rewriting which still supports integer addition constant folding.

TODO:
- [x] #3961, remove indirection in operation cloning
- [x] #4163, lazily include `ImplicitBuilder` in `Operation.__post_init__`
- [x] Remove `UnregisteredOp` case from `has_trait`
- [ ] Remove verification from `IntegerAttr.__init__` by assuming guarantees
- [ ] Move building irdl ops to creating operations
- [ ] Simplify data structure and remove extraneous work in region walking
- [ ] Simplify data structure and remove extraneous work in `replace_op`
- [ ] Add `__slots__` to data structures
- [ ] Delete any code (e.g. dialects/transforms/...) not required by workload, estimable by `pytest tests/transforms/test_constant_folding_simple.py  -W error --cov`

PR #4300, opened 2025-04-29
[WIP] bench: Specialised implementation of constant folding for performance evaluation

EdmundGoodman:
Implementing the most minimal implementation of pattern rewriting which still supports integer addition constant folding.

Replaces #4241 which had a different objective of cutting out functionality from the codebase, whereas this implementation inlines the code and only elides from the inlined code.

### DIFF
--- a/benchmarks/rewriting.py
+++ b/benchmarks/rewriting.py
@@ -24,6 +24,7 @@ from xdsl.traits import (
 )
 from xdsl.transforms.canonicalization_patterns.utils import const_evaluate_operand
 from xdsl.transforms.canonicalize import CanonicalizePass
+from xdsl.transforms.constant_folding_simple import ConstantFoldingSimplePass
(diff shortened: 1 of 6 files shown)


benchmarks/bench_utils.py
```diff
@@ -287,6 +285,8 @@ def dis_benchmark(
     benchmarks: dict[str, Benchmark],
 ):
     """Use dis to disassemble a benchmark."""
+    from bytesight import profile_bytecode
```
superlopuh, line 306: let's do this as a quick first PR?

EdmundGoodman: #4495

EdmundGoodman:
Add test_ suffix to transform name

EdmundGoodman:
Functionality merged by #4522

PR #4303, opened 2025-04-30
core: Create BuiltinAttribute class

math-fehr:
This class should be used for builtin attributes that are printed differently than "user-defined" attributes. This include for instance `vector`, `integer`, etc...

All non-`BuiltinAttribute` class should have the `dialect.name` syntax for their name

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -116,10 +116,12 @@ class Attribute(ABC):
         return res.getvalue()
 
 
-class BuiltinAttribute(Attribute):
+class BuiltinAttribute(Attribute, ABC):
     """
-    This class is used to mark builtin attributes. Builtin attributes are handled
-    differently when parsing and printing.
+    This class is used to mark builtin attributes.
+    Unlike other attributes in MLIR, printing and parsing of *Builtin*
+    attributes is handled directly by the parser.
+    Attributes outside of the `builtin` dialect should not inherit from `BuiltinAttribute`.
     """
 
     pass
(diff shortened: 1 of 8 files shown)


alexarice (approved):

xdsl/ir/core.py
```diff
@@ -116,6 +116,15 @@ def __str__(self) -> str:
         return res.getvalue()
 
 
+class BuiltinAttribute(Attribute):
+    """
+    This class is used to mark builtin attributes. Builtin attributes are handled
+    differently when parsing and printing.
+    """
```
superlopuh, line 123:
```suggestion
    """
    This class is used to mark builtin attributes. 
    Unlike other attributes in MLIR, printing and parsing of *Builtin* 
    attributes is handled directly by the parser.
    Attributes outside of the `builtin` dialect should not inherit from `BuiltinAttribute`.
    """
```

xdsl/ir/core.py
```diff
@@ -116,6 +116,15 @@ def __str__(self) -> str:
         return res.getvalue()
 
 
+class BuiltinAttribute(Attribute):
```
superlopuh, line 119: ?
```suggestion
class BuiltinAttribute(Attribute, ABC):
```

xdsl/irdl/attributes.py
```diff
@@ -87,6 +88,32 @@ def generic_constraint_coercion(args: tuple[Any]) -> AttrConstraint:
 ParameterDef = Annotated[_A, IRDLAnnotations.ParamDefAnnot]
 
 
+def check_attr_name(cls: type):
```
superlopuh, line 91: Could you please add some tests here?

superlopuh (approved):
nice

PR #4308, opened 2025-05-01
backend: (riscv) target-independent LIFO register queue

superlopuh:
This concludes the migration of the register queue itself to be basically target-independent. When the time comes, we can add fairly short target-specific queues that specify the registers reserved and available by the platform ABI.

### DIFF
--- a/xdsl/backend/register_queue.py
+++ b/xdsl/backend/register_queue.py
@@ -1,10 +1,12 @@
 from abc import ABC, abstractmethod
-from collections.abc import Sequence
+from collections import defaultdict
+from collections.abc import Iterable, Sequence
 from contextlib import contextmanager
-from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


xdsl/backend/register_queue.py
```diff
@@ -64,3 +66,143 @@ def exclude_register(self, reg: _T) -> None:
...
+        return ()
+
+    @classmethod
+    def default_available_registers(cls) -> Iterable[_T]:
+        """
+        The default registers to be made available when instantiating the queue.
+        """
+        return ()
+
+    @classmethod
```
superlopuh, line 111: from here and below, the code is a copy/paste from `RiscvRegisterQueue`, changing `RISCVRegisterType` to `_T`

emmau678 (approved):

PR #4309, opened 2025-05-01
backend: always exclude pre-allocated registers during regalloc

superlopuh:
In practice I don't think it's ever correct to exclude the preallocated registers, might as well drop this.


compor (approved):
We never used this, so it's fine.

PR #4314, opened 2025-05-02
backend: add x86-allocate-registers

superlopuh:
Adds the pass and a test, and modifies the end-to-end test to include register allocation in the pipeline.


qaco:
If I understand well, it is not ready for review ?

superlopuh:
Yeah, these are the overall changes, I'm opening small PRs for everything that's not x86-specific, and will open this one at the end (in about 3 PRs)

qaco (approved):

PR #4316, opened 2025-05-02
core: (ParsePropInAttrDict) allow a subset of properties to be parsed in attr-dict

### LINKED ISSUE #4306: Assembly format requires all properties to be in the assembly format string
Difference in behaviour with MLIR.

Take for example `transform.apply_registered_pass` op. It's is defined upstream as:

```

  let arguments = (ins TransformHandleTypeInterface:$target,
                       StrAttr:$pass_name,
                       DefaultValuedAttr<StrAttr, "\"\"">:$options);
  let results = (outs TransformHandleTypeInterface:$result);
  let assemblyFormat = [{
    $pass_name `to` $target attr-dict `:` functional-type(operands, results)
  }];
```

which might show the follow assembly format:

```
    transform.apply_registered_pass "canonicalize" to %1 {options = "invalid-option=1"} : (!transform.any_op) -> !transform.any_op
```

xDSL reports the following error:

```
$pass_name `to` $target attr-dict `:` functional-type(operands, results)
# |                                                                         ^
# |                                                                         options properties are missing from the declarative format. If this is intentional, consider using 'ParsePropInAttrDict' IRDL option.
```

I don't believe `ParsePropInAttrDict` is appropriate though.

alexarice:
Fixes #4306 

I'm not sure this is 100% what MLIR does in every case, but it passes all our tests at seems at least close to mlir behaviour in the cases I tested.

@erick-xanadu can't seem to add you as a reviewer for some reason, but would be good if you could check if this fixes the issues you were having.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -426,13 +426,11 @@ class AttrDictDirective(FormatDirective):
                 "dictionary."
             )
 
-        parsed = bool(res)
-
         props = tuple(k for k in res.keys() if k in self.expected_properties)
         for name in props:
             state.properties[name] = res.pop(name)
         state.attributes |= res
-        return parsed
+        return bool(res) or bool(props)
 
     def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> None:
         if not op.attributes.keys().isdisjoint(self.expected_properties):
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -150,19 +150,11 @@ class FormatProgram:
                 "Variadic or optional operand has no type or a single type "
                 operands.append(parser.resolve_operands(uo, ot, parser.pos))
 
-        # Get the properties from the attribute dictionary if no properties are
-        # defined. This is necessary to be compatible with MLIR format, such as
-        # `memref.load`.
-        if state.properties:
-            properties = state.properties
-        else:
-            properties = op_def.split_properties(state.attributes)
-
         return op_type.build(
             result_types=result_types,
             operands=operands,
             attributes=state.attributes,
-            properties=properties,
+            properties=state.properties,
             regions=state.regions,
             successors=state.successors,
         )
@@ -411,9 +403,9 @@ class AttrDictDirective(FormatDirective):
     printed twice otherwise.
     """
 
-    print_properties: bool
+    expected_properties: set[str]
     """
-    If this is set, also print properties as part of the attribute dictionary.
+    Properties that should be printed and parsed as part of this attr-dict.
     This is used to keep compatibility with MLIR which allows that.
     """
 
@@ -423,7 +415,7 @@ class AttrDictDirective(FormatDirective):
             if res is None:
                 res = {}
             else:
-                res = res.data
+                res = dict(res.data)
         else:
             res = parser.parse_optional_attr_dict()
         defined_reserved_keys = self.reserved_attr_names & res.keys()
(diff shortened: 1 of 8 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -433,23 +425,28 @@ def parse(self, parser: Parser, state: ParsingState) -> bool:
...
                 "dictionary."
             )
+
+        parsed = bool(res)
+
+        props = tuple(k for k in res.keys() if k in self.expected_properties)
+        for name in props:
+            state.properties[name] = res.pop(name)
         state.attributes |= res
-        return bool(res)
```
superlopuh, line 437: seems like a spurious diff

xdsl/parser/core.py
```diff
@@ -856,7 +856,7 @@ def _parse_generic_operation(self, op_type: type[Operation]) -> Operation:
             successors = []
 
         # Parse attribute dictionary
-        properties = self.parse_optional_properties_dict()
+        properties = self.parse_optional_properties_dict() or {}
```
superlopuh, line 859: I know this is pedantic but it avoids an extra empty dict creation
```suggestion
        properties = self.parse_optional_properties_dict()
        if properties is None:
            properties = {}
```

xdsl/irdl/operations.py
```diff
@@ -1235,16 +1235,16 @@ def verify(self, op: Operation):
...
+    ):
         """
         Remove all entries of an attribute dictionary that are defined as properties
         by the operation definition, and return them in a new dictionary.
         """
-        properties: dict[str, Attribute] = {}
         for property_name in self.properties.keys():
-            if property_name in attr_dict:
+            if property_name in attr_dict and property_name not in properties:
                 properties[property_name] = attr_dict.pop(property_name)
```
superlopuh, line 1247: this is sketchy, I would avoid modifying an argument if possible, let's return new properties instead?

superlopuh: Would it make sense to separate this change into its own PR? Seems like it's a matter of initializing the result with properties passed in, can probably be done independently of the other changes being reviewed.

xdsl/irdl/declarative_assembly_format.py
alexarice: I disagree

xdsl/irdl/operations.py
alexarice: This function is now only used in one place, so I could inline it if that is preferable, I'm not really sure what there is to separate into a separate PR

superlopuh: Yeah inlining also seems good to me.

xdsl/irdl/declarative_assembly_format.py
superlopuh: I mean is `parsed` used anywhere other than to return it? Why create a local value then not use it?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -433,23 +425,28 @@ def parse(self, parser: Parser, state: ParsingState) -> bool:
...
-                    "in a single dictionary"
-                )
-            op_def = op.get_irdl_definition()
-            dictionary = op.attributes | op.properties
-            defs = op_def.properties | op_def.attributes
-        else:
-            op_def = op.get_irdl_definition()
-            dictionary = op.attributes
-            defs = op_def.attributes
+        if any(name in op.attributes for name in self.expected_properties):
```
superlopuh, line 438: nit:
```suggestion
        if op.attributes & self.expected_properties:
```

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -780,23 +768,22 @@ def parse_format_directive(self) -> FormatDirective:
             return variable
         self.raise_error(f"unexpected token '{self._current_token.text}'")
 
-    def create_attr_dict_directive(self, with_keyword: bool) -> AttrDictDirective:
+    def create_attr_dict_directive(
+        self, with_keyword: bool, idx: int
+    ) -> AttrDictDirective:
         """Create an attribute dictionary directive, and update the parsing state."""
-        if self.has_attr_dict:
+        if self.attr_dict_idx != -1:
```
superlopuh, line 775: Would it be possible to instead move this logic into the `parse_format` method? It feels a little more pertinent there, as the `AttrDictDirective` itself doesn't need the index, and it's more of a property of the whole format than the individual directive.

xdsl/irdl/declarative_assembly_format.py
alexarice: res is modified after `parsed` is defined

math-fehr:
I'll look at it today, I'm trying to check what happens in MLIR

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: I've moved the logic to `verify_attr_dict`

xdsl/irdl/declarative_assembly_format.py
alexarice: Refactored to be nicer

superlopuh: Ty

superlopuh (approved):
Code looks good! Thank you for iterating.

math-fehr (approved):
I think you are doing something strictly better than MLIR, so that's good to me!
(What MLIR does is that it first merges attributes and property dictionary, then removes the properties that were already present in the textual format).

alexarice:
I think the main difference I could find is that in xDSL you can do:
```mlir
"test.op"() <{hello = "hello"}> {hello = "goodbye"} : () -> ()
```
as mlir would rewrite one of them (or throw an error?) Could probably make this all line up if we wanted but this feels like a major edge case/anti-pattern to begin with.

math-fehr:
Yeah, please don't follow MLIR here, this is probably something that MLIR should return an error for.

math-fehr:
So what you have now is perfect IMO

alexarice:
Great, I'll merge this and we can iterate if it didn't fix the original issue fully.

PR #4318, opened 2025-05-02
backend: add RegisterAllocatableOperation abstract base class

superlopuh:
This is half of the class, in a subsequent PR there will actually be an `allocate` method for each subclass to override.

### DIFF
--- a/xdsl/backend/register_allocatable.py
+++ b/xdsl/backend/register_allocatable.py
@@ -1,8 +1,43 @@
 import abc
-from collections.abc import Sequence
+from collections.abc import Iterator, Sequence
 from typing import NamedTuple
 
-from xdsl.ir import SSAValue
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import Operation, Region, SSAValue
+
+
+class RegisterAllocatableOperation(Operation, abc.ABC):
+    """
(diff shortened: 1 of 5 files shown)


xdsl/backend/register_allocatable.py
```diff
@@ -1,8 +1,43 @@
...
-from collections.abc import Sequence
+from collections.abc import Iterator, Sequence
 from typing import NamedTuple
 
-from xdsl.ir import SSAValue
+from xdsl.backend.register_type import RegisterType
+from xdsl.ir import Operation, Region, SSAValue
+
+
+class RegisterAllocatableOperation(Operation, abc.ABC):
```
qaco, line 9: If I understand well, for now we don't use it as a base class but we leverage its static method as an helper ?

superlopuh: It's used as a base class for `HasRegisterConstraints` below, which is used in the current riscv register allocation.

superlopuh: As you can see in `riscv_func`, the default implementation of `iter_used_registers` can be overridden to include things like registers not preserved during function calls

qaco: Thanks !

qaco (approved):

PR #4321, opened 2025-05-05
dialects: (builtin) remove TensorOrMemRefConstraint

alexarice:
This seemed to only be used in two places, both of which could trivially be replaced. Is anyone aware of the reason this exists? (e.g. is it used downstream?)


math-fehr (approved):
No clue how this works upstream, but here it's clear that this is some premature factoring at least.

superlopuh (approved):
Nice! I think we've added quite a lot of helpers since this was added, making this nicer.

Issue #4238, opened 2025-04-14
AttributeError: 'DenseResourceAttr' object has no attribute 'get_type'


Jimmy2027:
### Description

When trying to parse the following mlir code:
```mlir
module {
    %cst = arith.constant dense_resource<torch_tensor_4_3_torch.float32> : tensor<4x3xf32>
    }

{-#
  dialect_resources: {
    builtin: {
      torch_tensor_4_3_torch.float32: "0x04000000D2E86BBE4318F73D58574DBE71A7E4BEE2DA823E1BCA0CBFF4511E3F8AA53A3D4A5671BFBC744EBFD12B8ABF12ECD2BF"
    }
  }
#-}

```
like this:
```python
from xdsl.context import Context
from xdsl.dialects import get_all_dialects
from xdsl.parser import Parser

ctx = Context(True)


for dialect_name, dialect_factory in get_all_dialects().items():
    ctx.register_dialect(dialect_name, dialect_factory)

mlir_code = """
module {
    %cst = arith.constant dense_resource<torch_tensor_4_3_torch.float32> : tensor<4x3xf32>
    }

{-#
  dialect_resources: {
    builtin: {
      torch_tensor_4_3_torch.float32: "0x04000000D2E86BBE4318F73D58574DBE71A7E4BEE2DA823E1BCA0CBFF4511E3F8AA53A3D4A5671BFBC744EBFD12B8ABF12ECD2BF"
    }
  }
#-}

"""

parser = Parser(ctx, mlir_code)
module = parser.parse_module()
print(module)
```
I get this error:
`AttributeError: 'DenseResourceAttr' object has no attribute 'get_type'`
### Expected vs Actual Behavior

`DenseResourceAttr` is a `ParametrizedAttribute` and hence does not have the `get_type` method. For some reason it gets parsed as a `TypedAttribute`.

### Steps to Reproduce and a Minimal Working Example (MWE)

1. Run the above python code

### Additional Information

This is the full trace:
```
Traceback (most recent call last):
  File "/home/hjk/projects/tflite_to_tosa/iree/frontend-mess/xdsl_test/dense_resource_attr/minimal.py", line 27, in <module>
    module = parser.parse_module()
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/hjk/projects/tflite_to_tosa/iree/.venv/lib/python3.11/site-packages/xdsl/parser/core.py", line 130, in parse_module
[...]

PR #4322, opened 2025-05-05
core: (constraints) remove variance on `get_variable_extractors`

alexarice:
I'm pretty sure there was some type unsoundness created by having the parameter on `GenericAttrConstraint` be covariant and then having the extractors use this type parameter. This was highlighted by #4238 , where the `TypedAttributeConstraint` extractor encounters an attribute which is not a `TypedAttribute`.

This PR refactors `get_variable_extractors` to remove the variant parameter, adding in some new necessary checks within the extractor.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -85,7 +85,7 @@ from xdsl.utils.comparisons import (
     unsigned_upper_bound,
     unsigned_value_range,
 )
-from xdsl.utils.exceptions import DiagnosticException, VerifyException
+from xdsl.utils.exceptions import DiagnosticException, PyRDLError, VerifyException
 from xdsl.utils.hints import isa
 from xdsl.utils.isattr import isattr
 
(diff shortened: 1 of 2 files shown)


superlopuh (approved):

math-fehr (commented):
Just so I understand correctly.
We cannot have the covariant here because we don't know if the root attribute we are using the extractor for is correct, and thus if our extractor will return the right type?

alexarice:
Suppose A is a subtype of B

The covariance of `GenericAttrConstraint` means a `GenericAttrConstraint[A]` is a subtype of `GenericAttrConstraint[B]`. Therefore if I have a `GenericAttrConstraint[B]`, it is also a `GenericAttrConstraint[B]` and so I can call `get_variable_extractors` on it to get a bunch of `B` extractors, but `B` extractors are not `A` extractors so this is unsound

alexarice:
The linked issue is a case where an `arith.constant` was initialised with something that wasn't a `TypedAttribute`, and this got passed straight into the extractor of for `TypedAttributeConstraint` which assumed it was a `TypedAttribute`

math-fehr:
What can an extractor do besides grabbing a parameter from an attribute? Can you modify the attribute or something?

alexarice:
It should just be grabbing "subparts" of an attribute

alexarice:
It's just an extractor takes the attribute as an argument, so the variance effectively gets flipped

alexarice:
I'm surprised pyright doesn't complain tbh

math-fehr:
Okay, I need to look more at the related PR, because otherwise to me that looks correctly covariant no?
If you have `A` a subtype of `B`, then `Extractor[A]` should be an `Extractor[B]` no? e.g., if you return an `A`, the `A` is a `B`, so it's like you returned a `B`? Or am I saying something dumb here?

math-fehr:
Ohhhh, my bad! Shouldn't it be contravariant rather than invariant then?

alexarice:
It could be contravariant, it's currently invariant (which is fine). The problem this PR solves is that `get_variable_extractors` treats it as covariant (which I am surprised in retrospect that pyright is ok with).

math-fehr (approved):

PR #4325, opened 2025-05-05
dialects: (arith) move select to assembly format

alexarice:
Moves `arith.select` to assembly format and uses constraints instead of `_verify`. Might be worth upstreaming the assembly format here to MLIR, since there's been a FIXME for over 3 years.


superlopuh (approved):
Nice

PR #4330, opened 2025-05-06
core: Separate attributes and types in the context

math-fehr:
With this change, it should be possible to have an attribute and a type of the same name, which is necessary to handle the MLIR `emitc` dialect.

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -104,7 +104,7 @@ def test_get_attr():
     ctx.load_attr_or_type(DummyAttr)
 
     assert ctx.get_attr("test.dummy_attr") == DummyAttr
-    with pytest.raises(Exception):
+    with pytest.raises(ValueError):
         _ = ctx.get_attr("test.dummy_attr2")
 
     assert ctx.get_optional_attr("test.dummy_attr") == DummyAttr
@@ -118,7 +118,7 @@ def test_get_type():
     ctx.load_attr_or_type(DummyType)
 
     assert ctx.get_type("test.dummy_type") == DummyType
-    with pytest.raises(Exception):
+    with pytest.raises(ValueError):
         _ = ctx.get_type("test.dummy_attr")
 
     assert ctx.get_optional_type("test.dummy_type") == DummyType
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -226,7 +226,7 @@ def test_load_registered_dialect():
     assert list(ctx.loaded_dialects) == [testDialect]
     assert list(ctx.registered_dialect_names) == ["test"]
     assert list(ctx.loaded_types) == [DummyType]
-    assert list(ctx.loaded_attrs) == [DummyOp]
+    assert list(ctx.loaded_attrs) == [DummyAttr]
 
 
 def test_load_registered_dialect_not_registered():
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -26,8 +26,13 @@ class DummyAttr2(ParametrizedAttribute):
     name = "test.dummy_attr2"
 
 
-testDialect = Dialect("test", [DummyOp], [DummyAttr])
-testDialect2 = Dialect("test", [DummyOp2], [DummyAttr2])
+@irdl_attr_definition
+class DummyType(ParametrizedAttribute, TypeAttribute):
+    name = "test.dummy_type"
+
+
+testDialect = Dialect("test", [DummyOp], [DummyAttr, DummyType])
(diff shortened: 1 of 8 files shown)


xdsl/context.py
```diff
@@ -19,6 +19,9 @@ class Context:
     _loaded_attrs: dict[str, type[Attribute]] = field(
         default_factory=dict[str, type[Attribute]]
     )
+    _loaded_types: dict[str, type[Attribute]] = field(
```
alexarice, line 22:
```suggestion
    _loaded_types: dict[str, type[TypeAttribute]] = field(
```
Is this worth doing?

alexarice (approved):

tests/test_context.py
```diff
@@ -106,14 +111,28 @@ def test_get_attr():
...
 
 
+def test_get_type():
+    """Test `get_type` and `get_optional_type` methods."""
+    ctx = Context()
+    ctx.load_attr_or_type(DummyAttr)
+    ctx.load_attr_or_type(DummyType)
+
+    assert ctx.get_type("test.dummy_type") == DummyType
+    with pytest.raises(Exception):
```
superlopuh, line 121: Can you make this more specific?

xdsl/context.py
math-fehr: I was thinking of it, and I personally have no clue. I could add that if you think it's worth

xdsl/context.py
```diff
@@ -108,11 +119,20 @@ def load_op(self, op: "type[Operation]") -> None:
...
-        self._loaded_attrs[attr.name] = attr
+    def load_attr_or_type(self, attr: type[Attribute]) -> None:
+        """
+        Load an attribute or type definition.
+        Attribute or type names should be unique, but an attribute may have the same
+        name as a type.
+        """
+        if issubclass(attr, TypeAttribute):
+            if attr.name in self._loaded_types:
+                raise Exception(f"Type {attr.name} has already been loaded")
```
superlopuh, line 130: is there a better error here? At least a ValueError IMO, but maybe a custom subclass?

xdsl/context.py
```diff
@@ -166,6 +186,43 @@ def get_op(
...
+
+    def get_type(self, name: str) -> type[Attribute]:
+        """
+        Get a type definition from its name.
+        If the type is not registered, raise an exception unless unregistered
+        types are allowed in the context, in which case return an UnregisteredAttr.
+        """
+        if attr_type := self.get_optional_type(name):
+            return attr_type
+        raise Exception(f"Type {name} is not registered")
```
superlopuh: KeyError?

math-fehr: I added a new exception for the unregistered things (dialects, ops, types, attributes)

xdsl/context.py
math-fehr: I added a new one

xdsl/utils/exceptions.py
```diff
@@ -15,6 +15,24 @@
...
+    pass
+
+
+class AlreadyRegisteredConstructException(Exception):
+    """
+    An exception raised when a dialect, operation, type,
+    or attribute is registered twice.
+    """
+
+    pass
```
superlopuh, line 33: nit:
```suggestion
class UnregisteredConstructException(Exception):
    """
    An exception raised when a dialect, operation, type,
    or attribute is not registered.
    """


class AlreadyRegisteredConstructException(Exception):
    """
    An exception raised when a dialect, operation, type,
    or attribute is registered twice.
    """
```

xdsl/context.py
```diff
@@ -50,6 +58,13 @@ def loaded_attrs(self) -> "Iterable[type[Attribute]]":
         """
         return self._loaded_attrs.values()
 
+    @property
+    def loaded_types(self) -> Iterable[type[TypeAttribute]]:
+        """
+        Returns all the loaded types. Not valid across mutations of this object.
+        """
+        return self._loaded_types.values()
```
superlopuh, line 66: just looking at the codecov, this and the helper above aren't tested.

xdsl/context.py
```diff
@@ -73,21 +88,23 @@ def register_dialect(
         requested with `load_registered_dialect`.
         """
         if name in self._registered_dialects:
-            raise ValueError(f"'{name}' dialect is already registered")
+            raise AlreadyRegisteredConstructException(
```
superlopuh, line 91: would be nice to add a test for this also

superlopuh: and the ones below

superlopuh: ah but maybe codecov is out of date

superlopuh (approved):

PR #4332, opened 2025-05-06
core: Add nicer error messages for RangeOf

math-fehr:
This PR adds the following prefix on a verification error for range length constraints:
"incorrect length for range variable:"

### DIFF
--- a/tests/irdl/test_operation_definition.py
+++ b/tests/irdl/test_operation_definition.py
@@ -19,6 +19,7 @@ from xdsl.dialects.test import TestType
 from xdsl.ir import Attribute, Block, Region
 from xdsl.irdl import (
     AnyAttr,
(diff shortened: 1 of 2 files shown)


alexarice (approved):
This seems reasonable, I feel at some point we may want to rethink how the verification is working to try to get better messages, i.e. for this example something like:
"'test.same_length' expected `  2 results but got 1"

In any case, this is likely a local improvement

superlopuh (approved):

math-fehr:
> This seems reasonable, I feel at some point we may want to rethink how the verification is working to try to get better messages, i.e. for this example something like: "'test.same_length' expected ` 2 results but got 1"
> 
> In any case, this is likely a local improvement

I agree at some point we should find a way to make this nicer!

PR #4334, opened 2025-05-07
dialects: (ptr) add FromPtrOp

superlopuh:
PR 1/2 adding the canonicalization patterns from [here](https://github.com/llvm/llvm-project/pull/137469/files) that remove chains of to_ptr and from_ptr ops in a simpler way than removing reinterpret casts.

### DIFF
--- a/xdsl/dialects/ptr.py
+++ b/xdsl/dialects/ptr.py
@@ -91,6 +91,16 @@ class ToPtrOp(IRDLOperation):
     assembly_format = "$source attr-dict `:` type($source) `->` type($res)"
 
 
+@irdl_op_definition
+class FromPtrOp(IRDLOperation):
+    name = "ptr_xdsl.from_ptr"
+
+    source = operand_def(PtrType)
+    res = result_def(MemRefType)
+
+    assembly_format = "$source attr-dict `:` type($source) `->` type($res)"
+
+
 Ptr = Dialect(
     "ptr_xdsl",
     [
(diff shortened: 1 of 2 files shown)


xdsl/dialects/ptr.py
```diff
@@ -91,6 +91,16 @@ class ToPtrOp(IRDLOperation):
...
 
 
+@irdl_op_definition
+class FromPtrOp(IRDLOperation):
+    name = "ptr_xdsl.from_ptr"
+
+    source = operand_def(PtrType)
+    res = result_def(MemRefType)
+
+    assembly_format = "$source attr-dict `:` type($source) `->` type($res)"
```
alexarice, line 101: The `type($source)` isn't necessary but I guess it makes the intent clearer.

Also, `functional-type` would put brackets round the operand types, I'm not sure if this is preferred by mlir, or if there is precedent for doing it this way too.

alexarice (approved):
Is there a reason to prefer these ops to casts? Just a bit more type safety?

xdsl/dialects/ptr.py
superlopuh: I followed the syntax in MLIR, which has some more potential information on the ptr type itself, such as memory space, which we haven't added yet.

alexarice: Ah, it wasn't clear to me that this was based off an mlir op (especially as the dialect is called `ptr_xdsl`)

superlopuh: It's called that because they haven't merged their things yet, and we're living in the future :)

PR #4340, opened 2025-05-07
dialects: (ptr) add inits to ptr dialect

superlopuh:
Small fixup PR adding inits for a prettier API.

### DIFF
--- a/xdsl/dialects/ptr.py
+++ b/xdsl/dialects/ptr.py
@@ -12,8 +12,15 @@ Current deviations:
  conflict.
 """
 
-from xdsl.dialects.builtin import AnyAttr, IntegerAttrTypeConstr, MemRefType, UnitAttr
-from xdsl.ir import Dialect, ParametrizedAttribute, TypeAttribute
+from xdsl.dialects.builtin import (
+    AnyAttr,
+    IndexType,
+    IntegerAttrTypeConstr,
+    IntegerType,
+    MemRefType,
+    UnitAttr,
+)
+from xdsl.ir import Attribute, Dialect, ParametrizedAttribute, SSAValue, TypeAttribute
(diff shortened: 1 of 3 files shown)


xdsl/dialects/ptr.py
```diff
@@ -32,6 +39,9 @@
 class PtrType(ParametrizedAttribute, TypeAttribute):
     name = "ptr_xdsl.ptr"
 
+    def __init__(self):
```
alexarice, line 42: Does this do anything?

xdsl/dialects/ptr.py
```diff
@@ -43,6 +53,9 @@ class PtrAddOp(IRDLOperation):
 
     assembly_format = "$addr `,` $offset attr-dict `:` `(` type($addr) `,` type($offset) `)` `->` type($result)"
 
+    def __init__(self, addr: SSAValue, offset: SSAValue):
```
alexarice, line 56:
```suggestion
    def __init__(self, addr: SSAValue | Operation, offset: SSAValue | Operation):
```
Are we moving away from this?

alexarice (approved):

xdsl/dialects/ptr.py
superlopuh: It removes the option of passing arguments to the init function

xdsl/dialects/ptr.py
superlopuh: I never liked it, so it's my current quiet rebellion. Note that now that we have generic SSA values, we can actually do things like specify the types of values, which is impossible if we also allow Operation.

alexarice: fair

qaco (approved):

PR #4341, opened 2025-05-07
core: deprecate isattr

superlopuh:
Let's keep it until the next release.

### DIFF
--- a/tests/test_is_satisfying_hint.py
+++ b/tests/test_is_satisfying_hint.py
@@ -14,7 +14,7 @@ from xdsl.dialects.builtin import (
 from xdsl.ir import Attribute, ParametrizedAttribute, SSAValue
 from xdsl.irdl import BaseAttr, EqAttrConstraint, ParameterDef, irdl_attr_definition
 from xdsl.utils.hints import isa
-from xdsl.utils.isattr import isattr
+from xdsl.utils.isattr import isattr  # pyright: ignore[reportDeprecated]
 from xdsl.utils.test_value import create_ssa_value
 
 
@@ -394,10 +394,15 @@ def test_literal():
 
 
 def test_isattr():
-    assert isattr(IntAttr(1), BaseAttr(IntAttr))
-    assert not isattr(IntAttr(1), BaseAttr(StringAttr))
-    assert isattr(IntAttr(1), EqAttrConstraint(IntAttr(1)))
-    assert not isattr(IntAttr(1), EqAttrConstraint(IntAttr(2)))
+    # We're testing the deprecated function, so we need to suppress the warnings
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=DeprecationWarning)
+        assert isattr(IntAttr(1), BaseAttr(IntAttr))  # pyright: ignore[reportDeprecated]
+        assert not isattr(IntAttr(1), BaseAttr(StringAttr))  # pyright: ignore[reportDeprecated]
+        assert isattr(IntAttr(1), EqAttrConstraint(IntAttr(1)))  # pyright: ignore[reportDeprecated]
+        assert not isattr(IntAttr(1), EqAttrConstraint(IntAttr(2)))  # pyright: ignore[reportDeprecated]
 
 
 ################################################################################
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1382,7 +1382,7 @@ class DenseArrayBase(ParametrizedAttribute, BuiltinAttribute):
         if isinstance(data_type, IntegerType):
             _data = cast(Sequence[int] | Sequence[IntAttr], data)
             return DenseArrayBase.create_dense_int(data_type, _data)
-        elif AnyFloatConstr.matches(data_type):
+        elif isinstance(data_type, AnyFloat):
             _data = cast(Sequence[int | float] | Sequence[FloatData], data)
             return DenseArrayBase.create_dense_float(data_type, _data)
         else:
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -87,7 +87,6 @@ from xdsl.utils.comparisons import (
 )
 from xdsl.utils.exceptions import DiagnosticException, PyRDLError, VerifyException
 from xdsl.utils.hints import isa
-from xdsl.utils.isattr import isattr
 
 if TYPE_CHECKING:
     from _typeshed import ReadableBuffer, WriteableBuffer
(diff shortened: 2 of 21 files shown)


tests/test_is_satisfying_hint.py
```diff
@@ -14,7 +14,7 @@
 from xdsl.ir import Attribute, ParametrizedAttribute, SSAValue
 from xdsl.irdl import BaseAttr, EqAttrConstraint, ParameterDef, irdl_attr_definition
 from xdsl.utils.hints import isa
-from xdsl.utils.isattr import isattr
+from xdsl.utils.isattr import isattr  # pyright: ignore[reportDeprecated]
```
superlopuh, line 17: makes sure we don't break dependencies

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -362,9 +360,8 @@ def verify_(self) -> None:
                 )
 
         # typecheck required (only) block arguments
-        assert isattr(
+        assert (AnyTensorTypeConstr | MemRefType.constr()).matches(
```
superlopuh, line 363: There are a few of these, and they are all ugly, I guess this is why we had the `TensorOrMemRefOf`. It should get a little prettier with `TensorType.constr()` when we add one in the future.

xdsl/dialects/builtin.py
```diff
@@ -1383,7 +1382,7 @@ def from_list(
         if isinstance(data_type, IntegerType):
             _data = cast(Sequence[int] | Sequence[IntAttr], data)
             return DenseArrayBase.create_dense_int(data_type, _data)
-        elif isattr(data_type, AnyFloatConstr):
+        elif AnyFloatConstr.matches(data_type):
```
alexarice, line 1385: This one could be an isinstance

xdsl/dialects/csl/csl_stencil.py
alexarice: We could have `AnyTensorOrMemRef` without it being its own custom constraint

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -539,21 +536,21 @@ def parse(cls, parser: Parser):
...
         res_type = parser.parse_attribute()
-        if isattr(res_type, stencil.StencilTypeConstr):
+        if stencil.StencilTypeConstr.matches(res_type):
             return cls.build(
                 operands=[temp],
                 result_types=[res_type.get_element_type()],
                 properties=props,
             )
-        elif isattr(res_type, base(TensorType[Attribute])):
+        elif AnyTensorTypeConstr.matches(res_type):
```
alexarice, line 545: Things like this could be `isa(res_type, TensorType)`. It might be worth thinking about which we prefer

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -586,7 +583,7 @@ def verify_(self) -> None:
                     f"stencil.StencilType or memref.MemRefType but found {self.op.type}"
                 )
         else:
-            if not isattr(self.op.type, AnyTensorTypeConstr | MemRefType.constr()):
+            if not (AnyTensorTypeConstr | MemRefType.constr()).matches(self.op.type):
```
alexarice, line 586: Similarly this could be `isa(self.op.type, TensorType | MemRefType)` I think

alexarice (commented):
I feel that a lot of these could be `isa` but I don't know if they should be `isa`. I'm also unsure how I feel about this being so different to `isinstance` at the moment.

math-fehr (approved):
I think Alex found the main few nits, otherwise the future is now ;)

xdsl/dialects/builtin.py
superlopuh: Argument of type "FloatAttr[Unknown]" cannot be assigned to parameter "data_type" of type "AnyFloat" in function "create_dense_float"

alexarice: Works fine on my computer (maybe you did `isinstance(data_type, FloatAttr)`?)

superlopuh:
@alexarice happier with current state? I was wondering whetehr I should get rid of a bunch of `.matches` calls that can be replaced with `isa`, and decided against too much of a refactor.

xdsl/dialects/stencil.py
```diff
@@ -1225,7 +1226,9 @@ def matches(attr: TensorType[Attribute], other: Attribute) -> bool:
     def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
         ctx_attr = constraint_context.get_variable(self.name)
         if ctx_attr is not None:
-            if isa(attr, TensorType[Attribute]) and TensorIgnoreSizeConstraint.matches(
+            if isa(
+                attr, TensorType[Attribute]
```
alexarice, line 1230:
```suggestion
                attr, TensorType
```
unrelated but may as well change while we're here?

xdsl/dialects/stencil.py
```diff
@@ -1215,7 +1214,9 @@ def get_effects(cls, op: Operation):
 
 class TensorIgnoreSizeConstraint(VarConstraint[Attribute]):
     @staticmethod
-    def matches(attr: TensorType[Attribute], other: Attribute) -> bool:
+    def ranks_and_element_types_match(
+        attr: TensorType[Attribute], other: Attribute
```
alexarice, line 1218:
```suggestion
        attr: TensorType, other: Attribute
```

xdsl/dialects/stencil.py
```diff
@@ -1526,7 +1529,9 @@ def verify_(self) -> None:
                 op_type = types[j]
                 if op_type != res_type and not (
                     isa(op_type, TensorType[Attribute])
```
alexarice, line 1531:
```suggestion
                    isa(op_type, TensorType)
```

xdsl/parser/attribute_parser.py
```diff
@@ -522,7 +521,7 @@ def parse_shape(self) -> tuple[list[int] | None, Attribute]:
 
     def _parse_complex_attrs(self) -> ComplexType:
         element_type = self.parse_attribute()
-        if not isattr(element_type, base(IntegerType) | AnyFloatConstr):
+        if not (base(IntegerType) | AnyFloatConstr).matches(element_type):
```
alexarice, line 524: this one could be `isinstance` I think

alexarice (approved):
I'm happy with this, will leave it up to you whether you want to apply my suggestions

math-fehr:
Oh by the way, should we keep the term `match`, or should we go with something like `verify` like in `IntConstraint`? Alternatively, should we rename `IntConstraint.verify` to `IntConstraint.check`?

superlopuh:
Maybe even `matches`?

math-fehr:
Unknown if `matches` is better than `match` though, usually we use things like `verify` rather than `verifies` while programming no?

superlopuh:
Well `verify` makes sense to me as an effectful active verb, and would expect `verifies` to return a bool.

math-fehr:
I would personally say that wether it returns a bool or an exception, it should be spelled the same tbh. But maybe that spelling is `verifies`, idk

PR #4345, opened 2025-05-08
dialects: (arm_neon) add VectorWithArrangement.from_vec_type

emmau678:
Adding a lowering from arith.mulf to arm neon fmul (mixed: vector/scalar)


xdsl/backend/arm_neon/lowering/convert_arith_to_neon.py
superlopuh, line 1: I would put this in `xdsl/backend/arm/lowering/`, and we need `__init__.py` in every folder or it's not going to be happy.

xdsl/backend/arm_neon/lowering/convert_arith_to_neon.py
```diff
@@ -0,0 +1,40 @@
+from xdsl.dialects.arm_neon import NeonArrangement
+from xdsl.dialects.builtin import Float16Type, Float32Type, Float64Type, VectorType
+
+
+def get_arrangement_from_vec_type(vec_type: VectorType) -> NeonArrangement:
```
superlopuh, line 5: Could this be a dictionary mapping things like `vector<8xf16>` to `.H` and so on? We can then raise a single exception saying something like `Invalid vector type for ARM NEON: vector<17xi5>`.

superlopuh: Feels like it could also be defined on the NeonArrangement class itself as a static method.

emmau678: I made these updates, but I think I will leave it at that for this PR and open a new one for the next changes, since plans have changed a bit since I started this

xdsl/dialects/arm_neon.py
```diff
@@ -98,6 +103,15 @@ def map_to_num_els(self):
...
         return map[self.name]
 
+    @staticmethod
+    def get_arrangement_from_vec_type(vec_type: VectorType):
+        map_vec_to_arr = {
+            "vector<8xf16>": NeonArrangement.H,
+            "vector<4xf32>": NeonArrangement.S,
+            "vector<2xf64>": NeonArrangement.D,
+        }
+        return map_vec_to_arr[str(vec_type)]
```
superlopuh, line 113: Can you please make this a map from the instance of the vector type instead, and move the dictionary outside the method so that it's not created every time the function is called? Can you please add tests for supported and unsupported vectors also?

emmau678: I'm not sure I follow. I thought you couldn't use types as keys (unless you implement a separate equals method or something similar)

emmau678: > I'm not sure I follow. I thought you couldn't use types as keys (unless you implement a separate equals method or something similar)

e.g, this was my understanding of what you described, but it doesn't work because of this:
```
def get_arrangement_from_vec_type(vec_type: VectorType):
        map_vec_to_arr = {
            VectorType(Float16Type(), [8]): NeonArrangement.H,
            VectorType(Float32Type(), [4]): NeonArrangement.S,
            VectorType(Float64Type(), [2]): NeonArrangement.D,
        }
        return map_vec_to_arr[vec_type]
```

xdsl/dialects/arm_neon.py
```diff
@@ -94,9 +99,24 @@ class NeonArrangement(StrEnum):
...
+
+
+map_arrangement_to_num_els = {"D": 2, "S": 4, "H": 8}
+
+map_vec_to_arr: dict[str, NeonArrangement] = {
+    "vector<8xf16>": NeonArrangement.H,
+    "vector<4xf32>": NeonArrangement.S,
+    "vector<2xf64>": NeonArrangement.D,
+}
```
superlopuh, line 119: A few changes:

1. The convention in Python is to use UPPER_SNAKE_CASE for constants
2. These constants can be private and live on classes for namespacing reasons
3. rename helper to be a bit more consise
4. use property for the O(1) lookup for self
5. move other dict into class also
6. rename static method
7. use .get for optional lookup
8. use vector type as key

```suggestion
    _SIZE_BY_ARRANGEMENT = {"D": 2, "S": 4, "H": 8}

    @property
    def arrangement(self):
        return self._SIZE_BY_ARRANGEMENT[self.name]

    _ARRANGEMENT_BY_TYPE: dict[str, NeonArrangement] = {
        VectorType(f16, (8,)): NeonArrangement.H,
        VectorType(f32, (4,)): NeonArrangement.S,
        VectorType(f64, (2,)): NeonArrangement.D,
    }

    @staticmethod
    def from_arrangement(vec_type: VectorType):
        arrangement = NeonArrangement._ARRANGEMENT_BY_TYPE.get(vec_type)
        if arrangement is None:
            raise ValueError(f"Invalid vector type for ARM NEON: {vec_type}")
        return arrangement
```

emmau678: I think this is going to cause issues because it now treats _SIZE_BY_ARRANGEMENT as a member of the enum (this was why I previously had it outside the class or within a method). I can't find any other way of storing it that does not lead to errors.

emmau678: In addition, when trying to use NeonArrangement as the dictionary value type at this point in the code, it complains that "NeonArrangement" is not defined, I guess because the class has not been fully initialized?

xdsl/dialects/arm_neon.py
```diff
@@ -94,9 +99,24 @@ class NeonArrangement(StrEnum):
...
-    def map_to_num_els(self):
-        map = {"D": 2, "S": 4, "H": 8}
-        return map[self.name]
+    _SIZE_BY_ARRANGEMENT = {"D": 2, "S": 4, "H": 8}
+
+    @property
+    def arrangement(self):
+        return self._SIZE_BY_ARRANGEMENT[self.name]
+
+    _ARRANGEMENT_BY_TYPE: dict[str, NeonArrangement] = {
```
superlopuh, line 108:
```suggestion
    _ARRANGEMENT_BY_TYPE: dict[VectorType, NeonArrangement] = {
```

xdsl/dialects/arm_neon.py
superlopuh: Yeah that makes sense I didn't realise that enums broke this pattern, it should be fine outside of the class, but below the class definition. It makes sense that NeonArrangement would not be defined if the dictionary is placed above the definition in the file, as the file is executed top to bottom

xdsl/dialects/arm_neon.py
superlopuh: I don't understand this message, what doesn't work in the text above?

xdsl/dialects/arm_neon.py
emmau678: this name is not correct: ```def from_arrangement(vec_type: VectorType):``` as the arrangement is the return value so I propose to change it to from_vec_type (will push this with next commit)

superlopuh: Oh yeah that was a brain fart from me your name is great (although maybe `from_vector_type`?)

emmau678:
> Seems like something went wrong in a merge...

yep I'm trying to fix it (sorry!)

PR #4356, opened 2025-05-09
dialects: (builtin) move check to DenseElement verify

alexarice:
Moves a small check to the verify method instead of the constructor. This will make it easier to simplify the constructor situation

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -2102,6 +2102,14 @@ class DenseIntOrFPElementsAttr(
         # Product of dimensions needs to equal length
         return n == len(self)
 
+    def verify(self) -> None:
+        # zero rank type should only hold 1 value
+        data_len = len(self.get_values())
+        if not self.type.get_shape() and data_len != 1:
+            raise ValueError(
+                f"A zero-rank {self.type.name} can only hold 1 value but {data_len} were given."
+            )
+
     @staticmethod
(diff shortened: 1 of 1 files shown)


xdsl/dialects/builtin.py
```diff
@@ -2102,6 +2102,14 @@ def shape_is_complete(self) -> bool:
         # Product of dimensions needs to equal length
         return n == len(self)
 
+    def verify(self) -> None:
+        # zero rank type should only hold 1 value
+        data_len = len(self.get_values())
+        if not self.type.get_shape() and data_len != 1:
+            raise ValueError(
```
superlopuh, line 2109: `VerifyException`? Test?

alexarice: Already is one

alexarice: Can separate it out if you feel it makes sense

superlopuh: I mean we should probably have a filecheck test for this guy, instead of a test for the init.

alexarice: I'm struggling to trigger the error with filecheck as it hits a parsing error first, are you happy with this as is?

superlopuh: Ah interesting, sure

superlopuh (approved):

PR #4360, opened 2025-05-09
dialects: (builtin) deprecate vector_from_list

alexarice:
Switches over the last couple of uses to `create_dense_int`


xdsl/dialects/builtin.py
```diff
@@ -2227,6 +2227,7 @@ def from_list(
             return DenseIntOrFPElementsAttr.create_dense_int(new_type, new_data)
 
     @staticmethod
+    @deprecated("Please use `create_dense_{int/float}` instead.")
```
superlopuh, line 2230: same here, let's use the suggested replacement?

superlopuh (approved):

PR #4362, opened 2025-05-09
dialects: (toy) make f64 specific and use `create_dense_float`

alexarice:
Replace two uses of `tensor_from_list`. Also makes `toy.constant` specific to `f64` (as it already was but this wasn't checked).


math-fehr (approved):

alexarice:
I have realised that this makes the notebook out of date, would it be worth trying to fix this in the same PR (tbh I have no idea how to open a jupyter notebook)

superlopuh:
Can I push here?

superlopuh:
BTW the easiest way for you to fix this would be to open this PR in a GH CodeSpace and run the notebook there.

superlopuh (approved):

PR #4364, opened 2025-05-09
dialects: (builtin) remove DenseIntOrFPElementsAttr splat initialisation

alexarice:
Remove splat initialisation as it only seems to be used in a couple of tests.

If this has significant use downstream then I'm happy to reinstate this or add something similar to `create_dense_{int/float}`, but I personally think this is slightly confusing behaviour.

### DIFF
--- a/tests/dialects/test_builtin.py
+++ b/tests/dialects/test_builtin.py
@@ -358,12 +358,6 @@ def test_DenseIntOrFPElementsAttr_from_list():
     assert attr.type == AnyTensorType(f32, [2])
     assert len(attr) == 2
 
-    # splat initialization
-    attr = DenseIntOrFPElementsAttr.tensor_from_list([4], f32, [4])
-    assert attr.type == AnyTensorType(f32, [4])
-    assert tuple(attr.get_values()) == (4, 4, 4, 4)
-    assert len(attr) == 4
-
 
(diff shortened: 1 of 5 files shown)


math-fehr (commented):
For that one, should we have a specialized constructor like `from_splat`?

alexarice:
I was thinking we could either have a custom constructor, or allow a scalar to be passed into `create_dense_{int/float}` (as opposed to `Sequence[int | float]`), but I'm not convinced its necessary as the only uses are tests and an experimental transform and it's not hard to do the logic in the couple of places it's needed.

superlopuh:
If I understand correctly, MLIR has a separate splat representation, although I may be wrong. If you're looking into this bit of the code, maybe you could take a look at the MLIR implementation? CC @n-io

alexarice:
I agree that having proper constructor support would make sense if we actually represented these things as a single element

n-io:
I always thought that MLIR's way of keeping values as splat seemed preferable in the constructor, in the memory overhead involved, and in later establishing whether all values are actually the same and could be lowered as a single (splat) constant.

n-io:
I'm not against simplifying the logic, I'm just cautioning that there may be a complexity consideration involved.

alexarice:
In the meeting I think we came to the consensus that it would be good to keep this around, so I will close this PR

PR #4366, opened 2025-05-09
transformations: (convert-vector-to-ptr) Add support for vector loads and dynamic offset memrefs

kaylendog:
- Adds a new rewrite for `StoreOp`
- Implements support for dynamic offset memrefs.


superlopuh (commented):
On one hand, I think the code is fine, on the other I think this might be a nice opportunity to update `offset_calculations` in `convert_memref_to_ptr`, since that is also likely to be used at some point with a MemRef type with a dynamic index. How do you feel about another PR that uses the calculations added here to that function, and then rebasing this PR on top of that one? I'm a little hesitant to suggest this because of the dissertation time pressure, it would be fine to file an issue to do this later also.

superlopuh:
I didn't realise that #4384 did a part of this lowering! This PR can now be focused on just adding dynamic offset support, when you have the time to update it.

PR #4370, opened 2025-05-09
misc: remove APIs deprecated more than three months ago

superlopuh:


alexarice (approved):

xdsl/utils/deprecation.py
```diff
@@ -1,30 +0,0 @@
-import warnings
```
math-fehr, line 1: Wait, why removing this?
We should keep it for the next thing we deprecate, rather than having to find it again in the commits and adding it back.

math-fehr (commented):
Good for removing the deprecated stuff, but I wouldn't remove `deprecated` itself though?

xdsl/utils/deprecation.py
superlopuh: We've already migrated to using `deprecated` from `typing_extension`

math-fehr (approved):

xdsl/utils/deprecation.py
superlopuh: In fact it's already unused, all the ones I removed were not calling this function, and there are some remaining

alexarice: `deprecated` has been deprecated

PR #4373, opened 2025-05-10
dialects: (vector) Add `vector.bitcast` operation.

kaylendog:
Adds `BitCastOp` and tests.


tests/filecheck/mlir-conversion/with-mlir/dialects/vector/bitcast.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: xdsl-opt --print-op-generic %s | mlir-opt --mlir-print-op-generic --allow-unregistered-dialect | xdsl-opt | filecheck %s
+// RUN: xdsl-opt %s | mlir-opt --mlir-print-op-generic --allow-unregistered-dialect | xdsl-opt | filecheck %s
```
superlopuh, line 2:
```suggestion
// RUN: xdsl-opt %s | mlir-opt --allow-unregistered-dialect | xdsl-opt | filecheck %s
```

superlopuh (commented):
Looks great! Could you please add a check for bitwidths?

We probably don't want exactly the same error message as MLIR, but here's what it gives me:
```
<stdin>:3:11: error: 'vector.bitcast' op source/result bitwidth of the minor 1-D vectors must be equal
  %cast = vector.bitcast %vf : vector<2xf32> to vector<2xi64>
```

superlopuh:
Closing in favour of #5366

PR #4375, opened 2025-05-11
dialects: (smt) Add `smt.eq` and `smt.distinct`

math-fehr:


xdsl/dialects/smt.py
```diff
@@ -103,13 +113,134 @@ class XOrOp(VariadicBoolOp):
...
+    def parse(cls, parser: Parser) -> DistinctOp:
+        operands, attr_dict = _parse_same_operand_type_variadic_to_bool_op(parser)
+        op = DistinctOp(*operands)
+        op.attributes = attr_dict
+        return op
+
+    def print(self, printer: Printer):
+        _print_same_operand_type_variadic_to_bool_op(
+            printer, self.inputs, self.attributes
+        )
```
superlopuh, line 190: I'm surprised we can't handle this with the assembly format, what breaks there?

alexarice: I think we might have no way to access the type of a single operand. `type($inputs)` will print a comma separated list of all the inputs.

Kind of related but I wondered at some point whether it made sense to have an assembly format directive that directly mentioned a variable constraint. For example, here you could write `$T`.

xdsl/dialects/smt.py
```diff
@@ -103,13 +113,134 @@ class XOrOp(VariadicBoolOp):
...
+
+    def __init__(self, *operands: SSAValue):
+        super().__init__(operands=[operands], result_types=[BoolType()])
+
+
+@irdl_op_definition
+class EqOp(IRDLOperation):
```
alexarice, line 197: Would it be worth having a base class for these? They only differ by operation name and summary right?

alexarice (approved):
Needing the custom parsing/printing here is a shame. Does the MLIR version also use a custom parse/print method?

xdsl/dialects/smt.py
math-fehr: Exactly what Alex said :+1: 
I would be really happy if we could just write `$T`. We could do something like having an explicit marker for type variables, like `%T` for instance? So it's obvious what is compatible with MLIR or not

alexarice: I might cook up a prototype, had a look today and it should be fairly easy

math-fehr: Would be really cool!

xdsl/dialects/smt.py
math-fehr: Merged them into one

xdsl/dialects/smt.py
```diff
@@ -103,13 +113,134 @@ class XOrOp(VariadicBoolOp):
...
+
+    ```
+    and (distinct a b) (distinct a c) (distinct a d)
+        (distinct b c) (distinct b d) (distinct c d)
+    ```
+    """
+
+    name = "smt.distinct"
+
+    T: ClassVar = VarConstraint("T", irdl_to_attr_constraint(NonFuncSMTType))
```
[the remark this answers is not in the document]
math-fehr: Done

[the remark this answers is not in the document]
alexarice: Will `NonFuncSMTType` eventually be a union type?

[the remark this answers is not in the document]
math-fehr: Yes, but I can switch it then right?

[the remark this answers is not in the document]
alexarice: Sure, I think it would work anyway though

superlopuh (approved):

xdsl/dialects/smt.py
alexarice: As an update, parsing is easy, but printing is not atm as we don't store the values of the constraint variables anywhere (and don't store the extractors to get them either)

PR #4376, opened 2025-05-11
backend: ptr.store to x86

qaco:
Convert the ```ptr.store``` ops to x86 assembly.


tests/filecheck/backend/x86/convert_ptr_to_x86.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt -p convert-ptr-to-x86{arch=avx2} --verify-diagnostics --split-input-file  %s | filecheck %s
+// RUN: xdsl-opt -p convert-ptr-to-x86{arch=avx512} --verify-diagnostics --split-input-file  %s | filecheck %s
```
superlopuh, line 1: shouldn't we have two test files, one for each arch?

superlopuh (approved):

PR #4378, opened 2025-05-12
[WIP] analysis: (liveness) Port of the liveness analysis in MLIR 20.1.1

gabrielrodcanal:
**NOTE THIS IS AN UPDATED PR FOR THE LATEST VERSION OF XDSL CORRESPONDING TO MLIR 20.1.1. LINK TO THE ORIGINAL PR: https://github.com/xdslproject/xdsl/pull/3365. A NEW TEST FOR A FIX OF THE LIVENESS OF ARGUMENTS HAS BEEN ADDED.**

Port of the liveness analysis in MLIR: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Analysis/Liveness.cpp

It passes the tests provided in https://github.com/llvm/llvm-project/blob/main/mlir/test/Analysis/test-liveness.mlir

There are a few functions that had to be added in core xDSL to support the liveness analysis. Pre order walking was one of them that might need testing. Please, let me know if I should open a separate PR with these changes and corresponding tests before accepting this PR.

Finally, I placed the analysis under transforms for now, but I don't think it's the right place for it. Suggestions?

The pass needs some helpers in core. Whilst they are contained in this PR at the moment, I wiill be opening separate PRs for them that should be approved before this one.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1040,6 +1040,17 @@ class Operation(IRNode):
             return self.attributes[name]
         return None
 
+    def is_before_in_block(self, other_op: Operation) -> bool:
+        parent_block = self.parent_block()
+        assert isinstance(parent_block, Block)
+
+        if parent_block.get_operation_index(self) < parent_block.get_operation_index(
+            other_op
+        ):
+            return True
+        else:
+            return False
+
     def verify(self, verify_nested_ops: bool = True) -> None:
         for operand in self.operands:
             if isinstance(operand, ErasedSSAValue):
@@ -1787,6 +1798,13 @@ class Block(IRNode, IRWithUses):
         for op in self.ops:
             op.erase(safe_erase=safe_erase, drop_references=False)
 
+    def find_ancestor_op_in_block(self, other_op: Operation) -> Operation | None:
+        for op in self.ops:
+            if op.is_ancestor(other_op):
+                return op
+
+        return None
+
     def is_structurally_equivalent(
         self,
         other: IRNode,
@@ -2010,6 +2028,17 @@ class Region(IRNode):
             if self.parent is not None and self.parent.parent is not None
             else None
         )
+
+    def find_ancestor_block_in_region(self, block: Block) -> Block | None:
+        curr_block = block
+        while curr_block.parent_region() != self:
+            parent_op = curr_block.parent_op()
+            if not parent_op or not parent_op.parent_block():
+                return None
+            curr_block = parent_op.parent_block()
+            assert isinstance(curr_block, Block)
+
+        return curr_block
 
     @property
     def blocks(self) -> RegionBlocks:
--- a/xdsl/transforms/experimental/liveness.py
+++ b/xdsl/transforms/experimental/liveness.py
@@ -1,0 +1,450 @@
+from dataclasses import dataclass, field
+from typing import IO, cast
+
+from xdsl.dialects import builtin, func
+from xdsl.ir import Block, BlockArgument, Operation, Region, SSAValue
+from xdsl.printer import Printer
+
+
+class BlockInfoBuilder:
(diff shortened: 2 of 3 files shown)


xdsl/ir/core.py
```diff
@@ -1040,6 +1040,17 @@ def get_attr_or_prop(self, name: str) -> Attribute | None:
...
+        parent_block = self.parent_block()
+        assert isinstance(parent_block, Block)
+
+        if parent_block.get_operation_index(self) < parent_block.get_operation_index(
+            other_op
+        ):
+            return True
+        else:
+            return False
+
```
superlopuh, line 1053: Let's first merge this change in a separate PR? In MLIR, operation indices are cached, which makes this a constant-time comparison if they're cached, and O(length of block) otherwise. In xDSL, getting the operation index is always O(length of block). I would recommend an implementation like this:
```suggestion
    def is_before_in_block(self, other_op: Operation) -> bool:
        if (parent_block := self.parent_block()) is None or other_op.parent_block() is not parent_block:
            return False

        op = self
        while op is not None:
            if op is other_op:
                return True
            op = op.next_op
        return False
```

superlopuh: This way the cost is either the distance between the ops if one is before the other, or the length of the tail of the block.

gabrielrodcanal: PR #4385

xdsl/ir/core.py
```diff
@@ -2011,6 +2031,17 @@ def parent_region(self) -> Region | None:
...
+        curr_block = block
+        while curr_block.parent_region() != self:
+            parent_op = curr_block.parent_op()
+            if not parent_op or not parent_op.parent_block():
+                return None
+            curr_block = parent_op.parent_block()
+            assert isinstance(curr_block, Block)
+
+        return curr_block
+
```
gabrielrodcanal, line 2044: PR #4386

superlopuh (commented):
This is a prety huge PR to review, is there anything that could be pulled out into a smaller one, with a unit test instead of a filecheck test?

superlopuh:
It's just that reviewing effort feels quadratic in size, and I'm not sure when I'll have time to properly review this much code, a smaller PR would make things much easier as then I can be sure that I properly understand it.

xdsl/transforms/experimental/liveness.py
```diff
@@ -0,0 +1,462 @@
...
+from xdsl.printer import Printer
+
+
+class BlockInfoBuilder:
+    def __init__(self, block: Block):
+        self.out_values: set[SSAValue] = set()
+        self.def_values: set[SSAValue] = set()
+        self.use_values: set[SSAValue] = set()
+        self.in_values: set[SSAValue] = set()
+        self.block = block
```
superlopuh, line 17: It would be great to add this helper class in a first PR. Theer are a few things missing that would be worth doing, that seem unrelated to the rest of the changes:

1. A doc comment for this file, explaining what the purpose of it is, and what liveness is (linking out is fine)
2. a doc comment for this class, and for each of the fields. Some things are obvious and would only warrant a short doc comment (def_values are arguments of the block), while other things are tricky (can use_values contain values defined in a nested block?)
3. move the `#` comments into doc comments when appropriate
4. add tests for this thing specifically. TBH I still don't understand how it works and why things can be updated if it's initialised with a block that presumably doesn't change while the analysis is happening?

superlopuh: One more thing to consider is that we have a lightweight version of liveness that we use for register allocation: https://github.com/xdslproject/xdsl/blob/main/xdsl/backend/register_allocator.py, it would be great to combine these. Not sure if BlockInfoBuilder is all wee need for it, but it would be nice to consider what methods in this file could be used for that purpose, and whether some could be migrated in the first PR.

PR #4381, opened 2025-05-12
installation: (nix) include mlir-opt in flake

alexarice:
Adds `mlir-opt` v20.1.4 to the nix flake, allowing all the tests to be run without compiling mlir.

@superlopuh as far as I can tell, `brew` also installs v20.1.4 now (https://formulae.brew.sh/formula/llvm), should we update versions for next release?

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


superlopuh (approved):
Nice, definitely let's update

PR #4385, opened 2025-05-13
core: Check if an operation is before another in a block

gabrielrodcanal:
Simple check that tests whether an operation is before another in a block. Port of https://mlir.llvm.org/doxygen/IR_2Operation_8cpp_source.html#l00386


tests/test_is_before_in_block.py
```diff
@@ -0,0 +1,26 @@
...
+
+    assert not op2.is_before_in_block(op1)
+
+
+def test_is_before_in_block_same_op():
+    op1 = test.TestOp()
+    op2 = test.TestOp()
+    test.TestOp(regions=[Region(Block([op1, op2]))])
+
+    assert not op1.is_before_in_block(op1)
```
superlopuh, line 26:
```suggestion
def test_same_block():
    op1 = test.TestOp()
    op2 = test.TestOp()
    test.TestOp(regions=[Region(Block([op1, op2]))])

    assert op1.is_before_in_block(op2)
    assert not op2.is_before_in_block(op1)
    assert not op1.is_before_in_block(op1)
    
def test_different_blocks():
    op1 = test.TestOp()
    op2 = test.TestOp()
    test.TestOp(regions=[Region(Block([op1])), Region(Block([op2]))])

    assert not op1.is_before_in_block(op2)
    assert not op2.is_before_in_block(op1)
```

superlopuh: But TBH I'd move to an existing test file instead of creating a new one.

gabrielrodcanal: I agree it is a very simple test to deserve its own file. In which file would you place it, `test_ir`?

superlopuh (approved):
@math-fehr, are you happy with this?

math-fehr (approved):
That looks good to me, I'm just wondering if we should keep this in `core.py` or if we should have this in something like `utils/` or `ir/ordering.py` instead? Just to not add too many operations here.
I'm wondering what @superlopuh or @alexarice  think about this?

alexarice:
I have no thoughts on this

superlopuh:
Which class is this helper defined in in MLIR?

gabrielrodcanal:
> Which class is this helper defined in in MLIR?

It is defined in the Operation class: https://mlir.llvm.org/doxygen/IR_2Operation_8cpp_source.html#l00386 (updated the first post of the PR)

math-fehr:
Okay, maybe it's fine if we keep it there then!

PR #4386, opened 2025-05-13
core: Find the ancestor block of a block within a region

gabrielrodcanal:
Method to find the ancestor block of a block within a region. Returns the block passed as parameter if it lies in the region, or otherwise finds its ancestor that lies in the region.

Ported from: https://mlir.llvm.org/doxygen/IR_2Region_8cpp_source.html#l00154


tests/test_ancestor_block_in_region.py
```diff
@@ -0,0 +1,36 @@
...
+    blk_2 = Block([op_3])
+    op_2 = test.TestOp(regions=[Region(blk_2)])
+
+    # lvl 1
+    blk_1 = Block([op_2])
+    reg_1_1 = Region(blk_1)
+    reg_1_2 = Region()
+    test.TestOp(regions=[reg_1_1, reg_1_2])
+
+    assert reg_1_2.find_ancestor_block_in_region(blk_2) is not blk_1
```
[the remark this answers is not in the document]
gabrielrodcanal: Agreed. File suggestion?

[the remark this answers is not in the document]
superlopuh: BTW, I only saw this question because I thought to click on the resolved conversation. Letting the original commenter resolve their comment is good practice to make sure the suggestion was addressed as intended.

tests/test_ir.py
```diff
@@ -997,3 +997,37 @@ def test_replace_by_if():
...
+    blk_2 = Block([op_3])
+    op_2 = test.TestOp(regions=[Region(blk_2)])
+
+    # lvl 1
+    blk_1 = Block([op_2])
+    reg_1_1 = Region(blk_1)
+    reg_1_2 = Region()
+    test.TestOp(regions=[reg_1_1, reg_1_2])
+
+    assert reg_1_2.find_ancestor_block_in_region(blk_2) is not blk_1
```
superlopuh, line 1033:
```suggestion
    assert reg_1_2.find_ancestor_block_in_region(blk_2) is None
```

xdsl/ir/core.py
```diff
@@ -2011,6 +2011,21 @@ def parent_region(self) -> Region | None:
             else None
         )
 
+    def find_ancestor_block_in_region(self, block: Block) -> Block | None:
+        """
+        Returns 'block' if 'block' lies in this region, or otherwise finds
+        the ancestor of 'block' that lies in this region.
```
superlopuh, line 2017: Can you please add how one might get None instead?

xdsl/ir/core.py
```diff
@@ -2011,6 +2011,23 @@ def parent_region(self) -> Region | None:
...
+
+        Returns None if no ancestor block that lies in this region is found.
+        """
+        curr_block = block
+        while curr_block.parent_region() != self:
+            parent_op = curr_block.parent_op()
+            if not parent_op or not parent_op.parent_block():
+                return None
+            curr_block = parent_op.parent_block()
+            assert isinstance(curr_block, Block)
```
math-fehr, line 2027:
```suggestion
            curr_block = parent_op.parent_block()
            if curr_block is not None:
                return None
```
Should be more concise

math-fehr (approved):
Besides Sasha's comment, looks good to me!

xdsl/ir/core.py
math-fehr: Oh sorry, replace `is not None` with `is None` in my code!

gabrielrodcanal: Also, a check on `parent_op` needs to be kept, otherwise it could be `None` when calling `parent_block()`. I think the code is in its minimal form at the moment to keep Pyright happy. @math-fehr please, let me know if you agree and then I'll merge!

superlopuh: I have a counter proposal that also avoids isinstance:
``` python
        curr_block = block
        while (
            curr_block is not None
            and (parent_region := curr_block.parent_region()) is not None
        ):
            if parent_region is self:
                return curr_block
            curr_block = parent_region.parent_block()
```

math-fehr: I'm fine with your version or mine tbh

gabrielrodcanal: Thank you @superlopuh, that version seems correct too, but @math-fehr seems simpler and passes the tests. Let's go with that one, I think :)

superlopuh: Oh actually this works too, fun puzzle:
``` python
        curr_block = block
        while (parent_region := curr_block.parent_region()) is not None:
            if (
                parent_region is self
                or (curr_block := parent_region.parent_block()) is None
            ):
                return curr_block
```

superlopuh: I just got what Mathieu was going for, final offer:
``` python
        curr_block = block
        while curr_block is not None and curr_block.parent_region() is not self:
            curr_block = curr_block.parent_block()
        return curr_block
```

PR #4387, opened 2025-05-13
dialects: (builtin) add splat generators for DenseElementsAttr

alexarice:
Adds splat initialisers for `create_dense_{int/float}`, as discussed in xDSL meeting.


xdsl/dialects/builtin.py
```diff
@@ -2197,9 +2197,17 @@ def create_dense_index(
...
+        | Sequence[int]
+        | Sequence[IntegerAttr[_IntegerAttrType]],
     ) -> DenseIntOrFPElementsAttr[_IntegerAttrType]:
-        if len(data) and isinstance(data[0], IntegerAttr):
+        if isinstance(data, IntegerAttr):
+            data = data.value.data
+
+        if isinstance(data, int):
+            data = (data,) * prod(type.get_shape())
+        elif len(data) and isinstance(data[0], IntegerAttr):
```
superlopuh, line 2210: This feels weird to me. Could we not just get the bytes directly? Is there a way to avoid the roundtrip through int?

alexarice: It's possible there's a lot of better ways to do this, I wasn't really going for efficiency with this PR. I think by far the better optimisation would be normalising before duplicating the integer. 

By "avoid the roundtrip" do you mean just packing the input integer multiple times?

superlopuh: I'm OK to defer this to a different PR, but here we have direct access to the bytes, so could just replicate those directly instead of first getting integer, and then packing again.
```
>>> b'beef' * 8
b'beefbeefbeefbeefbeefbeefbeefbeef'
```

superlopuh: Sure, do you think you'll have the motivation to try this on top of the stack?

alexarice: I'll do it now

alexarice: #4393

xdsl/dialects/builtin.py
```diff
@@ -2229,8 +2237,16 @@ def create_dense_int(
...
         type: RankedStructure[_FloatAttrType],
-        data: Sequence[float] | Sequence[FloatAttr[_FloatAttrType]],
+        data: float
+        | FloatAttr[_FloatAttrType]
+        | Sequence[float]
+        | Sequence[FloatAttr[_FloatAttrType]],
     ) -> DenseIntOrFPElementsAttr[_FloatAttrType]:
+        if isinstance(data, FloatAttr):
+            data = data.value.data
+        assert not isinstance(data, int)  # WHY???
```
math-fehr, line 2247: I'm curious, do you have an idea why this is needed?
Otherwise, we should remove the comment and add a better one, or remove it

math-fehr (approved):
That's good to me

xdsl/dialects/builtin.py
alexarice: I forgot I added the comment, but I really don't know why. Somehow after ruling out `FloatAttr`, pyright decides that `data` could be an `int`. I feel this has come up before and is some weird intersection of the rules for floats including `int`. I'm also pretty sure that this code doesn't work without this line.

alexarice: Ok this is my bad, turns out python lets you pass an int to this function, so pyright was correct

math-fehr: Oh yes, isn't something like int being a subset of float?

alexarice: For function calls but not for `isinstance` checks or something?

superlopuh (approved):

n-io (approved):

PR #4389, opened 2025-05-13
dialects: (builtin) deprecate from_list

alexarice:
Final deprecation

### DIFF
--- a/docs/Toy/Toy_Ch3.ipynb
+++ b/docs/Toy/Toy_Ch3.ipynb
@@ -242,7 +242,7 @@
     "\n",
     "        assert isa(op.res.type, toy.TensorTypeF64)\n",
     "\n",
-    "        new_value = DenseIntOrFPElementsAttr.from_list(\n",
+    "        new_value = DenseIntOrFPElementsAttr.create_dense_float(\n",
     "            type=op.res.type, data=reshape_input_op.value.get_values()\n",
     "        )\n",
     "        new_op = toy.ConstantOp(new_value)\n",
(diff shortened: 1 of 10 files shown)


docs/Toy/Toy_Ch3.ipynb
```diff
@@ -242,7 +242,7 @@
     "\n",
     "        assert isa(op.res.type, toy.TensorTypeF64)\n",
     "\n",
-    "        new_value = DenseIntOrFPElementsAttr.from_list(\n",
+    "        new_value = DenseIntOrFPElementsAttr.create_dense_float(\n",
```
alexarice, line 245: Might need to check that the notebooks still make sense

xdsl/transforms/printf_to_llvm.py
```diff
@@ -103,7 +103,7 @@ def _construct_global(self, val: str):
             _key_from_str(val),
             constant=True,
             linkage="internal",
-            value=builtin.DenseIntOrFPElementsAttr.from_list(t_type, data),
+            value=builtin.DenseIntOrFPElementsAttr([t_type, builtin.BytesAttr(data)]),
```
alexarice, line 106: @AntonLydike could you check that this is reasonable or suggest a fix if it is not

superlopuh (approved):

xdsl/transforms/printf_to_llvm.py
AntonLydike: If tests pass then it's good enough for me!

math-fehr (approved):

n-io (approved):

PR #4391, opened 2025-05-13
core: Find the ancestor operation of an operation in a block

gabrielrodcanal:
Efficient method to find the ancestor operation of an operation in a block. Ported from https://mlir.llvm.org/doxygen/Block_8cpp_source.html#l00076


xdsl/ir/core.py
```diff
@@ -1774,6 +1774,14 @@ def drop_all_references(self) -> None:
         for op in self.ops:
             op.drop_all_references()
 
+    def find_ancestor_op_in_block(self, op: Operation) -> Operation | None:
```
superlopuh, line 1777: Can you please add a docstring to describe the behaviour?

tests/test_find_ancestor_op_in_block.py
```diff
@@ -0,0 +1,24 @@
+from xdsl.dialects import test
+from xdsl.irdl import Block, Region
+
+
+def test_find_ancestor_op_in_block_has_ancestor():
+    op1 = test.TestOp()
+    op2 = test.TestOp(regions=[Region(Block([op1]))])
+    # test.TestOp(regions=[Region(Block([op2]))])
```
superlopuh, line 8: commented code

xdsl/ir/core.py
```diff
@@ -1774,6 +1774,18 @@ def drop_all_references(self) -> None:
...
+        """
+        Traverse up the operation hierarchy starting from the owner of operand to
+        find the ancestor operation that resides in the block.
```
superlopuh, line 1780: Can you also add that it returns None if one is not found? Also what operand?

tests/test_ir.py
```diff
@@ -997,3 +997,23 @@ def test_replace_by_if():
...
+
+
+def test_find_ancestor_op_in_block_has_no_ancestor():
+    op1 = test.TestOp()
+    op2 = test.TestOp(regions=[Region(Block([op1]))])
+
+    blk_top = Block([op2])
+    op3 = test.TestOp(regions=[Region(blk_top)])
+
+    assert blk_top.find_ancestor_op_in_block(op3) is None
```
superlopuh, line 1019: I just realized these can be folded. Fewer functions is better for this sort of thing as each one has a bit of overhead when running the tests, which adds up over time.
```suggestion
def test_find_ancestor_op_in_block():
    op1 = test.TestOp()
    op2 = test.TestOp(regions=[Region(Block([op1]))])

    blk_top = Block([op2])
    op3 = test.TestOp(regions=[Region(blk_top)])

    assert blk_top.find_ancestor_op_in_block(op1) is op2
    assert blk_top.find_ancestor_op_in_block(op3) is None
```

superlopuh (approved):
@math-fehr @alexarice thoughts?

tests/test_ir.py
alexarice: I think it would be nice to see a more nested scenario. If op3 was in a block `b` then my understanding is that `b.find_ancestor_op_in_block(op1)` should return `op3` and not `op2`

alexarice (approved):
I left one comment but this looks fine to me (though I'm not sure I understand why this function is useful, for which I would probably need to read the other PRs).

tests/test_ir.py
gabrielrodcanal: That is correct. I can add another level of nesting, no problem

math-fehr (approved):

superlopuh (approved):

PR #4393, opened 2025-05-13
dialects: (builtin) optimise splat value initialisation

alexarice:
Optimises the splat value initialisation form #4387 so that:
- int normalisation only happens once
- packing to bytes only happens once

I also took the chance to make a function which returns the normalised integer or throws an error, which could also be used in `DenseArrayAttr`. Will happily rename this to a better name than `get_normalized_value` if someone can think of one.


xdsl/dialects/builtin.py
```diff
@@ -588,6 +588,20 @@ def normalized_value(
 
         return value
 
+    def get_normalized_value(self, value: int) -> int:
```
superlopuh, line 591: Probably worth either making private or adding tests.

superlopuh (approved):

xdsl/dialects/builtin.py
alexarice: it wouldn't work as private, I'll add tests

superlopuh (commented):
Same for float?

PR #4395, opened 2025-05-14
dialects: (emitc) add dialect

Jimmy2027:
wip for adding the emitc dialect

### DIFF
--- a/tests/filecheck/dialects/emitc/emitc_array.mlir
+++ b/tests/filecheck/dialects/emitc/emitc_array.mlir
@@ -1,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+"builtin.module"() ({
+  "emitc.global"() <{initial_value = dense<[[2, 1], [0, 0]]> : tensor<2x2xi32>, sym_name = "input_data", type = !emitc.array<2x2xi32>}> : () -> ()
+  "func.func"() <{function_type = () -> i32, sym_name = "main", sym_visibility = "public"}> ({
+    %0 = "emitc.constant"() <{value = 0 : i32}> : () -> i32
+    %1 = "emitc.get_global"() <{name = @input_data}> : () -> !emitc.array<2x2xi32>
+    "func.return"(%0) : (i32) -> ()
+  }) : () -> ()
+}) : () -> ()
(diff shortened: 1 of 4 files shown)


tests/filecheck/dialects/emitc/emitc_array.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: XDSL_ROUNDTRIP
+
+"builtin.module"() ({
```
superlopuh, line 3: no need for outer module

tests/filecheck/dialects/emitc/emitc_array.mlir
```diff
@@ -0,0 +1,10 @@
...
+  "emitc.global"() <{initial_value = dense<[[2, 1], [0, 0]]> : tensor<2x2xi32>, sym_name = "input_data", type = !emitc.array<2x2xi32>}> : () -> ()
+  "func.func"() <{function_type = () -> i32, sym_name = "main", sym_visibility = "public"}> ({
+    %0 = "emitc.constant"() <{value = 0 : i32}> : () -> i32
+    %1 = "emitc.get_global"() <{name = @input_data}> : () -> !emitc.array<2x2xi32>
```
superlopuh, line 7: let's add custom syntax for ops directly?

superlopuh (commented):
Nice! I would definitely start by adding one op or one attribute, and then adding more by similar groups. I'd also recommend doing things "vertically", as in the full functionality for each operation if possible, like custom syntax, roundtripping with mlir-opt, and adding helpful doc comments. It slows things down when adding things, but then leaves things in a state where we hopefully don't have to keep going back to fix things up and polish.

Jimmy2027:
yes sure, that makes sense!

xdsl/dialects/emitc.py
```diff
@@ -0,0 +1,423 @@
...
+
+    name = "emitc.array"
+
+    shape: ParameterDef[ArrayAttr[IntAttr]]
+    element_type: ParameterDef[AttributeCovT]
+
+    def __init__(
+        self,
+        element_type: AttributeCovT,
+        shape: Iterable[int | IntAttr],
```
superlopuh, line 51: Let's have these in the same order as the declaration?

Jimmy2027:
closing to iterate over smaller PRs

PR #4398, opened 2025-05-14
transformations: (convert-vector-to-ptr) share helper for offset calculations

superlopuh:
This will ensure consistency and hopefully correctness of lowering.


math-fehr (approved):

qaco:
One day too late : cool !

PR #4399, opened 2025-05-14
dialects: (builtin) make DenseArrayBase generic

alexarice:
It's time for this one.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -637,6 +637,7 @@ I16 = Annotated[IntegerType, i16]
 I8 = Annotated[IntegerType, i8]
 I1 = Annotated[IntegerType, i1]
 
+_IntegerTypeInvT = TypeVar("_IntegerTypeInvT", bound=IntegerType, default=IntegerType)
 
 SignlessIntegerConstraint = ParamAttrConstraint(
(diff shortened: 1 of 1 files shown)


alexarice:
Does make tests only run pyright on the files you changed?

xdsl/dialects/builtin.py
```diff
@@ -1388,11 +1388,25 @@ def from_params(handle: str | StringAttr, type: Attribute) -> DenseResourceAttr:
         return DenseResourceAttr([handle, type])
 
 
+DenseArrayT = TypeVar(
+    "DenseArrayT",
+    bound=IntegerType | AnyFloat,
+    default=IntegerType | AnyFloat,
+    covariant=True,
+)
+_IntegerTypeInvT = TypeVar("_IntegerTypeInvT", bound=IntegerType, default=IntegerType)
```
superlopuh, line 1397: is this the best place for this?

superlopuh (approved):

xdsl/dialects/builtin.py
alexarice: I'm not sure tbh, we don't really seem to have a naming scheme or place for type vars but I'll move it to just after `IntegerType`

PR #4401, opened 2025-05-14
dialects: (emitc) Add EmitC dialect with the EmitC_ArrayType

Jimmy2027:
Register the emitc dialect in xdsl/dialects/__init__.py and add the EmitC Array Type as a first step towards adding the full [emitc dialect](https://mlir.llvm.org/docs/Dialects/EmitC/)

### DIFF
--- a/tests/filecheck/dialects/emitc/emitc_types.mlir
+++ b/tests/filecheck/dialects/emitc/emitc_types.mlir
@@ -1,0 +1,12 @@
+// RUN: XDSL_ROUNDTRIP
+
+// CHECK: f32_2D = !emitc.array<4x2xf32>
+// CHECK-SAME: i32_1D = !emitc.array<10xi32>
+// CHECK-SAME: f64_1D = !emitc.array<5xf64>
+// CHECK-SAME: i1_3D = !emitc.array<3x4x5xi1>
+"test.op"() {
+  f32_2D = !emitc.array<4x2xf32>,
+  i32_1D = !emitc.array<10xi32>,
+  f64_1D = !emitc.array<5xf64>,
+  i1_3D = !emitc.array<3x4x5xi1>
+}: ()->()
--- a/xdsl/dialects/emitc.py
+++ b/xdsl/dialects/emitc.py
@@ -1,0 +1,90 @@
+"""
+Dialect to generate C/C++ from MLIR.
+
+The EmitC dialect allows to convert operations from other MLIR dialects to EmitC ops.
+Those can be translated to C/C++ via the Cpp emitter.
+
+See external [documentation](https://mlir.llvm.org/docs/Dialects/EmitC/).
+"""
+
+from collections.abc import Iterable
+
+from xdsl.dialects.builtin import (
+    ArrayAttr,
+    ContainerType,
+    IntAttr,
+    ShapedType,
+)
+from xdsl.ir import (
+    AttributeCovT,
+    Dialect,
+    ParametrizedAttribute,
+    TypeAttribute,
+)
+from xdsl.irdl import (
+    ParameterDef,
+    irdl_attr_definition,
+)
+from xdsl.parser import AttrParser
+from xdsl.printer import Printer
+from xdsl.utils.exceptions import VerifyException
+
+
+@irdl_attr_definition
+class EmitC_ArrayType(
+    ParametrizedAttribute, TypeAttribute, ShapedType, ContainerType[AttributeCovT]
+):
+    """EmitC array type"""
+
+    name = "emitc.array"
+
+    shape: ParameterDef[ArrayAttr[IntAttr]]
+    element_type: ParameterDef[AttributeCovT]
+
+    def __init__(
+        self,
+        shape: Iterable[int | IntAttr],
+        element_type: AttributeCovT,
+    ):
+        shape = ArrayAttr(
+            [IntAttr(dim) if isinstance(dim, int) else dim for dim in shape]
+        )
+        super().__init__([shape, element_type])
+
+    def verify(self) -> None:
+        if not self.shape.data:
+            raise VerifyException("EmitC array shape must not be empty")
+
+    def get_num_dims(self) -> int:
+        return len(self.shape.data)
+
+    def get_shape(self) -> tuple[int, ...]:
+        return tuple(i.data for i in self.shape.data)
+
+    def get_element_type(self) -> AttributeCovT:
+        return self.element_type
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser):
+        with parser.in_angle_brackets():
+            shape, type = parser.parse_ranked_shape()
+            if not shape:
+                raise parser.raise_error("EmitC array shape must not be empty")
+            return ArrayAttr(IntAttr(dim) for dim in shape), type
+
+    def print_parameters(self, printer: Printer) -> None:
+        with printer.in_angle_brackets():
+            printer.print_list(
+                self.shape, lambda dim: printer.print_string(f"{dim.data}"), "x"
+            )
+            printer.print_string("x")
+            printer.print_attribute(self.element_type)
+
+
+EmitC = Dialect(
+    "emitc",
+    [],
+    [
+        EmitC_ArrayType,
+    ],
+)
--- a/tests/dialects/test_emitc.py
+++ b/tests/dialects/test_emitc.py
@@ -1,0 +1,10 @@
+import pytest
+
+from xdsl.dialects.builtin import i32
+from xdsl.dialects.emitc import EmitC_ArrayType
+from xdsl.utils.exceptions import VerifyException
+
+
+def test_emitc_array_empty_shape():
+    with pytest.raises(VerifyException, match="EmitC array shape must not be empty"):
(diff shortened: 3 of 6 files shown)


superlopuh (commented):
Can you please add a test to check a roundtrip with mlir, like `tests/filecheck/mlir-conversion/with-mlir/dialects/arith/arith_attrs.mlir`, Also can you please make your test similar to that one (but omitting the implied builtin.module)

xdsl/dialects/emitc.py
superlopuh, line 1: Can you please add a docstring, with a link to the linalg implementation, and a description? You can see what the generated docs site looks like by clicking on the readthedocs CI job.

superlopuh: Sorry, I mistyped, I meant the MLIR dialect definition.

Jimmy2027: ah okay makes more sense :) will add that!

xdsl/dialects/emitc.py
```diff
@@ -0,0 +1,86 @@
...
+        return self.element_type
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
+        parser.parse_punctuation("<", " in array attribute")
+        shape, type = parser.parse_ranked_shape()
+        shape = ArrayAttr([IntAttr(dim) for dim in shape])
+        parser.parse_punctuation(">", " in array attribute")
+
+        return [shape, type]
```
superlopuh, line 70:
```suggestion
    def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
        with parser.in_angle_brackets():
            shape, type = parser.parse_ranked_shape()
            return ArrayAttr(IntAttr(dim) for dim in shape), type
```

xdsl/dialects/emitc.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def print_parameters(self, printer: Printer) -> None:
+        printer.print(
+            "<",
+            "x".join([*[str(e.data) for e in self.shape], *[str(self.element_type)]]),
+        )
+        printer.print(">")
```
superlopuh, line 77:
```suggestion
        with printer.in_angle_brackets():
            printer.print_list(
                self.shape,
                lambda dim: printer.print_string(f"{dim.data}"),
                "x"
            )
            if self.shape:
                printer.print_string("x")
            printer.print_attribute(self.element_type)
```

tests/filecheck/dialects/emitc/emitc_types.mlir
```diff
@@ -0,0 +1,12 @@
...
+// Test the emitc array type
+// CHECK: func.func public @identity_array(%input : !emitc.array<2x2xf32>) -> !emitc.array<2x2xf32> {
+// CHECK-NEXT:   func.return %input : !emitc.array<2x2xf32>
+// CHECK-NEXT: }
+func.func public @identity_array(
+  %input: !emitc.array<2x2xf32>
+) -> !emitc.array<2x2xf32> {
+  
+  func.return %input : !emitc.array<2x2xf32>
+}
```
superlopuh, line 12: Why not just a "test.op" with an attribute in a dictionary?

superlopuh: Also, can you please add an MLIR roundtrip test, also with a test op

Jimmy2027: sorry messed up my update this morning, should be there now

tests/filecheck/dialects/emitc/emitc_types.mlir
```diff
@@ -0,0 +1,17 @@
...
+"test.op"() {"array_attr" = !emitc.array<10xi32>}: ()->()
+// CHECK: array_attr = !emitc.array<10xi32>
+
+// Test EmitC_ArrayType with a different 1D shape and f64 element type
+"test.op"() {"array_attr" = !emitc.array<5xf64>}: ()->()
+// CHECK: array_attr = !emitc.array<5xf64>
+
+// Test EmitC_ArrayType with a 3D shape and i1 element type
+"test.op"() {"array_attr" = !emitc.array<3x4x5xi1>}: ()->()
+// CHECK: array_attr = !emitc.array<3x4x5xi1>
```
superlopuh, line 17: This is fine but I'd recommend something a little less verbose since we'll be adding more things:
```suggestion

// CHECK: f32_2D = !emitc.array<4x2xf32>
// CHECK-SAME: i32_1D = !emitc.array<10xi32>
// CHECK-SAME: f64_1D = !emitc.array<5xf64>
// CHECK-SAME: i1_3D = !emitc.array<3x4x5xi1>
"test.op"() {
  f32_2D = !emitc.array<4x2xf32>,
  i32_1D = !emitc.array<10xi32>,
  f64_1D = !emitc.array<5xf64>,
  i1_3D = !emitc.array<3x4x5xi1>,
}: ()->()
```
Also, MLIR puts the checks before, which I've tried to do in new test files.

Jimmy2027: right, thanks!

Jimmy2027:
trailing comma was the issue :)

xdsl/dialects/emitc.py
```diff
@@ -0,0 +1,84 @@
...
+        with parser.in_angle_brackets():
+            shape, type = parser.parse_ranked_shape()
+            return ArrayAttr(IntAttr(dim) for dim in shape), type
+
+    def print_parameters(self, printer: Printer) -> None:
+        with printer.in_angle_brackets():
+            printer.print_list(
+                self.shape, lambda dim: printer.print_string(f"{dim.data}"), "x"
+            )
+            if self.shape:
```
superlopuh, line 73: this is more of a nit, but it would be nice to have a test for an empty shape

superlopuh (approved):
Thank you!

xdsl/dialects/emitc.py
Jimmy2027: looks like this is a good point, an empty shape is not supported. I'm not sure if I can check for an error with lit, would it be better to add a pytest for this instead?

tests/filecheck/dialects/emitc/emitc_types.mlir
```diff
@@ -0,0 +1,14 @@
+// RUN: XDSL_ROUNDTRIP
+
+// CHECK: f32_2D = !emitc.array<4x2xf32>
+// CHECK-SAME: i32_1D = !emitc.array<10xi32>
+// CHECK-SAME: f64_1D = !emitc.array<5xf64>
+// CHECK-SAME: i1_3D = !emitc.array<3x4x5xi1>
+// CHECK-SAME: EmitC_ArrayType shape must not be empty
```
superlopuh, line 7: I would recommend adding a separate test for this, something like `tests/filecheck/dialects/emitc/emitc_types_invalid.mlir`, and passing `--verify-diagnostics`, you can do a search in the project for similar tests.

xdsl/dialects/emitc.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(
+        self,
+        shape: Iterable[int | IntAttr],
+        element_type: AttributeCovT,
+    ):
+        shape = ArrayAttr(
+            [IntAttr(dim) if isinstance(dim, int) else dim for dim in shape]
+        )
+        if len(shape.data) == 0:
+            raise ValueError("EmitC_ArrayType shape must not be empty")
```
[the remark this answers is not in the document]
superlopuh: Actually I'm curious, what does `mlir-opt` raise as an error in this case?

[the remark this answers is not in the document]
Jimmy2027: this :)
https://github.com/xdslproject/xdsl/actions/runs/15052998304/job/42312103517#step:9:279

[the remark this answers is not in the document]
superlopuh: Great, I would say let's use the mlir attribute name instead of the Python class name then we're GTG

xdsl/dialects/emitc.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def get_element_type(self) -> AttributeCovT:
+        return self.element_type
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser):
+        with parser.in_angle_brackets():
+            shape, type = parser.parse_ranked_shape()
+            if not shape:
+                raise parser.raise_error("EmitC_ArrayType shape must not be empty")
```
superlopuh, line 69: Oh we should have a separate parsing test also from verification, that is done with the `--parser-diagnostics` flag to `xdsl-op`, and we should add a `pytest` test for the verify exception

tests/filecheck/dialects/emitc/emitc_types.mlir
superlopuh: I think we need to yoink this line now and then we'll be gtg

xdsl/dialects/emitc.py
Jimmy2027: the empty shape verifying is only tested in a pytest now, don't think it is possible from a lit test since that fails already at parsing

superlopuh:
Thank you!

PR #4403, opened 2025-05-15
transformations: (convert-vector-to-ptr) cast first then ptr arithmetic

superlopuh:
Also fixes the name so that it's the same as the original memref, as opposed to either `%0` in the case of store, or the name of the result value in the case of load.


qaco:
Nice

qaco (approved):

PR #4404, opened 2025-05-15
dialects: (builtin) verify strided memref layout

### LINKED ISSUE #4400: transforms: (convert-memref-to-ptr) No error for stores/loads with dynamic strides
### Description

`convert-memref-to-ptr` does not error when handling dynamic strides, despite the current implementation not supporting them.

### Expected vs Actual Behaviour

Since the current implementation does not support dynamic strides, I would expect it to throw an error. Instead, it generates memory-unsafe IR.

### Steps to Reproduce and a Minimal Working Example (MWE)

Calling `xdsl-opt -p convert-memref-to-ptr` on the following MLIR code

```
builtin.module {
	func.func @load(%arr : memref<?x?xf32>, %0 : index, %1 : index) -> f32 {
		%res = memref.load %arr[%0, %1] : memref<?x?xf32>
		func.return %res : f32
	}
	func.func @store(%arr : memref<?x?xf32>, %0 : index, %1 : index, %val : f32) {
		memref.store %val, %arr[%0, %1] : memref<?x?xf32>
		func.return
	}
}
```
yields
```
builtin.module {
  func.func @load(%arr : memref<?x?xf32>, %0 : index, %1 : index) -> f32 {
    %pointer_dim_stride = arith.constant -1 : index
    %pointer_dim_offset = arith.muli %0, %pointer_dim_stride : index
    %pointer_dim_stride_1 = arith.addi %pointer_dim_offset, %1 : index
    %bytes_per_element = ptr_xdsl.type_offset f32 : index
    %scaled_pointer_offset = arith.muli %pointer_dim_stride_1, %bytes_per_element : index
    %res = ptr_xdsl.to_ptr %arr : memref<?x?xf32> -> !ptr_xdsl.ptr
    %offset_pointer = ptr_xdsl.ptradd %res, %scaled_pointer_offset : (!ptr_xdsl.ptr, index) -> !ptr_xdsl.ptr
    %res_1 = ptr_xdsl.load %offset_pointer : !ptr_xdsl.ptr -> f32
    func.return %res_1 : f32
  }
  func.func @store(%arr : memref<?x?xf32>, %0 : index, %1 : index, %val : f32) {
    %pointer_dim_stride = arith.constant -1 : index
    %pointer_dim_offset = arith.muli %0, %pointer_dim_stride : index
    %pointer_dim_stride_1 = arith.addi %pointer_dim_offset, %1 : index
    %bytes_per_element = ptr_xdsl.type_offset f32 : index
    %scaled_pointer_offset = arith.muli %pointer_dim_stride_1, %bytes_per_element : index
    %2 = ptr_xdsl.to_ptr %arr : memref<?x?xf32> -> !ptr_xdsl.ptr
    %offset_pointer = ptr_xdsl.ptradd %2, %scaled_pointer_offset : (!ptr_xdsl.ptr, index) -> !ptr_xdsl.ptr
    ptr_xdsl.store %val, %offset_pointer : f32, !ptr_xdsl.ptr
    func.return
  }
}
```

This is dangerous and incorrect code, particularly `%pointer_dim_stride` being equal to `-1`. This produces out-of-bounds memory accesses.

superlopuh:
Resolves #4400

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1568,7 +1568,7 @@ class StridedLayoutAttr(MemRefLayoutAttr, BuiltinAttribute, ParametrizedAttribut
 
     Strides must be positive and the offset must be non-negative. Both the strides and
     the offset may be dynamic, i.e. their value may not be known at compile time. This
-    is expressed as a ? in the assembly syntax and as ShapedType::kDynamic in the code.
+    is expressed as a ? in the assembly syntax and as `NoneAttr` in the code.
     Stride and offset values must satisfy the constraints above at runtime, the behavior
     is undefined otherwise.
     Contrary to MLIR, and we do not restrict offsets and strides to 64-bits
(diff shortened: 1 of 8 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1566,52 +1566,51 @@ class StridedLayoutAttr(MemRefLayoutAttr, BuiltinAttribute, ParametrizedAttribut
     An attribute representing a strided layout of a shaped type.
     See external [documentation](https://mlir.llvm.org/docs/Dialects/Builtin/#stridedlayoutattr).
 
-    Contrary to MLIR, we represent dynamic offsets and strides with
-    `NoneAttr`, and we do not restrict offsets and strides to 64-bits
```
superlopuh, line 1570: @math-fehr, what was the reason behind this?

math-fehr: I personally felt that encoding the dynamic offset by just using a special number is bad practice, because it's really easy to use the wrong one (`-1` versus `INT_MIN` versus `INT_MAX`).
If you want to use MLIR version that's fine to me, but I personally prefer `NoneAttr` here unless there is some benefit to follow MLIR.

math-fehr: I would personally prefer having a check that the values are non-negative then!
Also, having things like `NoneAttr` would technically allow us to write constraints that would check that there are no dynamic values for instance.

superlopuh: We could do that with -1 also but I currently don't have a strong opinion :) I'l update this PR

math-fehr: So why do you want to change it then?
My understanding is that this is equivalent to poison?

superlopuh:
Closing this for now, will add more info on the issue CC @kaylendog

PR #4405, opened 2025-05-15
backend: add BlockAllocator base class

superlopuh:
Moves some logic that is not RISC-V-specific to a new ABC.

### DIFF
--- a/xdsl/backend/register_allocator.py
+++ b/xdsl/backend/register_allocator.py
@@ -1,9 +1,12 @@
+import abc
 from collections.abc import Sequence
 from typing import cast
 
+from ordered_set import OrderedSet
+
 from xdsl.backend.register_queue import RegisterQueue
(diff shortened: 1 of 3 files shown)


xdsl/backend/register_allocator.py
```diff
@@ -107,3 +110,61 @@ def allocate_values_same_reg(self, vals: Sequence[SSAValue]) -> bool:
...
+
+
+def live_ins_per_block(block: Block) -> dict[Block, OrderedSet[SSAValue]]:
+    """
+    Returns a mapping from a block to the set of values used in it but defined outside of
+    it.
+    """
+    res: dict[Block, OrderedSet[SSAValue]] = {}
+    _ = _live_ins_per_block(block, res)
+    return res
```
superlopuh, line 148: these are just copy/pasted existing implementations

emmau678 (approved):

PR #4407, opened 2025-05-15
dialects (arm): Add fmul (vector) instruction

emmau678:
Add Floating-point Multiply (vector) instruction. There is an existing ```fmul``` instruction for mixed vector/scalar operands which was called ```fmulvec``` and now needs to be renamed (```fmulvecscalar```)

### DIFF
--- a/xdsl/dialects/arm_neon.py
+++ b/xdsl/dialects/arm_neon.py
@@ -7,7 +7,6 @@ from xdsl.dialects.arm.assembly import AssemblyInstructionArg, reg, square_brack
 from xdsl.dialects.arm.ops import ARMInstruction, ARMOperation
 from xdsl.dialects.arm.register import ARMRegisterType, IntRegisterType
 from xdsl.dialects.builtin import IntegerAttr, StringAttr, i8
-from xdsl.dialects.stencil import IndexAttr
 from xdsl.ir import (
     Attribute,
     Dialect,
@@ -207,8 +206,8 @@ class DSSFMulOp(ARMInstruction):
     arrangement = prop_def(NeonArrangementAttr)
 
     assembly_format = (
-      "$s1 `,` $s2 (`[` $scalar_idx^ `]`)? $arrangement attr-dict "
-      "`:` functional-type(operands, $d)"
+        "$s1 `,` $s2 (`[` $scalar_idx^ `]`)? $arrangement attr-dict "
+        "`:` functional-type(operands, $d)"
     )
 
     def __init__(
@@ -217,7 +216,7 @@ class DSSFMulOp(ARMInstruction):
         s2: Operation | SSAValue,
         *,
         d: NEONRegisterType,
-        scalar_idx: IndexAttr | None,
+        scalar_idx: IntegerAttr | None,
         arrangement: NeonArrangement | NeonArrangementAttr,
         comment: str | StringAttr | None = None,
     ):


superlopuh (commented):
How do you feel about making the scalar_idx optional instead?

xdsl/dialects/arm_neon.py
```diff
@@ -236,6 +236,59 @@ def assembly_line_args(self):
...
+    See external [documentation](https://developer.arm.com/documentation/100069/0606/SIMD-Vector-Instructions/FMUL--vector-).
+    """
+
+    name = "arm_neon.dss.fmulvec"
+    d = result_def(NEONRegisterType)
+    s1 = operand_def(NEONRegisterType)
+    s2 = operand_def(NEONRegisterType)
+    arrangement = prop_def(NeonArrangementAttr)
+
+    assembly_format = "$s1 `,` $s2 $arrangement attr-dict `:` `(` type($s1) `,` type($s2) `)` `->` type($d)"
```
alexarice, line 255:
```suggestion
    assembly_format = "$s1 `,` $s2 $arrangement attr-dict `:` functional-type(operands, $d)"
```
This should also work, though don't mind what is here now.

alexarice (commented):
I think I agree with @superlopuh

xdsl/dialects/arm_neon.py
```diff
@@ -179,34 +181,41 @@ def assembly_line(self):
...
     Encoding: FMUL <Vd>.<T>, <Vn>.<T>, <Vm>.<idx>.
-    Vd, Vn, Vm specify the regs. The <T> specifier determines element arrangement (size and count).
-    The <idx> specifier determines the index of Vm at which the second source operand (scalar) can be found,
-    preceded by a size specifier.
 
     See external [documentation](https://developer.arm.com/documentation/ddi0602/2024-12/SIMD-FP-Instructions/FMUL--vector---Floating-point-multiply--vector--?lang=en#T_option__4).
     """
 
-    name = "arm_neon.dss.fmulvec"
+    name = "arm_neon.dss.fmul"
```
superlopuh, line 202: nice, the `fmul` name below is no longer necessary

xdsl/dialects/arm_neon.py
```diff
@@ -231,7 +243,13 @@ def assembly_line_args(self):
...
             VectorWithArrangement(self.s1, self.arrangement),
             VectorWithArrangement(
-                self.s2, self.arrangement, index=self.scalar_idx.value.data
+                self.s2,
+                self.arrangement,
+                **(
+                    {"index": self.scalar_idx.value.data}
+                    if self.scalar_idx is not None
+                    else {}
+                ),
```
superlopuh, line 252:
```suggestion
                self.scalar_idx.value.data if self.scalar_idx is not None else None,
```

xdsl/dialects/arm_neon.py
```diff
@@ -7,6 +7,7 @@
 from xdsl.dialects.arm.ops import ARMInstruction, ARMOperation
 from xdsl.dialects.arm.register import ARMRegisterType, IntRegisterType
 from xdsl.dialects.builtin import IntegerAttr, StringAttr, i8
+from xdsl.dialects.stencil import IndexAttr
```
alexarice, line 10: This import looks suspicious

xdsl/dialects/arm_neon.py
```diff
@@ -179,34 +181,40 @@ def assembly_line(self):
...
+    assembly_format = "$s1 `,` $s2 (`[` $scalar_idx^ `]`)? $arrangement attr-dict `:` \
+    functional-type(operands, $d)"
 
     def __init__(
         self,
         s1: Operation | SSAValue,
         s2: Operation | SSAValue,
         *,
         d: NEONRegisterType,
+        scalar_idx: IndexAttr | None,
```
alexarice, line 218: Should this be `IntegerAttr[i8]`?

xdsl/dialects/arm_neon.py
```diff
@@ -179,34 +181,40 @@ def assembly_line(self):
...
     s1 = operand_def(NEONRegisterType)
     s2 = operand_def(NEONRegisterType)
-    scalar_idx = attr_def(IntegerAttr[i8])
-    arrangement = attr_def(NeonArrangementAttr)
+    scalar_idx = opt_prop_def(IntegerAttr[i8])
+    arrangement = prop_def(NeonArrangementAttr)
 
-    assembly_format = "$s1 `,` $s2 `[` $scalar_idx `]` $arrangement attr-dict `:` `(` type($s1) `,` type($s2) `)` `->` type($d)"
+    assembly_format = "$s1 `,` $s2 (`[` $scalar_idx^ `]`)? $arrangement attr-dict `:` \
+    functional-type(operands, $d)"
```
alexarice, line 210:
```suggestion
    assembly_format = (
      "$s1 `,` $s2 (`[` $scalar_idx^ `]`)? $arrangement attr-dict "
      "`:` functional-type(operands, $d)"
    )
```
Does this work? looks a bit cleaner to me if so.

xdsl/dialects/arm_neon.py
```diff
@@ -231,7 +242,9 @@ def assembly_line_args(self):
             VectorWithArrangement(self.d, self.arrangement),
             VectorWithArrangement(self.s1, self.arrangement),
             VectorWithArrangement(
-                self.s2, self.arrangement, index=self.scalar_idx.value.data
+                self.s2,
+                self.arrangement,
+                self.scalar_idx.value.data if self.scalar_idx is not None else None,
```
alexarice:
```suggestion
                self.scalar_idx and self.scalar_idx.value.data
```
Possibly cooking too hard

xdsl/dialects/arm_neon.py
```diff
@@ -179,34 +181,40 @@ def assembly_line(self):
...
     """
-    Floating-point multiply (mixed: first source operand is a vector, second is a scalar. Destination is a vector)
-    This instruction multiplies each of the floating-point values in the first source operand by the
-    second source operand and writes the resulting values to the corresponding lanes of the destination.
+    Floating-point multiply. Different instruction types supported:
+
+    1. Vector:  multiplies corresponding floating-point values in the vectors in the two
+    source NEON registers, and writes the result vector to the destination.
+
+    See external [documentation](https://developer.arm.com/documentation/100069/0606/SIMD-Vector-Instructions/FMUL--vector-).
```
alexarice, line 191: Points to an old version. Is this intended?

alexarice (approved):

xdsl/dialects/arm_neon.py
emmau678: it's the lateset version I can find online. the docs for neon are hard to navigate but as far as I can tell if you try to switch it to the latest version on the menu you just get a asm manual

alexarice: I see

xdsl/dialects/arm_neon.py
emmau678: yeah thanks for catching that, deleting it now

xdsl/dialects/arm_neon.py
emmau678: running into problems with this one (```Argument of type "IntegerAttr[IntegerType] | int | None" cannot be assigned to parameter "index" of type "int | None"```), not sure if I can get around it

alexarice: Ok, let's ignore this suggestion

superlopuh (approved):

PR #4410, opened 2025-05-15
dialects (arm): change attributes to properties uniformly and insert …

emmau678:
Applying some feedback I received on review of my last PR, which should be applied uniformly to update all instructions that had previously been added.

1. change ```arrangement``` to a property instead of an attribute.
2. Add ```scalar_idx``` to ```__init__``` method and add to list of properties.

### DIFF
--- a/xdsl/dialects/arm_neon.py
+++ b/xdsl/dialects/arm_neon.py
@@ -274,7 +274,7 @@ class DSSFmlaVecScalarOp(ARMInstruction):
     d = operand_def(SAME_NEON_REGISTER_TYPE)
     s1 = operand_def(NEONRegisterType)
     s2 = operand_def(NEONRegisterType)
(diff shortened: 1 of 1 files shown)


xdsl/dialects/arm_neon.py
```diff
@@ -275,7 +275,7 @@ class DSSFmlaVecScalarOp(ARMInstruction):
     s1 = operand_def(NEONRegisterType)
     s2 = operand_def(NEONRegisterType)
     scalar_idx = attr_def(IntegerAttr[i8])
```
alexarice, line 277:
```suggestion
    scalar_idx = prop_def(IntegerAttr[i8])
```

alexarice (approved):

superlopuh (approved):

PR #4411, opened 2025-05-15
dialects: Add `smt.exists` and `smt.forall` operations

math-fehr:
This adds the two quantifier operations to the `smt` dialect

### DIFF
--- a/xdsl/dialects/smt.py
+++ b/xdsl/dialects/smt.py
@@ -9,7 +9,6 @@ from typing_extensions import Self
 from xdsl.dialects.builtin import BoolAttr
 from xdsl.ir import (
     Attribute,
-    Block,
     Dialect,
     ParametrizedAttribute,
     Region,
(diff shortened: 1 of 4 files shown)


xdsl/dialects/smt.py
```diff
@@ -218,6 +229,73 @@ class EqOp(VariadicPredicateOp):
     name = "smt.eq"
 
 
+class QuantifierOp(IRDLOperation, ABC):
+    result = result_def(BoolType())
```
alexarice, line 233:
```suggestion
    result = result_def(BoolType)
```
?

xdsl/dialects/smt.py
```diff
@@ -218,6 +229,73 @@ class EqOp(VariadicPredicateOp):
...
+    body = region_def("single_block")
+
+    traits = traits_def(Pure())
+
+    assembly_format = "attr-dict-with-keyword $body"
+
+    def __init__(self) -> None:
```
alexarice, line 240: Would we not usually allow a Region/block to be put in the initialiser?

alexarice (approved):

xdsl/dialects/smt.py
math-fehr: :+1:

xdsl/dialects/smt.py
math-fehr: Yeah good point, I changed it to require the region

xdsl/dialects/smt.py
```diff
@@ -218,6 +228,73 @@ class EqOp(VariadicPredicateOp):
...
+        super().__init__(result_types=[BoolType()], regions=[body])
+
+    def verify_(self) -> None:
+        if not isinstance(yield_op := self.body.block.last_op, YieldOp):
+            raise VerifyException("region expects an `smt.yield` terminator")
+        if tuple(yield_op.operand_types) != (BoolType(),):
+            raise VerifyException(
+                "region yield terminator must have a single boolean operand, "
+                f"got {tuple(str(type) for type in yield_op.operand_types)}"
+            )
```
superlopuh, line 249: I'm surprised we don't have a trait for this but I guess this is also good.

PR #4417, opened 2025-05-16
dialects: Add `smt.declare_fun` and `!smt.func`

math-fehr:
Add the `smt.declare_fun` operation, which is used to declare constants and functions.
Also add the `smt.func` type

### DIFF
--- a/tests/filecheck/dialects/smt/ops.mlir
+++ b/tests/filecheck/dialects/smt/ops.mlir
@@ -1,5 +1,10 @@
 // RUN: XDSL_ROUNDTRIP
 
+%const1 = smt.declare_fun : !smt.bool
+%const_with_name = smt.declare_fun "bool_name": !smt.bool
+%func1 = smt.declare_fun : !smt.func<() !smt.bool>
+%func2 = smt.declare_fun : !smt.func<(!smt.bool) !smt.bool>
+
 %arg1 = smt.constant true
 %arg2 = smt.constant false
 %arg3 = smt.constant false
--- a/tests/filecheck/dialects/smt/ops.mlir
+++ b/tests/filecheck/dialects/smt/ops.mlir
@@ -1,43 +1,36 @@
 // RUN: XDSL_ROUNDTRIP
 
-%const1 = smt.declare_fun : !smt.bool
-%const_with_name = smt.declare_fun "bool_name" : !smt.bool
-%func1 = smt.declare_fun : !smt.func<() !smt.bool>
-%func2 = smt.declare_fun : !smt.func<(!smt.bool) !smt.bool>
-
 // CHECK: %const1 = smt.declare_fun : !smt.bool
 // CHECK-NEXT: %const_with_name = smt.declare_fun "bool_name" : !smt.bool
 // CHECK-NEXT: %func1 = smt.declare_fun : !smt.func<() !smt.bool>
 // CHECK-NEXT: %func2 = smt.declare_fun : !smt.func<(!smt.bool) !smt.bool
 
-%arg1 = smt.constant true
-%arg2 = smt.constant false
-%arg3 = smt.constant false
+%const1 = smt.declare_fun : !smt.bool
+%const_with_name = smt.declare_fun "bool_name" : !smt.bool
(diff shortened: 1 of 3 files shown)


tests/filecheck/dialects/smt/ops.mlir
```diff
@@ -1,31 +1,44 @@
 // RUN: XDSL_ROUNDTRIP
 
+%const1 = smt.declare_fun : !smt.bool
```
alexarice, line 3: I thought we were trying to prefer check statements before code? (Realise this is not followed everywhere)

xdsl/dialects/smt.py
```diff
@@ -46,6 +48,72 @@ class BoolType(ParametrizedAttribute, TypeAttribute):
...
+    ):
+        super().__init__([ArrayAttr[NonFuncSMTType](domain_types), range_type])
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        with parser.in_angle_brackets():
+            domain_types = parser.parse_comma_separated_list(
+                parser.Delimiter.PAREN, parser.parse_type
+            )
+            range_types = parser.parse_type()
```
alexarice, line 74:
```suggestion
            range_type = parser.parse_type()
```

xdsl/dialects/smt.py
```diff
@@ -46,6 +48,72 @@ class BoolType(ParametrizedAttribute, TypeAttribute):
...
+    """
+
+    name = "smt.declare_fun"
+
+    name_prefix = opt_prop_def(StringAttr, prop_name="namePrefix")
+    result = result_def(SMTType)
+
+    assembly_format = "($namePrefix^)? attr-dict `:` type($result)"
+
+    def __init__(self, result_type: SMTType, name_prefix: str | None = None):
```
alexarice, line 110:
```suggestion
    def __init__(self, result_type: SMTType, name_prefix: StringAttr | str | None = None):
```

xdsl/dialects/smt.py
```diff
@@ -308,5 +377,5 @@ def __init__(self, *values: SSAValue):
         ForallOp,
         YieldOp,
     ],
-    [BoolType],
+    [BoolType, FuncType],
```
alexarice, line 380:
```suggestion
    [BoolType, FuncType,],
```
adding this before @superlopuh does

alexarice (approved):

tests/filecheck/dialects/smt/ops.mlir
math-fehr: I changed it in the entire file

xdsl/dialects/smt.py
math-fehr: Done

xdsl/dialects/smt.py
```diff
@@ -46,6 +48,76 @@ class BoolType(ParametrizedAttribute, TypeAttribute):
...
+
+    def __init__(
+        self, result_type: SMTType, name_prefix: StringAttr | str | None = None
+    ):
+        if isinstance(name_prefix, str):
+            name_prefix = StringAttr(name_prefix)
+        properties: dict[str, Attribute] = (
+            {"namePrefix": name_prefix} if name_prefix else {}
+        )
+        super().__init__(result_types=[result_type], properties=properties)
```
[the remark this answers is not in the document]
superlopuh: also you don't need the whole `if` expression thing it should work the same way without it (unless you want that behaviour for an empty string in which case it's worth a comment IMO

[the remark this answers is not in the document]
math-fehr: What do you mean by that?
This is a `StringAttr`, so this should only be `True` if this is `None` (which is why I'm against `if name_prefix` and I prefer using `if name_prefix is not None` btw).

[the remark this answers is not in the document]
superlopuh: I'm confused, if name_prefix is there above? Point is if it's None it'll be handled properly by the init so there's no need to do any special casing.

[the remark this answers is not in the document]
math-fehr: Oh sorry I get it, fixed

superlopuh (approved):

PR #4420, opened 2025-05-16
dialects: Allow range constraints in ArrayOfConstraint

math-fehr:
ArrayOfConstraint can now take a range constraint. This means in particular that we can use a RangeVarConstraint in it, which is quite useful for some operations.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -228,6 +228,21 @@ class ArrayOfConstraint(GenericAttrConstraint[ArrayAttr[AttributeCovT]]):
     def get_unique_base(self) -> type[Attribute] | None:
         return ArrayAttr
 
+    @dataclass(frozen=True)
+    class _Extractor(VarExtractor[Attribute]):
+        inner: VarExtractor[Sequence[Attribute]]
+
+        def extract_var(self, a: Attribute) -> ConstraintVariableType:
+            assert isinstance(a, ArrayAttr)
+            a = cast(ArrayAttr[Attribute], a)
+            return self.inner.extract_var(a.data)
+
+    def get_variable_extractors(self) -> dict[str, VarExtractor[Attribute]]:
+        return {
+            k: self._Extractor(v)
+            for k, v in self.elem_range_constraint.get_variable_extractors().items()
+        }
+
 
 @irdl_attr_definition
 class StringAttr(Data[str], BuiltinAttribute):
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -51,6 +51,7 @@ from xdsl.irdl import (
     ConstraintVariableType,
     GenericAttrConstraint,
     GenericData,
+    GenericRangeConstraint,
     IntConstraint,
     IRDLAttrConstraint,
     IRDLGenericAttrConstraint,
@@ -58,6 +59,7 @@ from xdsl.irdl import (
     MessageConstraint,
     ParamAttrConstraint,
     ParameterDef,
+    RangeOf,
     VarExtractor,
     attr_constr_coercion,
     base,
@@ -137,25 +139,6 @@ class NoneAttr(ParametrizedAttribute, BuiltinAttribute):
     name = "none"
 
 
-@dataclass(frozen=True)
-class ArrayOfConstraint(AttrConstraint):
-    """
-    A constraint that enforces an ArrayData whose elements all satisfy
-    the elem_constr.
-    """
-
-    elem_constr: AttrConstraint
-
-    def __init__(self, constr: Attribute | type[Attribute] | AttrConstraint):
-        object.__setattr__(self, "elem_constr", attr_constr_coercion(constr))
-
-    def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
-        if not isinstance(attr, ArrayAttr):
-            raise VerifyException(f"expected ArrayData attribute, but got {attr}")
-        for e in cast(ArrayAttr[Attribute], attr).data:
-            self.elem_constr.verify(e, constraint_context)
-
-
 @irdl_attr_definition
 class ArrayAttr(
     GenericData[tuple[AttributeCovT, ...]], BuiltinAttribute, Iterable[AttributeCovT]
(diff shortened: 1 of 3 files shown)


math-fehr:
This is stacked so I can continue working on my branches without having a headache, but it doesn't really depend on anything. I'll rebase it to main when I'll merge it depending on which branch gets merged first

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
...
+        return ArrayOfConstraint(RangeOf(irdl_to_attr_constraint(args[0])))
 
     def __len__(self):
         return len(self.data)
 
     def __iter__(self) -> Iterator[AttributeCovT]:
         return iter(self.data)
 
+    @staticmethod
+    def constr(
```
alexarice, line 178: I've found it's sometimes nicer to have invariant typevars in static methods, though to be honest don't understand the whole system fully enough to say if it's the "right" thing to do

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
...
+        ),
+    ) -> GenericAttrConstraint[ArrayAttr[AttributeCovT]]:
+        return ArrayOfConstraint(constr)
+
+
+@dataclass(frozen=True)
+class ArrayOfConstraint(GenericAttrConstraint[ArrayAttr[AttributeCovT]]):
+    elem_range_constraint: GenericRangeConstraint[AttributeCovT]
+    """
+    A constraint that enforces an ArrayData whose elements all satisfy
```
alexarice, line 191:
```suggestion
    A constraint that enforces an ArrayData whose elements satisfy
```
"all" suggests the old behaviour to me

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
...
+    the elem_constr.
```
alexarice, line 192: elem_constr is no longer a parameter that exists

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
...
+    """
+
+    def __init__(
+        self,
+        constr: (
+            IRDLGenericAttrConstraint[Attribute] | GenericRangeConstraint[AttributeCovT]
+        ),
+    ):
+        if isinstance(constr, GenericRangeConstraint):
+            object.__setattr__(self, "elem_range_constraint", constr)
```
alexarice, line 202: why use `object.__setattr__`?

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
...
+            object.__setattr__(
+                self, "elem_range_constraint", RangeOf(irdl_to_attr_constraint(constr))
+            )
+
+    def verify(
+        self,
+        attr: Attribute,
+        constraint_context: ConstraintContext,
+    ) -> None:
+        if not isinstance(attr, ArrayAttr):
```
alexarice, line 213:
```suggestion
        if not isa(attr, ArrayAttr):
```

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
...
+            raise VerifyException(
+                f"expected ArrayAttr attribute, but got '{type(attr)}'"
+            )
+        attr = cast(ArrayAttr[Attribute], attr)
```
alexarice, line 217:
```suggestion
```

superlopuh (approved):

xdsl/dialects/builtin.py
```diff
@@ -183,14 +166,83 @@ def print_parameter(self, printer: Printer) -> None:
     @staticmethod
     def generic_constraint_coercion(args: tuple[Any]) -> AttrConstraint:
         assert len(args) == 1
-        return ArrayOfConstraint(irdl_to_attr_constraint(args[0]))
+        return ArrayOfConstraint(RangeOf(irdl_to_attr_constraint(args[0])))
```
alexarice, line 169: I think I'm getting slightly confused what the generic parameter on `ArrayAttr` means? Why is it a tuple

alexarice (commented):
I'm unsure how this change lines up to the generic parameter on `ArrayAttr` but other than that am happy.

alexarice (approved):
I've figured out the generic parameter thing now

xdsl/dialects/builtin.py
superlopuh: frozen

alexarice: even within the init?

alexarice: Is there not `super().__init__(constr)` or something instead?

superlopuh: not that I know of

math-fehr: `super().__init__(constr)` wouldn't work, because `super()` will refer to the parent class.
This initialization thing for frozen class kinda sucks ngl

xdsl/dialects/builtin.py
math-fehr: Both should work, right? Or do you see a case where this would fail?

xdsl/dialects/builtin.py
math-fehr: Done

xdsl/dialects/builtin.py
math-fehr: Done

xdsl/dialects/builtin.py
math-fehr: Done

xdsl/dialects/builtin.py
math-fehr: Done

xdsl/dialects/builtin.py
alexarice: I feel, at least at some point, that this style would sometimes force you to write `ArrayAttr[PutTypeHere].constr(PutTypeHere)` whereas the invariant version wouldn't

math-fehr: Ohhhhh, but wait, that's different then right?
We still want to use `CovT`, but just another version of it right? Like a `CovT2` so then it's unrelated to the actual generic type?

alexarice: Maybe that's also fine

math-fehr: Wait actually, I think you are right and `InvT` works as well as `CovT`

math-fehr: Yeah, my bad I'm dumb, I don't think `InvT` makes any sense in that context!

alexarice: It likely doesn't make a difference since the whole datatype is covariant anyway

math-fehr: Should be fixed now!

PR #4423, opened 2025-05-19
backend: add allocate_registers abstract method to RegisterAllocatableOperation

superlopuh:
Moves a bunch of per-op logic to the op rather than a shared class! Most importantly this removes the dependency on snitch for the RISC-V register allocator, but it also opens the door for future customisation in the dialect itself vs shared infrastructure.

### DIFF
--- a/xdsl/dialects/riscv_scf.py
+++ b/xdsl/dialects/riscv_scf.py
@@ -5,10 +5,14 @@ RISC-V SCF dialect
 from __future__ import annotations
 
 from abc import ABC
-from collections.abc import Sequence
+from collections.abc import Generator, Sequence
+from typing import cast
 
 from typing_extensions import Self
 
+from xdsl.backend.register_allocatable import RegisterAllocatableOperation
+from xdsl.backend.register_allocator import BlockAllocator
+from xdsl.backend.register_type import RegisterType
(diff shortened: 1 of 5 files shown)


xdsl/dialects/riscv_scf.py
```diff
@@ -118,6 +122,52 @@ def verify_(self):
                         f"variables types."
                     )
 
+    def iter_used_registers(self) -> Generator[RegisterType, None, None]:
+        # We know that all the registers for the inputs and outputs are the same, and
+        # that these registers will have been iterated earlier in the IR.
+        yield from ()
+
+    def allocate_registers(self, allocator: BlockAllocator) -> None:
```
superlopuh, line 130: this function is moved from the other file

xdsl/dialects/riscv_snitch.py
```diff
@@ -432,6 +434,38 @@ def verify_(self) -> None:
                         f"variables types."
                     )
 
+    def allocate_registers(self, allocator: BlockAllocator) -> None:
```
superlopuh, line 437: this function is moved from the other file

xdsl/dialects/riscv_scf.py
qaco: The loop-reservation logic is great !

qaco (approved):

PR #4424, opened 2025-05-20
dialects: (smt) Implement `smt.apply_func`

math-fehr:
`smt.apply_func` is the operation to apply a function to a list of arguments.

### DIFF
--- a/tests/filecheck/dialects/smt/ops.mlir
+++ b/tests/filecheck/dialects/smt/ops.mlir
@@ -10,6 +10,11 @@
 %func1 = smt.declare_fun : !smt.func<() !smt.bool>
 %func2 = smt.declare_fun : !smt.func<(!smt.bool) !smt.bool>
 
+// CHECK-NEXT:    smt.apply_func %func1() : !smt.func<() !smt.bool>
+// CHECK-NEXT:    smt.apply_func %func2(%const1) : !smt.func<(!smt.bool) !smt.bool>
+smt.apply_func %func1() : !smt.func<() !smt.bool>
+smt.apply_func %func2(%const1) : !smt.func<(!smt.bool) !smt.bool>
+
 // CHECK-NEXT:    %arg1 = smt.constant true
 // CHECK-NEXT:    %arg2 = smt.constant false
 // CHECK-NEXT:    %arg3 = smt.constant false
--- a/xdsl/dialects/smt.py
+++ b/xdsl/dialects/smt.py
@@ -17,13 +17,18 @@ from xdsl.ir import (
 )
 from xdsl.irdl import (
     AtLeast,
+    GenericAttrConstraint,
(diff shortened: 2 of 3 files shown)


xdsl/dialects/smt.py
```diff
@@ -117,6 +129,29 @@ def __init__(
...
+    RANGE: ClassVar = VarConstraint("RANGE", base(NonFuncSMTType))
+
+    func = operand_def(FuncType.constr(DOMAIN, RANGE))
+    args = var_operand_def(DOMAIN)
+
+    result = result_def(RANGE)
+
+    assembly_format = "$func `(` $args `)` attr-dict `:` type($func)"
+
+    def __init__(self, func: SSAValue[FuncType], args: Sequence[SSAValue]):
```
math-fehr, line 151: Let me know what you think of this.
I require a `SSAValue[FuncType]` as input, as I need to grab its range type.
I feel that now that we have generic types, this interface makes more sense than requiring an `SSAValue`, and then having an `if isinstance(value.type, FuncType)` with a custom error message.
What do you think @alexarice @superlopuh ?

alexarice: Looks good, this was the point of the typed SSAValues right?

tests/filecheck/dialects/smt/ops.mlir
```diff
@@ -10,6 +10,11 @@
 %func1 = smt.declare_fun : !smt.func<() !smt.bool>
 %func2 = smt.declare_fun : !smt.func<(!smt.bool) !smt.bool>
 
+// CHECK-NEXT:    smt.apply_func %func1() : !smt.func<() !smt.bool>
+// CHECK-NEXT:    smt.apply_func %func2(%const1) : !smt.func<(!smt.bool) !smt.bool>
+smt.apply_func %func1() : !smt.func<() !smt.bool>
+smt.apply_func %func2(%const1) : !smt.func<(!smt.bool) !smt.bool>
```
alexarice, line 16: Worth including the return value here?

alexarice (approved):

xdsl/dialects/smt.py
math-fehr: Oh yeah my bad, I forgot about that one!

tests/filecheck/dialects/smt/ops.mlir
math-fehr: I'm copying MLIR syntax here (I'll add the compatible tests once we have a version that includes SMT).
The return type is the second `smt.bool` in the `smt.func`

xdsl/dialects/smt.py
superlopuh: Yeah I like it

math-fehr: Done!

PR #4425, opened 2025-05-20
dialects: (smt) Implement `smt.implies` operation

math-fehr:


xdsl/dialects/smt.py
```diff
@@ -196,6 +197,29 @@ class XOrOp(VariadicBoolOp):
     name = "smt.xor"
 
 
+@irdl_op_definition
+class ImpliesOp(IRDLOperation):
```
alexarice, line 201: Am I missing something, why not use `VariadicBoolOp`? The docstring you gave even says this takes a variadic number of operands, though the definition does not agree with this

math-fehr: I need to fix the docstring, my bad

math-fehr: Done!

xdsl/dialects/smt.py
```diff
@@ -196,6 +197,27 @@ class XOrOp(VariadicBoolOp):
...
+    """
+    This operation performs a boolean implication. The semantics are equivalent
+    to the `=>` operator in the Core theory of the SMT-LIB Standard 2.7.
+    """
+
+    name = "smt.implies"
+
+    lhs = operand_def(base(BoolType))
```
alexarice, line 209:
```suggestion
    lhs = operand_def(BoolType)
```

xdsl/dialects/smt.py
```diff
@@ -196,6 +197,27 @@ class XOrOp(VariadicBoolOp):
...
+    rhs = operand_def(base(BoolType))
```
alexarice, line 210:
```suggestion
    rhs = operand_def(BoolType)
```

xdsl/dialects/smt.py
```diff
@@ -196,6 +197,27 @@ class XOrOp(VariadicBoolOp):
...
+    result = result_def(BoolType())
```
alexarice, line 211:
```suggestion
    result = result_def(BoolType)
```

alexarice (approved):
Have suggested some constraint simplifications but looks good.

PR #4428, opened 2025-05-20
core: (SSAValue) make `get` generic

alexarice:
Add a `type` argument to `SSAValue.get` allowing it to return `SSAValue`s with a more specialised generic paremeter.

This allows us to simplify a lot of cases where an SSAValue is obtained before checking its type.

### DIFF
--- a/xdsl/dialects/accfg.py
+++ b/xdsl/dialects/accfg.py
@@ -137,10 +137,10 @@ class LaunchOp(IRDLOperation):
         param_names: Iterable[str] | Iterable[StringAttr],
         state: SSAValue | Operation,
     ):
-        state_val: SSAValue = SSAValue.get(state)
+        state_val = SSAValue.get(state, type=StateType)
 
-        if not isinstance(state_val.type, StateType):
-            raise ValueError("`state` SSA Value must be of type `accfg.state`!")
+        # if not isinstance(state_val.type, StateType):
+        #     raise ValueError("`state` SSA Value must be of type `accfg.state`!")
 
         param_names_tuple: tuple[StringAttr, ...] = tuple(
             StringAttr(name) if isinstance(name, str) else name for name in param_names
--- a/xdsl/dialects/accfg.py
+++ b/xdsl/dialects/accfg.py
@@ -139,9 +139,6 @@ class LaunchOp(IRDLOperation):
     ):
         state_val = SSAValue.get(state, type=StateType)
 
-        # if not isinstance(state_val.type, StateType):
-        #     raise ValueError("`state` SSA Value must be of type `accfg.state`!")
-
         param_names_tuple: tuple[StringAttr, ...] = tuple(
             StringAttr(name) if isinstance(name, str) else name for name in param_names
         )
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -542,16 +542,22 @@ class SSAValue(Generic[AttributeCovT], IRWithUses, ABC):
         return name is None or cls._name_regex.fullmatch(name)
 
     @staticmethod
-    def get(arg: SSAValue | Operation) -> SSAValue:
+    def get(
+        arg: SSAValue | Operation, *, type: type[AttributeInvT] = Attribute
+    ) -> SSAValue[AttributeInvT]:
         "Get a new SSAValue from either a SSAValue, or an operation with a single result."
         match arg:
             case SSAValue():
-                return arg
+                if type == Attribute or isinstance(arg.type, type):
+                    return cast(SSAValue[AttributeInvT], arg)
+                raise ValueError(
+                    f"SSAValue.get: Expected type {type} but got SSAValue with type {arg.type}."
+                )
             case Operation():
                 if len(arg.results) == 1:
-                    return arg.results[0]
+                    return SSAValue.get(arg.results[0], type=type)
                 raise ValueError(
-                    "SSAValue.build: expected operation with a single result."
+                    "SSAValue.get: expected operation with a single result."
                 )
 
     def replace_by(self, value: SSAValue) -> None:
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -27,7 +27,7 @@ from typing import (
     overload,
 )
 
-from typing_extensions import Self, TypeVar
+from typing_extensions import Self, TypeForm, TypeVar
 
 from xdsl.traits import IsTerminator, NoTerminator, OpTrait, OpTraitInvT
 from xdsl.utils.exceptions import VerifyException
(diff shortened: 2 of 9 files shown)


xdsl/ir/core.py
```diff
@@ -542,16 +542,22 @@ def is_valid_name(cls, name: str | None):
...
+                raise ValueError(
+                    f"SSAValue.get: Expected type {type} but got SSAValue with type {arg.type}."
+                )
             case Operation():
                 if len(arg.results) == 1:
-                    return arg.results[0]
+                    return SSAValue.get(arg.results[0], type=type)
                 raise ValueError(
-                    "SSAValue.build: expected operation with a single result."
+                    "SSAValue.get: expected operation with a single result."
```
superlopuh, line 560: Uh oh, how do we not have tests for this! Could you please add one for this and the other error?

superlopuh (commented):
I like it

xdsl/ir/core.py
alexarice: sure, will do.

xdsl/ir/core.py
```diff
@@ -542,16 +542,22 @@ def is_valid_name(cls, name: str | None):
...
     @staticmethod
-    def get(arg: SSAValue | Operation) -> SSAValue:
+    def get(
+        arg: SSAValue | Operation, *, type: type[AttributeInvT] = Attribute
+    ) -> SSAValue[AttributeInvT]:
         "Get a new SSAValue from either a SSAValue, or an operation with a single result."
         match arg:
             case SSAValue():
-                return arg
+                if type == Attribute or isinstance(arg.type, type):
```
superlopuh, line 551: ?
```suggestion
                if type is Attribute or isinstance(arg.type, type):
```

alexarice:
I've converted various uses where we get a simplification due to this. I could split these off to be separate if preferred

xdsl/ir/core.py
```diff
@@ -542,16 +542,22 @@ def is_valid_name(cls, name: str | None):
         return name is None or cls._name_regex.fullmatch(name)
 
     @staticmethod
-    def get(arg: SSAValue | Operation) -> SSAValue:
+    def get(
+        arg: SSAValue | Operation, *, type: type[AttributeInvT] = Attribute
```
alexarice, line 546: is there any way to have a `TypeForm` here instead of a `type`? There was a case where it would be good for the type to be `MemRefType | UnrankedMemRefType` but the `isinstance` check below wasn't happy

xdsl/dialects/accfg.py
```diff
@@ -137,10 +137,10 @@ def __init__(
         param_names: Iterable[str] | Iterable[StringAttr],
         state: SSAValue | Operation,
     ):
-        state_val: SSAValue = SSAValue.get(state)
+        state_val = SSAValue.get(state, type=StateType)
 
-        if not isinstance(state_val.type, StateType):
-            raise ValueError("`state` SSA Value must be of type `accfg.state`!")
+        # if not isinstance(state_val.type, StateType):
+        #     raise ValueError("`state` SSA Value must be of type `accfg.state`!")
```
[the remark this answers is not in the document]
alexarice: Done

xdsl/ir/core.py
superlopuh: Why not, have you tried it? I'd actually do an isa then instead of isinstance

alexarice: I get 
```
Argument of type "TypeForm[AttributeInvT@get]" cannot be assigned to parameter "class_or_tuple" of type "_ClassInfo" in function "isinstance"
```
I could try `isa`

alexarice: It turns out the `init` for `ToTensorOp` (which is what I wanted this for, doesn't actually make any sense as it checks that an `UnrankedMemRefType` is a `ShapedType`

xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
```diff
@@ -497,7 +497,9 @@ def match_and_rewrite(self, op: AccessOp, rewriter: PatternRewriter, /):
             else:
                 memref_load_args.append(arg.idx)
 
-        load = memref.LoadOp.get(op.temp, memref_load_args)
+        load = memref.LoadOp(
```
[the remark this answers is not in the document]
superlopuh: I'm a little confused, isn't the op itself constrained to have an operand of MemRefType?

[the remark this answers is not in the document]
alexarice: sorry I replied to the wrong message, `LoadOp` expects a `MemRefType` but this function passes it an `SSAValue[StencilSomething]`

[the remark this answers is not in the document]
superlopuh: I think that's an actual bug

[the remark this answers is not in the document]
superlopuh: It just happens that they both have an `element_type` property and that it does the right thing but if we checked isinstance instead of casting things would fail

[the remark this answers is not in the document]
alexarice: Yeah, it for sure looks suspicious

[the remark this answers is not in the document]
superlopuh: My current feeling is that your fix is the way to go for rewrites that change the types of things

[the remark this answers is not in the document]
alexarice: :+1: I think this PR might be in a local maxima then

xdsl/ir/core.py
superlopuh: Worth a fix in a separate PR IMO

math-fehr: > no way to get a name out for the error message
What do you mean by this? Having an `isa` would be really cool here!

alexarice: I currently put `type.name` in the error message but this doesn't work if `type` is a `TypeForm`, so the cost of allowing a `TypeForm` would be to make the error message worse I think (unless you can see a way round this.

math-fehr (approved):
That's pretty nice!
I would just like to have `TypeForm` if possible, but otherwise yes!

math-fehr:
Yes:
image

math-fehr:
That's actually expected, because your `ArrayAttr` might come from another file right?
I would personally say that's fine, but it's up to you

alexarice:
```
>>> print(IndexType)
<class 'xdsl.dialects.builtin.IndexType'>
>>> print(IndexType | int)
xdsl.dialects.builtin.IndexType | int
```
I'm unsure what it uses to print without the `<class`

math-fehr:
I mean, we can probably write something like `if isinstance(x, type) print(type.__name__) else print(type)` ?

alexarice:
Updated to typeform and added another test

xdsl/ir/core.py
alexarice: #4431 for init fix

math-fehr:
Looks good to me! Thanks a lot!

math-fehr (approved):

superlopuh (approved):

PR #4429, opened 2025-05-20
dialects: (x86) migrate operations to consistent mnemonic

superlopuh:
This proved to be too weird, and also still ambiguous, as `imm` in the op name mnemonic could mean `immediate` or `inout`, `memory`, `memory`. Having one-letter mnemonics seems better. I'll propose a new letter for inout registers.


superlopuh:
Basically a new go at #4382

alexarice:
I think it'd help to see the full naming scheme before changing/reverting separate parts of it

superlopuh:
Fine I'll do it all in one PR

xdsl/dialects/x86/ops.py
```diff
@@ -316,43 +346,42 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
...
+    A base class for x86 operations that load from memory (m, r2) into a destination
+    register (d, r1).
     """
 
-    r1 = operand_def(R1InvT)
+    r2 = operand_def(R2InvT)  # memory source
     offset = attr_def(IntegerAttr, default_value=IntegerAttr(0, 64))
-
-    result = result_def(R2InvT)
+    r1 = result_def(R1InvT)  # destination register
```
superlopuh, line 359: note the `assembly_line_args` function below, the destination register is passed first, making it `r1`. This is the only major functional change in this file, the rest is just renaming.

xdsl/dialects/x86/ops.py
```diff
@@ -197,83 +224,86 @@ def assembly_line(self) -> str | None:
...
+class RS_Operation(
     Generic[R1InvT, R2InvT], X86Instruction, X86CustomFormatOperation, ABC
 ):
     """
     A base class for x86 operations that have two registers.
     """
 
-    r1 = operand_def(R1InvT)
+    r1_source = operand_def(R1InvT)
     r2 = operand_def(R2InvT)
```
alexarice, line 235: Should this be `s`?

xdsl/dialects/x86/ops.py
```diff
@@ -197,83 +224,86 @@ def assembly_line(self) -> str | None:
...
 ):
     """
-    A base class for x86 operations that have one register and one memory access with an optional offset.
+    A base class for x86 operations that have one inout register and one memory access
+    with an optional offset.
     """
 
-    r1 = operand_def(R1InvT)
+    r1_source = operand_def(R1InvT)
     r2 = operand_def(R2InvT)
```
alexarice, line 303: m?

alexarice (commented):
I started reviewing but think I have got confused and it's probably not worth putting the same comment on every op. Why don't the operands and results line up with the mnemonics, so that each `r` has a source and destination, a `d` appears only as an result, `s` appears only as operand etc.

qaco (approved):

xdsl/dialects/x86/ops.py
```diff
@@ -86,12 +113,15 @@ def parse(cls, parser: Parser) -> Self:
...
-        )
+        try:
+            return cls.build(
+                operands=operands,
+                result_types=result_types,
+                attributes=attributes,
+                regions=regions,
+            )
+        except ValueError:
+            raise parser.raise_error(f"{cls}", pos)
```
alexarice, line 124: Is this meant to be in this PR?

xdsl/dialects/x86/ops.py
```diff
@@ -421,69 +455,69 @@ def custom_print_attributes(self, printer: Printer) -> Set[str]:
         return {"immediate"}
 
 
-class M_MR_Operation(
+class MS_Operation(
     Generic[R1InvT, R2InvT], X86Instruction, X86CustomFormatOperation, ABC
 ):
     """
     A base class for x86 operations that have one memory reference and one register.
```
alexarice, line 462:
```suggestion
    A base class for x86 operations that have one memory reference and one source register.
```

alexarice: or something similar?

xdsl/dialects/x86/ops.py
```diff
@@ -637,63 +670,63 @@ def custom_parse_attributes(cls, parser: Parser) -> dict[str, Attribute]:
...
-        return {"immediate", "offset"}
+            print_immediate_value(printer, self.memory_offset)
+        return {"immediate", "memory_offset"}
 
 
-class M_M_Operation(Generic[R1InvT], X86Instruction, X86CustomFormatOperation, ABC):
+class M_Operation(Generic[R1InvT], X86Instruction, X86CustomFormatOperation, ABC):
     """
     A base class for x86 operations with a memory reference that's both a source and a
     destination
```
alexarice, line 688:
```suggestion
    A base class for x86 operations with a memory reference.
```

xdsl/dialects/x86/ops.py
```diff
@@ -815,224 +848,223 @@ def parse(cls, parser: Parser) -> Self:
         return op
 
 
-class RRROperation(
+class RSS_Operation(
     Generic[R1InvT, R2InvT, R3InvT], X86Instruction, X86CustomFormatOperation, ABC
 ):
     """
     A base class for x86 operations that have three registers.
```
alexarice, line 855:
```suggestion
    A base class for x86 operations that have one inout and two source registers.
```

xdsl/dialects/x86/ops.py
```diff
@@ -815,224 +848,223 @@ def parse(cls, parser: Parser) -> Self:
...
 
 # endregion
 
 
 @irdl_op_definition
-class RR_AddOp(R_RR_Operation[GeneralRegisterType, GeneralRegisterType]):
+class RS_AddOp(RS_Operation[GeneralRegisterType, GeneralRegisterType]):
     """
-    Adds the registers r1 and r2 and stores the result in r1.
+    Adds the registers r and s and stores the result in r.
```
alexarice, line 893: You're going to hate me but this would make more sense if the operands/results were called `r` and `s`

alexarice (approved):

xdsl/dialects/x86/ops.py
superlopuh: Ah no sorry good catch it was my debugging

xdsl/dialects/x86/ops.py
```diff
@@ -197,194 +227,198 @@ def assembly_line(self) -> str | None:
 # region: Operation Base Classes
 
 
-class R_RR_Operation(
+class RS_Operation(
     Generic[R1InvT, R2InvT], X86Instruction, X86CustomFormatOperation, ABC
 ):
     """
-    A base class for x86 operations that have two registers.
+    A base class for x86 operations that have two registers, where the contents of the
```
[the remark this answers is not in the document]
superlopuh: Definitely worth being consistent, I'll do a pass

xdsl/dialects/x86/ops.py
superlopuh: Train has left the station

compor (commented):
post-merge review: I think this conveys the use clearly enough.

I just wanted to say that names like "register_in" (or "inout") and the like, in my head, imply scope (as in high-level languages), which is not the case in assembly (despite our structured approach within MLIR, but I guess they are fine in comments.

superlopuh:
Yeah that is definitely my least favourite part of this naming scheme, I guess we can change it later if we come up with something better.

PR #1566, opened 2023-09-11
dialects: (riscv) canonicalization pattern added (x+x=x*2)

dshaaban01:

PR #4430, opened 2025-05-20
docs: (GETTING_STARTED.md) Add two new canonicalization suggestions

MDLC01:
Combined with https://github.com/xdslproject/xdsl/pull/1566, this ends up canonicalizing `x + x`  to `x << 1`, hence the updated test.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -47,6 +47,7 @@ You're welcome to come up with your own, or do one of the following:
 - `x + x * n -> x * (n + 1)` (where `n` is an immediate)
 - `x >> 0 -> x`
 - `x * 1 -> x`
+- `n * x -> x * n` (where `n` is an immediate)
 
 The patterns are defined in
 [xdsl/transforms/canonicalization_patterns/riscv.py](xdsl/transforms/canonicalization_patterns/riscv.py).
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -44,7 +44,7 @@ Here are some examples of first PRs from existing contributors:
 The first step is finding a missing optimisation pattern.
 You're welcome to come up with your own, or do one of the following:
 
-- `x * 2ⁱ -> x << i`
+- `x + x * n -> x * (n + 1)` (where `n` is an immediate)
 - `x >> 0 -> x`
 - `x * 1 -> x`
 
--- a/tests/filecheck/projects/riscv-backend-paper/bottom_up_f32.mlir
+++ b/tests/filecheck/projects/riscv-backend-paper/bottom_up_f32.mlir
@@ -91,13 +91,17 @@ func.func public @pooling_nchw_max_d1_s2_3x3(
 // CHECK-NEXT:      mv t2, zero
 // CHECK-NEXT:      # Constant folded riscv_cf.bge
 // CHECK-NEXT:  scf_body_{{\d}}_for:
-// CHECK-NEXT:      slli a2, t2, 1
+// CHECK-NEXT:      li a2, 2
+// CHECK-NEXT:      mul a2, t2, a2
 // CHECK-NEXT:      li t6, 18
 // CHECK-NEXT:      mul a2, a2, t6
-// CHECK-NEXT:      slli a2, a2, 3                               # multiply by element size
+// CHECK-NEXT:      li t6, 8
+// CHECK-NEXT:      mul a2, a2, t6                               # multiply by element size
 // CHECK-NEXT:      add a2, t1, a2
-// CHECK-NEXT:      slli t6, t2, 3
-// CHECK-NEXT:      slli t6, t6, 3                               # multiply by element size
+// CHECK-NEXT:      li t6, 8
+// CHECK-NEXT:      mul t6, t2, t6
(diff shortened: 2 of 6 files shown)


xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -517,3 +542,18 @@ def get_constant_value(value: SSAValue) -> riscv.Imm32Attr | None:
...
+
+    if v.value.data < 1:
+        return
+
+    # Since we use 32-bit signed integers, we can hardcode all 31 powers of 2.
+    powers_of_two = [1 << i for i in range(31)]
+    try:
+        return powers_of_two.index(v.value.data)
+    except ValueError:
+        return
```
alexarice, line 559:
```suggestion
if int.bit_count(v.value.data) == 1:
    return int.bit_length(v.value.data) - 1
```
Alternative to the list indexing

xdsl/transforms/canonicalization_patterns/riscv.py
```diff
@@ -58,6 +58,31 @@ def match_and_rewrite(self, op: riscv.MulOp, rewriter: PatternRewriter) -> None:
             rewriter.replace_matched_op(riscv.MVOp(op.rs2, rd=rd))
 
 
+class MultiplyImmediatePowerOfTwo(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: riscv.MulOp, rewriter: PatternRewriter) -> None:
+        if (log_rs1 := exact_log2(op.rs1)) is not None:
+            rd = cast(riscv.IntRegisterType, op.rd.type)
```
alexarice, line 65: It's a shame this cast is needed. It looks like a pyright bug to me after inspecting the types.

alexarice (approved):
Looks good to me.

An alternative to doing a "two-sided" rewrite pattern like this is to only supply one side, but have a canonicalisation pattern which converts `non-const * const` to `const * non-const`. Such a pattern could probably be a case of `MultiplyImmediates`. I think something similar is done in `arith` as an example. (Not necessarily asking for a change here but it's another option to keep in mind).

xdsl/transforms/canonicalization_patterns/riscv.py
MDLC01: That's better indeed. I was not aware Python had those methods. I still need to check that the value is positive though.

superlopuh:
Oh one more thing I forgot to say is that the convention in xDSL is to let he original commenter resolve the comments, as that makes sure that the comment was addressed as intended.

MDLC01:
Oh, sorry

tests/filecheck/projects/riscv-backend-paper/bottom_up_f32.mlir
```diff
@@ -91,17 +91,13 @@ func.func public @pooling_nchw_max_d1_s2_3x3(
...
 // CHECK-NEXT:      li t6, 18
 // CHECK-NEXT:      mul a2, a2, t6
-// CHECK-NEXT:      li t6, 8
-// CHECK-NEXT:      mul a2, a2, t6                               # multiply by element size
+// CHECK-NEXT:      slli a2, a2, 3                               # multiply by element size
 // CHECK-NEXT:      add a2, t1, a2
-// CHECK-NEXT:      li t6, 8
-// CHECK-NEXT:      mul t6, t2, t6
-// CHECK-NEXT:      li t5, 8
-// CHECK-NEXT:      mul t6, t6, t5                               # multiply by element size
```
superlopuh, line 104: These changes are scary and make me think that we might not want to actually add this change. The main issue is that other rewrites, like the one you proposed in the readme, rely on checking that an input is a product, and the check will fail here as it'll be a shift instead. I have a feeling that this optimisation should be in a different pass that optimises the code, as opposed to "canonicalization", which is designed to make the code more uniform to make downstream transformations easier.

GETTING_STARTED.md
```diff
@@ -44,7 +44,7 @@ Here are some examples of first PRs from existing contributors:
 The first step is finding a missing optimisation pattern.
 You're welcome to come up with your own, or do one of the following:
 
-- `x * 2ⁱ -> x << i`
+- `x + x * n -> x * (n + 1)` (where `n` is an immediate)
```
superlopuh, line 47: My current gut feeling is that the previous suggestion was not very good, as it's a valid optimisation (on some targets), but not a great canonicalization (since it actually makes code harder to process by the compiler than the input). I think the changes in the rest of the PR are already evidence of you getting the APIs etc, which was most of the point of the exercise, and that it would be ideal to just merge the change to this line, proposing a better canonicalization pattern.

superlopuh: BTW it's better because in the LHS, x is used twice, and in the rhs it's used once, which may unblock some optimisations that check the number of uses of an ssa value.

superlopuh (approved):
Yep, please update title, then should be good to go!

GETTING_STARTED.md
```diff
@@ -44,9 +44,10 @@ Here are some examples of first PRs from existing contributors:
...
```
alexarice, line 47: Would this canonicalisation still be good when `n` is not an immediate?

PR #4431, opened 2025-05-20
dialects: (bufferization) fix ToTensorOp init

alexarice:
Fixes the init method for `ToTensorOp`. Factors out the logic for converting to and from memref and tensor types to static methods on `TensorFromMemRefType`. 

Stacked on #4428

### DIFF
--- a/tests/irdl/test_attr_constraint.py
+++ b/tests/irdl/test_attr_constraint.py
@@ -2,7 +2,15 @@ from abc import ABC
 
 import pytest
 
-from xdsl.dialects.builtin import StringAttr
+from xdsl.dialects.bufferization import TensorFromMemRefConstraint
+from xdsl.dialects.builtin import (
+    MemRefType,
+    StringAttr,
+    TensorType,
+    UnrankedMemRefType,
+    UnrankedTensorType,
+    i32,
+)
 from xdsl.ir import Attribute, Data, ParametrizedAttribute
 from xdsl.irdl import (
     AllOf,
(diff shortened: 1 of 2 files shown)


superlopuh (commented):
Changes look great, but it feels like it would be better to unstack this, and merge this before the get change. It would great to have a couple of tiny tests for the new helpers also.

alexarice:
I'm not sure I understand the benefit of unstacking. This code uses the get change so unstacking would create a temporary version of this code.

superlopuh:
My perspective is that this change would partially be overwritten by the other change, but it has code that would not be overwritten, as in all the code that actually fixes the bug present in main. I don't feel strongly enough about this currently to insist :) I just think it is missing tests then it's GTG

superlopuh (approved):

tests/irdl/test_attr_constraint.py
```diff
@@ -153,3 +161,29 @@ class NoParamAttr(BaseNoParamAttr): ...
...
+    input: TensorType | UnrankedTensorType, output: MemRefType | UnrankedMemRefType
+):
+    assert TensorFromMemRefConstraint.tensor_to_memref(input) == output
+
+
+@pytest.mark.parametrize(
+    "input, output",
+    [
+        (MemRefType(i32, [2, 2]), TensorType(i32, [2, 2])),
+        (UnrankedMemRefType.from_type(i32), UnrankedTensorType(i32)),
```
alexarice, line 183: @superlopuh I don't really understand all the `memory_space` stuff, is there any case worth adding here?

superlopuh: Oh yeah, that's actually pretty crucial, and easily broken

alexarice: I expect it's currently broken on main, but maybe this is a good chance to fix it

superlopuh: I've just looked it up, tensor types don't have a memory space, so there's not much to do, I guess this should now be OK

PR #4432, opened 2025-05-20
core: (constraints) experiment with ditching extractors

alexarice:
Had a small attempt at doing constraints without the complicated extractor nonsense. This feels like it should be less efficient, but I imagine in practice it might just be fine.

Partially motivated because I was really struggling with implementing constraint variable directives, as there didn't seem to be a good way to tell the type of a constraint variable. This implementation of `variables` would fix that.


xdsl/irdl/constraints.py
```diff
@@ -256,23 +199,16 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
...
-    def get_variable_extractors(self) -> dict[str, VarExtractor[Attribute]]:
-        return merge_extractor_dicts(
-            self.attr_constraint.get_variable_extractors(),
-            {
-                v: self._Extractor(r)
-                for v, r in self.type_constraint.get_variable_extractors().items()
-            },
-        )
+    def variables(self) -> dict[str, ConstraintVarType]:
+        return self.type_constraint.variables() | self.attr_constraint.variables()
```
alexarice, line 203: Note that this isn't doing the duplicate key checking that the extractor implementation was doing.

math-fehr:
If I recall correctly (and tell me if I'm wrong here), but wasn't this kind of what we had before? I thought the goal of the extractor was just to not duplicate work (finding the type variable versus extracting the type variable)?

math-fehr:
What do you mean by that (btw my memory is quite bad about this part)

math-fehr:
In any case, to me that seems equivalent to what we had before I think? It's just that instead of creating classes to represent the extractors, we just encode this as a function in the constraint class.
So to me that looks good!

alexarice:
This should be functionally equivalent but might run `variables()` more times than strictly necessary.

math-fehr:
Okay, do you want me to review this now? Or should I wait?

alexarice:
I'll take a look through old PRs first if you want

math-fehr:
I don't think you have to, as you want!

xdsl/irdl/constraints.py
superlopuh: If we merge this version it would be nice to have it

xdsl/irdl/constraints.py
```diff
@@ -179,14 +129,18 @@ def verifies(self, attr: Attribute) -> TypeGuard[AttributeCovT]:
...
-        the value of the attribute which is verified by this constraint.
+        Returns a dictionary of the variables that can be extracted by this constraint.
         """
         return {}
 
+    def extract_var(self, attr: Attribute, var: str) -> ConstraintVariableType:
+        """
+        Extracts the value of an constraint variable `var` from the input attribute `attr`.
+        """
+        raise ValueError("Cannot extract variable from constraint")
```
superlopuh, line 142: It would be nice to give as much info for debugging as possible in these errors, otherwise one has to add them manually when it's raised.

xdsl/irdl/constraints.py
```diff
@@ -256,23 +199,16 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
...
+    def variables(self) -> dict[str, ConstraintVarType]:
+        return self.type_constraint.variables() | self.attr_constraint.variables()
+
+    def extract_var(self, attr: Attribute, var: str) -> ConstraintVariableType:
+        if var in self.attr_constraint.variables():
+            return self.attr_constraint.extract_var(attr, var)
+        else:
+            if not isinstance(attr, TypedAttribute):
+                raise PyRDLError(f"Inference expected {attr} to be a TypedAttribute")
+            return self.type_constraint.extract_var(attr.get_type(), var)
```
superlopuh, line 211: Feels like the pythonic way of doing this would be either to raise and catch a KeyError or return None if not found, instead of first checking whether it's in a dict then extracting.

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -3219,106 +3214,6 @@ class DefaultAttrDictOp(IRDLOperation):
     check_equivalence(program, generic, ctx)
 
 
-################################################################################
-#                                Extractors                                    #
-################################################################################
```
superlopuh, line 3224: what's the deal with these tests?

xdsl/irdl/constraints.py
```diff
@@ -819,23 +769,31 @@ def verify(
...
+        Returns a dictionary of the variables that can be extracted from the range length by this constraint.
+        """
+        return {}
+
+    def extract_var(
+        self, attrs: Sequence[Attribute], var: str
+    ) -> ConstraintVariableType:
         """
-        Get a dictionary of variables that can be solved using the length of the range.
+        Extracts the value of an constraint variable `var` from the input attribute range `attrs`.
```
superlopuh, line 788:
```suggestion
        Extracts the value of a constraint variable `var` from the input attribute range `attrs`.
```

xdsl/irdl/constraints.py
```diff
@@ -819,23 +769,31 @@ def verify(
...
         """
-        return dict()
+        raise ValueError("Cannot extract variable from constraint")
```
superlopuh, line 790: Same here would be great to have more info, and below

xdsl/irdl/constraints.py
```diff
@@ -90,67 +97,10 @@ def update(self, other: ConstraintContext):
...
-    The keys of the resulting dictionary is equal to the union of the keys
-    of the input dictionaries, with duplicated values being merged to a
-    single MergeExtractor
-    """
-    keys = set(key for d in dicts for key in d)
-    return {v: merge_extractors(v, *(d[v] for d in dicts if v in d)) for v in keys}
+class ConstraintVarType(StrEnum):
+    ATTRIBUTE = auto()
+    RANGE = auto()
+    INT = auto()
```
superlopuh, line 103: Where is this used?

alexarice: It's not used in this at all. I started doing this because I wanted to be able to implement an `AttributeConstraintVariableDirective` which would allow you to put something like `%T` in assembly format for a constraint variable `T`. The problem is that there's no easy way to have an operation definition and variable name `T` and tell whether `T` is a attribute variable, range variable, or int variable, which you need to know in order to tell how to parse. That said I imagine this should be a separate PR.

alexarice:
Initial feedback seems positive so I will try to adapt this into an actual PR when I get some time.

alexarice:
My memory now is that the old system did the following for inference:
- Make a context
- Run verification
- Pass the (now completed) context into the inference function.

At least I seem to remember that `verify` was the only function that instantiated variables.

xdsl/irdl/constraints.py
```diff
@@ -48,14 +56,13 @@ def get_range_variable(self, key: str) -> tuple[Attribute, ...] | None:
...
 
-    def set_variable(self, key: str, attr: Attribute):
-        self._variables[key] = attr
-
-    def set_range_variable(self, key: str, attr: tuple[Attribute, ...]):
-        self._range_variables[key] = attr
-
-    def set_int_variable(self, key: str, i: int):
-        self._int_variables[key] = i
+    def set_variable(self, key: str, attr: ConstraintVariableType):
```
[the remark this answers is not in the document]
alexarice: This might be less of a saving than I though though, as the `resolve_constraint_variables` function will need to deal with duplicates.

xdsl/irdl/constraints.py
math-fehr: I am not that sure about that?
We don't want to return `None`, because the contract is that we only extract a variable if we expect it to be there no?
I find catching `KeyError` a bit yuck in my opinion. To be honest this should rather be an assert, as we don't expect this error to be visible to the user, as it should never be triggered whatever the user inputs are.

math-fehr:
Btw, what do you all think about returning multiple variables instead of one?
Maybe I'm overthinking this, but returning a dictionary of variables would reduce the amount of calls we do, and would solve a lot of the problems we have with `if not in ... then ...`, as we would return `{}` when nothing is found?

math-fehr:
If this is scope creep let's just forget it btw

alexarice:
That might be a good idea, do we never actually ever resolve a single variable?

superlopuh:
Oh yeah in that case why not just call this with a context?

alexarice:
Mainly just that returning the variables felt cleaner than mutating a context, but maybe it's not too bad? Can more the duplicate key checking logic to the context as well which fixes the other issue I'm having.

alexarice:
Reworked this and should be ready for review

xdsl/dialects/builtin.py
```diff
@@ -228,20 +226,13 @@ def infer(self, context: ConstraintContext) -> ArrayAttr[AttributeCovT]:
     def get_unique_base(self) -> type[Attribute] | None:
         return ArrayAttr
 
-    @dataclass(frozen=True)
-    class _Extractor(VarExtractor[Attribute]):
-        inner: VarExtractor[Sequence[Attribute]]
+    def variables(self) -> set[str]:
+        return self.elem_range_constraint.variables()
```
superlopuh, line 230: how do you feel about this being an iterator instead? the caller can then call `set(constr.variables)` if they feel like it, and we avoid building and throwing away a ton of objects

alexarice: Would likely be fine as an iterator, sure

xdsl/irdl/constraints.py
superlopuh: Am I right in understanding that this is no longer relevant?

xdsl/dialects/builtin.py
math-fehr: Would an iterator really be better here? What would we gain?
I'm not sure that a `yield` is really faster than creating a small `set`?

alexarice: Yeah I started coding it up and you're just replacing set generation with python generator generation everywhere.

alexarice: I imagine if you really care about speed then the ideal is passing in a set which gets mutated, but that feels less nice to me.

superlopuh: I just think it's cleaner, the implementation should be simpler, and it's more direct

alexarice: just to be clear, you'd prefer:
```python
yield from a
yield from b
```
to 
```python
return a | b
```

math-fehr: Wait, what would happen when there are twice the same variable? The iterator way wouldn't work no?

alexarice: I think the idea is that it would duplicate the variable, and you generate one set from the iterator at the end which would remove the duplicate

math-fehr: I think this is quite yuck in my opinion. This is quite error-prone no?

superlopuh: I think it's the opposite, it lets you do things like detect that there are duplicate values, if you feel like it.

alexarice: I'm not sure I see the utility in that.

alexarice:
I'm becoming less sure again that this is the correct place to be checking for uniqueness of extracting constraint variables. I feel I would expect things like this to fail at verification rather than at parsing.

xdsl/dialects/builtin.py
superlopuh: I mean I also am not sure but it feels like the least opinionated way of implementing this function, and should also be more efficient and easy to read

alexarice:
I don't see why it would be inconsistent

superlopuh:
Oh yeah can't we just remove the extraction here entirely and verify instead?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -120,7 +116,7 @@ def parse(
             stmt.parse(parser, state)
 
         # Get constraint variables from the parsed operand and result types
-        self.resolve_constraint_variables(state)
+        self.resolve_constraint_variables(state, op_def)
```
superlopuh, line 119: This seems like a good place to catch any errors during extraction and re-raise it as a parsing error, with the start/end pos after parsing `self.stmts`?

alexarice:
> Oh yeah can't we just remove the extraction here entirely and verify instead?

There's definitely a couple of differences but you're right that it kind of feels like I'm just reverting the original change

superlopuh:
The real treasure was the friends we made along the way

superlopuh:
I don't recall exactly but it feels like the structure of the assembly format parsing is cleaner than when the extractors were added.

alexarice:
I'm more worried that I will have broken something downstream, where we have some nastier constraints, as I definitely remember introducing the previous system for a reason

superlopuh:
Downstream where? Would we not have added tests for the more complex cases?

alexarice:
https://github.com/xdslproject/inconspiquous/blob/35c4f9794f4d2aa32eff112e4b51b3dbc7030bc7/inconspiquous/constraints.py#L26 one example

alexarice:
I think it should be fine.

xdsl/irdl/constraints.py
```diff
@@ -48,17 +54,35 @@ def get_range_variable(self, key: str) -> tuple[Attribute, ...] | None:
     def get_int_variable(self, key: str) -> int | None:
         return self._int_variables.get(key)
 
-    def set_variable(self, key: str, attr: Attribute):
+    def set_attr_variable(self, key: str, attr: Attribute):
```
math-fehr, line 57: Ohhh, that's really nice!

xdsl/irdl/constraints.py
```diff
@@ -179,13 +140,18 @@ def verifies(self, attr: Attribute) -> TypeGuard[AttributeCovT]:
         except VerifyException:
             return False
 
-    def get_variable_extractors(self) -> dict[str, VarExtractor[Attribute]]:
+    def variables(self) -> set[str]:
         """
-        Get a dictionary of constraint variables to extractors for these variables,
-        which provide a method to obtain the value of each constraint variable from
-        the value of the attribute which is verified by this constraint.
+        Returns a set of the variables that can be extracted by this constraint.
```
math-fehr, line 145: I would also write something like:
`These variables are always expected to be set by the attribute constraint`

xdsl/irdl/constraints.py
```diff
@@ -179,13 +140,18 @@ def verifies(self, attr: Attribute) -> TypeGuard[AttributeCovT]:
...
         """
-        return {}
+        return set()
+
+    def extract_variables(self, attr: Attribute, constraint_context: ConstraintContext):
```
math-fehr, line 149:
```suggestion
    def extract_variables(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
```

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -159,19 +155,44 @@ def parse(
...
-            match v:
-                case Attribute():
-                    ctx.set_variable(k, v)
-                case int():
-                    ctx.set_int_variable(k, v)
-                case _:
-                    ctx.set_range_variable(k, tuple(v))
-
-        state.context = ctx
+    def resolve_constraint_variables(self, state: ParsingState, op_def: OpDef):
```
math-fehr, line 158: Can you add documentation to this function?

math-fehr (approved):
Very very nice!

alexarice:
@math-fehr @superlopuh updated this based on yesterdays conversation.

xdsl/irdl/constraints.py
```diff
@@ -48,17 +54,35 @@ def get_range_variable(self, key: str) -> tuple[Attribute, ...] | None:
     def get_int_variable(self, key: str) -> int | None:
         return self._int_variables.get(key)
 
-    def set_variable(self, key: str, attr: Attribute):
+    def set_attr_variable(self, key: str, attr: Attribute):
+        if key in self._variables and self._variables[key] != attr:
+            raise ValueError(
+                f"Variable {key} was set to different attributes:\n"
+                f"Possible values: {self._variables[key]}, {attr}"
+            )
```
alexarice, line 62: This isn't being used anymore, either I can revert it or we could keep this logic (and change it to a verify error) and simplify the logic in `VarConstraint`. Any thoughts?

superlopuh (approved):

xdsl/irdl/constraints.py
```diff
@@ -815,27 +724,28 @@ def verify(
...
         """
         Check if the range satisfies the constraint, or raise an exception otherwise.
-        The range can contain Nones, which represent an attribute not to be checked.
         """
         ...
 
-    def get_variable_extractors(
-        self,
-    ) -> dict[str, VarExtractor[Sequence[Attribute]]]:
+    def verify_length(self, length: int, constraint_context: ConstraintContext) -> None:
```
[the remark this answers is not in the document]
alexarice: Do you feel `RangeVarConstraint` should actually check the size in `verify_length` if the variable is available? I don't like this as I feel it makes makes the verification depend on the order you verify things.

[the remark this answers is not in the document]
superlopuh: I'm not sure about the general case but it feels like it wouldn't hurt to think about this when implementing custom range constraints. Adding the default implementation manually is a small effort, and would probably be useful.

math-fehr (approved):
Very nice!!

xdsl/irdl/constraints.py
alexarice: I think I wrote this in zulip, but the issue with moving all the logic here is that we would end up reverifying the constraint attached to `VarConstraint` each time the variable was encountered. I'm leaning towards keeping the logic in `VarConstraint` and verify, and reverting the changes to `ConstraintContext`

PR #4434, opened 2025-05-21
core: (constraints) improve inference debug message

alexarice:
just add some more information to some exception messages which should never appear.


superlopuh (approved):
ty

PR #4442, opened 2025-05-22
dialects: (builtin) add typed helper for conversion cast

superlopuh:
In a somewhat similar vein to the new `SSAValue.get`, adds a helper for casting.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1773,6 +1773,14 @@ class UnrealizedConversionCastOp(IRDLOperation):
             result_types=[result_type],
         )
 
+    @staticmethod
+    def cast_one(
+        input: SSAValue, result_type: AttributeInvT
+    ) -> tuple[UnrealizedConversionCastOp, SSAValue[AttributeInvT]]:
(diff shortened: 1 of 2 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1773,6 +1773,14 @@ def get(inputs: Sequence[SSAValue | Operation], result_type: Sequence[Attribute]
             result_types=[result_type],
         )
 
+    @staticmethod
+    def cast_one(
+        input: SSAValue, result_type: AttributeInvT
+    ) -> tuple[UnrealizedConversionCastOp, SSAValue[AttributeInvT]]:
+        op = UnrealizedConversionCastOp(operands=(input,), result_types=(result_type,))
+        res: SSAValue[AttributeInvT] = op.results[0]  # pyright: ignore[reportAssignmentType]
```
math-fehr, line 1781: Should you insert the op in the module?
By passing an `InsertPoint` or a `Builder` for instance

superlopuh: Hmm, I feel like this is a bit more flexible, I use it in a local branch where we have a "replace op with list of ops" pattern, which is nice because then they all get the same name hint

math-fehr (approved):

PR #4443, opened 2025-05-22
dialects: (x86) use typed SSA values for inference in inits

superlopuh:
This makes pyright happy and the API less broken.

### DIFF
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -273,13 +273,15 @@ class R_Operation(Generic[R1InvT], X86Instruction, X86CustomFormatOperation, ABC
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -273,13 +273,15 @@ class R_Operation(Generic[R1InvT], X86Instruction, X86CustomFormatOperation, ABC
 
     def __init__(
         self,
-        register_in: Operation | SSAValue | None = None,
+        register_in: SSAValue[R1InvT],
         *,
         comment: str | StringAttr | None = None,
         register_out: R1InvT | None = None,
```
superlopuh, line 279: this was a 🐛  whereas now it's a beautiful 🦋

PR #4444, opened 2025-05-22
backend: (x86) add x86 register queue

superlopuh:
I'm certain I didn't add all the registers that can be used during register allocation, would appreciate suggestions and any pointers to resources on what registers can be used when!

### DIFF
--- a/xdsl/dialects/x86/register.py
+++ b/xdsl/dialects/x86/register.py
@@ -226,22 +226,25 @@ See external # [documentation](https://wiki.osdev.org/X86-64_Instruction_Encodin
 """
 
 UNALLOCATED_AVX2 = AVX2RegisterType.unallocated()
-YMM0 = AVX2RegisterType.from_name("ymm0")
-YMM1 = AVX2RegisterType.from_name("ymm1")
-YMM2 = AVX2RegisterType.from_name("ymm2")
-YMM3 = AVX2RegisterType.from_name("ymm3")
-YMM4 = AVX2RegisterType.from_name("ymm4")
-YMM5 = AVX2RegisterType.from_name("ymm5")
-YMM6 = AVX2RegisterType.from_name("ymm6")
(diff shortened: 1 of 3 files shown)


xdsl/dialects/x86/register.py
```diff
@@ -226,22 +226,25 @@ def infinite_register_prefix(cls):
...
-YMM7 = AVX2RegisterType.from_name("ymm7")
-YMM8 = AVX2RegisterType.from_name("ymm8")
-YMM9 = AVX2RegisterType.from_name("ymm9")
-YMM10 = AVX2RegisterType.from_name("ymm10")
-YMM11 = AVX2RegisterType.from_name("ymm11")
-YMM12 = AVX2RegisterType.from_name("ymm12")
-YMM13 = AVX2RegisterType.from_name("ymm13")
-YMM14 = AVX2RegisterType.from_name("ymm14")
-YMM15 = AVX2RegisterType.from_name("ymm15")
+YMM = tuple(AVX2RegisterType.from_name(f"ymm{i}") for i in range(16))
```
qaco, line 229: So pythonic ;)

xdsl/dialects/x86/register.py
```diff
@@ -226,22 +226,25 @@ def infinite_register_prefix(cls):
...
+    YMM7,
+    YMM8,
+    YMM9,
+    YMM10,
+    YMM11,
+    YMM12,
+    YMM13,
+    YMM14,
+    YMM15,
+) = YMM
```
qaco, line 247: This whole diff is very pythonic ;)

qaco (approved):

superlopuh:
It looks like here we've actually hit an interesting thing of overlapping registers, I'd love to discuss it in our next meeting. The encoding for EBX and RBX is the same, so the queue is confused, which breaks the test. IT feels like either we need to move to different register types for 32/64 bits or we need a smarter queue.

PR #4447, opened 2025-05-23
dialects: (emitc) add emitc array verify and parsing checks

Jimmy2027:
Add additional checks to the emitc array shape and type verification as implemented in the [reference](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp)

### DIFF
--- a/tests/dialects/test_emitc.py
+++ b/tests/dialects/test_emitc.py
@@ -1,6 +1,6 @@
 import pytest
 
-from xdsl.dialects.builtin import i32
+from xdsl.dialects.builtin import MemRefType, i32
 from xdsl.dialects.emitc import EmitC_ArrayType
 from xdsl.utils.exceptions import VerifyException
 
@@ -10,3 +10,22 @@ def test_emitc_array_negative_dimension():
         VerifyException, match="EmitC array dimensions must have non-negative size"
     ):
         EmitC_ArrayType([-1], i32)
+
+
+def test_emitc_array_nested_array_type():
+    nested_array = EmitC_ArrayType([2], i32)
+    with pytest.raises(
+        VerifyException,
+        match="EmitC array element type cannot be another EmitC_ArrayType.",
+    ):
+        EmitC_ArrayType([1], nested_array)
+
+
+def test_emitc_array_unsupported_element_type():
+    unsupported_type = MemRefType(element_type=i32, shape=[1])
+
+    with pytest.raises(
+        VerifyException,
+        match=f"EmitC array element type '{unsupported_type}' is not a supported EmitC type.",
+    ):
+        EmitC_ArrayType([1], unsupported_type)
--- a/tests/dialects/test_emitc.py
+++ b/tests/dialects/test_emitc.py
@@ -1,6 +1,6 @@
 import pytest
 
-from xdsl.dialects.builtin import MemRefType, i32
+from xdsl.dialects.builtin import i32
 from xdsl.dialects.emitc import EmitC_ArrayType
 from xdsl.utils.exceptions import VerifyException
 
@@ -10,22 +10,3 @@ def test_emitc_array_negative_dimension():
         VerifyException, match="EmitC array dimensions must have non-negative size"
     ):
         EmitC_ArrayType([-1], i32)
-
-
-def test_emitc_array_nested_array_type():
-    nested_array = EmitC_ArrayType([2], i32)
-    with pytest.raises(
-        VerifyException,
-        match="EmitC array element type cannot be another EmitC_ArrayType.",
-    ):
-        EmitC_ArrayType([1], nested_array)
-
-
-def test_emitc_array_unsupported_element_type():
-    unsupported_type = MemRefType(element_type=i32, shape=[1])
-
-    with pytest.raises(
-        VerifyException,
-        match=f"EmitC array element type '{unsupported_type}' is not a supported EmitC type.",
-    ):
-        EmitC_ArrayType([1], unsupported_type)
--- a/tests/dialects/test_emitc.py
+++ b/tests/dialects/test_emitc.py
@@ -5,11 +5,6 @@ from xdsl.dialects.emitc import EmitC_ArrayType
 from xdsl.utils.exceptions import VerifyException
 
 
-def test_emitc_array_empty_shape():
-    with pytest.raises(VerifyException, match="EmitC array shape must not be empty"):
-        EmitC_ArrayType([], i32)
-
-
 def test_emitc_array_negative_dimension():
     with pytest.raises(
         VerifyException, match="EmitC array dimensions must have non-negative size"
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/emitc/emitc_types.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/emitc/emitc_types.mlir
@@ -2,11 +2,13 @@
 
 // CHECK: f32_2D = !emitc.array<4x2xf32>
 // CHECK-SAME: f64_1D = !emitc.array<5xf64>
+// CHECK-SAME: i1_0D = !emitc.array<0xi1>
 // CHECK-SAME: i1_3D = !emitc.array<3x4x5xi1>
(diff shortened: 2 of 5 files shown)


tests/filecheck/mlir-conversion/with-mlir/dialects/emitc/emitc_types.mlir
```diff
@@ -2,11 +2,13 @@
 
 // CHECK: f32_2D = !emitc.array<4x2xf32>
 // CHECK-SAME: f64_1D = !emitc.array<5xf64>
+// CHECK-SAME: i1_0D = !emitc.array<0xi1>
 // CHECK-SAME: i1_3D = !emitc.array<3x4x5xi1>
 // CHECK-SAME: i32_1D = !emitc.array<10xi32>
```
superlopuh, line 7: Can you please order these so they're in the same order as the input? Not a big deal but threw me off slightly

tests/dialects/test_emitc.py
superlopuh, line 1: TBH I don't think we need two tests for this, a filecheck one should be enough

xdsl/dialects/emitc.py
```diff
@@ -81,6 +115,60 @@ def print_parameters(self, printer: Printer) -> None:
             printer.print_attribute(self.element_type)
 
 
+def is_supported_integer_type(type_attr: Attribute) -> bool:
+    """Check if an IntegerType is supported by EmitC."""
+    if isinstance(type_attr, IntegerType):
+        return type_attr.width.data in [1, 8, 16, 32, 64]
+    return False
```
superlopuh, line 122: So we actually have a mechanism for this sort of thing, and that's an AttributeConstraint, which can be seen as a function from Attribute to bool, except that in the case of False it also has an error message. Could you rewrite these as constraints, and use them in the field definition directly? It might be worth looking at `PositiveIntAttr` in `test_attribute_definition.py`. You're welcome to translate everything in one go or just migrate this and ping me on Zulip so we can iterate.

tests/filecheck/mlir-conversion/with-mlir/dialects/emitc/emitc_types.mlir
Jimmy2027: yeah mlir-opt orders them, I left the original ordering in the test to bring this up

tests/dialects/test_emitc.py
Jimmy2027: the filecheck tests for the parsing though, not the verification

xdsl/dialects/emitc.py
Jimmy2027: oh nice, sure will do! Thanks a lot for your help :)

tests/dialects/test_emitc.py
superlopuh: but it's the same errors/code, no?

xdsl/dialects/emitc.py
```diff
@@ -70,6 +120,12 @@ def parse_parameters(cls, parser: AttrParser):
             shape, type = parser.parse_ranked_shape()
             if not shape:
                 raise parser.raise_error("EmitC array shape must not be empty")
+            if isinstance(type, EmitC_ArrayType):
+                raise parser.raise_error(
+                    "EmitC array element type cannot be another EmitC_ArrayType."
+                )
+            if not is_supported_emitc_type(type):
+                raise parser.raise_error(f"invalid array element type '{type}'")
```
superlopuh, line 128: These I think belong in verification, not parsing

xdsl/dialects/emitc.py
superlopuh: Actually I think this would also work as `I1 | I8 | I16` and so on

tests/dialects/test_emitc.py
superlopuh, line 1: I think we can delete this file, as these are all checked in lit tests

xdsl/dialects/emitc.py
```diff
@@ -117,8 +129,27 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
     match type_attr:
         case IntegerType():
             return _is_supported_integer_type(type_attr)
-        case _:
+        case IndexType():
+            return True
+        case EmitC_ArrayType():
+            elem_type: Attribute = type_attr.get_element_type()
```
superlopuh, line 135: have to cast unfortunately

xdsl/dialects/emitc.py
```diff
@@ -117,8 +129,27 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
...
+            return not isinstance(
```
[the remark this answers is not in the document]
Jimmy2027: what is the advantage of using `isa` here?

xdsl/dialects/emitc.py
Jimmy2027: ah right

xdsl/dialects/emitc.py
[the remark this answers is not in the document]
superlopuh: when a TypeVar has a default value, checking a value with isa results in a type with that default, as opposed to `Unknown`.

superlopuh (approved):

tests/dialects/test_emitc.py
```diff
@@ -10,3 +10,22 @@ def test_emitc_array_negative_dimension():
...
+
+
+def test_emitc_array_unsupported_element_type():
+    unsupported_type = MemRefType(element_type=i32, shape=[1])
+
+    with pytest.raises(
+        VerifyException,
+        match=f"EmitC array element type '{unsupported_type}' is not a supported EmitC type.",
+    ):
+        EmitC_ArrayType([1], unsupported_type)
```
superlopuh, line 31: Let's delete these?

Jimmy2027: right, sorry forgot about those

xdsl/dialects/emitc.py
```diff
@@ -117,8 +127,27 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
...
+            if isinstance(elem_type, EmitC_ArrayType):
+                return False
+            return is_supported_emitc_type(elem_type)
+        case TupleType():
+            return all(
+                not isinstance(t, EmitC_ArrayType) and is_supported_emitc_type(t)
+                for t in type_attr.types
+            )
+        case _:
+            return False
```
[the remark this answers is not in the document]
superlopuh: If you use VSCode you can navigate to "Run and Debug", select `xdsl-opt`, and press play, it should attach and hit breakpoints for you.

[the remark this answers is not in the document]
Jimmy2027: I realized that the tuple type is not supported yet in xdsl, should I remove that case for now?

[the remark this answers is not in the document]
superlopuh: Yeah let's remove it for now, can add later if you're feeling motivated :)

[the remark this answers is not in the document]
Jimmy2027: let's see :)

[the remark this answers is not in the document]
Jimmy2027: The nested array case is not used anywhere yet since the error is already thrown before calling the `is_supported_emitc_type` function. Trying to figure out how this works in MLIR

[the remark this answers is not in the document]
Jimmy2027: okay I had another look and I actually implemented the wrong function... What I want is [isValidElementType](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/EmitC/IR/EmitCTypes.td#L77) not the `isSupportedEmitCType`. Is it fine if I remove the `is_supported_emitc_type` here and add a `is_valid_element_type` instead? (probably as a method of the ArrayType) 
Sorry for this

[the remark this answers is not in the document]
superlopuh: Sure, glad we caught this

superlopuh:
Please ping when this is ready to re-review

superlopuh (approved):
Oh you already updated it! Great, merging

PR #4452, opened 2025-05-24
misc: swap comparison in disjoint set

jumerckx:
This was a bug in the disjoint set. Now, it should do union by size correctly.


jumerckx:
added the test

PR #4458, opened 2025-05-26
transformations: (pdl_interp) Refactor `PDLInterpRewritePattern`

jumerckx:
This refactor allows passing different interpreter functions, for example for eqsat interpreter functions.


jumerckx:
In the pass, we're calling `PatternRewriteWalker(rewrite_pattern).rewrite_module(op)` which seems good to keep as inlining all the logic with the worklist seems wrong?
Or do you mean to move `match_and_rewrite` inside `PDLInterpFunctions`?

PR #4459, opened 2025-05-26
backend: lowering of vector.broadcast to x86

qaco:
Convert ```vector.broadcast``` operations into x86 assembly instructions.

### DIFF
--- a/tests/filecheck/backend/x86/convert_vector_to_x86.mlir
+++ b/tests/filecheck/backend/x86/convert_vector_to_x86.mlir
@@ -49,3 +49,39 @@
 %rhs = "test.op"(): () -> vector<2xf128>
 %acc = "test.op"(): () -> vector<2xf128>
 %fma = vector.fma %lhs,%rhs,%acc: vector<2xf128>
+
+// -----
+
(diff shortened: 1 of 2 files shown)


qaco:
> I would propose instead lowering it to whatever assembly operations we need, and then to optimise later by leveraging the use-def chain, instead of doing analysis like this in the lowering. The idea is to make each transformation as simple as possible.

But the assembly instruction we need is broadcast, and we need the pointer to generate this instruction. I'm not sure I fully understand what you're proposing. Generate a scalar implementation of a matmul (without vectors) and then try to rebuild vectors from the scalars?

superlopuh:
No I mean: when we lower vector arithmetic, we lower each operation individually, without inspecting the use-def chain. We then rely on target-specific optimisations to recover the intended instruction.

superlopuh:
So here we would remove all ptr.load from the input or from the rewrite pattern, and fuse with x86 load assembly op

superlopuh:
Yeah

qaco:
> Yeah

Ok understood

superlopuh (approved):
Nice

PR #4463, opened 2025-05-27
dialects: (vector) tighten bounds on constraint variable

alexarice:
As discussed on zulip `VectorType.constr(_T) | _T` should only be valid if `_T` cannot have `VectorType` as a base. This PR fixes this for `vector.extract`.

### DIFF
--- a/xdsl/dialects/vector.py
+++ b/xdsl/dialects/vector.py
@@ -14,6 +14,7 @@ from xdsl.dialects.builtin import (
     DenseI64ArrayConstr,
     IndexType,
     IndexTypeConstr,
+    IntegerType,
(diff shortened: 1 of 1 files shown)


alexarice:
Added the same fix to `vector.insert`

superlopuh (approved):
Am I understanding correctly that for now we're just fixing this by inspection, and will add the verification at some point in the future?

alexarice:
> Am I understanding correctly that for now we're just fixing this by inspection, and will add the verification at some point in the future?

Yes, I think the blocker for verification is a way to tell if a constraint has any variables (including those that aren't always set)

PR #4474, opened 2025-05-28
API: migrate xDSL to param_def API for parametrized attributes

superlopuh:
This PR replaces the old API to a new one:

```py
def OldAttr:
    first: ParamDef[StringAttr]
    second: ParamDef[IntAttr, PositiveIntConstraint]

def NewAttr:
    first: StringAttr
    second: IntAttr = param_def(PositiveIntConstraint)
```

Most of the cases can be fixed in your favourite IDE with the following regex match and replace:

```
: ParameterDef\[(.*)\]
```

to

```
: $1
```

The constrained cases may be a little trickier.


xdsl/dialects/experimental/fir.py
```diff
@@ -226,15 +226,15 @@ class SequenceType(ParametrizedAttribute, TypeAttribute):
...
-    shape: ParameterDef[ArrayAttr[IntegerAttr | DeferredAttr | NoneType]]
-    type: ParameterDef[IntegerType | AnyFloat | ReferenceType]
-    type2: ParameterDef[IntegerType | AnyFloat | ReferenceType | NoneType]
+    shape: ArrayAttr[IntegerAttr | DeferredAttr | NoneType] = param_def()
+    type: IntegerType | AnyFloat | ReferenceType = param_def()
+    type2: IntegerType | AnyFloat | ReferenceType | NoneType = param_def()
 
     def __init__(
         self,
-        type1: ParameterDef[IntegerType | AnyFloat | ReferenceType],
```
superlopuh, line 235: 🙃

docs/marimo/irdl.py
```diff
@@ -459,14 +459,14 @@ class MyIntegerType(ParametrizedAttribute):
 
         # Only parameter of the type, with an `EqAttrConstraint` constraint.
         # Note the use of the attribute constraint coercion.
```
alexarice: I don't see the constraint coercion thing, maybe remove the comment?

docs/marimo/irdl.py
```diff
@@ -459,14 +459,14 @@ class MyIntegerType(ParametrizedAttribute):
...
-        width: ParameterDef[IntAttr]
+        width: IntAttr = param_def(constraint=EqAttrConstraint(IntAttr(32)))
 
 
     my_i32 = MyIntegerType([IntAttr(32)])
 
     # This will trigger an exception, since the attribute only expect a single attribute
     try:
-        MyIntegerType([IntAttr(32), IntAttr(64)])
+        MyIntegerType.new([IntAttr(32), IntAttr(64)])
```
alexarice: Lets use the new initialise (when merged)?

tests/irdl/test_attribute_definition.py
```diff
@@ -596,7 +592,7 @@ def test_nested_generic_constraint_fail():
 class NestedParamConstrAttr(ParametrizedAttribute):
     name = "test.nested_param_constr"
 
-    param: ParameterDef[NestedParamWrapperAttr[Annotated[IntData, PositiveIntConstr()]]]
+    param: NestedParamWrapperAttr[Annotated[IntData, PositiveIntConstr()]]
```
alexarice, line 574: Do we want to leave the annotated here? I guess we're checking that this still works at least for the moment?

tests/irdl/test_attribute_definition.py
```diff
@@ -676,45 +669,6 @@ def test_informative_constraint():
...
-    """
-    with pytest.raises(
-        PyRDLTypeError,
-        match=(
-            "Generic `Data` type 'test.missing_genericdata' cannot be converted to an "
-            "attribute constraint. Consider making it inherit from `GenericData` "
-            "instead of `Data`."
-        ),
-    ):
-        irdl_attr_definition(MissingGenericDataDataWrapper)
```
alexarice, line 689: why delete this?

tests/test_pyrdl.py
```diff
@@ -331,8 +330,47 @@ def test_constraint_vars_fail_underlying_constraint():
...
+    with pytest.raises(
+        PyRDLTypeError,
+        match=(
+            "Generic `Data` type 'test.missing_genericdata' cannot be converted to an "
+            "attribute constraint. Consider making it inherit from `GenericData` "
+            "instead of `Data`."
+        ),
+    ):
+        irdl_to_attr_constraint(MissingGenericDataData[int])
+
```
alexarice, line 378: ah it just moved.

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -77,7 +76,7 @@ class ExchangeDeclarationAttr(ParametrizedAttribute):
 
     name = "csl_stencil.exchange"
 
-    neighbor_param: ParameterDef[DenseArrayBase[I64]]
+    neighbor_param: DenseArrayBase[I64]
```
alexarice, line 78: Is something like this not where we would want to use:
```
neighbor_param = param_def(DenseArrayBase.constr(element_type = i64))
```
instead of annotated. If the idea is to merge this first then slowly work on deprecating Annotated then I'm ok with this.

xdsl/dialects/csl/csl_stencil.py
```diff
@@ -166,8 +165,8 @@ def __init__(
 @irdl_attr_definition
 class CoeffAttr(ParametrizedAttribute):
     name = "csl_stencil.coeff"
-    offset: ParameterDef[stencil.IndexAttr]
-    coeff: ParameterDef[FloatAttr[AnyFloat]]
+    offset: stencil.IndexAttr
+    coeff: FloatAttr[AnyFloat]
```
alexarice:
```suggestion
    coeff: FloatAttr
```
nit

xdsl/dialects/experimental/fir.py
alexarice: :thinking:

xdsl/irdl/attributes.py
```diff
@@ -171,15 +170,53 @@ def from_pyrdl(
...
+            parameters[field_name] = constraint
+
         for field_name, value in clsdict.items():
             if field_name == "name":
+                # Ignore name field
                 continue
             if isinstance(
                 value, FunctionType | PropertyType | classmethod | staticmethod
             ):
+                # Ignore functions
```
alexarice: Should we also ignore `ClassVar`. Feels like it may be useful to have `ClassVar` now that we can write `param_def(VarConstraint...)`

xdsl/irdl/attributes.py
```diff
@@ -171,15 +170,53 @@ def from_pyrdl(
...
+        for field_name, field_type in get_type_hints(
+            pyrdl_def, include_extras=True
+        ).items():
+            if field_name == "name" or field_name == "parameters":
+                continue
+
+            try:
+                constraint = irdl_to_attr_constraint(field_type, allow_type_var=True)
+            except TypeError:
+                raise PyRDLAttrDefinitionError(
```
alexarice, line 217: This throws away the original error which seems non-ideal?

xdsl/irdl/attributes.py
```diff
@@ -171,15 +170,53 @@ def from_pyrdl(
...
+                    f"Invalid field type {field_type} for field name {field_name}."
+                )
+
+            parameters[field_name] = constraint
+
         for field_name, value in clsdict.items():
```
alexarice: Can we fuse these two loops? Iterate over the type hints and then use `clsdict` as an actual dictionary?

alexarice: Or of course the other way round, using `get_type_hints` as a dictionary.

tests/irdl/test_attribute_definition.py
superlopuh: For now, yes, I think we can migrate this when we phase out the Annotated support.

xdsl/dialects/csl/csl_stencil.py
superlopuh: Yep exactly, especially the I64 things I'd like to play around with to make better.

xdsl/irdl/attributes.py
superlopuh: I was going to do that as a second step, but you're right that it would be nice.

xdsl/irdl/attributes.py
superlopuh: I'll fix it

- [x] fix it

xdsl/irdl/attributes.py
superlopuh: I'll take a look

- [x] take a look

xdsl/irdl/attributes.py
alexarice: Don't mind this being done as a second step

xdsl/dialects/csl/csl_stencil.py
superlopuh: #4695

docs/marimo/irdl.py
superlopuh: Removed

xdsl/irdl/attributes.py
superlopuh: It seems to me like we need it for the time being as the set of type annotations and values is disjoint, and we need to detect that somehow.

tests/irdl/test_attribute_definition.py
```diff
@@ -850,6 +803,27 @@ def test_deprecated_tuple_init():
         )
 
 
+@irdl_attr_definition
+class ParamAttrDefAttr2(ParametrizedAttribute):
+    name = "test.param_attr_def_attr"
+
+    arg1: Attribute = param_def(base(IntAttr))
+    arg2: BoolData = param_def()
```
superlopuh: Pyright made me put the `param_def()` here saying that fields without default values can't follow fields that have one.

alexarice: Thanks, I'll take a quick look

xdsl/irdl/attributes.py
alexarice: That seems reasonable, we could potentially (non-strictly) zip the two dictionaries together somehow but it's not clear this would actually be any cleaner. If you go with `param_def` being required then my understanding is that every type annotation should also be a value?

xdsl/irdl/attributes.py
```diff
@@ -91,9 +92,27 @@ def constr(constr: AttrConstraint) -> AttrConstraint:
...
+    ):
+        self.param = param
+
+
+def param_def(
+    constraint: GenericAttrConstraint[AttributeInvT] | None = None,
+    *,
+    default: None = None,
+    resolver: None = None,
+    init: Literal[False] = False,
```
alexarice: What do all the extra keyword arguments do?

docs/marimo/irdl.py
alexarice: Should be doable now?

xdsl/irdl/attributes.py
```diff
@@ -172,15 +171,54 @@ def from_pyrdl(
...
+                    )
+
+                try:
+                    if value.param is not None:
+                        constraint = irdl_to_attr_constraint(
+                            value.param, allow_type_var=True
+                        )
+                        parameters[field_name] &= constraint
+                except TypeError:
+                    raise PyRDLAttrDefinitionError(
```
alexarice, line 238: `from e` thing should also be here?

xdsl/irdl/attributes.py
```diff
@@ -91,9 +92,27 @@ def constr(constr: AttrConstraint) -> AttrConstraint:
...
+        self,
+        param: GenericAttrConstraint[Attribute] | None,
+    ):
+        self.param = param
+
+
+def param_def(
+    constraint: GenericAttrConstraint[AttributeInvT] | None = None,
+    *,
+    default: None = None,
```
alexarice: I think this argument might be what's causing the pyright default argument error. I think pyright sees this as the field having a default argument which is `None`. In the `dataclass_transform` documentation they have:
```python
default: Optional[Any] = ...
```

xdsl/irdl/attributes.py
superlopuh: Not sure what you mean by the very last bit.

alexarice: My understand is that with your current implementation then you should never have `param: Attribute` but should always have `param: Attribute = param_def()` so now the statement "set of type annotations and values is disjoint" isn't true as the second should be a subset of the first?

xdsl/irdl/attributes.py
superlopuh: They are part of the dataclass mechanism, this type signature lets them be picked up to do things like inits and eq and stuff. Which makes me think that I should try moving the types into the param_def again... I'll give that a quick go this morning.

superlopuh: - [x] give it a go

alexarice: Right but these aren't the right values?

alexarice: We surely don't want to set `init = False` on everything?

xdsl/irdl/attributes.py
superlopuh: Done

xdsl/irdl/attributes.py
superlopuh: I tried it without and it didn't work.

superlopuh: Thanks for helping with this

xdsl/irdl/attributes.py
superlopuh: You're right, it's now always True

docs/marimo/irdl.py
alexarice: Please address

docs/marimo/irdl.py
```diff
@@ -475,6 +474,15 @@ class MyIntegerType(ParametrizedAttribute):
         MyIntegerType([StringAttr("ga")])
```
alexarice: These also all look like they need updating (perhaps they need updating on main I'm not sure). Is this PR rebased past the initialiser stuff?

xdsl/irdl/attributes.py
```diff
@@ -91,9 +92,25 @@ def constr(constr: AttrConstraint) -> AttrConstraint:
...
+        param: GenericAttrConstraint[Attribute] | None,
+    ):
+        self.param = param
+
+
+def param_def(
+    constraint: GenericAttrConstraint[AttributeInvT] | None = None,
+    *,
+    init: Literal[True] = True,
+) -> AttributeInvT:
```
alexarice, line 114: Is there any point having the return type and cast if the type annotation is compulsory? Does having this give some extra protection against dodgy typing?

xdsl/irdl/attributes.py
alexarice: This is now the other way round, if everything requires a type annotation then we should be able to loop over `get_type_hints` and use `cls_dict` as a dictionary instead of having two loops.

docs/marimo/irdl.py
superlopuh: Oh I think we might ignore warnings in the marimo tests, I'll update

xdsl/irdl/attributes.py
superlopuh: Yeah I think so, pyright complains if you give an incompatible type

xdsl/irdl/attributes.py
superlopuh: We can't get rid of the second loop quite yet because we still have to check for Annotated, so I'd rather leave it as-is for now

xdsl/irdl/attributes.py
alexarice: Cool, lets leave this for now then.

xdsl/irdl/attributes.py
alexarice: I don't understand how Annotated effects this but happy to leave this and iterate later once Annotated is gone.

xdsl/irdl/attributes.py
```diff
@@ -91,9 +92,25 @@ def constr(constr: AttrConstraint) -> AttrConstraint:
 # |_|   \__,_|_|  \__,_|_| |_| |_/_/   \_\__|\__|_|
 #
 
-_A = TypeVar("_A", bound=Attribute)
 
-ParameterDef = Annotated[_A, IRDLAnnotations.ParamDefAnnot]
+# Field definition classes for `@irdl_param_attr_definition`
```
math-fehr: Make this a documentation instead of a comment?

xdsl/irdl/attributes.py
```diff
@@ -172,15 +169,54 @@ def from_pyrdl(
...
+                        )
+                        parameters[field_name] &= constraint
+                except TypeError as e:
+                    raise PyRDLAttrDefinitionError(
+                        f"Invalid constraint {value.param} for field name {field_name}."
+                    ) from e
+
+                continue
+
             # Constraint variables are allowed
```
math-fehr, line 244: Should that still exist? Is there a case where we still have a value that is an `Annotated`?

xdsl/irdl/attributes.py
```diff
@@ -172,15 +169,54 @@ def from_pyrdl(
             if key not in _PARAMETRIZED_ATTRIBUTE_DICT_KEYS
         }
 
-        # Check that all fields of the attribute definition are either already
-        # in ParametrizedAttribute, or are class functions or methods.
+        # The resulting parameters
+        parameters: dict[str, AttrConstraint] = {}
+
+        for field_name, field_type in get_type_hints(
```
math-fehr: Can we move this logic in a separate function?
Just something that returns the parameters value.

math-fehr (commented):
I had some nits comments, but otherwise that looks really good to me!

xdsl/irdl/attributes.py
alexarice: This hasn't got rid of all the Annotated, think the idea is to deprecate it in stages after this.

superlopuh: I don't actually know, but it feels like something that we can get rid of when we remove the Annotated mechanism?

xdsl/irdl/attributes.py
superlopuh: Hmm, why would you prefer it in a helper? What other scenarios might we call this in? Feels more readable inline to me...

math-fehr: I would prefer it in a helper to cleary distinguish between the multiple steps of the function.
Because the next two loops noly compute `parameters` right? I think it makes the entire function more readable, since you don't have to understand the two loops to understand the entire function

xdsl/irdl/attributes.py
math-fehr: Oh yeah my bad, I get it!

xdsl/irdl/attributes.py
superlopuh: I'd rather keep it in one function but I see your point, I'll try a refactor

alexarice: I think I would also prefer that the two loops that generate the parameters dict is a helper function that returns a dictionary.

superlopuh: OK so I think this is more readable now. I'd really like to avoid helper functions, I think they just make things harder to read by making the reader go to definition and back, and now everything is in one place that is close to fitting on one screen. Once we merge this, we can get rid of the Annotated case, which will further shrink this code, let's iterate then?

xdsl/irdl/attributes.py
superlopuh: Done

PR #4479, opened 2025-05-29
dialects: Add `smt.bv` type

math-fehr:
Stacked PRs:
 * #4482
 * #4481
 * #4480
 * __->__#4479

--- --- ---

### dialects: Add `smt.bv` type

It represent the SMT bitvector type

### DIFF
--- a/tests/dialects/test_smt.py
+++ b/tests/dialects/test_smt.py
@@ -5,6 +5,7 @@ from xdsl.dialects.smt import (
     AndOp,
     ApplyFuncOp,
     AssertOp,
+    BitVectorType,
     BoolType,
     ConstantBoolOp,
     DeclareFunOp,
@@ -23,6 +24,7 @@ from xdsl.dialects.smt import (
     YieldOp,
 )
 from xdsl.ir import Block, Region
+from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.test_value import create_ssa_value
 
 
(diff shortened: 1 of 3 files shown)


alexarice (approved):
Could alternatively have `NonFuncSMTConstr` that's defined once but this is fine. Should `smt.constant` be able to return a bitvector type too? Would be good to have a test of this if it should.

math-fehr:
> Could alternatively have `NonFuncSMTConstr` that's defined once but this is fine.

Yeah, I wasn't sure what we should do there. On one hand I understand why defining a new value for the constraint make sense. But on the other hand, I don't know if this would reduce the number of things to keep in mind, if that makes sense. Whatever you prefer.

> Should `smt.constant` be able to return a bitvector type too? Would be good to have a test of this if it should.

Nope, I still think that `smt.constant` is badly named in MLIR, because that's only for boolean constants. `smt.bv.constant` will handle bitvector constants.

math-fehr:
Also, I would argue that renaming the function `to_constr` instead of `irdl_to_attr_constraint` would remove that problem.

alexarice:
> I would say that I agree, but in general I'm not sure how far we should go with this. Like, should we do the same for every constraint that accepts a `AttrConstraint`? In any case, I'll just add your `Constr` for this PR.

I'd say anything user facing could take the irdl constraints

math-fehr:
That does make sense to me, we can work on changing this once we finish the constraints overhaul

PR #4480, opened 2025-05-29
core: Fix printing of non-builtin TypedAttribute

math-fehr:
core: Fix printing of non-builtin TypedAttribute

Turns out, we never actually used `TypedAttribute` on non-builtin
attributes, so this had to be fixed.


alexarice (commented):
I'm not sure I understand this, is printing "$print_without_type : type" wrong?

PR #4481, opened 2025-05-29
core: Add support for qualified attributes in assembly format

math-fehr:
Stacked PRs:
 * #4482
 * __->__#4481
 * #4480

--- --- ---

### core: Add support for qualified attributes in assembly format

The `qualified` directive has to be used when an attribute or property is
known to be a base attribute, and where we want to parse the entire attribute,
not just the parameters.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -3427,3 +3427,50 @@ def test_int_attr_verify_errors(program: str, error: str):
     with pytest.raises(VerifyException, match=error):
         op = parser.parse_operation()
         op.verify()
+
+
+@irdl_attr_definition
+class MyAttr(ParametrizedAttribute):
+    name = "test.my_attr"
+
+    param: ParameterDef[StringAttr]
+
+
+@irdl_op_definition
+class NonQualifiedAttrOp(IRDLOperation):
+    name = "test.non_qualified_attr"
+
+    attr = prop_def(MyAttr)
+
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1090,29 +1090,33 @@ class AttributeVariable(FormatDirective):
...
-            ParametrizedAttribute,
-        ):
-            attr = unique_base.new(unique_base.parse_parameters(parser))
-        elif issubclass(unique_base, Data):
-            attr = unique_base.new(  # pyright: ignore[reportUnknownVariableType]
-                unique_base.parse_parameter(parser)
-            )
         else:
-            raise ValueError("Attributes must be Data or ParameterizedAttribute.")
+            assert unique_base is not None
```
alexarice, line 1105: Why the assert here?

alexarice (commented):
Are you enforcing that `unique_base` is not `None` if `qualified` is true? If so why not just make things qualified by explicitly setting `unique_base` to `None`?

The MLIR version of this also seems to support `type` directives, and it doesn't feel like this infrastructure generalises nicely to that?

math-fehr:
Yes, I only implemented it for attributes (as I don't have the time to generalize it right now), but my guess is that I would extend it to types the same way?

For the unique_base, I somehow thought it was useful elsewhere. I can change this to remove is_qualified and just do that change if that's enough!

superlopuh (approved):

PR #4488, opened 2025-06-02
misc: enable PYI rule for ruff

superlopuh:
So PYI stands for the type stub extension file format, but it seems to contain a number of rules for the rest of the code. Enabling it improves the code quality IMO.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,10 +122,11 @@ exclude = [".venv", "docs/marimo"]
 [tool.ruff.lint]
 select = ["E", "F", "W", "I", "UP", "PT", "TID251", "INP", "PYI"]
 ignore = [
-    "E741",  # https://beta.ruff.rs/docs/rules/ambiguous-variable-name/
-    "PT006", # https://beta.ruff.rs/docs/rules/pytest-parametrize-names-wrong-type/
-    "PT011", # https://beta.ruff.rs/docs/rules/pytest-raises-too-broad/
-    "PT012", # https://beta.ruff.rs/docs/rules/pytest-raises-with-multiple-statements/
+    "E741",   # https://beta.ruff.rs/docs/rules/ambiguous-variable-name/
+    "PT006",  # https://beta.ruff.rs/docs/rules/pytest-parametrize-names-wrong-type/
+    "PT011",  # https://beta.ruff.rs/docs/rules/pytest-raises-too-broad/
+    "PT012",  # https://beta.ruff.rs/docs/rules/pytest-raises-with-multiple-statements/
+    "PYI041", # https://docs.astral.sh/ruff/rules/redundant-numeric-union
 ]
 
 [tool.ruff.lint.pycodestyle]
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -862,7 +862,9 @@ def test_parse_optional_int_error(text: str, allow_boolean: bool, allow_negative
         ("0.2", True, 0.2),
     ],
 )
-def test_parse_number(text: str, allow_boolean: bool, expected_value: float | None):
+def test_parse_number(
+    text: str, allow_boolean: bool, expected_value: int | float | None
+):
     parser = Parser(Context(), text)
     assert parser.parse_optional_number(allow_boolean=allow_boolean) == expected_value
 


pyproject.toml
```diff
@@ -154,19 +154,22 @@ max-line-length = 130
...
+"docs/irdl.ipynb" = ["ALL"]
+"docs/mlir_interoperation.ipynb" = ["E402"]
+"docs/Toy/Toy_Ch1.ipynb" = ["E402"]
+"docs/Toy/Toy_Ch2.ipynb" = ["E402"]
+"docs/Toy/Toy_Ch3.ipynb" = ["E402"]
+"docs/tutorial.ipynb" = ["ALL"]
+"docs/xdsl-introduction.ipynb" = ["ALL"]
 "scripts/**" = ["INP"]
+"tests/*" = ["E501", "INP"]
+"tests/test_declarative_assembly_format.py" = ["F811"]
```
superlopuh, line 170: I just sorted these lines in ascending format

pyproject.toml
```diff
@@ -154,19 +154,22 @@ max-line-length = 130
...
+"tests/test_frontend_type_conversion.py" = ["PYI042"]
+"xdsl/frontend/pyast/dialects/builtin.py" = ["PYI042"]
+"xdsl/utils/immutable_list.py" = ["PYI"]
```
superlopuh: these are new

tests/test_parser.py
```diff
@@ -862,9 +862,7 @@ def test_parse_optional_int_error(text: str, allow_boolean: bool, allow_negative
         ("0.2", True, 0.2),
     ],
 )
-def test_parse_number(
-    text: str, allow_boolean: bool, expected_value: int | float | None
-):
+def test_parse_number(text: str, allow_boolean: bool, expected_value: float | None):
```
superlopuh: this removes the `int | float` because int is a subtype of float 🤷

xdsl/builder.py
```diff
@@ -314,9 +314,10 @@ def __enter__(self) -> tuple[BlockArgument, ...]:
 
     def __exit__(
         self,
-        __exc_type: type[BaseException] | None,
-        __exc_value: BaseException | None,
-        __traceback: TracebackType | None,
+        exc_type: type[BaseException] | None,
```
superlopuh, line 317: it doesn't like double underscores in function names

xdsl/dialects/bufferization.py
```diff
@@ -1,4 +1,5 @@
-from collections.abc import Sequence, Set
+from collections.abc import Sequence
+from collections.abc import Set as AbstractSet
```
superlopuh, line 2: this is an interesting one that enforces the naming of `AbstractSet` to highlight that it's not `set`

xdsl/ir/core.py
```diff
@@ -1141,19 +1141,19 @@ def verify_(self) -> None:
     _OperationType = TypeVar("_OperationType", bound="Operation")
 
     @classmethod
-    def parse(cls: type[_OperationType], parser: Parser) -> _OperationType:
+    def parse(cls, parser: Parser) -> Self:
```
superlopuh, line 1144: this was an automated change

pyproject.toml
alexarice: are these disabling or enabling checks?

tests/test_parser.py
alexarice: Is there a way to turn this off, I feel even if they are technically the same `int | float` specifies the intent a lot better

pyproject.toml
alexarice: Looks like they're disabling, It would be nice to have a comment to say why

alexarice (approved):
Big fan of these changes overall.

pyproject.toml
superlopuh: Disabling, I'll add comments

tests/test_parser.py
superlopuh: I'll disable it

superlopuh: Actually I like their argument in the doc of the linting rule: https://docs.astral.sh/ruff/rules/redundant-numeric-union/ I think it makes sense to respect the decision to unify int and float in the type system in Python, even if I disagree with it.

pyproject.toml
superlopuh: added

tests/test_parser.py
alexarice: I don't even see an argument in that doc, it just says its redundant?

superlopuh: To someone new to Python typing this is unintuitive, and it makes sense for the linter to get rid of this construct because it's misleading. It suggests that without the `int |` an int could be passed there.

alexarice: I agree that it's highly unintuitive, but feel the intuition is worse the other way: why would a beginner expect a function with a `float` signature to be able to expect an `int`. At least this way round `int | float` suggests the correct meaning to me. (Maybe I'm missing something)

math-fehr:
I hate the `int | float` thing, but I guess that's a Python problem

tests/test_parser.py
alexarice: In fact, I would expect that the function with the `float` signature wouldn't necessarily work properly with integers, even though I know it can technically expect them.

alexarice: If this doesn't make sense then go with the lint and I'll do my crying internally

superlopuh: I see your point. Ultimately I don't think it matters enough to make this change automatically, I'll revert and ignore this rule.

PR #4491, opened 2025-06-04
dialects: (eqsat) Add eqsat egraph and yield operation

jumerckx:


superlopuh (commented):
Can you please add a filecheck test?

jumerckx:
I added an `eqsat.egraph` op to the existing filecheck test.

PR #4493, opened 2025-06-04
transformations: (eqsat) Add apply-eqsat-pdl-interp pass

jumerckx:
~Currently marked as draft as this PR is on top of `jumerckx/eqsat_pdl_interp`.~


superlopuh (approved):

jumerckx:
This was approved before so I assume I can merge?

superlopuh:
Yep! If you feel motivated it would be nice to add a test with the pdl_interp file separate, but this is not required, we can add one later.

jumerckx:
> it would be nice to add a test with the pdl_interp file separate

That was a good call, there was a bug that is now fixed. I've added a test.

PR #4494, opened 2025-06-04
transformations: (eqsat) Delete `pdl`-based eqsat in favor of `pdl_interp`

jumerckx:
Currently marked as draft as this pr builds on `jumerckx/eqsat_pass`.


superlopuh (commented):
Why change the name?

superlopuh:
I would actually propose not doing this, and instead changing the implementation of `apply-pdl` to silently lower `pdl` first then applying interpretation. In the case where the pdl functions are provided in a separate file, this lowering will be in-memory and won't affect the printed IR, which seems like a win, and it should make the whole interface more ergonomic, while allowing users to do this manually and call `apply-eqsat-pdl-interp` if they feel like it.

jumerckx:
Ok I think I see what you're getting at. To be sure: you propose to have `-apply-eqsat-pdl` depend on mlir-proper's `convert-pdl-to-pdl-interp`?

Is there code in xDSL that calls into MLIR that I can base this off of?

superlopuh:
Oh I forgot that we were relying on this... I think it's still worth doing, the code is in `xdsl/transforms/mlir_opt.py`, and the tests would need to move to the `with-mlir` folder. Eventually we'll probably want to duplicate this lowering somehow.

jumerckx:
Implemented `apply-eqsat-pdl` as discussed.

PR #4495, opened 2025-06-04
bench: switch to using bytesight implementation from github

EdmundGoodman:
Switch to using ByteSight implementation from GitHub in `pyproject.toml`, rather than duplicating code in xDSL repo. Later, this will switch to pointing at PyPI directly, but for now whilst the implementation is still a bit transient PyPI caching means it takes a few hours and a new patch release to test a single change.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -302,7 +302,7 @@ wheels = [
 [[package]]
 name = "bytesight"
 version = "0.1.2"
-source = { git = "https://github.com/EdmundGoodman/bytesight.git#ee773b979ba70d707a4d6532244de21496399c08" }
+source = { git = "https://github.com/EdmundGoodman/bytesight.git?rev=ee773b979ba70d707a4d6532244de21496399c08#ee773b979ba70d707a4d6532244de21496399c08" }
 
 [[package]]
 name = "certifi"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -50,6 +50,7 @@ bench = [
     "flameprof>=0.4",
     "orjson>=3.10.15",
     "pyinstrument>=5.0.1",
+    "bytesight",
```
superlopuh, line 53: let's pin git commit?

pyproject.toml
```diff
@@ -74,6 +75,7 @@ zip-safe = false
 
 [tool.uv.sources]
 lzstring2 = { git = "https://github.com/superlopuh/lz-string.git" }
+bytesight = { git = "https://github.com/EdmundGoodman/bytesight.git" }
```
EdmundGoodman: Per https://docs.astral.sh/uv/concepts/projects/dependencies/#git in response to https://github.com/xdslproject/xdsl/pull/4495/files#r2126658077

```suggestion
bytesight = { git = "https://github.com/EdmundGoodman/bytesight.git" , rev = "ee773b979ba70d707a4d6532244de21496399c08"}
```

pyproject.toml
EdmundGoodman: Done -- is it worth (in a different PR probably) doing the same thing to lzstring2?

superlopuh: Nah that's my repo and the aim is to remove it instead of ever updating it.

superlopuh (approved):

PR #4496, opened 2025-06-04
core: allow multiple levels of Generics in operation definitions

superlopuh:
The current implementation of checking for generic mappings only looks at one level of superclasses, and not recursively. This leads to a silent failure when doing multiple levels of subclassing like in the riscv dialect, where we progressively specialise the operation. The default constraint of the bound of the TypeVar is used instead of the specialised type variable, for which we don't have any tests so we didn't see it before. This PR adds a test to make sure that the constraints are more specific, as well as more tests for getting the type var mapping. As a bonus we can now handle multiple generic parents.

### DIFF
--- a/xdsl/utils/hints.py
+++ b/xdsl/utils/hints.py
@@ -149,48 +149,53 @@ def get_type_var_from_generic_class(cls: type[Any]) -> tuple[TypeVar, ...]:
 
 def get_type_var_mapping(
     cls: type[Any],
-) -> tuple[type[Any], dict[TypeVar, Any]]:
+) -> tuple[tuple[TypeVar, ...], dict[TypeVar, Any]]:
     """
-    Given a class that specializes a generic class, return the generic class and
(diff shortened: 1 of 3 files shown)


xdsl/utils/hints.py
```diff
@@ -149,48 +149,53 @@ def get_type_var_from_generic_class(cls: type[Any]) -> tuple[TypeVar, ...]:
 
 def get_type_var_mapping(
     cls: type[Any],
-) -> tuple[type[Any], dict[TypeVar, Any]]:
+) -> tuple[tuple[TypeVar, ...], dict[TypeVar, Any]]:
     """
-    Given a class that specializes a generic class, return the generic class and
-    the mapping from the generic class type variables to the specialized arguments.
+    Given a Generic class, return the TypeVars used to specialize it, and the mapping
+    from the generic class type variables to the specialized arguments.
```
math-fehr: Can you add the fact that this is also done for the parent classes?

PR #4497, opened 2025-06-04
[WIP] core: Add slots operations

EdmundGoodman:
This PR adds slotted variables to operation dataclasses. This gives some fair performance wins, particularly for specialised workloads.

There is something quite odd going on with the way the `Operation` class is constructed -- setting `slots=True` seems to clobber the dataclass init function getting called, so we manually disable it then implement its functionality where it is used in what used to be overloading `__init__` but is now called `populate`.


EdmundGoodman:
Passes 3.13 CI, but not 3.10 CI -- need to investigate further

superlopuh:
I think that we can probably drop 3.10 support, and have a policy of three most recent major versions of Python...

alexarice:
Perhaps the change between versions is caused by:

> Changed in version 3.11: If a field name is already included in the __slots__ of a base class, it will not be included in the generated __slots__ to prevent [overriding them](https://docs.python.org/3/reference/datamodel.html#datamodel-note-slots). Therefore, do not use __slots__ to retrieve the field names of a dataclass. Use [fields()](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) instead. To be able to determine inherited slots, base class __slots__ may be any iterable, but not an iterator.

?

math-fehr:
Do we really need to drop 3.10?
It still has at least a year of support, I would rather try to find a way to still enable it somehow.

superlopuh:
Let's first find a way to get this to work without dropping it, but I somehow doubt there are any users that require 3.10

superlopuh:
Let's not do this for now, and revisit when we migrate to 3.11?

superlopuh:
#4705

PR #4499, opened 2025-06-05
core: use attribute definition in irdl_to_attr_constraint

superlopuh:
This avoids us re-parsing it every time.


xdsl/transforms/experimental/convert_stencil_to_ll_mlir.py
```diff
@@ -658,7 +659,7 @@ def return_target_analysis(module: builtin.ModuleOp):
 
 
 class StencilTypeConversion(TypeConversionPattern):
-    @attr_type_rewrite_pattern
+    @attr_constr_rewrite_pattern(StencilTypeConstr)
```
superlopuh, line 662: We have to use the constraint explicitly as the one based on the type fails due to this being an abstract attribute superclass which doesn't have an IRDL definition

math-fehr (approved):

PR #4500, opened 2025-06-05
bench: add readme file to benchmarks directory

EdmundGoodman:
Add a readme file to describe the use of the benchmarking infrastructure. Could also be worth adding examples of using asv, and a workflow for performance diff of a patch?


superlopuh (approved):
Fantastic, ty

PR #4502, opened 2025-06-05
[WIP] core: Dialect resources and `DenseResourceAttr `

mamanain:
`torch-mlir` uses dialect resources to store network weights so it seems like a good idea to support this mechanism in xdsl.

Dialect resources are stored in file metadata - `{-# ... #-}` section - in the code data is stored using a dialect interface.

The mechanism is quite general in MLIR, in this case the resources are only referenced through `DenseResourceAttr `.


superlopuh:
Is all of this merged now?

mamanain:
Yep

PR #4503, opened 2025-06-05
core: Dialect interfaces,`OpAsmDialectInterface `

mamanain:
Defining dialect interfaces, `OpAsmDialectInterface ` for dialect resource management.

Implemented this just like op traits are implemented. So that it's easier to refactor in the future.

### DIFF
--- a/xdsl/dialect_interfaces.py
+++ b/xdsl/dialect_interfaces.py
@@ -14,24 +14,33 @@ class DialectInterface:
 
 
 class OpAsmDialectInterface(DialectInterface):
-    _blob_storage: dict[str, str] = {}
+    """
+    The OpAsm interface generally deals with making printed ir look less verbose.
+    It implements two main functionalities - Aliases and Resources.
 
-    ## Aliases - yet to be implemented
+    ## Aliases
+    If you have some commonly used types and attributes `AsmPrinter` can generate aliases for them through this interface.
+    For example, `!my_dialect.type<a=3,b=4,c=5,d=tuple,e=another_type>` and `#my_dialect.attr<a=3>`
+    can be aliased to `!my_dialect_type` and `#my_dialect_attr`, simplifying further references.
+
+    **This is not yet implemented in xdsl**
 
     ## Resources
+    Keep a dialect specific storage of blobs in the Dialect object.
 
-    # Keep a dialect specific storage of blobs in the Dialect object.
+    This functionality is usefull when we want to reference some objects
+    while keeping them outside of ir.
+    For example if we have a big dense array we might decide to store it in
+    dialect's storage and use a key to reference it inside the ir.
+    Or if we want some data to be shared across attributes we might wanna
+    store it in the storage and use the same key to reference it
+    in different places in the ir.
 
-    # This functionality is usefull when we want to reference some objects
-    # while keeping them outside of ir.
-    # For example if we have a big dense array we might decide to store it in
-    # dialect's storage and use a key to reference it inside the ir.
-    # Or if we want some data to be shared across attributes we might wanna
-    # store it in the storage and use the same key to reference it
-    # in different places in the ir.
+    This can help keep the ir clean and also give some performance improvements
+    compared to keeping these resources tied to ir objects.
+    """
 
-    # This can help keep the ir clean and also give some performance improvements
-    # compared to keeping these resources tied to ir objects.
+    _blob_storage: dict[str, str] = {}
 
     def declare_resource(self, key: str) -> str:
         """
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
     from xdsl.printer import Printer
 
 OpT = TypeVar("OpT", bound="Operation")
-DialectInterfT = TypeVar("DialectInterfT", bound=DialectInterface)
+DialectInterfaceT = TypeVar("DialectInterfaceT", bound=DialectInterface)
 
 
 @dataclass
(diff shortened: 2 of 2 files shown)


tests/test_dialect_interfaces.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from xdsl.dialect_interfaces import OpAsmDialectInterface
+
+
+def test_op_asm_interface():
```
superlopuh: Can you please also add a test for the general concept of a dialect interface? Probably with a test dialect interface defined in this file

xdsl/dialect_interfaces.py
```diff
@@ -0,0 +1,36 @@
+class DialectInterface:
```
superlopuh, line 1: Please be generous with doc strings here

xdsl/dialects/builtin.py
```diff
@@ -2624,6 +2625,10 @@ def print_without_type(self, printer: Printer):
 DenseIntElementsAttr: TypeAlias = DenseIntOrFPElementsAttr[IndexType | IntegerType]
 
 
+class BuiltinOpAsmDialectInterface(OpAsmDialectInterface):
```
superlopuh: Why subclass ?

xdsl/dialect_interfaces.py
```diff
@@ -0,0 +1,36 @@
...
+    pass
+
+
+class OpAsmDialectInterface(DialectInterface):
+    blob_storage: dict[str, str] = {}
```
superlopuh: Is this supposed to be accessible outside of this class? If yes, can you please add a docstring? I would probably lean towards making it private

xdsl/ir/core.py
```diff
@@ -79,6 +84,15 @@ def split_name(name: str) -> tuple[str, str]:
         except ValueError as e:
             raise ValueError(f"Invalid operation or attribute name {name}.") from e
 
+    def get_interface(self, inter: type[DialectInterfT]) -> DialectInterfT | None:
```
superlopuh: Also deserves doc strings

xdsl/dialects/builtin.py
mamanain: In MLIR this interface differs a little for Test and Builtin dialect in the alias functionality.

I thought it would be nice to set the tone of how the interface classes are usually structured - a base interface in `dialect_interfaces.py` and dialects inherit from them in the dialect definition files.

superlopuh: I see, let's subclass as we add the customisation?

xdsl/dialect_interfaces.py
```diff
@@ -0,0 +1,70 @@
...
+    Dialects will implement this interface and the inlining transformation will query them through the base interface.
+
+    The design logic tries to follow MLIR's dialect interfaces closely
+    https://mlir.llvm.org/docs/Interfaces/#dialect-interfaces
+    """
+
+    pass
+
+
+class OpAsmDialectInterface(DialectInterface):
```
superlopuh, line 16: It would be nice to add a doc string to make this page look a bit better:
https://xdsl--4503.org.readthedocs.build/4503/reference/dialect_interfaces/index.html#xdsl.dialect_interfaces.OpAsmDialectInterface.lookup

You can navigate to the updated pages from any PR from the readthedocs action result.

xdsl/ir/core.py
```diff
@@ -43,6 +44,7 @@
     from xdsl.printer import Printer
 
 OpT = TypeVar("OpT", bound="Operation")
+DialectInterfT = TypeVar("DialectInterfT", bound=DialectInterface)
```
math-fehr:
```suggestion
DialectInterfaceT = TypeVar("DialectInterfaceT", bound=DialectInterface)
```

xdsl/ir/core.py
```diff
@@ -79,6 +84,18 @@ def split_name(name: str) -> tuple[str, str]:
         except ValueError as e:
             raise ValueError(f"Invalid operation or attribute name {name}.") from e
 
+    def get_interface(self, inter: type[DialectInterfT]) -> DialectInterfT | None:
```
math-fehr:
```suggestion
    def get_interface(self, interface: type[DialectInterfaceT]) -> DialectInterfaceT | None:
```

xdsl/ir/core.py
```diff
@@ -79,6 +84,18 @@ def split_name(name: str) -> tuple[str, str]:
...
+        """
+        Return a class that implements the 'inter' interface if it exists.
+        """
+        for i in self._interfaces:
+            if isinstance(i, inter):
```
math-fehr:
```suggestion
            if isinstance(i, interface):
```

xdsl/ir/core.py
```diff
@@ -79,6 +84,18 @@ def split_name(name: str) -> tuple[str, str]:
...
+                return i
+        return None
+
+    def has_interface(self, inter: type[DialectInterfT]) -> bool:
+        return self.get_interface(inter) is not None
```
math-fehr:
```suggestion
    def has_interface(self, interface: type[DialectInterfaceT]) -> bool:
        return self.get_interface(interface) is not None
```

xdsl/ir/core.py
math-fehr: I would personally write `interface` entirely in most places, we don't lose that much space doing so, and `inter` and `interf` are not really common shortening tbh.

math-fehr (approved):
Hey!
Sorry for the delay, I got a bit busy lately.
The design looks good to me, I jsut put some nit comments about some names

PR #4507, opened 2025-06-06
core: deprecate attr_constr_coercion

alexarice:
It doesn't seem worth it to me to have both this and `irdl_to_attr_constraint`. Thoughts? Also moved `range_constr_coercion` and `single_range_constr_coercion` to the other file so that they don't need the local imports, though perhaps a better idea would be to move `irdl_to_attr_constraint`?

### DIFF
--- a/docs/marimo/irdl.py
+++ b/docs/marimo/irdl.py
@@ -170,7 +170,7 @@ def _(mo):
         r"""
     #### Attribute Constraint Coercion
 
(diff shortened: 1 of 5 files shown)


math-fehr (approved):
Good idea!

alexarice:
Any thoughts on how to structure the circular imports @math-fehr. Is this fine?

math-fehr:
Yeah that's what I would do too!

alexarice:
Great

PR #4508, opened 2025-06-06
bench: remove old bytecode profiling implementation and import-time dependency

EdmundGoodman:
Remove old bytecode profiling implementation and import-time dependency


uv.lock
superlopuh, line 1: let's revert this?

superlopuh (approved):

PR #4510, opened 2025-06-06
core: (constraints) ensure `AnyOf` cases are disjoint

alexarice:
I had a go at enforcing that the sets of bases are disjoint in `AnyOf` constraints. This is all the things that need to be changed. I think the only possible meaningful one is the `StencilType` change.

My hope is that this will allow:
- simplifying the verify method for `AnyOf`, as instead of the context copying stuff, we can just ensure that we are going down the correct branch first. An alternative to performing the check manually is to try to enforce that all constraints should check their base is correct before setting any variables (which most do).
- allow variables to be set by `AnyOf`. Extracting variables from anyof should be safe once branches are disjoint. This will allow us to get rid of `TypedAttributeConstraint`

### DIFF
--- a/tests/irdl/test_attr_constraint.py
+++ b/tests/irdl/test_attr_constraint.py
@@ -31,6 +31,7 @@ from xdsl.irdl import (
     eq,
     irdl_attr_definition,
 )
+from xdsl.utils.exceptions import PyRDLError
 
 
 def test_failing_inference():
@@ -78,6 +79,16 @@ class AttrC(Base):
         super().__init__(())
 
 
+@irdl_attr_definition
+class AttrD(Base):
+    name = "test.attr_d"
+
+    param: ParameterDef[AttrA | AttrC]
+
+    def __init__(self, param: AttrA | AttrC):
+        super().__init__((param,))
+
+
 @pytest.mark.parametrize(
     "constraint, expected",
     [
@@ -85,12 +96,14 @@ class AttrC(Base):
         (EqAttrConstraint(AttrB(AttrA())), {AttrB}),
         (BaseAttr(Base), None),
         (BaseAttr(AttrA), {AttrA}),
-        (EqAttrConstraint(AttrB(AttrA())) | AnyAttr(), None),
         (EqAttrConstraint(AttrB(AttrA())) | BaseAttr(AttrA), {AttrA, AttrB}),
-        (EqAttrConstraint(AttrB(AttrA())) | BaseAttr(AttrB), {AttrB}),
+        (
+            EqAttrConstraint(AttrD(AttrA())) | EqAttrConstraint(AttrD(AttrC())),
+            {AttrD},
+        ),
         (AllOf((AnyAttr(), BaseAttr(Base))), None),
         (AllOf((AnyAttr(), BaseAttr(AttrA))), {AttrA}),
-        (ParamAttrConstraint(AttrA, [BaseAttr(AttrB)]), {AttrA}),
+        (ParamAttrConstraint(AttrB, [BaseAttr(AttrA)]), {AttrB}),
         (ParamAttrConstraint(Base, [BaseAttr(AttrA)]), None),
         (VarConstraint("T", BaseAttr(Base)), None),
         (VarConstraint("T", BaseAttr(AttrA)), {AttrA}),
@@ -256,9 +269,9 @@ def test_memref_to_tensor(
             ParamAttrConstraint(AttrB, (BaseAttr(AttrA),)),
         ),
         (
-            ParamAttrConstraint(AttrB, (BaseAttr(AttrA),))
-            | ParamAttrConstraint(AttrB, (BaseAttr(AttrC),)),
-            ParamAttrConstraint(AttrB, (BaseAttr(AttrA) | BaseAttr(AttrC),)),
+            ParamAttrConstraint(AttrD, (BaseAttr(AttrA),))
+            | ParamAttrConstraint(AttrD, (BaseAttr(AttrC),)),
+            ParamAttrConstraint(AttrD, (BaseAttr(AttrA) | BaseAttr(AttrC),)),
         ),
         (
             ParamAttrConstraint(AttrB, (BaseAttr(AttrA),))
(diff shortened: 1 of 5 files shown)


tests/test_pyrdl.py
```diff
@@ -129,6 +145,32 @@ def test_any_attr_verify():
...
+    Check that an AnyOf constraint fails to verify if none of the constraints
+    verify.
+    """
+    constraint = BaseAttr(BoolData) | BaseAttr(IntData)
+
+    f = FloatData(10.0)
+
+    with pytest.raises(VerifyException) as e:
+        constraint.verify(f, ConstraintContext())
+    assert e.value.args[0] == f"Unexpected attribute {f}"
```
superlopuh: Not a fan of format strings in match cases
```suggestion
    with pytest.raises(VerifyException, match="Unexpected attribute 10.0"):
        constraint.verify(f, ConstraintContext())
```

xdsl/dialects/bufferization.py
```diff
@@ -88,6 +88,9 @@ def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None
             )
         return self.memref_constraint.verify(memref_type, constraint_context)
 
+    def get_bases(self) -> set[type[Attribute]] | None:
+        return {TensorType}
```
superlopuh: What about UnrangedTensorType?

xdsl/dialects/stencil.py
```diff
@@ -406,7 +406,7 @@ class TempType(
     name = "stencil.temp"
 
 
-StencilTypeConstr = StencilType[Attribute].constr()
+StencilTypeConstr = FieldType.constr() | TempType.constr()
```
superlopuh: this was unfortunately a blocker for @n-io, I'd be up for saying that we currently don't support this pattern if it's a blocker for this PR, but we should check in first whether this breaks anything. If we do go ahead, I would probably rather first refactor StencilType to be a Union instead of a superclass.

xdsl/utils/dialect_codegen.py
```diff
@@ -22,6 +22,7 @@
     VarOperandDef,
     VarResultDef,
 )
+from xdsl.utils.runtime_final import runtime_final
```
superlopuh: seems unrelated?

alexarice: `BaseAttr` uses the runtime final check to determine what's a `base` is.

tests/test_pyrdl.py
```diff
@@ -129,6 +145,31 @@ def test_any_attr_verify():
...
+def test_anyof_verify_fail():
+    """
+    Check that an AnyOf constraint fails to verify if none of the constraints
+    verify.
+    """
+    constraint = BaseAttr(BoolData) | BaseAttr(IntData)
+
+    f = FloatData(10.0)
+
+    with pytest.raises(VerifyException, match="Unexpected attribute 10.0"):
```
superlopuh: My b
```suggestion
    with pytest.raises(VerifyException, match=re.escape("Unexpected attribute #test.float<10.0>")):
```

xdsl/dialects/stencil.py
alexarice: what's the blocker? is there downstream code that this breaks?

alexarice:
If we want to go with this I should also add some tests

xdsl/dialects/stencil.py
superlopuh: I have been meaning to transform StencilType to this union for a while. I think nothing breaks, but it fundamentally makes the machinery less extensible. Now that the paper is submitted I would propose discussing this again. Could be a good topic for Monday's xDSL meeting.

n-io: I'm vague on details here - I *hope* that the reference/value types (Field/Temp, respectively) don't need extensibility. If all tests pass there should not be any issues.

superlopuh: OK let me quickly open a PR first to refactor this part of the code for consistency. I'll do this some time today.

superlopuh: I had a go and failed, happy to approve this change.

xdsl/irdl/constraints.py
```diff
@@ -439,6 +439,20 @@ def __init__(
...
+        for c in constrs:
+            b = c.get_bases()
+            if b is None:
+                raise PyRDLError(
+                    f"Constraint {c} cannot appear in an `AnyOf` constraint as its bases aren't known."
+                )
+            if not b.isdisjoint(bases):
+                raise PyRDLError(
+                    f"Constraints {constrs} do not have disjoint bases so cannot be put in an `AnyOf` constraint."
+                )
```
alexarice, line 467: So in the process of writing a test I have found I never updated `bases`

alexarice: This problem is more difficult than I appreciated. We have things like `i32 | i64` which I'm not sure what to do with. Of course we could do something funky with int constraints but that feels like a regression.
@math-fehr @superlopuh  any thoughts? Should branches with no variables be allowed to overlap?

superlopuh: I don't understand. i32 | i64 seem disjoint to me?

alexarice: They have the same base, which is what this pr uses to judge disjointness. Would be happy to consider a different disjointness criterion

superlopuh: Ah, hmm. We could consider special casing `|` on ParamAttrConstraint to fold the `|` into the children if the base is the same.

alexarice: It's not a `ParamAttrConstraint`, it's an `eq` constraint

superlopuh: We could consider making eq constraints ParamAttrConstraints :)

alexarice: Would still need to add an bunch more int constraint stuff for this. Doesn't feel like the way to go

superlopuh: Can you mark is as draft and push the changes that make things fail so that I can play with them?

alexarice: Done

tests/irdl/test_attr_constraint.py
```diff
@@ -60,35 +61,37 @@ class AttrA(Base):
...
 
-    param: ParameterDef[AttrA]
-
 
 @irdl_attr_definition
 class AttrC(Base):
     name = "test.attr_c"
 
+    param: ParameterDef[AttrA | AttrB]
+
```
alexarice, line 91: I renamed things to allow a test below but now realise that this has probably made the diff horiffic. Not sure how to fix really.

superlopuh: Not too bad

tests/test_pyrdl.py
```diff
@@ -130,6 +146,33 @@ def test_any_attr_verify():
     any_constraint.verify(IntData(0), ConstraintContext())
 
 
+def test_anyof_verify():
```
superlopuh: why move these?

xdsl/utils/dialect_codegen.py
superlopuh: Now I'm thinking this could be a nice separate PR

xdsl/parser/attribute_parser.py
```diff
@@ -744,19 +744,11 @@ def _parse_optional_builtin_parametrized_attr(self) -> Attribute | None:
 
     def _parse_dense_literal_type(
         self,
-    ) -> (
-        RankedStructure[IntegerType]
-        | RankedStructure[IndexType]
-        | RankedStructure[AnyFloat]
-        | RankedStructure[ComplexType]
-    ):
+    ) -> RankedStructure[IntegerType | IndexType | AnyFloat | ComplexType]:
```
superlopuh: this could also be a nice separate PR

tests/test_pyrdl.py
alexarice: I was moving the `LessThan` and `GreaterThan` closer to where they are now actually used but there's no point doing this, I'll revert

xdsl/parser/attribute_parser.py
alexarice: #4664

xdsl/utils/dialect_codegen.py
alexarice: #4663

tests/irdl/test_attr_constraint.py
superlopuh: If it's not too tedious, I think that it would be good to have separate tests for the case where the param is a union and not, so first rename the existing attrs (A(), B(), C(A)), then add D(A|B). WDYT? Or avoid renaming, and just add D(A|C)

alexarice: Sounds reasonable to me

alexarice: I went for `A, B(A), C, D(A|C)` which seems to keep the diff as small as possible.

xdsl/irdl/constraints.py
```diff
@@ -450,6 +450,32 @@ def __init__(
...
+                raise PyRDLError(
+                    f"Constraint {c} shares a base with a non-equality constraint "
+                    f"in {set(constrs[0:i])} in `AnyOf` constraint."
+                )
+
+            if isinstance(c, EqAttrConstraint):
+                eq_bases |= b
+            else:
+                if not b.isdisjoint(eq_bases):
+                    raise PyRDLError(
```
superlopuh, line 473: where are these errors tested?

superlopuh: Oh my bad

alexarice:
Rebased on top of Sasha's changes. @math-fehr would be good to get your opinions on these changes when you next get time.

math-fehr (approved):
That looks good to me!
I also think that it should be fairly easy to improve it in a later PR to only disallow cases where one of the base has a variable inside it (otherwise it is actually correct).

alexarice:
> That looks good to me! I also think that it should be fairly easy to improve it in a later PR to only disallow cases where one of the base has a variable inside it (otherwise it is actually correct).

Worth noting that the criteria here does allow things like #4665

math-fehr:
Yep that's pretty nice!

math-fehr:
Oh wait I see, if we only check for cases with variables, then it wouldn't allow to remove backtracking! Interesting

alexarice:
It would still remove the need for duplicating the constraintcontext

PR #4515, opened 2025-06-08
interpreter: (eqsat) Interpret for GetResult(s)Op

jumerckx:
~Draft until https://github.com/xdslproject/xdsl/pull/4514 gets merged.~

### DIFF
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -53,15 +53,19 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
             PDLInterpFunctions.run_get_result(self, interpreter, op, args).values[0],
         )
 
-        if (
-            result
-            and len(result.uses) == 1
-            and isinstance(
-                eclass_op := next(iter(result.uses)).operation, eqsat.EClassOp
-            )
+        if result is None:
+            return (None,)
+
+        if len(result.uses) == 1 and isinstance(
+            eclass_op := next(iter(result.uses)).operation, eqsat.EClassOp
         ):
             assert len(eclass_op.results) == 1
             result = eclass_op.results[0]
+        else:
+            raise InterpretationError(
+                "pdl_interp.get_result currently only supports operations with results"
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -34,3 +40,60 @@ def populate_known_ops(self, module: ModuleOp) -> None:
...
+        )
+
+        if result is None:
+            return (None,)
+
+        if len(result.uses) == 1 and isinstance(
+            eclass_op := next(iter(result.uses)).operation, eqsat.EClassOp
+        ):
+            assert len(eclass_op.results) == 1
+            result = eclass_op.results[0]
```
superlopuh: Seems unnecessary? Why not `eclass_op.result`?

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -34,3 +40,60 @@ def populate_known_ops(self, module: ModuleOp) -> None:
...
+        else:
+            raise InterpretationError(
+                "pdl_interp.get_result currently only supports operations with results"
+                " that are used by a single EClassOp each."
+            )
```
superlopuh, line 67: I would recommend flipping this, erroring out in a nested if, and carrying on downward on the happy path:
``` python
if len(result.uses) != 1 or not isinstance(...):
    raise ...

return (eclass_op.result,)
```

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -34,3 +40,60 @@ def populate_known_ops(self, module: ModuleOp) -> None:
...
+    def run_getresult(
+        self,
+        interpreter: Interpreter,
+        op: pdl_interp.GetResultOp,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        result = cast(
+            None | OpResult[Attribute],
+            PDLInterpFunctions.run_get_result(self, interpreter, op, args).values[0],
+        )
```
superlopuh: This is as many lines as the implementation of the original, let's just inline that here?

PR #4516, opened 2025-06-08
interpreter: (eqsat) Add backtracking mechanism for `GetDefiningOp`

jumerckx:
~Draft until https://github.com/xdslproject/xdsl/pull/4515 gets merged~

So backtracking is currently fully handled by the interpreter. I like the idea of trying to move logic into the IR itself but I think this is still a good place to start from?

### DIFF
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -138,7 +138,7 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
             if not self.visited:
                 if op != self.backtrack_stack[-1].gdo_op:
                     raise InterpretationError(
-                        "TODO: handle the case where a block contains multiple pdl_interp.get_defining_op."
+                        "Case where a block contains multiple pdl_interp.get_defining_op is currently not supported."
                     )
                 index = self.backtrack_stack[-1].index
                 self.visited = True
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -24,23 +24,68 @@ from xdsl.utils.scoped_dict import ScopedDict
 
 @dataclass
 class BacktrackPoint:
+    """
+    Represents a point in pattern matching where backtracking may be needed.
+
+    When a GetDefiningOpOp encounters an EClassOp with multiple operands,
+    we need to try matching against each operand. This class captures the
+    interpreter state so we can backtrack and try the next operand if
+    the current match fails.
+    """
+
     block: Block
+    """The block to return to when backtracking."""
+
     block_args: tuple[SSAValue, ...]
+    """Block arguments to restore when backtracking."""
+
     scope: ScopedDict[SSAValue, Any]
+    """Variable scope to restore when backtracking."""
+
     gdo_op: pdl_interp.GetDefiningOpOp
+    """The GetDefiningOpOp that created this backtrack point."""
+
     index: int
+    """Current operand index being tried in the EClassOp."""
+
     max_index: int
+    """Last valid operand index in the EClassOp (len(operands) - 1)."""
 
 
 @register_impls
 @dataclass
 class EqsatPDLInterpFunctions(PDLInterpFunctions):
+    """Interpreter functions for PDL patterns operating on e-graphs."""
+
     backtrack_stack: list[BacktrackPoint] = field(default_factory=list[BacktrackPoint])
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -1,26 +1,42 @@
...
+
+
+@dataclass
+class BacktrackPoint:
+    block: Block
+    block_args: tuple[SSAValue, ...]
+    scope: ScopedDict[SSAValue, Any]
+    gdo_op: pdl_interp.GetDefiningOpOp
+    index: int
+    max_index: int
```
superlopuh: Can you please add doc comments for the class, and each of the fields? I think more is better for this tricky part of the mechanism.

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -1,26 +1,42 @@
...
 
 
 @register_impls
 @dataclass
 class EqsatPDLInterpFunctions(PDLInterpFunctions):
+    backtrack_stack: list[BacktrackPoint] = field(default_factory=list[BacktrackPoint])
+    visited: bool = True
     known_ops: KnownOps = field(default_factory=KnownOps)
     eclass_union_find: DisjointSet[eqsat.EClassOp] = field(
```
superlopuh: Same here, more documentation would be fantastic. Ideally at a level where someone who isn't involved in the project (but is familiar with pdl) can come across this class and understand what each part does.

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -97,3 +113,56 @@ def run_getresults(
...
+    @impl(pdl_interp.GetDefiningOpOp)
+    def run_getdefiningop(
+        self,
+        interpreter: Interpreter,
+        op: pdl_interp.GetDefiningOpOp,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        defining_op = cast(
+            None | Operation,
+            PDLInterpFunctions.run_get_defining_op(self, interpreter, op, args).values[
```
superlopuh: same deal here, let's inline the implementation here?

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -97,3 +113,56 @@ def run_getresults(
...
+                0
+            ],
+        )
+        if isinstance(eclass_op := defining_op, eqsat.EClassOp):
```
superlopuh: Let's reduce the indentation levels by bailing early instead?
```suggestion
        if not isinstance(defining_op, eqsat.EClassOp):
            return (defining_op,)
```

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -97,3 +113,56 @@ def run_getresults(
...
+            if not self.visited:
+                if op != self.backtrack_stack[-1].gdo_op:
+                    raise InterpretationError(
```
superlopuh: can you please add a comment explaining how this condition is caused by multiple `get_defining_op`s in a block?

superlopuh: Actually more comments below would be great, I'm not 100% sure I understand the mechanism here

jumerckx:
I added comments. (https://github.com/xdslproject/xdsl/pull/4516/commits/0787f50991afcae1e6e8f98066ddc747646019cf)
> let's inline the implementation here

I did (https://github.com/xdslproject/xdsl/pull/4516/commits/14f69ac1afa308616029e29f2e0da3424cc3a7cb), but what's the rationale for not deduplicating?

superlopuh:
It just feels that for such a short function we're not gaining robustness or readabiity by calling to the other interpreter function.

xdsl/interpreters/eqsat_pdl_interp.py
superlopuh: This is more of a nit, feel free to merge as-is if you prefer

jumerckx:
In that case, I've also inlined the second call as well. (https://github.com/xdslproject/xdsl/pull/4516/commits/3b6e916cc5a32f4bdfa0af2769234668144edc4e)

xdsl/interpreters/eqsat_pdl_interp.py
jumerckx: I'd missed this one, should be OK now.

jumerckx:
@superlopuh good to merge?

superlopuh (approved):

PR #4517, opened 2025-06-08
interpreter: (eqsat) Interpret for `CreateOperationOp`

jumerckx:
Draft until https://github.com/xdslproject/xdsl/pull/4516 gets merged

### DIFF
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -16,6 +16,7 @@ from xdsl.interpreter import (
 )
 from xdsl.interpreters.pdl_interp import PDLInterpFunctions
 from xdsl.ir import Attribute, Block, Operation, OpResult, SSAValue
+from xdsl.rewriter import InsertPoint
 from xdsl.transforms.common_subexpression_elimination import KnownOps
 from xdsl.utils.disjoint_set import DisjointSet
 from xdsl.utils.exceptions import InterpretationError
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -212,6 +213,46 @@ def run_getdefiningop(
 
         return (defining_op,)
 
+    @impl(pdl_interp.CreateOperationOp)
+    def run_createoperation(
```
superlopuh: !! if we name things differently there will be two ways to call this function, I'm not sure how things might go wrong as a result but I'm sure that if they do they'll be a massive pain to debug
```suggestion
    def run_create_operation(
```

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -212,6 +213,46 @@ def run_getdefiningop(
...
+        self,
+        interpreter: Interpreter,
+        op: pdl_interp.CreateOperationOp,
+        args: tuple[Any, ...],
+    ) -> tuple[Any, ...]:
+        has_done_action_checkpoint = self.rewriter.has_done_action
+        (new_op,) = PDLInterpFunctions.run_create_operation(
+            self, interpreter, op, args
+        ).values
```
superlopuh:
```suggestion
        (new_op,) = super().run_create_operation(
            interpreter, op, args
        ).values
```

xdsl/interpreters/eqsat_pdl_interp.py
jumerckx: Agreed! I do think I haven't been consistent throughout all pdl_interp related code so it might be worthwhile to keep this for a follow up pr where I check everything?

superlopuh: I think a pass would be worth it, but also feels like fixing this here would make sense.

superlopuh (approved):

superlopuh:
Branch this was based on got merged, so I assume this is ready :)

PR #4518, opened 2025-06-08
core: replace generic_constraint_coercion with generic_constraint

superlopuh:
I thought I'd do a drive-by change here now that we've introduced TypeVarConstraint, to make `GenericData` take advantage of the new functionality.

### DIFF
--- a/xdsl/dialects/llvm.py
+++ b/xdsl/dialects/llvm.py
@@ -12,7 +12,6 @@ from xdsl.dialects.builtin import (
     ArrayAttr,
     ContainerType,
     DenseArrayBase,
-    IndexType,
     IntAttr,
     IntegerAttr,
     IntegerType,
@@ -129,7 +128,7 @@ class LLVMStructType(ParametrizedAttribute, TypeAttribute):
                 printer.print_list(self.types.data, printer.print_attribute)
 
     @classmethod
-    def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
+    def parse_parameters(cls, parser: AttrParser) -> tuple[StringAttr, ArrayAttr]:
         parser.parse_characters("<", " in LLVM struct")
         struct_name = parser.parse_optional_str_literal()
         if struct_name is None:
@@ -141,7 +140,7 @@ class LLVMStructType(ParametrizedAttribute, TypeAttribute):
             parser.Delimiter.PAREN, lambda: parse_llvm_type(parser)
         )
         parser.parse_characters(">", " to close LLVM struct parameters")
-        return [StringAttr(struct_name), ArrayAttr(params)]
+        return (StringAttr(struct_name), ArrayAttr(params))
 
 
 @irdl_attr_definition
@@ -150,7 +149,7 @@ class LLVMPointerType(
 ):
     name = "llvm.ptr"
 
-    type: ParameterDef[Attribute | NoneAttr]
+    type: ParameterDef[Attribute]
     addr_space: ParameterDef[IntAttr | NoneAttr]
 
     def print_parameters(self, printer: Printer) -> None:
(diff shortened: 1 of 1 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1144,7 +1146,7 @@ def pack(self, values: Sequence[tuple[float, float] | tuple[int, int]]) -> bytes
 
 
 @irdl_attr_definition
-class DictionaryAttr(GenericData[immutabledict[str, Attribute]], BuiltinAttribute):
+class DictionaryAttr(Data[immutabledict[str, Attribute]], BuiltinAttribute):
```
superlopuh: Not sure why this was a GenericData

tests/irdl/test_attribute_definition.py
```diff
@@ -729,10 +728,9 @@ def print_parameter(self, printer: Printer) -> None:
             printer.print_list(self.data, printer.print_attribute)
             printer.print_string("]")
 
-    @staticmethod
-    def generic_constraint_coercion(args: tuple[Any]) -> AttrConstraint:
-        assert len(args) == 1
-        return DataListAttr(irdl_to_attr_constraint(args[0]))
+    @classmethod
+    def generic_constraint(cls) -> AttrConstraint:
```
alexarice: Why not `.constr`?

superlopuh: Oh yeah

superlopuh: This turns out to be a bit tricky as ArrayAttr already has a `.constr` that is not compatible with this one, I'll take a look in a bit.

superlopuh: Fixed

xdsl/irdl/attributes.py
```diff
@@ -75,12 +75,12 @@ class GenericData(Data[_DataElement], ABC):
...
+    @classmethod
     @abstractmethod
-    def generic_constraint_coercion(args: tuple[Any]) -> AttrConstraint:
+    def constr(cls) -> AttrConstraint:
         """
-        Given the generic parameters passed to the generic attribute type,
-        return the corresponding attribute constraint.
+        Returns a constraint for this subclass.
+        Generic arguments are constrained via TypeVarConstraints.
         """
```
alexarice, line 83: I feel we had trouble before with having a very general `constr` on superclasses (or at least on Attribute). I seem to remember the issue being something along the lines of:
```
def constr(cls) -> AttrConstraint
```
vs
```
def constr(cls) -> GenericAttrConstraint[Self]
```
or something like this? Is there a reason this is resolved/ok?

superlopuh: Yep exactly, the problem is that we can't enforce that the generics are preserved, but this sort of "gives up" and doesn't preserve generics anyway, so there is no problem.

superlopuh: Note that this comes at the cost of specificity of ArrayAttr.constr, which I am proposing to remove, letting clients make an ArrayOfConstraint directly if they feel like it.

alexarice: This is more than giving up Generics, this gives up the base type too

superlopuh: I'll take a look at this now.

superlopuh: Yeah this is broken.

superlopuh: OK I went back to the way it was before, the problem with using .constr is that we need to specify types in the subclass, and have it be type-erased in the superclass.

tests/irdl/test_attribute_definition.py
superlopuh: Fix reverted

superlopuh:
@math-fehr thoughts on this?

xdsl/irdl/attributes.py
```diff
@@ -393,10 +393,12 @@ def irdl_to_attr_constraint(
         if len(args) != 1:
             raise Exception(f"GenericData args must have length 1, got {args}")
         origin = cast(type[GenericData[Any]], origin)
-        args = cast(tuple[Attribute], args)
+        type_var = get_type_var_from_generic_class(origin)[0]
```
alexarice: Will generic data always have exactly one generic?

xdsl/irdl/attributes.py
superlopuh: Turns out this works with static methods!

xdsl/irdl/attributes.py
superlopuh: Yep we enforce this on line 394

PR #4520, opened 2025-06-09
dialects: (stencil) replace IndexAttr.parse_nested_parameters

superlopuh:
Minor change to the API to make the typing a bit more explicit. I'm playing around with better Attribute inits in a branch and this change seemed like a quick separate refactor to consider.


n-io (approved):
LGTM 👍

PR #4524, opened 2025-06-10
dialects: (builtin) support printing and parsing tuple types

superlopuh:
Test and format from this: https://mlir.llvm.org/docs/Dialects/Builtin/#tupletype

### DIFF
--- a/xdsl/dialects/experimental/fir.py
+++ b/xdsl/dialects/experimental/fir.py
@@ -25,7 +25,6 @@ from xdsl.dialects.builtin import (
     IntegerType,
     StringAttr,
     SymbolRefAttr,
-    TupleType,
     UnitAttr,
 )
 from xdsl.ir import (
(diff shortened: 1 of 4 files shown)


xdsl/dialects/experimental/fir.py
```diff
@@ -139,44 +138,6 @@ class ReferenceType(ParametrizedAttribute, TypeAttribute):
...
-        if isinstance(self.type, TupleType):
-            printer.print("tuple<")
-            for idx, t in enumerate(self.type.types.data):
-                if idx > 0:
-                    printer.print(", ")
-                printer.print(t)
-            printer.print(">")
-        else:
-            printer.print(self.type)
-        printer.print(">")
```
superlopuh, line 156: The default behaviour should now print the same thing as this custom implementation. This dialect is experimental so I am mostly hoping to be correct here as there are no tests.

xdsl/printer.py
```diff
@@ -676,6 +677,12 @@ def print_int_or_question(value: IntAttr | NoneAttr) -> None:
             self.print_string(f"affine_set<{attribute.data}>")
             return
 
+        if isinstance(attribute, TupleType):
+            self.print_string("tuple<", indent=0)
```
alexarice: what's with the `indent`?

xdsl/parser/attribute_parser.py
```diff
@@ -600,6 +601,14 @@ def _parse_tensor_attrs(self) -> AnyTensorType | AnyUnrankedTensorType:
 
         return TensorType(type, shape)
 
+    def _parse_tuple_attrs(self) -> TupleType:
+        params = self.parse_optional_undelimited_comma_separated_list(
+            self.parse_optional_attribute, self.parse_attribute
+        )
+        if params is None:
+            params = []
```
alexarice, line 618: MLIR doesn't accept `tuple` (with no `<>`) so I imagine we shouldn't either.

xdsl/printer.py
superlopuh: It bypasses the check for adding indentation for nested printing, we currently do that by printing newlines between things, and adding an indent of 0 fast-tracks to printing the raw string.

xdsl/parser/attribute_parser.py
superlopuh: https://mlir.llvm.org/docs/Dialects/Builtin/#tupletype this seems to disagree. I can add an MLIR interop test to catch this sort of thing.

alexarice: I thought this would accept the angle brackets being omitted but I seem to be wrong

xdsl/printer.py
alexarice: I don't really understand why this is necessary.

superlopuh: I do it becasue of a perf measurement I did a while back, but I see that the `print_string` got rewritten and this is now not useful, I'll remove it.

alexarice (approved):

PR #4525, opened 2025-06-11
core: Parameter to load dialect if needed

mamanain:
When parsing dialect resources we might [have resources](https://github.com/xdslproject/xdsl/pull/4502/files#diff-33d76e9b12c370dfd56375282fd868782518fe1914d9c02c9c300bc82d31ca1bR163) for a registered dialect which doesn't have any operations parsed. So it is not a loaded and we need to construct the object. 

I thought maybe it would be better to do this inside the `get_optional_dialect` function?

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -265,13 +265,11 @@ def test_load_dialect_already_registered():
         ctx.load_dialect(testDialect)
 
 
-def test_load_if_needed():
+def test_get_dialect():
     ctx = Context()
     assert ctx.get_optional_dialect("test") is None
-    assert ctx.get_optional_dialect("test", load_if_needed=True) is None
     ctx.register_dialect("test", lambda: testDialect)
-    assert ctx.get_optional_dialect("test") is None
-    assert ctx.get_optional_dialect("test", load_if_needed=True) is testDialect
+    assert ctx.get_optional_dialect("test") is testDialect
 
 
 def test_get_optional_op_registered():
--- a/xdsl/context.py
+++ b/xdsl/context.py
@@ -287,21 +287,17 @@ class Context:
             raise UnregisteredConstructException(f"Dialect {name} is not registered")
         return dialect
 
-    def get_optional_dialect(
-        self, name: str, load_if_needed: bool = False
-    ) -> "Dialect | None":
+    def get_optional_dialect(self, name: str) -> "Dialect | None":
         """
-        Get a loaded dialect if it exists.
+        Get a dialect from its name if it exists.
+        If the dialect is not registered, return None.
+        """
+        if name in self._registered_dialects and name not in self._loaded_dialects:
+            self.load_registered_dialect(name)
 
-        `load_if_needed` parameter if set loads the corresponding registered
-        dialect if it wasn't loaded before.
-        With the parameter set function is equivalent to MLIR's `MLIRContext::getOrLoadDialect`.
-        """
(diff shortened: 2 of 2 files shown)


superlopuh:
How does MLIR handle this?

superlopuh:
This LGTM, happy to merge after @math-fehr's approval.

tests/test_context.py
```diff
@@ -265,6 +265,13 @@ def test_load_dialect_already_registered():
         ctx.load_dialect(testDialect)
 
 
+def test_load_if_needed():
+    ctx = Context()
+    ctx.register_dialect("test", lambda: testDialect)
+    assert ctx.get_optional_dialect("test") is None
+    assert ctx.get_optional_dialect("test", load_if_needed=True)
```
superlopuh:
```suggestion
    ctx = Context()
    assert ctx.get_optional_dialect("test") is None
    assert ctx.get_optional_dialect("test", load_if_needed=True) is None
    ctx.register_dialect("test", lambda: testDialect)
    assert ctx.get_optional_dialect("test") is None
    assert ctx.get_optional_dialect("test", load_if_needed=True) is testDialect
```

mamanain:
Oh. MLIR actually has a separate `getOrLoadDialect` function which is called in metadata parsing.

I guess it would be better to create the same function in xdsl instead?

superlopuh:
Is the implementation equivalent to the one added in this PR? I'd rather have one function with an option than duplicate the functionality.

superlopuh:
If so, it may warrant a comment and a link to the MLIR function.

mamanain:
I think it's the same:

``` c++
Dialect *MLIRContext::getOrLoadDialect(StringRef name) {
  Dialect *dialect = getLoadedDialect(name);
  if (dialect)
    return dialect;
  DialectAllocatorFunctionRef allocator =
      impl->dialectsRegistry.getDialectAllocator(name);
  return allocator ? allocator(this) : nullptr;
}
```

I'll insert a link.

tests/test_context.py
superlopuh: I think it's worth testing the behaviour of the unregistered case, and to actually check that the result is the one expected.

xdsl/context.py
```diff
@@ -287,7 +287,19 @@ def get_dialect(self, name: str) -> "Dialect":
             raise UnregisteredConstructException(f"Dialect {name} is not registered")
         return dialect
 
-    def get_optional_dialect(self, name: str) -> "Dialect | None":
+    def get_optional_dialect(
+        self, name: str, load_if_needed: bool = False
```
math-fehr: Can you remove this variable, and always assume that this is true?
So that this copies how attributes and types work.

xdsl/context.py
```diff
@@ -287,7 +287,19 @@ def get_dialect(self, name: str) -> "Dialect":
...
+    ) -> "Dialect | None":
+        """
+        Get a loaded dialect if it exists.
+
+        `load_if_needed` parameter if set loads the corresponding registered
+        dialect if it wasn't loaded before.
+        With the parameter set function is equivalent to MLIR's `MLIRContext::getOrLoadDialect`.
```
math-fehr: Can you more or less copy the attribute and type documentation there?

math-fehr (approved):
That makes sense to me, can you just remove "load_if_needed" and assume that it is always true?
Then that's fine for me!

PR #4526, opened 2025-06-11
dialects: `DenseResourceAttr` fixes

mamanain:
Make `DenseReourceAttr` behave more like in MLIR, allow it to be used as a constant's value.

Also some preparations for resource printing. Part of this bigger pr.


xdsl/parser/attribute_parser.py
```diff
@@ -852,7 +853,11 @@ def _parse_builtin_dense_resource_attr(self) -> DenseResourceAttr:
         resource_handle = self.parse_identifier(" for resource handle")
         self.parse_characters(">", " in dense_resource attribute")
         self.parse_characters(":", " in dense_resource attribute")
+
         type = self.parse_type()
+        if not isinstance(type, ShapedType):
+            self.raise_error(f"dense resource should have a shaped type, got: {type}")
```
superlopuh, line 859: let's add a test for this? But also, shouldn't this be a verification error?

xdsl/printer.py
superlopuh, line 1: this seems like a separate change?

mamanain: Well, there is `print_resource_handle` used in DenseResource printing, so they are somewhat intertwined.

superlopuh: I mean would the rest of the PR be affected without this change?

mamanain: Yes, it defines `print_resource_handle` which is used in the attribute's printing

xdsl/parser/attribute_parser.py
mamanain: I added this check here because otherwise pyright would complain that I'm passing an incorrect type to the attribute's constructor.

Also we do the same kind of checks in `_parse_builtin_densearray_attr`

xdsl/printer.py
superlopuh: I see, then I would suggest first opening a PR for this, that doesn't depend on DenseResourceAttr, with a test in `test_printer.py`

xdsl/parser/attribute_parser.py
superlopuh: OK, let's add tests though still :)

xdsl/printer.py
mamanain: You know, I thought about it for a second.  I'll just delete this printing logic for now and bring it back alongside resource printing.

Because otherwise this logic is useless. It would make sense to combine all printing features in one pr.

PR #4532, opened 2025-06-13
interpreter: (eqsat) `run_replace` and `apply_matches`

jumerckx:
~~I still have to add some more comments.~~

### DIFF
--- a/tests/interpreters/test_eqsat_pdl_interp_interpreter.py
+++ b/tests/interpreters/test_eqsat_pdl_interp_interpreter.py
@@ -656,10 +656,10 @@ def test_run_replace_same_eclass():
     )
 
     # Should return empty tuple
-    assert result.values == ()
+    assert not result.values
 
     # Should not add any merge todos since it's the same EClass
-    assert len(interp_functions.merge_list) == 0
+    assert not interp_functions.merge_list
 
 
 def test_run_replace_error_not_eclass_original():
@@ -767,7 +767,7 @@ def test_apply_matches():
     interp_functions.apply_matches()
 
     # Should have cleared the merge list
-    assert len(interp_functions.merge_list) == 0
+    assert not interp_functions.merge_list
 
     # Should have merged operands from to_replace into canonical
     assert len(canonical.operands) == initial_operand_count + len(to_replace.operands)
@@ -808,4 +808,4 @@ def test_apply_matches_with_known_ops_restore():
     assert interp_functions.known_ops[test_op] is test_op
 
     # Should have cleared the restore list
-    assert len(interp_functions.known_ops_restore_list) == 0
+    assert not interp_functions.known_ops_restore_list
--- a/tests/interpreters/test_eqsat_pdl_interp_interpreter.py
+++ b/tests/interpreters/test_eqsat_pdl_interp_interpreter.py
@@ -582,3 +582,230 @@ def test_run_finalize_with_backtrack_stack():
     assert result.terminator_value.values == ()
     assert result.values == ()
     assert len(interp_functions.backtrack_stack) == 0
+
+
+def test_run_replace():
+    """Test that run_replace correctly merges EClass operations."""
+    interpreter = Interpreter(ModuleOp([]))
+    ctx = Context()
+    interp_functions = EqsatPDLInterpFunctions(ctx)
+
+    # Create test operations
+    c0 = create_ssa_value(i32)
+    original_op = test.TestOp((c0,), (i32,))
+    replacement_op = test.TestOp((c0,), (i32,))
+
+    # Create EClass operations for both
+    original_eclass = eqsat.EClassOp(original_op.results[0], res_type=i32)
+    replacement_eclass = eqsat.EClassOp(replacement_op.results[0], res_type=i32)
+
+    # Add both EClass operations to union-find
+    interp_functions.eclass_union_find.add(original_eclass)
+    interp_functions.eclass_union_find.add(replacement_eclass)
+
+    # Create a ReplaceOp for testing
+    input_op_value = create_ssa_value(pdl.OperationType())
+    repl_value = create_ssa_value(pdl.ValueType())
+    replace_op = pdl_interp.ReplaceOp(input_op_value, [repl_value])
+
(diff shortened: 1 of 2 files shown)


xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -83,11 +91,28 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
...
+    def modification_handler(self, op: Operation):
+        """
+        Keeps `known_ops` up to date.
+        Whenever an operation is modified, for example when its operands are updated to a different eclass value,
+        the operation is added to `known_ops_restore_list`. At the end of `apply_matches`, all the (now updated)
+        operations in `known_ops_restore_list` are put back into `known_ops`.
+        """
+        if op in self.known_ops:
+            removed = self.known_ops.pop(op)
+            self.known_ops_restore_list.append(removed)
```
jumerckx, line 114: This whole approach is a bit finicky. I'm wondering if there's a better way.

superlopuh: How does egg do this?

xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -266,3 +345,26 @@ def run_finalize(
...
         return ReturnedValues(()), ()
+
+    def apply_matches(self):
+        todo = OrderedSet(
+            (self.eclass_union_find.find(todo.to_keep), todo.to_replace)
+            for todo in self.merge_list
+        )
+        self.merge_list.clear()
+        for to_keep, to_replace in todo:
+            operands = to_keep._operands  # pyright: ignore[reportPrivateUsage]
```
superlopuh: why the underscore?

xdsl/interpreters/eqsat_pdl_interp.py
jumerckx: I had a look at their implementation, egg keeps `known_ops` (called `memo` in their implementation) up to date in the rebuilding procedure (in `process_unions`).
Each parent e-node of one of the e-classes that is union-ed is put on a queue and is processed at this point.

When such a parent is processed, its children are updated to point to the union-ed e-class.
At this point, the updated e-node(/operation) is added to `memo`. The old entry is not removed but simply kept around AFAICT.

I think we should definitely have an option to rebuild like this. pdl rewritepatterns and other passes can register equalities and these can then be applied in combination with rebuilding. I’m not sure if that should be blocking for this pr though?

superlopuh: Definitely not blocking for this PR!

xdsl/interpreters/eqsat_pdl_interp.py
jumerckx: I'm working with the underlying tuples, creating a new tuple using `+`. But setting the new operands shouldn't use privateUsage, I've updated this.

superlopuh: I would much rather we did something like this:
```suggestion
            operands = tuple(to_keep._operands)
```

Or just something like this below:
```
new_operands = (*to_keep.operands, *to_replace.operands)
```

Changing the underlying representation of the operands shouldn't change this interpreter.

jumerckx: Ah right, thanks!

tests/interpreters/test_eqsat_pdl_interp_interpreter.py
```diff
@@ -582,3 +582,230 @@ def test_run_finalize_with_backtrack_stack():
...
+    # Call run_replace directly
+    result = interp_functions.run_replace(
+        interpreter, replace_op, (test_op, eclass_op.results[0])
+    )
+
+    # Should return empty tuple
+    assert result.values == ()
+
+    # Should not add any merge todos since it's the same EClass
+    assert len(interp_functions.merge_list) == 0
```
superlopuh:
```suggestion
    assert not interp_functions.merge_list
```

tests/interpreters/test_eqsat_pdl_interp_interpreter.py
```diff
@@ -582,3 +582,230 @@ def test_run_finalize_with_backtrack_stack():
...
+    repl_value = create_ssa_value(pdl.ValueType())
+    replace_op = pdl_interp.ReplaceOp(input_op_value, [repl_value])
+
+    # Call run_replace directly
+    result = interp_functions.run_replace(
+        interpreter, replace_op, (test_op, eclass_op.results[0])
+    )
+
+    # Should return empty tuple
+    assert result.values == ()
```
superlopuh:
```suggestion
    assert not result.values
```

tests/interpreters/test_eqsat_pdl_interp_interpreter.py
```diff
@@ -582,3 +582,230 @@ def test_run_finalize_with_backtrack_stack():
...
+    interp_functions.merge_list.append(MergeTodo(canonical, to_replace))
+
+    # Track initial operand count
+    initial_operand_count = len(canonical.operands)
+
+    # Apply the matches
+    interp_functions.apply_matches()
+
+    # Should have cleared the merge list
+    assert len(interp_functions.merge_list) == 0
```
superlopuh:
```suggestion
    assert not interp_functions.merge_list
```

tests/interpreters/test_eqsat_pdl_interp_interpreter.py
```diff
@@ -582,3 +582,230 @@ def test_run_finalize_with_backtrack_stack():
...
+
+    # Apply matches (with empty merge list to focus on restore functionality)
+    interp_functions.apply_matches()
+
+    # Should have restored the operation to known_ops
+    assert test_op in interp_functions.known_ops
+    assert interp_functions.known_ops[test_op] is test_op
+
+    # Should have cleared the restore list
+    assert len(interp_functions.known_ops_restore_list) == 0
```
superlopuh:
```suggestion
    assert not interp_functions.known_ops_restore_list
```

PR #4534, opened 2025-06-13
core: Print metadata in file

mamanain:
Last part of [this](https://github.com/xdslproject/xdsl/pull/4502/files) pr.

We start printing the parsed metadata in a special subsection inside an mlir file.

### DIFF
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -10,6 +10,7 @@ from typing import Any, cast
 
 from typing_extensions import TypeVar
 
+from xdsl.dialect_interfaces import OpAsmDialectInterface
 from xdsl.dialects.builtin import (
     AffineMapAttr,
     AffineSetAttr,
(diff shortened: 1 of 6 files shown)


xdsl/printer.py
```diff
@@ -904,3 +913,49 @@ def print_op(self, op: Operation) -> None:
             self.print_op_with_default_format(op)
         if scope:
             self.exit_scope()
+
+    def print_resource_handle(self, dialect: str, handle: str) -> None:
+        if dialect not in self._dialect_resources:
+            self._dialect_resources[dialect] = set()
```
superlopuh, line 919: Is the order of items leveraged? If so whould this be an ordered set?

superlopuh: Oh actually I see that they're passed into an interface, so sorting wouldn't do much.

superlopuh (approved):

PR #4535, opened 2025-06-14
transformations: (riscv) Add canonicalization pattern x*1->x

zykdtc:
Added the canonicalization pattern of multiply by 1 refer to the GETTING_STARTED guide.

### DIFF
--- a/tests/filecheck/backend/riscv/canonicalize.mlir
+++ b/tests/filecheck/backend/riscv/canonicalize.mlir
@@ -45,6 +45,12 @@ builtin.module {
   %multiply_immediate_l0 = riscv.mul %i1, %c0 : (!riscv.reg<a1>, !riscv.reg) -> !riscv.reg<a0>
   "test.op"(%multiply_immediate_l0) : (!riscv.reg<a0>) -> ()
 
+  %multiply_immediate_r1 = riscv.mul %c1, %i1 : (!riscv.reg, !riscv.reg<a1>) -> !riscv.reg<a0>
+  "test.op"(%multiply_immediate_r1) : (!riscv.reg<a0>) -> ()
+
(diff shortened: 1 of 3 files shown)


superlopuh:
Isn't this case already handled by the existing pattern that checks for constant arguments? It seems that you added a pattern from the list of existing patterns in the getting started guide, instead of one of the suggested patterns to add below.

zykdtc:
> Isn't this case already handled by the existing pattern that checks for constant arguments? It seems that you added a pattern from the list of existing patterns in the getting started guide, instead of one of the suggested patterns to add below.

Just find out this pattern appear at both the list of existing patterns and the suggested patterns. But I'm not sure if it already be handled since x*1 won't be simplifiled when I tested it before.

superlopuh:
You're right! Let's merge your new pattern into the existing pattern, though. Here is what I would recommend to change in the `MultiplyImmediates` pattern:

```
match (lhs.get_constant, rhs.get_constant):
  case (None, None):
    return
  case (l, None):
    replace self with mul(rhs, lhs)
  case (None, r):
    if r == 1:
        replace self with mov(lhs)
    elif r == 0:
        replace self with mov(zero)
    else:
        return
  case (l, r):
        replace self with li(l * r)
```

There are a few things off with the existing patterns:

1. the assumption is that if either of the operands are constant, then the rhs is a constant, this avoids checking separately for lhs and rhs being constant in patterns that look at their operands
2. there are separate patterns for zero, one, etc and that just feels wasteful, we might as well do the check in one rewrite

Would you mind doing this change?

zykdtc:
Sure, this design looks reasonable and more consice. I'll implement it and get rid of the redundant patterns for zero and one.

superlopuh (approved):
Beautiful!

PR #4536, opened 2025-06-14
core: Parser.parse_type returns a TypeAttribute

superlopuh:
Make the type system aware of this existing behaviour.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1167,9 +1167,9 @@ class DictionaryAttr(GenericData[immutabledict[str, Attribute]], BuiltinAttribut
 class TupleType(ParametrizedAttribute, BuiltinAttribute):
     name = "tuple"
 
-    types: ParameterDef[ArrayAttr[Attribute]]
+    types: ParameterDef[ArrayAttr[TypeAttribute]]
 
-    def __init__(self, types: list[Attribute] | ArrayAttr[Attribute]) -> None:
+    def __init__(self, types: list[TypeAttribute] | ArrayAttr[TypeAttribute]) -> None:
         if isinstance(types, list):
             types = ArrayAttr(types)
         super().__init__([types])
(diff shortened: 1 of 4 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1167,9 +1167,9 @@ def verify(self) -> None:
 class TupleType(ParametrizedAttribute, BuiltinAttribute):
     name = "tuple"
 
-    types: ParameterDef[ArrayAttr[Attribute]]
+    types: ParameterDef[ArrayAttr[TypeAttribute]]
```
superlopuh, line 1170: this is technically unnecessary but seems harmless. I needed to change this file because `list` is invariant, so I thought I'd change the base type also.

xdsl/parser/attribute_parser.py
```diff
@@ -276,6 +276,16 @@ def _parse_dialect_type_or_attribute_body(
...
             raise TypeError("Attributes are either ParametrizedAttribute or Data.")
 
+    @overload
+    def _parse_extended_type_or_attribute(
+        self, attr_or_dialect_name: str, is_type: Literal[False]
+    ) -> Attribute: ...
+
+    @overload
+    def _parse_extended_type_or_attribute(
+        self, attr_or_dialect_name: str, is_type: Literal[True] = True
```
alexarice: Do you need the default here?

alexarice (approved):

xdsl/parser/attribute_parser.py
```diff
@@ -400,7 +410,7 @@ def _parse_unregistered_attr_body(self, start_pos: Position | None) -> str:
         assert body is not None
         return body
 
-    def _parse_optional_builtin_parametrized_type(self) -> ParametrizedAttribute | None:
```
superlopuh, line 403: nothing was actually relying on this being a ParametrizedAttribute, and there's no `A & B` intersection type mechanism

xdsl/parser/attribute_parser.py
superlopuh: I'm not sure whether it was needed originally, but since this is a private method and this is actually an important argument to get right, I removed the default value, which seems to work fine.

PR #4543, opened 2025-06-16
core: (Printer) remove variadic prints

alexarice:
Removes all instances of using the variadic behaviour of `Printer.print`. All tests pass if `assert len(argv) == 1` is added to the top of this function.

I haven't added any sort of deprecation/change to `Printer.print` because I wasn't sure how and it didn't seem worth it if we were just going to deprecate the whole thing anyway.

### DIFF
--- a/xdsl/dialects/experimental/air.py
+++ b/xdsl/dialects/experimental/air.py
@@ -836,9 +836,9 @@ class WaitAllOp(IRDLOperation):
         super().__init__(operands=[async_dependencies], result_types=[AsyncTokenAttr()])
 
     def print(self, printer: Printer):
-        printer.print(" async ")
+        printer.print_string(" async ")
         if self.async_dependencies:
-            printer.print(self.async_dependencies, printer.print_ssa_value, ",")
+            printer.print_list(self.async_dependencies, printer.print_ssa_value)
 
     @classmethod
     def parse(cls, parser: Parser) -> WaitAllOp:
--- a/xdsl/dialects/irdl/irdl.py
+++ b/xdsl/dialects/irdl/irdl.py
@@ -135,7 +135,7 @@ class DialectOp(IRDLOperation):
         return DialectOp(sym_name, region)
 
     def print(self, printer: Printer) -> None:
-        printer.print(" @", self.sym_name.data, " ")
+        printer.print(f" @{self.sym_name.data} ")
         if self.body.block.ops:
             printer.print_region(self.body)
 
@@ -165,7 +165,7 @@ class TypeOp(IRDLOperation):
         return TypeOp(sym_name, region)
 
     def print(self, printer: Printer) -> None:
-        printer.print(" @", self.sym_name.data, " ")
+        printer.print(f" @{self.sym_name.data} ")
         if self.body.block.ops:
             printer.print_region(self.body)
 
@@ -220,7 +220,7 @@ class AttributeOp(IRDLOperation):
         return AttributeOp(sym_name, region)
 
     def print(self, printer: Printer) -> None:
-        printer.print(" @", self.sym_name.data, " ")
+        printer.print(f" @{self.sym_name.data} ")
         if self.body.block.ops:
             printer.print_region(self.body)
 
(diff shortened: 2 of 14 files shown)


xdsl/dialects/irdl/irdl.py
```diff
@@ -135,7 +135,7 @@ def parse(cls, parser: Parser) -> DialectOp:
         return DialectOp(sym_name, region)
 
     def print(self, printer: Printer) -> None:
-        printer.print(" @", self.sym_name.data, " ")
+        printer.print(f" @{self.sym_name.data} ")
```
alexarice, line 138: It feels like we should have a helper for printing symbols, this pattern appears in a couple of places

xdsl/dialects/memref_stream.py
```diff
@@ -345,11 +345,13 @@ def print(self, printer: Printer):
             if self.patterns.data:
                 printer.print_string("\npatterns = [")
                 with printer.indented():
-                    printer.print_list(
-                        self.patterns.data,
-                        lambda attr: printer.print("\n", attr),
-                        delimiter=",",
-                    )
+                    if len(self.patterns.data):
```
alexarice: This pattern appeared in a couple of places, I think what I've done is semantics preserving.

xdsl/dialects/memref_stream.py
```diff
@@ -345,11 +345,13 @@ def print(self, printer: Printer):
...
+                        printer.print_string("\n")
+                        printer.print_list(
+                            self.patterns.data,
+                            lambda attr: printer.print_attribute(attr),
+                            delimiter=",\n",
+                        )
```
superlopuh: nit:
```suggestion
                        printer.print_list(
                            self.patterns.data,
                            printer.print_attribute,
                            delimiter=",\n",
                        )
```

xdsl/dialects/riscv_scf.py
```diff
@@ -298,18 +298,24 @@ def verify_(self):
...
+        printer.print_string(" = ")
+        printer.print_ssa_value(pair[1])
+
     def print(self, printer: Printer):
         printer.print_string(" (")
         block_args = self.before_region.block.args
         printer.print_list(
             zip(block_args, self.arguments, strict=True),
-            lambda pair: printer.print(pair[0], " = ", pair[1]),
+            lambda pair: self._print_pair(printer, pair),
```
alexarice, line 312: I moved some things to staticmethods instead of doing hacky multiline lambdas

xdsl/dialects/experimental/air.py
```diff
@@ -836,9 +836,9 @@ def __init__(self, async_dependencies: list[SSAValue] | None):
         super().__init__(operands=[async_dependencies], result_types=[AsyncTokenAttr()])
 
     def print(self, printer: Printer):
-        printer.print(" async ")
+        printer.print_string(" async ")
         if self.async_dependencies:
-            printer.print(self.async_dependencies, printer.print_ssa_value, ",")
+            printer.print_list(self.async_dependencies, printer.print_ssa_value)
```
alexarice, line 841: drive by fix

xdsl/dialects/snitch_stream.py
```diff
@@ -279,11 +279,13 @@ def print(self, printer: Printer):
...
-                    printer.print_list(
-                        self.stride_patterns.data,
-                        lambda attr: printer.print("\n", attr),
-                        delimiter=",",
-                    )
+                    if len(self.stride_patterns.data):
+                        printer.print_string("\n")
+                        printer.print_list(
+                            self.stride_patterns.data,
+                            lambda attr: printer.print_attribute(attr),
```
superlopuh: nit:
```suggestion
                            printer.print_attribute,
```

xdsl/dialects/memref_stream.py
superlopuh: Oh this should just be the following, no?
```suggestion
                    if self.patterns.data:
```

xdsl/dialects/snitch_stream.py
```diff
@@ -279,11 +279,13 @@ def print(self, printer: Printer):
             if self.stride_patterns.data:
                 printer.print_string("\npatterns = [")
                 with printer.indented():
-                    printer.print_list(
-                        self.stride_patterns.data,
-                        lambda attr: printer.print("\n", attr),
-                        delimiter=",",
-                    )
+                    if len(self.stride_patterns.data):
```
superlopuh: Same here
```suggestion
                    if self.stride_patterns.data:
```

xdsl/dialects/memref_stream.py
alexarice: I think so but I wasn't 100% sure. I'll change it.

PR #4544, opened 2025-06-16
dialects: (builtin) remove SymbolNameAttr

alexarice:
As far as I can tell this isn't actually part of MLIR's builtin dialect, and we don't use it anywhere.

### DIFF
--- a/xdsl/tools/xdsl_tblgen.py
+++ b/xdsl/tools/xdsl_tblgen.py
@@ -330,7 +330,7 @@ class TblgenLoader:
             case "StrAttr":
                 return "BaseAttr(StringAttr)"
             case "SymbolNameAttr":
-                return "BaseAttr(SymbolNameAttr)"
+                return "BaseAttr(StringAttr)"
             case "UnitAttr":
                 return "EqAttrConstraint(UnitAttr())"
             case _:
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -265,17 +265,6 @@ class BytesAttr(Data[bytes], BuiltinAttribute):
 
 
 @irdl_attr_definition
-class SymbolNameAttr(ParametrizedAttribute, BuiltinAttribute):
-    name = "symbol_name"
-    data: ParameterDef[StringAttr]
-
-    def __init__(self, data: str | StringAttr) -> None:
-        if isinstance(data, str):
-            data = StringAttr(data)
-        super().__init__([data])
-
-
-@irdl_attr_definition
 class SymbolRefAttr(ParametrizedAttribute, BuiltinAttribute):
     name = "symbol_ref"
     root_reference: ParameterDef[StringAttr]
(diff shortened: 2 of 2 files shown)


xdsl/tools/xdsl_tblgen.py
```diff
@@ -330,7 +330,7 @@ def _resolve_prop_constraint(self, rec: TblgenRecord | str) -> str:
             case "StrAttr":
                 return "BaseAttr(StringAttr)"
             case "SymbolNameAttr":
```
superlopuh, line 332: why keep this?

superlopuh: Oh this is the tblgen... What is it in MLIR then?

alexarice: As far as I can tell it's exactly the same as `StrAttr` but maybe a tablegen expert could chime in. My understanding is that it is effectively a constraint rather than a new attribute definition.

alexarice: https://github.com/llvm/llvm-project/blob/299a55a88fae4fc423c440436b2632d2a6bd800a/mlir/include/mlir/IR/CommonAttrConstraints.td#L348 relevant definition

superlopuh: I think this is exactly the pattern you were talking about in the other PR: [op](https://github.com/llvm/llvm-project/blob/main/mlir/test/lib/Dialect/Test/TestOpsSyntax.td#L193-L197), [test](https://github.com/llvm/llvm-project/blob/main/mlir/test/mlir-tblgen/op-format.mlir#L29-L31).

alexarice: very interesting, maybe this and the other PR should wait until we know how this actually works.

superlopuh: That was my first thought but actually now I think that we can use the helper you added in the other PR for implementing SymbolNameAttr so it should be good to go anyway.

alexarice: My feeling is that if we can figure out `SymbolNameAttr` then adding a new public method to `Printer` shouldn't be necessary

math-fehr: Tablegen expert hat on:
`SymbolNameAttr` is literally just a `StringAttr`, but that gets printed as a symbol (like `@name`).
I think it also adds a method as well, but I may be wrong on that.

alexarice: Right, but this shouldn't be a separate attribute with a string parameter right? I feel this lines up with my current understanding that `SymbolNameAttr` is actually a constraint on an property field of an operation rather than an actual attribute.

superlopuh:
Let's close this while we figure out a plan?

alexarice:
You can close it if you like, but I still feel this change is correct, and haven't really seen anything to suggest otherwise.

superlopuh:
I just feel like we might as well implement your constraint suggestion instead

alexarice:
I agree, but that still involves removing this attribute

alexarice:
@math-fehr are you happy for this to be merged?

math-fehr (approved):
Yes sorry!

PR #4545, opened 2025-06-16
core: (Printer) add helper for symbol name printing

alexarice:
Adds a short helper for printing symbol names.

This could probably be rebased in front of variadic printing if that would be preferable.


alexarice:
I destacked it

xdsl/printer.py
```diff
@@ -964,3 +964,7 @@ def print_metadata(self, dialects: Iterable[Dialect]) -> None:
 
         self.print_string("  }\n")
         self.print_string("#-}")
+
+    def print_symbol_name(self, sym_name: StringAttr):
+        self.print_string("@")
+        self.print_string(sym_name.data)
```
superlopuh: Worth a separate test for new public function, IMO, as well as a doc string. OTOH I'm not entirely convinced that this is better than improving the SymbolNameAttr situation, which I think was originally meant for this purpose.

superlopuh: I guess we can always use this helper in that class implementation. I would just take a string here instead of a StringAttr.

alexarice: I feel I agree, but am also unsure how the mechanism used by MLIR works/how we would use this in xDSL

superlopuh: I actually kind of like it, feels semantically important, especially since we presumably want to handle spaces in strings etc properly. I know that we don't today, but by using a helper like this it increases the chances of the fixes being reused.

alexarice: I'm not sure I understand the points about spaces or why we necessarily want this to take a str instead of a StringAttr

alexarice:
```
> cat ../mlir/test.mlir
"builtin.module"() ({
  "func.func"() <{function_type = () -> (), sym_name = "@test"}> ({
    "func.return"() : () -> ()
  }) : () -> ()
}) : () -> ()
> mlir-opt ../mlir/test.mlir --allow-unregistered-dialect
module {
  func.func @"@test"() {
    return
  }
}
```
Symbol name printing seems slightly more difficult than I anticipated, when I get some time I'll dig into the cpp code and try to work out the correct behaviour for this function. Seems like a good opportunity to get it right. (Maybe we have code for `func.func` that already gets this right?)

superlopuh:
`print_identifier_or_string_literal` is relevant here, it seems that print_symbol should call that. Note that it takes a `str`.

superlopuh (approved):

PR #4546, opened 2025-06-16
core: (Parser) move Delimiter to separate module

alexarice:
I wish to use `Delimiter` in the printer in the next PR, and moving this class out of `Parser` creates an insane diff so I thought I'd factor this change out.

`utils` was the first place that came to mind but I'm happy to accept different suggestions.


superlopuh:
What's the motivation ?

alexarice:
#4547 is the motivation

alexarice:
I think this has been vetoed, I'll rework #4547

PR #4547, opened 2025-06-16
core: (Printer) add delimited

alexarice:
It's always annoyed me that there is `in_angle_brackets` but not the other corresponding helpers. I started trying to remedy this and created a huge diff by moving `Delimiter`, so have stacked this on #4546 

I think this will help with moving various `print` constructions to `print_string`.

### DIFF
--- a/xdsl/dialects/stim/stim_printer_parser.py
+++ b/xdsl/dialects/stim/stim_printer_parser.py
@@ -4,7 +4,6 @@ Full documentation can be found here: https://github.com/quantumlib/Stim/blob/ma
 """
 
 import abc
-from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import cast
 
@@ -15,23 +14,12 @@ from xdsl.utils.base_printer import BasePrinter
 
 @dataclass(eq=False, repr=False)
 class StimPrinter(BasePrinter):
-    @contextmanager
-    def in_braces(self):
-        self.print_string("{")
-        yield
-        self.print_string("}")
-
-    @contextmanager
-    def in_parens(self):
-        self.print_string("(")
-        yield
-        self.print_string(") ")
-
     def print_attribute(self, attribute: Attribute) -> None:
         if isinstance(attribute, ArrayAttr):
             attribute = cast(ArrayAttr[Attribute], attribute)
             with self.in_parens():
                 self.print_list(attribute, self.print_attribute)
+            self.print_string(" ")
             return
         if isinstance(attribute, FloatData):
             self.print_string(f"{attribute.data}")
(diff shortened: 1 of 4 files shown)


xdsl/dialects/stim/stim_printer_parser.py
```diff
@@ -15,23 +14,12 @@
...
-        self.print_string("(")
-        yield
-        self.print_string(") ")
-
     def print_attribute(self, attribute: Attribute) -> None:
         if isinstance(attribute, ArrayAttr):
             attribute = cast(ArrayAttr[Attribute], attribute)
             with self.in_parens():
                 self.print_list(attribute, self.print_attribute)
+            self.print_string(" ")
```
alexarice, line 22: Needed due to the extra space in the original `StimPrinter.in_parens`

xdsl/utils/base_printer.py
```diff
@@ -68,6 +70,27 @@ def print_list(
                 self.print_string(delimiter)
             print_fn(elem)
 
+    @contextmanager
+    def delimited(self, delimiter: Delimiter):
```
superlopuh: Could actually pass strings here instead, and avoid the delimiter API move?

alexarice: We could, but `Delimiter` really feels like the correct type to use here (at least to me). I feel there would be no discussion if it wasn't embedded in `Parser`

alexarice: If the api move is just too nasty then I'll accept doing something else

alexarice: Another option is to not have `delimited` but just have the separate functions for each delimiter type.

superlopuh: I'm just not entirely sure that it's pulling its weight TBH, we do all this switching on enums instead of doing the checking at pyright-time, the function taking delimiters could accept only the same strings and that would be just as safe.

superlopuh: Yeah thinking more about this, I would much rather make the base functions here just take two strings instead of a Delimiter, which is tied to the MLIR tokenizer anyway. I would then add the helpers below to the MLIR-specific printer, and encourage explicit params in other printers. `with printer.delimited("(", ")")` just feels more readable and nice IMO.

alexarice: The `in_parens` and `in_braces` are from `StimPrinter`. Do you feel strongly about the helpers being in `Printer` instead of `BasePrinter`?

alexarice: I've implemented the changes (though left things in BasePrinter for now).

An interesting side effect is that you can now do things like `Printer.delimited("myattrname<", ">")`

superlopuh (approved):
I like this

PR #4554, opened 2025-06-17
testing: use match for VerifyException tests

superlopuh:
There are a few places where we check the error message directly instead of using pytest's match, this changes this for VerifyExceptions.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -899,14 +899,10 @@ def test_parse_number(
     ],
 )
 def test_parse_optional_builtin_int_or_float_attr(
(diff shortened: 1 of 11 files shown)


tests/test_parser.py
```diff
@@ -903,7 +903,7 @@ def test_parse_optional_builtin_int_or_float_attr(
 ):
     parser = Parser(Context(), text)
     if expected_value is None:
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Hello"):
```
alexarice: ???

superlopuh: Oh yeah

PR #4557, opened 2025-06-17
dialects: (math_xdsl) `math` dialect extension for constant values

jumerckx:
MLIR does not have an upstream dialect that defines commonly used math constants (e, pi, ...). This is nice to have for doing symbolic rewrites on those.
(We have encountered such rewrites for example when porting over Herbie rewrite rules: https://github.com/jumerckx/herbie-pdl/blob/0cd62e83ee1c4aa80c523982f0e02a65f04dd682/rules.rkt#L339C1-L344C1)

@superlopuh proposed having these in an extension dialect to `math`. This is an attempt at doing so.

### DIFF
--- a/xdsl/dialects/math_xdsl.py
+++ b/xdsl/dialects/math_xdsl.py
@@ -1,5 +1,11 @@
 """
 The math_xdsl dialect contains extensions to the math dialect.
+Currently, it only contains the `math_xdsl.constant` operation and
+a related attribute, which can be used to define and use
+mathematical constants in IR.
+
+The idea is to upstream these extension to the math dialect in MLIR
+at some point when they have matured.
 """
 
 from enum import auto


xdsl/dialects/math_xdsl.py
```diff
@@ -0,0 +1,83 @@
...
+        symbol = ConstantAttr(Constant(symbol_str))
+
+        parser.parse_punctuation(":")
+        result_type = parser.parse_type()
+
+        return ConstantOp(symbol, result_type)
+
+    def print(self, printer: Printer):
+        printer.print_string(f" {self.symbol.data} : ")
+        printer.print_attribute(self.value.type)
```
superlopuh: what happens with the assembly format?

jumerckx: It just gets rid of all the junk around the symbol.
`%2 = "math_xdsl.constant"() <{symbol = #math_xdsl<constant m_2_sqrtpi>}> : () -> f32`
->
`%2 = math_xdsl.constant m_2_sqrtpi : f32`

superlopuh: I mean if you specify an `assembly_format = "$symbol attr-dict : type($value)"`?

jumerckx: of course, that works! I was first looking into having unicode representations and then I forgot about assembly_format for this simple form.

xdsl/dialects/math_xdsl.py
```diff
@@ -0,0 +1,69 @@
+"""
+The math_xdsl dialect contains extensions to the math dialect.
+"""
```
superlopuh: Can you please expand on this, and say that the idea is to eventually propose operations and attributes in this dialect to MLIR?

superlopuh (approved):

xdsl/dialects/math_xdsl.py
superlopuh: https://xdsl--4557.org.readthedocs.build/4557/reference/dialects/math_xdsl/

PR #4558, opened 2025-06-17
misc: remake lockfile and update transitive dependencies

superlopuh:
We might want to re-enable the remake lockfile bot, as our lockfile drifted progresively out of date as Renovate updated only the direct dependency, and not the transitive ones.


superlopuh:
Note that this only affects developers of xDSL, and not the dependencies on pypi, which are handled by the pyproject.toml, which is unchanged.

docs/blocks.py
```diff
@@ -1,29 +1,30 @@
 import textwrap
 import xml.etree.ElementTree as etree
+from collections.abc import Callable
 from typing import Any
 
 from pymdownx.blocks import BlocksExtension  # pyright: ignore[reportMissingTypeStubs]
 from pymdownx.blocks.block import (  # pyright: ignore[reportMissingTypeStubs]
     Block,
-    type_string,  # pyright: ignore[reportUnknownVariableType]
-    type_string_in,  # pyright: ignore[reportUnknownVariableType]
```
superlopuh, line 9: I added types to this repo so now we can all relax

alexarice (approved):

PR #4562, opened 2025-06-18
misc: (Printer) remove print uses, 'easy cases'

alexarice:
Removes a bunch of uses of `Printer.print`. This isn't complete but I thought doing this in batches would help the reviewing process.

Only non trivial thing I've done here is use some `printer.in_parens()` when possible.

### DIFF
--- a/xdsl/dialects/func.py
+++ b/xdsl/dialects/func.py
@@ -189,7 +189,8 @@ class FuncOp(IRDLOperation):
     def print(self, printer: Printer):
         if self.sym_visibility:
             visibility = self.sym_visibility.data
-            printer.print(f" {visibility}")
+            printer.print_string(" ")
+            printer.print_string(visibility)
 
         print_func_op_like(
             printer,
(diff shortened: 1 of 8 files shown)


xdsl/dialects/func.py
```diff
@@ -189,7 +189,8 @@ def parse(cls, parser: Parser) -> FuncOp:
     def print(self, printer: Printer):
         if self.sym_visibility:
             visibility = self.sym_visibility.data
-            printer.print(f" {visibility}")
+            printer.print_string(" ")
+            printer.print_string(visibility)
```
AntonLydike, line 193: Any reason why this isn't `printer.print_string(f" {visibility}")`?

xdsl/dialects/riscv_func.py
```diff
@@ -209,7 +209,8 @@ def parse(cls, parser: Parser) -> FuncOp:
     def print(self, printer: Printer):
         if self.sym_visibility:
             visibility = self.sym_visibility.data
-            printer.print(f" {visibility}")
+            printer.print_string(" ")
+            printer.print_string(visibility)
```
AntonLydike, line 213: same here

xdsl/dialects/func.py
alexarice: In my head two print_strings is simpler than a format string, as the format string is equivalent to `" " + str(visibility)` whereas this ensures that `visibility` is already a string

superlopuh: I would definitely fold into one string as a general pattern, to minimise the constant overheads of string printing

superlopuh (approved):

PR #3674, opened 2024-12-24
[WIP]: Proof of concept going from jax to riscv

mamanain:

PR #4564, opened 2025-06-18
frontend: Jax to linalg test

mamanain:
Simple test to check whether the import works as expected.

### DIFF
--- a/tests/frontend/test_jax_stablehlo.py
+++ b/tests/frontend/test_jax_stablehlo.py
@@ -1,0 +1,70 @@
+import numpy as np
+import pytest
+
+try:
+    import jax
+    import jax.numpy as jnp
+    from jax._src.interpreters import mlir
+    from jaxlib.mlir.dialects import (
+        stablehlo,  # pyright: ignore[reportMissingTypeStubs]
+    )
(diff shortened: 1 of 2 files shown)


tests/frontend/test_jax_stablehlo.py
superlopuh, line 1: I would actually propose merging this at the same time as the donate arguments PR, more as an integration test, to make sure that we can get the linalg on memrefs at the end, which is what we really care about in our microkernel compilation projects.

mamanain: But donate [arguments](https://github.com/xdslproject/xdsl/blob/main/xdsl/transforms/jax_use_donated_arguments.py) pr is already merged.

So you want to bring it into the test as well?

superlopuh: Oh sorry, I thought because #3674 was open and had the pass in the diff that wasn't yet merged. Let's close that PR and iterate here? Yep, I think it would be great as an additional test, so we have one test to check for linalg extraction, and another to check that we can go all the way to memref with the input reused.

superlopuh:
Let's merge this to the xdsl-jax repo once that's created later this year

PR #4565, opened 2025-06-18
dialects: (linalg) improve printing/verification of transpose and broadcast

alexarice:
This was the tricky `printer.print` case in linalg. I'm happy to split this into smaller changes if wanted.

I'm also unsure if `printer.print_string(str(x))` is how we want to be printing integers, but we don't seem to have a different function (there is `printer.print_float` for example)

### DIFF
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -876,7 +876,7 @@ class TransposeOp(IRDLOperation):
 
     hidden_region = region_def("single_block")
 
-    permutation = prop_def(DenseArrayBase)
+    permutation = prop_def(DenseArrayBase.constr(i64))
 
     def __init__(
         self,
@@ -931,19 +931,22 @@ class TransposeOp(IRDLOperation):
                 )
 
     def print(self, printer: Printer):
-        printer.print_string(" ins(")
-        printer.print(self.input)
-        printer.print_string(":")
-        printer.print(self.input.type)
-        printer.print_string(")")
-        printer.print_string(" outs(")
-        printer.print(self.init)
-        printer.print_string(":")
-        printer.print(self.init.type)
-        printer.print_string(") ")
-        printer.print_string("permutation")
-        printer.print_string(" = ")
-        printer.print(list(self.permutation.get_values()))
+        printer.print_string(" ins")
+        with printer.in_parens():
+            printer.print_ssa_value(self.input)
+            printer.print_string(":")
+            printer.print_attribute(self.input.type)
+        printer.print_string(" outs")
+        with printer.in_parens():
+            printer.print_ssa_value(self.init)
+            printer.print_string(":")
+            printer.print_attribute(self.init.type)
+        printer.print_string(" permutation = ")
+        with printer.in_square_brackets():
+            printer.print_list(
+                self.permutation.get_values(),
+                lambda x: printer.print_string(str(x)),
+            )
 
     @classmethod
     def parse(cls, parser: Parser) -> Self:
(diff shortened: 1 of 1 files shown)


xdsl/irdl/attributes.py
```diff
@@ -280,7 +280,7 @@ def irdl_attr_definition(cls: TypeAttributeInvT) -> TypeAttributeInvT:
 
 IRDLGenericAttrConstraint: TypeAlias = (
     GenericAttrConstraint[AttributeInvT]
-    | Attribute
+    | AttributeInvT
```
superlopuh: this definitely feels like a separate change to me, what's the advantage here?

xdsl/dialects/builtin.py
```diff
@@ -1691,6 +1697,24 @@ def iter_values(self) -> Iterator[float] | Iterator[int]:
...
+
+    def get_float_values(self) -> Sequence[float]:
+        """
+        Return all the values of the elements in this DenseArray,
+        checking that the elements are floats.
+        """
+        el_type = self.get_element_type()
+        assert isinstance(el_type, AnyFloat), el_type
+        return el_type.unpack(self.data.data, len(self))
+
```
superlopuh: I would really like to avoid this. Surely we can print an array of floats and ints if we know the type without parsing the type? Seems like a similar mechanism to floats and ints of a known type in the assembly format.

xdsl/irdl/attributes.py
superlopuh: If it's for the constr then let's add a separate test for this?

xdsl/dialects/builtin.py
alexarice: I don't understand this point, these just make the typing more precise. Ideally it would be overloads on `get_values` but this relies on a pyright bug being fixed, same as for DenseIntOrFPElementsAttr

xdsl/irdl/attributes.py
alexarice: The issue is that without this change we have 
```
DenseArrayBase.constr(i64) : GenericAttrConstraint[DenseArrayBase[IntegerType | AnyFloat]]
```

alexarice: You can fix this by using `EqAttrConstraint` but it seems like an oversight in the definition of `IRDLGenericAttrConstraint` to me. I can move this to a separate change though I'm not sure how to test it.

superlopuh: Sorry, I don't mean testing this, I mean testing the constr like the other tests in test_builtin. I think this would make sense in a separate PR.

alexarice: fair enough

alexarice: This change isn't needed for the `constr` definition though

superlopuh: BTW I'm playing with this locally and it works just as well with `DenseArrayBase[I64]`

superlopuh: And the pyright bug seems fixed...

alexarice: > BTW I'm playing with this locally and it works just as well with `DenseArrayBase[I64]`

We want to move away from `Annotated` though?

alexarice: > And the pyright bug seems fixed...

The overloading one? I will investigate immediately

xdsl/dialects/builtin.py
alexarice: This isn't strictly needed for this change, so I'll remove it for now.

PR #4575, opened 2025-06-18
dialects: (builtin) deprecate DenseArrayBase.create_dense_*

superlopuh:
Also drop support for Sequence of attributes, which is not used, and makes things worse IMO.


alexarice:
I think I'd like to see what the corresponding change is for `DenseIntOrFPElementsAttr` first? Are you working on this or should I give it a stab?

superlopuh:
Oh I was going to merge my current stack of PRs first for DenseArrayBase then think about other things. Feel free to look at that yourself, there's no rush on my end.

alexarice:
I'm having a look, I'd like the same interface for both attributes preferably so would like to see if the same thing is possible for both

PR #4576, opened 2025-06-18
misc: remove TypeVar from list of allowed values in IRDLGenericAttrConstraint

superlopuh:
With TypeForm, we already handle the case where the TypeVar is a type, it turns out that there were a bunch of uses of this in a context where the TypeVar used was not a generic parameter of the operation, which feels like a bug to me.

### DIFF
--- a/xdsl/irdl/attributes.py
+++ b/xdsl/irdl/attributes.py
@@ -284,7 +284,6 @@ IRDLGenericAttrConstraint: TypeAlias = (
     | type[AttributeInvT]
     | "TypeForm[AttributeInvT]"
     | ConstraintVar
-    | TypeVar
 )
 """
 Attribute constraints represented using the IRDL python frontend. Attribute constraints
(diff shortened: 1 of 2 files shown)


xdsl/irdl/attributes.py
```diff
@@ -353,7 +352,7 @@ def irdl_to_attr_constraint(
 
 @overload
 def irdl_to_attr_constraint(
-    irdl: Attribute | TypeVar | ConstraintVar,
+    irdl: Attribute | ConstraintVar,
     *,
     allow_type_var: bool = False,
     type_var_mapping: dict[TypeVar, AttrConstraint] | None = None,
```
alexarice, line 358: Are these overloads still needed since my earlier change?

alexarice (approved):

xdsl/irdl/attributes.py
superlopuh: I'll remove the overload in the next PR, this one is the easy one. I'll open the next one as a stacked PR for @math-fehr to look at both before merging, seems like a delicate-enough part of the API.

math-fehr (approved):

PR #4577, opened 2025-06-18
dialects: (builtin) undeprecate DenseIntOrFPElementsAttr.from_list

alexarice:
Trying to mimic the interface for `DenseArrayBase` proposed by @superlopuh . Everything seems to work straight away except:
- There was one case where FloatAttrs were actually used in csl
- I can't quite seem to get the typing for ComplexTypes right. It currently accepts a `ComplexType[AnyFloat] | ComplexType[IntegerType]` instead of `ComplexType`. I can fix this by adding another overload to ComplexType.pack, but this overload doesn't seem sound to me.

If we want to go ahead with this then my suggestion is that we leave deprecation of create_dense_* to a second PR.

cc @erick-xanadu for complex stuff.

### DIFF
--- a/tests/dialects/test_builtin.py
+++ b/tests/dialects/test_builtin.py
@@ -379,20 +379,6 @@ def test_DenseIntOrFPElementsAttr_fp_type_conversion():
     assert isinstance(value2, float)
     assert value2 == 5.0
 
-    t1 = FloatAttr(4.0, f32)
-    t2 = FloatAttr(5.0, f32)
-
-    check2 = DenseIntOrFPElementsAttr.create_dense_float(TensorType(f32, [2]), [t1, t2])
-
-    value3 = check2.get_attrs()[0].value.data
-    value4 = check2.get_attrs()[1].value.data
-
-    # Ensure type conversion happened properly during attribute construction.
-    assert isinstance(value3, float)
-    assert value3 == 4.0
-    assert isinstance(value4, float)
-    assert value4 == 5.0
-
 
 def test_DenseIntOrFPElementsAttr_splat():
     attr_int = DenseIntOrFPElementsAttr.create_dense_int(TensorType(i64, [3]), 4)
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -2522,81 +2522,26 @@ class DenseIntOrFPElementsAttr(
     @deprecated("Please use `create_dense_int` instead.")
     def create_dense_index(
         type: RankedStructure[IndexType],
-        data: Sequence[int] | Sequence[IntegerAttr[IndexType]],
+        data: Sequence[int],
     ) -> DenseIntOrFPElementsAttr[IndexType]:
         return DenseIntOrFPElementsAttr.create_dense_int(type, data)
 
     @staticmethod
     def create_dense_int(
-        type: RankedStructure[_IntegerAttrType],
-        data: int
-        | IntegerAttr[_IntegerAttrType]
-        | Sequence[int]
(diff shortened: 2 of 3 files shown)


xdsl/dialects/builtin.py
```diff
@@ -2622,109 +2567,58 @@ def create_dense_complex(
...
-            if len(new_data) == 1 and prod(shape) != 1:
-                new_data = new_data[0]
-            return DenseIntOrFPElementsAttr.create_dense_int(
-                TensorType(data_type, shape), new_data
-            )
+        # Normalise ints
+        if isinstance(t := type.get_element_type(), IntegerType):
+            data = tuple(t.get_normalized_value(cast(int, x)) for x in data)
+
+        b = type.element_type.pack(data)  # pyright: ignore[reportArgumentType]
```
alexarice, line 2615: I *think* this is fine due to the overloads

superlopuh: I have been relying on this, I think this is fine.

xdsl/dialects/builtin.py
```diff
@@ -2622,109 +2567,58 @@ def create_dense_complex(
...
-            | RankedStructure[IndexType]
-        ),
-        data: (
-            Sequence[int]
-            | Sequence[IntegerAttr[IndexType]]
-            | Sequence[IntegerAttr[IntegerType]]
-        ),
-    ) -> DenseIntOrFPElementsAttr: ...
+        type: (RankedStructure[_FloatAttrTypeInvT]),
+        data: (Sequence[float]),
```
superlopuh:
```suggestion
        data: Sequence[float],
```

xdsl/dialects/builtin.py
alexarice: In any case, the previous iteration of from_list was a bunch of casts

alexarice:
Any thoughts on what to do with complex dense elements @superlopuh? Is it fine that this function takes `ComplexType[IntegerType] | ComplexType[AnyFloat]` instead of `ComplexType`? I could probably make `create_dense_complex` call `from_list` with an ignore/cast (it already has an ignore/cast so maybe this isn't so bad?

superlopuh:
In my opinion it's fine

tests/dialects/test_builtin.py
```diff
@@ -379,20 +379,6 @@ def test_DenseIntOrFPElementsAttr_fp_type_conversion():
...
-
-    value3 = check2.get_attrs()[0].value.data
-    value4 = check2.get_attrs()[1].value.data
-
-    # Ensure type conversion happened properly during attribute construction.
-    assert isinstance(value3, float)
-    assert value3 == 4.0
-    assert isinstance(value4, float)
-    assert value4 == 5.0
-
```
superlopuh, line 382: why delete?

xdsl/transforms/csl_stencil_set_global_coeffs.py
superlopuh, line 1: could merge this separately in a tiny PR?

superlopuh (commented):
My current opinion is that we should just deprecate the other functions and migrate to from_list in one go

tests/dialects/test_builtin.py
alexarice: Because this tests the "creating DenseIntOrFPElementsAttr from FloatAttrs" which this removes

alexarice:
I'm happy to deprecate the other functions but that will be a big diff so would prefer to do it in multiple PRs

xdsl/transforms/csl_stencil_set_global_coeffs.py
alexarice: Sure, will make now

PR #4580, opened 2025-06-19
core: Ensure parent OpTraits is not modified

math-fehr:
Copy OpTraits when declaring a new operation.
This will ensure that modifying the traits of an operation will not modify the parent
traits of an operation.

### DIFF
--- a/tests/test_traits.py
+++ b/tests/test_traits.py
@@ -1001,3 +1001,29 @@ def test_modify_traits():
 
     with pytest.raises(VerifyException, match="Nope"):
         op.verify()
+
+
+def test_modify_parent_traits():
+    """
+    Test that adding traits do not modify parent traits when they are inherited.
+    """
+
+    class ParentOp(IRDLOperation, ABC):
+        name = "test.parent"
+
+        traits = traits_def(LargerOperandTrait())
+
+    @irdl_op_definition
+    class ModifiedOp(ParentOp):
+        name = "test.modified"
+
+    @irdl_op_definition
+    class NotModifiedOp(ParentOp):
+        name = "test.not_modified"
+
+    ModifiedOp.traits.add_trait(BitwidthSumLessThanTrait(64))
+
+    assert ModifiedOp.has_trait(LargerOperandTrait)
+    assert ModifiedOp.has_trait(BitwidthSumLessThanTrait(64))
+    assert NotModifiedOp.has_trait(LargerOperandTrait)
+    assert not NotModifiedOp.has_trait(BitwidthSumLessThanTrait(64))
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -795,6 +795,10 @@ class OpTraits(Iterable[OpTrait]):
     ) -> None:
         self._traits = traits
 
+    def copy(self) -> OpTraits:
+        """Returns a copy of this instance."""
+        return OpTraits(self._traits)
+
     @property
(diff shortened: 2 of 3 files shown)


superlopuh (approved):

tests/test_traits.py
```diff
@@ -1001,3 +1001,29 @@ def test_modify_traits():
...
+
+    @irdl_op_definition
+    class ModifiedOp(ParentOp):
+        name = "test.modified"
+
+    @irdl_op_definition
+    class NotModifiedOp(ParentOp):
+        name = "test.not_modified"
+
+    ModifiedOp.traits.add_trait(BitwidthSumLessThanTrait(64))
```
superlopuh, line 1024: Random thought, should this be the API instead?
```suggestion
    ModifiedOp.add_trait(BitwidthSumLessThanTrait(64))
```

alexarice: I agree `traits.add_trait` seems kind of redundant?

alexarice (commented):
I agree this is a problem but wonder if this is the correct approach, as it feels like we're adding a copy to every operation definition. An alternative could be to have `OpTraits` actually be immutable, and then to replace the whole `OpTraits` object whenever `add_trait` is called, which to me seems slightly cleaner?

superlopuh:
That was also my first thought, seems like it might as well be frozen (modulo cheating with lazy instantiation that should not be exposed to the user IMO).

superlopuh:
One consideration for this, is that we may still need to be careful when subclassing:

```py
from typing import ClassVar

class Hello:
    bla: ClassVar[int] = 42

class World(Hello):
    ...

print(Hello.bla) # 42
print(World.bla) # 42

Hello.bla = 100

print(Hello.bla) # 100
print(World.bla) # 100
```

math-fehr:
That was actually my first implementation.
The problem is that the traits are also put in `OpDef`, so if we mutate them they get mutated in `OpDef` as well.
If I would overwrite the traits, I would need to also overwrite the ones in `OpDef`, or to remove the traits from `OpDef`. I didn't know if we actually use the ones in `OpDef`, so I just decided to do this copy instead.

If you want I can pass more time thinking of this, and maybe removing the traits from `OpDef`? I just needed a quick solution because it prevented me from doing some refactor on another project. But since I found a workaround I can take more time.

math-fehr:
Also yeah, we should disable adding a trait to a parent operation because we don't handle that.

alexarice:
Can we store traits only in OpDef and have the relevant methods on operations get them from there? Feel weird that `OpDef` is mutable

alexarice:
(I guess it's mutable anyway with the overwriting)

superlopuh:
I would be happy to first merge this because it seems like the current implementation is a bug, and then consider changing implementations later. I would probably move the API to Operation though, if you are both happy with that.

alexarice:
I'm happy with whatever (especially as I don't ever see myself using this feature), but thought I'd throw the alternative out.

superlopuh:
@math-fehr what's the latest on this?

math-fehr:
I'll need a bit more time this week, but I'd like to finish it properly instead of merging this as is!

superlopuh:
@math-fehr Will you have time to iterate next week? I'd love to reduce the number of open PRs

superlopuh:
Ping

math-fehr:
Abandoning this for now since I am prioritizing other things

PR #4586, opened 2025-06-19
dependencies: update dependency to LLVM 20.1.7

superlopuh:
This also updates the docker image ƒrom a 10GB one to a 2GB one, so should make our MLIR CI time much faster.

### DIFF
--- a/.github/workflows/ci-mlir.yml
+++ b/.github/workflows/ci-mlir.yml
@@ -15,7 +15,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     # https://github.com/xdslproject/llvm-docker/pkgs/container/llvm
-    container: ghcr.io/xdslproject/llvm:20.1.1
+    container: ghcr.io/xdslproject/llvm:20.1.7
     strategy:
       matrix:
         python-version: ['3.10', '3.11', '3.12']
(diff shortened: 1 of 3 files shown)


superlopuh:
50 seconds faster to initialize containers!

alexarice:
https://github.com/NixOS/nixpkgs/pull/414615 llvm package is lagging behind so I'll just manually update that once that PR works its way through.

.github/workflows/ci-mlir.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     # https://github.com/xdslproject/llvm-docker/pkgs/container/llvm
-    container: ghcr.io/xdslproject/llvm:20.1.1
+    container: ghcr.io/xdslproject/llvm:20.1.7
     strategy:
       matrix:
         python-version: ['3.10', '3.11', '3.12']
```
alexarice, line 21: Aside: why are we not including '3.13'? Is it not officially released yet?

superlopuh: It's been released for like a year, we should probably include it: https://devguide.python.org/versions/

math-fehr (approved):

PR #4590, opened 2025-06-19
core: (Printer) add print_int

alexarice:
Series of PRs to try to improve standardisation of print_int and print_float
Stacked on #4589 

Adds a `print_int` function similar to `print_float`.

This deals with some tricky corners, such as `i1` being printed as `true/false` and that python booleans are apparently also python integers for function calls. This replaces the need for `type.print_without_type` that was on `IntegerType` and `IndexType`.

Also fixes a bug that `si1` and `ui1` were being printed as `true/false` whereas in mlir this is only true for `i1` (as far as I can observe anyway)

An alternative to this is to printing Bools as integers is to assert that the input is actually an integer. I'm not sure which is better.


xdsl/printer.py
```diff
@@ -380,6 +382,19 @@ def print_float(self, value: float, type: AnyFloat):
...
+            and type.width.data == 1
+            and type.signedness.data == Signedness.SIGNLESS
+        ):
+            if value:
+                self.print_string("true")
+            else:
+                self.print_string("false")
+        else:
+            self.print_string(f"{value:d}")
+
```
math-fehr: I think I would maybe rather just expose `print_int` without a type?
If we want to print an integer with a type, that usually means we already have an `IntegerAttr`, so we would use that.
Otherwise, we usually don't have a type? So I would just expose the `print_int` function without the type and move the "true"/"false" logic to `IntegerAttr`, what do you think?

math-fehr: From the next PR, it seems I am mistaken.
My question is then why we would not print `true` or `false` with `Signedness.SIGN` and `Signedness.UNSIGNED`? (Maybe that's an MLIR thing).

I still think that maybe we should have a `print_int` and a `print_typed_int`?

alexarice: My aim was to try to unify the interfaces of `print_int` and `print_float`, though for `print_float` the type isn't optional anyways so maybe this doesn't work. As of this PR the printing for `IntegerAttr` uses `printer.print_int` (as I think it should). An alternative could be to restore the `type.print_without_type` functions but have these call (an untyped) `print_int`. What do you think?

alexarice: As far as I can tell, `true` in MLIR always refers to `-1 : i1`

math-fehr: I don't think we need `type.print_without_type` tbh.
I think that's okay then that you pass the attribute, however, I would add a documentation comment that actually explains why you need the attribute (for printing as boolean)

alexarice: Sure, it occurs to me that I could have just had `type == i1` as the check)

tests/test_printer.py
```diff
@@ -875,7 +915,7 @@ def _test_float_attr(value: float, type: AnyFloat):
 
         io_attr = StringIO()
         printer.stream = io_attr
-        printer.print_float_attr(FloatAttr(value, type))
+        FloatAttr(value, type).print_without_type(printer)
```
alexarice: I'm a bit confused because I think this happened in the last PR?

alexarice: Git being weird, I'll rebase

superlopuh (approved):

alexarice:
@math-fehr would you still prefer `print_int` and `print_typed_int`?

math-fehr:
I'll live without it!

math-fehr (approved):

PR #4591, opened 2025-06-19
core: (Printer) cleanup print_complex

alexarice:
Series of PRs to try to improve standardisation of print_int and print_float
Stacked on #4590 

Slight clean up of `print_complex`, removing unnecessary TypeVars, using new `in_parens` function, and using `print_int`


math-fehr (approved):
Pretty!

PR #4593, opened 2025-06-19
dialects: (builtin) simplify and fix array printing

alexarice:
Removes the old `str` based printing and uses the new helpers instead. My understanding is that this is the "correct" way to print floats.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1588,15 +1588,13 @@ class DenseArrayBase(
                 return
             data = self.iter_values()
             printer.print_string(": ")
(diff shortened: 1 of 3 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1588,15 +1588,13 @@ def print_builtin(self, printer: Printer):
                 return
             data = self.iter_values()
             printer.print_string(": ")
-            # There is a bug in MLIR which will segfault when parsing DenseArrayBase type i1 as 0 or 1,
-            # therefore we need to print these as false and true
-            if self.elt_type == i1:
+            elt_type: IntegerType | AnyFloat = self.elt_type
```
alexarice, line 1591: Inlining this alias makes pyright angry

PR #4594, opened 2025-06-20
dialects: Add smt.bv.constant constructor

math-fehr:


alexarice (approved):
Would like to point out that 
```python
BVConstantOp.from_value_and_type(42, 32)
```
is more verbose than
```python
BVConstantOp(BitVectorAttr(42, 32))
```

math-fehr:
But Sasha plan was to remove `BitVectorAttr(42, 32)` in favor of `BitVectorAttr(42, BitVectorType(32))` right?

alexarice:
I see

PR #4595, opened 2025-06-20
dialects: (builtin) tighten `get_{int/float}_values` typing

alexarice:
Removes some asserts and strictens the typing on these functions.


alexarice:
These remove tonnes of casts everywhere. I'd really prefer to have them

alexarice:
Seems the "2 overload" approach works, so I'll work on deprecating this.

PR #4597, opened 2025-06-20
misc: (Printer) remove uses of `str` while printing

alexarice:
I'm considering `printer.print_string(str(...))` to basically be an antipattern now that `printer.print_int` exists. The only two cases left are for printing `BoolAttr` (which seems to exist only in tests?)

### DIFF
--- a/xdsl/dialects/experimental/dmp.py
+++ b/xdsl/dialects/experimental/dmp.py
@@ -184,13 +184,19 @@ class ExchangeDeclarationAttr(ParametrizedAttribute):
         )
 
     def print_parameters(self, printer: Printer) -> None:
-        printer.print_string("<at [")
-        printer.print_list(self.offset, lambda x: printer.print_string(str(x)))
-        printer.print_string("] size [")
-        printer.print_list(self.size, lambda x: printer.print_string(str(x)))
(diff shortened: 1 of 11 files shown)


alexarice:
`AffineMap` also has some funny `str` printing going on

xdsl/dialects/experimental/dmp.py
```diff
@@ -184,13 +184,19 @@ def source_area(self) -> ExchangeDeclarationAttr:
...
-        printer.print_list(self.offset, lambda x: printer.print_string(str(x)))
-        printer.print_string("] size [")
-        printer.print_list(self.size, lambda x: printer.print_string(str(x)))
-        printer.print_string("] source offset [")
-        printer.print_list(self.source_offset, lambda x: printer.print_string(str(x)))
-        printer.print_string(f"] to {list(self.neighbor)}>")
+        with printer.in_angle_brackets():
+            printer.print_string("at ")
+            with printer.in_square_brackets():
+                printer.print_list(self.offset, lambda x: printer.print_int(x))
```
superlopuh: nit:
```suggestion
                printer.print_list(self.offset, printer.print_int)
```
and below

PR #4599, opened 2025-06-20
dialects: (builtin) deprecate DenseIntOrFPElementsAttr.get_{int/float}_values

alexarice:
This was my attempt. There are unfortunately a few cases where the type wasn't known, so I added a `get_untyped_values` function for these. I feel this function should be just another overload but I'm pretty sure this triggers the weird pyright behaviour (I'll have a quick check though)

### DIFF
--- a/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
+++ b/xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
@@ -274,13 +274,19 @@ class ConvertMemRefGlobalOp(RewritePattern):
                     raise DiagnosticException(
                         f"Unsupported memref element type for riscv lowering: {element_type}"
                     )
-                ints = initial_value.get_int_values()
+                ints = cast(
+                    DenseIntOrFPElementsAttr[IntegerType], initial_value
+                ).get_values()
                 ptr = TypedPtr.new_int32(ints).raw
             case Float32Type():
-                floats = initial_value.get_float_values()
+                floats = cast(
+                    DenseIntOrFPElementsAttr[Float32Type], initial_value
+                ).get_values()
                 ptr = TypedPtr.new_float32(floats).raw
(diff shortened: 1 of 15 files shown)


alexarice:
I've had a check and it seems things only get weird when you have a `DenseIntOrFPElementsAttr[Unknown]` or `DenseIntOrFPElementsAttr[Any]`. It is perhaps ok to be weird in these cases so I'll remove `get_values_untyped`

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -274,13 +274,19 @@ def match_and_rewrite(self, op: memref.GlobalOp, rewriter: PatternRewriter):
...
-                floats = initial_value.get_float_values()
+                floats = cast(
+                    DenseIntOrFPElementsAttr[Float32Type], initial_value
+                ).get_values()
                 ptr = TypedPtr.new_float32(floats).raw
             case Float64Type():
-                floats = initial_value.get_float_values()
+                floats = cast(
+                    DenseIntOrFPElementsAttr[Float64Type], initial_value
+                ).get_values()
```
superlopuh: hmm

alexarice: FWIW I had the same issue in the "strengthen typing" PR (now closed). I'll have a play around and see if I can refactor this.

alexarice: Fixed (Kind of) (maybe)

superlopuh: I found a way to delete all this code so let's merge this and I'll do the logic change in a follow-up PR.

superlopuh (approved):

PR #4602, opened 2025-06-20
core: (Printer) deprecate Printer.print

alexarice:


xdsl/printer.py
```diff
@@ -112,6 +115,13 @@ def print(self, *argv: Any) -> None:
             text = str(arg)
             self.print_string(text)
 
+    def print_module_op(self, module_op: ModuleOp) -> None:
+        """
+        Prints a module followed by a newline.
+        """
+        self.print_op(module_op)
+        self._print_new_line()
+
```
superlopuh, line 123: seems like a separate change IMO

alexarice: I was so close

xdsl/dialects/transform.py
```diff
@@ -821,7 +821,8 @@ def parse(cls, parser: Parser) -> NamedSequenceOp:
     def print(self, printer: Printer):
         if self.sym_visibility:
             visibility = self.sym_visibility.data
-            printer.print(f" {visibility}")
+            printer.print_string(" ")
+            printer.print_string(visibility)
```
superlopuh, line 825: why split?

alexarice: Same as other PRs, I feel format strings are bad here for the same reason as `str`

xdsl/printer.py
superlopuh: I'm actually wondering whether we could just get rid of the trailing newline, or make it explicit at the callsite. Do you think we benefit form a separate module op helper?

xdsl/dialects/transform.py
superlopuh: I don't feel fantastically strongly about this

superlopuh: But respectfully disagree :)

xdsl/printer.py
alexarice: We should probably just get rid of this but I wasn't 100% sure how to fix the notebook

superlopuh:
Oh no I'm sorry

superlopuh:
Wrong tab

superlopuh:
My sincerest apologies

PR #4604, opened 2025-06-20
core: (Printer) deprecate Printer.print

alexarice:
attempt 2

### DIFF
--- a/xdsl/printer.py
+++ b/xdsl/printer.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass, field
 from itertools import chain
 from typing import Any, cast
 
-from typing_extensions import TypeVar
+from typing_extensions import TypeVar, deprecated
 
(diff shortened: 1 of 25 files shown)


xdsl/printer.py
```diff
@@ -112,6 +114,13 @@ def print(self, *argv: Any) -> None:
             text = str(arg)
             self.print_string(text)
 
+    def print_module_op(self, module_op: ModuleOp) -> None:
```
superlopuh: I still think this belongs in a separate PR, but I also think that we would be better off not providing this helper, and getting people to print their own new lines if they feel like it.

alexarice: Sure, will try to fix

alexarice: I edited the output of the Toy notebook to not include the newline, rather than adding an explicit print statement. Hopefully this is reasonable

superlopuh: Yep that's great.

PR #4607, opened 2025-06-21
misc: remove some duplicate declarations

superlopuh:
Tiny cleanup PR, Pyright in Zed was complaining about this for some reason, technically each variable can only be declared once in a scope.


alexarice (approved):
I remain confused on how you get pyright to say different things between different editors

PR #4608, opened 2025-06-21
transformations: use DenseArrayBase when lowering memref to riscv

superlopuh:


xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -263,41 +266,14 @@ def match_and_rewrite(self, op: memref.GlobalOp, rewriter: PatternRewriter):
...
-                )
+        if len(initial_value.data.data) % (self.xlen // 8):
+            raise DiagnosticException(
+                f"memref.global value size not divisible by xlen ({self.xlen}) not yet "
+                "supported."
+            )
 
-        text = ",".join(hex(i) for i in ptr.int32.get_list(42))
+        i32_attr = DenseArrayBase(i32, initial_value.data)
+        text = ",".join(hex(i) for i in i32_attr.iter_values())
```
alexarice, line 276: Can we not use the original `DenseIntOrFPElementsAttr`?

alexarice: This all feels a bit cursed tbh

superlopuh: No, the point is that RISCV assembly expects comma-separated hex values of 32-bit integers, I could do this manually with the struct package on the original data, but this feels more convenient.

alexarice (approved):
The simplification seems reasonable. I don't know much about the riscv infrastructure so maybe better for someone else to comment on that (though I guess if the tests still work?)

superlopuh:
This is already tested in a few places, I'm happy to take the risk here.

PR #4609, opened 2025-06-21
CI: use setup-python instead of uv python install

superlopuh:
These guys recommend it for some reason, I guess it doesn't hurt.

https://codspeed.io/docs/benchmarks/python


superlopuh:
I'll just merge this, let me know if you'd like me to revert.

PR #4610, opened 2025-06-21
[WIP] bench: add Codspeed micro benchmarking

superlopuh:
This tool allows you to catch regressions, I think it would be nice to find some core bits of functionality that we don't want to get slower. This is separate from the xdsl-bench which tracks runs over time.


tests/test_ci_benchmarks.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+import pytest
+
+from xdsl.dialects.arith import ConstantOp
+
+
+@pytest.mark.benchmark
+def test_const_verify():
+    ConstantOp.from_int_and_width(1, 32).verify()
```
alexarice, line 15: Can we not reuse some of the `xdsl-bench` stuff, instead of basing performance regressions on a test which makes and verifies one operation?

Makefile
```diff
@@ -57,6 +57,11 @@ filecheck: uv-installed
 pytest: uv-installed
 	uv run pytest tests -W error -vv
 
+# run pytest tests
```
alexarice, line 60: Maybe `run microbenchmark tests` or something?

alexarice (commented):
What benefit does this have compared to the rest of the benchmarking infrastructure?

tests/test_ci_benchmarks.py
superlopuh: Yep definitely would be nice to keep them coupled in some way, I first wanted to see whether this would work at all.

superlopuh:
The idea is that this happens live in the CI, meaning that this has only the benchmarks which we want to fail CI if they regress beyond some threshold (currently set to 10%), so it should be way more conservative than the xdsl-bench tests which are run overnight and therefore can be slower, less critical, and more informative.

alexarice:
I think I'm generally against adding random tooling (for instance I find all the codecov things to be spam) but won't block this if others feel it is useful to them.

superlopuh:
I find the codecov thing to be very useful when reviewing, as it quickly lets me find untested bits of code, sometimes catching bugs.

math-fehr:
Yeah, I would also personally say that having benchmarking in the CI is a bad thing.
We know whenever we are touching core data structures, so we should just be mindful of that and benchmark it ourselves when we are touching sensitive parts.

superlopuh:
Why is it a bad thing? I don't get it, surely this is the way to make sure that some specific benchmarks never regress?

superlopuh:
Will close for now, and will reopen when we have an obvious benchmark that we want to always keep fast, currently don't have one. Would also be nice for it to trigger only for some PRs that are more likely to hit the sensitive areas, maybe based on tags or files affected.

PR #4612, opened 2025-06-23
backend: (x86) convert arith.addi to x86 instructions

qaco:


tests/filecheck/backend/x86/convert_arith_to_x86.mlir
```diff
@@ -0,0 +1,22 @@
...
+%i0 = "test.op"(): () -> i32
+%i1 = "test.op"(): () -> i32
+%i2 = arith.addi %i0,%i1: i32
+
+// CHECK:      builtin.module {
+// CHECK-NEXT:   %i0 = "test.op"() : () -> i32
+// CHECK-NEXT:   %i1 = "test.op"() : () -> i32
+// CHECK-NEXT:   %i2 = builtin.unrealized_conversion_cast %i0 : i32 to !x86.reg
+// CHECK-NEXT:   %i2_1 = builtin.unrealized_conversion_cast %i1 : i32 to !x86.reg
+// CHECK-NEXT:   %i2_2 = x86.ds.mov %i2_1 : (!x86.reg) -> !x86.reg
```
superlopuh: no need for mov

xdsl/backend/x86/lowering/convert_arith_to_x86.py
```diff
@@ -0,0 +1,57 @@
...
+
+@dataclass
+class ArithAddiToX86(RewritePattern):
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: arith.AddiOp, rewriter: PatternRewriter):
+        if isinstance(op.lhs.type, builtin.ShapedType):
+            raise DiagnosticException(
+                "Lowering of arith.addi not implemented for ShapedType"
+            )
+        lhs_cast_op, lhs_x86 = UnrealizedConversionCastOp.cast_one(
```
superlopuh: we have helpers for this in riscv, maybe worth adding some here?

superlopuh: Oh actually those could be moved to a target-independent location in `xdsl/backend/utils.py` since the logic is not dependent on the register type.

tests/filecheck/backend/x86/convert_arith_to_x86.mlir
qaco: Are you sure ? ```x86.rs.add``` changes the value of its operand, whereas ```arith.addi``` produces a new (SSA) value. How do you represent this discrepancy of semantics ?

superlopuh: Oh no I didn't realize, mov is fine then

xdsl/backend/x86/lowering/convert_arith_to_x86.py
qaco: What should this helper do exactly?

superlopuh: `cast_operands_to_regs`
Used like this:
```py
        lhs, rhs = cast_operands_to_regs(rewriter)
```

superlopuh: There are a few helpers in that file that could be reused for the other backends.

qaco: Ok cool ! And how do you paremeterize the ISA/register types ? Should it be a second argument of the function ?

superlopuh: I don't think we need to, I would just loosen the restriction from riscv register types to `RegisterType`, since they all have an unallocated register.

qaco: Sorry it's not clear to me. Inside the ```cast_operands_to_regs``` function, we cast a SSAValue to another SSAValue reprensenting a register. In my case this register needs to be of type x86 because it then feeds x86 ops, but in other cases it needs to be of type riscv. How does the function choose which one is the relevant type ?

superlopuh: Ah yes I see what you mean yep it should take a function `register_type_for_type` as argument I think

qaco: Ok done

xdsl/backend/utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+from xdsl.backend.register_type import RegisterType
+from xdsl.dialects.builtin import (
+    UnrealizedConversionCastOp,
+)
+from xdsl.ir import Attribute, SSAValue
+from xdsl.pattern_rewriter import PatternRewriter
+
+
+def cast_values_to_regs(
```
superlopuh: can you please call this function from the riscv function? It would be great not to duplicate functionality. Since you're in this space, could you please add a short doc string here?

xdsl/backend/x86/lowering/convert_arith_to_x86.py
```diff
@@ -0,0 +1,57 @@
...
+from xdsl.pattern_rewriter import (
+    GreedyRewritePatternApplier,
+    PatternRewriter,
+    PatternRewriteWalker,
+    RewritePattern,
+    op_type_rewrite_pattern,
+)
+from xdsl.utils.exceptions import DiagnosticException
+
+from .helpers import scalar_type_to_register_type
```
superlopuh: please use absolute imports

qaco: Ok done

xdsl/backend/utils.py
qaco: Ok done

xdsl/backend/riscv/lowering/utils.py
```diff
@@ -177,7 +156,9 @@ def cast_operands_to_regs(rewriter: PatternRewriter) -> list[SSAValue]:
     if the operands were not already int registers
     """
 
-    new_ops, new_operands = cast_ops_for_values(rewriter.current_operation.operands)
```
superlopuh, line 180: cast_ops_for_values is now unused, let's delete it

qaco: Ok done

PR #4623, opened 2025-06-24
core: add simplification for & and | on AnyAttr

superlopuh:
The beginning of some constraint algebra, this is the easiest case.

### DIFF
--- a/tests/irdl/test_attr_constraint.py
+++ b/tests/irdl/test_attr_constraint.py
@@ -18,6 +18,7 @@ from xdsl.ir import Attribute, Data, ParametrizedAttribute
 from xdsl.irdl import (
     AllOf,
     AnyAttr,
+    AnyOf,
     AttrConstraint,
     BaseAttr,
     ConstraintContext,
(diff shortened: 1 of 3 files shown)


alexarice (commented):
This seems reasonable, though I wonder where this appears. If feels like we never want to actually write:
```
AnyAttr() | base(IntegerAttr)
```
Is there a place this arises internally somewhere? (Basically it's not clear to me that this shouldn't be a warning instead of a simplification)

superlopuh:
Nowhere yet, but I would like to do a similar change of ParamAttrConstraint, where some fields may be AnyAttr, and this would simplify things in that case.

tests/irdl/test_attr_constraint.py
```diff
@@ -217,3 +218,23 @@ def test_memref_to_tensor(
...
+            BaseAttr(AttrA) | BaseAttr(AttrB) | BaseAttr(AttrC),
+            AnyOf((BaseAttr(AttrA), BaseAttr(AttrB), BaseAttr(AttrC))),
+        ),
+        (
+            # Note the [Attribute] to provide a supertype of AttrA and AttrB
+            BaseAttr[Attribute](AttrA) & BaseAttr(AttrB) & BaseAttr(AttrC),
+            AllOf((BaseAttr(AttrA), BaseAttr(AttrB), BaseAttr(AttrC))),
+        ),
+        (AnyAttr() | BaseAttr(AttrA), AnyAttr()),
+        (AnyAttr() & BaseAttr(AttrA), BaseAttr(AttrA)),
```
alexarice, line 237: Could you add the symmetric cases?

alexarice (approved):
Ok, that sounds reasonable to me. Will this be:
`ParamAttrConstraint(type, (AnyAttr(),)) -> BaseAttr(type)`?

tests/irdl/test_attr_constraint.py
superlopuh: I actually think I cannot, I guess we could add it to the base constraint class but then subclasses could override it and I'm not sure how to guarantee that it works in the general case. I guess I could still add it for these cases if you think it would be worth doing.

superlopuh:
More like `P(A, (1, AnyAttr())) | P(A, (AnyAttr(), 2))) == P(A, (1, 2))`

superlopuh:
I changed my mind and added the symmetric case

PR #4617, opened 2025-06-23
[WIP] frontend: (pyast) development tracking removing PyAST custom dialects

EdmundGoodman:
#4624 adds support for registering `IntegerAttr` types, for example:

```python
p = FrontendProgram()
p.register_type(IntegerAttr[I64], i64)
with CodeContext(p):
    def test_identity(a: IntegerAttr[I64]) -> IntegerAttr[I64]:
        return a
```

This PR extends this by adding support for registering functions on these `IntegerAttr` types, supporting the following snippet:

```python
p = FrontendProgram()
p.register_type(IntegerAttr[I64], i64)
p.register_function(IntegerAttr.__add__, arith.AddiOp)
with CodeContext(p):
    def test_addi_overload(
        a: IntegerAttr[I64], b: IntegerAttr[I64]
    ) -> IntegerAttr[I64]:
        return a + b
```

This then allows us to deprecate PyAST's custom `builtin` and `arith` dialects.

PR #4624, opened 2025-06-24
frontend: (pyast) Add support for registering IntegerAttr types

EdmundGoodman:
Add support for registering `IntegerAttr` types to express fixed bit-width integers in emitted MLIR, for example the following code should be valid:

```python
p = FrontendProgram()
p.register_type(IntegerAttr[I64], i64)
with CodeContext(p):
    def test_identity(a: IntegerAttr[I64]) -> IntegerAttr[I64]:
        return a
```

Peeled from #4617 to avoid complications with registering and defining functions on `IntegerAttr`s

### DIFF
--- a/xdsl/frontend/pyast/program.py
+++ b/xdsl/frontend/pyast/program.py
@@ -4,6 +4,8 @@ from dataclasses import dataclass, field
 from io import StringIO
 from typing import Any
 
+from typing_extensions import TypeForm
+
 from xdsl.dialects.builtin import ModuleOp
 from xdsl.frontend.pyast.code_generation import CodeGeneration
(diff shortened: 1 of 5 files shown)


xdsl/frontend/pyast/program.py
```diff
@@ -48,7 +50,7 @@ class FrontendProgram:
 
     def register_type(
         self,
-        source_type: type,
+        source_type: type | TypeForm[Attribute],
```
superlopuh, line 53: Is this necessary?

EdmundGoodman:
Closed, as TypeForms are no longer needed to support `c_int` style types

PR #4627, opened 2025-06-24
core: add simplification for | on BaseAttr and ParamAttrConstraint

superlopuh:
Some more folding of things.

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -157,14 +157,14 @@ class GenericAttrConstraint(Generic[AttributeCovT], ABC):
     def __or__(
         self, value: GenericAttrConstraint[_AttributeCovT], /
     ) -> GenericAttrConstraint[AttributeCovT | _AttributeCovT]:
-        if isinstance(value, AnyAttr):
+        if isinstance(value, AnyAttr) or self == value:
             return value  # pyright: ignore[reportReturnType]
         return AnyOf((self, value))
 
     def __and__(
         self, value: GenericAttrConstraint[AttributeCovT], /
     ) -> GenericAttrConstraint[AttributeCovT]:
-        if isinstance(value, AnyAttr):
+        if isinstance(value, AnyAttr) or self == value:
             return self
         return AllOf((self, value))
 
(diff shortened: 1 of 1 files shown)


xdsl/irdl/constraints.py
```diff
@@ -652,6 +660,20 @@ def mapping_type_vars(
             tuple(c.mapping_type_vars(type_var_mapping) for c in self.param_constrs),
         )
 
+    def __or__(self, value: GenericAttrConstraint[_AttributeCovT], /):
+        if (
+            not isinstance(value, ParamAttrConstraint)
+            or self.base_attr is not cast(ParamAttrConstraint[Any], value).base_attr
```
alexarice, line 659:
```suggestion
            or self.base_attr is not cast(ParamAttrConstraint[Attribute], value).base_attr
```
would this fix the need for the `ignore`?

tests/irdl/test_attr_constraint.py
```diff
@@ -237,6 +237,28 @@ def test_memref_to_tensor(
         (AnyAttr() & BaseAttr(AttrA), BaseAttr(AttrA)),
         # Note the [Attribute] to provide a supertype of AttrA and Attribute
         (BaseAttr[Attribute](AttrA) & AnyAttr(), BaseAttr(AttrA)),
+        (BaseAttr(AttrA) | BaseAttr(AttrA), BaseAttr(AttrA)),
```
alexarice, line 240: This is an instance of the more general: `A | A = A` for any constraint `A` right?

alexarice (approved):
PR title suggests there will be some `&` stuff, is this for a follow up PR?

tests/irdl/test_attr_constraint.py
superlopuh: Yep, do you think we should have this as a general case?

xdsl/irdl/constraints.py
superlopuh: Nope for whatever reason Pyright is not happy with Attribute:

```
Type "Attribute" cannot be assigned to type variable "ParametrizedAttributeCovT@ParamAttrConstraint"
  Type "Attribute" is not assignable to upper bound "ParametrizedAttribute" for type variable "ParametrizedAttributeCovT@ParamAttrConstraint"
    "Attribute" is not assignable to "ParametrizedAttribute"Pyright[reportInvalidTypeArguments](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInvalidTypeArguments)
```

tests/irdl/test_attr_constraint.py
alexarice: I don't see why not, seems like a simpler rule than the one for `BaseAttr` anyway

PR #4628, opened 2025-06-24
frontend: (pyast) Rewrite tests to use xDSL dialects as opposed to PyAST shims

EdmundGoodman:
Currently, PyAST uses custom shim dialects for type hinting, with an example from the tests:

```python
from xdsl.frontend.pyast.dialects.builtin import f16, f32, f64, i1, i32, i64, index
def foo(x: i32) -> i32
    return x
```

This PR modifies the tests to leverage the newly added support for using the builtin xDSL dialects instead, for example:

```python
from xdsl.dialects.builtin import I32, IntegerAttr, i32
def foo(x: IntegerAttr[I32]) -> IntegerAttr[I32]
    return x
```

This unblocks the removal of these custom shim dialects in a later PR, as they are no longer used in the codebase.

One test file `arith.py` is left unmodified, as it requires adding methods (such as `__add__`) to the xDSL dialects to implement its test cases. As such, adoption of the new annotations for `arith.py` will be implemented in a separate PR.


tests/filecheck/frontend/dialects/builtin.py
```diff
@@ -1,46 +1,75 @@
...
 from xdsl.frontend.pyast.program import FrontendProgram
 
+si32 = IntegerType(32, signedness=Signedness.SIGNED)
+si64 = IntegerType(64, signedness=Signedness.SIGNED)
+ui32 = IntegerType(32, signedness=Signedness.UNSIGNED)
+ui64 = IntegerType(64, signedness=Signedness.UNSIGNED)
+SI32 = Annotated[IntegerType, si32]
+SI64 = Annotated[IntegerType, si64]
+UI32 = Annotated[IntegerType, ui32]
+UI64 = Annotated[IntegerType, ui64]
```
EdmundGoodman, line 34: I have a branch adding this to `xdsl/dialects/builtin.py` where it probably belongs. I had a chat with Mathieu, who said it's not there because almost everyone just uses signless integers, but he wouldn't object to a PR including it. Any opinions on this?

alexarice: My understanding is that all our uses of `Annotated` were unsound. I would be hesitant before adding more functionality to xDSL that relies on its use, though I assume you have discussed this with both @math-fehr and @superlopuh?

EdmundGoodman:
Closed, as we will use `ctypes` instead of `IntegerAttr`

PR #4631, opened 2025-06-25
core: don't print IR twice on verification

superlopuh:
Operation.verify already does op.emit_error, this PR removes the nested calls.


tests/filecheck/dialects/smt/invalid.mlir
```diff
@@ -3,27 +3,21 @@
 %arg1 = smt.constant true
 
 %and = smt.or %arg1
-// CHECK: operand at position 0 does not verify:
```
superlopuh, line 6: this only passed because the `operand at position 0 does not verify:` was printed below so many times

PR #4632, opened 2025-06-25
misc: print single line at the end of Diagnostic

superlopuh:


superlopuh:
Sorry yep this is required for a downstream change but I'll add a test to make it clear

superlopuh:
OK turns out this isn't the functionality that affected my downstream changes, I'd still like to merge it separately if you're ok with it.

tests/pattern_rewriter/test_pattern_rewriter.py
```diff
@@ -1771,7 +1771,6 @@ def test_error():
   -----------------------------------------------------------------------
   "test.op"() : () -> ()
 }) : () -> ()
-
```
alexarice, line 1774: Is this change actually related to the change below?

alexarice (approved):
Seems reasonable, shouldn't be using the python `print` outside the printer.

tests/pattern_rewriter/test_pattern_rewriter.py
superlopuh: yep, `print()` prints one newline, `print("\n")` prints two

PR #4633, opened 2025-06-25
backend: closed set of allocatable registers

superlopuh:
We currently provide two sets of registers per target, those that can be used for allocation, and those that cannot be used. Registers that aren't in those two sets actually also can't be used. So we could just provide one list of registers that can be used by the allocator. This PR makes this simplification.

### DIFF
--- a/xdsl/backend/riscv/register_stack.py
+++ b/xdsl/backend/riscv/register_stack.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
-from typing import override
+
+from typing_extensions import override
 
 from xdsl.backend.register_stack import RegisterStack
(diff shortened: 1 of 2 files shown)


tests/backend/x86/test_register_stack.py
```diff
@@ -26,12 +24,7 @@ def test_push_infinite_register():
 
 
 def test_push_register():
-    register_stack = X86RegisterStack()
-
-    with pytest.raises(
-        NotImplementedError, match="x86 general register type not implemented yet."
```
superlopuh, line 32: no need for this, we just ignore registers that aren't allocatable

compor (approved):

PR #4639, opened 2025-06-25
frontend: (pyast) add support for calling functions with variable arguments

EdmundGoodman:
Add support for calling functions with positional and keyword arguments defined as symref variables. This supports code generation for cases like this:

```python
from ctypes import c_int32

from xdsl.dialects import arith, builtin
from xdsl.frontend.pyast.context import CodeContext
from xdsl.frontend.pyast.program import FrontendProgram

def add_i32(operand1: c_int32, operand2: c_int32) -> c_int32: ...

p1 = FrontendProgram()
p1.register_type(c_int32, builtin.i32)
p1.register_function(add_i32, arith.AddiOp)
with CodeContext(p1):
    def foo(x: c_int32, y: c_int32) -> c_int32:
        return add_i32(x, operand2=y)

p1.compile()
print(p1.textual_format())
```

into

```mlir
builtin.module {
  func.func @foo(%x : i32, %y : i32) -> i32 {
    %0 = arith.addi %x, %y : i32
    func.return %0 : i32
  }
}
```

which then unblocks porting existing tests away from the custom PyAST dialects

### DIFF
--- a/xdsl/frontend/pyast/code_generation.py
+++ b/xdsl/frontend/pyast/code_generation.py
@@ -200,11 +200,8 @@ class CodeGenerationVisitor(ast.NodeVisitor):
         # Resolve function
         assert isinstance(node.func, ast.Name)
         func_name = node.func.id
-        for key, value in self.type_converter.globals.items():
-            if key == func_name:
-                source_func = value
-                break
-        else:
+        source_func = self.type_converter.globals.get(func_name, None)
+        if source_func is None:
             raise CodeGenerationException(
                 self.file,
                 node.lineno,
@@ -229,7 +226,7 @@ class CodeGenerationVisitor(ast.NodeVisitor):
                     self.file,
                     node.lineno,
                     node.col_offset,
-                    "Function arguments must be a declared variable.",
+                    "Function arguments must be declared variables.",
                 )
             args.append(arg_op := symref.FetchOp(arg.id, self.symbol_table[arg.id]))
             self.inserter.insert_op(arg_op)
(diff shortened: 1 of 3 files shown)


xdsl/frontend/pyast/code_generation.py
```diff
@@ -196,6 +196,65 @@ def visit_BinOp(self, node: ast.BinOp):
                 f"which does not overload '{overload_name}'.",
             )
 
+    def visit_Call(self, node: ast.Call):
+        # Resolve function
+        assert isinstance(node.func, ast.Name)
+        func_name = node.func.id
+        for key, value in self.type_converter.globals.items():
```
superlopuh: Is it a dict? Why enumerate if yes?

xdsl/frontend/pyast/code_generation.py
```diff
@@ -169,7 +169,7 @@ def visit_BinOp(self, node: ast.BinOp):
             function_name = f"{source_type.__qualname__}.{method_name}"
             op = self.type_converter.function_registry.resolve_operation(
                 module_name=source_type.__module__,
-                function_name=function_name,
+                method_name=function_name,
```
superlopuh, line 172: why this change?

xdsl/frontend/pyast/code_generation.py
```diff
@@ -196,6 +196,65 @@ def visit_BinOp(self, node: ast.BinOp):
...
+        for key, value in self.type_converter.globals.items():
+            if key == func_name:
+                source_func = value
+                break
+        else:
+            raise CodeGenerationException(
+                self.file,
+                node.lineno,
+                node.col_offset,
+                f"Function '{func_name}' is not defined in scope.",
```
superlopuh, line 209: It would be nice to have tests for these, I'm not sure how good coverage is for the existing exceptions but it would be nice to have good coverage going forward.

xdsl/frontend/pyast/code_generation.py
EdmundGoodman: This is because this function resolves methods on types, not standalone functions (unlike the new code for call operations), so it made sense to make it a bit more clear in the naming scheme

superlopuh (approved):
Very nice

PR #4640, opened 2025-06-25
frontend (pyast): Remove custom PyAST dialects from tests

EdmundGoodman:
Using the function call support added in #4639, we can remove use of the custom dialect shims in PyAST tests

### DIFF
--- a/tests/filecheck/backend/x86/convert_arith_to_x86.mlir
+++ b/tests/filecheck/backend/x86/convert_arith_to_x86.mlir
@@ -23,6 +23,29 @@
 
 // -----
 
+%i0 = "test.op"(): () -> i32
+%i1 = "test.op"(): () -> i32
+%i2 = arith.muli %i0,%i1: i32
+
+// CHECK:      builtin.module {
+// CHECK-NEXT:   %i0 = "test.op"() : () -> i32
+// CHECK-NEXT:   %i1 = "test.op"() : () -> i32
+// CHECK-NEXT:   %i0_1 = builtin.unrealized_conversion_cast %i0 : i32 to !x86.reg
+// CHECK-NEXT:   %i1_1 = builtin.unrealized_conversion_cast %i1 : i32 to !x86.reg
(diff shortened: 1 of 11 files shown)


tests/filecheck/frontend/dialects/arith.py
```diff
@@ -1,120 +1,150 @@
...
-            # We expect the type error here, since the function doesn't exist on f64
-            return a in b
-
-    p.compile(desymref=False)
-    print(p.textual_format())
-except CodeGenerationException as e:
-    print(e.msg)
+# try:
+#     with CodeContext(p):
+#         # NO-CHECK: Binary operation 'FloorDiv' is not supported by type '_Float64' which does not overload '__floordiv__'.
```
superlopuh, line 130: ?

xdsl/frontend/pyast/type_conversion.py
```diff
@@ -46,20 +46,20 @@ def get_operation_type(
     def resolve_operation(
         self,
         module_name: str,
-        function_name: str,
+        method_name: str,
```
superlopuh: why this change?

EdmundGoodman: https://github.com/xdslproject/xdsl/pull/4639#discussion_r2167237916

tests/filecheck/frontend/dialects/arith.py
EdmundGoodman: This case fails, then falls into the old implementation which doesn't throw the correct error. It can be re-enabled when the logic to also try the old code if the new code doesn't work is removed and replaced with just throwing an error when we delete the custom dialects

superlopuh: OK this is worth a comment in the code I think

superlopuh:
Maybe also the implementation

EdmundGoodman:
Close to instead merge #4643, which modifies the tests and removes the dialects at the same time

PR #4646, opened 2025-06-26
core: specialize DataListAttr

superlopuh:
We now have the ability to express the constraint type.

### DIFF
--- a/tests/irdl/test_attribute_definition.py
+++ b/tests/irdl/test_attribute_definition.py
@@ -44,6 +44,7 @@ from xdsl.irdl import (
     BaseAttr,
     ConstraintContext,
     ConstraintVar,
+    GenericAttrConstraint,
     GenericData,
     MessageConstraint,
     ParamAttrConstraint,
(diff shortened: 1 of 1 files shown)


tests/irdl/test_attribute_definition.py
```diff
@@ -745,6 +722,30 @@ def from_list(data: list[AttributeInvT]) -> ListData[AttributeInvT]:
...
+    """
+
+    elem_constr: GenericAttrConstraint[AttributeInvT]
+
+    def verify(
+        self,
+        attr: Attribute,
+        constraint_context: ConstraintContext,
+    ) -> None:
+        attr = cast(ListData[Attribute], attr)
```
alexarice: We should do an `isa` check here instead. (or at least an `isinstance` before the cast

alexarice (commented):
This is only used for testing, is there any need to specialise it?

superlopuh:
OK I opened this because it seemed like an obvious improvement, and would be an easy approve, but I don't think it's worth the time to iterate on it.

alexarice:
I made one comment :(

superlopuh:
I changed my mind, this is actually key to something I'm trying to do. I'll make the isa fix.

tests/irdl/test_attribute_definition.py
superlopuh: Done

PR #4651, opened 2025-06-26
frontend: (pyast) restructure pyast implementation and filecheck tests

EdmundGoodman:
This PR re-arranges implementation files and filecheck tests for PyAST into a better structure. No changes are made to the code other than import resolution


tests/filecheck/frontend/programs/frontend_program_exceptions.py
```diff
@@ -25,7 +26,7 @@
 # CHECK-NEXT:     p.compile()
 with CodeContext(p):
 
-    def foo():
+    def foo():  # pyright: ignore[reportRedeclaration]
```
superlopuh: Can you please update the pyproject.toml with the right exclude/ignore list? Also there should be some dead lines there as I forgot about it we deleted the builtin shims.

xdsl/frontend/pyast/README.md
```diff
@@ -59,16 +39,83 @@ p.compile()
...
+   - [ ] Correctly handle scope, rather than just individually ingesting single
+         functions
+   - [ ] Extend `Call` AST to support invoking `func.FuncOp`s as well as
+         functions registered to map to operations
+   - [ ] Loosen constraints enforced by `PythonCodeCheck` to leverage added
+         functionality
+2. Add sensible defaults for registering types and functions
+   - [ ] `int` -> `bigint.bigint`
+   - [ ] `float` -> `builtin.f64`
+   - [ ] `bool` -> `builtin.i1`
```
superlopuh: I'm actually not sure since our discussion, we might want to exclude this from this list of things we definitely know we want to do

tests/filecheck/frontend/programs/frontend_program_exceptions.py
EdmundGoodman: Have added these to pyproject.toml and removed the added excludes. However, there is also a `tool.pyright.ignore` key which afaict does nothing and only the exclude key should be used. I will make a separate PR to get rid of this.

pyproject.toml
```diff
@@ -106,6 +106,8 @@ extraPaths = ["tests"]
     "tests/dialects/test_memref.py",
     "tests/test_frontend_op_resolver.py",
     "tests/test_frontend_python_code_check.py",
+    "tests/filecheck/frontend/programs/code_generation_exceptions.py",
+    "tests/filecheck/frontend/programs/frontend_program_exceptions.py",
```
superlopuh, line 110: are the ignored frontend things below still there? Can't highlight them in PR UI

tests/filecheck/frontend/programs/frontend_program_exceptions.py
superlopuh: So exclude and ignore are similar in that when you're not looking at the files they're both ignored, and differnt in that when you are looking at the file the exclude shows no errors and ignore does show them. I put the files in one or the other on purpose, in that some files are inhenrently untypable, and others are not typed as a bug. The frontend things not being typed properly was always a bug to me.

pyproject.toml
superlopuh: Not sure what you mean by the thumbs up reaction  @EdmundGoodman

superlopuh (approved):

pyproject.toml
EdmundGoodman: I think they are still there? Not sure why they don't show up in the UI, maybe because they are the new name of an old file?

tests/filecheck/frontend/programs/frontend_program_exceptions.py
EdmundGoodman: Per IRL conversation, they seem to both show errors in open files, so migrate everything to exclude to avoid confusion

xdsl/frontend/pyast/utils/type_conversion.py
```diff
@@ -31,7 +31,7 @@ def insert(
 
     def get_operation_type(
```
superlopuh: this needs a rename, i think, and probably worth doing in a separate PR first

EdmundGoodman: Peeled into #4776

PR #4637, opened 2025-06-25
WIP: dialects: (omp) added ops for offloading

dk949:
**Operations:**

* TargetOp,
* MapBoundsOp,
* MapInfoOp,
* SIMDOp,
* TeamsOp,
* DistributeOp,

**Attributes:**

* MapBoundsType,
* VariableCaptureKindAttr,

All the ops are from [upstream](https://mlir.llvm.org/docs/Dialects/OpenMPDialect/ODS/), implemented by @mesham and myself.

PR #4653, opened 2025-06-26
dialects:(omp) added omp.target

dk949:
`omp.target` operation as defined upstream

Portion of #4637

### DIFF
--- a/xdsl/dialects/omp.py
+++ b/xdsl/dialects/omp.py
@@ -242,15 +242,6 @@ class TargetOp(IRDLOperation):
     irdl_options = [AttrSizedOperandSegments(as_property=True)]
     traits = traits_def(IsolatedFromAbove())
 
-    def verify_(self) -> None:
-        if (len(self.depend_kinds) if self.depend_kinds else 0) != len(
-            self.depend_vars
-        ):
-            raise VerifyException(
-                "Number of `depend_kinds` must match the number of `depend_vars`"
-            )
-        return super().verify_()
-
 
 OMP = Dialect(
     "omp",


xdsl/dialects/omp.py
```diff
@@ -164,6 +193,52 @@ class TerminatorOp(IRDLOperation):
     traits = traits_def(IsTerminator())
 
 
+@irdl_op_definition
+class TargetOp(IRDLOperation):
+    """
+    Implementation of upstream omp.target
+    https://mlir.llvm.org/docs/Dialects/OpenMPDialect/ODS/#omptarget-omptargetop
```
```suggestion
    See external [documentation](https://mlir.llvm.org/docs/Dialects/OpenMPDialect/ODS/#omptarget-omptargetop).
```

xdsl/dialects/omp.py
```diff
@@ -164,6 +193,52 @@ class TerminatorOp(IRDLOperation):
...
+    traits = traits_def(IsolatedFromAbove())
+
+    def verify_(self) -> None:
+        if (len(self.depend_kinds) if self.depend_kinds else 0) != len(
+            self.depend_vars
+        ):
+            raise VerifyException(
+                "Number of `depend_kinds` must match the number of `depend_vars`"
+            )
+        return super().verify_()
```
superlopuh: We actually have a pattern for this, could you please have a go at trying to use it? https://github.com/xdslproject/xdsl/blob/28e783afb9ebfcb0f41bddd60b772c7ccac6c2a6/xdsl/dialects/smt.py#L243

dk949: I've had a look, but I'm not sure how I could use a constraint to ensure the number of arguments/array length matches without also constraining the types to be the same (which they shouldn't be)

superlopuh: Sorry I sent the wrong link: https://github.com/xdslproject/xdsl/blob/2ae6c7735f2265795651e3b4806be93af23507fb/tests/irdl/test_operation_definition.py#L416

tests/filecheck/mlir-conversion/with-mlir/dialects/omp/ops.mlir
```diff
@@ -130,136 +130,150 @@ builtin.module {
...
+    }) : (memref<6xf32>, i64, i32, i1, memref<10xi8>, f64, i32) -> ()
+    func.return
+  }
 }
 
 // CHECK:       builtin.module {
-// CHECK-NEXT:    func.func @omp_parallel(%arg37 : memref<1xi32>, %arg38 : i1, %arg39 : i32) {
-// CHECK-NEXT:      "omp.parallel"(%arg37, %arg37, %arg38, %arg39) <{operandSegmentSizes = array<i32: 1, 1, 1, 1, 0, 0>, proc_bind_kind = #omp<procbindkind spread>}> ({
-// CHECK-NEXT:        "omp.parallel"(%arg37, %arg37, %arg39) <{operandSegmentSizes = array<i32: 1, 1, 0, 1, 0, 0>}> ({
+// CHECK-NEXT:    func.func @omp_parallel(%arg47 : memref<1xi32>, %arg48 : i1, %arg49 : i32) {
```
superlopuh: do you know why there are all these changes? Seems weird to me, I would expect all of these to be isolated from above

superlopuh: Can you please update the test with this command? `xdsl-opt tests/filecheck/mlir-conversion/with-mlir/dialects/omp/ops.mlir --print-op-generic | mlir-opt --mlir-print-op-generic | xdsl-opt | uvx filecheckize --mlir-anonymize` Should get rid of the noize, the test doesn't seem to actually test the use-def chain so losing the names should be preferable.

xdsl/dialects/omp.py
superlopuh: Might as well remove this verify_ method now. Ideally it would be good to have an `ops_invalid` filecheck file to test that the length constraint is respected.

superlopuh (approved):

PR #4658, opened 2025-06-27
dialects: (hw) add init to ModuleType, ModulePort and ParamDeclAttr

superlopuh:
One more steps towards typed inits everywhere.

### DIFF
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -514,6 +514,9 @@ class ModuleType(ParametrizedAttribute, TypeAttribute):
 
     ports: ParameterDef[ArrayAttr[ModulePort]]
 
+    def __init__(self, ports: ArrayAttr[ModulePort]):
+        super().__init__((ports,))
+
     @classmethod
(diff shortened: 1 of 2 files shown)


xdsl/dialects/hw.py
```diff
@@ -656,20 +662,18 @@ class ModuleArg(NamedTuple):
...
-                                arg.port_type,
-                                DirectionAttr(arg.port_dir),
-                            )
+            ArrayAttr(
+                tuple(
+                    ModulePort(
+                        (
+                            StringAttr(arg.port_name),
+                            arg.port_type,
+                            DirectionAttr(arg.port_dir),
```
alexarice: The title of the PR seems wrong, the init for `ModulePort` doesn't seem to be updated.

superlopuh: Updated

alexarice: Is it not worth adding an init for `ModulePort` while here?

superlopuh: Sure

PR #4665, opened 2025-06-27
core: (AnyOf) verify without backtracking

alexarice:
Thought I'd make a draft PR to further motivate the disjointness changes. Could probably be cleaned up a bit, but this is at least a proof of concept. 

Would it be possible to try some of the verification benchmarks on this? Would like to know if this actually improves anything/makes stuff worse/whether it's better to just iterate through tuples rather than making dictionaries and sets etc.

### DIFF
--- a/tests/filecheck/dialects/arith/arith_invalid.mlir
+++ b/tests/filecheck/dialects/arith/arith_invalid.mlir
@@ -19,9 +19,9 @@
 // -----
 
 %c = arith.constant 1 : si32
-// CHECK:  Unexpected attribute 1 : si32
+// CHECK: Expected attribute #builtin.signedness<signless> but got #builtin.signedness<signed>
 
 // -----
 
 %c = arith.constant 1 : ui32
-// CHECK:  Unexpected attribute 1 : ui32
+// CHECK: Expected attribute #builtin.signedness<signless> but got #builtin.signedness<unsigned>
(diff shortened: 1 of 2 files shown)


xdsl/irdl/constraints.py
```diff
@@ -524,6 +536,12 @@ def mapping_type_vars(
             tuple(c.mapping_type_vars(type_var_mapping) for c in self.attr_constrs)
         )
 
+    def __repr__(self) -> str:
+        return f"AnyOf(attr_constrs={self.attr_constrs})"
+
+    def __hash__(self):
+        return hash(self.attr_constrs)
+
```
alexarice: There's probably some nice dataclass ways to do this right?

tests/filecheck/dialects/arith/arith_invalid.mlir
```diff
@@ -19,9 +19,9 @@
 // -----
 
 %c = arith.constant 1 : si32
-// CHECK:  Unexpected attribute 1 : si32
+// CHECK: Expected attribute #builtin.signedness<signless> but got #builtin.signedness<signed>
 
 // -----
 
 %c = arith.constant 1 : ui32
-// CHECK:  Unexpected attribute 1 : ui32
```
alexarice, line 27: These seem to be the only test changes and in my opinion seem like a big improvement.

xdsl/irdl/constraints.py
superlopuh: I'm surprised this isn't already implemented by default

alexarice:
@superlopuh was there a way to try some benchmarking on this?

superlopuh:
I'm not actually sure if there's an easier way than to pick a relevant benchmark in the benchmarks directory and run it before and after on main

superlopuh:
I could never get asv to work nicely locally on my machine

alexarice:
Ok, I'm going to go ahead and merge this.

superlopuh:
I mean it would have been nice to run at least some sort of benchmark before/after to see the impact, but it's not a big deal.

PR #4666, opened 2025-06-27
core: add PyRDLTypeError for irdl_to_attr_constraint

superlopuh:
Makes our exceptions a little more precise. Also adds doc strings for existing PyRDL errors.

### DIFF
--- a/xdsl/utils/exceptions.py
+++ b/xdsl/utils/exceptions.py
@@ -48,25 +48,25 @@ class PassFailedException(DiagnosticException):
 
 class PyRDLError(Exception):
     """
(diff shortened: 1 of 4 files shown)


xdsl/utils/exceptions.py
```diff
@@ -47,15 +47,27 @@ class PassFailedException(DiagnosticException):
 
 
 class PyRDLError(Exception):
-    pass
+    """
+    An error in our Python to IRDL conversion.
```
math-fehr: Maybe "in our Python IRDL eDSL" instead?
Python to IRDL might refer to something else

math-fehr (approved):

PR #4668, opened 2025-06-27
tool: (xdsl-opt) add basic syntax highlighting

alexarice:
Thought I'd have a little experiment with some syntax highlighting to see how this could look.
My feelings are:
- this would be a great addition to a tool that we view as helpful for teaching/learning purposes.
- The various print functions are now a bit more semantic driven, so we're probably in a good place to implement this pretty easily.

I've implemented nice printing for ssa values at the moment, but though I'd throw up a draft PR before continuing as other people may have thought about this and their may be different ways of doing this/constraints that I haven't thought about.

### DIFF
--- a/tests/test_printer.py
+++ b/tests/test_printer.py
@@ -1180,11 +1180,9 @@ def test_syntax_printer():
     op = test.TestOp(result_types=(IntegerType(32),))
 
     printer.stream = StringIO()
-    with printer.colored(Colors.BLUE):
-        with printer.colored(Colors.RED):
-            printer.print_ssa_value(op.results[0])
+    printer.print_ssa_value(op.results[0])
 
-    assert "\x1b[34m%0\x1b[0m" == printer.stream.getvalue()
+    assert "\x1b[95m%0\x1b[0m" == printer.stream.getvalue()
 
 
 def assert_print_op(


superlopuh (commented):
How do you feel about rich instead? We already depend on it for xdsl-gui

alexarice:
I have 0 strong feelings about packages to use for doing coloured terminal text, this was just the first thing I found. I more wanted to gather opinions on whether this would be interesting to people and if this is sort of what people imagined it looking like before I spend any significant amount of time figuring this out.

alexarice:
I had a little look and `rich` feels hugely less ergonomic to me. It seems you're forced to do a tonne of string interpolation to wrap bits of text with bbcodes, instead of just being able to pass a colour into the print function. It's very possible that I'm missing something though, and that getting rich working properly would allow much better integration with the gui for example.

superlopuh:
I have 0 experience with rich but it seems to be what the cool kids are using so I'm surprised that it's less ergonomic, I'll take a quick look.

superlopuh:
?
```py
from rich.console import Console
console = Console()
console.print("Hello", style="bold magenta")
```

alexarice:
Can you specify the output stream with `Console`?

alexarice:
It seems like you can set the `file` for `Console`, so maybe this wouldn't be so bad.

superlopuh:
TBH I looked at the implementation of cprint and it feels like we could definitely just take the parts that we use and go with that for now. I had more of a look at rich docs and it's less nice than I was hoping for. Also @AntonLydike might have tips here as he added colour printing to riscemu which was quite nice.

alexarice:
I think `rich` is far more featureful (and heavy) than what we realistically need.

AntonLydike:
All I needed for rich text in any of my projects was this: <https://github.com/AntonLydike/aalib/blob/main/aalib/colors.py>

You can just use `FMT.RED | FMT.BOLD` in an fstring or a print, and it'll work...

alexarice:
Some further thoughts on this. I'm not a huge fan of importing the termcolor module every time we call `_print`. Does python cache these imports or does this actually do some big import operation every time the function is called? (Maybe I'm worrying over nothing). The other options here would be:
- Just have it as a compulsory dependency (It's pretty small and pure python as far as I can tell)
- Hand roll our own terminal ansi code functionality as suggested by Anton

superlopuh:
Feels like it might be worth trying doing it by hand at first and then seeing if we want the extra dependency if the API gets too weird.

alexarice:
I've updated this to not rely on an external package. I now realise the key issue with this will be that no one will be able to agree on colour choices and I'm not sure what to do about this.

AntonLydike:
> I've updated this to not rely on an external package. I now realise the key issue with this will be that no one will be able to agree on colour choices and I'm not sure what to do about this.

I think it's fine to just choose one imho, any colour is better than no colour

superlopuh:
I like it a lot, and I agree that some color is better than no color. I also think that it's nice to have initial functionality here, and to extend it later to be more full-featured. One thing that might be worth looking into is whether it would make sense to try a mechanism with context managers, so `printer.with_color(colors.MAGENTA):`, with state stored on the printer. This would probably avoid the print override.

alexarice:
That sounds very reasonable, but I think it wouldn't work if newlines are printed in the body of the contextmanager

alexarice:
Nevermind I just tested it and it seems to work fine, I'll implement this

alexarice:
1 issue: I would expect:
```python
with printer.with_color(Colors.BLUE):
    printer.print_string("I am blue")
    with printer.with_color(Colors.MAGENTA):
        printer.print_string("I am magenta")
    printer.print_string("But I am still blue")
```
which would need some sort of stack or at least a current colour variable. Can implement if wanted but maybe we don't want to open this door.

AntonLydike:
> I like it a lot, and I agree that some color is better than no color. I also think that it's nice to have initial functionality here, and to extend it later to be more full-featured. One thing that might be worth looking into is whether it would make sense to try a mechanism with context managers, so `printer.with_color(colors.MAGENTA):`, with state stored on the printer. This would probably avoid the print override.

Not sure this is ideal, as colouring in the terminal cannot easily be mapped onto contextmanagers, specifically the nesting part.

doing something like this:

```
with fmt.bold:
  with fmt.red:
    print("x")
  with fmt.blue:
    print("y")
```

should output a bold red "x", and then a bold blue "y". But upon exiting the `red` contextmanager, only a full reset can be issued, and hence a re-printing of the bold formatting has to be done. The printed stream would need to be:

`<fmt.bold><fmt.red>x<fmt.reset><fmt.bold><fmt.blue>y<fmt.reset>`

superlopuh:
`<fmt.bold><fmt.red>x<fmt.blue>y<fmt.reset>` What would this print? Just curious.

AntonLydike:
> `<fmt.bold><fmt.red>x<fmt.blue>y<fmt.reset>` What would this print? Just curious.

I think setting a different foreground color erases the previous color.

superlopuh:
OK so it sounds like we need to make the style a part of the colour in our system and then we're good?

superlopuh:
I think we basically just need a current variable, no stack needed as it's in the context manager state.

AntonLydike:
> I think we basically just need a current variable, no stack needed as it's in the context manager state.

no, consider Alex' example:


> ```python
> with printer.with_color(Colors.BLUE):
>     printer.print_string("I am blue")
>     with printer.with_color(Colors.MAGENTA):
>         printer.print_string("I am magenta")
>     printer.print_string("But I am still blue")
> ```

Needs to print `<fmt.blue>I am blue<fmt.magenta>I am magenta<fmt.???>But I am still blue`

Alternatively, replace blue with bold.

superlopuh:
I don't understand your least message, you say "needs to print" but then put a broken string.

alexarice:
This could be done by storing the last color in the context manager, I still wonder if this is worth the trouble, we likely never need nested colours

superlopuh:
WDYM? We print operations in blocks in regions all the time, if we want syntax colouring per kind of object in the IR then this is the only way to go IMO.

AntonLydike:
> I don't understand your least message, you say "needs to print" but then put a broken string.

Ah yes, I missed that this one works with the stack integrated withing context managers. Let me re-phrase what I think the problem is.


1. There are (at least) four different kinds of formatting control characters: reset, foreground color, background color and style (bold, underline, possibly more)
2. Foreground and background control characters overwrite previous state
3. style control does *not*

I think that the model you proposed would re-print the "previous level" formatting string upon exiting a context manager, hoping that this would re-instate the previous state.

I believe, that while this approach works for the "simple" foreground case, it breaks when combining more of the different kinds of control characters.

alexarice:
> WDYM? We print operations in blocks in regions all the time, if we want syntax colouring per kind of object in the IR then this is the only way to go IMO.

This doesn't make sense to me, I don't imagine printing a whole region or whole block a different colour? More generally I can't see any example where it would make sense to "nest" colours.
For what it's worth, I hadn't really considered that we would want anything but foreground colour, but I guess it would be good not to exclude the others

superlopuh:
IDK I feel like if we add this capability I can relatively easily come up with a nice use-case for nested colours.

alexarice:
Refactored to use a context manager which doesn't allow nesting, and tried this on diagnostics, though I can't see any sensible way to thread through the information on whether syntax highlighting is wanted.

I feel not nesting makes sense here it feels like the red "error highlighting" should override regular syntax highlighting.

alexarice:
Refactored to use subclasses for syntax highlighted printing on @superlopuh 's suggestion. Still unsure how best to (optionally) activate syntax highlighting for printing diagnostics

superlopuh (commented):
This looks great! Is there any way to test this? Can you try adding different colours for regions and blocks?

alexarice:
Should be able to just filecheck test it, I just hadn't bothered yet. I had no plans to colour entire regions, I don't see why you would want to do this.

superlopuh:
Here's the syntax highlighting for MLIR in my IDE:

Screenshot 2025-07-23 at 15 03 40

The regions actually have a rotating color, which helps tell where each region starts and ends. Same deal with attribute dict and array delimiters. I find it useful.

alexarice:
Ok, if I understand correctly, you're expecting the braces to be coloured, but not expecting a blanket colour for the operations contained in the region. I can have a look at that, but was just trying to keep this PR minimal and add highlighting for other constructs in chunks, but can try to do it in one go if wanted.

The rotating colours might be slightly harder (but certainly doable if that's something we really want).

superlopuh:
I think this clears up the misunderstanding we might have had from the beginning, this was always what I thought would be useful, a way to just colour the braces of the region, not its contents. With the context manager, I would expect this to be a three-line change. I think the rotating colours is a very low priority nice-to-have, if anything.

alexarice:
Right, and colouring a block will realistically be colouring the block label. Tbh colouring the braces in a region seems like it would require slightly more refactoring (unless we just override `in_braces`, which might not actually be the worst idea).

superlopuh:
Let's add some tests and merge this? I'd rather have it in and tested than an open PR TBH, we'll probably find some uses for it for printing in notebooks, for example.

positr0nium:
+1 on finding this super useful!

superlopuh:
@positr0nium where would you find the syntax highlighting most useful?

positr0nium:
Ideally there would be some ``pprint`` function that would highlight the types and operation names in different colors. I think this would already be sufficient to improve the situation a lot. It works similarly in Jax.

superlopuh:
If you have the bandwidth, would you mind creating a branch on top of this one, and making the pretty printing with the flag in the PR look like what you would expect? I think with this as a motivating use-case we could merge this infrastructure.

alexarice:
Added tests and made the diagnostics work through a class variable. Should be easily extendable for other parts of the syntax.

xdsl/utils/color_printer.py
```diff
@@ -0,0 +1,25 @@
...
+
+from xdsl.utils.base_printer import BasePrinter
+from xdsl.utils.colors import RESET, Colors
+
+
+@dataclass(eq=False, repr=False)
+class ColorPrinter(BasePrinter):
+    """
+    A printer for printing colored text to a terminal window.
+    """
```
superlopuh, line 13: Can you please update the docstring to describe the color nesting behavior?

alexarice: Done

PR #4670, opened 2025-06-27
dialects:(omp) add ops for defining device data mapping

dk949:
`omp.map.bounds` and `omp.map.info`.

Also adds `OpenMPOffloadMappingFlags`, which defines what each bit in `map_type` does.

### DIFF
--- a/xdsl/dialects/omp.py
+++ b/xdsl/dialects/omp.py
@@ -1,12 +1,17 @@
-from enum import auto
+from enum import IntFlag, auto
 from typing import ClassVar, cast
+
+from typing_extensions import TypeVar
 
 from xdsl.dialects.builtin import (
     ArrayAttr,
+    BoolAttr,
     DenseIntOrFPElementsAttr,
     IndexType,
     IntegerAttr,
     IntegerType,
+    Signedness,
+    StringAttr,
     SymbolRefAttr,
     UnitAttr,
     i1,
@@ -18,8 +23,10 @@ from xdsl.ir import (
     Attribute,
     Dialect,
     EnumAttribute,
+    ParametrizedAttribute,
     SpacedOpaqueSyntaxAttribute,
     StrEnum,
+    TypeAttribute,
 )
 from xdsl.irdl import (
     AnyAttr,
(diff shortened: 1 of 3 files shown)


xdsl/dialects/omp.py
```diff
@@ -69,6 +125,31 @@ class DependKind(StrEnum):
...
+    with printer.in_parens():
+        printer.print_string(e.data)
+
+
+def _parse_omp_enum(e: type[EnumAttribute[OmpEnumType]], parser: AttrParser):
+    parser.in_angle_brackets()
+    parser.parse_punctuation("(")
+    res = parser.parse_str_enum(e.enum_type)
+    parser.parse_punctuation(")")
+    return cast(OmpEnumType, res)
```
```suggestion
    with parser.in_angle_brackets():
        with parser.in_parentheses():
		res = parser.parse_str_enum(e.enum_type)
		return cast(OmpEnumType, res)
```

xdsl/dialects/omp.py
```diff
@@ -32,18 +39,67 @@
...
     var_operand_def,
 )
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
-from xdsl.traits import IsolatedFromAbove, IsTerminator, NoTerminator
+from xdsl.traits import IsolatedFromAbove, IsTerminator, NoMemoryEffect, NoTerminator
 from xdsl.utils.exceptions import VerifyException
 
 
+class OpenMPOffloadMappingFlags(IntFlag):
```
superlopuh, line 55: this is unused

xdsl/dialects/omp.py
superlopuh: But also I don't understand this pattern, why not rely on the built-in functionality here?

superlopuh: Also it seems untested and seems like it shouldn't work, as the parser has angle brackets, and the printer doesn't...

superlopuh: Can you please split this PR out, first implementing the attributes, with an `attrs.mlir` file with `test.op`s that have the atttributes in an attribute dictionary, and then use this PR to add the operations?

dk949: > Also it seems untested and seems like it shouldn't work, as the parser has angle brackets, and the printer doesn't...

Oops...

 I accidentally left a rouge `in_angle_brackets` without a `with`, I thought the code was being tested, so I'm not sure what's going on. Will check. 

The reason I'm not using default parsing/printing is that these `omp` enums are printed as `#omp<enum_name(enum_value)>` rather than `#omp<enum_name enum_value>`. If there's an existing attribute base class that does that, I'd be very happy to use it instead.

xdsl/dialects/omp.py
dk949: It would be used when setting or testing the value of `map_type` in `omp.map.info`. `map_type` is a bit set, so the enum is not actually used in the definition of the operation.

xdsl/dialects/omp.py
dk949: This attribute is being used [here](https://github.com/xdslproject/xdsl/pull/4670/files#diff-8202b3e31526ef6f4d1f245198503967a88c5b89452a6d2c44ed30b2129ca58cR144), it parses and prints just fine... I'm really not sure why.

dk949: Found it, it is because it is not used in a `with`, `contextmanager` swallows the exception :

```python
@contextmanager
def mgr():
    raise RuntimeError("e")
    yield

try:
    mgr()
except:
    print("exception")
else:
    print("no exception") 

# Prints "no exception"
```

dk949: I wanted to use `parser.in_parentheses`, (like with printer) but I can't seem to find it

alexarice: We don't seem to have it (but absolutely should add it)

superlopuh: This might be relevant here: `SpacedOpaqueSyntaxAttribute`

superlopuh: We should, could also do `with delimited("(", ")")`

alexarice: that's only on the printer I think

xdsl/dialects/omp.py
superlopuh: I'm confused I try to search for this name on the page and it's not referenced anywhere.

xdsl/dialects/omp.py
superlopuh: No longer!

dk949: Using the new method now

xdsl/dialects/omp.py
dk949: Sorry, I wasn't being very clear. Updated documentation accordingly. (Although that apparently also broke CI)

xdsl/dialects/omp.py
```diff
@@ -69,6 +136,28 @@ class DependKind(StrEnum):
...
+
+
+def _print_omp_enum(e: EnumAttribute[OmpEnumType], printer: Printer):
+    with printer.in_parens():
+        printer.print_string(e.data)
+
+
+def _parse_omp_enum(e: type[EnumAttribute[OmpEnumType]], parser: AttrParser):
+    with parser.in_parens():
+        return cast(OmpEnumType, parser.parse_str_enum(e.enum_type))
```
superlopuh: I would recommend just inlining these they seem pretty small

xdsl/dialects/omp.py
superlopuh: Ah I see makes sense.

superlopuh (approved):
Thank you!

PR #4671, opened 2025-06-27
core: (Printer) cleanup core printing functions

alexarice:
Cleans up some of the printing functions in `xdsl.printer`, using delimiter/indentation functions where possible, and using `print_int` for integers etc.

### DIFF
--- a/tests/filecheck/parser-printer/metadata.mlir
+++ b/tests/filecheck/parser-printer/metadata.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt %s | filecheck %s
+// RUN: xdsl-opt %s --split-input-file | filecheck %s
 
 builtin.module {
     func.func @constant_dense_resource() {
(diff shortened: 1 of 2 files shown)


tests/filecheck/parser-printer/metadata.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt %s | filecheck %s
```
alexarice, line 1: Drive by fix of this test

xdsl/printer.py
```diff
@@ -629,21 +619,32 @@ def print_metadata(self, dialects: Iterable[Dialect]) -> None:
...
+                                        resources.items(), key=lambda x: x[0]
+                                    )
+                                    for key, resource in sorted_elements[:-1]:
+                                        self.print_string(f'{key}: "{resource}",')
+                                    self.print_string(
+                                        f'{sorted_elements[-1][0]}: "{sorted_elements[-1][1]}"'
+                                    )
+                                self._print_new_line()
+                    self._print_new_line()
+            self._print_new_line()
```
alexarice, line 647: This was the biggest change, I liked using `self.indented` though wonder if I should use `delimited` to include the newline printing with the braces.

superlopuh: I think we can leave it as-is, I'm very curious to see how your colour printing may affect choices here.

PR #4673, opened 2025-06-27
dialects: (fir) Update to bring inline with operations in Flang 20.1.1

mesham:
Additional of the DummyScopeType and Operation, along with a missing optional operand elsewhere that brings HLFIR and FIR inline with the version used in Flang 20.1.1 that we are now working with.

These are experimental dialects, hence no tests

### DIFF
--- a/xdsl/dialects/experimental/fir.py
+++ b/xdsl/dialects/experimental/fir.py
@@ -191,6 +191,18 @@ class DeferredAttr(ParametrizedAttribute, TypeAttribute):
 
 
 @irdl_attr_definition
+class DummyScopeType(ParametrizedAttribute, TypeAttribute):
+    """
+    fir.dscope is a type returned by fir.dummy_scope operation.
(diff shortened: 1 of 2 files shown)


xdsl/dialects/experimental/fir.py
```diff
@@ -359,12 +384,89 @@ def parse_value():
...
+    Model of a Fortran LOGICAL intrinsic type, including the KIND type
+    parameter
+    """
+
+    name = "fir.logical"
+
+    size: ParameterDef[IntAttr]
+
+    def print_parameters(self, printer: Printer) -> None:
+        printer.print("<")
```
alexarice: Printer.print has been deprecated in favour of more strictly typed printing functions
```suggestion
        with printer.in_angle_brackets():
```

mesham: Yer I just noticed that - thanks for making those changes previously (just doing this now)

alexarice (approved):

PR #4674, opened 2025-06-27
dialects: (stencil) add typed inits to stencil and csl attrs

superlopuh:
Did a bunch in one go in related bits of code.

### DIFF
--- a/tests/dialects/test_stencil.py
+++ b/tests/dialects/test_stencil.py
@@ -607,7 +607,7 @@ def test_external_load():
 
 
 def test_external_store():
-    field = create_ssa_value(FieldType((5), f32))
+    field = create_ssa_value(FieldType(5, f32))
     memref = create_ssa_value(MemRefType(f32, ([5])))
 
     external_store = ExternalStoreOp.build(operands=[field, memref])
--- a/tests/dialects/test_stencil.py
+++ b/tests/dialects/test_stencil.py
@@ -596,7 +596,7 @@ def test_store_result():
 
 
 def test_external_load():
-    memref = create_ssa_value(MemRefType(f32, ([5])))
+    memref = create_ssa_value(MemRefType(f32, [5]))
     field_type = FieldType((5), f32)
 
     external_load = ExternalLoadOp.get(memref, field_type)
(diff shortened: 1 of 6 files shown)


tests/dialects/test_stencil.py
```diff
@@ -607,7 +607,7 @@ def test_external_load():
 
 
 def test_external_store():
-    field = create_ssa_value(FieldType((5), f32))
+    field = create_ssa_value(FieldType(5, f32))
     memref = create_ssa_value(MemRefType(f32, ([5])))
```
alexarice: Seems to be unnecessary parentheses here?

xdsl/dialects/csl/csl.py
```diff
@@ -444,13 +449,14 @@ def get_element_type(self):
 
     @staticmethod
     def from_type(child_type: Attribute) -> VariableOp:
-        return VariableOp(result_types=[VarType([child_type])])
+        assert isinstance(child_type, TypeAttribute)
```
alexarice, line 452: Why not tighten the type bound in the argument?

xdsl/dialects/experimental/dmp.py
```diff
@@ -302,7 +311,7 @@ def from_index_attrs(
     ):
         data_type = builtin.i64
         return ShapeAttr(
-            [
+            *[
```
alexarice: I think `*( ... )` will avoid creating the list.

tests/dialects/test_stencil.py
superlopuh: oh yeah

xdsl/dialects/csl/csl.py
superlopuh: I tried that and it was a rabbit hole that I decided to defer to a future side quest.

xdsl/dialects/experimental/dmp.py
superlopuh: probably will I'll make the change

xdsl/dialects/csl/csl.py
alexarice: Fair enough

xdsl/dialects/experimental/dmp.py
superlopuh: Done

PR #4676, opened 2025-06-28
core: (Parser) add delimiter functions

alexarice:
Similar to the printing functions (except using the Delimiter enum here)

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1051,19 +1051,18 @@ def test_parse_optional_complex_success(
 
 
 @pytest.mark.parametrize(
-    "delimiter, text",
+    "start, end, text",
     [
-        (Parser.Delimiter.ANGLE, "<1>"),
-        (Parser.Delimiter.SQUARE, "[1]"),
-        (Parser.Delimiter.BRACES, "{1}"),
-        (Parser.Delimiter.PAREN, "(1)"),
-        (Parser.Delimiter.METADATA_TOKEN, "{-# 1 #-}"),
-        (Parser.Delimiter.NONE, "1"),
+        ("<", ">", "<1>"),
+        ("[", "]", "[1]"),
+        ("{", "}", "{1}"),
+        ("(", ")", "(1)"),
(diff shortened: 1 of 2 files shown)


xdsl/parser/generic_parser.py
```diff
@@ -307,9 +307,25 @@ def parse_characters(self, text: str, context_msg: str = "") -> str:
         self.raise_error(f"'{text}' expected" + context_msg)
 
     @contextmanager
-    def in_angle_brackets(self):
-        self.parse_characters("<")
-        try:
+    def delimited(self, delimiter: Delimiter):
```
superlopuh: why not re-use the API of printer?
```suggestion
    def delimited(self, start: str, end: str):
```

alexarice: I could do, kind of makes me wonder what the point of `Parser.Delimiter` was

superlopuh: I think it made more sense when the parser and printer were only for MLIR syntax, this is more of a generic class, I feel like the Delimiter helper would be more helpful in a subclass where you might want to make it clear which delimiters are valid tokens.

superlopuh: I might be missing something

tests/test_parser.py
```diff
@@ -1050,6 +1050,80 @@ def test_parse_optional_complex_success(
...
+def test_delimiters(start: str, end: str, text: str):
+    parser = Parser(Context(), text)
+    with parser.delimited(start, end):
+        value = parser.parse_integer()
+
+    assert value == 1
+    assert parser.pos == len(text)
+
+
+def test_angle_brackets():
```
superlopuh, line 1072: supernit: feels like the tests below can be consolidated by passing the function as an argument

superlopuh (approved):

tests/test_parser.py
alexarice: It's a bit nasty with it being a function on parser which returns a context manager. I'd prefer just to leave it as is

PR #4677, opened 2025-06-28
dialect: Add PositiveIntConstr Constraint to tensor dialect

Jimmy2027:
Used to verify the ReassociationAttr

Maybe this should go somewhere it can be shared? Maybe the ShapeAttr could benefit from this as well.

### DIFF
--- a/tests/filecheck/dialects/tensor/invalid_ops.mlir
+++ b/tests/filecheck/dialects/tensor/invalid_ops.mlir
@@ -86,6 +86,6 @@ builtin.module {
 // -----
 
 // CHECK: The following constraints were not satisfied:
-// CHECK-NEXT: Expected positive integer, got -2.
+// CHECK-NEXT: expected integer >= 0, got -2
 // CHECK-NEXT: All inner arrays must be contiguous: [[-2 : i64, 3 : i64], [0 : i64, 1 : i64]]
 %res_collapse2 = tensor.collapse_shape %t0 [ [-2, 3], [0, 1] ] : tensor<4x1xf32> into tensor<4x1xf32>
--- a/tests/filecheck/dialects/tensor/invalid_ops.mlir
+++ b/tests/filecheck/dialects/tensor/invalid_ops.mlir
@@ -82,3 +82,10 @@ builtin.module {
 
 // CHECK: All inner arrays must be contiguous: [[2 : i64, 3 : i64], [0 : i64, 1 : i64]]
 %res_collapse2 = tensor.collapse_shape %t0 [ [2, 3], [0, 1] ] : tensor<4x1xf32> into tensor<4x1xf32>
+
+// -----
+
+// CHECK: The following constraints were not satisfied:
+// CHECK-NEXT: Expected positive integer, got -2.
+// CHECK-NEXT: All inner arrays must be contiguous: [[-2 : i64, 3 : i64], [0 : i64, 1 : i64]]
+%res_collapse2 = tensor.collapse_shape %t0 [ [-2, 3], [0, 1] ] : tensor<4x1xf32> into tensor<4x1xf32>
--- a/xdsl/dialects/tensor.py
+++ b/xdsl/dialects/tensor.py
@@ -48,7 +48,8 @@ class ContiguousArrayOfIntArray(AttrConstraint):
     """
     Enforce an ArrayAttr of ArrayAttr[IntegerAttr] to contain contiguous integer values across all inner arrays.
     For example: [[0, 1], [2, 3]] is valid, but [[3, 4], [0, 1]] is not.
-    An empty inner array is considered contiguous.
+    An empty inner array is considered contiguous. Incidentally, this also
+    verifies that all integers across all inner arrays are unique.
     """
 
     def verify(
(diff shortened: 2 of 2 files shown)


xdsl/dialects/tensor.py
```diff
@@ -48,7 +48,8 @@ class ContiguousArrayOfIntArray(AttrConstraint):
     """
     Enforce an ArrayAttr of ArrayAttr[IntegerAttr] to contain contiguous integer values across all inner arrays.
     For example: [[0, 1], [2, 3]] is valid, but [[3, 4], [0, 1]] is not.
-    An empty inner array is considered contiguous.
+    An empty inner array is considered contiguous. Incidentally, this also
```
Jimmy2027: this slipped inside, should I create a separate PR for this?

xdsl/dialects/tensor.py
```diff
@@ -73,6 +74,25 @@ def mapping_type_vars(
...
+    """
+    Enforce that an IntegerAttr is positive (> 0).
+    """
+
+    def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
+        if not isinstance(attr, IntegerAttr):
+            raise VerifyException(
+                f"Expected {IntegerAttr.name} attribute, but got {attr.name}."
+            )
+        if attr.value.data < 0:
```
alexarice: Either the docstring or this condition is incorrect

alexarice (commented):
I feel it would be nice to do this with an `IntConstraint` which could be used `IntAttrConstraint` (which already exists). The `IntConstraint` could live in `xdsl/irdl/constraints.py`

xdsl/dialects/tensor.py
Jimmy2027: yup sry docstring is incorrect

xdsl/dialects/tensor.py
```diff
@@ -73,6 +74,25 @@ def mapping_type_vars(
         return self
 
 
+class PositiveIntConstr(AttrConstraint):
+    """
+    Enforce that an IntegerAttr is positive (> 0).
```
Jimmy2027:
```suggestion
    Enforce that an IntegerAttr is positive (>= 0).
```

Jimmy2027:
Alright, will see how that would look

alexarice:
In fact the `IntConstraint` already exists https://github.com/xdslproject/xdsl/blob/37afb2e0cb1627fc688d24b0dae73a02d5cf43d2/xdsl/irdl/constraints.py#L777

So you should be able to enforce that an `IntegerAttr` is positive with:
```python
IntegerAttr.constr(value=IntAttrConstraint(AtLeast(0)))
```
(as an aside we could probably make this a bit more ergonomic)

Jimmy2027:
Would it make sense to define a PositiveIntAttr?

alexarice:
I think it would make sense to define it as an alias.

Jimmy2027:
maybe it might be good to define unsigned integer types?
```python
ui64 = IntegerType(64, Signedness.UNSIGNED)
ui32 = IntegerType(32, Signedness.UNSIGNED)
ui16 = IntegerType(16, Signedness.UNSIGNED)
ui8 = IntegerType(8, Signedness.UNSIGNED)
ui1 = IntegerType(1, Signedness.UNSIGNED)
UI64 = Annotated[IntegerType, ui64]
UI32 = Annotated[IntegerType, ui32]
UI16 = Annotated[IntegerType, ui16]
UI8 = Annotated[IntegerType, ui8]
UI1 = Annotated[IntegerType, ui1]

```

but that seems like a bigger change

alexarice:
I think deleting my branch closed this, but that wasn't my intention

Jimmy2027:
hmm github is kind of broken when not targeting the default branch, I'm not even able to change the target branch of this pr..

superlopuh (approved):

xdsl/dialects/tensor.py
```diff
@@ -216,7 +218,14 @@ def parse(cls, parser: Parser) -> Self:
 
 
 ReassociationAttr = Annotated[
-    ArrayAttr[ArrayAttr[IntegerAttr[I64]]],
+    ArrayAttr[
+        ArrayAttr[
+            Annotated[
+                IntegerAttr[I64],
+                IntegerAttr.constr(value=IntAttrConstraint(AtLeast(0))),
```
alexarice: Could you rebase onto the latest main? I worry that this will trigger the `AnyOf` disjointness check

tests/filecheck/dialects/tensor/invalid_ops.mlir
```diff
@@ -82,3 +82,10 @@ builtin.module {
 
 // CHECK: All inner arrays must be contiguous: [[2 : i64, 3 : i64], [0 : i64, 1 : i64]]
 %res_collapse2 = tensor.collapse_shape %t0 [ [2, 3], [0, 1] ] : tensor<4x1xf32> into tensor<4x1xf32>
+
+// -----
+
+// CHECK: The following constraints were not satisfied:
+// CHECK-NEXT: expected integer >= 0, got -2
+// CHECK-NEXT: All inner arrays must be contiguous: [[-2 : i64, 3 : i64], [0 : i64, 1 : i64]]
```
alexarice, line 90:
```suggestion
// CHECK-NEXT: All inner arrays must be positive and contiguous: [[-2 : i64, 3 : i64], [0 : i64, 1 : i64]]
```
Could we update this message?

alexarice (commented):
Looks good, I just left a couple of comments

tests/filecheck/dialects/tensor/invalid_ops.mlir
alexarice: or perhaps "non-negative", just worry the message could be confusing if the array was continuous but negative.

superlopuh: Oh yeah I saw that in some merge commit but couldn't find it in this PR, definitely should be non-negative.

alexarice: In retrospect it might be too hard to change this message as it's in a different constraint.

xdsl/dialects/tensor.py
Jimmy2027: done

alexarice: I was worried over nothing, this corresponds to an `AllOf` rather than an `AnyOf`

alexarice:
```suggestion
                IntegerAttr.constr(value=AtLeast(0)),
```
I think this is possible since rebasing? (and was the whole reason for the stacked PR that broke everything.

alexarice (approved):
1 quick comment, then happy for merge.

xdsl/dialects/tensor.py
Jimmy2027: right, sry pushed/pulled the wrong thing

Jimmy2027:
will re-word the commit message

Jimmy2027:
okay should be good now :)

PR #4680, opened 2025-06-29
dialects: (builtin) add typed inits to remaining builtin attrs

superlopuh:


xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
```diff
@@ -322,8 +323,7 @@ def match_and_rewrite(self, op: memref.SubviewOp, rewriter: PatternRewriter):
         source = op.source
         result = op.result
         source_type = source.type
-        assert isinstance(source_type, MemRefType)
-        source_type = cast(MemRefType, source_type)
+        assert isa(source_type, MemRefType)
```
alexarice: This looks a bit unrelated

xdsl/dialects/affine.py
```diff
@@ -317,7 +317,7 @@ def __init__(
                 raise ValueError(
                     "affine.store memref operand must be of type ShapedType"
                 )
-            memref_type = cast(MemRefType, memref.type)
+            memref_type = memref.type
```
alexarice: I in general prefer things like this to be inlined

xdsl/backend/riscv/lowering/convert_memref_to_riscv.py
superlopuh: This and the other thing below were drive-by fixes, I think they're incremental improvements even if there's more work to do

PR #4681, opened 2025-06-29
dialects: (llvm) add typed inits to llvm dialect

superlopuh:


xdsl/dialects/llvm.py
```diff
@@ -77,24 +77,24 @@
 """
 
 
-def _parse_llvm_type(parser: AttrParser):
+def _parse_llvm_type(parser: AttrParser) -> Attribute | None:
```
alexarice:
```suggestion
def _parse_optional_llvm_type(parser: AttrParser) -> Attribute | None:
```
maybe more informative

PR #4683, opened 2025-06-29
core: add diagnostic info to original error __notes__

superlopuh:
Leverage the notes feature of 3.11+ Pythons to append information to exceptions, in a way that also works for 3.10. This lets us avoid the excessive "This was caused by" and multiple stack traces, making the error messages more focused on the real cause. Also removes the need to create extra exceptions with dodgy exception types that may not have the expected init method.

### DIFF
--- a/xdsl/utils/diagnostic.py
+++ b/xdsl/utils/diagnostic.py
@@ -5,7 +5,6 @@ from io import StringIO
 from typing import NoReturn
 
 from xdsl.ir import Block, IRNode, Operation, Region
-from xdsl.utils.exceptions import DiagnosticException
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


xdsl/utils/diagnostic.py
```diff
@@ -41,4 +34,14 @@ def raise_exception(
         else:
             assert "xDSL internal error: get_toplevel_object returned unknown construct"
 
-        raise exception_type(message + "\n\n" + f.getvalue()) from underlying_error
```
superlopuh, line 44: This prevents us from subclassing exceptions with inits that aren't just taking a string, the proposed mechanism works for any exception.

math-fehr (approved):
Looks cool!

PR #4684, opened 2025-06-29
API: automatically generate typed inits for ParametrizedAttributes

superlopuh:
A PR many years in the making, leveraging the [`dataclass_transform`](https://peps.python.org/pep-0681/) decorator to indicate that our `irdl_attr_definition` is a kind of dataclass decorator. I added a new optional parameter to it to optionally not generate inits.

An important note to remember is that custom inits have to call `self.__post_init__()` at the end in order to be verified, I'll look into how to make sure that this is done later, for now I propose that we just try to remember to add it manually.


docs/marimo/irdl.py
```diff
@@ -462,7 +462,7 @@ class MyIntegerType(ParametrizedAttribute):
         width: ParameterDef[IntAttr]
 
 
-    my_i32 = MyIntegerType([IntAttr(32)])
+    my_i32 = MyIntegerType(IntAttr(32))
```
alexarice, line 465: Nearly got them all

tests/irdl/test_attribute_definition.py
```diff
@@ -222,7 +222,7 @@ class IndirectEnumData(  # pyright: ignore[reportUnusedClass]
             name = "test.non_identifier_enum"
 
 
-@irdl_attr_definition
+@irdl_attr_definition(init=False)
```
alexarice, line 225: Could you explain why this is needed?

tests/irdl/test_attribute_definition.py
```diff
@@ -907,11 +874,11 @@ class OveriddenInitAttr(ParametrizedAttribute):
...
-        match param:
-            case int():
-                super().__init__((IntData(param),))
-            case str():
-                super().__init__((StringData(param),))
+        object.__setattr__(
+            self,
+            "param",
+            IntData(param) if isinstance(param, int) else StringData(param),
+        )
```
alexarice: I'm not convinced this is a good test, we would basically never want to do this right? Maybe it should be removed if it's no longer relevant.

If it is relevant, why do we need the `setattr`, can we not call super.init or something?

xdsl/backend/register_type.py
```diff
@@ -22,7 +22,7 @@
 from xdsl.utils.exceptions import VerifyException
 
 
-@dataclass(frozen=True, init=False)
+@dataclass(frozen=True)
```
alexarice, line 25: Should this be `@irdl_attr_definition`?

alexarice: If not, why not

xdsl/dialects/accfg.py
```diff
@@ -86,7 +86,8 @@ class TokenType(ParametrizedAttribute, TypeAttribute):
     def __init__(self, accelerator: str | StringAttr):
         if not isinstance(accelerator, StringAttr):
             accelerator = StringAttr(accelerator)
-        super().__init__([accelerator])
+        object.__setattr__(self, "accelerator", accelerator)
+        self.__post_init__()
```
alexarice: Will this always be needed? I'm not sure I'm a fan of this pattern

xdsl/irdl/constraints.py
```diff
@@ -308,7 +308,7 @@ def mapping_type_vars(
         return res
 
 
-@dataclass(frozen=True, init=True)
+@dataclass(frozen=True)
```
alexarice: related change?

alexarice:
This is very cool though, definitely a big improvement :+1:

docs/marimo/irdl.py
superlopuh: Nearly...

superlopuh:
> when should you use irdl_attr_definition(init=False)

I'm not 100% sure but it seems like it's when you want an abstract superclass's init instead

> For things with custom inits, are they able to use the new function construct the class instead? I'm not a fan of the having to remember to call __post_init__

I can address this first before mergint the PR if you'd prefer, but I wanted to try first merging this and param_def and trying some dataclass magic in case we could actually have these inits instead of custom getters everywhere.

> Would it be able to explain what the dataclass_transform thing does? Are there advantages/disadvantages to this approach?

`dataclass_transform` tells type checkers that the annotation acts like `dataclasses.dataclass`, meaning it generates inits, eqs, and so on. I don't know of an alternative TBH, so cannot really do advantages/disadvantages.

xdsl/backend/register_type.py
superlopuh: OK so what I'm actually thinking we want here is some kind of `irdl_attr_trait` decorator for abstract superclasses and was hoping to begin this side quest later. But basically that's the answer, it's for Attribute superclasses that define fields.

tests/irdl/test_attribute_definition.py
superlopuh: I'm happy to remove it. @math-fehr ?

tests/irdl/test_attribute_definition.py
superlopuh: As I wrote on the main PR discussion this is for when you want to use the superclass's init.

xdsl/irdl/attributes.py
```diff
@@ -262,19 +263,39 @@ def get_type_index(cls: Any) -> int:
...
-        "'ParametrizedAttribute'"
-    )
+@overload
+def irdl_attr_definition(
+    cls: TypeAttributeInvT, *, init: bool = True
+) -> TypeAttributeInvT: ...
+
+
+@overload
+def irdl_attr_definition(
```
alexarice, line 273: What's with this overload? Where are we using this?

xdsl/irdl/constraints.py
superlopuh: Nope

xdsl/irdl/attributes.py
superlopuh: This is the new one we added, that lets you provide `init=False` if you want, without passing the type.

tests/irdl/test_attribute_definition.py
math-fehr: Yeah we can remove it, we should never be doing that

xdsl/backend/register_type.py
math-fehr: What do you need from `dataclass` in here? Just the `__init__`? Can you not directly use the one generated from `irdl_attr_definition`?

xdsl/dialects/accfg.py
math-fehr: Why not `super().__init__` it?

xdsl/dialects/stencil.py
```diff
@@ -390,6 +387,18 @@ class FieldType(
...
+        self,
+        bounds: (
+            Iterable[tuple[int | IntAttr, int | IntAttr]]
+            | int
+            | IntAttr
+            | StencilBoundsAttr
+        ),
+        element_type: _FieldTypeElement,
+    ) -> None:
+        super().__init__(bounds, element_type)
```
alexarice: This is surely wrong?

xdsl/irdl/attributes.py
alexarice: Where do we not pass the type?

xdsl/ir/core.py
```diff
@@ -452,9 +454,13 @@ def new(cls: type[Self], params: Sequence[Attribute]) -> Self:
         # We do this to allow users to redefine their own __init__.
```
alexarice, line 465: I'm now very confused, the init doesn't look modified

xdsl/irdl/attributes.py
superlopuh: First overload is for this:
```
@irdl_attr_definition # pass type directly
class Bla: ...
```

Second for this:
```
@irdl_attr_definition(init=False) # pass type to returned function
class Bla: ...
```

xdsl/dialects/accfg.py
superlopuh: Not needed!

xdsl/irdl/attributes.py
alexarice: I see, thanks

xdsl/ir/core.py
superlopuh: You mean why the change below despite me reverting the modification of the init? I can revert that as well, it's just one less function call and avoids the tuple creation I guess.

superlopuh: But this particular comment is somewhat straightforward, we need a method for the parser that lets us create arbitrary paramattrs from a tuple of the params.

alexarice: This isn't what I meant, but github didn't let me comment above, the init for `ParametrizedAttribute` doesn't look modified.

tests/irdl/test_attr_constraint.py
```diff
@@ -56,27 +56,18 @@ class Base(ParametrizedAttribute, ABC):
...
-        super().__init__((param,))
-
 
 @irdl_attr_definition
 class AttrC(Base):
     name = "test.attr_c"
 
-    def __init__(self):
-        super().__init__(())
-
```
alexarice, line 80: On main there is now an `AttrD` below that needs its init removing

superlopuh: Yep but with your change it no longer breaks if it's not removed! I'll remove it anyway.

alexarice: I think I've got really confused with all this stuff, shouldn't specifying a custom init without `init=False` in the decorator be an error?

xdsl/backend/register_type.py
alexarice: Ah right, I see

tests/irdl/test_attr_constraint.py
superlopuh: Not that I know of, seems to work fine.

xdsl/dialects/stencil.py
alexarice: These still don't look right to me, where are they coming from?

alexarice (approved):
1 comment then lets merge

xdsl/dialects/stencil.py
superlopuh: I pushed a change removing them, needed to pass `init=False` to the decorator

math-fehr (commented):
Forgot to push some comments

xdsl/ir/core.py
```diff
@@ -423,7 +425,18 @@ def print_parameter(self, printer: Printer):
 class ParametrizedAttribute(Attribute):
     """An attribute parametrized by other attributes."""
 
-    def __init__(self, parameters: Sequence[Attribute] = ()):
+    def __init__(self, *parameters: Attribute):
```
alexarice, line 428: This will need to be typed `Attribute | Sequence[Attribute]` right? likely with some overloads above?

superlopuh: Nah I think we want to discourage this completely, I just don't want to break the user's code, but I'm happy to break their type checking.

alexarice: Perhaps it's fine for the type to not include the `Sequence` as the deprecation warning should still fire if you run `make tests` or equivalent

superlopuh: I think so

PR #4686, opened 2025-06-30
API: use variadic init on ParametrizedAttribute

superlopuh:
A bit like #4684, but the init is not typed, the idea is that this initial step would give a nicer API for subclasses to call super init with, and potentially let us not use the dataclass generated init at all, using our existing init instead.


alexarice:
I'm not sure this is the order to merge these, feels like 90% of the changes here are just changing things that are removed by your PR.

superlopuh:
OK I'll make the change directly in my PR instead

PR #4687, opened 2025-06-30
API: make ParametrizedAttribute init variadic

alexarice:
This is what my change looks like on top of yours.


alexarice:
This was remade in the upstream PR

PR #4688, opened 2025-06-30
dialects:(omp) add `LoopWrapper` trait

dk949:
This is used by several operations upstream but currently only applies to `omp.wsloop` in xDSL.

Used to constrain operations which are only supposed to contain a `loop_nest` or another `LoopWrapper`.

### DIFF
--- a/tests/filecheck/dialects/omp/ops_invalid.mlir
+++ b/tests/filecheck/dialects/omp/ops_invalid.mlir
@@ -11,4 +11,16 @@ func.func @omp_ordered(%arg0 : i32, %arg1 : i32, %arg2 : i32, %arg3 : i64, %arg4
     return
 }
 
-// CHECK: Operation does not verify: Body of omp.wsloop operation body must consist of one loop nest
+
+// CHECK: Operation does not verify: omp.wsloop is not a LoopWrapper: has 2 ops, expected 1
+
+// -----
+
+func.func @omp_ordered(%arg0 : i32, %arg1 : i32, %arg2 : i32, %arg3 : i64, %arg4 : i64, %arg5 : i64, %arg6 : i64) {
+    "omp.wsloop"() <{operandSegmentSizes = array<i32: 0, 0, 0, 0, 0, 0, 0>, ordered = 0 : i64}> ({
+        "omp.terminator"() : () -> ()
+    }) : () -> ()
+    return
+}
+
+// CHECK: omp.wsloop is not a LoopWrapper: should have a single operation which is either another LoopWrapper or omp.loop_nest
--- a/xdsl/dialects/omp.py
+++ b/xdsl/dialects/omp.py
@@ -32,6 +32,7 @@ from xdsl.irdl import (
     AttrSizedOperandSegments,
     IntVarConstraint,
     IRDLOperation,
+    Operation,
     RangeOf,
     SameVariadicOperandSize,
     base,
@@ -48,8 +49,15 @@ from xdsl.irdl import (
 )
 from xdsl.parser import AttrParser
 from xdsl.printer import Printer
-from xdsl.traits import IsolatedFromAbove, IsTerminator, NoMemoryEffect, NoTerminator
+from xdsl.traits import (
+    IsolatedFromAbove,
+    IsTerminator,
+    NoMemoryEffect,
+    NoTerminator,
+    OpTrait,
+)
 from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
 
 
 class OpenMPOffloadMappingFlags(IntFlag):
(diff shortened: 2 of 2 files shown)


tests/filecheck/dialects/omp/ops_invalid.mlir
```diff
@@ -11,4 +11,16 @@ func.func @omp_ordered(%arg0 : i32, %arg1 : i32, %arg2 : i32, %arg3 : i64, %arg4
     return
 }
 
-// CHECK: Operation does not verify: Body of omp.wsloop operation body must consist of one loop nest
+
```
alexarice: Could you remove the extra newline?

xdsl/dialects/omp.py
```diff
@@ -141,6 +149,34 @@ class VariableCaptureKind(StrEnum):
...
+            )
+        if (num_regions := len(op.regions)) != 1:
+            raise VerifyException(
+                f"{op.name} is not a LoopWrapper: has {num_regions} region, expected 1"
+            )
+        if (num_ops := len(op.regions[0].block.ops)) != 1:
+            raise VerifyException(
+                f"{op.name} is not a LoopWrapper: has {num_ops} ops, expected 1"
+            )
+        assert (inner := op.regions[0].block.first_op), "Expected exactly 1 op"
```
alexarice:
```suggestion
        assert (inner := op.regions[0].block.first_op) is not None
```

xdsl/dialects/omp.py
```diff
@@ -141,6 +149,34 @@ class VariableCaptureKind(StrEnum):
...
+    Check that the omp operation is a loop wrapper as defined upstream.
+
+    See external [documentation](https://mlir.llvm.org/docs/Dialects/OpenMPDialect/#loop-associated-directives)
+    """
+
+    def verify(self, op: Operation) -> None:
+        if not op.has_trait(NoTerminator()):
+            raise VerifyException(
+                f"{op.name} is not a LoopWrapper: does not have NoTerminator trait"
+            )
```
alexarice: It feels strange to do this runtime check every time but I don't really see a way round this.

xdsl/dialects/omp.py
```diff
@@ -141,6 +149,34 @@ class VariableCaptureKind(StrEnum):
...
+        if (num_regions := len(op.regions)) != 1:
+            raise VerifyException(
+                f"{op.name} is not a LoopWrapper: has {num_regions} region, expected 1"
+            )
+        if (num_ops := len(op.regions[0].block.ops)) != 1:
+            raise VerifyException(
+                f"{op.name} is not a LoopWrapper: has {num_ops} ops, expected 1"
+            )
+        assert (inner := op.regions[0].block.first_op), "Expected exactly 1 op"
+        if not (inner.has_trait(LoopWrapper()) or isa(inner, LoopNestOp)):
```
alexarice:
```suggestion
        if not (inner.has_trait(LoopWrapper()) or isinstance(inner, LoopNestOp)):
```
`isinstance` is sufficient here.

alexarice (commented):
It would be nice to have tests for each of the ways this can fail (NoTerminator missing and more than one region/operation). These might need to be pytest tests instead so that "bad" operations can be defined.

xdsl/dialects/omp.py
superlopuh: Can subclass NoTerminator?

dk949: Sure, that should work. It will deviate from the upstream definition of the trait, but if we're OK with that I can try that instead.

superlopuh: I think either way works

superlopuh (approved):

xdsl/dialects/omp.py
```diff
@@ -141,6 +147,31 @@ class VariableCaptureKind(StrEnum):
     VLAType = "VLAType"
 
 
+class LoopWrapper(NoTerminator):
```
alexarice, line 150: Sorry, missed a bit of the discussion, is inheriting from traits compatible with how `has_trait` works?

tests/dialects/test_omp.py
```diff
@@ -0,0 +1,82 @@
...
+    traits = traits_def(omp.LoopWrapper())
+
+
+def dummy_wrapper_module(cls: type, *ops_list: list[Operation]):
+    return builtin.ModuleOp([cls(regions=[Region(Block(ops)) for ops in ops_list])])
+
+
+loop_op = omp.LoopNestOp(
+    operands=[[], [], []], regions=[Region(Block([omp.TerminatorOp()]))]
+)
```
alexarice, line 23: Nit: I think this could be a pytest fixture instead of just a global variable that's cloned everywhere.

tests/dialects/test_omp.py
```diff
@@ -0,0 +1,82 @@
...
+@irdl_op_definition
+class DummyLoopWrapper(IRDLOperation):
+    name = "dummy.loop_wrapper"
+    body = region_def()
+
+    traits = traits_def(omp.LoopWrapper())
+
+
+def dummy_wrapper_module(cls: type, *ops_list: list[Operation]):
+    return builtin.ModuleOp([cls(regions=[Region(Block(ops)) for ops in ops_list])])
```
alexarice, line 18: Should be no need to wrap this in a ModuleOp

alexarice (approved):
These tests look great, just missing one for a multi block region I think.

alexarice:
Ah nevermind I was too late

xdsl/dialects/omp.py
superlopuh: Yep

PR #4689, opened 2025-06-30
dialects: (llvm) Fixed bug in GlobalOp where body was always empty when initialised

mesham:
The llvm.GlobalOp operation can have a body associated with it, however the operation's initialisation in xDSL always set this to be empty. Therefore have added an optional body parameter to the function so that this can be set by the caller.

The xDSL MLIR filecheck test for globalop already has a body associated with it

### DIFF
--- a/xdsl/dialects/llvm.py
+++ b/xdsl/dialects/llvm.py
@@ -1430,7 +1430,7 @@ class GlobalOp(IRDLOperation):
             props["section"] = section
 
         if body is None:
-            body = Region([])
+            body = Region()
 
         super().__init__(properties=props, regions=[body])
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/llvm.py
```diff
@@ -1428,7 +1429,10 @@ def __init__(
                 section = StringAttr(section)
             props["section"] = section
 
-        super().__init__(properties=props, regions=[Region([])])
+        if body is None:
+            body = Region([])
```
alexarice:
```suggestion
            body = Region()
```

xdsl/dialects/llvm.py
```diff
@@ -1428,7 +1429,10 @@ def __init__(
...
+
+        super().__init__(properties=props, regions=[body])
```
alexarice:
```suggestion
        super().__init__(properties=props, regions=(body,))
```

PR #4690, opened 2025-06-30
dialects: (omp) add attributes for data mapping

dk949:
Attributes:

* ClauseRequiresKindAttr
* DataSharingClauseAttr
* DeclareTargetAttr
* DeclareTargetCaptureClauseAttr
* DeclareTargetDeviceTypeAttr# PR Template

These attributes help to define how data is mapped in OpenMP between the host and the device.

### DIFF
--- a/tests/filecheck/dialects/omp/attrs.mlir
+++ b/tests/filecheck/dialects/omp/attrs.mlir
@@ -46,8 +46,44 @@
                 #omp<procbindkind spread>,
                 // CHECK-SAME: #omp<procbindkind spread>
                 
-                #omp<orderkind concurrent>
+                #omp<orderkind concurrent>,
                 // CHECK-SAME: #omp<orderkind concurrent>
+
+                #omp<device_type (any)>,
+                // CHECK-SAME: #omp<device_type (any)>
+                #omp<device_type (host)>,
+                // CHECK-SAME: #omp<device_type (host)>
+                #omp<device_type (nohost)>,
+                // CHECK-SAME: #omp<device_type (nohost)>
+
+                #omp<clause_requires none>,
+                // CHECK-SAME: #omp<clause_requires none>
+                #omp<clause_requires reverse_offload>,
+                // CHECK-SAME: #omp<clause_requires reverse_offload>
+                #omp<clause_requires unified_address>,
+                // CHECK-SAME: #omp<clause_requires unified_address>
+                #omp<clause_requires unified_shared_memory>,
+                // CHECK-SAME: #omp<clause_requires unified_shared_memory>
+                #omp<clause_requires dynamic_allocators>,
+                // CHECK-SAME: #omp<clause_requires dynamic_allocators>
+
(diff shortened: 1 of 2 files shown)


xdsl/dialects/omp.py
```diff
@@ -226,6 +253,149 @@ def parse_parameter(cls, parser: AttrParser) -> VariableCaptureKind:
...
+    """
+
+    name = "omp.device_type"
+
+    @classmethod
+    def parse_parameter(cls, parser: AttrParser) -> DeclareTargetDeviceTypeKind:
+        with parser.in_parens():
+            return cast(
+                DeclareTargetDeviceTypeKind, parser.parse_str_enum(cls.enum_type)
+            )
```
alexarice:
```suggestion
            return parser.parse_str_enum(DeclareTargetDeviceTypeKind)
```

alexarice: and similar below

xdsl/dialects/omp.py
```diff
@@ -147,6 +149,31 @@ class VariableCaptureKind(StrEnum):
     VLAType = "VLAType"
 
 
+class DataSharingClauseKind(StrEnum):
+    private = auto()
```
alexarice: I believe we have the convention that all enum fields should be in All caps (even though this would mean that the string field has to be filled in manually.

tests/filecheck/dialects/omp/attrs.mlir
```diff
@@ -46,8 +46,44 @@
                 #omp<procbindkind spread>,
                 // CHECK-SAME: #omp<procbindkind spread>
                 
-                #omp<orderkind concurrent>
+                #omp<orderkind concurrent>,
                 // CHECK-SAME: #omp<orderkind concurrent>
 
+                #omp<device_type (any)>,
```
alexarice, line 52: Not a fan of all the `CHECK-SAME` in this file but I guess this is just following what is already there.

alexarice (approved):
Haven't checked against the upstream definition but this looks good to me

xdsl/dialects/omp.py
dk949: All caps will be OK here, since the fields are automatically made lower case by `StrEnum`.

xdsl/dialects/omp.py
dk949: Nice, that will simplify all the enum attributes a bit. Don't know why I didn't think to do this 😅

alexarice: it's a shame that pyright can't tell that `cls.enum_type` is `DeclareTargetDeviceTypeKind`. Feel we've hit similar issues in lots of places

tests/filecheck/dialects/omp/attrs.mlir
dk949: I was following the example in `arith/arith_attrs.mlir`, as it seemed like a nice, readable way to check all the attributes.

PR #4693, opened 2025-06-30
dialects: (math) Added fastmath property to fp operations in math that were missing it

mesham:
Several operations in the math dialect were missing the fastmath property, this was likely because in a previous version of MLIR that was correct but now all FP operations include the fastmath flag. Therefore have removed the distinction between the base unary/binary FP operation without, and the one with, fastmath, and all FP operations include fastmath now as a property.

Added MLIR based tests with fastmath flag in the IR

### DIFF
--- a/xdsl/dialects/math.py
+++ b/xdsl/dialects/math.py
@@ -25,7 +25,6 @@ from xdsl.irdl import (
     irdl_op_definition,
     operand_def,
     prop_def,
-    prop_def,
     result_def,
     traits_def,
 )


alexarice (approved):
Seems reasonable to me

xdsl/dialects/math.py
```diff
@@ -53,39 +53,23 @@ def __init__(self, operand: Operation | SSAValue):
...
-        )
-
-
-class FloatingPointLikeUnaryMathOperationWithFastMath(
-    FloatingPointLikeUnaryMathOperation, abc.ABC
-):
-    """A generic floating-point-like unary math operation with fastmath flags."""
-
-    fastmath = opt_prop_def(FastMathFlagsAttr)
+    fastmath = opt_prop_def(FastMathFlagsAttr, default_value=FastMathFlagsAttr("none"))
```
superlopuh: Feels like it should not be optional, as it has a default value, which comes pretty close to being the same thing. Does this work?
```suggestion
    fastmath = prop_def(FastMathFlagsAttr, default_value=FastMathFlagsAttr("none"))
```

xdsl/dialects/math.py
```diff
@@ -122,25 +106,7 @@ class FloatingPointLikeBinaryMathOperation(IRDLOperation, abc.ABC):
...
-        )
-
-
-class FloatingPointLikeBinaryMathOperationWithFastMath(
-    FloatingPointLikeBinaryMathOperation, abc.ABC
-):
-    """A generic floating-point-like binary math operation with fastmath flags."""
-
-    fastmath = opt_prop_def(FastMathFlagsAttr)
+    fastmath = opt_prop_def(FastMathFlagsAttr, default_value=FastMathFlagsAttr("none"))
```
superlopuh: same here
```suggestion
    fastmath = prop_def(FastMathFlagsAttr, default_value=FastMathFlagsAttr("none"))
```

xdsl/dialects/math.py
alexarice: optional with default and non optional with default have slightly different mlir behaviour

alexarice: Having said that I just looked up the tablegen and these shouldn't be optional

PR #4696, opened 2025-07-01
core: (operation) Added operation name to error raised when verifying properties and attributes

mesham:
Verification for properties and attributes on an operation raises an exception when trying to set a property/attribute that doesn't exist, however it just reports the property/attribute name that is trying to be set and not the operation that is being acted upon. This makes it quite difficult to track down issues in a large IR (e.g. when parsing). 

This PR adds the operation name to the error message, making it easier to debug issues

### DIFF
--- a/tests/irdl/test_operation_definition.py
+++ b/tests/irdl/test_operation_definition.py
@@ -651,14 +651,6 @@ def test_property_missing():
         op.verify()
 
 
-def test_no_optional_property():
-    """Test verification raises error when a property is missing"""
-
-    op = PropertyOp.create(properties={"attr": StringAttr("test")})
-
-    op.verify()
-
-
 def test_undefined_property():
     """Test verification raises error when a property is missing"""
 


tests/irdl/test_operation_definition.py
```diff
@@ -627,6 +639,41 @@ def test_property_setters():
...
+    ):
+        op.verify()
+
+
+def test_no_optional_property():
+    """Test verification raises error when a property is missing"""
+
+    op = PropertyOp.create(properties={"attr": StringAttr("test")})
+
+    op.verify()
```
superlopuh: No error raised?

superlopuh: Seems like this tests for no error raised, but this is functionality that is already relatively extensively covered, or am I misunderstanding?

mesham: Yer, I am just checking that no error message is raised when we set the mandatory property but omit setting the optional one on the operation. But if this is covered elsewhere then I can remove that?

superlopuh: Yep I'm not sure this is really necessary as a test, feels safe to delete for now.

PR #4697, opened 2025-07-01
dialects:(omp) add omp.simd operation

dk949:
Implements the upstream `omp.simd` operation.

I wasn't sure how to handle the "less than or equal to" constraint on `safelen` and `simdlen`. @superlopuh is there some way to do this without the `verify_`?

### DIFF
--- a/xdsl/dialects/omp.py
+++ b/xdsl/dialects/omp.py
@@ -38,6 +38,7 @@ from xdsl.irdl import (
     RangeOf,
     SameVariadicOperandSize,
     base,
+    eq,
     irdl_attr_definition,
     irdl_op_definition,
     operand_def,
@@ -56,6 +57,7 @@ from xdsl.traits import (
     IsTerminator,
     NoMemoryEffect,
     NoTerminator,
+    RecursiveMemoryEffect,
 )
 from xdsl.utils.exceptions import VerifyException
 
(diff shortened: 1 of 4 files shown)


xdsl/dialects/omp.py
```diff
@@ -593,6 +595,55 @@ def verify_(self) -> None:
...
+    private_vars = opt_operand_def()
+    reduction_vars = opt_operand_def()  # TODO: OpenMP_PointerLikeTypeInterface
+
+    alignments = opt_prop_def(
+        ArrayAttr.constr(RangeOf(base(IntegerAttr[i64]), length=ALIGN_COUNT))
+    )
+    order = opt_prop_def(OrderKindAttr)
+    private_syms = opt_prop_def(ArrayAttr[SymbolRefAttr])
+    reduction_syms = opt_prop_def(ArrayAttr[SymbolRefAttr])
+    simdlen = opt_prop_def(IntegerAttr[i64])
```
alexarice, line 629:
```suggestion
    simdlen = opt_prop_def(IntegerAttr.constr(type=i64))
```
I don't think what you have written actually enforces that the type is i64. Should be the same with the other `IntegerAttr`s

alexarice (approved):

xdsl/dialects/omp.py
dk949: It seems to, if I try to do `simdlen = 64 : ui64` in a test I get `Operation does not verify: Expected attribute i64 but got ui64`

alexarice: Fair enough

PR #4698, opened 2025-07-01
interpreter: (pdl_interp) Return `None` when `get_defining_op` doesn't find an operation

jumerckx:
If the value passed to it is a BlockArgument, for example, that's valid, it should just return None.
https://mlir.llvm.org/docs/Dialects/PDLInterpOps/#pdl_interpget_defining_op-pdl_interpgetdefiningopop

### DIFF
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -206,9 +206,6 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
         if not isinstance(args[0], OpResult):
             return (None,)
         else:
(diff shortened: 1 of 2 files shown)


superlopuh (commented):
Let's add tests for these?

superlopuh:
@math-fehr should SSAValues have `get_defining_op` like in MLIR?

math-fehr:
Can't we just have `value.owner if isinstance(value.owner, Operation) else None`?
That's better imo than having `value.get_defining_op` be a bit redundant

superlopuh:
Sure seems fine to me.

jumerckx:
I've added the tests

superlopuh (approved):

PR #4699, opened 2025-07-01
rewriting: Call `handle_operation_modification` after the modification is carried through.

jumerckx:
This is consistent with what the docstring states.

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/apply-eqsat-pdl/egg_example.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/apply-eqsat-pdl/egg_example.mlir
@@ -1,4 +1,5 @@
-// RUN: xdsl-opt %s -p apply-eqsat-pdl | filecheck %s
+// XFAIL: *
+//RUN: xdsl-opt %s -p apply-eqsat-pdl | filecheck %s
 
 func.func @impl() -> i32 {
   %two   = arith.constant 20  : i32
--- a/tests/filecheck/transforms/apply-eqsat-pdl-interp/egg_example.mlir
+++ b/tests/filecheck/transforms/apply-eqsat-pdl-interp/egg_example.mlir
@@ -1,3 +1,4 @@
+// XFAIL: *
 // RUN: xdsl-opt %s -p apply-eqsat-pdl-interp | filecheck %s
 
 func.func @impl() -> i32 {


xdsl/pattern_rewriter.py
```diff
@@ -142,12 +142,13 @@ def replace_all_uses_with(
         self, from_: SSAValue, to: SSAValue | None, safe_erase: bool = True
     ):
         """Replace all uses of an SSA value with another SSA value."""
-        for use in from_.uses:
-            self.handle_operation_modification(use.operation)
+        modified_ops = [use.operation for use in from_.uses]
```
math-fehr, line 145: Did you meant to use a set here, to instead only call the function once per operation?

math-fehr:
Do you think the CI error comes from this change?

jumerckx:
> Do you think the CI error comes from this change?

Ah yes, the eqsat stuff is actually what prompted this fix, but this will require some coordination with that stuff.
I'll put this as draft and push eqsat changes to this branch later.

jumerckx:
I've disabled the offending tests here.
For the fix I've opened https://github.com/xdslproject/xdsl/pull/4702 which can be merged afterwards?

xdsl/pattern_rewriter.py
superlopuh, line 145: Yeah seems like a set might be preferable
```suggestion
        modified_ops = {use.operation for use in from_.uses}
```

jumerckx: AFAICT, MLIR does the equivalent of a list here.

jumerckx: https://github.com/llvm/llvm-project/blob/5fe63ae9a3fb7502ce7ed076ea14def97d79e85e/mlir/include/mlir/IR/PatternMatch.h#L599C1-L606C4

superlopuh: I guess I'm fine either way it does feel a bit weird to call the callback multiple times for operations with multiple results. Does MLIR dedupe the operations in some other way?

superlopuh: Hmm no deduping in sight

jumerckx: It's not necessarily with operation with multiple results, but rather when an operation uses the same value multiple times.
I imagine there might be cases you want to get notified for each of those so I think it makes sense to not deduplicate.

superlopuh: Sorry that's what I meant, makes sense

PR #4700, opened 2025-07-01
backend: (x86_func) lower return without arguments

qaco:


xdsl/backend/x86/lowering/convert_func_to_x86_func.py
```diff
@@ -122,6 +122,7 @@ class LowerReturnOp(RewritePattern):
     @op_type_rewrite_pattern
     def match_and_rewrite(self, op: func.ReturnOp, rewriter: PatternRewriter):
         if len(op.arguments) == 0:
```
superlopuh:
```suggestion
        if not op.arguments:
```

PR #4703, opened 2025-07-02
backend: (x86) convert index to register

qaco:


xdsl/backend/x86/lowering/helpers.py
```diff
@@ -39,8 +40,9 @@ def vector_type_to_register_type(
...
 def scalar_type_to_register_type(value_type: Attribute) -> type[RegisterType]:
     assert not isinstance(value_type, ShapedType)
-    assert isinstance(value_type, FixedBitwidthType)
-    if value_type.bitwidth <= 64:
+    if (
+        isinstance(value_type, FixedBitwidthType) and value_type.bitwidth <= 64
+    ) or isinstance(value_type, IndexType):
         return x86.register.GeneralRegisterType
     else:
         raise DiagnosticException("Not implemented for bitwidth larger than 64.")
```
superlopuh, line 48: How about using `CompileTimeFixedBitwidthType` instead, and passing the bitwidth as a parameter? The usual name in MLIR is `index_bitwidth`, that way if the type is representable in less than the register size it gets put in a register.

qaco: Passing the bitwidth as a parameter of the function ```scalar_type_to_register_type``` ?

superlopuh: yeah, that way the 64 isn't hardcoded

superlopuh: It can be the default value though

qaco: But here the 64 is not the index bitwidth, it's the x86 register bitwidth, so it seems reasonable to hard-code it, no ?

superlopuh: Yeah actually you're right.

superlopuh (approved):

xdsl/backend/x86/lowering/helpers.py
superlopuh: I still think that it would be worth changing this function to take the index bitwidth as an argument, but that can be done in a separate PR. It's mostly the hardcoding of 64 that tripped me up, it feels like it should be an argument instead, but I guess not urgent.

qaco: Cool

PR #4709, opened 2025-07-03
dialects:(omp) add task based data mapping ops

dk949:
Also adds an extra constraint on mapped values which was missed in the previous PR: When the `map_type` is constrained to one of a number of flags, these constraints are on the `var_ptr` operand to `omp.map.info`, not on the result of `omp.map.info`. So we need to keep track of which flags have already been used with which operands.

**Operations added:**

* TargetEnterDataOp
* TargetExitDataOp
* TargetUpdateOp

### DIFF
--- a/xdsl/dialects/omp.py
+++ b/xdsl/dialects/omp.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from enum import IntFlag, auto
 from typing import ClassVar
 
@@ -131,9 +132,7 @@ def verify_map_vars(
     op_name: str,
     *,
     disallowed_types: OpenMPOffloadMappingFlags = OpenMPOffloadMappingFlags.NONE,
-    one_of: OpenMPOffloadMappingFlags = OpenMPOffloadMappingFlags.NONE,
 ):
-    mapped = dict[Operand, OpenMPOffloadMappingFlags]()
     for var in vars:
         if not isinstance(owner := var.owner, MapInfoOp):
             raise VerifyException(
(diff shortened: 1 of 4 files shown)


xdsl/dialects/omp.py
```diff
@@ -140,11 +142,16 @@ def verify_map_vars(
...
             if owner.map_type.value.data & t:
                 raise VerifyException(f"Cannot have map_type {t.name} in {op_name}")
-        if one_of and (owner.map_type.value.data & one_of).bit_count() != 1:
-            one_of_names = ", ".join(bit.name for bit in one_of if bit.name)
-            raise VerifyException(
-                f"{op_name} expected to have exactly one of {one_of_names} as map_type"
+        if one_of:
+            mapped[owner.var_ptr] = (
+                mapped.setdefault(owner.var_ptr, OpenMPOffloadMappingFlags.NONE)
+                | owner.map_type.value.data
```
superlopuh: What does this change do?

dk949: It's to keep track of the flags used with different operands to to `omp.map.info`. It checks for this case:

```mlir
%mapped1 = "omp.map.info"(%ptr) map_type = to
%mapped2 = "omp.map.info"(%ptr) map_type = from
"omp.target_update"(%mapped1, %mapped2) 
// ^ This should be an error, because %mapped1 and %mapped2 are mapping the same pointer with types TO and FROM at the same time
```

The way I did it previously only checked that the same `omp.map.info` op didn't use both types at the same time.

superlopuh: Why use setdefault and not a defaultdict? Also, would it not be clearer to separate this check out, and check separately for whether all the operands have the same flag?

dk949: The answer to the first question is: because I didn't know it existed, will change to using it instead.

For the second: I put it in this function because this is where I check the owners of the mapped values. I can move it into the `TargetUpdateOp` instead, since it's the only place that uses this functionality.

It would mean looping over the `map_vars` again and either doing an `assert` with `isinstance` or a `cast` on the owner. But it would simplify the logic significantly.

dk949:
Oops, forgot to change the `ops_invalid` test

PR #4714, opened 2025-07-04
dialects: Add a reshape_ops_utils.py

Jimmy2027:
Used by #4245

Inspired by the MLIR counterpart:
https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Utils/ReshapeOpsUtils.h

I'm thinking that in theory reshape utils can be shared by any dialect having reshape-like ops

### DIFF
--- a/tests/filecheck/dialects/tensor/invalid_ops.mlir
+++ b/tests/filecheck/dialects/tensor/invalid_ops.mlir
@@ -85,5 +85,5 @@ builtin.module {
 
 // -----
 
-// CHECK: Expected ArrayAttr[ArrayAttr[IntegerAttr[I64]]] but got array
+// CHECK: expected integer >= 0, got -2
 %res_collapse2 = tensor.collapse_shape %t0 [ [-2, 3], [0, 1] ] : tensor<4x1xf32> into tensor<4x1xf32>
--- a/tests/filecheck/dialects/tensor/invalid_ops.mlir
+++ b/tests/filecheck/dialects/tensor/invalid_ops.mlir
@@ -85,7 +85,5 @@ builtin.module {
 
 // -----
 
-// CHECK: The following constraints were not satisfied:
-// CHECK-NEXT: expected integer >= 0, got -2
-// CHECK-NEXT: All inner arrays must be contiguous: [[-2 : i64, 3 : i64], [0 : i64, 1 : i64]]
+// CHECK: Expected ArrayAttr[ArrayAttr[IntegerAttr[I64]]] but got array
 %res_collapse2 = tensor.collapse_shape %t0 [ [-2, 3], [0, 1] ] : tensor<4x1xf32> into tensor<4x1xf32>
--- a/xdsl/dialects/utils/reshape_ops_utils.py
+++ b/xdsl/dialects/utils/reshape_ops_utils.py
@@ -10,13 +10,22 @@ from typing_extensions import TypeVar
 
 from xdsl.dialects.builtin import I64, Annotated, ArrayAttr, IntegerAttr
 from xdsl.ir import Attribute
-from xdsl.irdl import AtLeast, AttrConstraint, ConstraintContext
+from xdsl.irdl import AtLeast, AttrConstraint, ConstraintContext, GenericAttrConstraint
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import isa
 
+_ReassociationAttrBase = ArrayAttr[
+    ArrayAttr[
+        Annotated[
+            IntegerAttr[I64],
+            IntegerAttr.constr(value=AtLeast(0)),
+        ]
+    ]
+]
+
 
 @dataclass(frozen=True)
-class ContiguousArrayOfIntArray(GenericAttrConstraint[ArrayAttr[ArrayAttr[IntegerAttr]]]):
+class ContiguousArrayOfIntArray(GenericAttrConstraint[_ReassociationAttrBase]):
     """
     Enforce an ArrayAttr of ArrayAttr[IntegerAttr] to contain contiguous integer values across all inner arrays.
     For example: [[0, 1], [2, 3]] is valid, but [[3, 4], [0, 1]] is not.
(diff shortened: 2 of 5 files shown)


xdsl/dialects/utils/reshape_ops_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+from xdsl.dialects.builtin import I64, Annotated, ArrayAttr, IntegerAttr
+from xdsl.ir import Attribute
+from xdsl.irdl import AtLeast, AttrConstraint, ConstraintContext
+from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
+
+
+@dataclass(frozen=True)
+class ContiguousArrayOfIntArray(AttrConstraint):
```
superlopuh: While we're here, let's fix this bit, it should be enough to have it like this:
```suggestion
class ContiguousArrayOfIntArray(GenericAttrConstraint[ArrayAttr[ArrayAttr[IntegerAttr]]]):
```
Then can remove the Annotated below, and use the constraint directly.

superlopuh:
Did you mean for this one to be a draft? I think modulo the generic constraint it should be good to merge.

Jimmy2027:
yeah it was a bit late and I wanted to have another look in the morning :)

also, I'm thinking now that the `ReassociationAttr` from [memref.py](https://github.com/xdslproject/xdsl/blob/main/xdsl/dialects/memref.py#L478) can be taken from the reshape utils

xdsl/dialects/utils/reshape_ops_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Enforce an ArrayAttr of ArrayAttr[IntegerAttr] to contain contiguous integer values across all inner arrays.
+    For example: [[0, 1], [2, 3]] is valid, but [[3, 4], [0, 1]] is not.
+    An empty inner array is considered contiguous.
+    """
+
+    def verify(
+        self, attr: Attribute, constraint_context: ConstraintContext | None = None
+    ) -> None:
+        if not isa(attr, ArrayAttr[ArrayAttr[IntegerAttr]]):
```
superlopuh: I would actually move the I64 thing in here, and have the _ReassociationAttrBase be a plain array of array of integer attr

xdsl/dialects/utils/reshape_ops_utils.py
```diff
@@ -0,0 +1,54 @@
...
+from xdsl.irdl import AtLeast, AttrConstraint, ConstraintContext, GenericAttrConstraint
+from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
+
+_ReassociationAttrBase = ArrayAttr[
+    ArrayAttr[
+        Annotated[
+            IntegerAttr,
+            IntegerAttr.constr(value=AtLeast(0)),
+        ]
```
superlopuh:
```suggestion
        IntegerAttr,
```

xdsl/dialects/utils/reshape_ops_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Enforce an ArrayAttr of ArrayAttr[IntegerAttr] to contain contiguous integer values across all inner arrays.
+    For example: [[0, 1], [2, 3]] is valid, but [[3, 4], [0, 1]] is not.
+    An empty inner array is considered contiguous.
+    """
+
+    def verify(
+        self, attr: Attribute, constraint_context: ConstraintContext | None = None
+    ) -> None:
+        if not isa(attr, ArrayAttr[ArrayAttr[IntegerAttr[I64]]]):
```
superlopuh: this is now missing the at least 0 bit

Jimmy2027:
the error message becomes quite a bit more generic: 99f0a30158037629e9350c8f1b24e814f5d70725
Not sure if that is better

xdsl/dialects/utils/reshape_ops_utils.py
```diff
@@ -0,0 +1,57 @@
...
+                    Annotated[
+                        IntegerAttr[I64],
+                        IntegerAttr.constr(value=AtLeast(0)),
+                    ]
+                ]
+            ],
+        ):
+            raise VerifyException(
+                f"Expected ArrayAttr[ArrayAttr[IntegerAttr[I64]]] but got {getattr(attr, 'name', type(attr))}"
+            )
```
superlopuh:
```suggestion
_CONTIGUOUS_ARRAY_TYPE_CONSTRAINT = irdl_to_attr_constraint(ArrayAttr[
	    ArrayAttr[
	        Annotated[
	            IntegerAttr[I64],
	            IntegerAttr.constr(value=AtLeast(0)),
	        ]
	    ]
	]
)

@dataclass(frozen=True)
class ContiguousArrayOfIntArray(
    GenericAttrConstraint[ArrayAttr[ArrayAttr[IntegerAttr]]]
):
    """
    Enforce an ArrayAttr of ArrayAttr[IntegerAttr] to contain contiguous integer values across all inner arrays.
    For example: [[0, 1], [2, 3]] is valid, but [[3, 4], [0, 1]] is not.
    An empty inner array is considered contiguous.
    """

    def verify(
        self, attr: Attribute, constraint_context: ConstraintContext | None = None
    ) -> None:
        _CONTIGUOUS_ARRAY_TYPE_CONSTRAINT.verify(attr)
        attr = cast(ArrayAttr[ArrayAttr[IntegerAttr]], attr)
```

superlopuh: Let's use the original verification error instead of adding our own one

Jimmy2027: yeah that looks better :) thanks!

tests/filecheck/dialects/tensor/invalid_ops.mlir
```diff
@@ -85,7 +85,5 @@ builtin.module {
 
 // -----
 
-// CHECK: The following constraints were not satisfied:
-// CHECK-NEXT: expected integer >= 0, got -2
-// CHECK-NEXT: All inner arrays must be contiguous: [[-2 : i64, 3 : i64], [0 : i64, 1 : i64]]
+// CHECK: expected integer >= 0, got -2
```
Jimmy2027, line 88: this test became a bit confusing because it does two things wrong but checks only for one error message

superlopuh: I think it's fine

superlopuh (approved):
Thank you!

PR #4716, opened 2025-07-04
interpreter: (eqsat) Don't backtrack in rewriter's `finalize`

jumerckx:
`finalize` is interpreted by going back to the last backtrackingpoint. This is correct for `finalize`s that occur in the matching portion of a pdl_interp file, but isn't for the finalize at the end of an actual rewrite.

Previously, this lead to very deep recursion, since `finalize` would not return from the call but rather jump back to a point in the matcher. We now keep track of whether the interpreter is in a matching context. If it isn't we return, no backtracking.

### DIFF
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -99,6 +99,10 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
     merge_list: list[MergeTodo] = field(default_factory=list[MergeTodo])
     """List of e-classes that should be merged by `apply_matches` after the pattern matching is done."""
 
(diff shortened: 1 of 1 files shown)


superlopuh:
Is there any way to test this at a high level? Not keeping track of `is_matching` or whatever, I'm thinking of a minimal pdl interp test that would fail before this PR and pass afterwards.

jumerckx:
> Is there any way to test this at a high level? Not keeping track of `is_matching` or whatever, I'm thinking of a minimal pdl interp test that would fail before this PR and pass afterwards.

As I understand it, that's not straightforward to do as execution was still correct before. The only way I see is to have an example where we hit Python's recursion limit but that would be a slow test.

PR #4717, opened 2025-07-05
core: make Region hashable again

superlopuh:
It seems like this functionality was deleted at some point.


superlopuh:
Yeah I was thinking that, I have a feeling I removed them in the first place, and regret doing so.

PR #4718, opened 2025-07-06
API: add region_utils and used_values_defined_above

superlopuh:
MLIR has a number of utils files with useful functions. I think we would benefit from this kind of pattern to make the surface area of core.py smaller. This file adds a function I'll need to port the vectorization infrastructure.

### DIFF
--- a/xdsl/ir/region_utils.py
+++ b/xdsl/ir/region_utils.py
@@ -1,0 +1,40 @@
+"""
+Collection of helpers for dealing with `Region`s, similar to `RegionUtils.cpp` in MLIR.
+"""
+
+from xdsl.ir import Region
+
+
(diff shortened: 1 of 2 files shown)


xdsl/ir/region_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    if limit is not None:
+        assert limit.is_ancestor(region), (
+            "expected isolation limit to be an ancestor of the given region"
+        )
+    else:
+        limit = region
+
+    # Collect proper ancestors of `limit` upfront to avoid traversing the region
+    # tree for every value.
+    proper_ancestors = set[Region]()
```
superlopuh, line 23: reason for the Region PR from before

math-fehr (approved):

PR #4719, opened 2025-07-07
dialects:(omp) updated all ops for MLIR 20.1.7

dk949:
Some operations had missing or incorrectly typed attributes for MLIR
20.1.7.

In some cases attribute constraints were missing/too wide.

All ops were missing verification for the number of block arguments.

### DIFF
--- a/xdsl/dialects/omp.py
+++ b/xdsl/dialects/omp.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from collections import defaultdict
 from enum import IntFlag, auto
 from typing import ClassVar
@@ -61,7 +62,6 @@ from xdsl.traits import (
     IsTerminator,
     NoMemoryEffect,
     NoTerminator,
-    OpTrait,
     Pure,
     RecursiveMemoryEffect,
 )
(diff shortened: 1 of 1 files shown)


xdsl/dialects/omp.py
```diff
@@ -229,6 +243,61 @@ def verify(self, op: Operation) -> None:
         return super().verify(op)
 
 
+class BlockArgOpenMPOp(IRDLOperation):
```
superlopuh: We use the `Operation` suffix as a naming convention for abstract base classes.
```suggestion
class BlockArgOpenMPOperation(IRDLOperation, ABC):
```

xdsl/dialects/omp.py
```diff
@@ -229,6 +244,61 @@ def verify(self, op: Operation) -> None:
...
+    Verifies that the operation has the appropriate number of block arguments corresponding to the following operands:
+    `host_eval`, `in_reduction`, `map`, `private`, `reduction`, `task_reduction`, `use_device_addr` and `use_device_ptr`.
+    """
+
+    def verify_(self) -> None:
+        if len(self.regions) < 1:
+            raise VerifyException(f"Expected {self.name} to have at least 1 region")
+        if len(self.regions[0].blocks) < 1:
+            raise VerifyException(
+                f"Expected {self.name} to have at least 1 block in it's first region"
```
```suggestion
                f"Expected {self.name} to have at least 1 block in its first region"
```

xdsl/dialects/omp.py
```diff
@@ -229,6 +244,61 @@ def verify(self, op: Operation) -> None:
...
+        return 0
+
+    def num_task_reduction_block_args(self) -> int:
+        return 0
+
+    def num_use_device_addr_block_args(self) -> int:
+        return 0
+
+    def num_use_device_ptr_block_args(self) -> int:
+        return 0
```
superlopuh: Will there be a use for these other than verification? Seems like it could be a single abstract method called `expected_block_arg_count` that all subclasses must implement, to make it clear that it's for verification purposes.

tests/filecheck/dialects/omp/ops.mlir
```diff
@@ -110,12 +110,6 @@ builtin.module {
         omp.yield
       }) : (index, index, index) -> ()
     }) : () -> ()
-    "omp.wsloop"() <{"inclusive", operandSegmentSizes = array<i32: 0, 0, 0, 0, 0, 0, 0>}> ({
```
alexarice, line 113: why is this no longer needed?

tests/filecheck/mlir-conversion/with-mlir/dialects/omp/ops.mlir
```diff
@@ -110,12 +110,6 @@ builtin.module {
         omp.yield
       }) : (index, index, index) -> ()
     }) : () -> ()
-    "omp.wsloop"() <{"inclusive", operandSegmentSizes = array<i32: 0, 0, 0, 0, 0, 0, 0>}> ({
```
alexarice, line 113: Similarly here, is this no longer valid IR? Can we replace the test with something meaningful?

xdsl/dialects/omp.py
```diff
@@ -495,26 +594,50 @@ class ParallelOp(IRDLOperation):
...
+    def num_reduction_block_args(self) -> int:
+        return len(self.reduction_vars)
+
 
 @irdl_op_definition
 class YieldOp(AbstractYieldOperation[Attribute]):
     name = "omp.yield"
 
-    traits = traits_def(IsTerminator())
+    assembly_format = "attr-dict (`(`$arguments^ `:` type($arguments)`)`)?"
```
alexarice: In the documentation the attr-dict is at the end of the format

dk949: Thanks, will double check

tests/filecheck/mlir-conversion/with-mlir/dialects/omp/ops.mlir
dk949: No, `inclusive` is no longer an attribute of `wsloop`.

dk949: Sorry the "no" was relating to whether we still need this test, not whether the tests can be replaced with something else. I'll look into that

xdsl/dialects/omp.py
dk949: Yes, i think that'd work. The way I've done it here is based on how upstream does it. 

Another alternative I've considered was making this a trait and using `getattr(op, "operand_name" , 0)`. This would rely on magic operand names, but imo would make it harder to forget to update the count in the future.

xdsl/dialects/omp.py
dk949: Yes, you are right, it's `scf.yield` that has it upfront.

xdsl/dialects/omp.py
superlopuh: Yeah it seems like making it an abstract count method, meaning every operation needs to implement it, would be nice. Should be lower overhead since it's just one function.

xdsl/dialects/omp.py
```diff
@@ -229,6 +244,45 @@ def verify(self, op: Operation) -> None:
...
+            raise VerifyException(
+                f"{op.name} expected to have {expected} block argument(s), got {actual}"
+            )
+
+    @staticmethod
+    def __get_len(op: Operation, name: str):
+        if field := getattr(op, name, None):
+            assert isinstance(field, VarOperand)
+            return len(field)
+        return 0
```
superlopuh: I would much rather the thing we discussed on the previous diff, getting ops to implement one well-typed function rather than using getattr

dk949: I can revert it back to how it was. 

I will say, however, that doing it this way has revealed a few of these I missed. I'd say this is much less error prone, and I do `assett` the type of the attribute

dk949: Not "how it was", but the single abstract method version.

superlopuh: If you mark it as `@abstractmethod` the type system will warn you if you miss anything, so should be pretty safe also, while catching any potential future in which the field names change.

superlopuh (approved):
Thank you!

PR #4720, opened 2025-07-07
backend:(csl) add extension point for the printer

dk949:
This extension point would allow someone to subclass `CslPrintContext` in order to change how existing operations are printed or to add support for new ones.

Most of the functionality of the printer can be modified by overriding existing methods, but the logic for printing operations is all in `print_block` and taking it out would be a much more substantial change.

The sub-classed printer can also not be passed to `print_to_csl` as an optional third argument.

### DIFF
--- a/tests/backend/test_csl.py
+++ b/tests/backend/test_csl.py
@@ -34,14 +34,13 @@ class NewPrinter(print_csl.CslPrintContext):
             case _:
                 return super().mlir_type_to_csl_type(type_attr)
 
-    def _print_op_override(self, op: Operation) -> bool:
+    def print_op(self, op: Operation):
         match op:
             case test.TestOp(res=res, ops=ops):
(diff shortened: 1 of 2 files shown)


dk949:
Sure that should be fine. The main reason I didn't do that is that it'd result in a 350 line change and affect the git history for that entire block. It will result in a cleaner interface for the child class though, like [mlir_type_to_csl_type](https://github.com/xdslproject/xdsl/pull/4720/files#diff-90a35d9a84b8771b7c8be6ec99cf0da276e68499fbc975332333cdfe2e5fb32dR30-R35) in the test.

dk949:
FYI here's a link to the diff with whitespace changes ignored: <https://github.com/xdslproject/xdsl/pull/4720/files?w=1>

n-io (approved):
LGTM, thanks

PR #4721, opened 2025-07-08
docs: open marimo notebooks in full screen

superlopuh:
I don't think there's a way to make it non-broken, so I propose just opening them full screen when clicking on the link.


scripts/gen_ref_pages.py
```diff
@@ -46,26 +46,21 @@
 
 docs_root = root / "docs"
 
+
+def create_marimo_app_url(code: str, mode: str = "read") -> str:
```
superlopuh, line 50: moved from blocks.py

superlopuh:
https://xdsl--4721.org.readthedocs.build/4721/marimo/

alexarice:
> https://xdsl--4721.org.readthedocs.build/4721/marimo/

Should there be some notebooks here?

superlopuh:
They're on the left

alexarice:
Right, this shows up very badly with a narrow monitor as you have to expand the menu. What's broken at the moment?

superlopuh:
I had to do an ugly hack to do with them taking up approximately the size of the page. There's no way to make them take the full width and height of the mkdocs window due to iframe limitations. The proposed approach makes the notebook as big as the browser window, which renders best in all setups that I've tried. The user can just navigate back to the documentation page to go to other notebooks, I don't think having the notebook embedded is much better than this proposed solution.

Try the existing situation here:

https://xdsl.readthedocs.io/stable/marimo/

alexarice:
I kind of like the existing to be honest, is it really that bad/ is this blocking other stuff?

superlopuh:
I think it's really not good, have you tried it on your laptop screen? It's really cramped for space. I also like that this gets rid of the copy/paste of the marimo extension which is one less thing to maintain in our docs.

superlopuh:
Just as an extra bit of information if it sways you, the small window size was a common comment from people reading the notebooks, with the recommended workaround being to right-click the background of the notbeook and open the iframe in a new tab, this approach avoids this entirely.

The reason I looked into this now BTW is that I was trying to export the notebooks using the official ecport APIs, which would let us avoid the artificial restriction on notebook size that we currently have with the wasm URL encoding, and would let us drop the dependency on the compression library I ported that we currently depend on.

docs/__init__.py
```diff
@@ -1,13 +0,0 @@
-"""
-marimo documentation package.
-
-Copied from https://github.com/marimo-team/marimo/blob/f25319284d439e8d809fab421f27e9226cb00a65/docs/__init__.py#L3
-
-Currently mkdocs-marimo doesn't ship the best way to embed notebooks in mkdocs, they use
-this mechanism instead, which we should replace with mkdocs-marimo when they update the
```
alexarice, line 7: Did they ever update anything? (guess not given this PR)

alexarice (approved):
Code looks fine to me, I don't feel hugely strongly so happy to go ahead if you feel this is an improvement

docs/__init__.py
superlopuh: They did but didn't fix the fundamental problem I had, which is that the code block errored out due to using a different encoding of wasm to the one they do in javascript. My understanding is that the recommendation is to use the export command to do this directly, but I couldn't find a nice way to do it during the mkdocs build process, hence this half-way PR to make sure we're happy with the interface. I'll let @math-fehr and @compor weigh in case I'm in the minority.

math-fehr (approved):
I agree it looks quite better on full screen, especially for laptops (which is our target audience during workshops)

PR #4722, opened 2025-07-08
CI: add back weekly remake lockfile

superlopuh:
Renovate only updates the direct dependency and not the transitive dependencies.

### DIFF
--- a/.github/workflows/remake-lockfile.yml
+++ b/.github/workflows/remake-lockfile.yml
@@ -1,0 +1,39 @@
+name: Remake UV Lockfile
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/remake-lockfile.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v3
+
+      - run: |
+          rm uv.lock
```
alexarice: Why not just run `uv lock -U`?

superlopuh: !!

PR #4723, opened 2025-07-08
documentation: (Toy) use xDSL's shape inference instead of a custom one

superlopuh:
A bit of cleanup. I'll try to clean up the Toy tutorial before moving it to a marimo notebook.


docs/Toy/toy/dialects/toy.py
```diff
@@ -121,11 +111,11 @@ def get_data(self) -> list[float]:
         return list(self.value.get_values())
 
 
-class InferAddOpShapeTrait(ToyShapeInferenceTrait):
-    @classmethod
-    def infer_shape(cls, op: Operation) -> None:
+class InferAddOpShapeInferencePattern(RewritePattern):
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter, /):
```
alexarice, line 115: Is it worth using the decorator here, or is the idea to keep the toy files as simple as possible?

superlopuh: I wanted to keep diff small, and at this point the operation is still not defined so using the decorator breaks as it eagerly evals the type annotation.

PR #4728, opened 2025-07-08
API: (irdl) remove unused IRDLAnnotations

superlopuh:
We new use field defs for this.

### DIFF
--- a/xdsl/irdl/error.py
+++ b/xdsl/irdl/error.py
@@ -1,9 +1,0 @@
-from enum import Enum
-
-
-class IRDLAnnotations(Enum):
-    ParamDefAnnot = 1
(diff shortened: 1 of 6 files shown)


xdsl/irdl/error.py
```diff
@@ -1,9 +0,0 @@
-from enum import Enum
-
-
-class IRDLAnnotations(Enum):
-    ParamDefAnnot = 1
-    AttributeDefAnnot = 2
-    OptAttributeDefAnnot = 3
-    SingleBlockRegionAnnot = 4
-    ConstraintVarAnnot = 5
```
alexarice, line 9: Interesting file

PR #4731, opened 2025-07-08
misc: IRNode = Operation | Region | Block

superlopuh:
There's a few places where types are actually one of a limited set, I propose to migrate to this pattern when this is the case. Two more places to do this: Attribute = ParametrizedAttribute | Data, and then some AST things in Toy.

As a drive-by fix, I replaced the recursive functions on IRNode with loopy ones, this isn't strictly necessary, but should be a bit faster.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -723,7 +723,7 @@ class ErasedSSAValue(SSAValue):
 @dataclass(init=False)
 class _IRNode(ABC):
     def is_ancestor(self, op: IRNode) -> bool:
-        "Returns true if the IRNode is an ancestor of another IRNode."
+        "Returns true if `self` is an ancestor of `op`."
         curr = op
         while curr is not None:
             if curr is self:
@@ -732,7 +732,10 @@ class _IRNode(ABC):
         return False
 
     def get_toplevel_object(self) -> IRNode:
-        """Get the operation, block, or region ancestor that has no parents."""
+        """
+        Get the ancestor of `self` that has no parent.
+        This can be an Operation, Block, or Region.
+        """
         current = self
         while (parent := current.parent_node) is not None:
             current = parent
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -21,6 +21,7 @@ from typing import (
     Generic,
     NoReturn,
     Protocol,
+    TypeAlias,
     cast,
     get_args,
     get_origin,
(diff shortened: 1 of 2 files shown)


xdsl/ir/core.py
```diff
@@ -720,20 +721,22 @@ def owner(self) -> Operation | Block:
 
 
 @dataclass(init=False)
-class IRNode(ABC):
+class _IRNode(ABC):
     def is_ancestor(self, op: IRNode) -> bool:
         "Returns true if the IRNode is an ancestor of another IRNode."
```
alexarice: If we're doing drive by fixes can we improve these docstrings? From reading this one I'd expect the roles of `op` and `self` to be flipped

xdsl/ir/core.py
```diff
@@ -720,20 +721,22 @@ def owner(self) -> Operation | Block:
...
-        return self.is_ancestor(parent)
+        curr = op
+        while curr is not None:
+            if curr is self:
+                return True
+            curr = curr.parent_node
+        return False
 
     def get_toplevel_object(self) -> IRNode:
         """Get the operation, block, or region ancestor that has no parents."""
```
```suggestion
        """Get the parent operation, block, or region ancestor that has no parents."""
```

xdsl/ir/core.py
```diff
@@ -720,20 +721,22 @@ def owner(self) -> Operation | Block:
...
-        if (parent := self.parent_node) is None:
-            return self
-        return parent.get_toplevel_object()
+        current = self
+        while (parent := current.parent_node) is not None:
+            current = parent
+        return current  # pyright: ignore[reportReturnType]
```
alexarice, line 742: Can pyright really not get this right?

xdsl/ir/core.py
```diff
@@ -720,20 +721,22 @@ def owner(self) -> Operation | Block:
...
+                return True
+            curr = curr.parent_node
+        return False
 
     def get_toplevel_object(self) -> IRNode:
         """Get the operation, block, or region ancestor that has no parents."""
-        if (parent := self.parent_node) is None:
-            return self
-        return parent.get_toplevel_object()
+        current = self
```
alexarice, line 739: Think the aliasing isn't needed btw, python is call by value

alexarice (approved):

xdsl/ir/core.py
superlopuh: This is for Pyright, otherwise we're assigning a supertype to a value of a specified type. It doesn't need you to provide a type annotation, but infers the type of the newly created value from the context.

xdsl/ir/core.py
superlopuh: This kind of makes sense, as here we've not constrained _IRNode to be one of Operation, Block, or Region.

PR #4734, opened 2025-07-08
core: add context to _consume_token assertion error

superlopuh:
Also use _parse_token in ToyParser, turns out I used the wong one inadvertantly.

### DIFF
--- a/docs/Toy/toy/frontend/parser.py
+++ b/docs/Toy/toy/frontend/parser.py
@@ -75,6 +75,13 @@ class ToyParser(GenericParser[ToyTokenKind]):
             self.raise_error(f"Expected '{pattern}'", token.span.start, token.span.end)
         return token
 
+    def pop_token(self, tokenType: ToyTokenKind) -> ToyToken:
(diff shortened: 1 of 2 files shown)


superlopuh:
I was also confused and decided to look in MLIR and it turns out that we inherited this API from there. I assume that it's useful for when an unexpected token is not an input error but a programmer error in some way, maybe when the text comes not from user input but from some other generated source? I wonder if @math-fehr knows the answer.

alexarice:
Do we use the functionality anywhere?

superlopuh:
It seems that we only use it in places where we are certain that the type is correct by already checking it, making the check redundant.

math-fehr:
Yeah, you are supposed to use `consume_token` whenever you already checked that the next token is the correct one.

superlopuh:
But then what's the purpose of passing it as an argument to the function? Should we ever do that/support it? Feels like that's a rare enough use-case that the caller can just assert it client-side instead of having a branchy helper.

PR #4735, opened 2025-07-09
misc: remove cast from builtin.py

superlopuh:
Much of the time it's a matter of using isa instead of isinstance, sometimes it's just a pyright ignore, when we already know from control flow that the value is of the right type.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -2233,17 +2233,12 @@ class MemRefType(
         layout: MemRefLayoutAttr | NoneAttr = NoneAttr(),
         memory_space: Attribute = NoneAttr(),
     ):
-        s: ArrayAttr[IntAttr]
-        if isa(shape, ArrayAttr[IntAttr]):
-            # Temporary cast until Pyright is fixed to not infer ArrayAttr[int] as a
-            # possible value for shape
-            s = shape
-        else:
-            s = ArrayAttr(
+        if not isa(shape, ArrayAttr[IntAttr]):
+            shape = ArrayAttr(
                 [IntAttr(dim) if isinstance(dim, int) else dim for dim in shape]
             )
         super().__init__(
-            s,
+            shape,
             element_type,
             layout,
             memory_space,


xdsl/dialects/builtin.py
```diff
@@ -153,11 +152,8 @@ def __init__(self, param: Iterable[AttributeCovT]) -> None:
 
     @classmethod
     def parse_parameter(cls, parser: AttrParser) -> tuple[AttributeCovT, ...]:
-        data = parser.parse_comma_separated_list(
-            parser.Delimiter.SQUARE, parser.parse_attribute
-        )
-        # the type system can't ensure that the elements are of type _ArrayAttrT
-        return cast(tuple[AttributeCovT, ...], tuple(data))
+        # This functionality is provided by the attribute parser.
+        raise NotImplementedError
```
math-fehr, line 156: Should we have an assertion instead?
I feel `NotImplementedError` looks like something could be implemented, while this is not the case.

math-fehr (approved):
Nice!

xdsl/dialects/builtin.py
superlopuh: Hmm, `...`?

alexarice: If we're doing this I'd prefer both the parse and print to be not implemented.

xdsl/dialects/builtin.py
```diff
@@ -2241,10 +2234,10 @@ def __init__(
         memory_space: Attribute = NoneAttr(),
     ):
         s: ArrayAttr[IntAttr]
-        if isinstance(shape, ArrayAttr):
+        if isa(shape, ArrayAttr[IntAttr]):
             # Temporary cast until Pyright is fixed to not infer ArrayAttr[int] as a
             # possible value for shape
-            s = cast(ArrayAttr[IntAttr], shape)
+            s = shape
         else:
```
alexarice: Can replace this by one `if not isa(shape, ArrayAttr): shape = ...`

alexarice: And remove the comment about casts

alexarice (approved):
Seems reasonable.

xdsl/dialects/builtin.py
superlopuh: I'll do a separate PR for this, feels like we missed a step here.

superlopuh:
Actually I'll first merge this then open a new one to address the printing and parsing.

PR #4739, opened 2025-07-09
backend: (x86) add 64-bit allocatable registers

superlopuh:
Now that `EAX` and friends aren't there, we can allocate general registers.


qaco (approved):

qaco:
I still encounter this ```DiagnosticException: Cannot allocate registers to the same register ['!x86.reg<rax>', '!x86.reg<rcx>']``` error !

superlopuh:
Can you please open a WIP PR for me to debug?

PR #4741, opened 2025-07-09
core: don't use try/finally for delimited helper

superlopuh:
If an error is raised in the yield, it is then swallowed, and another is raised with an expected delimiter, which is not ideal.


alexarice (approved):
Interesting, I thought this was needed for things like:
```python
printer.in_angle_brackets():
   ...
   return
...
```
but I guess not.

math-fehr (approved):

superlopuh:
It's really for things like closing file handles etc, so overkill/counterproductive here where we just want something to happen after the yielded scope is done.

PR #4744, opened 2025-07-09
dialects: (builtin) add UnknownLoc for unknown locations

superlopuh:
This is a step towards more location support, the next PR will introduce FileLineLoc, so this PR also prepares the way for that a little bit without changing the functionality.

### DIFF
--- a/xdsl/parser/attribute_parser.py
+++ b/xdsl/parser/attribute_parser.py
@@ -1261,9 +1261,7 @@ class AttrParser(BaseParser):
         Parse a location attribute, if present.
           location ::= `loc` `(` `unknown` `)`
         """
-        snapshot = self._current_token.span.start
         if not self.parse_optional_characters("loc"):
-            self._resume_from(snapshot)
             return None
 
         with self.in_parens():


tests/test_parser.py
```diff
@@ -970,7 +970,20 @@ def test_properties_retrocompatibility():
...
+
+    assert (
+        str(e.value)
+        == """\
+<unknown>:1:4
+loc(unexpected)
+    ^^^^^^^^^^
+    Unexpected location syntax.
+"""
+    )
```
math-fehr: Do we really want to test that? I feel that I would rather just check that we get `Unexpected location syntax`, or a just check `:1:4`, but not the entire error

superlopuh: Is there much of a downside? I actually added this test because of the delimited thing and `<unknown>:1:4` was printed despite the wrong error message. I can do just the error message but I have a feeling that for a small test like this it's kind of nice to have the full error checked.

xdsl/parser/attribute_parser.py
```diff
@@ -1261,15 +1262,15 @@ def parse_optional_location(self) -> LocationAttr | None:
...
-            self.parse_optional_characters("loc")
-            and self.parse_optional_punctuation("(")
-            and self.parse_optional_characters("unknown")
-            and self.parse_optional_punctuation(")")
-        ):
-            return LocationAttr()
-        self._resume_from(snapshot)
-        return None
+        if not self.parse_optional_characters("loc"):
+            self._resume_from(snapshot)
```
alexarice: I've not seen this resume used anywhere else in the code base, what's going on here?

math-fehr: Wait I missed that in my review.
I think that this should be removed? At least, if we don't parse `loc`, our location is already before `loc`, so this shouldn't be necessary.

tests/test_parser.py
math-fehr: I'll live with it, it's just that this is testing both the error messaging and the error. I don't think we test it that way everywhere else.
Maybe just matching `<unknown>:1:4`, then `Unexpected location syntax`? Not sure

superlopuh: I'll change it to just the error message.

PR #4748, opened 2025-07-10
misc: replace PipelinePass.iter_passes with parse_spec

superlopuh:
A minor change in the API that hopefully simplifies things a bit. I'd like to leverage this class a bit more, so I thought I'd simplify the point of use first.

### DIFF
--- a/xdsl/interpreters/transform.py
+++ b/xdsl/interpreters/transform.py
@@ -14,7 +14,6 @@ from xdsl.interpreter import (
     register_impls,
 )
 from xdsl.passes import ModulePass, PipelinePass
-from xdsl.utils import parse_pipeline
 
 
 @register_impls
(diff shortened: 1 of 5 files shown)


xdsl/interpreters/transform.py
```diff
@@ -45,13 +44,7 @@ def run_apply_registered_pass_op(
         args: PythonValues,
     ) -> PythonValues:
         pass_name = op.pass_name.data
-
-        schedule = tuple(
-            PipelinePass.iter_passes(
-                self.passes, parse_pipeline.parse_pipeline(pass_name)
```
alexarice, line 51: Is the idea to deprecate `parse_pipeline`?

alexarice (approved):

math-fehr (approved):

xdsl/interpreters/transform.py
superlopuh: It might be to eventually rename it/refactor it somehow, but no idea when.

PR #4749, opened 2025-07-10
dialects: (builtin) make container type non-optional and add a default

superlopuh:
Removes some need for casts.

### DIFF
--- a/xdsl/transforms/varith_transformations.py
+++ b/xdsl/transforms/varith_transformations.py
@@ -1,6 +1,6 @@
 import collections
 from dataclasses import dataclass
-from typing import Literal, cast
+from typing import Literal
 
 from xdsl.context import Context
 from xdsl.dialects import arith, builtin, varith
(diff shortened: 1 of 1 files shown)


xdsl/transforms/varith_transformations.py
```diff
@@ -173,12 +174,8 @@ def is_integer_like_type(t: Attribute) -> bool:
...
     container of integers ...
     """
-    if isinstance(t, builtin.IntegerType | builtin.IndexType):
-        return True
-    if isinstance(t, builtin.ContainerType):
-        elm_type = cast(builtin.ContainerType[Attribute], t).get_element_type()
-        return is_integer_like_type(elm_type)
-    return False
+    t = get_element_type_or_self(t)
+    return isinstance(t, builtin.IntegerType | builtin.IndexType)
```
alexarice, line 178: is this the same as `is_integer_like_type`?

superlopuh: It *is* `is_integer_like_type` :)

PR #4750, opened 2025-07-10
dialects: (builtin) add FileLineColLoc

superlopuh:
Just like MLIR.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -64,6 +64,7 @@ from xdsl.irdl import (
     irdl_op_definition,
     irdl_to_attr_constraint,
     opt_prop_def,
+    param_def,
     region_def,
     traits_def,
     var_operand_def,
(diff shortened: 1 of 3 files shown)


xdsl/dialects/builtin.py
```diff
@@ -715,7 +716,48 @@ def print_builtin(self, printer: Printer) -> None:
...
+
+    name = "file_line_loc"
+
+    filename: StringAttr = param_def()
+    line: IntAttr = param_def()
+    column: IntAttr = param_def()
+
+    @staticmethod
+    def get(filename: str, line: int, column: int):
+        return FileLineColLoc(StringAttr(filename), IntAttr(line), IntAttr(column))
```
alexarice: why `get` instead of a custom init? Maybe I missed part of the discussion on constructors for parametrized attributes?

alexarice (approved):
Nice

xdsl/dialects/builtin.py
superlopuh: I wanted to float this idea, happy to convert to init if you prefer

alexarice: I feel if we can get converters sorted then we would use an `init` right? If so then it makes sense to me to use an `init` instead of `get`. Maybe I'm missing an advantage of `get` though?

superlopuh: No you're right, I'll remove the get for now and wait for converters in the future.

PR #4751, opened 2025-07-10
API: PipelinePass -> PassPipeline

superlopuh:
So I believe that the current situation is kind of broken, in that it's a pass with no name, which feels weird. I think this makes a bit more sense, where the class is not a pass but has a similar API, and cannot be mixed up with other passes as far as the type system is concerned, until it's specified explicitly.


alexarice:
Just seen there's also a `PipelinePassSpec` that should probably also be renamed?

superlopuh:
That bit could do with a bit of an overhaul at some point, but also seems low priority for now.

alexarice:
I feel it could be nice to rename it at the same time as this, but will leave the decision to you

math-fehr (approved):
Maybe also adding documentation on the class would make this more clear?

PR #4752, opened 2025-07-10
interpreter: (pdl) Make pdl interpreter support memref.store operation

zykdtc:
By adding a `store_dict` attribute in class `RewritePattern()`, pdl interpreter could catch and store the `memref.store` operation with the corresponding SSAValue of memrefType.
Then check a special case that `memref.load` follows `memref.store`, and find the `memref.store` operation of the SSAValue used in `memref.load` from `store_dict`. Thus, we could bind load and store operation manually though store operation has no result.

### DIFF
--- a/tests/filecheck/transforms/apply-pdl/apply_pdl_store_load.mlir
+++ b/tests/filecheck/transforms/apply-pdl/apply_pdl_store_load.mlir
@@ -1,28 +1,13 @@
 // RUN: xdsl-opt %s -p apply-pdl | filecheck %s
 
 
-// CHECK:       builtin.module {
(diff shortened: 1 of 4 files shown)


zykdtc:
Yeah, sure. I do feel this way is kind of bypass for store operation and is not what pdl interpreter intend to do. While it is a good example of operations with zero result that need to be handled in eqsat. We could chat in Zulip about that or what we could do with tensors.

superlopuh:
I would also say that we might as well close this PR, I'm relatively sure that it's not possible to make this kind of pattern work without a much broader design discussion.

zykdtc:
Sure

PR #4753, opened 2025-07-10
core: add AffineMap inverse_and_broadcast_projected_permutation and drop_results

superlopuh:
Add some more helpers from MLIR's AffineMap.

### DIFF
--- a/tests/test_affine_builtins.py
+++ b/tests/test_affine_builtins.py
@@ -219,6 +219,33 @@ def test_inverse_permutation():
     )
 
 
+@pytest.mark.parametrize(
+    "input_map, expected_map",
+    [
+        (AffineMap.empty(), AffineMap.empty()),
+        (
+            AffineMap.from_callable(lambda d0, d1, d2: (d2, d1, d0)),
+            AffineMap.from_callable(lambda d0, d1, d2: (d2, d1, d0)),
+        ),
+        (
+            AffineMap.from_callable(lambda d0, d1, d2: (d1, d0)),
+            AffineMap.from_callable(lambda d0, d1: (d1, d0, 0)),
+        ),
+        (
+            AffineMap.from_callable(lambda d0, d1, d2: (d1, 0, d0)),
+            AffineMap.from_callable(lambda d0, d1, d2: (d2, d0, 0)),
+        ),
+    ],
+)
+def test_inverse_and_broadcast_projected_permutation(
+    input_map: AffineMap, expected_map: AffineMap
+):
+    inverse = input_map.inverse_and_broadcast_projected_permutation()
+    assert inverse == expected_map, f"{input_map} {expected_map}"
+    if inverse.is_projected_permutation():
+        assert inverse.inverse_and_broadcast_projected_permutation() == input_map
+
+
 def test_drop_dims():
     # (d0, d1, d2) -> (d1, d2) with [0,1,1] gives (d0, d1) -> (d0, d1)
     # (d0, d1, d2) -> (d2, d2) with [1,0,1] gives (d0, d1) -> (d1, d1)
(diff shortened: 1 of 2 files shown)


xdsl/ir/affine/affine_map.py
```diff
@@ -235,6 +235,39 @@ def inverse_permutation(self) -> AffineMap | None:
...
+        assert self.is_projected_permutation(allow_zero_in_results=True), f"{self}"
+        zero = AffineExpr.constant(0)
+        # Start with all the results as 0.
+        exprs = [zero] * self.num_dims
+        for i, res in enumerate(self.results):
+            # Skip zeros from input map. 'exprs' is already initialized to zero.
+            if isinstance(const_expr := res, AffineConstantExpr):
+                assert const_expr.value == 0, (
+                    "Unexpected constant in projected permutation"
+                )
```
alexarice: This shouldn't be necessary right? You've already done the assertion above?

xdsl/ir/affine/affine_map.py
```diff
@@ -235,6 +235,39 @@ def inverse_permutation(self) -> AffineMap | None:
             results=results,
         )
 
+    def inverse_and_broadcast_projected_permutation(self):
+        """
+        If `self` is a projected permutation, with possible constant 0 expression
+        results, returns the inverse permutation.
```
alexarice, line 242: Is calling this map an inverse standard terminology? It's not an inverse in the mathematical sense of an affine map

xdsl/ir/affine/affine_map.py
```diff
@@ -273,6 +306,36 @@ def drop_dims(self, unused_dims: Sequence[bool]) -> AffineMap:
             new_dims, new_symbols, result_num_dims, self.num_symbols
         )
 
+    def drop_results(self, unused_results: Sequence[bool]) -> AffineMap:
+        """
+        Given a sequence of `unused_results` indicating the resuts to drop,
```
```suggestion
        Given a sequence of `unused_results` indicating the results to drop,
```

xdsl/ir/affine/affine_map.py
```diff
@@ -273,6 +306,36 @@ def drop_dims(self, unused_dims: Sequence[bool]) -> AffineMap:
...
+        return a new map only with the new results.
+
+        Examples:
+        ```
+        (d0, d1, d2) -> (d1, d2) with [1,0] gives (d0, d1, d2) -> (d1)
+        (d0, d1, d2) -> (d2, d2) with [0,1] gives (d0, d1, d2) -> (d2)
```
alexarice: I don't understand these docs, why aren't both results dropped in both cases?

alexarice: oh, it's a sequence of bools not a sequence of indices

tests/test_affine_builtins.py
```diff
@@ -232,6 +259,17 @@ def test_drop_dims():
...
 
+def test_drop_results():
+    t = True
+    f = False
+    assert AffineMap.from_callable(lambda d0, d1, d2: (d1, d2)).drop_results(
+        [t, f]
+    ) == AffineMap.from_callable(lambda d0, d1, d2: (d2,))
+    assert AffineMap.from_callable(lambda d0, d1, d2: (d1, d2)).drop_results(
+        [f, t]
+    ) == AffineMap.from_callable(lambda d0, d1, d2: (d1,))
```
alexarice: Could we have an example with symbols (if that's valid)

xdsl/ir/affine/affine_map.py
alexarice: The test has `(d1,d2)` instead of `(d2, d2)`

alexarice (approved):

xdsl/ir/affine/affine_map.py
superlopuh: I guess so, in this case the map represents a permutation, which has in inverse permutation, which is what this is about.

xdsl/ir/affine/affine_map.py
superlopuh: Good catch

xdsl/ir/affine/affine_map.py
alexarice: True

PR #4754, opened 2025-07-11
dialects: (linalg) use `Operation` as abstract operation suffix

superlopuh:
Minor naming cleanup PR before more changes in the dialect.

### DIFF
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -1098,7 +1098,7 @@ class QuantizedMatmulOp(NamedOperation):
         )
 
 
-class PoolingOpsBase(NamedOperation, ABC):
+class PoolingOperation(NamedOperation, ABC):
     """Base class for linalg pooling operations."""
 
     PRINT_ATTRS_IN_FRONT: ClassVar[bool] = True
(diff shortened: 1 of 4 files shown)


xdsl/dialects/linalg.py
```diff
@@ -1098,7 +1098,7 @@ def hidden_region(args: tuple[BlockArgument, ...]) -> None:
         )
 
 
-class PoolingOpsBase(NamedOpBase, ABC):
+class PoolingOpsBase(NamedOperation, ABC):
```
alexarice:
```suggestion
class PoolingOperation(NamedOperation, ABC):
```
or change the PR title

xdsl/dialects/linalg.py
```diff
@@ -1144,7 +1144,7 @@ def hidden_region(args: tuple[BlockArgument, ...]) -> None:
         )
 
 
-class ConvOpsBase(NamedOpBase, ABC):
+class ConvOpsBase(NamedOperation, ABC):
```
alexarice: Same here?

xdsl/dialects/linalg.py
superlopuh: I thought I had found them all, my bad

superlopuh: Oh it's because one had OpBase, the others OpsBase, this is one reason they all need to be Operation.

PR #4755, opened 2025-07-11
core: (ParametrizedAttribute) add converters

alexarice:
I had a go. I'm not sure if this is the right way to do it but let's iterate.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1253,12 +1253,7 @@ class DictionaryAttr(_BuiltinData[immutabledict[str, Attribute]]):
 class TupleType(ParametrizedAttribute, BuiltinAttribute, TypeAttribute):
     name = "tuple"
 
-    types: ArrayAttr[TypeAttribute]
-
-    def __init__(self, types: list[TypeAttribute] | ArrayAttr[TypeAttribute]) -> None:
-        if isinstance(types, list):
-            types = ArrayAttr(types)
-        super().__init__(types)
+    types: ArrayAttr[TypeAttribute] = param_def(converter=ArrayAttr[TypeAttribute].get)
 
     def print_builtin(self, printer: Printer):
         printer.print_string("tuple")
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -429,10 +429,11 @@ class BitEnumAttribute(Generic[EnumType], Data[tuple[EnumType, ...]]):
 class ParametrizedAttribute(Attribute):
     """An attribute parametrized by other attributes."""
 
-    def __init__(self, *parameters: Attribute):
-        for (f, _), param in zip(
-            self.get_irdl_definition().parameters, parameters, strict=True
-        ):
+    def __init__(self, *parameters: Any):
+        irdl_def = self.get_irdl_definition()
+        for (f, d), param in zip(irdl_def.parameters, parameters, strict=True):
+            if d.converter is not None:
+                param = d.converter(param)
             object.__setattr__(self, f, param)
         super().__init__()
 
--- a/xdsl/irdl/attributes.py
+++ b/xdsl/irdl/attributes.py
@@ -98,21 +98,25 @@ class _ParameterDef:
     """
 
     param: AttrConstraint | None
+    converter: Callable[[Any], Attribute] | None
 
     def __init__(
         self,
         param: AttrConstraint | None,
+        converter: Callable[[Any], Attribute] | None,
     ):
         self.param = param
+        self.converter = converter
 
 
 def param_def(
     constraint: AttrConstraint[AttributeInvT] | None = None,
     *,
+    converter: Callable[[Any], AttributeInvT] | None = None,
     init: Literal[True] = True,
 ) -> AttributeInvT:
     """Defines a property of an operation."""
-    return cast(AttributeInvT, _ParameterDef(constraint))
+    return cast(AttributeInvT, _ParameterDef(constraint, converter))
 
 
 def check_attr_name(cls: type):
@@ -161,6 +165,7 @@ class ParamDef(NamedTuple):
     """
 
     constr: AttrConstraint
+    converter: Callable[[Any], Attribute] | None = None
 
 
 @dataclass
@@ -230,6 +235,8 @@ class ParamAttrDef:
                     f"Invalid field type {field_type} for field name {field_name}."
                 ) from e
 
+            converter: Callable[[Any], Attribute] | None = None
+
             if field_name in field_values:
                 value = field_values.pop(field_name)
                 if isinstance(value, _ParameterDef):
@@ -242,6 +249,8 @@ class ParamAttrDef:
                         raise PyRDLAttrDefinitionError(
                             f"Invalid constraint {value.param} for field name {field_name}."
                         ) from e
+                    if value.converter is not None:
+                        converter = value.converter
 
                 # Constraint variables are deprecated
                 elif get_origin(value) is Annotated or any(
@@ -259,7 +268,7 @@ class ParamAttrDef:
                         f"{field_name} is not a parameter definition."
                     )
 
-            parameters[field_name] = ParamDef(constraint)
+            parameters[field_name] = ParamDef(constraint, converter)
 
         for field_name, value in field_values.items():
             # Anything left is a field without an annotation or a constaint var.
--- a/tests/filecheck/utils/codegen/simple.py
+++ b/tests/filecheck/utils/codegen/simple.py
@@ -37,9 +37,18 @@ from xdsl.traits import (
 from xdsl.utils.dialect_codegen import dump_dialect_pyfile, generate_dynamic_attr_class
 
 types = [
-    ("Test_SingletonAType", ParamAttrDef(name="test.singleton_a", parameters=[])),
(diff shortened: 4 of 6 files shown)


xdsl/dialects/builtin.py
```diff
@@ -185,6 +186,14 @@ def constr(
     ) -> ArrayOfConstraint[AttributeInvT]:
         return ArrayOfConstraint(constr)
 
+    @staticmethod
+    def get(
```
alexarice: I thought `get` could be a good name for these sort of things as it mirrors the `SSAValue.get` interface.

xdsl/dialects/builtin.py
```diff
@@ -1197,12 +1206,7 @@ def verify(self) -> None:
 class TupleType(ParametrizedAttribute, BuiltinAttribute, TypeAttribute):
     name = "tuple"
 
-    types: ArrayAttr[TypeAttribute]
-
-    def __init__(self, types: list[TypeAttribute] | ArrayAttr[TypeAttribute]) -> None:
-        if isinstance(types, list):
-            types = ArrayAttr(types)
-        super().__init__(types)
+    types: ArrayAttr[TypeAttribute] = param_def(converter=ArrayAttr.get)
```
alexarice: Example

xdsl/ir/core.py
```diff
@@ -426,22 +426,13 @@ def print_parameter(self, printer: Printer):
...
-                stacklevel=2,
-            )
-            parameters = parameters[0]
-
-        for (f, _), param in zip(
-            self.get_irdl_definition().parameters, parameters, strict=True
-        ):
-            object.__setattr__(self, f, param)
+    def __init__(self, *parameters: Any):
+        irdl_def = self.get_irdl_definition()
```
alexarice, line 433: Removed the deprecated thing as it was throwing pyright errors, but could probably figure it out to keep it around a bit longer

xdsl/irdl/attributes.py
```diff
@@ -160,6 +164,7 @@ class ParamAttrDef:
 
     name: str
     parameters: list[tuple[str, AttrConstraint]]
+    converters: dict[str, Callable[[Any], Attribute]]
```
alexarice: I'm unsure if this is the correct place to store these. It's quite convenient though.

xdsl/dialects/builtin.py
superlopuh: Could this live on Data?

xdsl/ir/core.py
superlopuh: What were the pyright errors? I would ignore them probably

xdsl/dialects/builtin.py
alexarice: Maybe, I didn't think too hard about it

xdsl/ir/core.py
alexarice: I was very tired when I did this and didn't really read them

xdsl/ir/core.py
```diff
@@ -426,22 +426,13 @@ def print_parameter(self, printer: Printer):
...
+        for (f, _), param in zip(irdl_def.parameters, parameters, strict=True):
+            if (converter := irdl_def.converters.get(f)) is not None:
+                object.__setattr__(self, f, converter(param))
+            else:
+                object.__setattr__(self, f, param)
```
superlopuh: ?
```suggestion
            if (converter := irdl_def.converters.get(f)) is not None:
                param = converter(param)
            object.__setattr__(self, f, param)
```

xdsl/irdl/attributes.py
superlopuh: I think this is the best place to put it without doing some extensive perf profiling of alternatives once this is adopted.

alexarice: It was more whether it's morally the right place to put this

xdsl/dialects/builtin.py
alexarice: I might try to split this off

xdsl/irdl/attributes.py
math-fehr: Why not having a list here? Is there a reason we would need a `dict`?
Also, I think it would be nicer if the `parameters` would be something like a `ParamDef` class or something like this, rather than a `AttrConstraint`, so we can store the converter there directly.

math-fehr:
This looks quite nice, especially if pyright will be able to look at it!

alexarice:
> This looks quite nice, especially if pyright will be able to look at it!

I'm not sure there's a way to internalise it to a test, but I can at least confirm locally that pyright `reveal_type`s the correct initialiser type

xdsl/irdl/attributes.py
alexarice: I guess a `dict` just seemed natural to me? Is there a good reason to have a list? The issue with storing a `ParamDef` in the `parameters` is that this function does processing to the constraint in the `ParamDef` (i.e. converting from type to constraint) so this would involve unbundling the original `ParamDef` doing the processing then rebundling it. What's in the PR was at least the cleanest way I could see to hold all the data but I'd be open to other suggestions (I tried having parameters be a "bigger" tuple with an optional converter and that seemed more messy, but I could try again and follow through with it if wanted)

alexarice:
I'll try to rebase this past the `get` change  and clean it up next week. Seems like this approach is generally favoured

xdsl/irdl/attributes.py
math-fehr: I mean, the reason I would do that is because `parameters` are also lists, so I don't think there is any reasons to have both a list and a dict there?
Also, what I meant is just having a `list[tuple[str, tuple[AttrConstraint, Callable]]` for parameters, and creating a class for that `tuple`. But that's fine by me if you want two fields (like you have ni the PR). But I would just make that `dict` a `list`.

superlopuh: This feels worth spelling out, would you agree with these changes?

 1. make the value in the key/value pair list a ParamDef, so one level of indirection
 2. add an optional converter field to ParamDef

I actually considered proposing this and then thought that we are likely to have way fewer converters than other things, so having the ParamAttrDef have one dense structure and one sparse structure was probably the right way to go vs a dense structure of sparse structures. OTOH this is a bit more like dataclasses's Field, and it would be good for future extra features to have an obvious home...

alexarice: There's no such thing as a `ParamDef`. If you mean `_ParameterDef` then I disagree with the proposal, I think it's nice that the constraint field is actually a constraint by the time we get to this point.

For Mathieu, It would likely need to at least be a `list[tuple[str, tuple[AttrConstraint, Callable | None]]`. I really don't see what's wrong with the dictionary though tbh, we still have an ordering defined by the parameters field and this `converters` field feels like its just tagging on some extra info to me. I can change it to a list if really wanted, but just don't see the point?

alexarice:
I've added a test and marked this as ready for review. I didn't change the converters being a dictionary but am happy to if wanted.

xdsl/irdl/attributes.py
superlopuh: That's not what I mean, I mean creating a new `ParamDef` that doesn't exist that is essentially this:
```py
class ParamDef(NamedTuple):
  constr: AttrConstraint
  converter: Converter | None
```
So it's a resolved _ParameterDef basically. Then if we add new functionality like default attributes they can also live in this object.

xdsl/ir/core.py
alexarice: Adding context from zulip. Since its now valid for a parameter to be a tuple (which gets converted) there's no way to keep this warning without breaking stuff. I believe the plan is to wait a release, then remove the warning, then add converters.

alexarice:
Converting to draft again until we've removed deprecated things.

alexarice:
Rebased onto @superlopuh 's `ParamDef` idea. Seems a lot cleaner.

superlopuh (approved):
I think this code looks good for whenever we decide to merge it.

superlopuh (approved):
Still good

xdsl/dialects/builtin.py
```diff
@@ -1253,12 +1253,7 @@ def verify(self) -> None:
 class TupleType(ParametrizedAttribute, BuiltinAttribute, TypeAttribute):
     name = "tuple"
 
-    types: ArrayAttr[TypeAttribute]
-
-    def __init__(self, types: list[TypeAttribute] | ArrayAttr[TypeAttribute]) -> None:
-        if isinstance(types, list):
-            types = ArrayAttr(types)
-        super().__init__(types)
+    types: ArrayAttr[TypeAttribute] = param_def(converter=ArrayAttr[TypeAttribute].get)
```
math-fehr, line 1256: Instead of `.get`, should we actually use something like `.__init__`?

math-fehr (approved):
Vey cool!

xdsl/dialects/builtin.py
alexarice: The distinction is that `.get` allows an `ArrayAttr` to be passed in if you already have one.

alexarice: It's maybe more obvious with `StringAttr` where:
```python
StringAttr.__init__ : str -> StringAttr
```
vs
```python
StringAttr.get : str | StringAttr -> StringAttr
```

PR #4759, opened 2025-07-11
documentation: (Toy) use FileLineColLoc in Toy

superlopuh:
Another two year old TODO.


docs/Toy/toy/frontend/parser.py
```diff
@@ -23,6 +23,12 @@
 )
 
 
+def loc(token: ToyToken):
+    file = token.span.input.name
+    line, col = token.span.get_line_col()
+    return Location(file, line, col + 1)
```
superlopuh: I'm not sure why there's a disagreement between toy locations and our span locations...

superlopuh: I have a feeling that it's our span locations that are wrong, and that columns universally start at 1.

docs/Toy/toy/frontend/ir_gen.py
```diff
@@ -96,9 +104,7 @@ def ir_gen_module(self, module_ast: ModuleAST) -> ModuleOp:
 
     def loc(self, loc: Location):
```
alexarice, line 105: Could this not be a function on `Location`?

docs/Toy/toy/frontend/parser.py
alexarice: Columns should start at 0 right?

alexarice (commented):
Would be nice to figure out the off by 1 error before merging

docs/Toy/toy/frontend/ir_gen.py
superlopuh: It could, I was going for more of a diff minimisation in this PR.

superlopuh:
I'd rather just merge tbh, I feel like it's not so much an error as a change in the frame of reference. If we ever change the start column in xDSL proper we can just remove the +1 here.

superlopuh:
Let's first clean up location helper in the Lexer open PRs then merge this.

docs/Toy/toy/frontend/parser.py
superlopuh: #4769

docs/Toy/toy/frontend/ir_gen.py
superlopuh: I propose to keep this part here for now, and add helpers to go from Location to FileLineColLoc and back in another PR.

math-fehr (approved):

PR #4760, opened 2025-07-11
core: (Constraints) make int constraint IntAttr constraints

alexarice:
There's some nasty import loops, and the error messages aren't quite as good, but this should be worth it to simplify the infrastructure (hopefully).

I guess the alternative is to try what @superlopuh said and have AttrConstraint be generic on its input as well as output. I don't think this would capture range constraints but might be able to capture both int and attr constraints nicely, though perhaps this PR is good enough.

We could also consider giving `IntAttr` a nicer printing/parsing (say `int<n>`) as it is a builtin attribute.

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -21,6 +21,7 @@ from xdsl.ir import (
     ParametrizedAttribute,
     TypedAttribute,
 )
+from xdsl.irdl.core_attrs import IntAttr
 from xdsl.utils.exceptions import PyRDLError, VerifyException
 from xdsl.utils.runtime_final import is_runtime_final
 
@@ -42,26 +43,17 @@ class ConstraintContext:
     )
     """The assignment of constraint range variables."""
 
-    _int_variables: dict[str, int] = field(default_factory=dict[str, int])
-    """The assignment of constraint int variables."""
-
     def get_variable(self, key: str) -> Attribute | None:
         return self._variables.get(key)
 
     def get_range_variable(self, key: str) -> tuple[Attribute, ...] | None:
         return self._range_variables.get(key)
 
-    def get_int_variable(self, key: str) -> int | None:
-        return self._int_variables.get(key)
-
     def set_attr_variable(self, key: str, attr: Attribute):
         self._variables[key] = attr
 
     def set_range_variable(self, key: str, attrs: tuple[Attribute, ...]):
         self._range_variables[key] = attrs
-
-    def set_int_variable(self, key: str, i: int):
-        self._int_variables[key] = i
 
     @property
     def attr_variables(self) -> AbstractSet[str]:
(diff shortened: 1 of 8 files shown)


xdsl/irdl/constraints.py
```diff
@@ -1007,7 +879,9 @@ class RangeOf(GenericRangeConstraint[AttributeCovT]):
 
     constr: GenericAttrConstraint[AttributeCovT]
     _: KW_ONLY
-    length: IntConstraint = field(default_factory=AnyInt)
+    # Ideally we would like an AnyIntConstr here but this causes import loops
+    # `AnyAttr` won't infer anything so this *should* be safe.
```
superlopuh: If we manage to get the constraint variance to work out then this stops being a problem

alexarice: I saw the other PR but I'm not sure I understand how constraint variance solves this problem.

superlopuh: It solves it in that AnyConstr will then accept anything at all so you don't need an AnyInt anymore

alexarice: The issue here is that `AnyAttr` accepts too much rather than not enough

superlopuh: It seems safe to me to accept anything here, I thought the comment was about the pyright ignore

alexarice: I think it is safe. It would be unsafe below (on line 924) if `self.length.infer(context)` could return something that wasn't an `IntAttr`, but I believe this can't happen

math-fehr: Can we not define `AnyIntConstr` above this class? Why would this create a cycle of dependencies

math-fehr (approved):
Looks quite good, and definitely removes quite a lot of code!

xdsl/irdl/constraints.py
alexarice: `AnyIntConstr` depends on `IntAttr`.
`IntAttr` depends on `irdl_attr_definition` (or whatever the decorator is called)
that depends on `RangeOf`
`RangeOf` would depend on `AnyIntConstr`

alexarice: This can potentially be solved by `IntAttr` not using the decorator (as far as I'm aware the decorator doesn't do much for `Data`).

superlopuh: All it does is annotate with `runtime_final`. I have a feeling that it would be OK to have both ArrayAttr and IntAttr be a part of the ir. I think changes in this bit of code are worth doing slowly BTW, I wouldn't merge this too quickly if there are still things to explore.

math-fehr: Yeah, I think it would make sense to not use the decorator here, or to have `IntAttr` being a bit more "builtin" than the other attributes, just for that.

superlopuh:
Is this approach compatible with the `Literal[int]` mechanism we added since it was opened? Would it make sense to merge main into it and update it?

superlopuh:
Let's reopen this when you have time to iterate, I still think there's lots of things worth doing in constraints but I don't have the bandwidth to push it. Here I think the main thing that would be great is some constraints infrastructure that isn't to do with attributes, kind of similar to the int stuff we added to check for bitwidth on the type.

PR #4762, opened 2025-07-11
dialects: (emitc) Add PointerType

Jimmy2027:
Add the [emitc Pointertype](https://mlir.llvm.org/docs/Dialects/EmitC/#pointertype)


Jimmy2027:
not quite, a few remaining but should be pretty straight forward to add them :)

PR #4763, opened 2025-07-11
API: add GenericConstraint as supertype of other constraints

superlopuh:
As a kind of companion to #4760, I thought I'd try making a supertype of the constraints. I wonder how much logic we can share between the constraints without needing to unspecialcase integers? Might be a useful avenue to go down.

### DIFF
--- a/docs/marimo/irdl.py
+++ b/docs/marimo/irdl.py
@@ -346,13 +346,13 @@ def _(ConstraintContext, IntAttr, VerifyException, irdl_to_attr_constraint):
             object.__setattr__(self, "elem_constr", irdl_to_attr_constraint(constr))
 
         # Check that an attribute satisfies the constraints
-        def verify(self, attr: Attribute, constraint_context: ConstraintContext) -> None:
+        def verify(self, value: Attribute, constraint_context: ConstraintContext) -> None:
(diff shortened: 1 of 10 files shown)


math-fehr:
I'm not completely sure why we would go that route compared to the one that Alex just proposed?
Do you see any place where this would be better? The thing that is quite nice with Alex proposal is that we stay in the attribute world, and I think it's quite nice overall.

alexarice:
I guess a question is do you forsee/want there to be any other options for input to a constraint? My gut feeling is no but maybe I'm missing something. Do you feel that `GenericConstraint` is worth it just for sharing `verify` and `verifies` between range and attr constraints?

superlopuh:
To be honest, I did this to see if it would work, and it seems to. I moved the minimal amount of code to be shared but I'm sure that we could share more if we explore this route. I was hoping we could think about this and chat about it later. One potential thing that this unlocks is constraints that assume a tighter bound than Attribute on whatever it is they verify. I'm not sure of the other benefits.

superlopuh:
As discussed on Zulip, first we want to make sure that this is actually helpful for the other refactors.

superlopuh:
I still think this might be a good idea but we can do this later

PR #4766, opened 2025-07-12
core: (affine) add apply_permutation helper to AffineMap

superlopuh:
Reopen #4210, I realised that I actually would be useful to have a shortcut to apply permutations to non-int inputs, such as permuting tuples of attributes.

### DIFF
--- a/xdsl/ir/affine/affine_map.py
+++ b/xdsl/ir/affine/affine_map.py
@@ -6,6 +6,8 @@ from dataclasses import dataclass
 from inspect import getfullargspec
 from typing import cast
 
+from typing_extensions import TypeVar
+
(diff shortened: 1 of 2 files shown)


xdsl/ir/affine/affine_map.py
```diff
@@ -433,6 +437,26 @@ def is_projected_permutation(self, allow_zero_in_results: bool = False) -> bool:
         # Results are either dims or zeros and zeros can be mapped to input dims.
         return True
 
+    def apply_permutation(self, source: Sequence[_T]) -> tuple[_T, ...]:
+        """
+        Assert that `self` represents a permutation, and apply the permutation to
```
```suggestion
        Assert that `self` represents a projected permutation, and apply the permutation to
```
Maybe this is MLIR (mis)terminology, but the example is not a permutation

alexarice (approved):

PR #4767, opened 2025-07-12
core: (affine) add AffineMap.unused_dims and .unused_dims_bit_vector

superlopuh:
Another useful helper from MLIR's AffineMap.

### DIFF
--- a/xdsl/ir/affine/affine_map.py
+++ b/xdsl/ir/affine/affine_map.py
@@ -347,9 +347,22 @@ class AffineMap:
             if isinstance(expr, AffineDimExpr)
         }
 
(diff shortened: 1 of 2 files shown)


xdsl/ir/affine/affine_map.py
```diff
@@ -347,23 +347,45 @@ def used_dims(self) -> set[int]:
             if isinstance(expr, AffineDimExpr)
         }
 
+    def unused_dims(self) -> set[int]:
+        """
+        Return all dimensions used in the map as a set
```
```suggestion
        Return all dimensions unused in the map as a set
```
surely?

alexarice (approved):

PR #4768, opened 2025-07-12
API: unify PatternRewriter.insert and .insert_op

superlopuh:
Spotted this bit of weirdness where we actually have two operations on PatternRewriter, `insert` and `insert_op`. It feels like we should probably have just one, but they're different enough that it would be a bit of a painful transition. This PR makes it so that they at least both set `has_done_action`, maybe we should deprecate one of them in the future.

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -7,6 +7,8 @@ from dataclasses import dataclass, field
 from types import TracebackType
 from typing import ClassVar, TypeAlias, overload
 
+from typing_extensions import TypeVar
+
 from xdsl.dialects.builtin import ArrayAttr
 from xdsl.ir import Attribute, Block, BlockArgument, Operation, OperationInvT, Region
 from xdsl.rewriter import BlockInsertPoint, InsertPoint, Rewriter
@@ -42,6 +44,9 @@ class BuilderListener:
         self.block_creation_handler.extend(listener.block_creation_handler)
 
 
+InsertOpInvT = TypeVar("InsertOpInvT", bound=Operation | Sequence[Operation])
+
+
 @dataclass
 class Builder(BuilderListener):
     """
@@ -60,29 +65,15 @@ class Builder(BuilderListener):
         """
         return self.insert_op(op)
 
-    @overload
     def insert_op(
         self,
-        op: OperationInvT,
+        op: InsertOpInvT,
         insertion_point: InsertPoint | None = None,
-    ) -> OperationInvT: ...
-
-    @overload
-    def insert_op(
-        self,
-        op: Sequence[Operation],
-        insertion_point: InsertPoint | None = None,
-    ) -> None: ...
-
-    def insert_op(
-        self,
-        op: Operation | Sequence[Operation],
-        insertion_point: InsertPoint | None = None,
-    ) -> Operation | None:
+    ) -> InsertOpInvT:
         """Inserts `op` at the current insertion point."""
-        res, ops = (op, (op,)) if isinstance(op, Operation) else (None, op)
+        res, ops = (op, (op,)) if isinstance(op, Operation) else (op, op)
         if not ops:
-            return
+            return ops
 
         implicit_builder = ImplicitBuilder.get()
         if implicit_builder is not None and implicit_builder is not self:
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -58,17 +58,30 @@ class Builder(BuilderListener):
         """
         Inserts op at the current location and returns it.
         """
-        self.insert_op(op)
-        return op
+        return self.insert_op(op)
+
+    @overload
+    def insert_op(
+        self,
+        op: OperationInvT,
+        insertion_point: InsertPoint | None = None,
+    ) -> OperationInvT: ...
+
+    @overload
+    def insert_op(
+        self,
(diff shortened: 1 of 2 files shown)


xdsl/builder.py
```diff
@@ -55,22 +55,34 @@ class Builder(BuilderListener):
     """Operations will be inserted at this location."""
 
     def insert(self, op: OperationInvT) -> OperationInvT:
```
superlopuh, line 62: Should we deprecated this? There are 76 uses in xDSL, and most don't use the result, or can trivially be refactored.

math-fehr: I think we should just merge it with `insert_op`, which should be the same method anyway!

xdsl/builder.py
```diff
@@ -55,22 +55,34 @@ class Builder(BuilderListener):
...
+        Inserts op at the current location and returns it.
+        """
+        self.insert_op(op)
+        return op
+
+    def insert_op(
+        self,
+        op: Operation | Sequence[Operation],
+        insertion_point: InsertPoint | None = None,
+    ):
```
math-fehr: Actually, I'm relying *a lot* on the return value of `insert`.
Can we make sure to return the inserted `op` (and overload whenever this is a sequence or a single op?).
The use case I have is writing things like:
```
my_op = b.insert(AddOp(lhs, rhs)).res
```
Where you can one-line something quite nicely.

math-fehr (commented):
That's my bad, I was in the middle of a refactor I never finished I think.
Can we make it so that `insert` returns the inserted operations, which I'm relying on quite a lot.
Also, I would prefer to keep the name `insert` rather than `insert_op`, I'm using this operation so much I'd like to keep it quite short.

superlopuh:
@math-fehr is this what you were thinking?

alexarice:
I wasn't aware that `insert` existed, and now that I am I'm quite fond of its api. I feel there's been a few times where I've written:
```python
my_op = ...
rewriter.insert_op(my_op)
...
```
If `insert` is the preferred name, is there any objection to moving all instances of `insert_op` to `insert`? I imagine the overhead of returning something which is unused must be nonexistent?

alexarice:
Maybe disregard what I said, I realise this doesn't work as `insert_op` can take a sequence of `ops`

superlopuh:
The sequence thing messes with it a little bit... I would probably merge this first and then deprecate in a follow-up PR if that's OK with you, as this seems like a bug fix to me.

xdsl/builder.py
```diff
@@ -55,24 +55,49 @@ class Builder(BuilderListener):
...
+        op: OperationInvT,
+        insertion_point: InsertPoint | None = None,
+    ) -> OperationInvT: ...
+
+    @overload
+    def insert_op(
+        self,
+        op: Sequence[Operation],
+        insertion_point: InsertPoint | None = None,
+    ) -> None: ...
```
math-fehr: Why not returning the Sequence of operations here as well?
That's what it does anyway?

xdsl/builder.py
```diff
@@ -55,24 +55,49 @@ class Builder(BuilderListener):
...
+
+    def insert_op(
+        self,
+        op: Operation | Sequence[Operation],
+        insertion_point: InsertPoint | None = None,
+    ) -> Operation | None:
         """Inserts `op` at the current insertion point."""
+        res, ops = (op, (op,)) if isinstance(op, Operation) else (None, op)
+        if not ops:
+            return
```
math-fehr: If you change this to `return ops`, then our return type for sequences can be a sequence.

math-fehr: Or alternatively if you want to deprecate sequences later, I'm fine with that.

xdsl/builder.py
```diff
@@ -55,24 +55,49 @@ class Builder(BuilderListener):
     """Operations will be inserted at this location."""
 
     def insert(self, op: OperationInvT) -> OperationInvT:
+        """
+        Inserts op at the current location and returns it.
+        """
+        return self.insert_op(op)
+
+    @overload
```
math-fehr: So maybe my question is, why not adding this argument to `insert`, instead of creating `insert_op`?

math-fehr (commented):
The overload and return things are what I meant, thanks!
Otherwise, I think we maybe should just add the argument to `insert` instead of creating `insert_op`?

Also, can you add tests for it?

superlopuh:
@math-fehr I just wanted to minimise the impact of this PR so that the bug fix changes could be merged quickly without changing any of the function signatures and so that I could take the time to migrate the other things in a non-blocking way for my other work.

superlopuh:
Returning the input type seems to work well, I'd still rather do the migration in a separate PR if that's OK

xdsl/builder.py
```diff
@@ -55,22 +60,33 @@ class Builder(BuilderListener):
...
+        Inserts op at the current location and returns it.
+        """
+        return self.insert_op(op)
+
+    def insert_op(
+        self,
+        op: InsertOpInvT,
+        insertion_point: InsertPoint | None = None,
+    ) -> InsertOpInvT:
         """Inserts `op` at the current insertion point."""
```
alexarice: 1 comment, maybe this should say it returns the op(s)

PR #4769, opened 2025-07-13
misc: columns in debug messages start at 1

superlopuh:
This seems to be the convention in all IDEs and error messages, and can be verified by copy/pasting one of the changed messages in the filecheck tests to the terminal in VSCode, and command-clicking it, the new 1-indexed columns point to the right place.


alexarice:
Interestingly this isn't the default in emacs (but can be customized)

superlopuh:
Actually I want to go with a slightly different design.

superlopuh:
Please review #4770 first, then I'll make this PR address the new column numbers.

alexarice (approved):

PR #4772, opened 2025-07-13
misc: (lexer) replace get_lines_containing with get_line_start_and_index

superlopuh:
This feels like the more intuitive helper, and avoids creating a temporary throwaway lines instead when calculating line and column.


xdsl/utils/lexer.py
```diff
@@ -107,9 +108,19 @@ def len(self):
     def text(self):
         return self.input.content[self.start : self.end]
 
+    def get_line_start_and_index(self) -> LineStartAndIndex:
+        """
+        The start offset and 1-index of position in a file.
+        The line index is offset by `line_offset`, which is non-0 when the input file is
+        split.
```
alexarice:
```suggestion
        The start offset and 1-index of the start position of this span in a file.
```

xdsl/utils/lexer.py
```diff
@@ -107,9 +108,19 @@ def len(self):
...
+        """
+        line_start = self.input.get_start_of_line(self.start)
+        line_index_in_source = self.input.content.count("\n", 0, line_start) + 1
+        line_index = line_index_in_source + self.line_offset
+        return LineStartAndIndex(line_start, line_index)
+
     def get_line_col(self) -> tuple[int, int]:
-        info = self.input.get_lines_containing(self)
-        _lines, offset_of_first_line, line_no = info
+        offset_of_first_line, line_no = self.get_line_start_and_index()
```
alexarice, line 121: Can named tuples be destructured like this?

alexarice (approved):
This looks reasonable to me.

Something feel kind of yucky about all of this. It feels like there's a couple of methods doing pretty similar things and there's newline counting everywhere. I partially wonder if `Input` should be a list of lines but that feel like too drastic a change.

My only other idea would be to have a single function which converts a `Position` to a `LinePosition`, which would be a struct with (1-indexed) line a column numbers, and for all the other apis to go through this, so that the yuckyness is in one place.

xdsl/utils/lexer.py
superlopuh: Yep, this is an argument against using them for some people, but I find it convenient.

superlopuh:
Actually you're right this is yucky in a way that I think is fixable I'll mark this as draft for now.

superlopuh:
Sorry about the churn, I realised after your review that actually everything can be achieved relatively cleanly if you have the location ready, so we might as well use that. Closing in favour of #4759

PR #4775, opened 2025-07-14
dialects: (builtin) make `IntAttr` 'core'

alexarice:
Based on discussion in #4760, makes `IntAttr` a more core part of xDSL. Concretely:
- Moves `IntAttr` to a new file (I chose `xdsl.dialects.core` but will happily take other suggestions)
- Uses a `runtime_final` annotation instead of `irdl_attr_definition`.

This should allow #4760 to be simplified (I'll rebase now and see).


alexarice:
Moved this to `xdsl.irdl.core_attrs` (still maybe the wrong place) as the constraints stuff still got into a bit of an import loop with the irdl dialect finding in `dialects/__init__.py`

alexarice:
Thinking about this more, maybe `xdsl.dialects.core` was the correct place for this to live, and this was just exposing a bug with the way the irdl dialect import was set up in the `__init__.py`. Curious to know what everyone thinks

alexarice:
We decided we would explore other options first before committing to this, I'll leave the PR open for if we decide we'd like to explore this method again.

superlopuh:
I still think we might want to do this later, but let's close it for now

PR #4777, opened 2025-07-14
core: (ParametrizedAttribute) allow ClassVar

alexarice:
The param_def api allows us to add constraint variables, but we never actually allowed `ClassVar` to appear in the class, which we use heavily in operation definitions. There's three options:
- Allow any field with no type or a ClassVar type. As far as I'm aware this keeps closer to pythons dataclass model, as any annotationless field should be treated as a ClassVar. This is what I've implemented here.
- Only allow fields that are explicitly marked as `ClassVar`. This has the advantage that we catch "dodgy" param_def fields with no annotation
- Allow all fields but error if there's an annotationless `param_def`.

Any thoughts?

### DIFF
--- a/tests/irdl/test_attribute_definition.py
+++ b/tests/irdl/test_attribute_definition.py
@@ -836,18 +836,6 @@ def test_invalid_typed_field():
         irdl_attr_definition(InvalidTypedFieldTestAttr)
 
 
-class InvalidUntypedFieldTestAttr(ParametrizedAttribute):
-    name = "test.invalid_untyped_field"
-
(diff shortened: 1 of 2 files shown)


superlopuh:
I have a local branch with this done a bit differently, are you happy with me pushing that here?

alexarice:
So I realised I forgot to actually allow `ClassVar` annotations here. Would it possible to push yours separately so that its easier to compare?

superlopuh:
Yeah

alexarice:
The over one is better

PR #4778, opened 2025-07-14
core: (ParametrizedAttribute) Allow bare ClassVar in definition

alexarice:
Switches from using `get_type_hints` to `get_annotations` in order to allow the use of bare `ClassVar` in python 3.10.
Tests this by switching from `ConstraintVar` to `VarConstraint` in tests.


tests/irdl/test_attribute_definition.py
```diff
@@ -916,10 +915,12 @@ def test_generic_attr():
 class ConstraintVarAttr(ParametrizedAttribute):
     name = "test.constraint_var"
 
-    T = Annotated[IntegerType, ConstraintVar("T")]
+    T: ClassVar[GenericAttrConstraint[IntegerType]] = VarConstraint(
```
alexarice: So something goes wrong on python 3.10 without the type hint here. I can't really narrow down what it is except we use `get_type_hints` in attributes and directly use `__annotations__` in operations. I guess the options are:
- Put the type hint on and add this to the python 3.10 issue
- Use `__annotations__` instead (as far as I understand this might be bad for annotated things?)
- Put the type hint on, deprecate Annotated, then switch `__annotations__`?
- Press the deprecate 3.10 button?

superlopuh: Deprecating 3.10 is off the table, do you know what breaks when we use `__annotations__` instead? This feels like a nice-to-have to me, I would probably file an issue and move on until after we remove Annotated, indeed.

alexarice: It seems I have fixed it. There appears to be 3 ways to get typing annotations:
- directly use `__annotations__`. This seems to be the lowest level api, and in particular seems to resolve any "forward reference types" as strings, which isn't ideal. We use this in operations (but we don't rely on type annotations as much there)
- inspect.get_annotations (what I've used here) has an `eval_str` option which seems to resolve these forward references. Unfortunately it does not traverse subclasses (as `get_type_hints` does), but this seemed easily fixable
- typing.get_type_hints seems to be a higher level interface than annotations, and offers more features such as (optional) Annotation stripping and mro traversal. As this PR highlighted, it seemed to have a bug/different behaviour in 3.10 that meant that a bare `ClassVar` (i.e. without type subscript) caused an exception.

alexarice:
Updated the PR description post changes

superlopuh (approved):

PR #4779, opened 2025-07-14
core: allow constant ClassVar in attribute definitions

superlopuh:
Alternative implementation to #4777


alexarice:
This is definitely better. Are we sure we want to disallow "annotationless" classvars?

alexarice:
Do we make the same upper case restriction on operations? I feel it should be the same condition everywhere

superlopuh:
Yeah they're so similar to parameter definitions that it feels like a recipe for hard to debug bugs

superlopuh:
Yep, I moved the check out from operations that combines the check in one

alexarice (approved):
Lets go with this.

PR #4780, opened 2025-07-14
dialects: (builtin) add TensorType.constr

alexarice:
Adds a `constr` method for `TensorType`.

While writing this I realised I have forgot why we wanted these to be `classmethod`s, when they don't use the class information at all.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1405,6 +1405,18 @@ class TensorType(
     def get_element_type(self) -> AttributeCovT:
         return self.element_type
 
+    @classmethod
+    def constr(
+        cls,
+        *,
+        element_type: IRDLGenericAttrConstraint[AttributeInvT] | None = None,
+    ) -> GenericAttrConstraint[TensorType[AttributeInvT]]:
+        if element_type is None:
+            return BaseAttr[TensorType[AttributeInvT]](TensorType)
+        return ParamAttrConstraint[TensorType[AttributeInvT]](
+            TensorType, (AnyAttr(), element_type, AnyAttr())
+        )
+
 
 AnyTensorType: TypeAlias = TensorType[Attribute]
 AnyTensorTypeConstr = BaseAttr[TensorType[Attribute]](TensorType)


xdsl/dialects/builtin.py
```diff
@@ -1405,6 +1405,18 @@ def get_shape(self) -> tuple[int, ...]:
     def get_element_type(self) -> AttributeCovT:
         return self.element_type
 
+    @classmethod
```
superlopuh: Yeah the point is that we can't actually make them class methods IIRC, so we made them static methods. This is because there are things that classmethods let you write where the resulting constraint is inconsistent with the type signature, like `TensorType[A].constr()` which is actually a constraint for `TensorType[Attribute]` and will cause problems. Making the element_type non-optional may fix this issue.

alexarice: Oh right, I copied the `DenseArrayBase` one which is a `classmethod` (whoops)

xdsl/dialects/builtin.py
```diff
@@ -1405,6 +1405,17 @@ def get_shape(self) -> tuple[int, ...]:
     def get_element_type(self) -> AttributeCovT:
         return self.element_type
 
+    @staticmethod
+    def constr(
+        *,
+        element_type: IRDLGenericAttrConstraint[AttributeInvT] | None = None,
+    ) -> GenericAttrConstraint[TensorType[AttributeInvT]]:
+        if element_type is None:
+            return BaseAttr[TensorType[AttributeInvT]](TensorType)
```
superlopuh, line 1414: I'm not sure this is correct

xdsl/dialects/builtin.py
```diff
@@ -1405,6 +1405,17 @@ def get_shape(self) -> tuple[int, ...]:
     def get_element_type(self) -> AttributeCovT:
         return self.element_type
 
+    @staticmethod
+    def constr(
+        *,
+        element_type: IRDLGenericAttrConstraint[AttributeInvT] | None = None,
```
superlopuh, line 1411: Let's make this non-optional unless we need it? It's much trickier to deal with the cases where the attribute isn't provided

alexarice: This is pretty standard for all the `constr` methods we've done?

xdsl/dialects/builtin.py
alexarice: We've used this pattern in a lot of places, I believe it does work out (i.e. if `element_type` is none, then `AttributeInvT` must be `Attribute`)

superlopuh: Ah OK for some reason I thought it only worked if it isn't None. Happy to merge in this case

superlopuh: Worth a test maybe?

alexarice: I'm unsure, do we want a test for every `constr` method we define? (I can do it but wondering if this is what we really want)

superlopuh: Not sure, this is your call, it feels like we probably should have been adding small tests for this as we went but it seems not critical.

alexarice: I'm going to merge, as its blocking other things, but will think about this.

PR #4782, opened 2025-07-14
misc: cleanup ops using ConstraintVar

alexarice:
Move a couple of ops over to the new system.


alexarice:
Rebased onto the standardisation to simplify this.

PR #4784, opened 2025-07-14
dialects: (linalg) add LinalgOperation ABC

superlopuh:
This is the first change in a number that will let us operate on linalg operations whether or not they've been converted to generic, and actually to convert linalg ops to generic from within xDSL. Subsequent PRs will subclass the other Linalg ops to this interface.

### DIFF
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from abc import ABC
+from abc import ABC, abstractmethod
 from collections.abc import Mapping, Sequence
 from enum import auto
 from typing import ClassVar, cast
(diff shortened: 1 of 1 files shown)


xdsl/dialects/linalg.py
```diff
@@ -201,6 +168,52 @@ def get_static_loop_ranges(self) -> tuple[int, ...]:
...
+        iterator_types: Sequence[Attribute] | ArrayAttr[Attribute],
+        result_types: Sequence[Attribute] = (),
+        doc: StringAttr | None = None,
+        library_call: StringAttr | None = None,
+    ) -> None:
+        super().__init__(
+            operands=[inputs, outputs],
+            result_types=[result_types],
+            properties={
+                "indexing_maps": ArrayAttr(indexing_maps),
```
alexarice, line 206:
```suggestion
                "indexing_maps": ArrayAttr.get(indexing_maps),
```
I know its not the content of this change but just realised that this works as `ArrayAttr` is a `Sequence` so it rebuilds itself.

alexarice (approved):
Looks good

xdsl/dialects/linalg.py
superlopuh: Doesn't work since this takes a sequence and .get takes a tuple. We can revisit this at some point but I'd rather not take a Sequence here TBH

alexarice: I realised this as an issue when defining `Data.get`. Perhaps `ArrayAttr` should have its own `get` anyway.

PR #4785, opened 2025-07-14
misc: standardise `constr` methods

alexarice:
Tries to standardise the form of `constr` methods, such that:
- They are static methods
- They are more consistent with which arguments are positional/keyword.

I've left `StencilType` alone for this PR as it seems to actually rely on its `constr` being a class method

### DIFF
--- a/xdsl/dialects/smt.py
+++ b/xdsl/dialects/smt.py
@@ -134,7 +134,6 @@ class FuncType(ParametrizedAttribute, TypeAttribute):
 
     @staticmethod
     def constr(
-        *,
         domain: GenericRangeConstraint[NonFuncSMTType],
         range: GenericAttrConstraint[NonFuncSMTType],
     ) -> GenericAttrConstraint[FuncType]:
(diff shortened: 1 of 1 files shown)


tests/dialects/test_builtin.py
```diff
@@ -766,7 +766,7 @@ def test_integer_type_repr():
 
 
 def test_vector_constr():
-    constr = VectorType.constr(i32)
+    constr = VectorType.constr(element_type=i32)
```
superlopuh: If we're standardised then hopefully we can make the base always positional and the rest always kw? Feels like a readable balance of things.

alexarice: `element_type` is a parameter, not a base

alexarice: There's an argument its the "important one" though. I'd be happy to go the other way with this if we want to allow a single position "important" parameter

superlopuh: Sorry yeah I got confused there, but this is what I meant, I feel like it's the most important thing to pass and would make sense to special-case IMO.

alexarice: ok, I'll try to standardise the other way round

superlopuh (approved):

alexarice:
I might just check that @math-fehr is ok with the changes to `smt` dialect (I can revert if not) before merging

math-fehr (approved):
I would personally disagree in the statement "the arguments are always keywords", but it's not what you implemented so it's fine by me!
I would say we should just treat it like usual functions? Not necessarily requiring keyword arguments, unless we think it's prettier?
To be honest, I would say that even MemRefType.constr(i32, Shape) would be readable enough, I wouldn't necessarily require a keyword (the same way I wouldn't in most functions).

I still approve the PR, just mentioning this in general

alexarice:
Sorry, I shouldn't have edited the PR description, we decided to standardise the other way. I'm not sure how I feel about `MemRefType.constr(i32, Shape)`, but based on this I'll revert the smt function type change, as that has a clear ordering.

PR #4786, opened 2025-07-14
x86: integration test vector -> x86

qaco:


tests/filecheck/projects/libxsmm/vect_matmul_kernel.mlir
```diff
@@ -0,0 +1,59 @@
+// RUN: xdsl-opt -p convert-vector-to-ptr,convert-memref-to-ptr{lower_func=true},convert-ptr-type-offsets,canonicalize,convert-func-to-x86-func,convert-vector-to-x86{arch=avx2},convert-ptr-to-x86{arch=avx2},convert-arith-to-x86,reconcile-unrealized-casts,x86-allocate-registers %s
```
superlopuh:
```suggestion
// RUN: xdsl-opt -p convert-vector-to-ptr,convert-memref-to-ptr{lower_func=true},convert-ptr-type-offsets,canonicalize,convert-func-to-x86-func,convert-vector-to-x86{arch=avx2},convert-ptr-to-x86{arch=avx2},convert-arith-to-x86,reconcile-unrealized-casts,canonicalize,x86-allocate-registers,canonicalize -t x86-asm %s | filecheck %s
```

superlopuh:
Why f64?

qaco:
> Why f64?

limit the code size of our example (ie in avx2 handling ```vector<4xf64>``` instead of ```vector<8xf32>```).

PR #4787, opened 2025-07-15
misc: use ValueError in core.py

superlopuh:
I think we should rarely use raise Exception, and we may want to consider a more precise error for illegal IR manipulation in the future, but this seems like a safe improvement.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -666,7 +666,7 @@ class SSAValue(Generic[AttributeCovT], IRWithUses, ABC):
         If safe_erase is False, then replace its uses by an ErasedSSAValue.
         """
         if safe_erase and len(self.uses) != 0:
-            raise Exception(
+            raise ValueError(
                 "Attempting to delete SSA value that still has uses of result "
                 f"of operation:\n{self.owner}"
             )
@@ -1054,7 +1054,7 @@ class Operation(_IRNode):
     def add_region(self, region: Region) -> None:
         """Add an unattached region to the operation."""
         if region.parent:
-            raise Exception(
+            raise ValueError(
                 "Cannot add region that is already attached on an operation."
             )
         self.regions += (region,)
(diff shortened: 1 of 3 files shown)


xdsl/ir/core.py
```diff
@@ -1054,7 +1054,7 @@ def create(
     def add_region(self, region: Region) -> None:
         """Add an unattached region to the operation."""
         if region.parent:
-            raise Exception(
+            raise ValueError(
```
alexarice, line 1057: Places like this that didn't require a change are presumably untested. This seems like a good opportunity to remedy this?

xdsl/ir/core.py
```diff
@@ -1589,7 +1589,7 @@ def insert_arg(self, arg_type: Attribute, index: int) -> BlockArgument:
         Returns the new argument.
         """
         if index < 0 or index > len(self._args):
-            raise Exception("Unexpected index")
+            raise ValueError("Unexpected index")
```
alexarice, line 1592:
```suggestion
            raise ValueError(f"Cannot insert argument into block {self} at index {index}"
            f"index must be in range [0,{len(self._args)}]")
```
Lets improve this? (above is just a suggestion)

xdsl/ir/core.py
superlopuh: I think this can be a great first issue for new members, I have a bit too much going on right now: #4793

alexarice: Sounds good, I understand it wasn't the point of the PR, but PRs like this do highlight where stuff isn't tested.

xdsl/ir/core.py
superlopuh: I added it to the issue

xdsl/ir/core.py
superlopuh: Yep the issue is definitely useful, might be a good thing for MLIR school attendees to add

PR #4788, opened 2025-07-15
misc: use ValueError in xdsl_opt_main.py

superlopuh:
I was surprised that there wasn't a better Exception subclass for command-line options, but ValueError seems better to me than Exception.

### DIFF
--- a/tests/xdsl_opt/test_xdsl_opt.py
+++ b/tests/xdsl_opt/test_xdsl_opt.py
@@ -14,7 +14,7 @@ from xdsl.context import Context
 from xdsl.dialects import builtin, get_all_dialects
 from xdsl.passes import ModulePass
 from xdsl.transforms import get_all_passes
-from xdsl.utils.exceptions import DiagnosticException
+from xdsl.utils.exceptions import DiagnosticException, ParseError
 from xdsl.xdsl_opt_main import xDSLOptMain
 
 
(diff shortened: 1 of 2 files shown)


xdsl/xdsl_opt_main.py
```diff
@@ -371,7 +371,7 @@ def output_resulting_program(self, prog: ModuleOp) -> str:
         """Get the resulting program."""
         output = StringIO()
         if self.args.target not in self.available_targets:
-            raise Exception(f"Unknown target {self.args.target}")
+            raise ValueError(f"Unknown target {self.args.target}")
```
alexarice: This is missing a test above

superlopuh: What do you mean?

alexarice: I meant that there seemed to be cases in `test_error_on_run` for all the errors apart from this one, and it seems it would be trivial to add a test that triggers this error?

superlopuh: Oh it's just tested in its own test but I can merge them if that's what you mean?

superlopuh: Can merge with wrong pass also

superlopuh: I'll do this

alexarice: No it's fine

superlopuh: I tried this, and it actually raises an error not at runtime, but at init time, so it should stay a separate test.

superlopuh: Actually I refactored the tests to a similar pattern so that we can extend them in the future

tests/xdsl_opt/test_xdsl_opt.py
```diff
@@ -44,57 +44,83 @@ def test_empty_program():
...
+    "args, expected_error",
+    [
+        (
+            ["tests/xdsl_opt/empty_program.mlir", "-t", "wrong"],
+            "invalid choice: 'wrong'",
+        ),
+    ],
+)
+def test_error_on_argparse(
+    capsys: pytest.CaptureFixture[str], args: list[str], expected_error: str
```
superlopuh, line 117: Found out about this helper while googling for how to test argparse, we might want to use it in other tests in the future

tests/xdsl_opt/test_xdsl_opt.py
```diff
@@ -44,57 +44,83 @@ def test_empty_program():
...
         opt.run()
 
 
 @pytest.mark.parametrize(
-    "args, expected_error",
+    "args, error_type, expected_error",
     [
         (
             ["tests/xdsl_opt/empty_program.mlir", "-p", "wrong"],
+            Exception,
```
alexarice: Should this be a `ValueError`?

superlopuh: Probably, I can update it

tests/xdsl_opt/test_xdsl_opt.py
alexarice: I'm not sure I understand what this does, captures errors automatically?

superlopuh: Yep exactly, it captures stdout and stderr

PR #4789, opened 2025-07-15
testing: add diagnostic on register allocation error

superlopuh:
Turns out that the diagnostic error I added in register allocation wasn't actually being printed in a pretty way, this PR adds both IR printing and a test to make sure that we see the diagnostic with IR.

### DIFF
--- a/tests/backend/test_register_allocation.py
+++ b/tests/backend/test_register_allocation.py
@@ -1,8 +1,10 @@
 import re
 from collections.abc import Sequence
+from typing import ClassVar
 
 import pytest
 
+from xdsl.backend.block_naive_allocator import BlockNaiveAllocator
 from xdsl.backend.register_allocatable import (
     HasRegisterConstraints,
     RegisterAllocatableOperation,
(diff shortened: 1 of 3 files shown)


xdsl/utils/diagnostic.py
```diff
@@ -30,6 +30,7 @@ def raise_exception(self, ir: IRNode, underlying_error: Exception) -> NoReturn:
                 p.print_string("\n")
             case Block():
                 p.print_block(toplevel)
+                p.print_string("\n")
```
superlopuh, line 33: the last operation in a block is not terminated with a newline, so the diagnostic is not printed without this change.

alexarice: This might be left over from the `printer.print` changes

tests/backend/test_register_allocation.py
```diff
@@ -327,3 +334,40 @@ def test_allocate_values_same_reg():
...
+        ),
+    ) as e:
+        allocator.allocate_block(block)
+
+    assert getattr(e.value, "__notes__") == [
+        """
+^0:
+  %0 = "test.inout"(%1) : (!test.reg<a0>) -> !test.reg<a1>
+  ^^^^^^^^^^^^^^^^^----
+  | Error allocating op
```
alexarice, line 370: Should the rest of the error appear here? Not hugely familiar with the diagnostics system

superlopuh: Hmm, maybe it should, currently it's just the message that's added in `add_message`, but maybe we should also add the original exception's description...

superlopuh: Seems like a separate PR question to me.

PR #4794, opened 2025-07-15
dialects: (snitch-runtime) remove `ConstraintVar`

alexarice:
Seems to be the only generic operation we have, but it's interesting that this seems to work.

### DIFF
--- a/xdsl/dialects/snitch_runtime.py
+++ b/xdsl/dialects/snitch_runtime.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
-from typing import Annotated, Generic
+from typing import Annotated, ClassVar, Generic
 
 from typing_extensions import TypeVar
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/snitch_runtime.py
```diff
@@ -322,11 +325,11 @@ class DmaStart1DBaseOperation(SnitchRuntimeBaseOperation, Generic[_T], ABC):
...
     """
 
-    T = Annotated[Attribute, ConstraintVar("T"), _T]
+    _T: ClassVar = VarConstraint("T", TypeVarConstraint(_T, AnyAttr()))
+
     dst = operand_def(_T)
     src = operand_def(_T)
-    # Pylance was complaining about the below.
-    # size = operand_def(Annotated[Attribute, i32])
+
```
alexarice, line 326: Wait, I'm an idiot, this wasn't even using `T`

superlopuh (approved):

PR #4795, opened 2025-07-15
dialects: (x86) add get_constant_value

superlopuh:
This helper will be used in canonicalization patterns to fold constants, wanted to add it separately first.

### DIFF
--- a/xdsl/transforms/canonicalization_patterns/x86.py
+++ b/xdsl/transforms/canonicalization_patterns/x86.py
@@ -1,4 +1,5 @@
 from xdsl.dialects import x86
+from xdsl.ir import Attribute, OpResult, SSAValue
 from xdsl.pattern_rewriter import (
     PatternRewriter,
     RewritePattern,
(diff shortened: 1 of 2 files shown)


xdsl/transforms/canonicalization_patterns/x86.py
```diff
@@ -11,3 +12,14 @@ class RemoveRedundantDS_Mov(RewritePattern):
     def match_and_rewrite(self, op: x86.DS_MovOp, rewriter: PatternRewriter) -> None:
         if op.destination.type == op.source.type and op.destination.type.is_allocated:
             rewriter.replace_matched_op((), (op.source,))
+
+
+def get_constant_value(value: SSAValue) -> Attribute | None:
```
alexarice:
```suggestion
def get_constant_value(value: SSAValue) -> IntegerAttr | None:
```
It seems that immediates are always integer attrs?

alexarice (approved):

xdsl/transforms/canonicalization_patterns/x86.py
superlopuh: I guess they are for now, I'll restrict the type

PR #4796, opened 2025-07-15
core: (Constraints) deprecate `ConstraintVar`

alexarice:
`param_def` api means that this can finally be scrapped.

### DIFF
--- a/tests/irdl/test_operation_definition.py
+++ b/tests/irdl/test_operation_definition.py
@@ -168,16 +168,17 @@ def test_attr_verify():
         op.verify()
 
 
-# TODO: remove this test once the Annotated API is deprecated
-@irdl_op_definition
-class ConstraintVarOp(IRDLOperation):
-    name = "test.constraint_var_op"
+with pytest.deprecated_call():
+    # TODO: remove this test once the Annotated API is deprecated
+    @irdl_op_definition
+    class ConstraintVarOp(IRDLOperation):
+        name = "test.constraint_var_op"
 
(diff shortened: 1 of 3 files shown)


alexarice:
Not sure there's any way I can make a nice diff until things are merged unfortunately.

alexarice:
This is now unstacked

superlopuh:
My first thought is that this isn't deprecating, it's deleting, which could be OK, but I think it would actually be good to do one release with it deprecated, and to remove it in the next one?

alexarice:
That seems reasonable, I'd have to see if I could still get the PRs further down the chain to work, as the partial motivation of this was to make `from_pyrdl` simpler

alexarice:
Made a version that actually does deprecation

tests/irdl/test_operation_definition.py
```diff
@@ -168,89 +167,6 @@ def test_attr_verify():
...
-# TODO: remove this test once the Annotated API is deprecated
-@irdl_op_definition
-class ConstraintVarOp(IRDLOperation):
-    name = "test.constraint_var_op"
-
-    T = Annotated[IntegerType | IndexType, ConstraintVar("T")]
-
-    operand = operand_def(T)
-    result = result_def(T)
-    attribute = attr_def(T)
```
superlopuh: I would propose keeping the test around while suppressing the deprecation warnings until we delete the mechanism, because things should work even if they're deprecated.

```py
    with pytest.deprecated_call():
```

alexarice: Added it back

PR #4797, opened 2025-07-15
testing: use match argument for error testing

superlopuh:
I think this is a nice linting error. I also moved the message testing to the match clause, which is important for when the exception type is not initialised with just a string.


EdmundGoodman (approved):
lgtm -- but I think there are still some more cases which need changing as pre-commit is now failing

PR #4799, opened 2025-07-15
core: (ParametrizedAttribute) simplify `from_pyrdl`

alexarice:
Reduces the two loops to a single loop. This will make future refactors (e.g. converters) easier, as `parameters` is only set in one place.


alexarice:
Unstacked and added the constraint var checking back

alexarice:
Rebased past constraint var deprecation

PR #4802, opened 2025-07-15
dialects:(omp) add private and reduction declarations

dk949:
Adding the `PrivateClauseOp` and `DeclareReductionOp` operations.


xdsl/dialects/omp.py
```diff
@@ -583,6 +584,61 @@ def num_block_args(self) -> int:
...
+    cleanup_region = region_def()
+
+    traits = traits_def(IsolatedFromAbove(), SymbolOpInterface())
+
+    def verify_(self) -> None:
+        if len(self.alloc_region.blocks) > 1:
+            raise VerifyException(
+                f"{self.name} should have at most 1 block in alloc_region"
+            )
+        return super().verify_()
```
```suggestion
```

xdsl/dialects/omp.py
```diff
@@ -583,6 +584,61 @@ def num_block_args(self) -> int:
...
+    dealloc_region = region_def()
+
+    traits = traits_def(IsolatedFromAbove())
+
+    def verify_(self) -> None:
+        if len(self.alloc_region.blocks) < 1:
+            raise VerifyException(
+                f"alloc_region of {self.name} has to have at least 1 block"
+            )
+        return super().verify_()
```
```suggestion
```

tests/filecheck/dialects/omp/ops.mlir
```diff
@@ -224,6 +224,22 @@ builtin.module {
     }) : () -> ()
     func.return
   }
+  "omp.private"() <{data_sharing_type = #omp.data_sharing_type {type = private}, sym_name = "p1", type = i32}> ({
```
superlopuh: Could you please add the assembly format for these functions?

dk949: I'm having a bit of trouble with it due to the second and third region being optional in the assembly format, but not actually being optional regions.

Opened a thread on zulip about this (thread is about `PrivateClauseOp`, but they both have the same issue).

dk949: I could submit a separate PR adding the assembly format once #4807 (or something like it) is merged?

superlopuh: Ah nice I think we can probably merge that quite quickly, would be a nice improvement

dk949: So that's been merged, but there's another issue. The `sym_name` attribute is expected to be `SymbolNameAttr` upstream, which doesn't typically matter, except that it gets printed differently depending if you do generic or custom assembly.

In generic format it's printed as a `StringAttr` (which is what we have in the dialect right now), but when doing custom assembly it's printed as a `SymbolRefAttr`.

It appears that `SymbolNameAttr` has recently been removed (0fce942b1ecbc85299eb6d626b434ab6a8499ea2)

alexarice: For what it's worth, that SymbolNameAttr wouldn't have helped you as it didn't have any special assembly format support. 

The idea was to instead introduce a `SymbolNameConstr` which you would act as a `StringAttr` but have special assembly format support, but I never got round to implementing it (partially because nothing has needed it yet)

alexarice: I might have time today to implement this.

dk949: Could you tag me in it if you do implement it?

I'll be working on 2 more ops for `omp` today. Maybe we can have this one merged without assembly format and I'll do another PR when the assembly format support is merged, @superlopuh ?

superlopuh: I just feel like it's much easier to read and understand the tests if they're written in the custom format, so if Alex says that he can implement this today then I'd rather wait for that. If Alex can't promise this then I'd be happy to merge and migrate later.

alexarice: https://github.com/xdslproject/xdsl/pull/4811 I promise.

dk949: Thank you, Alex!

dk949:
Sorry, I messed up one last thing. Upstream requires `BlockArgOpenMPOpInterface` operations to have *at least* that many block args not exactly that many. This is something `flang` takes advantage of.

xdsl/dialects/omp.py
```diff
@@ -520,7 +522,7 @@ class WsLoopOp(BlockArgOpenMPOperation):
         ArrayAttr.constr(RangeOf(base(SymbolRefAttr), length=REDUCTION_COUNT))
     )
     reduction_mod = opt_prop_def(ReductionModifierAttr)
-    reduction_byref = opt_prop_def(DenseIntOrFPElementsAttr[i1])
+    reduction_byref = opt_prop_def(DenseArrayBase[i1])
```
alexarice: Are these part of a different change? As an aside, it is very confusing to me that this typechecks, `i1` is an instantiation of `IntegerType` so how comes it can be a generic type parameter?

dk949: Oh, they are. I think I got confused last night and thought I was still working on the 20.0.7 fixes PR, my bad, I can take this out and push as a separate PR (which is where I made this mistake).

alexarice: Would be good if they can be separated :+1:

dk949: No problem

dk949: Took out the unrelated changes

xdsl/dialects/omp.py
```diff
@@ -583,6 +585,76 @@ def num_block_args(self) -> int:
...
+    copy_region = region_def()
+    dealloc_region = region_def()
+
+    traits = traits_def(IsolatedFromAbove())
+
+    assembly_format = """
+        $data_sharing_type $sym_name `:` $type
+        `alloc` $alloc_region
+        (`copy` $copy_region^)?
+        (`dealloc` $dealloc_region^)?
```
alexarice, line 647: This looks very different to the upstream docs, what's going on? (upstream has `init` instead of `alloc` and its optional)

xdsl/dialects/omp.py
```diff
@@ -583,6 +585,76 @@ def num_block_args(self) -> int:
...
+    See external [documentation](https://mlir.llvm.org/docs/Dialects/OpenMPDialect/ODS/#ompdeclare_reduction-ompdeclarereductionop).
+    """
+
+    name = "omp.declare_reduction"
+
+    sym_name = prop_def(SymbolNameConstraint())
+    var_type = prop_def(TypeAttribute, prop_name="type")
+
+    alloc_region = region_def()
+    init_region = region_def()
```
alexarice, line 601:
```suggestion
    initializer_region = region_def()
```
Nit: Would it make more sense to stick with upstream naming?

alexarice (commented):
Bit confused about why private clause looks different (maybe stuff has changed between master/20.1.7?
Other than that I'm happy for this to be merged.

xdsl/dialects/omp.py
dk949: This was copies straight from upstream <https://github.com/llvm/llvm-project/blob/llvmorg-20.1.7/mlir/include/mlir/Dialect/OpenMP/OpenMPOps.td#L109-L113>

Looks like there was another breaking change since 20.1.7

alexarice (approved):

PR #4805, opened 2025-07-16
dialects: (x86) optimize out addition of 0

superlopuh:


xdsl/dialects/x86/ops.py
```diff
@@ -1026,6 +1034,8 @@ class RS_AddOp(RS_Operation[GeneralRegisterType, GeneralRegisterType]):
 
     name = "x86.rs.add"
 
+    traits = traits_def(Pure(), RS_AddOpHasCanonicalizationPatterns())
```
alexarice, line 1037: Is this definitely pure? Doesn't it modify the input register `r`, or do you enforce that `r` isn't used again after this op?

alexarice (approved):
If the plan is to add a lot of these I wonder if it would be worth doing something like I did for `arith`, with general functions for specifying units/operations?

xdsl/dialects/x86/ops.py
superlopuh: I was just thinking about this, I think we actually need to add a verification for this, it should be possible to do in an automated way with the register constraints that we use for regalloc

superlopuh:
Maybe, although I think it would probably be worth adding some by hand first to see what kinds of patterns emerge.

xdsl/dialects/x86/ops.py
alexarice: So these operations don't work like the "destination passing style" things? (Not that I'm 100% sure on those things.

PR #4807, opened 2025-07-16
WIP: core:(assembly-format) allow empty region blocks to be optional in assembly format

dk949:
This allows non-optional regions to be anchorable by treating missing regions as empty.

As far as I can tell this is consistent with upstream.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -930,6 +930,12 @@ class RegionVariable(RegionDirective, VariableDirective):
         state.last_was_punctuation = False
         state.should_emit_space = True
 
+    def is_anchorable(self) -> bool:
+        return True
+
+    def set_empty(self, state: ParsingState):
+        state.regions[self.index] = Region()
+
 
 @dataclass(frozen=True)
 class VariadicRegionVariable(RegionDirective, VariadicVariable):


dk949:
I'm not very familiar with this part of the codebase so would really appreciate any tips on what I may have missed.

(will add tests in a little bit)

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -930,6 +930,12 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
         state.last_was_punctuation = False
         state.should_emit_space = True
 
+    def is_anchorable(self) -> bool:
+        return True
```
alexarice, line 934: When setting `is_anchorable` you need to also implement `is_present` which the optional group uses to determine whether to print.

alexarice (commented):
This seems like a good change if it agrees with what MLIR does. It would be nice to get a simple test of a region being used as an anchor in `tests/irdl/test_declarative_assembly_format.py`

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -2140,6 +2140,40 @@ class OptionalRegionOp(IRDLOperation):
...
+        ),
+    ],
+)
+def test_optional_groups_empty_regions(format: str, program: str, generic_program: str):
+    """Test the parsing of empty regions in an optional group"""
+
+    @irdl_op_definition
+    class EmptyRegionOp(IRDLOperation):
+        name = "test.empty_region_group"
+        irdl_options = [AttrSizedRegionSegments]
```
alexarice: Why do we need this option?

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -2140,6 +2140,40 @@ class OptionalRegionOp(IRDLOperation):
...
+@pytest.mark.parametrize(
+    "format, program, generic_program",
+    [
+        (
+            "(`keyword` $maybe_empty^)? attr-dict",
+            "test.empty_region_group",
+            '"test.empty_region_group"() ({}) : () -> ()',
+        ),
+        (
+            "(`keyword` $maybe_empty^)? attr-dict",
```
alexarice: Could we inline the format? it's the same for both.

dk949:
> This seems like a good change if it agrees with what MLIR does. It would be nice to get a simple test of a region being used as an anchor in `tests/irdl/test_declarative_assembly_format.py`

The one thing I'm currently unsure of is whether an anchored region is allowed to be the first directive in an optional group. The way I have it now, it's not

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -2140,6 +2140,40 @@ class OptionalRegionOp(IRDLOperation):
...
+            'test.empty_region_group keyword {\n  "test.op"() : () -> ()\n}',
+            '"test.empty_region_group"() ({ "test.op"() : () -> ()}) : () -> ()',
+        ),
+    ],
```
alexarice, line 2158: Could we add one test with a single empty block? Occured to me that I'm not sure how MLIR treats this.

alexarice (commented):
This is looking great. Thanks for fixing this.

alexarice:
> 
> The one thing I'm currently unsure of is whether an anchored region is allowed to be the first directive in an optional group. The way I have it now, it's not

It's irrelevant whether the first element is the anchor. The important thing is that the first element is marked as optionally parsable, meaning that it successfully parses the empty string

tests/irdl/test_declarative_assembly_format.py
dk949: Do you mean something like:

```mlir
test.empty_region_group keyword { }
```

It currently roundtrips to:

```mlir
test.empty_region_group
```

Which I believe is consistent with MLIR

alexarice: I meant something like `test.empty_region_group keyword { ^bb1: }`

alexarice: I'm having a play round with MLIR now and am less sure about what the intended behaviour here should be.

alexarice: What was the example op from openmp that uses this feature?

dk949: This roundtrips with no changes. I'm not 100% sure what upstream behaviour of this is, but I expect the same?

tests/irdl/test_declarative_assembly_format.py
dk949: Removed

tests/irdl/test_declarative_assembly_format.py
dk949: Done

tests/irdl/test_declarative_assembly_format.py
alexarice: Seems hard to tell what the upstream behaviour is, as most things require a terminator.

dk949: Yup. My expectation is that since `keyword {}` becomes nothing, `keyword {^0:}` should retain the empty block, but it's just a guess.

alexarice: Lets go with this for now and if we find evidence otherwise we can change it (shouldn't effect the ops you care about I think)

dk949: > What was the example op from openmp that uses this feature?

`omp.private` uses this:

```mlir
"builtin.module"() ({
  "omp.private"() <{data_sharing_type = #omp<data_sharing_type{type = private}>, sym_name = "p1", type = i32}> ({
  ^bb0(%arg0: i32):
    %0 = "arith.constant"() <{value = 0 : i32}> : () -> i32
    "omp.yield"(%0) : (i32) -> ()
  }, {
  }, {
  }) : () -> ()
}) : () -> ()
```

Becomes:

```mlir
module {
  omp.private {type = private} @p1 : i32 alloc {
  ^bb0(%arg0: i32):
    %c0_i32 = arith.constant 0 : i32
    omp.yield(%c0_i32 : i32)
  }
}
```

PR #4808, opened 2025-07-16
dialects: (x86) store and load at constant offset canonicalization pattern

superlopuh:
Added the patterns to the abstract superclass ops, seems safe to me for now but I'm not 100% sure the rules are the same for all memory operations.


qaco:
> Added the patterns to the abstract superclass ops, seems safe to me for now but I'm not 100% sure the rules are the same for all memory operations.

Not sure indeed. x86 can be odd. We may want to document this behaviour at least in the naming of the base classes (adding "WithOffset" or something). Or with a dedicated trait ?

superlopuh:
Well the base classes always have a memory_offset, so I think it's fine? I'm leaning towards merging this for now and seeing about the rest later.

qaco:
> Well the base classes always have a memory_offset, so I think it's fine? I'm leaning towards merging this for now and seeing about the rest later.

yeay you're right

PR #4811, opened 2025-07-16
core: (SymbolNameConstraint) add assembly format symbol names

alexarice:
Adds a new constraint `SymbolNameConstr`, which takes the role that `SymbolNameAttr` does in MLIR (note that `SymbolNameAttr` is not an attribute in MLIR, but some tablegen construct that I think translates to a constraint).
The assembly format then uses this information to parse/print the attribute as a symbol. Made this work with optional attributes and optional groups.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -21,7 +21,7 @@ from xdsl.dialects.builtin import (
     MemRefType,
     ModuleOp,
     StringAttr,
-    SymbolNameConstr,
+    SymbolNameConstraint,
     UnitAttr,
     i32,
 )
@@ -724,8 +724,8 @@ def test_symbol_name_variable(program: str, generic_program: str):
     class SymbolNameOp(IRDLOperation):
         name = "test.symbol"
 
-        sym_name = prop_def(SymbolNameConstr())
-        attr_sym_name = attr_def(SymbolNameConstr())
+        sym_name = prop_def(SymbolNameConstraint())
+        attr_sym_name = attr_def(SymbolNameConstraint())
 
         assembly_format = "$sym_name $attr_sym_name attr-dict"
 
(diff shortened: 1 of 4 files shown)


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -705,6 +706,69 @@ class TypedAttributeOp(IRDLOperation):
...
+            '"test.symbol"() <{sym_name = "123"}> {attr_sym_name = "456"} : () -> ()',
+        ),
+    ],
+)
+def test_symbol_name_variable(program: str, generic_program: str):
+    @irdl_op_definition
+    class SymbolNameOp(IRDLOperation):
+        name = "test.symbol"
+
+        sym_name = prop_def(SymbolNameConstr())
```
alexarice: We could likely do something like `SymbolNameAttr = SymbolNameConstr()` to make this look more like the MLIR if we want.

xdsl/dialects/builtin.py
```diff
@@ -255,6 +255,30 @@ def print_builtin(self, printer: Printer):
         printer.print_string_literal(self.data)
 
 
+@dataclass(frozen=True)
+class SymbolNameConstr(GenericAttrConstraint[StringAttr]):
```
alexarice: This should likely be `SymbolNameConstraint`, I'll change it now. If we really want we could call it `SymbolNameAttr` instead, but it feels dishonest.

alexarice: @superlopuh any thoughts on naming?

dk949 (approved):
Test behaviour looks good wrt. upstream

Thank you!

xdsl/dialects/builtin.py
superlopuh: I like this

tests/irdl/test_declarative_assembly_format.py
superlopuh: I don't think this is necessary

PR #4812, opened 2025-07-16
frontend: (pyast) add new decorator-based API for frontend code generation

EdmundGoodman:
This PR adds a new decorator-based API for frontend code generation. This will supersede the old context manager API.

An example of this new API is as follows:

```python
from xdsl.dialects.arith import AddfOp, MulfOp
from xdsl.dialects.builtin import f64
from xdsl.frontend.pyast.context import PyASTContext

# `FrontendContext` encapsulates the mapping from Python to xDSL constructs
ctx = PyASTContext()
ctx.register_type(float, f64)
ctx.register_function(float.__add__, AddfOp)
ctx.register_function(float.__mul__, MulfOp)

# Functions can be parsed in a context to yield a lazy wrapper for the IR
# representation of that function
@ctx.parse_program
def test_arith(x: float, y: float, z: float) -> float:
    return x + y * z

# The lazy wrapper can still be called as if it were a native Python function
print(test_arith(1.0, 2.0, 3.0))
# CHECK: 7.0

# But the wrapper also provides a property to get an IR representation
print(test_arith.module)
# CHECK:       builtin.module {
# CHECK-NEXT:  func.func @test_arith(%x : f64, %y : f64, %z : f64) -> f64 {
# CHECK-NEXT:    %0 = arith.mulf %y, %z : f64
# CHECK-NEXT:    %1 = arith.addf %x, %0 : f64
# CHECK-NEXT:    func.return %1 : f64
# CHECK-NEXT:  }
# CHECK-NEXT:}
```

### DIFF
--- a/xdsl/frontend/builder.py
+++ b/xdsl/frontend/builder.py
@@ -1,0 +1,17 @@
+from attr import dataclass
+
+from xdsl.dialects.builtin import ModuleOp
+from xdsl.frontend.pyast.utils.type_conversion import FunctionRegistry, TypeRegistry
+
+
+@dataclass
+class PyASTBuilder:
+    """Build an operation ."""
+
+    type_registry: TypeRegistry
+    function_registry: FunctionRegistry
+    desymref: bool
+
+    @property
+    def module(self) -> ModuleOp:
+        return ModuleOp([])
--- a/xdsl/frontend/builder.py
+++ b/xdsl/frontend/builder.py
@@ -49,7 +49,3 @@ class PyASTBuilder:
             module.verify()
 
         return module
-
-    # @property
-    # def module(self) -> ModuleOp:
-    #     return ModuleOp([])
--- a/xdsl/frontend/builder.py
+++ b/xdsl/frontend/builder.py
@@ -1,7 +1,16 @@
-from attr import dataclass
+import ast
+from dataclasses import dataclass
+from typing import Any
 
 from xdsl.dialects.builtin import ModuleOp
-from xdsl.frontend.pyast.utils.type_conversion import FunctionRegistry, TypeRegistry
+from xdsl.frontend.pyast.code_generation import CodeGeneration
+from xdsl.frontend.pyast.utils.python_code_check import PythonCodeCheck
+from xdsl.frontend.pyast.utils.type_conversion import (
+    FunctionRegistry,
+    TypeConverter,
+    TypeRegistry,
+)
+from xdsl.transforms.desymref import Desymrefier
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


xdsl/frontend/pyast/program.py
```diff
@@ -17,6 +20,30 @@
...
+
+
+@dataclass
+class PyASTProgram(Generic[P, R]):
+    """Wrapper to associate an IR representation with a Python function."""
+
+    name: Final[str]
+    func: Final[Callable[P, R]]
+    _builder: Final[PyASTBuilder]
+    _module: ModuleOp | None = None
```
superlopuh: Can you please add some small doc strings for these?

xdsl/frontend/pyast/context.py
```diff
@@ -1,12 +1,107 @@
...
+
+    def parse_program(
+        self,
+        decorated_func: Callable[P, R] | None = None,
+        *,
+        desymref: bool = True,
+    ) -> Callable[[Callable[P, R]], PyASTProgram[P, R]] | PyASTProgram[P, R]:
+        """Get a program wrapper by decorating a function."""
+
+        def decorator(func: Callable[P, R]) -> PyASTProgram[P, R]:
```
superlopuh, line 65: Would be nice to preserve doc strings etc from the original, let's add a test for the doc string also?
```suggestion
        @functools.wraps(decorated_func)
        def decorator(func: Callable[P, R]) -> PyASTProgram[P, R]:
```
https://docs.python.org/3/library/functools.html#functools.wraps

xdsl/frontend/builder.py
```diff
@@ -0,0 +1,51 @@
...
+    """Build an operation ."""
+
+    type_registry: TypeRegistry
+    function_registry: FunctionRegistry
+
+    file: str
+    globals: dict[str, Any]
+    ast: ast.FunctionDef
+
+    desymref: bool
```
superlopuh: Can you please add a one-line doc string for each of these (maybe a bit longer for desymref), and for the functions below?

xdsl/frontend/pyast/context.py
EdmundGoodman: Because the `decorated_func` can be `None` to support invocation as both `ctx.parse_program` and `ctx.parse_program()`, the `@functools.wraps` decorator cannot be used.

I've instead done something which I think should be equivalent directly calling `functools.update_wrapper` within the decorator.

superlopuh: Nice, can you please print `test_arith.__doc__` to make sure it works?

EdmundGoodman: Standalone expressions are not currently supported by the AST parser, so any test cases including docstrings fail, but have tested it manually without the parsing step and the docstrings are preserved, and added `assert program.__doc__ == func.__doc__` which should also guarantee this

superlopuh (approved):

PR #4816, opened 2025-07-17
dialects: (builtin) remove simple uses of `I{n}`

alexarice:
As the title says, looking towards deprecating `Annotated` things, this PR removes the easy uses of `I64` etc.

Also allows `IRDLAttrConstraints` in `IntegerAttr.constr` so that you can write `IntegerAttr.constr(i64)`.

As an aside, this PR has made me doubt the utility of `IntegerAttr` being generic.


alexarice:
Sure np

alexarice:
Using the new `Literal` stuff instead

PR #4817, opened 2025-07-17
dialects: (all) use SymbolNameConstraint for sym_name props/attrs

alexarice:
Actually use the new `SymbolNameConstraint` for things called `sym_name`. In all these cases there's no functional difference as they don't use assembly format, but I personally believes this conveys intent better.

I did not change `symref.declare` as this would actually cause functionality change. I'm not sure who this dialects "belongs" to, but would be good to get their input.

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -12,6 +12,7 @@ from xdsl.dialects.builtin import (
     Float64Type,
     FunctionType,
     StringAttr,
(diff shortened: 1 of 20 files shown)


superlopuh:
It belongs to the people, @EdmundGoodman and I are currently iterating on that bit of the code, we can probably clean it up as we go instead of in this PR

alexarice:
The better question would be "who uses `symref`" and would they mind/prefer if the symbol name parsed/printed as a symbol.

PR #4819, opened 2025-07-17
dialects:(omp) fix things missed in #4719

dk949:
* changes type of (in_)reduction_byref from `dense` to `array`
* `BlockArgOpenMPOperation` too constraining, needs to allow at more block args than required, but disallow Fewer.


superlopuh (commented):
Feels like we should have MLIR interop tests for these?

PR #4820, opened 2025-07-17
core: (AnyOf) allow variable inference and use for arith.constant

alexarice:
With the new(ish) `AnyOf` mechanisms/restrictions, we can now infer variables from "any of" constraints when that variable appears in all branches. 

This can be used in arith.constant in place of `TypedAttributeConstraint` (which I intend to deprecate)

If wanted, I can split this into two PRs:
- Change AnyOf (finding a suitable test to add)
- Change arith.constant

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -31,9 +31,8 @@ from xdsl.ir import Attribute, BitEnumAttribute, Dialect, Operation, SSAValue
 from xdsl.irdl import (
     AnyAttr,
     AnyOf,
-    BaseAttr,
     IRDLOperation,
-    TypedAttributeConstraint,
+    ParamAttrConstraint,
     VarConstraint,
     base,
     irdl_attr_definition,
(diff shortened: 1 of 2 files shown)


xdsl/dialects/arith.py
```diff
@@ -133,13 +132,12 @@ class ConstantOp(IRDLOperation):
...
     result = result_def(_T)
     value = prop_def(
-        TypedAttributeConstraint(
-            IntegerAttr.constr(SignlessIntegerConstraint | IndexTypeConstr)
-            | BaseAttr(FloatAttr)
-            | BaseAttr(DenseIntOrFPElementsAttr)
-            | BaseAttr(DenseResourceAttr),
-            _T,
+        ParamAttrConstraint(
+            IntegerAttr, (AnyAttr(), _T & (SignlessIntegerConstraint | IndexTypeConstr))
```
alexarice, line 136: Here is the issue I was discussing at the meeting. I would like to write:
```python
IntegerAttr.constr(_T & (SignlessIntegerConstraint | IndexTypeConstr))
```
but `_T & ...` has type `AttrConstraint` rather than anything more specific so cannot be passed into the `.constr` method, which expects an `AttrConstraint[_IntegerTypeT]`

xdsl/irdl/constraints.py
```diff
@@ -519,10 +519,12 @@ def __or__(
         return AnyOf((*self.attr_constrs, value))
 
     def variables(self) -> set[str]:
-        if len(self.attr_constrs) == 1:
-            return self.attr_constrs[0].variables()
-        else:
+        if len(self.attr_constrs) == 0:
```
superlopuh:
```suggestion
        if not self.attr_constrs:
```

xdsl/dialects/arith.py
alexarice: This could maybe be solved with a bunch of overloads on `IntegerAttr.constr` but it's not clear that it's worth it to me when we can just do the above for this one function.

superlopuh (approved):

PR #4821, opened 2025-07-17
misc: don't construct a new DenseArrayBase.constr(i32)

superlopuh:
Seems wasteful, we should probably do a full rewrite of this part of the code to be more performant.

### DIFF
--- a/xdsl/irdl/operations.py
+++ b/xdsl/irdl/operations.py
@@ -1350,7 +1350,7 @@ def get_variadic_sizes_from_attr(
             f"Expected {size_attribute_name} {container_name} in {op.name} operation."
         )
     attribute = container[size_attribute_name]
-    if not DenseArrayBase.constr(i32).verifies(attribute):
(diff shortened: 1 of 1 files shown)


alexarice:
I was too slow

superlopuh:
I would say that we should probably not verify here at all, and I have a trace that shows that this is a significant proportion of time but it's 1.49GB...

math-fehr:
Sasha just showed me a trace where this is the main bottleneck.
Btw, `type(attr) is DenseArrayBase` might even be faster.
Why not verifying? What happens when it's wrong then?

superlopuh:
Screenshot 2025-07-17 at 16 35 30

superlopuh:
I would do it once per operation, not on every attribute read

alexarice:
Wait, does this happen on every attribute read?

superlopuh:
Yeah

superlopuh:
This is why we need to rewrite this part it's by far the slowest part of xDSL

PR #4827, opened 2025-07-18
API: add EqIntConstraint

superlopuh:
I'm still not sure which way we'll go on the int constraint front, but if we get rid of IntConstraint then we can replace this trivially, and if we don't then it'll be useful.

### DIFF
--- a/tests/irdl/test_int_constraint.py
+++ b/tests/irdl/test_int_constraint.py
@@ -2,7 +2,7 @@ import re
 
 import pytest
 
-from xdsl.irdl import AtLeast, ConstraintContext
+from xdsl.irdl import AtLeast, ConstraintContext, EqIntConstraint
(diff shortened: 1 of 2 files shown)


tests/irdl/test_int_constraint.py
```diff
@@ -28,3 +28,12 @@ def test_at_least():
         AtLeast(2).verify(1, ConstraintContext())
     with pytest.raises(VerifyException):
         AtLeast(2).verify(0, ConstraintContext())
+
+
+def test_eq():
+    one_constr = EqIntConstraint(1)
+    one_constr.verify(1, ConstraintContext())
+    with pytest.raises(VerifyException):
```
alexarice: match statement?

alexarice (approved):

PR #4829, opened 2025-07-18
[RFC] core: (assembly-format) custom directives

alexarice:
Prototype implementation of custom assembly format directives.
Feel I've been procrastinating a while with trying to find a "perfect" way to do this, when realistically we would benefit a lot from a half decent way of doing this.

Thought I'd see what people think of this sort of interface. Will need a few more tests for multiple parameters/all the errors.

### DIFF
--- a/tests/irdl/test_declarative_assembly_format.py
+++ b/tests/irdl/test_declarative_assembly_format.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import textwrap
 from collections.abc import Callable
+from dataclasses import dataclass
 from io import StringIO
 from typing import Annotated, ClassVar, Generic
 
@@ -74,11 +75,15 @@ from xdsl.irdl import (
 )
 from xdsl.irdl.declarative_assembly_format import (
     AttrDictDirective,
+    FormatDirective,
     FormatProgram,
     OperandsDirective,
+    ParsingState,
+    PrintingState,
     PunctuationDirective,
     ResultsDirective,
     TypeDirective,
+    VariadicOperandVariable,
 )
 from xdsl.parser import Parser
 from xdsl.printer import Printer
@@ -3614,3 +3619,97 @@ def test_qualified_attr():
     op = parser.parse_operation()
     assert isinstance(op, QualifiedAttrOp)
     assert op.attr == MyAttr(StringAttr("test"))
+
+
+################################################################################
+#                             Custom directives                                #
+################################################################################
+
+
+@irdl_op_definition
+class CustomDirectiveOp(IRDLOperation):
+    name = "test.custom"
+
+    class Hello(FormatDirective):
+        def parse(self, parser: Parser, state: ParsingState) -> bool:
+            parser.parse_keyword("hello")
+            return True
+
+        def print(
+            self, printer: Printer, state: PrintingState, op: IRDLOperation
+        ) -> None:
+            if state.should_emit_space or not state.last_was_punctuation:
+                printer.print_string(" ")
+            state.last_was_punctuation = False
+            state.should_emit_space = True
+            printer.print_string("hello")
+
+    assembly_format = "custom<Hello>() attr-dict"
+
+
+@pytest.mark.parametrize(
+    "program",
+    [
+        "test.custom hello",
+        "test.custom hello {attr = 1 : i32}",
+    ],
+)
+def test_custom_directive(program: str):
+    ctx = Context()
+    ctx.load_op(CustomDirectiveOp)
+    ctx.load_dialect(Test)
+    check_roundtrip(program, ctx)
+
+
+@irdl_op_definition
+class CustomDirectiveWithParamOp(IRDLOperation):
+    name = "test.custom_param"
+
+    ops = var_operand_def()
+
+    @dataclass(frozen=True)
+    class Bars(FormatDirective):
+        """We print the operands with bars between, because why not."""
+
+        var: VariadicOperandVariable
+
+        def parse(self, parser: Parser, state: ParsingState) -> bool:
+            first = parser.parse_optional_unresolved_operand()
+            if first is None:
+                operands = []
+            else:
+                operands = [first]
+                while parser.parse_optional_punctuation("|"):
+                    operands.append(parser.parse_unresolved_operand())
+            state.operands[self.var.index] = operands
+            return bool(operands)
+
+        def print(
+            self, printer: Printer, state: PrintingState, op: IRDLOperation
+        ) -> None:
+            operand = getattr(op, self.var.name)
+            if not operand:
+                return
+            if state.should_emit_space or not state.last_was_punctuation:
+                printer.print_string(" ")
+            printer.print_list(operand, printer.print_ssa_value, delimiter=" | ")
+            state.last_was_punctuation = False
+            state.should_emit_space = True
+
+    assembly_format = "custom<Bars>($ops) (`:` type($ops)^)? attr-dict"
+
+
+@pytest.mark.parametrize(
+    "program",
+    [
+        "test.custom_param",
+        "test.custom_param %0 : i32",
+        "test.custom_param %0 | %1 : i32, i32",
+        "test.custom_param %0 | %1 | %2 : i32, i32, i32",
+    ],
+)
+def test_custom_directive_param(program: str):
+    ctx = Context()
+    ctx.load_op(CustomDirectiveWithParamOp)
+    ctx.load_dialect(Test)
+    check_roundtrip(program, ctx)
--- a/xdsl/irdl/declarative_assembly_format_parser.py
+++ b/xdsl/irdl/declarative_assembly_format_parser.py
@@ -5,6 +5,7 @@ https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-for
 
 from __future__ import annotations
 
+import inspect
 import re
 from collections.abc import Set as AbstractSet
 from dataclasses import dataclass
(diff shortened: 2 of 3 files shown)


tests/irdl/test_declarative_assembly_format.py
```diff
@@ -3614,3 +3619,97 @@ def test_qualified_attr():
...
+
+        def parse(self, parser: Parser, state: ParsingState) -> bool:
+            first = parser.parse_optional_unresolved_operand()
+            if first is None:
+                operands = []
+            else:
+                operands = [first]
+                while parser.parse_optional_punctuation("|"):
+                    operands.append(parser.parse_unresolved_operand())
+            state.operands[self.var.index] = operands
```
alexarice: I think it would be nice for directive to implement getters and setters for things like this. I propose to do that in a separate PR

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -3614,3 +3619,97 @@ def test_qualified_attr():
...
+            parser.parse_keyword("hello")
+            return True
+
+        def print(
+            self, printer: Printer, state: PrintingState, op: IRDLOperation
+        ) -> None:
+            if state.should_emit_space or not state.last_was_punctuation:
+                printer.print_string(" ")
+            state.last_was_punctuation = False
+            state.should_emit_space = True
```
alexarice: Realistically it's probably not necessary to set these most of the time, as we know where the format is being used.

xdsl/irdl/operations.py
```diff
@@ -914,6 +914,7 @@ class OpDef:
     or is already used by the operation, so we need to use a different name.
     """
     assembly_format: str | None = field(default=None)
+    custom_directives: dict[str, Any] = field(default_factory=dict[str, Any])
```
alexarice: My old friend import loops stopping this from being `dict[str, FormatDirective]`

superlopuh: Have you tried importing if `TYPE_CHECKING`?

superlopuh: and using `lambda: {}`?

tests/irdl/test_declarative_assembly_format.py
```diff
@@ -3614,3 +3619,97 @@ def test_qualified_attr():
...
+################################################################################
+#                             Custom directives                                #
+################################################################################
+
+
+@irdl_op_definition
+class CustomDirectiveOp(IRDLOperation):
+    name = "test.custom"
+
+    class Hello(FormatDirective):
```
superlopuh: Feels like we want a mechanism for sharing custom format directives between operations

xdsl/irdl/operations.py
alexarice: I tried a few things, one issue is that you can't even import `FormatDirective` from within `from_pyrdl` to check that the class is actually a format directive.

alexarice: you're welcome to have a play around and see if you can figure it out.

tests/irdl/test_declarative_assembly_format.py
alexarice: MLIR doesn't allow this (not that this means we shouldn't). I guess the alternative here is to have a `custom_directives` field where you can pass in arbitrary directives?

alexarice: You could also subclass some top-level directive to use it in multiple places

superlopuh: Since we're parsing the directives with inspect, maybe we can also do a `@directive_definition` like for operations and attributes? And then add a `custom_directives` field with those? Or maybe a trait?

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -705,6 +706,39 @@ def parse_typeable_directive(self) -> TypeableDirective:
...
+        """
+
+        with self.in_angle_brackets():
+            name = self.parse_identifier()
+            if name not in self.op_def.custom_directives:
+                self.raise_error(f"Custom directive {name} cannot be found.")
+        directive = self.op_def.custom_directives[name]
+        if not issubclass(directive, FormatDirective):
+            self.raise_error(f"{name} is not a format directive.")
+        param_types = inspect.get_annotations(directive, eval_str=True)
```
superlopuh: it would be great to do this once per directive definition and not once per op definition

xdsl/irdl/operations.py
superlopuh: I had a play around and bumped into some issues with inspecting and initialising objects, this feels like more motivation for me to do a small refactor here to reduce a bit of dynamism

tests/irdl/test_declarative_assembly_format.py
alexarice: I'm not sure what this would entail exactly.

superlopuh: I see at least two things, one is a trait that takes a number of custom format directive subclasses, and makes them discoverable during the assembly format parsing. Another is to move logic for inspecting the class to the class itself, somehow. I think it could be done with an annotation or with a superclass, that checks the fields at the point of definition and not when parsing the format directive.

alexarice:
@superlopuh it feels like a big question to me is whether we expect/want to share custom directives between different ops. I think I made this PR on the assumption that we wouldn't, but if we do then I agree that a reasonable way forward could be:
- Have a decorator for custom directives that does the inspection
- Have a different mechanism than inner classes for including a custom directive. Ideas seem to be:
  - directives field
  - trait
  - could even be an argument to `irdl_op_defintion`

superlopuh:
The main driver for custom directives that I can see is our backend dialects where it would make sense to have shared directives between all the different dialects for RISCV/ARM/x86/whatever.

alexarice:
This makes sense. I'll have a think about how to refactor this.

superlopuh:
Oh actually a great motivating use-case is `DynamicIndexList` which is also used a lot in MLIR.

tests/irdl/test_declarative_assembly_format.py
math-fehr: > MLIR doesn't allow this (not that this means we shouldn't). I guess the alternative here is to have a custom_directives field where you can pass in arbitrary directives?

You can have custom directives that are reused between operations no?

alexarice: Seems I was mistaken about this. I'll rework this PR a bit to not use inner classes/allow sharing

xdsl/irdl/operations.py
```diff
@@ -914,6 +915,9 @@ class OpDef:
     or is already used by the operation, so we need to use a different name.
     """
     assembly_format: str | None = field(default=None)
+    custom_directives: dict[str, type[FormatDirective]] = field(
+        default_factory=dict[str, Any]
```
alexarice:
```suggestion
        default_factory=dict[str, type[FormatDirective]]
```

xdsl/irdl/operations.py
```diff
@@ -1018,6 +1022,27 @@ def wrong_field_exception(field_name: str) -> PyRDLOpDefinitionError:
...
+                    field_names.add("directives")
+                    from xdsl.irdl.declarative_assembly_format import FormatDirective
+
+                    if not isinstance(value, tuple):
+                        raise PyRDLOpDefinitionError(
+                            f"pyrdl operation definition '{pyrdl_def.__name__}' "
+                            "directives field should be a tuple."
+                        )
+                    custom_directives: dict[str, type[FormatDirective]] = {}
+                    for directive in value:  # pyright: ignore[reportUnknownVariableType]
```
alexarice: Is there a nicer way to do this? I considered adding a case for `type[...]` to `isa`

alexarice:
Moved to having a `directives` field

xdsl/irdl/operations.py
alexarice: Seems to be fixed now

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: Do you feel strongly about this? Is the motivation just performance because I'm not actually sure how slow `get_annotations` is?

superlopuh: It evals once per field, which was quite visible on old traces of attribute definition, but I guess we can do this part later. The non-performance argument for it is that I feel like it's nice when the error occurs close in code to the thing that's broken, here if something goes wrong it's not at the point of definition, but at the point of use.

xdsl/irdl/operations.py
superlopuh:
```suggestion
        default_factory=lambda: {}
```

xdsl/irdl/operations.py
superlopuh: Could just do it in a different bit of logic? It would make sense to me to make this a ClassVar that's empty by default, and to just use normal Python to enumerate over it instead of being leet with introspection

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: The alternative here I guess is to have:
- Add a `CustomDirective` subclass with a classvar of `dict[str, type[FormatDirective]`
- a decorator `@irdl_custom_directive` to do the inspection which sets this.
- Use this instead of inspect in the above function.
?
I agree with the point about the better error handling. I think it should never happen that we have an unused custom directive, but it still seems nice that the directive itself would error instead of the operation that uses it. If the above sounds reasonable I can implement it.

xdsl/irdl/operations.py
alexarice: clever

xdsl/irdl/declarative_assembly_format_parser.py
superlopuh: I think `@irdl_custom_directive` would be great, not sure what you mean with that classvar exactly but it sounds about right. It'll also make it easier to test I think.

superlopuh:
Would it make sense to try to implement DynamicIndexList as a way to proving out the design/seeing what the usage would look like in practice? https://mlir.llvm.org/docs/Dialects/Vector/#vectorextract-vectorextractop Not sure if it should be in the same PR but might be useful to see anyway.

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: I mean that the use of `inspect` above is getting the parameters of the custom directive. I'm proposing to store the types of these parameters in a ClassVar of the CustomDirective instead.

alexarice:
> Would it make sense to try to implement DynamicIndexList as a way to proving out the design/seeing what the usage would look like in practice? https://mlir.llvm.org/docs/Dialects/Vector/#vectorextract-vectorextractop Not sure if it should be in the same PR but might be useful to see anyway.

Absolutely, I was just thinking that once we've got a candidate design it would be good to try this to see what it looks like/if it works in practice. I think stacking on top of this PR would be good.

xdsl/irdl/operations.py
```diff
@@ -870,7 +872,7 @@ def lazy_traits_def(future_traits: Callable[[], tuple[OpTrait, ...]]):
 _OPERATION_DICT_KEYS = {key for cls in Operation.mro()[:-1] for key in cls.__dict__}
 
 
-def _is_const_classvar(field_name: str, annotation: Any) -> bool:
+def is_const_classvar(field_name: str, annotation: Any) -> bool:
```
superlopuh: utils? also we should probably reuse for attributes if we're making this one public

alexarice: Sure, should I make a new PR?

superlopuh (approved):
Very nice

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -693,6 +693,33 @@ def parse_typeable_directive(self) -> TypeableDirective:
...
+
+        with self.in_angle_brackets():
+            name = self.parse_identifier()
+            if name not in self.op_def.custom_directives:
+                self.raise_error(f"Custom directive {name} cannot be found.")
+        directive = self.op_def.custom_directives[name]
+        param_types = directive.parameters
+        params = list[FormatDirective]()
+        with self.in_parens():
+            self.parse_comma_separated_list
```
math-fehr:
```suggestion
```
👀

math-fehr (approved):
Very nice!

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: Interesting

PR #4830, opened 2025-07-18
API: add IntConstraint.mapping_type_vars

superlopuh:
Useful whether or not we merge IntConstraint with AttrConstraint


alexarice (commented):
Should probably have tests but I guess these can't be added without a `IntTypeVarConstraint`? Would it be worth adding these together to allow for tests?

superlopuh:
I'll add a test

alexarice:
This will need updating after #4832 is merged I imagine.

superlopuh:
Probably, although I want to first finish the I32 port then try to figure out the most elegant end result, I'll mark is a draft for now.

PR #4831, opened 2025-07-18
core: tighten type bounds on constraint & operator

superlopuh:


xdsl/irdl/constraints.py
```diff
@@ -165,12 +165,10 @@ def __or__(
             return value  # pyright: ignore[reportReturnType]
         return AnyOf((self, value))
 
-    def __and__(
-        self, value: AttrConstraint[AttributeCovT], /
-    ) -> AttrConstraint[AttributeCovT]:
+    def __and__(self, value: AttrConstraint, /) -> AttrConstraint[AttributeCovT]:
```
superlopuh, line 168: In the absense of an intersection type, I propose just taking the first type to be the bound, as opposed to all of the constraints needing to be of the same type

alexarice (approved):
This is probably reasonable. I think I considered it but didn't like that `&` wasn't symmetric but it's probably fine. Can you clean up `arith.constant` with this a bit?

PR #4832, opened 2025-07-18
core: split out RangeOf (with length) and AnyRangeOf (without length)

superlopuh:
This will unblock the IntConstraint removal, at the expense of a small breaking change.


alexarice (commented):
Is there a secondary benefit of this over solving an import loop? Otherwise moving `IntAttr` to a different file seems kind of simpler to me :shrug:

superlopuh:
I disagree on the simplicity side of things, I think the IntAttr move is very surprising and should be avoided if possible.

alexarice:
I'm not sure I understand, the `IntAttr` change was like 30 lines with no external API change

superlopuh:
I'm happy to make this more palateable. I think it's quite a big disruption to have something more core than builtin, and if we have a more ergonomic way of avoiding it it would be preferred.

tests/irdl/test_range_constraint.py
```diff
@@ -85,6 +87,9 @@ def test_verify_correct_length():
     # variables
     assert range_len_constr.variables() == {"ATTR", "LENGTH"}
 
+    # variables_from_length
+    assert range_len_constr.variables_from_length() == {"LENGTH"}
```
superlopuh, line 91: I forgot to test this and it was broken

alexarice:
This should still work right? The outer constraint sets length known to true then the inner constraint can infer?

superlopuh:
Yep we have a test for this, which had to be fixed with the `variables_from_length`

xdsl/dialects/builtin.py
```diff
@@ -58,6 +58,7 @@
     ParamAttrConstraint,
     RangeConstraint,
     RangeOf,
+    # RangeOf,
```
alexarice:
```suggestion
```

xdsl/dialects/builtin.py
```diff
@@ -206,7 +207,9 @@ def __init__(
             object.__setattr__(self, "elem_range_constraint", constr)
         else:
             object.__setattr__(
-                self, "elem_range_constraint", RangeOf(irdl_to_attr_constraint(constr))
+                self,
+                "elem_range_constraint",
+                RangeOf(irdl_to_attr_constraint(constr)),
```
alexarice: why is there a diff here, seems like nothing has changed.

xdsl/dialects/omp.py
```diff
@@ -697,8 +697,7 @@ class TargetOp(BlockArgOpenMPOperation):
     depend_vars = var_operand_def(
         RangeOf(
             AnyAttr(),  # TODO: OpenMP_PointerLikeTypeInterface
-            length=DEP_COUNT,
-        )
+        ).of_length(DEP_COUNT)
```
alexarice, line 700: This is a sort of case where we could imagine having a `length` keyword arg on a `var_operand_def`, but I imagine this would cause import loops at the moment.

xdsl/dialects/builtin.py
superlopuh: I was wondering about that, I must have touched it in the refactor somehow, I'll revert

PR #4834, opened 2025-07-19
core: speed up parsing of large constant tensors

ice-tong:
# Motivation
- When I use xDSL to parse an MLIR module with large constant tensor, it takes 1 minute to parse a 500 MB MLIR file. I profiled it and found that the 'StringLiteral.bytes_contents' took 80% of the time to parse the long hex string. We can speed this up by returning early if there are no escape sequences.

### DIFF
--- a/tests/test_pass_from_spec.py
+++ b/tests/test_pass_from_spec.py
@@ -28,6 +28,8 @@ class CustomPass(ModulePass):
 
     optional_bool: bool = False
 
+    annotation: "str" = "default_value"
+
     def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
         pass
 
@@ -61,6 +63,7 @@ def test_pass_instantiation():
                 "str-thing": ("hello world",),
                 "literal": ("maybe",),
                 # "optional" was left out here, as it is optional
+                "annotation": ("other_value",),
             },
         )
     )
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -12,6 +12,7 @@ from typing import (
     Union,
     get_args,
     get_origin,
+    get_type_hints,
 )
 
 from typing_extensions import Self, TypeVar
@@ -88,6 +89,8 @@ class ModulePass(ABC):
 
         required_fields = cls.required_fields()
 
+        field_types = get_type_hints(cls)
+
         # iterate over all fields of the dataclass
         for op_field in fields:
             # ignore the name field and everything that's not used by __init__
@@ -101,9 +104,10 @@ class ModulePass(ABC):
                 raise ValueError(f'Pass {cls.name} requires argument "{op_field.name}"')
 
             # convert pass arg to the correct type:
+            field_type = field_types[op_field.name]
             arg_dict[op_field.name] = _convert_pass_arg_to_type(
                 spec_arguments_dict.pop(op_field.name),
-                op_field.type,
+                field_type,
             )
             # we use .pop here to also remove the arg from the dict
 


superlopuh (commented):
Nice! It would be great to add a performance test for this.

@EdmundGoodman, could you please suggest the best way to track this in our benchmarking setup?

xdsl/utils/mlir_lexer.py
```diff
@@ -64,6 +64,10 @@ def string_contents(self):
 
     @property
     def bytes_contents(self) -> bytes:
+        # NOTE: Early return if no escape sequences are present.
+        if "\\" not in self.text[1:-1]:
+            return self.text[1:-1].encode()
+
```
alexarice, line 73: Would it make sense to do something like split the string on escape characters instead of special casing?

superlopuh: Probably, so scanning for the next index of `\` and inspecting the next char, instead of iterating over each character? That seems like a better approach anyway.

alexarice: `"\\" == r"\"` right? so should just be collect up to the next `\` and encode that, rather than doing 1 character at a time (which I assume but don't know is the slow bit.

EdmundGoodman:
> Nice! It would be great to add a performance test for this.
> 
> @EdmundGoodman, could you please suggest the best way to track this in our benchmarking setup?

We currently have some tests in `benchmarks/lexer.py` that exercise similar things, but are bottlenecked by consuming the regex rather than getting the byte contents (e.g. `uv run benchmarks/lexer.py Lexer.dense_attr_hex viztracer`), so don't measure this particular change. I think the best place to add this would be adjacent to them in that same file.

There are some instructions on how to do this in the benchmarks README (https://github.com/xdslproject/xdsl/blob/main/benchmarks/README.md#adding-new-benchmarks). As this is a new workload, we'll probably want to add it to the `benchmarks/workload.py` workload builder (ideally in a way that can be procedurally generated like the others), and then add a new benchmark for it following the above instructions.

@ice-tong if you have a link to what your big MLIR file looks like I am happy to help out adding this into our benchmarking suite

ice-tong:
Hi @superlopuh @alexarice  @EdmundGoodman , thanks for your quick feedback on this PR!

> @ice-tong if you have a link to what your big MLIR file looks like I am happy to help out adding this into our benchmarking suite

I'm sorry I can't share the MLIR I actually use, but I've created a script to test the speed increase introduced by this PR change. Run the following script to see the 5x speed-up.

```python
from typing import Tuple
import time
from contextlib import contextmanager

import numpy as np

from xdsl.context import Context
import xdsl.dialects.builtin as builtin_dialect
import xdsl.dialects.arith as arith_dialect
import xdsl.dialects.func as func_dialect
from xdsl.utils.mlir_lexer import StringLiteral

TEST_SHAPE = (1, 3, 1080, 1080)

CTX = Context()
CTX.load_dialect(builtin_dialect.Builtin)
CTX.load_dialect(arith_dialect.Arith)
CTX.load_dialect(func_dialect.Func)


def create_large_constant_tensor_case(
    tensor_shape: Tuple[int, ...],
    num_add_ops: int = 10,
) -> "builtin_dialect.ModuleOp":
    """Create a module with a function that adds multiple large constant tensors.
    
    Args:
        tensor_shape (Tuple[int, ...]): The shape of the constant tensors to be created.
        num_add_ops (int): The number of addition operations to perform with the constant tensors.

    Returns:
        ModuleOp: The created module containing the function with addition operations.
    """
    tensor_type = builtin_dialect.TensorType(
        shape=TEST_SHAPE, element_type=builtin_dialect.Float32Type()
    )
    function_type = builtin_dialect.FunctionType.from_lists(
        inputs=[tensor_type], outputs=[tensor_type]
    )

    func_op = func_dialect.FuncOp(name="main", function_type=function_type)

    input_val = func_op.args[0]
    output_val = input_val

    for _ in range(num_add_ops):
        # Create a constant tensor with random values
        const_array = np.random.random(tensor_shape).astype(np.float32)
        dense_elements_attr = builtin_dialect.DenseIntOrFPElementsAttr.from_list(
            type=tensor_type,
            data=const_array.flatten().tolist()
        )
        const_op = arith_dialect.ConstantOp(value=dense_elements_attr, value_type=tensor_type)
        func_op.body.block.add_op(const_op)

        # Add the constant to the input value
        add_op = arith_dialect.AddfOp(operand1=input_val, operand2=const_op, result_type=tensor_type)
        func_op.body.block.add_op(add_op)

        # Update input_val to the result of the addition
        input_val = add_op.result
        output_val = add_op.result

    return_op = func_dialect.ReturnOp(output_val)
    func_op.body.block.add_op(return_op)

    module_op = builtin_dialect.ModuleOp([func_op])
    module_op.verify()
    return module_op


@contextmanager
def _hack_bytes_contents():
    ori_bytes_contents = StringLiteral.bytes_contents

    def _hack_bytes_contents(self: "StringLiteral") -> bytes:
        # NOTE: Early return if no escape sequences are present.
        if "\\" not in self.text[1:-1]:
            return self.text[1:-1].encode()
        return ori_bytes_contents.fget(self)

    try:
        StringLiteral.bytes_contents = property(_hack_bytes_contents)
        yield
    finally:
        StringLiteral.bytes_contents = ori_bytes_contents


def main():
    """The main entry point for the script."""
    start_t = time.perf_counter()
    test_module_op = create_large_constant_tensor_case(TEST_SHAPE, num_add_ops=10)
    end_t = time.perf_counter()
    print(f"Module creation cost time: {end_t - start_t:.4f} seconds")

    parser = Parser(CTX, str(test_module_op))
    start_t = time.perf_counter()
    parser.parse_module()
    end_t = time.perf_counter()
    print(f"Parsing the created module back cost time: {end_t - start_t:.4f} seconds")
    
    with _hack_bytes_contents():
        parser = Parser(CTX, str(test_module_op))
        start_t = time.perf_counter()
        parser.parse_module()
        end_t = time.perf_counter()
        print(f"Parsing with bytes content patch cost time: {end_t - start_t:.4f} seconds")
            

if __name__ == "__main__":
    main()
```

Run this script, we got:
```
Module creation cost time: 1.2368 seconds
Parsing the created module back cost time: 11.8795 seconds
Parsing with bytes content patch cost time: 2.1197 seconds
```

EdmundGoodman:
> Hi @superlopuh @alexarice @EdmundGoodman , thanks for your quick feedback on this PR!
> 
> > @ice-tong if you have a link to what your big MLIR file looks like I am happy to help out adding this into our benchmarking suite
> 
> I'm sorry I can't share the MLIR I actually use, but I've created a script to test the speed increase introduced by this PR change. Run the following script to see the 5x speed-up.
> 
> ```python
> from typing import Tuple
> import time
> from contextlib import contextmanager
> 
> import numpy as np
> 
> from xdsl.parser import Parser
> from xdsl.context import Context
> import xdsl.dialects.builtin as builtin_dialect
> import xdsl.dialects.arith as arith_dialect
> import xdsl.dialects.func as func_dialect
> from xdsl.utils.mlir_lexer import StringLiteral
> 
> TEST_SHAPE = (1, 3, 1080, 1080)
> 
> CTX = Context()
> CTX.load_dialect(builtin_dialect.Builtin)
> CTX.load_dialect(arith_dialect.Arith)
> CTX.load_dialect(func_dialect.Func)
> 
> 
> def create_large_constant_tensor_case(
>     tensor_shape: Tuple[int, ...],
>     num_add_ops: int = 10,
> ) -> "builtin_dialect.ModuleOp":
>     """Create a module with a function that adds multiple large constant tensors.
>     
>     Args:
>         tensor_shape (Tuple[int, ...]): The shape of the constant tensors to be created.
>         num_add_ops (int): The number of addition operations to perform with the constant tensors.
> 
>     Returns:
>         ModuleOp: The created module containing the function with addition operations.
>     """
>     tensor_type = builtin_dialect.TensorType(
>         shape=TEST_SHAPE, element_type=builtin_dialect.Float32Type()
>     )
>     function_type = builtin_dialect.FunctionType.from_lists(
>         inputs=[tensor_type], outputs=[tensor_type]
>     )
> 
>     func_op = func_dialect.FuncOp(name="main", function_type=function_type)
> 
>     input_val = func_op.args[0]
>     output_val = input_val
> 
>     for _ in range(num_add_ops):
>         # Create a constant tensor with random values
>         const_array = np.random.random(tensor_shape).astype(np.float32)
>         dense_elements_attr = builtin_dialect.DenseIntOrFPElementsAttr.from_list(
>             type=tensor_type,
>             data=const_array.flatten().tolist()
>         )
>         const_op = arith_dialect.ConstantOp(value=dense_elements_attr, value_type=tensor_type)
>         func_op.body.block.add_op(const_op)
> 
>         # Add the constant to the input value
>         add_op = arith_dialect.AddfOp(operand1=input_val, operand2=const_op, result_type=tensor_type)
>         func_op.body.block.add_op(add_op)
> 
>         # Update input_val to the result of the addition
>         input_val = add_op.result
>         output_val = add_op.result
> 
>     return_op = func_dialect.ReturnOp(output_val)
>     func_op.body.block.add_op(return_op)
> 
>     module_op = builtin_dialect.ModuleOp([func_op])
>     module_op.verify()
>     return module_op
> 
> 
> @contextmanager
> def _hack_bytes_contents():
>     ori_bytes_contents = StringLiteral.bytes_contents
> 
>     def _hack_bytes_contents(self: "StringLiteral") -> bytes:
>         # NOTE: Early return if no escape sequences are present.
>         if "\\" not in self.text[1:-1]:
>             return self.text[1:-1].encode()
>         return ori_bytes_contents.fget(self)
> 
>     try:
>         StringLiteral.bytes_contents = property(_hack_bytes_contents)
>         yield
>     finally:
>         StringLiteral.bytes_contents = ori_bytes_contents
> 
> 
> def main():
>     """The main entry point for the script."""
>     start_t = time.perf_counter()
>     test_module_op = create_large_constant_tensor_case(TEST_SHAPE, num_add_ops=10)
>     end_t = time.perf_counter()
>     print(f"Module creation cost time: {end_t - start_t:.4f} seconds")
> 
>     parser = Parser(CTX, str(test_module_op))
>     start_t = time.perf_counter()
>     parser.parse_module()
>     end_t = time.perf_counter()
>     print(f"Parsing the created module back cost time: {end_t - start_t:.4f} seconds")
>     
>     with _hack_bytes_contents():
>         parser = Parser(CTX, str(test_module_op))
>         start_t = time.perf_counter()
>         parser.parse_module()
>         end_t = time.perf_counter()
>         print(f"Parsing with bytes content patch cost time: {end_t - start_t:.4f} seconds")
>             
> 
> if __name__ == "__main__":
>     main()
> ```
> 
> Run this script, we got:
> 
> ```
> Module creation cost time: 1.2368 seconds
> Parsing the created module back cost time: 11.8795 seconds
> Parsing with bytes content patch cost time: 2.1197 seconds
> ```

Amazing, thank you so much, that's really helpful. I've made a new PR adapting this into the style of our benchmarking suite #4840 , so we should have better visibility on the perf impact of the change

EdmundGoodman:
New benchmark on main reflects these new changes!

Before:
```
$ uv run python3 benchmarks/parser.py Parser.large_constant_tensor run
Test Parser.large_constant_tensor ran in: 10.06508s
```

After:
```
$ uv run python3 benchmarks/parser.py Parser.large_constant_tensor run
Test Parser.large_constant_tensor ran in: 1.75765s
```

alexarice (approved):
Very cool, I think we could go ahead and merge this, and consider more general solutions in a future PR (which should let us get a good comparison)

superlopuh:
I disagree, I think it would be great to try the approach @alexarice proposed here to avoid the special case, and to speed up processing when there are escaped characters.

xdsl/utils/mlir_lexer.py
superlopuh: @ice-tong, could you please try this approach? It feels like scanning for next index of `\` should speed up both the case with and without escape characters, with less special-case logic.

ice-tong: @alexarice @superlopuh That sounds good! I will try this approach these days.

ice-tong:
Hi @superlopuh, @alexarice, @EdmundGoodman,

Apologies for the delay on this PR. I had some time today to implement the optimization @alexarice suggested earlier.

Initially, I tried the simplest approach—scanning directly until the next `\\` appeared before encoding the characters. However, this didn’t yield any performance improvement, and I couldn’t pinpoint the reason (see commit: https://github.com/xdslproject/xdsl/pull/4834/commits/4020155ed263ab94bf1ff706844a1d622d6cc16c).

I then switched to using regex to match escape characters, which successfully achieved the expected acceleration (see commit: https://github.com/xdslproject/xdsl/pull/4834/commits/c51d10e62c25268aeb66fe245bd138c439c28be2).

Before:
```
$ uv run python3 benchmarks/parser.py Parser.large_constant_tensor run
Test Parser.large_constant_tensor ran in: 0.56134s
```

After:
```
$ uv run python3 benchmarks/parser.py Parser.large_constant_tensor run
Test Parser.large_constant_tensor ran in: 0.07453s
```

Please take a look at this PR when you have time, and let me know your thoughts or suggestions.

Thank you!

superlopuh:
Thank you for iterating on this! No worries at all about the delay. I'm hoping to avoid regular expressions here, could you please try adapting this sketch to rely just on the string methods?

``` python
text = r"hello\nworld!"
index = 0
end = len(text)
res = ""
try:
    while (next_backslash_index := text.index("\\", index)) < end:
        print(next_backslash_index, text[next_backslash_index+1])
        res += text[index:next_backslash_index]
        if text[next_backslash_index+1] == "n":
            res += "\n"
        else:
            assert False, next_backslash_index
        index = next_backslash_index + 2
except ValueError:
    res += text[index:]
print(res)
```

I dropped most of the logic we need in the function but this seems to work for this super simplified `\n` folding

ice-tong:
> Thank you for iterating on this! No worries at all about the delay. I'm hoping to avoid regular expressions here, could you please try adapting this sketch to rely just on the string methods?
> 
> ```python
> text = r"hello\nworld!"
> index = 0
> end = len(text)
> res = ""
> try:
>     while (next_backslash_index := text.index("\\", index)) < end:
>         print(next_backslash_index, text[next_backslash_index+1])
>         res += text[index:next_backslash_index]
>         if text[next_backslash_index+1] == "n":
>             res += "\n"
>         else:
>             assert False, next_backslash_index
>         index = next_backslash_index + 2
> except ValueError:
>     res += text[index:]
> print(res)
> ```
> 
> I dropped most of the logic we need in the function but this seems to work for this super simplified `\n` folding

@superlopuh Thank you for your excellent implementation! I’ve reworked the string method-based version inspired by your approach, and it’s working effectively. Please see commit [6d24cca](https://github.com/xdslproject/xdsl/pull/4834/commits/6d24cca3c5aa98142f1f3e6d134df9109fc6b926) for details.

Before:
```
$ uv run python3 benchmarks/parser.py Parser.large_constant_tensor run
Test Parser.large_constant_tensor ran in: 0.56377s
```

After:
```
$ uv run python3 benchmarks/parser.py Parser.large_constant_tensor run
Test Parser.large_constant_tensor ran in: 0.07198s
```

tests/utils/test_mlir_lexer.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from xdsl.utils.lexer import Input
+from xdsl.utils.mlir_lexer import StringLiteral
+
+
+def _create_string_literal(input_str: str) -> StringLiteral:
+    """Create a StringLiteral from a given input string."""
+    input_data = Input(input_str, "test.mlir")
+    return StringLiteral(0, len(input_str), input_data)
```
superlopuh: since this is used only in the test below, let's inline it? It'll simplify the `parametrize` call which I'm finding a bit hard to read

xdsl/utils/mlir_lexer.py
```diff
@@ -64,26 +64,58 @@ def string_contents(self):
 
     @property
     def bytes_contents(self) -> bytes:
+        # NOTE: This function is used to convert the string literal into a bytes object and
+        # handle escape sequences. There are two types of escape sequences we process:
+        # 1. Known escape sequences like \n, \t, \\, and \"
+        # 2. Hexadecimal escape sequences like \x00, \x0E, \00, \0E, etc.
```
superlopuh: Let's add a doc comment here:
```suggestion
        """
        Convert the string literal into a bytes object and handle escape sequences. 
        There are two types of escape sequences we process:
        1. Known escape sequences like \n, \t, \\, and \"
        2. Hexadecimal escape sequences like \x00, \x0E, \00, \0E, etc.
        """
```

xdsl/utils/mlir_lexer.py
```diff
@@ -64,26 +64,58 @@ def string_contents(self):
...
-                bytes_contents += c0.encode()
+
+        # Initialize the last index to track where the last escape sequence ended.
+        last_index = 0
+
+        # Scan through the string literal text to find escape sequences.
+        while (backslash_index := text_contents.find("\\", last_index)) > -1:
+            bytes_contents += text_contents[last_index:backslash_index].encode()
+
+            if len(text_contents) <= backslash_index + 1:
```
superlopuh: Let's factor out `len(text_contents)` to a local value to avoid calling the function in a loop?

xdsl/utils/mlir_lexer.py
```diff
@@ -64,26 +64,58 @@ def string_contents(self):
...
+                raise ParseError(self, "Incomplete escape sequence at end of string.")
```
superlopuh, line 96: can you please add a unit test for this?

xdsl/utils/mlir_lexer.py
```diff
@@ -64,26 +64,58 @@ def string_contents(self):
...
-                        bytes_contents += b"\t"
-                    case "\\":
-                        bytes_contents += b"\\"
-                    case '"':
-                        bytes_contents += b'"'
-                    case _:
-                        c1 = next(iter_string)
-                        bytes_contents += int(c0 + c1, 16).to_bytes(1, "big")
+                # If the escape sequence is not recognized, we raise an error.
+                raise ParseError(
```
superlopuh, line 116: Can you please add a unit test for this?

tests/utils/test_mlir_lexer.py
ice-tong: Done by https://github.com/xdslproject/xdsl/pull/4834/commits/dcf7f34756732fa8b42b66f37cb09cafe183bee8

xdsl/utils/mlir_lexer.py
ice-tong: Done by https://github.com/xdslproject/xdsl/pull/4834/commits/dcf7f34756732fa8b42b66f37cb09cafe183bee8

xdsl/utils/mlir_lexer.py
ice-tong: Done by https://github.com/xdslproject/xdsl/pull/4834/commits/dcf7f34756732fa8b42b66f37cb09cafe183bee8

xdsl/utils/mlir_lexer.py
ice-tong: Done by https://github.com/xdslproject/xdsl/pull/4834/commits/dcf7f34756732fa8b42b66f37cb09cafe183bee8

xdsl/utils/mlir_lexer.py
ice-tong: Done by https://github.com/xdslproject/xdsl/pull/4834/commits/dcf7f34756732fa8b42b66f37cb09cafe183bee8

tests/utils/test_mlir_lexer.py
```diff
@@ -0,0 +1,47 @@
...
+    """Test that the StringLiteral contains the expected bytes."""
+    string_literal = StringLiteral(
+        0, len(input_string), Input(input_string, "test.mlir")
+    )
+    assert string_literal.bytes_contents == expected_bytes, (
+        f"Expected {expected_bytes!r}, got {string_literal.bytes_contents!r}"
+    )
+    assert string_literal.string_contents == expected_bytes.decode("utf-8"), (
+        f"Expected {expected_bytes.decode('utf-8')!r}, got {string_literal.text!r}"
+    )
```
superlopuh, line 27: There's no need for assert messages in pytest it generates decent ones for you anyway IMO
```suggestion
    assert string_literal.bytes_contents == expected_bytes
    assert string_literal.string_contents == expected_bytes.decode("utf-8")
```

xdsl/utils/mlir_lexer.py
```diff
@@ -64,26 +64,63 @@ def string_contents(self):
...
+                    self,
+                    f"Invalid escape sequence: {text_contents[backslash_index : backslash_index + 2]}",
+                )
 
+        # Add the remaining part of the string after the last escape sequence.
+        if last_index < len(text_contents):
```
superlopuh:
```suggestion
        if last_index < text_contents_len:
```

superlopuh (approved):
Thank you very much! I'll merge as soon as the last suggestion is in

xdsl/utils/mlir_lexer.py
ice-tong: Done by https://github.com/xdslproject/xdsl/pull/4834/commits/d99c53667da84c56b65691d5b22513eb153e95be

ice-tong:
> Thank you very much! I'll merge as soon as the last suggestion is in

@superlopuh Thank you so much for taking the time to review my PR and providing such detailed and constructive feedback. Your suggestions were incredibly helpful and have significantly improved the quality of the code. I’ve learned a lot from this process, and I truly appreciate your guidance and support!

superlopuh:
Thank you for taking the time to make this improvement to xDSL!

PR #4836, opened 2025-07-19
core: fix RangeOf variables

superlopuh:
Small bug fix to propagate the variables properly.


alexarice (approved):
Strange that this got lost at some point

PR #4837, opened 2025-07-19
core: add RangeLengthConstraint

superlopuh:
Unlike `RangeOf`, which takes an attribute constraint and a length constraint, this constraint takes a range constraint and a length constraint.

### DIFF
--- a/tests/irdl/test_range_constraint.py
+++ b/tests/irdl/test_range_constraint.py
@@ -3,17 +3,21 @@ from collections.abc import Sequence
 import pytest
 from typing_extensions import TypeVar
 
+from xdsl.dialects.builtin import StringAttr
 from xdsl.ir import Attribute
 from xdsl.irdl import (
     AnyAttr,
     AnyInt,
     AttrConstraint,
+    BaseAttr,
     ConstraintContext,
     IntVarConstraint,
     RangeConstraint,
+    RangeLengthConstraint,
     RangeOf,
     VarConstraint,
 )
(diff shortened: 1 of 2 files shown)


tests/irdl/test_range_constraint.py
```diff
@@ -44,3 +48,58 @@ def test_range_of_variables():
...
+    assert range_len_constr.can_infer({"ATTR"}, length_known=True)
+    assert range_len_constr.can_infer({"ATTR", "LENGTH"}, length_known=False)
+
+    # infer
+    assert range_len_constr.infer(
+        ConstraintContext(_variables={"ATTR": world}), length=2
+    ) == (
+        world,
+        world,
+    )
```
superlopuh, line 101: One thing I'm not sure about is whether inference should verify the length here, feels like it would be a bit weird if we could infer a range of an unsupported length. This would be a change on the test, and the implementation of infer would have to call verify in the case where `length` is not `None`.
```suggestion
    ctx = ConstraintContext(_variables={"ATTR": world})
    assert range_len_constr.infer(ctx, length=2) == (
        world,
        world,
    )
    assert ctx.get_int_variable("LENGTH") == 2
```

alexarice: I don't see why this would be necessary, as far as I know, we don't require inference to verify things (unless required by pyright typechecking usually).

alexarice (approved):
This seems a lot nicer than `RangeOf` and `AnyRangeOf` to me.

PR #4838, opened 2025-07-19
dialects: (fir) FIR compatibility with Flang 20.7.1

mesham:
An additional fortran variable flag and optional properties in the global operation that are present in Flang 20.7.1 but were missing in our FIR dialect

### DIFF
--- a/xdsl/dialects/experimental/fir.py
+++ b/xdsl/dialects/experimental/fir.py
@@ -1744,7 +1744,7 @@ class GlobalOp(IRDLOperation):
     sym_name = prop_def(SymbolNameConstraint())
     symref = prop_def(SymbolRefAttr)
     type = prop_def(Attribute)
-    initVal = opt_prop_def(Attribute)
+    initVal = opt_prop_def()
     constant = opt_prop_def(UnitAttr)
     target = opt_prop_def(UnitAttr)
     linkName = opt_prop_def(StringAttr)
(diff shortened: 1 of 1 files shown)


xdsl/dialects/experimental/fir.py
```diff
@@ -1743,8 +1744,12 @@ class GlobalOp(IRDLOperation):
     sym_name = prop_def(SymbolNameConstraint())
     symref = prop_def(SymbolRefAttr)
     type = prop_def(Attribute)
-    linkName = opt_prop_def(StringAttr)
+    initVal = opt_prop_def(Attribute)
```
alexarice, line 1747:
```suggestion
    initVal = opt_prop_def()
```
This is the default

xdsl/dialects/experimental/fir.py
```diff
@@ -1743,8 +1744,12 @@ class GlobalOp(IRDLOperation):
...
     constant = opt_prop_def(UnitAttr)
+    target = opt_prop_def(UnitAttr)
+    linkName = opt_prop_def(StringAttr)
+    data_attr = opt_prop_def(Attribute)
```
alexarice, line 1751:
```suggestion
    data_attr = opt_prop_def()
```

xdsl/dialects/experimental/fir.py
mesham: Thanks :)

alexarice:
As discussed on zulip it seems the default argument for `opt_prop_def` is missing. I'll address this in a different PR but happy for this one to be merged as it originally was.

PR #4840, opened 2025-07-19
bench: incorporate benchmark proposed in #4834

EdmundGoodman:
#4834 discusses speeding up parsing by improving the way byte contents of strings are retrieved. This PR incorporates the benchmark proposed in the PR into the benchmarking suite for xDSL

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -71,7 +71,9 @@ uv run benchmarks/lexer.py Lexer.empty_program dis
         importlib.reload(xdsl.dialects.newdialect)
     ```
 
-2. Associate the method with a name in the main code at the bottom of the file
+2. If the test operates on an xDSL operation or other workload which needs to be set up, this is best added in `benchmarks/workloads.py` as a new class method on `WorkloadBuilder`.
+
+3. Associate the method with a name in the main code at the bottom of the file
    
    For example, to add this new method, you might write
 
--- a/benchmarks/workloads.py
+++ b/benchmarks/workloads.py
@@ -153,6 +153,8 @@ class WorkloadBuilder:
         Returns:
             The created module containing the function with addition operations.
         """
+        assert num_add_ops >= 0
+        random.seed(RANDOM_SEED)
         tensor_type = TensorType(shape=tensor_shape, element_type=Float32Type())
         function_type = FunctionType.from_lists(
             inputs=[tensor_type], outputs=[tensor_type]
--- a/benchmarks/workloads.py
+++ b/benchmarks/workloads.py
@@ -2,16 +2,20 @@
 """Workloads for benchmarking xDSL."""
 
 import random
+from math import prod
 
-from xdsl.dialects.arith import AddiOp, ConstantOp
+from xdsl.dialects.arith import AddfOp, AddiOp, ConstantOp
 from xdsl.dialects.builtin import (
     DenseIntOrFPElementsAttr,
+    Float32Type,
+    FunctionType,
     IntegerAttr,
     ModuleOp,
     TensorType,
     i8,
     i32,
 )
+from xdsl.dialects.func import FuncOp, ReturnOp
 from xdsl.dialects.test import TestOp
 from xdsl.ir import Operation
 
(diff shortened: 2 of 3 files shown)


benchmarks/README.md
```diff
@@ -71,7 +71,9 @@ uv run benchmarks/lexer.py Lexer.empty_program dis
         importlib.reload(xdsl.dialects.newdialect)
     ```
 
-2. Associate the method with a name in the main code at the bottom of the file
+2. If the test operates on an xDSL operation or other workload which needs to be set up, this is best added in `benchmarks/workloads.py` as a new class method on `WorkloadBuilder`.
+
+3. Associate the method with a name in the main code at the bottom of the file
```
alexarice: Maybe we should have a separate quick PR for the unrelated docstring changes?

benchmarks/workloads.py
```diff
@@ -133,3 +137,54 @@ def large_dense_attr_hex(cls, x: int = 1024, y: int = 1024) -> str:
...
+        func_op = FuncOp(name="main", function_type=function_type)
+
+        input_val = func_op.args[0]
+        output_val = input_val
+
+        for _ in range(num_add_ops):
+            # Create a constant tensor with random values
+            dense_elements_attr = DenseIntOrFPElementsAttr.from_list(
+                type=tensor_type,
+                data=[random.random() for _ in range(prod(tensor_shape))],
```
alexarice, line 172: having randomness in tests feels like a bad idea to me, if it really needs to "look" random, can we at least fix a seed?

EdmundGoodman: Yes, sorry -- other benchmarks do this, but here I forgot it in the copy-paste job across

benchmarks/README.md
EdmundGoodman: #4841

benchmarks/workloads.py
superlopuh: Does it need to look random? I don't think it adds anything, I feel like DEADBEEF a bunch of times would be fine

superlopuh: Or floats from 0 to size.

EdmundGoodman: It doesn't particularly need to be random -- this was just lifted from the example code. However, I don't really see any difference between a fixed set of random numbers generated from a seed and an arbitrary selection made during development, other than perhaps being slightly more representative as Python might be able to do strange optimisation trickery for allocating big homogeneous lists?

We could change it to this if you particularly want to?

```suggestion
                data=[float(x) for x in range(prod(tensor_shape))],
```

alexarice: Is the time taken to generate this list part of the benchmark time? (Basically does the time taken to generate random numbers matter?) If not I would be happy with the random numbers

EdmundGoodman: No -- the workload is generated as a member property of the benchmark class (`WORKLOAD_LARGE_CONSTANT_TENSOR`), and the benchmark only records parsing that pre-generated member property

superlopuh: I guess random is fine, let's see if it gets annoying

benchmarks/workloads.py
```diff
@@ -133,3 +137,56 @@ def large_dense_attr_hex(cls, x: int = 1024, y: int = 1024) -> str:
...
+            The created module containing the function with addition operations.
+        """
+        assert num_add_ops >= 0
+        random.seed(RANDOM_SEED)
+        tensor_type = TensorType(shape=tensor_shape, element_type=Float32Type())
+        function_type = FunctionType.from_lists(
+            inputs=[tensor_type], outputs=[tensor_type]
+        )
+
+        func_op = FuncOp(name="main", function_type=function_type)
```
alexarice, line 163: It feels like the `func_op` isn't really necessary to what we're trying to show here, but I guess it's fine to keep it.

alexarice (approved):
This seems reasonable to me, I left one comment about the func op.

My understanding is that the slow thing is the parsing of the module that this function generates. Is this benchmark testing that?

benchmarks/workloads.py
EdmundGoodman: I think lets just keep it this way for now for simplicity and to get this PR merged to unblock #4834

PR #4843, opened 2025-07-21
dialects: (pdl_interp) `apply_constraint`

jumerckx:


xdsl/dialects/pdl_interp.py
```diff
@@ -417,6 +422,48 @@ def __init__(
...
+    is_negated = prop_def(
+        BoolAttr, prop_name="isNegated", default_value=BoolAttr(False, i1)
+    )
+    args = var_operand_def(AnyPDLTypeConstr)
+    rresults = var_result_def(AnyPDLTypeConstr)
+    true_dest = successor_def()
+    false_dest = successor_def()
+    irdl_options = [ParsePropInAttrDict()]
+
+    assembly_format = "$name `(` $args `:` type($args) `)` `:` type($rresults) attr-dict `->` $true_dest `, ` $false_dest"
```
alexarice:
```suggestion
    assembly_format = "$name `(` $args `:` type($args) `)` (`:` type($rresults)^)? attr-dict `->` $true_dest `,` $false_dest"
```
Why give a different assembly format to upstream?

alexarice: Also there was an extra space in your comma directive (this is automatically added by the assembly format printer, but allows it to be parsed without the space)

xdsl/dialects/pdl_interp.py
```diff
@@ -417,6 +422,48 @@ def __init__(
...
+class ApplyConstraintOp(IRDLOperation):
+    """
+    See external [documentation](https://mlir.llvm.org/docs/Dialects/PDLInterpOps/#pdl_interpapply_constraint-pdl_interpapplyconstraintop).
+    """
+
+    name = "pdl_interp.apply_constraint"
+    traits = traits_def(IsTerminator())
+    constraint_name = prop_def(StringAttr, prop_name="name")
+    is_negated = prop_def(
+        BoolAttr, prop_name="isNegated", default_value=BoolAttr(False, i1)
```
alexarice:
```suggestion
        BoolAttr, prop_name="isNegated", default_value=BoolAttr.from_bool(False)
```

xdsl/dialects/pdl_interp.py
```diff
@@ -417,6 +422,48 @@ def __init__(
...
+    assembly_format = "$name `(` $args `:` type($args) `)` `:` type($rresults) attr-dict `->` $true_dest `, ` $false_dest"
+
+    def __init__(
+        self,
+        constraint_name: str | StringAttr,
+        args: Sequence[SSAValue],
+        true_dest: Block,
+        false_dest: Block,
+        res_types: Sequence[AnyPDLType] = (),
+        is_negated: bool = False,
```
alexarice:
```suggestion
        is_negated: bool | BoolAttr = False,
```
We tend to allow attributes to be passed in when we already have them.

xdsl/dialects/pdl_interp.py
```diff
@@ -417,6 +422,48 @@ def __init__(
...
+    name = "pdl_interp.apply_constraint"
+    traits = traits_def(IsTerminator())
+    constraint_name = prop_def(StringAttr, prop_name="name")
+    is_negated = prop_def(
+        BoolAttr, prop_name="isNegated", default_value=BoolAttr(False, i1)
+    )
+    args = var_operand_def(AnyPDLTypeConstr)
+    rresults = var_result_def(AnyPDLTypeConstr)
+    true_dest = successor_def()
+    false_dest = successor_def()
```
alexarice, line 439: Would it be worth sticking to the upstream names of `match_dest` and `failure_dest`?

alexarice (commented):
I don't have a huge familiarity with this dialect but left some comments anyway in case they are helpful. Please don't wait for my tick to merge.

xdsl/dialects/pdl_interp.py
jumerckx: Thanks for the review! And yes!

jumerckx: ah, but upstream is `trueDest` and `falseDest`, they just don't use those names in the example code.

alexarice: I see, (MLIR docs continue to be actual carnage (Maybe I should try to make MLIR PRs for these things))

tests/interpreters/test_pdl_interp_interpreter.py
```diff
@@ -771,3 +772,70 @@ def test_get_defining_op_block_argument():
...
+    # Test with non-existent constraint
+    apply_constraint_op_nonexistent = pdl_interp.ApplyConstraintOp(
+        "non_existent_constraint",
+        (c0,),
+        true_dest,
+        false_dest,
+        (pdl.AttributeType(),),
+        is_negated=BoolAttr.from_bool(False),
+    )
+    with pytest.raises(InterpretationError):
```
alexarice: Could this have a match clause?

xdsl/interpreters/pdl_interp.py
```diff
@@ -403,6 +412,26 @@ def call_func(
...
+        assert len(args) == 1
+        constraint_name = op.constraint_name.data
+
+        if constraint_name not in self.native_constraints:
+            raise InterpretationError(f"Unknown constraint function: {constraint_name}")
+        passed, *results = self.native_constraints[constraint_name](
+            *args,
+        )
+        is_negated = op.is_negated.value
+        terminator = op.true_dest if passed != bool(is_negated) else op.false_dest
```
alexarice:
```suggestion
        terminator = op.true_dest if passed != is_negated else op.false_dest
```
Does this work?

alexarice: Might even work with `op.is_negated` instead of `op.is_negated.value`?

jumerckx: I can drop `.value`, but without the `bool` a false value is represented as `-1`.

alexarice: ah of course

xdsl/interpreters/pdl_interp.py
```diff
@@ -34,6 +35,14 @@ class PDLInterpFunctions(InterpreterFunctions):
 
     _rewriter: PatternRewriter | None = field(default=None)
 
+    native_constraints: ClassVar[
+        dict[str, Callable[..., tuple[bool, tuple[Any, ...]]]]
+    ] = {}
+    """
+    The functions that can be used in `pdl_interp.apply_constraint`. Note that we do
+    not verify that the functions are used with the correct types.
+    """
```
alexarice: Is having this as a class var the done thing here? I'm not sure why but having a mutable class variable seems strange to me?

alexarice (approved):
As far as I can tell this looks good, but I don't really know anything about the interpreter, so maybe wait for a second review.

xdsl/interpreters/pdl_interp.py
jumerckx: I'd copied the `ClassVar` from `PDLMatcher`. I can remove it, I'm not even sure what it really does.

alexarice: A `ClassVar` is shared between all instances of this class, compared to a field which can be set independently for each object. I'm unsure what is right for this situation, but it seemed surprising to me (as I've said I really don't know how this bit of infrastructure works)

superlopuh: Yep should definitely not be a ClassVar

superlopuh: Neither on PDLMatcher

jumerckx: ~for pdl: https://github.com/xdslproject/xdsl/pull/4844~

xdsl/interpreters/pdl_interp.py
```diff
@@ -34,6 +35,12 @@ class PDLInterpFunctions(InterpreterFunctions):
 
     _rewriter: PatternRewriter | None = field(default=None)
 
+    native_constraints: dict[str, Callable[..., tuple[bool, tuple[Any, ...]]]] = {}
```
superlopuh:
```suggestion
    native_constraints: dict[str, Callable[..., tuple[bool, tuple[Any, ...]]]] = field(default_factory=lambda: {})
```

xdsl/dialects/pdl_interp.py
```diff
@@ -417,6 +421,50 @@ def __init__(
         super().__init__(operands=[lhs, rhs], successors=[trueDest, falseDest])
 
 
+@irdl_op_definition
+class ApplyConstraintOp(IRDLOperation):
```
superlopuh, line 425: This PR is quite big, could you please first merge this op, then the interpreter changes?

jumerckx: ok 👍

jumerckx:
I've split off the interpreter implementation: https://github.com/xdslproject/xdsl/pull/4845

xdsl/interpreters/pdl_interp.py
jumerckx: Actually, for pdl, the changes are more involved than just removing the ClassVar, a new PDLMatcher is created in each `match_and_rewrite`.
I've closed the pr for now but can look into reworking it if you want.

alexarice (approved):

PR #4844, opened 2025-07-21
interpreter: (pdl) remove ClassVar for `native_constraints`

jumerckx:
as discussed [here](https://github.com/xdslproject/xdsl/pull/4843#pullrequestreview-3037675332)

PR #4846, opened 2025-07-21
interpreter: (pdl) move native_constraints to instance attribute

superlopuh:
An attempt at #4844.

CC @jumerckx

### DIFF
--- a/xdsl/interpreters/pdl.py
+++ b/xdsl/interpreters/pdl.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from dataclasses import dataclass, field
-from typing import IO, Any, ClassVar
+from typing import IO, Any
 
 from xdsl.context import Context
 from xdsl.dialects import pdl
@@ -31,7 +31,9 @@ class PDLMatcher:
     the corresponding xDSL object.
     """
 
-    native_constraints: ClassVar[dict[str, Callable[..., bool]]] = {}
+    native_constraints: dict[str, Callable[..., bool]] = field(
+        default_factory=lambda: {}
+    )
     """
     The functions that can be used in `pdl.apply_native_constraint`. Note that we do
     not verify that the functions are used with the correct types.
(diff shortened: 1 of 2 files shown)


jumerckx:
This looks right to me, but I'm not a stakeholder in the pdl interpreter.

As as sidenote, maybe the `native_constraints` for `pdl_interp` should move somewhere else as well, currently they're stored in `PDLInterpFunctions` but it might make sense to put it in a structure alongside the interpreter like here...

alexarice:
> This looks right to me, but I'm not a stakeholder in the pdl interpreter.

I feel you're the primary user, at least from recent PRs

xdsl/interpreters/pdl.py
```diff
@@ -31,7 +31,9 @@ class PDLMatcher:
     the corresponding xDSL object.
     """
 
-    native_constraints: ClassVar[dict[str, Callable[..., bool]]] = {}
+    native_constraints: dict[str, Callable[..., bool]] = field(
+        default_factory=lambda: {}
+    )
```
alexarice, line 36: Why are we storing these in two places?

xdsl/interpreters/pdl.py
```diff
@@ -217,9 +219,16 @@ class PDLRewritePattern(RewritePattern):
     functions: PDLRewriteFunctions
     pdl_rewrite_op: pdl.RewriteOp
     interpreter: Interpreter
+    native_constraints: dict[str, Callable[..., bool]] = field(
+        default_factory=lambda: {}
```
alexarice: I don't think the field declaration is necessary, this is always set in the `__init__` anyway

xdsl/interpreters/pdl.py
alexarice: Nevermind I can see it now.

alexarice (approved):
This looks reasonably to me, although I'm far from an expert in this part of the repository.

superlopuh:
I think for PDL interp they're in the perfect place

superlopuh:
Modulo me eventually wanting to shift the state to some interpretation state object

PR #4849, opened 2025-07-22
dialects: (eqsat) add verification for eclass operands

superlopuh:


jumerckx:
This is not blocking, but another useful verification to have could be to verify an eclass does not contain the same operand twice.

jumerckx (approved):

PR #4852, opened 2025-07-22
transformations: (eqsat-add-costs) add default cost

superlopuh:
In the presence of cycles, our toy cost model falls over, as we calculate costs top to bottom. This PR adds a parameter to both the helper functions and the pass to provide a default value when it cannot be computed. The default behaviour is unchanged, but we now have an escape hatch to guarantee that we can extract something out of nothing (even if the order and dominance aren't guaranteed to be preserved).

### DIFF
--- a/tests/filecheck/transforms/eqsat-add-costs-with-default.mlir
+++ b/tests/filecheck/transforms/eqsat-add-costs-with-default.mlir
@@ -1,10 +1,10 @@
 // RUN: xdsl-opt -p eqsat-add-costs{default=1000} --verify-diagnostics --split-input-file %s | filecheck %s
 
 //      CHECK:    func.func @recursive(%a : index) -> index {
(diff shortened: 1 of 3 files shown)


jumerckx:
I'm not too familiar with this code, but it seems like both eclassops as well as other operations get a cost?
In other frameworks, costs are associated with eclasses only. As long as there is an acyclic extraction embedded in the e-graph, it's possible to (greedily) figure out costs for e-classes.
To me it seems a bit arbitrary to try associate a cost with a cycle.

We can tackle this issue later as it pops up as well of course.

superlopuh:
Here the expressions get a cost, and eclasses get an index of the operand with the least cost. It's more of a placeholder until we implement a better extraction.

jumerckx (approved):

PR #4857, opened 2025-07-22
documentation: (marimo) add equality saturation notebook

### LINKED ISSUE #3174: documentation: (eqsat) expand on top-level documentation in equality saturation dialect file

superlopuh:
Resolves #3174

The notebook works locally even if the version online doesn't since we just changed the API. The docs only show the last stable release's docs website anyway, so it'll automatically be fixed by the time it is visible on the website.


jumerckx:
Could we have two modules, one with the pdl_interp code, and one with the payload, such that we only print the full module where it matters.
I was messing around with
```
input_module = Parser(ctx, pdl_program).parse_module()
input_func = Parser(ctx, input_program).parse_module().body.op
input_func.detach()
input_module.body.first_block.insert_op_before(input_func, input_module.body.first_block.first_op)
```
But this becomes annoying with the fact that you can't mutate values in marimo.
Perhaps the python interface to the eqsat pass should optionally take a separate pdl_interp module?

superlopuh:
Yeah that's what I was thinking as well, I didn't want to open that PR at the same time as yours as there will definitely be conflicts but I can do it if it's not too annoying for you.

jumerckx:
Up to you. I can handle the conflicts so feel free. Otherwise I can also create a follow up pr later.

jumerckx (approved):

PR #4859, opened 2025-07-22
[WIP] frontend: (pyast) development tracking branch for PDL rewrite frontend DSL

EdmundGoodman:
This branch tracks the development of the PDL rewrite frontend DSL

### DIFF
--- a/tests/filecheck/frontend/programs/pdl.py
+++ b/tests/filecheck/frontend/programs/pdl.py
@@ -1,0 +1,42 @@
+"""Tests for PDL rewriting Python DSL.
+
+These tests exercise the API providing an evaluatable Python DSL for expressing
+the semantics of rewrites, which can then be lowered to the PDL dialect using
+the PyAST frontend.
+
+The following code snippets show early suggestions for the structure of this API:
+
+```python
+# Derived from <https://github.com/xdslproject/xdsl/pull/1137>
+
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
This looks already very cool. 😄

tests/filecheck/frontend/programs/pdl.py
```diff
@@ -0,0 +1,73 @@
...
+    """Shim to avoid `Expr` AST node required for methods."""
+    return Rewriter.erase_op(operation)
+
+
+ctx = PyASTContext()
+ctx.register_type(arith.ConstantOp, pdl.OperationType())
+ctx.register_function(erase_op, pdl.EraseOp)
+
+
+@ctx.parse_program(pdl_rewrite=True)
```
superlopuh: This feels like the wrong API to me. `parse_program` should not know about its use-cases, and instead we should create new function wrappers that call `parse_program` and them process the results.

EdmundGoodman: Have rewritten it to be just `ctx.pdl_rewrite` instead, which then invokes the builder with the correct configuration

EdmundGoodman:
No longer under active development, so closed for now. We now have some PDL support with the example of simple erasure, and possibly more to come as PyAST evolves in the future!

PR #4865, opened 2025-07-23
core: Allow non-keyword enums in StrEnum

math-fehr:
Stacked PRs:
 * #4867
 * #4866
 * __->__#4865
 * #4878

--- --- ---

### DIFF
--- a/tests/irdl/test_attribute_definition.py
+++ b/tests/irdl/test_attribute_definition.py
@@ -205,16 +205,17 @@ def test_indirect_enum_guard():
             name = "test.indirect_enum"
 
 
-def test_identifier_enum_guard():
-    with pytest.raises(
-        ValueError,
-        match="All StrEnum values of an EnumAttribute must be parsable as an identifer.",
-    ):
+def test_attribute_def_with_non_identifier_enum():
+    """
+    Test the definition of an EnumAttribute with a non-identifier enum is
+    allowed.
+    """
 
-        class IndirectEnumData(  # pyright: ignore[reportUnusedClass]
-            EnumAttribute[TestNonIdentifierEnum]
-        ):
-            name = "test.non_identifier_enum"
+    @irdl_attr_definition
+    class TestNonIdentifierEnumAttr(  # pyright: ignore[reportUnusedClass]
+        EnumAttribute[TestNonIdentifierEnum]
+    ):
+        name = "test.non_identifier_enum"
 
 
 @irdl_attr_definition(init=False)
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -33,7 +33,6 @@ from typing_extensions import Self, TypeVar
 from xdsl.dialect_interfaces import DialectInterface
 from xdsl.traits import IsTerminator, NoTerminator, OpTrait, OpTraitInvT
 from xdsl.utils.exceptions import VerifyException
-from xdsl.utils.mlir_lexer import MLIRLexer
 from xdsl.utils.str_enum import StrEnum
 
 # Used for cyclic dependencies in type hints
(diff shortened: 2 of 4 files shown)


math-fehr:
This was sadly needed to support the attribute defined in https://github.com/xdslproject/xdsl/pull/4866

tests/irdl/test_attribute_definition.py
```diff
@@ -262,9 +255,7 @@ def test_typed_attribute():
     ):
 
         @irdl_attr_definition
-        class TypedAttr(  # pyright: ignore[reportUnusedClass]
-            TypedAttribute
-        ):
+        class TypedAttr(TypedAttribute):  # pyright: ignore[reportUnusedClass]
```
alexarice: unrelated change?

tests/irdl/test_attribute_definition.py
```diff
@@ -182,6 +182,11 @@ class TestNonIdentifierEnum(StrEnum):
     Spaced = "left right"
 
 
+@irdl_attr_definition
+class TestNonIdentifierEnumAttr(EnumAttribute[TestNonIdentifierEnum]):
```
alexarice: I feel this is missing the actual test?

tests/irdl/test_attribute_definition.py
math-fehr: My bad, my autoformatter set it that way for some reasons.

tests/irdl/test_attribute_definition.py
math-fehr: I moved it to a function, so only the function should fail would this fail!

xdsl/ir/core.py
```diff
@@ -309,12 +300,27 @@ class MyEnumAttribute(EnumAttribute[MyEnum], SpacedOpaqueSyntaxAttribute):
...
+        # They only need to be quoted if they are not valid bare identifiers.
+        for value in tuple(cls.enum_type):
+            if MLIRLexer.bare_identifier_suffix_regex.fullmatch(str(value)):
+                cls.enum_to_string[value] = str(value)
+            else:
+                cls.enum_to_string[value] = f'"{str(value)}"'
+
     def print_parameter(self, printer: Printer) -> None:
-        printer.print_string(self.data.value)
+        printer.print_string(self.enum_to_string[self.data])
```
superlopuh: ?
```suggestion
        printer.print_identifier_or_string_literal(self.data.value)
```

alexarice: I don't think this is correct with the caching

alexarice: (I would be ok with removing the caching and just using this for simplicity)

xdsl/parser/base_parser.py
```diff
@@ -293,12 +293,19 @@ def parse_str_enum(self, enum_type: type[_EnumType]) -> _EnumType:
...
+            if val in enum_type.__members__.values():
+                self._consume_token(MLIRTokenKind.BARE_IDENT)
+                return enum_type(val)
+            return None
+
+        if self._current_token.kind != MLIRTokenKind.STRING_LIT:
             return None
 
-        val = self._current_token.text
+        val = StringLiteral.from_span(self._current_token.span).string_contents
```
superlopuh: Feels like we might want a `parse_optional_identifier_or_string_literal` helper for use here and in the attribute dictionary parsing?

xdsl/ir/core.py
```diff
@@ -309,12 +300,27 @@ class MyEnumAttribute(EnumAttribute[MyEnum], SpacedOpaqueSyntaxAttribute):
     """
 
     enum_type: ClassVar[type[StrEnum]]
+    enum_to_string: dict[StrEnum, str]
```
superlopuh: I'm not sure of the value of this performance optimisation, can't we wait until we see enum printing to be slow before speeding it up? Isn't printing more of a debug thing anyway?

superlopuh (approved):

PR #4866, opened 2025-07-23
dialects: Add #llvm.FramePointerKindAttr

math-fehr:
dialects: Add #llvm.FramePointerKindAttr


tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/attrs.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: mlir-opt %s --mlir-print-op-generic | xdsl-opt --print-op-generic | filecheck %s
+
+// CHECK: frame_all = #llvm.framePointerKind<all>
+// CHECK: frame_non_leaf = #llvm.framePointerKind<"non-leaf">
+// CHECK: frame_none = #llvm.framePointerKind<none>
+// CHECK: frame_reserved = #llvm.framePointerKind<reserved>
+"test.op"() {
+    frame_all = #llvm.framePointerKind<all>,
+    frame_non_leaf = #llvm.framePointerKind<"non-leaf">,
```
alexarice, line 9: why did they do this

xdsl/dialects/llvm.py
```diff
@@ -1502,6 +1502,29 @@ def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
...
+
+@irdl_attr_definition
+class FramePointerKindAttr(EnumAttribute[FramePointerKind]):
+    """LLVM Frame Pointer Kind."""
+
+    name = "llvm.framePointerKind"
+
+    def print_parameter(self, printer: Printer) -> None:
+        with printer.in_angle_brackets():
+            super().print_parameter(printer)
```
alexarice, line 1520: It surprises me that printing the parameters in angle brackets isn't the default behaviour

alexarice (approved):

tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/attrs.mlir
math-fehr: I just don't know at this point :(
Took me a while to even find that they handle it that way.

xdsl/dialects/llvm.py
math-fehr: The reason is kinda wild.

There are two ways to print an attribute. Either you use the "dialect format", which is `#dialect<enum<...>>`, or you use the "pretty format", which is `#dialect.enum<...>`.
You can only use the "pretty format" if you use the angle brackets.

So, why would you even use the "dialect format"? Well, if you have an operation that knows statically the attribute type, then you only print the `...` instead of `#dialect<enum ...>`. In particular, that means that you automatically print `all` instead of `#dialect<enum all>`. If you had angle brackets, this would be `<all>` instead of `#dialect<enum<all>>`

superlopuh (approved):

xdsl/dialects/llvm.py
math-fehr: (Btw, I checked and it is the default behavior of attributes in our codebase, but not enums)

PR #4872, opened 2025-07-24
dialects: (tosa) Add support for `tosa.add`

jakedves:
Adds support for `tosa.add` and tests (roundtrip test, and testing of the verification step for some values)

### DIFF
--- a/tests/dialects/test_tosa.py
+++ b/tests/dialects/test_tosa.py
@@ -1,9 +1,6 @@
 import pytest
 
 from xdsl.dialects.builtin import (
-    Float32Type,
-    IntegerType,
-    Signedness,
     TensorType,
     i32,
     f32,
--- a/tests/dialects/test_tosa.py
+++ b/tests/dialects/test_tosa.py
@@ -6,7 +6,6 @@ from xdsl.dialects.builtin import (
 from xdsl.dialects.tosa import AddOp, are_tosa_broadcastable
 from xdsl.utils.test_value import create_ssa_value
 
-
 t_i = TensorType(i32, [1, 2, 3, 4])
 t_f = TensorType(f32, [1, 2, 3, 4])
 t_flat = TensorType(i32, [1, 1, 1, 1])
--- a/tests/dialects/test_tosa.py
+++ b/tests/dialects/test_tosa.py
@@ -1,0 +1,33 @@
+import pytest
+
+from xdsl.dialects.builtin import (
+    Float32Type,
+    IntegerType,
+    Signedness,
+    TensorType,
+    i32,
+    f32,
+)
+from xdsl.dialects.tosa import AddOp
+from xdsl.utils.test_value import create_ssa_value
+from xdsl.utils.exceptions import VerifyException
+
+
+tensor_i = create_ssa_value(TensorType(i32, [1, 2, 3, 4]))
+tensor_flat = create_ssa_value(TensorType(i32, [1, 1, 1, 1]))
+tensor_f = create_ssa_value(TensorType(f32, [1, 2, 3, 4]))
+
+
+def test_valid_cases():
+    AddOp(operands=[tensor_i, tensor_i], result_types=[tensor_i.type]).verify_()
+    AddOp(operands=[tensor_i, tensor_flat], result_types=[tensor_i.type]).verify_()
+
+
+def test_invalid_cases():
+    with pytest.raises(VerifyException):
+        AddOp(operands=[tensor_i, tensor_f], result_types=[tensor_f.type]).verify_()
+
+    with pytest.raises(VerifyException):
+        AddOp(
+            operands=[tensor_i, tensor_flat], result_types=[tensor_flat.type]
+        ).verify_()
--- a/tests/dialects/test_tosa.py
+++ b/tests/dialects/test_tosa.py
@@ -1,29 +1,41 @@
-import pytest
-
 from xdsl.dialects.builtin import (
     TensorType,
     f32,
     i32,
 )
-from xdsl.dialects.tosa import AddOp
-from xdsl.utils.exceptions import VerifyException
+from xdsl.dialects.tosa import AddOp, are_tosa_broadcastable
 from xdsl.utils.test_value import create_ssa_value
 
-tensor_i = create_ssa_value(TensorType(i32, [1, 2, 3, 4]))
-tensor_flat = create_ssa_value(TensorType(i32, [1, 1, 1, 1]))
-tensor_f = create_ssa_value(TensorType(f32, [1, 2, 3, 4]))
+
+t_i = TensorType(i32, [1, 2, 3, 4])
+t_f = TensorType(f32, [1, 2, 3, 4])
+t_flat = TensorType(i32, [1, 1, 1, 1])
+t_small = TensorType(i32, [1, 2])
+t_large = TensorType(i32, [4, 5, 6, 7])
+
+tensor_i = create_ssa_value(t_i)
+tensor_f = create_ssa_value(t_f)
+tensor_flat = create_ssa_value(t_flat)
+tensor_small = create_ssa_value(t_small)
+tensor_large = create_ssa_value(t_large)
+
+
+def test_are_tosa_broadcastable():
+    # test same
+    assert are_tosa_broadcastable(t_i, t_i, t_i)
+
+    # test broadcasting
+    assert are_tosa_broadcastable(t_i, t_flat, t_i)
+    assert are_tosa_broadcastable(t_flat, t_i, t_i)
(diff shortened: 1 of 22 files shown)


xdsl/dialects/tosa.py
```diff
@@ -68,11 +70,71 @@ class RescaleOp(IRDLOperation):
...
+        both tensors must have the same rank. They should either have the same number of elements per dim, or
+        if there is only one element it can be broadcast implcitly.
+        """
+        t1 = self.in1.type
+        t2 = self.in2.type
+        t_out = self.output.type
+
+        assert isa(t1, TensorType)
+        assert isa(t2, TensorType)
+        assert isa(t_out, TensorType)
```
superlopuh: This one shouldn't be necessary. During rewriting, patterns can rewrite the values to a different type, so it's nto safe to assume that their type doesn't change. This isn't the case for results, so we preserve the type from `result_def`. Pyright should be happy with this line removed.

xdsl/dialects/tosa.py
```diff
@@ -68,11 +70,71 @@ class RescaleOp(IRDLOperation):
...
+            raise VerifyException(
+                "'tosa.add' rank mismatch between input and output tensors"
+            )
+
+        if not (
+            t1.get_element_type() == t2.get_element_type() == t_out.get_element_type()
+        ):
+            raise VerifyException(
+                "'tosa.add' element type mismatch between inputs and output tensors"
+            )
```
superlopuh: Can you please move this into the constraint of the operation? Ideally with a `VarConstraint` and `TensorType.constr`

xdsl/dialects/tosa.py
```diff
@@ -68,11 +70,71 @@ class RescaleOp(IRDLOperation):
...
+            if dim_in1 != dim_in2 and 1 not in (dim_in1, dim_in2):
+                raise VerifyException(
+                    f"'tosa.add' shapes mismatch along axis: {dim_in1} and {dim_in2}"
+                )
+
+            # check that t_out is made from only broadcasting the operand types
+            if dim_out != max(dim_in1, dim_in2):
+                raise VerifyException(
+                    "'tosa.add' incompatible result type from operand types"
+                )
```
superlopuh: Can you please put this logic into a helper with a unit test?

xdsl/dialects/tosa.py
jakedves: Does `verify_` get called during each stage of rewriting? I thought only at the end where you should still expect `TensorType`? Pyright complained about unknown types for all the following calls without it

superlopuh (commented):
Can you please add a rountrip test with MLIR?

xdsl/dialects/tosa.py
superlopuh: No it's just at the end of the pass, I'm not sure what Pyright was complaing about but we can also leave it in.

superlopuh:
I would recommend running `make precommit-install` to avoid formatting CI issues.

jakedves:
Should be everything @superlopuh ? Roundtrip test in `tests/filecheck/dialects/tosa/ops.mlir`

tests/dialects/test_tosa.py
```diff
@@ -0,0 +1,29 @@
...
+from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.test_value import create_ssa_value
+
+tensor_i = create_ssa_value(TensorType(i32, [1, 2, 3, 4]))
+tensor_flat = create_ssa_value(TensorType(i32, [1, 1, 1, 1]))
+tensor_f = create_ssa_value(TensorType(f32, [1, 2, 3, 4]))
+
+
+def test_valid_cases():
+    AddOp(operands=[tensor_i, tensor_i], result_types=[tensor_i.type]).verify_()
```
superlopuh: The idea is that other ops should reuse the same verification helper, so I would recommend creating a new function that takes three tensor types and an operation name and raises the appropriate exception.

superlopuh: Ah sorry you added one! Can you please move the invalid cases into a filecheck test called `ops_invalid.mlir` with `--verify-diangostics`, and remove this pytest?

xdsl/utils/type.py
```diff
@@ -53,3 +53,26 @@ def have_compatible_shape(lhs_type: Attribute, rhs_type: Attribute) -> bool:
         dim1 == DYNAMIC_INDEX or dim2 == DYNAMIC_INDEX or dim1 == dim2
         for dim1, dim2 in zip(lhs_type.get_shape(), rhs_type.get_shape())
     )
+
+
+def are_tosa_broadcastable(t_in1: Attribute, t_in2: Attribute, t_out: Attribute):
```
superlopuh: if this is tosa specific then it should be in the tosa dialect. I think maybe eventually we might want such a helper in this file, but then it shouldn't reference tosa

tests/filecheck/dialects/tosa/ops.mlir
```diff
@@ -3,10 +3,12 @@
 %0 = "test.op"() : () -> tensor<12x34xi32>
 %1 = tosa.clamp %0 {max_fp = 1.000000e+00 : f32, max_int = 1 : i64, min_fp = 0.000000e+00 : f32, min_int = 0 : i64} : (tensor<12x34xi32>) -> tensor<12x34xi32>
 %2 = tosa.rescale %0 {double_round = false, input_zp = 127 : i32, multiplier = array<i32: 1073741824>, output_zp = -1 : i32, per_channel = false, scale32 = true, shift = array<i8: 30>} : (tensor<12x34xi32>) -> tensor<12x34xi32>
+%3 = tosa.add %0, %0 : (tensor<12x34xi32>, tensor<12x34xi32>) -> tensor<12x34xi32>
```
alexarice: It would be good to have a broadcasting example here.

xdsl/dialects/tosa.py
```diff
@@ -1,23 +1,52 @@
...
+
+    # check ranks are broadcastable
+    in_shapes = zip(t_in1.get_shape(), t_in2.get_shape())
+
+    if not all(dim1 == dim2 or dim1 == 1 or dim2 == 1 for dim1, dim2 in in_shapes):
+        return False
+
+    # check output shape is constructed from input shapes
+    shapes = zip(t_in1.get_shape(), t_in2.get_shape(), t_out.get_shape())
+    return all(dim_out == max(dim1, dim2) for dim1, dim2, dim_out in shapes)
```
superlopuh: Would be good to have a doc string and I feel that it's clearer to do the whole check in two steps:
```suggestion
def are_tosa_broadcastable(lhs: Attribute, rhs: Attribute, out: Attribute):
    """
    Returns `True` if lhs and rhs have compatible shapes with broadcasting: the dimensions have to match, or one of them must be equal to 1, in which case the corresponding resulting dimension must be equal to the non-1 dimension.
    
    e.g.: `[1, 2, 3] & [4, 2, 1] -> [4, 2, 3]`
    """
    if (
        not isinstance(lhs, ShapedType)
        or not isinstance(rhs, ShapedType)
        or not isinstance(out, ShapedType)
    ):
        return False

    lhs_shape = lhs.get_shape()
    rhs_shape = rhs.get_shape()
    out_shape = out.get_shape()
    
    ranks_equal = len(lhs_shape) == len(rhs_shape) == len(out_shape)
    if not ranks_equal:
        return False

    # Expected dimensions match output dimensions
    # Input dimensions are equal or broadcast
    return all(
        max(l, r) == o and ((l == r) or (min(l, r) == 1))
        for l, r, o in zip(lhs_shape, rhs_shape, out_shape)
    )
```

alexarice: For readability, I'd be tempted to just go for `(l == r == o) or (l == 1 and r == o) or (r == 1 and l == o)`

superlopuh: might have to be careful with parentheses here but I like this formulation also

superlopuh (approved):
Looks great! Thanks for iterating

xdsl/dialects/tosa.py
```diff
@@ -68,11 +104,44 @@ class RescaleOp(IRDLOperation):
...
+
+    name = "tosa.add"
+
+    T: ClassVar = VarConstraint("T", AnyAttr())
+
+    in1 = operand_def(TensorType.constr(T))
+    in2 = operand_def(TensorType.constr(T))
+    output = result_def(TensorType.constr(T))
+
+    assembly_format = "$in1 `,` $in2 attr-dict `:` `(` type($in1) `,` type($in2) `)` `->` type($output)"
```
```suggestion
    assembly_format = "operands attr-dict `:` functional-type(operands, results)"
```
is there any reason not to use the upstream assembly format?

alexarice: and to use upstream operand names for that matter.

PR #4876, opened 2025-07-24
transformations: (convert-ptr-to-x86) fold constant offsets

superlopuh:
I'm not sure how to optimise out the moves that get inserted when lowering addition to x86 (which is lowered to an inout register so needs an extra move for SSA/register safety), so let's add the constant offset special case to the lowering.


superlopuh:
Actually I'm not longer sure that we want this in, let's see if we can avoid it

superlopuh:
We went with a different approach

PR #4880, opened 2025-07-25
dialects: (x86) add shufps operation

superlopuh:
Feels like a useful instruction to have, allows us to represent AVX vector shuffles.


superlopuh:
Ah I hadn't noticed that distinction. So if we want to implement both shuffles, it sounds like we would need two operations, an IRS that only operates on AVX registers, and an IDSS that operates on AVX registers?

emmau678:
> Ah I hadn't noticed that distinction. So if we want to implement both shuffles, it sounds like we would need two operations, an IRS that only operates on AVX registers, and an IDSS that operates on AVX registers?

as I understand, yes. Although the register distinction is possibly a bit tricky because both SSE and AVX operate on the xmm regs, it's the instructions that are different.

qaco (approved):

qaco:
> looks good to me, I just have one question: if the goal is to represent AVX shuffle, why is the SSE version implemented here? As I understand, the AVX version does not have the "inout" register operand in the same sense as SSE

I hadn't notice either ! To answer Sasha’s question, it seems to me that AVX instructions are backward-compatible with SSE registers (but in a non-destructive way) so I guess we don't need the SSE one at all ?

superlopuh:
Coolio I'll rework it to the AVX version

PR #4882, opened 2025-07-25
dialects: (emitc) Add OpaqueAttr

Jimmy2027:
Add the [emitc OpaqueAttr](https://mlir.llvm.org/docs/Dialects/EmitC/#opaqueattr)


tests/filecheck/dialects/emitc/emitc_attrs.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: XDSL_ROUNDTRIP
+
+//===----------------------------------------------------------------------===//
+// OpaqueAttr
+//===----------------------------------------------------------------------===//
+
+// CHECK: opaque_attr = #emitc.opaque<"some_value">
+// CHECK-NEXT: #emitc.opaque<"\"quoted_attr\"">
```
Jimmy2027: in [mlir](https://github.com/llvm/llvm-project/blob/adb2421202e4014b4986a9f1eb40833dc7ec25ad/mlir/test/Dialect/EmitC/attrs.mlir#L9) the quotes are expected to be escaped, but this seems unrelated to the OpaqueAttr. 
Maybe something to implement in the StringAttr?

superlopuh: Is this resolved?

Jimmy2027: not really, not sure what to do about this.. Should I add a comment in the test? Could be confusing since the behaviour is different than mlir-opt

superlopuh: I mean we can handle escaped strings, you mean that it should be `\22` instead of `\"`?

Jimmy2027: yeah thats probably what the user would expect

superlopuh: I have a feeling that if xDSL can parse the mlir-opt escape and vice versa there's nothing to worry about

superlopuh: Can you please add a roundtrip test?

Jimmy2027: you're right `mlir-opt` is able to parse that, but `xdsl-opt` does not like the "\22quoted_attr\22":
`xdsl.utils.exceptions.VerifyException: "\22quoted_attr\22" should be of base attribute string`

superlopuh: I think that's a separate bug then can you please have a go at fixing it? It should be a matter of handling numbers in escape sequences in `StringLiteral.bytes_contents`

Jimmy2027: the StringLiteral.bytes_contents already handles numbers in escape sequences, I think the issue is in the `_lex_string_literal`.
Is this going in the right direction?
https://github.com/xdslproject/xdsl/pull/4893

superlopuh: Definitely looks like the right direction to me, let's iterate there on precise implementation.

Jimmy2027: fixed in 1547fb431857a1ff1d1d544f7b554f1dbcee63f0

PR #4883, opened 2025-07-25
dialects: (vector) use custom assembly for vector.extract

alexarice:
Example use of custom directives.

I believe the change is spacing is actually a bugfix, as it agrees with upstream documentation, but I could likely split out this change or revert it if wanted.

This has strengthened my belief that it would be a good idea to define some getters and setters on the common format directives.


alexarice:
@superlopuh still happy with this?

alexarice:
I've looked again and now think that I may have been wrong about the colon spacing, let me iterate

superlopuh (approved):

PR #4884, opened 2025-07-25
misc: (FormatDirective) add getters and setters

alexarice:
Adds utility methods to some directives that isolates setting and getting the appropriate variable.
The main purpose of this is for use as the parameter to a custom directive.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -20,6 +20,8 @@
               buildInputs = [
                 uv
                 nodejs_22
+                llvmPackages_20.mlir
+                llvmPackages_20.tblgen
               ];
             };
           }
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -20,8 +20,6 @@
               buildInputs = [
                 uv
                 nodejs_22
-                llvmPackages_20.mlir
-                llvmPackages_20.tblgen
```
superlopuh: !!

alexarice: Whoops, they are broken because nix is still on 20.1.6 for some reason.

alexarice: I'll split it

PR #4885, opened 2025-07-25
misc: return verified attribute/range from `verify`

superlopuh:
I occasionally have to do many stages of verification, and we don't have a nice way of doing that today. This PR unlocks a new pattern where you can use the verified attribute with the narrowed type without casting after verification.

### DIFF
--- a/tests/irdl/test_attribute_definition.py
+++ b/tests/irdl/test_attribute_definition.py
@@ -9,7 +9,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from enum import auto
 from io import StringIO
-from typing import Annotated, ClassVar, Generic, TypeAlias
+from typing import Annotated, ClassVar, Generic, TypeAlias, cast
 
 import pytest
 from typing_extensions import TypeVar, override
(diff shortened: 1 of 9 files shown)


tests/irdl/test_attribute_definition.py
```diff
@@ -613,8 +613,7 @@ class InformativeAttr(ParametrizedAttribute):
 
 
 def test_informative_attribute():
-    okay = InformativeAttr(NoneAttr())
-    okay.verify()
```
superlopuh, line 617: unnecessary call to verify called at init

tests/irdl/test_attribute_definition.py
```diff
@@ -684,13 +683,11 @@ def verify(
         self,
         attr: Attribute,
         constraint_context: ConstraintContext,
-    ) -> None:
-        if not isa(attr, ListData):
-            raise VerifyException(
-                f"Expected {attr} to be instance of {ListData.__name__}"
-            )
+    ) -> ListData[AttributeInvT]:
+        attr = BaseAttr(ListData).verify(attr, constraint_context)
```
superlopuh, line 687: one potential use-case for this pattern is to do multiple stages of verification less verbosely

xdsl/irdl/constraints.py
```diff
@@ -336,9 +341,10 @@ def verify(
         self,
         attr: Attribute,
         constraint_context: ConstraintContext,
-    ) -> None:
+    ) -> AttributeCovT:
         if attr != self.attr:
             raise VerifyException(f"Expected attribute {self.attr} but got {attr}")
+        return self.attr
```
superlopuh, line 347: A bit of sleight of hand

superlopuh:
Don't need this for now

PR #4886, opened 2025-07-25
misc: (utils) move `is_const_classvar` to utils

alexarice:
Makes the error message slightly worse but makes things a bit more consistent.

### DIFF
--- a/tests/irdl/test_attribute_definition.py
+++ b/tests/irdl/test_attribute_definition.py
@@ -57,10 +57,7 @@ from xdsl.irdl import (
 )
 from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
-from xdsl.utils.exceptions import (
-    PyRDLAttrDefinitionError,
-    VerifyException,
-)
+from xdsl.utils.exceptions import PyRDLAttrDefinitionError, VerifyException
 from xdsl.utils.hints import isa
 
 
--- a/tests/irdl/test_attribute_definition.py
+++ b/tests/irdl/test_attribute_definition.py
@@ -57,7 +57,7 @@ from xdsl.irdl import (
 )
 from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
-from xdsl.utils.exceptions import PyRDLAttrDefinitionError, VerifyException
+from xdsl.utils.exceptions import PyRDLAttrDefinitionError, PyRDLError, VerifyException
 from xdsl.utils.hints import isa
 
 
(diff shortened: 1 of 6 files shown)


alexarice:
Could throw an exception in `is_const_classvar` for class vars which are not uppercase

tests/irdl/test_attribute_definition.py
```diff
@@ -1088,7 +1088,9 @@ def test_class_var_fail():
     """Test that lowercase ClassVar fields are not allowed."""
     with pytest.raises(
         PyRDLAttrDefinitionError,
-        match='Invalid ClassVar name "constant", must be uppercase.',
+        match=re.escape(
+            "Invalid field type typing.ClassVar[int] for field name constant."
```
superlopuh: I feel like the old error message was more instructive, I'm not sure how I would fix my defintion with the new error message.

alexarice: I agree

xdsl/utils/classvar.py
```diff
@@ -0,0 +1,28 @@
...
+     * `"ClassVar[MyType]"`.
+    """
+    return (
+        get_origin(annotation) is ClassVar
+        or annotation is ClassVar
+        or (isinstance(annotation, str) and annotation.startswith("ClassVar"))
+    )
+
+
+def is_const_classvar(field_name: str, annotation: Any) -> bool:
```
superlopuh: Maybe we actually want to refactor this and change the Operation definition rather than Attribute?

superlopuh: Or we could go with your suggestion instead and raise the error here.

alexarice: Any of these seem fine to me, do you have a preference?

superlopuh: I like your suggestion most

xdsl/utils/classvar.py
```diff
@@ -0,0 +1,35 @@
...
+    The type annotation can be one of
+     * `ClassVar[MyType]`,
+     * `ClassVar`, or
+     * `"ClassVar[MyType]"`.
+    This function throws an exception on a non UPPER_CASE class variable.
+    """
+    if not is_classvar(annotation):
+        return False
+    if not field_name.isupper():
+        raise PyRDLError(f'Invalid ClassVar name "{field_name}", must be uppercase.')
```
superlopuh: Could pass the error type in? Would then keep the errors more precise. More of a nit.

alexarice: Sure

tests/irdl/test_attribute_definition.py
```diff
@@ -57,7 +57,10 @@
 )
 from xdsl.parser import AttrParser, Parser
 from xdsl.printer import Printer
-from xdsl.utils.exceptions import PyRDLAttrDefinitionError, VerifyException
+from xdsl.utils.exceptions import (
+    PyRDLAttrDefinitionError,
+    VerifyException,
+)
```
superlopuh: supernit rogue diff not very important

alexarice:
```suggestion
from xdsl.utils.exceptions import PyRDLAttrDefinitionError, VerifyException
```

PR #4888, opened 2025-07-25
dialects: (vector) add vector.shuffle

superlopuh:
Adds the op and the constraint.


tests/filecheck/dialects/vector/vector_ops.mlir
```diff
@@ -120,3 +120,12 @@ func.func @insert_poison_idx(%a: vector<4x5xf32>, %b: f32) {
   vector.insert %b, %a[-1, 0] : f32 into vector<4x5xf32>
   return
 }
+
+// CHECK-LABEL: @shuffle
+func.func @shuffle(%a : vector<2xf32>, %b : vector<1x16xf32>, %c : vector<2x16xf32>, %d : vector<f32>) -> (vector<2xf32>, vector<3x16xf32>, vector<4xf32>, vector<2xf32>) {
+  %0 = vector.shuffle %a, %a[0, 3] : vector<2xf32>, vector<2xf32>
```
superlopuh: @alexarice do you know how to get it to print `[0, 3]` instead of `[0 : i64, 3 : i64]`?

alexarice: mask should be a `DenseArrayBase` not an `ArrayAttr`

superlopuh: I pushed an attempt at DenseArrayBase it fails in a super obscure location, would appreciate help with this when you have time.

xdsl/dialects/vector.py
```diff
@@ -1261,18 +1455,19 @@ def verify_(self):
...
         PrintOp,
-        CreateMaskOp,
-        ExtractOp,
-        ExtractElementOp,
-        InsertOp,
-        InsertElementOp,
+        ShuffleOp,
+        StoreOp,
         TransferReadOp,
         TransferWriteOp,
```
superlopuh: sorted these ascending for easier diffs in the future

tests/filecheck/dialects/vector/vector_ops.mlir
superlopuh: Got it to print but there's an extra space vs MLIR

superlopuh: Ah no MLIR also prints a space the spec lies

xdsl/dialects/vector.py
```diff
@@ -153,6 +162,194 @@ def get(
...
+            else result_trailing
+        )
+        return VectorType(element_type, ArrayAttr(shape))
+
+    def mapping_type_vars(
+        self, type_var_mapping: Mapping[TypeVar, AttrConstraint | IntConstraint]
+    ) -> AttrConstraint[VectorType]:
+        return ShuffleResultConstraint(
+            self.element_constr.mapping_type_vars(type_var_mapping)
+        )
```
alexarice, line 236: I feel this is cooking a bit too much. It feels to me that rolling a custom constraint for one operation is basically always going to be more complicated than just having a custom verify. In this case is it impossible to get the inference through with the predefined constraints?

superlopuh: Yeah that's the problem, we need both the verification and the result type inference. We don't have a Python hook for that so it's either this custom constraint and assembly format or custom printing&parsing&verification.

xdsl/dialects/vector.py
```diff
@@ -153,6 +162,194 @@ def get(
...
+            self.element_constr, shape=VarConstraint(_V1_SHAPE, _IntArrayConstr)
+        )
+
+    def v2(self) -> AttrConstraint[VectorType]:
+        return VectorType.constr(
+            self.element_constr, shape=VarConstraint(_V2_SHAPE, _IntArrayConstr)
+        )
+
+    def mask(self) -> AttrConstraint[DenseArrayBase[I64]]:
+        return VarConstraint(_MASK, _MaskConstr)
```
math-fehr: I would personally inline these methods to the `v1` `v2` contsraints in the operation below, just so it's easy to see what constraints the operands actually have.

PR #4890, opened 2025-07-26
core: OptionalUnitAttrVariable is not optional-like

superlopuh:
Small bug fix

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -271,6 +271,8 @@ class Directive(ABC):
     def is_anchorable(self) -> bool:
         """
         Can appear as an anchor in an optional group.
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -271,6 +271,8 @@ def is_present(self, op: IRDLOperation) -> bool:
     def is_anchorable(self) -> bool:
         """
         Can appear as an anchor in an optional group.
+        Anchorable directives that are not optional-like can't appear as the first
```
alexarice: I feel this has nothing to do with being anchorable?

superlopuh: You're right I moved the comment.

alexarice (approved):

PR #4891, opened 2025-07-26
dialects: (linalg) Default result is empty list rather than none for varadics

mesham:
This fixes a bug in several operations in the linalg dialect, where the default for results in the init method was None to signify no results. However as results are a varadic this will fail verification as instead a list must be provided, and no results should be an empty list. Therefore changing the default value from None to [].

Also minor update to linalg.transpose and broadcast operations, previously a single result type was passed to init method. But results are varadic so overly restrictive - and all other linalg operations pass a sequence of result types. Therefore changed this to a sequence of result types to be consistent and improve generality.


mesham:
Closed as reworking to fix specific linalg operations that need this

PR #4892, opened 2025-07-26
dialects: (linalg) Broadcast and transpose op init methods take sequence of result types

mesham:
This is a bug fix for the broadcast and transpose linalg methods, where the result type is varadic but the default result init method is None, which results in an error. This therefore fixes that by defaulting to an empty list.

This also ensures these two operations match MLIR behaviour which is when the outs type is a tensor then the result type of the operation is a tensor, otherwise if the outs type is a memref then there is no result type. The parsing routing did not handle this, so this nuance is now added (the verification already picked this up, as the only result type can be a tensor). This meant we could not use memrefs as output from these operations when passed to MLIR, which we can now.

I have updated the filecheck test to include no result type with memref as outs type

### DIFF
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -936,8 +936,14 @@ class TransposeOp(IRDLOperation):
         input: SSAValue,
         init: SSAValue,
         permutation: Attribute,
-        result: Sequence[Attribute] | None = [],
+        result: Attribute | None = None,
     ):
+        if result is None:
+            if isa(input.type, TensorType):
+                result = (input.type,)
+            else:
+                result = ()
+
         arg_types = NamedOperation.body_arg_types((input, init))
 
         @Builder.implicit_region(arg_types)
@@ -1021,7 +1027,7 @@ class TransposeOp(IRDLOperation):
             input,
             init,
             DenseArrayBase.from_list(i64, permutation),
-            [result] if isa(result, TensorType) else [],
+            (result,) if isa(result, TensorType) else None,
         )
         return transpose
 
@@ -1302,8 +1308,14 @@ class BroadcastOp(IRDLOperation):
         input: SSAValue,
         init: SSAValue,
         dimensions: Attribute,
-        result: Sequence[Attribute] | None = [],
+        result: Attribute | None = None,
     ):
+        if result is None:
+            if isa(input.type, TensorType):
+                result = (input.type,)
+            else:
+                result = ()
+
         arg_types = NamedOperation.body_arg_types((input, init))
 
         @Builder.implicit_region(arg_types)
@@ -1392,7 +1404,7 @@ class BroadcastOp(IRDLOperation):
             input,
             init,
             DenseArrayBase.from_list(i64, dimensions),
-            [result] if isa(result, TensorType) else [],
+            (result,) if isa(result, TensorType) else None,
         )
         return broadcast
 
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -1018,7 +1018,7 @@ class TransposeOp(IRDLOperation):
         parser.parse_punctuation("(")
         init = parser.parse_operand()
         parser.parse_punctuation(":")
-        result = parser.parse_type()
+        parser.parse_type()
         parser.parse_punctuation(")")
         parser.parse_keyword("permutation")
         parser.parse_punctuation("=")
@@ -1029,7 +1029,6 @@ class TransposeOp(IRDLOperation):
             input,
             init,
             DenseArrayBase.from_list(i64, permutation),
-            result if isa(result, TensorType) else None,
         )
         return transpose
 
@@ -1397,7 +1396,7 @@ class BroadcastOp(IRDLOperation):
         parser.parse_punctuation("(")
         init = parser.parse_operand()
         parser.parse_punctuation(":")
-        result = parser.parse_type()
+        parser.parse_type()
         parser.parse_punctuation(")")
         parser.parse_keyword("dimensions")
         parser.parse_punctuation("=")
@@ -1408,7 +1407,6 @@ class BroadcastOp(IRDLOperation):
             input,
             init,
             DenseArrayBase.from_list(i64, dimensions),
-            result if isa(result, TensorType) else None,
         )
         return broadcast
 
--- a/xdsl/dialects/linalg.py
+++ b/xdsl/dialects/linalg.py
@@ -943,6 +943,8 @@ class TransposeOp(IRDLOperation):
                 results = (input.type,)
             else:
                 results = ()
+        else:
+            results = (result,)
 
         arg_types = NamedOperation.body_arg_types((input, init))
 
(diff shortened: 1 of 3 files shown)


xdsl/dialects/linalg.py
```diff
@@ -936,7 +936,7 @@ def __init__(
         input: SSAValue,
         init: SSAValue,
         permutation: Attribute,
-        result: Attribute | None = None,
+        result: Sequence[Attribute] | None = [],
     ):
         arg_types = NamedOperation.body_arg_types((input, init))
```
superlopuh: I would much rather keep the default value None as then it's unambiguously unspecified, whereas the empty list could be provided by the user or the default. One consideration is that if the user did not specify the result then we can actually infer it in most cases. I would recommend inferring it here, for example:
```suggestion
        result: Attribute | None = None,
    ):
    if result is None:
        if isa(input.type, TensorType):
            result = (input.type,)
        else:
            result = ()
    arg_types = NamedOperation.body_arg_types((input, init))
```

superlopuh: WIth this I think the parsing changes can be safely reverted, if you're happy with this suggestion.

mesham: > WIth this I think the parsing changes can be safely reverted, if you're happy with this suggestion.

Thanks :) I've followed this for the init method, but I think we do need the parser code still. The reason is, the following MLIR will not parse with mlir-opt (error is cannot name an operation with no results error):

`%51 = linalg.transpose ins(%49 : memref<16x64xf32>) outs(%50 : memref<64x16xf32>) permutation = [1, 0]`

Whereas for tensortype the return type is needed (which makes sense, as tensors are immutable). Therefore the parser needs to pass None as return type for result=memref, and TensorType for result=tensor when _result_ is parsed.

I've put an explicit test in the parser for this, I suppose it could just pass None regardless and the init method would sort it (as the input type will be TensorType) - I think it's a bit clearer making this explicit but can just pass None there regardless of _result_ if you would prefer :)

xdsl/dialects/linalg.py
```diff
@@ -938,6 +938,12 @@ def __init__(
         permutation: Attribute,
         result: Attribute | None = None,
     ):
+        if result is None:
+            if isa(input.type, TensorType):
+                results = (input.type,)
+            else:
+                results = ()
+
```
alexarice, line 948:
```suggestion
else:
    results = (result,)

```
This might fix the errors?

xdsl/dialects/linalg.py
```diff
@@ -1021,7 +1027,7 @@ def parse(cls, parser: Parser) -> Self:
             input,
             init,
             DenseArrayBase.from_list(i64, permutation),
-            result,
+            (result,) if isa(result, TensorType) else None,
```
alexarice: Shouldn't need to change anything here?

xdsl/dialects/linalg.py
mesham: Thanks, yup I realised I missed this :)

xdsl/dialects/linalg.py
mesham: Similarly to my comment above, I think we do need something here to swallow memref type (pass as None) as mlir-opt requires a return type for tensor for this operation, but requires there is not a return type for memref. I could pass None regardless, and the type would be handled but I think an explicit check is a bit clearer (but happy to do whatever people think is best)

superlopuh (approved):
Nice, thank you

xdsl/dialects/linalg.py
```diff
@@ -1021,7 +1029,7 @@ def parse(cls, parser: Parser) -> Self:
             input,
             init,
             DenseArrayBase.from_list(i64, permutation),
-            result,
+            result if isa(result, TensorType) else None,
```
alexarice: This still doesn't make sense to me, why input `None` if `result` is a non-tensor type?

mesham: Thanks - MLIR handles these operations differently depending on if we pass memref or tensor as operands, for a tensor there is a return type (I guess because tensors are immutable) whereas when passing memref there is no return type. Therefore the following results in errors when parsed by MLIR:

`%res = linalg.transpose ins(%49 : memref<16x64xf32>) outs(%50 : memref<64x16xf32>) permutation = [1, 0]`
`linalg.transpose ins(%49 : tensor<16x64xf32>) outs(%50 : tensor<64x16xf32>) permutation = [1, 0]`

Whereas the following are correct:

`linalg.transpose ins(%49 : memref<16x64xf32>) outs(%50 : memref<64x16xf32>) permutation = [1, 0]`
`%res = linalg.transpose ins(%49 : tensor<16x64xf32>) outs(%50 : tensor<64x16xf32>) permutation = [1, 0]`

Furthermore, in the verify method the operation correctly ensures that if a return type is present then it is a tensor (a return type of memref is an error). So when parsing _result_ (which is the type of _init_ ), this needs to be passed as the return type if it is a tensor, and otherwise None (denoting no return type) needs to be passed. If we don't, and it's a memref, then this it then results in a verify error in xDSL, and the output MLIR fails mlir-opt .

alexarice: Thanks, I see now, in this case is it possible to always pass `None` here? It feels like this argument is never necessary to me.

superlopuh: Then we might silently change the return type when parsing if for whatever reason it's incorrect. Not sure how big a deal that is.

alexarice: > Then we might silently change the return type when parsing if for whatever reason it's incorrect. Not sure how big a deal that is.

I'm not sure what this means, the return type doesn't appear in the parsed input

xdsl/dialects/linalg.py
```diff
@@ -1304,6 +1311,14 @@ def __init__(
         dimensions: Attribute,
         result: Attribute | None = None,
```
alexarice, line 1312: I think I'd prefer that the `result` argument wasn't there, but I'd be happy to leave that for this PR, feel I've done enough nitting.

alexarice (approved):

xdsl/dialects/linalg.py
```diff
@@ -938,6 +938,14 @@ def __init__(
         permutation: Attribute,
         result: Attribute | None = None,
     ):
+        if result is None:
+            if isa(input.type, TensorType):
+                results = (input.type,)
```
alexarice: This should be `init.type` right?

alexarice (approved):
Thanks, I'm now happy if CI is green

xdsl/dialects/linalg.py
mesham: Thanks, I've had a chat with Alex about this via Zulip - we have agreed to remove passing the return type to the _init_ method from the parse method and for _init_ to handle it :) I've made the changes here

xdsl/dialects/linalg.py
mesham: Yup, one of the tests picked this up - i've fixed that now

PR #4893, opened 2025-07-26
core: (mlir) Add support for unicode parsing

Jimmy2027:
Still wip, not sure if this is the right way to do this. Needs to be improved to handle longer unicode sequences


Jimmy2027:
yeah actually you're right, I think we only need to handle ascii:
https://godbolt.org/z/MvdWbM1hK

Jimmy2027:
yeah probably, that does look neater. Will have a look!

Jimmy2027:
looks better now, reusing the functionality of StringLiteral :)

xdsl/utils/mlir_lexer.py
```diff
@@ -478,57 +478,35 @@ def _lex_prefixed_ident(self, start_pos: Position) -> MLIRToken:
...
+        if span.text == '""':
+            return MLIRToken(MLIRTokenKind.STRING_LIT, span)  # empty string literal
 
-            # newline character in string literal (not allowed)
-            if current_char in ["\n", "\v", "\f"]:
-                raise ParseError(
-                    Span(start_pos, self.pos, self.input),
-                    "Newline character not allowed in string literal.",
-                )
+        lit = StringLiteral.from_span(span)
```
superlopuh: I'm worried about the performance impact of this, could you please measure perf before/after? Ideally we'd avoid actually creating the string, and would just calculate the end index instead.

Jimmy2027: yeah it does get considerably slower.
on main:
```
Test Lexer.empty_program ran in: 0.00007s
Test Lexer.constant_100 ran in: 0.00180s
Test Lexer.constant_1000 ran in: 0.01695s
Test Lexer.dense_attr ran in: 0.00671s
Test Lexer.dense_attr_hex ran in: 0.00678s
```
on #4893
```
Test Lexer.empty_program ran in: 0.00007s
Test Lexer.constant_100 ran in: 0.00187s
Test Lexer.constant_1000 ran in: 0.01755s
Test Lexer.dense_attr ran in: 0.16036s
Test Lexer.dense_attr_hex ran in: 0.15370s
```
Not sure where that comes from though, the string is also recreated in main with the regex, or am I missing something?

superlopuh: This particular line is quite expensive, as it transcodes the string. In main it just checks that it's OK. Is it possible to be smart enough with the regex to one-shot the verification?

Jimmy2027: hmm do you mean the check in the __post_init__ of the StringLiteral? Other than that I don't see much happening here, maybe I should create a StringLiteral instead of creating a Span first... Isn't the more expensive part l. 508?
https://github.com/xdslproject/xdsl/pull/4893/files#diff-3a31e53ac5c7f21a3bcc90c9fb062055818c105269b8c7f4e04f8d5156a2c0e7R508

superlopuh: Sorry I mean what we do here is skip all the valid characters in a string literal, until either the next unescaped `"` or an invelid character such as a newline. Would it be possible to express this in a regular expression directly? The case you care about is `\\\d\d` no? I wonder if there's a way to just throw more regex at the problem.

superlopuh: it's quite easy to verify what is expensive by running 

```
python benchmarks/lexer.py Lexer.dense_attr_hex viztracer
```

Jimmy2027: I did some quick manual experimentation and actually what seems to be slowing down the benchmark is the new regex. Will have a look at the viztracer as well

Jimmy2027: > Sorry I mean what we do here is skip all the valid characters in a string literal, until either the next unescaped `"` or an invelid character such as a newline. Would it be possible to express this in a regular expression directly? The case you care about is `\\\d\d` no? I wonder if there's a way to just throw more regex at the problem.

hmm yeah true, let me think about it

Jimmy2027: I think this regex should do that `"(?:\\(?:["nt\\]|[0-9A-Fa-f]{2})|[^"\\\n\v\f])*"` but the slowdown seems to come from the regex pattern match, the benchmark doesn't even hit l. 505

Jimmy2027: updated the regex in 966e09e8613865f216bab22f8208be6a68546606, change in the benchmark is non-significant now:
```
Test Lexer.empty_program ran in: 0.00007s
Test Lexer.constant_100 ran in: 0.00178s
Test Lexer.constant_1000 ran in: 0.01690s
Test Lexer.dense_attr ran in: 0.00629s
Test Lexer.dense_attr_hex ran in: 0.00633s
```

superlopuh: Nice! I'll play around with this locally soon

PR #4895, opened 2025-07-26
core: move default_value into AttributeVariable directive

superlopuh:
Makes the implementation a bit more straightforward, and makes some upcoming factoring a little easier.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1186,10 +1186,7 @@ class AttributeVariable(FormatDirective):
         raise ValueError("Attributes must be Data or ParameterizedAttribute!")
 
     def is_present(self, op: IRDLOperation) -> bool:
-        if self.is_property:
-            attr = op.properties.get(self.name)
-        else:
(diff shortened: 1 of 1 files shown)


math-fehr (approved):

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1184,17 +1185,6 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
-    """
-    An attribute variable with default value, with the following format:
-      result-directive ::= dollar-ident
-    The directive will request a space to be printed right after.
-    """
-
-    default_value: Attribute
-
     def is_present(self, op: IRDLOperation) -> bool:
         if self.is_property:
```
alexarice: Would you be happy to switch this to `self.get` too?

alexarice (approved):

xdsl/irdl/declarative_assembly_format.py
alexarice: Don't mind if this is a different PR

PR #4898, opened 2025-07-26
CI: use --allow-unregistered-dialect in LLVM attrs test

superlopuh:
I can't figure out hoew go get the CI to catch this.


math-fehr:
Oh wait that's my bad, but how does this still pass the CI currently?

superlopuh:
The problem is that the version distributed via apt seems to include test.op functionality, whereas nix and brew ones don't.

PR #4899, opened 2025-07-26
dialects: Add missing llvm.func properties

math-fehr:
These are the properties that are required to parse a simple C++ file that was passed to `clang` then `mlir-translate`.

In particular, these properties are not verified, and are currently just added to allow roundtripping of the IR.

### DIFF
--- a/tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/func.mlir
+++ b/tests/filecheck/mlir-conversion/with-mlir/dialects/llvm/func.mlir
@@ -27,22 +27,22 @@
 
 // CHECK:       builtin.module {
 // CHECK-NEXT:    "llvm.func"() <{
-// CHECK:            CConv = #llvm.cconv<ccc>,
-// CHECK:            arg_attrs = [{llvm.noundef}, {llvm.noundef}],
-// CHECK:            frame_pointer = #llvm.framePointerKind<"non-leaf">,
-// CHECK:            function_type = !llvm.func<i32 (i32, i32)>,
-// CHECK:            linkage = #llvm.linkage<"external">,
-// CHECK:            no_inline,
-// CHECK:            no_unwind,
-// CHECK:            optimize_none,
-// CHECK:            passthrough = [["no-trapping-math", "true"]],
-// CHECK:            sym_name = "add",
-// CHECK:            target_cpu = "x86-64",
-// CHECK:            target_features = #llvm.target_features<["+mmx"]>,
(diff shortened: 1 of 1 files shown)


tests/filecheck/dialects/llvm/func.mlir
```diff
@@ -0,0 +1,49 @@
...
+// CHECK:            no_unwind,
+// CHECK:            optimize_none,
+// CHECK:            passthrough = [["no-trapping-math", "true"]],
+// CHECK:            sym_name = "add",
+// CHECK:            target_cpu = "x86-64",
+// CHECK:            target_features = #llvm.target_features<["+mmx"]>,
+// CHECK:            tune_cpu = "generic",
+// CHECK:            unnamed_addr = 0 : i64,
+// CHECK:            visibility_ = 0 : i64
+// CHECK:         }> ({
```
alexarice, line 45: What's with the check-next vs check here? do they get reordered?

xdsl/dialects/llvm.py
```diff
@@ -1563,6 +1577,7 @@ def __init__(
         visibility: int | IntegerAttr[IntegerType] = 0,
         sym_visibility: str | StringAttr | None = None,
         body: Region | None = None,
+        other_props: dict[str, Attribute | None] | None = None,
```
alexarice, line 1580:
```suggestion
        other_props: dict[str, Attribute] | None = None,
```
don't see the point in the `None`. it could just be omitted from the dictionary

alexarice: Could make it keyword variadic too if you wanted.

alexarice (approved):

tests/filecheck/dialects/llvm/func.mlir
math-fehr: Actually I should check if `CHECK-SAME` works with this filecheck version.
Otherwise, `CHECK` will check for the next occurence, so it will stay at the same line. `CHECK-NEXT` will check that the occurence will be on the next line.

superlopuh: We use CHECK-SAME in places

xdsl/dialects/llvm.py
superlopuh: I don't have a very strong view on this but it can be quite convenient to not have to do the whole `if thing is not None: props["thing"] = thing`

superlopuh (approved):

PR #4900, opened 2025-07-26
core: factor out parse_attr in AttributeVariable directive

superlopuh:
This makes it easier to specialise subclasses of AttributeVariable.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1132,7 +1132,7 @@ class AttributeVariable(FormatDirective):
         else:
             state.attributes[self.name] = attr
 
-    def parse(self, parser: Parser, state: ParsingState) -> bool:
+    def parse_attr(self, parser: Parser) -> Attribute | None:
         unique_base = self.unique_base
 
         if self.is_symbol_name:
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1200,29 +1211,16 @@ class OptionalAttributeVariable(AttributeVariable):
...
-        if attr is None:
-            return False
-        if self.is_property:
-            state.properties[self.name] = attr
+            return parser.parse_optional_attribute()
         else:
-            state.attributes[self.name] = attr
-        return True
-
-    def is_present(self, op: IRDLOperation) -> bool:
```
alexarice, line 1221: why remove this?

superlopuh: an error, thanks for catching this

superlopuh: Ah no sorry not an error it's already present in the superclass, I moved this definition into the superclass as a fix

alexarice: It's not quite the same though right?

alexarice: Oh I see you've changed the superclass definition

alexarice (approved):

PR #4903, opened 2025-07-27
core: factor out UniqueBase and UniqueTypeAttributeVariable directives

superlopuh:
Thought I'd do these two in one go.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1118,13 +1118,9 @@ class AttributeVariable(FormatDirective):
     """The attribute name as it should be in the attribute or property dictionary."""
     is_property: bool
     """Should this attribute be put in the attribute or property dictionary."""
-    unique_base: type[Attribute] | None
-    """The known base class of the Attribute, if any."""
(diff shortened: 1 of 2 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1118,13 +1118,9 @@ class AttributeVariable(FormatDirective):
...
     is_property: bool
     """Should this attribute be put in the attribute or property dictionary."""
-    unique_base: type[Attribute] | None
-    """The known base class of the Attribute, if any."""
-    unique_type: Attribute | None
-    """The known type of the Attribute, if any."""
     is_optional: bool
     """Is this attribute optional in the operation definition."""
-    default_value: Attribute | None = None
+    default_value: Attribute | None
```
alexarice, line 1123: Is this a drive-by change?

superlopuh: Not really, although I guess I could have decoupled it, but you can't have function arguments without default values after arguments with default values, and the automatically generated inits are in order from superclass to subclass, so moving unique_base and unique_type into subclasses made providing this default value impossible.

PR #4881, opened 2025-07-25
dialects: (tosa) add support for `tosa.sub/mul`

jakedves:
Adds support and test cases for the `tosa.sub` and `tosa.mul` operations

PR #4904, opened 2025-07-27
dialects: (tosa) implement trig operations

jakedves:
Adds operations for `tosa.cos` and `tosa.sin` and a positive roundtrip test case

### DIFF
--- a/xdsl/dialects/tosa.py
+++ b/xdsl/dialects/tosa.py
@@ -5,6 +5,7 @@ from xdsl.dialects.builtin import (
     I32,
     I64,
     AnyAttr,
+    AnyFloatConstr,
     BoolAttr,
     DenseArrayBase,
     FloatAttr,
@@ -181,6 +182,45 @@ class MulOp(ElementwiseBinaryOperation):
     )
 
 
+class ElementwiseTrigOperation(IRDLOperation, ABC):
+    """
+    Abstract base class for elementwise trig operations on tensors of floating-point types
+    """
+
+    T: ClassVar = VarConstraint("T", AnyFloatConstr)
+
+    input1 = operand_def(TensorType.constr(T))
+    result = result_def(TensorType.constr(T))
+
+    traits = traits_def(
+        Pure(),
+    )
+
+    assembly_format = "operands attr-dict `:` functional-type(operands, results)"
+
+
+@irdl_op_definition
+class SinOp(ElementwiseTrigOperation):
+    """
+    TOSA dialect operation computing sin(x) for each element in a tensor
+
+    See external [documentation](https://mlir.llvm.org/docs/Dialects/TOSA/#tosasin-mlirtosasinop)
+    """
+
+    name = "tosa.sin"
+
+
+@irdl_op_definition
+class CosOp(ElementwiseTrigOperation):
+    """
+    TOSA dialect operation computing cos(x) for each element in a tensor
+
+    See external [documentation](https://mlir.llvm.org/docs/Dialects/TOSA/#tosacos-mlirtosacosop)
+    """
+
+    name = "tosa.cos"
+
+
 TOSA = Dialect(
     "tosa",
     [
@@ -189,6 +229,8 @@ TOSA = Dialect(
         AddOp,
         SubOp,
         MulOp,
+        SinOp,
+        CosOp,
     ],
     [],
 )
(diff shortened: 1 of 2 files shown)


alexarice (commented):
MLIR Tosa seems to have the following structure:
- Tosa_ElementwiseOp is the superclass for everything you have added so far and has the assembly format
- Tosa_ElementwiseUnaryOp is the superclass of these two functions
- AddOp inherits directly from Tosa_ElementwiseOp

Did you have a thoughts on what structure you wanted to take here wrt class hierarchy? I feel a nice structure could be:

- ElementwiseOperation as the superclass for everything here, which could have the assembly format and possibly a shared verification
- ElementwiseUnaryOperation in place of ElementWiseTrigOperation
- ElementwiseBinaryOperation as the superclass for AddOp and friends

@superlopuh do you have any thoughts here?

jakedves:
That sounds good to me, hadn't looked at the MLIR implementation but the reason I made a separate class for trig operations is because of the description of the inputs:

`input1: tosa-conformant tensor of floating-point values`

vs for the other unary operations:

`input1: tosa-conformant tensor of number values`

Wasn't sure how that difference should be handled in xDSL, but in the MLIR they have:

```
def Tosa_FloatTensor : TosaTensorOf<[AnyFloat]>;

...
let arguments = (ins
    Tosa_FloatTensor:$input1
);
```

alexarice:
By the way, I'm not saying we should necessarily follow MLIR's class structure blindly, but it seems to me in cases where we've ended up with roughly the same abstractions it could be nice to have them share the name so that someone trying to follow the xDSL and MLIR implementations can link them up. On the other hand if we have good reasons to go with a different structure then that's absolutely fine from my end.

superlopuh:
@jakedves might be a good use-case for generics kind of how we've used them in the x86 dialect? The subclasses can specialise the type when specifying the generic arguments of the abstract superclass.

superlopuh (approved):
I like it

xdsl/dialects/tosa.py
```diff
@@ -176,11 +185,50 @@ class MulOp(ElementwiseBinaryOperation):
...
+
+    input1 = operand_def(TInv)
+    result = result_def(TInv)
+
+    traits = traits_def(
+        Pure(),
+    )
+
+    assembly_format = "operands attr-dict `:` functional-type(operands, results)"
+
```
```suggestion
```
This is all part of the superclass

xdsl/dialects/tosa.py
```diff
@@ -176,11 +185,50 @@ class MulOp(ElementwiseBinaryOperation):
     name = "tosa.mul"
 
     traits = traits_def(
```
alexarice, line 187: do traits compose with superclass traits like this? (Genuine question I do not know)

xdsl/dialects/tosa.py
```diff
@@ -176,11 +185,50 @@ class MulOp(ElementwiseBinaryOperation):
...
+
+    traits = traits_def(
+        Pure(),
+    )
+
+    assembly_format = "operands attr-dict `:` functional-type(operands, results)"
+
+
+@irdl_op_definition
+class SinOp(ElementwiseUnaryOperation[TensorType[AnyFloat]]):
```
alexarice, line 205: Are we expecting operations where the inputs are not `TensorType[AnyFloat]`?

jakedves: Other unary ops e.g. [`tosa.abs`](https://mlir.llvm.org/docs/Dialects/TOSA/#tosaabs-mlirtosaabsop) work on "tensors of number value"

superlopuh: I would probably do that bit when we add these operations

xdsl/dialects/tosa.py
jakedves: I believe yes [as per here](https://github.com/xdslproject/xdsl/pull/4881#discussion_r2231422773)

alexarice: Great

xdsl/dialects/tosa.py
jakedves: Should I remove it for now then?

alexarice: I'm happy to leave it, I was just checking that we weren't prematurely overgeneralising. (Maybe @superlopuh feels differently)

superlopuh: I think now it's in let's leave it, implementation looks good.

PR #4905, opened 2025-07-27
[WIP] core: add generic arguments for IntegerType width and signedness

superlopuh:
Note that this isn't actually how I would like to implement this, but more of a proof of concept that it can be done.

Importantly it works without any kind of EnumConstraint which makes me think that it's safe to remove the IntConstraint and that we'll find a way to map the type var for int/Literal to an AttrConstraint directly.


superlopuh:
Merged!

PR #4906, opened 2025-07-28
core: Abstract `uses` field from `IRWithUses`

math-fehr:
Stacked PRs:
 * #4909
 * #4908
 * #4907
 * __->__#4906

--- --- ---

### core: Abstract `uses` field from `IRWithUses`

This is a refactor which will make it easier to make `IRWithUses` use
a linked list.

### DIFF
--- a/tests/test_use.py
+++ b/tests/test_use.py
@@ -24,8 +24,8 @@ def test_main():
     assert isinstance(module, ModuleOp)
 
     op1, op2 = list(module.ops)
-    assert op1.results[0].uses == {Use(op2, 0), Use(op2, 1)}
-    assert op2.results[0].uses == set()
+    assert set(op1.results[0].uses) == {Use(op2, 0), Use(op2, 1)}
+    assert not op2.results[0].uses
 
     print("Done")
 
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -536,21 +536,45 @@ class Use:
     """The index of the operand using the value in the operation."""
 
 
+@dataclass
+class IRUses(Iterable[Use]):
+    """
+    Multi-pass iterable of the uses of an IR value (SSAValue or Block).
+    """
+
+    ir: IRWithUses
+
+    def __iter__(self):
+        return self.ir._uses.__iter__()  # pyright: ignore[reportPrivateUsage]
+
+    def __len__(self):
+        return len(self.ir._uses)  # pyright: ignore[reportPrivateUsage]
+
+    def __bool__(self) -> bool:
+        """Returns `True` if there are operations in this block."""
+        return bool(self.ir._uses)  # pyright: ignore[reportPrivateUsage]
+
+
 @dataclass(eq=False)
 class IRWithUses(ABC):
     """IRNode which stores a list of its uses."""
 
-    uses: set[Use] = field(init=False, default_factory=set[Use], repr=False)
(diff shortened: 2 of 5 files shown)


superlopuh (approved):

tests/test_use.py
```diff
@@ -24,8 +24,8 @@ def test_main():
     assert isinstance(module, ModuleOp)
 
     op1, op2 = list(module.ops)
-    assert op1.results[0].uses == {Use(op2, 0), Use(op2, 1)}
-    assert op2.results[0].uses == set()
+    assert set(op1.results[0].uses) == {Use(op2, 0), Use(op2, 1)}
+    assert set(op2.results[0].uses) == set()
```
alexarice:
```suggestion
    assert not op2.results[0].uses
```
Tiny nit?

xdsl/ir/core.py
```diff
@@ -536,21 +536,52 @@ class Use:
...
+    def __bool__(self) -> bool:
+        """Returns `True` if there are operations in this block."""
+        return bool(self.ir._uses)  # pyright: ignore[reportPrivateUsage]
+
+    def copy(self) -> Iterable[Use]:
+        """
+        Returns a copy of the uses.
+        This is useful to avoid modifying the original uses while iterating over them.
+        """
+        return self.ir._uses.copy()  # pyright: ignore[reportPrivateUsage]
```
alexarice: Is this correct? Does the output not need to be wrapped in some sort of `IRUses` object?

xdsl/ir/core.py
```diff
@@ -536,21 +536,52 @@ class Use:
     """The index of the operand using the value in the operation."""
 
 
+@dataclass
+class IRUses(Iterable[Use]):
+    """
+    Multi-pass iterable of the uses of an IR value (SSAValue or Block).
+    """
+
+    ir: IRWithUses
```
alexarice, line 545: Maybe it will become more apparent in further PRs but why take the IR here instead of the set.

alexarice (approved):

xdsl/ir/core.py
superlopuh: TBH it might be more straightforward and more like the end result to let clients build a tuple themselves rather than call copy

math-fehr: Fine by me, but should I do this as a last commit in the stack, so this commit does not change any other code?

xdsl/ir/core.py
math-fehr: Yes!

alexarice: Is this "yes it will become clear"?

xdsl/ir/core.py
alexarice: I think it would be nicer to do here but if it's a tonne of work then I don't mind it being at the end

math-fehr: Lemme try that here actually

xdsl/ir/core.py
math-fehr: Actually wait, it might be unrelated.
So the `set` disappears after, in favor of a pointer to the next use.

But `BlockOps` also refers directly to `block` instead of the operation list, so I'm not sure if we should change both or not if we want that?

alexarice: If it's going to disappear anyway then this seems fine

xdsl/ir/core.py
math-fehr: I added this in that PR

PR #4907, opened 2025-07-28
core: Add useful methods to `IRWithUses` class

math-fehr:
Stacked PRs:
 * #4909
 * #4908
 * __->__#4907

--- --- ---

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -554,6 +554,13 @@ class IRUses(Iterable[Use]):
         """Returns `True` if there are operations in this block."""
         return bool(self.ir._uses)  # pyright: ignore[reportPrivateUsage]
 
+    def get_length(self) -> int:
+        """
+        Returns the number of uses.
+        We do not expose it as `__len__` as it is expensive to compute `O(len)`.
+        """
+        return len(self.ir._uses)  # pyright: ignore[reportPrivateUsage]
+
 
 @dataclass(eq=False)
 class IRWithUses(ABC):
@@ -575,6 +582,32 @@ class IRWithUses(ABC):
         """Remove a use of the value."""
         assert use in self._uses, "use to be removed was not in use list"
         self._uses.remove(use)
+
+    def has_one_use(self) -> bool:
+        """Returns true if the value has exactly one use."""
+        return len(self._uses) == 1
+
+    def has_more_than_one_use(self) -> bool:
+        """Returns true if the value has more than one use."""
+        return len(self._uses) > 1
+
+    def get_unique_use(self) -> Use | None:
+        """
+        Returns the single use of the value, or None if there are no uses or
+        more than one use.
+        """
+        if self.has_one_use():
+            return next(iter(self._uses))
+        return None
+
+    def get_user_of_unique_use(self) -> Operation | None:
+        """
+        Returns the user of the single use of the value.
+        If there are no uses or more than one use, returns None.
+        """
+        if (use := self.get_unique_use()) is not None:
+            return use.operation
+        return None
 
 
 @dataclass(eq=False)
@@ -679,7 +712,7 @@ class SSAValue(Generic[AttributeCovT], IRWithUses, ABC):
         If safe_erase is True, then check that no operations use the value anymore.
         If safe_erase is False, then replace its uses by an ErasedSSAValue.
         """
-        if safe_erase and len(self.uses) != 0:
+        if safe_erase and self.uses:
             raise ValueError(
                 "Attempting to delete SSA value that still has uses of result "
                 f"of operation:\n{self.owner}"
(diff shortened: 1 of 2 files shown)


superlopuh (approved):

xdsl/ir/core.py
```diff
@@ -583,6 +587,32 @@ def remove_use(self, use: Use):
         assert use in self._uses, "use to be removed was not in use list"
         self._uses.remove(use)
 
+    def has_single_use(self) -> bool:
+        """Returns true if the value has exactly one use."""
+        return len(self._uses) == 1
+
+    def get_single_use(self) -> Use | None:
```
superlopuh: What do you think about this name instead? Feels like it's a bit more obvious what's going on.
```suggestion
    def get_only_use(self) -> Use | None:
```

alexarice: I like `get_only_use`. I also like `get_unique_use`

xdsl/ir/core.py
```diff
@@ -583,6 +587,32 @@ def remove_use(self, use: Use):
...
+        """
+        Returns the single use of the value, or None if there are no uses or
+        more than one use.
+        """
+        if len(self._uses) == 1:
```
```suggestion
        if self.has_single_use():
```

xdsl/ir/core.py
alexarice: I like that the name matches up with `has_single_use` though. Maybe `has_unique_use` and `get_unique_use` would work?

xdsl/ir/core.py
```diff
@@ -583,6 +587,32 @@ def remove_use(self, use: Use):
...
+            return next(iter(self._uses))
+        return None
+
+    def get_single_use_user(self) -> Operation | None:
+        """
+        Returns the user of the single use of the value.
+        If there are no uses or more than one use, returns None.
+        """
+        if len(self._uses) == 1:
+            return next(iter(self._uses)).operation
```
```suggestion
        if (use := self.get_single_use()) is not None:
            return use.operation
```
Tbh this helper function seems of questionable utility to me.

alexarice (approved):
A lot of tiny comments.

xdsl/ir/core.py
```diff
@@ -547,9 +547,6 @@ class IRUses(Iterable[Use]):
     def __iter__(self):
         return self.ir._uses.__iter__()  # pyright: ignore[reportPrivateUsage]
 
-    def __len__(self):
```
alexarice: Ah, should this be in the next PR?

xdsl/ir/core.py
math-fehr: Would you both be fine with `has_one_use` (following MLIR) https://mlir.llvm.org/doxygen/classmlir_1_1Value.html and `get_unique_use`?
It wouldn't be the same, but at least we would follow MLIR standards there

alexarice: I'd be fine with that

xdsl/ir/core.py
math-fehr: That's because there are so many places in the code where you write
```
if (use := x.get_single_use()) is not None and isinstance(use.operation, MyOp)
```
Which we can really easily shorten to
```
if isinstance(x.get_single_use_user(), MyOp)
```
which I believe is much nicer to read/use (because the lack of monad-like syntax in Python

alexarice: Imagine having to use a language without monads.

superlopuh: 🌯

xdsl/ir/core.py
math-fehr: Thanks, it's fixed now

xdsl/ir/core.py
```diff
@@ -576,6 +583,32 @@ def remove_use(self, use: Use):
...
+    def get_unique_use(self) -> Use | None:
+        """
+        Returns the single use of the value, or None if there are no uses or
+        more than one use.
+        """
+        if self.has_one_use():
+            return next(iter(self._uses))
+        return None
+
+    def get_unique_use_user(self) -> Operation | None:
```
superlopuh: Surely `get_unique_user`? You don't use a use?

alexarice: You can have a unique user without unique use

math-fehr: What I want to convey is that this is not a unique user, but a unique use, and then get the user.
Because `get_unique_user` would (to me), mean that you may have multiple uses, as long as they have only a single user.
So I'm not sure how to rename this correctly

alexarice: `get_user_of_unique_use` is the most explicit

math-fehr: That's fine by me

superlopuh: I think that reads much better

xdsl/ir/core.py
alexarice: I got distracted by the monad discussion, I still stand by the code suggestion

alexarice: (I guess this code gets nuked later though)

superlopuh: Yeah I would not implement it either way once the implementation is changed

math-fehr: I changed it, but yeah it's getting nuked in 2 commits anyway

superlopuh (approved):

PR #4908, opened 2025-07-28
core: Remove __len__ from IRUses

math-fehr:
Stacked PRs:
 * #4909
 * __->__#4908
 * #4907

--- --- ---

### core: Remove __len__ from IRUses

__len__ is costly to compute (since it will be moved to a linked list),
so it should not be easy to call.

### DIFF
--- a/tests/test_ir.py
+++ b/tests/test_ir.py
@@ -161,8 +161,8 @@ def test_op_operands_assign():
     op.operands = [val2, val1]
     op.verify()
 
-    assert len(val1.uses) == 1
-    assert len(val2.uses) == 1
+    assert val1.has_one_use()
+    assert val2.has_one_use()
     assert tuple(op.operands) == (val2, val1)
 
 
@@ -180,8 +180,8 @@ def test_op_operands_indexing():
     op.operands[0] = val2
     op.verify()
 
-    assert len(val1.uses) == 0
-    assert len(val2.uses) == 2
+    assert not list(val1.uses)
+    assert len(list(val2.uses)) == 2
     assert tuple(op.operands) == (val2, val2)
 
 
--- a/xdsl/transforms/canonicalization_patterns/cf.py
+++ b/xdsl/transforms/canonicalization_patterns/cf.py
@@ -516,10 +516,8 @@ class SimplifySwitchFromSwitchOnSameCondition(RewritePattern):
         block = op.parent_block()
         if block is None:
             return
-        preds = block.uses
-        if len(preds) != 1:
+        if (pred := block.get_unique_use()) is None:
             return
-        pred = next(iter(preds))
         switch = pred.operation
         if not isinstance(switch, cf.SwitchOp):
             return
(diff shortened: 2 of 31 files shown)


tests/test_ir.py
```diff
@@ -161,8 +161,8 @@ def test_op_operands_assign():
     op.operands = [val2, val1]
     op.verify()
 
-    assert len(val1.uses) == 1
-    assert len(val2.uses) == 1
+    assert len(list(val1.uses)) == 1
```
alexarice: Why are we preferring `len(list(*.uses))` to `*.get_length()`

alexarice: Also this one should be `has_single_use` anyway

math-fehr: Yes, fixed!

tests/test_ir.py
```diff
@@ -180,8 +180,8 @@ def test_op_operands_indexing():
     op.operands[0] = val2
     op.verify()
 
-    assert len(val1.uses) == 0
-    assert len(val2.uses) == 2
+    assert len(list(val1.uses)) == 0
```
alexarice:
```suggestion
    assert not val1.uses
```

tests/test_ir.py
```diff
@@ -180,8 +180,8 @@ def test_op_operands_indexing():
...
+    assert len(list(val2.uses)) == 2
```
alexarice, line 184:
```suggestion
    assert val2.uses.has_more_than_one_use()
```

xdsl/transforms/canonicalization_patterns/cf.py
```diff
@@ -98,9 +98,11 @@ def collapse_branch(
...
 
     new_operands = tuple(
-        successor_operands[operand.index]
-        if isinstance(operand, BlockArgument) and operand.owner is successor
-        else operand
+        (
+            successor_operands[operand.index]
+            if isinstance(operand, BlockArgument) and operand.owner is successor
+            else operand
+        )
```
alexarice: Unrelated formatting change?

alexarice (approved):

tests/test_ir.py
alexarice: Hmm, I guess this one isn't the same

math-fehr: Yeah this one I wouldn't replace it, as we actually want to know how many uses there are.

tests/test_ir.py
math-fehr: Done

xdsl/transforms/canonicalization_patterns/cf.py
math-fehr: Done, sorry I missed that one

alexarice:
Just had a thought, do we want to leave this in with a deprecation message?

math-fehr:
Good point, we could if you want, that's easy to do

alexarice:
> Good point, we could if you want, that's easy to do

This is for sure something that will need changes downstream, so would be nice I think (as you say I don't think it blocks the next PR)

PR #4909, opened 2025-07-28
core: Use linked list for SSAValue uses

math-fehr:
Stacked PRs:
 * __->__#4909

--- --- ---

### DIFF
--- a/tests/test_use.py
+++ b/tests/test_use.py
@@ -1,7 +1,6 @@
 from xdsl.context import Context
 from xdsl.dialects.builtin import Builtin, ModuleOp
 from xdsl.dialects.test import Test, TestOp
-from xdsl.ir import Use
 from xdsl.parser import Parser
 
 test_prog = """
@@ -24,8 +23,11 @@ def test_main():
     assert isinstance(module, ModuleOp)
 
     op1, op2 = list(module.ops)
-    assert set(op1.results[0].uses) == {Use(op2, 0), Use(op2, 1)}
-    assert not op2.results[0].uses
+    assert set((use.operation, use.index) for use in op1.results[0].uses) == {
+        (op2, 0),
+        (op2, 1),
+    }
+    assert set(op2.results[0].uses) == set()
 
 
 def test_uses_methods():
@@ -58,7 +60,9 @@ def test_uses_methods():
     assert not res2.has_more_than_one_use()
 
     assert res0.get_unique_use() is None
-    assert res1.get_unique_use() == Use(op2, 0)
+    assert (use := res1.get_unique_use()) is not None
+    assert use.operation == op2
+    assert use.index == 0
     assert res2.get_unique_use() is None
 
     assert res0.get_user_of_unique_use() is None
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -525,15 +525,37 @@ class TypedAttribute(ParametrizedAttribute, ABC):
     def print_without_type(self, printer: Printer): ...
 
 
-@dataclass(frozen=True)
+@dataclass
 class Use:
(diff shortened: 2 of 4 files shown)


math-fehr:
@EdmundGoodman, please let me know if you think I correctly changed the constant folding benchmark!

tobiasgrosser:
Very cool!

xdsl/ir/core.py
```diff
@@ -921,6 +956,13 @@ class Operation(_IRNode):
     _operands: tuple[SSAValue, ...] = field(default=())
     """The operation operands."""
 
+    _operand_uses: tuple[Use, ...] = field(default=())
+    """
+    The uses for each operands.
```
superlopuh:
```suggestion
    The uses for each operand.
```

superlopuh (approved):

xdsl/ir/core.py
```diff
@@ -535,6 +535,18 @@ class Use:
     index: int
     """The index of the operand using the value in the operation."""
 
+    prev_use: Use | None = None
+    """The previous use of the value in the use list."""
```
alexarice, line 539: Should these be private? Feels like something which shouldn't be user modified

tests/test_use.py
```diff
@@ -24,8 +23,13 @@ def test_main():
...
     op1, op2 = list(module.ops)
-    assert set(op1.results[0].uses) == {Use(op2, 0), Use(op2, 1)}
-    assert not op2.results[0].uses
+    assert any(
+        use.operation == op2 and use.index == 0 for use in set(op1.results[0].uses)
+    )
+    assert any(
+        use.operation == op2 and use.index == 1 for use in set(op1.results[0].uses)
+    )
+    assert set(op2.results[0].uses) == set()
```
alexarice, line 30: Are these changes necessary? I guess the issue is we can't compare `Use` anymore.
What about:
```python
assert set((o.operation, o.index) for o in op1.results[0].uses) == {(op2, 0), (op2, 1)}
```
?

superlopuh:
What would the advantage be? I like that the linked list is intrusive here conceptually.

alexarice:
> What would the advantage be? I like that the linked list is intrusive here conceptually.

It feels to me like the user should not have (at least mutable) access to the linked list pointers. Separating out the `Use` structure from the linked list node class would keep the `Use` api the same (though maybe this doesn't really matter.)

xdsl/ir/core.py
math-fehr: Yeah I'll make these private, that makes sense

alexarice:
One more point, separating out `Use` would allow it to be immutable.

tests/test_use.py
math-fehr: Yes, it looks much nicer I agree

alexarice: Perhaps with a better naming choice than o

xdsl/ir/core.py
```diff
@@ -573,38 +598,59 @@ def uses(self) -> IRUses:
...
         If there are no uses or more than one use, returns None.
         """
-        if (use := self.get_unique_use()) is not None:
-            return use.operation
-        return None
+        if (first_use := self.first_use) is None:
+            return None
+        if first_use._next_use is not None:  # pyright: ignore[reportPrivateUsage]
+            return None
+        return first_use.operation
```
alexarice: Why not stick with the above?

math-fehr: Oh I think it happen when I changed it higher in the stack, let me put it back again

math-fehr: Oh I think it happen when I changed it higher in the stack, let me put it back again

PR #4911, opened 2025-07-28
core: add special handling of DenseArrayBase in assembly format

superlopuh:
Adds functionality present in MLIR, where the special assembly format for DenseArrayBase of a specific subclass is printed prettily.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1261,7 +1261,7 @@ class DenseArrayAttributeVariable(AttributeVariable):
         assert isa(attr, DenseArrayBase), attr
         with printer.in_square_brackets():
             if isa(attr, DenseArrayBase[IntegerType]):
-                printer.print_list(attr.get_values(), printer.print_int)
+                printer.print_list(attr.iter_values(), printer.print_int)
             elif isa(attr, DenseArrayBase[AnyFloat]):
                 printer.print_list(
                     attr.iter_values(),
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1258,7 +1258,6 @@ class DenseArrayAttributeVariable(AttributeVariable):
             )
 
     def print_attr(self, printer: Printer, attr: Attribute) -> None:
-        assert isa(attr, DenseArrayBase), attr
         with printer.in_square_brackets():
             if isa(attr, DenseArrayBase[IntegerType]):
                 printer.print_list(attr.iter_values(), printer.print_int)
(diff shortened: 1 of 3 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1215,6 +1223,52 @@ def print_attr(self, printer: Printer, attr: Attribute) -> None:
...
+            if self.is_optional:
+                elements = parser.parse_optional_comma_separated_list(
+                    parser.Delimiter.SQUARE, parser.parse_float
+                )
+                if elements is None:
+                    return None
+            else:
+                elements = parser.parse_comma_separated_list(
+                    parser.Delimiter.SQUARE, parser.parse_float
+                )
```
alexarice, line 1255: This feels like an unnecessary number of cases, I guess it's needed for pyright to be happy?

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1215,6 +1223,52 @@ def print_attr(self, printer: Printer, attr: Attribute) -> None:
...
+            return DenseArrayBase(
+                self.elt_type, BytesAttr(self.elt_type.pack(elements))
+            )
+
+    def print_attr(self, printer: Printer, attr: Attribute) -> None:
+        assert isa(attr, DenseArrayBase), attr
+        with printer.in_square_brackets():
+            if isa(attr, DenseArrayBase[IntegerType]):
+                printer.print_list(attr.get_values(), printer.print_int)
```
alexarice: This is maybe a good opportunity to add overloads to `iter_values` so we can use it for both?

superlopuh: Unfortunately that's not the issue, as it's about Pyright knowing that the value type which is one of int or float corresponds to the input to the printing function, which is int or float, but I couldn't figure out how to do it.

superlopuh: Oh you mean to avoid a temporary tuple?

alexarice: Wait, I'm confused, why can't we use `iter_values` here. (I thought it was because it didn't have overloads but it does have overloads).

alexarice: I didn't mean merging the branches, I just meant using `iter_values` on both branches

superlopuh: Yeah I misunderstood

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -1215,6 +1223,52 @@ def print_attr(self, printer: Printer, attr: Attribute) -> None:
...
+            else:
+                elements = parser.parse_comma_separated_list(
+                    parser.Delimiter.SQUARE, parser.parse_float
+                )
+            return DenseArrayBase(
+                self.elt_type, BytesAttr(self.elt_type.pack(elements))
+            )
+
+    def print_attr(self, printer: Printer, attr: Attribute) -> None:
+        assert isa(attr, DenseArrayBase), attr
```
alexarice: The assert feels slightly unnecessary as its rechecked again below.

PR #4912, opened 2025-07-28
core: add ConstraintConvertible for values that correspond to Attribute classes

superlopuh:
I first tried to make this Data-specific, but the typing didn't work out. The bound of the new class would have to be `Data[Any]`, but then Pyright complains in the IRDLAttrConstraint definition as you can't have more constrained bounds for union members.


tests/test_pyrdl.py
```diff
@@ -1,6 +1,7 @@
 """Unit tests for IRDL."""
 
 from dataclasses import dataclass
+from enum import StrEnum, auto
```
alexarice: `StrEnum` needs to come from `xdsl.utils`

alexarice (commented):
Could you quickly fix this pyright error first before I review properly?

xdsl/irdl/attributes.py
```diff
@@ -384,6 +402,7 @@ def decorator(cls: TypeAttributeInvT) -> TypeAttributeInvT:
 - An instance of `Attribute` representing an equality constraint on an attribute.
 - A type representing a specific attribute class.
 - A TypeForm that can represent both unions and generic attributes.
+- An instance or subclass of AttributeData, a Python class that has a corresponding Attribute class.
```
alexarice: ?

xdsl/irdl/attributes.py
```diff
@@ -371,11 +371,29 @@ def decorator(cls: TypeAttributeInvT) -> TypeAttributeInvT:
...
+        """The constraint for this instance."""
+
+
 IRDLAttrConstraint: TypeAlias = (
     AttrConstraint[AttributeInvT]
     | AttributeInvT
     | type[AttributeInvT]
     | "TypeForm[AttributeInvT]"
+    | type[ConstraintConvertible[AttributeInvT]]
+    | ConstraintConvertible[AttributeInvT]
```
alexarice, line 405: So is the idea that you can just pass in the enum directly instead of the Attribute that wraps the enum?

xdsl/irdl/attributes.py
superlopuh: Oh a doc string from a different iteration of this

xdsl/irdl/attributes.py
superlopuh: Yep exactly

alexarice: Would you be able to point at where this is used in the big PR (or push a follow up PR). It's not clear to me how this all fits together yet. Shouldn't `Attribute` itself be `ConstraintConvertible`?

superlopuh: This is the case for when enums are used as parameters to generic classes, these are mapped with irdl_to_attr_constraint. We could have a separate function for this case also, but I don't see the disadvantage to accepting them here.

alexarice: Will `int` also be `ConstraintConvertible`?

superlopuh: No because we can't retroactively conform it to that, it'll be special cased, and the assumption is that you always want to use IntAttr for int, but you don't know which attribute to use for a given enum

alexarice: Hmm, I'm not sure I really understand this then, is this abstract class just for enums?

superlopuh: I guess it could also be used for affine maps or other things that are wrapped by data in the future, I just couldn't figure out how to get the type system to be happy with just a subclass of Data

PR #4915, opened 2025-07-28
dialects: (memref) use DynamicIndexList

alexarice:
I'm for sure doing something wrong with the spacing here, but I cannot figure it out. Might come back to it with fresh eyes tomorrow.

I can't find why `memref.subview` had `attr-dict-with-keyword` before, but as far as I can tell it shouldn't.

P.S. a lot of these ops do not agree with current MLIR, I've made no attempt to fix them here.


math-fehr (approved):
Yeah that's really really cool!

alexarice:
Fixed the spacing differences

superlopuh (approved):

PR #4918, opened 2025-07-29
core: remove deprecated isattr

superlopuh:
Deprecated for months, should be safe to remove.


alexarice:
There's a test that needs removing I think

PR #4921, opened 2025-07-29
core: (Assembly Format) add ref directive

alexarice:
Adds ref directives, which allow a variable to be passed to a custom directive even if it has already appeared in the assembly format.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format_parser.py
+++ b/xdsl/irdl/declarative_assembly_format_parser.py
@@ -598,7 +598,9 @@ class FormatParser(BaseParser):
         """
         Parse a type directive with the following format:
           type-directive ::= `type` `(` typeable-directive `)`
-        `type` is expected to have already been parsed
+        `type` is expected to have already been parsed.
+        If `ref` is `True`, then this directive can have already in the
+        assembly format already.
         """
         with self.in_parens():
             return TypeDirective(self.parse_typeable_directive(ref))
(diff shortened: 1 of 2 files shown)


superlopuh:
Can you add documentation for this in the code?

alexarice:
Did a pass over all the docstrings

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -369,21 +369,24 @@ def verify_successors(self):
                 )
 
     def _parse_optional_operand(
-        self, variable_name: str, top_level: bool
+        self, variable_name: str, top_level: bool, ref: bool
```
math-fehr: ### 
```suggestion
        self, variable_name: str, top_level: bool, inside_ref: bool
```
maybe?

xdsl/irdl/declarative_assembly_format_parser.py
```diff
@@ -432,13 +440,16 @@ def parse_optional_typeable_variable(self) -> TypeableDirective | None:
 
     def parse_optional_variable(
         self,
+        ref: bool,
```
math-fehr:
```suggestion
        inside_ref: bool,
```

math-fehr (approved):
Looks good to me!

xdsl/irdl/declarative_assembly_format_parser.py
alexarice: I could do. I think If I'm changing this I'd also like to change `top_level` to `inside_type` for consistency. Would that be OK?

alexarice: (Could split this out first?)

PR #4924, opened 2025-07-29
dialects: (dlti) Add TargetDeviceSpecAttr and TargetSystemSpecAttr 

mesham:
Added TargetDeviceSpecAttr and TargetSystemSpecAttr attributes to the DLTI dialect

### DIFF
--- a/xdsl/dialects/dlti.py
+++ b/xdsl/dialects/dlti.py
@@ -69,7 +69,7 @@ class DLTIEntryMap(ParametrizedAttribute, ABC):
             printer.print_list(self.entries, print_entry)
 
     def verify(self) -> None:
-        if len(set(self.entries)) != len(self.entries):
+        if len({entry.key for entry in self.entries}) != len(self.entries):
             raise VerifyException("duplicate DLTI entry key")
 
 


xdsl/dialects/dlti.py
```diff
@@ -73,5 +68,67 @@ def print_entry(entry: DataLayoutEntryAttr):
         with printer.in_angle_brackets():
             printer.print_list(self.entries, print_entry)
 
+    def verify(self) -> None:
+        keys: Sequence[Attribute] = []
+        for entry in self.entries:
+            entry.verify()
+            if entry.key in keys:
+                raise VerifyException("duplicate DLTI entry key")
+            keys.append(entry.key)
```
superlopuh:
```suggestion
        if len(set(self.entries)) != len(self.entries):
            raise VerifyException("duplicate DLTI entry key")
```

xdsl/dialects/dlti.py
```diff
@@ -33,20 +35,13 @@ def verify(self) -> None:
             raise VerifyException("key must be a string or a type attribute")
 
 
-@irdl_attr_definition
-class DataLayoutSpecAttr(ParametrizedAttribute):
+class DLTIEntryMap(ParametrizedAttribute):
```
superlopuh: nit:
```suggestion
class DLTIEntryMap(ParametrizedAttribute, ABC):
```

xdsl/dialects/dlti.py
```diff
@@ -33,20 +35,13 @@ def verify(self) -> None:
...
-    An attribute to represent a data layout specification.
-
-    A data layout specification is a list of entries that specify (partial) data
-    layout information. It is expected to be attached to operations that serve as
-    scopes for data layout requests.
-
-    https://mlir.llvm.org/docs/Dialects/DLTIDialect/#datalayoutspecattr
+    Many DLTI dialect operations contain arrays of DataLayoutEntryInterface,
+    with these representing different things such as data layout or information
+    about the target hardware. This is the base class that these operations extend
```
superlopuh:
```suggestion
    about the target hardware. This is the base class that these operations extend.
```

xdsl/dialects/dlti.py
mesham: Nice! Thanks, I thought there must be a nicer way to do it. I had to slightly tweak as entries is an arrayattr, so creating a set directly also took account of the value (which we don't want, its the unique keys that we care about), hence I use set comprehension to create a set of the keys explicitly and then check it against the length.

superlopuh (approved):
Thank you!

math-fehr (approved):
Thanks a lot!

PR #4929, opened 2025-07-30
core: (Assembly Format) simplify typing

alexarice:
The assembly format parser is currently slightly confusing in that it stores various fields as `X | None | Sequence[X]` where `None` means the field isn't assigned yet, a non-variadic field must be set to `X`, a variadic field is set to `Sequence[X]` and an optional field is set to either an empty `Sequence[X]` or `X` (but not `None`). 

I propose simplifying this to just set every field to a (possible length 1) tuple, which I think simplifies things when resolving (though does mean that some things must be wrapped in singleton tuples)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dev = [
     "nbval<0.12",
     "filecheck==1.0.2",
     "lit<19.0.0",
-    "marimo==0.14.13",
+    "marimo==0.14.15",
     "pre-commit>=4.0.0,<5.0.0",
     "ruff==0.12.7",
     "nbconvert>=7.7.2,<8.0.0",
--- a/tests/filecheck/dialects/dlti/attrs.mlir
+++ b/tests/filecheck/dialects/dlti/attrs.mlir
@@ -7,9 +7,178 @@
     entry2 = #dlti.dl_entry<i32, i32>
 } : () -> ()
 
+// -----
+
 // CHECK: spec1 = #dlti.dl_spec<>
 // CHECK: spec2 = #dlti.dl_spec<"str" = i32, i32 = i32>
 "test.op"() {
     spec1 = #dlti.dl_spec<>,
     spec2 = #dlti.dl_spec<"str" = i32, i32 = i32>
 } : () -> ()
+
+// -----
+
+// CHECK:      module attributes {
+// CHECK-SAME:  dlti.target_system_spec = #dlti.target_system_spec<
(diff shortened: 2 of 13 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -707,14 +680,14 @@ def is_anchorable(self) -> bool:
...
-        field: list[_T | None | Sequence[_T]],
+        field: list[None | Sequence[_T]],
         field_name: str,
         set_to: Sequence[_T],
     ) -> str | None:
         if self.variadic_index is None:
             if len(set_to) != len(field):
                 return f"Expected {len(field)} {field_name} but found {len(set_to)}"
-            field[:] = set_to
+            field[:] = map(lambda x: (x,), set_to)
```
superlopuh: I'd recommend avoiding map always basically. Not sure if this works without the tuple but worth a try
```suggestion
            field[:] = tuple((x,) for x in set_to)
```

alexarice: I think this will work, but what's the issue with map? This just makes a tuple then iterates over it right?

superlopuh: It's significantly slower and seems to be kept around for legacy/stability reasons, iterator expressions are generally preferable in my understanding.

xdsl/irdl/declarative_assembly_format.py
```diff
@@ -707,14 +680,14 @@ def is_anchorable(self) -> bool:
...
-        field: list[_T | None | Sequence[_T]],
+        field: list[None | Sequence[_T]],
         field_name: str,
         set_to: Sequence[_T],
     ) -> str | None:
         if self.variadic_index is None:
             if len(set_to) != len(field):
                 return f"Expected {len(field)} {field_name} but found {len(set_to)}"
-            field[:] = set_to
+            field[:] = tuple((x,) for x in set_to)
```
superlopuh: Does it work without the tuple? Could you do the maps below also?

alexarice: I don't think it works without but can check

math-fehr (approved):
Looks simpler to me, nice!

xdsl/irdl/declarative_assembly_format.py
alexarice: It seems to work actually, I didn't realise you needed brackets

PR #4933, opened 2025-07-30
core: (Operation) optimise accessors

alexarice:
Test at hardcoding all the operation accessors. It's unclear whether this is actually worth it.

### DIFF
--- a/xdsl/irdl/operations.py
+++ b/xdsl/irdl/operations.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from abc import ABC
+from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass, field
 from enum import Enum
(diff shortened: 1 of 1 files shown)


xdsl/irdl/operations.py
```diff
@@ -1266,6 +1266,24 @@ def get_op_constructs(
...
+    """
+    match construct:
+        case VarIRConstruct.OPERAND:
+            return VarOperand(args)  # pyright: ignore[reportArgumentType]
+        case VarIRConstruct.RESULT:
+            return VarOpResult(args)  # pyright: ignore[reportArgumentType]
+        case _:
+            return args  # pyright: ignore[reportReturnType]
+
+
```
superlopuh: with the SSAValues change this is no longer necessary

alexarice: I don't see how to avoid it

superlopuh: slicing args will return SSAValues, so there's no need to wrap them

alexarice: oh nice

alexarice:
Not sure if it's the most representative benchmark but on main:
```
Test ConstantFolding.1000 ran in: 0.0604 ± 0.00443s
```
and on this branch
```
Test ConstantFolding.1000 ran in: 0.0532 ± 0.00392s
```

PR #4935, opened 2025-07-30
dialects: (tosa) Add support for `tosa.matmul`

jakedves:
Adds support and test cases for `tosa.matmul`

### DIFF
--- a/xdsl/dialects/tosa.py
+++ b/xdsl/dialects/tosa.py
@@ -1,11 +1,9 @@
 from abc import ABC
-from collections.abc import Sequence
 from typing import ClassVar, Generic
 
-from typing_extensions import Self, TypeVar
+from typing_extensions import TypeVar
 
 from xdsl.dialects.builtin import (
-    I1,
     I32,
     I64,
     AnyAttr,
@@ -18,33 +16,19 @@ from xdsl.dialects.builtin import (
     StringAttr,
     TensorType,
 )
-from xdsl.ir import Attribute, Block, Dialect, Operation, Region, SSAValue
-from xdsl.ir.core import Block
+from xdsl.ir import Attribute, Dialect
 from xdsl.irdl import (
     IRDLOperation,
     ParsePropInAttrDict,
     VarConstraint,
     irdl_op_definition,
-    lazy_traits_def,
     operand_def,
     opt_prop_def,
     prop_def,
-    region_def,
     result_def,
     traits_def,
-    var_operand_def,
-    var_result_def,
 )
-from xdsl.parser import Parser
-from xdsl.printer import Printer
-from xdsl.traits import (
-    Commutative,
-    HasParent,
-    IsTerminator,
-    Pure,
-    RecursiveMemoryEffect,
-    SingleBlockImplicitTerminator,
-)
+from xdsl.traits import Commutative, Pure
 from xdsl.utils.exceptions import VerifyException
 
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/tosa.py
```diff
@@ -223,6 +223,70 @@ class CosOp(ElementwiseUnaryOperation[TensorType[AnyFloat]]):
...
+        sb = self.b.type.get_shape()
+        s_az = self.a_zp.type.get_shape()
+        s_bz = self.b_zp.type.get_shape()
+
+        if len(sa) != 3 or len(sb) != 3:
+            raise VerifyException("'tosa.matmul' Expected operand tensors of rank 3")
+
+        if sa[0] != 1 or sb[0] != 1:
+            raise VerifyException(
+                "'tosa.matmul' Expected leading rank of input tensors to be 1"
```
superlopuh: Unless TOSA uses a different convention, the rank is the length of the shape, which contains dimensions, right?
```suggestion
                "'tosa.matmul' Expected leading dimension of input tensors to be 1"
```

superlopuh (approved):

xdsl/dialects/tosa.py
```diff
@@ -223,6 +223,70 @@ class CosOp(ElementwiseUnaryOperation[TensorType[AnyFloat]]):
...
 
 
+@irdl_op_definition
+class MatMulOp(IRDLOperation):
+    """
+    TOSA dialect operation for computing 2D matmuls. Expects 3D tensors as input with leading rank of 1 element, e.g.
+
+    `tensor<1x14x19xf32> * tensor<1x19x28xf32> -> tensor<1x14x28xf32>`
+
+    The following two operands are the zero-point which are used for quantized operations, can be set to 0.0 for no effect.
```
```suggestion
    The operands `a_zp` and `b_zp` are the zero-point which are used for quantized operations, can be set to 0.0 for no effect.
```
I feel this is slightly more clear

xdsl/dialects/tosa.py
```diff
@@ -223,6 +223,70 @@ class CosOp(ElementwiseUnaryOperation[TensorType[AnyFloat]]):
...
+    )
+
+    def verify_(self) -> None:
+        if not (
+            isinstance(self.a.type, ShapedType)
+            and isinstance(self.b.type, ShapedType)
+            and isinstance(self.a_zp.type, ShapedType)
+            and isinstance(self.b_zp.type, ShapedType)
+        ):
+            raise VerifyException("'tosa.matmul' Expected operands to be shaped types")
```
alexarice: I feel these could safely be asserts? This should be checked by the constraints

xdsl/dialects/tosa.py
```diff
@@ -223,6 +223,70 @@ class CosOp(ElementwiseUnaryOperation[TensorType[AnyFloat]]):
...
+    """
+
+    name = "tosa.matmul"
+
+    T: ClassVar = VarConstraint("T", AnyAttr())
+
+    a = operand_def(TensorType.constr(T))
+    b = operand_def(TensorType.constr(T))
+    a_zp = operand_def(TensorType.constr(T))
+    b_zp = operand_def(TensorType.constr(T))
```
alexarice, line 245: I'm unsure that `a_zp` and `b_zp` need to have the same element type as `a` and `b`? At least in the tablegen `a` and `b` have `Tosa_Tensor3D` and `a_zp` and `b_zp` have type `Tosa_ScalarIntOrFloatTensor`

alexarice (approved):
This looks good to me. Left a couple of very minor comments

xdsl/dialects/tosa.py
alexarice: I'm not actually so sure on this now, maybe the only difference between these is the rank

jakedves: I wasn't sure either -- the only examples I had available were from their test cases in MLIR

PR #4940, opened 2025-07-31
frontend: (pyast): add support for expr AST node

EdmundGoodman:
This PR adds support for the expr ast node by passing through walking its value. This is required for building out the PDL rewrite frontend and peeled off #4859


superlopuh:
As we peel off changes I think we should add tests that motivate the individual bits of infrastructure, otherwise it basically requires reviewing the big PR anyway

EdmundGoodman:
Closed as will be handled in single rewriting PR

PR #4941, opened 2025-07-31
frontend: (pyast) add transformation pass to map functions to pdl rewrites

EdmundGoodman:
This PR adds a transformation pass which converts `func.func` operations generated by the frontend into pdl rewrite blocks. It is peeled from #4859

### DIFF
--- a/xdsl/transforms/func_to_pdl_rewrite.py
+++ b/xdsl/transforms/func_to_pdl_rewrite.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 
 from xdsl.context import Context
 from xdsl.dialects import builtin, func, pdl, test
+from xdsl.ir import Block, Region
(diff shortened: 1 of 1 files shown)


superlopuh (commented):
Do we need this as a pass exposed to xdsl-opt? I think this can be a hidden transformation as part of the PDL infrastructure

EdmundGoodman:
> Do we need this as a pass exposed to xdsl-opt? I think this can be a hidden transformation as part of the PDL infrastructure

It probably doesn't need to be -- does this then mean reverting the changes to xdsl/transforms/__init__.py?

superlopuh:
It's more that it feels like the wrong bit to pull out from the target PR, let's iterate there instead to figure out first how things will fit together?

EdmundGoodman:
Closed as will be handled in single rewriting PR

PR #4942, opened 2025-07-31
core: speed up checking for uses

superlopuh:
Before:

```
❯ python benchmarks/rewriting.py RemoveUnused.is_trivially_dead dis
===========================================
 Events recorded: 58
 Opcodes recorded: 40
 Uninstrumented: 350.35 ns
 Uninstrumented per opcode: 8.76 ns/op
 Instrumented: 11895.62 ns
 Instrumented per opcode: 297.39 ns/op
 Elapsed: 4825.74 ns
 Elapsed per opcode: 120.64 ns/op
 Opcode overhead: 0.00 ns/op
 Tracing overhead per opcode: 111.88 ns/op
===========================================
```

After:

```
❯ python benchmarks/rewriting.py RemoveUnused.is_trivially_dead dis ===========================================
 Events recorded: 39
 Opcodes recorded: 27
 Uninstrumented: 220.84 ns
 Uninstrumented per opcode: 8.18 ns/op
 Instrumented: 7753.91 ns
 Instrumented per opcode: 287.18 ns/op
 Elapsed: 2951.75 ns
 Elapsed per opcode: 109.32 ns/op
 Opcode overhead: 0.00 ns/op
 Tracing overhead per opcode: 101.15 ns/op
===========================================
```

### DIFF
--- a/xdsl/interpreters/eqsat_pdl_interp.py
+++ b/xdsl/interpreters/eqsat_pdl_interp.py
@@ -151,7 +151,7 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
         if result.has_one_use():
             if isinstance(eclass_op := result.get_user_of_unique_use(), eqsat.EClassOp):
                 result = eclass_op.result
-        elif result.uses:  # multiple uses
+        else:
             for use in result.uses:
                 if isinstance(use.operation, eqsat.EClassOp):
                     raise InterpretationError(
@@ -183,7 +183,7 @@ class EqsatPDLInterpFunctions(PDLInterpFunctions):
                 ):
                     assert len(eclass_op.results) == 1
                     result = eclass_op.results[0]
-            elif result.uses:  # multiple uses
+            else:
                 for use in result.uses:
                     if isinstance(use.operation, eqsat.EClassOp):
                         raise InterpretationError(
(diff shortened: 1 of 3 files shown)


xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -151,7 +151,7 @@ def run_get_result(
         if result.has_one_use():
             if isinstance(eclass_op := result.get_user_of_unique_use(), eqsat.EClassOp):
                 result = eclass_op.result
-        elif result.uses:  # multiple uses
+        else:
```
superlopuh, line 154: if there are no uses then the for loop is trivially empty

xdsl/transforms/dead_code_elimination.py
```diff
@@ -34,7 +34,9 @@ def is_trivially_dead(op: Operation):
     """
     Returns if the operation has no observable effect.
     """
-    return all(not result.uses for result in op.results) and would_be_trivially_dead(op)
+    return all(
+        result.first_use is None for result in op.results
+    ) and would_be_trivially_dead(op)
```
alexarice, line 39: This just looks like a manual code inline?

superlopuh: Yep. I feel that we should expose the `__bool__` method for convenience and still inline things for performance when we care about them. In most of the code this is not an issue but this function for example will run on all operations in a module on every `canonicalize` pass so feels a bit more performance sensitive than the average function. I didn't remove the uses of `bool` in tests for example, where that feels more readable.

superlopuh:
I agree, but I also think that it's worth it for this code, I'll merge it as-is.

PR #4943, opened 2025-07-31
dialects: (x86_func) print visibility and .text directive

superlopuh:
We currently just print a bare label which isn't enough for the assembler/linker.


xdsl/dialects/x86_func.py
```diff
@@ -112,11 +115,25 @@ def print(self, printer: Printer):
             reserved_attr_names=("sym_name", "function_type", "sym_visibility"),
         )
 
-    def assembly_line(self) -> str | None:
-        if self.body.blocks:
-            return f"{self.sym_name.data}:"
-        else:
-            return None
+    def print_assembly(self, printer: AssemblyPrinter) -> None:
```
superlopuh, line 118: pretty similar to our implementation for `riscv_func.func`

tests/filecheck/dialects/x86_func/x86_func_asm.mlir
```diff
@@ -0,0 +1,35 @@
+// RUN: xdsl-opt -t riscv-asm --split-input-file --verify-diagnostics %s | filecheck %s
```
qaco, line 1: It should be x86 rather than riscv (what about raising an error when we try to generate riscv assembly from x86 dialect ?)

superlopuh: Oh yeah that's my bad, although the fact that they both work is interesting they both call the generic AssemblyPrinter.

qaco (approved):

PR #4948, opened 2025-08-01
frontend: (pyast): pdl rewrite example for simple erasure case

EdmundGoodman:
This PR provides an example of using PyAST frontend infrastructure for expressing PDL rewrites, in the simple case of erasing a single operation


xdsl/transforms/__init__.py
superlopuh, line 1: let's remove this

xdsl/frontend/pypdl/transforms/func_to_pdl_rewrite.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+
+@dataclass
+class FuncToPdlRewritePattern(RewritePattern):
+    """Rewrite pattern that transforms a function into a PDL rewrite operation."""
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(
+        self, op: func.FuncOp | func.ReturnOp, rewriter: PatternRewriter
```
superlopuh: I would recommend splitting this into two separate rewrites

xdsl/frontend/pypdl/transforms/func_to_pdl_rewrite.py
```diff
@@ -0,0 +1,58 @@
...
+                body=Region(
+                    Block([pdl_root, pdl.RewriteOp(root=rewrite_root, body=func_body)])
+                ),
+            )
+            rewriter.replace_matched_op(pdl_pattern)
+        else:
+            rewriter.erase_matched_op()
+
+
+class FuncToPdlRewrite(ModulePass):
```
superlopuh: I don't think we need a pass for this

xdsl/frontend/pypdl/builder.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@dataclass
+class PyPDLRewriteBuilder(PyASTBuilder):
+    """Builder for PDL rewrites from aspects of a Python function."""
+
+    def build(self) -> ModuleOp:
+        """Build a PDL rewrite from the builder state."""
+        module = super().build()
+        FuncToPdlRewrite().apply(ctx=None, op=module)
```
superlopuh: let's just inline whatever logic we need here

tests/filecheck/frontend/pypdl/pdl.py
```diff
@@ -0,0 +1,73 @@
...
+
+ctx = PyPDLContext()
+ctx.register_type(arith.ConstantOp, pdl.OperationType())
+ctx.register_function(erase_op, pdl.EraseOp)
+
+
+@ctx.pdl_rewrite
+def constant_replace(matched_operation: arith.ConstantOp) -> arith.ConstantOp:
+    erase_op(matched_operation)
+    return matched_operation
```
superlopuh, line 48: why return the matched operation?

tests/filecheck/frontend/pypdl/pdl.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def erase_op(operation: Operation) -> None:
+    """Shim to avoid `Expr` AST node required for methods."""
+    return Rewriter.erase_op(operation)
+
+
+ctx = PyPDLContext()
+ctx.register_type(arith.ConstantOp, pdl.OperationType())
+ctx.register_function(erase_op, pdl.EraseOp)
```
superlopuh, line 41: shouldn't these always be registered for things to work? I would recommend putting this into the init of PyPDLContext

tests/filecheck/frontend/pypdl/pdl.py
```diff
@@ -0,0 +1,73 @@
...
+  %2 = pdl.operation "arith.constant" {"value" = %1} -> (%0 : !pdl.type)
+  pdl.rewrite %2 {
+    pdl.erase %2
+  }
+}
+```
+"""
+
+from xdsl.dialects import arith, builtin, pdl
+from xdsl.frontend.pypdl.context import PyPDLContext
```
superlopuh: Ideally the API would look something like this:
```suggestion
from xdsl.frontend import pypdl
```

and lower:

``` python
ctx = pypdl.PyPDLContext()
...
pypdl.erase_op(op)
```

This seems a bit more stable as an API, to not expose users to our directory structure. So I'd put the context and whatever functions we expect users to reach for directly into the `__init__` but things like helper rewrite patterns into a submodule

xdsl/transforms/__init__.py
EdmundGoodman: Ruff complains if you remove it:

```
xdsl/frontend/pypdl/transforms/func_to_pdl_rewrite.py:1:1: INP001 File `xdsl/frontend/pypdl/transforms/func_to_pdl_rewrite.py` is part of an implicit namespace package. Add an `__init__.py`
```

xdsl/frontend/pypdl/context.py
```diff
@@ -0,0 +1,30 @@
...
+from dataclasses import dataclass
+from inspect import currentframe
+
+from xdsl.frontend.pyast.context import PyASTContext
+from xdsl.frontend.pyast.program import P, PyASTProgram, R
+from xdsl.frontend.pypdl.builder import PyPDLRewriteBuilder
+
+
+@dataclass
+class PyPDLContext(PyASTContext):
```
superlopuh: Subclassing PyASTContext kind of exposes the user to PyAST, which I'm not sure we want. Would it make sense to make this a new class that holds a PyAST context instead? We might want to expose some way on PyAST builder to run either passes or a callback on the module after it's extracted.

tests/filecheck/frontend/pypdl/pdl.py
EdmundGoodman: This is a limitation of the pyast implementation currently -- functions must return a value as the None type isn't handled properly. In a future patch I'll try to fix this

xdsl/transforms/__init__.py
superlopuh: Sorry I meant the diff in this file, all good

tests/filecheck/frontend/pypdl/pdl.py
superlopuh: Ah OK sounds good

superlopuh: Might be worth a comment even if it's very temporary?

tests/filecheck/frontend/pypdl/pdl.py
EdmundGoodman: I think this might be a decision for a later PR, as picking sensible default values is not completely trivial. It should definitely have all of the pdl operations, but also some amount of arith/builtin/scf which is less obvious to decide

xdsl/frontend/pypdl/transforms/func_to_pdl_rewrite.py
EdmundGoodman: Since the new plan is to apply passes for post processing, this pass still needs to exist, but is no longer registered to be usable through `xdsl-opt`

xdsl/frontend/pypdl/builder.py
EdmundGoodman: New approach of applying a list of passes supercedes this approach

tests/filecheck/frontend/pypdl/pdl.py
EdmundGoodman: Have exposed `PyPDLContext` through the `__init__` file

xdsl/frontend/pypdl/context.py
EdmundGoodman: Imo `PyPDLContext` is a specialised type of `PyASTContext` which offers additional functionality: doing PDL rewrites. This could be restructured if you have strong opinions, but I don't really see what to move it to?

xdsl/frontend/pyast/utils/builder.py
```diff
@@ -31,8 +33,28 @@ class PyASTBuilder:
...
+
+    def _apply_post_transforms(
+        self,
+        module: ModuleOp,
+        context: Context,
+        verify: bool = True,
+    ) -> None:
+        """Apply post transforms to a module."""
+        for transform in self.post_transforms:
+            transform.apply(context, module)
```
superlopuh: transforms can mutate the context, I would recommend copying it first before applying the transforms

superlopuh: Ah you do it below all good

xdsl/frontend/pyast/context.py
```diff
@@ -125,13 +125,19 @@ def decorator(func: Callable[P, R]) -> PyASTProgram[P, R]:
...
+            if not desymref:
+                kwargs["post_transforms"] = []
+
             builder = PyASTBuilder(
                 type_registry=self.type_registry,
                 function_registry=self.function_registry,
                 file=func_file,
                 globals=func_globals,
                 function_ast=func_ast,
-                desymref=desymref,
```
superlopuh: let's do this change in a separate first PR?

xdsl/frontend/pypdl/context.py
superlopuh: I would recommend moving it to `__init__.py`, removing this file, moving the `PyASTContext` to a variable on this, and adding a helper functions for things like registering attrs and ops.

xdsl/transforms/__init__.py
superlopuh: Actually the diff is back, let's not expose this in this file?

xdsl/frontend/pypdl/builder.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass, field
+
+from xdsl.frontend.pyast.utils.builder import PyASTBuilder
+from xdsl.frontend.pypdl.transforms.func_to_pdl_rewrite import FuncToPdlRewrite
+from xdsl.passes import ModulePass
+from xdsl.transforms.desymref import FrontendDesymrefyPass
+
+
+@dataclass
+class PyPDLRewriteBuilder(PyASTBuilder):
```
superlopuh: Would passing post_transforms and post_verify as arguments work? I don't see advantages to subclassing in this case.

xdsl/frontend/pyast/context.py
```diff
@@ -125,13 +125,19 @@ def decorator(func: Callable[P, R]) -> PyASTProgram[P, R]:
             func_file, func_globals, func_ast = self._get_func_info(
                 currentframe(), func, decorated_func
             )
+
+            # Override default post transforms if desymref is unset
+            kwargs: dict[str, Any] = {}
+            if not desymref:
+                kwargs["post_transforms"] = []
```
superlopuh: it should pass `self.post_transforms` here, as users should set those on the ctx directly

superlopuh: Actually the whole thing could be a PassPipeline instead and then it's easy to do verification between passes not as a property but as a lambda

xdsl/frontend/pyast/utils/builder.py
```diff
@@ -31,8 +33,28 @@ class PyASTBuilder:
...
+
+    post_verify: bool = True
+    """Whether to verify each post processing transformation pass."""
+
+    def _apply_post_transforms(
+        self,
+        module: ModuleOp,
+        context: Context,
+        verify: bool = True,
+    ) -> None:
```
superlopuh: What's the advantage of this helper over inlining the implementation? I can't think of a case of using it outsside of `build`

xdsl/frontend/pypdl/builder.py
superlopuh: Let's delete this file

xdsl/frontend/pypdl/__init__.py
```diff
@@ -0,0 +1,5 @@
+from xdsl.frontend.pypdl.context import PyPDLContext
+
+__all__ = [
+    "PyPDLContext",
+]
```
superlopuh: let's inline PyPDLContext here

PR #4951, opened 2025-08-01
core: inherit traits from parent classes

superlopuh:
This kind of enables us to properly use abstract superclasses as interfaces, sort of, in a way that I thought already worked. In this case we have a `HasRegisterConstraintsTrait`, which is declared as a trait in `HasRegisterConstraints`. This is now inherited by all operations, which enables proper verification of linearity of inout registers. Once this gets merged, we can look into doing something similar for canonicalization, to avoid the extra `HasCanonicalizationPatternsTrait` class that we now add instead of just a function on the operation itself.


superlopuh:
I marked this as draft but the core mechanism is ready to review, it's just stacked on the fix to an x86 canonicalization which I'd like to merge first.

tests/filecheck/dialects/riscv_snitch/assembly_emission.mlir
superlopuh, line 1: I would recommend not looking at any of the riscv/x86 filecheck files which move some used values around to make sure that we only use inout things once.

tests/test_traits.py
```diff
@@ -193,15 +193,18 @@ class LargerOperandOp(IRDLOperation, ABC):
 class TestCopyOp(LargerOperandOp):
     name = "test.test_copy"
 
-    traits = OpTraits(LargerOperandOp.traits.traits | {BitwidthSumLessThanTrait(64)})
```
superlopuh, line 196: no need for this manual union

xdsl/ir/core.py
```diff
@@ -914,18 +914,18 @@ class OpTraits(Iterable[OpTrait]):
...
-    ) -> None:
-        self._traits = traits
+    def __init__(self, gen_traits: Callable[[], tuple[OpTrait, ...]]) -> None:
+        self.gen_traits = gen_traits
+        self._traits = None
 
     @property
     def traits(self) -> frozenset[OpTrait]:
         """Returns a copy of this instance's traits."""
-        if callable(self._traits):
```
superlopuh, line 942: I've been meaning to get rid of this for a while, it's a much more expensive check than the new mechanism, expensive enough to be quite noticeable in traces I've looked at

xdsl/ir/core.py
```diff
@@ -914,18 +914,18 @@ class OpTraits(Iterable[OpTrait]):
     at the first use.
     """
 
-    _traits: frozenset[OpTrait] | Callable[[], tuple[OpTrait, ...]]
+    gen_traits: Callable[[], tuple[OpTrait, ...]]
+    _traits: frozenset[OpTrait] | None
```
math-fehr: Can you document both variables to explain why they exist?

PR #4952, opened 2025-08-02
dialects: (dlti) Add initialisation convenience from dictionary and python types

mesham:
DLTI dialect initialises the entry map operations based on a dictionary or list, and the entries themselves are initialised based on str or attribute (for key), and attribute, str, int or float for value. This is for convenience when creating these in code.

Pyright test added for this, along with testing verification of aspects in the dialect too

### DIFF
--- a/tests/dialects/test_dlti.py
+++ b/tests/dialects/test_dlti.py
@@ -31,131 +31,73 @@ DictValueType: TypeAlias = Mapping[
 def test_data_layout_entry():
     # Test passing strings for key and value
     entry = DataLayoutEntryAttr("test", "V")
-    assert isinstance(entry.key, StringAttr)
-    assert isinstance(entry.value, StringAttr)
-    assert entry.key.data == "test"
-    assert entry.value.data == "V"
+    assert entry.key == StringAttr("test")
+    assert entry.value == StringAttr("V")
 
     # Test passing string for key and int for value
     entry = DataLayoutEntryAttr("test2", 12)
-    assert isinstance(entry.key, StringAttr)
-    assert entry.key.data == "test2"
-    assert isinstance(entry.value, IntAttr)
-    assert entry.value.data == 12
+    assert entry.key == StringAttr("test2")
+    assert entry.value == IntAttr(12)
 
     # Test passing string for key and float for value
     entry = DataLayoutEntryAttr("test3", FloatAttr(99.45, Float32Type()))
-    assert isinstance(entry.key, StringAttr)
-    assert isinstance(entry.value, FloatAttr)
-    assert entry.key.data == "test3"
-    assert round(entry.value.value.data, 2) == 99.45  # pyright: ignore
+    assert entry.key == StringAttr("test3")
+    assert entry.value == FloatAttr(99.45, Float32Type())
 
     # Test passing type for key and string for value
     entry = DataLayoutEntryAttr(i32, "test")
-    assert isinstance(entry.key, TypeAttribute)
-    assert isinstance(entry.value, StringAttr)
     assert entry.key == i32
-    assert entry.value.data == "test"
+    assert entry.value == StringAttr("test")
 
     # Test passing StringAttr for key and value
     entry = DataLayoutEntryAttr(StringAttr("k"), StringAttr("v"))
-    assert isinstance(entry.key, StringAttr)
-    assert isinstance(entry.value, StringAttr)
-    assert entry.key.data == "k"
-    assert entry.value.data == "v"
+    assert entry.key == StringAttr("k")
+    assert entry.value == StringAttr("v")
 
 
 def test_incorrect_data_layout_entry():
     with pytest.raises(VerifyException):
-        entry = DataLayoutEntryAttr("", "V")
-        entry.verify()
-
-
-def generic_test_entry_equals_defn(
-    dlti_entry: DataLayoutEntryAttr,
-    comparison_entry: tuple[
-        StringAttr | TypeAttribute | str, Attribute | str | int | DictValueType
-    ],
-):
-    assert isinstance(dlti_entry, DataLayoutEntryAttr)
-
-    comparison_entry_key = comparison_entry[0]
-    comparison_entry_value = comparison_entry[1]
-
-    dlti_entry_key = dlti_entry.key
-    dlti_entry_value = dlti_entry.value
-
-    if isinstance(comparison_entry_key, str) or isinstance(
-        comparison_entry_key, StringAttr
-    ):
-        assert isinstance(dlti_entry_key, StringAttr)
-        assert (
-            dlti_entry_key.data == comparison_entry_key.data
-            if isinstance(comparison_entry_key, StringAttr)
-            else comparison_entry_key
-        )
-    else:
-        assert isinstance(dlti_entry_key, TypeAttribute)
-        assert dlti_entry_key == comparison_entry_key
-
-    if isinstance(comparison_entry_value, str) or isinstance(
-        comparison_entry_value, StringAttr
-    ):
-        assert isinstance(dlti_entry_value, StringAttr)
-        assert (
-            dlti_entry_value.data == dlti_entry_value.data
-            if isinstance(comparison_entry_value, StringAttr)
-            else comparison_entry_value
-        )
-    elif isinstance(comparison_entry_value, int) or isinstance(
-        comparison_entry_value, IntAttr
-    ):
-        assert isinstance(dlti_entry_value, IntAttr)
-        assert (
-            dlti_entry_value.data == dlti_entry_value.data
-            if isinstance(comparison_entry_value, IntAttr)
-            else comparison_entry_value
-        )
-    elif isinstance(comparison_entry_value, FloatAttr):
-        assert isinstance(dlti_entry_value, FloatAttr)
-        assert dlti_entry_value.value.data == dlti_entry_value.value.data
-    else:
-        pytest.fail("Unknown comparison type")
+        DataLayoutEntryAttr("", "V")
 
 
 def generic_specification_test(
     dlti_class: type[DLTIEntryMap],
     contents: ArrayAttr[DataLayoutEntryAttr] | DictValueType,
-    comparison_entries: list[
-        tuple[StringAttr | TypeAttribute | str, Attribute | str | int]
-    ],
+    comparison_entries: list[Attribute],
 ):
     spec = dlti_class(contents)
     assert isa(spec.entries, ArrayAttr[DataLayoutEntryAttr])
     assert len(spec.entries) == len(comparison_entries)
 
-    for dlti_entry, comparison in zip(spec.entries.data, comparison_entries):
-        generic_test_entry_equals_defn(dlti_entry, comparison)
+    for idx, (dlti_entry, comparison_entry) in enumerate(
+        zip(spec.entries.data, comparison_entries)
+    ):
+        assert dlti_entry.key == StringAttr("key_" + str(idx))
+        assert dlti_entry.value == comparison_entry
 
 
 @pytest.mark.parametrize(
     "entries",
     [
-        [("key1", "value1"), ("key2", 23), (i32, 43)],
-        [("k", IntAttr(23)), (i32, FloatAttr(2.4, Float32Type()))],
+        [StringAttr("value1"), IntAttr(23), IntAttr(43)],
+        [IntAttr(23), FloatAttr(2.4, Float32Type())],
     ],
 )
 def test_data_layout_spec(
-    entries: list[tuple[StringAttr | TypeAttribute | str, Attribute | str | int]],
+    entries: list[Attribute],
 ):
     # Test initialisation from a dictionary
     generic_specification_test(
-        DataLayoutSpecAttr, {e[0]: e[1] for e in entries}, entries
+        DataLayoutSpecAttr,
+        {"key_" + str(idx): e for idx, e in enumerate(entries)},
+        entries,
     )
     # Test initialisation from an array attribute of data layout entries
     generic_specification_test(
         DataLayoutSpecAttr,
-        ArrayAttr([DataLayoutEntryAttr(e[0], e[1]) for e in entries]),
+        ArrayAttr(
+            [DataLayoutEntryAttr("key_" + str(idx), e) for idx, e in enumerate(entries)]
+        ),
         entries,
     )
 
@@ -163,21 +105,25 @@ def test_data_layout_spec(
 @pytest.mark.parametrize(
     "entries",
     [
-        [("key1", "value1"), ("key2", 23), (i32, FloatAttr(9.3, Float32Type()))],
-        [("k", IntAttr(23)), (i32, FloatAttr(2.4, Float32Type()))],
+        [StringAttr("value1"), IntAttr(23), IntAttr(43)],
+        [IntAttr(23), FloatAttr(2.4, Float32Type())],
     ],
 )
 def test_target_device_spec(
-    entries: list[tuple[StringAttr | TypeAttribute | str, Attribute | str | int]],
+    entries: list[Attribute],
 ):
     # Test initialisation from a dictionary
     generic_specification_test(
-        TargetDeviceSpecAttr, {e[0]: e[1] for e in entries}, entries
+        TargetDeviceSpecAttr,
+        {"key_" + str(idx): e for idx, e in enumerate(entries)},
+        entries,
     )
     # Test initialisation from an array attribute of data layout entries
     generic_specification_test(
         TargetDeviceSpecAttr,
-        ArrayAttr([DataLayoutEntryAttr(e[0], e[1]) for e in entries]),
+        ArrayAttr(
+            [DataLayoutEntryAttr("key_" + str(idx), e) for idx, e in enumerate(entries)]
+        ),
         entries,
     )
 
@@ -185,21 +131,25 @@ def test_target_device_spec(
 @pytest.mark.parametrize(
     "entries",
     [
-        [("key1", "value1"), ("key2", 23), (i32, "test")],
-        [("k", IntAttr(23)), (i32, FloatAttr(2.4, Float32Type()))],
+        [StringAttr("value1"), IntAttr(23), IntAttr(43)],
+        [IntAttr(23), FloatAttr(2.4, Float32Type())],
     ],
 )
 def test_target_system_spec(
-    entries: list[tuple[StringAttr | TypeAttribute | str, Attribute | str | int]],
+    entries: list[Attribute],
 ):
     # Test initialisation from a dictionary
     generic_specification_test(
-        TargetSystemSpecAttr, {e[0]: e[1] for e in entries}, entries
+        TargetSystemSpecAttr,
+        {"key_" + str(idx): e for idx, e in enumerate(entries)},
+        entries,
     )
     # Test initialisation from an array attribute of data layout entries
     generic_specification_test(
         TargetSystemSpecAttr,
-        ArrayAttr([DataLayoutEntryAttr(e[0], e[1]) for e in entries]),
+        ArrayAttr(
+            [DataLayoutEntryAttr("key_" + str(idx), e) for idx, e in enumerate(entries)]
+        ),
         entries,
     )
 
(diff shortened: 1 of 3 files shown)


xdsl/dialects/dlti.py
```diff
@@ -47,12 +71,39 @@ class DLTIEntryMap(ParametrizedAttribute, ABC):
...
             entry = parser.parse_attribute()
+            end_pos = parser.pos
+            if not (
+                isinstance(entry, str)
+                or isinstance(entry, StringAttr)
+                or isinstance(entry, TypeAttribute)
+            ):
+                parser.raise_error(
+                    "key must be a string or a type attribute", pos, end_pos
+                )
```
superlopuh: If it can only be a StringAttr or TypeAttribute then we should constrain it at the point of declaration instead of in the parser

superlopuh: if you added this check for type reasons then I would actually recommend using `DataLayoutEntryAttr.new(entry, value)` below

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,270 @@
...
+    entry = DataLayoutEntryAttr(StringAttr("k"), StringAttr("v"))
+    assert isinstance(entry.key, StringAttr)
+    assert isinstance(entry.value, StringAttr)
+    assert entry.key.data == "k"
+    assert entry.value.data == "v"
+
+
+def test_incorrect_data_layout_entry():
+    with pytest.raises(VerifyException):
+        entry = DataLayoutEntryAttr(12, "V")  # pyright: ignore
```
superlopuh: we tend not to add tests like this since the assumption is that the user is using a type checker that will prevent them from calling the function with the wrong type anyway, the overall approach is that programs with pyright errors have undefined behaviour

xdsl/dialects/dlti.py
```diff
@@ -30,6 +37,23 @@ class DataLayoutEntryAttr(ParametrizedAttribute):
...
+            value = StringAttr(value)
+        elif isinstance(value, int):
+            value = IntegerAttr(value, i32)
+        elif isinstance(value, float):
+            value = FloatAttr(value, Float32Type())
+
+        super().__init__(key, value)
+
     def verify(self) -> None:
         if not isinstance(self.key, StringAttr | TypeAttribute):
```
superlopuh, line 83: feels like this should also be a constraint on the key directly?

superlopuh: Maybe also with a new `NonEmptyStringAttr` constraint to encode non-empty string?

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,270 @@
...
+    )
+    # Test initialisation from an array attribute of data layout entries
+    generic_specification_test(
+        MapAttr, ArrayAttr([DataLayoutEntryAttr(e[0], e[1]) for e in entries]), entries
+    )
+
+
+def test_duplicate_spec_entries():
+    with pytest.raises(VerifyException):
+        entry = DataLayoutSpecAttr(
```
superlopuh: if this raises then the cases below won't be covered

xdsl/dialects/dlti.py
superlopuh: Following up on this from our conversation in Zulip, I forgot about our inability to handle TypeAttribute in unions, but I would still recommend getting rid of this error case and replacing the init call with `new`

xdsl/dialects/dlti.py
```diff
@@ -30,6 +37,23 @@ class DataLayoutEntryAttr(ParametrizedAttribute):
     key: Attribute
     value: Attribute
```
superlopuh: This we could constrain here directly
```suggestion
    value: StringAttr | IntegerAttr | FloatAttr[Float32Type]
```

xdsl/dialects/dlti.py
```diff
@@ -47,12 +71,39 @@ class DLTIEntryMap(ParametrizedAttribute, ABC):
...
+        if isinstance(contents, dict):
+            contents = ArrayAttr(
+                [DataLayoutEntryAttr(k, v) for k, v in contents.items()]
+            )
+        elif isinstance(contents, list):
+            contents = ArrayAttr(contents)
+
+        assert isinstance(contents, ArrayAttr)
+
+        super().__init__(contents)
```
superlopuh: I would recommend making a few changes:
1. only allowing two posible types, the actual attribute to just assign or a pretty python type
2. Only allowing immutable types as inputs as these will be covariant, and will avoid pyright headaches down the road
```suggestion
        contents: ArrayAttr[DataLayoutAttr]
        | Mapping[StringAttr | TypeAttribute | str, StringAttr | IntegerAttr | FloatAttr[Float32Type] | str | int | float],
    ):
        if not isa(contents, ArrayAttr[DataLayoutAttr]):
            contents = ArrayAttr(
                [DataLayoutEntryAttr(k, v) for k, v in contents.items()]
            )

        super().__init__(contents)
```

This requires `from typing import Mapping` at the top of the file, it's the immutable protocol version of `dict`.

tests/dialects/test_dlti.py
mesham: Thanks, I've removed this test

xdsl/dialects/dlti.py
mesham: I've removed the check here and use a cast to cast the Attribute returned from parser.parse_attribute into StringAttr | TypeAttribute to keep pyright happy

tests/dialects/test_dlti.py
mesham: Broken into separate tests :)

xdsl/dialects/dlti.py
mesham: The value can be anything and is not just strings, integers or floats (e.g. it can be types, any of the attributes of this operation etc). So I think we do need the "wildcard" Attribute here

xdsl/dialects/dlti.py
mesham: I think this is what we discussed in terms of needing Attribute on the key (for now)?

xdsl/dialects/dlti.py
mesham: I've moved to using Mapping as suggested, and have removed list[DataLayoutEntryAttr], now have  ArrayAttr[DataLayoutAttr] and the mapping as suggested here.

xdsl/dialects/dlti.py
superlopuh: Yep

xdsl/dialects/dlti.py
superlopuh: Ah, in that case I would vote against this init helper completely, I think it makes sense to convert str to StringAttr automatically, and int to IntAttr, but IntegerAttr and FloatAttr are much trickier, in terms of user expectations of what the bitwidth should be, and signedness for integers. Would it be terribly annoying to ask to first add code that used the existing automatically-generated inits and then to see how the helpers would improve it in practice? I feel like adding special cases for some types before seeing how common they would be in practice is really difficult to get right in general

mesham: Yer, that's a very good point and likely to cause some confusion. Following this, as suggested, I've updated to remove creation of FloatAttr from float, and now when passed int it creates IntAttr instead of IntegerAttr.

xdsl/dialects/dlti.py
```diff
@@ -89,6 +130,12 @@ class DataLayoutSpecAttr(DLTIEntryMap):
 
     name = "dlti.dl_spec"
 
+    def __init__(
+        self,
+        contents: ArrayAttr[DataLayoutEntryAttr] | DictValueType,
+    ):
+        return super().__init__(contents)
```
superlopuh: Does something break if this is not added? I would expect things to work without this override

xdsl/dialects/dlti.py
```diff
@@ -47,12 +75,25 @@ class DLTIEntryMap(ParametrizedAttribute, ABC):
...
     @classmethod
     def parse_parameters(
         cls, parser: AttrParser
     ) -> tuple[ArrayAttr[DataLayoutEntryAttr]]:
         def parse_entry() -> DataLayoutEntryAttr:
             entry = parser.parse_attribute()
+            entry = cast(StringAttr | TypeAttribute, entry)
             parser.parse_punctuation("=")
             value = parser.parse_attribute()
             return DataLayoutEntryAttr(entry, value)
```
superlopuh:
```suggestion
parser.parse_punctuation("=")
value = parser.parse_attribute()
return DataLayoutEntryAttr.new(entry, value)
```

xdsl/dialects/dlti.py
```diff
@@ -47,12 +75,25 @@ class DLTIEntryMap(ParametrizedAttribute, ABC):
     # In MLIR, this is a DataLayoutEntryInterface.
     entries: ArrayAttr[DataLayoutEntryAttr]
 
+    def __init__(
+        self,
+        contents: ArrayAttr[DataLayoutEntryAttr] | DictValueType,
+    ):
+        if not isa(contents, ArrayAttr[DataLayoutEntryAttr]):
+            assert isinstance(contents, Mapping)
```
superlopuh: Does pyright complain if this line is removed?

mesham: Yes, for whoever instantiates the object it reports `error: Expected 0 positional arguments` if this is removed

xdsl/dialects/dlti.py
superlopuh: Nice I think the new version should avoid confusion

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,264 @@
...
+        )
+        entry.verify()
+
+
+def test_duplicate_target_device_spec_entries():
+    with pytest.raises(VerifyException):
+        entry = TargetDeviceSpecAttr(
+            ArrayAttr([DataLayoutEntryAttr("k", "v"), DataLayoutEntryAttr("k", 12)])
+        )
+        entry.verify()
```
superlopuh: these should not be necessary, as attributes are automatically verified in the init
```suggestion
```

xdsl/dialects/dlti.py
mesham: So would I, however pyright reports `error: Expected 0 positional arguments` if this is removed

xdsl/dialects/dlti.py
mesham: Yup, `Argument type is partially unknown` without this

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,264 @@
...
+]
+
+
+def test_data_layout_entry():
+    # Test passing strings for key and value
+    entry = DataLayoutEntryAttr("test", "V")
+    assert isinstance(entry.key, StringAttr)
+    assert isinstance(entry.value, StringAttr)
+    assert entry.key.data == "test"
+    assert entry.value.data == "V"
```
superlopuh: Could you please condense the tests a little bit now that we've iterated on implementation? Could you please test directly for key/value below instead of unwrapping data?
```suggestion
    # String key and value
    entry = DataLayoutEntryAttr("test", "V")
    assert entry.key == StringAttr("test")
    assert entry.value == StringAttr("V")
```

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,264 @@
...
+    assert isinstance(entry.key, StringAttr)
+    assert isinstance(entry.value, StringAttr)
+    assert entry.key.data == "k"
+    assert entry.value.data == "v"
+
+
+def test_incorrect_data_layout_entry():
+    with pytest.raises(VerifyException):
+        entry = DataLayoutEntryAttr("", "V")
+        entry.verify()
```
superlopuh:
```suggestion
```

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def generic_test_entry_equals_defn(
```
superlopuh: I have a suspicion that this test might actually have as much code as an inlined test for all the cases we care about... If you have the time, could you please try to make it a little dumber and see if you're still happy with the coverage? I'm not sure we need to be exhaustive for all the combinations of cases, and could probably do with just representative examples

superlopuh: One thing that it seems like we could actually parametrize on is the class of the attribute, and then inline all the arguments into the test directly.

xdsl/dialects/dlti.py
superlopuh: Can you try adding `@datclass(frozen=True)` on `DLTIEntryMap`?

xdsl/dialects/dlti.py
mesham: Thanks, code updated now to follow this :)

xdsl/dialects/dlti.py
mesham: I don't think this works with Pyright, as the error `Argument of type "Mapping[StringAttr | TypeAttribute | str, Attribute | str | int | DictValueType]" cannot be assigned to parameter "entries" of type "ArrayAttr[DataLayoutEntryAttr]" in function "__init__"`

I think this is because the generated __init__ for these subclasses is typed just `ArrayAttr[DataLayoutEntryAttr]` which follows the type of `entries` variable, whereas we want `ArrayAttr[DataLayoutEntryAttr] | DictValueType` for the argument  which is then passed to the super class __init__ method for handling

superlopuh: OK I think I'll have to clone this and play with it locally, I'm fairly certain that there's some dataclass annotation that should work, maybe with `init=False`, but it's annoyingly fiddly

mesham: Yer, I tried init=False but that had the same issue - it feels like it should work and seems a bit nasty having to duplicate the init method

tests/dialects/test_dlti.py
mesham: Thanks, removed the verify throughout the tests :)

tests/dialects/test_dlti.py
mesham: Ah, nice - I have done this, make it much shorter :)

tests/dialects/test_dlti.py
mesham: I have done this, the code here is significantly shorter now - I have removed the `generic_test_entry_equals_defn` method as that is no longer needed, there is still the `generic_specification_test` method as that avoids duplication but it's pretty simple now. 

As suggested, I parameterise just on the value and the key is automatically generated (more interesting keys, such as types are already tested in `test_data_layout_entry` method so I could safely remove that from here as you suggested

xdsl/dialects/dlti.py
superlopuh: Ah sorry, we need `(init=False)` on the `irdl_attr_definition` itself, not the superclass

xdsl/dialects/dlti.py
superlopuh: Ah then it might be worth flipping the check and doing `if isinstance(contents, Mapping):` instead

tests/dialects/test_dlti.py
superlopuh: If we take a step back, what are these tests testing? The way I would expect a test for an init to look like is to create the same structure with and without the sugar and to check that they're equal, is there something else that we're trying to check for here?

xdsl/dialects/dlti.py
mesham: Done :)

xdsl/dialects/dlti.py
mesham: Ah cool, applied to the code :)

tests/dialects/test_dlti.py
mesham: It is checking that these operations are created correctly/as expected whether they are created from a dictionary or from an ArrayAttr. The definition of "correctly" is that the keys and values, including their order, that are contained within the elements of _entries_ match what is expected.

The reason I am doing it this way is to enforce this "contract" - currently ArrayAttr goes straight through and is set to _entries_ in the operation so it's fairly obviously correct, but nowhere else in the tests does it ensure/enforce that when passing an ArrayAttr to the operation the values are set correctly. So having this explicit check enforces correctness in the future as if someone was to change the code and change the handling of ArrayAttr for whatever reason in the init method, then if the expected behaviour changed this test would fail.

I could instead initialise one operation from a dict and the other from an ArrayAttr and check they are equal, but if they are both wrong in the same way then the test would pass but would still be wrong.

superlopuh: How about using `.new` as the thing to compare against? If `new` is not correct we have bigger problems, so it feels safe to use in these tests.

mesham: If you want, so you are thinking I create a third instance with .new and then do an equality check on the .entries of this compared to the .entry of what we are creating (or could I do it directly on the created attribute itself)? I assume that equality checking between ArrayAttr is supported ?

superlopuh: Yeah I think that would work nicely and be a bit clearer. Equality should just work on all attributes automatically.

xdsl/dialects/dlti.py
```diff
@@ -30,13 +41,31 @@ class DataLayoutEntryAttr(ParametrizedAttribute):
...
+            key = StringAttr(key)
+
+        if isinstance(value, str):
+            value = StringAttr(value)
+        elif isinstance(value, int):
+            value = IntAttr(value)
+        elif isinstance(value, Mapping):
+            value = MapAttr(value)
+
+        super().__init__(key, value)
```
alexarice: Any thoughts on trying the new `converter` functionality here?

mesham: > Any thoughts on trying the new `converter` functionality here?

What's that? Happy to use it if you can point me to it :)

tests/dialects/test_dlti.py
mesham: Sounds good, done :)

xdsl/dialects/dlti.py
alexarice: See the builtin `tuple` type. Essentially each parameter can define a `converter` method which essentially does these `if isinstance(key, str): key = StringAttr(key)` clauses. We haven't really explored this fully yet as we added it quite recently.

I'll leave it up to you whether you want to try it, it's always something we can change later/iterate on if wanted.

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,181 @@
...
+    "entries",
+    [
+        [StringAttr("value1"), IntAttr(23), IntAttr(43)],
+        [IntAttr(23), FloatAttr(2.4, Float32Type())],
+    ],
+)
+def test_map_attr(
+    entries: list[Attribute],
+):
+    generic_specification_test(MapAttr, entries)
```
superlopuh: I would recommend using this two-step parametrization, which produces a product of the different parametrize entries:
```suggestion
@pytest.mark.parametrize(
    "cls", [DataLayoutSpecAttr, TargetDeviceSpecAttr, TargetSystemSpecAttr, MapAttr]
)
@pytest.mark.parametrize(
    "entries",
    [
        [StringAttr("value1"), IntAttr(23), IntAttr(43)],
        [IntAttr(23), FloatAttr(2.4, Float32Type())],
    ],
)
def test_target_device_spec(
    cls: type[DLTIEntryMap],
    entries: list[Attribute],
):
    generic_specification_test(TargetDeviceSpecAttr, entries)
```
Then I would recommend inlining the `generic_specification_test` function.

tests/dialects/test_dlti.py
```diff
@@ -0,0 +1,181 @@
...
+        TargetSystemSpecAttr(
+            ArrayAttr([DataLayoutEntryAttr("k", "v"), DataLayoutEntryAttr("k", 12)])
+        )
+
+
+def test_duplicate_map_attr_entries():
+    with pytest.raises(VerifyException):
+        MapAttr(
+            ArrayAttr([DataLayoutEntryAttr("k", "v"), DataLayoutEntryAttr("k", 12)])
+        )
```
superlopuh: Here I would also recommend parametrizing on the class

xdsl/dialects/dlti.py
mesham: @alexarice  I've moved over to using converter for the three fields here, happy to keep with the converter or to go back to what we had before (whatever you think is best). Equally can change how it's structured if there is a better way

xdsl/dialects/dlti.py
```diff
@@ -45,7 +103,15 @@ class DLTIEntryMap(ParametrizedAttribute, ABC):
...
-    entries: ArrayAttr[DataLayoutEntryAttr]
+    entries: ArrayAttr[DataLayoutEntryAttr] = param_def(
+        converter=DLTITypeConverters.convert_entry_map_type
+    )
+
+    def __init__(
+        self,
+        contents: ArrayAttr[DataLayoutEntryAttr] | DictValueType,
+    ):
+        super().__init__(contents)
```
superlopuh: Is this necessary?

mesham: If we are using the converter approach I think so, as the __init__ method generated by dataclass is only typed ArrayAttr[DataLayoutEntryAttr] , whereas we also want to potentially pass a dictionary (leaving dataclass to auto-generate this results in pyright and runtime errors)

tests/dialects/test_dlti.py
mesham: Done :)

tests/dialects/test_dlti.py
mesham: Done :)

xdsl/dialects/dlti.py
superlopuh: Oh this is because `dataclass` doesn't know  about our `param_def` field and our converters... I think to avoid this kind of thing we'll have to add an abstract attr definition annotation in the future, I don't think it's a blocker CC @alexarice, @math-fehr

xdsl/dialects/dlti.py
superlopuh: With the new converters this init is generated automatically, so it can safely be deleted

xdsl/dialects/dlti.py
alexarice: Not a blocker on my end at all, just thought it could be a cool place to try stuff out.

superlopuh: Sorry I thought I'd ping you to highlight this reason to add the abstract superclass annotation maybe before we have an explicit AttributeInterface thing or whatever, since our dataclass thing is currently a bit of a leaky abstraction.

alexarice: Can't you just drop the annotation on the superclass? Not sure I fully understand the problem.

mesham:
> Looks great modulo the init on `DataLayoutEntryAttr`, thank you very much for iterating on this!

Thanks for all your help - removed the init method from `DataLayoutEntryAttr`  :)

xdsl/dialects/dlti.py
superlopuh: Oh yeah you're right, removing dataclass fixed this locally

xdsl/dialects/dlti.py
mesham: done :)

xdsl/dialects/dlti.py
mesham: Applied to the code, __init__ method now removed :)

superlopuh (approved):
Looks perfect, thanks again

PR #4954, opened 2025-08-02
transformations: (memref-stream-interleave) extract factor helpers

superlopuh:
As seein in #3724, this PR extracts some helper functions to bring us one step closer to a parametrizeable `memref-stream-interleave` pass, where the user may have better information than the heuristic as to which factor and iterator to choose.


superlopuh:
Draft until I get annotations working

superlopuh:
#4955

math-fehr (approved):

PR #4955, opened 2025-08-02
core: add support for string types in Pass definitions

superlopuh:
When we import `annotations` from `__future__`, the types are turned to strings, and our current pass validation code breaks. This PR proposes to just eval the strings when checking the type.

### DIFF
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -101,9 +101,13 @@ class ModulePass(ABC):
                 raise ValueError(f'Pass {cls.name} requires argument "{op_field.name}"')
 
             # convert pass arg to the correct type:
+            field_type = op_field.type
+            if isinstance(field_type, str):
+                # If the type was provided as a string, eval the string to get the type
+                field_type = eval(field_type, None)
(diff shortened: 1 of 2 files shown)


xdsl/passes.py
```diff
@@ -101,9 +101,13 @@ def from_pass_spec(cls, spec: PipelinePassSpec) -> Self:
                 raise ValueError(f'Pass {cls.name} requires argument "{op_field.name}"')
 
             # convert pass arg to the correct type:
+            field_type = op_field.type
+            if isinstance(field_type, str):
+                # If the type was provided as a string, eval the string to get the type
+                field_type = eval(field_type, None)
```
math-fehr: Instead of an `eval` here, can we use `field_types = typing.get_type_hints(cls)` before the loop, and then use `field_types[op_field.name]` to get the actual type?
That should be nicer than calling `eval`? I'm not sure there would be too much of a runtime cost from that?

superlopuh: Sure I'll refactor it. It does just call eval for you but I agree that it's nicer

PR #4956, opened 2025-08-03
interactive: only show individual rewrites in "condense" mode

superlopuh:
I'd like to abstract this bit of the infrastructure, and the abstraction only makes sense if we put individual rewrites and the passes that change the input together. I have a feeling this is actually better UX, because then in uncondensed mode the user doesn't pay the penalty of lots of copying and applying patterns on every click.


alexarice (approved):
Sounds reasonable.

PR #4957, opened 2025-08-03
dialects: (tosa) add support for `tosa.cond_if`, `tosa.yield`

jakedves:
Adds support for the TOSA conditional and yield operations, although doesn't follow the MLIR printing as xDSL dynamically provides SSA names. The `IfOp` class takes a lot of inspiration from the `scf.if` implementation, but I didn't think it made sense to write shared functions for them.


superlopuh (commented):
I just realized that the with-mlir filecheck got out of sync with the new operations we've been adding, could you please extend it with these and all the other missing ops?

jakedves:
Didn't add `mlir-conversion` tests for the `cond_if` as that's targeting a newer MLIR (the current version didn't have any examples with the operands list and had a custom syntax)

superlopuh:
Hmm, we should be in sync with the MLIR version indicated in the README, do you have requirements to use a different version of MLIR?

PR #4959, opened 2025-08-04
interactive: display pass spec in gui always

superlopuh:
Currently our individual rewrites are displayed differently, but I'm not sure that the extra logic is pulling its weight, especially since they are printed differently once they are clicked on and added to the pass pipeline.


dshaaban01 (approved):
Yep, way better!

alexarice (approved):

PR #4960, opened 2025-08-04
dialects: (vector) vector broadcast from vector

superlopuh:


xdsl/dialects/vector.py
```diff
@@ -132,6 +132,10 @@ def get(
 
 @irdl_op_definition
 class BroadcastOp(IRDLOperation):
+    """
+    See external [documentation](https://mlir.llvm.org/docs/Dialects/Vector/#vectorbroadcast-vectorbroadcastop).
+    """
+
     name = "vector.broadcast"
     source = operand_def()
```
alexarice, line 140: Could we not do something like:
```
T: ClassVar = VarConstraint("T", AnyVectorElement)

...
source = operand_def(T | VectorType[T])
```
?

alexarice (approved):
Seems fine though would be nice to use constraints here if we can.

xdsl/dialects/vector.py
superlopuh: The problem is that there isn't really a constraint on vector elements, so we can't prove that these are disjoint...

alexarice: I thought vector elements were a fairly small set of types

superlopuh: In MLIR there's a `VectorElementTypeInterface`, we don't have any constraints, in either case it seems like it's an open thing, which we currently don't support in AnyOf.

alexarice: Maybe we should just revert the `AnyOf` change.

superlopuh: I think it's worth chatting about but I wouldn't rush into reverting it.

PR #4961, opened 2025-08-04
core: Add permute method on Block

emmau678:
Add a method on Block for permuting the operations based on a list of ints specifying the new order


xdsl/ir/core.py
```diff
@@ -1681,6 +1681,15 @@ def parent_block(self) -> Block | None:
     def __repr__(self) -> str:
         return f"Block(_args={repr(self._args)}, num_ops={len(self.ops)})"
 
+    def permute(self, orderings: list[int]) -> None:
```
```suggestion
    def permute(self, orderings: Sequence[int]) -> None:
```
I don't see the need to take a `list` specifically here.

xdsl/ir/core.py
```diff
@@ -1681,6 +1681,15 @@ def parent_block(self) -> Block | None:
...
+        list_ops: list[Operation] = []
```
```suggestion
        tuple_ops: tuple(self.ops)
```

xdsl/ir/core.py
```diff
@@ -1681,6 +1681,15 @@ def parent_block(self) -> Block | None:
...
+        for op in self.ops:
+            op.detach()
+            list_ops.append(op)
```
```suggestion
```

xdsl/ir/core.py
alexarice: Feels like it makes sense to use a tuple here

xdsl/ir/core.py
```diff
@@ -1681,6 +1681,15 @@ def parent_block(self) -> Block | None:
...
+        reordered_ops = [list_ops[i] for i in orderings]
```
```suggestion
```

alexarice: Can just directly iterate below instead of making a temporary list

xdsl/ir/core.py
```diff
@@ -1681,6 +1681,15 @@ def parent_block(self) -> Block | None:
...
+        for op in reordered_ops:
+            self.add_op(op)
```
```suggestion
        for i in orderings:
            self.add_op(tuple_ops[i])
```

tests/test_core.py
superlopuh, line 1: can you please move this test to the other Block helper unit tests?

tests/test_core.py
```diff
@@ -0,0 +1,26 @@
...
+
+def test_permute_block_ops():
+    a = ConstantOp(IntegerAttr.from_int_and_width(1, 32), i32)
+    b = ConstantOp(IntegerAttr.from_int_and_width(2, 32), i32)
+
+    # Operations on these constants
+    c = AddiOp(a, b)
+    d = AddiOp(a, b)
+    e = AddiOp(c, d)
+    f = PrintOp.get(e)
```
superlopuh: can you please use test ops instead?

superlopuh:
I suggested adding it on Block directly, since I feel like it could be a little tricky to get right, with all the detaching etc. It could also be written more efficiently if one wanted to modify the linked list thing directly. I would also be happy with it on the rewriter if you feel that's better, or in a utils file if you think that would be more appropriate. The problem would then be discoverability.

superlopuh:
Emma, could you please handle cases where the new indices passed in aren't a valid permutation? + tests

alexarice:
> I suggested adding it on Block directly, since I feel like it could be a little tricky to get right, with all the detaching etc. It could also be written more efficiently if one wanted to modify the linked list thing directly. I would also be happy with it on the rewriter if you feel that's better, or in a utils file if you think that would be more appropriate. The problem would then be discoverability.

I don't feel too strongly about this, feel free to go with whatever

xdsl/ir/core.py
```diff
@@ -1681,6 +1681,13 @@ def parent_block(self) -> Block | None:
     def __repr__(self) -> str:
         return f"Block(_args={repr(self._args)}, num_ops={len(self.ops)})"
 
+    def permute(self, orderings: Sequence[int]) -> None:
+        tuple_ops: tuple[Operation, ...] = tuple(self.ops)
```
superlopuh:
```suggestion
        tuple_ops = tuple(self.ops)
```

tests/test_core.py
emmau678: I can only find ones that look similar in test_operation_builder.py so I will move there for now, please let me know if I have missed another file

superlopuh: `test_ir.py` would be my first instinct

tests/irdl/test_operation_builder.py
```diff
@@ -759,3 +760,24 @@ def test_parent_pointers():
...
+    # Create Block from operations
+    block0 = Block([a, b, c, d, e, f])
+    orderings = [2, 5, 0, 4, 1, 3]
+    block0.permute(orderings)
+
+    expected_order = [c, f, a, e, b, d]
+    expected_iter = iter(expected_order)
+    for actual_op in block0.ops:
+        expected_op = next(expected_iter)
+        assert actual_op is expected_op
```
superlopuh: Does this work?
```suggestion
    assert tuple(block0.ops) == (c, f, a, e, b, d)
```

emmau678: yep thank you!

math-fehr:
Looks good to me, but I really think this should be added to some `utils/` file somewhere.
In my opinion, this is such a special use case that I probably wouldn't even put it in `xdsl` but in the dialect/pass using it, hence why I'm suggesting `utils/`

emmau678:
thanks all for the comments. I chatted with @superlopuh and we are going to move this to our experiments repo instead, so I'll close this PR

PR #4962, opened 2025-08-04
core: unify VarOperand and VarOpResult in to SSAValues

superlopuh:
I'm not sure that we benefit from these being separate

### DIFF
--- a/xdsl/dialects/cf.py
+++ b/xdsl/dialects/cf.py
@@ -18,7 +18,6 @@ from xdsl.irdl import (
     AttrSizedOperandSegments,
     IRDLOperation,
     Successor,
-    VarOperand,
     attr_def,
     irdl_op_definition,
     operand_def,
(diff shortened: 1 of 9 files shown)


xdsl/dialects/cf.py
```diff
@@ -395,11 +394,11 @@ def __init__(
         )
 
     @property
-    def case_operand(self) -> tuple[VarOperand, ...]:
-        cases: list[VarOperand] = []
+    def case_operand(self) -> tuple[Sequence[SSAValue], ...]:
```
alexarice: Why not wrap with `SSAValues` here?

xdsl/dialects/omp.py
```diff
@@ -134,7 +135,7 @@ def __iter__(self):
 
 
 def verify_map_vars(
-    vars: VarOperand,
+    vars: Sequence[SSAValue],
```
alexarice: Same here

alexarice (commented):
This seems reasonable to me. The alternative here is to have `SSAValues` as you have defined it, but then have type aliases:
`VarOperand = SSAValues`
`VarOpResult = SSAValues[OpResult[Attribute]]`
Thoughts? Should be a much smaller diff.

superlopuh:
Ah sure I can do that

PR #4963, opened 2025-08-04
core: Operation operands and results are SSAValues

superlopuh:
This means that `.types` are always available, including for slices of operands and results. I propose discussing removing the helper wrapper after this PR is merged.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -939,22 +939,28 @@ class OpTraits(Iterable[OpTrait]):
     An operation's traits.
     Some operations have mutually recursive traits, such as one is always the parent
     operation of the other.
-    For this case, the operation's traits can be declared lazily, and resolved only
-    at the first use.
+    For this case, the operation's traits can be declared lazily and resolved only at
+    the first use.
     """
 
-    _traits: frozenset[OpTrait] | Callable[[], tuple[OpTrait, ...]]
+    gen_traits: Callable[[], tuple[OpTrait, ...]]
+    """
+    Factory method that lazily populates the traits on first use.
+    """
+    _traits: frozenset[OpTrait] | None
+    """
+    The traits of this operation, can be updated via `add_trait`.
+    """
 
-    def __init__(
-        self, traits: frozenset[OpTrait] | Callable[[], tuple[OpTrait, ...]]
-    ) -> None:
-        self._traits = traits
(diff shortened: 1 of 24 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -608,7 +608,7 @@ def parse_types(self, parser: Parser, state: ParsingState) -> bool:
     def parse_single_type(self, parser: Parser, state: ParsingState) -> None:
         state.operand_types[self.index] = (parser.parse_type(),)
 
-    def get(self, op: IRDLOperation) -> VarOperand:
+    def get(self, op: IRDLOperation) -> SSAValues:
```
alexarice: Unrelated change?

xdsl/transforms/experimental/lower_hls.py
```diff
@@ -327,10 +326,10 @@ def match_and_rewrite(self, op: ParallelOp, rewriter: PatternRewriter, /):
...
         hls_pipeline_op: Operation = PragmaPipelineOp(ii)
 
-        lb: VarOperand = op.lowerBound
-        ub: VarOperand = op.upperBound
-        step: VarOperand = op.step
-        res: VarOpResult = op.res
+        lb = op.lowerBound
+        ub = op.upperBound
+        step = op.step
+        res = op.res
```
alexarice: Also unrelated change?

math-fehr (approved):
This adds the overhead we want to measure, right?
Can we just measure it in that PR?

xdsl/ir/core.py
```diff
@@ -971,7 +985,7 @@ class Operation(_IRNode):
     access to the operand SSAValue.
     """
 
-    results: tuple[OpResult, ...] = field(default=())
+    results: SSAValues[OpResult] = field(default=SSAValues())
```
alexarice, line 994: What do you think about adding a setter for results that can wrap an arbitrary tuple of ssa values?

alexarice (approved):
This seems like a good improvement to me

xdsl/ir/core.py
superlopuh: I'm not sure of the advantage since this should almost never happen

alexarice: you're right, it only seems to happen in "internal" things

PR #4964, opened 2025-08-04
core: Deprecate VarOperand.types

alexarice:


math-fehr:
Did we actually measure if this has a performance impact or not?
I would say that we should only deprecate it if we want to delete `VarOperand`, otherwise this is still a nice API call? Unless you feel different about it?

alexarice:
This likely isn't worth doing. We've solved the original problem now so don't really need this, we can revisit if someone finds this is actually performance penalty I guess

PR #4966, opened 2025-08-04
transformations: add split-load-extract for manual optimization

superlopuh:
This transformation doesn't always make sense to perform, but it does in the case of vectorisation of matmul for NEON and AVX. it feels like this might make sense eventually as a transform thing but until then I propose we add a pass.

### DIFF
--- a/tests/filecheck/transforms/vector-split-load-extract.mlir
+++ b/tests/filecheck/transforms/vector-split-load-extract.mlir
@@ -1,0 +1,34 @@
+// RUN: xdsl-opt -p vector-split-load-extract %s | filecheck %s
+
+// CHECK-LABEL: @do_split
+func.func @do_split(%ptr : !ptr_xdsl.ptr) {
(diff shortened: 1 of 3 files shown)


qaco:
I agree it is an optimization, but I don't understand in which (healthy) case we generate such a vector/extract pattern ?

superlopuh:
This is a result of the default vectorization of linalg.matmul with no scheduling. I'd be glad to have this done as part of a schedule instead of in a pass, but until we find such a schedule this addition seems OK to me? We can always remove it once we find an alternative.

qaco (approved):

qaco:
> This is a result of the default vectorization of linalg.matmul with no scheduling. I'd be glad to have this done as part of a schedule instead of in a pass, but until we find such a schedule this addition seems OK to me? We can always remove it once we find an alternative.

No no it's fine no need to remove. I juste wanted to be sure I'm ok with the rationale.

PR #4968, opened 2025-08-05
transformations: (pdl_interp) optimize matcher by moving equality checks

jumerckx:
~This pass detects `check_operation_name` operations applied on the result of a `get_defining_op` and duplicates these as close to the `get_defining_op` as possible (in a block right after the `is_not_null`).~

~Opening this as a draft for now as this still requires tests and it might be superseded by a pass that moves equality constraints up.~

This pr now contains the optimisation that tries to move `pdl_interp.are_equal` as close as possible to their respective `pdl_interp.get_defining_op`.


xdsl/transforms/eqsat_optimize_pdl_interp.py
```diff
@@ -0,0 +1,155 @@
...
+
+        assert (are_equal_block := are_equal_op.parent_block())
+        blocks_to_move = [are_equal_block]
+
+        # We can always pick only one of the predecessors to move up.
+        # Forked paths will never define values that are used after paths
+        # are joined again, since there are no block arguments/phi nodes.
+        # The final destination of the blocks will always be along a
+        # non-split path at the post-dominator frontier:
+        assert (incoming_edge := next(iter(are_equal_block.uses), None)) is not None
```
superlopuh:
```suggestion
        assert (incoming_edge := are_equal_block.get_unique_use()) is not None
```

xdsl/transforms/eqsat_optimize_pdl_interp.py
```diff
@@ -0,0 +1,155 @@
...
+        incoming_edges: list[Use] = [incoming_edge]
+
+        outgoing_edges = [are_equal_op._successor_uses[0]]  # pyright: ignore[reportPrivateUsage]
```
superlopuh: Why use the private API here?

xdsl/transforms/eqsat_optimize_pdl_interp.py
```diff
@@ -0,0 +1,155 @@
...
+
+        block_dependencies: set[SSAValue] = set()
+        for op in are_equal_block.ops:
+            for result in op.operands:
+                block_dependencies.add(result)
```
superlopuh:
```suggestion
        block_dependencies: set[SSAValue] = {
            result
            for op in are_equal_block.ops
            for result in op.operands
        }
```

xdsl/transforms/eqsat_optimize_pdl_interp.py
superlopuh: Actually could this be `get_user_of_unique_use()`? Do we ever use the use itself, as opposed to the user?

jumerckx: The thing is there can be multiple users, we're fine as long as we find one.

xdsl/transforms/eqsat_optimize_pdl_interp.py
jumerckx: Out of confusion. But what should it be here?
`are_equal_op.true_dest.successors[0].uses[...?]` or am I missing something?

xdsl/transforms/eqsat_optimize_pdl_interp.py
superlopuh: ah then how about `first_use`?

xdsl/transforms/eqsat_optimize_pdl_interp.py
superlopuh: I guess so

jumerckx: But then we'd need to iterate over the uses to find the one coming from `are_equal_op`?

superlopuh: hmm, indeed, maybe actually we just need to expose this as public API? CC @math-fehr

xdsl/transforms/eqsat_optimize_pdl_interp.py
jumerckx: That works!

xdsl/transforms/eqsat_optimize_pdl_interp.py
```diff
@@ -0,0 +1,154 @@
...
+class EqsatOptimizePDLInterp(ModulePass):
+    name = "eqsat-optimize-pdl-interp"
+
+    def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
+        matcher = SymbolTable.lookup_symbol(op, "matcher")
+        assert isinstance(matcher, pdl_interp.FuncOp)
+        assert matcher is not None, "matcher function not found"
+
+        mo = MatcherOptimizer(matcher)
+        mo.optimize()
```
superlopuh, line 29: this seems to be unused? Do we want to expose this as a pass? If yes, can you please add a test?

xdsl/transforms/eqsat_optimize_pdl_interp.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@dataclass
+class MatcherOptimizer:
+    matcher: pdl_interp.FuncOp
```
superlopuh: Can you please add a doc string here explaining what's happening? Maybe also another one at the top of the file that motivates this, that way we first introduce the need and then explain this concrete implementation of the optimisation.

xdsl/transforms/eqsat_optimize_pdl_interp.py
```diff
@@ -0,0 +1,154 @@
...
+        ):
+            assert outgoing.operation.parent_block() is block
+            if incoming.operation is insertion_edge.operation:
+                insertion_edge = outgoing
+                continue
+            insertion_dest = insertion_edge.operation.successors[insertion_edge.index]
+            insertion_edge.operation.successors[insertion_edge.index] = block
+            outgoing_dest = outgoing.operation.successors[outgoing.index]
+            outgoing.operation.successors[outgoing.index] = insertion_dest
+            if incoming.operation != insertion_edge.operation:
```
jumerckx: There's a bug here.
By replacing the successor here, the original use in the block is replaced by a new one. At this point, the original use can still be part of `incoming_edges`, but the operation it points to is now outdated. This is the case when subsequent blocks are part of `blocks_to_move`, so quite often. I'm surprised things didn't error, I would be surprised if this doesn't change semantics of the matcher.

`incoming` should always contain the up-to-date use of `block`, this means we probably need to keep `incoming_edges` up-to-date as we go.

superlopuh: I'm not sure I understand, you mean the bug is in this code or somewhere in the rewriting functionality?

jumerckx: The bug is in this code, I was using the use as an edge that is kept up-to-date w.r.t to the block, but (of course) it's kept up to date w.r.t the operation.
I wrote this comment mainly for myself to understand the problem better, I think I know how to fix it.

jumerckx:
Converted to draft as we'll try do this in the pdl-to-pdl-interp pass directly

jumerckx:
closing in favor of a custom pdl-to-pdl-interp pass.

PR #4969, opened 2025-08-05
interactive: pass context to iter_condensed_passes

superlopuh:
No need to create a new context every time, and lets the user choose the dialects to look for. Importantly, it removes the dependency on get_all_dialects in iter_condensed_passes, which currently prevents the app from accessing dialects not in core xDSL.

### DIFF
--- a/xdsl/interactive/passes.py
+++ b/xdsl/interactive/passes.py
@@ -2,7 +2,7 @@ from collections.abc import Callable
 from typing import NamedTuple
 
 from xdsl.context import Context
-from xdsl.dialects import builtin, get_all_dialects
+from xdsl.dialects import builtin
 from xdsl.ir import Dialect
 from xdsl.passes import ModulePass
 from xdsl.transforms.mlir_opt import MLIROptPass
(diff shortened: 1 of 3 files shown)


xdsl/interactive/passes.py
```diff
@@ -37,14 +37,10 @@ def get_new_registered_context(
 
 
 def iter_condensed_passes(
+    ctx: Context,
     input: builtin.ModuleOp,
     all_passes: tuple[tuple[str, type[ModulePass]], ...],
 ):
-    ctx = Context(True)
-
-    for dialect_name, dialect_factory in get_all_dialects().items():
```
alexarice, line 45: Wow, I thought I got rid of all uses of `get_all_dialects()` when I was doing stuff downstream. I obviously missed one.

xdsl/interactive/passes.py
```diff
@@ -63,11 +59,12 @@ def iter_condensed_passes(
 
 
 def get_condensed_pass_list(
```
alexarice, line 61: aside, is this function really necessary?

superlopuh: Nope and it's leaving soon

PR #4972, opened 2025-08-05
core: move out KIND_BY_PUNCTUATION_SPELLING in MLIR lexer

superlopuh:
Saves 6% = 1-0.782s/0.834s when loading arith

### DIFF
--- a/xdsl/utils/mlir_lexer.py
+++ b/xdsl/utils/mlir_lexer.py
@@ -170,41 +170,14 @@ class MLIRTokenKind(Enum):
     FILE_METADATA_BEGIN = "{-#"
     FILE_METADATA_END = "#-}"
 
-    @staticmethod
-    def get_punctuation_spelling_to_kind_dict() -> dict[str, MLIRTokenKind]:
-        return {
-            "->": MLIRTokenKind.ARROW,
-            ":": MLIRTokenKind.COLON,
-            ",": MLIRTokenKind.COMMA,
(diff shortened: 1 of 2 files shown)


xdsl/utils/mlir_lexer.py
```diff
@@ -170,41 +170,14 @@ class MLIRTokenKind(Enum):
     FILE_METADATA_BEGIN = "{-#"
     FILE_METADATA_END = "#-}"
 
-    @staticmethod
-    def get_punctuation_spelling_to_kind_dict() -> dict[str, MLIRTokenKind]:
-        return {
```
superlopuh, line 175: creates a new dictionary each time we check whether a token is punctuation then throws it away

xdsl/utils/mlir_lexer.py
```diff
@@ -250,6 +223,30 @@ def get_string_literal_value(self, span: Span) -> str:
         return StringLiteral.from_span(span).string_contents
 
 
+KIND_BY_PUNCTUATION_SPELLING = {
```
alexarice, line 226: Worth considering having this as a ClassVar on `MLIRTokenKind`?

superlopuh: It's not happy when I try to do this directly, as the enums are not fully initialised until the end of the class definition, and doing some kind of setting a class var after the class definition also seems a little off.

PR #4973, opened 2025-08-05
core: add OpSelector for easier matching

superlopuh:
Not sure about naming or placement in code. We probably want two systems, one nice query system for navigating IR, and this hammer that's not very elegant but necessary in cases, it would be nice for this API to be available but not super prominent.


alexarice (approved):
Nice

math-fehr (commented):
Looks good to me, but I would move it outside of `core.py` to the `utils` folder (because that's kind of niche, and I would like `core.py` to stay small-ish).

Also, for later, you will probably want to change your selector to be recursive, so you don't have to travel an entire region if you know your operation is not in it.

PR #4974, opened 2025-08-06
misc: use contextmanager annotation for ImplicitBuilder

superlopuh:
We never actually used the stack, it's only ever one current builder, the others can be stored in the contextmanager scope.

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -4,8 +4,7 @@ import contextlib
 import threading
 from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass, field
-from types import TracebackType
-from typing import ClassVar, TypeAlias, overload
+from typing import TypeAlias, overload
 
 from typing_extensions import TypeVar
 
@@ -75,7 +74,7 @@ class Builder(BuilderListener):
         if not ops:
             return ops
 
-        implicit_builder = ImplicitBuilder.get()
+        implicit_builder = _implicit_builder.builder
         if implicit_builder is not None and implicit_builder is not self:
             raise ValueError(
                 "Cannot insert operation explicitly when an implicit builder exists."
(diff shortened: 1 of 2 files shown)


xdsl/builder.py
```diff
@@ -263,30 +262,24 @@ def assert_implicit():
 
 
 @dataclass
-class _ImplicitBuilderStack(threading.local):
+class _ImplicitBuilder(threading.local):
     """
     Stores the stack of implicit builders for use in @Builder.implicit_region, empty by
```
alexarice: docstring needs updating

xdsl/builder.py
```diff
@@ -263,30 +262,24 @@ def assert_implicit():
...
-    stack: list[Builder] = field(default_factory=list[Builder])
+    builder: Builder | None = None
 
-    def push(self, builder: Builder) -> None:
-        self.stack.append(builder)
 
-    def get(self) -> Builder | None:
-        if len(self.stack):
-            return self.stack[-1]
+_implicit_builder = _ImplicitBuilder()
```
alexarice: If this is a constant, should it be `_IMPLICIT_BUILDER`?

xdsl/builder.py
```diff
@@ -263,30 +262,24 @@ def assert_implicit():
...
 
-    def pop(self, builder: Builder) -> Builder:
-        popped = self.stack.pop()
-        assert popped is builder
-        return popped
 
-
-class ImplicitBuilder(contextlib.AbstractContextManager[tuple[BlockArgument, ...]]):
+@contextlib.contextmanager
+def ImplicitBuilder(
```
alexarice, line 278: Feels strange to have a function named like a class.

alexarice (approved):
Seems reasonable

xdsl/builder.py
superlopuh: It's not really a constant as it's the shared mutable state for the mechanism

alexarice: True

xdsl/builder.py
superlopuh: Yeah I wanted minimal disruption in this PR as I think it might make sense to rethink the API altogether.

PR #4975, opened 2025-08-06
transformations: (eqsat) eqsat-create-egraphs

jumerckx:
This pass creates an `eqsat.egraph` operation for each function. The egraph contains eclasses wrapping the function arguments as well as all other operations that only depend on eclasses (except `func.return`).

### DIFF
--- a/tests/filecheck/transforms/eqsat-create-egraphs.mlir
+++ b/tests/filecheck/transforms/eqsat-create-egraphs.mlir
@@ -1,0 +1,19 @@
+// RUN: xdsl-opt -p eqsat-create-egraphs %s | filecheck %s
+
+func.func @test(%x : index) -> (index) {
+  %c2 = arith.constant 2 : index
+  %res = arith.muli %x, %c2 : index
+  func.return %res : index
+}
+
+// CHECK:      func.func @test(%x : index) -> index {
+// CHECK-NEXT:   %res = eqsat.egraph -> index {
+// CHECK-NEXT:     %x_1 = eqsat.eclass %x : index
+// CHECK-NEXT:     %c2 = arith.constant 2 : index
+// CHECK-NEXT:     %c2_1 = eqsat.eclass %c2 : index
+// CHECK-NEXT:     %res_1 = arith.muli %x_1, %c2_1 : index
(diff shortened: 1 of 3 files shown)


xdsl/transforms/eqsat_create_egraphs.py
```diff
@@ -0,0 +1,99 @@
...
+    egraph_block = Block()
+    egraph_body = Region(egraph_block)
+    egraph_values: OrderedSet[OpResult] = OrderedSet(())
+
+    def create_eclass(val: SSAValue):
+        eclass_op = eqsat.EClassOp(val)
+        egraph_block.add_op(eclass_op)
+        new_val = eclass_op.results[0]
+        egraph_values.add(new_val)
+        val.replace_by_if(new_val, lambda u: u.operation != eclass_op)
```
superlopuh: nit:
```suggestion
        val.replace_by_if(new_val, lambda u: u.operation is not eclass_op)
```

tests/filecheck/transforms/eqsat-create-egraphs.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: xdsl-opt -p eqsat-create-egraphs %s | filecheck %s
+
+func.func @test(%x : index) -> (index) {
+  %c2 = arith.constant 2 : index
+  %res = arith.muli %x, %c2 : index
+  func.return %res : index
+}
```
superlopuh: can you please add a test case with a for loop and a `"test.op"`?

superlopuh: I would put these in separate functions.

Also, can you please put the check lines above the functions in this file? We've been trying to do that for new tests since that's the convention in MLIR

jumerckx: I added two new tests, or did you want only one new one with a loop containing a `test.op`?

superlopuh (approved):
Thank you!

tests/filecheck/transforms/eqsat-create-egraphs.mlir
superlopuh: This is perfect

PR #4976, opened 2025-08-06
dialects: (tensor) Add splat operation

mesham:
Splat operation in tensor dialect to broadcast a constant into a tensor

### DIFF
--- a/tests/filecheck/dialects/tensor/invalid_ops.mlir
+++ b/tests/filecheck/dialects/tensor/invalid_ops.mlir
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt --verify-diagnostics --split-input-file %s | filecheck %s
+// RUN: xdsl-opt --parsing-diagnostics --verify-diagnostics --split-input-file %s | filecheck %s
 
 // Verification checks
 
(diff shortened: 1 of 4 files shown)


tests/filecheck/dialects/tensor/invalid_ops.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: xdsl-opt --verify-diagnostics --split-input-file %s | filecheck %s
+// RUN: xdsl-opt --parsing-diagnostics --verify-diagnostics --split-input-file %s | filecheck %s
```
superlopuh, line 1: Why this change?

mesham: The way in which we do the element type in the operation (below) is picked up as an error by the parser, not by the verifier. So that last test, passing f32 to a tensor of element type i32, is picked up by the parser

```
SPLAT_TYPE: ClassVar = VarConstraint("SPLAT_TYPE", AnyAttr())
input = operand_def(SPLAT_TYPE)
result = result_def(TensorType.constr(SPLAT_TYPE))
```

I do it this way for the assembly-format, having an operand_def(Attribute) instead it complains that `input` has no type, I see the `FromElementsOp` does the same so have adopted that same approach here.

superlopuh: Ah I see, that makes sense

PR #4980, opened 2025-08-06
frontend: (pyast) add post transforms list to builder

EdmundGoodman:
This PR replaces the fixed desymref pass on the pyast builder with a configurable list of post transformations to apply to the generated IR

### DIFF
--- a/xdsl/frontend/pyast/context.py
+++ b/xdsl/frontend/pyast/context.py
@@ -32,6 +32,13 @@ class FuncInfo(NamedTuple):
     """The Python AST representation of the function."""
 
 
+def default_pipeline_callback(
+    _previous_pass: ModulePass, module: ModuleOp, _next_pass: ModulePass
+) -> None:
+    """Default callback to verify the module after each transformation pass."""
+    module.verify()
+
+
 @dataclass
 class PyASTContext:
     """Encapsulate the mapping between Python and IR types and operations."""
@@ -47,7 +54,9 @@ class PyASTContext:
     )
     """An ordered list of passes to apply to the built module."""
 
-    post_callback: Callable[[ModulePass, ModuleOp, ModulePass], None] | None = None
+    post_callback: Callable[[ModulePass, ModuleOp, ModulePass], None] | None = (
+        default_pipeline_callback
+    )
     """Callback to run between post transforms."""
 
     ir_context: Context = field(


xdsl/frontend/pyast/utils/builder.py
```diff
@@ -1,15 +1,26 @@
...
+
+
+class PostTransform(NamedTuple):
+    """Transformation to apply to built IR."""
+
+    module_pass: ModulePass
+    """The module pass to apply."""
+
+    verify: bool = True
+    """Whether to verify the result of the pass."""
```
EdmundGoodman: Currently some post transforms we want to be able to do (such as making pld rewrites which rely on test ops instead of pdl match ops) generate code which does not verify. This wrapper tuple lets you selectively disable verification on these intermediate passes.

This might be something to think about avoiding in future as it breaks the contract of passes necessarily mapping verifiable to verifiable IR, but for now I think this is what we want

xdsl/frontend/pyast/context.py
```diff
@@ -38,6 +41,16 @@ class PyASTContext:
...
 
+    post_transforms: list[PostTransform] = field(
+        default_factory=lambda: [PostTransform(FrontendDesymrefyPass())]
+    )
+    """An ordered list of passes to apply to the built module."""
+
+    ir_context: Context = field(
+        default_factory=lambda: Context(allow_unregistered=True)
+    )
+    """The xDSL context to use when applying transformations to the built module."""
```
EdmundGoodman, line 65: I think the xdsl context for the post transformations should be part of the pyast context. I wonder if there is a nice way for registering types/operations to automagically pull in their associated dialects to the context?

xdsl/frontend/pyast/context.py
```diff
@@ -38,6 +41,16 @@ class PyASTContext:
     function_registry: FunctionRegistry = field(default_factory=FunctionRegistry)
     """Mappings between functions and their operation types."""
 
+    post_transforms: list[PostTransform] = field(
+        default_factory=lambda: [PostTransform(FrontendDesymrefyPass())]
+    )
+    """An ordered list of passes to apply to the built module."""
```
superlopuh: Let's make this a PassPipeline instead?

superlopuh: Ah no this can be a list of passes, and the builder should have a PassPipeline IMO

xdsl/frontend/pyast/context.py
superlopuh: not right now but this could be a good argument to add this sort of link

xdsl/frontend/pyast/context.py
```diff
@@ -52,20 +65,26 @@ def register_function(
         """Associate a method on an object in the source code with its IR implementation."""
         self.function_registry.insert(function, ir_constructor)
 
+    def register_post_transform(
+        self, transform: ModulePass, verify: bool = True
```
superlopuh: I would say we should either verify all or none of the passes

xdsl/frontend/pyast/utils/builder.py
EdmundGoodman: Verifying no passes whilst we have this constraint, then aim to verify all later when we can make better transforms

xdsl/frontend/pyast/context.py
EdmundGoodman: Parking this for now

superlopuh (approved):

superlopuh:
This is missing verification, can you add it as a field on context, and pass a verification callback to the pipeline init if it's True?

EdmundGoodman:
> This is missing verification, can you add it as a field on context, and pass a verification callback to the pipeline init if it's True?

Added support for setting custom callback functions for the pass pipeline on the pyast context

superlopuh (approved):

PR #4982, opened 2025-08-07
dialects: (x86) Update x86 instructions with Memory Effect traits

emmau678:
Add Memory Effect traits to the instructions in the x86 dialect


xdsl/backend/register_type.py
```diff
@@ -187,3 +190,25 @@ def infinite_register(cls, index: int) -> Self:
...
+    def get_effects(cls, op: Operation) -> set[EffectInstance]:
+        effects = set[EffectInstance]()
+        if any(
+            isinstance(r.type, RegisterType) and r.type.is_allocated
+            for r in chain(op.results)
+        ):
+            effects.add(EffectInstance(MemoryEffectKind.WRITE))
+        if any(
+            isinstance(r.type, RegisterType) and r.type.is_allocated
+            for r in chain(op.operands)
```
superlopuh:
```suggestion
        if any(
            isinstance(r.type, RegisterType) and r.type.is_allocated for r in op.results
        ):
            effects.add(EffectInstance(MemoryEffectKind.WRITE))
        if any(
            isinstance(r.type, RegisterType) and r.type.is_allocated for r in op.operands
```

superlopuh: Not sure why this was here but I don't think it's necessary

PR #4983, opened 2025-08-07
[WIP] core: allow results directive with SameVariadicResultSize option

superlopuh:
@alexarice do you have any pointers wrt how to fix this? MLIR supports this for [memref.extract_strided_metadata](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefextract_strided_metadata-memrefextractstridedmetadataop)


alexarice:
You'd need to rework the logic in the directive, it currently works under the assumption that there is at most one variadic

superlopuh:
Addressed in #5059

PR #4984, opened 2025-08-07
transformations: (convert-memref-to-ptr) split get_target_ptr into more helpers

superlopuh:
They'll be useful for more memref to ptr lowerings I'm planning to add.

### DIFF
--- a/xdsl/transforms/convert_memref_to_ptr.py
+++ b/xdsl/transforms/convert_memref_to_ptr.py
@@ -85,11 +85,12 @@ def get_strides_offset(
     Given SSA values for indices, and constant strides, insert the arithmetic ops that
     create the combined index offset.
     The length of indices and strides must be the same.
-    Strides must not contain `-1`.
+    Strides must be positive.
     """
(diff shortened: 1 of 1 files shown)


xdsl/transforms/convert_memref_to_ptr.py
```diff
@@ -19,61 +19,62 @@
...
+_index_type = builtin.IndexType()
 
-def get_target_ptr(
-    target_memref: SSAValue,
-    memref_type: memref.MemRefType[Any],
-    indices: Iterable[SSAValue],
-    builder: Builder,
-) -> SSAValue:
+
+def get_constant_strides(memref_type: builtin.MemRefType) -> tuple[Sequence[int], int]:
```
superlopuh, line 56: offset isn't in the name because I'd like to remove it in the subsequent PR, but I can add it here in the meantime if anyone feels strongly about it

superlopuh:
I think I should fix the offset situation first, I'll mark this as draft in the meantime

PR #4985, opened 2025-08-07
transformations: (convert-memref-to-ptr) ignore offset when indexing

superlopuh:
The idea is to apply the offset when lowering ops that affect the
memref, such as subview and reinterpret_cast.


qaco:
That's risky, no ? Shouldn't the nullication of the offset be the responsability of previous passes ?

superlopuh:
What do you mean by this? And by previous passes?

superlopuh:
As long as everyone uses the same convention it should be fine, I think.

PR #4988, opened 2025-08-08
core: (Operation) Move verification to verify

alexarice:
Seemed to be a random bit of verification in a `post_init` function.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1138,8 +1138,7 @@ class Operation(_IRNode):
         self._successor_uses = new_uses
 
     def __post_init__(self):
-        assert self.name != ""
-        assert isinstance(self.name, str)
+        pass
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


xdsl/ir/core.py
```diff
@@ -1138,8 +1138,7 @@ def successors(self, new: Sequence[Block]):
         self._successor_uses = new_uses
 
     def __post_init__(self):
-        assert self.name != ""
-        assert isinstance(self.name, str)
```
superlopuh, line 1142: Why the move? This only runs one, and is defiintely a programmer error if it triggers, I don't see the purpose of verifying this immutable constant after every pass

alexarice: Would run 0 times with `--disable-verify`

superlopuh: Yeah but the programmer error vs user error still stands, it only makes sense to throw verification errors for badly constructed ops, not badly defined ones.

alexarice: fair enough

superlopuh: Actually how about moving this to __init_subclass__?

alexarice: It gets angry with `IRDLOperation` not having a name, I'll just leave it I think

superlopuh: kk

PR #4989, opened 2025-08-08
core: (Operation) remove some implicit builder dynamicism

alexarice:
Possibly very controversial as @math-fehr doesn't like things being core, but the dynamic roundabout of setting a post_init in a different file feels much worse for me.

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import threading
 from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass, field
 from typing import TypeAlias, overload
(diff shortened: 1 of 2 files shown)


alexarice:
I got a bit sidetracked I think, these changes probably don't make sense

superlopuh:
I mean they kind of do but the advantage of the current system is that it's 0-cost if you don't use the ImplicitBuilder, this includes more instructions into a potentially very hot function.

alexarice:
Even with the very dynamic __post_init__?

superlopuh:
It's fully opt-in, so the deal is that either you don't use ImplicitBuilder, so you can expect things to be somewhat fast, or you use it and you forget about efficiency in the name of convenience. I'm not sure that there's a way to make this both fast and safe TBH

alexarice:
fair enough

PR #4991, opened 2025-08-08
core: (Operation) Use deletion on descriptors

alexarice:
Less sure about this one but descriptors seem to have `delete` functionality, which would avoid a check in attribute setting. Unsure whether this is worth it as we probably use `op.attr = None` in other places.


alexarice:
I think this likely isn't worth doing, I wasn't even aware of the `del` operator

PR #4993, opened 2025-08-08
tool: (xdsl-opt) Disable attribute verify on `--disable-verify`

alexarice:
I'm guessing that verification is run on Attribute post_init instead of being recursively called from operation verifications as Attributes should be immutable? 

It would be nice to have these not run when `--disable-verify` is on, and this is the best way I could think of doing it. This does feel a bit hacky though, and It won't effect benchmarking. I wonder if it's just worth taking the hit and recursively activating the attribute verification.

### DIFF
--- a/xdsl/xdsl_opt_main.py
+++ b/xdsl/xdsl_opt_main.py
@@ -19,6 +19,10 @@ from xdsl.utils.lexer import Span
 
 
 def _empty_post_init(self: Attribute):
+    """
+    An empty 'post_init' function.
+    Used to replace the default 'post_init' on 'Attribute' when verification is disabled.
+    """
     pass
 
 
--- a/xdsl/xdsl_opt_main.py
+++ b/xdsl/xdsl_opt_main.py
@@ -9,12 +9,17 @@ from typing import IO, Any
 
 from xdsl.context import Context
 from xdsl.dialects.builtin import ModuleOp
+from xdsl.ir import Attribute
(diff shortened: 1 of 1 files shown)


xdsl/xdsl_opt_main.py
```diff
@@ -17,6 +18,10 @@
 from xdsl.utils.lexer import Span
 
 
+def _empty_post_init(self: Attribute):
+    pass
```
superlopuh: Can you please add documentation here?

alexarice: Done

xdsl/xdsl_opt_main.py
```diff
@@ -54,6 +63,9 @@ def __init__(
 
         self.ctx.allow_unregistered = self.args.allow_unregistered_dialect
 
+        if self.args.disable_verify:
+            Attribute.__post_init__ = _empty_post_init
```
superlopuh, line 67: I'm wondering whether there's a way to make this more targeted. I wouldn't want to break a custom post-init of a subclass of Attribute

alexarice: This shouldn't change subclass post-inits right? In any case I feel that calling verification from operation verification instead of on attribute creation may be more principled, even if its less efficient

superlopuh: Oh yeah you're right it should still work.

superlopuh:
LGTM modulo @math-fehr's approval

PR #5007, opened 2025-08-10
core: add Universe and plugin discovery

superlopuh:
Adds a Universe which is a kind of meta-context of available dialects and passes. This class provides a `get_multiverse` getter, which traverses the environment for the right kind of plugins, and aggregates all passes and dialects into a single dict.

This adds about 25ms to the launch time of xdsl-opt on my computer. I personally think this is worth it for the ability of `xdsl-opt`, `xdsl-gui`, and other tools to automatically discover things.


EdmundGoodman (approved):
Looks exciting!! 

Doesn't block this PR but could we add an example of creating a plugin/using a plugin to xdsl-template once it's made public?

superlopuh:
Yeah defo, is the template set up with renovate? It would be great to add it so that it stays up to date with xDSL APIs.

superlopuh:
Just checked, I'll migrate things in the renovate PR if this gets merged.

docs/marimo/linalg_snitch.py
```diff
@@ -31,7 +31,7 @@ def _():
     from xdsl.interpreters.utils.ptr import TypedPtr
     from xdsl.ir import Attribute, Block, Region, SSAValue
     from xdsl.passes import PassPipeline
-    from xdsl.tools.command_line_tool import get_all_dialects
+    from xdsl.dialects import get_all_dialects
```
alexarice, line 34: Separate change?

xdsl/universe.py
```diff
@@ -0,0 +1,91 @@
+"""
+xDSL leverages Python [plugins](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/) for user-provided compiler infrastructure.
+
+Python packages that want to extend the passes and dialects available in `xdsl-opt` can
+implement their own `Universe`, which will be discovered by `Multiverse` at runtime.
+
+To opt into this behavior first add an AbstractUniverse subclass somewhere in your
```
alexarice:
```suggestion
To opt into this behavior first add an Universe subclass somewhere in your
```

docs/marimo/linalg_snitch.py
superlopuh: Kind of it's a tiny fix to importing from the wrong place, I'm not sure if ruff has a way to catch this kind of thing automatically.

math-fehr: Well, we can catch things like `mlir.ir.core`, so surely we can catch that one as well right?
But to be honest, if we shouldn't import it from there, then maybe we should move `get_all_dialects` to `xdsl.dialects`? (And I think it would make more sense).

tests/dialects/test_universe.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+import pytest
+
+from xdsl.ir import Dialect
+from xdsl.universe import Universe
+
+
+def test_multiverse():
```
math-fehr, line 9: Can you add a documentation comment on that test, so that we don't forget in the future what this does?

tests/my_plugin/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[project]
+name = "my_plugin"
+description = "Test plugin for pass and transform discovery"
```
math-fehr:
```suggestion
description = "Plugin to test pass and transform discovery"
```

docs/marimo/linalg_snitch.py
alexarice: This isn't the same as `mlir.ir.core`. The problem is that python reexports everything that you import in a file, so `get_all_dialects` is already in `xdsl.dialects` but because `xdsl.tools.command_line_tool` has `from xdsl.dialects import get_all_dialects`, it's valid python to do the above. I often find that it's quite easy to import the "wrong file" with pyright and couldn't find a ruff option to fix this.

superlopuh: I mean here it's a kind of implicit re-export that no longer works because the command line tool no longer imports this. It's not quite the same as the `xdsl.ir.core` since that is to block importing from the place of definition, instead here what I'd like to do is block everything that is not the place of definition. I think it's a nice to have though, we can just fix these in the notebooks as we go.

PR #5009, opened 2025-08-10
CI: lint benchmarks with pyright

superlopuh:
It turns out that we didn't check the benchmarks with pyright until now, so we didn't spot some API changes. It would be nice to add some tests in the future also, but for the time being I'll just remove the failing test which we no longer care about, fix some API changes, and add the benchmarks folder to be checked with pyright.


superlopuh:
I'll merge this now to make xdsl-bench happy but am happy to make post-merge changes in a follow-up PR

PR #5011, opened 2025-08-11
transformations: (apply-eqsat-pdl) Add option to apply patterns one by one

jumerckx:
As opposed to converting all the patterns to pdl_interp together. This helps with debugability of performance problems in the matching.

~TODO: clean up print statements, add tests~


xdsl/transforms/apply_eqsat_pdl.py
```diff
@@ -1,49 +1,209 @@
...
+
+    When True: Each pattern is converted to PDL_interp separately and applied individually
+    in each iteration.
+
+    When False (default): All patterns are converted together, potentially producing a more efficient
+    matcher by reusing equivalent expressions.
+    """
+
+    optimize_matcher: bool = False
+
```
superlopuh: Can you add a doc string here also?

xdsl/transforms/apply_eqsat_pdl.py
```diff
@@ -1,49 +1,209 @@
...
-            pdl_module = op
+            return op
+
+    def _extract_patterns(self, pdl_module: builtin.ModuleOp) -> list[pdl.PatternOp]:
+        """Extract all PDL patterns from the module."""
+        patterns: list[pdl.PatternOp] = []
+        for pattern_op in pdl_module.walk():
+            if isinstance(pattern_op, pdl.PatternOp):
+                patterns.append(pattern_op)
+        return patterns
```
superlopuh: Can we expect these to always be at the top level in their module?

jumerckx: I'm not sure. You mean that we wouldn't need to recurse with the walk?

superlopuh: Yeah

superlopuh: Not sure if that's a major concern but could speed things up quite a bit

xdsl/transforms/apply_eqsat_pdl.py
```diff
@@ -1,49 +1,204 @@
...
+        rewriters_module_ops: list[builtin.ModuleOp] = []
+        if all_rewriter_ops:
+            rewriters_module = builtin.ModuleOp(cast(list[Operation], all_rewriter_ops))
+            rewriters_module.attributes["sym_name"] = builtin.StringAttr("rewriters")
+            rewriters_module_ops.append(rewriters_module)
+
+        all_ops = cast(list[Operation], matchers) + cast(
+            list[Operation], rewriters_module_ops
+        )
+        return builtin.ModuleOp(all_ops)
```
superlopuh: Could you please do this via a builder pattern as opposed to a list pattern? So inserting the ops into the module as they're created, instead of appending lists and then inserting them?

superlopuh: I'm also not sure that this particular bit benefits from being in a helper

xdsl/transforms/apply_eqsat_pdl.py
```diff
@@ -1,49 +1,184 @@
...
+        rewriter_func = rewriter_module.body.first_block.first_op
+        assert isinstance(rewriter_func, pdl_interp.FuncOp)
+
+        return matcher, rewriter_func
+
+    def _apply_individual_patterns(
+        self, ctx: Context, op: builtin.ModuleOp, pdl_module: builtin.ModuleOp
+    ) -> None:
+        """Apply patterns individually in separate iterations."""
+        patterns = self._extract_patterns(pdl_module)
```
superlopuh:
```suggestion
        patterns = tuple(
            pattern
            for pattern_op in pdl_module.ops
            if isinstance(pattern, pdl.PatternOp)
        )
```

superlopuh (approved):

superlopuh:
Before we merge this, if you put the patterns in separate modules, and then run pdl-to-pdl-interp, do they get merged anyway or do they stay separate?

superlopuh:
It might be a big difference in perf if we only do the to-mlir-opt-and-back once

jumerckx:
Interesting idea, I just tried:
```
module {
  pdl.pattern {...}
}
module {
  pdl.pattern {...}
}
```
but `convert-pdl-to-pdl-interp` doesn't handle that well, it simply creates an empty matcher and empty rewrites module.

PR #5017, opened 2025-08-12
backend: (x86) add Arch enum and use for register type selection

superlopuh:
The aim of this PR is to unify the register type getters. These shouldn't be separate, as we need a way to generically handle lowering target-independent values passed in functions and as loop iter arguments. The best way I could come up with to model this is the `Arch` enum, although it feels like that's a bit of a misnomer. It would be nice to polish this later and get it to be more in line with how LLVM represents the same concept in its backend, hopefully in an extensible way. I think this will do for now, though.

### DIFF
--- a/xdsl/backend/x86/lowering/helpers.py
+++ b/xdsl/backend/x86/lowering/helpers.py
@@ -83,7 +83,7 @@ class Arch(StrEnum):
         return self._scalar_type_for_type(value_type)
 
 
-_ARCH_BY_NAME = {str(case).lower(): case for case in Arch}
+_ARCH_BY_NAME = {str(case): case for case in Arch}
 """
 Handled architectures in x86 backend.
 """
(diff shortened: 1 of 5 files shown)


qaco:
It is nice !

xdsl/backend/x86/lowering/helpers.py
```diff
@@ -9,49 +10,91 @@
...
+        match self, vector_size:
+            case ((Arch.AVX2 | Arch.AVX512), 256):
+                return x86.register.AVX2RegisterType
+            case Arch.AVX512, 512:
+                return x86.register.AVX512RegisterType
+            case _, 128:
+                return x86.register.SSERegisterType
+            case _:
+                raise DiagnosticException(
+                    f"The vector size {vector_size} and target architecture {self} are inconsistent."
```
qaco: Suggestion : The vector size {vector_size} and -> The vector size {vector_size} bits and

PR #5018, opened 2025-08-12
core: unify SSAValue and Block name logic in IRWithName

superlopuh:
A small refactor to share logic between IR nodes with names. Hopefully with nicer error messages.

Motivation for this PR is to extract the valid name extraction function for an upcoming PR.


superlopuh:
draft due to stacking otherwise ready to review

PR #5019, opened 2025-08-12
rewriting: add helper to add a listener to set inserted op name hints

superlopuh:
This will unlock more builder and implicit builder patterns, letting people set default ssa values for inserted operations.

### DIFF
--- a/xdsl/builder.py
+++ b/xdsl/builder.py
@@ -266,7 +266,7 @@ class Builder(BuilderListener):
             )
 
 
-class _SSAValueNameHandler:
+class SSAValueNameSetter:
     """
     Helper structure for setting name hints for inserted ops.
     """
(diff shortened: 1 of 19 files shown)


xdsl/builder.py
```diff
@@ -258,6 +266,40 @@ def assert_implicit():
...
+        return self._name
+
+    @name_hint.setter
+    def name_hint(self, name: str):
+        self._name = SSAValue.extract_valid_name(name)
+
+    def register(self, builder_listener: BuilderListener):
+        """
+        Adds a handler to the builder_listener instance to set the name hint of all
+        results of inserted operations, as long as the result name is not None.
```
jumerckx:
```suggestion
        results of inserted operations, as long as the result name is None.
```
I think?

jumerckx (approved):

xdsl/builder.py
superlopuh: I'll reword this completely so it's less ambiguous

superlopuh: I had to rework the logic to make it easier to reset the name hint for the builder to None

math-fehr (approved):
That makes a lot of sense!

PR #5021, opened 2025-08-12
installation: try renovate nix support

alexarice:
I guess we try the default config first and see what happens. From reading the docs this should try to update the lock file once a week. I'm not 100% sure if this is a great idea without any CI but I guess I can manually test the commits for now.

Is there a way to get renovate to test the config in this PR?


alexarice:
Ok, we'll give it a go then and see what happens

PR #5022, opened 2025-08-12
core: (Operation) simplify `irdl_op_verify_arg_list`

alexarice:
Step one of simplifying some of the variadics infrastructure. This removes the last use of the return value for `get_variadic_sizes`, which should allow that method to be simplified. We instead rely on the operation accessors to get the various variadic ranges.

I've kept the logic to put the slice in the error message instead of referring to it by construct name, as I fear the resulting diff from changing all the tests would be very large.

### DIFF
--- a/xdsl/irdl/operations.py
+++ b/xdsl/irdl/operations.py
@@ -1481,7 +1481,7 @@ def irdl_op_verify_arg_list(
         elif not isinstance(args, Sequence):
             arg_types = (args.type,)
         else:
-            arg_types = tuple(a.type for a in cast(Sequence[SSAValue], args))
+            arg_types = args.types
         length = len(arg_types)
         try:
             arg_def.constr.verify(arg_types, constraint_context)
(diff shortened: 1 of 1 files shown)


xdsl/irdl/operations.py
```diff
@@ -1469,43 +1469,32 @@ def irdl_op_verify_arg_list(
...
+    idx = 0
+
+    for arg_name, arg_def in defs:
+        args = getattr(op, arg_name)
+        if args is None:
+            arg_types = ()
+        elif not isinstance(args, Sequence):
+            arg_types = (args.type,)
+        else:
+            arg_types = tuple(a.type for a in cast(Sequence[SSAValue], args))
```
superlopuh: ?
```suggestion
            arg_types = args.types
```

alexarice: Right, this was mainly unchanged but this should work now.

alexarice: It would seem not

PR #5023, opened 2025-08-12
dependencies: remove `jax`, `riscv`, `bench` groups from dev dependencies group

jumerckx:
These are often not needed and prevent running with pypy.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.12.8
     hooks:
-    -   id: ruff
+    -   id: ruff-check
         types_or: [ python, pyi, jupyter ]
         args: [ --fix, --exit-non-zero-on-fix ]
     -   id: ruff-format
--- a/tests/filecheck/backend/x86/convert_ptr_to_x86.mlir
+++ b/tests/filecheck/backend/x86/convert_ptr_to_x86.mlir
@@ -51,7 +51,7 @@
 
 // -----
 
-// CHECK: The vector size and target architecture are inconsistent.
+// CHECK: The vector size (1024 bits) and target architecture `avx512` are inconsistent.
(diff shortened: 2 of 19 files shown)


superlopuh (approved):
LGTM but I'd like to check with @alexarice and @math-fehr first that they agree that this is a good idea.

superlopuh (commented):
needs a uv sync

math-fehr:
SO does that mean that if we just install the `dev` version, we won't be able to run all the tests?

superlopuh:
We already have checks for whether packages are installed, so it's more that you only have to install the core dev dependencies for the base installation, and install `--all-extras` if you break CI for some reason.

superlopuh:
I'll clone this and try to fix the CI quickly

Makefile
```diff
@@ -11,7 +11,9 @@ VENV_DIR ?= .venv
 export UV_PROJECT_ENVIRONMENT=$(if $(VIRTUAL_ENV),$(VIRTUAL_ENV),$(VENV_DIR))
 
 # allow overriding which extras are installed
-VENV_EXTRAS ?= --extra gui --extra dev --extra jax --extra riscv --extra docs --extra bench
+VENV_EXTRAS ?= --all-extras
+VENV_GROUPS ?= --all-groups
```
superlopuh, line 15: we now have extras and groups but make venv is parametrizable

superlopuh:
With my latest changes: 

1. `make venv` installs everything, same as `uv sync --all-extras --all-groups`
2. `xdsl[docs]` and `xdsl[bench]` no longer works, but I think that's actually an improvement
3. `uv sync && make tests-functional` works as checked by `ci-core` (pyright will break due to mossing dependencies)
4. The recommended workflow is still to install all dependencies, but at least people playing with pypy will be able to install only the core dependencies necessary to make things run, and not JAX and other optional things

math-fehr:
I think that's fine for me, but please update the README.md to reflect that, so it's clear at least!

superlopuh:
@math-fehr thoughts on new README? I'm not sure that it makes sense to keep the `pip` instructions.

PR #5025, opened 2025-08-12
dialects: (builtin) MemReftype layout constraint is MemRefLayoutAttr | NoneAttr

jorendumoulin:
The layout attribute of a `MemRefType` should allow for any attribute that implements the `MemRefLayoutAttr` base class

Made possible through #5240


superlopuh (approved):
🎊

PR #5026, opened 2025-08-12
dialects: (emitc) Add AddOp

Jimmy2027:
Add the emitc [Add Op](https://mlir.llvm.org/docs/Dialects/EmitC/#emitcadd-emitcaddop)


tests/filecheck/dialects/emitc/emitc_ops.mlir
```diff
@@ -18,3 +22,19 @@ emitc.call_opaque "test" ()  : () -> ()
 // CHECK-GENERIC:  %0 = "emitc.call_opaque"() <{callee = "blah"}> : () -> i64
 // CHECK-GENERIC-NEXT: "emitc.call_opaque"(%0) <{callee = "foo", args = [0 : index, dense<[0, 1]> : tensor<2xi32>, 0 : index]}> : (i64) -> ()
 // CHECK-GENERIC-NEXT: "emitc.call_opaque"() <{callee = "test"}> : () -> ()
+
+//===----------------------------------------------------------------------===//
+// AddOp
+//===----------------------------------------------------------------------===//
+
+%add_int = "emitc.add" (%i32_lhs, %i32_rhs) : (i32, i32) -> i32
```
superlopuh: why not custom format here?

superlopuh (commented):
can you please add MLIR interop tests?

tests/filecheck/dialects/emitc/emitc_ops.mlir
Jimmy2027: hmm good point, guess I just copied the [MLIR emitC tests](https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/EmitC/ops.mlir#L63)

Jimmy2027: switched to custom format in 1557658a4d33580f763293fdc5b36c701523791c

Jimmy2027:
added mlir interop tests in b3df0425a120efefe3b81184449c0291446cd0a5 and a python test to get better coverage in dde7bff65a90b4d4e5b9b3995c5280eef5c4b9ef

tests/filecheck/mlir-conversion/with-mlir/dialects/emitc/emitc_ops.mlir
```diff
@@ -1,5 +1,9 @@
-// RUN: xdsl-opt %s | mlir-opt | xdsl-opt | filecheck %s
-// RUN: xdsl-opt %s --print-op-generic | mlir-opt --mlir-print-op-generic | xdsl-opt | filecheck %s
+// RUN: xdsl-opt %s | mlir-opt --allow-unregistered-dialect | xdsl-opt | filecheck %s
+// RUN: xdsl-opt %s --print-op-generic | mlir-opt --allow-unregistered-dialect --mlir-print-op-generic | xdsl-opt | filecheck %s
```
superlopuh: I just added new macros for these to make this easier:
```suggestion
// RUN: MLIR_ROUNDTRIP
// RUN: MLIR_GENERIC_ROUNDTRIP
```

xdsl/dialects/emitc.py
```diff
@@ -302,6 +304,73 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
...
+
+    assembly_format = "operands attr-dict `:` functional-type(operands, results)"
+
+    def __init__(
+        self,
+        lhs: SSAValue,
+        rhs: SSAValue,
+        result_type: Attribute | None = None,
+    ):
+        if result_type is None:
```
Jimmy2027: on second thought I'm not sure if this is a good idea... I had implemented this because I thought it would be a nice feature but now I'm thinking it could hide mistakes from the user

tests/filecheck/mlir-conversion/with-mlir/dialects/emitc/emitc_ops.mlir
Jimmy2027: nice!

Jimmy2027:
looks like I marked this as ready too soon, some type check functionality is missing, sry @superlopuh 
need to have another look

xdsl/dialects/emitc.py
superlopuh: Yeah I think this is minor because the most important thing is the user has the flexibility to do the thing they want, but it definitely feels like we can't infer the correct type in general.

Jimmy2027: yup, removed it :)

Jimmy2027:
I had to add a few things to make it match the mlir functionality:
- add tensor_type.has_static_shape in af021b9fee2356958ff4ac874ff15dc374021994
- add a emitc_type and EmitC_BinaryOp in 2e2a3b08efb225ec42a8db66f1db74e5df899f8e

superlopuh:
Ready to review?

Jimmy2027:
yeah let's have another go 😀

xdsl/dialects/builtin.py
```diff
@@ -1417,6 +1417,15 @@ def get_shape(self) -> tuple[int, ...]:
...
 
+    def has_static_shape(self) -> bool:
+        """
+        Check if all dimensions have static (non-dynamic) shapes.
+
+        Returns:
+            True if all dimensions are static, False if any dimension is dynamic (-1).
+        """
+        return all(dim.data != -1 for dim in self.shape.data)
+
```
superlopuh: Can you please open a separate PR adding this to ShapedType directly? Can you also use `DYNAMIC_INDEX` instead of `-1` in the doc, as there's a general hope of that being an implementation detail even if we're not very careful with it in practice.

xdsl/dialects/emitc.py
```diff
@@ -296,12 +305,100 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
...
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> AttrConstraint:
+        # No type variables to map in this constraint
+        return self
+
+
+emitc_type = EmitCType()
+
+
+class EmitC_BinaryOp(IRDLOperation, abc.ABC):
```
superlopuh: We use Operation as a convention for abstract superclasses
```suggestion
class EmitC_BinaryOperation(IRDLOperation, abc.ABC):
```

xdsl/dialects/emitc.py
```diff
@@ -296,12 +305,100 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
...
+        self, attr: Attribute, constraint_context: ConstraintContext | None = None
+    ) -> None:
+        if not is_supported_emitc_type(attr):
+            raise VerifyException(f"Type {attr} is not a supported EmitC type")
+
+    def mapping_type_vars(
+        self, type_var_mapping: dict[TypeVar, AttrConstraint]
+    ) -> AttrConstraint:
+        # No type variables to map in this constraint
+        return self
```
superlopuh: This is backwards, we should refactor `is_supported_emitc_type` to a constraint IMO, instead of wrapping it in a constraint. I think this is also worth doing in a separate PR before opening this one.

xdsl/dialects/builtin.py
Jimmy2027: sure, will do! Ah nice yeah having a variable for that makes sense

xdsl/dialects/emitc.py
Jimmy2027: just to be sure, you're thinking of something like this right?
```python
EmitCTypeConstraint = AnyOf(
    [
        supported_emitc_integer_types,
        supported_emitc_float_types,
        IndexType(),
        EmitC_OpaqueType,
        supported_emitc_pointer_wide_types,
        EmitC_LValueType,
        EmitC_ArrayType,
        EmitC_PointerType,
        TensorType,
        TupleType,
    ]
)
```

I'm not sure how to define this since EmitCTypeConstraint needs to include EmitC_PointerType, but EmitC_PointerType uses EmitC_LValueType which in turn uses EmitCTypeConstraint, creating a circular dependency... Is there any way to define a constraint lazily? Something like this:
`    value_type: Annotated[TypeAttribute, lambda: get_emitc_type_constraint()]`

superlopuh: Eventually maybe but for the time being I just meant inlining the function definition of `is_supported_emitc_type` into here and calling `EmitCTypeConstr().verifies(attr)` when calling it manually. We can refactor to make the implementation prettier later.

PR #5027, opened 2025-08-13
core: Make get_canonicalization_patterns a method

math-fehr:
Previously, this method was a class method, which is not necessary at all.

### DIFF
--- a/docs/Toy/toy/dialects/toy.py
+++ b/docs/Toy/toy/dialects/toy.py
@@ -366,8 +366,7 @@ class ReturnOp(IRDLOperation):
 
 
 class ReshapeOpHasCanonicalizationPatternsTrait(HasCanonicalizationPatternsTrait):
-    @classmethod
-    def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
+    def get_canonicalization_patterns(self) -> tuple[RewritePattern, ...]:
         from ..rewrites.optimise_toy import (
             FoldConstantReshapeOpPattern,
             ReshapeReshapeOpPattern,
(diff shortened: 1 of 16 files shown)


superlopuh:
If we're doing a refactoring here, I think it would make sense to agree on a design to move to. I think it would make sense to make it a method that takes an operation, so `canonicalize` looks like this:

```py
        traits = op.get_traits_of_type(HasCanonicalizationPatternsTrait)
        if trait is None:
            return
        patterns = tuple(
            trait.get_canonicalization_patterns(op)
            for trait in traits
        )
```

We could actually make the transition smoother by first supporting both the current `HasCanonicalizationPatternsTrait` and a new trait with a new name, and deprecate the old one for a month or whatever to give people time to migrate.

math-fehr:
> If we're doing a refactoring here, I think it would make sense to agree on a design to move to. I think it would make sense to make it a method that takes an operation, so `canonicalize` looks like this:
> 
> ```python
>         traits = op.get_traits_of_type(HasCanonicalizationPatternsTrait)
>         if trait is None:
>             return
>         patterns = tuple(
>             trait.get_canonicalization_patterns(op)
>             for trait in traits
>         )
> ```
> 
> We could actually make the transition smoother by first supporting both the current `HasCanonicalizationPatternsTrait` and a new trait with a new name, and deprecate the old one for a month or whatever to give people time to migrate.

I still don't really get thi, because if you get the trait from the op, then you already have all the information you want?

alexarice:
I don't really see the point in this, does it enable something you want to do. Don't feel we should really be allowing patterns to be added based on the specific object instance?

PR #5030, opened 2025-08-14
core: add HasCanonicalizationPatternsInterface

superlopuh:
My alternative proposal to #5027, introducing a new `CanonicalizationPatternsTrait` and `HasCanonicalizationPatternsInterface`.

If the `CanonicalizationPatternsTrait` is initialised with a tuple, these are always used, and there are no constraints on the operation directly.
If it is instead initialised with `None`, then it will query the interface's implementation for the patterns instead.

### DIFF
--- a/xdsl/interfaces.py
+++ b/xdsl/interfaces.py
@@ -9,6 +9,7 @@ This can be more convenient than adding the traits explicitly.
 
 import abc
 from dataclasses import dataclass
+from typing import cast
 
 from xdsl.ir import Operation
 from xdsl.irdl import traits_def
@@ -17,7 +18,7 @@ from xdsl.traits import HasCanonicalizationPatternsTrait
 
 
 @dataclass(frozen=True)
-class HasCanonicalizationPatternsInterfaceTrait(HasCanonicalizationPatternsTrait):
+class _HasCanonicalizationPatternsInterfaceTrait(HasCanonicalizationPatternsTrait):
     """
     Gets the canonicalization patterns from the operation's implementation
     of `HasCanonicalizationPatternsInterface`.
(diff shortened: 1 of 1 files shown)


docs/Toy/toy/dialects/toy.py
superlopuh, line 1: This file is now an example of a dialect using the tuple init directly for lower overhead and ceremony.

xdsl/dialects/riscv_cf.py
superlopuh, line 1: This file is an example of using the Interface instead, lowering the number of things to subclass for canonicalization to work.

xdsl/traits.py
```diff
@@ -454,16 +454,53 @@ def get_result_types(cls, op: Operation) -> tuple[Attribute, ...]:
...
+    Note: this class will be deprecated soon, please use
+    `HasCanonicalizationPatternsTrait` instead.
+    """
+
+    def get_patterns(
+        self,
+        op: type[Operation],
+    ) -> tuple[RewritePattern, ...]:
+        return type(self).get_canonicalization_patterns()
+
```
superlopuh: I made this PR backwards compatible so that we can deprecate this class in the next one if we go down this route.

alexarice:
Also thoughts on calling this `HasCanonicalizer` like the tablegen flag?

superlopuh:
I'll open new PRs for bits of this one to reduce the diff. Would you like a separate PR introducing the mechanism and use in Toy? I feel that it's nice to have a practical example of it but I guess we could also do that in a test.

alexarice:
This wasn't really a request for splitting the PR. More that there's lots of changes that look completely unnecessary here (Moving rewrite patterns around)

superlopuh:
They are necessary though due to circular dependencies otherwise being introduced. It's easier to open some more PRs to make the moves first in preparation though.

alexarice:
I'd prefer not to move to this interface if it requires circular dependencies

superlopuh:
Let me just make the changes so you can assess them

PR #5031, opened 2025-08-14
rewriting: drop _op from insert_op, erase_op, and replace_op

superlopuh:
As that's almost always the thing that's called, the _op can safely be implicit. As far as I could find, these are all the ops on Builder/PatternRewriter with _op suffixes. In the future, I'd like to also migrate Rewriter, but that can safely be done later IMO.


math-fehr:
If we would go that route, I would actually be in favor of having `insert`, and thus `replace`, `erase`, etc...
My main reasonning is that these operations are used quite a lot (I have dozens of `b.insert(...)` in a row), and would like to shorten them as much as possible while keeping them readable.
Because operations are also ended with `Op`, writing `b.insert(MyOp(...))` also make it alerady clear what you are inserting. Ops are also the thing you most commonly insert.

superlopuh:
Can you push the API you want to this branch (I mean rename the functions to what you think is best) and then I'll fix all the errors?

alexarice:
@math-fehr would you rename `replace_matched_op` to `replace_matched` and similar for `erase`?
I still feel `insert_op` is slightly more simple, even though all ops end in `Op`.

math-fehr:
I would personally like to get rid of `replace_matched_op` and the like. Since the `op` is passed to the rewrite pattern anyway, and the insertion point should be set at that operation, these functions are actually not that useful IMO. But that would need a bit more refactor I believe?
For `erase_op`, I would actually be fine having `erase` instead.

superlopuh:
It would be quite helpful if you could change the names of the ops in this branch to what you would like so we can discuss a concrete API, I'll fix up the errors.

sueszli (changes requested):
Great idea.

But given how this is a really core piece of the library, wouldn't it make sense to leave a stub that warns about this function being deprecated and suggesting the new alternative instead?

alexarice:
What happened to this? Did we ever make a decision?

superlopuh:
Yeah kind of, I need to find the time to make the change or inspire someone to do it. I have come around to the idea that it's fine to drop the `op` suffix for op-related things since it's the most common thing to operate on by far. I also think it's more important than when I first started playing around with this to have a proper hierarchy of Builder/Rewriter/PatternRewriter like in MLIR to make more robust helpers, so it's growing in urgency, but still less so than my current backend work as far as my schedule is concerned. Would love some help here if people feel like volunteering.

alexarice:
So would the preference here to be changing all `insert_op`/`replace_op`/etc to `insert`/`replace`?

superlopuh:
That's the idea, I think it's reasonable.

alexarice:
Seems reasonable to me, I guess the steps are:
- Add `replace`/`erase` (is there another?)
- Big find and replace
- Deprecate `insert_op`/`erase_op`/`replace_op`

Was there a consensus on the various `matched` ops? Are they staying?

superlopuh:
Oh yeah, I think they should go, and that's probably a pre-requisite.

superlopuh:
I mean `PatternRewriter.current_operation` maybe could stay, but `replace_matched_op` should go IMO.

benchmarks/rewriting.py
```diff
@@ -197,7 +197,7 @@ def ignore_time_pattern_rewriter_insert_op(self) -> None:
         Exercise inserting an operation and running any required callbacks. This
         is used to effect the results of rewriting.
         """
-        self.pattern_rewriter.insert_op((self.sub_op,), self.insert_point)
+        self.pattern_rewriter.insert((self.sub_op,), self.insert_point)
```
superlopuh, line 200: this will cause the benchmark to break for the next four commits

xdsl/builder.py
```diff
@@ -110,6 +110,7 @@ def insert(
 
         return op
 
+    @deprecated("Use .insert(op, insertion_point) instead")
     def insert_op(
```
superlopuh, line 114: deprecation 1

xdsl/pattern_rewriter.py
```diff
@@ -184,7 +189,7 @@ def replace_uses_with_if(
         for op in tracking.modified_ops:
             self.handle_operation_modification(op)
 
-    @deprecated("Please use `replace_op(op, new_op)`")
+    @deprecated("Use .replace(op, new_op) instead")
```
superlopuh, line 192: deprecation update

xdsl/pattern_rewriter.py
```diff
@@ -251,7 +256,18 @@ def replace_op(
                         res.name_hint = name_hint
 
         # Then, erase the original operation
-        self.erase_op(op, safe_erase=safe_erase)
+        self.erase(op, safe_erase=safe_erase)
+
+    @deprecated("Use .replace(op, new_ops, new_results, safe_erase) instead")
```
superlopuh, line 261: deprecation 2

xdsl/pattern_rewriter.py
```diff
@@ -147,6 +147,11 @@ def erase_op(self, op: Operation, safe_erase: bool = True):
         self.handle_operation_removal(op)
         Rewriter.erase_op(op, safe_erase=safe_erase)
 
+    @deprecated("Use .erase(op, safe_erase) instead")
```
superlopuh, line 150: deprecation 3

tests/pattern_rewriter/test_pattern_rewriter.py
```diff
@@ -2134,3 +2134,66 @@ def match_and_rewrite(self, op: test.TestOp, rewriter: PatternRewriter):
         op_removed=1,
         op_replaced=1,
     )
+
+
+def test_erase_op_deprecated():
```
superlopuh, line 2139: tests for deprecated functions, to make sure they still work

sueszli (approved):

docs/notebooks/ir_gen.py
```diff
@@ -296,7 +296,7 @@ def emit_ir(expr: Expr) -> ModuleOp:
         result = emit_op(expr, builder, arg_values)
 
         # Insert a return statement at the end of the function.
-        builder.insert(ReturnOp(result))
+        builder.insert_op(ReturnOp(result))
```
alexarice, line 299: Isn't this the wrong way round?

docs/Toy/Toy_Ch3.ipynb
```diff
@@ -291,7 +291,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": ".venv",
```
alexarice, line 294: Unrelated change?

xdsl/pattern_rewriter.py
alexarice: Can likely get rid of this one now?

alexarice (commented):
I see I'm too late but here's some comments anyway. Aside: has github's ui got super slow for anyone else? firefox is really struggling to display this PR

docs/Toy/Toy_Ch3.ipynb
superlopuh: yep will revert

superlopuh:
Yep GH UI is struggling for me also. I'll open a cleanup PR for the first two comments. It's only been two months since we deprecated the matched_op helpers, and I generally like to wait three months before dropping those, but I'll also do a cleanup pass over the whole project for the older deprecations.

PR #5032, opened 2025-08-14
dependencies: remove [riscv] optional extra dependency

superlopuh:
Move riscemu to the docs, which is where it's used anyway.


superlopuh:
I'll need to merge #5023 first for this to work

alexarice (approved):

PR #5033, opened 2025-08-14
dialects: (builtin) Add has_static_shape method to TensorType

Jimmy2027:
Usage example can be found in #5026

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -109,6 +109,12 @@ class ShapedType(Attribute, ABC):
     def element_count(self) -> int:
         return prod(self.get_shape())
 
+    def has_static_shape(self) -> bool:
+        """
+        Check if all dimensions have static (non-dynamic) shapes.
+        """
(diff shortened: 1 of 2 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1417,6 +1417,12 @@ def get_shape(self) -> tuple[int, ...]:
     def get_element_type(self) -> AttributeCovT:
         return self.element_type
 
+    def has_static_shape(self) -> bool:
+        """
+        Check if all dimensions have static (non-dynamic) shapes.
+        """
+        return all(dim.data != DYNAMIC_INDEX for dim in self.shape.data)
+
```
superlopuh: Why not shaped type?

Jimmy2027: good point :) fixed in [623f8fd](https://github.com/xdslproject/xdsl/pull/5033/commits/623f8fd09c0c5c51a16a90d6f33e15704b110297)

PR #5036, opened 2025-08-15
CI: lint markdown files

superlopuh:
I occasionally find myself fixing these up and I feel like it would be worth keeping things consistent in the CI.
I tried to add this as a pre-commit hook, but it timed out when installing on my computer, so I would propose just checking it in the CI instead.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,6 @@
+{
+    "MD013": {
+        "line_length": 88,
+        "code_blocks": false
+    }
+}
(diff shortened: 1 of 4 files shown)


.markdownlint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "MD013": {
+        "line_length": 88,
```
superlopuh, line 3: same line length as our python code

README.md
```diff
@@ -165,11 +172,12 @@ Furthermore, all python code must run through [pyright](https://github.com/micro
 without errors. Pyright can be run on all staged files through the
 makefile using `make pyright`.
 
+<!-- markdownlint-disable MD013 -->
```
math-fehr: Why do we need that here?

math-fehr (approved):

README.md
superlopuh: I got it to work without this

PR #5037, opened 2025-08-15
documentation: add make docs-serve-fast for easier debugging

superlopuh:
Generating our full API reference is too slow for hot reloading.


scripts/gen_ref_pages.py
superlopuh, line 1: I recommend hiding white space changes for this file

alexarice (approved):

PR #5039, opened 2025-08-16
dialects: (emitc) Add EmitCTypeConstraint

Jimmy2027:
Convert the is_supported_emitc_type function to an AttrConstraint.

As proposed in #5026

### DIFF
--- a/xdsl/dialects/emitc.py
+++ b/xdsl/dialects/emitc.py
@@ -7,8 +7,8 @@ Those can be translated to C/C++ via the Cpp emitter.
 See external [documentation](https://mlir.llvm.org/docs/Dialects/EmitC/).
 """
 
-from collections.abc import Iterable, Sequence
-from typing import Generic, Literal, cast
+from collections.abc import Iterable, Mapping, Sequence
+from typing import Generic, Literal
 
 from typing_extensions import TypeVar
 
@@ -37,12 +37,16 @@ from xdsl.ir import (
     TypeAttribute,
 )
 from xdsl.irdl import (
+    AttrConstraint,
+    ConstraintContext,
+    IntConstraint,
     IRDLOperation,
     ParsePropInAttrDict,
     irdl_attr_definition,
     irdl_op_definition,
     irdl_to_attr_constraint,
     opt_prop_def,
+    param_def,
     prop_def,
     var_operand_def,
     var_result_def,
@@ -209,6 +213,55 @@ class EmitC_ArrayType(
             printer.print_attribute(self.element_type)
 
 
+class EmitCTypeConstraint(AttrConstraint):
+    """
+    Check if a type is supported by EmitC.
+    See [MLIR implementation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L62).
+    """
+
+    def verify(
+        self, attr: Attribute, constraint_context: ConstraintContext | None = None
+    ) -> None:
+        if isa(attr, TensorType):
+            elem_type = attr.get_element_type()
+            if isinstance(elem_type, EmitC_ArrayType):
+                raise VerifyException("EmitC type cannot be a tensor of EmitC arrays")
+            self.verify(elem_type, constraint_context)
+            return
+
+        if isa(attr, EmitC_ArrayType):
+            elem_type = attr.get_element_type()
+            if isinstance(elem_type, EmitC_ArrayType):
+                raise VerifyException("nested EmitC arrays are not allowed")
+            self.verify(elem_type, constraint_context)
+            return
+
+        if isinstance(attr, EmitC_PointerType):
+            self.verify(attr.pointee_type, constraint_context)
+            return
+
+        if isinstance(attr, TupleType):
+            for t in attr.types:
+                if isinstance(t, EmitC_ArrayType):
+                    raise VerifyException(
+                        "EmitC type cannot be a tuple of EmitC arrays"
+                    )
+                self.verify(t, constraint_context)
+            return
+
+        assert constraint_context is not None
+        EmitCArrayElementTypeConstr.verify(attr, constraint_context)
+
+    def mapping_type_vars(
+        self, type_var_mapping: Mapping[TypeVar, AttrConstraint | IntConstraint]
+    ) -> AttrConstraint:
+        # No type variables to map in this constraint
+        return self
+
+
+EmitCTypeConstr = EmitCTypeConstraint()
+
+
 @irdl_attr_definition
 class EmitC_LValueType(ParametrizedAttribute, TypeAttribute):
     """
@@ -218,18 +271,13 @@ class EmitC_LValueType(ParametrizedAttribute, TypeAttribute):
     """
 
     name = "emitc.lvalue"
-    value_type: TypeAttribute
+    value_type: Attribute = param_def(EmitCTypeConstr)
 
     def verify(self) -> None:
         """
         Verify the LValueType.
         See [MLIR implementation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L1095)
         """
-        # Check that the wrapped type is valid. This especially forbids nested lvalue types.
-        if not is_supported_emitc_type(self.value_type):
-            raise VerifyException(
-                f"!emitc.lvalue must wrap supported emitc type, but got {self.value_type}"
-            )
         if isinstance(self.value_type, EmitC_ArrayType):
             raise VerifyException("!emitc.lvalue cannot wrap !emitc.array type")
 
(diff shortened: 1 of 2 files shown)


xdsl/dialects/emitc.py
```diff
@@ -137,18 +253,14 @@ class EmitC_LValueType(ParametrizedAttribute, TypeAttribute):
     """
 
     name = "emitc.lvalue"
-    value_type: TypeAttribute
+    value_type: Annotated[TypeAttribute, emitc_type_constr]
```
superlopuh: We're trying to get rid of Annotated, so let's not introduce it here.
```suggestion
    value_type: EmitCTypeConstr
```
Also this is the naming convention we've gone with

xdsl/dialects/emitc.py
```diff
@@ -50,6 +54,118 @@
...
+
+def _is_supported_integer_type(type_attr: Attribute) -> bool:
+    """
+    Check if an IntegerType is supported by EmitC.
+    See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L96).
+    """
+    return (
+        isinstance(type_attr, IntegerType)
+        and type_attr.width.data in _SUPPORTED_BITWIDTHS
+    )
```
superlopuh:
```suggestion
import functools
import operator

EmitCIntegerTypeBitwidthConstr = reduce(
    operator.or, 
    (
        EqAttrConstraint(bw) for bw in (1, 8, 16, 32, 64)
    )
)

EmitCIntegerTypeConstr = ParamAttrConstr(
	IntegerAttr,
	(
		EmitCIntegerTypeBitwidthConstr,
		AnyAttr()
	)
)
```

xdsl/dialects/emitc.py
```diff
@@ -50,6 +54,118 @@
...
+def is_supported_float_type(type_attr: Attribute) -> bool:
+    """
+    Check if a type is a supported floating-point type in EmitC.
+    See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L117)
+    """
+    match type_attr:
+        case Float16Type() | BFloat16Type() | Float32Type() | Float64Type():
+            return True
+        case _:
+            return False
```
superlopuh:
```suggestion
EmitCFloatTypeConstr = (
    BaseAttr(Float16Type)
    | BaseAttr(BFloat16Type)
    | BaseAttr(Float32Type)
    | BaseAttr(Float64Type)
)
"""
Supported floating-point type in EmitC.
See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L117)
"""
```

superlopuh: Can you please make similar changes to the other helpers?

xdsl/dialects/emitc.py
Jimmy2027: right... thanks!

xdsl/dialects/emitc.py
Jimmy2027: would `AnyOf([i1, i8, i16, i32, i64])` not work here?

xdsl/dialects/emitc.py
Jimmy2027: will do, thanks!

Jimmy2027: actually, what is the best way to call the constraints then in EmitCTypeConstraint? Couldn't figure out how to handle the circular dependency mentioned in the PR description

xdsl/dialects/emitc.py
superlopuh: No we added a constraint that all elements of AnyOf must have separate types, this lets us avoid backtracking during verification

xdsl/dialects/emitc.py
superlopuh: I'm not sure I get it but adding `from __future__ import annotations` at the top of the file is worth trying.

xdsl/dialects/emitc.py
Jimmy2027: ah okay I see

xdsl/dialects/emitc.py
Jimmy2027: don't think `from __future__ import annotations` helps for variables here

Jimmy2027: the issue becomes clear in b038107ff429e9494b1eb3ec795a70053fd177ba where EmitCTypeConstr needs to be defined in https://github.com/xdslproject/xdsl/pull/5039/commits/b038107ff429e9494b1eb3ec795a70053fd177ba#diff-a6fc816eccf354215a90db862d5edae1ba9491317599aac0e26621e61ce39660R167

xdsl/dialects/emitc.py
```diff
@@ -50,6 +62,97 @@
 from xdsl.utils.exceptions import VerifyException
 from xdsl.utils.hints import isa
 
+EmitCIntegerTypeBitwidthConstr = functools.reduce(
+    operator.or_, (EqAttrConstraint(bw) for bw in (i1, i8, i16, i32, i64))
```
superlopuh:
```suggestion
    operator.or_, (EqAttrConstraint(bw) for bw in (1, 8, 16, 32, 64))
```

xdsl/dialects/emitc.py
```diff
@@ -60,7 +163,8 @@ class EmitC_ArrayType(
     name = "emitc.array"
 
     shape: ArrayAttr[IntAttr]
-    element_type: AttributeCovT
+
+    element_type: EmitCArrayElementTypeConstr
```
superlopuh:
```suggestion
    element_type: AttributeCovT = param_def(EmitCArrayElementTypeConstr)
```

xdsl/dialects/emitc.py
```diff
@@ -137,169 +226,39 @@ class EmitC_LValueType(ParametrizedAttribute, TypeAttribute):
     """
 
     name = "emitc.lvalue"
-    value_type: TypeAttribute
+    value_type: EmitCTypeConstr
```
superlopuh:
```suggestion
    value_type: Attribute = param_def(EmitCTypeConstr)
```

xdsl/dialects/emitc.py
superlopuh: I feel like the diff is already big enough to be difficult to reason about. Can you please open a smaller PR with only the things that seem to work (like integers and floats), so that we can iterate on the thing that doesn't work in isolation?

xdsl/dialects/emitc.py
superlopuh: Ah my bad I just realised this isn't expressible yet I'll add a new constraint quickly.

superlopuh: #5040

xdsl/dialects/emitc.py
```diff
@@ -50,6 +62,97 @@
...
+See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L96).
+"""
+
+
+EmitCFloatTypeConstr = (
+    BaseAttr(Float16Type)
+    | BaseAttr(BFloat16Type)
+    | BaseAttr(Float32Type)
+    | BaseAttr(Float64Type)
+)
```
superlopuh:
```suggestion
EmitCFloatType = Float16Type | BFloat16Type | Float32Type | Float64Type
EmitCFloatTypeConstr = irdl_to_attr_constraint(EmitCFloatType)
```

I realise that we already merged this in the other PR but it might make sense to define both the type and the corresponding constraint, and to use one or the other as convenient. For example, for the array element type it might be more convenient to just use the type directly in the attribute definition.

xdsl/dialects/emitc.py
```diff
@@ -209,6 +213,55 @@ def print_parameters(self, printer: Printer) -> None:
...
+            elem_type = attr.get_element_type()
+            if isinstance(elem_type, EmitC_ArrayType):
+                raise VerifyException("EmitC type cannot be a tensor of EmitC arrays")
+            self.verify(elem_type, constraint_context)
+            return
+
+        if isa(attr, EmitC_ArrayType):
+            elem_type = attr.get_element_type()
+            if isinstance(elem_type, EmitC_ArrayType):
+                raise VerifyException("nested EmitC arrays are not allowed")
```
superlopuh: let's add this on the ArrayType directly?

xdsl/dialects/emitc.py
```diff
@@ -209,6 +213,55 @@ def print_parameters(self, printer: Printer) -> None:
...
+        if isinstance(attr, TupleType):
+            for t in attr.types:
+                if isinstance(t, EmitC_ArrayType):
+                    raise VerifyException(
+                        "EmitC type cannot be a tuple of EmitC arrays"
+                    )
+                self.verify(t, constraint_context)
+            return
+
+        assert constraint_context is not None
```
superlopuh: Why this assert?

Jimmy2027: ah right sorry forgot to remove this, instead the function arg should require it

xdsl/dialects/emitc.py
Jimmy2027: moved in [5f67915](https://github.com/xdslproject/xdsl/pull/5039/commits/5f679157cc7c3ca93f247da7ef9e0872c9d73559)

PR #5040, opened 2025-08-17
core: add IntSetConstraint

superlopuh:
Allows us to specify a number of possible int values.

CC @Jimmy2027

### DIFF
--- a/xdsl/dialects/emitc.py
+++ b/xdsl/dialects/emitc.py
@@ -20,6 +20,7 @@ from xdsl.dialects.builtin import (
     FloatAttr,
     IndexType,
     IntAttr,
+    IntAttrConstraint,
     IntegerAttr,
     IntegerType,
     ShapedType,
@@ -36,7 +37,9 @@ from xdsl.ir import (
     TypeAttribute,
 )
 from xdsl.irdl import (
+    IntSetConstraint,
     IRDLOperation,
+    ParamAttrConstraint,
     ParsePropInAttrDict,
     irdl_attr_definition,
     irdl_op_definition,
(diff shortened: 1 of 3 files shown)


xdsl/irdl/constraints.py
```diff
@@ -834,6 +834,29 @@ def infer(self, context: ConstraintContext) -> int:
...
+    values: frozenset[int]
+
+    def verify(
+        self,
+        i: int,
+        constraint_context: ConstraintContext,
+    ) -> None:
+        if i not in self.values:
+            raise VerifyException(
+                f"Invalid value {i}, expected one of {list(self.values)}"
```
alexarice:
```suggestion
                f"Invalid value {i}, expected one of {set(self.values)}"
```
?

xdsl/dialects/emitc.py
```diff
@@ -213,18 +216,18 @@ class EmitC_SizeT(ParametrizedAttribute, TypeAttribute):
...
-def _is_supported_integer_type(type_attr: Attribute) -> bool:
-    """
-    Check if an IntegerType is supported by EmitC.
-    See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L96).
-    """
-    return (
-        isinstance(type_attr, IntegerType)
-        and type_attr.width.data in _SUPPORTED_BITWIDTHS
-    )
+EmitCIntegerConstr = ParamAttrConstraint(
```
alexarice, line 224: Why not use `IntegerAttr.constr` here?

alexarice (approved):

xdsl/dialects/emitc.py
superlopuh: Doesn't seem to exist

xdsl/irdl/constraints.py
superlopuh: Ye

xdsl/dialects/emitc.py
alexarice: Interesting, fair enough

alexarice:
Thinking about this again, would it not be better to have a `int` any_of constraint instead?

superlopuh:
IDK, we could have either, this is partially a result of me trying to do the I64 stuff in a clean way, and I'm trying to avoid having IntConstraints for the time being. Maybe this is misguided, but I guess we can always add the other one later.

Jimmy2027:
nice, thanks :) Also for the example, I was missing the AttrConstraint.verifies for #5039

PR #5041, opened 2025-08-17
core: replace GenericData.generic_constraint with generic_args_constraint

superlopuh:
This mostly reverts #4518, commit 28e783afb9ebfcb0f41bddd60b772c7ccac6c2a6.

I didn't realize that the previous mechanism was more expressive than the one I proposed, so I propose rolling it back, with some minor changes.

What this will let us do is have more control within the attribute itself of how to convert the generic parameters.


alexarice:
I wonder if it would be instead possible to make `GenericData` subclass from `ConstraintConvertible`. I'm not sure if this is the direction you wanted to go with all this stuff but it feels like it could be neat to basically have a lot of the current functionality of `irdl_to_attr_constraint` in `ConstraintConvertible` methods.

superlopuh:
Sorry I'm still cleaning it up, and making a bit more of a mess. I'll open a WIP PR with the full flow working and comment on the things I'm unhappy with, which will hopefully make it easier to discuss. I think I'm letting the perfect be the enemy of the good here, in that it would be good to merge something that lets us get rid of annotated, and then improve the situation with things working, as opposed to first trying to get them to be perfect and then seeing that things don't quite work.

alexarice:
Would it be worth trying to review the big PR (if it still exists)?

superlopuh:
Here it is: #4905, I'll open a stack of just the int-related changes, I think.

superlopuh:
Ended up not being necessary. We might want to revisit this if we ever attempt to get rid of mapping_type_vars.

PR #5045, opened 2025-08-18
dialects: (emitc) Add EmitCFloatTypeConstr

Jimmy2027:
Needed for #5039
Using the private _FloatType is probably not the right way, should there be a public alternative?

### DIFF
--- a/xdsl/dialects/emitc.py
+++ b/xdsl/dialects/emitc.py
@@ -27,7 +27,6 @@ from xdsl.dialects.builtin import (
     StringAttr,
     TensorType,
     TupleType,
-    _FloatType,
 )
 from xdsl.ir import (
     Attribute,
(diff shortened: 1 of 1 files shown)


xdsl/dialects/emitc.py
```diff
@@ -276,6 +285,8 @@ def is_supported_emitc_type(type_attr: Attribute) -> bool:
     match type_attr:
         case IntegerType():
             return EmitCIntegerConstr.verifies(type_attr)
+        case _FloatType():
+            return is_supported_float_type(type_attr)
```
superlopuh:
```suggestion
_constrs = EmitCIntegerConstr | EmitCFloatTypeConstr
if _constrs.verifies(type_attr):
    return True
match type_attr:
```

xdsl/dialects/emitc.py
```diff
@@ -229,17 +231,24 @@ class EmitC_SizeT(ParametrizedAttribute, TypeAttribute):
...
     """
     Check if a type is a supported floating-point type in EmitC.
     See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L117)
     """
-    match type_attr:
-        case Float16Type() | BFloat16Type() | Float32Type() | Float64Type():
-            return True
-        case _:
-            return False
+    return EmitCFloatTypeConstr.verifies(type_attr)
```
superlopuh:
```suggestion
```

xdsl/dialects/emitc.py
Jimmy2027: yeah, much better :)

PR #5046, opened 2025-08-18
core: add get_int_constr helper to convert int type expressions to constraints

superlopuh:
Another step towards generic integer types.

### DIFF
--- a/tests/irdl/test_int_constraint.py
+++ b/tests/irdl/test_int_constraint.py
@@ -95,6 +95,12 @@ def test_get_int_constr():
     assert get_int_constraint(
         Literal[1] | Literal[2, 3]  # noqa
     ) == IntSetConstraint(frozenset((1, 2, 3)))
+    assert get_int_constraint(
+        (Literal[1] | Literal[2]) | Literal[3]  # noqa
+    ) == IntSetConstraint(frozenset((1, 2, 3)))
+    assert get_int_constraint(
+        Literal[1] | (Literal[2] | Literal[3])  # noqa
+    ) == IntSetConstraint(frozenset((1, 2, 3)))
 
     with pytest.raises(PyRDLTypeError, match="Unexpected int type: <class 'str'>"):
         get_int_constraint(str)  # pyright: ignore[reportArgumentType]


xdsl/irdl/attributes.py
```diff
@@ -595,6 +600,51 @@ def eq(irdl: AttributeInvT) -> AttrConstraint[AttributeInvT]:
     return irdl_to_attr_constraint(irdl)
 
 
+def int_constr(arg: "int | TypeForm[int]") -> IntConstraint:
```
alexarice: Perhaps `int_to_constraint` would be more descriptive?

xdsl/irdl/attributes.py
```diff
@@ -595,6 +600,51 @@ def eq(irdl: AttributeInvT) -> AttrConstraint[AttributeInvT]:
...
+
+    if isclass(arg) and issubclass(arg, int):
+        return AnyInt()
+
+    if get_origin(arg) is Literal:
+        literal_args = get_args(arg)
+
+        if all(isinstance(literal_arg, int) for literal_arg in literal_args):
+            if len(literal_args) == 1:
+                return EqIntConstraint(get_args(literal_args)[0])
```
alexarice: Isn't this duplicated logic?

xdsl/irdl/attributes.py
```diff
@@ -595,6 +600,51 @@ def eq(irdl: AttributeInvT) -> AttrConstraint[AttributeInvT]:
...
+            ints = frozenset(get_args(union_arg)[0] for union_arg in union_args)
+            return IntSetConstraint(ints)
+
+    if isinstance(arg, TypeVar):
+        if arg.__bound__ is None:
+            raise PyRDLTypeError(
+                "Type variables used in IRDL are expected to be bound."
+            )
+        base = int_constr(arg.__bound__)
+        return IntTypeVarConstraint(arg, base)
```
alexarice: So a union of a `TypeVar` and `int` won't work right?

superlopuh: I'll add TypeVar support in a PR that leverages it

xdsl/irdl/attributes.py
superlopuh: Actually how about `get_int_constraint`?

xdsl/irdl/attributes.py
superlopuh: Not really, one is `1`, the other is `Literal[1]`. It's a kind of optimisation though to not create a singleton set constraint.

xdsl/irdl/attributes.py
alexarice: Sure

xdsl/irdl/attributes.py
alexarice: The above case covers `Literal[1]` as well right?

superlopuh: Ah no you're right my bad

xdsl/irdl/attributes.py
alexarice: Sounds good, lets get `I64` and friends deprecated first

alexarice: / using not annotated

xdsl/irdl/attributes.py
alexarice: Seems the second logic was broken

xdsl/irdl/attributes.py
superlopuh: We'll actually need it for that but that's two PRs down the line

alexarice: ok sure

tests/irdl/test_int_constraint.py
```diff
@@ -78,3 +80,14 @@ def test_mapping_type_vars():
...
+
+
+def test_get_int_constr():
+    assert get_int_constraint(int) == AnyInt()
+    assert get_int_constraint(Literal[1]) == EqIntConstraint(1)
+    assert get_int_constraint(Literal[2]) == EqIntConstraint(2)
+    assert get_int_constraint(Literal[2, 3]) == IntSetConstraint(frozenset((2, 3)))
+
+    assert get_int_constraint(
+        Literal[2] | Literal[3]  # noqa
```
alexarice, line 92: what's with the comment here?

xdsl/irdl/attributes.py
```diff
@@ -595,6 +599,38 @@ def eq(irdl: AttributeInvT) -> AttrConstraint[AttributeInvT]:
...
+                return IntSetConstraint(ints)
+
+    if get_origin(arg) is Union:
+        union_args = get_args(arg)
+        if all(
+            (get_origin(union_arg) is Literal) and len(get_args(union_arg)) == 1
+            for union_arg in union_args
+        ):
+            ints = frozenset(get_args(union_arg)[0] for union_arg in union_args)
+            return IntSetConstraint(ints)
```
alexarice, line 634: Why do we want to support `Literal[1] | Literal[2]` but not `Literal[1] | Literal[2,3]`?

superlopuh: Oh bother I didn't think of that I'll try to add it

alexarice: This was a genuine question, it's not clear to me which of these is actually going to appear. If we want to support all of these, I really feel it would make sense to have recursive function that builds a set of ints from an `int | TypeForm[int]`

tests/irdl/test_int_constraint.py
superlopuh: Our ruff linter doesn't like this syntax, presumably because it makes it harder to handle correctly than `Literal[2, 3]`, which it proposes as a fixit.

xdsl/irdl/attributes.py
superlopuh: The issue with that is that I'd rather not have an `AnyOfIntConstraint` as that would re-introduce backtracking, which is a little gross. We can add it if it unlocks something specific but for now I'd rather raise an error instead.

alexarice: `(Literal[1] | Literal[2]) | Literal[3]`?

alexarice: This doesn't need an `AnyOfIntConstraint`

superlopuh: Added

superlopuh: Python folds the Union for us automatically but I guess it's good to test that this actually happens

alexarice: So it does

xdsl/irdl/attributes.py
```diff
@@ -595,6 +599,43 @@ def eq(irdl: AttributeInvT) -> AttrConstraint[AttributeInvT]:
     return irdl_to_attr_constraint(irdl)
 
 
+def get_int_constraint(arg: "int | TypeForm[int]") -> IntConstraint:
```
alexarice, line 602: One last thing: Is there a reason to use a quoted type here? something to do with `TypeForm`?

superlopuh:
Thank you for all the small nits!

xdsl/irdl/attributes.py
superlopuh: Yeah my understanding is that `"a" | b` would actually call `__or__` on `str`, which would be weird, and they didn't want to allow `a | "b"` but not the other way round so Pyright tells you to just quote the whole thing.

alexarice: @superlopuh any comment on this? (Can address in next PR)

alexarice: Can't it just be unquoted completely though?

PR #5047, opened 2025-08-18
core: add get_constraint helper

superlopuh:
A union of the int converter and attr converter functions. Will be used in the next PR.

### DIFF
--- a/xdsl/irdl/attributes.py
+++ b/xdsl/irdl/attributes.py
@@ -637,7 +637,8 @@ def get_optional_int_constraint(arg: Any) -> IntConstraint | None:
 
 def get_int_constraint(arg: "int | TypeForm[int]") -> IntConstraint:
     """
-    Converts an int or an int type to the corresponding constraint.
+    If the input is an int or an int type, return the corresponding constraint,
+    otherwise raise `PyRDLTypeError`.
     """
     if (ic := get_optional_int_constraint(arg)) is not None:
         return ic


xdsl/irdl/attributes.py
```diff
@@ -633,6 +634,15 @@ def get_int_constraint(arg: "int | TypeForm[int]") -> IntConstraint:
             )
             return IntSetConstraint(ints)
 
+
+def get_int_constraint(arg: "int | TypeForm[int]") -> IntConstraint:
```
alexarice, line 638:
```suggestion
def get_int_constraint(arg: int | TypeForm[int]) -> IntConstraint:
```
I still don't understand why this doesn't work

xdsl/irdl/attributes.py
```diff
@@ -653,3 +663,15 @@ def single_range_constr_coercion(
...
+def get_constraint(
+    arg: "IRDLAttrConstraint | int | TypeForm[int]",
+) -> AttrConstraint | IntConstraint:
+    """
+    Converts the input expression, constraint, or type to the corresponding constraint.
+    """
+    if (ic := get_optional_int_constraint(arg)) is not None:
+        return ic
+
+    return irdl_to_attr_constraint(arg)  # pyright: ignore[reportArgumentType]
```
alexarice, line 677: Is this an issue with generics?

superlopuh: Yeah there's no guarantee that `arg` is not an int thing at this point

xdsl/irdl/attributes.py
superlopuh:
```
xdsl/irdl/attributes.py:638: in <module>
    def get_int_constraint(arg: int | TypeForm[int]) -> IntConstraint:
                                      ^^^^^^^^
E   NameError: name 'TypeForm' is not defined
```

alexarice: It works in other places though?

superlopuh: It's not in CPython, which tries to evaluate the types of function arguments.

alexarice: I'll have a mess around

PR #5048, opened 2025-08-18
core: (Operation) simplify `get_variadic_sizes` to `verify_variadic_size`

alexarice:
We no longer need the output of `get_variadic_sizes`. This refactors these methods slightly, simplifying them/improving some error messages, and splitting off the `same_size` portion so that it can be reused in `OperandsDirective` in the (near) future.

### DIFF
--- a/benchmarks/rewriting.py
+++ b/benchmarks/rewriting.py
@@ -14,7 +14,7 @@ from xdsl.dialects.arith import (
 from xdsl.dialects.builtin import Builtin, IntegerAttr, IntegerType, ModuleOp
 from xdsl.ir import Region
 from xdsl.ir.post_order import PostOrderIterator
-from xdsl.irdl import VarIRConstruct, get_variadic_sizes
+from xdsl.irdl import VarIRConstruct, verify_variadic_size
 from xdsl.parser import Parser as XdslParser
 from xdsl.pattern_rewriter import PatternRewriter, Worklist
 from xdsl.rewriter import InsertPoint
@@ -188,14 +188,14 @@ class PatternRewriting(RewritingMicrobenchmarks):
         """
         self.pattern_rewriter.insert_op((self.sub_op,), self.insert_point)
 
-    def time_get_variadic_sizes(self) -> None:
-        """Time `get_variadic_sizes`.
+    def time_verify_variadic_size(self) -> None:
+        """Time `verify_variadic_size`.
 
-        Exercise getting the variadic size of an operation, including a
+        Exercise verifying the variadic size of an operation, including a
         significant amount of logic to check lengths and types. This is invoked
-        whenever an operation argument is defined with `irdl_op_arg_definition`.
+        4 times whenever an operation is verified.
         """
-        get_variadic_sizes(self.add_op, self.add_op_def, self.add_op_construct)
+        verify_variadic_size(self.add_op, self.add_op_def, self.add_op_construct)
 
 
 class Canonicalization(RewritingMicrobenchmarks):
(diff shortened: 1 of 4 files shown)


benchmarks/rewriting.py
```diff
@@ -188,14 +188,14 @@ def ignore_time_pattern_rewriter_insert_op(self) -> None:
...
+    def time_verify_variadic_size(self) -> None:
+        """Time `verify_variadic_size`.
 
-        Exercise getting the variadic size of an operation, including a
+        Exercise verifying the variadic size of an operation, including a
         significant amount of logic to check lengths and types. This is invoked
-        whenever an operation argument is defined with `irdl_op_arg_definition`.
+        4 times whenever an operation is verified.
         """
-        get_variadic_sizes(self.add_op, self.add_op_def, self.add_op_construct)
```
alexarice, line 198: Not sure if this is reasonable.

tests/filecheck/dialects/scf/for_args_number.mlir
```diff
@@ -4,7 +4,7 @@
   %lb = "arith.constant"() {"value" = 0 : index} : () -> index
   %ub = "arith.constant"() {"value" = 42 : index} : () -> index
   %step = "arith.constant"() {"value" = 7 : index} : () -> index
-// CHECK: Expected at least 3 operands, got 4
```
alexarice, line 7: Not sure what was going on with this error message before

tests/irdl/test_operation_builder.py
```diff
@@ -349,7 +349,7 @@ def test_same_size_operand_builder2():
     with pytest.raises(
         VerifyException,
         match=re.escape(
-            "Operation does not verify: Operation has 3 operands for 2 variadic operands marked as having the same size."
```
alexarice: Might try to revert this, not sure it was actually an improvement.

xdsl/irdl/operations.py
```diff
@@ -222,6 +222,12 @@ class AttrSizedSegments(IRDLOption, ABC):
     as_property: bool = False
     """Name of the attribute containing the segment sizes."""
 
+    def container(self, op: Operation) -> dict[str, Attribute]:
+        if self.as_property:
+            return op.properties
+        else:
+            return op.attributes
```
alexarice: Seems reasonable to have this helper here.

tests/filecheck/dialects/scf/for_args_number.mlir
superlopuh: weird

xdsl/irdl/operations.py
superlopuh: I wonder if this would be worth merging first with the other tiny changes before reviewing core logic changes?

alexarice: Can do, a lot of the logic change is actually just reordering the functions I think, I'll try to reduce the diff a bit

alexarice:
rebased

benchmarks/rewriting.py
math-fehr: Why? Do you mean because we run it 4 times?

math-fehr (approved):
Thanks for the clean-up!

benchmarks/rewriting.py
alexarice: I meant I didn't know if I had done the changes to the benchmark correctly

superlopuh: Seems good to me

superlopuh (approved):

PR #5050, opened 2025-08-18
dialects: (builtin) add generic argument to IntAttr

superlopuh:


xdsl/dialects/builtin.py
```diff
@@ -185,8 +186,11 @@ def print_builtin(self, printer: Printer):
     @staticmethod
     @override
     def constr(
-        constr: IRDLAttrConstraint[AttributeInvT] | RangeConstraint[AttributeInvT],
+        constr: IRDLAttrConstraint[AttributeInvT]
+        | RangeConstraint[AttributeInvT]
+        | IntConstraint,
     ) -> ArrayOfConstraint[AttributeInvT]:
```
math-fehr: This is ill-typed?
If you pass an int constraint, it should return something related to the int

tests/irdl/test_attribute_definition.py
```diff
@@ -662,8 +662,9 @@ def print_parameter(self, printer: Printer) -> None:
     @staticmethod
     @override
     def constr(
-        constr: AttrConstraint[AttributeCovT],
+        constr: AttrConstraint[AttributeCovT] | IntConstraint,
     ) -> DataListAttr[AttributeCovT]:
+        assert isinstance(constr, AttrConstraint)
```
alexarice: This change either feels unrelated, or should be in a different PR

tests/test_pyrdl.py
```diff
@@ -421,6 +422,16 @@ def test_irdl_to_attr_constraint():
...
         TestEnumAttr(TestEnum.A)
     )
+    assert irdl_to_attr_constraint(IntAttr) == BaseAttr(IntAttr)
+    assert irdl_to_attr_constraint(IntAttr[int]) == IntAttrConstraint(
+        int_constraint=AnyInt()
+    )
+    assert irdl_to_attr_constraint(IntAttr[Literal[1]]) == IntAttrConstraint(
+        int_constraint=EqIntConstraint(value=1)
+    )
+    assert irdl_to_attr_constraint(IntAttr[2]) == IntAttrConstraint(
```
alexarice, line 432: I feel this is where I'm getting confused. Is this really valid python? If it is, why can't arbitrary constraints be put in as generic arguments? Is there special casing for integers?

xdsl/dialects/builtin.py
alexarice: Yeah, this confuses me too, what does `ArrayOfConstraint.constr(EqIntConstraint(2))` even mean?

alexarice: I guess the point is that `GenericData` enforces the type of a `constr` method?

xdsl/irdl/attributes.py
```diff
@@ -81,7 +81,7 @@ class GenericData(Data[_DataElement], ABC):
 
     @staticmethod
     @abstractmethod
-    def constr(constr: AttrConstraint) -> AttrConstraint:
+    def constr(constr: AttrConstraint | IntConstraint) -> AttrConstraint:
```
alexarice: I'm not a huge fan of this. Is this what the other PR was trying to solve?

xdsl/irdl/attributes.py
```diff
@@ -665,6 +667,14 @@ def single_range_constr_coercion(
     return SingleOf(irdl_to_attr_constraint(attr))
 
 
+@overload
+def get_constraint(arg: "int | TypeForm[int]") -> IntConstraint: ...
+
+
+@overload
+def get_constraint(arg: IRDLAttrConstraint) -> AttrConstraint: ...
```
alexarice: This should be in the previous PR I think

tests/test_pyrdl.py
superlopuh: The special casing is for integers, strings, and enum values

alexarice: absolutely cursed, but ok

superlopuh: Image

alexarice: I can't reproduce, this works for me:
```python
from typing import Generic, TypeVar

A = TypeVar("A")

class MyOtherClass:
    pass

class MyClass(Generic[A]):
    field: A

MyClass[MyOtherClass()]
```

xdsl/dialects/builtin.py
superlopuh: So this was kind of the motivation for reverting the constr change. Some of the GenericData will have attributes inside, and I propose to now add an int version. How do we type this? Either all GenericData must accept the union and verify themselves (this PR), or they accept what they accept and we YOLO (not a fan), or we revert to them parsing the type arguments themselves in a helper that is not `.constr`.

xdsl/irdl/attributes.py
superlopuh: Yep

tests/test_pyrdl.py
superlopuh: This runs but is Pyright happy?

xdsl/dialects/builtin.py
superlopuh: #5041

tests/test_pyrdl.py
alexarice: yeah, it's happy

tests/irdl/test_attribute_definition.py
```diff
@@ -661,10 +660,8 @@ def print_parameter(self, printer: Printer) -> None:
 
     @staticmethod
     @override
-    def constr(
-        constr: AttrConstraint[AttributeCovT],
-    ) -> DataListAttr[AttributeCovT]:
-        return DataListAttr(constr)
+    def constr() -> DataListAttr:
```
superlopuh: I'll make this change in a separate PR

xdsl/irdl/attributes.py
```diff
@@ -512,12 +512,26 @@ def irdl_to_attr_constraint(
...
     # GenericData case
     if isclass(origin) and issubclass(origin, GenericData):
-        args = get_args(irdl)
+        args: tuple[IRDLAttrConstraint | int | TypeForm[int], ...] = get_args(irdl)
         if len(args) != 1:
             raise PyRDLTypeError(f"GenericData args must have length 1, got {args}")
-        constr = irdl_to_attr_constraint(args[0])
 
-        return cast(AttrConstraint[AttributeInvT], origin.constr(constr))
+        inner = get_constraint(args[0], allow_type_var=allow_type_var)
```
alexarice: Can this be merged with below?

xdsl/irdl/attributes.py
```diff
@@ -665,13 +679,39 @@ def single_range_constr_coercion(
...
     arg: "IRDLAttrConstraint | int | TypeForm[int]",
+    *,
+    allow_type_var: bool = False,
+) -> AttrConstraint | IntConstraint: ...
+
+
+def get_constraint(
+    arg: "IRDLAttrConstraint | int | TypeForm[int]",
+    *,
+    allow_type_var: bool = False,
```
alexarice: Quickly split this out?

tests/irdl/test_attribute_definition.py
superlopuh: #5051

xdsl/irdl/attributes.py
superlopuh: Also in #5051 but can split out again if you'd like

superlopuh: I'll do a separate PR after 5051 is merged

xdsl/dialects/builtin.py
```diff
@@ -370,6 +375,13 @@ def __bool__(self) -> bool:
         """Returns True if value is non-zero."""
         return bool(self.data)
 
+    @staticmethod
+    @override
+    def constr(constr: IntConstraint | None = None) -> AttrConstraint[IntAttr]:
```
alexarice, line 380: Does `IntConstraint` need to be generic like `AttrConstraint` is? Possibly for future.

superlopuh: Yeah I think this might be a good idea in the future but for now feels unneccessary

PR #5051, opened 2025-08-18
core: GenericData.constr no parameter

superlopuh:
A more flexible setup separating concerns of defining the constraints and specifying generic arguments.


xdsl/irdl/attributes.py
```diff
@@ -665,13 +679,39 @@ def single_range_constr_coercion(
     return SingleOf(irdl_to_attr_constraint(attr))
 
 
+@overload
```
alexarice: This still looks unrelated

PR #5053, opened 2025-08-18
core: Add `AtMost` integer contraint

mudit2812:
Add `AtMost` integer constraint to allow constraining integers to upper bounds.

### DIFF
--- a/tests/irdl/test_int_constraint.py
+++ b/tests/irdl/test_int_constraint.py
@@ -54,8 +54,6 @@ def test_at_most():
     AtMost(2).verify(1, ConstraintContext())
     with pytest.raises(VerifyException, match="Expected integer <= 2, got 3"):
         AtMost(2).verify(3, ConstraintContext())
-    with pytest.raises(VerifyException, match="Expected integer <= 2, got 4"):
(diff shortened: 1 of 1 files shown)


tests/irdl/test_int_constraint.py
```diff
@@ -41,6 +42,22 @@ def test_at_least():
...
+    AtMost(1).verify(0, ConstraintContext())
+    with pytest.raises(VerifyException):
+        AtMost(0).verify(1, ConstraintContext())
+
+    AtMost(2).verify(2, ConstraintContext())
+    AtMost(2).verify(1, ConstraintContext())
+    with pytest.raises(VerifyException):
+        AtMost(2).verify(3, ConstraintContext())
+    with pytest.raises(VerifyException):
+        AtMost(2).verify(4, ConstraintContext())
```
superlopuh: Can you please add a match statement here?
```suggestion
    with pytest.raises(VerifyException, match="Expected integer <= 2, got 3"):
        AtMost(2).verify(3, ConstraintContext())
```

superlopuh: Also I don't think we need two tests

PR #5055, opened 2025-08-18
core: (irdl_to_attr_constraint) slight refactor/cleanup

alexarice:
Diff looks horrible because of reorderings, but I think this cleans up the logic a bit.

### DIFF
--- a/xdsl/irdl/attributes.py
+++ b/xdsl/irdl/attributes.py
@@ -9,7 +9,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Hashable, Sequence
 from dataclasses import dataclass
 from inspect import get_annotations, isclass
-from types import FunctionType, GenericAlias, UnionType
+from types import FunctionType, UnionType
 from typing import (
     TYPE_CHECKING,
     Annotated,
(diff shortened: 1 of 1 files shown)


alexarice:
(I'm happy to continue rebasing this over further changes)

xdsl/irdl/attributes.py
```diff
@@ -510,62 +497,17 @@ def irdl_to_attr_constraint(
...
-            AttrConstraint[AttributeInvT],
-            origin.constr().mapping_type_vars(type_var_mapping),
-        )
-
-    # Generic ParametrizedAttributes case
-    # We translate it to constraints over the attribute parameters.
-    if (
-        isclass(origin)
-        and issubclass(origin, ParametrizedAttribute)
-        and issubclass(origin, Generic)
```
superlopuh, line 541: How do you feel about doing `if issubclass(Generic)` first and then branch into data and parametrized attribute?

alexarice: Is this not what I've done? Or do you mean just to keep the ordering closer to what it is now

superlopuh: Yeah sorry keep ordering closer

alexarice: Reduced the diff as much as I can by avoiding any reordering.

PR #5057, opened 2025-08-18
dialects: (bufferization) rename to_memref to to_buffer

Hatsunespica:
According to the MLIR upstream, the old `to_memref` operation is renamed to `to_buffer`

https://mlir.llvm.org/docs/Dialects/BufferizationOps/#bufferizationto_buffer-bufferizationtobufferop

This PR is intended to change the name.

### DIFF
--- a/tests/test_pyrdl.py
+++ b/tests/test_pyrdl.py
@@ -8,6 +8,7 @@ from typing import Literal
 import pytest
 from typing_extensions import Self, TypeVar
 
+from xdsl.dialects.builtin import IntAttr, IntAttrConstraint
 from xdsl.ir import Attribute, Data, ParametrizedAttribute
 from xdsl.irdl import (
     AllOf,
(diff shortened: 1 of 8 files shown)


Hatsunespica:
> Are both currently supported by MLIR? Should we also support both, noting that `to_memref` is deprecated?

I don't think MLIR has `to_memref` anymore. You can try with the most recent MLIR on this exmaple:

```
func.func @test_to_buffer(%arg0: tensor<2xi64>) -> (memref<2xi64>) {
  %0 = bufferization.to_buffer %arg0 : tensor<2xi64> to memref<2xi64>
  return %0 : memref<2xi64>
}
```

It works with `to_buffer`, but not for `to_memref`

superlopuh:
This seems like something for when we update our MLIR version, the docs are generated for main, and we're a little behind.

Hatsunespica:
> This seems like something for when we update our MLIR version, the docs are generated for main, and we're a little behind.

Aha, so I should close this PR and wait for updating MLIR??

Hatsunespica:
> This seems like something for when we update our MLIR version, the docs are generated for main, and we're a little behind.

The only test case failed right now is the one with `mlir-conversion/dialects/bufferizatoin/ops.mlir`, MLIR-based Testing. I can keep this test as the original one so it works with the current MLIR version, and change it later when updating MLIR version?

superlopuh:
No I think we should update them together. Let's close this PR for now. I'm happy to update the version of LLVM/MLIR if both `brew` and `apt` install newer versions now.

Hatsunespica:
> No I think we should update them together. Let's close this PR for now. I'm happy to update the version of LLVM/MLIR if both `brew` and `apt` install newer versions now.

OK, let's do that. Once you updated both version, let me know so I can reopen PR here and submit my proposed change

PR #5059, opened 2025-08-18
core: ((Operands|Results)Directive) refactor to allow multiple variadics

alexarice:
Verifies the test from #4983 

All ended up a bit messier than I would have liked, an alternative could be to store Accessors in the `op_def`. I can probably also split this into multiple PRs

Probably needs a corresponding `operands` and `operand types` test.

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -31,18 +31,22 @@ from xdsl.ir import (
     TypedAttribute,
 )
 from xdsl.irdl import (
+    BaseAccessor,
     ConstraintContext,
     IRDLOperation,
     IRDLOperationInvT,
     OpDef,
     OptionalDef,
     Successor,
+    VarIRConstruct,
     VarOperand,
+    get_construct_defs,
     is_const_classvar,
+    verify_variadic_same_size,
 )
 from xdsl.parser import Parser, UnresolvedOperand
 from xdsl.printer import Printer
-from xdsl.utils.exceptions import PyRDLError
+from xdsl.utils.exceptions import PyRDLError, VerifyException
 from xdsl.utils.hints import isa
 from xdsl.utils.mlir_lexer import PunctuationSpelling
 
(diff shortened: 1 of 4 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -867,8 +887,10 @@ def parse_types(self, parser: Parser, state: ParsingState) -> bool:
             or []
         )
 
-        if s := self.set_types(state, types):
-            parser.raise_error(s, at_position=pos_start, end_position=parser.pos)
+        try:
+            self.set_types(state, types)
+        except VerifyException as e:
+            parser.raise_error(str(e), at_position=pos_start, end_position=parser.pos)
```
math-fehr, line 894: Maybe adding a context on this?
Like starting the string with "While parsing operation types" or something like this?

alexarice:
@math-fehr did you have any thoughts on how best to route the accessors to the directive? I've currently put the op_type in the ParsingState but that feels a bit yucky. Could alternatively put them all in the `OpDef` though I'm not sure if this is conceptually the correct place for them to live.

superlopuh (approved):
Can we merge this straight to main? Should I close my PR?

alexarice:
It's not on top of your PR, it's on top of mine.

xdsl/irdl/declarative_assembly_format.py
alexarice: Do you mean just adding to the string? I can't see a specific option for context on parse errors

superlopuh: I think it would be worth seeing the output of this code before changing it. I would expect it to be fairly clear what the problem is without more context, but seeing the error message as it's printed would help make that decision.

alexarice:
```mlir
tosa.add %0 : ...
         ^^^
         Expected 2 operands, but got 1
```

alexarice:
```
tosa.add %0, %1 : (tensor<?xi32>) -> tensor<?xi32>
                   ^^^^^^^^^^^^^
                   Expected 2 operand types, but got 1
```
I feel it's pretty clear tbh

PR #5063, opened 2025-08-19
documentation: Add a simple expression parser to marimo

tobiasgrosser:
This PR adds a simple expression parser to the Marimo notebook.


Moxinilian:
Is there a way to have the parser live in a separate file/package? That would make maintenance much more convenient I think! Maybe by pushing the parser project from the other repository to pypi?

tobiasgrosser:
@Moxinilian, I agree that it would be desirable if we could pull in external packages. At the moment, even installing xdsl breaks, so consider this PR mostly as a test of mkdocs-marimo. I first want it to actually work before I start to see if we can get some more advanced packaging tricks in. 

 I suggest you develop your code in your repository. The moment `mkdocs-marimo` basic work, we can look into better packaging strategies.

@superlopuh offered to look into the failure here, so II hand the token to him for now.

tobiasgrosser:
This has been superseeded by https://github.com/xdslproject/xdsl/pull/5103.

PR #5065, opened 2025-08-19
dialects: (builtin) make IntegerType generic over width type

superlopuh:
Also adds functionality to `get_optional_int_constraint` to construct a constraint from a TypeVar.

I updated the EmitC dialect as it's the only one that cares just about the bitwidth and not signedness.

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -54,6 +54,7 @@ from xdsl.traits import (
     Pure,
 )
 from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.hints import isa
 from xdsl.utils.str_enum import StrEnum
 from xdsl.utils.type import get_element_type_or_self, have_compatible_shape
 
@@ -1319,8 +1320,7 @@ class TruncIOp(IRDLOperation):
         super().__init__(operands=[op], result_types=[target_type])
 
     def verify_(self) -> None:
-        assert isinstance(self.input.type, IntegerType)
-        assert isinstance(self.result.type, IntegerType)
+        assert isa(self.input.type, IntegerType)
         if not self.result.type.width.data < self.input.type.width.data:
             raise VerifyException(
                 "Destination bit-width must be smaller than the input bit-width"
@@ -1342,8 +1342,7 @@ class ExtSIOp(IRDLOperation):
         super().__init__(operands=[op], result_types=[target_type])
 
     def verify_(self) -> None:
-        assert isinstance(self.input.type, IntegerType)
-        assert isinstance(self.result.type, IntegerType)
+        assert isa(self.input.type, IntegerType)
         if not self.result.type.width.data > self.input.type.width.data:
             raise VerifyException(
                 "Destination bit-width must be larger than the input bit-width"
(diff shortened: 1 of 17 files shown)


xdsl/dialects/emitc.py
```diff
@@ -216,14 +214,13 @@ class EmitC_SizeT(ParametrizedAttribute, TypeAttribute):
...
 """
-Constraint for the bitwidth parameter of integer types supported by EmitC.
+Type for integer types supported by EmitC.
+See external [documentation](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/EmitC/IR/EmitC.cpp#L96).
 """
 
-EmitCIntegerConstr = ParamAttrConstraint(
-    IntegerType, (IntAttrConstraint(EmitCIntegerBitwidthConstr), None)
-)
+EmitCIntegerTypeConstr = irdl_to_attr_constraint(EmitCIntegerType)
```
superlopuh, line 224: CC @Jimmy2027

xdsl/dialects/builtin.py
```diff
@@ -888,13 +892,7 @@ def from_bool(value: bool) -> BoolAttr:
...
     def print_builtin(self, printer: Printer) -> None:
         # boolean shorthands
-        if (
-            isinstance(
-                (ty := self.get_type()),
-                IntegerType,
-            )
-            and ty.width.data == 1
-        ):
+        if isa((ty := self.get_type()), IntegerType[1]):
```
superlopuh, line 895: Small improvement

xdsl/dialects/arith.py
```diff
@@ -1319,8 +1320,7 @@ def __init__(self, op: SSAValue | Operation, target_type: IntegerType):
         super().__init__(operands=[op], result_types=[target_type])
 
     def verify_(self) -> None:
-        assert isinstance(self.input.type, IntegerType)
-        assert isinstance(self.result.type, IntegerType)
```
alexarice, line 1323: Where did the result type check go? Just not needed?

alexarice (approved):

xdsl/dialects/arith.py
superlopuh: Yeah results are typed already so it's not necessary

xdsl/dialects/builtin.py
math-fehr: This doesn't consider signed/unsigned, but I think it was a bug before

xdsl/dialects/emitc.py
```diff
@@ -217,14 +215,13 @@ class EmitC_SizeT(ParametrizedAttribute, TypeAttribute):
     name = "emitc.size_t"
 
 
-EmitCIntegerBitwidthConstr = get_int_constraint(Literal[1, 8, 16, 32, 64])
+EmitCIntegerType = IntegerType[Literal[1, 8, 16, 32, 64]]
```
math-fehr, line 218: Very cool

math-fehr (approved):
Looks cool!

xdsl/dialects/builtin.py
superlopuh: Do we care? Should we print 0/1 for non-signless?

math-fehr: I think we do care, we should not print true/false for non-signless, so your implementation is correct

superlopuh: OK I'll fix this after merge

math-fehr: Wait fix what? What you wrote is correct

PR #5067, opened 2025-08-19
backend: integration harness

qaco:
Add an harness allowing to generate an executable from the integration test.


tests/filecheck/projects/libxsmm/main.c
superlopuh, line 1: Can you please add a README.md to document how to use this?

qaco: Yes it is done !

superlopuh: TY

PR #5070, opened 2025-08-19
dialects: (emitc) Add EmitCPointerWideTypeConstr

Jimmy2027:
based on #5069


xdsl/dialects/emitc.py
```diff
@@ -261,7 +254,7 @@ def is_integer_index_or_opaque_type(
     return (
         EmitCIntegerTypeConstr.verifies(type_attr)
         or isinstance(type_attr, IndexType)
-        or is_pointer_wide_type(type_attr)
+        or EmitCPointerWideTypeConstr.verifies(type_attr)
```
alexarice, line 257:
```suggestion
        or isinstance(type_attr, EmitCPointerWideType)
```
This might be even simpler? Could likely do similar for uses of `EmitCFloatTypeConstr`

superlopuh: isinstance won't work in general as it doesn't check for our generic attributes for IntegerType so it might be better to stick with this for now

superlopuh (approved):

xdsl/dialects/emitc.py
superlopuh: Also I think this bit of code will disappear very soon

Jimmy2027: yup :)

alexarice (approved):

PR #5072, opened 2025-08-19
dialects: (emitc) add EmitCArrayElementTypeConstr

Jimmy2027:
reorders the type definitions in emitc.py to be able to define the constraints before they're used.
Happy to open a PR for 27d476e5057eedc75df6380cb61e820a8ac9ca02 if needed.

### DIFF
--- a/tests/dialects/test_emitc.py
+++ b/tests/dialects/test_emitc.py
@@ -1,6 +1,6 @@
 import pytest
 
-from xdsl.dialects.builtin import IntegerType
+from xdsl.dialects.builtin import i32
 from xdsl.dialects.emitc import EmitC_ArrayType, EmitC_CallOpaqueOp
 from xdsl.utils.exceptions import VerifyException
 
@@ -9,7 +9,7 @@ def test_emitc_array_negative_dimension():
     with pytest.raises(
         VerifyException, match="EmitC array dimensions must have non-negative size"
     ):
-        EmitC_ArrayType([-1], IntegerType(32))
+        EmitC_ArrayType([-1], i32)
 
 
 def test_call_opaque_with_str_callee():
--- a/xdsl/dialects/emitc.py
+++ b/xdsl/dialects/emitc.py
@@ -29,6 +29,7 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.ir import (
     Attribute,
+    AttributeCovT,
     Dialect,
     ParametrizedAttribute,
     SSAValue,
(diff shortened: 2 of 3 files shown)


tests/dialects/test_emitc.py
superlopuh, line 1: why this change?

xdsl/dialects/emitc.py
```diff
@@ -91,7 +182,7 @@ def verify(self) -> None:
             )
 
         # Check that the element type is a supported EmitC type.
-        if not self._is_valid_element_type(element_type):
+        if not EmitCArrayElementTypeConstr.verifies(element_type):
             raise VerifyException(
                 f"EmitC array element type '{element_type}' is not a supported EmitC type."
             )
```
superlopuh, line 189: Should be done automatically
```suggestion
```

superlopuh: I don't think the check above is necessary either, as it'll already be excluded by the constraint

tests/dialects/test_emitc.py
Jimmy2027: pyright was not happy:
https://github.com/xdslproject/xdsl/actions/runs/17071946484/job/48402989448

xdsl/dialects/emitc.py
Jimmy2027: ah right, forgot to remove them in this pr, sorry

tests/dialects/test_emitc.py
superlopuh: This feels like a usability problem, since we definitely want users to be able to assign i32 there... I'm surprised that `i32` is `IntegerType[int]`, though

Jimmy2027: yeah looks like the type checker is a bit limited there... not sure how to fix this

superlopuh: Let's merge the rest of the changes but keep the `.verify` function and then I can play with it locally?

Jimmy2027: sounds good, thanks!

Jimmy2027: once this is fixed [07a9622](https://github.com/xdslproject/xdsl/pull/5072/commits/07a962200fb02b116dfd45779bac70f4c747790c) can be reverted

superlopuh: Locally this works for me...

Jimmy2027: you mean pyright does not complain? functionally this should be equivalent, its just the type inference being limited

superlopuh: Yeah it doesn't complain

Jimmy2027: interesting

Jimmy2027: fails for me with both pyright 1.1.404 and 1.1.403 on 1fa610f017466fa033d68c85f9a92aeaaa9e3f94

superlopuh: Could you please open a new PR with the changes you'd like to make so we can iterate there?

superlopuh: The change I made locally is just replace the verify with the new type and that seems to be OK Pyright-wise

Jimmy2027: sure will do!

PR #5076, opened 2025-08-19
documentation: build eqsat notebook to readonly marimo app

superlopuh:
This notebook isn't reactive, so it's a simple enough migration.

Interestingly, it looks like the race condition for notebook updates affects the text displayed when switching between notebooks also.


superlopuh:
converting to draft while I debug how to make it only install xdsl and not all the other things from pypi

math-fehr (approved):

tobiasgrosser:
@superlopuh, can you briefly describe what the issue here is?

superlopuh:
There are two issues: one is that when the eqsat notebook page opens the xdsl cell outputs get duplicated, and the other is that switching between the notebook tabs keeps the contents of the first text cell of the other notebook in the newly opened one.

PR #5077, opened 2025-08-20
dialects: (complex) Add complex dialect

jumerckx:
This passes the filecheck tests from MLIR


xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+
+    def verify_(self):
+        if len(self.value.data) != 2:
+            raise VerifyException("complex.constant must have an array of 2 values")
+        elem_type = self.complex.type.element_type
+        # TODO: this check might be wrong, x can be something other than float?
+        if any(cast(FloatAttr, x).type != elem_type for x in self.value.data):
+            raise VerifyException(
+                "complex.constant value types must match the complex element type"
+            )
```
jumerckx: I don't yet fully grasp types/constraints in xDSL, but `cast(FloatAttr, x)` is wrong as `x` can be something else.
e.g. `%cst_i1 = complex.constant [1 : i1, 0 : i1] : complex<i1>`

What should the cast be, something related to `ComplexElementCovT`? And probably there needs to be an explicit check as well?

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+    irdl_op_definition,
+    operand_def,
+    prop_def,
+    result_def,
+    traits_def,
+)
+from xdsl.traits import ConstantLike, Pure, SameOperandsAndResultType
+from xdsl.utils.exceptions import VerifyException
+
+# Type constraint for our new ComplexType
```
alexarice:
```suggestion
```

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+ComplexTypeConstr = base(ComplexType)
+
+
+class ComplexUnaryOp(IRDLOperation, abc.ABC):
```
alexarice: This doesn't quite agree with the MLIR superclass of the same name. For example there `AbsOp` inherits from `ComplexUnaryOp`. I'm happy to keep this superclass but feel it should be named something else. I also believe we tend to use `*Operation` for superclasses and `*Op` for final operations.

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+    """Base class for unary operations on complex numbers."""
+
+    T: ClassVar = VarConstraint("T", ComplexTypeConstr)
+    complex = operand_def(T)
+    result = result_def(T)
+    fastmath = prop_def(FastMathFlagsAttr, default_value=FastMathFlagsAttr("none"))
+
+    traits = traits_def(Pure(), SameOperandsAndResultType())
```
alexarice: Having both the constraints and `SameOperandsAndResultType()` feels unnecessary to me.

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+    operand_def,
+    prop_def,
+    result_def,
+    traits_def,
+)
+from xdsl.traits import ConstantLike, Pure, SameOperandsAndResultType
+from xdsl.utils.exceptions import VerifyException
+
+# Type constraint for our new ComplexType
+ComplexTypeConstr = base(ComplexType)
```
alexarice: It seems that both `ComplexUnaryOp` and `ComplexArithmeticOp` should be taking only complex numbers over float types?
```suggestion
ComplexTypeConstr = ComplexType.constr(AnyFloat)
```

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+            operands=[lhs_ssa, rhs],
+            result_types=[lhs_ssa.type],
+            properties={"fastmath": fastmath},
+        )
+
+    def verify_(self):
+        if self.lhs.type != self.rhs.type:
+            raise VerifyException("lhs and rhs of binary op must have the same type")
+        if self.lhs.type != self.result.type:
+            raise VerifyException("result type must match operand types")
```
alexarice: This is covered by the constraints

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+    traits = traits_def(Pure())
+
+    assembly_format = "$lhs `,` $rhs attr-dict `:` type($lhs)"
+
+    def __init__(self, lhs: SSAValue | Operation, rhs: SSAValue | Operation):
+        super().__init__(operands=[lhs, rhs], result_types=[IntegerType(1)])
+
+    def verify_(self):
+        if self.lhs.type != self.rhs.type:
+            raise VerifyException("lhs and rhs of compare op must have the same type")
```
alexarice: This is also covered by constraints

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+
+@irdl_op_definition
+class ConjOp(ComplexUnaryOp):
+    name = "complex.conj"
+
+
+@irdl_op_definition
+class ConstantOp(IRDLOperation):
+    name = "complex.constant"
+    value = prop_def(ArrayAttr)
```
alexarice:
```suggestion
    value = prop_def(ArrayOfConstraint(RangeOf(ComplexTypeConstr).of_length(2))
```

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+    assembly_format = "$value attr-dict `:` type($complex)"
+
+    def __init__(self, value: ArrayAttr, result_type: ComplexType):
+        super().__init__(properties={"value": value}, result_types=[result_type])
+
+    def verify_(self):
+        if len(self.value.data) != 2:
+            raise VerifyException("complex.constant must have an array of 2 values")
+        elem_type = self.complex.type.element_type
+        # TODO: this check might be wrong, x can be something other than float?
```
alexarice: See above constraint

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+        imag_type = self.imaginary.type
+        if real_type != imag_type:
+            raise VerifyException(
+                "real and imaginary parts of complex.create must have the same type"
+            )
+        complex_element_type = self.complex.type.element_type
+        if real_type != complex_element_type:
+            raise VerifyException(
+                "operands of complex.create must match the element type of the result"
+            )
```
alexarice: This custom verify is unnecessary again

xdsl/dialects/complex.py
alexarice: I've put comments on this below, but it looks like you should use `ComplexTypeConstr = ComplexType.constr(AnyFloat)` everywhere?

xdsl/dialects/complex.py
superlopuh: Does this work?
```suggestion
ComplexTypeConstr = irdl_to_attr_constraint(ComplexType[AnyFloat])
```

xdsl/dialects/complex.py
jumerckx: In that case, which of the two is preferred?
```
complex = operand_def(ComplexTypeConstr)
result = result_def(ComplexTypeConstr)
traits = traits_def(Pure(), SameOperandsAndResultType())
```
or getting rid of the trait?

alexarice: I would get rid of the trait personally

superlopuh: I would say get rid of the trait, and we might want to remove the trait definition altogether at some point.

xdsl/dialects/complex.py
jumerckx: That gives a Pylance error:
```
Argument of type "Literal[2]" cannot be assigned to parameter "length_constr" of type "IntConstraint" in function "of_length"
```

xdsl/dialects/complex.py
jumerckx: > Does this work?

yes, cool!

jumerckx: I would still keep `ComplexType.constr` for other cases in this file though?

xdsl/dialects/complex.py
superlopuh: We have to pass `EqIntConstraint` here

xdsl/dialects/complex.py
superlopuh: It depends on where you need it TBH

xdsl/dialects/complex.py
```diff
@@ -0,0 +1,381 @@
...
+@irdl_op_definition
+class Atan2Op(ComplexBinaryOp):
+    name = "complex.atan2"
+
+
+@irdl_op_definition
+class BitcastOp(IRDLOperation):
+    name = "complex.bitcast"
+    operand = operand_def(AnyAttr())
+    result = result_def(AnyAttr())
```
superlopuh:
```suggestion
    operand = operand_def()
    result = result_def()
```

xdsl/dialects/complex.py
jumerckx: I've gone with `.constr` everywhere for clarity.

xdsl/dialects/complex.py
jumerckx: I've renamed this to `ComplexUnaryComplexResultOperation` and `ComplexUnaryRealResultOperation`.

xdsl/dialects/complex.py
jumerckx: Here the array needs to hold attributes of the element type, not complex attributes themselves. I tried:
```py
    # In contrast to other operations, `complex.constant` can
    # have any complex result type, not just floating point:

    T : ClassVar = VarConstraint("T", AnyFloatConstr | base(IntegerType))
    value = prop_def(
        ArrayOfConstraint(RangeOf(T).of_length(EqIntConstraint(2)))
    )
    complex = result_def(ComplexType.constr(T))
```
but this gives:
```
attribute f64 expected from variable 'T', but got 1.000000e-01 : f64
```
I guess I'm constraining the value, not the type of the value here?

jumerckx:
Woops, I'd forgotten this pr. All feedback should've been addressed.
I had to change the `ArrayOfConstraint` to:
```py
class ConstantOp(IRDLOperation):
    name = "complex.constant"
    T: ClassVar = VarConstraint("T", AnyFloatConstr | base(IntegerType))
    value = prop_def(
        ArrayOfConstraint(
            RangeOf(
                AnyOf(
                    [
                        ParamAttrConstraint(IntegerAttr, (AnyAttr(), T)),
                        ParamAttrConstraint(FloatAttr, (AnyAttr(), T)),
                    ]
                )
            ).of_length(EqIntConstraint(2))
        )
    )
```
I *think* that does the right thing?

superlopuh (approved):

PR #5078, opened 2025-08-20
transformations: (canonicalize) handle multiple canonicalization patterns traits

superlopuh:
Pulls out one bit of functionality from #5030, allowing multiple traits that provide canonicalization patterns.

### DIFF
--- a/tests/transforms/test_canonicalization.py
+++ b/tests/transforms/test_canonicalization.py
@@ -1,0 +1,49 @@
+from xdsl.context import Context
+from xdsl.dialects.builtin import ModuleOp
+from xdsl.ir import Operation
+from xdsl.irdl import IRDLOperation, irdl_op_definition, traits_def
+from xdsl.pattern_rewriter import PatternRewriter, RewritePattern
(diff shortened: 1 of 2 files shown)


math-fehr:
Quick question, are you implementing it this way in order to support well inheritance/extensibility?
Is there a reason we couldn't do this by just extending the trait instead?

superlopuh:
I don't know of a way to extend the trait safely. Imagine an operation defined in xDSL with two dependencies adding canonicalization patterns to it. As far as each of them is concerned the op doesn't have a trait to add things to, so they each add their own trait, but their combination no longer works as expected.

math-fehr:
Yeah, but the idea is that adding a trait without checking if it is already there is quite bad.
Instead, they should just check if the trait is there, and add to the trait (or create a new trait)

superlopuh:
I'm confused, why is this preferable? For memory effects we can have multiple traits that declare multiple effects. Why not do the same for canonicalization patterns?

alexarice (approved):

PR #5081, opened 2025-08-20
core: make `operation._successor_uses` public

jumerckx:
context: https://github.com/xdslproject/xdsl/pull/4968#discussion_r2269571403

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -1157,12 +1157,12 @@ class Operation(_IRNode):
     def successors(self, new: Sequence[Block]):
         new = tuple(new)
         new_uses = tuple(Use(self, idx) for idx in range(len(new)))
-        for successor, use in zip(self._successors, self._successor_uses):
+        for successor, use in zip(self._successors, self.successor_uses):
             successor.remove_use(use)
         for successor, use in zip(new, new_uses):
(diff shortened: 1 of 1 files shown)


superlopuh (commented):
@math-fehr WDYT?

math-fehr (commented):
Can you make it a `@property` instead?
Just to avoid people modifying it

xdsl/ir/core.py
```diff
@@ -2147,7 +2151,7 @@ def __getitem__(self, idx: int | slice) -> Block | Sequence[Block]:
 
     def __setitem__(self, idx: int, successor: Block) -> None:
         successors = self._op._successors  # pyright: ignore[reportPrivateUsage]
-        successor_uses = self._op._successor_uses  # pyright: ignore[reportPrivateUsage]
+        successor_uses = self._op.successor_uses
```
jumerckx: Or should internals access `_successor_uses` directly?

superlopuh: In this case I'd go for private

superlopuh: Since it's likely to be at least a little perf sensitive

superlopuh (approved):

math-fehr (approved):

PR #5084, opened 2025-08-20
CI: try adjusting renovate settings to fix dependency updating

superlopuh:
According to this comment this might be the setting we want to get both lockfile updates on direct dependencies and weekly transitive dependency updates: https://github.com/renovatebot/renovate/discussions/19285


superlopuh:
I'll merge this now and we can fix things later if it doesn't work

PR #5085, opened 2025-08-20
transformations: (convert-memref-to-ptr) add get_offset_pointer helper

superlopuh:
No functional change.

### DIFF
--- a/xdsl/transforms/convert_memref_to_ptr.py
+++ b/xdsl/transforms/convert_memref_to_ptr.py
@@ -5,7 +5,7 @@ from typing import cast
 from xdsl.builder import Builder
 from xdsl.context import Context
 from xdsl.dialects import arith, builtin, func, memref, ptr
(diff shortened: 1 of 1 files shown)


xdsl/transforms/convert_memref_to_ptr.py
```diff
@@ -19,6 +19,39 @@
...
+    return bytes_offset.result
+
+
+def get_offset_pointer(
+    pointer: SSAValue,
+    bytes_offset: SSAValue,
+    builder: Builder,
+) -> SSAValue:
+    """
+    Returns the pointer scaled by an offset in bytes.
```
TaoBi22: Is scaled the right term? Maybe 
```suggestion
    Returns the provided pointer offset by a given number of bytes.
```

TaoBi22 (approved):
LGTM!

PR #5087, opened 2025-08-21
dialects: (emitc) use EmitCArrayElementType in ArrayType

Jimmy2027:
Was part of #5072, keeping it separate to investigate a type checker issue in test_emitc.py (ref [discussion](https://github.com/xdslproject/xdsl/pull/5072/commits/07a962200fb02b116dfd45779bac70f4c747790c#r2285734120))

### DIFF
--- a/xdsl/dialects/emitc.py
+++ b/xdsl/dialects/emitc.py
@@ -29,7 +29,6 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.ir import (
     Attribute,
-    AttributeCovT,
     Dialect,
     ParametrizedAttribute,
     SSAValue,
@@ -146,19 +145,19 @@ class EmitC_ArrayType(
     ParametrizedAttribute,
     TypeAttribute,
     ShapedType,
-    ContainerType[AttributeCovT],
+    ContainerType[EmitCArrayElementType],
 ):
     """EmitC array type"""
 
     name = "emitc.array"
 
     shape: ArrayAttr[IntAttr]
-    element_type: AttributeCovT
+    element_type: EmitCArrayElementType
 
     def __init__(
         self,
         shape: Iterable[int | IntAttr],
-        element_type: AttributeCovT,
+        element_type: EmitCArrayElementType,
     ):
         shape = ArrayAttr(
             [IntAttr(dim) if isinstance(dim, int) else dim for dim in shape]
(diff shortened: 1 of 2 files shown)


Jimmy2027:
Don't understand why this passes now, but I'll take it :)
`git diff 1fa610f017466fa033d68c85f9a92aeaaa9e3f94` shows nothing considerable in emitc.py, maybe c90c45408712864476f5bac129e6bb7c2421637e fixed something?

xdsl/dialects/emitc.py
```diff
@@ -146,19 +145,19 @@ class EmitC_ArrayType(
     ParametrizedAttribute,
     TypeAttribute,
     ShapedType,
-    ContainerType[AttributeCovT],
+    ContainerType[EmitCArrayElementType],
```
alexarice, line 148: This seems less powerful than before? Did you no longer want this to be a generic type. You could get the specialisation and keep the generic parameter by adding a new `TypeVar` with bound and default of `EmitCArrayElementType` and using that in place of `AttributeCovT`

superlopuh: Could also do this in a later PR IMO since it's not used at all rn.

superlopuh (approved):

xdsl/dialects/emitc.py
Jimmy2027: hmm what would be the benefit of that? `EmitC_ArrayType` only supports the elem_type EmitCArrayElementType so why not restrict it here?

alexarice: It would allow you to restrict to say an array of `EmitCFloatType`, though perhaps as sasha says this can be implemented if/when needed

Jimmy2027: and that would not work with the `ContainerType[EmitCArrayElementType]`? In theory a `EmitC_ArrayType[EmitCFloatType]` is also a `EmitC_ArrayType[EmitCArrayElementType]`

superlopuh: @Jimmy2027 It's your call basically, I think if we keep it generic then we should add a test that somehow exercises that, otherwise we can add it later.

superlopuh: Sorry GH UI didn't show your comments when I opened the tab

superlopuh: To expand a bit it, subclassing from `Generic[EmitCArrayElementTypeCovT]` would let you express `EmitC_ArrayType[Float32Type]` for things that only accepts this subclass. By using a type variable in your class definition I think you implicitly defined a generic class, but I would advise doing explicitly or not at all.

Jimmy2027: interesting, will try to open a PR on top of this then. I think this can go in as is though

Jimmy2027: does this make sense?
https://github.com/xdslproject/xdsl/pull/5089/commits/68e40db1925221688b51c1a240bd1fdb152244dc

PR #5088, opened 2025-08-21
API: add `ModulePass.apply_to_clone` helper

superlopuh:
This bit of sugar helps in all contexts that need to track changes via assignment as opposed to deep mutation, such as marimo, gui, and exhaustive exploration of rewrites.

### DIFF
--- a/xdsl/passes.py
+++ b/xdsl/passes.py
@@ -64,6 +64,18 @@ class ModulePass(ABC):
     @abstractmethod
     def apply(self, ctx: Context, op: builtin.ModuleOp) -> None: ...
 
+    def apply_to_clone(
+        self, ctx: Context, op: builtin.ModuleOp
+    ) -> tuple[Context, builtin.ModuleOp]:
+        """
+        Creates deep copies of the inputs, and returns the result of calling `apply` on
+        them.
+        """
+        ctx = ctx.clone()
+        op = op.clone()
+        self.apply(ctx, op)
+        return ctx, op
+
     @classmethod
     def from_pass_spec(cls, spec: PipelinePassSpec) -> Self:
         """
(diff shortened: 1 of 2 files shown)


xdsl/passes.py
```diff
@@ -64,6 +64,18 @@ class ModulePass(ABC):
...
     def apply(self, ctx: Context, op: builtin.ModuleOp) -> None: ...
 
+    def apply_to_clone(
+        self, ctx: Context, op: builtin.ModuleOp
+    ) -> tuple[Context, builtin.ModuleOp]:
+        """
+        Creates deep copies of the inputs, and returns the result of calling `apply` on
+        them.
+        """
+        ctx = ctx.clone()
```
alexarice, line 74: why is the context clone needed?

superlopuh: Technically the pass could modify it in some observable way, this guarantees that the inputs are not changed. In practice I don't think there are any scenarios today that would be affected, but it feels like a safe default. In MLIR I think dialects can be registered and unregistered by a pass, and the verification makes sure that all the ops are gone after a rewrite, for example.

xdsl/passes.py
```diff
@@ -64,6 +64,18 @@ class ModulePass(ABC):
     @abstractmethod
     def apply(self, ctx: Context, op: builtin.ModuleOp) -> None: ...
 
+    def apply_to_clone(
+        self, ctx: Context, op: builtin.ModuleOp
+    ) -> tuple[Context, builtin.ModuleOp]:
+        """
+        Creates deep copies of the inputs, and returns the result of calling `apply` on
+        them.
```
math-fehr:
```suggestion
        Creates deep copies of the module and the context, and returns
        the result of calling `apply` on them.
```
Just to be more explicit about the "inputs" part

PR #5090, opened 2025-08-21
core: add `CanonicalizationPatternsTrait`

superlopuh:
The idea is to replace the `HasCanonicalizationPatternsTrait`, together with the upcoming interface as defined in #5030.


superlopuh:
Actually let me propose another change first I think this order is wrong

PR #5096, opened 2025-08-22
core: allow Literal enum cases in irdl_to_attr_constraint

superlopuh:
Turns out that you need the TypeAlias annotation for Pyright to reason about the type as a type specifically. When I added the annotation it started complaining that the generic arguments are not `Literal`, which we actually didn't support for enum cases. This PR adds support for it and updates Builtin's `I_` types and values.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -728,26 +728,24 @@ class IntegerType(
         return f[format_index]
 
 
-i64 = IntegerType[64, Signedness.SIGNLESS](64)
-i32 = IntegerType[32, Signedness.SIGNLESS](32)
-i16 = IntegerType[16, Signedness.SIGNLESS](16)
-i8 = IntegerType[8, Signedness.SIGNLESS](8)
-i1 = IntegerType[1, Signedness.SIGNLESS](1)
-I64 = IntegerType[64, Signedness.SIGNLESS]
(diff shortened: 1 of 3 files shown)


xdsl/dialects/builtin.py
```diff
@@ -728,27 +728,25 @@ def format(self) -> str:
...
-i32 = IntegerType[32, Signedness.SIGNLESS](32)
-i16 = IntegerType[16, Signedness.SIGNLESS](16)
-i8 = IntegerType[8, Signedness.SIGNLESS](8)
-i1 = IntegerType[1, Signedness.SIGNLESS](1)
-I64 = IntegerType[64, Signedness.SIGNLESS]
-I32 = IntegerType[32, Signedness.SIGNLESS]
-I16 = IntegerType[16, Signedness.SIGNLESS]
-I8 = IntegerType[8, Signedness.SIGNLESS]
-I1 = IntegerType[1, Signedness.SIGNLESS]
+I64: TypeAlias = IntegerType[Literal[64], Literal[Signedness.SIGNLESS]]
```
alexarice, line 731: why change these?

alexarice (approved):

xdsl/dialects/builtin.py
superlopuh: Pyright started complaining when I added `: TypeAlias`

PR #5098, opened 2025-08-22
dialects: (builtin) Add StaticShapeConstraint

Jimmy2027:


xdsl/dialects/builtin.py
```diff
@@ -130,6 +130,30 @@ def strides_for_shape(shape: Sequence[int], factor: int = 1) -> tuple[int, ...]:
...
+                f"expected static shape, but got '{attr.get_shape()}'"
+            )
+
+    def mapping_type_vars(
+        self, type_var_mapping: Mapping[TypeVar, AttrConstraint | IntConstraint]
+    ) -> StaticShapeConstraint:
+        return self
+
+    def get_bases(self) -> set[type[Attribute]] | None:
+        return {ShapedType}
```
alexarice: This should only contain `final` superclasses

Jimmy2027: yup sry, thats why I marked it as draft, didn't have time to update it yet

Jimmy2027: returning None here should be correct no?

Jimmy2027:
> This seems reasonable to me, though I do wonder if this should be a constraint on an `ArrayAttr[IntAttr]` directly rather than a constraint on shaped types.

yeah not sure how useful this is.. #5026 uses it like this:
https://github.com/xdslproject/xdsl/pull/5026/files#diff-a6fc816eccf354215a90db862d5edae1ba9491317599aac0e26621e61ce39660R173

superlopuh (commented):
Can you please add a test?

superlopuh:
Actually another way of doing this would be to use a `MessageConstraint(RangeOf(AtLeast(0)))` constraint or something like it. The best thing to do is probably to add the tests first with this custom constraint, and then try to make them pass by replacing the constraint class with a helper funtion.

Jimmy2027:
> Actually another way of doing this would be to use a `MessageConstraint(RangeOf(AtLeast(0)))`

That would remove the `DYNAMIC_INDEX` variable though. I guess this would be a usecase for an `NotEqual` constraint :)

Jimmy2027:
added a NotEqual constraint in [70d1fc4](https://github.com/xdslproject/xdsl/pull/5098/commits/70d1fc4cbe31f077fff7414d7867f093337cccfe) but I think it might be more useful in general to have a Not constraint here, as you had already proposed in the past. Will experiment a bit with that

xdsl/irdl/constraints.py
```diff
@@ -765,6 +765,46 @@ def mapping_type_vars(
...
+
+        if constraint_satisfied:
+            raise VerifyException(
+                f"expected attribute to not satisfy {self.constraint}, but {attr} does"
+            )
+
+    def variables(self) -> set[str]:
+        return self.constraint.variables()
+
+    def get_bases(self) -> set[type[Attribute]] | None:
```
alexarice: This isn't correct, in fact these are the only bases we know the constraint doesn't verify

xdsl/irdl/constraints.py
```diff
@@ -765,6 +765,46 @@ def mapping_type_vars(
...
+        return self.constraint.get_bases()
+
+    def can_infer(self, var_constraint_names: AbstractSet[str]) -> bool:
+        return False
```
alexarice: This should be the default

Jimmy2027:
> I'm unsure about the not constraint, to do it properly requires backtracking and I don't know if this is something we really want to support. At the very least we should enforce that no variables appear in a not constraint if we do want to go ahead with this.

yeah wanted to experiment with it, I'm also happy to keep the original constraint class in  [dbf1977](https://github.com/xdslproject/xdsl/pull/5098/commits/dbf19773f2b5766daa964753a1cb19315abd5a7f) and move the conversation about the `Not` constrain somewhere else

superlopuh:
Actually I think we can safely add a NotIntConstraint without backtracking, might need a better name though. Maybe we want to special case it with a `StaticShapeConstraint` instead?

Jimmy2027:
I still think a Not constraint with some limitations would be useful, proposed something in #5167 
While the discussion is going on there, should I maybe move the static shape constraint as initially proposed here to `emitc.py` as its the only user? Then maybe we can get  #5026 in as well and have less open threads

superlopuh:
In the short term, I think the way forward is a `StaticDimensionConstraint`, and we can discuss Not separately. I would probably also discuss it on Zulip instead of GH as there are quite a lot of things to consider there, and it's probably a better place to have a deeper conversation.

Jimmy2027:
> In the short term, I think the way forward is a `StaticDimensionConstraint`

True, that makes sense

xdsl/irdl/constraints.py
```diff
@@ -919,6 +919,23 @@ def mapping_type_vars(
         return self
 
 
+@dataclass(frozen=True)
+class NotEqual(IntConstraint):
```
Jimmy2027: NotEqualIntConstraint

PR #5060, opened 2025-08-18
[WIP] documentation: add new notebook leveraging mkdocs-marimo 

superlopuh:
Run with `make docs-serve-fast`

PR #5102, opened 2025-08-23
documentation: enable islands to use local wheel

tobiasgrosser:
This PR adds a test notebook to add support for the local xdsl wheel in Marimo Islands. We deliberately make this a marimo-island file, as this file is meant for marimo debugging.

We also exclude our test notebook from the python tests as these otherwise trigger `SyntaxError: 'await' outside async function` errors. 

The code is taken from https://github.com/xdslproject/xdsl/pull/5060.

### DIFF
--- a/docs/marimo/expressions.py
+++ b/docs/marimo/expressions.py
@@ -20,7 +20,6 @@ async def _():
             url = re.sub('([^/])/([a-f0-9-]+)', '\\1/', url, count=1)
             buildnumber = re.sub('.*--([0-9+]+).*', '\\1', url, count=1)
             if buildnumber != url:
-                print(buildnumber)
                 url = url + buildnumber + "/"
 
             return url


Makefile
```diff
@@ -108,7 +112,7 @@ tests-marimo: uv-installed
 
 # run all tests
 .PHONY: tests-functional
-tests-functional: pytest tests-toy filecheck tests-marimo
+tests-functional: pytest tests-toy filecheck tests-marimo
```
tobiasgrosser:
```suggestion
tests-functional: pytest tests-toy filecheck tests-marimo
```

docs/marimo/expressions.py
```diff
@@ -7,6 +7,26 @@
 @app.cell
 def _():
     import marimo as mo
+
+    # Use the locally built xDSL version
+    import micropip
+    import asyncio
+
+    def get_url():
```
tobiasgrosser:
```suggestion
    # Get the current notebook URL, drop the 'blob' URL components that seem to be added,
    # and add the buildnumber that a makethedocs PR build seems to add. This allows to load
    # the wheel both locally and when deployed to makethedocs. 
    def get_url():
```

docs/marimo/expressions.py
```diff
@@ -7,6 +7,28 @@
 @app.cell
 def _():
     import marimo as mo
+
+    # Use the locally built xDSL version
+    import micropip
+    import asyncio
```
tobiasgrosser:
```suggestion

```

tobiasgrosser:
This is now ready-to-review. @superlopuh @Moxinilian, a review would be appreciated.

docs/marimo/expressions.py
```diff
@@ -7,6 +7,27 @@
...
+        if buildnumber != url:
+            print(buildnumber)
+            url = url + buildnumber + "/"
+
+        return url
+
+    await micropip.install("xdsl @ " + get_url() + "/xdsl-0.0.0-py3-none-any.whl")
+
+    from xdsl.printer import Printer
+
```
superlopuh: Instead of conditionally importing and ignoring for tests, could you please test for whether this is running in pyodide with
``` py
import sys

if sys.platform == 'emscripten':
    ...
```
That way the notebook will work both with local development and when running online

tobiasgrosser: I tried, but this did not work. AFAIU this is a syntax-error which is checked for statically, not dynamically.

tobiasgrosser: Obviously, any ideas how to keep our tests working would be much appreciated.

superlopuh:
Did you get it to work locally? I'm happy to merge this if it works

docs/marimo/expressions.py
tobiasgrosser: I am now trying to add `async` before the function.

superlopuh: I'll have a bit of time to look into it tomorrow morning if that's OK, I'd love for the notebooks to keep being tested.

tobiasgrosser:
OK, it seems this has all now been resolved. I just did not look carefully enough before. I merge this for now, but a post-merge review and intensive testing would be appreciated.

docs/marimo/expressions.py
```diff
@@ -5,8 +5,30 @@
 
 
 @app.cell
-def _():
+async def _():
     import marimo as mo
+
+    # Use the locally built xDSL wheel when running in Marimo
+    if mo.running_in_notebook():
```
superlopuh, line 12: This doesn't avoid this in the logs, and the rendering being off when the page first opens
```
xdsl @ s/sasha/Developer/xdslproject/xdsl/.venv/in/xdsl-0.0.0-py3-none-any.whl was not installed: micropip is only available in WASM notebooks.INFO    -  [marimo] on_page_markdown /reference/
```

tobiasgrosser: Proposed fix in https://github.com/xdslproject/xdsl/pull/5105.

PR #5103, opened 2025-08-23
documentation: add expression parser

tobiasgrosser:
Incorporate Theo's expression parser from https://github.com/xdslproject/list-lang/ and make it available at 
https://xdsl.readthedocs.io/latest/marimo/expressions/.

This PR can be tested at: https://xdsl--5103.org.readthedocs.build/5103/marimo/expressions/.

### DIFF
--- a/xdsl/frontend/listlang/main.py
+++ b/xdsl/frontend/listlang/main.py
@@ -74,11 +74,9 @@ class Located(Generic[T]):  # noqa: UP046
 
 
 class ListLangType:
-    def __str__(self) -> str:
-        ...
+    def __str__(self) -> str: ...
 
-    def xdsl(self) -> Attribute:
-        ...
+    def xdsl(self) -> Attribute: ...
 
 
 @dataclass
@@ -225,8 +223,7 @@ def _parse_opt_expr_atom(
         if not isinstance(cond.value.typ, ListLangBool):
             raise ParseError(
                 cond.loc.pos,
-                f"expected {ListLangBool()} type for condition, "
-                f"got {cond.value.typ}",
+                f"expected {ListLangBool()} type for condition, got {cond.value.typ}",
             )
 
         then_block = Block()
@@ -311,8 +308,7 @@ class BinaryOp:
         builder: Builder,
         lhs: Located[TypedExpression],
         rhs: Located[TypedExpression],
-    ) -> TypedExpression:
-        ...
+    ) -> TypedExpression: ...
 
 
 class Multiplication(BinaryOp):
@@ -409,8 +405,7 @@ class Comparator(BinaryOp):
             arith.CmpiOp(lhs.value.value, rhs.value.value, self.arith_code)
         )
         cmpi_op.result.name_hint = (
-            f"{lhs.value.value.name_hint}_"
-            f"{self.arith_code}_{rhs.value.value.name_hint}"
+            f"{lhs.value.value.name_hint}_{self.arith_code}_{rhs.value.value.name_hint}"
         )
         return TypedExpression(cmpi_op.result, ListLangBool())
 


tobiasgrosser:
@superlopuh @Moxinilian, the expression parser works now on the website as expected and with relatively low latency: https://xdsl--5103.org.readthedocs.build/5103/marimo/expressions/.

Open issues:
- `list-lang` uses Python 3.12. Things break for 3.11 and below due to the use of too-fancy type annotations.
- We did not factor the `mo.state` trick into some more self-contained object.
- This currently copies @Moxinilian's code into the xdsl repo. I am unsure how to resolve the deployment issue. I can see the following options.
  1) depend on`list-lang` repo
  2) pull list-lang in only during CI
  3) create a full separate marimo setup in the `list-lang` repo
  4) keep copying `list-lang` into this repo
  5) move list-lange development to xdsl repo
  
I would appreciate your comments on which of the options is preferable.

pyproject.toml
```diff
@@ -113,12 +113,12 @@ reportUnnecessaryTypeIgnoreComment = true
 typeCheckingMode = "strict"
 extraPaths = ["tests"]
 "include" = ["docs", "xdsl", "tests", "benchmarks"]
-"exclude" = ["docs/marimo"]
+"exclude" = ["docs/marimo", "xdsl/frontend/expression/main.py"]
```
superlopuh: Why exclude it?

tobiasgrosser: Because the current formatting that @Moxinilian has chosen is different to what we have. I guess I can reformat when copying.

Moxinilian:
Could you expand on the challenges of just depending on list-lang as a regular Python package?

Moxinilian:
By the way, why are you still supporting Python 3.11? It seems even Ubuntu has moved on by now.

tobiasgrosser:
> By the way, why are you still supporting Python 3.11? It seems even Ubuntu has moved on by now.

Because there are HPC systems that are not running on Ubuntu and are updated a bit slower. When we last checked two months ago, Archer2 was still running Python 3.10.

tobiasgrosser:
> Could you expand on the challenges of just depending on list-lang as a regular Python package?

I just tried and it did not work:

> pip install list-lang
> ERROR: Could not find a version that satisfies the requirement list-lang (from versions: none)

I guess we would need to ensure that list-lang has pypi releases and then we would need to sync them in lock-step with us evolving the documentation. This seems quite some overhead. In theory we could push a lot of things into independent repositories. In practice, managing a large collection of repositories has quite some maintenance cost. Why exactly do you feel it is a good idea to have this parser in its own repository?

Moxinilian:
What did you just try? I'm trying to understand the challenges involved. Did you try to depend on it as a git dependency?

Moxinilian:
The reason I think it is good outside of the xDSL repository is because this is the sort of reusable pieces that do not need to have a hard connection to the xDSL repository. In the same way that the hardware compiler is perfectly fine in its own repository, this also is, to me. That way it's easy to depend on for other people without needing to hook into internal and arguably unrelated details of xDSL.

tobiasgrosser:
I listed in an earlier comment the following options I see:

> -  This currently copies @Moxinilian's code into the xdsl repo. I am unsure how to resolve the deployment issue. I can see the following options.
>    1) depend onlist-lang repo
>    2) pull list-lang in only during CI
>    3) create a full separate marimo setup in the list-lang repo
>    4) keep copying list-lang into this repo
>    5) move list-lange development to xdsl repo

> What did you just try? I'm trying to understand the challenges involved. Did you try to depend on it as a git dependency?

I tried option `1)`, meaning I micropip install the `list-lang` package in the notebook. The first step for this option was to check if there exists a `list-lang` package. I run `pip install list-lang` which reported that such a package does not yet exist.
(I am also unsure we want to create such a package).

> Did you try to depend on it as a git dependency?

Are you suggesting I make list-lang a git subrepository of the xdsl repository? Or rather, for me to add a script to our CI setup that has a hardcoded hash and then pulls the list-lang changes in?

> The reason I think it is good outside of the xDSL repository is because this is the sort of reusable pieces that do not need to have a hard connection to the xDSL repository. In the same way that the hardware compiler is perfectly fine in its own repository, this also is, to me. That way it's easy to depend on for other people without needing to hook into internal and arguably unrelated details of xDSL.

I think having the hardware compiler as a separate repository is a great idea. In fact, I would love to have more separate repositories. However, doing so is not super trivial and I am unsure that trying to figure this out before the MLIR (Un)School is the best use of all of our time.

> That way it's easy to depend on for other people without needing to hook into internal and arguably unrelated details of xDSL.

In which way would people `need to hook into internal details of xdsl` if the file lives in `xdsl.frontend.expressions` or `xdsl.frontend.list-lang`? I would be happy to develop things in the xdsl repo for the next two weeks and then have a separate conversation how we can pull things after we have a working implementation.

Moxinilian:
Feel free to pick the option you prefer. I was suggesting to depend on list-lang for the notebooks by using a git dependency (`list-lang @ git+https://github.com/xdslproject/list-lang.git` in pip install or pyproject), and eventually publish it as a PyPi package if this becomes release-worthy.

superlopuh:
What if we change the APIs of builder (we're planning to do this soon)? Suddenly we have lock-step dependencies and a bad time. I would recommend making the parser as small as possible, have it clearly documented as documentation-only, and to iterate on it in xDSL. We review things quickly enough for this not to be too big an overhead, and it's a massive headache reduction in terms of dependency management.

tobiasgrosser:
I fixed the formatting and proposed a patch to `list-lang` to support Python down to 3.10 (the change is easy). I'll merge this for now, such that we can move ahead with the notebook. I agree with @superlopuh that the extra-repo has quite some cost at the moment. Happy to discuss this further.

PR #5111, opened 2025-08-27
core: (AnyOf) relax the disjointness condition

alexarice:
This was one of the options for making `any_of` a bit more permissive. The downside of such an approach is that it gives you the opportunity to shoot yourself in the foot with interfaces (though tbh I can't think of a good example).

On the other hand it's likely to alleviate some of the pain points people have had with the current implementation.

### DIFF
--- a/tests/irdl/test_attr_constraint.py
+++ b/tests/irdl/test_attr_constraint.py
@@ -81,23 +81,27 @@ class AttrD(Base):
     param: AttrA | AttrC
 
 
+class AttrE(ParametrizedAttribute):
+    name = "test.attr_e"
+
+
 @pytest.mark.parametrize(
     "constraint, expected",
     [
         (AnyAttr(), None),
         (EqAttrConstraint(AttrB(AttrA())), {AttrB}),
-        (BaseAttr(Base), None),
+        (BaseAttr(Base), {Base}),
         (BaseAttr(AttrA), {AttrA}),
         (EqAttrConstraint(AttrB(AttrA())) | BaseAttr(AttrA), {AttrA, AttrB}),
         (
             EqAttrConstraint(AttrD(AttrA())) | EqAttrConstraint(AttrD(AttrC())),
             {AttrD},
         ),
-        (AllOf((AnyAttr(), BaseAttr(Base))), None),
+        (AllOf((AnyAttr(), BaseAttr(Base))), {Base}),
         (AllOf((AnyAttr(), BaseAttr(AttrA))), {AttrA}),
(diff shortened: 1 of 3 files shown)


math-fehr (approved):
My understanding is that we should not really have an operation inheriting from two different parent classes, so to me that should just work?

alexarice:
I'm sure there's ways to break it, but it feels a bit like you have to be trying

math-fehr:
Actually, one thing that could happen is:
```
class HasTwoResults(ABC, Operation): ...
class HasOneArgument(ABC, Operation): ...
class MyOp(HasTwoResults, HasOneArgument): ...
```
And then having an `AnyOf(HasTwoResults, HasOneArgument)`
I feel like this *could* happen, but at the same time this change would make this much easier for some AnyOf machinery?

Actually, what's the use case of this PR?

alexarice:
1. I think it should resolve things like https://github.com/xdslproject/xdsl/pull/5025
2. I hit this problem when thinking about trying to make `Attribute` a union, as I ended up with `TypeAttribute` as also a union of two superclasses, which broke any_of

alexarice:
I guess the other option is to have this behaviour behind a "trust_me" flag

math-fehr:
Oh I see. I guess in the most common case, you would only have one of the `ABC` classes, so that we can actually check.
Otherwise, I would not try to split Attribute into the both subclasses, I feel that this would break quite some stuff (like inheritance, as you saw). Also, not sure what it would give?

alexarice:
Yeah, I feel the Attribute split is a different discussion (I'm also not sure it gives you that much)

superlopuh:
I stil feel a bit off about this, I'd love to have a bit of time to play with this if that's OK with you. I think limiting to one abstract superclass is probably best.

superlopuh:
```python
class OtherBase(ParametrizedAttribute):
    name = "other_base"


def test_ok():
    constr = irdl_to_attr_constraint(Base | OtherBase)

    assert isinstance(AttrA(), Base)
    assert isinstance(AttrA(), Base | OtherBase)
    assert isa(AttrA(), Base | OtherBase)
    assert constr.verifies(AttrA())
```

This fails on the last line.

alexarice:
What's `AttrA` here? Does it derive from Base?

superlopuh:
Yep

superlopuh:
Sorry this is in tests/irdl/test_attr_constraint.py, I added it locally to play with it

alexarice:
You're right, this approach doesn't work, lets go with the other one.

superlopuh:
I accidentally deleted the remote instead of my local clone, am I right in understanding that we don't need this any more?

alexarice:
I assume not

PR #5117, opened 2025-08-29
dialects: (pdl_interp) Fix some small type errors.

jumerckx:
Small fixes I had to make while porting `pdl-to-pdl-interp`

### DIFF
--- a/xdsl/dialects/pdl_interp.py
+++ b/xdsl/dialects/pdl_interp.py
@@ -496,7 +496,7 @@ class RecordMatchOp(IRDLOperation):
         self,
         rewriter: str | SymbolRefAttr,
         root_kind: str | StringAttr | None,
-        generated_ops: list[StringAttr] | None,
+        generated_ops: ArrayAttr[StringAttr] | None,
         benefit: int | IntegerAttr[I16],
         inputs: Sequence[SSAValue],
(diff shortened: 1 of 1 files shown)


xdsl/dialects/pdl_interp.py
```diff
@@ -495,8 +495,8 @@ class RecordMatchOp(IRDLOperation):
     def __init__(
         self,
         rewriter: str | SymbolRefAttr,
-        root_kind: str | StringAttr,
-        generated_ops: list[OperationType] | None,
+        root_kind: str | StringAttr | None,
+        generated_ops: list[StringAttr] | None,
```
alexarice: Could we drive-by fix the TODO below? From what I can see this attribute is meant to be optional.

jumerckx: Yes, thanks!

superlopuh: As a nit, I'd rather take an optional `ArrayAttr[StringAttr]` here, it's low overhead for the caller, and simplifies this init. If not `ArrayAttr` then Sequence would also be preferable to `list`

jumerckx: Right, will do

PR #5122, opened 2025-09-01
docs: Use slider to select passes

math-fehr:
Adds an example to the expressions tutorial where we can use a slider to explore the effect of different pass pipelines on a module.


tobiasgrosser:
90% of this has now been upstreamed. Let's close this and open new PRs for the remaining passes for the documentation.

PR #5133, opened 2025-09-03
interpreter: (arith, tensor) add support for more features

Moxinilian:


xdsl/interpreters/arith.py
```diff
@@ -157,3 +169,10 @@ def run_floordivsi(
         (lhs, rhs) = args
         assert rhs != 0
         return (lhs // rhs,)
+
+    @impl(arith.IndexCastOp)
+    def run_indexcast(
+        self, interpreter: Interpreter, op: arith.IndexCastOp, args: PythonValues
+    ):
+        assert len(args) == 1
+        return args
```
math-fehr: Can you either handle the case when we cast to smaller bit-widths, or raise an exception saying you don't handle it?

xdsl/interpreters/tensor.py
```diff
@@ -14,16 +17,25 @@
...
+
 
 @register_impls
 class TensorFunctions(InterpreterFunctions):
     @impl(tensor.EmptyOp)
     def run_empty(
-        self, interpreter: Interpreter, op: tensor.EmptyOp, args: tuple[Any, ...]
-    ) -> tuple[Any, ...]:
+        self, interpreter: Interpreter, op: tensor.EmptyOp, args: tuple[int, ...]
+    ) -> tuple[ShapedArray[Any]]:
```
math-fehr, line 28: For this type, and the following one, I have the suspicion that @superlopuh didn't add them because we don't actually check them.
So when we call this interpretation function, if we actually write a wrong type, we will have a silent or hard to track bug because of it.
That's why we have all these `assert isinstance(...)` everywhere.
I can't know for sure, only @superlopuh knows that AFAIK.

xdsl/interpreters/tensor.py
```diff
@@ -47,3 +61,51 @@ def run_reshape(
             raise InterpretationError("Mismatch between static shape and new shape")
         result = ShapedArray(input.data_ptr, static_shape)
         return (result,)
+
+    @impl(tensor.InsertOp)
+    def run_insert(
+        self,
+        interpreter: Interpreter,
+        op: tensor.InsertOp,
+        args: tuple[T | ShapedArray[T] | int, ...],
```
math-fehr, line 70: Btw, this can be written that way: `tuple[T, ShapedArray[T], *tuple[int, ...]]`

xdsl/interpreters/tensor.py
Moxinilian: Well in any case the existing implementation does not fully check for the types either. Are you sure this is what you want?

math-fehr: As Sasha is not there, I would like to stick with the style that was used.
Also, I feel that not fully checking the type is a bug/mistake in some sense. In any case, I would like to avoid introducing inconsistencies if all the code is written that way. When sasha is back I can check with him what we should do there.

Moxinilian: The reason I am not sticking with the style is that otherwise I get type errors. The hack that was used previously cannot be used anymore with the new operations (as the value type becomes relevant). If this does not work for you, could you suggest specific code changes? I'm not sure what to do from there.

Moxinilian: This was resolved offline.

xdsl/interpreters/tensor.py
Moxinilian: As discussed offline this is not allowed in Python 3.10 :(

math-fehr (approved):
Good to me now!

xdsl/interpreters/tensor.py
superlopuh: I'm back

superlopuh: Not sure what hack you meant here. Do you mean the use of `Any`?

math-fehr: I think he meant the place you use `cast` to ensure the array had `float` inside

superlopuh: My thinking there was that an explicit cast makes it clear that we're in YOLO mode, typed functions are kind of a lie, there's no type system checks that the usage is correct

superlopuh: I'd like a system that uses irdl and some kind of context for mapping of types to runtime values but until then I think all interpreters should be tuples of Any

PR #5138, opened 2025-09-04
dialects: (builtin) Add default `AnyAttr` for `element_type` in `TensorType.constr`

mudit2812:
#5061 added a shape constraint to `TensorType.constr`, but the default constraint for the element type was left incomplete. The default `element_type` is `None`, but this should be normalized to `AnyAttr()`. This PR just makes that change.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -1466,9 +1466,10 @@ class TensorType(
     ) -> AttrConstraint[TensorType[AttributeInvT]]:
         if element_type is None and shape is None:
             return BaseAttr[TensorType[AttributeInvT]](TensorType)
+        element_type_constr = AnyAttr() if element_type is None else element_type
         shape_constr = AnyAttr() if shape is None else shape
         return ParamAttrConstraint[TensorType[AttributeInvT]](
-            TensorType, (shape_constr, element_type, AnyAttr())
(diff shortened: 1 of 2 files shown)


xdsl/dialects/builtin.py
```diff
@@ -1466,9 +1466,10 @@ def constr(
     ) -> AttrConstraint[TensorType[AttributeInvT]]:
         if element_type is None and shape is None:
             return BaseAttr[TensorType[AttributeInvT]](TensorType)
+        element_type_constr = AnyAttr() if element_type is None else element_type
         shape_constr = AnyAttr() if shape is None else shape
         return ParamAttrConstraint[TensorType[AttributeInvT]](
-            TensorType, (shape_constr, element_type, AnyAttr())
+            TensorType, (shape_constr, element_type_constr, AnyAttr())
```
superlopuh: I'm surprised this is needed, what happens if you run the test with this change reverted?

mudit2812: Actually, it just works. I guess this change isn't needed. Should I close the PR?

superlopuh: Let's merge the test? It'll be a kind of documentation for the constraint also.

mudit2812: Reverted the change to `TensorType.constr` but kept the test change. Updated the PR title and description accordingly 👍

PR #5141, opened 2025-09-04
interpreter: add support for i1 tensors

Moxinilian:


Moxinilian:
It's because struct packing does not support i1

xdsl/interpreters/builtin.py
```diff
@@ -27,6 +27,8 @@ def xtype_for_el_type(
     el_type: Attribute, index_bitwidth: Literal[32, 64]
 ) -> PackableType[Any]:
     match el_type:
+        case builtin.i1:
+            return ptr.int8
```
superlopuh, line 31: I just replaced this with `builtin.i1` locally and it works fine?

Moxinilian: It didn't work fine on my machine. Feel free to update it.

PR #5144, opened 2025-09-04
Add reset button, and fix printer

math-fehr:


superlopuh:
It would be good to keep to our PR format with the `label: ` prefix for consistency.

PR #5150, opened 2025-09-05
docs: Some improvement on the marimo SSA section

math-fehr:


superlopuh (approved):

superlopuh:
Why the `_` prefix?

math-fehr:
It's just a prefix that I think makes sense for temporary values? And that's allowed in MLIR

PR #5157, opened 2025-09-06
documentation: disable marimo docs export

superlopuh:
This unused and broken path has been slowing me down recently when trying to add features to the worse-but-working readonly marimo export, I propose adding the md export back in once the features work.


superlopuh:
Discussed with @math-fehr who agrees this is the way to go, let's try to add it back after the workshop.

PR #5158, opened 2025-09-06
documentation: substitute xDSL wheel import when building marimo

superlopuh:
Marimo has a pytest integration that breaks in the presence of async cells. In order to make the notebooks with pytest cells possible to run on the documentation website, we need to have all cells be non-`async`, which is not possible with micropip. This PR adds the async import only when building the notebook.

Currently requires that all notebooks contain exactly the following cell, which will be replaced with an async cell:

```python
def _():
    from xdsl.utils import marimo as xmo
    return xmo
```

### DIFF
--- a/docs/marimo/expressions.py
+++ b/docs/marimo/expressions.py
@@ -4,49 +4,15 @@ __generated_with = "0.15.0"
 app = marimo.App(width="medium")
 
 
-@app.cell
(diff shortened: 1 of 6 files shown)


tobiasgrosser (approved):
LGTM. Thank you, @superlopuh!

PR #5167, opened 2025-09-07
[WIP] core: Add a Not Constraint

Jimmy2027:
wip to discuss the addition of a Not Constraint,
discussion started in #5098

### DIFF
--- a/xdsl/irdl/constraints.py
+++ b/xdsl/irdl/constraints.py
@@ -170,6 +170,10 @@ class AttrConstraint(Generic[AttributeCovT], ABC):
             return self
         return AllOf((self, value))  # pyright: ignore[reportReturnType]
 
+    def __invert__(self) -> Not[AttributeCovT]:
+        """Return the negation of this constraint (syntactic sugar for Not(self))."""
+        return Not(self)
+
     @abstractmethod
     def mapping_type_vars(
         self, type_var_mapping: Mapping[TypeVar, AttrConstraint | IntConstraint]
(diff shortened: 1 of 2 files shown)


superlopuh:
Can you please make the CI pass? Then would be nice to chat on the Zulip about this to have a bit more of a high-bandwidth conversation. It seems like the general approach should work but would be good to have @math-fehr and @alexarice's takes.

math-fehr (approved):
That looks good to me! I think we need that constraint, even though that can easily become a footgun if we are not careful enough

xdsl/irdl/constraints.py
```diff
@@ -765,6 +768,59 @@ def mapping_type_vars(
...
+    - Does not provide base information and therefore cannot be used inside
+      `AnyOf` constraints.
+    """
+
+    constraint: AttrConstraint[AttributeCovT]
+    """The constraint to negate."""
+
+    def __post_init__(self) -> None:
+        """Validate that the inner constraint doesn't contain variables."""
+        if self.constraint.variables():
```
alexarice, line 788: As mentioned in the xdsl meeting. I don't believe this ensures that the constraint doesn't mention any variables, just that it doesn't guarantee solving variables.

superlopuh:
Would be good to get @alexarice's tick before merging. It would be nice to have a way to avoid the backtracking. This seems related to the anyof PR, a bit of functionality we're missing is a check for whether running the constraint will modify the context.

xdsl/irdl/constraints.py
alexarice: As an example `Not(Not(VarConstraint("x", AnyAttr()))` is allowed by this atm

alexarice (commented):
I'm generally hesitant about checks on constraints that ensure there are no variables, as I feel at least part of the motivation of this system was to eventually be fully compatible/upstream to irdl, where *every* constraint is a variable. I guess in irdl this could be replaced with some sort of check that the constraint only has a single use, but I still thought I'd voice my unease at this.

xdsl/irdl/constraints.py
Jimmy2027: the inner constraint `Not(VarConstraint("x", AnyAttr())` would throw an error no?
I'm getting this:
`xdsl.utils.exceptions.PyRDLError: Not constraint cannot contain variables, but inner constraint VarConstraint(name='x', constraint=AnyAttr()) has variables: {'x'}. `

alexarice: Sorry, it was a bad example: I think `Not(AnyOf(VarConstraint("x", base(IntegerAttr)), VarConstraint("y", base(FloatAttr))))` is a concrete example.

superlopuh:
I still think we might want to merge this at some point, but let's close it for now, until we have a better conviction of when this is safe to use.

PR #5169, opened 2025-09-07
documentation: check for arith interpretation output

superlopuh:
Checks the printed output instead of a value.


tobiasgrosser (approved):
LGTM.

PR #5215, opened 2025-09-11
core: Add a NotEqualIntConstraint

Jimmy2027:
Add a constraint that constrains an integer to not be equal to a given value.
Might be generalized to a general "Not" constraint, see https://github.com/xdslproject/xdsl/pull/5167

### DIFF
--- a/tests/dialects/test_builtin.py
+++ b/tests/dialects/test_builtin.py
@@ -978,5 +978,5 @@ def test_not_equal_int_constraint():
     constraint.verify(3, ConstraintContext())
 
     # Test with integer attribute equal to 5
-    with pytest.raises(VerifyException, match="expected integer != 5, got 5"):
+    with pytest.raises(VerifyException, match="expected integer != 5"):
         constraint.verify(5, ConstraintContext())


xdsl/irdl/constraints.py
```diff
@@ -858,6 +858,23 @@ def mapping_type_vars(
...
+@dataclass(frozen=True)
+class NotEqualIntConstraint(IntConstraint):
+    """Constrain an integer to not be equal to a given value."""
+
+    value: int
+    """The value the integer must not be equal to."""
+
+    def verify(self, i: int, constraint_context: ConstraintContext) -> None:
+        if i == self.value:
+            raise VerifyException(f"expected integer != {self.value}, got {i}")
```
math-fehr: This message is a bit redundant, because you are printing the same value twice (i and self.value are equal)
Maybe just remove "got i"?

Jimmy2027: good point :)

tests/dialects/test_builtin.py
```diff
@@ -965,3 +966,17 @@ def test_array_of_constraint():
...
+# NotEqualIntConstraint
+################################################################################
+def test_not_equal_int_constraint():
+    constraint = NotEqualIntConstraint(5)
+
+    # Test with integer attribute not equal to 5
+    constraint.verify(3, ConstraintContext())
+
+    # Test with integer attribute equal to 5
+    with pytest.raises(VerifyException, match="expected integer != 5, got 5"):
```
superlopuh:
```suggestion
    with pytest.raises(VerifyException, match="expected integer != 5"):
```

PR #5216, opened 2025-09-11
backend: add cost model interface

adutilleul:


xdsl/backend/asm_perf_reporter.py
```diff
@@ -0,0 +1,79 @@
...
+import subprocess
+from abc import ABC, abstractmethod
+from shutil import which
+from tempfile import NamedTemporaryFile
+
+from xdsl.backend.assembly_printer import AssemblyPrinter
+from xdsl.dialects.builtin import ModuleOp
+
+
+class AssemblyPerformanceReporter(ABC):
```
superlopuh: How do you feel about a design that just takes a single `Block` and returns a `float`? Since we're estimating the throughput of iteration, I assume?

xdsl/backend/asm_perf_reporter.py
```diff
@@ -0,0 +1,81 @@
...
+                    f"Block operation {op} should be an assembly instruction"
+                )
+                op.print_assembly(printer)
+
+        try:
+            report = self.produce_report()
+            return self.process_report(report)
+        finally:
+            if os.path.exists(self.src_path):
+                os.remove(self.src_path)
```
superlopuh: One day, I hope to implement a block throughput estimator in xDSL. When that day comes, most of this code won't be compatible, since it assumes that the estimation will be done via a command-line call. Can you please move it into the MCAReporter, and leave only `estimate_throughput` as an abstract function?

Can you please also add a doc string to the file, the class, and the function?

xdsl/backend/asm_perf_reporter.py
superlopuh, line 1: What do you think about calling this file `block_throughput_cost_model.py`? Then the abstract class below could be `BlockThroughputCostModel`, which seems like an intuitive name for what it does.

xdsl/backend/block_throughput_cost_model.py
```diff
@@ -0,0 +1,127 @@
...
+
+    arch: str
+    src_path: str
+
+    def __init__(self, arch: str):
+        self.arch = arch
+
+    @abstractmethod
+    def name(self) -> str:
+        pass
```
superlopuh: Let's move these to the `ExternalBlockThroughputCostModel`? Also, it would be nice to make it clear that `name` is actually the name of the command-line tool to call, maybe `bin_name`? Or `tool_name`? Not sure what's best but something more explicit would be nice.

tests/backend/test_asm_reporter.py
```diff
@@ -0,0 +1,23 @@
+import shutil
+
+import pytest
+
+from xdsl.backend.block_throughput_cost_model import MCABlockThroughputCostModel
+from xdsl.builder import Builder
+from xdsl.dialects import x86_func
+
+llvm_mca_available = shutil.which("llvm-mca") is not None
```
superlopuh: Might as well use this?
```suggestion
llvm_mca_available = MCABlockThroughputCostModel.is_installed()
```

xdsl/backend/block_throughput_cost_model.py
```diff
@@ -0,0 +1,127 @@
...
+    This class provides an interface for estimating the throughput of a block
+    of assembly-like operations for a specific microarchitecture.
+    """
+
+    arch: str
+    src_path: str
+
+    def __init__(self, arch: str):
+        self.arch = arch
+
```
superlopuh: Let's move this to the class below?

superlopuh:
Thank you!

superlopuh (approved):

xdsl/backend/block_throughput_cost_model.py
```diff
@@ -0,0 +1,127 @@
...
+        Args:
+            block: The block of operations to analyze.
+
+        Returns:
+            The estimated throughput as a float, or None if the estimation fails.
+        """
+        pass
+
+
+class ExternalBlockThroughputCostModel(BlockThroughputCostModel):
```
qaco, line 44: This design reminds me something ;) It's nice !

PR #5217, opened 2025-09-11
frontend: add support for bool/int/float literals

### LINKED ISSUE #5212: frontend: Add support for float and int literals in pyast frontend

AlexanderViand:
closes #5212

This adds support for bool/int/float literals to the PyAst -> IR translation. Currently, only "scalar" literals are supported, i.e., no lists/dense constants.

* bools and floats are translated to `arith.constant` with type `i1` and `f64`, respectively
* ints are translated to `bigint.constant` of type `!bigint.bigint`

### DIFF
--- a/.github/workflows/ci-mlir.yml
+++ b/.github/workflows/ci-mlir.yml
@@ -39,6 +39,9 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Mark workspace as safe for git
+      run: git config --global --add safe.directory $GITHUB_WORKSPACE
+
     - name: Install the package locally and nbval
       run: make venv
 
--- a/.github/workflows/markdown-formatting.yml
+++ b/.github/workflows/markdown-formatting.yml
@@ -13,7 +13,7 @@ jobs:
     - uses: actions/checkout@v6
 
     - name: markdownlint-cli2-action
-      uses: DavidAnson/markdownlint-cli2-action@v22
+      uses: DavidAnson/markdownlint-cli2-action@v23
       with:
         config: '.markdownlint.json'
         globs: '**/*.{md,markdown}'


xdsl/dialects/bigint.py
```diff
@@ -48,6 +49,28 @@ def __init__(
         super().__init__(operands=[operand1, operand2], result_types=[bigint])
 
 
+@irdl_op_definition
+class ConstantOp(IRDLOperation):
```
superlopuh, line 63: Can you please split this out into its own PR with an addition to the filecheck test?

AlexanderViand:
@superlopuh I found a few more limitations on the way, will (ab)use this branch/PR as a scratch pad for now and clean up into separate PRs later :)

- [x] Added the `bigint.to_int` we discussed
- [x] Function arguments were restricted to being named variables --> now allows any expr
- [x] Assignments weren't implemented, so there wasn't even a great way to *get* named variables --> added assignment handling

Still TODO: 
- [ ] Canonicalize `bigint.to_int <some bigint.constant> : i32` --> `arith.constant : i32`
      Oh, and look, `bigint` (if it was in C++ MLIR, would be another candidate for something that would beneift from generic constant folding!)
- [ ] Add support for `x.foo` calls
- [ ] ....

superlopuh:
Hi @AlexanderViand, any updates on your plans?

AlexanderViand:
Ohh, thanks for reminding me of this! I totally forgot about this after the mlir-school… I’ll add it back to my ToDo list and see if I have some time before or during the LLVM Developer Meeting to continue hacking on this!

xdsl/dialects/bigint.py
superlopuh: Hi @AlexanderViand, I'm wondering whether you might have time for this in the next few weeks?

AlexanderViand: Yes, I was planning to get back to this (and the other issue) when I'm in Cambridge (week of Feb 16th)

superlopuh: I'll be away but probably still looking at PRs, looking forward!

xdsl/frontend/pyast/code_generation.py
```diff
@@ -385,6 +387,28 @@ def visit_Compare(self, node: ast.Compare) -> None:
...
+            # python floats are always f64, so we use arith.constant : f64
+            const_op = arith.ConstantOp(builtin.FloatAttr(node.value, 64))
+            self.inserter.insert_op(const_op)
+        else:
+            raise CodeGenerationException(
+                self.file,
+                node.lineno,
+                node.col_offset,
+                f"Unsupported constant '{node.value}' of type '{type(node.value).__qualname__}'.",
+            )
```
superlopuh: can you please use a match statement?

superlopuh: Actually, I'm now thinking that this should be customizable by the frontend user, and should live in the context. For example, if you're building something that is aware of python type weirdness, you'd want to distinguish between `c_double` and `float`. Could you please add a mechanism to the context to provide a function to call for a given value type?

AlexanderViand: So add a `LiteralRegistry`, just like there's a `TypeRegistry` and `FunctionRegistry`?

superlopuh: Yep exactly

xdsl/frontend/pyast/README.md
superlopuh: if you happen to have the time, it would be great to update this snippet to not use CodeContext any more, and to demonstrate the functionality of register_literal

xdsl/frontend/pyast/utils/type_conversion.py
```diff
@@ -102,6 +103,41 @@ def resolve_attribute(
...
+
+
+def default_literal_registry() -> LiteralRegistry:
+    """Return a literal registry with the default frontend literal lowering."""
+    registry = LiteralRegistry()
+    registry.insert(bool, lambda v: arith.ConstantOp(builtin.IntegerAttr(int(v), 1)))
+    registry.insert(int, lambda v: bigint.ConstantOp(v))
+    registry.insert(float, lambda v: arith.ConstantOp(builtin.FloatAttr(v, 64)))
+    return registry
+
```
superlopuh: I would prefer not to have this tbh, let's instead have the default registry be empty for now and register things in the tests for the time being?

AlexanderViand: I can make the default be empty, but its probably nice to keep this around (renamed) as a helper?

superlopuh: I would say probably not, as then only the literals are registered and not the other ops. My vision here is that we can ship multiple frontends, or our clients can, and then I think it's less confusing if it's all registered in one place instead of shared between some defaults and some extensions.

AlexanderViand:
@superlopuh  Updated the README and removed the default literal registry :)

EDIT: and I have *no* idea why the CI fails all of a sudden.... 🤷

PR #5223, opened 2025-09-12
transformations: (eqsat-add-costs) Support external costs from a JSON file

osmanyasar05:
This PR adds support for reading operation costs from a JSON file. If a cost is not found in the JSON file, it falls back to the default cost. An error is raised if unregistered operations are not allowed and the JSON file contains an operation that is not registered in the current context.


xdsl/transforms/eqsat_add_costs.py
```diff
@@ -96,5 +104,20 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+
+        cost_dict: dict[str, int] = {}
+
+        if self.cost_file is not None:
+            assert os.path.exists(self.cost_file)
+            with open(self.cost_file) as file:
+                external_costs = json.load(file)
+                for dialect in external_costs:
+                    for op in external_costs[dialect]:
+                        if ctx.get_optional_op(f"{dialect}.{op}") is None:
```
superlopuh: It's a shame to not be O(1) here, can you please change the dict and the logic to just map ops to costs?

superlopuh (approved):
Thank you!

PR #5228, opened 2025-09-13
documentation: minor fixes of PDL notebook

superlopuh:
- add a tick
- remove 3.0 and 6.0, and print python function
- nicer format fow miscompilation

### DIFF
--- a/docs/marimo/pdl.py
+++ b/docs/marimo/pdl.py
@@ -206,10 +206,10 @@ def _(mo):
 
 
 @app.cell(hide_code=True)
-def _(mo):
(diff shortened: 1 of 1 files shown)


Jimmy2027:
I think the solution of exercise 3 is wrong, should be matching `arith.divf` instead of `arith.addf` and replacing it with const 1. Might be worth fixing here as well, the rest looks good to me

superlopuh:
My bad I hand't merged main, this was already resolved in https://github.com/xdslproject/xdsl/pull/5208

PR #5232, opened 2025-09-14
core: add folding mechanism

jumerckx:
This pr includes
* a `HasFolderInterface` operation interface
* ~~a `get_constant_value` in the `ConstantLike` trait~~ (moved to #5233)
* ~~a `ConstantMaterializationInterface ` dialect interface~~ (moved to #5234)
* an implementation of folding for arith's SignlessIntegerBinaryOperation.

~~I'm opening this as a draft to check whether this seems like the right direction for a folding implementation. If it is, I can split this up in separate PRs.~~

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -28,7 +28,7 @@ from xdsl.dialects.builtin import (
     VectorType,
 )
 from xdsl.dialects.utils import FastMathAttrBase, FastMathFlag
-from xdsl.interfaces import ConstantLikeInterface
+from xdsl.interfaces import ConstantLikeInterface, HasFolderInterface
 from xdsl.ir import (
     Attribute,
     BitEnumAttribute,
@@ -182,7 +182,7 @@ class ConstantOp(IRDLOperation, ConstantLikeInterface):
         return self.value
 
 
-class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
+class SignlessIntegerBinaryOperation(IRDLOperation, HasFolderInterface, abc.ABC):
     """A generic base class for arith's binary operations on signless integers."""
 
     T: ClassVar = VarConstraint("T", signlessIntegerLike)
@@ -224,6 +224,23 @@ class SignlessIntegerBinaryOperation(IRDLOperation, abc.ABC):
         See external [documentation](https://en.wikipedia.org/wiki/Identity_element).
         """
         return False
+
+    def fold(self):
+        lhs = self.get_constant(self.lhs)
+        rhs = self.get_constant(self.rhs)
+        if lhs is not None and rhs is not None:
+            if isinstance(lhs, IntegerAttr) and isinstance(rhs, IntegerAttr):
+                lhs = cast(IntegerAttr, lhs)
+                rhs = cast(IntegerAttr, rhs)
+                assert lhs.type == rhs.type
+                result = self.py_operation(lhs.value.data, rhs.value.data)
+                if result is not None:
+                    return (IntegerAttr(result, lhs.type),)
+        if (
+            isinstance(rhs, IntegerAttr)
+            and self.is_right_unit(cast(IntegerAttr, rhs)) == 0
+        ):
+            return (self.lhs,)
 
     def __init__(
         self,
@@ -361,6 +378,20 @@ class AddiOp(SignlessIntegerBinaryOperationWithOverflow):
     @staticmethod
     def is_right_unit(attr: IntegerAttr) -> bool:
         return attr.value.data == 0
+
+    def fold(self):
+        lhs = self.get_constant(self.lhs)
+        rhs = self.get_constant(self.rhs)
+        if lhs is not None and rhs is not None:
+            if isinstance(lhs, IntegerAttr) and isinstance(rhs, IntegerAttr):
+                lhs = cast(IntegerAttr, lhs)
+                rhs = cast(IntegerAttr, rhs)
+                assert lhs.type == rhs.type
+                result = self.py_operation(lhs.value.data, rhs.value.data)
+                if result is not None:
+                    return (IntegerAttr(result, lhs.type),)
+        if isinstance(rhs, IntegerAttr) and self.is_right_unit(cast(IntegerAttr, rhs)):
+            return (self.lhs,)
 
 
 @irdl_op_definition
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -234,7 +234,7 @@ class SignlessIntegerBinaryOperation(IRDLOperation, HasFolderInterface, abc.ABC)
                 result = self.py_operation(lhs.value.data, rhs.value.data)
                 if result is not None:
                     return (IntegerAttr(result, lhs.type),)
-        if isa(rhs, IntegerAttr) and self.is_right_unit(rhs) == 0:
+        if isa(rhs, IntegerAttr) and self.is_right_unit(rhs):
             return (self.lhs,)
 
     def __init__(
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -229,17 +229,12 @@ class SignlessIntegerBinaryOperation(IRDLOperation, HasFolderInterface, abc.ABC)
         lhs = self.get_constant(self.lhs)
         rhs = self.get_constant(self.rhs)
         if lhs is not None and rhs is not None:
-            if isinstance(lhs, IntegerAttr) and isinstance(rhs, IntegerAttr):
-                lhs = cast(IntegerAttr, lhs)
-                rhs = cast(IntegerAttr, rhs)
+            if isa(lhs, IntegerAttr) and isa(rhs, IntegerAttr):
                 assert lhs.type == rhs.type
                 result = self.py_operation(lhs.value.data, rhs.value.data)
                 if result is not None:
                     return (IntegerAttr(result, lhs.type),)
-        if (
-            isinstance(rhs, IntegerAttr)
-            and self.is_right_unit(cast(IntegerAttr, rhs)) == 0
-        ):
+        if isa(rhs, IntegerAttr) and self.is_right_unit(rhs) == 0:
             return (self.lhs,)
 
     def __init__(
@@ -378,20 +373,6 @@ class AddiOp(SignlessIntegerBinaryOperationWithOverflow):
     @staticmethod
     def is_right_unit(attr: IntegerAttr) -> bool:
         return attr.value.data == 0
-
-    def fold(self):
-        lhs = self.get_constant(self.lhs)
-        rhs = self.get_constant(self.rhs)
-        if lhs is not None and rhs is not None:
-            if isinstance(lhs, IntegerAttr) and isinstance(rhs, IntegerAttr):
-                lhs = cast(IntegerAttr, lhs)
-                rhs = cast(IntegerAttr, rhs)
-                assert lhs.type == rhs.type
-                result = self.py_operation(lhs.value.data, rhs.value.data)
-                if result is not None:
-                    return (IntegerAttr(result, lhs.type),)
-        if isinstance(rhs, IntegerAttr) and self.is_right_unit(cast(IntegerAttr, rhs)):
-            return (self.lhs,)
 
 
 @irdl_op_definition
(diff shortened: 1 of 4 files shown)


xdsl/dialects/arith.py
```diff
@@ -352,6 +371,24 @@ def py_operation(lhs: int, rhs: int) -> int | None:
...
+            assert (
+                result := self.py_operation(lhs.value.data, rhs.value.data)
+            ) is not None
+            # TODO: Argument of type "Attribute" cannot be assigned to parameter "value_type"
+            # of type "IntCovT@__init__" in function "__init__"
+            return (IntegerAttr(result, type),)
+        if isinstance(lhs, IntegerAttr) and lhs.value.data == 0:
+            return (self.rhs,)
+        if isinstance(rhs, IntegerAttr) and rhs.value.data == 0:
+            return (self.lhs,)
```
jumerckx: In MLIR, only `x+0 --> x` is checked, not the other direction. I'm not sure why that's the case.

xdsl/dialects/arith.py
```diff
@@ -1373,6 +1410,18 @@ def verify_(self) -> None:
...
+    def materialize_constant(
+        self, ctx: Context, value: Attribute, type: TypeAttribute
+    ) -> Operation:
+        if not isinstance(value, IntegerAttr | FloatAttr):
+            raise ValueError("expected IntegerAttr or FloatAttr")
+        # TODO: Argument type is partially unknown
+        #   Argument corresponds to parameter "value" in function "__init__"
+        #   Argument type is "IntegerAttr[Unknown] | FloatAttr[Unknown]"PylancereportUnknownArgumentType
+        return ConstantOp(value=value, value_type=type)
+
```
jumerckx: TODO: I'm doing something wrong here with the type checking

xdsl/interfaces.py
```diff
@@ -49,3 +50,35 @@ class HasCanonicalizationPatternsInterface(Operation, abc.ABC):
     @abc.abstractmethod
     def get_canonicalization_patterns(cls) -> tuple[RewritePattern, ...]:
         raise NotImplementedError()
+
+
+class _HasFolderInterfaceTrait(OpTrait):
+    # TODO: Do we even need a trait if there's already an interface?
+    def verify(self, op: Operation) -> None:
+        return
```
jumerckx, line 95: `HasCanonicalizationInterface` does something similar but that's maybe just for backwards compatibility?

xdsl/interfaces.py
```diff
@@ -49,3 +50,35 @@ class HasCanonicalizationPatternsInterface(Operation, abc.ABC):
...
+    @abc.abstractmethod
+    def fold(self) -> Sequence[OpFoldResult] | None:
+        """
+        Attempts to fold the operation. The fold method cannot modify the IR.
+        Returns either an existing SSAValue or an Attribute for each result of the operation.
+        When folding is unsuccessful, returns None.
+        """
+        # TODO: should we support fold methods that mutate the operation?
+        #   That would require fold to return an additional flag
+        #   indicating whether the operation was modified?
```
jumerckx: Should we support fold methods that mutate the operation? That would require fold to return an additional flag indicating whether the operation was modified?

xdsl/dialects/arith.py
superlopuh: https://mlir.llvm.org/docs/Canonicalization/

xdsl/dialects/arith.py
```diff
@@ -127,6 +137,15 @@ def __init__(self, flags: None | Sequence[IntegerOverflowFlag] | Literal["none"]
         super().__init__(flags)
 
 
+class ArithConstantLike(ConstantLike):
+    """ConstantLike implementation for arith.constant operations."""
+
+    def get_constant_value(self, op: Operation) -> Attribute:
+        """Get the constant value from an arith.constant operation."""
+        assert isinstance(op, ConstantOp)
+        return op.value
```
superlopuh: The way we've set it up in xDSL is that traits are the mechanism by which behaviour is found, and interfaces are a convenience mechanism to implement traits. Here my gut feeling is to add a `ConstantLikeInterface` to add this method to the operation directly.

superlopuh: Doing this in a separate PR will hopefully explain how to define interfaces and make doing the rest of this PR easier.

xdsl/dialects/arith.py
jumerckx: Hmm, should we match this? This means folding only works when code is canonicalized. I'm wondering what the implications for eqsat are.

superlopuh: Aren't we doing canonicalization on steroids?

xdsl/dialects/arith.py
jumerckx: https://github.com/xdslproject/xdsl/pull/5233

xdsl/dialects/arith.py
jumerckx: I guess you're right 👍

xdsl/traits.py
```diff
@@ -52,6 +52,25 @@ def get_constant_value(cls, op: Operation) -> Attribute:
         raise NotImplementedError()
 
 
+OpFoldResult = SSAValue | Attribute
```
superlopuh: let's add a doc string here

xdsl/interfaces.py
superlopuh: Does MLIR allow this? I prefer nonmutating

xdsl/traits.py
jumerckx: Ah, I've just removed this in favor of putting `SSAValue | Attribute` explicitly because there was an issue with circular imports.

xdsl/interfaces.py
jumerckx: Yes, referenced here for example: https://github.com/llvm/llvm-project/pull/88508
I agree with non-mutating, as this avoids some awkward logic

xdsl/dialects/arith.py
```diff
@@ -362,6 +379,20 @@ def py_operation(lhs: int, rhs: int) -> int | None:
     def is_right_unit(attr: IntegerAttr) -> bool:
         return attr.value.data == 0
 
+    def fold(self):
+        lhs = self.get_constant(self.lhs)
+        rhs = self.get_constant(self.rhs)
+        if lhs is not None and rhs is not None:
+            if isinstance(lhs, IntegerAttr) and isinstance(rhs, IntegerAttr):
```
superlopuh:
```suggestion
            if isa(lhs, IntegerAttr) and isa(rhs, IntegerAttr):
```

superlopuh: But also, it feels like this should replace the caller of `py_operation` and `is_right_unit`, right?

xdsl/interfaces.py
```diff
@@ -78,3 +83,47 @@ class ConstantLikeInterface(Operation, abc.ABC):
...
+    """
+
+    traits = traits_def(_HasFolderInterfaceTrait())
+
+    def get_constant(self, operand: SSAValue) -> Attribute | None:
+        if not isinstance(operand, OpResult):
+            return None
+        operand_op = operand.owner
+        if (t := operand_op.get_trait(ConstantLike)) is not None:
+            return t.get_constant_value(operand_op)
```
superlopuh: meganit
```suggestion
        if isinstance(operand_op := operand.owner, Operation) and (operand_op := operand_op.get_trait(ConstantLike)) is not None:
            return t.get_constant_value(operand_op)
```

xdsl/builder.py
```diff
@@ -115,6 +118,41 @@ def insert_op(
 
         return op
 
+    def try_fold(
```
superlopuh: who calls this?

xdsl/dialects/arith.py
jumerckx: I've now implemented a more generic `fold` that uses `py_operation` and `is_right_unit`.

xdsl/builder.py
jumerckx: no one yet. This is what I would call in eqsat's pdl_interp.create_operation

superlopuh: Where is it called in MLIR?

jumerckx: createOrFold (which is part of opBuilder), and legalizeWithfold in the legalize pass.
xdsl's builder does not have a create method for operations, should it have one?

superlopuh: @math-fehr Should we have `insert_or_fold` or something like this?

superlopuh: Maybe folding can be a flag on Builder.insert? then the API could be replicated on all the insert_after and replace methods?

jumerckx: I like that

jumerckx: But then, either a `Context` needs to be accessible in Builder, or `Operation` needs to have a reference to its actual `Dialect`.

superlopuh: I think a `Context` on `Builder` makes sense, would like to hear @math-fehr's opinion

jumerckx: Here's an attempt at that: https://github.com/xdslproject/xdsl/pull/5239

jumerckx: Another consideration when adding a folding flag to `insert_op` is that now `insert_op` should be able to return a collection of SSAValues when the results of the operation to insert have been folded into existing SSAValues.
Perhaps having a separate `insert_op_with_folding` would be better? Or a `builder.fold` method that is applied on operations that are already inserted?

EDIT: Sasha mentioned that it's possible to have different return types based on the argument types so this is not a big problem

jumerckx:
I've split off the contentious builder changes:
* adding context to builder: https://github.com/xdslproject/xdsl/pull/5239
* WIP: adding some form of folding to builder: https://github.com/xdslproject/xdsl/compare/jumerckx/builder_context...jumerckx/builder_fold

This pr now only includes the trait and interface, and an implementation for some arith operations

xdsl/dialects/arith.py
```diff
@@ -225,6 +225,18 @@ def is_right_unit(attr: IntegerAttr) -> bool:
...
+        lhs = self.get_constant(self.lhs)
+        rhs = self.get_constant(self.rhs)
+        if lhs is not None and rhs is not None:
+            if isa(lhs, IntegerAttr) and isa(rhs, IntegerAttr):
+                assert lhs.type == rhs.type
+                result = self.py_operation(lhs.value.data, rhs.value.data)
+                if result is not None:
+                    return (IntegerAttr(result, lhs.type),)
+        if isa(rhs, IntegerAttr) and self.is_right_unit(rhs) == 0:
+            return (self.lhs,)
```
superlopuh: Can you please make the implementation more like `SignlessIntegerBinaryOperationZeroOrUnitRight` and `SignlessIntegerBinaryOperationConstantProp`, and replace those implementations with a call to `fold`?

xdsl/interfaces.py
superlopuh: Yeah all traits need a verify method but sometimes it's a noop

xdsl/interfaces.py
superlopuh: Let's remove this comment? Maybe instead add to the doc comment, that this is not mutating, in contrast to MLIR?

xdsl/dialects/arith.py
jumerckx: You mean to unify those `SignlessIntegerBinaryOperationZeroOrUnitRight ` and `SignlessIntegerBinaryOperationConstantProp` by having `fold` do both constprop and the zero/unit simplification?

superlopuh: I guess so, it's what it's there for right? This is until folding is somehow incorporated into canonicalization directly, might as well not duplicate the logic when possible for now.

jumerckx: In MLIR, these simple fold patterns are not implemented as canonicalizations but the canonicalizer calls the folder.
I guess we should do the same but maybe we should keep that for a follow up pr?

As for the implementation now, I'm wondering if it even makes sense to try have a generic fold for many arith operations, only the most basic behaviour is shared.

superlopuh: I agree, there's also an issue that we only saw later in that it's very expensive to fold in this way, inlining the code seems like a good idea. I'm happy for this to be in the next PR.

superlopuh (approved):

PR #5234, opened 2025-09-14
core: add ConstantMaterializationInterface

jumerckx:
This is a dialect interface containing `materialize_constant` to create a constant operation from a dialect given a constant attribute.
I had to use `if TYPE_CHECKING` to handle a circular import issue.

(on top of https://github.com/xdslproject/xdsl/pull/5233)

### DIFF
--- a/xdsl/dialects/arith.py
+++ b/xdsl/dialects/arith.py
@@ -4,9 +4,12 @@ import abc
 from collections.abc import Mapping, Sequence
 from typing import ClassVar, Literal, cast
 
+from xdsl.context import Context
+from xdsl.dialect_interfaces import ConstantMaterializationInterface
 from xdsl.dialects.builtin import (
     AnyFloat,
     AnyFloatConstr,
+    ComplexType,
     ContainerOf,
     DenseIntOrFPElementsAttr,
     DenseResourceAttr,
@@ -28,7 +31,14 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.dialects.utils import FastMathAttrBase, FastMathFlag
 from xdsl.interfaces import ConstantLikeInterface
-from xdsl.ir import Attribute, BitEnumAttribute, Dialect, Operation, SSAValue
+from xdsl.ir import (
+    Attribute,
+    BitEnumAttribute,
+    Dialect,
+    Operation,
+    SSAValue,
+    TypeAttribute,
+)
 from xdsl.irdl import (
     AnyAttr,
     AnyOf,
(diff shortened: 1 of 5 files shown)


superlopuh (approved):

superlopuh:
Oh let's update the `constant_fold_interp` pass while we're at it?

jumerckx:
Done!

jumerckx:
That was overoptimistic, the interpreter returns actual python objects instead of attributes. Will look into this tomorrow.

jumerckx:
Essentially half of the problem remains: we don't have a way to go from a Python object to the correct Attribute. We can again go best effort (`float`=>`FloatAttr`, `int`=>`IntAttr`, ...). Or perhaps we can extend the interface compared to MLIR to support this use case as well, e.g. having a `materialize_constant_from_python` that does this conversion?

xdsl/dialects/arith.py
```diff
@@ -1376,6 +1384,32 @@ def verify_(self) -> None:
...
 
+class ArithConstantMaterializationInterface(ConstantMaterializationInterface):
+    def materialize_constant(self, value: Attribute, type: Attribute) -> Operation:
+        if not isinstance(
+            value,
+            IntegerAttr | FloatAttr | DenseIntOrFPElementsAttr | DenseResourceAttr,
+        ):
+            raise ValueError(
+                "expected IntegerAttr, FloatAttr, DenseIntOrFPElementsAttr, or DenseResourceAttr"
+            )
```
superlopuh: I think it would be better to fail with a verification error here
```suggestion
```

xdsl/dialects/arith.py
```diff
@@ -1376,6 +1384,32 @@ def verify_(self) -> None:
...
+        elif isinstance(value, DenseIntOrFPElementsAttr):
+            value = cast(
+                DenseIntOrFPElementsAttr[
+                    IntegerType | IndexType | AnyFloat | ComplexType
+                ],
+                value,
+            )
+        # DenseResourceAttr doesn't need type casting
+
+        return ConstantOp(value=value, value_type=type)
```
superlopuh: Let's build the op here instead, and skip the type checking

superlopuh:
Nah I'll just fix the interpreter to use attributes instead of python objects. I'll probably need to do this before the end of the year anyway. But here we give the type to use in the function? I don't think it needs any more information that it already has, I'm happy with asserting that it's one of the cases that we handle here, and not handling the others for now.

jumerckx:
> But here we give the type to use in the function? I don't think it needs any more information that it already has

Can we assume that if a ConstantLike operation returns a result of a certain type, the constant attribute corresponds to that type? I'd assumed not.

superlopuh:
Let's modify constant_op_for_value to return IntegerAttr here? It's a minor change. The assumption is unsafe but it's localized to this pass.

jumerckx:
`build` doesn't verify does it? Should I add an explicit verify after the operation is built?

superlopuh:
I think this is fine, verification between passes should be enough.

PR #5240, opened 2025-09-16
core: (AnyOf) allow for abstract classes without bases

jorendumoulin:


tests/irdl/test_attr_constraint.py
```diff
@@ -284,13 +284,6 @@ def test_constraint_simplification(lhs: AttrConstraint, rhs: AttrConstraint):
 @pytest.mark.parametrize(
     "c1, c2, msg",
     [
-        (
-            AnyAttr(),
-            BaseAttr(AttrA),
-            re.escape(
-                "Constraint AnyAttr() cannot appear in an `AnyOf` constraint as its bases aren't known"
-            ),
-        ),
```
superlopuh, line 293: We should still fail here, no? Can you please add more tests to check that none of the branches overlap?

xdsl/irdl/constraints.py
```diff
@@ -465,12 +466,18 @@ def __init__(
...
-                    f"Constraint {c} cannot appear in an `AnyOf` constraint as its bases aren't known."
-                )
+                if abstr_constr is None:
+                    abstr_constr = c
+                    continue
+                else:
+                    raise PyRDLError(
+                        "Only one abstract constraint is allowed in `AnyOf` constraint,"
+                        f" found {c} when {abstr_constr} was already present."
+                    )
```
superlopuh: we have to check here that none of the bases or eq bases satisfy the abstract constraint

xdsl/irdl/constraints.py
```diff
@@ -506,14 +513,23 @@ def __init__(
...
+        if constr is not None:
+            constr.verify(attr, constraint_context)
+            return
+        # Try abstract constraint if present
+        if self._abstr_constr is not None:
+            try:
+                self._abstr_constr.verify(attr, constraint_context)
+                return
+            except VerifyException:
+                pass
```
superlopuh:
```suggestion
        self._abstr_constr.verify(attr, constraint_context)
        return
```

xdsl/irdl/constraints.py
superlopuh: Now that I think about it, I don't think we can do any abstract constraint, it has to be a non-final class, which would let us check `issubclass` for bases and `isinstance` for eq_bases

alexarice:
How does this differ to https://github.com/xdslproject/xdsl/pull/5111?

xdsl/irdl/constraints.py
```diff
@@ -491,6 +503,23 @@ def __init__(
                     based_constrs[base] = c
                 bases |= b
 
+        # check for overlaps with the abstract constraint
+        if abstr_constr:
+            # equality constraints should not overlap
+            for attr, constr in eq_constrs.items():
+                if isinstance(attr, abstr_constr.attr):
+                    raise PyRDLError(
+                        f"Equality constraint {constr} overlaps with the abstract "
```
superlopuh: It feels like it's cleaner to reconstruct the equality constraint for the error message here, instead of making the rest of the logic more complex with the dictionary

xdsl/irdl/constraints.py
```diff
@@ -460,17 +461,28 @@ def __init__(
...
+                            f"Abstract constraint in `AnyOf` must be a `BaseAttr` "
+                            f"with a non-final attribute class, got {c} instead."
+                        )
+                    abstr_constr = c
+                    continue
+                else:
+                    raise PyRDLError(
+                        "Only one abstract constraint is allowed in `AnyOf` constraint,"
+                        f" found {c} when {abstr_constr} was already present."
+                    )
```
superlopuh: nit: would be nice to keep indentation level as low as possible
```suggestion
                if abstr_constr is not None:
                    raise PyRDLError(
                        "Only one abstract constraint is allowed in `AnyOf` constraint,"
                        f" found {c} when {abstr_constr} was already present."
                    )
                if not isinstance(c, BaseAttr) or is_runtime_final(c.attr):
                    raise PyRDLError(
                        f"Abstract constraint in `AnyOf` must be a `BaseAttr` "
                        f"with a non-final attribute class, got {c} instead."
                    )
                abstr_constr = c
                continue
```

xdsl/irdl/constraints.py
superlopuh: Actually I guess it's not a big deal, your call

xdsl/irdl/constraints.py
```diff
@@ -460,17 +461,28 @@ def __init__(
             irdl_to_attr_constraint(constr) for constr in attr_constrs
         )
 
-        eq_constrs = set[Attribute]()
+        eq_constrs = dict[Attribute, AttrConstraint[AttributeCovT]]()
         based_constrs = dict[type[Attribute], AttrConstraint[AttributeCovT]]()
 
-        bases = set[Attribute]()
```
superlopuh, line 466: was this just wrong before? I can't figure out how the typing makes sense

xdsl/irdl/constraints.py
```diff
@@ -491,6 +503,23 @@ def __init__(
                     based_constrs[base] = c
                 bases |= b
 
+        # check for overlaps with the abstract constraint
+        if abstr_constr:
```
superlopuh: nit:
```suggestion
        if abstr_constr is not None:
```

xdsl/irdl/constraints.py
```diff
@@ -460,17 +461,28 @@ def __init__(
...
 
-        eq_constrs = set[Attribute]()
+        eq_constrs = dict[Attribute, AttrConstraint[AttributeCovT]]()
         based_constrs = dict[type[Attribute], AttrConstraint[AttributeCovT]]()
 
-        bases = set[Attribute]()
-        eq_bases = set[Attribute]()
+        bases = set[type[Attribute]]()
+        eq_bases = set[type[Attribute]]()
+        abstr_constr: AttrConstraint[AttributeCovT] | None = None
```
superlopuh, line 475: Would this work? Seems a little simpler
```suggestion
        abstr_constr: type[Attribute] | None = None
```

jorendumoulin: This would work, yes, but then the error message clashes a bit in style with other ones:
For this it seemed nicer to keep track of the constraint rather than just the attribute class

```
"Non-equality constraint BaseAttr(AttrA) overlaps with the abstract constraint <class 'tests.irdl.test_attr_constraint.Base'> in `AnyOf` constraint."
```

vs

```
"Non-equality constraint BaseAttr(AttrA) overlaps with the abstract constraint BaseAttr(Base) in `AnyOf` constraint."
```

xdsl/irdl/constraints.py
jorendumoulin: I think it was wrong before, yes

xdsl/irdl/constraints.py
superlopuh: I guess it's a slightly different thing, we can make the error message look however we want, like for example reconstructing the constraint. I would optimise for simpler and more efficient code in the verification at the expense of slower code in the failure case if that makes sense.

xdsl/irdl/constraints.py
superlopuh: Great that you're fixing it!

alexarice: Surprising that this worked before

xdsl/irdl/constraints.py
```diff
@@ -460,17 +461,27 @@ def __init__(
             irdl_to_attr_constraint(constr) for constr in attr_constrs
         )
 
-        eq_constrs = set[Attribute]()
+        eq_constrs = dict[Attribute, AttrConstraint[AttributeCovT]]()
```
alexarice: Why is this needed? Should be no need to actually ever call the constraint right?

xdsl/irdl/constraints.py
alexarice: I think having the `AttrConstraint` here is fine. This could potentially be a `ParamAttrConstraint` or other `AnyOf` right?

xdsl/irdl/constraints.py
```diff
@@ -460,17 +461,27 @@ def __init__(
...
+        abstr_constr: AttrConstraint[AttributeCovT] | None = None
         for i, c in enumerate(constrs):
             b = c.get_bases()
             if b is None:
-                raise PyRDLError(
-                    f"Constraint {c} cannot appear in an `AnyOf` constraint as its bases aren't known."
-                )
+                if abstr_constr is not None:
+                    raise PyRDLError(
+                        "Only one abstract constraint is allowed in `AnyOf` constraint,"
```
alexarice: I worry that the terminology "abstract constraint" won't mean much to someone whose not very familiar with these mechanisms.

xdsl/irdl/constraints.py
```diff
@@ -506,14 +534,20 @@ def __init__(
...
-            raise VerifyException(f"Unexpected attribute {attr}")
-        constr.verify(attr, constraint_context)
+        if constr is not None:
+            constr.verify(attr, constraint_context)
+            return
+        # Try abstract constraint if present
+        if self._abstr_constr is not None:
+            self._abstr_constr.verify(attr, constraint_context)
+            return
+        raise VerifyException(f"Unexpected attribute {attr}")
```
alexarice, line 556: Ideally we should check if the base class aligns with the abstract base class and then pass it to the constraint if it does and only throw the "Unexpected attribute" error otherwise?

xdsl/irdl/constraints.py
```diff
@@ -460,17 +461,27 @@ def __init__(
...
             if b is None:
-                raise PyRDLError(
-                    f"Constraint {c} cannot appear in an `AnyOf` constraint as its bases aren't known."
-                )
+                if abstr_constr is not None:
+                    raise PyRDLError(
+                        "Only one abstract constraint is allowed in `AnyOf` constraint,"
+                        f" found {c} when {abstr_constr} was already present."
+                    )
+                if not isinstance(c, BaseAttr) or is_runtime_final(c.attr):
```
alexarice, line 484: I would still prefer that we don't special case `BaseAttr` here, and instead modify `get_bases` to also return non final bases. (Should be able to pull some code from #5111).

alexarice:
The issue I thought I'd found with `BaseAttr(Attribute)` seems like it's actually already resolved in this PR.

xdsl/irdl/constraints.py
superlopuh: I would rather first merge a more restrictive version if we are all happy that it's correct, and then look at more powerful constraints.

xdsl/irdl/constraints.py
jorendumoulin: this was to use the constraint in an error message but this seems to be confusing, so i reverted this change and reconstructed the constraint in the error message instead

xdsl/irdl/constraints.py
jorendumoulin: A second point where this would be mildly inconvenient is at 
`self._abstr_constr.verify(attr, constraint_context)`,
where I would then have to duplicate the check and related error message from `BaseAttr`, so to me it makes more sense to just store the constraint rather than the `type[Attribute]`

xdsl/irdl/constraints.py
jorendumoulin: Lacking better naming, I added some explanation as to what an 'abstract constraint' means in the docstring

xdsl/irdl/constraints.py
jorendumoulin: What do you mean with _aligns_? Is this just `isinstance`? Because that is what the `constr.verify` checks anyway

xdsl/irdl/constraints.py
jorendumoulin: Right now the check for whether a constraint is abstract is based on `get_bases` being None which would then fail. Would this be better we just check on `is_runtime_final`?

xdsl/irdl/constraints.py
superlopuh: OK

xdsl/irdl/constraints.py
superlopuh: I think one problem with this phrasing is that the constraint is not abstract, the class is, maybe rephrasing it like this would make it clearer?
```suggestion
                        f"Cannot form AnyOf constraint with {abstr_constr} and {c}, as they cannot be verified as disjoint"
```

xdsl/irdl/constraints.py
superlopuh: I would also suggest doing this in a separate PR if that's OK? Seems orthogonal to this change unless I'm missing something

superlopuh (approved):
TYVM!

PR #5244, opened 2025-09-16
core: add `parse_dimension_list`

jakedves:
Added the `DimensionList` custom directive, needed for some ops in the `mesh`/`shard` dialect's `assembly_format`, and the `parse_dimension_list` API to the Attribute Parser [(MLIR equivalent)](https://github.com/llvm/llvm-project/blob/92f5d8df361bb1bb6dea88f86faeedfd295ab970/mlir/lib/AsmParser/TypeParser.cpp#L542).

Happy to discuss and iterate if not right fit for xDSL.

Added test cases for public APIs in `dimension_list.py`

### DIFF
--- a/xdsl/parser/attribute_parser.py
+++ b/xdsl/parser/attribute_parser.py
@@ -507,13 +507,15 @@ class AttrParser(BaseParser):
         characters, it will split the token. For instance, 'x1' will be split
         into 'x' and '1'.
         """
+        if self._parse_optional_shape_delimiter() is not None:
+            return
+
         token = self._current_token
         tk = token.kind
 
         err_val = tk.name if tk != MLIRTokenKind.BARE_IDENT else token.text
 
-        self.expect(
-            self._parse_optional_shape_delimiter,
+        self.raise_error(
             f"Expected 'x' in shape delimiter, got {err_val}",
         )
 


xdsl/parser/attribute_parser.py
```diff
@@ -504,21 +504,41 @@ def parse_shape_delimiter(self) -> None:
...
+        else:
+            if self._current_token.kind in accepted_token_kinds:
+                dim = self.parse_shape_dimension()
+
+                while self._current_token.text[0] == "x":
+                    self.parse_shape_delimiter()
+                    dim = self.parse_shape_dimension()
+                    dims.append(dim)
+
+        return dims
```
superlopuh: Why change the logic here? Won't the function below work? Why the need for `trailing_x`?

xdsl/dialects/utils/dimension_list.py
```diff
@@ -0,0 +1,68 @@
...
+from xdsl.utils.hints import isa
+
+
+def parse_empty_dimension_list(parser: Parser) -> bool:
+    if parser.parse_optional_characters("["):
+        parser.parse_characters("]")
+
+        return True
+
+    return False
```
superlopuh: Why this function? What happens if you call parse_dimension_list with an empty list?

tests/test_dialect_utils.py
```diff
@@ -229,6 +231,46 @@ def test_parse_dynamic_index_list_with_custom_delimiter():
...
+
+    parser = Parser(ctx, "5x5x5x5")
+    assert not parse_empty_dimension_list(parser)
+
+    with pytest.raises(
+        ParseError,
+        match=re.escape("']' expected"),
+    ):
+        parser = Parser(ctx, "[5, 10]")
+        parse_empty_dimension_list(parser)
```
superlopuh: Can you please check for the expected output instead of truthy/falsey?

tests/test_dialect_utils.py
```diff
@@ -229,6 +231,46 @@ def test_parse_dynamic_index_list_with_custom_delimiter():
...
+    dims = DenseArrayBase[I64].from_list(i64, [8, dyn_index, 2, dyn_index])
+    print_dimension_list(printer, dims)
+    assert stream.getvalue() == "8x?x2x?"
+
+    # Test case 3: Empty
+    stream = StringIO()
+    printer = Printer(stream)
+    dims = DenseArrayBase[I64].from_list(i64, [])
+    print_dimension_list(printer, dims)
+    assert stream.getvalue() == "[]"
```
superlopuh: please use `pytest.mark.parametrize`

xdsl/dialects/utils/dimension_list.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def print_dimension_list(printer: Printer, dims: DenseArrayBase[IntegerType]) -> None:
```
superlopuh: Feels like we should have a `print_dimension_list` on the printer if we're adding it to the parser? I would recommend splitting this PR, first adding print/parse dimension list, and then adding the custom directive.

xdsl/printer.py
```diff
@@ -409,6 +409,17 @@ def print_int(self, value: int, type: IntegerType | IndexType | None = None):
         else:
             self.print_string(f"{value:d}")
 
+    def print_dimension_list(self, dims: Sequence[int]):
+        if len(dims) == 0:
+            self.print_string("[]")
+            return
```
superlopuh: Let's not print the `[]` here, and use this helper in the other types that print dimensions in this format. The easiest way to find them is to search for "?" in the project.

xdsl/parser/attribute_parser.py
```diff
@@ -504,21 +504,48 @@ def parse_shape_delimiter(self) -> None:
...
+        """
+        Parses an empty dimension list, which has it's own syntax:
+          empty-dimension-list ::= `[` `]`
+        Note that a regular dimension list has no square brackets.
+        """
+        if self.parse_optional_characters("["):
+            self.parse_characters("]")
+            return True
+
+        return False
```
superlopuh: This should be done by the caller, not all dimension lists are printed with `[]`

xdsl/parser/attribute_parser.py
```diff
@@ -504,21 +504,48 @@ def parse_shape_delimiter(self) -> None:
...
+
+        while self._current_token.kind in accepted_token_kinds:
+            dim = self.parse_shape_dimension()
+            dims.append(dim)
+
+            # on last iteration, 'x' may not exist, e.g.: 5x3x2
+            if len(self._current_token.text) > 0 and self._current_token.text[0] == "x":
+                self.parse_shape_delimiter()
+
+        return dims
```
superlopuh: Would something like this work?

```
dims = []

while token is int or ?:
    dims.append(parse_dim())
    if not parse_optional_shape_delimiter():
        break

return dims
```

This way we consume the trailing x if there is one, and limit the shape delimiter logic to one function.

jakedves:
The way we have parsing of the dimension list works, but means we get different error messages to MLIR, e.g.:
```illegal_array_missing_x = !emitc.array<10>```

Should produce the error:
```Expected 'x' in shape delimiter, got GREATER```

 in MLIR and the test cases, but now we accept a missing 'x' so this is deferred to `parse_shape` and becomes:
```Expected shape type.```

File: xdsl/printer.py [hunk 9]
```diff
@@ -409,6 +409,13 @@ def print_int(self, value: int, type: IntegerType | IndexType | None = None):
         else:
             self.print_string(f"{value:d}")
 
+    def print_dimension_list(self, dims: Sequence[int]):
+        self.print_list(
```
superlopuh: Can you please add a short description here?

xdsl/parser/attribute_parser.py
```diff
@@ -485,6 +485,21 @@ def parse_shape_dimension(self, allow_dynamic: bool = True) -> int:
 
         return int_token.kind.get_int_value(int_token.span)
 
+    def parse_optional_shape_delimiter(self) -> bool:
```
superlopuh: let's call this in `parse_shape_delimiter` below, using `expect`

xdsl/parser/attribute_parser.py
```diff
@@ -504,20 +519,46 @@ def parse_shape_delimiter(self) -> None:
...
         """
-        dims: list[int] = []
-        while self._current_token.kind in (
-            MLIRTokenKind.INTEGER_LIT,
-            MLIRTokenKind.QUESTION,
-        ):
-            dim = self.parse_shape_dimension()
-            dims.append(dim)
+        dims = self.parse_dimension_list()
+        if len(dims) > 0:
```
superlopuh:
```suggestion
        if dims:
```

superlopuh: We generally avoid this check as it's potentially expensive

tests/test_parser.py
```diff
@@ -1207,3 +1207,18 @@ def test_parse_identifier_or_str_literal(input: str, expected: str | None):
             parser.parse_identifier_or_str_literal()
     else:
         assert parser.parse_identifier_or_str_literal() == expected
+
+
+@pytest.mark.parametrize(
+    "input, expected",
+    [
+        ("2x3x4", [2, 3, 4]),
```
superlopuh: Can you please add the empty case here?
```suggestion
    [
        ("", []),
```

superlopuh: Oh and probably `"x"`

xdsl/parser/attribute_parser.py
```diff
@@ -485,6 +485,21 @@ def parse_shape_dimension(self, allow_dynamic: bool = True) -> int:
...
+    def parse_optional_shape_delimiter(self) -> bool:
+        """
+        Parse 'x', a shape delimiter. Note that if 'x' is followed by other
+        characters, it will split the token. For instance, 'x1' will be split
+        into 'x' and '1'.
+        """
+        if self._current_token.kind != MLIRTokenKind.BARE_IDENT:
+            return False
+
+        if self._current_token.text[0] != "x":
```
superlopuh, line 497:
```suggestion
        if self._current_token.text != "x":
```

superlopuh: Ah no ignore me, `xf32` is a valid identifier

xdsl/parser/attribute_parser.py
```diff
@@ -504,20 +519,46 @@ def parse_shape_delimiter(self) -> None:
...
+        dims: list[int] = []
+        accepted_token_kinds = (MLIRTokenKind.INTEGER_LIT, MLIRTokenKind.QUESTION)
+
+        # empty case
+        if self._current_token.kind not in accepted_token_kinds:
+            return []
+
+        # parse first number
+        dim = self.parse_shape_dimension()
+        dims.append(dim)
```
superlopuh, line 537:
```suggestion
        accepted_token_kinds = (MLIRTokenKind.INTEGER_LIT, MLIRTokenKind.QUESTION)

        # empty case
        if (first_dim := self.parse_optional_shape_dimension) is None:
            return []

        # parse first number
        dims = [first_dim]
```

xdsl/parser/attribute_parser.py
```diff
@@ -504,20 +519,46 @@ def parse_shape_delimiter(self) -> None:
...
+
+        # Two approaches possible here, could look for `x` `num` and only parse
+        # when both are found (requires peek() functionality), or could parse
+        # the `x` always, and then `undo` if we don't see a number following
+        # which is this approach
```
superlopuh: I like this approach, let's remove this commend and add one below to make it clear whey we're resuming from a different position

tests/test_printer.py
```diff
@@ -954,6 +954,22 @@ def _test_attr_print(expected: str, attr: FloatAttr):
     _test_attr_print("0xfff0000000000000 : f64", FloatAttr(float("-inf"), 64))
 
 
+@pytest.mark.parametrize(
+    "dims, expected",
+    [
+        ([1, 2, 3], "1x2x3"),
```
superlopuh:
```suggestion
    [
        ([], ""),
        ([1, 2, 3], "1x2x3"),
```

xdsl/parser/attribute_parser.py
jakedves: Could do this, but it requires making a `parse_optional_shape_dimension`. This is fine but then if we overwrite `parse_shape_dimension` to use this with `expect()` we will lose some specific error information

xdsl/parser/attribute_parser.py
jakedves: This needs to be `text[0]` to handle splitting of a token like `5xi32` which will initially be tokenised as `5` `xi32`

xdsl/parser/attribute_parser.py
```diff
@@ -485,24 +485,66 @@ def parse_shape_dimension(self, allow_dynamic: bool = True) -> int:
...
+        """
+        token = self._current_token
+        tk = token.kind
+
+        err_val = tk.name if tk != MLIRTokenKind.BARE_IDENT else token.text
+
+        self.expect(
+            self._parse_optional_shape_delimiter,
+            f"Expected 'x' in shape delimiter, got {err_val}",
+        )
```
superlopuh: Let's build the error string only in the failing case, it tends to be annoyingly expensive, unfortunately
```suggestion
		if self._parse_optional_shape_delimiter() is None:
		    return
		
        token = self._current_token
        tk = token.kind

        err_val = tk.name if tk != MLIRTokenKind.BARE_IDENT else token.text

        self.raise(
            f"Expected 'x' in shape delimiter, got {err_val}",
        )
```

xdsl/parser/attribute_parser.py
superlopuh: Ah OK in that case let's keep it as-is

superlopuh (approved):
Thank you! Would be good to get a tick from either @alexarice or @math-fehr before merging, just to make sure they agree with me on the approach

PR #5245, opened 2025-09-17
documentation: fix `__init__` handling in docs generation

jumerckx:


scripts/gen_ref_pages.py
```diff
@@ -35,6 +36,9 @@ def gen_reference():
 
         if parts[-1] == "__init__":
             parts = parts[:-1]
+            if not parts:
+                # skip the root __init__.py
+                continue
```
jumerckx, line 41: This ignores the root `xdsl/__init__.py` so I believe no documentation is being generated for it:
```
class LazyVersion:
    """
    Resolving the version dynamically is slow, hence this lazy wrapper to get the version
    only when needed.
    """

    _version: str | None

    def __init__(self):
        self._version = None

    def __str__(self):
        if self._version is None:
            from importlib.metadata import PackageNotFoundError, version

            try:
                self._version = version("xdsl")
            except PackageNotFoundError:
                # package is not installed
                self._version = "Error: xDSL not installed"

        return self._version


__version__ = LazyVersion()

```

superlopuh (approved):
Thank you very much!

PR #5246, opened 2025-09-18
[wip] gui: improve pass argument screen

dshaaban01:
This PR updates the UI for providing arguments to passes. Previously, arguments were entered as a single inline string, which was error-prone and unfriendly. With this change, each argument now has its own labeled input field, types (int, str, etc.) are clearly indicated as placeholders in the input boxes (and don't need to be deleted/replaced which was how it was previously done), and default values are provided (if they exist).

The new layout makes it easier to understand what values are expected and reduces the risk of formatting errors.

###Old:
image

### New:
**Example of a pass with no default values.**
image
**Example of a pass with default values.**
image

TODO:
- Center the argument screen (textual being weird)
- Fix failing test


superlopuh:
Thanks for opening this! Will try to play with this before too long

PR #5249, opened 2025-09-18
dialects: (csl) Add assembly_format for variables

dk949:
This aims to make `csl.variable`, `csl.load_var` and `csl.store_var` a bit easier to read.

The custom assembly is not based on anything upstream, suggestions are welcome.

### DIFF
--- a/xdsl/dialects/csl/csl.py
+++ b/xdsl/dialects/csl/csl.py
@@ -433,7 +433,7 @@ class VariableOp(IRDLOperation):
     default = opt_prop_def(ParamAttr)
     res = result_def(VarType)
 
-    assembly_format = "`(`($default^)?`)` `:` type($res) attr-dict"
+    assembly_format = "`(` $default `)` `:` type($res) attr-dict"
 
     def get_element_type(self):
         assert isinstance(self.res.type, VarType)


xdsl/dialects/csl/csl.py
```diff
@@ -433,6 +433,8 @@ class VariableOp(IRDLOperation):
     default = opt_prop_def(ParamAttr)
     res = result_def(VarType)
 
+    assembly_format = "`(`($default^)?`)` `:` type($res) attr-dict"
```
alexarice:
```suggestion
    assembly_format = "`(` $default `)` `:` type($res) attr-dict"
```
does this work?

xdsl/dialects/csl/csl.py
```diff
@@ -505,6 +509,10 @@ class StoreVarOp(IRDLOperation):
 
     traits = traits_def(MemoryWriteEffect())
 
+    assembly_format = (
+        "$var `:` type($var) `=` $new_value `:` type($new_value) attr-dict"
```
alexarice, line 513: I'm not a huge fan of using `=` in assembly format but I know other ops do it.

dk949: I agree, but it seemed like the most sensible choice for a scalar store/assignment operation :shrug:

PR #5250, opened 2025-09-18
core: Add directive for `custom<DImensionList>` in assembly format

jakedves:
There is a custom directive used by the mesh/shard dialect that would be useful to support in the `assembly_format`. This PR implements the `DimensionList(CustomDirective)` class

### DIFF
--- a/tests/test_dialect_utils.py
+++ b/tests/test_dialect_utils.py
@@ -12,13 +12,15 @@ from xdsl.dialects.utils import (
     parse_dynamic_index_with_type,
     parse_dynamic_index_without_type,
     print_dynamic_index_list,
+    parse_empty_dimension_list_directive,
+    print_empty_dimension_list_directive,
     split_dynamic_index_list,
 )
 from xdsl.dialects.utils.dynamic_index_list import verify_dynamic_index_list
 from xdsl.ir import Dialect, SSAValue
 from xdsl.parser import Parser, UnresolvedOperand
 from xdsl.printer import Printer
-from xdsl.utils.exceptions import VerifyException
+from xdsl.utils.exceptions import ParseError, VerifyException
 from xdsl.utils.test_value import create_ssa_value
 
 ctx = Context()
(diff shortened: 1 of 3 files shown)


superlopuh (commented):
Are there any operations that use this directive already? Could you use it in one of those or add it first before adding this change?

superlopuh (commented):
Oh I thought you meant that this existed for upstream MLIR dialects. Where is `mesh` from?

jakedves:
It's been renamed to shard in the documentation but does exist as mesh in the current release of MLIR

https://github.com/llvm/llvm-project/blob/llvmorg-20.1.7/mlir/test/Dialect/Mesh/ops.mlir

superlopuh:
I think it would make most sense to add https://mlir.llvm.org/docs/Dialects/Shard/#shardgrid-shardgridop first then look at mesh, if that works for you?

superlopuh:
Ooh

jakedves:
I think this is the same op, just renamed

tests/filecheck/dialects/mesh/ops.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: XDSL_ROUNDTRIP
+
+mesh.mesh @mesh0(shape = 2x2x4)
+mesh.mesh @mesh1(shape = 4x?)
+mesh.mesh @mesh2(shape = ?x4)
+mesh.mesh @mesh3(shape = ?x?)
+mesh.mesh @mesh4(shape = 3)
+mesh.mesh @mesh5(shape = ?)
```
superlopuh, line 8: I don't see any empty lists here. Are they supported by the MLIR version? If not, should our version support it? It feels inconsistent that the empty list is `[]` and non-empty is not delimited...

jakedves: My understanding is that yes this syntax is supported in [mlir for general dimension lists](https://github.com/llvm/llvm-project/blob/2c3453afd2d28d1a4d802af071c15b8620896ed3/mlir/test/IR/custom-print-parse.mlir#L5) but this operation in particular doesn't allow for an empty dimension list, so added the verify and [`ops_invalid`](https://github.com/xdslproject/xdsl/blob/d183c8026101bf575355a8e612eb383668e29c66/tests/filecheck/dialects/mesh/ops_invalid.mlir) tests cases for this

superlopuh: OK at least we're consistent with MLIR 😅

superlopuh (approved):

PR #5254, opened 2025-09-19
core: add `Folder` class

jumerckx:
Instead of forcing a context field into Builder, move folding in its own class with methods `try_fold`, `insert_with_fold`, and `replace_with_fold`.
Will add tests in a bit.


xdsl/folder.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+
+        if not isinstance(op, HasFolderInterface):
+            return None
+        folded = op.fold()
+        if folded is None:
+            return None
+        results: list[SSAValue] = []
+        new_ops: list[Operation] = []
+        for val, original_result in zip(folded, op.results):
```
superlopuh:
```suggestion
        for val, original_result in zip(folded, op.results, strict=True):
```

xdsl/folder.py
```diff
@@ -0,0 +1,80 @@
...
+            if isinstance(val, SSAValue):
+                results.append(val)
+            else:
+                assert isinstance(val, Attribute)
+                dialect = self.context.get_dialect(op.dialect_name())
+                interface = dialect.get_interface(ConstantMaterializationInterface)
+                if not interface:
```
superlopuh:
```suggestion
                if interface is None:
```

xdsl/folder.py
```diff
@@ -0,0 +1,80 @@
...
+        Otherwise, returns None.
+        """
+        results = self.try_fold(op)
+        if results is not None:
+            values, new_ops = results
+            builder.insert_op(new_ops)
+            return values, new_ops
+        else:
+            builder.insert(op)
+            return op.results, list[Operation]()
```
superlopuh: Feels a bit weird to me to have a `not` in an if statement where both branches are kind of "success" conditions. Also not sure why not return the list literal
```suggestion
        if results is None:
            builder.insert(op)
            return op.results, []
        else:
            values, new_ops = results
            builder.insert_op(new_ops)
            return values, new_ops
```

xdsl/folder.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def replace_with_fold(
+        self, op: Operation, safe_erase: bool = True
+    ) -> tuple[list[SSAValue], list[Operation]] | None:
```
superlopuh: Feels a little weird not to pass a PatternRewriter here, it might be important for the listener to be notified of the replacement

xdsl/folder.py
jumerckx: the list literal conflicts with the `Sequence[SSAValue]` which is there because op.results isn't a list.
I'm still not entirely sure we actually need new_ops to be returned here though...

superlopuh: Oh yeah I would not return them

tests/test_folder.py
```diff
@@ -0,0 +1,129 @@
...
+    ctx = Context()
+    from xdsl.dialects import arith
+
+    ctx.load_dialect(arith.Arith)
+
+    # Create constants: 0 and 5
+    one_const = ConstantOp.from_int_and_width(1, i32)
+    five_const = ConstantOp.from_int_and_width(5, i32)
+
+    # Create an AddiOp: %result = arith.addi %zero, %five : i32
```
superlopuh: Is this a TODO?

tests/test_folder.py
```diff
@@ -0,0 +1,129 @@
...
+    builder.insert(five_const)
+
+    # Create an AddiOp: %result = arith.addi %zero, %five : i32
+    # Adding zero should fold to just the other operand
+    addi_op = AddiOp(one_const.result, five_const.result)
+    builder.insert(addi_op)
+
+    # Replace with fold
+    folder = Folder(ctx)
+    folded_values = folder.insert_with_fold(addi_op, builder)
```
superlopuh: not replace?

xdsl/folder.py
```diff
@@ -0,0 +1,77 @@
...
+from xdsl.interfaces import HasFolderInterface
+from xdsl.ir import Attribute, Operation, SSAValue, TypeAttribute
+from xdsl.pattern_rewriter import PatternRewriter
+
+
+@dataclass
+class Folder:
+    context: Context
+
+    def try_fold(self, op: Operation) -> tuple[list[SSAValue], list[Operation]] | None:
```
superlopuh, line 16: I thought folding could only produce one operation?

jumerckx: The docs aren't super clear, but the fold method can return a constant attribute for each of the operation's results. I think confusion stems from:
```
`fold` has the restriction that no new operations may be created, and only the
root operation may be replaced (but not erased). It allows for updating an
operation in-place, or returning a set of pre-existing values (or attributes) to
replace the operation with.
```
(https://mlir.llvm.org/docs/Canonicalization/#canonicalizing-with-the-fold-method)

tests/test_folder.py
```diff
@@ -0,0 +1,129 @@
...
+    from xdsl.dialects import arith
+
+    ctx.load_dialect(arith.Arith)
+
+    # Create constants: 0 and 5
+    one_const = ConstantOp.from_int_and_width(1, i32)
+    five_const = ConstantOp.from_int_and_width(5, i32)
+
+    # Create an AddiOp: %result = arith.addi %zero, %five : i32
+    # Adding zero should fold to just the other operand
```
math-fehr: The comment mentions zero, but you add one

tests/test_folder.py
```diff
@@ -0,0 +1,129 @@
...
+from xdsl.dialects.arith import AddiOp, ConstantOp
+from xdsl.dialects.builtin import IntegerAttr, i32
+from xdsl.dialects.test import TestOp
+from xdsl.folder import Folder
+from xdsl.ir import Block, OpResult
+from xdsl.rewriter import InsertPoint
+
+
+def test_try_fold_foldable_operation():
+    """Test that try_fold correctly folds an AddiOp with zero."""
```
math-fehr:
```suggestion
    """Test that try_fold correctly folds an AddiOp with two constants."""
```

tests/test_folder.py
```diff
@@ -0,0 +1,129 @@
...
+    assert result is not None
+    values, new_ops = result
+    assert len(values) == 1
+    assert len(new_ops) == 1
+    assert isinstance(new_op := new_ops[0], ConstantOp)
+    assert isinstance(new_op.value, IntegerAttr)
+    assert new_op.value.value.data == 6
+
+
+def test_insert_with_fold():
```
math-fehr, line 41: Can you update the documentation / function names?
This check for `insert_and_fold`, while the other was checking for `try_fold`

tests/test_folder.py
math-fehr: I don't think we should allow an `insert_and_fold` with something that is already inserted.

xdsl/folder.py
math-fehr: Yes, so the second part of the tuple should be an `Operation | None` probably?

jumerckx: I might be misunderstanding, but this is `try_fold` not `fold`. `try_fold` takes the attributes and creates constant operations from them, so there can be multiple.

jumerckx:
Sorry about the messy/wrong docstrings, should be better now.

tests/test_folder.py
jumerckx: Oh. This code should've been `replace_with_fold` but I've now added an exception in `insert_with_fold` and added an extra test.

xdsl/folder.py
superlopuh: Ah that makes sense. Feels like it belongs in a comment, maybe in the docstring below?

xdsl/folder.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Try to fold the given operation.
+        Returns a tuple the list of SSAValues that replace the results of the operation,
+        and a list of new operations that were created during folding.
```
math-fehr:
```suggestion
        and a list of new operations that were created during folding (which
        includes created constant operations).
```

tests/test_folder.py
superlopuh: Just a little reminder about this comment, should be removed? Did you want to add another test? Seems like it would be worth adding tests for folding with left 0 and right 0

superlopuh: Ah sorry I just realised that this is a test for the folder not arith, really. Then let's just get rid of the comment?

xdsl/folder.py
jumerckx: Currently the list of new operations consists of **only** the created constant operations.

While MLIR allows mutating the operation being folded in `fold`, [we don't](https://github.com/xdslproject/xdsl/pull/5232#discussion_r2347381608). This means the fold method in xDSL is slightly more restrictive. Perhaps `fold` should be able to return an `Operation` to signal that `try_fold` should replace it?

tests/test_folder.py
jumerckx: I'm tired, seems like I missed some of those comments. Now they should be GONE!

xdsl/folder.py
math-fehr: Oh I see, then yes maybe let's change the documentation to clearly say only constant operations are created.

PR #5258, opened 2025-09-19
interpreter: (arith) Add support for arith.cmpf

positr0nium:
Title says it all - seemed to be missing.

### DIFF
--- a/tests/interpreters/test_arith_interpreter.py
+++ b/tests/interpreters/test_arith_interpreter.py
@@ -239,14 +239,6 @@ def test_cmpi(
     assert ret[0] == fn(lhs_value, rhs_value)
 
 
-def o(x: float, y: float) -> bool:
-    return not isnan(x) and not isnan(y)
-
-
-def u(x: float, y: float) -> bool:
-    return isnan(x) or isnan(y)
-
-
 @pytest.mark.parametrize("lhs_value", [1.5, 0.5, -1.5, 127.5, float("nan")])
 @pytest.mark.parametrize("rhs_value", [1.5, 0.5, -1.5, 127.5, float("nan")])
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


positr0nium:
Not sure if the cases are assigned correctly, will figure it out in the next days :)

superlopuh:
Here are the cmpf comparison operators: https://github.com/xdslproject/xdsl/blob/main/xdsl/dialects/arith.py#L86-L103

superlopuh:
I'll mark this as draft in the meantime. I would recommend running `make tests` locally before opening PRs, I find it easier to troubleshoot over Zulip before opening the PR than in GH, but that's more of a nit.

superlopuh:
Have you run `make precommit-install` work on your machine?

tests/filecheck/dialects/irdl/pyrdl-to-irdl/cmath-conversion.py
```diff
@@ -1,6 +1,7 @@
 # RUN: python %s | filecheck %s
 
 from xdsl.dialects.cmath import Cmath
+
```
superlopuh: seems like a spurious change

tests/interpreters/test_arith_interpreter.py
```diff
@@ -238,6 +239,51 @@ def test_cmpi(
     assert ret[0] == fn(lhs_value, rhs_value)
 
 
+def o(x, y):
```
superlopuh: We have a fully-typed project, can you please add type annotations everywhere?

superlopuh: Also, I would advise running `make tests` locally before opening PRs, and pinging on the Zulip if there are any unexpected errors, I find that it's an easier place to iterate.

xdsl/interpreters/arith.py
```diff
@@ -161,6 +161,55 @@ def run_cmpi(self, interpreter: Interpreter, op: arith.CmpiOp, args: PythonValue
...
+            return isnan(x) or isnan(y)
+
+        x = args[0]
+        y = args[1]
+
+        match op.predicate.value.data:
+            case 0:
+                return (False,)
+            case 1:
+                return ((x == y) and o(x, y),)
```
superlopuh: nit: let's avoid extra functions
```suggestion
        x = args[0]
        y = args[1]
        
        nan = isnan(x) or isnan(y)

        match op.predicate.value.data:
            case 0:
                return (False,)
            case 1:
                return ((x == y) and not nan,)
```

tests/interpreters/test_arith_interpreter.py
```diff
@@ -238,6 +239,85 @@ def test_cmpi(
     assert ret[0] == fn(lhs_value, rhs_value)
 
 
+def o(x: float, y: float) -> bool:
+    return not isnan(x) and not isnan(y)
+
+
+def u(x: float, y: float) -> bool:
+    return isnan(x) or isnan(y)
```
superlopuh: let's inline these below? might as well just have one value for the nan comparison, no need for helper functions IMO

superlopuh (approved):
Thank you!

PR #5259, opened 2025-09-19
add canonicalization pattern x&x->x

khorcarol:


superlopuh (commented):
Looks great! Can you please suggest a new cannicalization pattern for the next person?

docs/index.md
```diff
@@ -64,7 +64,7 @@ The first step is finding a missing optimisation pattern.
 You're welcome to come up with your own, or do one of the following:
 
 - `x + x * n -> x * (n + 1)` (where `n` is an immediate)
-- `x & x = x`
+- `x / x = 1` where x != 0
```
superlopuh: I'm not sure this will work, since either x is a constant, in which case we would want to pre-compute the result, and if it's not a constant we don't know if it's 0 or not. Maybe another logical operator?

docs/index.md
```diff
@@ -64,7 +64,7 @@ The first step is finding a missing optimisation pattern.
 You're welcome to come up with your own, or do one of the following:
 
 - `x + x * n -> x * (n + 1)` (where `n` is an immediate)
-- `x & x = x`
+- `x & 1 -> x`
```
superlopuh: This one is also tricky, since bitwise and should actually be with all ones, and not just `1`, and immediate values in riscv are not 32-bit. How about this:
```suggestion
- `n & m -> k` (where `n`, `m`, and `k` are immediates, and k == n & m)
```

khorcarol: Right, that makes sense! Thanks!

PR #5260, opened 2025-09-21
misc: fix cross-platform llvm-mca call

superlopuh:
Doesn't work on my ARM mac.

I propose hard-coding it for now and making it parametrizable in the future.

### DIFF
--- a/xdsl/backend/block_throughput_cost_model.py
+++ b/xdsl/backend/block_throughput_cost_model.py
@@ -108,7 +108,12 @@ class MCABlockThroughputCostModel(ExternalBlockThroughputCostModel):
         return "llvm-mca"
 
     def cmd(self) -> list[str]:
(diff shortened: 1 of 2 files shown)


xdsl/backend/block_throughput_cost_model.py
```diff
@@ -108,7 +108,12 @@ def tool_name(self) -> str:
         return "llvm-mca"
 
     def cmd(self) -> list[str]:
-        return [self.tool_name(), f"-mcpu={self.arch}", self.src_path]
+        return [
+            self.tool_name(),
+            "-mtriple=x86_64-unknown-linux-gnu",
```
qaco: So if we are on mac we have mtriple=linuxx86 and mcpu=neon ?

superlopuh: Yeah. Actually I'll add this as a parameter and make it less hacky, one sec.

PR #5261, opened 2025-09-21
backend: (riscv) make RISC-V assembly printing accept any register type

superlopuh:
Another step in the journey towards unifying the register printing infrastructure across backends. Only LabelAttr to go in terms of riscv-specific things.

### DIFF
--- a/xdsl/dialects/riscv.py
+++ b/xdsl/dialects/riscv.py
@@ -6,7 +6,7 @@ from collections.abc import Set as AbstractSet
 from io import StringIO
 from typing import IO, Annotated, Generic, Literal, TypeAlias
 
-from typing_extensions import Self, TypeVar, assert_never
+from typing_extensions import Self, TypeVar
 
 from xdsl.backend.assembly_printer import (
     AssemblyPrintable,
@@ -65,7 +65,6 @@ from xdsl.traits import (
     Pure,
 )
 from xdsl.utils.exceptions import VerifyException
-from xdsl.utils.hints import isa
 
 
 def is_non_zero(reg: IntRegisterType) -> bool:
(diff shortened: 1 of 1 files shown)


superlopuh:
This is separate from verification, and only affects assembly printing. Verification should be done by the compiler, not printer, IMO.

xdsl/dialects/riscv.py
```diff
@@ -447,7 +446,7 @@ def print_op_type(self, printer: Printer) -> None:
 
 
 AssemblyInstructionArg: TypeAlias = (
-    IntegerAttr | LabelAttr | SSAValue | IntRegisterType | str
+    IntegerAttr | LabelAttr | SSAValue | RegisterType | str
```
alexarice, line 449: Right, is this only used in this file? (Just want to check that this isn't used in an `operand_def` or something)

xdsl/dialects/riscv.py
```diff
@@ -496,28 +495,18 @@ def assembly_line(self) -> str | None:
 
 
 def _assembly_arg_str(arg: AssemblyInstructionArg) -> str:
-    if isa(arg, IntegerAttr):
+    if isinstance(arg, SSAValue):
+        if not isinstance(t := arg.type, RegisterType):
+            raise ValueError(f"Unexpected register type {t}")
+        return t.register_name.data
+    elif isinstance(arg, IntegerAttr):
         return f"{arg.value.data}"
```
alexarice, line 503: unrelated to this PR, but we should really consider using `printer.print_int` here if that's an option I feel

superlopuh: Yeah defo I just want to get there with smaller changes

xdsl/dialects/riscv.py
superlopuh: Yep, and note that it can't be used in an operand_def since it includes SSAValue and str

PR #5263, opened 2025-09-22
ci: (Nix) add new workflow

alexarice:
Attempt to add a simple nix ci which enters a nix shell and runs 'make tests'

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -1,0 +1,23 @@
+# This workflow enter a nix shell and run the main test suite.
+# It's primarily used to check that the nix environment works
+
+name: CI - Nix based testing
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


alexarice:
I'm sure there are more clever things you could do with caching things, but it feels like they would be minimal gain.

superlopuh:
Very nice! I'm not sure that the gain would be minimal, but I'm happy to do that later.

superlopuh:
The nice thing about this is that it should catch the `test` dialect not being included in brew failures.

alexarice:
I more meant that I can't see that caching would gain much as neither nix or uv is building anything I think (Just downloading stuff). I guess it would save on the nix evaluation time, but that seems minimal to the time to download mlir.

Perhaps I'm misunderstanding github action caching and it's actually far faster to reload the cache than redownload things.

superlopuh:
Yeah obviously caching is more important if you need to build things rather than just download, but seems like a win still to have it cached if possible.

PR #5270, opened 2025-09-23
misc: allow shadowing in ScopedDict

superlopuh:
I added this check for when I added it ito contain the SSA to value mapping in the interpreter, where assigning to an SSA value twice is a bug, but it turns out that there are lots more uses for a scoped dictionary where shadowing is allowed. We'll need to just be careful when using it not to use it incorrectly, but this seems like a better tradeoff.


jumerckx:
Just out of interest, why not have a field `allow_shadow`? Purely for performance reasons?

superlopuh:
Yeah I guess, we can add it back later if we feel like it, but it feels like it's a bit overprotective for now

ed741:
When parsing an AST it used to be possible to detect a variable name re-declaration error by trying to insert the variable name into the ScopedDict and it would raise an error if it was already locally defined which could be propagated up. With this update it is now impossible (without accessing local variables) to even check if the variable name is free (not in _local_scope), and so redefinition errors in input programs can not be as easily checked. 
@superlopuh

jumerckx:
Perhaps `_local_scope` should just be made public?

xdsl/utils/scoped_dict.py
```diff
@@ -69,10 +68,6 @@ def __setitem__(self, key: _Key, value: _Value):
         Assign key to current scope. Raises InterpretationError if key already
         assigned to.
         """
```
ed741, line 70: This doc string is wrong (even more so now, since no exceptions will be raised)

ed741:
> Perhaps `_local_scope` should just be made public?

IMO it's best not to expose that since it might get messy letting users write directly into `_local_scope` or replace it entirely.
A `contains_local(self, key: _Key) -> bool` method would give back all the functionality without exposing anything extra.

PR #5273, opened 2025-09-24
core: Update `DYNAMIC_INDEX` to match C++ MLIR

jakedves:
In C++ MLIR dynamic indices are specified by kDynamicShape which represents the smallest integer in `i64`: `-(2 ** 63)`. This PR updates xDSL to use this value as it is required for printing some ops in generic format, e.g.:

```
mesh.mesh @mesh0(shape = ?x2)
```

ran with `mlir-opt --mlir-print-op-generic` becomes:

```
"builtin.module"() ({
  "mesh.mesh"() <{shape = array<i64: -9223372036854775808, 2>, sym_name = "mesh0"}> : () -> ()
}) : () -> ()
```

Meaning this is required for generic MLIR compatibility. The previous code often used `-1` as a value instead of `DYNAMIC_INDEX` so this also updates that, but it's possible that some untested code isn't updated in this PR.

### DIFF
--- a/xdsl/dialects/builtin.py
+++ b/xdsl/dialects/builtin.py
@@ -101,9 +101,10 @@ if TYPE_CHECKING:
     from xdsl.printer import Printer
 
 
-DYNAMIC_INDEX = -1
+DYNAMIC_INDEX = -9_223_372_036_854_775_808
 """
 A constant value denoting a dynamic index in a shape.
+Equal to -(2 ** 63) which is used in C++ MLIR.
 """
 
 
--- a/tests/dialects/test_builtin.py
+++ b/tests/dialects/test_builtin.py
@@ -8,6 +8,7 @@ from typing_extensions import TypeVar
 
 from xdsl.dialects.arith import ConstantOp
 from xdsl.dialects.builtin import (
+    DYNAMIC_INDEX,
     AnyFloat,
     ArrayAttr,
     ArrayOfConstraint,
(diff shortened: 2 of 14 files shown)


xdsl/dialects/builtin.py
```diff
@@ -101,9 +101,10 @@
     from xdsl.printer import Printer
 
 
-DYNAMIC_INDEX = -1
+DYNAMIC_INDEX = -9_223_372_036_854_775_808
```
superlopuh: Seems a little neater:
```suggestion
DYNAMIC_INDEX = -(2 ** 63)
```

superlopuh (approved):
Nice, do you know -1 worked at all?

superlopuh:
Is it because we normalised it when re-parsing the MLIR generic format?

jakedves:
Previously xDSL didn't have any ops that would ever print `-1` even in generic format, so MLIR would parse and handle `?` on it's own, e.g. generic memref:
```
%0 = "arith.constant"() <{value = 4 : index}> : () -> index
%1 = "memref.alloc"(%0) <{operandSegmentSizes = array<i32: 1, 0>}> : (index) -> memref<?x12xi32>
```
It's fine when it's still a question mark, but for some ops it's used in the attribute as the actual value, which isn't compatible as MLIR doesn't expect `-1`, e.g. this would crash:
```
"mesh.mesh"() <{shape = array<i64: -1, 2>, sym_name = "mesh0"}> : () -> ()
```

superlopuh:
I see. Great, let's merge

PR #5277, opened 2025-09-24
dialects: (x86_scf) ForRofOperation: RegisterAllocatableOperation

superlopuh:
Now supports register allocation. Mostly a copy/paste of riscv_scf implementation. Makes me think that this should actually go into an `asm` dialect, to share between backends.

### DIFF
--- a/xdsl/dialects/x86_scf.py
+++ b/xdsl/dialects/x86_scf.py
@@ -1,10 +1,14 @@
 from __future__ import annotations
 
 from abc import ABC
-from collections.abc import Sequence
+from collections.abc import Generator, Sequence
+from typing import cast
 
 from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


xdsl/dialects/x86_scf.py
```diff
@@ -112,6 +116,52 @@ def verify_(self):
                         f"variables types."
                     )
 
+    def iter_used_registers(self) -> Generator[RegisterType, None, None]:
+        # We know that all the registers for the inputs and outputs are the same, and
+        # that these registers will have been iterated earlier in the IR.
+        yield from ()
+
+    def allocate_registers(self, allocator: BlockAllocator) -> None:
```
qaco, line 124: Neat !

PR #5287, opened 2025-09-28
core: (assembly-format) Add optional anchor support for `Custom<DynamicIndexList>`

jakedves:
Adds support for `custom<DynamicIndexList>` to be used as an aqchorable directive, as in C++ MLIR. This lets it be used in optional clauses, which [some MLIR operations](https://github.com/llvm/llvm-project/blob/6146a88f60492b520a36f8f8f3231e15f3cc6082/mlir/include/mlir/Dialect/Mesh/IR/MeshOps.td#L305) use.

e.g. you can have assembly formats as the following:

```assembly_format = (`keyword` custom<DynamicIndexList>($dynamics, $statics)^)?```

Not super familiar with this code so open to feedback

### DIFF
--- a/xdsl/dialects/utils/dynamic_index_list.py
+++ b/xdsl/dialects/utils/dynamic_index_list.py
@@ -259,9 +259,7 @@ class DynamicIndexList(CustomDirective):
         dynamic_empty = len(self.dynamic_position.get(op)) == 0
 
         static_vals = self.static_position.get(op)
-        static_empty = (
-            static_vals == DenseArrayBase.from_list(i64, []) or static_vals is None
-        )
+        static_empty = static_vals == DenseArrayBase.from_list(i64, [])
 
         return not (dynamic_empty and static_empty)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: xdsl/dialects/utils/dynamic_index_list.py [hunk 1]
```diff
@@ -248,3 +248,18 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
         print_dynamic_index_list(
             printer, self.DYNAMIC_INDEX, dynamic, static.get_values()
         )
+
+    def set_empty(self, state: ParsingState):
+        self.dynamic_position.set(state, [])
+        self.static_position.set(state, DenseArrayBase.from_list(i64, []))
+
+    def is_present(self, op: IRDLOperation) -> bool:
+        dynamic_empty = len(self.dynamic_position.get(op)) == 0
```
superlopuh:
```suggestion
        dynamic_empty =not self.dynamic_position.get(op))
```

xdsl/dialects/utils/dynamic_index_list.py
```diff
@@ -248,3 +248,18 @@ def print(self, printer: Printer, state: PrintingState, op: IRDLOperation) -> No
...
+
+        static_vals = self.static_position.get(op)
+        static_empty = static_vals == DenseArrayBase.from_list(i64, [])
```
superlopuh:
```suggestion
        static_empty = not static_vals.data
```

PR #5292, opened 2025-09-30
dialects: (mesh) Implement `mesh.sharding` 

jakedves:
`mesh.sharding` is the main operation used in the mesh dialect. Implemented with xDSL roundtrip, MLIR conversion, and simple verify test cases.

Couldn't get the parsing to infer `i16` from untyped array syntax but don't think the workaround is too horrible for the syntax it wanted parsed/printed.

### DIFF
--- a/xdsl/dialects/mesh.py
+++ b/xdsl/dialects/mesh.py
@@ -11,15 +11,19 @@ from xdsl.dialects.builtin import (
     BytesAttr,
     DenseArrayBase,
     FlatSymbolRefAttr,
+    IntegerAttr,
     SymbolNameConstraint,
+    i16,
     i64,
 )
-from xdsl.dialects.utils.dimension_list import DimensionList
+from xdsl.dialects.utils import DimensionList, DynamicIndexList
 from xdsl.ir import (
     Attribute,
     Dialect,
     EnumAttribute,
+    OpaqueSyntaxAttribute,
     ParametrizedAttribute,
+    SpacedOpaqueSyntaxAttribute,
     TypeAttribute,
     VerifyException,
 )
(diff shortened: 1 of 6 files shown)


xdsl/dialects/mesh.py
```diff
@@ -1,17 +1,113 @@
...
+    def print_parameters(self, printer: Printer) -> None:
+        def print_sublist(sublist: MeshAxesAttr):
+            with printer.in_square_brackets():
+                printer.print_list(sublist.get_values(), printer.print_int)
+
+        with printer.in_square_brackets():
+            printer.print_list(
+                self.axes.data,
+                print_sublist,
+            )
```
superlopuh, line 117: Can you please add this one separately, adding a docstring for the class and an `attrs.mlir` filecheck file?

xdsl/dialects/mesh.py
```diff
@@ -1,17 +1,113 @@
...
+class MeshAxesArrayAttr(ParametrizedAttribute, OpaqueSyntaxAttribute):
+    name = "mesh.axisarray"
+
+    axes: ArrayAttr[MeshAxesAttr]
+
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
+        axes = parser.parse_comma_separated_list(
+            parser.Delimiter.SQUARE,
+            parser.parse_attribute,
```
superlopuh: feels like this could be parse_integer?

But could also abstract this out into a function that parses a DenseArrayBase[I16] and parse a list of those directly?
```python
                elements = parser.parse_comma_separated_list(
                    parser.Delimiter.SQUARE, parser.parse_integer
                )
            return DenseArrayBase(
                self.elt_type, BytesAttr(self.elt_type.pack(elements))
            )
```

PR #5293, opened 2025-09-30
dialects: (tosa) update `tosa.mul`, `tosa.matmul` for LLVM @ 20.1.7

jakedves:
This PR implements the first part of #4957 which is to update the `mlir-conversion` tests for the current target of MLIR. This also required updating the `tosa.mul` and `tosa.matmul` ops to match that version.


superlopuh:
I'm a litle concerned that this is going to conflict with my changes to TOSA in the LLVM 21 branch. I refactored the filecheck tests somewhat significantly, bringing them more in line with the MLIR tests, and I would like to go with that approach if possible. Do you think you could take a look at that branch and see how to similarly refactor the tests for these changes? Although alternatively it might be less work to finish the updates to that branch and avoid updating these ops twice.

jakedves:
Ah yes forgot about that, will give them a look and see how much I can use what's from already been done there

PR #5301, opened 2025-10-02
dialects (tosa): Update test cases with MLIR-styling and ops (`tosa.mul`, `tosa.matmul`) for MLIR 20.1.7

jakedves:
Updates any invalid tosa ops for MLIR version 20.1.7.

Also rewrites test cases to be MLIR-style.

Left in TODOs which are intended for MLIR v21


xdsl/dialects/tosa.py
```diff
@@ -258,13 +278,17 @@ class MatMulOp(IRDLOperation):
     def verify_(self) -> None:
         assert isinstance(self.a.type, ShapedType)
         assert isinstance(self.b.type, ShapedType)
-        assert isinstance(self.a_zp.type, ShapedType)
-        assert isinstance(self.b_zp.type, ShapedType)
+
+        # TODO: uncomment for MLIR v21
```
jakedves, line 282: @superlopuh not sure if you want things like this remaining, I can remove otherwise

superlopuh: This is fine! If the other PR weren't already open and in flight I would ask for an issue but here that seems overkill since we'll fix it soon anyway.

PR #5308, opened 2025-10-03
dialects: (pdl) add `isNegated` property to ApplyNativeConstraintOp

jumerckx:


xdsl/dialects/pdl.py
```diff
@@ -182,6 +183,9 @@ class ApplyNativeConstraintOp(IRDLOperation):
 
     name = "pdl.apply_native_constraint"
     constraint_name = prop_def(StringAttr, prop_name="name")
+    is_negated = prop_def(
+        BoolAttr, prop_name="isNegated", default_value=BoolAttr.from_bool(False)
+    )
```
superlopuh, line 188: Can you add a test, please? Also with a roundtrip with MLIR?

jumerckx: (I have to add a test with MLIR still)

jumerckx: In MLIR, `pdl.rewrite`'s region isn't optional:
```
let regions = (region AnyRegion:$bodyRegion);

let assemblyFormat = [{
    ($root^)? (`with` $name^ (`(` $externalArgs^ `:` type($externalArgs) `)`)?)?
              ($bodyRegion^)?
    attr-dict-with-keyword
  }];
```
But the assembly format for the region is an optional group meaning it shouldn't get printed when the region is empty (default case).
https://mlir.llvm.org/docs/DefiningDialects/Operations/#optional-groups
> When a non-variadic region anchors a group, the detector for printing the group is if the region is empty. 

This behaviour seems to be missing in xDSL?

(This is causing difficulties with an interop test)

superlopuh: We could add this somewhat easily, I can try adding it today.

superlopuh: Oh it seems to already be working for parsing, but not for printing, if I'm reading the code correctly. What happens when you use this format?

superlopuh: OK no I see it's a little weirder than this... I'm looking into it

superlopuh: Merged!

tests/filecheck/mlir-conversion/with-mlir/dialects/pdl/pdl_native_constraint.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: xdsl-opt --print-op-generic %s | mlir-opt --mlir-print-op-generic | xdsl-opt | filecheck %s
+// RUN: xdsl-opt %s | mlir-opt --mlir-print-op-generic | xdsl-opt | filecheck %s
```
superlopuh:
```suggestion
// RUN: MLIR_ROUNDTRIP
// RUN: MLIR_GENERIC_ROUNDTRIP
```

PR #5309, opened 2025-10-04
transformations: (convert-pdl-to-pdl-interp) Extract predicates from pattern

jumerckx:
In this pr, only patterns with a single root are supported.

### DIFF
--- a/tests/transforms/test_convert_pdl_to_pdl_interp.py
+++ b/tests/transforms/test_convert_pdl_to_pdl_interp.py
@@ -1,5 +1,7 @@
 from typing import cast
 
+import pytest
+
 from xdsl.builder import ImplicitBuilder
 from xdsl.dialects import pdl
 from xdsl.dialects.builtin import IntegerType, StringAttr, f32, i32
@@ -736,14 +738,17 @@ def test_non_tree_predicates_attribute_literal():
     assert attr_pos.value == expected_attr
 
 
-def test_constraint_simple():
+@pytest.mark.parametrize("is_negated", [True, False])
+def test_constraint_simple(is_negated: bool):
     """Test ApplyNativeConstraintOp basic functionality"""
     body = Region([Block()])
     block = body.first_block
     with ImplicitBuilder(block):
         root = pdl.OperationOp("op1").op
         # Simple constraint that doesn't create conflicts
-        pdl.ApplyNativeConstraintOp("simple_constraint", [root], []).res
+        pdl.ApplyNativeConstraintOp(
+            "simple_constraint", [root], [], is_negated=is_negated
+        ).res
         pdl.RewriteOp(None, name="rewrite")
 
     pattern = pdl.PatternOp(1, "pattern", body)
@@ -771,6 +776,7 @@ def test_constraint_simple():
     ).name == "simple_constraint"
     assert constraint_pred.a == TrueAnswer()
     assert len(q.arg_positions) == 1  # Should have the root operation as argument
+    assert q.is_negated is is_negated
 
 
 def test_result_op_non_tree():


xdsl/transforms/convert_pdl_to_pdl_interp/predicate.py
superlopuh: Let's do changes to this file as a first tiny PR, these should take tuples already IMO

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -0,0 +1,582 @@
...
+    """An op accepting a value at an optional index."""
+
+    parent: SSAValue
+    index: int | None
+
+
+class PatternAnalyzer:
+    """Analyzes PDL patterns and extracts predicates"""
+
+    def detect_roots(self, pattern: pdl.PatternOp) -> list[OpResult[pdl.OperationType]]:
```
superlopuh, line 37: After a PR in `predicate.py`, let's open a PR for just this function? Easier PRs will make it easier to iterate and I think we'll merge it faster that way.

xdsl/transforms/convert_pdl_to_pdl_interp/predicate.py
jumerckx: https://github.com/xdslproject/xdsl/pull/5314

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
jumerckx: https://github.com/xdslproject/xdsl/pull/5315

jumerckx:
I split off `_extract_type_predicates` https://github.com/xdslproject/xdsl/pull/5318
For the other methods in `PatternAnalyzer` is a bit harder as they recursively call each other...

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -38,6 +66,325 @@ def detect_roots(self, pattern: pdl.PatternOp) -> list[OpResult[pdl.OperationTyp
...
+        inputs: dict[SSAValue, Position],
+    ) -> list[PositionalPredicate]:
+        """Extract predicates for an operand or operand group"""
+        predicates: list[PositionalPredicate] = []
+
+        # Get the defining operation
+        defining_op = operand_value.owner
+        is_variadic = isinstance(operand_value.type, pdl.RangeType)
+
+        if isinstance(defining_op, pdl.OperandOp | pdl.OperandsOp):
```
superlopuh: let's do similar `match` statements as appropriate for the rest of the helpers?

superlopuh (approved):

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -66,3 +423,168 @@ def _extract_type_predicates(
...
+
+                    # Find the furthest position (deepest)
+                    furthest_pos = max(
+                        arg_positions, key=lambda p: p.get_operation_depth() if p else 0
+                    )
+
+                    # Create the constraint predicate
+                    result_types = tuple(r.type for r in op.res)
+                    # TODO: is_negated is not part of the dialect definition yet
+                    is_negated = False
```
superlopuh: We added it already, let's add the logic here as well?

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -66,3 +423,168 @@ def _extract_type_predicates(
...
+        best_root = rewriter.root if rewriter.root is not None else roots[0]
+
+        # Downward traversal from the best root
+        root_pos = OperationPosition(depth=0)
+        predicates.extend(
+            self.analyzer.extract_tree_predicates(best_root, root_pos, inputs)
+        )
+
+        predicates.extend(self.analyzer.extract_non_tree_predicates(pattern, inputs))
+        return predicates, best_root, inputs
```
superlopuh: Couldn't this be added in a separate PR?

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -1,16 +1,44 @@
...
+# Pattern Analysis
+# =============================================================================
+
+
+@dataclass(frozen=True)
+class OpIndex:
+    """An op accepting a value at an optional index."""
+
+    parent: SSAValue
+    index: int | None
```
superlopuh: is this used?

superlopuh:
Looking at the code coverage, it looks like there are a bunch of cases that aren't covered. Could you please generate some tests for those? It's a little annoying but I'm certain will help when debugging in the future.

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
jumerckx: https://github.com/xdslproject/xdsl/pull/5308
It's not merged yet due to the optional region assemblyformat

superlopuh: Ah I see, my bad, I'll try to fix that soon, happy to merge with TODO in the meantime

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
jumerckx: Not yet, moved out of this pr

jumerckx:
(Not quite there yet with the coverage)

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -66,3 +414,130 @@ def _extract_type_predicates(
                 pass
 
         return predicates
+
+    def extract_non_tree_predicates(
```
superlopuh, line 420: what calls this function?

jumerckx: It's called after all tree_predicates have been extracted and it tries to find extra predicates that aren't unearthed by traversing operation->operands->...

For example, a literal pdl.attribute could be defined in the matcher code and used in the rewrite.
`extract_non_tree_predicates` will process this operation even if it doesn't lead to a new predicate, because it still needs to map the pdl.attribute operation onto a position:
```py
case pdl.AttributeOp():
    if op.output not in inputs:
        if op.value:
            # Create literal position for constant attribute
            attr_pos = AttributeLiteralPosition(
                value=op.value, parent=None
            )
            inputs[op.output] = attr_pos
```

There's some fishy business in `case pdl.ApplyNativeConstraintOp()` which I'm looking at now and trying to get coverage for.

jumerckx: > what calls this function?

But so it will be called in `PredicateTreeBuilder` in a follow up pr.

jumerckx: > There's some fishy business in case pdl.ApplyNativeConstraintOp() which I'm looking at now and trying to get coverage for.

I think it's a bug that I ported over from MLIR: https://github.com/llvm/llvm-project/pull/162265

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -38,6 +57,335 @@ def detect_roots(self, pattern: pdl.PatternOp) -> list[OpResult[pdl.OperationTyp
...
+            # Has variadic results - check minimum
+            result_count_pred = Predicate.get_result_count_at_least(min_results)
+            predicates.append(
+                PositionalPredicate(
+                    q=result_count_pred.q, a=result_count_pred.a, position=op_pos
+                )
+            )
+
+        # Process attributes
+        for attr_name, attr in zip(op_op.attributeValueNames, op_op.attribute_values):
```
superlopuh: These are expected to have the same length, right?
```suggestion
        for attr_name, attr in zip(op_op.attributeValueNames, op_op.attribute_values, strict=True):
```

superlopuh (approved):

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
jumerckx: yes, I keep forgetting about this, thanks!

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
superlopuh: How would you test this?

PR #5310, opened 2025-10-04
dialects: (math) folders for floating point operations

jumerckx:
I mostly limited myself to simple floating point operations that have a clear python counterpart. Only does folding with constant operands. No other identities are taken into account.

Open questions:
I'm wondering if these folders should work for types other than f64 (like I currently made them to)?
Also, any proposal on how to test it. Just with a single arbitrary input and checking they match the python function?

### DIFF
--- a/xdsl/dialects/math.py
+++ b/xdsl/dialects/math.py
@@ -357,7 +357,7 @@ class AtanOp(FloatingPointLikeUnaryMathOperation, HasFolderInterface):
 
 
 @irdl_op_definition
-class CbrtOp(FloatingPointLikeUnaryMathOperation, HasFolderInterface):
+class CbrtOp(FloatingPointLikeUnaryMathOperation):
     """
     The cbrt operation computes the cube root. It takes one operand of
     floating point type (i.e., scalar, tensor or vector) and returns one result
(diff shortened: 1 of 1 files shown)


jumerckx:
seems like `math.cbrt` and other functions don't exist in Python 3.10.

superlopuh:
@jumerckx sorry for letting this slip. Is this something that you would still like to add? Does MLIR have a similar folding mechanism? It feels to me like some of this should only happen with fastmath. Testing should be regular filecheck for canonicalization IMO.

jumerckx:
> Is this something that you would still like to add?

I don't have an immediate need for this, and you're probably right RE: fastmath.
MLIR has a bunch of (non-obvious to me) folding logic, e.g.
```cpp
  // addf(x, -0) -> x
  if (matchPattern(adaptor.getRhs(), m_NegZeroFloat()))
    return getLhs();
```
and then uses APFloat for generic constant folding.
We could copy these special folders already and postpone full constant folding until xDSL gets an APFloat analogue.
But I don't mind if this is closed for now.

superlopuh:
Yep let's close it for now

PR #5316, opened 2025-10-05
dialects: (tosa) Add `tosa.cond_if`, `tosa.yield` operations

jakedves:
Adds the operations for `tosa.cond_if` and `tosa.yield`. Also adds MLIR and xDSL roundtrip test cases

### DIFF
--- a/xdsl/dialects/tosa.py
+++ b/xdsl/dialects/tosa.py
@@ -2,10 +2,9 @@ from abc import ABC
 from collections.abc import Sequence
 from typing import ClassVar, Generic
 
-from typing_extensions import Self, TypeVar
+from typing_extensions import TypeVar
 
 from xdsl.dialects.builtin import (
-    I1,
     I8,
     I32,
     I64,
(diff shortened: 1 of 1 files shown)


xdsl/dialects/tosa.py
```diff
@@ -410,6 +423,81 @@ def __init__(
...
+        printer.print_ssa_value(self.cond)
+        printer.print_string(" -> (")
+        printer.print_list(
+            self.output.types,
+            printer.print_attribute,
+        )
+        printer.print_string(") ")
+        printer.print_region(self.true_region)
+        printer.print_string(" else ")
+        printer.print_region(self.false_region)
```
superlopuh: This seems simple enough to use an assembly format, could you please give that a go? do you know what would be needed for it to work?

jakedves: I would imagine like this:
```assembly_format = "$cond `->` type($output) $true_region `else` $false_region attr-dict"```

As well as giving the `cond` a specific type: `TensorType(i1, [])` so that we can exclude `type($cond)`.

But it seems to be parsing too much too early:
```
xdsl.utils.exceptions.ParseError: test.mlir:5:38
%1 = tosa.cond_if %0 -> (tensor<3x4xf32>) {
                                          ^
                                          Expected '->'
```

jakedves: Ah think I've got it, just testing now

PR #5318, opened 2025-10-06
transformations: (convert_pdl_to_pdl_interp) _extract_type_predicates

jumerckx:


xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -27,3 +37,35 @@ def detect_roots(self, pattern: pdl.PatternOp) -> list[OpResult[pdl.OperationTyp
...
+                    q=type_constraint.q, a=type_constraint.a, position=type_pos
+                )
+            )
+        elif isinstance(type_op, pdl.TypesOp) and type_op.constantTypes:
+            type_constraint = Predicate.get_type_constraint(type_op.constantTypes)
+            predicates.append(
+                PositionalPredicate(
+                    q=type_constraint.q, a=type_constraint.a, position=type_pos
+                )
+            )
```
superlopuh: Can you please use a match statement here?

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -27,3 +37,35 @@ def detect_roots(self, pattern: pdl.PatternOp) -> list[OpResult[pdl.OperationTyp
             if isinstance(op, pdl.OperationOp) and op.op not in used
         ]
         return roots
+
+    def _extract_type_predicates(
+        self,
+        type_value: Operation | SSAValue,
```
superlopuh: I'm a little suspicious of this, why is it better for it to be passed in as either a value or an op like this rather than just the op, letting the caller resolve the value to the op if needed?

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
jumerckx: done

PR #5319, opened 2025-10-06
misc: (DisjointSet) Add `rooted_union` method to DisjointSet

jumerckx:
Sometimes it's useful to force a particular set to become the representative when `union`ing two sets. E.g. in eqsat when merging an `EClassOp` with a `ConstantEClassOp`, we want the `ConstantEClassOp` to be representative.
(`ConstantEClassOp` is yet to be merged)

### DIFF
--- a/xdsl/utils/disjoint_set.py
+++ b/xdsl/utils/disjoint_set.py
@@ -75,6 +75,30 @@ class IntDisjointSet:
 
         return root
 
+    def rooted_union(self, representative: int, other: int) -> bool:
+        """
+        Merges the set with the given representative and the set containing `other`.
+        The `representative` must be a root of its set (i.e. `ds[representative] == representative`).
+        Returns True if the sets were merged, False if they were already the same set.
+
+        In contrast to `union`, this does not do union by size - the `other` set is always
+        attached to the `representative` set. This is useful when we want to control
+        which element becomes the representative of the merged set.
+        """
+        if self[representative] != representative:
+            raise ValueError("The given representative is not a root of its set.")
+
+        other_root = self[other]
+        if representative == other_root:
+            return False
+
+        rep_count = self._count[representative]
+        other_count = self._count[other_root]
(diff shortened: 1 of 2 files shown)


xdsl/utils/disjoint_set.py
```diff
@@ -75,6 +75,30 @@ def __getitem__(self, value: int) -> int:
 
         return root
 
+    def rooted_union(self, representative: int, other: int) -> bool:
```
superlopuh: `*` to force keyword arguments for clarity; how about `new_root` as a clearer way to communicate that it'll be the new root?
```suggestion
    def rooted_union(self, *, new_root: int, other: int) -> bool:
```

superlopuh: Is `rooted_union` used by another project?

jumerckx: Egg seems to do this by default:
```rs
/// Given two leader ids, unions the two eclasses making root1 the leader.
pub fn union(&mut self, root1: Id, root2: Id) -> Id {
    *self.parent_mut(root2) = root1;
    root1
}

```

jumerckx: To me, `new_root` implies the value isn't a root already, which in this implementation, it needs to be.
If you think it's still clearer I'm fine with it.

Apart from naming, perhaps the arguments should consistently either
* have to be representatives
* need not be representatives
Currently I implemented it as `representative` needing to be representative, and `other` not. wdyt?

superlopuh: Let's not make that a requirement, indeed

superlopuh: Yeah naming I'm not really sure, maybe `union_left` then, and keep the names the same?

superlopuh: as in `lhs, rhs`

jumerckx: donezo

PR #5320, opened 2025-10-06
dialects: (hw) Modernise dialect with properties instead of attributes

gabrielrodcanal:
Attributes are replaced by properties to be compliant with latest MLIR versions (tested with 22.0.0).

### DIFF
--- a/xdsl/dialects/hw.py
+++ b/xdsl/dialects/hw.py
@@ -35,11 +35,11 @@ from xdsl.ir import (
 )
 from xdsl.irdl import (
     IRDLOperation,
-    attr_def,
     irdl_attr_definition,
     irdl_op_definition,
     lazy_traits_def,
-    opt_attr_def,
+    opt_prop_def,
+    prop_def,
     region_def,
     traits_def,
     var_operand_def,
@@ -812,10 +812,10 @@ class HWModuleOp(IRDLOperation):
 
     name = "hw.module"
 
-    sym_name = attr_def(SymbolNameConstraint())
-    module_type = attr_def(ModuleType)
-    sym_visibility = opt_attr_def(StringAttr)
-    parameters = opt_attr_def(ArrayAttr[ParamDeclAttr])
+    sym_name = prop_def(SymbolNameConstraint())
+    module_type = prop_def(ModuleType)
+    sym_visibility = opt_prop_def(StringAttr)
+    parameters = opt_prop_def(ArrayAttr[ParamDeclAttr])
 
     body = region_def("single_block")
 
@@ -836,7 +836,7 @@ class HWModuleOp(IRDLOperation):
         parameters: ArrayAttr[ParamDeclAttr] = ArrayAttr([]),
         visibility: str | StringAttr | None = None,
     ):
-        attributes: dict[str, Attribute] = {
+        properties: dict[str, Attribute] = {
             "sym_name": sym_name,
             "module_type": module_type,
             "parameters": parameters,
@@ -845,16 +845,16 @@ class HWModuleOp(IRDLOperation):
         if visibility:
             if isinstance(visibility, str):
                 visibility = StringAttr(visibility)
-            attributes["sym_visibility"] = visibility
+            properties["sym_visibility"] = visibility
 
         return super().__init__(
-            attributes=attributes,
+            properties=properties,
             regions=[body],
         )
 
     def verify_(self) -> None:
         if self.parameters is not None:
-            # FIXME: once xDSL supports typed attributes, check that parameter
+            # FIXME: once xDSL supports typed properties, check that parameter
             # types are consistent with their default values
             param_names = [param.port_name.data for param in self.parameters.data]
             if len(set(param_names)) != len(param_names):
@@ -895,7 +895,7 @@ class HWModuleOp(IRDLOperation):
         )
 
         if attrs is not None:
-            module_op.attributes.update(attrs.data)
+            module_op.properties.update(attrs.data)
 
         return module_op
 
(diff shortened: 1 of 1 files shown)


gabrielrodcanal:
The method `parse_optional_attr_dict_with_reserved_attr_names` was used to parse attributes provided their names. There is currently not an equivalent method for properties @alexarice @superlopuh

xdsl/dialects/hw.py
```diff
@@ -845,16 +845,16 @@ def __init__(
...
         return super().__init__(
-            attributes=attributes,
+            properties=properties,
             regions=[body],
         )
 
     def verify_(self) -> None:
         if self.parameters is not None:
-            # FIXME: once xDSL supports typed attributes, check that parameter
+            # FIXME: once xDSL supports typed properties, check that parameter
```
superlopuh: how old is this fixme? I'm pretty sure we support typed properties and attributes...

superlopuh:
Two thoughts:
1. We want to stick to being compatible with the same mlir-opt version in all the dialects, these changes seem to actually be to one of the dialects not in upstream MLIR, and in CIRCT, so it's probably OK for now, but we'll need to add some kind of mechanism to check compatibility, probably by moving this to a different repo
2. My understanding is that the attributes vs properties logic is a bit more interesting, in that with the right flag properties are actually printed in the attribute dict. [We handle this correctly in custom assembly format](https://github.com/xdslproject/xdsl/blob/4d1c1bf07a/xdsl/irdl/declarative_assembly_format.py#L480-L501), but not in the exposed printer APIs, could you please add a new function to Printer and Parser with the right logic? I expect it to be a fairly straightforward move of logic from the declarative assembly format

xdsl/dialects/hw.py
gabrielrodcanal: Quite old (28th of March 2024). It appeared for the first time in commit [df1c825](https://github.com/xdslproject/xdsl/blob/df1c825772361f9cdcaa464649829fedad65b5b0/xdsl/dialects/hw.py#L658)

superlopuh: OK from my understanding of the issue it's actually to do with attribute the class not attribute the field of an operation, so it should still refer to attributes after the changes in this PR. Could you please revert this line? This also seems like a good thing to tackle if you're working in this space

gabrielrodcanal:
> Two thoughts:
> 
> 1. We want to stick to being compatible with the same mlir-opt version in all the dialects, these changes seem to actually be to one of the dialects not in upstream MLIR, and in CIRCT, so it's probably OK for now, but we'll need to add some kind of mechanism to check compatibility, probably by moving this to a different repo
> 2. My understanding is that the attributes vs properties logic is a bit more interesting, in that with the right flag properties are actually printed in the attribute dict. [We handle this correctly in custom assembly format](https://github.com/xdslproject/xdsl/blob/4d1c1bf07a/xdsl/irdl/declarative_assembly_format.py#L480-L501), but not in the exposed printer APIs, could you please add a new function to Printer and Parser with the right logic? I expect it to be a fairly straightforward move of logic from the declarative assembly format

gabrielrodcanal:
> Two thoughts:
> 
> 1. We want to stick to being compatible with the same mlir-opt version in all the dialects, these changes seem to actually be to one of the dialects not in upstream MLIR, and in CIRCT, so it's probably OK for now, but we'll need to add some kind of mechanism to check compatibility, probably by moving this to a different repo
> 2. My understanding is that the attributes vs properties logic is a bit more interesting, in that with the right flag properties are actually printed in the attribute dict. [We handle this correctly in custom assembly format](https://github.com/xdslproject/xdsl/blob/4d1c1bf07a/xdsl/irdl/declarative_assembly_format.py#L480-L501), but not in the exposed printer APIs, could you please add a new function to Printer and Parser with the right logic? I expect it to be a fairly straightforward move of logic from the declarative assembly format

Respect to the second thought, it turns out it's not necessary to change anything in the dialect. Just by keeping `parse_optional_attr_dict_with_reserved_attr_names` roundtripping works fine.

However, I've seen that some tests try to add custom attributes, such as here https://github.com/xdslproject/xdsl/blob/946a256487d3339ac884118ef677666b896e49cb/tests/filecheck/dialects/hw/hw_instance.mlir#L26 How is this usually handled in xDSL when properties are expected? Is compatibility for attributes added too? Or maybe we don't really need support for this and the test can be updated accordingly?

superlopuh:
I'm a bit confused by your question, what doesn't work? This is one of those things that the custom syntax must support, all ops can have extra attributes added that are not in the definition. I have a feeling that a bunch of our custom printers/parsers don't handle it properly but we should.

gabrielrodcanal:
I think there might be a little misunderstanding on the support of attributes and properties on my end. As far as I know MLIR supports both attributes and properties at the same time, correct? Does that mean that for a case like https://github.com/xdslproject/xdsl/blob/946a256487d3339ac884118ef677666b896e49cb/tests/filecheck/dialects/hw/hw_instance.mlir#L26
properties and attributes need to be parsed separately? As far as I know it is possible to have custom attributes but not custom properties. Sorry if it's basic, but I've never had to mix attributes and properties before.

superlopuh:
It's a little weird, but since properties were added later and as a kind of optimisation over attributes, there are some places where they behave in a slightly weird way. One case of this is `memref.load`, where because it's annotated with `ParsePropInAttrDict`, the `nontemporal` property is printed in the attr dict in the custom format.

superlopuh:
Hi @gabrielrodcanal, what's the latest here? Should we iterate on this or close it if you're no longer working on it?

gabrielrodcanal:
Hi @superlopuh, I was investigating an idea around this that I decided not to pursue, so we can close it.

PR #5325, opened 2025-10-07
[WIP] dialects: (pdl) make rewrite op body non-optional

superlopuh:
I'll extract the assembly format changes to another PR once we're happy with them

### DIFF
--- a/xdsl/irdl/declarative_assembly_format.py
+++ b/xdsl/irdl/declarative_assembly_format.py
@@ -1053,6 +1053,42 @@ class OptionalRegionVariable(RegionDirective, OptionalVariable):
         self.set(state, None)
 
 
+@dataclass(frozen=True)
+class AnchorRegionVariable(RegionDirective, VariableDirective):
+    """
+    A special case for non-optional anchor regions, which treat the empty region as a special case.
+    """
+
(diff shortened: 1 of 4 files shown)


xdsl/irdl/declarative_assembly_format.py
```diff
@@ -971,6 +977,13 @@ def parse(self, parser: Parser, state: ParsingState) -> bool:
         self.set(state, parser.parse_region())
         return True
 
+    def parse_optional(self, parser: Parser, state: ParsingState) -> bool:
+        region = parser.parse_optional_region()
+        if region is None:
+            region = Region()
+        self.set(state, region)
+        return True
+
```
alexarice, line 987: I feel this will be inherited by `OptionalRegionVariable` and do the wrong thing?

alexarice: Also it always returns true, shouldn't it return False when the region is None?

superlopuh: yep

superlopuh: Fixed

math-fehr (approved):

PR #5330, opened 2025-10-08
core: ignore `__type_params__` field when creating attributes

dk949:
In python 3.12, the `__type_params__` field is used to hold the list of a class' generic arguments. Since xDSL targets 3.10, this is not ignored when inspecting the fields of a class decorated with `@irdl_attr_definition` causing an error.

This feels a little hacky, better suggestions welcome.


dk949:
Changed it to ignore all magic dunder methods/fields

PR #5331, opened 2025-10-08
transformations: (convert-pdl-to-pdl-interp) Add _optimize_tree and _insert_exit

jumerckx:


xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -788,3 +788,40 @@ def _nodes_match(self, node: MatcherNode, predicate: OrderedPredicate) -> bool:
         return (
             node.position == predicate.position and node.question == predicate.question
         )
+
+    def _insert_exit_node(self, root: MatcherNode) -> MatcherNode:
+        """Insert exit node at end of failure paths"""
+        curr = root
+        while curr.failure_node:
```
superlopuh: We try to specify this explicitly in case objects decided to implement `__bool__` in the future and cause unexpected behaviour
```suggestion
        while curr.failure_node is not None:
```

xdsl/transforms/convert_pdl_to_pdl_interp/conversion.py
```diff
@@ -789,3 +789,40 @@ def _nodes_match(self, node: MatcherNode, predicate: OrderedPredicate) -> bool:
...
+            # Convert switch to bool node
+            answer, child = next(iter(root.children.items()))
+            bool_node = BoolNode(
+                position=root.position,
+                question=root.question,
+                success_node=child,
+                failure_node=root.failure_node,
+                answer=answer,
+            )
+            return bool_node
```
superlopuh, line 826: Weird! I would have expected this kind of optimisation to be done in the IR but I guess we can do that later, and some redundancy is OK

PR #5332, opened 2025-10-08
[WIP] installation: enable python 3.14

alexarice:
Attempt at getting python 3.14 working.


pyproject.toml
```diff
@@ -38,6 +38,7 @@ dev = [
 ]
 gui = ["textual>=6,<7", "pyclip==0.7"]
 jax = ["jax==0.6.0", "numpy==2.2.6"]
+riscv = ["riscemu==2.2.7"]
```
alexarice: I'm not sure this is correct but it feels like this shouldn't have been in the `docs` group?

xdsl/frontend/listlang/main.py
```diff
@@ -794,7 +794,6 @@ def program_to_mlir_string(code: str):
     arg_parser.add_argument(
         "filename",
         nargs="?",
-        type=argparse.FileType("r"),
```
alexarice, line 797: This has been deprecated in python 3.14, by the docs it sounds like this should just be removed and that you should check you actually got a readable file later?

xdsl/irdl/attributes.py
```diff
@@ -405,7 +399,7 @@ def constr(self) -> AttrConstraint[AttributeCovT]:
     AttrConstraint[AttributeInvT]
     | AttributeInvT
     | type[AttributeInvT]
-    | "TypeForm[AttributeInvT]"
+    | TypeForm[AttributeInvT]
```
alexarice: This seems unfortunate, but python 3.14 does not like trying to union a string and a type. I'm not entirely sure what the ramifications of making this change are.

pyproject.toml
```diff
@@ -37,7 +37,7 @@ dev = [
     "sympy==1.14.0",
 ]
 gui = ["textual>=6,<7", "pyclip==0.7"]
-jax = ["jax==0.6.0", "numpy==2.2.6"]
```
alexarice: @superlopuh is xdsl now jax free? Jax doesn't work with python 3.14 yet

alexarice:
This all works locally (and on the nix ci) and I'm not really sure what would be causing the errors in the MLIR CI. Anyone have any ideas?

pyproject.toml
superlopuh: Oh I forgot to officially remove this

xdsl/irdl/attributes.py
superlopuh: We just need to `from __future__ import annotations` here I think , 3.14 is the future in question IIUC

pyproject.toml
superlopuh: #5334

xdsl/frontend/listlang/main.py
superlopuh: happy to just drop this line from the tutorial code

pyproject.toml
superlopuh: No it is, we no longer have riscv as a supported optional dependency, I moved it to docs since we still use it in the notebooks

xdsl/irdl/attributes.py
alexarice: This didn't seem to make any difference unfortunately

alexarice:
I really have no idea what is happening here. The mlir ci seems to fail to build marimo because of a dependency, yet the python ci seems to build this dependency with no issues.

alexarice:
Ok, it's something to do with the container, I guess it makes sense to push forward with #5238 for now and revisit this afterwards

xdsl/irdl/attributes.py
superlopuh: Not sure what you mean, what is the error if you add this import?

alexarice: Same error, can't apply `|` to `Typing.union` and `str`

diegorusso: There ~was~ is a regression in 3.14: https://github.com/xdslproject/xdsl/issues/5396

diegorusso:
I've just released pyperformance 1.13.0 where the `xdsl` benchmarks has been re-enabled

https://pypi.org/project/pyperformance/1.13.0/

PR #5335, opened 2025-10-09
Syntax highlighting adjustments

positr0nium:
slightly adjusted the coloring scheme. Types are now printed in green, operation names in cyan and operations that induce control flow in magenta. Wrote a pprint wrapper function (pprint) to reduce boilerplate when calling the syntax printer.

### DIFF
--- a/xdsl/ir/core.py
+++ b/xdsl/ir/core.py
@@ -2109,14 +2109,6 @@ class Block(_IRNode, IRWithUses, IRWithName):
 
         return True
 
-    def __str__(self) -> str:
-        from xdsl.printer import Printer
-
-        res = StringIO()
-        printer = Printer(stream=res)
-        printer.print_block(self)
-        return res.getvalue()
-
 
 @dataclass
 class _RegionBlocksIterator(Iterator[Block]):


positr0nium:
image

xdsl/syntax_printer.py
```diff
@@ -9,10 +9,28 @@ class SyntaxPrinter(Printer, ColorPrinter):
...
-        with self.colored(Colors.RED if op in self.diagnostic.op_messages else None):
-            return super().print_op(op)
+    def print_attribute(self, attribute: Attribute):
+        if isinstance(attribute, TypeAttribute):
+            with self.colored(Colors.GREEN):
+                return super().print_attribute(attribute)
+        return super().print_attribute(attribute)
+
+
+def pprint(obj: Operation | Region | Block):
```
superlopuh: Let's just use the SyntaxPrinter in `__str__` in these classes?

xdsl/syntax_printer.py
```diff
@@ -9,10 +9,28 @@ class SyntaxPrinter(Printer, ColorPrinter):
...
 
-    def print_ssa_value(self, value: SSAValue) -> str:
-        with self.colored(Colors.BRIGHT_MAGENTA):
-            return super().print_ssa_value(value)
+    def print_op_name(self, name: str):
+        dialect = name.split(".")[0]
+        if dialect in ["scf", "cf", "func"]:
+            color = Colors.MAGENTA
+        else:
+            color = Colors.CYAN
```
superlopuh: let's avoid per-dialect logic for now

positr0nium: I feel like having control flow blocks highlighted in a distinguished color helps understanding what is going on in the code more easily. It's also common for many high-level languages.

xdsl/syntax_printer.py
positr0nium: That could also be an option but it would make serializing a bit more cumbersome because there needs to be a dedicated call without the "coloring instructions" within the string. I'd say visualition happens more often than serialization so, it could be worth it to have ``__str__`` doing the highlighting.

xdsl/syntax_printer.py
superlopuh: I would propose having that discussion separately once we've merged syntax highlighting at all

xdsl/syntax_printer.py
superlopuh: We already have a dedicated call that we expect clients to use most of the time which is `Printer().print_op(op)`, `print(op)` is already kind of sugar IMO, so having a pretty print there makes sense I think.

positr0nium:
Ok, so there seem to be some Jax related errors related to to modifying the ``__str__`` method. I thought more about this serialization thing and it might be more convenient to implement a ``Operation.serialize`` method. The ``Printer``-based method only prints to the console, which can in principle be captured into a string but this introduces like 5-10 lines of boilerplate, which is relatively cumbersome to remember compared to a single method call.

superlopuh:
I was confused at first when seeing the error since we actually removed that part of the code. I just merged main into Alex's branch, and that branch into yours, hopefully the errors will be more specific to this.

superlopuh:
Looking at the test failures, I actually think that the tests are what should be fixed. I would rather have pretty printing by default, and the testing be more verbose by calling into the Printer directly instead of Operation's `__str__` method.

positr0nium:
As discussed on Zulip, the syntax highlights are deactivated again for ``__str__`` . The ``--syntax-highlight`` feature for xdsl-opt was already established in a previous commit.

xdsl/ir/core.py
```diff
@@ -2109,6 +2109,14 @@ def is_structurally_equivalent(
...
         return True
 
+    def __str__(self) -> str:
+        from xdsl.printer import Printer
+
+        res = StringIO()
+        printer = Printer(stream=res)
+        printer.print_block(self)
+        return res.getvalue()
+
```
superlopuh: let's revert this for now, we can try it again in a future PR

xdsl/ir/core.py
```diff
@@ -2728,5 +2728,13 @@ def is_structurally_equivalent(
             return False
         return True
 
+    def __str__(self) -> str:
```
alexarice, line 2731: This feels like a separate change?

xdsl/syntax_printer.py
```diff
@@ -9,10 +9,12 @@ class SyntaxPrinter(Printer, ColorPrinter):
     A printer for printing syntax-highlighted mlir code to a terminal.
     """
```
alexarice, line 11: Any reason for removing the colouring of SSA values? Are you following some particular colour scheme? For reference I was roughly following https://github.com/llvm/llvm-project/blob/8b3a124ad87d1e808852644090ea5d1117fe2f9f/mlir/utils/emacs/mlir-mode.el#L36

xdsl/syntax_printer.py
```diff
@@ -9,10 +9,12 @@ class SyntaxPrinter(Printer, ColorPrinter):
...
 
-    def print_ssa_value(self, value: SSAValue) -> str:
-        with self.colored(Colors.BRIGHT_MAGENTA):
-            return super().print_ssa_value(value)
+    def print_op_name(self, name: str):
+        with self.colored(Colors.CYAN):
+            return super().print_op_name(name)
 
-    def print_op(self, op: Operation) -> None:
-        with self.colored(Colors.RED if op in self.diagnostic.op_messages else None):
```
alexarice: Is the decision to give up on the diagnostics?

PR #5339, opened 2025-10-09
dialects: (eqsat) ConstantEClassOp

jumerckx:


xdsl/dialects/eqsat.py
```diff
@@ -88,6 +128,9 @@ def verify_(self) -> None:
                     )
 
 
+AnyEClassOp = EClassOp | ConstantEClassOp
```
superlopuh, line 140: can you please add a doc string?

xdsl/dialects/eqsat.py
```diff
@@ -36,6 +44,38 @@
 """
 
 
+@irdl_op_definition
+class ConstantEClassOp(IRDLOperation, ConstantLikeInterface):
```
superlopuh, line 48: Can you add a doc string here also, please? Probably also in the `EClassOp`, and mention each other there

jumerckx: I've added docstrings with links but that means the in-editor docs on hover look stupid.
image
Which way do you prefer, or is there a fix for this?

superlopuh: They don't look that stupid to me, just like regular markdown, which is pretty readable. I can live with this.

xdsl/dialects/eqsat.py
```diff
@@ -36,8 +44,49 @@
...
+    result = result_def(T)
+    value = prop_def()
+    min_cost_index = opt_attr_def(IntAttr)
+
+    def get_constant_value(self):
+        return self.value
+
+    def __init__(self, const_arg: OpResult):
+        if (trait := const_arg.owner.get_trait(ConstantLike)) is None:
+            raise DiagnosticException(
```
superlopuh, line 72: I just realized that this isn't tested, should we add a test for this?

xdsl/dialects/eqsat.py
jumerckx: ok

xdsl/dialects/eqsat.py
jumerckx: Done

PR #5340, opened 2025-10-09
transformations: (apply-eqsat-pdl-interp) Support folding

jumerckx:
~~TODO: test~~

### DIFF
--- a/tests/interpreters/test_eqsat_pdl_interp_interpreter.py
+++ b/tests/interpreters/test_eqsat_pdl_interp_interpreter.py
@@ -1164,6 +1164,65 @@ def test_eclass_union_different_constants_fails():
         interp_functions.eclass_union(const_eclass1, const_eclass2)
 
 
+def test_eclass_union_constant_with_regular():
+    """Test that eclass_union of ConstantEClassOp with regular EClassOp results in ConstantEClassOp containing both operands."""
+    ctx = Context()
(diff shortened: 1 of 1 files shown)


xdsl/interpreters/eqsat_pdl_interp.py
```diff
@@ -315,31 +332,62 @@ def run_create_operation(
...
             )
-            assert isinstance(arg.owner, eqsat.EClassOp), (
-                "pdl_interp.create_operation currently only supports creating operations with operands that are EClassOp results."
+            assert isinstance(arg.owner, eqsat.AnyEClassOp), (
+                "pdl_interp.create_operation currently only supports creating operations with operands that are eclass results."
             )
             updated_operands.append(self.eclass_union_find.find(arg.owner).result)
         args = (*updated_operands, *args[len(op.input_operands) :])
         (new_op,) = super().run_create_operation(interpreter, op, args).values
+        if cast(Operation, new_op).get_attr_or_prop("unsound") is None and (
```
superlopuh, line 341: this feels like a Herbie thing, right? What does "unsound" mean in this context? Should we generalize it somehow?

jumerckx: As discussed, this should become redundant with proper analysis-guided rule application. It might make sense to keep this in temporarily as a stopgap until we get there?

superlopuh (approved):

PR #5342, opened 2025-10-10
core: handle modulo and division in AffineExpr.simplify

superlopuh:
Ports the simplification logic for affine expressions from MLIR. Does not implement semi-affine map simplification. I'll split this up into digestible PRs.

### DIFF
--- a/xdsl/ir/affine/affine_expr.py
+++ b/xdsl/ir/affine/affine_expr.py
@@ -628,7 +628,7 @@ class SimpleAffineExprFlattener:
 
         A floordiv is thus flattened by introducing a new local variable q, and
         replacing that expression with 'q' while adding the constraints
-        `c * q <= expr <= c * q + c - 1` to local_var_cst (done by
+        `c * q <= expr <= c * q + c - 1` to `local_var_cst` (done by
         `add_local_floor_div_id`).
 
         A ceildiv is similarly flattened:


xdsl/ir/affine/affine_expr.py
```diff
@@ -620,10 +621,143 @@ def visit_constant_expr(self, expr: AffineConstantExpr) -> None:
...
     def visit_div_expr(self, expr: AffineBinaryOpExpr, *, is_ceil: bool) -> None:
-        raise NotImplementedError("Simplifying div expr not yet implemented")
+        """
+        Handles floor and ceil division for affine expressions.
+
+        `t = expr floordiv c   <=> t = q, c * q <= expr <= c * q + c - 1`
+
+        A floordiv is thus flattened by introducing a new local variable q, and
+        replacing that expression with 'q' while adding the constraints
+        `c * q <= expr <= c * q + c - 1` to local_var_cst (done by
```
alexarice:
```suggestion
        `c * q <= expr <= c * q + c - 1` to `local_var_cst` (done by
```

xdsl/ir/affine/affine_expr.py
```diff
@@ -620,10 +621,143 @@ def visit_constant_expr(self, expr: AffineConstantExpr) -> None:
...
-        raise NotImplementedError("Simplifying div expr not yet implemented")
+        """
+        Handles floor and ceil division for affine expressions.
+
+        `t = expr floordiv c   <=> t = q, c * q <= expr <= c * q + c - 1`
+
+        A floordiv is thus flattened by introducing a new local variable q, and
+        replacing that expression with 'q' while adding the constraints
+        `c * q <= expr <= c * q + c - 1` to `local_var_cst` (done by
+        `add_local_floor_div_id`).
```
alexarice: Why do we need a new variable `q`? Can you not just use `t` in its place?

qaco (approved):

PR #5343, opened 2025-10-10
core: Check irdl_options only contains IRDLOptions 

dk949:
The `RecursiveMemoryEffect` trait was previously specified in the `irdl_options` options list, not in `traits`.

### DIFF
--- a/tests/irdl/test_operation_definition.py
+++ b/tests/irdl/test_operation_definition.py
@@ -156,7 +156,10 @@ class InvalidIRDLOpts(IRDLOperation):
 
 def test_invalid_irdl_options():
     """Check that irdl_options only contains IRDLOptions"""
-    with pytest.raises(PyRDLOpDefinitionError):
+    with pytest.raises(
+        PyRDLOpDefinitionError,
+        match="All values in irdl_options should inherit IRDLOption",
+    ):
         irdl_op_definition(InvalidIRDLOpts)
 
 


superlopuh:
Can you please fix the PR title to follow xDSL conventions? Can you please add a test it feels like the easier thing here is to add a test to the IRDL infrastructure to raise an error if there's a thing in the irdl_options that's not an IRDL option.

dk949:
> Can you please fix the PR title to follow xDSL conventions? Can you please add a test it feels like the easier thing here is to add a test to the IRDL infrastructure to raise an error if there's a thing in the irdl_options that's not an IRDL option.

I had a quick look, it seems that the options are just cast to a `list[IRDLOption]` [here](https://github.com/xdslproject/xdsl/blob/main/xdsl/irdl/operations.py#L991). It should probably instead actually verify that this is what the type is, I'll add the check.

superlopuh:
That would be fantastic, thank you!

tests/irdl/test_operation_definition.py
```diff
@@ -149,6 +149,17 @@ def test_invalid_field():
...
 
 
+class InvalidIRDLOpts(IRDLOperation):
+    name = "test.invalid_field"
+    irdl_options = [42]
+
+
+def test_invalid_irdl_options():
+    """Check that irdl_options only contains IRDLOptions"""
+    with pytest.raises(PyRDLOpDefinitionError):
```
superlopuh: Can you please add a match statement here?

dk949:
Oops, I was sure I'd merged this 2 weeks ago

PR #5351, opened 2025-10-13
core: Add dataflow analysis core

jumerckx:


jumerckx:
Closing since everything in this pr has already made its way into main

PR #5352, opened 2025-10-13
core: (Analysis) add `ChangeResult`

jumerckx:


xdsl/analysis/dataflow.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import annotations
```
superlopuh, line 5: Can you please add a doc string here describing what kind of things will go into the file?

PR #5356, opened 2025-10-13
core: (analysis) add `AnalysisState`

jumerckx:
`DataFlowSolver ` and `DataFlowAnalysis ` are declared here because I wanted to use them in the test. Follow up prs will complete the implementation

### DIFF
--- a/xdsl/analysis/dataflow.py
+++ b/xdsl/analysis/dataflow.py
@@ -112,10 +112,10 @@ class AnalysisState(ABC, Generic[AnchorInvT]):
     and evolve as the analysis iterates.
     """
 
-    anchor: LatticeAnchor
+    anchor: AnchorInvT
     dependents: set[tuple[ProgramPoint, DataFlowAnalysis]]
 
-    def __init__(self, anchor: LatticeAnchor):
+    def __init__(self, anchor: AnchorInvT):
         self.anchor = anchor
         self.dependents = set()
 


xdsl/analysis/dataflow.py
```diff
@@ -90,3 +95,63 @@ def __str__(self) -> str:
...
+AnalysisT = TypeVar("AnalysisT", bound="DataFlowAnalysis")
+
+
+class AnalysisState(ABC):
+    """
+    Base class for all analysis states. States are attached to lattice anchors
+    and evolve as the analysis iterates.
+    """
+
+    anchor: LatticeAnchor
```
superlopuh: Would it make sense to make AnalysisState generic and specify this? On your WIP branch some subclasses use SSAValue specifically as the anchor, seems like it might be worth keeping this around in the type system

xdsl/analysis/dataflow.py
```diff
@@ -90,3 +95,63 @@ def __str__(self) -> str:
         if isinstance(self.entity, Operation):
             return f"before op '{self.entity.name}'"
         return f"at end of block '{self.entity.name_hint or id(self.entity)}'"
+
+
+LatticeAnchor: TypeAlias = SSAValue | Block | ProgramPoint | GenericLatticeAnchor
```
superlopuh, line 100: worth a doc comment

xdsl/analysis/dataflow.py
```diff
@@ -90,3 +95,63 @@ def __str__(self) -> str:
...
+StateType = TypeVar("StateType", bound="AnalysisState")
+AnalysisT = TypeVar("AnalysisT", bound="DataFlowAnalysis")
```
superlopuh:
```suggestion
AnalysisStateInvT = TypeVar("StateType", bound="AnalysisState")
DataFlowAnalysisInvT = TypeVar("AnalysisT", bound="DataFlowAnalysis")
```
For consistency with the rest of the project

xdsl/analysis/dataflow.py
jumerckx: I agree. I'm not too familiar with Python's type system though. Is there a way for `AnalysisStateInvT` to capture the particular typevar as well?
i.e.:
```py
AnchorInvT = TypeVar("AnchorInvT", bound=LatticeAnchor)

AnalysisStateInvT = TypeVar("AnalysisStateInvT", bound="AnalysisState[AnchorInvT]")
DataFlowAnalysisInvT = TypeVar("DataFlowAnalysisInvT", bound="DataFlowAnalysis")

class AnalysisState(ABC, Generic[AnchorInvT]):
	...
```
doesn't work.

superlopuh: Not sure, but I also don't see why you would want to do this

jumerckx: I might be misunderstanding, but when I make `AnalysisState` Generic, I get errors saying that type arguments must be provided.
This works, but loses the fact that the anchortype is the same in argument and return value.
```py
    def get_or_create_state(
        self, anchor: LatticeAnchor, state_type: type[AnalysisState[Any]]
    ) -> AnalysisState[Any]:
```

jumerckx:
```py
AnchorInvT = TypeVar("AnchorInvT", bound=LatticeAnchor, default=LatticeAnchor)

AnalysisStateInvT = TypeVar("AnalysisStateInvT", bound="AnalysisState")

def get_or_create_state(
        self, anchor: LatticeAnchor, state_type: type[AnalysisStateInvT]
    ) -> AnalysisStateInvT:
```
I think this does what it needs to: the type of the result matches `state_type`?

superlopuh:
```suggestion
    anchor: AnchorInvT
```

PR #5357, opened 2025-10-13
core: (analysis) add `DataFlowSolver`

jumerckx:


tests/analysis/test_dataflow.py
```diff
@@ -180,3 +180,85 @@ def test_analysis_state_on_update():
...
+    def visit(self, point: ProgramPoint) -> None:
+        pass
+
+
+def test_data_flow_solver_init(solver: DataFlowSolver):
+    assert isinstance(solver.context, Context)
+    assert solver._analyses == []  # pyright: ignore[reportPrivateUsage]
+    assert not solver._worklist  # pyright: ignore[reportPrivateUsage]
+    assert not solver._analysis_states  # pyright: ignore[reportPrivateUsage]
+    assert not solver._is_running  # pyright: ignore[reportPrivateUsage]
```
superlopuh: I don't think it's a good idea in general to test private APIs like this. The idea is that the tests should be for the observable behaviour, and still give us flexibility to change the implementation when needed. Can you please change these tests to check for this specifically?

PR #5358, opened 2025-10-13
core: (analysis) add `DataFlowAnalysis`

jumerckx:


xdsl/analysis/dataflow.py
```diff
@@ -240,3 +240,28 @@ def visit(self, point: ProgramPoint) -> None:
         solver when a dependency of this analysis at this point has changed.
         """
         ...
+
+    def get_or_create_state(
+        self, anchor: LatticeAnchor, state_type: type[AnalysisStateInvT]
```
superlopuh, line 245: Does something like this work? Otherwise the return type is unbound, it doesn't really mean anything
```suggestion
        self, anchor: AnchorInvT, state_type: type[AnalysisState[AnchorInvT]]
```

jumerckx: Yes!

jumerckx: Actually, no!

It's more important that the return type has the same AnalysisState as the `state_type` provided. Is there a way to encode this? i.e. `get_or_create_state(self, anchor:A, state_type:type[S]) where S:AnalysisState[A] -> S`

jumerckx: https://returns.readthedocs.io/en/latest/pages/hkt.html

jumerckx: Assuming we can't have both, I'm in favour of `state_type: type[AnalysisStateInvT] -> AnalysisStateInvT`. At that point I think it doesn't make much sense to keep `AnalysisState` generic, or does it?

superlopuh: I think it still does, I'm in favour of your last suggestion

jumerckx: Ok, this is how it's now set up

PR #5360, opened 2025-10-13
core: (analysis) add `PredecessorState` for dead code analysis

jumerckx:


xdsl/analysis/dead_code_analysis.py
```diff
@@ -1,14 +1,33 @@
...
     DataFlowAnalysis,
     DataFlowSolver,
+    GenericLatticeAnchor,
     LatticeAnchor,
     ProgramPoint,
 )
-from xdsl.ir import Operation
+from xdsl.ir import Block, Operation, SSAValue
+
+_ValueRange = Sequence[SSAValue]
```
superlopuh: let's just inline this

PR #5363, opened 2025-10-14
dialects: (x86) add fma ops with memory operand

emmau678:
Add Vfmadd ops to x86 dialect with memory as the second source operand

### DIFF
--- a/xdsl/dialects/x86/ops.py
+++ b/xdsl/dialects/x86/ops.py
@@ -42,11 +42,14 @@ from xdsl.backend.register_allocatable import (
 )
 from xdsl.backend.register_type import RegisterAllocatedMemoryEffect, RegisterType
 from xdsl.dialects.builtin import (
+    I32,
     IntegerAttr,
     IntegerType,
     ModuleOp,
     Signedness,
     StringAttr,
+    i32,
+    i64,
 )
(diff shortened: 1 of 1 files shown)


xdsl/dialects/x86/ops.py
```diff
@@ -1056,6 +1057,68 @@ def get_register_constraints(self) -> RegisterConstraints:
...
+    """
+    A base class for x86 operations that have one register that is read and written to,
+    one source register and one memory source operand.
+    """
+
+    register_in = operand_def(R1InvT)
+    register_out = result_def(R1InvT)
+    source1 = operand_def(R2InvT)
+    memory = operand_def(R4InvT)
+    memory_offset = attr_def(IntegerAttr, default_value=IntegerAttr(0, 64))
```
superlopuh: From some spelunking, it looks like the maximum bitwidth for this instruction is 32, can you please change it to this to make sure that we respect this spec?
```suggestion
    memory_offset = attr_def(IntegerAttr[I32], default_value=IntegerAttr(0, 32))
```
The verification will happen automatically.

Can you also change the init below to `IntegerAttr[I32]`?

qaco:
Nice !

